PR #238, opened 2022-03-17
Update data-fetching.md

colinthedev:
Just a minor suggestion regarding clarity.


sarah11918:
Thanks, @colinthedev ! This was a really helpful suggestion!

I updated the link to a relative one, to make it match the rest of our link formats, and linked specifically to the relevant part of the page. I also just used the more generic wording, since both of those possibilities are available at that page so we don't need to list them both here.

Much more accurate, and cleaner here now! Thanks again!

PR #245, opened 2022-03-20
Fix broken links

gaeulbyul:
I noticed that some of the links on the webpage are broken or pointing to the English page instead of the already translated one. This commit will fix it.


sarah11918:
Thank you, @gaeulbyul , these look good! Thank you for such attention to detail.  :)

PR #246, opened 2022-03-20
add trailing slash for sidebar links

manigandham:
The production build/hosting use a trailing slash but the nav links don't. This cases an extra 301 redirect request for every link.

This fixes the problem by adding the slash to the end of the sidebar links. I checked all the links and nothing is broken with this change.

image

image


FredKSchott:
Nice! Great catch!

PR #248, opened 2022-03-21
Beautiful install instructions toggle

sarah11918:
A "proof-of-concept" (is it?) instruction toggle for Quick/Manual set up all on the main Installation page. (Includes necessary corresponding changes: removing manual setup page entirely, updating config.ts sidebar entry.)

Someone who can style can fix this, once they've finished laughing.


FredKSchott:
Did some work to improve it! I'm not sure yet if I *love* this new approach, but it felt like an interesting way to solve the problem of having some common UI shared across multiple pages.

FredKSchott:
I have decided I like this! It would go well with the `spa`/`turbo` mode that Nate has played with to enable SPA-style navigation for Astro sites. This would mean that clicking the toggle wouldn't cause a full page reload.

FredKSchott:
@sarah11918 if you're 👍 then feel free to merge!

sarah11918:
@FredKSchott I do like it! FYI: this would push to `next`, not main as is, but I think it's fine to go to either. Do you have a preference?

FredKSchott:
v0.25.0 will be cut imminently, so `next` should get merged to main today as well! I'll merge into `next` to reduce conflicts

PR #249, opened 2022-03-22
Fix white on light grey text in DocSearch modal

delucis:
Just noticed that the hovered item in the search modal with the light theme currently uses an illegible text colour. This PR fixes that.

| before | after |
|--------|-------|
| Screen Shot 2022-03-22 at 10 09 11 | Screen Shot 2022-03-22 at 10 09 26 |


sarah11918:
Looks good! I'm going to merge this now so that text is legible (!), but would think we'd also be happy to consider other styling (dark purple text?) for the highlighted option at some point.

PR #250, opened 2022-03-22
Bring back dynamic values & props in components

sarah11918:
I know there's a larger rewrite/page to come, but I'm just submitting to have on our radar some content I'd originally suggested for core-concepts/astro-components that was decided out of scope for core-concepts.

This might or might not be where we eventually decide this info should live, but we've seen recently 2 or 3 people with questions that, had they seen this, they might not have needed to ask.  :) So I wanted to bring it back so it goes somewhere!

I think the thinking in removing it was that this is bording on "teaching people how to JSX" and so more suited to a "How-to Guide" placement. Since I've had to type it out a couple of times in Discord recently (and once on Twitter!  :S ) maybe it needs to come back in some form sooner rather than later? Maybe even back in astro-components.md until it has a better home?


sarah11918:
OK, I'm now going to push for this as at least a temporary solution. 

We wanted to get slots out of the sidebar itself and into an Astro Component page of some kind, and until Astro Component Syntax is written this is the one we have. I also think you might want that dynamic content re-introduced in one of those two pages. 

So, my suggestion is to just throw it in here, where people will think to look for it and where it will be helpful. I know @FredKSchott  you intend to work on writing Astro Component Syntax and Markdown Syntax for reference this week. Having these here will at least be reminders of the content that needs to pick a file to live in.  

(Accordingly, I didn't bother "paring down" the slot content for now. We probably will want all of that somewhere, so now it can just be plucked and placed somewhere else, if/when the time comes.)

src/pages/en/core-concepts/astro-components.md
```diff
@@ -101,6 +101,206 @@ const myFavoritePokemon = [/* ... */];
...
+
+
+### Component Props
+
+An Astro component can define and accept props. These props then become available to the component template for rendering HTML. Props are available on the `Astro.props` global in your frontmatter script.
+
+Here is an example of a component that receives a `greeting` prop and a `name` prop. Notice that the props to be received are destructured from the global `Astro.props` object.
+
+```astro
+// src/components/GreetingHeadline.astro
```
FredKSchott, line 178: this comments should always be inside the `---` code fence, so that the code is valid. Otherwise syntax highlighting can get very odd.

FredKSchott: also, can you show an example of how this component is used? Maybe something like:


```astro
---
// Example: GreetingHeadline.astro
// Usage: <GreetingHeadline greeting="Howdy" name="Partner" />
const { greeting, name } = Astro.props
---
<h2>{greeting}, {name}!</h2>
```

natemoo-re: Correct `// comments` can only go inside of the `---` fence.

src/pages/en/core-concepts/astro-components.md
```diff
@@ -101,6 +101,206 @@ const myFavoritePokemon = [/* ... */];
...
+  name: string;
+  greeting?: string;
+}
+
+const { greeting = "Hello", name } = Astro.props
+---
+<h2>{greeting}, {name}!</h2>
+```
+
+This component, when imported and rendered in other Astro components, layouts or pages, can be passed these props as attributes:
```
FredKSchott, line 205: ah yea, I commented on this above. This feels too late in the section to bring up usage. I think just showing usage in the snippet suggestion I made is enough, but if you want to keep this section you should move it up to the top of the section, maybe even the first thing you talk about as a way to define what you mean when you say "props"

FredKSchott:
I had to rebase and re-push to fix the branch after the merge to `next`. Left some comments, but otherwise LGTM!

sarah11918:
Tightened up the code examples. Good suggestions. Shipping!

PR #251, opened 2022-03-22
Move built-in components into Runtime API

sarah11918:
Per instructions in config.ts sidebar: some light editing/updating of former built-in components page, moved into Runtime API


FredKSchott:
LGTM!

PR #252, opened 2022-03-23
Customizing Astro - page skeleton

sarah11918:
I wasn't sure the exact level of detail we're going for on this page (since the details are in configuration reference), so it's mostly a page skeleton (just to get the ball rolling) of the *kinds* of customizations. Gives us a placeholder and a starting point.


FredKSchott:
I just pushed a commit with what I'd imagined this page to focus on, would love your edits! 

Also, looks like this branch got messed up a bit! Not sure why or if it was me, but I'm happy to repush from my local branch if you need me to to fix

FredKSchott:
I got scared of the merge conflict, and was happy enough with my edit that I ended up merging directly into main. Would still love your feedback on it, if you are able to create another PR to `main` with any edits that you'd recommend: https://docs.astro.build/en/guides/configuring-astro/

PR #253, opened 2022-03-24
Updated content to inc YAML content

aFuzzyBear:
Applied a small part explaining about YAML in the frontmatter, and how it can be used to organise the page's meta data

### DIFF
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -78,6 +78,7 @@ An example blog post `content` object might look like:
   "author": "Matthew Phillips",
   "description": "Astro 0.18 is our biggest release since Astro launch.",
   "draft": false,
+  "keywords": ["astro", "release", "announcement"]
   **/
   "astro": {
     "headers": [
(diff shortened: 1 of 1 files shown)


sarah11918:
I'd like to consider something more along the lines of this. (See my changes)

I like the points being made explicit that a) you can use these props in ANY astro component, not just layouts and b) the (meta)data can be richer, and be used in more elaboarte interesting ways.

I don't think it's the mandate of this document to explain/describe YAML itself, nor do I think a second content object provides much more value than the one already there. BUT, I did adjust the earlier example to include a keyword array, to make it more obvious that these properties can be of varied data types, in keeping with the point you raise.

I considered making this into example of usage in a component (not of showing the resulting object), but a) we do have an example of using frontmatter in a layout component, and I think it would just need to change soon anyway from this week's RFC where the content prop object will slightly change? 

I think just the explicit note/description is a solid addition to the page, and I think your example will be perfect to keep on the stack for a larger tutorial on "Builing an Astro blog".

Thoughts?

aFuzzyBear:
I think that these changes are more than sufficient, I was just thinking that for those who dont know, looking over the page would of been difficult to understand if it could be possible. I like the way that this has been presented, thank you buddy 💚.

I would say that this is safe as is to merge, however more eyes if possible,

sarah11918:
Knowing that the Markdown content prop is scheduled to change soon with the new RFC anyway, I'm gonna call this safe to merge. This page will have to be revisited soon to update for that, and if other eyes don't like the change, we'll deal with it then!  :)

PR #255, opened 2022-03-24
Server-side rendering docs

matthewp:
This documents how to setup and use SSR features in Astro.

### DIFF
--- a/src/pages/en/reference/adapter-reference.md
+++ b/src/pages/en/reference/adapter-reference.md
@@ -1,0 +1,159 @@
+---
+layout: ~/layouts/MainLayout.astro
+title: Astro Adapter API (experimental)
+---
+
+Astro is designed to make it easy to deploy to any cloud provider for SSR (server-side rendering). This ability is provided by __adapters__, which are [integrations](/en/reference/integrations-reference/).
+
+> Server-side rendering in Astro is *experimental*. If you are interested in building an adapter for a host now is the perfect time to help shape these APIs. If you are worried about breaking changes this might be a little too soon for you.
+
+## What is an adapter
+
+An adapter is a special kind of [integration](/en/reference/integrations-reference/) that provides an entrypoint for server-side rendering. An adapter does two things:
+
+- Implements host-specific APIs for handling requests.
+- Configures the build according to host conventions.
+
+## Building an Adapter
+
+An adapter is an [integration](/en/reference/integrations-reference/) and can do anything that an integration can do.
+
+An adapter __must__ call the `setAdapter` API in the `astro:config:done` hook like so:
+
+```js
+export default function createIntegration() {
+  return {
+    name: '@matthewp/my-adapter',
+    hooks: {
+      'astro:config:done': ({ setAdapter }) => {
+        setAdapter({
+          name: '@matthewp/my-adapter',
+          serverEntrypoint: '@matthewp/my-adapter/server.js'
+        });
+      },
+    },
+  };
+}
+```
+
+The object passed into `setAdapter` is defined as:
+
+```ts
+interface AstroAdapter {
+	name: string;
+	serverEntrypoint?: string;
+	exports?: string[];
+}
+```
+
+The properties are:
+
+* __name__: A unique name for your adapter, used for logging.
+* __serverEntrypoint__: The entrypoint for server-side rendering.
+* __exports__: An array of named exports when used in conjunction with `createExports` (explained below).
+
+### Server Entrypoint
+
+Astro's adapter API attempts to work with any type of host, and gives a flexible way to conform to the host APIs.
+
+#### Exports
+
+Some serverless hosts expect you to export a function, such as `handler`:
+
+```js
+export function handler(event, context) {
+  // ...
+}
+```
+
+With the adapter API you achieve this by implementing `createExports` in your `serverEntrypoint`:
+
+```js
+import { App } from 'astro/app';
+
+export function createExports(manifest) {
+  const app = new App(manifest);
+
+  const handler = (event, context) => {
+    // ...
+  };
+
+  return { handler };
+}
+```
+
+And then in your integration, where you call `setAdapter`, provide this name in `exports`:
+
+```diff
+export default function createIntegration() {
+  return {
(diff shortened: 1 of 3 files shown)


src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
+---
+layout: ~/layouts/MainLayout.astro
+title: Astro Adapter API (experimental)
+---
+
+Astro is designed to make it easy to deploy to any cloud provider for SSR (server-side rendering). This ability is provided by __adapters__, which are [integrations](./integrations-reference/).
```
sarah11918, line 6:
```suggestion
Astro is designed to make it easy to deploy to any cloud provider for SSR (server-side rendering). This ability is provided by __adapters__, which are [integrations](/en/reference/integrations-reference/).
```

src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
...
+
+> Server-side rendering in Astro is *experimental*. If you are interested in building an adapter for a host now is the perfect time to help shape these APIs. If you are worried about breaking changes this might be a little too soon for you.
+
+## What is an adapter
+
+An adapter is a special kind of [integrations](./integrations-reference/) that provide an entrypoint for server-side rendering. An adapter does two things:
```
sarah11918, line 12:
```suggestion
An adapter is a special kind of [integration](/en/reference/integrations-reference/) that provides an entrypoint for server-side rendering. An adapter does two things:
```

src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
...
+
+- Implements host-specific APIs for handling requests.
+- Configures the build according to host conventions.
+
+## Building an Adapter
+
+An adapter is an [integration](./integrations-reference/) and can do anything that an integration can do.
```
sarah11918, line 19:
```suggestion
An adapter is an [integration](/en/reference/integrations-reference/) and can do anything that an integration can do.
```

src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
...
+
+The properties are:
+
+* __name__: A unique name for your adapter, used for logging.
+* __serverEntrypoint__: The entrypoint for server-side rendering.
+* __exports__: An array of named exports when used in conjunction with `createExports` (explained below).
+
+### Server Entrypoint
+
+Astro's adapter API attempts to work with any type of host, and gives a flexible way to confirm to the host APIs.
```
sarah11918, line 57: "way to confirm to the host APIs."

(Sorry, I'm not familiar with the proper language to describe this, so I'm just checking that this is intentional wording. In plain English, I'd be expecting "to confirm "X" to the host APIs" or "to provide confirmation to the host APIs")

src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
...
+
+Some serverless hosts expect you to export a function, such as `handler`:
+
+```js
+export function handler(event, context) {
+  // ...
+}
+```
+
+With the adapter API you achieve these by implementing `createExports` in your `serverEntrypoint`:
```
sarah11918, line 69:
```suggestion
With the adapter API you achieve this by implementing `createExports` in your `serverEntrypoint`:
```

src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
...
+
+  const handler = (event, context) => {
+    // ...
+  };
+
+  return { handler };
+}
+```
+
+And then in your integration, where you call `setAdapter` provide this name in `exports`:
```
sarah11918, line 85:
```suggestion
And then in your integration, where you call `setAdapter`, provide this name in `exports`:
```

src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
...
+        });
+      },
+    },
+  };
+}
+```
+
+#### Start
+
+Some hosts expect you to *start* the server yourselves, for example by listening to a port. For these types of hosts, the adapter API allows you to export a `start` function which will be called when the bundle script is ran.
```
sarah11918, line 106:
```suggestion
Some hosts expect you to *start* the server yourselves, for example by listening to a port. For these types of hosts, the adapter API allows you to export a `start` function which will be called when the bundle script is run.
```

src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
...
+    );
+  });
+}
+```
+
+The following methods are provided:
+
+##### app.match(request)
+
+This method is used to determine a request is matched by the Astro app's routing rules.
```
sarah11918, line 143:
```suggestion
This method is used to determine if a request is matched by the Astro app's routing rules.
```

src/pages/en/reference/adapter-reference.md
matthewp: Oops, I meant __conform__

sarah11918: Ha! That was my next guess!  :)

src/pages/en/reference/adapter-reference.md
```diff
@@ -0,0 +1,159 @@
...
+    event.respondWith(
+      app.render(event.request)
+    );
+  });
+}
+```
+
+The following methods are provided:
+
+##### app.match(request)
```
sarah11918, line 141: Just a thought that maybe this would flow better if these two were swapped in order? app.render seems to be the more common method, and the one used in the example above?

If the first thing you're going to say about app.match(request) is that you can usually use the other one (that you haven't described yet) instead, then maybe it makes sense to describe app.render first? And, right in there, you can say that it can be used for 404s if you provide a 404.astro file?

 Then, when you get to app.match, you've already talked about the more commonly used method, and you can just say to use this if you want to handle 404s differently?

matthewp: Good point, I'll switch this.

PR #260, opened 2022-03-25
Why Astro? - initial draft 

sarah11918:
First draft of the "intro-type" page that will go first in Core Concepts in the sidebar.


FredKSchott:
Closing this for now, to be picked back up post release-week. @natemoo-re is working on a blog post for next week that will touch on some of these ideas, Once that's out it would be good to pick this back up (post release-week!) and see if we can tie the two posts together.

PR #261, opened 2022-03-25
Static Assets - added section for public folder

sarah11918:
In the course of editing all the current pages, I thought an example of referencing some files in public would be useful.

### DIFF
--- a/src/pages/en/guides/imports.md
+++ b/src/pages/en/guides/imports.md
@@ -45,7 +45,7 @@ Reference them in your Astro component template via absolute URL paths, from the
 ```
 
 ## `src/`
- The following applies to files in your `src/` directory, which Astro is ultimately responsible for.
+ The following applies to files located in your `src/` directory, which are processed by Astro.
 
 ### JavaScript & ESM
 
--- a/src/pages/en/guides/imports.md
+++ b/src/pages/en/guides/imports.md
@@ -17,9 +17,37 @@ Astro uses Vite as its internal build system. Vite provides Astro with built-in
 - Markdown (`.md`)
 - WASM (`.wasm`)
 
-Any files in your `public/` directory are copied into the final build, untouched by Vite or Astro. The following applies to files in your `src/` directory, which Astro is ultimately responsible for.
+## `public/`
+Any files in your `public/` directory are copied into the final build, untouched by Vite or Astro. 
 
-## JavaScript & ESM
+Reference them in your Astro component template via absolute URL paths, from the `public/` directory.
+
+```
+├── public/
+│   ├── images/
+|   |   └── photo1.jpg
+│   ├── styles/
+|   |   └── global.css
+|   └── utils.js
+```
+
+```astro
+---
+// Astro component script
+---
+<head><link rel="stylesheet" href="/styles/global.css"></head>
+<body>
+  <img src="/images/photo1.jpg"> 
+
+  <script type="module" src="/nav-utils.js"></script>
+</body>
+
+```
+
+## `src/`
+ The following applies to files in your `src/` directory, which Astro is ultimately responsible for.
+
+### JavaScript & ESM
 
 Astro was designed for JavaScript’s native ES Module (ESM) syntax. ESM lets you define explicit imports & exports that browsers and build tools can better understand and optimize for. If you're familiar with the `import` and `export` keywords in JavaScript, then you already know ESM!
 
@@ -35,7 +63,7 @@ import { getUser } from './user.js';
 
 All browsers now support ESM, so Astro is able to ship this code directly to the browser during development.
 
-## TypeScript
+### TypeScript
 
 Astro includes built-in support to build TypeScript files (`*.ts`) to JavaScript. Astro components also support TypeScript in the frontmatter script section.
 
@@ -43,7 +71,7 @@ Note that this built-in support is build only. By default, Astro does not type-c
 
 <!-- To integrate type checking into your development/build workflow, add the [@snowpack/plugin-typescript](https://www.npmjs.com/package/@snowpack/plugin-typescript) plugin. -->
 
-## JSX
+### JSX
 
 Astro includes built-in support to build JSX files (`*.jsx` & `*.tsx`) to JavaScript.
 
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/imports.md
```diff
@@ -17,9 +17,37 @@ Astro uses Vite as its internal build system. Vite provides Astro with built-in
...
+<body>
+  <img src="/images/photo1.jpg"> 
+
+  <script type="module" src="/nav-utils.js"></script>
+</body>
+
+```
+
+## `src/`
+ The following applies to files in your `src/` directory, which Astro is ultimately responsible for.
```
aFuzzyBear, line 48:
```suggestion
 The following applies to files that are located within your `src/` directory, which Astro is wholly responsible for.
```

src/pages/en/guides/imports.md
```diff
@@ -17,9 +17,37 @@ Astro uses Vite as its internal build system. Vite provides Astro with built-in
...
+
+### JavaScript & ESM
 
 Astro was designed for JavaScript’s native ES Module (ESM) syntax. ESM lets you define explicit imports & exports that browsers and build tools can better understand and optimize for. If you're familiar with the `import` and `export` keywords in JavaScript, then you already know ESM!
```
aFuzzyBear, line 52:
```suggestion
Astro was designed for JavaScript’s native ES Module (ESM) syntax. ESM lets you define explicit _imports_ & _exports_ that browsers and build tools can better understand and optimize for. If you're familiar with the `import` and `export` keywords in JavaScript, then you already know ESM!
```

src/pages/en/guides/imports.md
```diff
@@ -35,23 +63,23 @@ import { getUser } from './user.js';
 
 All browsers now support ESM, so Astro is able to ship this code directly to the browser during development.
```
aFuzzyBear, line 64:
```suggestion
ESM is now native to all browsers, this lets Astro ship code directly to the browser during development, using Astro's own Hot-Module Reloading (HMR).
```

src/pages/en/guides/imports.md
sarah11918: Actually, I like *neither* of those "Astro is responsible for" options.  :)

What about changing that to something descriptive:

"The following applies to files located within your `src/` directory, which are processed by Astro."  ??

aFuzzyBear (commented):
a couple of slight amendments from me,

src/pages/en/guides/imports.md
sarah11918: I don't think this helps the paragraph since import and export are inlined later. Italics isn't the *best* for accessibility (bold is better) and I'm not sure I'd do either one here, since I think "regular words" suit this description, followed up by the inline emphasis.

src/pages/en/guides/imports.md
sarah11918: Changed! Better, thanks! (But, it's 2 diff sentences, so I just edited myself.)

src/pages/en/guides/imports.md
sarah11918, line 48:
```suggestion
 The following applies to files located in your `src/` directory, which are processed by Astro.
```

src/pages/en/guides/imports.md
```diff
@@ -17,9 +17,37 @@ Astro uses Vite as its internal build system. Vite provides Astro with built-in
...
+│   ├── images/
+|   |   └── photo1.jpg
+│   ├── styles/
+|   |   └── global.css
+|   └── utils.js
+```
+
+```astro
+---
+// Astro component script
```
FredKSchott, line 36: If there's nothing in the front matter, it's safe to remove:

```diff
- ---
- // Astro component script
- ---
<head>
  <link rel="stylesheet" href="/styles/global.css">
</head>
<body>
  <img src="/images/photo1.jpg">
  <script type="module" src="/nav-utils.js"></script>
</body>
```

src/pages/en/guides/imports.md
```diff
@@ -17,9 +17,37 @@ Astro uses Vite as its internal build system. Vite provides Astro with built-in
...
+<head><link rel="stylesheet" href="/styles/global.css"></head>
+<body>
+  <img src="/images/photo1.jpg"> 
+
+  <script type="module" src="/nav-utils.js"></script>
+</body>
+
+```
+
+## `src/`
```
FredKSchott, line 47: I don't think that this change should force all the existing headers to be nested. I still think the main organization of this page is grouped "by file type".

Can we keep the rest of this page as-is, and instead rename the `public/` header to something like `public/ vs. src/` or `Static Assets in public/`. 

And then rephrase the section/intro a bit to highlight that the rest of the page is about ESM imports, which only work in the `src/` directory.

sarah11918: Made a tweak to see what you think: I've put public at the bottom now, emphasizing that the page is mostly about src files, but with a quick link to the bottom.

FredKSchott: I took a pass and rewrote a bit more. Kept the `public/` reference and moved it higher onto the page, but also made it shorter / less highlighted. I think it works best as a "hey FYI public exists" mention but not a full-fledged "Here's how public works" given that we've already covered that in other parts of the site.

sarah11918: No problem! Sorry I kept getting the intent of this page wrong. I was motivated by having a clear `publc/` vs `/src` comparison, as we do see this come up in support, so I thought this might be a page to do that. 

But, I think this can make for its own content (eventually) somewhere else. My `public/` blurb will just go off and join the others in the Land of Misfit Examples for future, more tutorial-like material.  :)

Fixed a pre-existing typo, and otherwise, page looks great!

FredKSchott: lets see if this improves the support issues that you see, happy to take another look if it becomes an issue.

My hope is that this makes that point clear, but another thing that we could do is to add some bold to the paragraph, it make it pop more:

> Remember that you can also place any static asset in the [`public/` directory](/en/core-concepts/project-structure/#public) of your project, and Astro will copy the directly into your final build. `public/` files are not built or bundled by Astro, which means that any type of file is supported. You can reference a `public/` file by a URL path directly in your HTML templates.

PR #269, opened 2022-03-27
Adds an AgnosticUI + Astro starter

roblevintennis:
AgnosticUI is a UI component library where one stylesheet is used for many frameworks. These UI components that work in React, Vue 3, and Svelte so I think that using alongside Astro is a very compelling polyglot microfrontend sort of use case.

### DIFF
--- a/src/pages/en/integrations/integrations.md
+++ b/src/pages/en/integrations/integrations.md
@@ -57,5 +57,6 @@ Here are some production sites, repositories, blog posts and videos from the com
 
 [p13rnd/centauri](https://github.com/p13rnd/centauri) - Astro template: Tailwind, Svelte and authentication with Supabase
 
+[AstroAgnosticUIStarter](https://github.com/AgnosticUI/AstroAgnosticUIStarter) — Experimental starter template for leveraging Astro + AgnosticUI to render React, Vue 3, and Svelte.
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @roblevintennis! I just checked this out and read your blog post. Looks neat!

I did want to let you know, though, that we just released Astro v.0.25 and our entire rendering situation is now different! You can check out the migration guide in our docs, but basically we use "integrations" now, and the astro.config.mjs file is new and different.  https://docs.astro.build/en/migrate/

I don't think this necessarily stops us from adding your site to the list (since I'm sure there are others there that are not yet updated either!) but I did want to let you know!

roblevintennis:
Thanks @sarah11918 I appreciate you checking it out! I'll have to update the template to use the integration approach next spare cycle 👍🏽

sarah11918:
@roblevintennis Now that I'm sitting down to properly deal with this page...

The plan is to replace this page entirely with pages on our main https://astro.build site vs the docs site. I think the best place for this would be on our "Themes" page: https://astro.build/themes/  You can see that there's a link there for adding your own theme which will take you to GitHub issues for that site. There, you'll be able to submit the information necessary so that we can create a proper "card" for you to match the rest of the site. https://github.com/withastro/astro.build/issues/new/choose

I know the basic information requested when that page was built was:  a preview image, title, short description, and links to the public repo and live demo (if available), so that should be enough to file the issue and get your site in the queue there.

So, I'm going to close this PR here, and send you off to the Themes page on the main website! (I am going to share your article in our Discord, though, unless you'd like to pop in and put it in our Showcase channel!)

PR #271, opened 2022-03-28
Deleted all themes pages & references to them

sarah11918:
Now that we have https://astro.build/themes/ the theme page in the docs is redundant.

I have removed the themes.md page from all the languages that had one, as well as removed/updated any links to those pages throughout the site and in the left sidebar nav.

(Also while I was there, updated a couple of /es/ files that referenced the recently-removed renderer-reference page, since one fetchContent page was giving an error, expecting that file to exist.)


FredKSchott:
Nice! LGTM!

PR #273, opened 2022-03-29
update configuration page

FredKSchott:


src/pages/en/reference/configuration-reference.md
```diff
@@ -8,24 +8,15 @@ setup: |
   import Since from '../../../components/Since.astro';
 ---
 
-To configure Astro, add an `astro.config.mjs` file to the root of your project.
+The following reference covers all supported configuration options in Astro. To learn more configuring Astro, read our guide on [Configuring Astro](/en/guides/configuring-astro/).
```
sarah11918, line 11:
```suggestion
The following reference covers all supported configuration options in Astro. To learn about more configuring Astro, read our guide on [Configuring Astro](/en/guides/configuring-astro/).
```

src/pages/en/reference/configuration-reference.md
```diff
@@ -92,22 +83,28 @@ The value can be either an absolute file system path or a path relative to the p
...
+**Default:** `[]`
 </p>
 
-Set the UI framework renderers for your project. Framework renderers are what power Astro's ability to use other frameworks inside of your project, like React, Svelte, and Vue.
+Add Integrations to your project to extend Astro.
+
+Integrations are your one-stop shop to add new frameworks (like Solid.js), new features (like sitemaps), and new libraries (like Partytown and Turbolinks).
 
-Setting this configuration will disable Astro's default framework support, so you will need to provide a renderer for every framework that you want to use.
+Setting this configuration will disable Astro's default integration, so it is recommended to provide a renderer for every framework that you use:
```
sarah11918, line 98: I'm a little unsure about this: 

- if the default is an empty array, I would wonder, is there something *else* I have to worry about being disabled?  (I know some of our starter templates have some integrations configured, but adding a new integration shouldn't disable those existing integrations, right?)

- I'd also be wondering, how is it only *recommended* to provide a renderer for every framework? Isn't it *required*? (And, although I know the function these these framework integrations perform is rendering, just want to make sure that it's a conscious choice to refer to them informally as renderers. The alternative would be to reinforce the word integration here.)

I get that we use this similar phrasing when describing adding the Markdown plugins, as the default needs to be re-added. But with an empty integrations array as a default, there doesn't seem to be anything to disable. So maybe this isn't needed?

sarah11918:
One typo fixed, one comment re: a sentence I'm unclear about, but otherwise, All looks good!

src/pages/en/reference/configuration-reference.md
FredKSchott: Ah, this is outdated! I'll update it in the main astro repo and re-run

FredKSchott:
Merging this in since it's better than what we have today. 
i'll re-run docgen once the main astro repo docs are updated.

PR #279, opened 2022-03-29
Editor Setup page w VS Code Extension info

sarah11918:
Initial Editor Setup page draft, to highlight the Astro VS code extension

PR #281, opened 2022-03-30
Prevent `<Badge>` from wrapping across lines & tweak colours

delucis:
This PR fixes the funky rendering spotted in https://github.com/withastro/docs/pull/279#issuecomment-1082456004 and also slightly tweaks the colours of non-accented badges.


sarah11918:
I wasn't actually seeing any wrapping problems on my phone, or in any smaller screen views, but I do like the grey background of the community button better here!

Looks great to me!

delucis:
Double checked and it looks like a Firefox thing? I also can’t get the current badge to break like that in Chrome, Edge, or Safari.

FredKSchott:
LGTM regardless!

PR #282, opened 2022-03-30
Add directives reference page

### LINKED ISSUE #272: Add documentation for `class:list`
Astro has an utility called `class:list` to allow to conditionally add classes to something, it was added in 0.21 but has never been documented

It'd be cool to have documentation for it because the VS Code extension does provide a completion for it, which will inevitably leads users to use it (and wonder how to use it)

For info on usage, see [this fixture](https://github.com/withastro/astro/blob/d84bfe719a546ad855640338d5ed49ad3aa4ccb4/packages/astro/test/fixtures/astro-class-list/src/pages/component.astro) and the [PR for it](https://github.com/withastro/astro/pull/1612)

zadeviggers:
This is a work in progress pull request that adds a single page documenting all the available directives in Astro.

Todo list:
- [x] Document `is:raw` - Can someone please explain to me what it does?
- [x] Add more information about the `hoist` script directive
- [x] Document `class:list`
- [x] Document `global`
- [x] Link to directives reference page from more pages
- [x] Read-over to check flow, spelling, grammar, etc
- [x] Update to discuss new `is:new` and `is:global` directives.

Closes #272


sarah11918 (commented):
(I guess I need a comment here so that GitHub thinks I've "reviewed" it. But, see my earlier comment.)

zadeviggers:
> In general, I would need to be convinced that this is a useful reorganzing of material. (Though, there are some good content additions in here!)

Well, there is quite a bit of missing content around directives in the docs at the moment, and it's all very scattered (I had to use the vscode project wide search with a custom regex in order to find the many small scattered snippets about them), so I think that organizing all the information around directives is a very important thing to do.

I did also notice that adding a page documenting all the directives was on the [v1 beta->stable roadmap](https://github.com/withastro/rfcs/discussions/161#discussion-3972352), so I figured we may as well get started on it now.

> Specifically, I personally would be hesitant to remove `client:directives` from `framework-components`. The idea of this page (at the moment) is where you go for finding out how to use these framework components...

I agree, but the reason I moved them was to avoid having any duplication of content.

> One option I see for right now would be to propose the new material (e.g. `set:html` etc, in the place where it would currently exist in the existing structure of the docs right now, and we can deal with content first. Then, we can revisit the idea of where it belongs.

I agree that getting the content on the docs as soon as possible is important, but I think that having it all organised into one place in a logical order is just as important.

src/pages/en/core-concepts/component-hydration.md
```diff
@@ -90,6 +90,8 @@ Start importing the component JS at page load and hydrate when the import comple
 
 If more than one renderer is included in the Astro [config](/en/reference/configuration-reference), `client:only` needs a hint to know which renderer to use for the component. For example, `client:only="react"` would make sure that the component is hydrated in the browser with the React renderer. For custom renderers not provided by `@astrojs`, use the full name of the renderer provided in your Astro config, i.e. `<client:only="my-custom-renderer" />`.
 
+📚 See our [directives reference](/en/reference/directives-reference#ui-framework-components) page for more infomation on all of the `client:` directives.
```
sarah11918, line 93: This page can probably be removed entirely now. It was split into "framework-components" and "partial-hydration." When doing so, we should also make a Netlify redirect to partial-hydration. (I think anyone linking to Astro's hydration previously was probably referring more to the concept, than the hydration of the framework components themselves, so that's my recommendation.)

I'll make a separate PR for this.  :)

src/pages/en/core-concepts/astro-components.md
```diff
@@ -335,6 +334,9 @@ To send JavaScript to the browser without [using a framework component](/en/core
   // Processed! Bundled! ESM imports work, even to npm packages.
 </script>
 ```
+
+📚 See our [directives reference](/en/reference/directives-reference#script-and-style-tags) page for more information about the directives available  on `<script>` tags.
```
sarah11918, line 338: 1. Is "directives" a standard term that we expect readers to understand? 

In framework-components, we explain directives as "a component attribute to define how your component should be rendered and hydrated." Is there some similar phrase we could add here so that "directive" doesn't just sneak up out of no where? (I'm not sure I would have known what a "directive" was when I started with Astro, but I did know and understand the word "attribute.") 

I checked the entire page and we do mention "special Astro directives" earlier in this page, but also don't explain/define, and maybe we should in at least one place? Here or there? Also note that we don't mention anything about directives in the section directly above, re CSS styles. So maybe we should there, too? This page doesn't indicate that there are any `<style>` directives available.

2. (slightly lower down) This page mentions "hoist" and I think that's going away, if it hasn't already. Probably safe for the purposes of this PR to consider it gone?

src/pages/en/core-concepts/framework-components.md
```diff
@@ -78,41 +78,9 @@ the user scrolls down and the component is visible on the page -->
 
 ### Available Hydration Directives
 
-#### `client:load`
+There are serveral hydration directives available for UI framework components, such as `client:load`, `client:idle`, and more!
```
sarah11918, line 81:
```suggestion
There are serveral hydration directives available for UI framework components: `client:load`, `client:idle`, `client:visible`, `client:media={QUERY}` and `client:only=" "`
```

Can we maybe meet in the middle, and since there are only 5 of them, list them all here??  ;)

src/pages/en/core-concepts/framework-components.md
```diff
@@ -78,41 +78,9 @@ the user scrolls down and the component is visible on the page -->
...
-#### `client:only=" "`
-
-Start importing the component JS at page load, similar to `client:load`.
-
- >⚠️ This component will be **skipped** at build time, and to assist the client, you should specify which renderer to use from the array in your [`astro.config.mjs` configuration](/en/reference/configuration-reference).
- >
- > e.g. `<client:only="react" />` or `<client:only="my-custom-renderer" />`
- 
- 💡 *Useful for components that are entirely dependent on client-side APIs.* 
+📚 See our [directives reference](/en/reference/directives-reference#ui-framework-components) page for the full list of hydration directives, and their usage.
```
sarah11918, line 83:
```suggestion
📚 See our [directives reference](/en/reference/directives-reference#ui-framework-components) page for a full description of these hydration directives, and their usage.
```

src/pages/en/guides/styling.md
```diff
@@ -84,7 +84,7 @@ const backgroundColor = "rgb(24 121 78)";
 <h1>Hello</h1>
 ```
 
-Any *serializable* front matter variable is supported, including props passed to your component through `Astro.props`.
+📚 See our [directives reference](/en/reference/directives-reference#definevarsvariables) page to learn more about `define:vars`.
```
sarah11918, line 87: Just pointing out that earlier on this page, in Global styles, we refer to global as an attribute, not a directive, and we don't link to the directives page. (Side note: maybe the Global header should be one level higher, so that it shows up in the sidebar with scoped?) Should this be treated the same way?

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
+---
+layout: ~/layouts/MainLayout.astro
+title: Directives Reference
+---
+
+Directives are special properties that can be passed to HTML elements, Astro components, And UI Framework components. Directives can only be used inside of Astro components.
```
sarah11918, line 6:
```suggestion
Directives are special properties that can be passed to HTML elements, Astro components, And UI Framework components. 

Directives are Astro-specific attributes, used in your `.astro` component to "direct" Astro as it renders your HTML template.
```

(Is that accurate to say? I like linking directives with attributes because I think that's a more commonly-known word to less-experienced developers.)

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+
+The directives on this page are originised by what they can be used on.
```
sarah11918, line 8:
```suggestion
This page describes the options available for directing Astro to render different types of content on your page.
```

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+
+💡 *Useful for components that are entirely dependent on client-side APIs.* 
+
+>⚠️ This component will be **skipped** at build time, and to assist the client, you should specify which renderer to use from the array in your [`astro.config.mjs` configuration](/en/reference/configuration-reference).
+>
+> e.g. `<client:only="react" />` or `<client:only="my-custom-renderer" />`
+
+## Script and Style tags
+
+These directives can be used on HTML `<script>` and `<style>` tags.
```
sarah11918, line 52:
```suggestion
The following directives can be used on both HTML `<script>` and `<style>` tags.
```

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+    color: var(--textColor);
+  }
+</style>
+
+<script define:vars={{ message }}>
+  alert(message);
+</script>
+```
+
+### `hoist`
```
sarah11918, line 82: As mentioned, I *think* `hoist` is going away soon enough that maybe it should just be removed from this page right now? Check with @natemoo-re

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+
+Start importing the component JS at page load, similar to `client:load`.
+
+💡 *Useful for components that are entirely dependent on client-side APIs.* 
+
+>⚠️ This component will be **skipped** at build time, and to assist the client, you should specify which renderer to use from the array in your [`astro.config.mjs` configuration](/en/reference/configuration-reference).
+>
+> e.g. `<client:only="react" />` or `<client:only="my-custom-renderer" />`
+
+## Script and Style tags
```
sarah11918, line 50: I'm not sure how I feel about this organization. There's one directive that can be used on both, then (because hoist is going away) there's one that can be used on style. (And then below, there's the everything section)

I think this is in part why originally, this new organization didn't sit well with me. I'm not convinced that this is a useful section grouping, in part because if I look in the sidebar, I'd click here to look for CSS or script stuff, and I'm honestly not sure that I would further look and notice "Everything."

So, I'll continue reading the rest of this for content, but I don't know that this is page is just yet best organized for people who will be scanning sidebars and trying to go directly to particular sections of content.

src/pages/en/reference/directives-reference.md
sarah11918, line 52: (NB: I originally edited this to say "used on both" script and style tags, but then when I read further, I realized that this entire section wasn't just things that can be used on both. That's what led to my long comment above.)

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+
+These directives can be used on HTML `<script>` and `<style>` tags.
+
+### `define:vars={variables}`
+
+Pass variables into a `<script>` or `<style>` tag. Any *serializable* front matter variable is supported, including props passed to your component through `Astro.props`.
+
+>⚠️ `define:vars` cannot be used on a `<script>` tag that makes use of the `hoist` directive.
```
sarah11918, line 58: How will these two warning notes need to change if hoist is now the default behaviour? (Also related to whether we need to continue to describe hoist if it's no longer going to be a specific directive that you add, but is in fact default behaviour. Do we need to document that adding this makes the script not hoisted? So, what do we actually need to say about hoist?)

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+</script>
+```
+
+>⚠️ `hoist` Always needs to be used in conjunction with `type="module"`.
+>
+> e.g. `<script hoist type="module">`
+
+### `global`
+
+Make the contents of a `<style>` tag apply globaly on pages where the component is included by disabling Astro's CSS scoping system.
```
sarah11918, line 101: "apply globally on pages where the component is included" -- just checking, is this exactly what the global tag does? These styles take affect only where that component is included? I had thought global made styles global to your whole site, no matter which component they were written in (because we have a warning about that in the CSS/Styling page that says "this affects all the h1's on your site).

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+
+This is equivalent to wrapping all of the selectors within a `<style>` tag with `:global()`.
+
+```astro
```
sarah11918, line 105: The examples here show that the two forms are analogous, but neither of them describes what the styling/CSS effect produced is. This is a case where I think the original page itself has more information than the part on the directives page, so probably requires deciding what is going to go where. (And, I'm not convinced having this here is better than having it on the CSS page itself, making readers click through.)

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+
+The opposite of `set:html`. `set:text` ensures that any HTML content passed to it is escaped.
+
+```astro
+---
+const potentialyDangerousContent = await fetchUserGeneratedContent();
+---
+<Fragment set:text={potentialyDangerousContent}>
+```
+
```
sarah11918, line 159: I like the examples above, but I feel like I'd want to know what the default behaviour is, if neither of these are used. (Is there a default? Is it context dependent?) If one is going to escape and one isn't, then why do I need more than one directive?  :)

In the first example, not using any directive did NOT escape, but set:html did. So, I might be wondering when I would use/need set:text, and not just leave it to the default behaviour. So I think some kind of further note might be useful here.

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+## Everything
+
+These directives are avalible on HTML elements, UI Framework components, Astro components, the works!
+
+### `is:raw`
+
+Instructs the Astro compiler to treat any children of that element as text, similar to the default behavior of `<script>` and `<style>` which don't support expressions. This means that all special Astro templating syntax will be ignored.
```
sarah11918, line 166:
```suggestion
Instructs the Astro compiler to treat any children of that element as text. This means that all special Astro templating syntax will be ignored.
```
(I found that extra note a little confusing. Does the sentence work just as well without?)

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+<Katex is:raw>Some conflicting {syntax} here</Katex>
+```
+
+### `class:list`
+
+Serializes a JavaScript expression to a string of CSS class names. Similar to the [clsx](https://github.com/lukeed/clsx) helper library.
+
+The final `class` string is passed to the component/element as the `class` prop.
+
+>⚠️ `class:list` will not be parsed if its value is not an expression.
```
sarah11918, line 186: I would move the warning below the expression. Show how it does work first, then a warning about how it doesn't.

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,195 @@
...
+The opposite of `set:html`. `set:text` ensures that any HTML content passed to it is escaped.
+
+```astro
+---
+const potentialyDangerousContent = await fetchUserGeneratedContent();
+---
+<Fragment set:text={potentialyDangerousContent}>
+```
+
+## Everything
```
sarah11918, line 160: I do love your description of "the works"! :)

But, I don't think "Everything" is a useful header. And, if I've already scanned the sidebar and found what I was looking for (script, css, framework), then I wouldn't think to keep looking and notice "Everything."

So, I'm open to suggestions for this. Two thoughts come to mind, either renaming to something like "universal" (not great, but you get what I mean) and/or moving these directives up to the TOP and starting out with, "Here's some stuff you can use on everything!" then let the page get into specific use cases.

sarah11918:
Hey @zadeviggers ! I added a bunch of specific comments (and, those that are suggestions are just when it's easier to type inline to express an idea. Not requiring you to accept them literally; just giving a thought!)

Once again, thanks for all your work putting this together, and don't interpret the fact that I commented on a lot of little things as me not being appreciative or not liking what you've done! I do just have some specific thoughts, much of them around organization to make things easy to spot for the reader, so calling attention to things in various places looks like a lot of comments!  :)

src/pages/en/guides/styling.md
zadeviggers: I think that it's better to use the correct terminology here, and on one of the first pages explain what a directive is (but that's out of the scope of this PR)

src/pages/en/reference/directives-reference.md
zadeviggers: I think this is redundant now, given that there's now also `is:inline`.

src/pages/en/reference/directives-reference.md
zadeviggers: Already done!

src/pages/en/reference/directives-reference.md
zadeviggers: Hmm, I'm not actually sure on this one. I'll ask someone in discord.

src/pages/en/reference/directives-reference.md
zadeviggers: I was originally worried about content duplication, which is why I moved it here. I've added a link back to the styling page - is that good enough for now?

src/pages/en/reference/directives-reference.md
zadeviggers: Alright, Fred confirmed that they only apply to pages where the component is included. ([Message link](https://discord.com/channels/830184174198718474/830184175176122389/959898750183149568))

src/pages/en/reference/directives-reference.md
zadeviggers: Yeah I think you're right.

src/pages/en/core-concepts/astro-components.md
zadeviggers: Quite possibly, but I think changing either of those things are out of this PR's scope.

I think that for 1, the change should be made in the earlier reference in the page.

FredKSchott:
The content itself is great, but the "copy" like in the introduction need some work ("And UI Framework components." for example, is a fragment).

If you'd like help with the copy, I'd be happy to give it a pass and then merge this with our changes combined together?

src/pages/en/reference/directives-reference.md
```diff
@@ -0,0 +1,219 @@
+---
+layout: ~/layouts/MainLayout.astro
+title: Directives Reference
+---
+
+And UI Framework components. Directives can only be used inside of Astro components.
```
sarah11918, line 6: Did you mean to leave this in? It seems like a fragment from the previous draft.

zadeviggers: Whoops, yep

zadeviggers:
> If you'd like help with the copy, I'd be happy to give it a pass and then merge this with our changes combined together?

Yes, I would really appreciate that thank you!

FredKSchott (approved):
This looks great! I actually don't have much to add, but I'll do a bit of cleanup on the wording post-merge. Thanks for the awesome contributions!

PR #283, opened 2022-03-30
Changed all astro.config references to .mjs

sarah11918:
Includes the docgen file, but NOT the configuration-reference file, since it is auto-generated.

NOTE: I think the docgen file will address the top of the page config example in /reference/configuration-reference but this page ALSO contains `astro.config.js` in the section "project root." So, whatever generates that description should be updated.

### DIFF
--- a/src/pages/en/guides/configuring-astro.md
+++ b/src/pages/en/guides/configuring-astro.md
@@ -29,7 +29,7 @@ export default {}
 
 ## Supported Config File Types
 
-Astro supports several file formats for its JavaScript configuration file: `astro.config.mjs`, `astro.config.mjs`, `astro.config.cjs` and `astro.config.ts`. 
+Astro supports several file formats for its JavaScript configuration file: `astro.config.js`, `astro.config.mjs`, `astro.config.cjs` and `astro.config.ts`. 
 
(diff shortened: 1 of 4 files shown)


src/pages/en/guides/configuring-astro.md
```diff
@@ -29,15 +29,15 @@ export default {}
 
 ## Supported Config File Types
 
-Astro supports several file formats for its JavaScript configuration file: `astro.config.js`, `astro.config.mjs`, `astro.config.cjs` and `astro.config.ts`. 
+Astro supports several file formats for its JavaScript configuration file: `astro.config.mjs`, `astro.config.mjs`, `astro.config.cjs` and `astro.config.ts`.
```
Princesseuh, line 32: Duplicated here

sarah11918: D'oh! Good catch. Thanks.

sarah11918:
```suggestion
Astro supports several file formats for its JavaScript configuration file: `astro.config.js`, `astro.config.mjs`, `astro.config.cjs` and `astro.config.ts`. 
```

FredKSchott:
LGTM!

PR #284, opened 2022-03-30
WIP: Rename integrations to extensions

FredKSchott:
This is an idea that I've been kicking around and mentioned to some core maintainers, but I wanted to see what the change would look like on our docs site.

The reasoning: "Integrations" is both a mouthful (4 syllables) and a pretty meaningless word. It's not bad, but could it be better? "Extensions" came out of the idea of bring more meaning to the word: You add an extension because you want to "extend" Astro with some new feature, power, etc. Seeing this change on the docs site, and I think it's better.

One downside / reason not to make the change: Obviously it will cost us in thrash, even if we make this change in a backwards-compatible way. One bright-side is that we are about to do a config object refactor, so our users will already be updating their config file one more time before v1.0.0-beta.

Example: https://deploy-preview-284--astro-docs-2.netlify.app/en/guides/integrations-guide/ (URLs not yet updated, but will be if we decide to merge this PR)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -163,7 +163,7 @@ packages:
     resolution: {integrity: sha512-4YM4hM02ILvvsSQljAlwLY7OWymlhbikY7O/gsZqELVGMXixemA6dI3GTr2K6duU4ktPBJ3JexuU/OFyb5jvtQ==}
     hasBin: true
     dependencies:
-      '@astrojs/svelte-language-integration': 0.1.2_typescript@4.6.2
+      '@astrojs/svelte-language-extension': 0.1.2_typescript@4.6.2
       '@vscode/emmet-helper': 2.8.4
       lodash: 4.17.21
       source-map: 0.7.3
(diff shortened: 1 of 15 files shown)


sarah11918:
I'm probably going to be in the minority here, but I actually do prefer "integrations"! Maybe not for any good reasons, maybe even for the wrong ones! I just like it.  :)

I don't feel strongly enough to disapprove of this, but I actually think integration is a simpler word than extension, despite the extra syllable. The pronunciation is simpler, the verb form more closely matches the noun. (And, this is totally just a me thing, but this word in particular always makes me second-guess myself -tion or -sion. You just happened to pick a word I hate spelling!) It feels more "universally international" just as a straight up word. (Putting the meanings aside.)

As for semantics, I guess integration makes me think of a more overall, total unit whereas extensions make me think add-on. So, yes, these add-ons extend Astro. But also, they integrate into the whole. So I don't think either one *doesn't* work. Integration is more about incorporating and absorbing into the whole, and extending does give that "tacking on" or "reaching out" image. It's "inclusion" vs "addition."

Certainly, if you were going to change, I agree with doing it now! And, there are benefits to picking a word that makes you happy.  :)

FredKSchott:
Update: I think it's too close to release to make this change, people are already writing blog posts and recording videos. 

I still do like extensions more, so I think we'll have one last chance to change this when we release the v1.0 release candidate. I'll resubmit this for that, with plenty of time to message.

Thanks @sarah11918 for your feedback!

PR #290, opened 2022-03-31
Typescript - draft of page

sarah11918:
OK, so not being a TypeScript user myself... 

Here's at least the information from around the docs, and from https://vitejs.dev/guide/features.html#typescript

I tried to examine a bunch of projects/starters using TypeScript, and most don't configure very much at all. I even included a configuration from Astro Ink that I haven't seen anyone else used, but I'm not sure is even really necessary.

So, forgive my ineperience with TS, but here's some stuff to get that page rolling....


FredKSchott:
Took a pass at it and cleaned up a lot about the config/setup part. I'll create a PR back to astro since I think we need to update all of the example `tsconfig.json` files anyway.

if I missed anything, I'd love a followup PR!

PR #296, opened 2022-04-02
chore: update dependencies

JuanM04:
- Updated Astro 1.0.0 beta
- Updated all the other dependencies except for React (it throws a warning about hooks - may be related to the Docsearch component)


FredKSchott:
A little scared about this one, but if tests are passing and it's looking good... LGTM! I'll give the docs site one last one through post-release to make sure that we didn't break anything

PR #302, opened 2022-04-04
docs(cli): update reference for specify root

SebasG22:
The latest Astro package uses `root` instead of project-root flag.
https://github.com/withastro/astro/blob/main/packages/astro/src/cli/index.ts#L39

This took me a while when using others configurations that are not the default one (current directory)


natemoo-re (approved):
LGTM! Thank you so much!

PR #306, opened 2022-04-04
Fix anchor links

delucis:
Nesting the anchor inside the heading was causing it to be picked up by Google, but also could have a11y consequences if it was read by a screenreader as part of the heading.

This PR moves the anchor link out of the heading element and follows the guidance in this article to also make the links more accessible: https://amberwilson.co.uk/blog/are-your-anchor-links-accessible/

TLDR: This uses a structure like:

```html
<h2 id="id">Exciting New Feature</h2>
<a href="#id">
  <span aria-hidden="true">#</span>
  <span class="sr-only">Section titled Exciting New Feature</span>
</a>
```

1. Moving the link after the heading means the link itself isn’t read out in a screen reader table of contents, won’t be picked up by search engine crawlers, and as a bonus means we don’t need to strip it out when building our own table of contents.

2. The visible link icon (`#` in the sample above, an SVG icon added in this PR) is marked as `aria-hidden` to hide it from screen readers.

3. Some useful text content is included for screen readers only, so that tabbing to the anchor link will tell the user the title of the section they are entering.

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -44,11 +44,8 @@ export default defineConfig({
 						{
 							properties: { class: 'anchor-link' },
 							behavior: 'after',
-							group: h('div.heading-wrapper'),
-							content: (heading) => [
-								h(`span.anchor-icon.level-${heading.tagName}`, { ariaHidden: 'true' }, AnchorLinkIcon),
-								h('span.sr-only', `Section titled ${toString(heading)}`),
-							],
+							group: ({ tagName }) => h(`div.heading-wrapper.level-${tagName}`),
+							content: (heading) => [h(`span.anchor-icon`, { ariaHidden: 'true' }, AnchorLinkIcon), h('span.sr-only', `Section titled ${toString(heading)}`)],
 						},
 					],
 				],
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -3,6 +3,18 @@ import type { AstroMarkdownOptions } from '@astrojs/markdown-remark';
 import { defineConfig } from 'astro/config';
 import preact from '@astrojs/preact';
 import react from '@astrojs/react';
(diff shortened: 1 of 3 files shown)


sarah11918:
Nicely done, @delucis !  I don't think it's a problem, but just an observation that on mobile, the link icons are **after** the header, whereas on medium/large screen views, they are **before**. And, fwiw, on my phone in landscape (wide enough to be 2 column view) in Kiwi chromium-based browser, the links weren't visible until tapping on the header. Then, they showed up (on the left).

None of those are problems, but maybe it's more desirable for the links to also be on the left in mobile view, too? (I'm not sure whether that's standard, expected behaviour or not.) Not enough to block this, though, I don't think!

sarah11918:
Looking good! Seeing both now, I think I'm personally fine either way re: only showing the links on hover. But my gut still says show until/unless we decide we don't want it.

zadeviggers (commented):
Woah, this is way better than my original implementation!

I would personally prefer if on desktop the links were hidden except on hover, but that's just my personal preference.

delucis:
> Woah, this is way better than my original implementation!

I'm standing on the shoulders of giants! 😉

> I would personally prefer if on desktop the links were hidden except on hover, but that's just my personal preference.

Maybe I could reinstate the behaviour but at a higher breakpoint. If only there was `@media (interaction: mouse)` or something.

sarah11918:
I like it!  @zadeviggers have you had a chance to check out the hover preview?

zadeviggers (approved):
Everything looks good, including the hover detection! (Tested on Firefox stable on Windows 10, and Chrome developer channel on Chrome OS)

sarah11918:
OK, let's get this money!

PR #307, opened 2022-04-04
Expand i18n support

### LINKED ISSUE #280: Translation not possible on some elements
Translation is not possible on theses elements :

- `Top navigation` title
- `Site navigation` title
- `Table of content` title
- Search input in the top bar
  - `Press '/' to search`
- The entire search modal component
- The `<sponsor/>` component 
  - the text "Sponsored by"
  - the aria-label on the sponsor
- The sidebar content placeholder when no translation found
  - "No Translations Found"
  - "View in English"
- "Skip to Content" text
- Avatar list component
  - Contributor image alt & title
  - Additionnal contributor text
- config.ts `SITE` and `OPEN_GRAPH` constants for defaults values in each languages

> ### Edit: more found !
> - `InstallLayout` layout component
>   - `title` & `description` is overwriten
>   - the entire content is static on this layout
> - the `configuration-reference` is auto-generated in english but no options si available for i18n's translations

It's all i got when searching in the code, may be more hidden in corners that i didn't found !

delucis:
Mostly addresses #280 

Makes it possible to internationalise the following UI:

- [X] Skip link text

- [X] Sidebar placeholder content when no translation found
  - [X] "No Translations Found"
  - [X] "View in English"

- [X] Site title & description meta

- [X] `InstallLayout` component
    The shared introduction to the installation guides is now in `install/_intro.md`. Translators can use the same pattern of `_intro.md`, `auto.md`, and `manual.md` with the `InstallLayout` to build the same kind of experience.

- [X] DocSearch box & modal
	Because DocSearch contains a *lot* of microcopy and has [its own i18n API](https://docsearch.algolia.com/docs/api/#translations), I’ve separated translations for it in a separate `i18n/docsearch.ts` file. That way we can follow the same format as the component expects and don’t need to reinvent the wheel.

- [X] `title` attribute on navigation menus
  - [X] “Top navigation” title
  - [X] “Site navigation” title
  - [X] “Table of content” title

#### Sponsorship Ad

@FredKSchott Would we be good to support i18n of the “Sponsored by Netlify” ad in the sidebar as requested in #280 or is it safer to keep that only in English?

#### Open Graph metadata

I can’t make up my mind how to handle this. Currently `src/config.ts` includes an `OPEN_GRAPH` object with a default OG image URL, Twitter handle, etc. The only bit that really needs i18n is the `alt` text for the OG image. We could move it to the translations file, but then it isn’t together with the image URL if that changes, which seems like a maintenance headache. I’ve left it as is because I think the impact is small, but happy to hear suggestions!

#### Avatar list

This isn’t currently in use. Should I revive it? Otherwise it could be removed.

**Update:** I’ve opened #315 to suggest using this on the homepage — of course including i18n options.


delucis:
Just resolved conflicts after the upgrade to Astro v1. Should be good to go if someone has a moment to take a look.

tony-sull (approved):
LGTM! 🚀  Great work here, the i18n support here is looking very clean and I'm always happy to see a `.tsx` file replaced with `.astro` 😄

PR #310, opened 2022-04-05
Updating french `getting-started` translation

Chrissdroid:
Change the current `src/pages/fr/getting-started.md` to reflect latest version and avoid 404 links

Added a banner at the top to suggest user to help with the translation if he want to


src/pages/fr/getting-started.md
```diff
@@ -1,59 +1,97 @@
...
+## Rejoindre notre communauté
 
-Cette documentation reflète toujours la dernière version stable d'Astro. Dès que nous auront atteint l'étape de la v1.0, nous ajouterons la possibilité de parcourir la documentation versionnée.
+Venez sur [notre Discord](https://astro.build/chat) pour partager vos créations et obtenir l'aide d'une communauté active et conviviale !
 
-## Rester informé
+💬 Dites bonjour dans notre salon `#introduce-yourself` !
 
-Le compte Twitter [@astrodotbuild](https://twitter.com/astrodotbuild) est la source officielle d'information de l'équipe Astro.
+💬 Demandez de l'aide a l'équipe de support dans notre salon `#support` !
```
Olyno, line 86:
```suggestion
💬 Demandez de l'aide à l'équipe de support dans notre salon `#support` !
```

Olyno (changes requested):
This revivew should fix most of the spelling mistakes, and change some sentences that, in my opinion, didn't make much sense.

**Note:** The contribution section is missing.

sarah11918:
Hi @Chrissdroid @Olyno !

I just wanted to let you know that I'm going to separately make a PR myself for the config.ts changes because a) this preview won't automatically deploy via Netlify, and I can't merge anything that I can't see deployed and b) it's better to separate these two parts of the process anyway. (I'm sorry, that 2nd one is on me: I did ask you to do it.)

I don't think the deploy issue will be a huge one going forward. I just noticed that the Netlify logs are requesting that these depoys be authorized because they are coming from "untrusted deploys"... so it's just eventually making sure that the PRs look "trusted" going forward, however we fix that. But, I'd like the sidebars to get updated sooner rather than later, so I'm going to do that separately elsewhere.

(And, don't worry, now that I'm seeing the logs, I can see that there are other newer contributors' PR branches that are having the same issue, so we'll figure out what's going on there as we start to accept initial translations again. It's not just you!)  😄

sarah11918:
> **Note:** The contribution section is missing.

I would expect this, because it requires Astro's GitHub authorization token. It should just be missing from contributor's local pages, I think, but otherwise you should be able to work like normal.  :)

sarah11918:
I'm going to merge this now, to get it out of our hands so we can start thinking about the i18n sites fresh. Thank you both for this! I will just merge the conflict now that we're changing `npm init astro` to `npm create astro@latest` and then we have a fresh page up until we decide what happens next! 🥳

PR #314, opened 2022-04-06
Fix: logo's link to astro.build clickable on mobile view

sarah11918:
I noticed that the smaller screen version of the docs site couldn't click through to astro.build via the logo as it did in larger screen sizes. @delucis helped me narrow down that it was a negative `z-index` value making the link unclickable.

We both tested, and removing the `z-index` property entirely didn't seem to affect other parts of the header for either of us.


bholmesdev (approved):
You fixed ze index!

PR #318, opened 2022-04-06
Temp fix: escape markdown in client:media directive example

sarah11918:
Emergency fix replacing braces so that the page at least builds.


sarah11918:
Getting that money on my own!

PR #319, opened 2022-04-06
fix: escape `sr-only` labels

natemoo-re:
Found the culprit! The auto-generated `sr-only` labels were not escaping any curly braces. Fixed this by adding the `is:raw` directive to any labels that are created.


astro.config.ts
```diff
@@ -15,6 +15,11 @@ const AnchorLinkIcon = h(
 		d: 'M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z',
 	})
 );
+const createSROnlyLabel = (text: string) => {
+	const node = h('span.sr-only', `Section titled ${text}`);
+	node.properties['is:raw'] = true;
+	return node;
+};
```
delucis, line 22: `h` can take properties directly as the second argument:

```suggestion
const createSROnlyLabel = (text: string) => h('span.sr-only', { 'is:raw': true }, `Section titled ${text}`);
```

delucis (commented):
Made one suggestion for simplicity’s sake, but otherwise LGTM!

PR #321, opened 2022-04-06
Fix syntax highlighting config

jhuleatt:
source: https://docs.astro.build/en/reference/configuration-reference/#markdownsyntaxhighlight


natemoo-re (approved):
LGTM! Thanks for the fix. ❤️

PR #322, opened 2022-04-07
fix encontre to ache

victor0x16:
Hey there,

A tiny change to improve undertanding 🚀 !

### DIFF
--- a/src/pages/pt-BR/getting-started.md
+++ b/src/pages/pt-BR/getting-started.md
@@ -14,7 +14,7 @@ Como alternativa, leia nosso [Guia de Instalação](/install/auto/) para uma gui
 
 ## Aprenda Astro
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Thank you @victor0x16 ! I'm going to bring this to the attention of our volunteer translator group to try to get a consensus!

rafaelcanical:
"encontre" is more used in European portuguese, and "ache" in Brazilian portuguese. I'm a native speaker of European portuguese, I would find it weird and immediately know that it is written in brazilian portuguese, but I would understand it. Since there are way less european portuguese speakers I think the correct decision is to change it to "ache". The minority (european portuguese speakers) will still understand it ;)

sarah11918:
Well, since this is the `pt-br` page, even if there is a difference, I think the Brazilians should have it!  😄

PR #327, opened 2022-04-07
WIP: update migration guide for clarity

natemoo-re:
Taking some feedback from Discord, our Migration Guides are great about noting what has changed but are lacking a bit on the "migration" piece.

This PR is intended to refresh our migration guide and clarify exactly _how_ one can expect to migrate from deprecated APIs to their replacements.

### DIFF
--- a/src/pages/en/migrate.md
+++ b/src/pages/en/migrate.md
@@ -4,9 +4,15 @@ title: Migration Guide
 description: How to migrate your project to latest version of Astro.
 ---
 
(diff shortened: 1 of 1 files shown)


FredKSchott (approved):
LGTM! I think ZERO breaking changes is a bit tough to hit, we've already seen a few bug fixes that technically change existing behavior. I might rewrite this a bit post-merge to soften this, and also give us a section on the page to call out any changes that the reader should be aware of.

PR #328, opened 2022-04-07
Remove unnecessary backslash

MarcusOtter:
I added this in [august 2021](https://github.com/withastro/astro/pull/1148/files#r690872040) because there was [a bug](https://github.com/withastro/astro/issues/1147) that would render `\n` as a literal newline, even inside code blocks. This should no longer be the case, so I'm removing it.

Also, now I get to be in the face pile :)


sarah11918:
GO, TEAM FACEPILE!

sarah11918:
LGTM!

PR #329, opened 2022-04-08
Add max-width to docs

Yan-Thomas:
On bigger screens, the navbar and sidebar are infinitely spaced, personally making it harder to navigate through the docs. This PR limits the body width to 1600px (we can define a better breakpoint if necessary) and centers the page. 

On a 3000-ish wide screen, without max-width we get the result below:
zyCTy2q

Now, with max-width and margin-auto:
qeuUjLL

There really isn't a consent on this, Vue.js docs has max-width, React.js docs does not, Bootstrap docs unconstrain the main content and adds a margin to the sidebar so it doesn't go all the way to the right.

Feedback if this should be merged or not is much appreciated!

### DIFF
--- a/src/layouts/MainLayout.astro
+++ b/src/layouts/MainLayout.astro
@@ -25,6 +25,8 @@ const lang = getLanguageFromURL(url.pathname);
 		<style>
 			body {
 				width: 100%;
+				max-width: 1600px;
+				margin: auto;
 				display: grid;
 				grid-template-rows: var(--theme-navbar-height) 1fr;
 				--gutter: 0.5rem;
--- a/src/layouts/MainLayout.astro
+++ b/src/layouts/MainLayout.astro
@@ -25,8 +25,9 @@ const lang = getLanguageFromURL(url.pathname);
 		<style>
 			body {
 				width: 100%;
-				max-width: 1600px;
-				margin: auto;
+				max-width: 100em;
+				margin-left: auto;
+				margin-right: auto;
 				display: grid;
 				grid-template-rows: var(--theme-navbar-height) 1fr;
 				--gutter: 0.5rem;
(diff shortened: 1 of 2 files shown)


sarah11918:
Thanks @Yan-Thomas ! This is helpful and great feedback, and I hope others will help figure out the best thing to do.  😄 

@aFuzzyBear @delucis

delucis (commented):
Checked this and overall this is OK I think. I might go with a slightly higher `max-width` (110em for example), but that’s a matter of taste.

Could I suggest setting specific left & right margins so this doesn’t impact anywhere someone might want to set the top/bottom margin?

```css
margin-left: auto;
margin-right: auto;
```

Yan-Thomas:
I agree with you. I'm going to change the margin property and instead of using 1600px, I will change that for 100em. I had totally forgotten some people use bigger font-sizes and using the `em` unit on breakpoints does make the page less "squished" according to my tests, thanks!

About the breakpoint size itself, I'm going to leave this way for now and see if most people prefer 110em, 100em or another size.

delucis:
Sounds good to me 👍

FredKSchott:
I'd like to keep it the way it is! One objective reason to keep it that I don't see listed here is that it lets you do stuff like this:

Screen Shot 2022-04-11 at 4 08 35 PM


So, if you wanted to embed side-by-side content like a diff or code editor, that could take up even more of the screen on large screens. You couldn't really do that if the sidebars are hugging the content, without it looking odd if no large-width content was on the page.

sarah11918:
I'm closing this, based on Fred's stated preference for not changing it.  :)

As we spend more time on the look/function of the site with the updates to internationalization, we will try to keep screen sizes in mind for future choices! And if it becomes a problem, do bring it up again!

PR #330, opened 2022-04-08
Update Japanese translation of Getting Started page

kyosuke:
I have matched the current getting started English page.

### DIFF
--- a/src/pages/ja/getting-started.md
+++ b/src/pages/ja/getting-started.md
@@ -1,75 +1,106 @@
 ---
+setup: |
+    import Button from '../../components/Button.astro'
+    import ContributorList from '../../components/ContributorList.astro'
 layout: ~/layouts/MainLayout.astro
 title: はじめに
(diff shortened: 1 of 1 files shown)


sarah11918:
Thank you, @kyosuke ! There are some Netlify deploy issues right now, so we might have to wait to merge this until I figure out what's going on there. But you can (or, if not, I will) post in the i18n docs channel to let people know that this PR can be reviewed by other Japanese speakers!

kyosuke:
@sarah11918 Thank you for your reply! I will wait for the fix. After that, I will post a review request in the i18n docs channel.

sarah11918:
Hi @kyosuke ! I had to update a merge conflict since we are now using `npm create astro@latest` instead of `npm init astro`. But, I'm going to merge this now, so that we have it as a new page to start from. And, we'll start fresh with all the i18n stuff with our new plan as soon as we figure it out!  Thank you again for this!  :)

PR #332, opened 2022-04-08
Added pt-BR translation in `translations.ts`

caioferrarezi:
Inspired by #331, I also made my brazilian portuguese contribution and fixed a little typo in the text :D


kaentos:
Hello, first of all I wanna say I'm an European Portuguese, so some translations might sound better in PT-PT than PT-BR, but here is my feedback about your changes:

'rightSidebar.onThisPage': 'Nessa página', -> 'Esta página', or 'Nesta página',
'rightSidebar.editPage': 'Edite essa página', -> 'Edite esta página',
'rightSidebar.translatePage': 'Traduza essa página', -> 'Traduza esta página',

Someone should still check if my recommendation of changes are better.

caioferrarezi:
Hi, @kaentos! 

Actually, I think you're absolutely right. This doesn't sound bad for us but just a little bit more "natural" instead. However since this could be read by a Portuguese person like you, it seems to fit better in the documentation. 

Thanks for the feedback!

Yan-Thomas:
Looks great for me. I also prefer using "Nesta/Esta" instead of "Nessa/Esta" since it is the right grammar in this case.

sarah11918:
Thanks everyone! It's great that you all came to the same conclusion! 😅 

Just a quick note that there's something going on with Netlify deploys, so until I know exactly what the situation is there, this might be delayed. But, it sounds like the translation looks good to everyone, so I'll assume this will be fine to merge, and thank you all for getting it right!

PR #333, opened 2022-04-08
stop auto-directing based on language, always show english by default

FredKSchott:
Based on a convo in Discord. See the "files changed" tab for context and reasoning, left as a comment in the codebase.


FredKSchott:
confirmed working!

PR #336, opened 2022-04-09
Change `global` to `is:global` for style tags

markteekman:
As per migration to v0.26's New Markdown API (https://docs.astro.build/en/migrate/#new-markdown-api) 🙂


zadeviggers:
This is already done in #282, but that's on hold until @FredKSchott can do a final edit, so it might be a good idea to merge this anyway so that the information is there?

Merging this would, however, mean that more edits are required in #282, which isn't  necessarily a good thing.

FredKSchott:
TY! Lets merge this one since its so small, taking a look at #282 later today

PR #337, opened 2022-04-09
Minimal sidebars in i18n pages

sarah11918:
This comments out all but the Getting Started page in all non-/en/ pages. Practically none, if any at all, of these pages are really up-to-date after so many English doc content and structural changes.

This will let us start with the plan discussed in the docs channel Friday night: to slowly add in translated pages, and eventually on a language-by-language basis be able to re-enable automatic language redirect.


sarah11918:
Closing this, since I don't think it's the direction we'll end up taking!

PR #347, opened 2022-04-12
Add PostCSS usage notes to page

aFuzzyBear:
A couple of users have asked recently about advice on setting and consuming postcss within their projects.

Since we can use `@container` and `@layer` queries, for backwards support users would need to install `postcss-preset-env` to their projects.

This short addition to the Styling Guide should suffice.

### DIFF
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -170,26 +170,14 @@ Use `<style lang="less">` in `.astro` files.
 Astro comes with PostCSS included as part of [Vite](https://vitejs.dev/guide/features.html#postcss). To configure PostCSS for your project, create a `postcss.config.js` file in the project root. You can import plugins using `require()`.
 
 ```js
-//./postcss.config.js
-// Import a PostCSS plugin (https://www.postcss.parts/)
-const postcssPresetEnv = require('postcss-preset-env');
+// ./postcss.config.js
 
-module.export={
-  // Plugins as an Array
-  plugins:[
-   postcssPresetEnv({
-     //Plugin options
-   }),
-  autoprefixer({
-    ...options
-  }),
+module.exports = {
+  plugins: [
+    require('postcss-preset-env'),
+    require('autoprefixer'),
   ],
-  // Plugins as an object
-  plugins:{
-    postcssPresetEnv:{...options},
-    autoprefixer:{...options},
-  }
-}
+};
 ```
 
 
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -125,7 +125,7 @@ Because this approach uses the `public/` directory, it skips the normal CSS proc
 
 ## CSS Integrations
 
-Astro comes with support for adding popular CSS libraries, tools and frameworks to your project like PostCSS, Tailwind and more! 
+Astro comes with support for adding popular CSS libraries, tools and frameworks to your project like Tailwind and more! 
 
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/styling.md
```diff
@@ -161,7 +165,32 @@ Use `<style lang="less">` in `.astro` files.
 - **Vue**: `<style lang="scss">`
 - **Svelte**: `<style lang="scss">`
 
-Additionally, PostCSS is supported as an [integration](/en/guides/integrations-guide/).
+## PostCSS
+
+Astro comes with PostCSS included as part of [Vite](https://vitejs.dev/guide/features.html#postcss). To configure PostCSS for your project. Make a `postcss.config.js` in the root of the project directory. Import plugins from [npm](https://www.npmjs.com/search?q=postcss). Using `require()` we apply additional plugins to the configuration file.
```
delucis, line 170: Some grammar, some opinion, feel free to pick n mix 😉 

```suggestion
Astro comes with PostCSS included as part of [Vite](https://vitejs.dev/guide/features.html#postcss). To configure PostCSS for your project, create a `postcss.config.js` file in the project root. You can import plugins using `require()`.
```

src/pages/en/guides/styling.md
```diff
@@ -161,7 +165,32 @@ Use `<style lang="less">` in `.astro` files.
...
+  autoprefixer({
+    ...options
+  }),
+  ],
+  // Plugins as an object
+  plugins:{
+    postcssPresetEnv:{...options},
+    autoprefixer:{...options},
+  }
+}
```
delucis, line 192: What do you think about trimming this example down a bit for clarity?

Maybe something like:

```suggestion
// ./postcss.config.js

module.exports = {
  plugins: [
    require('postcss-preset-env'),
    require('autoprefixer'),
  ],
};
```

I’m not sure we need to show every possible configuration style, just something to get folks started.

src/pages/en/guides/styling.md
```diff
@@ -161,7 +165,32 @@ Use `<style lang="less">` in `.astro` files.
...
+```
 
 
 ---
```
delucis, line 196: I find the horizontal rule after the code block looks a bit off, so maybe remove this too? But this is just me nitpicking.

```suggestion
```

image

delucis (commented):
Made a few comments on detail, but this looks like a good addition to me @aFuzzyBear!

sarah11918:
Just checking, is the intention here about installing *any* PostCSS plugin, or about `postcss-preset-env` in particular in order to support those queries?

And, this example is showing two different ways of installing the same plugins. Is there any reason why, for Astro, we might recommend one over the other, or do both work equally well? Is one way more standard than the other? Is one way going to work better in an Astro project?

src/pages/en/guides/styling.md
sarah11918: Shorter is often good for cognitive load! Do we feel the need to show what it looks like with options?

src/pages/en/guides/styling.md
sarah11918: We typically don't use these in the rest of the docs, so I'd probably remove for consistency. The code sample gives a nice separation itself, I think.

src/pages/en/guides/styling.md
delucis: I don’t think either of these plugins have commonly used options. PostCSS’s own README literally doesn’t show how to use options 😆

PR #349, opened 2022-04-14
Added notes for pnpm users when deploying to Netlify

RafidMuhymin:
Netlify requires additional settings if using `pnpm`. This PR adds notes for `pnpm` users when deploying to Netlify.

### DIFF
--- a/src/pages/en/guides/deploy.md
+++ b/src/pages/en/guides/deploy.md
@@ -213,6 +213,16 @@ Create a new `netlify.toml` file at the top level of your project repository wit
   publish = "dist"
 ```
 
+If you are using `pnpm`, then set the following settings instead:
+
+```toml
+[build.environment]
+  NPM_FLAGS = "--version" # prevent Netlify npm install
+[build]
+  command = 'npx pnpm i --store=node_modules/.pnpm-store && npm run build'
+  publish = 'dist'
+```
+
 Push the new `netlify.toml` file up to your hosted git repository. Then, set up a new project on [Netlify](https://netlify.com) for your git repository. Netlify will read this file and automatically configure your deployment.
 
 ### Netlify Website UI
--- a/src/pages/en/guides/deploy.md
+++ b/src/pages/en/guides/deploy.md
@@ -213,7 +213,7 @@ Create a new `netlify.toml` file at the top level of your project repository wit
   publish = "dist"
 ```
 
-If you are using `pnpm`, then set the following settings instead:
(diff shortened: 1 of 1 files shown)


delucis (commented):
Looks good @RafidMuhymin. How about including that link you shared in Discord to the Netlify pnpm guide?

johnnytest4real:
[Credit to this person](https://community.cloudflare.com/t/add-pnpm-to-pre-installed-cloudflare-pages-tools/288514/5) for linking the fix, which actually came from [Vite's Netlify config](https://github.com/vitejs/vite/blob/155fd11b783eddd33f0cd7e411eea40a3585217a/netlify.toml).

Relatedly, I'm wondering if there's a similar config that can be created for other hosts that don't fully support pnpm (e.g., Cloudflare Pages).

src/pages/en/guides/deploy.md
```diff
@@ -213,6 +213,16 @@ Create a new `netlify.toml` file at the top level of your project repository wit
   publish = "dist"
 ```
 
+If you are using `pnpm`, then set the following settings instead:
```
delucis, line 216:
```suggestion
Using `pnpm`? [Netlify needs some extra set-up to support `pnpm`.](https://answers.netlify.com/t/using-pnpm-and-pnpm-workspaces/2759) We suggest starting with the following settings in your `netlify.toml`:
```

delucis (commented):
Just to follow up my previous comment, I think adding a link like this could be helpful to let people dive deeper if they need to.

src/pages/en/guides/deploy.md
sarah11918:
```suggestion
Using [`pnpm` with Netlify?](https://answers.netlify.com/t/using-pnpm-and-pnpm-workspaces/2759) Use the following settings instead:
```

delucis: ✂️💨 Scissors to the rescue! LGTM.

sarah11918:
The change above is also good for SEO. What do you y'all think?

Othewise, @RafidMuhymin LGTM! Merge when you're happy with it!

src/pages/en/guides/deploy.md
delucis: Ah, good question. I think I’d say “on”?

sarah11918: Updated in my suggestion, thanks! @RafidMuhymin can make the final call.  :)

PR #354, opened 2022-04-19
Update Non HTML Pages Documentation

RafidMuhymin:
Updated documentation on **File Routes** (**Non HTML Pages Routes**) with more description and code example.


delucis (commented):
Given this is bringing over content from the announcement post, how about including something like this paragraph from there:

> To create a file route inside of your Astro project, create a new JavaScript or TypeScript file inside of your src/pages directory. File routes leverage Astro’s existing file-based router, so be sure to include the final built file extension in the filename.

That emphasises that these routes are inside `src/pages` in case that’s not clear.

Otherwise +1 from me on including this example here over linking to the announcement.

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -71,14 +71,28 @@ This is my page, written in **Markdown.**
 
 > ⚠️ This feature is currently only supported with the `--experimental-static-build` CLI flag. This feature may be refined over the next few weeks/months as SSR support is finalized.
 
-Non-HTML pages, like `.json` or `.xml`, can be built from `.js` and `.ts`. 
+Non-HTML pages, like `.json` or `.xml`, or even non-text assets like images can be built using **File Routes**. **File Routes** must end with the `.js` or `.ts` extension.
```
sarah11918, line 74:
```suggestion
Non-HTML pages, like `.json` or `.xml`, or even non-text assets like images can be built using **File Routes**. 

**File Routes** must end with the `.js` or `.ts` extension and the source file must exist within the `src/pages/` directory.
```
(Does something like this address @delucis ' comment? Is this clear enough to indicated that it can be in any subdirectory of /pages?)

delucis: Yes, this sounds really good 👍

sarah11918:
@RafidMuhymin see my suggestion above, otherwise LGTM. Feel free to merge.

PR #352, opened 2022-04-17
Update script information in components page

zadeviggers:
When helping out in a support thread, I noticed that these docs are out of date.

PR #356, opened 2022-04-20
Removing references to script hoist in translated docs

tony-sull:
Follow-up to #352, removing examples using `<script hoist>` in translated docs

### DIFF
--- a/src/pages/es/core-concepts/astro-components.md
+++ b/src/pages/es/core-concepts/astro-components.md
@@ -297,7 +297,7 @@ Sin embargo, si deseas que todos tus scripts se extraigan de los componentes y s
 Un **script izado** se ve así:
 
 ```astro
-<script hoist>
+<script>
   // Una secuencia de comandos en línea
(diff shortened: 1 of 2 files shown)


src/pages/es/core-concepts/astro-components.md
```diff
@@ -297,15 +297,15 @@ Sin embargo, si deseas que todos tus scripts se extraigan de los componentes y s
 Un **script izado** se ve así:
 
 ```astro
-<script hoist>
+<script>
   // Una secuencia de comandos en línea
```
delucis, line 301: Same here: looks like the `hoist` was added at some point, but the comment wasn’t updated.

delucis (changes requested):
Looks like we’ll need German & Spanish translators to fix these comments.

tony-sull:
False start!  Looks like it makes more sense to fix this as part of a larger translation update leading up to 1.0 - closing this out for now

PR #357, opened 2022-04-20
Make OG image translatable

delucis:
See https://discord.com/channels/830184174198718474/853350631389265940/966255545390624800 for context

Moves the Open Graph image source and alt text into the translations dictionary to support translating alt text. Moving the `src` here too is primarily to keep these two bits of content together so they get updated together, but in theory it would also support localising the OG image if it included text.


tony-sull (approved):
LGTM!  🚀

PR #360, opened 2022-04-20
Update `npm init astro` instructions to `npm create astro@latest`

bholmesdev:
### What changed
- Replace all uses of `npm init astro` with `npm create astro@latest`
- Add `pnpm create` and `yarn create` instructions where applicable
- Update "npm" and "pnpm" to be all lowercase where applicable

### Motivation

Users were getting cached versions of `create-astro` when running without the `@latest` flag. Using [Vite's docs as a guide](https://vitejs.dev/guide/#trying-vite-online), we're replacing all uses of `npm init astro` with `npm create astro@latest` across the docs.

**Why `create` instead of `init`?** You can run `yarn create` and `pnpm create`, but the same isn't true for "init." This should help with having consistent instructions across package managers.


src/pages/de/installation.md
```diff
@@ -16,17 +16,17 @@ Zu Demonstrationszwecken verwenden wir in den folgenden Beispielen [`npm`](https
...
-npm init astro
+# Mit npm
+npm create astro@latest
 
-# Yarn
+# yarn
 yarn create astro
 
-# Pnpm
-pnpm create astro
```
bholmesdev, line 29: npm and pnpm prefer to be all lowercase. Changed yarn's casing for consistency as well

src/pages/en/guides/deploy.md
```diff
@@ -433,19 +433,19 @@ If you don't have an existing Astro site, you can create one by running:
 
 ```bash
 # Make a new project directory, and navigate directly into it
-$ mkdir my-astro-project && cd $_
```
bholmesdev, line 436: Remove $ for better copy-paste-ability!

Issue #103, opened 2021-12-24
📘 DOC: Lint for broken links in all languages


jasikpark:
Currently we have a `yarn lint:linkcheck:local` command that can check for broken links, but we don't have any way to switch what language you're using without JS, so it only checks the English pages.

I want to write a script using the programmatic version of [blc](https://github.com/stevenvachon/broken-link-checker) to check all translations of the docs for broken links!

hippotastic:
Sure @sarah11918, why not! I'll have a look 🙂

hippotastic:
I've implemented it myself. See my PR above. The output looks like this:

```
/da/getting-started/
  ❌ https://docs.astro.build/quick-start
  ❌ https://docs.astro.build/installation
  ❌ https://docs.astro.build/reference/builtin-components

/de/core-concepts/component-hydration/
  🔗 https://docs.astro.build/de/core-concepts/component-hydration/#mycomponent-clientonly-

/de/core-concepts/routing/
  🔗 https://docs.astro.build/reference/api-reference#getstaticpaths

/de/getting-started/
  ❌ https://docs.astro.build/reference/builtin-components

[...]

*** Found 76 broken links.
```

When a link leads to a 404 page, the target href is prefixed by a ❌.
When the target page exists, but the link contains an URL hash that does not exist on that page, the prefix is 🔗.

hippotastic:
@sarah11918 This issue can be closed now, my PR got merged. 🎉

sarah11918:
YAY! 🥳

PR #361, opened 2022-04-20
Add support for i18n docs to lint:linkcheck script

hippotastic:
This fixes issue #103 by rewriting the broken link check scripts.

Because the previously used `broken-link-checker` package didn't work as expected for multi-language checks and the API docs did not match the actual behavior of the component, I rolled my own link checking logic based on the popular `htmlparser2` library instead.

All link checks are now run locally and are based the output of the `astro build` command. The script parses the `sitemap.xml` in the build output directory, builds an index of all pages contained in the sitemap, checks for broken links (including invalid URL hashes) and outputs them in an easy-to-use format grouped by page.


scripts/lint-linkcheck.mjs
```diff
@@ -0,0 +1,179 @@
...
+		var total = brokenLinks.length;
+
+		if (total > 0) {
+			var lastPage;
+			brokenLinks.forEach(brokenLink => {
+				if (lastPage !== brokenLink.page) {
+					console.log(`\n${brokenLink.page.pathname}`);
+					lastPage = brokenLink.page;
+				}
+				console.log(`  ${brokenLink.isMissingHash ? '🔗' : '❌'} ${brokenLink.href}`);
```
delucis, line 154: When reading the report in my terminal I didn’t understand what these emojis meant. The link icon actually read as “link” to me rather than hash. How about using a `#`?

```suggestion
				console.log(`  ${brokenLink.isMissingHash ? '#️⃣' : '❌'} ${brokenLink.href}`);
```

scripts/lint-linkcheck.mjs
```diff
@@ -0,0 +1,179 @@
...
+			});
+
+			console.log(`\n*** Found ${total} broken ${total === 1 ? 'link' : 'links'}.\n`);
```
delucis, line 157: Further to the above, another help would be distinguishing the count here:

```
    ❌ Found 49 broken page links
    #️⃣ Found 27 broken fragment links
```

That would help associate each emoji with some text explaining what they’re supposed to mean.

delucis (changes requested):
This looks fantastic @hippotastic!

One thing: the docs repo uses `pnpm` so I think you need to remove `package-lock.json` and run `pnpm i` to update `pnpm-lock.yaml`.

Made a couple of suggestions below, but in general looks great!

hippotastic:
Thank you for the detailed review and great feedback! I updated my PR to address your valid suggestions.

Unfortunately, I found that none of the emojis get displayed correctly in Windows PowerShell and the old terminal, so I replaced them with plain text prefixes and added colors instead to allow faster scanning of the results. I also added a breakdown of the broken link types in the summary as you suggested.

delucis (commented):
Love it! The plain text solution is actually way better than emojis.

How would you feel about using [`kleur`](https://www.npmjs.com/package/kleur) instead of `chalk` for the terminal colours? It’s already in the transitive dependencies for the repo and is supposed to be smaller/faster than `chalk` (with a more or less interchangeable API I believe).

hippotastic:
Done. I usually prefer `chalk` because `kleur` is lacking bright colors (they have a [long outstanding PR](https://github.com/lukeed/kleur/pull/26) that would solve this, but don't seem to merge it due to confusion with bold, even though I can see a clear difference between bright and bold in my terminal).

However, as I now know that `kleur` is already being used by Astro, I'm happy to go with that, too.

delucis:
Oh interesting! I'm on Mac and just saw bold rather than bright colours so I guess that explains the potential for confusion.

Anyway this looks great! Awesome work!

hippotastic:
@tony-sull Could you please have a look if this can be merged?

It's been reviewed by Chris already and I was informed to mention you now. Hope that's alright!

tony-sull (approved):
Oh wow, this is excellent!  Big fan of the color coded plaintext output, ran this locally and the output looks excellent and very clear 🥳

Issue #362, opened 2022-04-21
Outdated Documentation | RSS


RafidMuhymin:
The documentation on `RSS` is outdated now, the code example is still using the deprecated `Astro.fetchContent` API https://docs.astro.build/en/guides/rss/

PR #363, opened 2022-04-21
updated RSS to replace fetchContent with glob

sarah11918:
(And, a couple of other pases where fetchContent was in use.)

Should close: https://github.com/withastro/docs/issues/362


aFuzzyBear (approved):
This looks good buddy, 
Safe to merge this 👊

PR #364, opened 2022-04-21
Update astro-components.md to warn usage of named slots

leader22:
Refs https://github.com/withastro/astro/issues/2238

Is this reasonable to add here?

### DIFF
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -237,7 +237,7 @@ const { name } = Astro.props
 
 // src/components/Person.astro
 
-<Wrapper name = "Astro">
+<Wrapper name="Astro">
   <h2>I am a person.</h2>
   <p>Here is some stuff about me.</p>
 </Wrapper>
@@ -268,13 +268,15 @@ const { name } = Astro.props
 
 // src/components/Person.astro
 
-<Wrapper name = "Astro">
+<Wrapper name="Astro">
   <img src="https://my.photo/astro.jpg" slot="after-header">
   <h2>I am a person.</h2>
   <p slot="after-footer">Here is some stuff about me.</p>
 </Wrapper>
-
 ```
+
+> ⚠️ Named slots are available only if the parent component and child component are both Astro components (even though some frameworks like Vue, Svelte support it).
+
 #### Fallback Content for Slots
 Slots can also render **fallback content**. When there are no matching children passed to a `<slot>`, a `<slot>` element will render its own placeholder children.
 
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -245,7 +245,7 @@ const { name } = Astro.props
 
 #### Named Slots
 
-Slots can also be **named**. Rather than a single `<slot>` element which renders _all_ children, named slots allow you to specify multiple places where children should be placed.
+Inside of an Astro component, slots can also be **named**. Rather than a single `<slot>` element which renders _all_ children, named slots allow you to specify multiple places where children should be placed.
 
 ```astro
 // src/components/Wrapper.astro
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -245,7 +245,7 @@ const { name } = Astro.props
 
 #### Named Slots
 
-Slots can also be **named**. Rather than a single `<slot>` element which renders _all_ children, named slots allow you to specify multiple places where children should be placed.
+Slots can also be **named**. Rather than a single `<slot>` element which renders _all_ children, named slots allow you to specify multiple places where (only certain child elements from your Astro component) should be placed.
 
 ```astro
 // src/components/Wrapper.astro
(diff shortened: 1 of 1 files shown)


sarah11918:
@delucis @aFuzzyBear can I get some thoughts on this? 

I know in general we try to keep the Astro pages about Astro stuff. And, at this point in the docs, in this "Astro components" page, we haven't even discussed any non-astro stuff on your site yet at all. So, I'm not *entirely* sure about the call-out here. BUT, I could see adjusting the current text of the page around it:

Named Slots
[Section titled Named Slots](https://docs.astro.build/en/core-concepts/astro-components/#named-slots)
Slots can also be named. Rather than a single <slot> element which renders all children, named slots allow you to specify multiple places where **(only certain child elements from your `.astro` component)** should be placed.

QUESTION: can named slots be used with Markdown children, too, or JUST .astro components?

sarah11918:
@RafidMuhymin Do you have thoughts on this?

RafidMuhymin:
We need to investigate these things. Anyone from the core member will be able to help better.

delucis:
In this context I _might_ just fix the typos. I think I’ve seen scenarios where passing a framework component to a slot actually does work, so it may not be as clear cut — or at the very least we need core to confirm as @RafidMuhymin suggests.

RafidMuhymin:
@delucis I don't think @leader22 is saying that if we pass framework components inside named slots then they were working. I think the problem is that `Svelte`, `Vue` have their own slots like Astro, and when trying to pass content/components via the slots in framework components to the parent Layout which is an Astro component, then those components aren't getting passed properly.

RafidMuhymin:
@leader22 If my point on the comment above is correct, then how did you experiment that it's not working? Because if you have passed the Vue/Svelte component inside the main slot, then it's normal that the slots inside those framework components won't be passed to the Layout component, as it's not possible to use a slot inside another slot.

leader22:
Thank you for your review all. 😉 

My original case was like this code below,

```astro
---
// pages/index.astro
import Layout from "../components/layout.svelte";
---

<Layout>
  <div slot="main">Main</div>
  <div slot="sub">Sub</div>
</Layout>
```

and

```svelte
<script>
// components/layout.svelte
</script>

<div class="layout">
  <div><slot name="main" /></div>
  <div><slot name="sub" /></div>
</div>
```

ends up with rendering nothing(= empty default slot).

To make this case works, `Layout` component should be `.astro`(not `.svelte|.vue`).

So at least for me, @sarah11918 's suggestion looks good.

RafidMuhymin:
@leader22 oh, that makes sense. Thank you for the explanation. Yeah, it can be documented. I think when the compiler is identifying the slot it's trying to pass the slot to a parent Astro component. But as it's not being able to find an Astro component it's not doing anything.

leader22:
Sorry, I should include the context first. 😅 

The remaining problem is how should the documentation be worded. Should we still avoid referring to Framework components?

sarah11918:
@leader22 Would the wording I suggested above be OK? I think I would prefer not to mention other framework components if we can, but instead say very clearly ASTRO COMPONENTS when describing. Do you think that would be clear enough?

leader22:
@sarah11918 

> I think I would prefer not to mention other framework components if we can, but instead say very clearly ASTRO COMPONENTS when describing.

Sure. 

But what we should care about here is **parent** slot (must be an `.astro` component, not `.svelte` not `.vue`), instead of **child**(there is no restriction, any types of component can be passed) content for the slot.

Do you have any ideas to express this nuance?

sarah11918:
Do you mean something like...

Slots can also be named. Rather than a single element which renders all children, named slots in `.astro` components allow you to specify multiple places where ...

leader22:
My two cents... https://github.com/withastro/docs/pull/364/commits/3c0c0a4c4dd1bcfd4416741a00bd7ee664e51d00

Finally I realized that simple is best. 😹

sarah11918:
@leader22 If that wording satisfies you, it satisfies me!  My job is to keep things simple! 😅 

I'll merge this now, and if you have second thoughts later, we can absolutely revisit it with another PR!

Jutanium:
Sorry to revive this, but the current wording tripped me up and I ended up down a whole rabbit hole until I found this.
The current wording says `inside an Astro component` and I _was_ using named slots inside an Astro component - it isn't about where the named slots are _used_ but abut the component you're using it _with_. Explicit is better than simple in this scenario.

sarah11918:
I think the communication issue might be in how to distinguish betweent "Astro Component" ie the `.astro` file in which you're rendering a slot vs "Astro Component" ie the the component being rendered in the HTML template.

When the original author mentioned being satisfied with the wording, I took that to mean: "Seeing "in an Astro coponent" is enough of a cue for me to look at my example and think, "Hey, my Layout component is a Svelte component, so maybe I shouldn't expect this to work like an Astro component." which is maybe the more desirable outcome, vs annotating unusual cases in the docs? (Full disclosure: I'm not sure how common of a pattern to expect this to be.)

Anyway, that explains the motivation here, and I'll move on to the open PR.  :)

leader22:
@Jutanium I feel sorry to cause a misunderstanding. 😓
@sarah11918 Yes, I think so too. Since this page is about Astro component as template and how to define the slot itself. How to use (and combine with other stuff) may be an another concern. (And it was a little bit difficult for me to express without referring framework components).

I'll comment on some thoughts in #371 later.

Issue #359, opened 2022-04-20
New i18n process


delucis:
Opening an issue to keep track of the things we need to get the new i18n process ready. This is partly for myself as I can tackle a bunch of this, but also so others can point out other improvements and requirements or generally give feedback — or even jump in and help out! The more the merrier.

## To-do

- [ ] #1057
- [ ] #1058
- [ ] #1059
- [ ] #1481
- [x] Can we manually “clear” a language from the overview issue if a contributor confirms it is actually up-to-date?
- [x] #1060

    Hippo re: keywords:
    > if anyone wonders what needs to happen for a previously translated page to switch to the "needs updating" state: If changes are committed to the English source page after the translation got committed, the new commits are parsed and if there is a commit that doesn't include any of the terms (minor|typo|broken link|i18nReady), this commit is considered "major" and therefore triggers the state change of the translated versions.

    Hippo re: “clearing”:
    > Similar to the explanation in the script in lines 271-272, you can achieve this by editing the summary issue and setting the respective translation’s UTC date in the state data to the current date.

<details>
<summary>

## Done
</summary>

### Done in #365

- [x] **Write a guide on contributing translations.**

  - [x] Cover the two sources of translations: pages and UI strings.
  - [x] Explain [the new Astro i18n policy posted on Discord](https://discord.com/channels/830184174198718474/867764597838184478/966314002974122024).
  - [x] Instructions on keeping forks updated. App option recommended by @sarah11918: https://github.com/wei/pull

- [x] **Add a fallback content system** to surface English content for languages that aren’t fully translated.

  - [x] Fetch English pages and inject their content where it is missing in another language
  - [x] Add a translatable notice explaining that this page isn’t yet translated and that’s why you’re seeing the English version.
  - [x] Add a link to instructions on how to get involved translating content to the notice.
  - [x] Side quest: this would make all pages reliably “available” across languages. We should update the language switcher select menu to preserve the current page when switching between languages (currently it always links to Getting Started).

- [x] **Sidebars!** (related to above, see [Discord thread](https://discord.com/channels/830184174198718474/966314911078707200)).

  - [x] Include all pages in sidebar even if they’re not translated and will show fallback content (seems to be the consensus on Discord).
  - [x] Indicate in the sidebar that pages are in the fallback language, e.g. with a `[EN]` suffix or something.

[...]

PR #365, opened 2022-04-21
i18n overhaul

delucis:
This PR takes most of the structural steps to get #359 off the ground.

### Summary

It’s a big PR and touches *a lot*, so here’s a summary of what’s going on:

- All non-English pages except for “Getting Started” have been removed. (They’re archived in `old-translations/` for now for reference, but we can delete them at some point in the future.)

- Every language now has its own directory in `src/i18n`. This includes translations for UI strings as well as the navigation sidebar (which used to live in `src/config.ts`).

- Every language now uses the same navigation sidebar structure as English. If translations are found in `src/i18n/{lang}/nav.ts`, these will be used instead of the English label. If a page is not yet available in this language, the page is still generated, but using fallback content from the English version. These are marked with an <sup>EN</sup> suffix in the sidebar.

    For example, here’s [`/de/editor-setup/`](https://deploy-preview-365--astro-docs-2.netlify.app/de/editor-setup/). Its sidebar text is translated — “Editor-Einrichtung” — but the page is not and shows a notice explaining this. This notice can be translated in `src/i18n/{lang}/ui.ts`.

    image

- `src/i18n/README.md` contains a guide on how to contribute translations

- A new script provides automated scaffolding of all of the required files for a new language: `pnpm run add-language`

    https://user-images.githubusercontent.com/357379/164994712-65fc1242-fb26-4e07-9886-5f0ad04fb01f.mov

#### Smaller details

- 404 pages are now translatable in the same way as other UI strings and are generated automatically for each language

- Bare language URLs redirect to Getting Started, e.g. `/de` → `/de/getting-started`

- Following [feedback on Discord](https://discord.com/channels/830184174198718474/867764597838184478/966981090831126559), this system uses consistent slugs across locales

- The language switcher now preserves the current page when the language changes instead of always redirecting to “Getting Started”

- Had to upgrade to the latest Astro beta to get some bug fixes for the markdown `Content` component.

- SEO: added `hreflang` alternate links to connect up translations across languages and made sure that pages displaying fallback content use the English URL as their canonical.

### Key areas for review

- Evaluate `src/pages/[lang]/[...fallback].astro`. Is this approach OK for rendering the fallback pages? Any improvements possible?

- I ran into a few performance hurdles, see [`adb545d`](https://github.com/withastro/docs/pull/365/commits/adb545da17a43e0c6f8626f8676696a7d9eba961) for one detailed explanation.

- What is still needed in `src/i18n/README.md`? We need a list of supported languages obviously, but otherwise?

- Should this PR also prune the number of supported languages or will we do that later? Or, in the other direction, should this PR preserve potentially out-of-date translations for now instead of moving them to `old-translations/`?

- How does the translation file structure in `src/i18n` feel? German (de) is a good language to check as it has bits of everything translated.

### Constraints introduced

- The fallback content system relies on `Astro.glob`, so all pages need to be `.md` files rather than `.astro`. _Might_ be possible once the metadata returned by `Astro.glob` for Astro files is improved to expand that, but for now we’ll need to stick to Markdown.

- Because slugs are shared across locales, Markdown files need to have the same name across the different language directories and these need to match the slugs in `src/i18n/en/nav.ts`. I haven't added any tooling around that, but we could.


src/pages/[lang]/[...fallback].astro
```diff
@@ -0,0 +1,30 @@
...
+			const doesNotNeedFallback = pagesByLang[lang]?.some(p => p.url.endsWith(slug));
+			if (doesNotNeedFallback) continue;
+			paths.push({ params: { lang, fallback: slug } });
+		}
+	}
+	return paths;
+}
+
+const { fallback } = Astro.params
+const englishPages = await Astro.glob('../en/**/*.md');
```
Comment (FredKSchott) [inline src/pages/[lang]/[...fallback].astro:27]: FYI you should be able to do this:

```
const { fallback } = Astro.params
const foo = await import(`../en/${fallback}.md`);
```

however, it looks like its not returning the right thing due to a Vite bug, so `Astro.glob()` is fine until that gets resolved.

src/pages/[lang]/[...fallback].astro
```diff
@@ -0,0 +1,30 @@
...
+			if (lang === 'en') continue;
+			const doesNotNeedFallback = pagesByLang[lang]?.some(p => p.url.endsWith(slug));
+			if (doesNotNeedFallback) continue;
+			paths.push({ params: { lang, fallback: slug } });
+		}
+	}
+	return paths;
+}
+
+const { fallback } = Astro.params
```
Comment (FredKSchott) [inline src/pages/[lang]/[...fallback].astro:26]: `fallback` is a bit of a misleading name here, what about `slug` instead?

FredKSchott (approved):
@delucis this looks great, thank you for tackling something so huge in scope with such an attention to detail! Those two things are usually not easy to do together :)

I've given it a look and I don't have too many comments! Given the scope I'd prefer to merge this sooner rather than later to prevent merge conflicts. 

I want to take a look at some of the perf issues you've run into, and maybe suggest a couple of follow up PRs. Sadly, some of them may be in Vite instead of Astro, and how they eagerly scan/build files in a project.

FredKSchott:
Suggested follow-ups, based on the convo above:
- a lint script that warns if a language file (ex: `/de/foobar.md`) doesn't have a matching English file (ex: `/en/foobar.md`). We could run this in CI to make sure all PRs are passing.
- some perf improvements (@FredKSchott to PR)
- decide on final official languages to support (@FredKSchott to post in Discord)

RE project layout: At this level of complexity, I would suggest that we begin to look into moving all markdown content out of `src/pages` and into a separate `src/docs` or `src/content` directory, and then converting all `src/pages/` to be just the page "shells" that are responsible for importing that content using `Astro.glob()`. But I'm not sure if it sounds like we have perf issues lingering that might make this hard, based on the above. Maybe more investigation is needed.

PR #367, opened 2022-04-22
Fix external link to `astro.new`

mayank99:
Without `https` it evaluates to `https://docs.astro.build/en/guides/typescript/astro.new`.


sarah11918:
We have to stop meeting like this...

PR #370, opened 2022-04-23
API route clarification

Jutanium:
- Specify what an API route is
- Add the example filename to match the previous example
- Add `async` in the function declaration so that it's a working example

### DIFF
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -119,7 +119,7 @@ if(!product) {
 
 #### API Routes
 
-An API route is a .js or .ts file in the src/pages/ folder that takes a request and returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response).
+An API route is a `.js` or `.ts` file within the `src/pages/` folder that takes a [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) and returns a `[Response]`(https://developer.mozilla.org/en-US/docs/Web/API/Response).
 
 __[id].js__
 ```js
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -119,12 +119,13 @@ if(!product) {
 
 #### API Routes
 
-A [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) can also be returned from an API route.
+An API route is a .js or .ts file in the src/pages/ folder that takes a request and returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response).
 
+__[id].js__
 ```js
 import { getProduct } from '../db';
 
-export function get({ id }) {
+export async function get({ id }) {
   const product = await getProduct(id);
 
   if(!product) {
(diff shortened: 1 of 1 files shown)


sarah11918:
@matthewp Can you please verify that this change is good by you? If you're happy with it, feel free to merge.

src/pages/en/guides/server-side-rendering.md
```diff
@@ -119,12 +119,13 @@ if(!product) {
 
 #### API Routes
 
-A [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) can also be returned from an API route.
+An API route is a .js or .ts file in the src/pages/ folder that takes a request and returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response).
```
aFuzzyBear, line 122:
```suggestion
An API route is either a `.js` or `.ts` file within the `src/pages/` folder which  takes a [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) and returns a `[Response]`(https://developer.mozilla.org/en-US/docs/Web/API/Response).
```

aFuzzyBear: Just a small formatting nit

aFuzzyBear (commented):
Aye, lets wait for the big yin to sign off on this =), looks good @Jutanium

src/pages/en/guides/server-side-rendering.md
sarah11918:
```suggestion
An API route is a `.js` or `.ts` file within the `src/pages/` folder that takes a [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) and returns a `[Response]`(https://developer.mozilla.org/en-US/docs/Web/API/Response).
```

Aaand.... this is what I would do! (I don't think "either" needs to be there, and I feel like "that" is more in the voice of these docs than "which.")

Jutanium:
I can't imagine @matthewp would have a problem with this seeing that I took his language from his blog post, and since the code is broken I think we should merge it and change later if there's a problem.

src/pages/en/guides/server-side-rendering.md
aFuzzyBear: English is a romantic language

“Speak the speech, I pray you, as I pronounc'd it to you, trippingly on the tongue.”

sarah11918:
OK, then I'm commiting the code formatting and extra link from Fuzzy, and I'll take the blame for this one...

src/pages/en/guides/server-side-rendering.md
```diff
@@ -119,12 +119,13 @@ if(!product) {
 
 #### API Routes
 
-A [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) can also be returned from an API route.
+An API route is a `.js` or `.ts` file within the `src/pages/` folder that takes a [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) and returns a `[Response]`(https://developer.mozilla.org/en-US/docs/Web/API/Response).
```
sarah11918, line 122:
```suggestion
An API route is a `.js` or `.ts` file within the `src/pages/` folder that takes a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) and returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response).
```

PR #371, opened 2022-04-24
Clarify named slots behavior

Jutanium:
The existing wording was ambiguous; direct + explicit is best in this counterintuitive scenario


sarah11918:
So, forgive me, and I'll loop @leader22 in on this convo too, but I think this change makes more sense in the context of the other situation we described, but not necessarily in the general sense?

Two things that I believe are true but am willing to be corrected on: the **slot's child** can be anything, and a slot doesn't need to have a parent in the html template itself. So, I found talking about parent and child here as a general description to be a little more confusing: I think my mind would jump to the slot's child, not the child of the component rendering the slot (because there might not even be one, so that wouldn't be my first thought.)

Does that make any sense??  :)

As per my comment in the other thread, my main goal is (a. obviously not confusing people! but also b.) not necessarly to document all the use cases. That's why I had suggested "If we scream ASTRO loudly and often enough, while they ae reading this page on Astro components, does that give people enough prompting to think about the fact that they have a Svelte etc. component?"

I was also wondering, if this is a really common use pattern, should it go somewhere in the page about framework components, which is where we're nudging people to read about using React, Svelted etc.? Thoughts?

leader22:
> So, forgive me, and I'll loop @leader22 in on this convo too,

Don't worry, Thank you for your efforts. 😸 

> but I think this change makes more sense in the context of the other situation we described, but not necessarily in the general sense?

Agree, now I feel some verbosity about these changes.

> I was also wondering, if this is a really common use pattern, should it go somewhere in the page about framework components, which is where we're nudging people to read about using React, Svelted etc.? Thoughts?

This is exactly what I've thought today!

My current suggestions are

- Keep this page simple like

```diff
- Inside of an Astro component, slots can also be **named**. Rather than a single `<slot>` element which renders _all_ children, named slots allow you to specify multiple places where children should be placed.
+ Slots can also be **named**. Rather than a single `<slot>` element which renders _all_ children, named slots allow you to specify multiple places where children should be placed in between Astro components.
```

- And update https://github.com/withastro/docs/blob/main/src/pages/en/core-concepts/framework-components.md#nesting-framework-components
  - I gave it a try: #373
  - (BTW I think this should be my first PR...!)

How do you think?

Jutanium:
@sarah11918 I agree with you that my wording of `parent` vs `child` might be confusing, but I disagree that the warning should go in the frameworks component page and _not_ here: after all, this is the only place in the docs that is focused on Slots.

Jutanium:
(Also to clarify what I did mean about `parent` and `child` - in the context of slots, the child is the component that can take slots, and the parent is the page/component that includes the child and passes slots. But I agree that we can choose simpler language that doesn't assume this terminology knowledge!)

src/pages/en/core-concepts/astro-components.md
```diff
@@ -275,6 +275,8 @@ const { name } = Astro.props
 </Wrapper>
 ```
 
+This is a feature of Astro components - it can't be used when including a framework component inside a `.astro` file, even if that framework supports named slots.
```
aFuzzyBear, line 278: This bit is true, but the way its written it infers the wrong meaning, Astro uses the html `<template>` `<slot>` elements to pass through children into components. Its just that slots when in use in UI-land tend to be abstractions over this, hence the incompatibility between ui frameworks using `props.children`  and Astro using `<slots/>`

aFuzzyBear:
I would also like to see the new syntax for the slots api, where by we can pass through a callback function into the slot to render dynamic slots, I was taking a look around the github, and couldn't see the finalised api of this albeit this has been merged into Astro since 0.26, I'll ping @natemoo-re  to see if he could provide us with an example for this to include into the documentation,

sarah11918:
With fresh morning eyes, can we solve this whole thing via some combination of:
- when introducing / referring to slots, sprinkling in every now and then something that makes it explicit... e.g "Slots in Astro..."
 - a call-out like: "Using Astro slots with other [UI framework components](link) ? " and then explain this slot behaviour over there? (specifics to be hashed out in the other PR)

@Jutanium @leader22

Jutanium:
Yep, as long as it's a clear and explicit disclaimer on the UI Framework Components page, I'm happy. #373 is a good example of that

sarah11918:
@Jutanium OK, assume #373 will be merged! 

Given that, are you happy to be the one to make those changes here? Make a call out in our style of " > emoji " and just throw some more Astros on the page?  :smile:

aFuzzyBear:
Could we hold of on the push for this, I would like to see the inclusion of the callback api within the slots being documented, just waiting for a definitive example from Nate,

sarah11918:
Hey @aFuzzyBear , is this related to `Astro.slots`? Is there more to document than what's here in reference? https://docs.astro.build/en/reference/api-reference/#astroslots

aFuzzyBear:
fml ... that's what I was looking for !! I'm out

sarah11918:
@Jutanium & I will finish this off tomorrow (Wednesday)!

sarah11918:
@Jutanium Did we decide that this PR can be closed, because we're going to tackle the wording for slots in the other PR? (And, we've added something to Framework Components already?)

Jutanium:
I lost track of the other PR, but it's fine if this one's closed!

Issue #374, opened 2022-04-24
Typo: "directly" to "directory"


a-trost:
One line in the [Static Assets page](https://docs.astro.build/en/guides/imports/#supported-file-types) says

> Remember that you can also place any static asset in the [`public/` directory](/en/core-concepts/project-structure/#public) of your project, and Astro will copy the _directly_ into your final build.
(emphasis mine)

I assume this is supposed to say "copy the _directory_ into your final build."

Opening a pull request with this change.

sarah11918:
Thanks for this! I'm closing the issue because I'm handling the pr now.  :)

sarah11918:
Thanks for this! It has been fixed by another pr so I'm closing this one. (The intended original wording was "copy them directly" by the way!)

PR #375, opened 2022-04-24
Fix Typo (#374)

a-trost:


sarah11918:
Thanks for this! This has been fixed by another PR, so I'm closing this one. (The original intended wording was "copies them directly" by the way. But this would have worked, too!)

PR #376, opened 2022-04-24
Fix broken links

delucis:
Used the new link checker to do an audit for broken links and fixed them. Scoped this to the `/en/` pages given the likely future upheaval that awaits other languages.

I did also remove one link completely because I don’t think anything like what it linked to exists any more.

#### Before

```
*** Found 69 broken links in total:
  [404] 47 broken page links
  [ # ] 22 broken fragment links
```

#### After

```
*** Found 51 broken links in total:
  [404] 35 broken page links
  [ # ] 16 broken fragment links
```

PR #379, opened 2022-04-25
Fix 404 broken links

Rolanddoda:
Links for `<Code />` and `<Prism />` lead to 404


Rolanddoda:
Ahh, I actually saw that those links are fixed by #376 so I am closing this one.

PR #381, opened 2022-04-26
Improve docs rtl language support and acessibility

Yan-Thomas:
Right to left languages have a problem with some components styling, this PR fixes the problems I have encountered:
* "on this page" component border direction
* sponsors component direction 
* adding dir and lang attributes in HTML for sponsors and logo components

image

The expected behaviour would be something like what React.js docs does:

image

Fortunately, [CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) help us fix this!

image

As you can see, I added explicit lang and dir attributes to sponsors and logo components, the lang attribute makes clear for screen readers it is in English, and not the person's specific language ([See Language of Parts on W3 website](https://www.w3.org/WAI/WCAG21/Understanding/language-of-parts.html)). Sure, I could have used `direction: ltr` in CSS, but that wouldn't work if somehow CSS didn't load properly.

For those concerned about logical properties support, [here is the compatibility table on Can I Use](https://caniuse.com/css-logical-props). This same CSS file uses both `:is` and `:not` who have similar compatibility percentages, as you can [see here](https://caniuse.com/css-matches-pseudo), so, I'm guessing that compatibility isn't a real issue in this case.

Probably there are more problems like this in docs, please feel free to warn me about them!

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -104,7 +104,7 @@ h5 {
 	--icon-size: 1rem;
 }
 
-.heading-wrapper:not(:first-child) {	
+.heading-wrapper:not(:first-child) {
 	margin-block: 0;
 }
 .heading-wrapper:not(:first-child):is(.level-h2, .level-h3) {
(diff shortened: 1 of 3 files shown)


X7md:
lgtm 🚀

delucis (commented):
LGTM! Nice work!

sarah11918:
Excellent! Thanks, @Yan-Thomas !

Issue #298, opened 2022-04-04
[i18n] Support for external link in another language


Chrissdroid:
Adding a component to specify what language the link is referring to

like so :

```md
see docs for the <External-link lang="fr" href="https://developer.mozilla.org/fr/docs/Web/API/Request">Request</External-link> api in french
```

which is transformed in something like so :

see docs for the [**Request**⁽ᴱᴺ⁾](https://developer.mozilla.org/fr/docs/Web/API/Request) api in french

Yan-Thomas:
Maybe we don't need a custom component for this. I just need to be sure if there is any downsides using `<a>` tag instead of Markdown link syntax, but with CSS pseudo-elements I can make exactly what you want with a few lines of css:

```
a::after {
	display: inline-block;
	content: attr(hreflang);
	font-size: 0.7rem;
	position: relative;
	bottom: 0.2rem;
}
```

What happens is that when an `a` tag has the `hreflang` attribute (this attribute specifies the linked document language), the `content` attribute will get the value (hreflang is case-insensitive, so as far as I know we have to write it in uppercase to be like that in the markup but that doesn't affect screen readers/SEO).

Final result:

![image](https://user-images.githubusercontent.com/61414485/165198874-c9b37e30-79fd-48a1-b480-e54bdec75890.png)

delucis:
@Yan-Thomas 

You can even get this with Markdown links to avoid complicating authoring:

```css
a[href^="en"]::after {
  content: "en";
}
```

Requires a CSS declaration for each language but we could probably auto-generate it once #365 is merged as that adds `src/i18n/languages.ts` with a list of all the languages.

Whether the content is upper or lowercase is not so important because we could style that with CSS too:

```css
a::after {
  text-transform: uppercase;
  vertical-align: super;
}
```

Yan-Thomas:
Thanks @delucis, clever solution! Also, I had totally forgotten about `vertical-align` and `text-transform` 🤣

I did some tests based on your approach using `<span>` and it really is better acessibility-wise (NVDA on Windows). I also totally forgot that CSS `content` isn't readable for screen reader users, I was really tired yesterday 😓 
 
The `hreflang` attribute AFAIK is very specific (only used by Google SEO) and would just be a convenient way to express that the link is in a different language. If we can make this work without it, then be it!

sarah11918:
@Yan-Thomas and everyone, do you think we can close the existing PR then, and continue to discuss here what solution we want for this?

[...]

PR #382, opened 2022-04-26
Add link language indicator

Yan-Thomas:
This PR is a solution for #298

Links for pages with different languages, indicated by `hreflang` attribute receive an language indicator based on the value attributed to `hreflang`. 

The only downside I know is that you have to use `<a>` tags instead of Markdown syntax, but since this is only for different language links, maybe this isn't much of a problem.

Any link with `hreflang` gets this indicator, see this example:

image

HTML inside Markdown:
`🏗️ Add new <a href="/en/core-concepts/astro-pages" hreflang="en">Astro (.astro) pages</a>`

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -190,6 +190,15 @@ a {
 	gap: 0.5rem;
 }
 
+a::after {
+	display: inline-block;
+	content: attr(hreflang);
+	font-size: 0.7rem;
+	text-transform: uppercase;
+	position: relative;
+	bottom: 0.2rem;
+}
+
 article > section :is(ul, ol) > * + * {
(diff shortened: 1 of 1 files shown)


sarah11918:
Just so there's a comment here, too: the discussion in the original (linked) issue #298 and in the Discord has us all not sure whether the low-tech solution is actually easiest for translators, and almost the same result!

e.g. `[routing (en)](/en/reference/routing)` --> a link that reads `routing (en)`

Also suggested is using a component.

So, I dont think we have consensus yet on how to handle (if at all) mentioning that links are going off to an English site.

PR #384, opened 2022-04-26
Fix initial manual page experience

marbiano:
Because of `---` being present on the default frontmatter comment, I had a confusing initial experience while trying to copy and paste the content for my first page, in VSCode with the Astro plugin installed. I spent like 20 minutes digging around until I could understand what was going on. This PR fixes that initial experience by providing a valid synonym that doesn't include the triple dashes. More info about the issue in here: https://github.com/withastro/language-tools/issues/248

### DIFF
--- a/src/pages/en/install/manual.md
+++ b/src/pages/en/install/manual.md
@@ -51,7 +51,7 @@ For this guide, copy-and-paste the following code snippet (including `---` dashe
 
 ```astro
 ---
-// Welcome to Astro! Everything between these "---" code fences
+// Welcome to Astro! Everything between triple-dashed lines
 // is your "component front matter". It never runs in the browser.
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @marbiano , and thanks for this! We do use the phrase "code fence" so I think I'd prefer to keep that in the documentation.

Do you think it would be clear enough to say , "everything between these code fences" and just NOT mention the dashes? Or what about "everything between these triple-dash code fences" ? (Of course, we can revert to typing the dashes out again at some point if the editor eventually handles this as intended.) What are your thoughts?

sarah11918:
Then it's good to go! Thanks again @marbiano , and welcome to Team Docs as a first-time contributor! 🥳

PR #385, opened 2022-04-26
remove unneccesary astro glob

FredKSchott:
Follow up to #365, I don't know if this actually helped our perf in dev, but it can't hurt :) Passing `props` from `getStaticPaths` is the official way to pass data down to the page so that you don't need to re-request it a second time.

There are two more perf improvements that I'd like to look into, separate from this, but that we appear to be blocked on:
1. Removing `[slug]` and `[...slug]`: Tony is planning to look into moving `docs.astro.build -> astro.build/docs` next week, which we should combine with a big "finalize our slugs and remove all of our old redirects".
2. Vite seems to be building all of our markdown files 3 times, during dev. No idea why this is, but it will need investigation!

cc @delucis


FredKSchott:
Hmm, this is a weird bug:

```
 error   Unable to render SidebarToggle!
  
  This component likely uses @astrojs/react or @astrojs/preact,
  but Astro encountered an error during server-side rendering.
```

@delucis possibly related to this? https://github.com/withastro/astro/issues/2480

delucis:
I actually tried this approach initially! But there appears to be a bug with framework components when you pass the content this way. (Errors on build but appears to work in dev, IIRC.)

My wild guess was: Astro uses imports to know to load various assets. When the "import" (via `Astro.glob`) is inside `getStaticPaths` some of these don't get picked up. I noticed it with this error but also certain (component-specific, I think) styles.

Obviously would be good to fix this, but FWIW I don't think I actually saw a huge perf difference between the two approaches. Maybe the glob is cached or something. The bigger issue was https://github.com/withastro/docs/pull/365/commits/adb545da17a43e0c6f8626f8676696a7d9eba961

delucis:
> @delucis possibly related to this? [withastro/astro#2480](https://github.com/withastro/astro/issues/2480)

Hmm, the errors I was getting on that issue were quite different although I don't want to rule it out that they share something.

FredKSchott:
> The bigger issue was [adb545d](https://github.com/withastro/docs/commit/adb545da17a43e0c6f8626f8676696a7d9eba961)

Agreed, and to make matters worse I think that's being compounded by the same "why is Vite building every markdown file 3 times" issue 🤦

delucis:
As this is still failing, I’ve marked it as a draft to help us triage the PRs tab. I’d love for this to work, so keeping this around for the glorious future in which the bug is fixed.

PR #386, opened 2022-04-27
Capturing Migration guide content

sarah11918:
Small bits of info that didn't make it from the Migration guide to the docs themselves and/or minor wording changes to reflect updates. e.g.
- Removing a snowpack reference
- Component Imports on top
- renderers -> integrations
- removing an old component-hydration file (and setting a netlify redirect) because its contents had been split up and reworked into partial-hydration and framework-components


src/pages/en/reference/cli-reference.md
```diff
@@ -54,10 +54,10 @@ Specifies the path to the project root. If not specified the current working dir
 The root is used for finding the Astro configuration file.
 
 ```shell
-astro --root examples/snowpack dev
+astro --root myRootFolder/myProjectFolder dev
 ```
 
-### `--reload`
+### `--reload
```
FredKSchott, line 60: Can you update all of these headers to not use the "code tick" and just be plain text instead?

src/pages/en/core-concepts/astro-components.md
```diff
@@ -41,18 +41,22 @@ import Button from './Button.astro';
...
 - Importing data, like a JSON file
 - fetching content from an API or database
 - creating variables that you will reference in your template
 
+
+>⚠️ In Astro v0.21, a bug was introduced that requires imports inside components to be at the top of your component script.
+
 ```astro
 ---
+// Import components at the top of your script
```
FredKSchott, line 59:
```suggestion
```astro
---
// Note: Imports must live at the top of your file.
```

FredKSchott: I generally try to avoid calling things bugs in our docs, if they can instead be framed as "requirements". 

The migration guide is a bit different since it's literally recording changed behavior between specific versions, which is why it probably used the term "new bug" more liberally.

FredKSchott (approved):
LGTM with two small comments

zadeviggers:
Everything looks fine to me

PR #388, opened 2022-04-27
Edit: Add full hooks API reference for integrations

bholmesdev:
**Motivation:** Hackathon participants were confused how integrations should be built. This PR introduces detailed docs and examples for each API hook!

### DIFF
--- a/src/pages/en/reference/integrations-reference.md
+++ b/src/pages/en/reference/integrations-reference.md
@@ -43,9 +43,8 @@ interface AstroIntegration {
 
 ### astro:config:setup
 
-**When it's run:** on initialization, before either the [Vite](https://vitejs.dev/config/) or [Astro config](/en/reference/configuration-reference/) have resolved.
-
-**Use case:** for extending the project config. This includes updating the [Astro config](/en/reference/configuration-reference/), applying [Vite plugins](https://vitejs.dev/guide/api-plugin.html), adding component renderers, and injecting scripts onto the page.
+**When:** On initialization, before either the [Vite](https://vitejs.dev/config/) or [Astro config](/en/reference/configuration-reference/) have resolved.    
+**Why:** To extend the project config. This inludes updating the [Astro config](/en/reference/configuration-reference/), applying [Vite plugins](https://vitejs.dev/guide/api-plugin.html), adding component renderers, and injecting scripts onto the page.
 
 #### "config" option
 
@@ -100,18 +99,17 @@ A callback function to add a component framework renderer (i.e. React, Vue, Svel
 
 A callback function to inject a string of JavaScript content onto every page.
 
-The **`stage`** denotes how this script (the `content`) should be inserted. Some stages are inserted purely, while others are picked up by Vite:
+The **`stage`** denotes how this script (the `content`) should be inserted. Some stages allow inserting scripts without modification, while others allow optimization during [Vite's bundling step](https://vitejs.dev/guide/build.html):
 
-- `"before-hydration"`: Imported client-side, before the hydration script runs. Processed & resolved by Vite.
-- `"head-inline"`: Injected into a script tag in the `<head>` of every page. Not processed or resolved by Vite.
-- `"page"`: Injected into the JavaScript bundle of every page. Processed & resolved by Vite.
-- `"page-ssr"`: Injected into the frontmatter of every Astro page. Processed & resolved by Vite.
+- `"head-inline"`: Injected into a script tag in the `<head>` of every page. **Not** optimized or resolved by Vite.
+- `"before-hydration"`: Imported client-side, before the hydration script runs. Optimized and resolved by Vite.
+- `"page"`: Injected into the JavaScript bundle of every page. Optimized and resolved by Vite.
+- `"page-ssr"`: Injected into the frontmatter of every Astro page. Optimized and resolved by Vite.
 
 ### astro:config:done
 
-**When it's run:** after the Astro config has resolved and other integrations have run their `astro:config:setup` hooks.
-
-**Use case:** to retrieve the final config for use in other hooks.
+**When:** After the Astro config has resolved and other integrations have run their `astro:config:setup` hooks.  
+**Why:** To retrieve the final config for use in other hooks.
 
 #### "config" option
 
@@ -121,9 +119,8 @@ A read-only copy of the user-supplied [Astro config](/en/reference/configuration
 
 ### astro:server:setup
 
-**When it's run:** just after the Vite server is created in "dev" or "preview" mode, but before the `listen()` event is fired. [See Vite's createServer API](https://vitejs.dev/guide/api-javascript.html#createserver) for more.
-
-**Use case:** to update Vite server options and middleware.
+**When:** Just after the Vite server is created in "dev" or "preview" mode, but before the `listen()` event is fired. [See Vite's createServer API](https://vitejs.dev/guide/api-javascript.html#createserver) for more.  
+**Why:** To update Vite server options and middleware.
 
 #### "server" option
 
@@ -146,9 +143,8 @@ import
 
 ### astro:server:start
 
-**When it's run:** after the server's `listen()` event has fired.
-
-**Use case:** to intercept network requests at the specified address. If you intend to use this address for middleware, consider using `astro:server:setup` instead.
+**When:** Just after the server's `listen()` event has fired.  
+**Why:** To intercept network requests at the specified address. If you intend to use this address for middleware, consider using `astro:server:setup` instead.
 
 #### "address" option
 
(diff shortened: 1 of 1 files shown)


src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,159 @@ interface AstroIntegration {
...
+
+**When it's run:** just after the Vite server is created in "dev" or "preview" mode, but before the `listen()` event is fired. [See Vite's createServer API](https://vitejs.dev/guide/api-javascript.html#createserver) for more.
+
+**Use case:** to update Vite server options and middleware.
+
+#### "server" option
+
+**Type:** [`ViteDevServer`](https://vitejs.dev/guide/api-javascript.html#vitedevserver)
+
+An mutate-able instance of the Vite server used in "dev" and "preview" mode. For instance, this is [used by our Partytown integration](https://github.com/withastro/astro/tree/main/packages/integrations/partytown) to inject the Partytown server as middleware:
```
sarah11918, line 132: Up to you if you prefer the easy-on-the eyes mutate-able over mutatable, but either way, it'll be "A" one of those.

src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,159 @@ interface AstroIntegration {
...
+
+```js
+'astro:server:setup': ({ server }) => {
+  server.middlewares.use(
+    sirv(partytownLibDirectory, {
```
sarah11918, line 137: just checking: serv?

src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,159 @@ interface AstroIntegration {
...
-
-<!-- TODO: Detailed Hooks Reference
-
 ## Hooks
 
 ### astro:config:setup
+
+**When it's run:** on initialization, before either the Vite or Astro config have resolved.
+
+**Use case:** for extending the project config. This includes updating the Astro config, applying Vite plugins, adding component renderers, and injecting scripts or HTML elements onto the page.
```
roblevintennis, line 48: [Vite plugins](https://vitejs.dev/guide/api-plugin.html)

src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,159 @@ interface AstroIntegration {
...
+- `clientEntrypoint` - path to a file that executes on the client whenever your component is used. This is mainly for rendering or hydrating your component with JS.
+- `serverEntrypoint` - path to a file that executes during server-side requests or static builds whenever your component is used. These should render components to static markup, with hooks for hydration where applicable. [React's `renderToString` callback](https://reactjs.org/docs/react-dom-server.html#rendertostring) is a classic example.
+
+#### "injectScript" option
+
+**Type:** `(stage: InjectedScriptStage, content: string) => void;`
+
+A callback function to inject a string of JavaScript content onto every page.
+
+The **`stage`** denotes how this script (the `content`) should be inserted. Some stages are inserted purely, while others are picked up by Vite:
```
roblevintennis, line 103: Overall, this doc update is really fantastic!

I seem to get a bit hung up on some of the Vite specific stuff and I could be wrong but there's some assumed background knowledge there. I don't think you need to increase the scope to this doc, but linking to Vite resources to help get any prereq knowledge may be helpful.

> while others are picked up by Vite

What's the significance of this. Probably dumb question for those w/more understanding of Vite, but this is an example from this noob ;-)

src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,159 @@ interface AstroIntegration {
...
+
+- `"before-hydration"`: Imported client-side, before the hydration script runs. Processed & resolved by Vite.
+- `"head-inline"`: Injected into a script tag in the `<head>` of every page. Not processed or resolved by Vite.
+- `"page"`: Injected into the JavaScript bundle of every page. Processed & resolved by Vite.
+- `"page-ssr"`: Injected into the frontmatter of every Astro page. Processed & resolved by Vite.
```
aFuzzyBear, line 108: Could I ask for a bit of clarity around this; `page-ssr` this inject code directly into every `src/page/**/*.astro` file's frontmatter section, if so does it add to the head or the tail of the block?

src/pages/en/reference/integrations-reference.md
aFuzzyBear: Mutateable would be better here

src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,159 @@ interface AstroIntegration {
...
+
+**When it's run:** when the dev server is closed.
+
+**Use case:** to run any cleanup events you may trigger during the `astro:server:setup` or `astro:server:start` hooks.
+
 ### astro:build:start
+
+**When it's run:** after the `astro:config:done` event, but before the production build begins.
+
+**Use case:** to setup any global objects or clients you may need during a production build. This can also extend build configuration options in the [experimental adapter API](/en/reference/adapter-reference/).
```
aFuzzyBear, line 167:
```suggestion
**Use case:** To setup global objects or clients which maybe needed during a production build. This can also extend the build configuration options in the [experimental adapter API](/en/reference/adapter-reference/).
```

aFuzzyBear (commented):
Much needed, a great improvement on the API, thanks ben this would really help alot of folk out,

roblevintennis:
It also occurs to me that a nice follow up to this PR would be an example of settings up a dev env for local testing of said integration. For example, should I create a `/integration_test` that's a literal astro project? Then do I do `npm pack` and point to that tarball?  Where exactly should the integration files go and what are the conventions there?

src/pages/en/reference/integrations-reference.md
bholmesdev: `mutable` wins, my college education loses!

src/pages/en/reference/integrations-reference.md
bholmesdev: I can throw the `sirv` import at the top of the example for clarity, but I assumed it was a non-important detail 🤷

bholmesdev: Update: changing `sirv` to `partytownServer` to avoid confusion 👍

src/pages/en/reference/integrations-reference.md
bholmesdev: Asking the team about this! I felt pretty uneasy writing this section, because I honestly don't know which I would choose and when. Definitely re-drafting and adding example use cases 👍

FredKSchott:
This is great! The only big feedback that I'd have is to add an example below each hook that just shows all of the options, so that the reader understands what `“config” option` means. 
Example:

```
## astro:config:setup

When it’s run: ...
Use case: ...

'''
'astro:config:setup': ({ 
  config,
  command,
  updateConfig,
  addRenderer,
  injectScript,
}) => {
  // ...
},
'''
```

Open to other ideas here as well, as long as it solves the "what does `“config” option` mean?" problem.

src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,161 @@ interface AstroIntegration {
...
 ```
 
-
-<!-- TODO: Detailed Hooks Reference
-
 ## Hooks
 
 ### astro:config:setup
+
+**When it's run:** on initialization, before either the [Vite](https://vitejs.dev/config/) or [Astro config](/en/reference/configuration-reference/) have resolved.
```
FredKSchott, line 46: Nit: can you follow the format that we use on https://deploy-preview-388--astro-docs-2.netlify.app/en/reference/configuration-reference/

```diff
- **When it's run:** ...
- 
- **Use case:** ...
+ **When it's run:** ...  (note: 2 spaces at the end for a newline)
+ **Use case:** ...
```

Also, "When it's run" and "Use case" are almost in two different voices. Stylistically, I like to shorten these so that they are easier for the readers brain to "key" to when scanning the page. Example:



 
```md
**Lifecycle:** ... (alt: "Stage")
**Purpose:** ...
```

Finally, it could be interesting to do something like what Rollup does with their hooks docs (ex: https://rollupjs.org/guide/en/#buildend). This "type" item would actually also solve my comment I made about wanting to see an example with all of the options so that users knew what "config" option was referring to:

```
Type: `({config, command, updateConfig, addRenderer, injectScript}) => void`
Previous Hook: None
Next Hook: `astro:config:done` (clickable link to it)
```

src/pages/en/reference/integrations-reference.md
FredKSchott: Yea, agreed that these two sound the same here. Here's how I'd write them, @aFuzzyBear lmk if this is more clear:

```
- `"page"`: Similar to `head-inline`, except that the injected snippet is handled by Vite and bundled with any other `<script>` tags defined inside of Astro components on the page. The script will be loaded with a `<script type="module">` in the final page output.
- `"page-ssr"`: Injected into the frontmatter of every Astro page component. This is not commonly used, however it can be useful for injecting a CSS `import` into every page via its frontmatter.
```

src/pages/en/reference/integrations-reference.md
bholmesdev: I tried lifecycle but it didn't make the most sense in context. I landed on simply "when" and "why!"

src/pages/en/reference/integrations-reference.md
aFuzzyBear: That is indeed alot clearer thank you guys 💚

Holben888: Makes sense to me too!

src/pages/en/reference/integrations-reference.md
bholmesdev: Totally understand! I updated this to say:

> The **`stage`** denotes how this script (the `content`) should be inserted. Some stages allow inserting scripts without modification, while others allow optimization during [Vite's bundling step](https://vitejs.dev/guide/build.html)

Let me know if this helps!

aFuzzyBear (commented):
Its nae mare skewwhiff

bholmesdev:
@roblevintennis Don't wanna lose your detailed suggestion! I'd suggest moving this [to a docs issue](https://github.com/withastro/docs/issues/new) so we can follow-up here. I'll look into this myself for some best practices with `pnpm` 😁

src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,206 @@ interface AstroIntegration {
...
+
+**Type**: `AstroConfig`
+
+A read-only copy of the user-supplied [Astro config](/en/reference/configuration-reference/). This is resolved _before_ any other integrations have run. If you need a copy of the config after all integrations have completed their config updates, [see the `astro:config:done` hook](#astroconfigdone).
+
+#### "command" option
+
+**Type:** `'dev' / 'build'`
+
+- `dev` - project was executed with `astro dev` or `astro preview`
```
aFuzzyBear, line 71:
```suggestion
- `dev` - Project's executed with `astro dev` or `astro preview`
```

src/pages/en/reference/integrations-reference.md
```diff
@@ -40,30 +39,206 @@ interface AstroIntegration {
...
+- `build` - project was executed with `astro build`
+
+#### "updateConfig" option
+
+**Type:** `(newConfig: Record<string, any>) => void;`
+
+A callback function to update the user-supplied [Astro config](/en/reference/configuration-reference/). Any config you provide **will be merged with the user config + other integration config updates,** so you are free to omit keys!
```
aFuzzyBear, line 78:
```suggestion
Callback function to update the user-supplied [Astro config](/en/reference/configuration-reference/). Any configurations you provide **will be merged with the user config + other integration config updates,** so you are free to omit keys! 
```

aFuzzyBear (commented):
If possible could we get this merged before Close of Play?

FredKSchott (approved):

PR #392, opened 2022-04-28
Add start script to manual install instructions

sarah11918:
Our current instructions only include dev/build/preview scripts, but then we send the user back to Step 3 of the automatic install instructions: "run start."

(This also matches the package.json scripts generated by `create-astro` / in our starter templates.)


delucis (approved):
Well done catching that! LGTM 👍

PR #394, opened 2022-04-28
Consistent reference to islands architecture

sarah11918:
Page had a mix of island and islands when using the phrase "islands architecture." Now it does not.  :)


delucis (approved):
These islands are looking better than ever! 🏝️

PR #395, opened 2022-04-28
Improve handling of RTL languages

delucis:
This PR stores each language’s writing direction in `src/i18n/languages.ts` to avoid needing to specify it in Markdown frontmatter (which is what Arabic has done up until now).

This allows us to layout RTL-language pages correctly even when they contain LTR fallback content. After #365 the site navigation was on the right on the Arabic “Getting Started” page as expected, but on the left for fallback pages. Making this a per-language and not a per-page setting makes this consistent.

Just to demonstrate what this fixes, here are two Arabic pages as they render currently:

<table>
<tr>
	<th>Arabic content
	<th>Fallback content
<tr>
	<td>
image
</td>
	<td>
image</td>

</table>

This commit also updates the `add-language` script to collect writing direction.

### DIFF
--- a/scripts/add-language.mjs
+++ b/scripts/add-language.mjs
@@ -23,6 +23,11 @@ class LanguageScaffolder {
 	#tag = '';
 	/** Language name (e.g. English, Português do Brasil, etc.) */
 	#name = '';
+	/**
+	 * Language writing direction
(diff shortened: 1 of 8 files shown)


delucis:
Closing this as it complicates our languages dictionary and won't be needed for now.

PR #396, opened 2022-04-28
Add script to check translation slugs match English slugs

delucis:
Follow up PR to #365. @FredKSchott suggested:

> a lint script that warns if a language file (ex: `/de/foobar.md`) doesn't have a matching English file (ex: `/en/foobar.md`). We could run this in CI to make sure all PRs are passing.

This PR adds:

- A script that checks slugs as proposed above
- A GitHub action that will run this linter & the broken link checker for PRs and commits to the main branch

This slug linter will error in CI if it finds a slug that doesn’t have an English counterpart. The broken link checker currently just logs its output and won’t error (we can think about changing that once the 41 broken links it reports currently have been cleaned up).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+      branches: [main]
+
+# Automatically cancel in-progress actions on the same branch
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup PNPM
+        uses: pnpm/action-setup@v2.2.1
+
+      - name: Setup Node.js 16.x
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+          cache: 'pnpm'
+
+      - name: Install Dependencies
+        run: pnpm install
+
+      - name: Lint
+        run: pnpm run lint
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@types/react": "^17.0.43",
     "astro": "^1.0.0-beta.17",
     "bcp-47-normalize": "^2.1.0",
+    "fast-glob": "^3.2.11",
     "htmlparser2": "^7.2.0",
     "kleur": "^4.1.4",
     "node-fetch": "^3.2.3",
--- a/package.json
+++ b/package.json
@@ -8,13 +8,14 @@
     "build": "astro build",
     "preview": "astro preview",
     "format": "prettier -w .",
-    "lint": "lint:linkcheck",
+    "lint": "pnpm run lint:slugcheck && pnpm run lint:linkcheck",
     "add-language": "node ./scripts/add-language.mjs",
     "lint:a11y": "start-test 'yarn build && yarn preview' 3000 'yarn lint:a11y:local'",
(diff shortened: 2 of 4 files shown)


hippotastic:
Looks good! I've tested it locally and it seems to work well.

The only nitpick I have is that you're chaining both linters with `&&` in `package.json`, and that will cause the linkcheck linter to be skipped in case the slug linter finds any errors. I'd rather like to have separate tasks in the CI workflow that can fail independently from each other.

I'd suggest defining three jobs in `ci.yml` instead: One `build` job that performs the build process needed for any linters that may need it in the future, and two lint jobs. Here's an example implementation:

```yml
name: CI

on:
  push:
    branches: [main]
  pull_request:
    branches: [main]

# Automatically cancel in-progress actions on the same branch
concurrency:
  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.head_ref || github.ref }}
  cancel-in-progress: true

defaults:
  run:
    shell: bash

jobs:
  # Build the docs using `pnpm run build`, and then upload `dist/` as a build artifact,
  # allowing any dependent jobs to use the output without having to re-run the build.
  build:
    name: Build Docs
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3

      - name: Setup PNPM
        uses: pnpm/action-setup@v2.2.1

      - name: Setup Node
        uses: actions/setup-node@v3
        with:
          node-version: 16
          cache: 'pnpm'

      - name: Install dependencies
        run: pnpm install

      - name: Build Docs
        run: pnpm run build

      - name: Compress Build Output
        run: tar -czf dist.tar.gz dist

      - name: Upload Build Output
        uses: actions/upload-artifact@v3
        with:
          name: dist
          path: dist.tar.gz
          if-no-files-found: error

  # Run `pnpm run lint:linkcheck:nobuild` on the build output generated by the `build` job
  linkcheck:
    name: Check for Broken Links
    runs-on: ubuntu-latest
    needs:
      - build
    steps:
      - name: Checkout
        uses: actions/checkout@v3

      - name: Setup PNPM
        uses: pnpm/action-setup@v2.2.1

      - name: Setup Node
        uses: actions/setup-node@v3
        with:
          node-version: 16
          cache: 'pnpm'

      - name: Install dependencies
        run: pnpm install

      - name: Download Build Output
        uses: actions/download-artifact@v3
        with:
          name: dist

      - name: Extract Build Output
        run: tar -xzf dist.tar.gz dist
      
      - name: Lint
        run: pnpm run lint:linkcheck:nobuild

  # Run `pnpm run lint:slugcheck` (this job does not depend on the build and can run in parallel)
  slugcheck:
    name: Check for Mismatching Slugs
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3

      - name: Setup PNPM
        uses: pnpm/action-setup@v2.2.1

      - name: Setup Node
        uses: actions/setup-node@v3
        with:
          node-version: 16
          cache: 'pnpm'

      - name: Install dependencies
        run: pnpm install

      - name: Lint
        run: pnpm run lint:slugcheck
```

hippotastic:
Additional note: In the future, our broken link checker should also set an exit code of 1 in case any broken links were found.

I'm currently not doing this because we have several existing broken links in the docs, and we probably don't want the PR checks to fail every time. This will be solved once I've updated the broken link checker to keep its state. Once we have access to the current state of broken links in the `main` branch, we can then compare the broken links in a PR to those, and only fail the PR check if NEW broken links were introduced.

delucis:
Thanks @hippotastic! What do you think about focusing this PR only on running the slug checker and then you can add the link checker as part of your PR? Would that make most sense?

I also wonder if we can avoid some of the duplicate work between the tasks above (all the checkout, setup, dependencies etc.), I’ll look into that.

delucis:
Awesome! OK, I’ll update this to just run the slug check for now and leave the rest to you. Thanks 😄

PR #397, opened 2022-04-28
Polished component and especially slot content

sarah11918:
Team effort, to revamp the slot material to address some slot issues raised in the community re: mixing with other frameworks.

Also, added some new touches re: our fancy new directives reference! (So... check that!)


sarah11918:
@Jutanium Here are the fruits of our orangey labours! (Where are our socks, indeed?)

Jutanium:
🧦 🧦 ❤️

src/pages/en/core-concepts/astro-components.md
```diff
@@ -215,12 +218,21 @@ const name = "Astro"
 
 ### Slots
 
-The `<slot>` element is a placeholder for HTML which will be passed in from outside of the component by "children" (as they are called in React or Preact). Slots are a way of passing data into an Astro component and are useful when you will want to reuse an "outer" component, rendered "around" data coming from an external source.
-
-A component that uses the `<slot />` element can be thought of as a reusable "wrapper" around other content, and this pattern is the basis of an Astro [Layout component](/en/core-concepts/layouts).
+Astro components can accept HTML content in your component template. This "child" content will be injected into its `<slot />` element.
 
+This pattern is the basis of an Astro layout component: an entire page of HTML content can be “wrapped” with `<Layout></Layout>` tags and sent to the Layout component to render inside of common page elements.
```
aFuzzyBear, line 223:
```suggestion
```
Move this to a better location under the code block

src/pages/en/core-concepts/astro-components.md
```diff
@@ -236,20 +248,25 @@ const { name } = Astro.props
   <slot />  <!-- children will go here -->
   <Footer />
 </div>
+```
+
```
aFuzzyBear, line 252:
```suggestion
This pattern forms the basis of an Astro layout component. Whereby different islands can be placed within your `<Layout>` along side HTML markup that are coming from external sources. To do this we place `<slots>` and *named* slots `<slots name="">` within our layout to direct where the content goes. 
```

src/pages/en/core-concepts/astro-components.md
```diff
@@ -236,20 +248,25 @@ const { name } = Astro.props
...
+
+#### Named Slots
 
-// src/components/Person.astro
+You can pass multiple pieces of content to your Astro component, and place them precisely, using named slots.
```
aFuzzyBear, line 256:
```suggestion
You can precisely pass multiple pieces of content directly into your Astro component, by *naming* the slots. 
```

aFuzzyBear (commented):
Just a few thoughts, Ill be back , need to rest eyes,

src/pages/en/core-concepts/astro-components.md
delucis: I prefer the existing wording here — “passing precisely” seems confusing to me. But maybe the emphasis is helpful: `using **named slots**`?

src/pages/en/core-concepts/astro-components.md
```diff
@@ -215,12 +218,21 @@ const name = "Astro"
 
 ### Slots
 
-The `<slot>` element is a placeholder for HTML which will be passed in from outside of the component by "children" (as they are called in React or Preact). Slots are a way of passing data into an Astro component and are useful when you will want to reuse an "outer" component, rendered "around" data coming from an external source.
-
-A component that uses the `<slot />` element can be thought of as a reusable "wrapper" around other content, and this pattern is the basis of an Astro [Layout component](/en/core-concepts/layouts).
+Astro components can accept HTML content in your component template. This "child" content will be injected into its `<slot />` element.
```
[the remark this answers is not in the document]
delucis: Agree that this needs some work — I find that first sentence hard to parse. “HTML content in your component template?” Sure my component template is full of HTML content! 🤔 

How about something like

```suggestion
If you want to pass HTML content into your component template, you can use the `<slot />` element to control where that “child” content is rendered.
```

src/pages/en/core-concepts/astro-components.md
aFuzzyBear: Still got to finish reviewing this, 😬

src/pages/en/core-concepts/astro-components.md
```diff
@@ -215,12 +218,21 @@ const name = "Astro"
...
+import Wrapper from '../components/Wrapper.astro';
+---
+<Wrapper name="Astro">
+  <h2>I am a person.</h2>
+  <p>Here is some stuff about me.</p>
+</Wrapper>
+
+
 // src/components/Wrapper.astro
 ---
```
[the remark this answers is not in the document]
delucis: I guess there’s also a question about whether it would more helpful to see `Wrapper` first and then the page that uses it?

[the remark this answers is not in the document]
sarah11918: Both things I'd thought about.
a. They were together initially, so the formatting issue has always been there, but I have no issues splitting them up.

b. This was an explicit choice to swap the order (not saying it's right, just saying it was an intentional choice) and I'll add a comment below to explain the thinking behind this change before going point by point through all the (great, helpful!) comments.  :)

sarah11918:
Quick note, since I'm not yet sitting down to really tackle this, just to give some context to steer the convo by the time I get there:  😄 

Based on feedback/issues re: using slots with other framework components that also have a concept of slots, Dan and I tried to reorganize the logical flow of material to see if that would help preempt that particular problem. Essentially:

FROM: slots are placeholders, components with slots are wrappers

TO: some components can have "children" html elements that get passed to them; that stuff goes in to a component's slot.

So, just the context to know that that's where these changes came from, but of course, I want the explanation to work just in general, and not to be optimized for one thing we can see coming.

Love that everyone's looking at this so critically, and if general consensus is that changing this section's logic is NOT preferred, we can go back to the original for now, and the drawing board for later. That's an option, too!  (I still want the other changes sprinkled throughout the file for polish, though!)

src/pages/en/core-concepts/astro-components.md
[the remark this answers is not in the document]
delucis: Got it! I’m not that concerned on the order question if it’s been thought about, but do think the formatting change would be good. Team work! 🙌

src/pages/en/core-concepts/astro-components.md
sarah11918: Yes, I had been thinking about this.

sarah11918:
OK, @Jutanium @delucis @aFuzzyBear ... I'm "slotting" you in for another round! 

Can I suggest, DON'T revisit the comments above, and DO start reading at "Slots" to see, with fresh eyes and standing on its own, what you think??

Jutanium:
Let's Get That Merged

sarah11918:
🥳  Let he who complains, open the next PR!

PR #398, opened 2022-04-28
Remove unused imports from TabBox.astro

### LINKED ISSUE #387: TabBox.astro component uses the Markdown component
Just reminding us that when the Markdown component is no longer built-in, we'll need to address this.

delucis:
Closes #387

Turned out `TabBox.astro` _imports_ the Markdown component, but didn’t actually use it, so this was an easy fix.


sarah11918:
Well, that worked out well! Thanks, @delucis ! 🥳

Issue #224, opened 2022-03-15
a11y contrast issues after theme update


delucis:
@FredKSchott Looks like deb66f5 re-introduced a few a11y contrast issues that had been fixed in #198. We can’t use the orange accent for links against a light background.

Even in dark mode links containing inline code now fail an audit because the code bg is just _slightly_ lighter:

<img width="340" alt="image" src="https://user-images.githubusercontent.com/357379/158362143-cf9a620d-c98c-43c5-866d-15af1bbb3192.png">

The dark theme fix could be fairly simple: set `--theme-code-inline-bg` to a darker shade than the background gradient (although even at full black that doesn’t quite pop like the lighter shade — would work better with a slightly lighter bottom gradient shade).

In the light theme we need to choose a different colour for links (i.e. use `--theme-text-accent` not `--theme-accent`). I know the orange really pops and I love it too, but it’s a long way from being accessible (a ratio of 4.5 is the minimum to pass an AA audit):

<img width="229" alt="image" src="https://user-images.githubusercontent.com/357379/158364547-77e7b40e-5ae2-43b9-a69a-a54a755d70c6.png">

Maybe a darker orange or a purple? 

---

If anyone’s interested in how to do an audit, I find Firefox’s built-in Accessibility panel in dev tools the easiest to use quickly. You can select “Contrast” from the “Check for issues” drop-down menu and it will scan the page for any colour contrast issues, listing any text with insufficient contrast.

<img width="268" alt="image" src="https://user-images.githubusercontent.com/357379/158365264-ef3df0e0-7a15-4908-adef-d4aa7110087a.png">

On the docs site, the Framework Components page is a good litmus test as it contains most possible combos of background and text colours.

sarah11918:
Just musing out loud, but I had it in my head that sometimes the numbers didn't always coincide with what was otherwise agreed-upon as "easiest to read" by readers, and I thought orange (and in particular, orange-white) was a particular culprit here.

I think some standards are scheduled to change (but I don't recall a timeline) to reflect some of the issues with current ratios, and of course, if we are required in some way to pass particular standards now, then I suppose we have to adhere to the letter of them!

delucis:
[...]

PR #198, opened 2022-03-09
fix(a11y): Pass colour contrast audit

delucis:
This PR addresses colour contrast accessibility issues with the current docs theme in light mode.

It adds:

- Darker text accent colour (links etc.).

  I chose a darker, but also marginally _redder_ hue as that helps with contrast without needing to go too dark and lose vibrancy. I think it still vibes with the accent orange, but feedback welcome.

- Slightly darker code block background colour.

- Tweaked two text colours used for syntax highlighting to improve contrast

Tested using in-browser colour contrast tools in Chrome and Firefox.

I have more cosmetic improvements I’d like to propose, but I’ll add those in a separate PR to make this (hopefully) quick to merge and address the accessibility issues.

PR #406, opened 2022-04-29
fix(a11y): Purple links in light mode for accessible contrast

delucis:
I know we discussed this in #198, but seeing as #224 was closed on the basis that I’m “ON THE CASE, ON THE REG,” here I am! On the case!

The current link colour provides insufficient colour contrast in light mode by *quite a bit*. [Guidelines](https://webaim.org/articles/contrast/) require a minimum contrast of 4.5:1 for text content, but the orange links are only around 2.5:1, even lower for inline code snippets within links.

This PR switches links in the body content to use a purple instead of the current orange, which provides a passing colour ratio:

| Before | After |
|---|---|
| image | image |

This change only applies in light mode. In dark mode the orange links work well and don’t need changing.

**Why purple?** The issue with orange is that to get enough contrast you need to reduce lightness quite a bit and it loses its vibrancy. It also introduces two slightly different shades of orange between links & logotype, which kind of ends up looking like a mistake. Purple fits with the rest of the theme and provides more contrast without seeming too muted.

This change likely does make links stand out from the rest of the page slightly less. If that’s a concern we could consider other ways to highlight links like a more pronounced underline instead of relying solely on colour.

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -184,7 +184,7 @@ small,
 }
 
 a {
-	color: var(--theme-accent);
+	color: var(--theme-accent-secondary);
 	text-underline-offset: 0.08em;
 	align-items: center;
 	gap: 0.5rem;
--- a/public/theme.css
+++ b/public/theme.css
@@ -57,7 +57,7 @@
 :root {
 	color-scheme: light;
 	--theme-accent: hsla(var(--color-orange), 1);
-	--theme-accent-secondary: hsla(302, 60%, 38%, 1);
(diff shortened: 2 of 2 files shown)


aFuzzyBear (approved):
Purpflect

Chrissdroid:
LGTM

sarah11918:
I'm so confused? Did I not comment on this? (I remember saying I was having trouble in sunlight?) And, wasn't someone going to just make an executive decision on this?

At any rate, LGTM! 💜  (broken orange heart emoji)

PR #405, opened 2022-04-29
Markdown guide updates

JuanM04:
- Added "Importing Markdown"
- Removed references to markdown parsers
- Updated plugins sections
- Added notes section
- Formatted

[Rendered page](https://deploy-preview-405--astro-docs-2.netlify.app/en/guides/markdown-content/)

PR #411, opened 2022-04-29
Update Markdown component guidance

FredKSchott:
- Added Markdown component warnings to the relevant sections
- Removed an outdated piece of docs around markdown customization (no longer possible in v1.0-beta)


sarah11918:
Waiting on PR #405 (close to done) before this one!

src/pages/en/guides/markdown-content.md
```diff
@@ -203,7 +203,7 @@ back to it later.
 
 ## Markdown Component
 
-Astro has a dedicated component used to let you render Markdown in `.astro` files. 
+> NOTE: The `<Markdown>` component does not work in SSR and may be removed before v1.0. It should should be avoided if possible. To use Markdown in your templates, use a seperate `.md` file and then `import` it into your template as a component. Read this [RFC Discussion](https://github.com/withastro/rfcs/discussions/179) to learn more.
```
sarah11918, line 206:
```suggestion
> NOTE: The `<Markdown>` component does not work in SSR and may be removed before v1.0. It should should be avoided if possible. To use Markdown in your templates, use a seperate `.md` file and then [`import` Markdown](/en/guides/markdown-content#importing-markdown) into your template as a component.
```
You could do something like this, now that #405 is merged.

sarah11918:
@FredKSchott This will have conflicts to resolve now that #405 is merged. And, a reminder in my comment above that we can now link directly to info about importing Markdown in the docs, but otherwise, LGTM!

src/pages/en/guides/markdown-content.md
sarah11918: (Unless your intention really was to point to the Discussion/rationale, of course! But, I do think having the link to how to import Markdown content, now that we have it, works here.)

FredKSchott: Agreed! That was a bit of a placeholder

FredKSchott:
Thanks @sarah11918 !

PR #415, opened 2022-04-30
added link: directives reference back to scripts

sarah11918:
The directives-reference page linked back to info about CSS styling, but not to our info about scripts in Astro components.


Princesseuh (approved):
Looks good to me!

PR #416, opened 2022-04-30
Minor additions to components - mostly links

sarah11918:
Mostly adding lists to things that now exist (integrations, directives page) and an example of `set:html`

### DIFF
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -100,7 +100,7 @@ const myFavoritePokemon = [/* ... */];
   {myFavoritePokemon.map((data) => <li>{data.name}</li>)}
 <ul>
 
-<!--Use a template directive to inject an unescaped HTML string into an element: -->
+<!-- Use a template directive to inject an unescaped HTML string into an element: -->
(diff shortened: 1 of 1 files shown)


delucis:
This all LGTM! I do wonder if that code block you added the directive example to is getting a little unwieldy? Maybe we need to think about breaking it up at some point.

src/pages/en/core-concepts/astro-components.md
```diff
@@ -100,6 +100,10 @@ const myFavoritePokemon = [/* ... */];
   {myFavoritePokemon.map((data) => <li>{data.name}</li>)}
 <ul>
 
+<!--Use a template directive to inject an unescaped HTML string into an element: -->
```
sarah11918, line 103:
```suggestion
<!-- Use a template directive to inject an unescaped HTML string into an element: -->
```

sarah11918:
> This all LGTM! I do wonder if that code block you added the directive example to is getting a little unwieldy? Maybe we need to think about breaking it up at some point.

I'm OK to think about that at some point. I don't think it's blocking for now?  :)

delucis:
> I don't think it's blocking for now? :)

Agreed!

PR #418, opened 2022-05-01
Translate ThemeToggle labels and add system theme sync

### LINKED ISSUE #106: 📘 DOC: Add support for clearing your theme selection in the docs
Currently, if you select light theme or dark theme in the theme toggle, it is then fixed, and there is no way to return to `prefers-color-scheme` without going into Inspect Element and clearing the local storage key.

I'd like to propose using <https://cssdb.org/>'s Theme Switcher that adds an indeterminate state to the theme toggle.

light mode:
image

dark mode:
image

"let the browser choose" mode:
image

### LINKED ISSUE #399: Theme Switcher a11y and i18n concerns
The [theme switcher component](https://github.com/withastro/docs/blob/main/src/components/RightSidebar/ThemeToggleButton.tsx) uses `aria-label` with template strings to clarify to screen readers the current theme you're using. The problem with this is that it makes harder for translators to just translate it as part of [`ui.ts`](https://github.com/withastro/docs/blob/main/src/i18n/en/ui.ts).

If we can simplify the current component to work without template strings in a way that allow us to translate "Use dark/light theme" would be great. If not, I suggest we change the solution to Heydon Pickering's [Theme Switcher component on Inclusive Components' blog](https://inclusive-components.design/a-theme-switcher/)

Yan-Thomas:
This PR makes two things possible:
- Add translatable labels to ThemeToggle, used in its `aria-label` attribute. See pt-BR `ui.ts` as an example. Fixes #399 
- Remove the selected theme from `localStorage` when the selected theme is equal to your system's preferred color scheme, syncing up with your system. This allows you to let the browser decide what's best. Fixes #106


sarah11918:
@jasikpark Would you kindly take a look at this, with Issue #106 in mind and let us know what you think?

Yan-Thomas:
Made a few changes to fix the merge conflicts, sorry for not fixing it earlier 😅

sarah11918:
Hey @Yan-Thomas I know this has been stuck here a little while. :sweat_smile: 

So, I'm going to shamelessly throw some names in here to see if you can get some attention! Do we have any thoughts from:

@jasikpark @aFuzzyBear @natemoo-re @bholmesdev  ?

aFuzzyBear (approved):
I like the improvements made, I do also like the way you've added the touch of making it i18n compt. 
Great work to the Sidebar, 👊 you really have took too this wee part of the docs site and made it better, thank you @Yan-Thomas 💚💚

sarah11918:
@Yan-Thomas I can see that you've changed `ui.ts` for pt-BR. Will this need to be changed for all other languages, too?

Yan-Thomas:
Hi @sarah11918, AFAIK it will fallback to the English version for languages that didn't update `ui.ts` yet. But yes, we just have to add `themeToggle.useDark/Light` labels after.

sarah11918:
Great! I put out a call in docs-i18n general channel for 1 person to update their own language, so I'll merge this now!

bholmesdev:
Belated LGTM!

Issue #400, opened 2022-04-28
"On this page" doesn't indicate which part you're currently reading


Yan-Thomas:
Something that I feel is missing on the current "On this page" component is the dynamic focus on headings based on where you're on the page, like what React.js Tutorial does in the image below:
![EXDlhQP](https://user-images.githubusercontent.com/61414485/165864998-55be262d-a3e3-4c15-855f-f88d217c4794.gif)

This can be achieved using [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API), the downside is that it means adding more JS to the docs of a framework that has "Ship Less JavaScript" as catchprase😅.

Some pages are big and this functionality helps me keep track of where I am as I read, but the million dollar question is: Is this really valuable to most of us and the extra JavaScript sprinkles are worth it?

sarah11918:
So I took a look and NuxtJS docs do have this. In mobile mode, they don't even have "on this page" at all.

We have to re-do all the comparisons for Astro vs X page anyway... would you be up for looking at e.g. https://nuxtjs.org/docs/get-started/installation as a first step and seeing whether their "on this page" that shows where you are in the page seems to affect performance very much, @Yan-Thomas ? How much work would it be to try adding it to our docs and examining the difference in performance?

Yan-Thomas:
Hey @sarah11918 !

About Nuxt.js docs, they add this "Table of Contents" dropdown that sticks to the page as you scroll on mobile:

![Screenshot_20220430-185543_Samsung Internet.jpg](https://user-images.githubusercontent.com/61414485/166123973-14ed3a73-74af-4c49-806f-63e8b0edaa01.jpg)

I think I can easily PR this tomorrow and show how it affects page load through Lighthouse. Also, since we have preview deploys, anyone can use other metrics to test the performance impact as well!

PR #419, opened 2022-05-01
Add visual indicator to the table of contents

Yan-Thomas:
Adds styling indicating the part of the docs you're currently reading as you scroll the page. This only works for desktop, on mobile you can't see the table of contents as you scroll, so I didn't include it. Solution for #400 

See the gif as an example:
Ojm0jKO

The performance difference on Lighthouse (looks like the preview deploy is slower but the other metrics are the same, so I think it doesn't really impact the overall performance, needs further testing):

| Preview  | Official |
| image  | image  |

This is not 100% yet. I think there are some improvements we can make to make the transition between headings smoother, I just don't know how yet.

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -389,11 +389,14 @@ h2.heading {
 
 .header-link:hover,
 .header-link:focus,
-.header-link:focus-within {
+.header-link:focus-within,
+.current-header-link {
 	border-inline-start-color: var(--theme-accent-secondary);
 }
+
 .header-link:hover a,
-.header-link a:focus {
+.header-link a:focus,
+.current-header-link {
 	color: var(--theme-text);
 	text-decoration: underline;
 }
--- a/src/components/RightSidebar/TableOfContents.tsx
+++ b/src/components/RightSidebar/TableOfContents.tsx
@@ -23,8 +23,44 @@ const TableOfContents: FunctionalComponent<Props> = ({ headers = [], labels }) =
 			}));
 		};
 
+		const addObservers = () => {
+			const observerOptions = {
+				root: null,
+				rootMargin: '0px',
+				threshold: [1],
+			};
+
+			const setCurrent = (element) => {
+				const tocItems = document.querySelectorAll('.sidebar-nav #toc li');
+
+				element.map((e) => {
+					const currentHeadingLink = document.querySelector(`.sidebar-nav #toc li a[href="#${e.target.id}"]`);
+
+					if (e.isIntersecting === true && currentHeadingLink !== null) {
+						tocItems.forEach((item) => {
+							item.classList.remove('current-header-link');
+
+							if (item.contains(currentHeadingLink)) {
+								item.classList.add('current-header-link');
+							}
+						});
+					}
+				});
+			};
+
+			const observeHeadings = new IntersectionObserver(setCurrent, observerOptions);
+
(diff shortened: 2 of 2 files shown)


sarah11918:
Video looks snazzy, @Yan-Thomas ! 💅 

Let's see what @hippotastic and @delucis think, too! (I don't see anything here that shouldn't also work in all the i18n languages... and I think this should even target RTL correctly, too?)

delucis:
Nice work @Yan-Thomas! I think you’re good on the performance side — that difference appears to be mainly because of the feedback UI Netlify injects into previews. If you use the preview for a specific commit, Netlify won‘t load that UI and you get comparable performance to the current site. Try running the test on https://626f1e99ed55da0008f8d7b8--astro-docs-2.netlify.app/en/getting-started/ for example.

I’m going to dig into the component now and I’ll let you know if I come across anything that isn’t perfectly cromulent.

hippotastic:
I love the little indicator! Very nice idea. I also noticed the quirk mentioned by @delucis, however. Especially on short pages like [Getting Started](https://deploy-preview-419--astro-docs-2.netlify.app/en/getting-started/), it is very noticeable that the indicator highlights a different section than the one I clicked on. It's sometimes even 4 sections off for me - probably depends on the browser window dimensions.

I also noticed that the highlighted section tends to jump somewhere completely different when I change the scroll direction. If you scroll up, it suddenly seems to highlight the top section on the page, whereas before it always highlights the bottom one. I'd prefer to always highlight the top section, no matter which direction I scroll to. This should also fix difference between clicked section & highlighted section in most cases - maybe except for the bottom sections of a page, where the ones above could still get highlighted. I'm not sure if it's easy enough to implement, but in case there are multiple visible sections on the page, the script could check the current URL fragment in the browser's navigation bar and prefer the section with that fragment if visible.

Yan-Thomas:
Thanks everyone!

I have a mental draft of how to fix both of these problems. When I go back home I'm gonna test this and commit if it works properly:
- Get the intersecting heading's siblings and check if the one on top is still intersecting (this way, the one on top is prioritized), if there are no more headings below the current one, the one on the bottom is highlighted instead.

- As @hippotastic suggested, check the current URL fragment and highlight accordingly.

delucis:
@Yan-Thomas I’ve been playing around. Mind if I open a PR to your branch?

Yan-Thomas:
Sure @delucis !

delucis:
Ok folks, this LGTM but seeing as @Yan-Thomas kindly accepted my changes, I can't really be the one to sign off on this. Want to take a last look @sarah11918 or @hippotastic?

hippotastic:
This is a huge improvement, well done!

I still have the issue that the very last section doesn't always get picked up, and using the location hash as discussed should solve this. If any of you would like to try adding this to get it to 💯, I'd really appreciate that. If not, I still think that the new functionality is great and could also be released as is.

sarah11918:
Tested a few different screen sizes, light and dark mode, and it's looking sharp!  LGTM! I'm happy to approve, and if we notice anything later, it can be revisited.

hippotastic:
Yeah, let's approve this. We can always refine it even more at a later point in time. This is a really nice visual improvement as it is right now.

delucis:
Aaaaaand it’s live! (Or will be in minute or so it takes Netlify to build.)

Thanks so much for championing this idea @Yan-Thomas! 🎉

PR #424, opened 2022-05-02
Added explicit warning re: importing CSS files at the top

sarah11918:
In response to a support thread, this both makes it explicit that an imported CSS file must follow the same rules as any other ESM import and links back to the page/section explaining the Astro component script.


delucis (approved):
Thanks for catching that @sarah11918! LGTM 👍

sarah11918:
Haha, I went to triple-check the Netlify preview one last time before merging, and was confused that I wasn't seeing the nice "You are here" in the right sidebar! 😂

PR #426, opened 2022-05-03
Improve "On this page" scrollbar

Yan-Thomas:
The docs are currently using the default scrollbar on the [`RightSidebar.astro`](https://github.com/withastro/docs/blob/main/src/components/RightSidebar/RightSidebar.astro) component. The default scrollbar is fine for the page itself, but on the table of contents, it creates an unpleasant look on larger pages. 

This PR adds the same custom scrollbar used on the [`LeftSidebar.astro`](https://github.com/withastro/docs/blob/main/src/components/LeftSidebar/LeftSidebar.astro) component as a starting point. 

I think it is worth discussing if we should maintain this more discrete scrollbar styling or change for one that is always visible.

| Astro docs  (current) | Nuxt.js docs | Astro docs (PR) |
| ------------- | ------------- |  ------------- |
| image  | image  | image |

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -8,6 +8,26 @@
 	outline: none;
 }
 
+/* Firefox */
+* {
+	scrollbar-color: var(--theme-dim-light) transparent;
+}
+
+/* Webkit */
+
+::-webkit-scrollbar,
+::-webkit-scrollbar-track {
+	/* same as --theme-bg-gradient but without 'fixed' */
+	background: linear-gradient(180deg, var(--theme-bg-gradient-top), var(--theme-bg-gradient-top) var(--theme-navbar-height), var(--theme-bg-gradient-bottom));
+}
+
+::-webkit-scrollbar-thumb {
+	background-color: var(--theme-dim-light);
+	border: 4px solid transparent;
+	background-clip: content-box;
+	border-radius: 10px;
+}
+
 :root {
 	--user-font-scale: 1rem - 16px;
 	--max-width: calc(100% - 1rem);
--- a/src/components/LeftSidebar/LeftSidebar.astro
+++ b/src/components/LeftSidebar/LeftSidebar.astro
@@ -71,23 +71,4 @@ for (const section of sidebarSections) {
 		max-height: 100vh;
 		padding-bottom: var(--theme-navbar-height);
 	}
-
-	/* Scrollbar Design Technique */
-	/* Firefox: */
-	.nav-groups {
-		scrollbar-color: var(--theme-dim-light) transparent;
-	}
-	/* Webkit: */
-	.nav-groups::-webkit-scrollbar,.nav-groups::-webkit-scrollbar-track {
-		background-color: transparent;
-	}
-	.nav-groups:active::-webkit-scrollbar-thumb,.nav-groups:focus-within::-webkit-scrollbar-thumb,.nav-groups:focus::-webkit-scrollbar-thumb,.nav-groups:hover::-webkit-scrollbar-thumb {
-		background-color: var(--theme-dim-light);
-		border: 4px solid transparent;
-		background-clip: content-box;
-		border-radius: 10px;
-	}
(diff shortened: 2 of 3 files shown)


delucis (commented):
If we’re using the same styles across left and right sidebars could it make sense to consolidate in `public/index.css`?

I only tested really quickly but I think it might even be safe to target `*` instead of a class:

```css
* {
  scrollbar-color: var(--theme-dim-light) transparent;
}
/* etc. */
```

sarah11918:
Just realized that I can't adequately preview this, because I have a scrollbar extension installed to make mine BIGGER and UGLIER for when I'm using my Chromebook in touch-screen tablet mode. 😆 

So, I will defer to @delucis on this one!

delucis:
I can’t actually reproduce this issue locally either as I think it doesn‘t impact macOS? I already get minimal scrollbars across Firefox, Chrome, Safari, and Edge. Do we have any docs reviewers using Linux or Windows?

So my review is just the suggestion to consolidate the styles that do this, but I can’t actually confirm what they’re doing, because they aren’t doing anything for me 😆

sarah11918:
I guess we'll call @aFuzzyBear in on this one, since none of the rest of us can see it!

aFuzzyBear:
Hey @Yan-Thomas  this looks really good, thank you for doing this.
@delucis  you are right to think about consolidating the styles across the stylesheets, I was thinking could we go one further and target the `::-webkit-scrollbar` to make the master scrollbar the same aesthetically as those that appear on the page

Yan-Thomas:
Thanks everyone!

I agree with consolidating the styles into one file. Regarding changing the master scrollbar, the only thing I would not like is it to only be visible on hover/focus. Maybe we can by default have scrollbar-color/background-color with `--theme-dim-light` and then we override the specific styles on the sidebar components for them to be invisible by default?

Yan-Thomas:
I made one extra change: I made that all the scrollbars are always visible. I did some research and found out hidden scrollbars are bad for UX/accessibility.

public/index.css
```diff
@@ -8,6 +8,26 @@
...
+* {
+	scrollbar-color: var(--theme-dim-light) transparent;
+}
+
+/* Webkit */
+
+::-webkit-scrollbar,
+::-webkit-scrollbar-track {
+	/* same as --theme-bg-gradient but without 'fixed' */
+	background: linear-gradient(180deg, var(--theme-bg-gradient-top), var(--theme-bg-gradient-top) var(--theme-navbar-height), var(--theme-bg-gradient-bottom));
```
delucis, line 21: What does changing this from `transparent` fix? I’m seeing a slightly off scrollbar background, which I think wouldn’t be there if we were using `transparent`:

image

But maybe that’s needed on other platforms?

delucis (commented):
Thanks for you determination, @Yan-Thomas!

I don’t have any strong feelings either way re: always displaying scrollbars, so I’ll let some other folks chime in.

I’ve left one small question, but the the consolidation looks good 👍

sarah11918:
I'm ok if @aFuzzyBear says it looks OK!

public/index.css
Yan-Thomas: What happens is that the main scrollbar looks like this when it is as transparent:
image

I did some research but I didn't find a solution for this, my monitor has low color contrast so I didn't even notice the difference in color, nice catch!

`transparent` still works for the (Right/Left)Sidebar components though, we can override the `background` if we don't find a global solution for this.

aFuzzyBear: @delucis  I dont know if @Yan-Thomas  has fixed this before I had reviewed this, but I cant seem to see this trackbar any more, as a result Im pretty confident that this has conversation has been resolved 🙂

aFuzzyBear (approved):
@Yan-Thomas  👏👏👏👏
This is the PR of the week imo, fantastic work on this.
It looks 😻😻
My favorite part of this is the intersection observer, really do like the modifications you made to it.
Great job 
🎉💚👊

PR #429, opened 2022-05-03
Add note to i18n guide for folks who can’t access Discord

delucis:
Discord is blocked in some places, so this suggests an alternative engagement path for those impacted.


sarah11918:
Great idea. ✅

PR #431, opened 2022-05-03
Feat: Add translation status script and workflow

hippotastic:
Adds a script that creates or updates a special "Translation Status Overview" summary issue on GitHub that provides an overview of the current Astro documentation translation status.

Also adds a GitHub workflow that runs this script on every push to the `main` branch.

An example of the summary issue maintained by this script can be seen here:
https://github.com/hippotastic/astro-docs/issues/2

### DIFF
--- a/scripts/github-translation-status.mjs
+++ b/scripts/github-translation-status.mjs
@@ -347,10 +347,10 @@ class GitHubTranslationStatus {
 			const outdated = arrContent.filter(content => content.translations[lang].isOutdated);
 			lines.push('<details>');
 			lines.push(
-				`<summary><b>` +
+				`<summary><strong>` +
 				`${lang}: ` +
 				`${missing.length} missing, ${outdated.length} needs updating` +
-				`</b></summary>`
+				`</strong></summary>`
 			);
 			lines.push(``);
 			if (missing.length > 0) {
--- a/scripts/github-translation-status.mjs
+++ b/scripts/github-translation-status.mjs
@@ -1,0 +1,474 @@
+import fs from 'fs';
+import glob from 'fast-glob';
+import dedent from 'dedent-js';
+import simpleGit from 'simple-git';
+import issues from './lib/github-issues.mjs';
+import output, { dedentMd } from './lib/output.mjs';
+
+/**
+ * Creates or updates a special summary issue on GitHub that provides an overview of
+ * the current Astro documentation translation status.
+ * 
+ * This code is designed to be run on every push to the `main` branch.
+ */
+class GitHubTranslationStatus {
+	constructor ({
+		pageSourceDir,
+		sourceLanguage,
+		targetLanguages,
+		githubToken,
+		githubRepo,
+		githubRefName,
+		issueTitle,
+	}) {
+		this.pageSourceDir = pageSourceDir;
+		this.sourceLanguage = sourceLanguage;
+		this.targetLanguages = targetLanguages;
+		this.githubToken = githubToken;
+		this.githubRepo = githubRepo;
+		this.githubRefName = githubRefName;
+		this.issueTitle = issueTitle;
+		this.git = simpleGit();
+
+		if (!this.githubToken) {
+			if (output.isCi) {
+				output.error('Missing GITHUB_TOKEN. Please add the following lines to the task:\n' +
+					'    env:\n' +
+					'      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}'
+				);
+				process.exit(1);
+			} else {
+				output.warning('You are not running this script from a GitHub workflow. ' +
+					'Performing a dry run without actually updating the issue.');
+			}
+		}
+	}
+
+	/**
+	 * Performs the following steps:
+	 * 
+	 * - Update the state based on the current contents of the repo:
+	 *   - Detect if any pages existing in our index were changed:
+	 *     - Determine which commits were made to the repo since the previous head SHA
+	 *       (which was stored in the state during the last run of this script)
+	 *       - During this check, ignore commits marked as "minor" or "typo"
+	 *     - If any of these commits affected pages contained in our index,
+	 *       update the state to remember when these pages were changed
+	 * - Update the GitHub issue with the new human-friendly summary and JSON metadata
+	 */
+	async update () {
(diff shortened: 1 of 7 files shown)


hippotastic:
Would love you to have a look at this @delucis & @sarah11918 😄

hippotastic:
Note: If you want to test this locally outside of GitHub, you will need to set the env variable `GITHUB_REPOSITORY` to `withastro/docs`. This variable is normally set by GitHub workflows. Then you can run `pnpm run github:translation-status` to see what it would do. It will perform a dry run in this case and not create an issue because there is no `GITHUB_TOKEN` (which also gets set by the workflow), but you can see the generated state data and human-friendly issue summary.

sarah11918:
Looks wonderful, @hippotastic ! I'd be super happy with this as a display for our translators!

scripts/lib/github-issues.mjs
```diff
@@ -0,0 +1,230 @@
+/**
+ * Note: This module is only here to avoid including octokit.js with hundreds of dependencies,
+ * which would slow down the install task of our GitHub workflows.
+ * 
+ * If we want to switch to octokit.js at one point, we should make this code independent from
+ * our scripts by creating separate GitHub actions with their own repos and compilation steps.
+ */
+
+import fetch from 'node-fetch';
```
delucis, line 9: Wondering if we could run this on Node 18 and use the new `fetch` global? This is fine too though.

scripts/github-translation-status.mjs
```diff
@@ -0,0 +1,462 @@
...
+		// if we do not want a translated page to become outdated) and the actual commit messages
+		pageData.lastChange = latest(oldPageData.lastChange, gitHistory.lastCommitDate);
+		pageData.lastCommitMsg = gitHistory.lastCommitMessage;
+		pageData.lastMajorChange = latest(oldPageData.lastMajorChange, gitHistory.lastMajorCommitDate);
+		pageData.lastMajorCommitMsg = gitHistory.lastMajorCommitMessage;
+		
+		return pageData;
+	}
+
+	tryGetFrontMatterBlock (filePath) {
```
delucis, line 281: Not sure if this would simplify your life much, but could you use [`gray-matter`](https://www.npmjs.com/package/gray-matter) here? That’s how Astro parses frontmatter, so it’s already a dependency.

scripts/github-translation-status.mjs
```diff
@@ -0,0 +1,462 @@
...
+		const arrContent = this.getTranslationStatusByContent({ pages });
+		const lines = [];
+
+		this.targetLanguages.forEach(lang => {
+			const missing = arrContent.filter(content => content.translations[lang].isMissing);
+			const outdated = arrContent.filter(content => content.translations[lang].isOutdated);
+			lines.push('<details>');
+			lines.push(
+				`<summary><b>` +
+				`${lang}: ` +
```
delucis, line 351: It would be cool to display the full language name here and not just the two-letter tag, but our languages dictionary is a TypeScript file, so it’s a bit of a pain to import in this context. Maybe we could just maintain the list of language names manually here like we do the list of target languages? But this is a cosmetic detail, so I’m happy to see this merged and have this improved upon at some point in the future.

scripts/github-translation-status.mjs
```diff
@@ -0,0 +1,462 @@
...
+				`${missing.length} missing, ${outdated.length} needs updating` +
+				`</b></summary>`
```
delucis, line 353: (Can you tell I’m nitpicking because I can’t think of anything else to say? 😅)

`<strong>` conveys something semantically unlike `<b>`. I’d suggest `<h4>`, but I see that introduces quite a bit of excess space between each item, but maybe that’s OK? Up to you.

```suggestion
				`<summary><strong>` +
				`${lang}: ` +
				`${missing.length} missing, ${outdated.length} needs updating` +
				`</strong></summary>`
```

scripts/github-translation-status.mjs
```diff
@@ -0,0 +1,462 @@
...
+			);
+			lines.push(``);
+			if (missing.length > 0) {
+				lines.push(`##### Missing`);
+				lines.push(...missing.map(content =>
+					`- [${content.subpath}](${content.githubUrl})`
```
delucis, line 359: Bonus: would it be helpful to offer a GitHub create link here?

For example:

`https://github.com/withastro/docs/new/main/src/pages/{LANG}?filename={PAGE/SLUG.md}`

The `filename` param seems to support `/` to create subdirectories, so you just need the language and then the full slug including directories if present can be passed to `filename`.

It’s even possible to pre-fill the body:

```js
const createUrl = new URL(`https://github.com/withastro/docs/new/main/src/pages/${lang}`);
createUrl.searchParams.set('filename', content.subpath);
createUrl.searchParams.set('value', `---
layout: ~/layouts/MainLayout.astro
title:
description:
---
`);
```

delucis (commented):
Absolutely stunning work @hippotastic! 🏆 

Really struggled to say anything of value here. I have made some suggestions, but I think basically all of this could just be noted down as possibilities for polish in the future. Happy to see this merged as is 👍

hippotastic:
Amazing review with lots of valuable feedback and ideas @delucis, thank you so much! I'd love to address your feedback after merging this PR though to get this into production asap so we can start monitoring the behavior on the "real thing" right away. I'd really be happy to keep improving this!

scripts/github-translation-status.mjs
hippotastic: This is an amazing idea!

scripts/github-translation-status.mjs
```diff
@@ -0,0 +1,462 @@
...
+				return;
+			const lang = pathParts[0];
+			const subpath = pathParts.splice(1).join('/');
+
+			// Ignore pages with languages not contained in our configuration
+			if (!pages[lang])
+				return;
+
+			// Attempt to get old data for the current page from our index
+			const oldPageData = { ...(oldPages[lang] && oldPages[lang][subpath]) };
```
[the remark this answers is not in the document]
hippotastic: This is actually not an issue. You can spread both `false` and `undefined` and it will just behave like an empty object. :)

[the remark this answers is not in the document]
delucis: 🤯

PR #432, opened 2022-05-04
Need to revert to HTML elements to use ImportMetaEnv component

sarah11918:
Thanks to @delucis comment in Docs channel, I realized that we used the ImportMetaEnv component for the new additions to the Environment Variables page, but did NOT wrap these in HTML elements.

This further necessitated using HTML instead of Markdown to create the list, the link, all the inline code snippets etc.

I'm inferring the desired code rendering pattern is:
 `<code>import.meta...</code> (<code>Boolean</code>):` (with a separate inline code inside the brackets)

If not, easy to fix. 

So, please check the Netlify deploy preview, because this one is all about how it looks on te page!

### DIFF
--- a/src/pages/en/guides/environment-variables.md
+++ b/src/pages/en/guides/environment-variables.md
@@ -24,14 +24,14 @@ In this example, <code>PUBLIC_ANYBODY</code> (accessible via <ImportMetaEnv path
 
 Astro includes a few environment variables out-of-the-box:
 <ul>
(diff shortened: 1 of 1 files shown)


aFuzzyBear (approved):
This looks like the its working properly now... great catch buddy 👊

delucis:
Whoops guess I missed those. Thanks for catching this!

aFuzzyBear:
Sarah and her hawkeye vision,

PR #434, opened 2022-05-04
Document API route signature changes

matthewp:
Depends on https://github.com/withastro/astro/pull/3272


sarah11918:
(Gah, had the wrong window open).

@matthewp This all looks great when it's time to merge!

My only question is, looking at the page itself, do you think visually the SSR stuff is separated/stands out enough from the non-SSR stuff? I'll leave that to be your call. Just wanted to make sure that it's also functioning for people who might be scanning the docs.

matthewp:
@sarah11918 That's a good call out! I sort of feel that eventually we should peel out the API Route docs into their own top-level Core Concept outside of Pages. There I think there would be more breathing room for separately talking about SSR and SSG.

PR #437, opened 2022-05-04
Add API reference docs for Astro.response

matthewp:
Implementation: https://github.com/withastro/astro/pull/3289


sarah11918:
LGMT.

delucis:
I’m merging this seeing as the feature has shipped and it would be good to have some docs out there. We can iterate on this in the future if we want to tweak anything.

PR #441, opened 2022-05-04
i18nReady: true for first batch of pages

sarah11918:
Set first batch of mostly core pages as i18nReady!

These pages should now be automatically listed at https://github.com/withastro/docs/issues/438 for translator reference.

Thanks, @hippotastic !


delucis (approved):
Let’s merge and find out!

PR #443, opened 2022-05-05
Improve toc indicator precision and add auto-scroll

Yan-Thomas:
The table of contents indicator logic received some changes to be more precise and now "on this page" auto-scrolls, this way the indicator is always visible.

- Auto-scroll (look how new toc links appear as you scroll down the page):
KuKYDum

- Precision comparison between the old and new logic (Getting Started):

Old
wDRNvrb

New
RVSMZXZ 

*Yes, I do love sidebars.*

### DIFF
--- a/src/components/RightSidebar/TableOfContents.tsx
+++ b/src/components/RightSidebar/TableOfContents.tsx
@@ -19,28 +19,61 @@ const TableOfContents: FunctionalComponent<Props> = ({ headers = [], labels }) =
 		if (!toc.current) return;
 
 		const setCurrent: IntersectionObserverCallback = (entries) => {
+			const intersectingEntries = [];
+
 			for (const entry of entries) {
-				if (entry.isIntersecting) {
-					setCurrentID(entry.target.id);
-					break;
-				}
+				if (entry.isIntersecting) intersectingEntries.push(entry);
+			}
+
+			if (intersectingEntries.length === 1) {
+				setCurrentID(intersectingEntries[0].target.id);
+				scrollIntoView();
+			} else if (intersectingEntries.length > 1) {
+				const sortedEntries = intersectingEntries.sort((a, b) => {
+					return getHeadingIndex(a.target.id) - getHeadingIndex(b.target.id);
+				});
+				setCurrentID(sortedEntries[0].target.id);
+				scrollIntoView();
 			}
 		};
 
 		const observerOptions: IntersectionObserverInit = {
 			// Negative top margin accounts for `scroll-margin`.
 			// Negative bottom margin means heading needs to be towards top of viewport to trigger intersection.
-			rootMargin: '-100px 0% -66%',
+			rootMargin: '-100px 0px -40% 0px',
 			threshold: 1,
 		};
 
 		const headingsObserver = new IntersectionObserver(setCurrent, observerOptions);
 
-		// Observe all the headings in the main page content.
-		document.querySelectorAll('article :is(h1,h2,h3)').forEach((h) => headingsObserver.observe(h));
+		// removes the headings from "on this page" on mobile using [id] selector.
+		const headings = document.querySelectorAll('article :is(h1,h2[id],h3)');
 
-		// Stop observing when the component is unmounted.
-		return () => headingsObserver.disconnect();
+		const getHeadingIndex = (id) => Array.from(headings).findIndex((heading) => heading.id === id);
+
+		let scrollPosition = 0;
+		let scrollDirection = 10;
+
+		const setScrollDirection = () => {
+			if (document.body.getBoundingClientRect().top > scrollPosition) scrollDirection = -10;
+			else scrollDirection = 10;
+			scrollPosition = document.body.getBoundingClientRect().top;
+		};
+
+		window.addEventListener('scroll', setScrollDirection);
+
+		const scrollIntoView = () => {
+			const scrollingContext = document.querySelector('.sidebar-nav-inner');
+			if (window.matchMedia('(prefers-reduced-motion)')) scrollingContext.scrollBy(0, scrollDirection);
+		};
+
+		// Observe all the headings in the main page content.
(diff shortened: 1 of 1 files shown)


aFuzzyBear (commented):
Looks pretty decent to me Yan, a good late night post, 
Thank you for these... 👊

src/components/RightSidebar/TableOfContents.tsx
```diff
@@ -19,28 +19,61 @@ const TableOfContents: FunctionalComponent<Props> = ({ headers = [], labels }) =
...
+				setCurrentID(sortedEntries[0].target.id);
+				scrollIntoView();
 			}
 		};
 
 		const observerOptions: IntersectionObserverInit = {
 			// Negative top margin accounts for `scroll-margin`.
 			// Negative bottom margin means heading needs to be towards top of viewport to trigger intersection.
-			rootMargin: '-100px 0% -66%',
+			rootMargin: '-100px 0px -40% 0px',
```
aFuzzyBear, line 43:
```suggestion
			rootMargin: '-100px 0px -40% 0px',
```
After a bit of playing, it appears a wee bit off, I'd need to spin this up and test it properly,

aFuzzyBear (changes requested):
Intersection Obs needs a wee bit of fine tuning...

src/components/RightSidebar/TableOfContents.tsx
```diff
@@ -19,28 +19,61 @@ const TableOfContents: FunctionalComponent<Props> = ({ headers = [], labels }) =
...
+			else scrollDirection = 10;
+			scrollPosition = document.body.getBoundingClientRect().top;
+		};
+
+		window.addEventListener('scroll', setScrollDirection);
+
+		const scrollIntoView = () => {
+			const scrollingContext = document.querySelector('.sidebar-nav-inner');
+			if (window.matchMedia('(prefers-reduced-motion)')) scrollingContext.scrollBy(0, scrollDirection);
+		};
```
delucis, line 68: Wondering if we could simplify all this with the `scrollIntoView` browser API: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

Need to double check, but I bet we could use a `ref`, pass it to the current list item, and then scroll it each time it changes in another `useEffect`. Anything that avoids the `scroll` event seems like a good idea if possible.

(Guess we'd need to disable on mobile to avoid scrolling up to the table of contents.)

delucis (changes requested):
Thanks for trying to improve this @Yan-Thomas. You're really becoming our sidebar guy!

I don't think this is quite there yet and I've left a couple of comments. Let me know if you want to bounce ideas around or have any questions.

src/components/RightSidebar/TableOfContents.tsx
Yan-Thomas: I have tried to use `scrollIntoView`, it was my first attempt at implementing this, but it only scrolls the page scrollbar, not the sidebar one, so it doesn't help us in this case.

Yan-Thomas:
I did some changes trying to polish it. Let me know if something still sounds off to you guys!

src/components/RightSidebar/TableOfContents.tsx
```diff
@@ -19,28 +19,61 @@ const TableOfContents: FunctionalComponent<Props> = ({ headers = [], labels }) =
...
 			}
 		};
 
 		const observerOptions: IntersectionObserverInit = {
 			// Negative top margin accounts for `scroll-margin`.
 			// Negative bottom margin means heading needs to be towards top of viewport to trigger intersection.
-			rootMargin: '-100px 0% -66%',
-			threshold: 1,
+			rootMargin: '-100px 0px -66% 0px',
+			threshold: [0.75, 1],
```
delucis, line 44: Interesting. What did you find the second `threshold` value helped with?

I’m seeing the highlighted menu item jump back and forth now, presumably as each heading crosses its second threshold:

https://user-images.githubusercontent.com/357379/167094798-be238cab-3889-41a6-8b27-ddb08790798d.mov

Yan-Thomas: I'm actually confused right now 😅. For me any of these problems didn't happen! I added the second threshold value to maintain the heading on top selected even if it isn't totally visible but most part of it is.

Maybe removing the `0.75` from the threshold solves this? I think if I make any changes I can't properly test it since I didn't had the same problems. 

Can you guys test changing some values (probably threshold and bottom margin?) so we can iterate over what looks best?

src/components/RightSidebar/TableOfContents.tsx
delucis: Ah, you’re probably right! I think we could probably still move this logic to only run on intersection somehow rather than tracking scroll events/direction. Like: if a heading is current and it’s not in the viewport, scroll the parent container.

src/components/RightSidebar/TableOfContents.tsx
delucis: Haha, yeah this stuff is tricky! This only shows up I think when two headings are close together so 0.75 for one is passed before 1 for the other. For me just `1` on its own works well?

sarah11918:
Hey @aFuzzyBear , do you have any spare attention to give to this again today? :green_heart:

Yan-Thomas:
I think it should be better if we close this PR. I have been working on this for quite some time now and it really is hard to improve this in a way that works well for users with different screen sizes. Also, other approaches I found seem overkill and/or bad for performance. If you guys think the auto-scroll is really valuable I can make a different PR just for this feature, or just edit this PR.

sarah11918:
OK, then we'll shorten our PR list, but are always happy to have your ideas and contributions when you have them! Maybe you just need another trip to the showers!  ;)

I'll close this for now then, but RightSidebar is never forgotten! 💜

delucis:
Yeah thanks again for your dedication here @Yan-Thomas! 🙌

PR #444, opened 2022-05-05
i18n(pt-BR): Updated Getting Started and UI/Docsearch.

Yan-Thomas:
Updates the pt-BR getting started page and `docsearch.ts` and `ui.ts` texts.

Feel free to review and discuss changes here!

### DIFF
--- a/src/i18n/pt-BR/ui.ts
+++ b/src/i18n/pt-BR/ui.ts
@@ -1,6 +1,15 @@
 import { UIDictionary } from '../translation-checkers';
 
 export default UIDictionary({
+	'a11y.skipLink': 'Pular para o Conteúdo',
+	'navbar.a11yTitle': 'Navegação Superior',
+	'site.title': 'Documentação do Astro',
+	'site.description': 'Construa websites mais rápidos com menos Javascript no lado do cliente.',
+	'site.og.imageSrc': '/default-og-image.png?v=1',
+	'site.og.imageAlt': 'a logo do astro em um estrelado pedaço do espaço, com um planeta roxo parecido com Saturno flutuando à direita',
+	'leftSidebar.a11yTitle': 'Navegação do site',
+	'leftSidebar.noTranslations': 'Não foram encontradas traduções',
+	'leftSidebar.viewInEnglish': 'Veja em Inglês',
 	'leftSidebar.learnTab': 'Aprenda',
 	'leftSidebar.referenceTab': 'Referência',
 	'rightSidebar.onThisPage': 'Nesta página',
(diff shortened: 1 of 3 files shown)


src/i18n/pt-BR/ui.ts
```diff
@@ -1,6 +1,15 @@
 import { UIDictionary } from '../translation-checkers';
 
 export default UIDictionary({
+	'a11y.skipLink': 'Pular para o Conteúdo',
+	'navbar.a11yTitle': 'Navegação Superior',
+	'site.title': 'Documentação do Astro',
+	'site.description': 'Construa websites mais rápidos com menos Javascript no lado do cliente.',
```
caioferrarezi, line 7: What if we add a conector like this? Maybe it would feel more natural:
```suggestion
	'site.description': 'Construa websites mais rápidos e com menos Javascript no lado do cliente.',
```

src/i18n/pt-BR/ui.ts
```diff
@@ -9,8 +18,17 @@ export default UIDictionary({
 	'rightSidebar.editPage': 'Edite esta página',
 	'rightSidebar.translatePage': 'Traduza esta página',
 	'rightSidebar.joinCommunity': 'Junte-se à comunidade',
+	'rightSidebar.a11yTitle': 'Índice',
 	'articleNav.nextPage': 'Próxima página',
 	'articleNav.prevPage': 'Página anterior',
 	'since.addedIn': 'Adicionado em:',
 	'since.new': 'Novo',
+	'install.autoTab': 'Interface de Linha de Comando Automática',
+	'install.manualTab': 'Instalação Manual',
```
caioferrarezi, line 27: Should we translate the word `setup` as `Configuração`?

src/pages/pt-BR/getting-started.md
```diff
@@ -1,52 +1,120 @@
...
+
+
+> Tem um projeto antigo? Siga o [guia de migração](/pt-BR/migrate) para atualizar para a v1.0 beta!
 
-Astro é um moderno construtor de site estático. Aprenda mais sobre Astro através da [nossa página inicial](https://astro.build/) ou do [nosso post de lançamento](https://astro.build/blog/introducing-astro). Esta página é uma visão geral da documentação do Astro e de todos os recursos relacionados.
 
 ## Experimente Astro
 
-A maneira mais fácil de experimentar Astro é através da execução do comando `npm create astro@latest` em um novo diretório na sua máquina. Nosso assistente para interface de linha de comandos (Wizard CLI) irá te ajudar a iniciar um novo projeto Astro.
+Nós fizemos com que fosse o mais fácil possível dar os seus primeiros passos com o Astro seja no navegador ou em sua máquina!
```
caioferrarezi, line 18:
```suggestion
Nós fizemos com que fosse o mais fácil possível dar os seus primeiros passos com o Astro, seja no navegador ou em sua máquina!
```

src/i18n/pt-BR/ui.ts
Yan-Thomas: Maybe we should change the wording a bit? I want to be clear that the website is faster because we're sending less Javascript to the client-side, if we add `e` it looks a bit like it's two separate things and not the reason why it is faster.

src/i18n/pt-BR/ui.ts
Yan-Thomas: In this specific case I think `Instalação` works better since this tab is about installing Astro manually, but yeah, sometimes `Configuração` is better. I just tend to associate it more with changing settings than with adding/installing/creating something and because of that, I opted for `Instalação`.

Thanks for all the reviews ❤️

src/i18n/pt-BR/ui.ts
caioferrarezi: Nice, I agree!

src/i18n/pt-BR/ui.ts
caioferrarezi: I understand! `Instalação` seems to fit better in this case as the first step to obtain Astro. I was asking because I think it's important to find out a consistent translation throughout the docs, I'll keep my head up to your explanation as I work on the other pages!

Thanks!

caioferrarezi (approved):

delucis:
I see @caioferrarezi approved these changes @Yan-Thomas. Are they ready to merge?

delucis:
Thanks for your work on this! Launching i18n pt-BR edition 🚀

PR #446, opened 2022-05-05
Polish to the automated i18n issue

delucis:
This PR adds a few little details to our new automated i18n tracker.

You can see a live preview of the changes here: https://github.com/delucis/docs/issues/1

- Adds a “Create page” link next to missing content:
   image
   This links to a GitHub UI for creating a new file with the correct name and pre-fills it with the basic Markdown frontmatter

- Displays each language’s full name in the todos list:
   image

- Fixes the i18n guide link in the intro (h/t @hippotastic)

- Adds the emoji to the subheadings in each language’s todo section
   image image

PR #447, opened 2022-05-05
Prevent create page button breaking across lines

delucis:
Fixes mobile rendering of the “Create page” buttons in the i18n tracking issue. #446 added these but used normal spaces in the button text causing it to wrap across lines on narrow viewports. This replaces those with non-breaking spaces so the whole button always wraps together.


hippotastic:
The diff on GitHub looks exactly the same, but I simply believe you that these shiny new spaces are 100% non-breaking! LGTM

PR #451, opened 2022-05-05
Make GitHub link in sidebar translatable & fix `UIString` typings

delucis:
Quick fix for a bit of UI text @Yan-Thomas spotted didn’t yet support translation.

While doing it I also noted I’d broken the typings for the `UIString` component during the big i18n overhaul, so patched that up.


sarah11918:
LGTM!

PR #453, opened 2022-05-05
Edit: make RSS canonicalUrl required

bholmesdev:
We discovered that `canonicalUrl` did _not_ work as expected in the `@astrojs/rss` package. So, we're asking users to apply this value manually as a required prop.

**Changes made**
- rename `canonicalUrl` option to `site`
- add note that `import.meta.env.SITE` only works in the latest 1.0 beta
- add `site` to all code examples, with explainer comments where applicable

### DIFF
--- a/src/pages/en/guides/rss.md
+++ b/src/pages/en/guides/rss.md
@@ -2,8 +2,6 @@
 layout: ~/layouts/MainLayout.astro
 title: RSS
 description: An intro to RSS in Astro
-setup : |
-  import ImportMetaEnv from '~/components/ImportMetaEnv.astro';
 ---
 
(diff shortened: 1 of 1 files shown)


bholmesdev:
@sarah11918 Deploy preview looks fine to me! Shouldn't be a problem in a code fence from my understanding, but lmk if there's something crazy I don't know about for production

sarah11918:
@bholmesdev Gah! I replied in the other PR by mistake... short version, this is what this current Netlify deploy preview looks like:

Can you look carefully at this? Because this is what I see:

When it's in commented out code in the code fence, you're gonna need that component (as I gave instructions for dealing with, in the other PR):
Screenshot 2022-05-05 18 35 05

When it's just a string in the code fence, it seems to be fine:
Screenshot 2022-05-05 18 39 03

bholmesdev:
Totally understood! Thanks for walking through this nuanced / wacky issue @sarah11918 🙏

PR #454, opened 2022-05-05
import and use ImportMetaEnv component

sarah11918:
@bholmesdev - please check this, and adjust as necessary to get it to render properly.  This might require fiddling, but the ideas are:

1. Where ever you're seeing `{{}}` in the preview render instead of `import.meta.env`, then you need to import (I've done) and use (I've done) this component instead of typing the phrase out directly. The rest of the expression can be passed as the`path=" "` attribute, and it will be formatted as code.

2. This component must be inside an HTML element, and can't be written inside regular Markdown. I've tried the simplest version first (just `<p>` tags enclosing the single line it's in). If this isn't sufficient, or screws up the rest of the rendering, then you'll have to "work your way out" converting larger and larger chunks to HTML until you have the rendering you need.

(Or, you figure out whether it's ABSOLUTELY NECESSARY to write the phrase import.meta.env. Dealer's choice.) Hope this helps.

### DIFF
--- a/src/pages/en/guides/rss.md
+++ b/src/pages/en/guides/rss.md
@@ -2,6 +2,8 @@
 layout: ~/layouts/MainLayout.astro
 title: RSS
 description: An intro to RSS in Astro
+setup : |
+  import ImportMetaEnv from '~/components/ImportMetaEnv.astro';
 ---
 
 Astro supports fast, automatic RSS feed generation for blogs and other content websites. For more information about RSS feeds in general, see [aboutfeeds.com](https://aboutfeeds.com/).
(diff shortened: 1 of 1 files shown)


sarah11918:
@bholmesdev Of the 3 places it's typed, this is the only one that IS a problem in the deploy preview, as far as I can tell? Can you check this really closely?

Screenshot 2022-05-05 18 35 05

sarah11918:
The two other places, in a code fence but yet not commented out ( 🤪 ) seem to be fine:

Screenshot 2022-05-05 18 39 03

PR #455, opened 2022-05-05
Adding missing slash before link to avoid wrong linking

Chrissdroid:
Adding a small slash to avoid being redirected to `https://docs.astro.build/en/core-concepts/framework-components/en/reference/directives-reference/#client-directives` and not `https://docs.astro.build/en/reference/directives-reference/#client-directives`


delucis:
Just a note that we *should* have included “broken link” in this commit message to avoid the i18n automation treating this as a “major” change. Shouldn’t matter because this page hasn’t been translated yet, but flagging so we remember (hopefully!) for the future.

sarah11918:
Oops... I guess I need to consult our guide...

PR #457, opened 2022-05-05
Adding slashes for tags when needed according to HTML's spec

Chrissdroid:
Updating files from the `/en/` docs regarding self closing tags technically wrong (according to spec) when specified in the docs

- changing `<slot>` to `<slot />`
- ~changing `` to ``~ — reverted
- changing `<Markdown>` to `<Markdown />`
- ~changing `<meta>` to `<meta />`~ — reverted

(contain also change for `migrate.md` removing useless spaces at end of line, (ty my IDE))

### DIFF
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -9,7 +9,7 @@ i18nReady: true
 
 Astro component syntax is a superset of HTML. The syntax was [designed to feel familiar to anyone with experience writing HTML or JSX](/en/comparing-astro-vs-other-tools/#astro-vs-jsx), and adds support for including components and JavaScript expressions. You can spot an Astro component by its file extension: `.astro`.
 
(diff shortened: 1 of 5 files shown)


sarah11918:
I don't think the `<meta>` and `` tags should have a closing slash. But the components (slot, Markdown) definitely should!

Chrissdroid:
When I was browsing for information on this, it said that img and meta *has* to have an self closing according to the spec

But it's from the XML part of the HTML spec so both are "suggested" as correct

I can revert these changes if you prefer the other syntax

delucis:
I’ll leave the final call on this to you @sarah11918. I’m OK either way — I agree that this is most important for the Astro-specific components.

sarah11918:
Well, how's this for the basis of a decision?  Our new basics starter uses `<meta>` tags.  :)

So, I'm going to say let's only change the components for now, and we'll work on a proper style gude where we can make these decisions, and make them available to everyone.

sarah11918:
@Chrissdroid For now, we're only going to use trailing slashes in components, not HTML elements. But, we'll pick something for the style guide. So, I'll merge this when you take them out of img and meta, please.  :)

Chrissdroid:
I'll do !

delucis (approved):
LGTM!

PR #458, opened 2022-05-05
Lighthouse audits

sarah11918:
This page reruns (as of 2022-05-05) Lighthouse audits for all "Astro vs X" comparisons and:

- uploads and now links to an individual `public/lighthouse/.html` file for each performance audit, since it's no longer possible to link to a URL for a particular test result
- uploads but does NOT link to .html pages for the other comparisons that previously did NOT have a doc site test comparison (except Elder, because they don't seem to have a separate docs site outside of GitHub repo).
- reports the current Performance number on the page for each comparison (only the ones that previously reported one)
- comments out any reference to "only ships xxx kB of JS" in each reference, but otherwise leaves the comparison text the same.

TODO (for this, or in future):
- decide which, if any other measures we want to report to replace the commented out text. (Total Bytes? Total Blocking Time? Starts rending in x seconds? --> these are all things we can collect from webpagetest.org if we want.)
- review/edit any existing prose/text (not done at this point)
- further to discussions in docs channel, think about WHETHER we want to continue to support this page as-is, or whether we prefer a different kind of comparison page (some examples and ideas posted in #docs channel)


delucis:
Smart! Good thinking on including those Lighthouse reports that way.

I think I’d include the Lighthouse score comparison for 11ty, Gatsby & Hugo seeing as you’ve done the hard work running those? Doesn’t do any harm. And I’m good to merge otherwise and revisit your other questions as part of a broader discussion?

sarah11918:
Looping @FredKSchott in on this, so he can see the baby steps we've taken.

Again, I think this is safe to merge, as long as Fred's OK with the concept, since it just updates existing performance scores. Then we tackle what to REALLY do with this page going forward.

sarah11918:
OK, for the sake of updated-ness/accuracy, I'm going to merge this, but on the stack is how to best handle this page going forward!

delucis:
No adding the Lighthouse score for the sites missing them currently?

sarah11918:
ALRIGHT, CHRIS!  ;)

I was getting lazy... it's more than just adding the scores, it's copying "Real World example..." to create that entire section for these that weren't done before. But yeah, I can do that.  💜

delucis:
Want me to if you’re feeling lazy? It IS Friday.

delucis (approved):
No mis-slander here! ✅

sarah11918:
OK, thanks! Then NOW I'll merge Astro vs X, now, with more comparisons!

All the original TODOs still apply for the future...

PR #466, opened 2022-05-06
Fix a heap of broken links

delucis:
This PR fixes _almost_ every single broken link reported by our linter. The only exception are three on the old Spanish getting started page given that’s about to be replaced by #461.

These links fall into four categories:

1. The Lighthouse HTML files in `public/`. Files in `public/` don’t end up in the sitemap so our linter was erroneously reporting links to these pages as missing. I’ve converted these to Astro files in `src/pages` to fix that.

2. Broken links on `en/` pages. There were a handful of these. Mostly `#` links where some heading text changed breaking the link.

3. Broken links on translated pages. Almost all of these were links on out-of-date getting started pages to `/installation` or `/quick-start` which no longer exist. I’ve replaced these with the new URLs.

4. A couple of broken links on our _new_ translated pages. Both `es` & `pt-BR` install guides didn’t realise the need to translate a `#` link. All the more reason to get that CI turned on, so we catch that kind of thing at the PR stage.


sarah11918:
I only carefully looked at the /en/ pages, knowing the others will soon disappear, but LGTM! Thanks for solving my Lighthouse issues @delucis !

delucis:
Great! And I think because this says “broken links” we should be fine on that front too. Otherwise it’ll be another trip to Dr. Hippo.

PR #476, opened 2022-05-07
Feat: Add link checker to CI workflow

hippotastic:
This adds our link checker to the CI workflow. It shows up independently from the slug checker in the PR & push checks.

If any broken links are found, it returns an error code, which causes the CI check to show a red cross. It also attempts to find the Markdown source files containing the broken links and annotates the lines containing the broken links. The annotations can be seen in the "changed files" view of PRs and commits.

### DIFF
--- a/scripts/lint-linkcheck.mjs
+++ b/scripts/lint-linkcheck.mjs
@@ -235,7 +235,8 @@ class BrokenLinkChecker {
 
 		// Always output a summary first because GitHub only displays the first 10 annotations
 		const totalLines = annotations.length;
-		core.error(`Found ${totalLines} ${totalLines === 1 ? 'line' : 'lines'} containing broken links in Markdown sources, please check changed files view`);
+		core.error(`Found ${totalLines} ${totalLines === 1 ? 'line' : 'lines'} containing ` +
+			`broken links in Markdown sources, please check changed files view`);
 
 		// Now output all line annotations
 		annotations.forEach(annotation => core.error(annotation.message, annotation.location));
--- a/scripts/lint-linkcheck.mjs
+++ b/scripts/lint-linkcheck.mjs
@@ -2,14 +2,16 @@ import path from 'path';
 import fs from 'fs';
 import kleur from 'kleur';
 import htmlparser2 from 'htmlparser2';
+import core from '@actions/core';
 
 /**
  * Contains all link checking logic.
  */
 class BrokenLinkChecker {
-	constructor ({ baseUrl, buildOutputDir }) {
+	constructor ({ baseUrl, buildOutputDir, pageSourceDir }) {
 		this.baseUrl = baseUrl;
 		this.buildOutputDir = buildOutputDir;
+		this.pageSourceDir = pageSourceDir;
 	}
 
 	/**
@@ -26,8 +28,19 @@ class BrokenLinkChecker {
 		// Find all broken links
 		const brokenLinks = this.findBrokenLinks(pages);
 
-		// Output the result
+		// Output the result to the console
 		this.outputResult(brokenLinks);
+
+		if (brokenLinks.length > 0) {
+			// If we're being run by a GitHub CI workflow, try to output annotations
+			// that show the locations of the broken links in the source files
+			if (process.env.CI) {
+				this.outputSourceFileAnnotations(brokenLinks);
+			}
+
+			// Let our caller know that we found errors
+			process.exitCode = 1;
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


hippotastic:
PTAL @delucis

scripts/lint-linkcheck.mjs
```diff
@@ -173,6 +187,61 @@ class BrokenLinkChecker {
 		console.log();
 	}
 
+	outputSourceFileAnnotations (brokenLinks) {
+		const annotations = [];
+
+		// Collect all unique pathnames that had broken links
+		const pathnames = [...new Set(brokenLinks.map(brokenLink => brokenLink.page.pathname))];
```
delucis, line 194: Nit, but I don't think you need to spread the Set?

```suggestion
		const pathnames = new Set(brokenLinks.map(brokenLink => brokenLink.page.pathname));
```

scripts/lint-linkcheck.mjs
```diff
@@ -181,12 +250,55 @@ class BrokenLinkChecker {
...
+	 * Prevents false positive partial matches (like an href of `/en/install` matching
+	 * an input containing `/en/install/auto`) by requiring the characters surrounding a match
+	 * not to be a part of URLs in Markdown.
+	 */
+	indexOfHref (input, href) {
+		let i = input.indexOf(href);
+		while (i !== -1) {
+			// Get the characters surrounding the current match (if any)
+			let charBefore = input.substring(i - 1, i);
+			let charAfter = input.substr(i + href.length, 1);
```
delucis, line 285: `substr` is deprecated I believe. I'd suggest using `slice` for both of these but `substring` works too.

scripts/lint-linkcheck.mjs
```diff
@@ -181,12 +250,55 @@ class BrokenLinkChecker {
...
+			// If both characters are not a part of URLs in Markdown,
+			// we have a proper (non-partial) match, so return the index
+			if ((charBefore + charAfter).match(/^[\s"'()[\],.]*$/))
```
delucis, line 288: This wouldn't match an absolute URL right? So if someone linked to `https://docs.astro.build/404`, we'd fail to annotate it.

delucis (commented):
On the whole this looks really solid! I've left a couple of small comments.

Out of curiosity, did you do any test runs to see what the annotations look like? I'm curious to see.

scripts/lint-linkcheck.mjs
hippotastic: The regex is there to check the character immediately before and after the broken link, not the broken link URL itself. So if your broken link is contained in a markdown file like `See the [Install Guide](https://docs.astro.build/404) for more details.`, the regex will look at the opening and closing round brackets and succeed, making it properly annotate it.

The only reason this check is in there is to prevent partial URL matches from occuring in which case it would annotate lines that don't contain an error. For example, if `https://docs.astro.build/en/install` is a broken link, and `https://docs.astro.build/en/install/auto` is not, we do not want to annotate the perfectly valid `https://docs.astro.build/en/install/auto` just because a substring matches the broken link.

delucis: Ahhh, OK. It's just that the logging for the checker always outputs the full absolute URL, so I was assuming you then were converting that to a path. (So in my example I thought you might be looking for `/404` when a file actually includes the absolute URL.)

scripts/lint-linkcheck.mjs
hippotastic: Interesting, I didn't know about the deprecated status. I'm actually going to use the array-style accessors then because I'm only interested in one character max each, and this perfectly captures that:
```js
let charBefore = input[i - 1] || '';
let charAfter = input[i + href.length] || '';
```

scripts/lint-linkcheck.mjs
hippotastic: Haha, good eye! Yeah, I wouldn't need to spread it in this case. It's just a personal preference to keep everything arrays in my code as I don't have to think twice if I need to access `.length` or `.size`, or if it's `.push` or `.add`.

scripts/lint-linkcheck.mjs
delucis: Oh yeah, good idea!

delucis:
Looks good to me! Any reason we should wait before merging?

hippotastic:
I added the results of our discussion to the "minor changes" commit. I sent you a screenshot how the annotations look like on Discord. Think we should be ready to merge now. :)

PR #479, opened 2022-05-08
docs: add shim config option, add missing args docs

thepassle:
depends on https://github.com/withastro/astro/pull/3317

### DIFF
--- a/src/pages/en/reference/adapter-reference.md
+++ b/src/pages/en/reference/adapter-reference.md
@@ -42,7 +42,9 @@ The object passed into `setAdapter` is defined as:
 interface AstroAdapter {
 	name: string;
 	serverEntrypoint?: string;
+	shim?: string[];
 	exports?: string[];
+	args?: any;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/pages/en/reference/adapter-reference.md
```diff
@@ -42,15 +42,19 @@ The object passed into `setAdapter` is defined as:
...
 }
 ```
 
 The properties are:
 
 * __name__: A unique name for your adapter, used for logging.
 * __serverEntrypoint__: The entrypoint for server-side rendering.
+* __shim__: Optional array of import specifiers pointing to polyfills/shims.
 * __exports__: An array of named exports when used in conjunction with `createExports` (explained below).
+* __args__: Optional args to pass to your server handler.
```
sarah11918, line 57: If it's correct to say "arguments" in the property description, then I would suggest using the unabbreviated word here.

thepassle: The `args` property already existed, it was just undocumented. Changing it to `arguments` now would be a breaking change.

delucis: I think Sarah means in the explanation of the option (rather than changing the property name).

delucis:
```suggestion
* __args__: Optional arguments to pass to your server handler.
```

delucis:
Good thinking @sarah11918 — done! @thepassle please mark this as ready for review once https://github.com/withastro/astro/pull/3317 is merged. Thanks!

PR #480, opened 2022-05-08
Clarify render props limitation

SteveALee:
Hopefully I got it right.


src/pages/en/core-concepts/framework-components.md
```diff
@@ -87,7 +87,7 @@ the user scrolls down and the component is visible on the page -->
 
 ### Available Hydration Directives
 
-There are serveral hydration directives available for UI framework components: `client:load`, `client:idle`, `client:visible`, `client:media={QUERY}` and `client:only=" "`.
+There are serveral hydration directives available for UI framework components: `client:load`, `client:idle`, `client:visible`, `client:media={QUERY}` and `client:only="FRAMEWORK"`.
```
sarah11918, line 90:
```suggestion
There are serveral hydration directives available for UI framework components: `client:load`, `client:idle`, `client:visible`, `client:media={QUERY}` and `client:only={FRAMEWORK}`.
```
I'm not sure this is best, but I think it's better. Open to suggestions? Our Template Directives page uses `client:only={string}` and `client:media={string}` so it seems like these two directives should at least match?

sarah11918:
Will ask @Jutanium to take a look at this wording, since it's a change to his original.

Also always happy to havce @delucis and @aFuzzyBear weigh in on nuance.

src/pages/en/core-concepts/framework-components.md
delucis: This is tricky because you’d most likely be writing `client:only="react"` in actual code. That applies for both of these but I guess there is a chance you have a file that exports your breakpoints and then pass a variable to `client:media` whereas `client:only` is always just going to be a string. I do agree consistency makes sense here though 👍

Jutanium:
I don't understand the change - what is clarified in the new wording?

src/pages/en/core-concepts/framework-components.md
```diff
@@ -133,7 +133,7 @@ This allows you to build entire "apps" in your preferred JavaScript framework an
 
 Each framework has its own patterns for nesting: `children` props and [render props](https://reactjs.org/docs/render-props.html) for React and Solid; `<slot />` with or without names for Svelte and Vue, for example. 
 
-Note, however, that you can't pass render props or named slots to a framework component from a `.astro` file, even if the framework component supports it. This is due to a limitation in Astro's compiler.
+Note, Astro components are always rendered to static HTML, even if any included framework components are hydrated. This means attribute passing patterns from the Astro component to framework components that require client runtime behaviour will not work. For example you can't pass React "render props" or named slots.
```
delucis, line 136: Sorry for taking so long to get a review in @SteveALee. I think you made a great effort at improving this thing that makes my head hurt 😅

After re-reading your changes several times it clicked and it’s smart, correct, and makes sense, but I’d love it if we can make this simpler and clearer somehow. What do you think of this?

```suggestion
Note, Astro components are always rendered to static HTML, even if framework components are hydrated. This means passing React "render props" or named slots to framework components from an Astro component will not work. Astro components can’t provide the client runtime behavior that those patterns require.
```

As always, Sarah, I’m wouldn’t be surprised if you spotted an opportunity to clarify something there.

src/pages/en/core-concepts/framework-components.md
sarah11918: Yes, and I could even get behind `client:only="framework"` (lowercase) because you won't be typing REACT when you use it.  :)  ... but then maybe that means `client:media="query"` and I don't know how THAT looks...

I don't know how "legal" it would be to do "*framework*"  and "*query*" italicized...? Now I'm just making things up.

delucis:
> I don't understand the change - what is clarified in the new wording?

We ran in to people misreading the current wording. Not knowing that a “render prop” is a very specific kind of prop, some people understood this as “you can’t pass props from Astro to a framework component.” So I believe the intention was to give a little more context explaining what exactly you can’t pass: things that would need to render on the client.

Jutanium:
@delucis Agh sorry Chris, I looked at the OP's wording not seeing yours. I'm happy with what you have now!

src/pages/en/core-concepts/framework-components.md
sarah11918:
```suggestion
Note: Astro components are always rendered to static HTML, even when they include framework components that are hydrated. So, Astro components can only pass props to components via attributes and children that are known at build time. This means that passing React's "render props" or named slots to framework components from an Astro component will not work. Astro components can’t provide the client runtime behavior that those patterns require.
```
I agree, @SteveALee 's contribution was excellent. And, as always, Chris is my muse... making me realize exactly what I want. (Even if I then need someone other than me to execute it.)

The thing I want here is, even if not exactly as above, some kind of statement about what props we CAN pass. I think that would help make it clear, "Hey, we're not saying no props!" 

I'm still deciding between "Note:" and one of our emojis (warning triangle? lightbulb? We're admittedly not terribly consistent about that.) But, gimme a line about the props we CAN pass, and I think we have a winner.

delucis: Hmm, I get where you're coming from but this section is really specifically about this one kind of prop that's a problem, so I'm not sure we really need to list what is possible?

I'm also ducking a little as it's quite hard to phrase what you ask well. It's not really about being "known" at build time, it's about rendering being possible at both build time and on the client. Frameworks can do that but Astro can't so Astro can't pass a function that says "render me two p tags with this content" to a framework because that function cannot exist outside build time.

What would be a pithy way to explain that? This is what the "client runtime" bit is trying to point at currently but maybe it's not clear enough.

In the meantime here are your improvements minus the extra sentence.

```suggestion
Note: Astro components are always rendered to static HTML, even when they include framework components that are hydrated. This means that passing React's "render props" or named slots to framework components from an Astro component will not work. Astro components can’t provide the client runtime behavior that those patterns require.
```

sarah11918: I guess I was really hoping you were going to have a way to say that that could fit in a sentence. 😅  

My hope was that someone seeing just this section, who maybe hadn't diligently read the whole way through all the examples leading up to this, but just ended up here, wouldn't think there's no prop passing at all to framework components. That's why I was hoping we could fit in a very quick "This will work, but this won't" even if it's a bit outside the scope of this exact section to address a specific problem.

I'm trying to figure out why we get people in support who have read this here and think that we can't pass any props, when there are example right above that CLEARLY show it happening. So if we had HAD a sentence that could go here, that would be what I'd ideally want. If no such simple sentence exists (because I agree, it's not about having a huge explanation at this point), then I say we take what we have and see how it works in practice? Wait to see if someone else PRs again? 😄 

I haven't seen a comment from @SteveALee today. How do you feel about the latest version above, Steve? Does this do what you were hoping?

delucis: So I'm not 100% on "do rendering" but maybe this is in the right direction?

```suggestion
Note: Astro components are always rendered to static HTML, even when they include framework components that are hydrated. This means that you can only pass props that don't do any HTML rendering. Passing React's "render props" or named slots to framework components from an Astro component will not work. Astro components can’t provide the client runtime behavior that those patterns require.
```

SteveALee: The only issue I was addressing was that client:only was specified as taking whitespace as the parameter. However, it requires a string specifying the framework. 

I copied the style you used earlier of a capitalised placeholder name, FRAMEWORK

The only solution  I can think of for fully specifying the syntax would be good old backus naur.

SteveALee: Otherwise, I'm getting lost on what the semantics and syntax are here. Makes my head hurt too :)

src/pages/en/core-concepts/framework-components.md
delucis: Haha, don’t think we can italicize inside code? I’d say go with your `{FRAMEWORK}` for now and hope folks understand that’s a placeholder syntax.

SteveALee: I thought of "string" but that's not precise enough. Need semantics not syntax. Caps seemed to be how you were representing a placeholder so I followed. Not my favourite or very common :( mabe use <framework> as good old Backus–Naur form but might be confused with HTML markup :(

SteveALee:
Sorry, I just cleaned up my mess after trying to use the mobile GitHub app to comment to the two threads here :(

+1 to @delucis [latest version](https://github.com/withastro/docs/pull/480#discussion_r872041460) for clearing up the various compounded issues. Nice!

Perhaps a very slight niggle with 'that don't do any HTML rendering'. Rendering WHEN? But that is explained in final sentence. I only raise it as the time of rendering is something that is critical and separates all the SPA, SSG, SSR, Edge behaviours. Some folks seem to be having problems getting their head around the distinction. Anyway no need to change it.

src/pages/en/core-concepts/framework-components.md
sarah11918: "do rendering" ...  "inject/create/parse new DOM/HTML elements"  ?

As an alternative to this unicorn sentence, I'm now wondering whether instead linking to reactjs.org for "render props" might do the job of making it more clear that it's "a thing" and doesn't just refer to passing props in general?

SteveALee: Please see my [comment](https://github.com/withastro/docs/pull/480#issuecomment-1125880472) in wrong place.

Re unicorn - I seem to recall previous para about react had such a link - maybe the two need two be worked on together?

sarah11918: Serves me right for just looking at the app on my phone... I probably had this same concern/solution up there already!

OK, when I'm at the computer, I'll scan the page as a whole because, yes, my main concern is people coming away thinking that they can't pass props at all. So maybe I have to do a check that we've organized the info in such a way that best sets them up for success.

delucis:
Thanks @SteveALee based on this I've merged in the most recent suggestion. We can keep thinking about improvements, but I'll let Sarah have a look at the page holistically first. Great team work folks!

delucis:
Huh, Netlify build appears to be using `yarn` instead of `pnpm` for some reason and failing. Saw the same on #535 the other day but not sure what's causing it... 🤔

sarah11918:
LGTM!

I'm happy with the latest version! @SteveALee I think this was a great addition to the section, and I think the sooner we get it out there for the world, the better!

Thanks @delucis for the great edits and care taken!

I will put evaluating the page "on my stack" as I'll be using it very soon to build the new tutorial. That'll be my chance to give it a thorough anaylsis.

src/pages/en/core-concepts/framework-components.md
aFuzzyBear: `{FRAMEWORK}` is what should go here

aFuzzyBear:
There is something on the `api-reference.md` page that is causing the test to bork somehow, @delucis up for some investigating my dear Watson

delucis:
> There is something on the `api-reference.md` page that is causing the test to bork somehow, @delucis up for some investigating my dear Watson

I think it's that Netlify is choosing to build with yarn for some reason. This PR doesn't change the API reference page and it builds fine on the main branch so it's a bit odd. Saw the same behaviour in #535.

delucis:
Update: _all_ our Netlify deploys build with yarn! So it's not that but that also seems wrong... Traveling all day so if someone can figure this out that'd be great. Otherwise I'll look tomorrow.

delucis:
> do you think some combination of those recent things could be causing this, just to point Tony in the right direction?

It's as good a guess as anything although I don't understand what in this PR would make those start misbehaving when those changes are present on other branches that are building fine. It's kind of weird!

sarah11918:
Checks pass, @delucis! 

I think before this, we got to a place we were happy with the content? Can we just confirm that, then I think this is mergable?

sarah11918:
Final update from main branch, and checks are passing. So I'm merging, then we can always amend if we decide we don't like it later! Thanks @SteveALee @delucis

PR #483, opened 2022-05-08
i18n(de): Add translations for `i18n/de/ui.ts`

hippotastic:
Adds missing German strings to UI dictionary.

### DIFF
--- a/src/i18n/de/ui.ts
+++ b/src/i18n/de/ui.ts
@@ -1,19 +1,42 @@
 import { UIDictionary } from '../translation-checkers';
 
 export default UIDictionary({
+	'a11y.skipLink': 'Zum Inhalt springen',
+	'navbar.a11yTitle': 'Hauptnavigation',
+	// Site settings
+	'site.title': 'Astro-Dokumentation',
+	'site.description': 'Erzeuge schnellere Websites mit weniger ausgeliefertem JavaScript.',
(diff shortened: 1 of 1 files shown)


Pukimaa (changes requested):
This is some great work! But could you give me an url with a site which includes `<Since>` because in your example you wrote "NEW" and in the translation "Neu" which should be "NEU" If your example is correct. I know not your translation but if that would be changed it would be nice!

hippotastic:
@Pukimaa Are you referring to the line `// Used in `<Since>`: Added in: v0.24.0 [NEW]`? This is just a comment that has been copied over 1:1 from the English original. The string that's actually being used is below and correctly states "Neu". 😄

hippotastic:
Oh, and the English original file also doesn't use "NEW" in all uppercase in the actual string: https://github.com/withastro/docs/blob/1317bb6aee86a4d9f63015be524006e873266738/src/i18n/en/ui.ts#L29

Pukimaa (approved):
Thanks for the explanation for the "New" Tag. I would say it's ready to merge.

hippotastic:
@delucis We're ready to merge here :)

delucis (approved):
Looks good!

You can see an example of the `Since` component here: https://deploy-preview-483--astro-docs-2.netlify.app/de/reference/configuration-reference/#serverhost

(There's no "new" badge as that is included if the version number where the feature was added is recent enough. It is rendered uppercase using CSS text-transform.)

PR #495, opened 2022-05-09
Don’t italicise keywords in syntax highlighter

delucis:
We had a user run into an issue where they mistook `|` for `/` because it was being italicised by our syntax highlighter (see [Discord support thread](https://discord.com/channels/830184174198718474/973143587044864040)). That doesn’t seem great, so I suggest removing that italic styling.

Here’s the offending code sample:

| before | after |
|---|---|
| Screen Shot 2022-05-09 at 16 31 49 | Screen Shot 2022-05-09 at 16 30 58 |

It does impact elsewhere, like this below. Anyone know if there’s a way to target those things separately?

| before | after |
|---|---|
| Screen Shot 2022-05-09 at 16 33 26 | Screen Shot 2022-05-09 at 16 34 51 |

### DIFF
--- a/syntax-highlighting-theme.ts
+++ b/syntax-highlighting-theme.ts
@@ -149,7 +149,6 @@ export const tokens: IRawThemeSetting[] = [
 		scope: 'keyword',
 		settings: {
 			foreground: red[0],
-			fontStyle: 'italic',
 		},
 	},
 	{


sarah11918:
I have to agree. I was just going over a code sample with `||` and it really did look like (weak) `//`.

So, I'm +1 on the change.

delucis:
Do you think we should wait for @bholmesdev’s eyes — given he led on setting the theming up — or do we just take an executive decision?

bholmesdev:
@delucis _yes_, this de`theme`itely looks good!

Pukimaa:
This seems like a small but impactful detail! It should clear things up while not looking off elsewhere...if that makes sense 😅

PR #498, opened 2022-05-09
Use a more targeted npm link for integrations

delucis:
Instead of linking to the npm front page, this changes the integrations guide link to use a targeted search for the recommended astro integration keyword (`astro-component`): https://www.npmjs.com/search?q=keywords:astro-component&ranking=popularity

Thoughts on whether that is helpful?

### DIFF
--- a/src/pages/en/guides/integrations-guide.md
+++ b/src/pages/en/guides/integrations-guide.md
@@ -6,7 +6,7 @@ title: Using Integrations
 i18nReady: true
 ---
 
-**Astro Integrations** add new functionality and behaviors for your project with only a few lines of code. You can write a custom integration yourself, or grab popular ones from [npm](https://www.npmjs.com/). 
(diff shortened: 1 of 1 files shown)


sarah11918:
LGTM!

Maybe this means we need added content somewhere about installing npm packages meant for other UI frameworks? On framework-components, maybe? A new section? People do seem to be asking about installing stuff FOR REACT etc. and how to use it. (And then side-note: if we are NOT calling these "Integrations", then how ARE we distinguishing between "npm stuff for Astro" and "npm stuff for your other frameworks"?)

But I do agree, when talking about Astro Integrations, this link is perfect. 🚀

delucis:
I think we should stick to using the term “Integration” specifically for a thing you import and pass to the `integrations` array in your `astro.config.mjs`. Anything else is a “module” or “package” and can usually be used “as you normally would” (with the exception of Vue plugins that document a `App.use(plugin)` pattern that isn’t compatible with Astro — [explained very well in this Discord post](https://discord.com/channels/830184174198718474/872579324446928896/972760021588906004)). Obviously “as you normally would” is useless if you have never done something before, so this is probably something to improve in each framework’s docs (so I guess in the integration READMEs?) or in some future tutorial? Like, “here is how to use a date-formatting module with your framework”.

Issue #335, opened 2022-04-09
Astro.self docs are missing


FredKSchott:
Nothing here: https://docs.astro.build/en/reference/api-reference/#astroself

sarah11918:
Adding for reference: https://discord.com/channels/830184174198718474/830184175176122389/949369789900525578

aFuzzyBear:
See PR #499

PR #499, opened 2022-05-09
🎭 Added text describing `Astro.self` behaviour

aFuzzyBear:
Addressing Issue #335 

Adding Documentation surrounding the `Astro.self` global. 

Appended text to the `src/pages/api-references.md` 

Demonstrated the nature of using `Astro.self` in a recursive use-case of generating subheading from the `frontmatter.headings` also provided example of `prop drilling`.

### DIFF
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -158,6 +158,29 @@ You could pass a callback function that renders our the message:
 <div>Hello world!</div>
 ```
 
+### `Astro.self`
+
+`Astro.self` allows Astro components to be recursively called. This behaviour lets you call the same Astro component over again by defining `<Astro.self>` within the component template.  Allowing you to easily iterate over large data stores and nested data-structures.
+
+```astro
+---
+const {frontmatter, nestedHeaders} = Astro.props 
+const Headers = frontmatter.headers || nestedHeaders
+---
+<ul>
+  {Headers.map((item)=>{
+    //If there is a nested data-structure
+    //We invoke `<Astro.self>`
+    if(Array.isArray(item)=>{
+      //We can also pass props through with the recursive call
+      return (<Astro.self nestedHeaders={item.subHeadings}>) 
+    })
+    <li>
+      {item}
+    <li>
+  })}
+</ul>
+```
 
 ## `getStaticPaths()`
 


sarah11918:
Yay, thanks @aFuzzyBear !! Tackling issues...

I have questions... 😅 

Let's start with: Looking at Nate's tweet here: https://discord.com/channels/830184174198718474/830184175176122389/949369789900525578 suggests that there are two different ways you can use Astro.self. (Either by defining a component first, or just using it as a component itself.)

1. Are both these ways still possible? (stuff could have changed) and if so
2. Do you see any value in having both in an example? and
3. Is one way more "Astro-y"/recommended than the other?

src/pages/en/reference/api-reference.md
```diff
@@ -158,6 +158,29 @@ You could pass a callback function that renders our the message:
 <div>Hello world!</div>
 ```
 
+### `Astro.self`
+
+`Astro.self` allows Astro components to be recursively called. This behaviour lets you call the same Astro component over again by defining `<Astro.self>` within the component template.  Allowing you to easily iterate over large data stores and nested data-structures.
```
sarah11918, line 163:
```suggestion
`Astro.self` allows Astro components to be recursively called. This behaviour lets you call the same Astro component over again by defining `<Astro.self>` within the component template. 
```
Do you think maybe these two sentences say pretty much the same thing? Can this be condensed into one sentence?

b) And, I don't think I'd say we've "defined" Astro.self? I know you use "invoke" somewhere else. Depending on the sentence, maybe "called" or "rendered"? Something else?

sarah11918, line 163:
```suggestion
within the component template, allowing you to easily iterate over large data stores and nested data-structures.
```
I love the last bit, but it's not a sentence on its own. It's either going to be after a comma fromt whatever you decide for before, or it would need to be something like:

"This allows you to easily iterate..."

src/pages/en/reference/api-reference.md
```diff
@@ -158,6 +158,29 @@ You could pass a callback function that renders our the message:
...
+
+```astro
+---
+const {frontmatter, nestedHeaders} = Astro.props 
+const Headers = frontmatter.headers || nestedHeaders
```
delucis, line 168: It might be useful for the example to simplify this. Or even better can we just inline the data we're thinking about so that it's easy to see?

So declare some recursive data in the front matter and then use the template Astro.self to render it?

Oh wait. Or we can't do that can we? Because the self needs to receive props. In that case maybe show the component in a separate code block being passed the data?

src/pages/en/reference/api-reference.md
```diff
@@ -158,6 +158,51 @@ You could pass a callback function that renders our the message:
...
+  <li>A</li>
+  <li>
+    <ul class="nested-headings">
+      <li>B</li>
+      <li>C</li>
+    </ul>
+  </li>
+  <li>D</li>
+</ul>
+```
```
delucis, line 205: Sorry for vacillating here @aFuzzyBear. I think maybe having this just be “list” instead of the more specific headings might be clearer?

~~~suggestion
```astro
---
// NestedList.astro
const { items } = Astro.props;
---
<ul class="nested-list">
  <li>{items.map((item) => {
    if (Array.isArray(item)) {
      // If there is a nested data-structure we render `<Astro.self>`
      // and can pass props through with the recursive call
      return <Astro.self items={item} />;
    } else {
      return item;
    }
  })}</li>
</ul>
```

This component could then be used like this:

```astro
---
import NestedList from './NestedList.astro';
---
<NestedList items={['A', ['B', 'C'], 'D']} />
```

And would render HTML like this:

```html
<ul class="nested-list">
  <li>A</li>
  <li>
    <ul class="nested-list">
      <li>B</li>
      <li>C</li>
    </ul>
  </li>
  <li>D</li>
</ul>
```
~~~

delucis, line 205: Sorry for vacillating here @aFuzzyBear. I think maybe having this just be “list” instead of the more specific headings might be clearer?

~~~suggestion
```astro
---
// NestedList.astro
const { items } = Astro.props;
---
<ul class="nested-list">
  <li>{items.map((item) => {
    if (Array.isArray(item)) {
      // If there is a nested data-structure we render `<Astro.self>`
      // and can pass props through with the recursive call
      return <Astro.self items={item} />;
    } else {
      return item;
    }
  })}</li>
</ul>
```

This component could then be used like this:

```astro
---
import NestedList from './NestedList.astro';
---
<NestedList items={['A', ['B', 'C'], 'D']} />
```

And would render HTML like this:

```html
<ul class="nested-list">
  <li>A</li>
  <li>
    <ul class="nested-list">
      <li>B</li>
      <li>C</li>
    </ul>
  </li>
  <li>D</li>
</ul>
```
~~~

aFuzzyBear: no problem in the slightest, list was my first thought, but thoughts of, perhaps a usable example, but given he code demo, list feels like it would suit the generic format you've provided for us dude

src/pages/en/reference/api-reference.md
```diff
@@ -158,6 +158,51 @@ You could pass a callback function that renders our the message:
 <div>Hello world!</div>
 ```
 
+### `Astro.self`
+
+within the component template, allowing you to easily iterate over large data stores and nested data-structures.
```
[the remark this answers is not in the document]
aFuzzyBear: 😵‍💫🧐
I think I made a oops, I had tried to resolve both the editions from Sarah, but only one got applied 😑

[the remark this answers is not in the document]
delucis:
```suggestion
`Astro.self` allows Astro components to be recursively called. This behaviour lets you render an Astro component from within itself by using `<Astro.self>` in the component template. This can be helpful for iterating over large data stores and nested data-structures.
```

sarah11918:
Then, I am satisfied! Carry on, folks!

delucis:
Over to you @aFuzzyBear — sorry for meddling, but I think this is OK now!

aFuzzyBear:
🎉🎉🎉
Great work team 👊💚

PR #508, opened 2022-05-09
Don’t convert “--” to “—”

delucis:
Disable `remark-smartypants`’s `dashes` option so that `--` is preserved (important for CLI flags for example.)


sarah11918:
LGTM!

PR #515, opened 2022-05-10
Feat: Allow skipping issue search by configuring issue number

hippotastic:
This PR adds support for two new optional env variables to the translation status script:
- `ISSUE_NUMBER`: If set, the script skips the issue search and uses the given issue number. If this optional env variable is set in the workflow and the referenced issue is not found, the script will throw an error and abort.
- `ISSUE_TITLE`: I'm making this configurable through env variables as well in case we want to change the issue title in the future. This is especially important because the script validates the issue title to avoid accidentally changing the wrong issue.

I'm also updating the workflow to set the new `ISSUE_NUMBER` env variable to our existing issue #438. This should hopefully reduce the amount of rate limiting that occurs in our workflow.

### DIFF
--- a/.github/workflows/translation-status.yml
+++ b/.github/workflows/translation-status.yml
@@ -36,3 +36,4 @@ jobs:
         run: pnpm run github:translation-status
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ISSUE_NUMBER: 438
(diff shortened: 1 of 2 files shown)


hippotastic:
@delucis PTAL

delucis (approved):
Thanks for the quick response on this!

PR #520, opened 2022-05-10
Fix: Make translation status work again in forks

hippotastic:
My previous commit that added a fixed issue number to the translation status workflow caused workflow runs in forks to fail because an issue with number 438 does not exist there.

This PR fixes that by checking the repository owner and only using our hardcoded issue number if it's `withastro`. Workflow runs in forks will revert to the previous behavior which searches for the correct issue number based on the issue title.


hippotastic:
@delucis PTAL

delucis (approved):
In that case, this LGTM 👍

PR #529, opened 2022-05-11
Add caveat about potential <Markdown /> deprecation

tinymachine:
This same caveat appears in the docs section on Markdown but not in the API section.


delucis (approved):
Great idea, thank you! 🚀

PR #530, opened 2022-05-11
Use language-specific link from docs logo

delucis:
The docs logo at the top-left currently always links to `https://docs.astro.build/` which gets redirected to the English version of the docs. This change means that we link to the current language, e.g. `https://docs.astro.build/fr/` to keep users on their current language if they use that link.

H/T to @hippotastic for spotting this.

### DIFF
--- a/src/components/Header/Header.astro
+++ b/src/components/Header/Header.astro
@@ -46,7 +46,7 @@ const t = useTranslations(Astro);
 					></path>
 				</svg>
 			</a>
-			<a href={new URL((lang || '') + '/', Astro.site).href}>
+			<a href={`/${lang || 'en'}/getting-started/`}>
 				<h1 class="sr-only">Docs</h1>
 				<svg xmlns="http://www.w3.org/2000/svg" width="226" height="102" viewBox="0 0 226 102" fill="none">
 					<path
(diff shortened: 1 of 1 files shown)


hippotastic:
I noticed two things while looking at the deploy preview:

1. In my opinion, the docs logo link should point directly to the `getting-started` page to prevent an additional redirect whenever clicking it. There is no page at e.g. `https://docs.astro.build/en/`, it will always redirect to `https://docs.astro.build/en/getting-started/`.
2. I'd suggest making both logo links relative (so `<a href="/">` rather than `<a href="https://astro.build/">`). This would make them work correctly in the deploy preview as well because it is served from a different domain.

Chrissdroid (commented):
I do agree with @hippotastic's points and the more we avoid redirection, the better search engines will treat us. It avoid redirection bounces and gain this in better loading and caching from the browser

delucis:
Thanks for the suggestions! Agree with both and accepted @Chrissdroid’s proposed change 🙌

Chrissdroid (approved):
LGTM! 👍

hippotastic (approved):
LGTM!

sarah11918:
Lovely work, @delucis and everyone! Thoughtful as ever, in every respect. LGTM!

PR #533, opened 2022-05-11
Convert linkcheck script to TypeScript

hippotastic:
As discussed with @delucis, I've started converting my scripts used in GitHub workflows to TypeScript. I've decided to use the `tsm` loader which is already used by Astro as well.

I migrated all options from our previous `jsconfig.json` over to the new `tsconfig.json`, and then added new TS options on top to get strict checks. 

Special note: To make the TypeScript compiler output more predictable, I changed the `module` setting from `esnext` to `ES2020`. Using `esnext` is not recommended by the TypeScript docs.

I tested my changes in VS Code (the `.astro` extension support seems to work just as well as before), ran a build, and also tested running the new script from a GitHub workflow without any issues.


scripts/lint-linkcheck.ts
```diff
@@ -187,16 +217,24 @@ class BrokenLinkChecker {
...
-		const annotations = [];
+	outputSourceFileAnnotations (brokenLinks: BrokenLink[]) {
+		const annotations: {
+			message: string;
+			location: {
+				file: string;
+				startLine: number;
+				startColumn: number;
+				endColumn: number;
+			}
```
delucis, line 228: Looks like we can use `@actions/core`’s [exported type](https://github.com/actions/toolkit/blob/c4ae214c261c0d684c6bcfe6632b773b82482bad/packages/core/src/core.ts#L40-L71) for this:

```suggestion
			location: AnnotationProperties;
```

(See next suggestion for adding the import.)

scripts/lint-linkcheck.ts
```diff
@@ -4,11 +4,41 @@ import kleur from 'kleur';
 import htmlparser2 from 'htmlparser2';
 import core from '@actions/core';
```
delucis, line 5:
```suggestion
import core, { AnnotationProperties } from '@actions/core';
```

tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+	"compilerOptions": {
+		"allowJs": true,
+		"declaration": true,
+		"emitDeclarationOnly": true,
```
delucis, line 5: Do we need these? We’re not ever building with `tsc` so I would have thought emitting `.d.ts` files was not really relevant.

tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+		"strict": true,
+		"baseUrl": ".",
+		"moduleResolution": "node",
+		"module": "ES2020",
+		"target": "ES2020",
+		"paths": {
+			"~/*": ["src/*"]
+		},
+		"outDir": "./dist",
+		"declarationDir": "./dist/types",
```
delucis, line 18: Just double checking: does anything break without this? We didn’t need it with jsconfig so just wondering. It does align with the build output but could get out of sync with `astro.config.mjs` in theory, so I wouldn’t include it unless we definitely need it.

delucis (commented):
Mostly looks good @hippotastic — thank you 🙌 

I’ve left one suggestion and have a couple of questions on the tsconfig, but the actual script itself is 💯

tsconfig.json
hippotastic: I have actually used `tsc` a couple of times already as kind of a TS linter. You can just run it from the repo root and it will output all issues from all scripts instead of having to go through them manually in VS Code. If you remove those options, it's not possible to run `tsc` anymore and it will error.

tsconfig.json
hippotastic: Yes, if you remove these, `tsc` breaks. It would overwrite the source files and errors out.

delucis (approved):

PR #534, opened 2022-05-11
CSS Styling page final polish

sarah11918:
Some minor fixes to this page. 

Notably, the link at the top of the page mentioning Tailwind now goes to our Tailwind Integration (instead of tailwind.org itself). We need to get users there more quickly, because that's where (not here!) we document instructions for each specific integrations, and Tailwind is a BIG one!

### DIFF
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -125,7 +125,7 @@ Because this approach uses the `public/` directory, it skips the normal CSS proc
 
 ## CSS Integrations
 
-Astro comes with support for adding popular CSS libraries, tools and frameworks to your project like Tailwind and more! 
+Astro comes with support for adding popular CSS libraries, tools and frameworks to your project like [Tailwind][tailwind] and more! 
 
 📚 See the [Integrations Guide](/en/guides/integrations-guide/) for instructions on installing, importing and configuring these integrations.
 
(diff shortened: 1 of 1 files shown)


Chrissdroid:
The tailwind link is only used once at the start of the page, would it be interesting to make a section inside "CSS integrations" for more visibility toward this Integration ?

sarah11918:
I'm not opposed to suggestions! The biggest issue with this page, as I see it (and am happy to get feedback on):

- OMG FRIGGIN' TAILWIND. It's a BIG DEAL. And to some people, maybe CSS *is* Tailwind? This page really needs to point people to the integration, and maybe it's not yet successful at that?

Otherwise, the purpose of this page SHOULD be to explain the style tag, scoped styles, bundling and all that stuff. But, the fear is that people are going to expect Tailwind info here. And we DO need a nice "off-boarding" for them.

Even if we use the "CSS + UI" filter on the astro.build/integrations page, I don't think we get a css-heavy focused list of things. So, eventually, I think this page could *maybe* be more geared towards `<style>`? Not, "all your CSS ideas"

But for now, I *do* want to push the double period to end a sentence typo fix, and replace the link to tailwind with the integration link! We need to at least pretend to send people there from this page!  😄

sarah11918:
@Chrissdroid I linked the word Tailwind again when it shows up in the Integrations section. There's already a link right there to the regular integrations, so with the way the page is now, I think that's as "linky" as we can get?

Chrissdroid:
LGTM! It's the best we can do for now on this page 👍

sarah11918:
>  It's the best we can do for now on this page 👍

Such cruel, cold comfort... but I'll take it.

PR #538, opened 2022-05-12
docs(zh-CN): translate editor setup and migrate

kecrily:

- [ ] New or updated content
- [x] Translated content
- [ ] Something else!

#### Description

Translate editor setup and migrate


delucis (dismissed):
Nice! Thank you so much @kecrily 

I think the changes in `pnpm-lock.yaml` don’t need to be included in this PR, would you mind removing them?

kecrily:
Ok, but do we also need to update the pnpm version to v7?

delucis:
> Ok, but do we also need to update the pnpm version to v7?

I think eventually, but currently we build using Node v14 which isn’t compatible with pnpm@7, so we need to move both together at some point.

delucis (approved):
Just checked the content here using Google Translate and it looks good to me. I can’t speak to any language details, but at least the translation matches the original content 🙌

PR #539, opened 2022-05-12
perf(a11y): add searchbox role

kecrily:

- [ ] New or updated content
- [ ] Translated content
- [x] Something else!

#### Description

Aria-label attribute cann't be used on a span with no valid role attribute. So I add `searchbox` role for it.

### DIFF
--- a/src/components/Header/Search.tsx
+++ b/src/components/Header/Search.tsx
@@ -49,9 +49,7 @@ export default function Search({ lang = 'en', labels }: Props) {
 					<path d="M21 21l-6-6m2-5a7 7 0 11-14 0 7 7 0 0114 0z" stroke="currentColor" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round" />
 				</svg>
 				<span className="search-placeholder">{labels.button} <span class="sr-only">({labels.shortcutLabel})</span></span>
-				<span className="search-hint" role="searchbox" aria-label={labels.shortcutLabel}>
-					<kbd aria-hidden="true">/</kbd>
-				</span>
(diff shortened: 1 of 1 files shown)


delucis:
Thanks! Would you mind explaining how you tested the accessibility with and without the `aria-role`? My understanding is that span receives a default “text container” role.

Firefox’s accessibility tree shows this:

image

And Chrome this:

image

That suggests that at least those browsers are taking the `aria-label` into account. Obviously if the label is missing from screen reader output we should fix that, but we probably need a better role as the text there is describing a shortcut to activate search rather than the `searchbox` input itself.

kecrily:
I used the narrator that comes with Windows for testing. Its presence or absence does not seem to affect the result; the narrator will always read the button role of the previous level.

This issue was reported by [Microsoft's A11y extension](https://accessibilityinsights.io/docs/general/telemetry/).

delucis (commented):
Thanks so much for looking into that. Given that, and given the difficulty of finding an appropriate `role`, what do you think of this solution? Move the shortcut label *into* the placeholder span and hide its visibility. So the button is labelled “Search (Press / to search)” and we don’t worry about the `<kbd>` element.

delucis:
Thanks @kecrily! Out of curiosity did you check that this works as expected with your A11y tool?

kecrily:
Yes, now the A11y tool can't find any problems. Astro docs did a great job on the A11y! 🌟

delucis (approved):
Fantastic. Thanks so much for doing that research and taking the time to work on it 🙌

PR #545, opened 2022-05-12
Code as headings

### LINKED ISSUE #509: Code as headings, or not?
We're trying to determine a consistent style to use.

Currently, and (explicitly by request), the CLI page does NOT use formatted code for the various `--` headings. 

However, on Project Struture, the directories DO use formatted code directories.

Do we have a preference? Are we concerned with consistency? (And sorry, more of these issues will come up as more translators are actively in the docs wanting explicit instructions, and finding these inconsistencies.) 😅 

Screenshot 2022-05-09 18 45 12
Screenshot 2022-05-09 18 44 43

Chrissdroid:

#### Description

- What does this PR change? Adds backticks in the `reference/` directory.

Closes #509 

Before:
image

After:
image

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -103,6 +103,11 @@ h2 {
 	margin-top: 2rem;
 }
 
+:is(h1, h2, h3, h4, h5) code {
+  font-size: 1em;
+  font-weight: 400;
+}
+
 h1 {
 	font-size: 3.25rem;
 	font-weight: 800;
--- a/src/pages/en/reference/adapter-reference.md
+++ b/src/pages/en/reference/adapter-reference.md
@@ -118,7 +118,7 @@ export function start(manifest) {
 }
 ```
 
-#### astro/app
+#### `astro/app`
 
 This module is used for rendering pages that have been prebuilt through `astro build`. Astro uses the standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) and [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) objects. Hosts that have a different API for request/response should convert to these types in their adapter.
 
@@ -139,7 +139,7 @@ export function start(manifest) {
 
 The following methods are provided:
 
-##### app.render(request)
+##### `app.render(request)`
 
 This method calls the Astro page that matches the request, renders it, and returns a Promise to a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object. This also works for API routes, that do not render pages.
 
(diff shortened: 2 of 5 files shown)


sarah11918:
I mean... right??

delucis:
In general I feel positive about this change, except maybe these smaller headings at the bottom of the Adapter API reference page?

image

They feel kind of too small. Whereas currently the bolding helps them stand out:

image

The `code` in headings don’t look bold because we only ship regular and italic weights of IBM Plex Mono (they get kind of faux-bolded by the browser but it’s minimal and doesn’t look great).

We _could_ also ship a bold weight of the monospace font, but the alternative solution would be shipping _no_ code font and use system fonts like we do for all the other text in the docs? That way there’ll usually be a bold monospace variant available.

As an example here’s the same snippet with `font-family: monospace` on my (macOS) machine:

image

sarah11918:
I agree they seem a little small, and would be tempted to want them a tiny bit bigger instead of necessarily bolder myself, but I'm OK with whatever people prefer!

delucis:
> I agree they seem a little small, and would be tempted to want them a tiny bit bigger

Yeah, that’s also a possibility. By default `code` is set to a smaller font-size than surrounding text to balance, but maybe we need to experiment with some heading-specific code styles.

delucis:
@Chrissdroid Do you want to try adding the following to the docs styles to see how it feels? Bumps the font size back up to the size of other text helping code-only headings stand out and also disables bold so we don’t get the browser faux-bolding the monospace font.

```css
:is(h1, h2, h3, h4, h5) code {
  font-size: 1em;
  font-weight: 400;
}
```

In a couple of quick tests this looked like an improvement to me.

Chrissdroid:
Tell me if it's better for you @sarah11918, on my side it's clearly better for reading

delucis:
Thanks @Chrissdroid!

So, what do we think now? Consider the small standalone headings like [the bottom of the adapter page](https://deploy-preview-545--astro-docs-2.netlify.app/en/reference/adapter-reference/#apprenderrequest) and mixed code and text headings like [the top of the runtime API page](https://deploy-preview-545--astro-docs-2.netlify.app/en/reference/api-reference/) or (smaller) in [the integrations reference](https://deploy-preview-545--astro-docs-2.netlify.app/en/reference/integrations-reference/#config-option).

sarah11918:
I've only looked on desktop so far, both light and dark mode, but *I* like this. I find it very readable.

The only thing that startled me at first was the huge ASTRO global.  :)  But, I don't hate it? It is very "Astro-y"

sarah11918:
Just checked it on my phone, and It's BIG, yeah. But it's CLEAR. I think it's a winner.

Chrissdroid:
It looks pretty great like this, it's more readable this way 👍

sarah11918:
LGTM!

delucis:
Yay! Also LGTM! We can always revert if someone chokes on these `code` tags. Thanks for doing this @Chrissdroid 🙌

PR #547, opened 2022-05-13
Code as heading: final pass

Chrissdroid:

- [x] New or updated content

#### Description

- What does this PR change? Final pass for code used in headings, adding backticks to them
- Did you change something visual? see #509

### DIFF
--- a/src/pages/en/guides/typescript.md
+++ b/src/pages/en/guides/typescript.md
@@ -64,7 +64,7 @@ import Layout from '@layouts/Layout.astro';
 }
 ```
 
-## Component `Props`
+## Component Props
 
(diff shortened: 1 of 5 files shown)


src/pages/en/guides/typescript.md
```diff
@@ -64,7 +64,7 @@ import Layout from '@layouts/Layout.astro';
 }
 ```
 
-## Component Props
+## Component `Props`
```
delucis, line 67: I wouldn’t include this one. `props` lowercase is the code version, here this is just referring to the general concept of “props” rather than a specific variable named “props”.

```suggestion
## Component Props
```

delucis (changes requested):
Looks good @Chrissdroid!

I haven’t conducted an exhaustive search for other places that might need the code ticks adding, but we can add them as we spot them.

Just one change needed I think. Thank you!

src/pages/en/guides/typescript.md
Chrissdroid: Yeah, I was not sure about this one, thanks for the clarification 👍

delucis (approved):

PR #558, opened 2022-05-14
change class to className

victor0x16:
React uses `className`.


delucis (approved):
Well spotted! Thanks for the PR 💜

PR #571, opened 2022-05-16
i18n(de): Add `core-concepts/partial-hydration.md`

hippotastic:

- [ ] New or updated content
- [X] Translated content
- [ ] Something else!

#### Description

- Adds German translation for `core-concepts/partial-hydration.md`.
- Note: Since the diagram linked in the original article is only available in English, I have created a component that recreates the diagram using grid layout. This allows easy translation and makes it accessible.


hippotastic:
Regarding the failed check: The links are fine. The build process just had a random error.

src/components/IslandsDiagram.astro
```diff
@@ -0,0 +1,79 @@
...
+		align-items: center;
+		text-align: center;
+		padding: 2rem;
+		background: hsl(0, 0%, 96%);
+		border: 2px solid hsl(0, 0%, 80%);
+		color: hsl(0, 0%, 40%);
+	}
+	.diagram .app {
+		font-size: 1.05rem;
+	}
```
delucis, line 41: I’m not sure this is doing much? I’d either bump up the font size a little more (`1.25rem` looks more like the original image?) or just remove this and have them all be the same size.

delucis: Yeah, I noticed that too. OK we’ll roll with it!

delucis (approved):
Looks good on a quick look and merging to unblock usage of the IslandsDiagram component in other languages. Thanks @hippotastic!

PR #573, opened 2022-05-16
docs(zh): translate markdown and layout

kecrily:

- [ ] New or updated content
- [x] Translated content
- [ ] Something else!

#### Description

translate markdown and layout


sarah11918:
Whew! *We* have been busy, @kecrily !  😄 

I've been fixing some links that passed hippo's OLD link checker, but not the newest version. BUT, this next batch will be easier for you to do than me.  :)

I updated all the links to be /zh-CN/ instead of /en/. BUT now that more pages are translated, we also have to make sure all the header slugs are localized too (e.g.  #组件-script instead of #the-component-script). When it was just one or two of them, it took work, but I could hunt down the localized equivalent. THIS PR though has a lot of them.  😅  So, I'm going to throw this back to you to update!  💜

kecrily:
It should be fine now

PR #604, opened 2022-05-18
Fix: Place theme toggle title attribute correctly

hippotastic:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

- Fixes a bug in the theme toggle control that prevented its `title` attributes from being used - instead, the title of the entire table of contents aka right sidebar popped up. Bug was confirmed on Chrome and Firefox:
  before-fix
- This PR fixes this by moving the title attribute up in the tree to make it work.
  after-fix

(In both screenshots, please imagine a mouse cursor pointing to the little sun icon.)

**Important:** I don't know if it would be a good idea to also move the `aria-label` attribute. I've kept it at the actual input element because this is the element with the semantic `checked` attribute and I was hoping that a screen reader would read this as something like "Checked option: Use dark theme". If anyone with access to a screen reader can verify if this is the correct solution, that would be awesome.

### DIFF
--- a/src/components/RightSidebar/ThemeToggleButton.tsx
+++ b/src/components/RightSidebar/ThemeToggleButton.tsx
@@ -49,14 +49,13 @@ const ThemeToggle: FunctionalComponent<Props> = ({ labels }) => {
 				const checked = t === theme;
 				const themeLabel = t === 'light' ? labels.useLight : labels.useDark;
 				return (
-					<label class={checked ? 'checked' : ''}>
+					<label title={themeLabel} class={checked ? 'checked' : ''}>
 						{icon}
 						<input
 							type="radio"
 							name="theme-toggle"
 							checked={checked}
 							value={t}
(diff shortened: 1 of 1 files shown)


hippotastic:
Could you have a look with your screen reader, @Yan-Thomas?

Yan-Thomas:
That's a nice addition for accessibility and UX! LGTM!

Just to be clear, the `title` attribute doesn't affect screen readers (at least not NVDA) but is great to have `title` for the specific icons since this helps people with bad sight or who are just confused about "what the hell this sun/moon button does?"

NVDA reads the theme toggle like this: "Use dark theme - radio button checked, 1 of 2". So, we don't need to move the `aria-label` attribute.

hippotastic:
@Yan-Thomas Thank you so much for checking this, and also for the helpful explanation!

@sarah11918 We're ready to merge this! :)

PR #611, opened 2022-05-19
Reworking the Routing Rules - Caveats section

tony-sull:

- [x] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

Related to [PR #3407](https://github.com/withastro/astro/pull/3407)

Tweaking the Routing Rules section to highlight the full rule set used to prioritize which route handles a URL when it matches multiple `src/pages` files


src/pages/en/core-concepts/routing.md
```diff
@@ -128,13 +128,16 @@ In this example, a request for `/withastro/astro/tree/main/docs/public/favicon.s
...
+It's possible for multiple routes to match the same URL path. Astro needs to know which route should be used to build the page. To do so, it sorts them according to the following rules:
 
-Static routes without path params will take precedence over all other routes, and will not match queries for dynamic path params. Similarly, named path routes take precedence over catch-all routes, and will not match queries for catch-all path params.
+- Static routes without path parameters will take precedence over all other routes
+- Dynamic routes using named parameters take precedence over rest parameters
+- Rest parameters have the lowest priority
+- Ties are resolved alphabetically
 
-Consider the following project:
+For example each of these routes would match `/post/create`:
```
sarah11918, line 140: If these routes now ALL do match `/post/create`, then we need to update the lines underneath the code structure. (Because we list out which ones do vs don't match.

sarah11918:
Just need to update the lines underneath the diagram, @tony-sull , but otherwise, LGTM! 🥳

src/pages/en/core-concepts/routing.md
delucis: The code block also reads `posts` (plural) but this sentence and those below all use `post` (singular).

tony-sull: Thanks for the quick review @sarah11918 and @delucis!  Looking back with a fresh set of eyes I actually think it might make sense to have the rules *after* the example project.  Thoughts on my last revision?

src/pages/en/core-concepts/routing.md
```diff
@@ -145,9 +141,18 @@ Consider the following project:
...
 
-- `pages/post/create.astro` - Will match `/post/create`
-- `pages/post/[pid].astro` - Will match `/post/1`, `/post/abc`, etc. But not `/post/create`
-- `pages/post/[...slug].astro` - Will match `/post/1/2`, `/post/a/b/c`, etc. But not `/post/create`, `/post/1`, `/post/abc`
+Astro needs to know which route should be used to build the page. To do so, it sorts them according to the following rules:
+
+- Static routes without path parameters will take precedence over all other routes
+- Dynamic routes using named parameters take precedence over rest parameters
+- Rest parameters have the lowest priority
+- Ties are resolved alphabetically
```
delucis, line 149: Given this is a priority list, would it make sense for it to be numbered:
```suggestion
1. Static routes without path parameters will take precedence over all other routes
2. Dynamic routes using named parameters take precedence over rest parameters
3. Rest parameters have the lowest priority
4. Ties are resolved alphabetically
```

sarah11918: I like the numbering in general, but I'm not sure I see ties resolved alphabetically as a "4", since I guess this could happen at any stage?

I would be OK with or without numbering, and if numbered, then a slight preference for "Ties resolved" to be more of a note than a 4th step. But these are all weak preferences.

Otherwise, LGTM!

delucis: You're right! Then I think maybe bullet points are ok. I don't want to break up these 4 things into 3 numbers and a note.

delucis:
I guess this is good to merge! Thanks @tony-sull!

aFuzzyBear (approved):
I got nothing to say, other than it's great to see Tony helping out in the docs, your really did straighten the rules here, 👊

PR #629, opened 2022-05-22
Improve site typography, spacing and mobile navigation

hippotastic:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

This PR contains a major overhaul of the site's CSS styles, especially focused on improving the typography, spacing and mobile browsing experience. The following has changed:
- Fix broken scrolling when opening the mobile sidebar and then increasing viewport width
- Add two-step collapsing header logo, with the clickable "Docs" part now being visible until there is no more space available
- Make scroll anchors respect the sticky mobile navbar, no more invisible / partially visible headings stuck under the navbar
- Harmonize font sizes, increase readability (the headings now use a less bulky size scaling, the base font size has been set to the widespread common size of 16px rather than a dynamic value between 14px and 16px which caused strange mobile font sizes)
- Slightly increase tap areas for navigation links to make them more accessible
- Slightly reduce spacings to allow more content to be visible
- Move language fallback notice above mobile in-page TOC to prevent surprises by seeing an English TOC
- Move mobile hamburger menu to the right to make it reachable with the thumb
- Hide mobile anchor link icons on headings by default to remove visual noise, but show them on hover and on tap (now the entire heading can be tapped)

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -56,6 +56,14 @@ body {
 	font-size: 16px;
 	line-height: 1.5;
 	max-width: 100vw;
+	--gutter: 0.5rem;
+	--doc-padding: 0.5rem;
+}
+
+@media (min-width: 50em) {
+	body {
+		--doc-padding: 2rem;
+	}
 }
 
 nav ul {
@@ -408,7 +416,6 @@ button {
 h2.heading {
 	font-size: 1rem;
 	font-weight: 700;
-	padding: 0.1rem 1rem;
 	text-transform: uppercase;
 	margin-bottom: 0.5rem;
 	margin-top: 1.5rem;
@@ -416,13 +423,17 @@ h2.heading {
 
 .header-link {
 	font-size: 1rem;
-	padding: 1px 0 1px 0;
-	padding-inline-start: 1rem;
-	border-inline-start: 4px solid var(--theme-divider);
+	padding: 0.1rem 0 0.1rem 0;
 	transition: border-inline-start-color 100ms ease-out, background-color 200ms ease-out;
 }
-.current-header-link {
-	background-color: var(--theme-bg-accent);
+
+.header-link a {
+	display: inline-flex;
+	gap: 0.5em;
+	width: 100%;
+	font: inherit;
+	color: var(--theme-text-light);
+	text-decoration: none;
 }
 
 .header-link:hover,
@@ -442,35 +453,38 @@ h2.heading {
 .header-link:hover svg {
 	opacity: 0.8;
 }
-.header-link a {
-	display: inline-flex;
-	gap: 0.5em;
-	width: 100%;
-}
-.current-header-link a {
-	color: var(--theme-text);
-}
 
 .header-link.depth-3 {
-	padding-left: 2rem;
+	padding-inline-start: 1rem;
 }
 .header-link.depth-4 {
-	padding-left: 3rem;
+	padding-inline-start: 2rem;
 }
 
-[dir='rtl'] .header-link.depth-3 {
-	padding-left: 0;
-	padding-right: 2rem;
-}
-[dir='rtl'] .header-link.depth-4 {
-	padding-left: 0;
-	padding-right: 3rem;
-}
+/* Apply different TOC styling for wide viewports showing the right sidebar */
+@media (min-width: 72em) {
+	h2.heading {
+		padding-left: calc(1rem + 4px);
+	}
+	/* Add line and padding on the left side */
+	.header-link {
+		padding-inline-start: 1rem;
+		border-inline-start: 4px solid var(--theme-divider);
+	}
+	.header-link.depth-3 {
+		padding-inline-start: 2rem;
+	}
+	.header-link.depth-4 {
+		padding-inline-start: 3rem;
+	}
 
-.header-link a {
-	font: inherit;
-	color: var(--theme-text-light);
-	text-decoration: none;
+	/* Highlight TOC header link matching the current scroll position */
+	.current-header-link {
+		background-color: var(--theme-bg-accent);
+	}
+	.current-header-link a {
+		color: var(--theme-text);
+	}
 }
 
 /* Screenreader Only Text */
--- a/public/index.css
+++ b/public/index.css
@@ -517,12 +517,17 @@ h2.heading {
 	white-space: normal;
 }
 
-/* Respect the sticky navbar height when calculating scroll targets */
+/* Add a bit of scroll padding to the top of scroll targets */
+.content > section :target {
+	padding-top: 1rem;
+}
+.content > section > :first-child {
+	padding-top: 0;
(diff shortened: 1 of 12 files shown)


Yan-Thomas:
Amazing improvements, "The Fantastic 🦛"! LGTM.

sarah11918:
Just looking at this now, and seems lovely @hippotastic ! 🎉  This is just an initial scan of the preview between tasks, so not a comprehensive review yet, but there are two things I can see immediately:  

(Note these seem to be currently true on the existing site, too. So this isn't a "you" thing, this is a "now that I'm looking at it" thing that I think are entirely within the scope to fix with these other changes.)  😄 

1. The hover text of EVERY item at the top in a smaller screen width appears to say "Top Navigation" -- the logo, the language switcher, the search bar...


2. In smaller screen sizes, we lose the "docs" logo to nav back to the docs template, and ONLY have the Astro logo, which takes you "back" to astro.build, not Docs home. While on the one hand this makes sense, on the other, my instinct is always to think of a logo taking me "home" on the site I'm currently on. 

So, we need a solution here because I think this works well on desktop (having both, having them animate separately so it's clear that they're separate things) but none of those things that make it GREAT on desktop are relevant to the mobile site.

kevinzunigacuellar:
Amazing work! 🚀

sarah11918:
We very recently just received feedback that someone thought it was SO GREAT and MUCH USER EMPATHY to have 2 logos linking both to astro.build as well as docs home.  😅 

So, I'm going to say, no, I don't think that's our solution.

What else have you got for me??  ;)

hippotastic:
Well, then I'd prefer to keep this out of the scope of my PR, to be honest. I intentionally didn't include possibly controversial topics in there to help getting the fixes out to our readers faster. 😅

public/index.css
```diff
@@ -165,41 +184,26 @@ h5 {
...
 	}
 }
 
-@media (hover: hover) {
-	.heading-wrapper > .anchor-link {
-		opacity: 0;
-	}
+.heading-wrapper > .anchor-link {
+	opacity: 0;
+}
```
delucis, line 197: I don’t think it’s a great idea to hide the anchor links universally. On touch devices where “hover” isn’t really a concept, it makes the anchor links quite hard to discover, except maybe by accident. Once you have accidentally discovered them, you need to poke at some invisible thing on the page, hoping to make the link appear, _then_ click the icon. (Also, with this solution hiding only by opacity it is possible to click on the invisible icon and have the page scroll to the heading, which is a confusing experience in itself.)

I understand the aesthetic concern, which in general you’ve done amazing work on in this PR, but in the end if aesthetic changes hurt UX, I don’t think that’s a worthwhile trade-off. Happy if you want to propose tweaks to how these appear that you think is prettier, but I think we need to stick to hiding them only for `@media (hover: hover)`.

delucis (commented):
In general this is fantastic work @hippotastic! Really love how clean the TOC feels on mobile here. We’ve just been chatting on Discord about ways to further improve the header, but here’s one more point I’d like to push back on. Otherwise I think everything else looks great!

sarah11918:
Yes, just to have it stated here for "the record" @hippotastic , the other concerns are concerns I have, but we don't have to tackle them here.  :)

public/index.css
hippotastic: I've solved it with 0.5 opacity by default on mobile. 😄

hippotastic:
Alright, everything should be fine now! The Docs part of the logo stays on screen now for a lot longer, and unless you have a very narrow phone, it should be visible. I hope we're ready to give this to our users now. I'm exhausted. 😅

hippotastic:
Okay, now even the most narrow phones should be able to display the Docs part of the logo. 😂

public/index.css
```diff
@@ -165,41 +184,32 @@ h5 {
...
 
-	.heading-wrapper.level-h3 .anchor-icon {
-		--icon-baseline: 0.15rem;
-	}
+.heading-wrapper > .anchor-link {
+	opacity: 0;
 }
 
-@media (hover: hover) {
+@media (hover: none) {
```
delucis, line 199: Sorry I thought you'd done but noticed we're relying on this media query now instead of `hover: hover`. On some phones (Sarah's for example) the `hover` media query isn't set so the hover links are never revealed. We need to use the inverse logic so:

1. reveal by default
2. progressively enhance by hiding links for devices that say they are hover-capable

hippotastic: Oh wow, I didn't know that there are devices that don't follow this part of the spec (yet?), but still understand hover: hover. I use this media query in multiple locations and will recheck, thanks!

hippotastic: Done. As there were too many locations and too much logic to rewrite otherwise, I just used the actual inverse query, which also does what you described.

delucis: Will that still work if the hover media flag isn't provided by the user agent? I'll trust you but we should wait for @sarah11918 to confirm on her devices as we know it was an issue on those. Thanks!

delucis: Ohhh maybe I misunderstood your first post above. I don't think Sarah's phone _would_ understand `hover: hover` so that's why we can use it. A media query that isn't known/understood is just ignored so: reveal by default, hide only for browsers that understand and match the query.

hippotastic: The `not` queries are special - if their contents do not match or are not understood by the browser, their contents are applied. So we should be perfectly fine. :)

delucis: Ok! 🤞

hippotastic: Ok, so I just tested it in Chrome, and it appears that I was wrong. When I insert a rule like "not screen and (yabba: hut)", it should always apply occording to my logic. However, it doesn't apply. So I'll actually have to rework all the rules. Stay tuned! 😁

delucis: 🙌

hippotastic: It's properly fixed now. :)

src/components/Header/Header.astro
```diff
@@ -56,31 +53,41 @@ const t = useTranslations(Astro);
...
+	.nav-wrapper {
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		gap: 0.75rem;
+		width: 100%;
+		max-width: calc(100vw - 1.1rem);
+		padding-left: 1.5rem;
+		padding-right: 1.5rem;
+	}
```
delucis, line 85: Nice try @hippotastic! These styles were introduced in this PR so we can also fix them here ;-)

```suggestion
		padding-left: 1rem;
		padding-right: 1.5rem;
	}
	
	@media (min-width: 50em) {
		padding-left: 1.5rem;
	}
```

hippotastic: You can't compare them like this, as I moved the hamburger menu to the right, the visual paddings have remained the same. :) I've checked it side by side between old and new site.

PR #648, opened 2022-05-25
migration guide improvement re: `<script type="module">`

sarah11918:
The migration guide mentions to remove the `hoist` attribute from `<script>` tags when upgrading, but doesn't mention removing `type="module"` which causes unexpected behaviour when not removed.

Changes: 
Adds mention that, like hoist, type="module" should be removed as its default behaviour now
Also, links directly to the section in docs that does fully document current behaviour.

NOTE: No need to fully explain behaviour here on this page. It's well documented at the (now added) link in docs proper. And, this page is just for "steps to take to upgrade" so it only needs to exist in "do this!" form.

- [ X] New or updated content

### DIFF
--- a/src/pages/en/migrate.md
+++ b/src/pages/en/migrate.md
@@ -63,7 +63,7 @@ Read [RFC0017](https://github.com/withastro/rfcs/blob/main/proposals/0017-markdo
 
 This includes a few changes to be aware of:
 
-- **BREAKING:** `<script hoist>` is the new default `<script>` behavior. The `hoist` attribute has been removed.
+- **BREAKING:** `<script hoist type="module">` is the new default `<script>` behavior. The `hoist` and `type="module"` attributes hav been removed.
 - New `<script is:inline>` directive, to revert a `<script>` tag to previous default behavior (unbuilt, unbundled, untouched by Astro).
 - New `<style is:inline>` directive, to leave a style tag inline in the page template (similar to previous `<script>` behavior).
 - New `<style is:global>` directive to replace `<style global>` in a future release.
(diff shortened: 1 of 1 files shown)


src/pages/en/migrate.md
```diff
@@ -63,19 +63,21 @@ Read [RFC0017](https://github.com/withastro/rfcs/blob/main/proposals/0017-markdo
 
 This includes a few changes to be aware of:
 
-- **BREAKING:** `<script hoist>` is the new default `<script>` behavior. The `hoist` attribute has been removed.
+- **BREAKING:** `<script hoist type="module">` is the new default `<script>` behavior. The `hoist` and `type="module"` attributes have been removed.
```
delucis, line 66:
```suggestion
- **BREAKING:** `<script hoist>` is the new default `<script>` behavior. The `hoist` attribute has been removed. To use the new default behaviour, make sure there are no other attributes on the `<script>` tag. For example remove `type="module"` if you were using it before.
```

aFuzzyBear, line 66:
```suggestion
- **BREAKING:** `<script hoist>` and `<script type="module">` is the new default `<script>` behavior. Both `hoist` and `type="module"` attributes have been removed. 
You can still use `type="module"` however this would be treated as `<script is:inline>`
```

aFuzzyBear (commented):
I think this should be do,

src/pages/en/migrate.md
Chrissdroid: I prefer this one over the suggestion from Fuzzy because of how the default is removed from the script.

From the WIP version of `debugging.md`:
> If you have included the [`is:inline` directive](https://docs.astro.build/en/reference/directives-reference/#isinline) or any other attribute in your tag, this default behavior is removed.

sarah11918:
Everyone giving me basically the same edit is great! Thanks everyone!

PR #650, opened 2022-05-25
Moves the Routing page's sidebar entry into Learn > Basics

sarah11918:
This only moves where the Routing page shows up in the Left Sidebar. The file stays in the same location, so no links should have to change. (No RIP link checker! 😅 )

Moving into Learn for now, tabling a later discussion on WHERE EXACTLY.

Also removing placeholders for hypothetical pages that have been scrapped and will not be written.

- [X ] New or updated content
- [X ] Changes to the docs site code


delucis (approved):
LGTM!

PR #651, opened 2022-05-25
minor: Change the syntax for some `project-structure.md` sentences to be more accurate

Chrissdroid:

#### Description

- What does this PR change? Change some of the syntax from `project-structure.md` to suit more in the given context, some suggested by @jbrains

### DIFF
--- a/src/pages/en/core-concepts/project-structure.md
+++ b/src/pages/en/core-concepts/project-structure.md
@@ -85,7 +85,7 @@ It is a common convention to store your CSS or Sass files in a `src/styles` dire
 
 ### `public/`
 
-The `public/` directory is for files and assets which shouldn't be processed during Astro's build process. These files will be copied into the build folder untouched.
+The `public/` directory is for files and assets that do not need to be processed during Astro's build process. These files will be copied into the build folder untouched.
 
 This behavior makes `public/` ideal for common assets like images and fonts, or special files such as `robots.txt` and `manifest.webmanifest`.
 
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/project-structure.md
```diff
@@ -85,19 +85,19 @@ It is a common convention to store your CSS or Sass files in a `src/styles` dire
 
 ### `public/`
 
-The `public/` directory is for files and assets that do not need to be processed during Astro's build process. These files will be copied into the build folder untouched.
+The `public/` directory is for files and assets which shouldn't be processed during Astro's build process. These files will be copied into the build folder untouched.
```
sarah11918, line 88: I prefer the original! 

You could say "that you do not want processed during Astro's build", but I think that's equally as good as the original. 

"Should" can be problematic in English. This sentence COULD actually mean... they SHOULDN'T get processed, but you never know what can happen..."  :) So, I'd prefer to keep it very clear. These are files that don't need processing, or these are files that you don't want processed.

src/pages/en/core-concepts/project-structure.md
```diff
@@ -85,19 +85,19 @@ It is a common convention to store your CSS or Sass files in a `src/styles` dire
...
 
- 💡 *As a general rule, any CSS or JavaScript that you write yourself should live in your `src/` directory.*
+> 💡 *As a general rule, any CSS or JavaScript that you write yourself should live in your `src/` directory.*
 
 ### `package.json`
 
 This is a file used by JavaScript package managers to manage your dependencies. It also defines the scripts that are commonly used to run Astro (ex: `npm start`, `npm run build`).
 
-For help creating a new `package.json` file for your project, check out the [manual setup](/en/install/manual/) instructions.
+For help regarding the creation of a new `package.json` file for your project, check out the [manual setup](/en/install/manual/) instructions.
```
sarah11918, line 100: This is not in keeping with the style of the document, and it would be considered too verbose/formal in English. The original stays!

sarah11918:
Thanks, @Chrissdroid I've made some comments on the changes that I do NOT accept as is, and the rest are great!

src/pages/en/core-concepts/project-structure.md
Chrissdroid: Ohh, I see

Chrissdroid:
Reverted !

sarah11918:
Great! Thanks! Just waiting for the checks!

PR #687, opened 2022-05-29
update the config-reference generated header for i18n

sarah11918:
This file is auto-generated (never manually edited) in English, but we decided that translations would be manual.

Adding `i18nReady: true` to the generated header so that this page shows up in our pinned status tracking issue for translators.

Paging Dr. @delucis for approval. 😄

### DIFF
--- a/scripts/docgen.mjs
+++ b/scripts/docgen.mjs
@@ -11,6 +11,7 @@ const HEADER = `---
 
 layout: ~/layouts/MainLayout.astro
 title: Configuration Reference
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM! We will also need to re-run the docgen script to compile this change to the Markdown file:

```
node scripts/docgen.mjs
```

We can probably do that in a separate PR though as it will include all recent changes not just the `i18nReady` flag. Happy to tackle that when I land tomorrow if you like.

Issue #620, opened 2022-05-20
a11y issues/improvements


mayank99:
List of accessibility issues I ran into when using the docs site:
- The table of contents ("on this page") section is at the very end of the document. This also includes the theme switcher.
  - I'm guessing this was done to avoid having to tab through the whole list before getting to the main content, but it arguably serves no purpose anymore. One way to fix this would be add another skip link dedicated for this table. 
- "Astro" and "Docs" links at the top left lack visible focus indicator.
- "Aside" components (e.g. alert or informational) use the `<blockquote>` element which is semantically incorrect.
  - Some of them are also italicized not just stylistically but also semantically using the `<em>` element, which seems wrong.
- Code blocks are focusable, even though there is no interaction within them.

Also, some of the borders/underlines get messed up when using Windows forced high contrast mode. e.g. you can't tell which tab is active.
![](https://user-images.githubusercontent.com/9084735/169622849-7de3885f-8679-4b7e-b41b-8e057ca869d9.png)
![](https://user-images.githubusercontent.com/9084735/169622824-bf8132fb-7295-4514-8ac2-7b0c5ab87cd1.png)

There are also some forced contrast issues in the search popup but I'm guessing the UI is not customizable so I'd have to file that under the algolia repo.

---

P.S. I would love to create PRs to fix some of these things myself.

delucis:
Awesome! Thanks so much or putting together this list 💜

- I suspect the table of contents DOM position was just to have the three columns in source order as we switch it out to before the main article content on mobile. So tab order actually varies depending on viewport width.

- The code blocks are made focusable here: https://github.com/withastro/docs/blob/3ffa4f660f7b24245bfc39feeab279acf5eba84f/public/make-scrollable-code-focusable.js
   I think the idea is that scrollable elements should be able to receive focus and some code blocks overflow so need this?

- #401 also suggests replacing `<blockquote>` usage.

Happy to look at PRs and thanks again for the attention to detail!

mayank99:
@delucis I've created #692 to fix some of the issues. It's not much but it's a start. I might tackle forced colors mode in a separate PR.

> I suspect the table of contents DOM position was just to have the three columns in source order as we switch it out to before the main article content on mobile. So tab order actually varies depending on viewport width.

I've fixed the order on desktop but left it as-is on mobile, so now it comes before the main content on both. We might be possibly able to use the same DOM element but I left the duplication for now.

[...]

PR #692, opened 2022-05-29
fix: Improve a11y

mayank99:
### What kind of changes does this PR include?

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

### Description

General accessibility improvements to the issues found in #620. No visible changes other than the focus indicators described below.

#### `MainLayout.astro`

Reordered right sidebar to come before the main page content. This makes the DOM order more sensible. No CSS changes were necessary because it already has a grid-column set.

This also affects the "More" list and the theme button, which I feel like do not belong inside a "table of contents" section, but I didn't want to make huge visual changes.

Also removed the `title` attribute from both sidebars because it didn't make sense to me and was causing duplicate announcements in NVDA.

#### `LeftSidebar.astro` and `RightSidebar.astro`

Fixed accessible name. Previously, NVDA was announcing the entire contents of the `nav` instead of just "Site navigation" or "Table of contents".

#### `Header.astro`

Added visible focus indicator to logo

| Before | After |
| --- | --- |
|  |  |

#### `LanguageSelect.css`

Updated focus indicator of select button to be more prominent

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/components/Header/Header.astro
+++ b/src/components/Header/Header.astro
@@ -128,9 +128,12 @@ const t = useTranslations(Astro);
 
 	.logo a:hover,
 	.logo a:focus {
-		outline: none;
 		opacity: 1;
 		transform: translateY(-2px);
+	}
+
+	.logo a:focus:not(:focus-visible) {
+		outline: none;
 	}
 
 	.logo h1 {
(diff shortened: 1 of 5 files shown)


delucis (approved):
Excellent work! Thanks @mayank99 💜

PR #710, opened 2022-06-02
Use eslint

### LINKED ISSUE #634: Use new ESLint plugin to dogfood it
@FredKSchott [suggested on Discord](https://discord.com/channels/830184174198718474/830184175176122389/978400069042704464) we add the new ESLint plugin to dogfood it and give feedback to the author. Plugin repo is here: https://github.com/ota-meshi/eslint-plugin-astro

ota-meshi:

- [ ] New or updated content
- [ ] Translated content
- [X] Something else!

#### Description

close #634

This PR introduces ESLint to the project. And, add the npm script `"lint:eslint"` to run ESLint.
I also made some simple fixes to some of the warned files.

However, the eslint configuration may need to be changed. Currently, it is based on the recommended configuration of eslint and each eslint plugin.

There are no visual changes.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,5 +2,4 @@ dist
 node_modules
 .github
 .changeset
-*.md
 /src/pages/lighthouse
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,6 @@
+dist
+node_modules
+.github
+.changeset
+*.md
+/src/pages/lighthouse
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,6 +12,15 @@ module.exports = {
   rules: {},
   overrides: [
     {
+      files: ["*.astro"],
+      parser: "astro-eslint-parser",
+      parserOptions: {
+        parser: "@typescript-eslint/parser",
+        extraFileExtensions: [".astro"],
+      },
+      rules: {},
+    },
+    {
       files: ['*.ts'],
       parser: '@typescript-eslint/parser',
       extends: ['plugin:@typescript-eslint/recommended'],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,24 @@
+module.exports = {
+  env: {
+    node: true,
+    es2022: true,
+    browser: true,
+  },
+  extends: ['eslint:recommended', 'plugin:astro/recommended'],
+  parserOptions: {
+    ecmaVersion: 'latest',
+    sourceType: 'module',
+  },
+  rules: {},
+  overrides: [
+    {
+      files: ['*.ts'],
+      parser: '@typescript-eslint/parser',
+      extends: ['plugin:@typescript-eslint/recommended'],
+      rules: {
+        '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_', destructuredArrayIgnorePattern: '^_' }],
+        '@typescript-eslint/no-non-null-assertion': 'off',
+      },
+    },
+  ],
+};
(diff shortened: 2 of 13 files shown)


ota-meshi:
If you are using VS Code and ESLint extension, you can also check in the editor by adding the following settings to `.vscode/settings.json`.

```json
{
  "eslint.validate": [
    "javascript",
    "javascriptreact",
    "astro",
    "typescript",
    "typescriptreact"
  ]
}
```

https://github.com/ota-meshi/eslint-plugin-astro#visual-studio-code

.eslintignore
```diff
@@ -0,0 +1,6 @@
+dist
+node_modules
+.github
+.changeset
+*.md
```
delucis, line 5: Do we need this? I tried removing it and seemed to run fine — I assume ESLint ignores Markdown files by default?

```suggestion
```

.eslintrc.js
```diff
@@ -0,0 +1,33 @@
...
+  overrides: [
+    {
+      files: ["*.astro"],
+      parser: "astro-eslint-parser",
+      parserOptions: {
+        parser: "@typescript-eslint/parser",
+        extraFileExtensions: [".astro"],
+      },
+      rules: {},
+    },
```
delucis, line 22: Just a couple of questions about the eslint plugin.

Is it possible for `plugin:astro/recommended` to set this for us? Or can overrides like this only be set by a user?

And do you always need `@typescript-eslint` for Astro files? If so, maybe that could be a dependency of `astro-eslint`?

delucis (commented):
Hi @ota-meshi!

Thanks so much for taking the time to add this 🙌 

I have a couple of questions about the config/plugin, but in general this looks great.

.eslintignore
ota-meshi: You're right. It is not necessary. I copied it from `.prettierignore` and forgot to remove it.

.eslintrc.js
ota-meshi: This parser configuration is actually already included in `"plugin:astro/recommended"`. If we do not change the current configuration, we may not need to add the parser configuration to `.eslintrc.*`.
However, the parser configuration is cannot merged with other configurations, so if we don't explicitly configure it in `overrides`, a simple configuration change will break it. For example, adding a shared configuration with a parser configuration to `extends` will prevent eslint from parsing `.astro`.
Therefore, I recommend explicitly configuring the parser configs for `overrides`.

eslint-plugin-astro depends on `@typescript-eslint/parser` as peerDeps. I think this is necessary because we can always write TypeScript in `.astro`. Are there many astro users who don't actually use TypeScript? In that case, we may need to consider changing it so that it doesn't depend on `@typescript-eslint/parser`.

delucis: Thanks for the clarification. I think there are definitely people who don’t write TypeScript with Astro (or don’t _know_ they are writing TypeScript!) I don’t think you need to stop depending on `@typescript-eslint/parser` though, the easier thing for people less familiar with TS would be if they didn’t have to manually install `@typescript-eslint`, so moving it from `peerDependencies` to `dependencies` seems like a good idea to me.

But we can continue this discussion on your repo. For this PR, this looks good to me!

delucis (approved):

.eslintrc.js
ota-meshi: That's common argument in the eslint plugin community 😅 . If we use `dependencies`, they may be placed deep in `node_modules` and not use a version of the parser that you did not intend. Therefore, it is safer to make it `peerDependencies`.
Even now, eslint-plugin-vue still depends on vue-eslint-parser with `dependencies`, but users often complain that an unintended version is used :sweat:.

delucis: 😅! In general I prefer to use dependencies to simplify installation except for clear cases where the dependency is a bigger part of the user’s usage. So Astro or React would be peer deps because they’re the big important piece, and in this case, `eslint` should be a peer dep, but specific parsers seem safe in dependencies. I’ll trust you on this though!

PR #715, opened 2022-06-03
ci: Add check to run ESLint on PRs/commits

delucis:

- [ ] New or updated content
- [ ] Translated content
- [x] Something else!

#### Description

#710 added a script to allow linting `.astro` and `.js`/`.ts` files. Ideally we should keep on top of any lint errors before they proliferate, so this PR adds a job to our CI workflow to run the new `pnpm run lint:eslint` command for PRs/commits.

This could introduce a little extra friction to PRs if people need to fix lint errors before merging, but we’ve only enabled a fairly relaxed set of linter rules. You can see these rules here: https://eslint.org/docs/rules/ (marked with ✓). As this and the Astro-ESLint parser is new we may need to tweak the rules as we go, but my idea is that we can try and come up with a best practice config for others to use by dogfooding this. This also doesn’t impact Markdown files, so I don’t think we’re introducing extra friction for content authors, it’s more for layout, component, and other infrastructure code.

(You can see the new CI check running in the checks on this PR.)


sarah11918:
LGTM! Happy to be using @ota-meshi 's fantastic contribution! 🥳

PR #720, opened 2022-06-05
Fix scrollbar overlapping issue on Firefox

Yan-Thomas:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description
This PR adds padding to the `RightSidebar` component, so the component scrollbar (and the page scrollbar) don't overlap (if they're overlapping you can only drag the page scrollbar). This change only affects Firefox, since Chrome scrollbars automatically add padding to elements, unlike Firefox. This is important because some people can't use the mouse scroll wheel or are in their sister's laptops where the touchpad doesn't work well (this is not personal experience)

Here's a gif demonstrating what happens:

|  Before  | PR  |
|------------|-------|
|  |   |

Before merging, I want to confirm two things:

- Does this same problem happen in other non-Chromium browsers like Safari?
- Maybe we should add padding not only in browsers where scrollbars overlap but to all of them, just because this way both sidebars have the same padding **AND** we avoid having styles for specific browsers?

### DIFF
--- a/src/components/RightSidebar/RightSidebar.astro
+++ b/src/components/RightSidebar/RightSidebar.astro
@@ -4,7 +4,9 @@ import ThemeToggleButton from './ThemeToggleButton.tsx';
 import ContributeMenu from './ContributeMenu.astro';
 import CommunityMenu from './CommunityMenu.astro';
 import { useTranslations } from "../../i18n/util.ts";
+
 const t = useTranslations(Astro);
+
 const { content, githubEditUrl } = Astro.props;
 const headers = content.astro?.headers;
 ---
--- a/src/components/RightSidebar/RightSidebar.astro
+++ b/src/components/RightSidebar/RightSidebar.astro
@@ -14,19 +14,14 @@ const headers = content.astro?.headers;
 <nav class="sidebar-nav" aria-label={t('rightSidebar.a11yTitle')}>
 	<div class="sidebar-nav-inner">
 		{headers && (
-			<TableOfContents
-				client:media="(min-width: 50em)"
-				headers={headers}
-				labels={{ onThisPage: t('rightSidebar.onThisPage'), overview: t('rightSidebar.overview') }}
-			/>
+		<TableOfContents client:media="(min-width: 50em)" headers={headers} labels={{ onThisPage:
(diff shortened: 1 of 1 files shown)


src/components/RightSidebar/RightSidebar.astro
```diff
@@ -14,19 +14,14 @@ const headers = content.astro?.headers;
...
 		<ContributeMenu editHref={githubEditUrl} />
-		<CommunityMenu/>
+		<CommunityMenu />
 		<div style="margin: 2rem 0; text-align: center;">
-			<ThemeToggleButton 
-				client:visible 
-				labels={{  useLight: t('themeToggle.useLight'), useDark: t('themeToggle.useDark')  }} 
-			/>
+			<ThemeToggleButton client:visible labels={{ useLight: t('themeToggle.useLight'), useDark:
+				t('themeToggle.useDark') }} />
```
delucis, line 24: Ooh, this formatting is a little odd. Is this the auto-formatting the VS Code extension now provides?

src/components/RightSidebar/RightSidebar.astro
```diff
@@ -37,10 +32,18 @@ const headers = content.astro?.headers;
...
 	}
-</style>
+
+	/* style applied only on Firefox, this avoids scrollbars overlapping. */
+	@supports (-moz-appearance: none) {
+		.sidebar-nav {
+			padding-inline-end: 1.5rem;
+		}
+	}
+</style>
```
delucis, line 49: Looks good!

I tested in Safari and it seems to follow Chromium in automatically adding padding around scrollbars.

I also tried adding this universally but it’s visually pretty unsatisfactory as you get loads of extra space in odd places, so I think your solution is probably best! 🙌

delucis (commented):
Thanks @Yan-Thomas! So happy you didn’t have to survive difficulties like borrowing laptops with bad touchpads. We should leave slightly broken input devices as Sarah’s expertise 💜

src/components/RightSidebar/RightSidebar.astro
Yan-Thomas: Maybe it's from Gitpod? I will fix this real quick.

delucis (approved):
Perfect!

PR #735, opened 2022-06-07
Change `ThemeToggleButton` positioning

Yan-Thomas:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

Changes made:
- Moved ThemeToggleButton component to the Header directory (since it doesn't even appear in the RightSidebar now).
- Moved ThemeToggleButton to the header on larger viewports.
- Moved ThemeToggleButton to the bottom of the pop-up menu on smaller viewports.
- Added extra bottom padding to the RightSidebar since it doesn't have the ThemeToggle padding anymore, so the Community Menu has some good whitespace to breath.

mcAHM76

### DIFF
--- a/src/components/Header/ThemeToggleButton.css
+++ b/src/components/Header/ThemeToggleButton.css
@@ -27,6 +27,24 @@
 	opacity: 1;
 }
 
+@media (max-width: 50em) {
+	.theme-toggle {
+		margin-top: 32px;
+	}
+	.hide-toggle-on-smaller-screens {
+		display: none;
+	}
+}
+
+@media (min-width: 50em) {
+	.theme-toggle {
+		display: none;
+	}
+	.hide-toggle-on-smaller-screens {
+		display: inline-flex;
+	}
+}
+
 input[name='theme-toggle'] {
 	position: absolute;
 	opacity: 0;
--- a/src/components/RightSidebar/RightSidebar.astro
+++ b/src/components/RightSidebar/RightSidebar.astro
@@ -1,6 +1,5 @@
 ---
 import TableOfContents from './TableOfContents.tsx';
-import ThemeToggleButton from './ThemeToggleButton.tsx';
 import ContributeMenu from './ContributeMenu.astro';
 import CommunityMenu from './CommunityMenu.astro';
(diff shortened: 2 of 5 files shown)


sarah11918:
Tried it at a view sizes in the browser, and on my phone, and this is looking nice, @Yan-Thomas !

src/components/Header/ThemeToggleButton.css
```diff
@@ -27,6 +27,24 @@
...
+}
+
+@media (min-width: 50em) {
+	.theme-toggle {
+		display: none;
+	}
+	.hide-toggle-on-smaller-screens {
+		display: inline-flex;
+	}
+}
```
delucis, line 46: We have an edge case at 800px wide where the margin-top persists and the toggle is already in the top bar:

image

I think we can avoid the `max-width` query and just progressively enhance from mobile first:

```suggestion
.theme-toggle {
	margin-top: 32px;
}
.hide-toggle-on-smaller-screens {
	display: none;
}


@media (min-width: 50em) {
	.theme-toggle {
		display: none;
		margin-top: 0;
	}
	.hide-toggle-on-smaller-screens {
		display: inline-flex;
	}
}
```

(The `margin-top` should probably be in the main `.theme-toggle` declaration above but I can’t do that in a suggestion here.)

src/components/RightSidebar/RightSidebar.astro
```diff
@@ -41,7 +34,7 @@ const headers = content.astro?.headers;
 	.sidebar-nav-inner {
 		height: 100%;
 		padding: 0;
```
delucis, line 36: Looks like we now have `padding` twice so we can probably remove this?

```suggestion
```

delucis (changes requested):
Awesome work @Yan-Thomas! Thanks so much for tackling this. There are a couple of small issues, but this looks like a great improvement 🚀

Yan-Thomas:
Thanks, @sarah11918 and @delucis!!

I have applied the suggestions, let me know if there's anything else to solve.

delucis:
Thanks so much for this @Yan-Thomas — amazing improvement 💜

Issue #595, opened 2022-05-17
`Astro.slots` example is not working


jcha0713:
The example code for `Astro.slots` doesn't seem to work, and I'm not sure whether the given example is wrong or the api has a bug.

The code I tested is:

```astro
---
// components/Message.astro

let html: string = '';
if (Astro.slots.has('default')) {
  html = await Astro.slots.render('default', Astro.props.messages)
}
---
<Fragment set:html={html} />
```

```astro
---
// pages/index.astro

import Message from '../components/Message.astro'
---
<div>
  <Message messages={['Hello', 'world!']}>{(messages) => messages.join(' ')}</Message>
</div>

<!-- 
must be renders as
<div>Hello world!</div>
-->

```

This example gives me an error that says `message.join is not a function`.
If I have made any mistake, please let me know.
Thank you!

link to the doc: https://docs.astro.build/en/reference/api-reference/#astroslots
minimal reproduction: https://codesandbox.io/s/distracted-shamir-hpxuel?file=/src/pages/index.astro

sarah11918:
I'm confirming the behaviour in the CodeSandbox. I'm getting `messages.join is not a function`

Furthermore, when I play around with variations of this, I'm getting `messages is not defined`, so I think the error is less about not being able to perform the join, and more about `messages` itself being undefined (and THAT's why `messages.join` is not a function.)

sarah11918:
@aFuzzyBear I think you were going to take a look at this one?

delucis:
Raising an issue on the main Astro repo for this as it is likely a bug. Here’s a Stackblitz reproduction:

https://stackblitz.com/edit/github-csutf1?file=src%2Fpages%2Findex.astro,src%2Fcomponents%2FMessage.astro

sarah11918:
Example in the docs has been commented out. When https://github.com/withastro/astro/issues/3562 is fixed, we can go back and revive the example.

Untill then, labelled as "on hold."

Princesseuh:
Hello! An update on this, the function works as expected, however the second argument of `Astro.render.slots` is an array of argument to pass, as such when trying to pass an array as a single parameter, it needs to be an item of the array:

So in this case, it needs to be the following:
```ts
html = await Astro.slots.render('default', [Astro.props.messages])
```

(Note the square brackets around `Astro.props.messages`!)

Jutanium:
[...]

PR #740, opened 2022-06-08
commenting out broken astro.self example

sarah11918:
In response to https://github.com/withastro/docs/issues/595 re: non-working code example 

- [ x] New or updated content

The offending code example has been commented out, pending a bug fix.https://github.com/withastro/astro/issues/3562

As noted in the code itself, in order to comment out code that had its own code comments... I had to get creative. Please take care when reverting.


delucis (approved):
Very creative! Thanks for putting that together so quickly.

LGTM!

PR #742, opened 2022-06-09
fix Astro:self example

### LINKED ISSUE #577: Astro:self example does not work
The current example in https://docs.astro.build/en/reference/api-reference/#astroself does not work.

Reproduction: https://codesandbox.io/s/astro-self-example-3qv8f2?file=/src/components/NestedList.astro

This causes an error _"unexpected return"_.

```astro
---
const { items } = Astro.props;
---
<ul class="nested-list">
  <li>{items.map((item) => {
    if (Array.isArray(item)) {
      // If there is a nested data-structure we render `<Astro.self>`
      // and can pass props through with the recursive call
      return <Astro.self items={item} />;
    } else {
      return item;
    }
  })}</li>
</ul>
```

The only way I got it to work was by returning a fragment/element directly from the mapping and do the recursive breaking conditional check _inside_ the fragment/element.

```astro
---
const { items } = Astro.props;
---
<ul class="nested-list">
  {items.map((item) => (
    <li>
      {Array.isArray(item) ? (
        <Astro.self items={item} />
      ) : (
        item
      )}
    </li>
  ))}
</ul>
```

ekafyi:
Closes #577

- [x] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

Update `Astro:self` code example. The existing example does not work due to a known existing bug. Reproduction: https://codesandbox.io/s/astro-self-example-3qv8f2?file=/src/components/NestedList.astro

No visual change.

### DIFF
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -196,15 +196,19 @@ tick tick tick astro
 const { items } = Astro.props;
 ---
 <ul class="nested-list">
-  <li>{items.map((item) => {
-    if (Array.isArray(item)) {
-      // If there is a nested data-structure we render `<Astro.self>`
-      // and can pass props through with the recursive call
-      return <Astro.self items={item} />;
-    } else {
-      return item;
-    }
-  })}</li>
+  {items.map((item) => (
+    <li>
+      {/* 
+        If there is a nested data-structure we render `<Astro.self>`
+        and can pass props through with the recursive call
+      */}
+      {Array.isArray(item) ? (
+        <Astro.self items={item} />
+      ) : (
+        item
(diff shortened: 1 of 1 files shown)


delucis (commented):
Looks great @ekafyi! Excited you’re going to become a docs contributor! 🥳 

I’ve made on suggestion for the comment syntax.

@sarah11918 We need the ternary instead of if/else here because we’re explicitly avoiding using `return`, so that looks good to me 👍

ekafyi:
Done! @sarah11918 @delucis

sarah11918:
Let's Get That Money, @ekafyi !  (Oh wait... you already did! 💅 )

PR #743, opened 2022-06-09
Convert bullet points to subsections in Importing Markdown

### LINKED ISSUE #669: Reformat markdown-content page for better visual organization
See the content/comment here: https://github.com/withastro/docs/pull/668#issue-1249627394

Specifically, we should start by replacing bullet points with headers (level 4) and see whether that fixes both the ability to scan the contents better, and also that some of these items are long enough that they could now do with multiple paragraphs to visually space out all the content. (And that's harder to manage within a bullet point.)

If that's not the instant solution, then we just figure out how to make the page nicer to read.  :)

This should NOT be marked as a minor change when it happens, so that it triggers our "Languages need updating" Issue.

felixsanz:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

Converted bullet points from Importing Markdown section into subsections. Fixes #669

N/A

### DIFF
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -198,7 +198,7 @@ author: Leon
 
 ## Importing Markdown
 
-You can import Markdown files directly into your Astro files! You can import one specific page with `import` or multiple with `Astro.glob()`
+You can import Markdown files directly into your Astro files! You can import one specific page with `import` or multiple with `Astro.glob()`.
 
 ```astro
 ---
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks, @felixsanz ! This is looking much cleaner!

@delucis @aFuzzyBear - With `<h3>`'s, each of these exported properties becomes their own entry in the right sidebar. What do we think about this? I'm not sure whether we want this, especially given what it would do on mobile.

One option would be to add something like `### Exported Properties` and then the individual properties themselves can all become `<h4>` and would NOT show up in the right sidebar. Thoughts?

felixsanz:
@sarah11918 @delucis Done!

aFuzzyBear:
this does make it look alot cleaner, wanna check the toc component on desktop when I'm back on,

felixsanz:
> this does make it look alot cleaner, wanna check the toc component on desktop when I'm back on,

i'm on desktop, they don't show now:

delucis (approved):
LGTM — thanks Felix!

PR #746, opened 2022-06-09
Chore: Updating to latest Astro beta release

tony-sull:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

- Updates `astro` and all `@astrojs` dependencies to the latest release
- Updates all external dependencies that only had a minor version bump
- Updates `@docsearch/react` with support for react v18

To-do in a future PR: 
- update all dependencies to the latest
- hook up dependabot to track all outdated dependencies


delucis:
Looks like we have caught a rare silent Astro bug! 🦋 

Check out the bottom of this page: https://deploy-preview-746--astro-docs-2.netlify.app/en/migrate/#deprecated-script-default-processing

Seems like Markdown rendering broke halfway through but without actually erroring. May happen elsewhere and we just caught this cause some pages are trying to link to the content that’s now missing.

delucis:
The Markdown where it is choking looks like this. Everything after the heading is omitted.

````md
### Deprecated: `<script>` Default Processing

Previously, all `<script>` elements were read from the final HTML output and processed + bundled automatically. This behavior is no longer the default. Starting in 0.24, you must opt-in to `<script>` element processing via the `hoist` attribute. The `type="module"` is also required for hoisted modules.

```astro
<script>
  // Will be rendered into the HTML exactly as written!
  // ESM imports will not be resolved relative to the file.
</script>
<script type="module" hoist>
  // Processed! Bundled! ESM imports work, even to npm packages.
</script>
```
````

@hippotastic I know you’ve been tackling Astro + Markdown issues, does this smell familiar to you?

hippotastic:
@delucis Yes, I think I fixed this bug already. Note that our Docs repo is still on 1.0.0-beta.17. Can you try updating to the latest beta and see if that fixes the issue?

hippotastic:
Oops, sorry about that. I just quickly glanced at the error and didn't check the full context myself. 😂

This error seems to be introduced not by Astro, but by our docs site code here:
https://github.com/withastro/docs/blob/a09930dac00e058936a3970dadd929056109b963/astro.config.ts#L28

When building the screen reader text for the heading, we neet to escape the `<` into a `&lt;` as well. A simple global regex string replacement of `text` will do.

hippotastic:
By the way, I cannot recommend to merge this upgrade yet. I'm just working on fixing yet another Markdown bug (see our dev channel) which causes HTML comments inside fenced code blocks to be broken:
image

This must be fixed on the Markdown tokenizer level and is tracked in this issue: https://github.com/withastro/astro/issues/3529

delucis:
Great to know on both counts — thanks @hippotastic! 💜

tony-sull:
Thanks for the quick investigation here, @delucis and @hippotastic!

Good to know that #3529 will be a blocker for updating docs.  I flipped this PR over to a draft, just need to track down a break related to escaping `.sr-only` text then we can leave this as a draft until the last markdown issue is unblocked :+1:

hippotastic:
Sorry, I somehow pasted the wrong related issue ID. The HTML comment issue is tracked here: https://github.com/withastro/astro/issues/3460

delucis:
In order to get this merged I’m leaving `@astrojs/sitemap` on 0.1.2 as I think we can do some cool stuff with 0.2.0 regarding i18n, but I don’t want to hold this up for that.

tony-sull:
Oh wow, just came back to merge this in after the latest and you already took care of it!  Thanks @delucis!

PR #747, opened 2022-06-09
Fix recommended import styles for js, ts, jsx and tsx files

delucis:

- [X] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

This PR updates how we recommend importing JavaScript, Typescript, JSX and TSX modules based on consultation with @Princesseuh last week.

The one potential point of confusion is importing `.ts` which in general should be done _without_ an extension to bring in types etc., but in `<script>` tags should _include_ the extension. I’ve gone with recommend the basics, flag up the `<script>` tag exception, but I’d like to know if we need to maybe explain in the standard case that omitting the extension is good because we get types etc.

### DIFF
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -49,13 +49,11 @@ To use a framework component, import it from its relative path in your Astro com
 ```astro
 ---
 import MyReactComponent from '../components/MyReactComponent.jsx';
-import MySvelteComponent from '../components/MySvelteComponent.svelte';
 ---
 <html>
   <body>
-    <h1>Use React & Svelte components directly in Astro!</h1>
+    <h1>Use React components directly in Astro!</h1>
     <MyReactComponent />
-    <MySvelteComponent />
   </body>
 </html>
 ```
@@ -77,8 +75,8 @@ Most directives will render the component on the server at build time. Component
 ```astro
 ---
 // Example: hydrating framework components in the browser.
-import InteractiveButton from '../components/InteractiveButton';
-import InteractiveCounter from '../components/InteractiveCounter';
+import InteractiveButton from '../components/InteractiveButton.jsx';
+import InteractiveCounter from '../components/InteractiveCounter.jsx';
 ---
 <!-- This component's JS will begin importing when the page loads -->
 <InteractiveButton client:load />
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -48,7 +48,7 @@ To use a framework component, import it from its relative path in your Astro com
 
 ```astro
 ---
-import MyReactComponent from '../components/MyReactComponent';
+import MyReactComponent from '../components/MyReactComponent.jsx';
 import MySvelteComponent from '../components/MySvelteComponent.svelte';
 ---
 <html>
@@ -106,7 +106,7 @@ You can import and render components from multiple frameworks in the same Astro 
 ---
 // src/pages/MyAstroPage.astro
 // Example: Mixing multiple framework components on the same page.
-import MyReactComponent from '../components/MyReactComponent';
+import MyReactComponent from '../components/MyReactComponent.jsx';
 import MySvelteComponent from '../components/MySvelteComponent.svelte';
 import MyVueComponent from '../components/MyVueComponent.vue';
 ---
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -44,7 +44,7 @@ export default defineConfig({
 
 Use your JavaScript framework components in your Astro pages, layouts and components just like Astro components! All your components can live together in `/src/components`, or can be organized in any way you like.
 
-To use a framework component, import it from its relative path (including file extension) in your Astro component script. Then, use the component alongside other components, HTML elements and JSX-like expressions in the component template.
+To use a framework component, import it from its relative path in your Astro component script. Then, use the component alongside other components, HTML elements and JSX-like expressions in the component template.
 
 ```astro
 ---
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -53,8 +53,9 @@ import MySvelteComponent from '../components/MySvelteComponent.svelte';
 ---
 <html>
   <body>
-    <h1>Use React components directly in Astro!</h1>
+    <h1>Use React & Svelte components directly in Astro!</h1>
     <MyReactComponent />
+    <MySvelteComponent />
   </body>
 </html>
 ```
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -49,6 +49,7 @@ To use a framework component, import it from its relative path in your Astro com
 ```astro
 ---
 import MyReactComponent from '../components/MyReactComponent';
+import MySvelteComponent from '../components/MySvelteComponent.svelte';
 ---
 <html>
   <body>
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -48,7 +48,7 @@ To use a framework component, import it from its relative path (including file e
 
 ```astro
 ---
-import MyReactComponent from '../components/MyReactComponent.jsx';
+import MyReactComponent from '../components/MyReactComponent';
 ---
 <html>
   <body>
@@ -75,8 +75,8 @@ Most directives will render the component on the server at build time. Component
 ```astro
 ---
 // Example: hydrating framework components in the browser.
-import InteractiveButton from '../components/InteractiveButton.jsx';
-import InteractiveCounter from '../components/InteractiveCounter.jsx';
+import InteractiveButton from '../components/InteractiveButton';
+import InteractiveCounter from '../components/InteractiveCounter';
 ---
 <!-- This component's JS will begin importing when the page loads -->
 <InteractiveButton client:load />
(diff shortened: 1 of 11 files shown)


sarah11918:
Looks great, and we should also update our strong wording in a couple of places that components must be imported using relative file path **including full extension**. I guess now the instruction is "except for ..."

- framework-components
- troubleshooting

src/pages/en/guides/troubleshooting.md
```diff
@@ -70,7 +70,7 @@ Then check your import statement:
 
 - Does your import have the same name as the imported component? (Check your component name and that it [follows the `.astro` syntax](/en/comparing-astro-vs-other-tools/#astro-vs-jsx).)
 
-- Have you included the extension in the import? (Check that your imported file contains an extension. e.g. `.astro`, `.md`, `.jsx`, `.vue`)
+- Have you included the extension in the import? (Check that your imported file contains an extension. e.g. `.astro`, `.md`, `.vue`, `.svelte`)
```
sarah11918, line 73:
```suggestion
- Have you included the extension in the import? (Check that your imported file contains an extension. e.g. `.astro`, `.md`, `.vue`, `.svelte`. Note: no file extension is required for `.jsx` files.)
```

delucis: Do we need to also mention `.tsx`?

```suggestion
- Have you included the extension in the import? (Check that your imported file contains an extension. e.g. `.astro`, `.md`, `.vue`, `.svelte`. Note: no file extension is required for `.jsx` and `.tsx` files.)
```

sarah11918: Sounds good. Given that this section is specifically about importing **components**, I'm assuming we don't need to worry about mentioning `.js`, `.ts` here.

src/pages/en/core-concepts/framework-components.md
```diff
@@ -44,11 +44,11 @@ export default defineConfig({
 
 Use your JavaScript framework components in your Astro pages, layouts and components just like Astro components! All your components can live together in `/src/components`, or can be organized in any way you like.
 
-To use a framework component, import it from its relative path (including file extension) in your Astro component script. Then, use the component alongside other components, HTML elements and JSX-like expressions in the component template.
+To use a framework component, import it from its relative path in your Astro component script. Then, use the component alongside other components, HTML elements and JSX-like expressions in the component template.
 
 ```astro
 ---
-import MyReactComponent from '../components/MyReactComponent.jsx';
+import MyReactComponent from '../components/MyReactComponent';
```
sarah11918, line 51:
```suggestion
import MyReactComponent from '../components/MyReactComponent';
import MySvelteComponent from '../components/MySvelteComponent.svelte';
```

(and update the templating)

Do we want to do this, right off the bat, so at least it's OBVIOUS that sometimes there's a file extension, sometimes there isn't? 

I would really want some kind of call out below that explicitly says when an extension is/isn't required. But, I worry about a single example of React, right off the bat, when it's the ONLY one that does NOT require the file extension. It will be too easy for people to get the wrong pattern at a glance?

delucis: Good shout

sarah11918:
Just looking for a few other places where it might be helpful to hit the "no extension thing." (Bold emphasis just for highlighting a suggested place for additional not, not to be bold in final copy). These are the two other places an additional note could be helpful?

Imports Page:
JSX / TSX
[Section titled JSX / TSX](https://deploy-preview-747--astro-docs-2.netlify.app/en/guides/imports/#jsx--tsx)
import { MyComponent } from './MyComponent';
Astro includes built-in support for JSX (*.jsx and *.tsx) files in your project. **Import them into any `.astro` component without file extension, and the** JSX syntax is automatically transpiled to JavaScript.

TypeScript Page
Astro ships with built-in support for [TypeScript](https://www.typescriptlang.org/). You can import .ts and .tsx files in your Astro project **(no file extension required)**, and even write TypeScript code directly inside your [Astro component](https://deploy-preview-747--astro-docs-2.netlify.app/en/core-concepts/astro-components/#the-component-script).

Princesseuh:
In my opinion, such heavy emphasis shouldn't be put on the no extension thing, as importing with a `.jsx` / `.js` (for TSX and TS respectively) extension is also totally valid 

Additionally, soon we might migrate to using TS  4.7's node16 module resolution which requires the extension (so .js for .ts, .jsx for .tsx) so it's best to not hammer the no extension thing too hard (or we'll run in the same situation we have now where people use .ts extensions because they think it was required)

Princesseuh:
Importing a .ts file with .js should work, it was fixed a week ago: https://github.com/withastro/astro/pull/3518

delucis:
> Importing a .ts file with .js should work, it was fixed a week ago: https://github.com/withastro/astro/pull/3518

Ah, OK. I tried in the docs repo and we’re stuck on an old beta version at the moment. I see in that issue that that doesn’t work in `<script>` tags, yet. Hmm.

You know what? The current advice doesn’t _break_ anything — you just get some red squigglies importing TS with the extension. I think it might be a good idea to leave it as is and wait for `script` tags to reach parity so we can give consistent advice everywhere about using `.js`/`.jsx` for Typescript and everything else remains unchanged.

src/pages/en/guides/imports.md
```diff
@@ -39,15 +39,25 @@ JavaScript can be imported using normal ESM `import` & `export` syntax. This wor
...
 
-**Astro doesn't perform any type checking itself.** Type checking should be taken care of outside of Astro, either by your IDE or through a separate script. The [Astro VSCode Extension](/en/editor-setup/) automatically provides TypeScript hints and errors in your open files.
+**Astro doesn't perform any type checking itself.** Type checking should be taken care of outside of Astro, either by your IDE or through a separate script. For type checking Astro files, the [`astro check` command](/en/reference/cli-reference/#astro-check) is provided.
 
-📚 Read more about [TypeScript support in Astro.](/en/guides/typescript/)
+:::note[TypeScript and file extensions]
+Per [TypeScript's module resolution rules](https://www.typescriptlang.org/docs/handbook/module-resolution.html), the `.ts` or `.tsx` file extensions should not be used when importing TypeScript files. Instead TypeScript recommend to either use the `.js` and `.jsx` file extensions respectively or completely omit the file extension
```
delucis, line 51: Trying to decide whether “TypeScript” is plural or singular for conjugation purposes makes my head hurt, so I’ve rewritten that bit and tried to simplify a little as well.

```suggestion
Per [TypeScript's module resolution rules](https://www.typescriptlang.org/docs/handbook/module-resolution.html), `.ts` and `.tsx` file extensions should not be used when importing TypeScript files. Instead, either use `.js`/`.jsx` file extensions or completely omit the file extension.
```

src/pages/en/guides/imports.md
```diff
@@ -39,15 +39,25 @@ JavaScript can be imported using normal ESM `import` & `export` syntax. This wor
...
+Astro includes built-in support for [TypeScript](https://www.typescriptlang.org/). You can import `.ts` and `.tsx` files directly in your Astro project, and even write TypeScript code directly inside your [Astro component script](/en/core-concepts/astro-components/#the-component-script) and any [hoisted script tags](/en/core-concepts/astro-components/#client-side-scripts).
 
-**Astro doesn't perform any type checking itself.** Type checking should be taken care of outside of Astro, either by your IDE or through a separate script. The [Astro VSCode Extension](/en/editor-setup/) automatically provides TypeScript hints and errors in your open files.
+**Astro doesn't perform any type checking itself.** Type checking should be taken care of outside of Astro, either by your IDE or through a separate script. For type checking Astro files, the [`astro check` command](/en/reference/cli-reference/#astro-check) is provided.
```
delucis, line 48: I see that this removed the reference to the VS Code Extension. Do we want to keep something about that here too? That it will provide in-editor feedback?

src/pages/en/guides/typescript.md
```diff
@@ -130,7 +130,7 @@ To see type errors in your editor, please make sure that you have the [Astro VS
 ```
 
 :::note
-`astro check` only checks types within `.astro` files, and `tsc --noEmit` only checks types within `.ts` and `.tsx` files.
+`astro check` only checks types within `.astro` files, and `tsc --noEmit` only checks types within `.ts` and `.tsx` files. To check types within Svelte and Vue files, the [`svelte-check`](https://www.npmjs.com/package/svelte-check) and the [`vue-tsc`](https://www.npmjs.com/package/vue-tsc) packages may be used respectively
```
delucis, line 133: Nit: period at the end of the sentence and slight word re-ordering to make “respectively” a little smoother.

```suggestion
`astro check` only checks types within `.astro` files, and `tsc --noEmit` only checks types within `.ts` and `.tsx` files. To check types within Svelte and Vue files, you can use the [`svelte-check`](https://www.npmjs.com/package/svelte-check) and the [`vue-tsc`](https://www.npmjs.com/package/vue-tsc) packages respectively.
```

src/pages/en/guides/troubleshooting.md
```diff
@@ -73,7 +73,7 @@ Then check your import statement:
 
 - Does your import have the same name as the imported component? (Check your component name and that it [follows the `.astro` syntax](/en/comparing-astro-vs-other-tools/#astro-vs-jsx).)
 
-- Have you included the extension in the import? (Check that your imported file contains an extension. e.g. `.astro`, `.md`, `.jsx`, `.vue`)
+- Have you included the extension in the import? (Check that your imported file contains an extension. e.g. `.astro`, `.md`, `.vue`, `.svelte`. Note: File extensions are not required for `.js(x)` and `.ts(x)` files only.)
```
delucis, line 76: Given the back and forth on this line (did you? you do/don’t need to!) maybe we just want to add a little extra highlight to the not here?

```suggestion
- Have you included the extension in the import? (Check that your imported file contains an extension. e.g. `.astro`, `.md`, `.vue`, `.svelte`. Note: File extensions are **not** required for `.js(x)` and `.ts(x)` files only.)
```

delucis (commented):
Thanks so much for getting this done @Princesseuh — I’ve made some suggestions but the actual content is 🔥

src/pages/en/guides/imports.md
delucis: Actually, just tried to write this and found I don’t like how it pairs with this paragraph, so let’s scrap my suggestion 👍

delucis (commented):
LGTM! (I can’t approve because it’s “my” PR, although it’s really yours now @Princesseuh 😂)

Princesseuh (approved):

PR #748, opened 2022-06-09
Practice what we preach: remove extensions from ts(x) imports 🧹

delucis:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

#747 updates our advice on how to import JS/TS/JSX/TSX files. We had a bit of a mixed bag, so this PR removes any remaining extensions from imports like that so we’re now complying with our own advice.


sarah11918:
LGTM!

PR #759, opened 2022-06-10
New individual deploy pages

sarah11918:
- [X ] New or updated content

Initial test of separation out of various Deployment instructions using the same tab component as we do with auto/manual install.

We might not choose to go in this same direction, but this will help in the sort term with organizing that page, keeping a consistent look until we have a grander vision. It emphasizes our main sponsors Netlify and Vercel, and lets us put emphasis on keeping those pages clean, updated and helpful.

(Note: Link checker going off the rails because of changing up some page routes. Will eventually fix.)

PR #761, opened 2022-06-11
Add Aside component

### LINKED ISSUE #401: Add a breakout/admonition/aside component
Currently we abuse the `<blockquote>` element to highlight bits of information, sometimes creatively combined with the `Badge` component. However these aren't usually actually quotes, so we shouldn't strictly speaking use `<blockquote>` to do this.

Many design systems include a component for this that usually supports different variations to highlight warnings, notes etc. Docusaurus calls them [“admonitions”](https://docusaurus.io/docs/markdown-features/admonitions) and the BBC's GEL system calls them [“breakout boxes”](https://bbc.github.io/gel/components/breakout-boxes/) for example. Docusaurus provides a Markdown shortcut to show these too, which we could mimic if we wanted.

Would something like this be useful? I built a component like this for another project based on the Astro docs template which could be a starting point: https://github.com/delucis/bgio-effects/blob/latest/docs/src/components/Aside.astro

delucis:

- [X] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

- Close #401

- Add an `<Aside>` Astro component to replace our abuse of the `<blockquote>` element

   This is an accessible component, based on the [recommended markup from the BBC’s GEL design system](https://bbc.github.io/gel/components/breakout-boxes/#recommended-markup).

   The component has note, tip, info, caution, danger and learn variants with colour, iconography, and default labelling distinct for each.

- Add an Astro integration/remark plugin that allows us to use a simple syntax to use the component in Markdown and also avoid needing to import it in the frontmatter all the time.

   ```md
   :::tip
   It’s best to avoid using `<blockquote>` for things that aren’t quotes.
   :::
   ```

   The syntax also supports custom titles for the asides:

   ```md
   :::caution[Deprecated]
   Using `<blockquote>` for notes is deprecated.
   :::
   ```

- Update all English pages (except for Deploy as that’s already being revised in #759) to use the new component wherever we were using a blockquote or similar styling currently. I used my judgement to decide which variant to use and where possible included custom titles.

   Here are the resulting usage stats:

   | variant | used |
   |---------|------|
   | caution | 21   |
   | note    | 21   |
   | tip     | 20   |
   | learn   | 3    |
   | info    | 2    |
   | danger  | 0    |

#### Questions

- How does this look? Are we happy this is a helpful way to help distinguish information on pages?

- What do we think about the variants? Check out how they look in the deploy preview. Do we want to get rid of “info” in favour of just “note” (neutral) or “tip” (positive, energetic)?

- What do we think about the “learn” variant? I found a couple of places where it could be used, but there are far more that use the 📚 books emoji as bullet points where this didn’t feel quite right. It might be better to get rid of this variant and think about an alternative component for linking to other learning material.

#### Screenshots

| light | dark |
|---|---|
| Components in light mode on a large screen | Components in dark mode on a large screen |

| light | dark |
|---|---|
| Components in light mode on a small screen | Components in dark mode on a small screen


package.json
```diff
@@ -35,17 +35,23 @@
     "eslint": "^8.16.0",
     "eslint-plugin-astro": "^0.8.1",
     "fast-glob": "^3.2.11",
+    "hastscript": "^7.0.2",
     "htmlparser2": "^7.2.0",
     "kleur": "^4.1.4",
+    "mdast": "^3.0.0",
```
hippotastic, line 41: Based on the deprecation message, this package has been renamed to `remark`. Can you find a way to make it work without the deprecated package and use remark instead?

src/components/Aside.astro
```diff
@@ -0,0 +1,126 @@
+---
+import UIString from './UIString.astro';
+export interface Props {
+	type?: 'note' | 'tip' | 'info' | 'caution' | 'danger' | 'learn';
+	title?: string;
+}
+
+const { type = 'note', title } = Astro.props as Props;
+const id = 'aside-' + (Math.random() * 10e16).toString(32);
```
hippotastic, line 9: I'm not particularly happy with the idea of having a random output in an Astro component. If it's only used for `aria-labelledby`, can you instead change that to `aria-label={title}` and thereby get rid of the need for an ID completely?

hippotastic (commented):
This is amazing, thank you so much @delucis! This is a perfect use case for directives and I love both your clean implementation and the way the result looks. Great job! 💯

I had a few minor comments as usual. :)

sarah11918:
Beautiful! Amazing work, Chris!

Just quickly, I might prefer also big able to see a left border, and not full width on mobile, too. (At first, I didn't realize that was intentional, and thought it was just another "Sarah phone" thing.) At full width, it might have the effect of breaking content into "sections" when they weren't meant to be quite so broken up?

But, that's a minor point, and maybe there's a good reason for it!

Will look more closely over the weekend!

package.json
delucis: I’m only using the types. Looked into it and `remark` uses `mdast`’s typings still… really odd. I can get around this by replacing `mdast` with `@types/mdast`. Odd, but works!

Yan-Thomas:
LGTM. Great work, @delucis!!

I just have a nitpick:

I felt there is too little margin between the icon and the accompanying text, mainly on the "Note" one, since its icon takes more space than the others. Something like 0.1~0.3rem margin between the icon and the text looks better to me, or maybe changing the icon to something that isn't a rectangle could work too? Docusaurus for example uses an circle-shaped exclamation point icon for "Note". I'm not sure if anyone else feels like this, though.

src/components/Aside.astro
delucis: I think the issue is that we then would need to _hide_ the actual text node with the title from screenreaders if we did this or the title would be repeated, so I’d prefer this pattern if possible. What’s your concern with randomising the IDs? That there’s a small chance of conflicts? Would a more robust ID library like `nanoid` reassure you?

hippotastic: No, my point is that I like to have stable build output in general.

I often do full directory comparisons of the build output to catch if an upgrade of some dependency introduced changes to the HTML output. This becomes harder with each piece of randomness added to the output.

I honestly wonder if there is any actual need for an aria label in the current case. The contained text should be perfectly accessible by default, don't you think?

delucis: I believe the labelling (which I [based on the BBC’s GEL markup](https://bbc.github.io/gel/components/breakout-boxes/#recommended-markup)) is motivated by this: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#labeling_section_content

A user might be offered to hear a list of sections and labelling gives the section a title.

How about I hash the title and contents of the Aside?

hippotastic: Hmm, I see. Hashing wouldn't work I'm afraid, because the exact same Aside content could be repeated multiple times on a page.

What about my idea of using `aria-label={title}` on the container and then just adding `aria-hidden="true"` to the title <p> to solve your concern of it being read out twice?

delucis: Seems to work well with VoiceOver on Mac — and the Accessibility Tree in Firefox dev tools also looks OK. Would love some more tests. @Yan-Thomas Want to try this out with your screenreader?

delucis:
> @sarah11918: Just quickly, I might prefer also big able to see a left border, and not full width on mobile, too. (At first, I didn't realize that was intentional, and thought it was just another "Sarah phone" thing.) At full width, it might have the effect of breaking content into "sections" when they weren't meant to be quite so broken up?

My thinking was to match how our code blocks behave currently, making use of the full width on narrow devices so we don’t squish the content too much. Happy to revisit this if you think it’s too disruptive in context.

> @Yan-Thomas: I felt there is too little margin between the icon and the accompanying text, mainly on the "Note" one, since its icon takes more space than the others.

Yes, I’m also wondering about this. Couldn’t quite make up my mind about the icon for “note”. If anyone thinks another one of these could be better, I’d love to try a variation: https://primer.style/octicons/

Otherwise we could definitely add a small margin. Potentially _just_ for the note icon.

src/pages/en/core-concepts/routing.md
```diff
@@ -20,13 +20,17 @@ src/pages/about/me.astro     -> mysite.com/about/me
...
+There is no separate "routing config" to maintain in an Astro project. Static pages are created by placing files in the `/src/pages/` directory.
+:::
 
 ## Dynamic routes
 
 A single Astro Page component can also specify dynamic route parameters in its filename to generate multiple routes that match a given criteria. You can create several related pages at once, such as author pages, or a page for each blog tag. Named parameters allow you to specify values for "named" levels of these route paths, and rest parameters allow for more flexible "catch-all" routes.
 
-> 💡 Even dynamically-created pages and routes are generated at build time.
+:::note
+Even dynamically-created pages and routes are generated at build time.
```
aFuzzyBear, line 32:
```suggestion
Dynamically-created pages and routes are generated at build time.
```

aFuzzyBear (approved):
⭐⭐⭐⭐⭐
This is a really nice and well implemented UX feature. Better descriptive boxes will help sign post information easier to the end users. 

Great work on this PR Chris. Really solid work dude


👊👩‍💻💜

Yan-Thomas:
> EDIT: Just pushed a change tweaking the view boxes of some of the SVG icons to try and improve how they space. Let me know if this is better!

It looks way better, thanks for your hard work!!

hippotastic:
Just a heads up: When merging this, all translated pages will become outdated. I think that's okay, I only wanted to mention it so we know what's happening.

src/pages/en/core-concepts/routing.md
delucis: LOL. Out of scope Fuzzy! 😂

src/components/Aside.astro
Yan-Thomas: I can test this out tomorrow (I'm on my sister's laptop so I don't have my screen reader here, unfortunately). Sorry for the late response, this mention didn't notify me for some reason?! - I'm not even sure if you still need to test this out :sweat_smile:

delucis: Yes, if you could, a test would be helpful! I could only test on macOS VoiceOver on my end and the more tests the better 💜

Yan-Thomas (approved):
Tested with NVDA, the output was "Tip, complementary landmark, (...)". LGTM.

sarah11918:
Just so there's a record in this PR:  @delucis & I had a long voice chat about this PR:

1. Some changes we decided to make:
- Our "tip" icon has changed from a light bulb to.... a rocket! 🚀  We like the playfulness, and tie-in with Astro. Tips are either leading to other sites, or things you *can* do (but don't *have* to do) to super-charge your project. We feel like the rocket works in both cases, and is totally on-brand.
- Sarah got fussy and wanted the mobile view to NOT take up the full page, but rather look like the desktop view with the solid left border, and defined width. This change ALSO helps us make sure we're not over-using these Asides, and with the additional padding and margin, we control the inner content and never let the Aside get too tall because we're just babbling too much inside these Asides.
3. Sarah's tasks:
- Check **every, single Aside** (since Chris dutifully changed every single aside!) to make sure none are visually off-putting (out of place in the section, styling (e.g. lists?), padding... especially on mobile. And if so, make a note what to change: the component, OR is this a sign of flaws in content or structure of the page itself?
- Check each Aside is the "proper" category
- In particular: what's the deal with "Learn More" - the few of them (3) that are NOT regular, external links with the stack of books: should these be a different category entirely? Should these instead be the regular book icon + link (which could eventually be its own thing)

delucis:
Thanks for the thorough notes @sarah11918 💜 

Just pushed the discussed changes so should be ready for your fine-tooth comb!

sarah11918:
OK, my suggestions:

- No Learn More Aside. Turn them all into stack of books + link to match the others. We will figure out what we want to do for that specific pattern separately. The Aside is too "distracty" for just a simple read more.

-No Info Aside. We can make do with Note and Tip. of the pro that are there:

1. TypeScript: I think this is equivalent to our Stack of books + link pattern.
2. Migrate to v24: Let's make it a Note.


- Other specific Asides

Troubleshooting: Not sure that this is your problem? -- Not sure how I feel about this one yet, but now-blocking

There is one place a warning immediately follows a tip. (Just before Intellisense header.) Fine for now, non-blocking, but I'll examine content there in case that's a content smell.

I have a list of Asides that look a little long/overwhelming on mobile. Non-blocking for this PR, but I will tackle content of those going foreward.

I love that this Pr is helping with content, too! 💜

Last thought: if we are down to purple notes and "fiery" cautions/danger/warnings, is green the best color for our tips? I only ask because it's not really a theme color. Would some form of orange or pink be playful yet on brand? Or, an 80s arcade lighter blue/teal? Opinions?

src/pages/en/guides/integrations-guide.md
```diff
@@ -63,7 +60,9 @@ If you see an error on startup, make sure that you:
 
 Astro recently launched an **experimental** `astro add` command to automate the setup of integrations.
 
-> We will always ask for confirmation before updating any of your files, but it never hurts to have a version-controlled backup just in case.
+:::note
```
sarah11918, line 63:
```suggestion
:::warning
```
What do you think about making this a warning?

src/pages/en/core-concepts/framework-components.md
```diff
@@ -129,13 +135,17 @@ import MySvelteButton from '../components/MySvelteButton.svelte';
...
 
 Each framework has its own patterns for nesting: `children` props and [render props](https://reactjs.org/docs/render-props.html) for React and Solid; `<slot />` with or without names for Svelte and Vue, for example.
 
-Note: Astro components are always rendered to static HTML, even when they include framework components that are hydrated. This means that you can only pass props that don't do any HTML rendering. Passing React's "render props" or named slots to framework components from an Astro component will not work, because Astro components can’t provide the client runtime behavior that those patterns require.
+:::note
+Astro components are always rendered to static HTML, even when they include framework components that are hydrated. This means that you can only pass props that don't do any HTML rendering. Passing React's "render props" or named slots to framework components from an Astro component will not work, because Astro components can’t provide the client runtime behavior that those patterns require.
+:::
```
sarah11918, line 148:
```suggestion
```

sarah11918:
All the official "Aside" stuff is great! I like the teal! LGTM!

For the sake of the poor translators, I'll go through and tackle now some of the problematic aside *content* that is too long etc. 

So, be prepared for comments/suggestions... but at this point I believe it will all be content stuff, as I try to make sure these call-out asides are appropriately-focused. We're now in content-editing mode on this PR, as far as I'm concerned! 🥳

src/pages/en/guides/markdown-content.md
```diff
@@ -308,7 +314,9 @@ import {Content as PromoBanner} from '../components/promoBanner.md';
 
 ## Markdown Component
 
-> NOTE: The `<Markdown />` component does not work in SSR and may be removed before v1.0. It should be avoided if possible. To use Markdown in your templates, use a separate `.md` file and then [`import` Markdown](/en/guides/markdown-content/#importing-markdown) into your template as a component.
+:::caution[Deprecated]
+The `<Markdown />` component does not work in SSR and may be removed before v1.0. It should be avoided if possible. To use Markdown in your templates, use a separate `.md` file and then [`import` Markdown](/en/guides/markdown-content/#importing-markdown) into your template as a component.
```
sarah11918, line 318:
```suggestion
The `<Markdown />` component does not work in SSR and will be moved to its own package before v1.0. Consider [importing Markdown content](/en/guides/markdown-content/#importing-markdown) instead.
```

src/pages/en/guides/markdown-content.md
```diff
@@ -345,7 +353,9 @@ const expressions = 'Lorem ipsum';
 
 ### Remote Markdown
 
-> NOTE: The `<Markdown />` component does not work in SSR and may be removed before v1.0. It should be avoided if possible. To use Markdown in your templates, use a separate `.md` file and then `import` it into your template as a component. Read this [RFC Discussion](https://github.com/withastro/rfcs/discussions/179) to learn more.
+:::caution[Deprecated]
+The `<Markdown />` component does not work in SSR and may be removed before v1.0. It should be avoided if possible. To use Markdown in your templates, use a separate `.md` file and then `import` it into your template as a component. Read this [RFC Discussion](https://github.com/withastro/rfcs/discussions/179) to learn more.
```
sarah11918, line 357:
```suggestion
The `<Markdown />` component does not work in SSR and will be moved to its own package before v1.0. Consider [importing Markdown content](/en/guides/markdown-content/#importing-markdown) instead.
```

src/pages/en/reference/directives-reference.md
```diff
@@ -208,7 +212,9 @@ const message = "Astro is awesome!";
 </script>
 ```
 
->⚠️ Using `define:vars` on a `<script>` or `<style>` tag implies the `is:inline` directive, which means your scripts or styles won't be bundled and will be inlined directly into the HTML. See the [dedicated section](#isinline) on `is:inline` for more details.
+:::caution
+Using `define:vars` on a `<script>` or `<style>` tag implies the `is:inline` directive, which means your scripts or styles won't be bundled and will be inlined directly into the HTML. See the [dedicated section](#isinline) on `is:inline` for more details.
```
sarah11918, line 216:
```suggestion
Using `define:vars` on a `<script>` or `<style>` tag implies the [`is:inline` directive](#isinline), which means your scripts or styles won't be bundled and will be inlined directly into the HTML.
```

src/pages/en/reference/api-reference.md
```diff
@@ -466,7 +468,9 @@ Astro includes several built-in components for you to use in your projects. All
 
 ### `<Markdown />`
 
-> NOTE: The `<Markdown />` component does not work in SSR and may be removed before v1.0. It should should be avoided if possible. To use Markdown in your templates, use a seperate `.md` file and then [`import` Markdown](/en/guides/markdown-content/#importing-markdown) into your template as a component.
+:::caution[Deprecated]
+The `<Markdown />` component does not work in SSR and may be removed before v1.0. It should should be avoided if possible. To use Markdown in your templates, use a seperate `.md` file and then [`import` Markdown](/en/guides/markdown-content/#importing-markdown) into your template as a component.
```
sarah11918, line 472:
```suggestion
The `<Markdown />` component does not work in SSR and will be moved to its own package before v1.0. Consider [importing Markdown content](/en/guides/markdown-content/#importing-markdown) instead.
```

src/pages/en/guides/integrations-guide.md
```diff
@@ -13,22 +11,21 @@ i18nReady: true
...
 In this example, we will add the `@astrojs/react` integration to add React support to your Astro project. The process for adding any other framework (Preact, Vue, Svelte or Solid.js) is almost identical and can be followed using the same steps outlined below.
 
-<blockquote>
-  <Badge variant="accent">Feeling adventurous?</Badge>
-
-  Astro recently launched an **experimental** `astro add` command to automate this process! Instead of the steps below, you can run `npx astro add react`. That's it!
-
-  Skip down to [Automatic Integration Setup](/en/guides/integrations-guide/#automatic-integration-setup) for more details.
+:::tip[Feeling adventurous?]
+Astro recently launched an **experimental** `astro add` command to automate this process! Instead of the steps below, you can run `npx astro add react`. That's it!
```
sarah11918, line 25:
```suggestion
Astro provides an `astro add` command to automate this process for official Astro integrations! Instead of the steps below, you can run `npx astro add react`. That's it!
```

src/pages/en/guides/integrations-guide.md
```diff
@@ -13,22 +11,21 @@ i18nReady: true
 - Add new features to your project, like automatic sitemap generation.
 - Write custom code that hooks into the build process, dev server, and more.
 
-> Integrations are still new, and the API has not yet been finalized. Only official Astro integrations (those published to `@astrojs/` on npm) are currently supported to protect users from breaking changes.
->
-> **To enable 3rd-party integrations:** Run Astro with the `--experimental-integrations` CLI flag, or include `experimental: { integrations: true }` in your Astro config file. 
+:::note[Experimental status]
+Integrations are still new, and the API has not yet been finalized. Only official Astro integrations (those published to `@astrojs/` on npm) are currently supported to protect users from breaking changes.
```
sarah11918, line 15:
```suggestion
Only official Astro integrations (those published to `@astrojs/` on npm) are currently supported to protect users from breaking changes.
```

src/pages/en/guides/integrations-guide.md
```diff
@@ -13,22 +11,21 @@ i18nReady: true
...
 
 In this example, we will add the `@astrojs/react` integration to add React support to your Astro project. The process for adding any other framework (Preact, Vue, Svelte or Solid.js) is almost identical and can be followed using the same steps outlined below.
 
-<blockquote>
-  <Badge variant="accent">Feeling adventurous?</Badge>
-
-  Astro recently launched an **experimental** `astro add` command to automate this process! Instead of the steps below, you can run `npx astro add react`. That's it!
-
-  Skip down to [Automatic Integration Setup](/en/guides/integrations-guide/#automatic-integration-setup) for more details.
+:::tip[Feeling adventurous?]
```
sarah11918, line 24:
```suggestion
:::tip[Quick Method]
```

src/pages/en/guides/data-fetching.md
```diff
@@ -71,9 +71,13 @@ const weather = json.data
...
 <p>Weather: {weather.getCityByName.weather.summary.description}</p>
 ```
-> 💡 Remember, all data in Astro components is fetched when a component is rendered.
+
+:::note
+Remember, all data in Astro components is fetched when a component is rendered.
 
 Your deployed Astro site will fetch data **once, at build time**. In dev, you will see data fetches on component refreshes. If you need to re-fetch data multiple times client-side, use a [framework component](/en/core-concepts/framework-components/) or a [client-side script](/en/core-concepts/astro-components/#client-side-scripts) in an Astro component.
+:::
+
```
sarah11918, line 80:
```suggestion
:::note
Remember, all data in Astro components is fetched when a component is rendered.
:::

Your deployed Astro site will fetch data **once, at build time**. In dev, you will see data fetches on component refreshes. 

If you need to re-fetch data multiple times client-side, use a [framework component](/en/core-concepts/framework-components/) or a [client-side script](/en/core-concepts/astro-components/#client-side-scripts) in an Astro component.

```

src/pages/en/guides/styling.md
```diff
@@ -277,8 +283,9 @@ For advanced use cases, you can import a direct URL reference for a CSS file ins
 
 This is not recommended for most users. Instead, place your CSS files inside of `public/` to get a consistent URL reference.
 
-> ⚠️WARNING⚠️:
-> Importing a smaller CSS file with `?url` may return the base64 encoded contents of the CSS file as a data URL, but only in your final build. You should either write your code to support encoded data URLs (`data:text/css;base64,...`) or set the [`vite.build.assetsInlineLimit`](https://vitejs.dev/config/#build-assetsinlinelimit) config option to `0`  to disable this feature.
+:::caution
+Importing a smaller CSS file with `?url` may return the base64 encoded contents of the CSS file as a data URL, but only in your final build. You should either write your code to support encoded data URLs (`data:text/css;base64,...`) or set the [`vite.build.assetsInlineLimit`](https://vitejs.dev/config/#build-assetsinlinelimit) config option to `0`  to disable this feature.
```
sarah11918, line 287:
```suggestion
Write your code to support encoded data URLs (`data:text/css;base64,...`) or set the [`vite.build.assetsInlineLimit`](https://vitejs.dev/config/#build-assetsinlinelimit) config option to `0` in order to avoid returning the base64 encoded contents of the CSS file as a data URL in your final build.
```

src/pages/en/core-concepts/framework-components.md
```diff
@@ -129,13 +135,17 @@ import MySvelteButton from '../components/MySvelteButton.svelte';
...
 
 Each framework has its own patterns for nesting: `children` props and [render props](https://reactjs.org/docs/render-props.html) for React and Solid; `<slot />` with or without names for Svelte and Vue, for example.
 
-Note: Astro components are always rendered to static HTML, even when they include framework components that are hydrated. This means that you can only pass props that don't do any HTML rendering. Passing React's "render props" or named slots to framework components from an Astro component will not work, because Astro components can’t provide the client runtime behavior that those patterns require.
+:::note
+Astro components are always rendered to static HTML, even when they include framework components that are hydrated. This means that you can only pass props that don't do any HTML rendering. Passing React's "render props" or named slots to framework components from an Astro component will not work, because Astro components can’t provide the client runtime behavior that those patterns require.
```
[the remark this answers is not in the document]
sarah11918: OK, then I'm fine leaving it all in.

hippotastic:
Love the new pages! However, in dark mode, the boxes are a bit too distracting in my opinion.

What about adding a new `--theme-aside-bg-opacity` var and setting it to `0.2` instead of the `0.4` we currently use in the dark theme? The light theme can remain unchanged as the boxes are a lot less distracting there.

src/pages/en/guides/integrations-guide.md
delucis: One small tweak here just to clarify that it’s a default not zero support. (Next paragraph does imply this but could be confusing reading these side by side.)

```suggestion
Only official Astro integrations (those published to `@astrojs/` on npm) are currently supported by default to protect users from breaking changes.
```

src/pages/en/guides/data-fetching.md
delucis: You guess right 😅

delucis: I’ll leave it as is for now. If you really dislike it, we can come back and rework this into some kind of section with this information in.

src/pages/en/guides/styling.md
delucis: I find this less clear than the current text. Current version is “small css can cause base64 URL, so…” whereas this change would leave a vague threat of base64 causing a problem. I’d like to keep this closer to the current version if you don’t mind? Here’s a slightly slimmer version keeping the current structure:

```suggestion
Importing a smaller CSS file with `?url` may return the base64 encoded contents of the CSS file as a data URL in your final build. Either write your code to support encoded data URLs (`data:text/css;base64,...`) or set the [`vite.build.assetsInlineLimit`](https://vitejs.dev/config/#build-assetsinlinelimit) config option to `0`  to disable this feature.
```

src/pages/en/guides/markdown-content.md
```diff
@@ -362,7 +372,9 @@ const content = await fetch('https://raw.githubusercontent.com/withastro/docs/ma
 
 ### Nested Markdown
 
-> NOTE: The `<Markdown />` component does not work in SSR and may be removed before v1.0. It should be avoided if possible. To use Markdown in your templates, use a separate `.md` file and then `import` it into your template as a component. Read this [RFC Discussion](https://github.com/withastro/rfcs/discussions/179) to learn more.
+:::caution[Deprecated]
+The `<Markdown />` component does not work in SSR and may be removed before v1.0. It should be avoided if possible. To use Markdown in your templates, use a separate `.md` file and then `import` it into your template as a component. Read this [RFC Discussion](https://github.com/withastro/rfcs/discussions/179) to learn more.
```
delucis, line 376: I’m assuming this should match the same suggestions made above

```suggestion
The `<Markdown />` component does not work in SSR and will be moved to its own package before v1.0. Consider [importing Markdown content](/en/guides/markdown-content/#importing-markdown) instead.
```

src/pages/en/reference/api-reference.md
```diff
@@ -506,7 +510,9 @@ import { Prism } from '@astrojs/prism';
 <Prism lang="js" code={`const foo = 'bar';`} />
 ```
 
-> **`@astrojs/prism`** is built-in as part of the `astro` package. No need to install as a separate dependency just yet! However, note that we do plan to extract `@astrojs/prism` to a separate, installable package in the future.
+:::note
+**`@astrojs/prism`** is built-in as part of the `astro` package. No need to install as a separate dependency just yet! However, note that we do plan to extract `@astrojs/prism` to a separate, installable package in the future.
```
delucis, line 514: :::caution[Deprecated]
**`@astrojs/prism`** will be extracted to a separate, installable package in the future.

delucis:
Thanks for trying that out @hippotastic! I did try something similar and didn’t like how the note variant then had a different purple background from other purple elements on the page, but we could in theory override the opacity back to `0.4` just for that variant.

I’ll wait for @sarah11918 to give her perspective because when we chatted, the feeling was that we liked how these popped. (I’d also like to get this merged so I can use it in the next PR, but that’s _my_ problem 😅_

sarah11918:
I am going to go with, and defer to @delucis ! We had the "distracting" conversation live, and we just kept coming back to... "it fits." I think Hippo's solution has *more* pop than the subtler option we considered, but I remain Team OG as long as Chris does!

(with the caveat of course that none of these decisions need ever be permanent)

Your comment re: several in proximity is valid, and I made some changes in this Pr to mitigate that, and will probably revisit in future.

I'm sticking with being on brand, and trust Chris to go with his informed gut!

src/pages/en/guides/styling.md
sarah11918: Aww... no vague threat? OK!

PR #777, opened 2022-06-16
Update: Markdown `<Content/>` Description 

aFuzzyBear:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=vH5D7px)

- [x] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

- Updated `Markdown/Imported Components#Content` with addition code content to demonstrate how to use the `<Content/>` component when used with `Astro.glob` 

**Before:**
image

**After**
image


src/pages/en/guides/markdown-content.md
```diff
@@ -312,6 +312,29 @@ import {Content as PromoBanner} from '../components/promoBanner.md';
 <PromoBanner />
 ```
 
+When used with `Astro.glob()`, you can simply pass the `<Content/>` component from the _Exported Properties_ of each returned markdown file. This will then render out the Markdown file and its contents in its entirety into your template.
```
delucis, line 315: Tried to rework this a bit given the code example is quite specific. I also removed “simply” because that can be quite subjective and potentially frustrating for someone if they’re finding the process anything but simple 😅 

```suggestion
When using `getStaticPaths` and `Astro.glob()` to generate pages from Markdown files, you can pass the `<Content/>` component through the page’s `props`. You can then retrieve the component from `Astro.props` and render it in your template. 
```

src/pages/en/guides/markdown-content.md
```diff
@@ -301,7 +301,7 @@ const posts = await Astro.glob('./posts/**/*.md');
 
 #### `Content`
 
-A component that renders the contents of the Markdown file. Here is an example:
+A component that returns the full rendered contents of the Markdown file. Here is an example:
```
delucis, line 304: Nit, but would it be better to stick to “renders” rather than “returns” here? Like, you don’t do `const contents = Component()` so I’m not sure “returning” is quite the right concept here (even if that is what technically happens under the hood).

```suggestion
A component that renders the full contents of the Markdown file. Here is an example:
```

src/pages/en/guides/markdown-content.md
```diff
@@ -312,6 +312,29 @@ import {Content as PromoBanner} from '../components/promoBanner.md';
...
+    params: { 
+      slug: post.frontmatter.slug 
+    },
+    props: {
+      post
+    },
+  }))
+}
+
+const { post:{ Content } } = Astro.props
```
delucis, line 332: Another nit, but I’m not a huge fan of multiple destructuring. Do you think this might be easier to read/understand?

```suggestion
const { Content } = Astro.props.post
```

delucis (commented):
Nice work @aFuzzyBear! What do you think of these comments?

src/pages/en/guides/markdown-content.md
aFuzzyBear: "To return or not to return, it really depends on how you throw...." - Fuzzy

PR #778, opened 2022-06-16
Update "Fragments & Multiple Elements" examples

### LINKED ISSUE #369: Better explanation for "Fragments & Multiple Elements"
I am reading the docs and I stumbled here:

image

The first section(Dynamic HTML) is easy to understand.

The second section though, it's a bit confusing. Compared to the first one, it has multiple elements.
So at first, I thought that when "looping", if I have to display HTML with a single root element I write it as is. However, if I have to write multiple root elements I must wrap them inside a Fragment.

**But is true??**

In `index.astro` I pasted the code from the Docs and removed the Fragments:

```html
<ul>
  {items.map((item) => (
    <li>Red {item}</li>
    <li>Blue {item}</li>
    <li>Green {item}</li>
  ))}
</ul>
```

And everything seems to be working fine without any issue.

So either the docs should get updated or it's a bug (but since it works why does it have to be a bug?)

sarah11918:
- [X ] New or updated content

Fixes #369 

This updates our Muliple Elements example to avoid telling people they MUST use a wrapping element when they don't have to.

Further adds an example of using Fragment to show when it might be useful to do so, since they don't appear to be necessary.


src/pages/en/core-concepts/astro-components.md
```diff
@@ -154,25 +154,26 @@ const items = ["Dog", "Cat", "Platypus"];
...
-      <li>Green {item}</li>
-    </>
+    <li>Red {item}</li>
+    <li>Blue {item}</li>
+    <li>Green {item}</li>
   ))}
 </ul>
 ```
 
+However, Astro also supports using either `<Fragment> </Fragment>` or `<> </>`. This may be useful to add [`client:*` directives](/en/reference/client-directives/) and avoid wrapper elements, as in the following example:
```
sarah11918, line 172:
```suggestion
However, Astro also supports using either `<Fragment> </Fragment>` or `<> </>`. This may be useful to add [`client:*` directives](/en/reference/directives-reference/) and avoid wrapper elements, as in the following example:
```

delucis (commented):
I thought we decided to keep the fragment wrapping example? The reason being that without the `<></>` this code breaks if you leave out the wrapping `()` which is less easy for people to spot as important when following this example.

We can probably relax the phrasing to be less insistent (“must” etc.) though.

src/pages/en/core-concepts/astro-components.md
```diff
@@ -154,25 +154,26 @@ const items = ["Dog", "Cat", "Platypus"];
...
 <ul>
   {items.map((item) => (
-    <>
-      <li>Red {item}</li>
-      <li>Blue {item}</li>
-      <li>Green {item}</li>
-    </>
+    <li>Red {item}</li>
+    <li>Blue {item}</li>
+    <li>Green {item}</li>
```
delucis, line 167: Sounds like we agreed to keep this code sample with the Fragment wrapper

```suggestion
    <>
      <li>Red {item}</li>
      <li>Blue {item}</li>
      <li>Green {item}</li>
    </>
```

src/pages/en/core-concepts/astro-components.md
```diff
@@ -154,25 +154,26 @@ const items = ["Dog", "Cat", "Platypus"];
 
 #### Fragments & Multiple Elements
 
-Remember: an Astro component template can render multiple elements with no need to wrap everything in a single `<div>` or `<>`.
-
-However, when using an Astro JSX-like expression to dynamically create elements, you must wrap these multiple elements inside of a **Fragment** just like you would in JavaScript or JSX. Astro supports using either `<Fragment> </Fragment>` or `<> </>`.
+An Astro component template can render multiple elements with no need to wrap everything in a single `<div>` or `<>`, unlike JavaScript or JSX.
```
delucis, line 157: May be overkill but what do you think of an extra example here?

````suggestion
An Astro component template can render multiple elements with no need to wrap everything in a single `<div>` or `<>`, unlike JavaScript or JSX.

```astro
---
// Template with multiple elements
---
<p>No need to wrap elements in a single containing element.</p>
<p>Astro supports multiple root elements in a template.</p>
```

However, when using an expression to dynamically create multiple elements, you should wrap these elements inside a **Fragment** as you would in JavaScript or JSX. Astro supports using either `<Fragment> </Fragment>` or the shorthand `<> </>`.
````

File: src/pages/en/core-concepts/astro-components.md [hunk 4]
```diff
@@ -154,25 +154,26 @@ const items = ["Dog", "Cat", "Platypus"];
...
-      <li>Green {item}</li>
-    </>
+    <li>Red {item}</li>
+    <li>Blue {item}</li>
+    <li>Green {item}</li>
   ))}
 </ul>
 ```
 
+However, Astro also supports using either `<Fragment> </Fragment>` or `<> </>`. This may be useful to add [`client:*` directives](/en/reference/directives-reference/) and avoid wrapper elements, as in the following example:
```
delucis, line 172: I like this example! Given the updates agreed above we could maybe update the text like this?

````suggestion
Fragments can also be useful to avoid wrapper elements when adding [`client:*` directives](/en/reference/directives-reference/), as in the following example:
````

File: src/pages/en/core-concepts/astro-components.md [hunk 5]
```diff
@@ -154,25 +154,26 @@ const items = ["Dog", "Cat", "Platypus"];
...
+
+```astro
+<Fragment set:html={htmlString} />
```
delucis, line 175: Not sure if we need this clarifying context here or only in the directives reference, but this might help explain what is going on.

```suggestion
---
const htmlString = '<p>Raw HTML content</p>';
---
<Fragment set:html={htmlString} />
```

delucis (commented):
Took the liberty of trying to write the changes as suggestions. Please feel free to cut/edit/ignore as appropriate 🙌

src/pages/en/core-concepts/astro-components.md
sarah11918: (Probably going to amend with a lower-case fragment in the text, as I don't think it would normally be capitalized when not the component itself?)

src/pages/en/core-concepts/astro-components.md
```diff
@@ -173,6 +181,14 @@ const items = ["Dog", "Cat", "Platypus"];
 </ul>
 ```
 
+Fragments can also be useful to avoid wrapper elements when adding [`client:*` directives](/en/reference/directives-reference/), as in the following example:
```
delucis, line 184: Looks good now, but hadn’t noticed this before. I don’t think you can ever add a _client_ directive to a fragment because they’re not hydrated. I think only `set:html` or `set:text` make sense with fragments so we can probably do this:

```suggestion
Fragments can also be useful to avoid wrapper elements when adding [`set:*` directives](/en/reference/directives-reference/#sethtml), as in the following example:
```

PR #786, opened 2022-06-17
♻️ clarify that a string or number can be used

shaunchander:
It's a bit confusing that `id: 3` is a number whereas the other two are strings. When I first read this I thought it was a bug until I read above where it mentioned that strings and numbers can be used as parameters. I think it'd be nice to reflect that in the code block as well just to make it concrete.

- [x] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

### DIFF
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -274,7 +274,7 @@ export async function getStaticPaths() {
   return [
     { params: { id: '1' } },
     { params: { id: '2' } },
-    { params: { id:  3  }  }  // Can be a number too!
+    { params: { id:  3  } }  // Can be a number too!
   ];
 }
 
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -274,7 +274,7 @@ export async function getStaticPaths() {
   return [
     { params: { id: '1' } },
     { params: { id: '2' } },
-    { params: { id:  3 // Can be a number too! Only strings and numbers are supported values. } }
+    { params: { id:  3  }  }  // Can be a number too! Only strings and numbers are supported values.
   ];
 }
 
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -274,7 +274,7 @@ export async function getStaticPaths() {
   return [
     { params: { id: '1' } },
     { params: { id: '2' } },
-    { params: { id:  3  }  }  // Can be a number too! Only strings and numbers are supported values.
+    { params: { id:  3  }  }  // Can be a number too!
   ];
 }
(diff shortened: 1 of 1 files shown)


src/pages/en/reference/api-reference.md
```diff
@@ -274,7 +274,7 @@ export async function getStaticPaths() {
   return [
     { params: { id: '1' } },
     { params: { id: '2' } },
-    { params: { id:  3 } }
+    { params: { id:  3 // Can be a number too! Only strings and numbers are supported values. } }
```
sarah11918, line 277:
```suggestion
    { params: { id:  3  }  }  // Can be a number too! Only strings and numbers are supported values.
```
I'm going to ask @delucis to confirm, but I think we would typically only comment after the complete line of code, not within it. That would be my instinct.

Also, I could go either way re: repeating "Only strings and numbers are supported" in the code comment. I do like that it's clear! (Especially for those who only skim and read the code samples. 😅 ) 

I suspect that just "Can be a number too!" would also be fine, given we have just said that same thing above. I think the key thing to point out is that no, we didn't make a typo! We meant for that to be a number.  ;)

sarah11918:
Thanks for this, @shaunchander ! I agree, it's maybe not clear that we *meant* for that to be a number, so I think it will be great to call this out!

We just want to get the wording and syntax right, so I've made some nit-picky comments above for us to resolve first, but I'm looking forward to having this improvement in the docs very soon!  🥳

src/pages/en/reference/api-reference.md
```diff
@@ -274,7 +274,7 @@ export async function getStaticPaths() {
   return [
     { params: { id: '1' } },
     { params: { id: '2' } },
-    { params: { id:  3 } }
+    { params: { id:  3  }  }  // Can be a number too! Only strings and numbers are supported values.
```
sarah11918, line 277:
```suggestion
    { params: { id:  3  }  }  // Can be a number too!
```
OK, I'm going to make an executive decision and say that after testing at a few different screen sizes, I dislike the longer comment making an otherwise narrow code sample need to be scrollable!

So, I suggest we just point out that it can be a number. And, @shaunchander if you can also "update branch" to sync up with main, then I'm happy to merge!

src/pages/en/reference/api-reference.md
```diff
@@ -274,7 +274,7 @@ export async function getStaticPaths() {
   return [
     { params: { id: '1' } },
     { params: { id: '2' } },
-    { params: { id:  3 } }
+    { params: { id:  3  }  }  // Can be a number too!
```
sarah11918, line 277:
```suggestion
    { params: { id:  3  } }  // Can be a number too!
```
Extra space?

PR #787, opened 2022-06-17
♻️ reword the dynamic JSX section

shaunchander:
I found some of the headings in this section to be oddly worded, "dynamic" is used a lot here but it's also a bit misleading (for example I thought "Dynamic Attributes" at first meant actual dynamic HTML attributes like `<marklar dynamic-attribute-* />`. People coming from React or have used JSX know it's dynamic in nature so I think we can simplify the headings here a bit.

I also made minor changes to the sentence structure and wording of the descriptions for each heading.

#### What kind of changes does this PR include?

- [ ] Minor content fixes (broken links, typos, etc.)
- [x] New or updated content
- [ ] Translated content
- [ ] Changes to the docs site code
- [ ] Something else!

#### Description

- What does this PR change? Give us a brief description.
- Did you change something visual? A before/after screenshot can be helpful.

PR #788, opened 2022-06-17
⚡️ add class:list HTML directive example to the guide

shaunchander:
The HTML directives (`class:list`, `set:html`, and `set:text`) are super useful when working with Astro components.

Previously they were only accessible in Template Directives page which is located in the reference section (and thus somewhat hard to access/know about). Given the importance of these directives (and the extreme usefulness of `class:list`), I think it's best if we have a dedicated section in the guide to introduce newcomers to these useful Astro features!

- [x] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description


delucis:
Hi @shaunchander — thanks for this!

I noticed it also includes the same changes you added in #787. Would you mind rebasing to separate those out?

sarah11918:
Hi @shaunchander ! Just following up on @delucis ' comment from a few days ago: you have two open PRs and this one seems to include the changes from the other one as well. Are you able to separate to include only the new proposed content here in this PR so we can handle each one separately? Thanks!

shaunchander:
> Hi @shaunchander ! Just following up on @delucis ' comment from a few days ago: you have two open PRs and this one seems to include the changes from the other one as well. Are you able to separate to include only the new proposed content here in this PR so we can handle each one separately? Thanks!

Oh yeah, sorry it completely slipped my mind!

shaunchander:
@sarah11918 @delucis fixed!

sarah11918:
Hi @shaunchander ! 

I agree that these directives are useful, but it can be diffiicult to maintain the same examples and wording in two separate locations. We prefer to have as little duplicated content so that when things need to be updated, we are not needing to change in multiple places.

So I've been thinking about what to do with this suggestion, and I do agree these directives could be better surfaced on this page. And, I think you've identified a great spot for them, in the section about props.

I think what would fit best here is a mention, with a link to the reference page, about the "handy/helpful HTML attributes available in Astro components, such as `class:list={...}` etc." without dupliacating the explanation and examples that are on the reference page.

Would you like to propose something to function in that way?

shaunchander:
@sarah11918 I think that's fair, though I do also think the directives should be shown in full just for sake of example. We could make a single section titled "Directives" and say that "You can use a special directive syntax to do XYZ, the directive syntax looks `like:this`. An example would be `class:list` which does ABC or `set:html`, which does EFG".

We can then link to the reference section to show all other directives. 

This way we not only introduce directives in a friendly manner but we also guide users to check out the reference to see full implementation details of said directives.

sarah11918:
@shaunchander  So here are some follow up thoughts about this:

- We do already have at https://docs.astro.build/en/core-concepts/astro-components/#the-component-template an overall description (including directives) of what you can include in your Astro templatate, and each of these is represented in the code sample there. There is an example of `set:html` in that code example. There's also another example of using `set:html` in the Fragment section below.
- Another place where we do address the directives on a page, but leave the heavy lifting for the directives page, is: https://docs.astro.build/en/core-concepts/framework-components/#available-hydration-directives So, this is the precedent we have established for that. It's not an entirely analogous situation because these client: directives are pretty essential to using a UI framework component (ie much of the time you do want SOME kind of hydration, and if you don't know about these, your component won't even work). But even on this page, we really don't go into details about the options. (We do, in the section above, explain that there are these different hydration methods because they are so key to the component functioning. )

So my initial reaction is that we've probably given these *adequate* exposure here, although maybe given that we see `set:html` in a real use-case in the 2nd example, and it's the only one shown on this page, maybe the first one could be a `class:list` one-liner? What about PR'ing a change to update the example there? Then that one is seen on this page, too, and yet there are two small-ish examples that can alert people to the fact that these exist, and they can find more info on the specific reference page?

shaunchander:
> 

Sure sounds good, I think, above all, `class:list` should be highlighted as it is quite literally a hidden gem in the docs.

shaunchander:
@sarah11918 I created a new header for `class:list` right under "Dynamic HMTL" as well as back-linked to the reference documentation for more details on the directive.

sarah11918:
Hi @shaunchander ,

I was thinking more like around line 106 where it says this:

```
<!-- Use a template directive to inject an unescaped HTML string into an element: -->
<p set:html={rawHTMLString} />
```
We just updated THAT to show an element with `class:list` instead of `set:html` (with an appropriate comment). Again, the point here is just to show that it exists.

The section for Dynamic Attributes just a little further down from that DOES show passing a class name with a variable, so we don't need an entire new heading for that here. The above kind of example is all we're intending to include on this page. 

There's a somewhat related issue #995 that might lead to some more content being added to the Styling Guide page, and that may lead to another mention of `class:list` as well. So I think just changing lines 106 and 107 here on this page would be most helpful!

shaunchander:
> Hi @shaunchander ,
> 
> I was thinking more like around line 106 where it says this:
> 
> ```
> 
> <p set:html={rawHTMLString} />
> ```
> 
> We just updated THAT to show an element with `class:list` instead of `set:html` (with an appropriate comment). Again, the point here is just to show that it exists.
> 
> The section for Dynamic Attributes just a little further down from that DOES show passing a class name with a variable, so we don't need an entire new heading for that here. The above kind of example is all we're intending to include on this page.
> 
> There's a somewhat related issue #995 that might lead to some more content being added to the Styling Guide page, and that may lead to another mention of `class:list` as well. So I think just changing lines 106 and 107 here on this page would be most helpful!

Sounds good, I'll do that then 👍

src/pages/en/core-concepts/astro-components.md
```diff
@@ -104,7 +104,7 @@ const myFavoritePokemon = [/* ... */];
 <ul>
 
 <!-- Use a template directive to inject an unescaped HTML string into an element: -->
```
sarah11918, line 106:
```suggestion
<!-- Use a template directive to build class names from multiple strings or even objects! -->
```

src/pages/en/core-concepts/astro-components.md
```diff
@@ -104,7 +104,7 @@ const myFavoritePokemon = [/* ... */];
...
-<p set:html={rawHTMLString} />
+<p set:html={rawHTMLString} class:list=["add", "dynamic", {"classNames": true}] />
```
sarah11918, line 107:
```suggestion
<p class:list={["add", "dynamic", {classNames: true}]} />
```

sarah11918 (commented):
Thanks for your patience, @shaunchander !

We discussed this, and decided that having only class list made the example a little cleaner, so I took out the set:html part and updated the comment accordingly.

I'm glad that we now have an example of class list on this page!  🥳

Issue #793, opened 2022-06-17
SSR page could explain what the adapter does / why it is necessary


Jutanium:
Currently, we state that you _need_ to use an adapter, and then describe _how_ to install the adapter.

`To enable SSR you need to use an adapter. The following adapters are available today with more to come in the future... In this example we will use`

In the next section, Features, we do explain that adapters provide additional features.

Still, we never quite address head-on what an adapter _does_ and _why_ we need one for SSR mode (while we don't for the default static site generator mode).

Aside from being an opportunity for deeper understanding, if we add a line or two answering that, we can reduce the potential for confusion if someone discovers the deno/netlify/vercel adapter and thinks they need to install it because they're _deploying_ to that platform (when in fact they can deploy fine without it if they're not using SSR).

PR #794, opened 2022-06-17
Clarify why you need an adapter

Jutanium:
Proposes a fix to https://github.com/withastro/docs/issues/793

- Change "is enabled" to "can be enabled", as SSR is not enabled by default
- Describe what a runtime is
- Add an explanation for why we need an adapter: it provides a runtime
- Clarification that if you don't need SSR, you don't need an adapter

### DIFF
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -4,7 +4,7 @@ title: Server-side Rendering
 i18nReady: true
 ---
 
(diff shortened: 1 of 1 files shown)


sarah11918:
LGTM, Dan!

Squash. Merge. Go forth and spread the word about why you need an adapter.

Jutanium:
❤️🚀🤹‍♀️🧩

PR #798, opened 2022-06-18
Make fallback content notice sticky & dismissible

### LINKED ISSUE #646: links to `#anchor` when the target page is not translated
A general comment about links to `#anchor` when the target page is not translated: the user will not see the message "This page hasn't been translated yet". This might surprise the user, because they are not expecting to see English and do not see the warning. Maybe the warning needs to be sticky/absolutely-positioned?

_Originally posted by @jbrains in https://github.com/withastro/docs/pull/463#discussion_r880881262_

delucis:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

Make the notice indicating a page is not translated yet sticky. Closes #646

### DIFF
--- a/src/components/FallbackNotice.astro
+++ b/src/components/FallbackNotice.astro
@@ -1,9 +1,73 @@
 ---
 import Aside from './Aside.astro';
 import UIString from './UIString.astro';
+import { useTranslations } from '../i18n/util';
+const t = useTranslations(Astro);
 ---
 
-<Aside>
-	<UIString key="fallbackContent.notice" />
-	<a href="https://github.com/withastro/docs/blob/main/src/i18n#readme"><UIString key="fallbackContent.linkText" /></a>
-</Aside>
+<div class="fallback-notice">
+	<input type="checkbox" id="fallback-notice-close-checkbox" class="sr-only" aria-hidden="true">
+	<label for="fallback-notice-close-checkbox" aria-hidden="true">
+		<UIString key="fallbackContent.closeLabel" />
+	</label>
+	<Aside type="caution" title={t('fallbackContent.title')}>
+		<UIString key="fallbackContent.notice" />
+		<a href="https://github.com/withastro/docs/blob/main/src/i18n#readme">
+			<UIString key="fallbackContent.linkText" />
+		</a>
+	</Aside>
+</div>
+
+<style>
+	.fallback-notice {
+		position: fixed;
+		bottom: .5rem;
+		left: 50%;
+		transform: translateX(-50%);
+		width: calc(100% - 1rem);
+		max-width: 35rem;
+		z-index: 5;
+		background: var(--theme-bg-gradient-bottom);
+	}
+
+	@media (min-width: 37.75em) {
+		.fallback-notice {
+			bottom: 1.5rem;
+			z-index: 50;
+		}
+	}
+
+	@media (min-width: 50em) {
+		.fallback-notice {
+			bottom: 2.5rem;
+		}
+	}
+
+	.fallback-notice :global(aside) {
+		box-shadow: 0 0 var(--theme-glow-blur) var(--theme-glow-diffuse);
+	}
+
+	label {
+		position: absolute;
+		top: 0;
+		right: 0;
+		font-size: .875rem;
+		padding: .375rem .625rem;
+		color: var(--theme-text-lighter);
+	}
+
+	:global(.theme-dark) label {
+		color: var(--theme-text);
+	}
+
+	/* Highlight label for keyboard users */
+	input[type="checkbox"]:focus-visible~label {
+		outline: 2px solid currentColor;
+	}
+
+	/* Hide fallback notice when checkbox is checked */
+	input[type="checkbox"]:checked,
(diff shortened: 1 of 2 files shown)


sarah11918:
Looks great on the screen widths I tested, light and dark mode! @Yan-Thomas ? @hippotastic ?

Yan-Thomas:
Since I'm on my sister's laptop I can't do a more detailed accessibility test, but I would expect the "Not Translated" aside to be focused the moment I open the page. This way, keyboard-only users could more easily close it and then read the page, since the aside gets in the way of the text and is at the very end of the focusable elements list. Besides that, it looks good to me. Great work as always, @delucis :raised_hands:

delucis:
The notice hasn't moved in the source — it's still at the top of the page, so I'm not sure of focusing on load. But I'd definitely wait for a trademark Yan screenreader test 💜

src/components/FallbackNotice.astro
```diff
@@ -1,9 +1,73 @@
...
+import { useTranslations } from '../i18n/util';
+const t = useTranslations(Astro);
 ---
 
-<Aside>
-	<UIString key="fallbackContent.notice" />
-	<a href="https://github.com/withastro/docs/blob/main/src/i18n#readme"><UIString key="fallbackContent.linkText" /></a>
-</Aside>
+<div class="fallback-notice">
+	<input type="checkbox" id="fallback-notice-close-checkbox" class="sr-only" aria-hidden="true">
```
Yan-Thomas, line 9: The close button when focused is being read as "blank". I removed aria-hidden and added the button role to the input and now it is read as "Close, button". I'm not 100% sure if `role="button` is necessary, but I think it's better than "Close, checkbox, not checked"
```suggestion
	<input type="checkbox" id="fallback-notice-close-checkbox" class="sr-only" role="button">
```

Yan-Thomas:
By the way, I just noticed I can't close the aside by pressing "enter" or using the directional keys (since it's a checkbox). The aside isn't blocking the user view so much and can be reached out easily going through the landmarks keyboard shortcut in NVDA, so I'm not sure if this is something we can consider "immediate action" (when it's recommended to auto-focus).

Hi @mayank99, could you please share your thoughts about this? Would help us a lot! 🙌

hippotastic:
Thank you so much for your testing! Does it work better with a radio button instead of a checkbox? I know that the keyboard inputs can be different depending on the form element type.

Yan-Thomas:
> Does it work better with a radio button instead of a checkbox? I know that the keyboard inputs can be different depending on the form element type.

TIL: You can check checkboxes using "Space", but not "Enter", for some reason. 😅  
Pressing "Space" closes it normally, so I think we don't necessarily need to change the input type, but yeah, a radio button would allow users to use the directional keys (but not Enter!).

We could just allow the user to check the checkbox by pressing enter using a `keypress` JavaScript event, but It kinda defeats the purpose of avoiding JavaScript here.

mayank99:
Is there a good reason to avoid JavaScript here? I haven't seen a checkbox being used as a button in a long time.

If a keyboard user focuses the checkbox but it's announced as a button, they would likely expect <kbd>Enter</kbd> to perform the action.

delucis:
Thanks for the testing everyone! I was actually trying to remove the close button entirely for screenreaders as it seemed that the ability to dismiss the content only really made sense for a sighted user. Strange that it still gets read as "blank" 🤔

Thinking further, I'm not sure how best to solve this issue for screenreader users as following an anchor link will wouldn't read this notice.

We can use JS if necessary obviously — just thought it might not be needed in this case!

mayank99:
> Strange that it still gets read as "blank"

When a focusable element has `aria-hidden`, the screenreader has nothing to announce but it still needs to announce something to indicate the focus, so "blank" makes sense.

> I'm not sure how best to solve this issue for screenreader users as following an anchor link still wouldn't read this notice.

Maybe using an `<output>` element (which is an `aria-live` region) would help?

kevinzunigacuellar:
I love Chris's work for the Astro docs but the content notice doesn't have my vote. I think it does more harm than good in terms of user experience. Users might get frustrated or annoyed with additional forced interactions.

sarah11918:
Thatnks for that comment, @kevinzunigacuellar , I hadn't considered that!

delucis:
Thanks Kevin! Appreciate the feedback.

What do we think then folks? Accept that jumping to the middle of an untranslated page may just have this unfortunate side effect of not reporting why that is?

One option that would work only for larger screens would be to move the notice to the right sidebar:

image

A limited bandaid, but maybe less disruptive?

delucis:
We could do that, but I’d also always want a manual dismiss option. The whole problem is that it covers UI so we need a way to get rid of it as fast as possible if someone wants. (So Kevin is right that there’s an annoyance to that.)

sarah11918:
If only we had a #vote channel! Maybe see if we can rally teamdocs/i18n-gang for more opinions/ideas in a docs channel?

delucis:
Very good idea @hippotastic!

delucis:
In fact, I like it enough to close this for now in favour of that solution instead. We can always revive this PR later if we decide we want it after all.

PR #809, opened 2022-06-21
docs: add target property to open in playground buttons

EmaSuriano:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

I was reading the docs and clicked on Open in CodeSandbox which made a redirect and took me from the documentation (which is not ideal).

By providing `target="_blank"` to the "Open in Playground" buttons, the user will have a new tab opened and keep the documentation in the original one. No visual changes.

### DIFF
--- a/src/pages/en/getting-started.md
+++ b/src/pages/en/getting-started.md
@@ -24,8 +24,8 @@ Visit [astro.new](https://astro.new/) for the easiest way to "try before you buy
 Or, **instantly launch our basic starter project** with just one click of a button:
 
 <div style="display: flex; flex-wrap: wrap; gap: 0.5rem;">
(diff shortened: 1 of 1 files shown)


delucis:
Hi @EmaSuriano — thanks for your interest in contributing to the Astro docs! Really appreciate it.

In general we prefer not to open in a new window/tab by default as that may not be what a user wants — and if they do they can use "open in new tab" to get that behaviour. The inverse — opening in the current tab even with `target="_blank"` set — is not available to users I don't think, so I'd prefer to stick with the current set up.

Thanks again for the idea though and if you have any further ideas on how to improve the docs, don't hesitate to use our issues and discussions to suggest them! 🙌

PR #810, opened 2022-06-21
Prefetch pages to speed up navigation

### LINKED ISSUE #813: Add support for route prefetching
Rich Harris selected this repo's docs as [an example](https://twitter.com/rich_harris/status/1539084063558062080?s=21&t=RPHzjZ_2MXac-Z2PXAOgNA) of how SSR routing is less performant than CSR, and it's true that the UX for these docs is noticeably worse than the SPA example.

This appears to only be the case because the request upon clicking actually goes down the wire and has not been precached by the page after render. It is possible to prefetch links on a page such that they are cached in advance and will respond 304 on an actual click, reducing the slowdown to only the difference between disk read and memory read (negligible).

See [next/link](https://nextjs.org/docs/api-reference/next/link) for an example of how this should be done.

matthewp:

- [ ] New or updated content
- [ ] Translated content
- [x] Something else!

#### Description

- Adds prefetch link for faster page navigation.
- Please try out the preview URL and see if it feels faster for you.
- Closes #813

Comparison between the two

### DIFF
--- a/src/components/LeftSidebar/SidebarContent.astro
+++ b/src/components/LeftSidebar/SidebarContent.astro
@@ -32,6 +32,7 @@ const lang = getLanguageFromURL(Astro.canonicalURL.pathname);
 			<ul>
 				{section.children.map(({ slug, text, isFallback }) => (
 					<li class="nav-link">
+						<link rel="prefetch" href={`${Astro.site.pathname}${lang}/${slug}/`}>
 						<a
 							href={`${Astro.site.pathname}${lang}/${slug}/`}
 							aria-current={`${currentPageMatch.endsWith(slug) ? 'page' : 'false'}`}


orenelbaum:
From trying this out on my phone it seems faster. It doesn't seem as fast as it can be for whatever reason, compare it to https://markojs.com/docs/ for example, but on my phone it's fast enough that real users won't care at all.

Yan-Thomas:
I tested both on my smartphone (with a slow 3G connection) and Chrome Dev Tools (simulated slow 3G connection) and I didn't feel the extra prefetching payload degraded the performance. Also, I think it would be nice if we added `prefetch` links to [`InstallGuideTabGroup.astro`](https://github.com/withastro/docs/blob/main/src/components/TabGroup/InstallGuideTabGroup.astro) as well.

delucis:
So in Chrome I’m seeing `DOMContentLoaded` 10–15% earlier when using a simulated “Slow 3G” connection. Hard to tell any difference without throttling, but I think in slower network conditions it does seem to speed things up.

Couple of weird things:

- I don’t seem to see any prefetching happen in Firefox. Maybe they skip it because its too many links?
- Chrome re-pre-fetches everything on each page load, which seems… wasteful?

orenelbaum:
> * I don’t seem to see any prefetching happen in Firefox. Maybe they skip it because its too many links?
> * Chrome re-pre-fetches everything on each page load, which seems… wasteful?

I hope that you guys don't mind that I shared the preview deploy on the Twitter thread but Rich actually addressed those points
https://twitter.com/Rich_Harris/status/1539254229071712258
https://twitter.com/Rich_Harris/status/1539254472035254273

I'm not sure what rich meant by adding some js, maybe prefetching on hover

delucis:
Thanks for sharing @orenelbaum! Yes, I assume by JS he'd mean prefetch on hover or visible or something.

I also kind of agree with him that Firefox's behaviour is probably correct if unhelpful in this case.

ctjlewis:
Very nice work, thanks for implementing.

ctjlewis:
@matthewp Can I get a Discord invite? Would like to help out more with this issue.

Issue #805, opened 2022-06-19
PT-BR pages Google and Algolia Search issues


Yan-Thomas:
The Brazilian Portuguese docs are having some weird problems, the first one is that Algolia Search doesn't find any of the pt-BR pages, so any search results in "No results ...", as you can see in the gif below:

![5EnDbvd](https://user-images.githubusercontent.com/61414485/174502795-4c543cd8-46cc-4f0c-b9d0-600c52bccd73.gif)

The second one, Google Search is indexing the Brazilian Portuguese pages, but for some reason, some of them have their title/description in English. The page content itself is in Brazilian Portuguese, and I can see that the meta attributes are properly translated (Astro Components pt-BR page source code):

![image](https://user-images.githubusercontent.com/61414485/174503792-1cdff8ea-1a6e-42ac-9edc-59653604af1b.png)

I'm using Google's "in cache" feature to see when the page was supposedly last crawled, here's a list of pages I tested and the behaviors I found:

### Hidratação Parcial  (Partial Hydration)
- [x] Correct title and description
- [x] Includes pt-BR as part of the link breadcrumb
[Cache](https://webcache.googleusercontent.com/search?q=cache:qlw0utFTggUJ:https://docs.astro.build/pt-BR/core-concepts/partial-hydration/+&cd=1&hl=pt-BR&ct=clnk&gl=br): May 10, 2022

### Solução de Problemas (Troubleshooting)
- [x] Correct title and description
- [ ] Includes pt-BR as part of the link breadcrumb
[Cache](https://webcache.googleusercontent.com/search?q=cache:H7j_XHEvgKUJ:https://docs.astro.build/pt-BR/guides/troubleshooting/+&cd=2&hl=pt-BR&ct=clnk&gl=br): June 10, 2022
![image](https://user-images.githubusercontent.com/61414485/174503471-08fac311-58ee-4966-81c7-634211651b85.png)

### Componentes Astro (Astro Components)
- [ ] Correct title and description
- [x] Includes pt-BR as part of the link breadcrumb
[Cache](https://webcache.googleusercontent.com/search?q=cache:21X7leXNzMMJ:https://docs.astro.build/pt-BR/core-concepts/astro-components/+&cd=1&hl=pt-BR&ct=clnk&gl=br): 19 June, 2022
![image](https://user-images.githubusercontent.com/61414485/174503609-b7bf3416-66b9-4eb2-8f6d-7a8e5cbe10dd.png)

[...]

PR #833, opened 2022-06-24
minor: Use lowercase language tags for canonical URLs

delucis:

#### Description

@Yan-Thomas reported strange search engine indexing of `pt-BR` pages in #805 and we (@hippotastic and I) suspect this is caused by the fact we use `/pt-BR/` canonical URLs even though Netlify will always redirect those to `/pt-br/`. (The same applies for other languages with a regional identifier in their tag: `zh-CN` and `zh-TW`.)

This PR touches a lot to fix this unfortunately, but I think it’s all necessary:

1. Rename our directories to use lowercase language tags, moving _all_ `pt-BR`, `zh-CN` and `zh-TW` pages and i18n files.

2. Update our languages dictionary to use the lowercase format (e.g. `pt-br`).

3. In places where we require a correctly formatted tag (`lang` & `hreflang` attributes, sitemap, etc.), apply a simple normalizer function to re-uppercase the region identifier.

4. Rewrite _all_ `pt-BR`, `zh-CN` and `zh-TW` links.

5. Use the lowercase format in the translation status checker.

> **Warning**
> Before merging, I think I need to manually update the i18n Overview Status issue data (@hippotastic can confirm).

This will hopefully address #805 but I’d suggest leaving it open until we can confirm (Google will need to recrawl).


hippotastic:
Wow, thanks for this PR @delucis! On it :)

hippotastic (commented):
Sorry it took a while! I did a full comparison of the build output of the current docs site vs. your PR, ran the link checker on the output, manually checked the sitemap output, the SEO-related HTML tag and attributes, as well as all changed links inside documents.

Other than a few places where your search & replace pass apparently changed a bit too much (see comments), this looks really great! Awesome job!

hippotastic:
Regarding your note on replacing the JSON state data in the i18n Status Overview issue, you're spot on. Luckily it's only two root-level keys that need to be changed, everything else should stay the same.

Also, when merging, we need to ensure the "minor" prefix is contained in the squashed commit message.

If those two points are taken care of, we should be able to merge this and everything should work - and in case it doesn't, we can still fix it then. 😄

PR #838, opened 2022-06-24
Fix recursive 404

nikhilhenry:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

The current 404 page on the docs site fails to redirect to the homepage. This creates a recursive 404 loop and hampers the UX. 

This PR addresses this issue by utilising the `getLanguageFromURL()` function to identify the correct locale and redirect to the appropriate homepage (with that locale). 
No visual changes have been made in this PR.

### DIFF
--- a/src/pages/[lang]/404.astro
+++ b/src/pages/[lang]/404.astro
@@ -1,11 +1,9 @@
 ---
-import languages from '../../i18n/languages';
+import { getLanguageFromURL } from '~/util';
 import { useTranslations } from '../../i18n/util';
 import SplashLayout from '../../layouts/SplashLayout.astro';
 
-export const getStaticPaths = () => Object.keys(languages).map(lang => ({ params: { lang } }));
-
-const { lang } = Astro.params;
+const lang = getLanguageFromURL(Astro.canonicalURL.pathname) || 'en';
 const t = useTranslations(Astro);
 ---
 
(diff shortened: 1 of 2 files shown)


src/pages/[lang]/404.astro
```diff
@@ -1,11 +1,11 @@
...
-import languages from '../../i18n/languages';
+import languages from '~/i18n/languages';
+import { getLanguageFromURL } from '~/util';
 import { useTranslations } from '../../i18n/util';
 import SplashLayout from '../../layouts/SplashLayout.astro';
 
 export const getStaticPaths = () => Object.keys(languages).map(lang => ({ params: { lang } }));
-
-const { lang } = Astro.params;
+const lang = getLanguageFromURL(Astro.canonicalURL.pathname) || 'en';
```
Comment (delucis) [inline src/pages/[lang]/404.astro:8]: The `lang` is available in params everywhere except on the root 404, so I think we can just add a fallback value here and avoid importing `getLanguageFromURL`.

```suggestion
const { lang = 'en' } = Astro.params;
```

delucis (commented):
Thanks for this @nikhilhenry! I think we can actually do this even more simply.

src/pages/[lang]/404.astro
```diff
@@ -1,11 +1,10 @@
 ---
-import languages from '../../i18n/languages';
+import languages from '~/i18n/languages';
```
Comment (delucis) [inline src/pages/[lang]/404.astro:2]: Amazing, thanks @nikhilhenry!

I'm just going to revert this switch to using the alias for consistency with the rest of the codebase, but this is great!


```suggestion
import languages from '../../i18n/languages';
```

delucis (approved):

delucis:
Woo! Thanks for another great contribution @nikhilhenry 🥳

PR #849, opened 2022-06-25
Sidebar: Highlight subpages

### LINKED ISSUE #844: Highlight Deploy/Install pages inside "subpages"
When we move from the main deploy page to any of the deployment services guides or from the automatic to manual installation guide, we end up having no page highlighted as the current one in the left sidebar. This affects other things: `LeftSidebar` doesn't scroll to the "Deploy" page link and the new "open current page section on mobile" feature doesn't work because there isn't any page marked as "current". Theoretically yes, they aren't on the same page, but the impression is that we are on a subsection of the same page (even more on the Install guides), and conveying this to the users through the left sidebar would be nice.

Example:
image

Yan-Thomas:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description
Closes #844 

This PR highlights subpages (like `install/manual` and `deploy/netlify`) as part of the same page. This allows other sidebar features:tm: to work with them, like the "scroll sidebar to current page" and "open current page on mobile" features.

| docs.astro.build  |  PR  |
|  current-page-not-highlighting | current-page-highlight |

NOTE: I moved the individual deploy pages to be part of `/guides/` like the main deploy page, this simplifies the solution and just feels right that the individual deploy pages are still part of "guides" like the main page. We can undo only this and would just mean adding one more if statement to my solution.

Also, this solution is kinda hardcoded, if we can find a better solution that works without specifying the pages that have subpages would be very nice. This is a simple PR so we can scrap it and go for a better solution without worries.


src/util.ts
```diff
@@ -13,6 +13,15 @@ export function removeTrailingSlash(path: string) {
 	return path.replace(/[/\\]+$/, '');
 }
 
+/** Remove the subpage segment of a URL string */
+export function removeSubpageSegment(path:string) {
+	// Include new pages with subpages as part of this if statement.
+	if (path.includes('install/' || 'deploy/')) {
```
delucis, line 19: This will never test if the path includes `deploy/`. Maybe do it with a regular expression?

But the fact it’s still working makes we wonder if we even need this utility function 🤔 

```suggestion
	if (/(?:install|deploy)\//.test(path)) {
```

delucis (commented):
Thanks @Yan-Thomas! Looks like there are a couple of broken links in `src/pages/en/guides/deploy.md` that need updating to the new deploy page location (agree with moving the guides).

One thing I’m not quite happy with is adding `aria-current` to a link which isn’t actually the current page. Might need to think of an alternative like a `data-current-parent` for parent pages.

src/util.ts
Yan-Thomas: lol, I don't even know why I tried to use "OR" inside `.includes()`. 
The reason why it still works is that I changed `.endsWith()` to `.includes()` here:

```jsx
<a href={`${Astro.site.pathname}${lang}/${slug}/`} aria-current={`${currentPageMatch.includes(removeSubpageSegment(slug)) ? 'page' : 'false'}`}>
```

Deploy subpages, like `en/guides/deploy/netlify` includes the slug `guides/deploy` but since the install parent page is `install/auto` we need the utility function. We can completely remove the utility function if we change `install/auto` to only `install`, how do feel about this?

Also, I just found that I was using this utility function in places I didn't even needed it :sweat_smile:

Yan-Thomas: I think we can do something similar to deploy as well: change the main page to `index.md` and place it inside the `deploy` folder (Maybe it can be kinda weird to have multiples `index.md` pages, so we can just leave `deploy.md` and `install.md` outside its subpages folder)

Yan-Thomas: > We can completely remove the utility function if we change install/auto to only install, how do you feel about this?

Actually, this opens the possibility for future bugs in the current state of things. We need to know it's a URL segment by including `/` as part of `currentPageMatch`, so we can use `includes()` without worrying about matching stuff that shouldn't match. We would still change the automatic installation page to be only `install`, but `currentPageMatch` would not remove the trailing slash, so we are sure it's an entire segment of the URL and not something like `guides/install-integrations`. This way we can just compare the URL fragment and know it is a subpage without a hardcoded utility function mentioning which pages have subpages.

I have this new solution staged and working as intended, I just want to be sure if I can change `install/auto` to `install` just to avoid committing these changes and then having to rollback :sweat_smile:.

Yan-Thomas:
Thanks for the suggestion @delucis! This is now working fine, but I still would like to make this work without the hardcoded utility function. This would need to restructure the install pages and document a pattern for pages with subpages as a whole, so we can use the URL's path to determine if something is a subpage.

delucis (approved):
LGTM, @Yan-Thomas!

As discussed on Discord, we'll keep the utility helper for now so we don't need to move the install guides. Thanks so much for another excellent sidebar improvement 🚀

PR #860, opened 2022-06-26
docs(zh): add glossary

kecrily:

- [ ] New or updated content
- [ ] Translated content
- [x] Something else!

#### Description

add glossary


sarah11918:
Thanks, @kecrily ! I'm going to ping @delucis here because of the link checks, so he can guide you through them for adding this glossary.

delucis:
Hi @kecrily! Who is the target audience of this glossary? Is it for other translators or for people reading the docs?

Currently the “Check for Mismatched Slugs” CI only allows pages with slugs that match the slugs in `src/pages/en` so if you’re looking to add an extra public page we might need to rethink that as well as thinking about where to link to this from.

If this is more intended for translators, I think I’d suggest moving this to `src/i18n/zh-cn`.

delucis:
Perfect. Thanks!

PR #862, opened 2022-06-26
docs: fix <scripts> typo

kecrily:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

No <scripts>, just <script>


sarah11918:
You're right, @kecrily ! The tag is absolutely only `<script>` singular, but hmm... now the sentence (in English) won't read properly unless we add an 's' after, which I'm not sure I love, or we rewrite the sentence.

My first thought is to change the English to: 

📚 Learn more about [using a client-side `<script>` in Astro components](/en/core-concepts/astro-components/#client-side-scripts)

What does everyone think of that? If we do that, then this PR should *only* change the English, and we should let the translators craft the new sentence in their own languages.

kecrily:
I thought that the content included by the code block is not considered singular or plural. It should just refer to a specific thing.

sarah11918:
@kecrily When this code word is wrtten inside an English sentence, we would typically read the word inside the tag as if it were just a regular word in the sentence. So, it would sound a little off in English if we didn't decide whether it is singular or plural, and make sure the overall sentence reads like a normal sentence. 

This might not be true for every language, but especially since I don't know, this is a good reason to let every language handle it themselves, I think. For the languages where it doesn't matter, they might not need to do anything but remove the "s." But other languages might need to change their wording, like English does.

I think in English, I'd prefer to add wording to force the sentence to read singular. What about everyone else?

delucis:
Makes sense @sarah11918!

The other option would be to drop the tag styling and just use “client-side scripts”, but we’d still want to scope this PR to English and let other languages catch up with the right wording for them.

sarah11918:
I'm OK with either option, although part of me *maybe* has a slight preference for actually writing a `<script>` tag:

- it's a little more visual
- it's also its own shout out "Hey! Remember you can totally do script stuff!" in this section about whether or not you can hydrate an Astro component. 

So in this case, I think it might be helpful for people who are scanning and their brain sees:
"Can I hydrate an Astro component?"
...
`<script>`

(That's almost the answer right there, alone!  😄 )

src/pages/en/core-concepts/framework-components.md
```diff
@@ -214,7 +214,7 @@ Astro components are always rendered to static HTML, even when they include fram
 
 [Astro components](/en/core-concepts/astro-components/) are HTML-only templating components with no client-side runtime. But, you can use a `<script>` tag in your Astro component template to send JavaScript to the browser that executes in the global scope.
 
-📚 Learn more about [client-side `<scripts>` in Astro components](/en/core-concepts/astro-components/#client-side-scripts)
+📚 Learn more about [client-side `<script>` in Astro components](/en/core-concepts/astro-components/#client-side-scripts)
```
sarah11918, line 217:
```suggestion
📚 Learn more about about using [client-side `<script>` tags in Astro components](/en/core-concepts/astro-components/#client-side-scripts)
```
I think this might be my preference.

sarah11918:
Thanks @kecrily ! And, thank you for being my test case for a new GitHub browser extension with more features so I could revert the changes in the other languages for this! ;)

PR #868, opened 2022-06-27
chore: format code style

kecrily:

- [ ] New or updated content
- [ ] Translated content
- [x] Something else!

#### Description

add `pre-push` git hook


delucis:
Thanks for this @kecrily. I’m not 100% sold on this. In my experience, pre-push checks can be a barrier to less technically comfortable contributors: they try to push, are blocked, and are then alone on their machines to try and figure out how to get unblocked and share changes. I actually don’t mind seeing checks fail in a PR where myself and other maintainers can help out if someone isn’t sure how to address them.

Sorry that you spent time on this. If in doubt, it’s always best to post in [Discussions](https://github.com/withastro/docs/discussions) or on Discord and we can chat about new features like this before you invest the time.

If you think it would be helpful, maybe we could improve `CONTRIBUTING.md` to better explain how to run the link checker etc. manually if you want?

kecrily:
Maybe we can keep the `prettier` and `eslint  . --fix` without causing errors

delucis:
That’s an idea, although I guess if you’re fixing, it should be pre-commit rather than pre-push? I also wouldn’t include Prettier yet as it seems to touch _a lot_ of files currently. I think we’d want to rethink our settings first (line length for example).

One other negative: What if someone doesn’t install dependencies? I guess people might pull, edit Markdown, and push — without ever opening the terminal.

kecrily:
> One other negative: What if someone doesn’t install dependencies? I guess people might pull, edit Markdown, and push — without ever opening the terminal.

It is indeed possible.

Let me change this PR to files change formatting with Prettier.

delucis:
Oh right! Didn’t see that coming 😅 

Could we update this setting then to use `100` instead of `180`?

https://github.com/withastro/docs/blob/a4586477eb4ee1c0e1872534354303945e777279/.prettierrc#L2

I think we’ve avoided auto-formatting with Prettier partially because `180` is too long. `100` is what the `withastro/astro` repo uses, so seems a safer bet.

delucis:
Oh, one more thing! (Sorry — should have caught it before.)

We don’t want Prettier to format `pnpm-lock.yaml`. Can we add it to `.prettierignore`?

PR #870, opened 2022-06-27
Add collapsible "on this page" in smaller screens

Yan-Thomas:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

This PR adds a sticky "on this page" navbar for mobile/tablet users:

m1Lec97

I'm marking this as a draft because there are some blocking issues I couldn't fix:

1 - As you can see, some elements are kinda breaking out from the parent element width, so this causes a very ugly effect:
image

I think to fix this we would need to have a way to make "on this page" cover more than its parent width but I'm not sure what would be the best approach here.

2 - When the header also becomes sticky the navbar isn't sticky anymore. Probably some very specific CSS behavior I couldn't find an answer for 😅.


src/components/PageContent/PageContent.astro
```diff
@@ -57,6 +58,13 @@ const t = useTranslations(Astro);
 	.block {
 		display: block;
 	}
+	.sticky-nav {
+		width: 100%;
+		position: sticky;
+		z-index: 2;
+		top: 0;
+		background: var(--theme-bg-gradient);
+	}
```
delucis, line 67: Thanks so much for this @Yan-Thomas! I’m going to try to get this merged TODAY 🤞

Your first issue is easy to address by extending the nav left a bit like you suggested:

```suggestion
		margin-inline-start: -2rem;
		padding-inline-start: 2rem;
	}
```

Taking a look now to see if I can figure out the sticky thing — probably something to do with a parent `position` value for something. Sticky is tricky!

src/components/PageContent/PageContent.astro
```diff
@@ -57,6 +58,13 @@ const t = useTranslations(Astro);
 	.block {
 		display: block;
 	}
+	.sticky-nav {
+		width: 100%;
```
delucis, line 62: Ah, tweaking margin and padding also impacts width so I’d probably suggest:

```suggestion
		--sticky-nav-overlap: 2rem;
		width: calc(100% + var(--sticky-nav-overlap);
		margin-inline-start: calc(-1 * var(--sticky-nav-overlap);
		padding-inline-start: var(--sticky-nav-overlap);
```

delucis:
@Yan-Thomas hope you don’t mind I ran with this. Let me know what you think!

Yan-Thomas:
Thank you so much for all these amazing improvements, @delucis! 🙌 💛 
LGTM!!

sarah11918:
I've been letting this PR just chug along, seeing what you all have been doing with it. After Chris' demo it's *quite impressive*! (And, I hear this is working to make our Lighthouse scores 100s, so that's great, too!) I like that it's sticky; I think that's great for navigation!

1. I did have a "Sarah's Phone" issue: in Chrome (and Kiwi, another Chromium-based browser):
The workflow that seems to bork is:
- expand menu
- select one of the nav links to go to a section
- the navigation component disappears completely, and neither scrolling up nor down makes it visible again.

2. I'm not 100% convinced on the initial loaded state of collapsed, and e.g. ON THIS PAGE > Overview

It makes me think that what is on this page is.... an overview! the Right arrow could be interpreted as a symbol for "one level deeper" navigation instead of "a collapsed menu." Random possible thoughts for dealing with this:

- NOT ever have "Overview" show up there, so that the > symbol looks more like something meant to be expanded?
- use a different symbol or word or styling that makes it look more like, on mobile a thing that can expand?
- start expanded (didn't we talk about this before, where I thought that the initial list expanded was actually helpful for the reader to know they're on the correct or a useful page? I still think this is a plus) BUT the fact that it's sticky when expanded means on initial scroll, you might be taken aback by the small amount of screen real estate actually scrolling (if any at all, when the list is long enough!) until you figure out you can collapse it.

Those are my initial thoughts!

delucis:
Remind me of your phone + browser combo, Sarah? I can’t seem to get that to happen for me.

Re: 2. I’ve just hidden the “Overview” text from the summary. PTAL.

The next approach I’d suggest would be better styling. I went with minimal changes for this PR, but it could look a little more like a button for example to help prompt people if you think it’s still not clear.

delucis:
Example of something like what I mean by “like a button”:

| | |
|---|---|
| image | image |

delucis:
Or even:

<table>
<tr>
	<td>image
	<td>image
	<td>image
</table>

sarah11918:
Samsung Galaxy note 8, Chrome (and Kiwi); Android 9, OneUI 1.0

sarah11918:
UPDATE It's *closer* to working on my phone!

The update is that now, when I click a TOC link and the navigation TOC disappears, I can scroll all the way to the top, and now it's there.

Visually, looking 🔥 !

sarah11918:
LGTM!  NWTWWHB!

I know on Chrome mobile it doesn't have *all* the features (yet?) that it does in FF, but it doesn't break anything, and still gives the very slick collapsable menu (just non sticky) available at the top.

Plus, LIGHTHOUSE!!  🟢 💯

delucis:
Just to leave a note as I merge:

There’s an outstanding bug that once a link has been clicked and the page is a `...url#heading`, the sticky behaviour seems to break in _several_ mobile browsers. Needs more investigation.

PR #890, opened 2022-06-30
Replace `react` with `preact/compat`

delucis:

- [ ] New or updated content
- [ ] Translated content
- [ ] Something else!

#### Description

- Drop React and use `preact/compat` instead

- ~~Currently throws a `Cannot find module 'react'` error. This seems to be happening when something tries to load Docsearch’s UMD distribution. I wonder if this is due to `@docsearch/react` [setting the UMD build as `main` in its `package.json`](https://github.com/algolia/docsearch/blob/d7be594fc62c8ee8573dbaf5ffad3f4bb9149002/packages/docsearch-react/package.json#L22)? Not sure. Can we tell Vite some how to always ignore UMD and bundle the ESM distribution?~~

- ~~Currently working in dev but fails when building with this error:~~

   ~~`[vite]: Rollup failed to resolve import "react" from "node_modules/.pnpm/@docsearch+react@3.1.0/node_modules/@docsearch/react/dist/esm/DocSearchModal.js".`~~

- Currently working after adding Preact’s Vite preset plugin


tony-sull (approved):
LGTM!  Very cool to see so much JS shake out of the final bundle size 👀

PR #902, opened 2022-07-01
Improve GitHub PR template

delucis:

- Something else!

#### Description

- Adds a prompt for an issue number if the PR should close one
- Switches the list to a “Delete any that don’t apply” to get around the “tasks UI” GitHub associates with checklists

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,6 +11,7 @@
 
 #### Description
 
+- Closes # <!-- Add an issue number if this PR will close it. -->
 - What does this PR change? Give us a brief description.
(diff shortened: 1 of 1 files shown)


sarah11918:
LGTM! Thank you!

sarah11918 (approved):
At some point we may want to consider adding points like:

- only submit content changes in English; we will make sure page is reflected in all languages
- "minor" doesn't just mean "small/tiny" - it means "doesn't trigger a translation request"

But I can't think of exactly what I want yet, and I want to write this stuff into CONTRIBUTING.md first anyway, so that's to think about another day!

delucis:
Yes, we could also say something like “Please read our CONTRIBUTING.md before opening a PR” or something rather than including too much in this template. GitHub will also show a link but still could be helpful.

PR #908, opened 2022-07-02
Redirect to the new deploy guides location

delucis:

- Something else!

#### Description

#849 moved the deploy guides from `/[lang]/deploy/[service]` to `/[lang]/guides/deploy/[service]` causing some 404s from old links like in Discord threads. This Netlify redirect makes sure we redirect from the old URLs to the new ones.


sarah11918:
LGTM!

PR #927, opened 2022-07-05
Improve sticky TOC on mobile devices and medium viewports

hippotastic:

#### Description

- Improves the sticky TOC on mobile devices like iOS Safari:
  - The TOC was not sticky on iOS Safari before, now it is
  - Scroll anchors are no longer obscured by sticky elements
  - You can now tap on top of the screen to go to the beginning of the page again
  - The dynamically resizing Safari tab bar now works again instead of being stuck in the expanded state

- Improves the sticky TOC on all devices with medium viewports:
  - Instead of the TOC being below the article heading and scrolling/stopping out of sync with the left sidebar, it is now always at the top of the main column (as discussed on Discord) and scrolls in sync with the left sidebar.

- Cleanup: Reduced magic numbers across the CSS code by moving them to clearly named variables in `public/theme.css`.

### DIFF
--- a/public/theme.css
+++ b/public/theme.css
@@ -11,7 +11,7 @@
 	--min-spacing-inline: 1rem;
 	/* Vertical spacing around the article content and the right sidebar */
 	--doc-padding-block: 0.5rem;
-	--max-width: calc(100%);
+	--max-width: 100%;
 	--cur-viewport-height: 100vh;
 }
 
--- a/public/theme.css
+++ b/public/theme.css
@@ -1,3 +1,44 @@
+/*
+	Define common sizes and spacings
+*/
+:root {
+	--theme-navbar-height: 6rem;
+	--theme-sticky-toc-height: 4rem;
+	/*
+		Minimum visual horizontal spacing from the edges of the viewport,
+		and between vertically arranged elements
+	*/
+	--min-spacing-inline: 1rem;
+	/* Vertical spacing around the article content and the right sidebar */
+	--doc-padding-block: 0.5rem;
+	--max-width: calc(100%);
+	--cur-viewport-height: 100vh;
+}
+
+@media (min-width: 50em) {
+	:root {
+		--min-spacing-inline: 1.5rem;
+		--doc-padding-block: 1rem;
+		--max-width: 46em;
+	}
+}
+
+@media (min-width: 72em) {
+	:root {
+		--doc-padding-block: 2rem;
+	}
+}
+
+/* Use dynamic viewport height if the unit is supported by the browser */
+@supports (height: 100dvh) {
+	:root {
+		--cur-viewport-height: 100dvh;
+	}
+}
+
+/*
+	Define common fonts and colors
+*/
 :root {
 	--font-fallback: -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial, sans-serif, Apple Color Emoji, Segoe UI Emoji;
 	--font-body: system-ui, var(--font-fallback);
@@ -5,15 +46,13 @@
 		'Liberation Mono', 'Nimbus Mono L', Monaco, 'Courier New', Courier, monospace;
 
 	/*
-   * Variables with --color-base prefix define
-   * the hue, and saturation values to be used for
-   * hsla colors.
-   *
-   * ex:
-   *
-   * --color-base-{color}: {hue}, {saturation};
-   *
-   */
+		Variables with --color-base prefix define
+		the hue, and saturation values to be used for
+		hsla colors.
+
+		Example:
+		--color-base-{color}: {hue}, {saturation};
+	*/
 
 	--color-base-white: 0, 0%;
 	--color-base-black: 240, 100%;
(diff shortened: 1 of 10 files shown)


hippotastic:
@delucis @Yan-Thomas PTAL :)

sarah11918:
I don't know why my evil phone was not also pinged for a PTAL, but fwiw, you have appeased even Sarah's phone with this upgrade, Hippo! I have none of my earlier issues re: completely disappearing TOC if you clicked on a section to navigate there.

hippotastic:
> I don't know why my evil phone was not also pinged for a PTAL, but fwiw, you have appeased even Sarah's phone with this upgrade, Hippo! I have none of my earlier issues re: completely disappearing TOC if you clicked on a section to navigate there.

Oops, my bad! I wanted to ping you, too. I just talked to Chris earlier today that I was really curious what issues your phone could come up with!

src/components/LeftSidebar/LeftSidebar.astro
```diff
@@ -72,14 +72,13 @@ for (const section of sidebarSections) {
...
 		width: 100%;
-		margin-inline-end: 1rem;
 		padding-top: 0rem;
 	}
 	.nav-groups {
 		height: 100%;
 		overflow-x: visible;
 		overflow-y: auto;
-		max-height: calc(100vh - var(--theme-navbar-height));
+		max-height: calc(var(--cur-viewport-height) - var(--theme-navbar-height));
```
delucis, line 81: I’m seeing one tricky bug that I think _might_ come back to this. Basically on mobile where we have the issue that `100vh` can be misleading, I’m sometimes seeing the header pull down _below_ the top of the screen when the menu is open. Can’t reproduce in the iOS simulator and can’t figure out how to capture my phone’s screen, but the flow is:

1. Scroll down the page a little so the address bar vanishes
2. Open the hamburger menu
3. Scroll up the nav menu

Result (note the dark empty space at the top):



Browser: Firefox
OS: Android 11

Can’t reproduce in Chrome on my phone so may be FF only? (This could also be a genuine bug with mobile FF when the address bar is at the bottom of the screen.)

delucis (commented):
Looks really good! A couple of small things and one bigger question that isn’t a complete blocker, but would be good to investigate.

public/theme.css
```diff
@@ -1,19 +1,58 @@
...
+	--theme-navbar-height: 6rem;
+	--theme-sticky-toc-height: 4rem;
+	/*
+		Minimum visual horizontal spacing from the edges of the viewport,
+		and between vertically arranged elements
+	*/
+	--min-spacing-inline: 1rem;
+	/* Vertical spacing around the article content and the right sidebar */
+	--doc-padding-block: 0.5rem;
+	--max-width: calc(100%);
```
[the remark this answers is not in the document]
hippotastic: I don't know! I just moved this line from `index.css` here to centralize all the variables that make up our theme. I *think* we should be fine to remove it.

src/components/LeftSidebar/LeftSidebar.astro
hippotastic: Hmm, can you reproduce this issue on the old site with this browser as well?

Whenever I noticed some hiccups of the left sidebar (and the hamburger menu is just a differently styled left sidebar), I checked the old page and saw that it also had the same issues. Unfortunately, the "incorrect height of sidebar" issues are not easy to fix, and if I wasn't able to, I let them be because at least I didn't make things worse. 😅

delucis: Not present on the old site. With the current production version, the address bar never hides so I guess there’s never this bug that `100vh` mismatches with the actual viewport height.

You don’t see this is iOS Safari, I guess?

hippotastic: No, it works fine there! I don't have any device to reproduce this with. 😬

delucis: Haha! Now you know how it feels 🤪 

Let me try using `100dvh` on supported browsers locally and see if that fixes it (https://caniuse.com/viewport-unit-variants).

hippotastic: My var already uses dvh if supported 😅

delucis: Doh… just wrote out an identical `@supports` block to what you had 😆 

Maybe we want `lvh` or `svh` then? 🤔

delucis: Nope same behaviour with both.

delucis: OK. So not sure what the problem is here. Let’s leave it as BFMPTWWHB (Better For Most People Than What We Had Before). If we figure out a fix in the future, that’s great, but FF Mobile has a market share of something like 0.4% and it’s a harmless rendering glitch so should be OK.

delucis (approved):
LGTM! Thanks so much Hippo! A great improvement 🚀

PR #951, opened 2022-07-08
Use fixed header navigation on all devices

hippotastic:

#### Description

- Instead of having our header navigation disappear when scrolling on wide viewports, this PR changes all of our headers to a fixed position, effectively unifying our header behavior for all devices & viewport widths.
- As discussed on Discord, this allows the Search bar to be always easily accessible without having to scroll up.
- Unifying the header positioning also allowed to clean up our CSS code quite a bit, removing the now unnecessary grid layout and improving the browser's ability to keep the current scroll position when resizing the viewport. It's still possible for the content to jump a bit when any of the sidebars disappear, but it's a lot better than before.

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -69,7 +69,6 @@ body {
 	font-family: var(--font-body);
 	font-size: 16px;
 	line-height: 1.5;
-	max-width: 100vw;
 }
 
 nav ul {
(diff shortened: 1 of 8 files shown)


hippotastic:
Summoning @delucis, @Yan-Thomas and @sarah11918 's evil phone for a quick review if everything is working well. :)

sarah11918:
LGTMP!

Yan-Thomas:
Nice work @hippotastic, looks great in my devices! Just a thing: we will have to change the "scroll to current sidebar item" script a bit now that the header is always visible. I can fix this quickly, mind if I commit this change in your PR?

hippotastic:
> Nice work @hippotastic, looks great in my devices! Just a thing: we will have to change the "scroll to current sidebar item" script a bit now that the header is always visible. I can fix this quickly, mind if I commit this change in your PR? 

Not sure what script you're referring to - at least I didn't know of this functionality, or otherwise I would have fixed it myself! 😄 You're welcome to fix this and commit to my PR. Thank you! 💜

delucis (approved):
LGTM!

PR #953, opened 2022-07-08
Replace references to vite/client with astro/client

matthewp:

- New or updated content

#### Description

In https://github.com/withastro/astro/pull/3851 we are replacing `vite/client` types with our own `astro/client`. This is so that we can make changes to the types you see in frontmatter.

This updates our docs to reflect the new location.


delucis:
Thanks @matthewp! Could you remove the changes from non-`/en/` pages? We [usually leave those to our translators](https://github.com/withastro/docs/blob/main/CONTRIBUTING.md#making-prs-pull-requests) so they can make sure all the surrounding context gets updated correctly (some of the updated code samples for examples still mention “Vite” in their accompanying comments for example).

matthewp:
@delucis If you'd like, but the changes are almost definitely correct and not making the changes means that they are for sure wrong. Would it be better to merge this and if for some reason there's a problem fix that than to leave in docs that we know are wrong?

delucis (approved):
Thanks @matthewp! LGTM 🙌

PR #965, opened 2022-07-10
Add type assertion to Astro.props

wilkomanger:

#### Description

This type assertion was missing from the "TypeScript" page, while it's present
at other places. The assertion is necessary to have proper type info in the component itself.


Yan-Thomas:
Thanks so much for this @wilkomanger, nice catch! I think you aren't in our Discord yet, we would love to have you there https://astro.build/chat 🙌 Welcome to the Docs FacePile! 🥳

PR #954, opened 2022-07-08
Add official integration READMEs to docs

delucis:
#### What kind of changes does this PR include?

- New or updated content
- Changes to the docs site code

#### Description

- Closes #945  
- Add a script to fetch the official integration READMEs and generate a page for each one
- Add a CI job to run the script nightly and open a PR with changes
- Add links to the new pages from the existing integrations guide

##### To do

- [x] Think about what we want these pages to look like.
  - Are we really fans of `<details>`?
  - What’s missing on these auto-generated pages?
  - Is there anything we want to remove?
  - Do we want to make changes to the actual READMEs now we’re seeing them in this context?
- [x] Make the integrations navigation component pretty à la deploy guide
- [x] Make an integration page layout to include the nav component in the footer like for deploy guides
- [x] Add a remark plugin to map video URLs to video embeds (see https://github.com/withastro/docs/pull/954#issuecomment-1179374828)
- [x] Fix absolute links when using the footnote/reference/whatever it's called format
- [x] Remove table of contents as we have that in “On this page”. ~~Do we need to tag these specially in the integration READMEs? A `hide-in-docs` class or something?~~ Done by assuming all of these start with a link to `#why...`
- [x] Add complete Changelog sections to the actual READMEs
- [x] Do a pass for weird formatting differences (Astro is fussier than GitHub at Markdown rendering)
- [x] Add `<details>` styling
- [x] Add package info to header (latest version, npm link, etc.)
- [x] Fix current sidebar item highlighting for integration pages

PR #966, opened 2022-07-11
Fix import in translation status script

delucis:

- Something else!

#### Description

In #954 I extracted a function from the translation status checker to a separate module so I could re-use it in the page generator. I messed up the updated import inside the translation status script, leaving out the `.mjs` extension which is required.

This PR fixes that.


hippotastic:
LGTM!

PR #968, opened 2022-07-11
Add margin for slotted content

Yan-Thomas:

#### Description

This PR adds a base `1.25rem` (we can discuss and change this!) flow margin for slotted content, improving the reading experience:

| docs.astro.build  |  Deploy Preview  |
| image |  image |

Notes:
- I'm under the assumption that we don't want a lot of slotted text for SEO reasons, so the margin is small because we are not going to slot too much text. And if we do want to slot a lot of text, this can be solved as a separate PR.
- Slotted content from Astro is inside a `<astro-slot>` element, I believe this isn't going to change, but are there better approaches? Also, other components can easily override this rule, so I guess conflicting CSS isn't a real problem here.


sarah11918:
This looks better to me, Yan! I'll look around the preview, to make sure it doesn't throw anything else off, but I really do like this change here!

delucis (approved):
LGTM! Thanks @Yan-Thomas

PR #980, opened 2022-07-13
Improve "edit this page" link on integrations and fallback pages

### LINKED ISSUE #979: Improve "Edit this page" link on integration pages
We show an "Edit this page" link in the right sidebar of our docs pages which currently links to the generating Markdown file:

https://github.com/withastro/docs/blob/81713b98f4261c0eb9245ffac5873b72c996a325/src/layouts/MainLayout.astro#L18

On our new integration pages, e.g. https://docs.astro.build/en/guides/integrations-guide/tailwind/, this points to the auto-generated files. We could improve this by pointing to the actual source README in withastro/astro. The `githubURL` field in these pages' frontmatter should be helpful!

(We should _not_ do this for translations because those won't be auto-generated.)

Yan-Thomas:

#### Description

- Closes #979 

The edit link now points to the package README file rather than the docs' copy of the file. This pattern for custom edit links should work for any future pages that includes a `githubUrl` frontmatter property.

EDIT: Also improved the fallback pages edit link, now it points to the English page instead of resulting in a 404.

Note: For translated pages, we still point to the file in our docs, since they are manually translated. Although the `IntegrationsNav` and `DeployGuidesNav` components don't support different languages yet, I have done manual tests to sure we are pointing to the right file.


src/layouts/MainLayout.astro
```diff
@@ -15,11 +15,11 @@ const isFallback = !!Astro.params.fallback;
 const url = new URL(Astro.request.url);
 const currentPage = url.pathname;
 const currentFile = `src/pages${currentPage.replace(/\/$/, '')}.md`;
-const githubEditUrl = `https://github.com/withastro/docs/blob/main/${currentFile}`;
 const t = useTranslations(Astro);
 const formatTitle = (content) => (content.title ? `${content.title} 🚀 ${t('site.title')}` : t('site.title'));
 const lang = getLanguageFromURL(url.pathname);
 const bcpLang = normalizeLangTag(lang);
+const githubEditUrl = content.githubURL && lang === 'en' ? `${content.githubURL}README.md` : `https://github.com/withastro/docs/blob/main/${currentFile}`;
```
delucis, line 22: Thanks for taking this on @Yan-Thomas!

What do you think about doing this? Currently the edit link on non-English pages that use the fallback content results in a 404 because the translated version doesn’t exist yet. This change would link to the source README also for fallback pages.

```suggestion
const githubEditUrl = content.githubURL && (lang === 'en' || isFallback)
  ? `${content.githubURL}README.md`
  : `https://github.com/withastro/docs/blob/main/${currentFile}`;
```

Yan-Thomas: Nice catch, thanks! 🙌

Yan-Thomas: I think the issue still persists, I'll take a look at this problem 🤔

Yan-Thomas: Now I think the issue is solved (I hope at least 😅)

delucis (approved):
Oh nice! I hadn’t even considered the other fallback pages here. LGTM!

PR #989, opened 2022-07-14
Add note about Netlify functions to deploy guide

### LINKED ISSUE #988: Add note about Netlify Functions
A Netlify user was confused how "Netlify Background Functions" worked in Astro

The answer was: "just follow the netlify docs" (aka use a `netlify` folder in your repo) but it wasn't clear that this would work natively in Astro. 

It would be nice to add a mention to our integration guide/readme about how the `netlify` folder works, and that you can just use it as expected!

delucis:

- New or updated content

#### Description

- Closes #988 

- Add a note regarding Netlify Functions to our deploy guide. #988 suggested adding this toe the adapter README, but the [original issue](https://discord.com/channels/830184174198718474/995986253784825967/995989611409068042) was raised by someone with a static build, so I thought this might be the best catch-all location. (Maybe we also need to confirm at some point if `netlify/functions` works when you’re using the adapter for SSR Astro itself generates functions. Shouldn’t block this though.)

### DIFF
--- a/src/pages/en/guides/deploy/netlify.md
+++ b/src/pages/en/guides/deploy/netlify.md
@@ -130,6 +130,10 @@ You can [specify your Node.js version in Netlify](https://docs.netlify.com/confi
 - a [`.nvmrc`](https://github.com/nvm-sh/nvm#nvmrc) file in your base directory.
 - a `NODE_VERSION` environment variable in your site's settings using the Netlify project dashboard.
 
+## Using Netlify Functions
+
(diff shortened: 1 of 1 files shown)


FredKSchott:
@delucis I just realized that Netlify has a blog post about this!

https://www.netlify.com/blog/deploy-an-astro-site-with-forms-serverless-functions-and-redirects/

I think this is all we'd need to link to in the integration readme. Actually, I bet each integration could have a "links" section at the bottom, that links out to relevant off-site tutorials and guides on how to use the integration!

FredKSchott (approved):
LGTM with 1 comment for how we could incorporate an official tutorial that Netlify wrote into this readme.

delucis:
Confirmed that Netlify’s guides are for static sites only at the moment so I’ve added them here rather than the adapter README.

PR #993, opened 2022-07-14
Document how to escape special characters in Markdown

### LINKED ISSUE #987: Document how to escape curly braces in Markdown
We currently don't document how to prevent a text like `{some content}` inside a Markdown file from being treated as a JS expression.

It ~~works~~ should work by replacing the opening brace with its HTML entity: `&#123;`

I think we should document this in our Markdown section.

delucis:

- New or updated content

#### Description

- Closes #987
- Documents how to escape special characters in Markdown (e.g. `{` and `<`) using HTML entities.
- Currently blocked by https://github.com/withastro/astro/issues/3916, so this build will fail, but once that bug is fixed, the content will look like:
   image


delucis:
**Update:** Still blocked in legacy Astro-flavored Markdown. We would need to update to use MDX to document this :-/

sarah11918:
Love the simple English changes!  🥳  LGTM!

PR #1001, opened 2022-07-14
Remove "imports on top" warning

natemoo-re:

- New or updated content

#### Description

- Removes tips/warnings that import statements must go on top of your `.astro` files!
- Replaces old debugging tip with new "unexpected export" tip
- Blocked by https://github.com/withastro/compiler/pull/463

### DIFF
--- a/src/pages/en/guides/troubleshooting.md
+++ b/src/pages/en/guides/troubleshooting.md
@@ -13,7 +13,7 @@ Here are some common error messages you might see in the terminal, what they mig
 
 ### Transform failed with "Unexpected `export`"
 
-A current limitation in Astro is that `export` statements (other than `getStaticPaths`) are not supported in `.astro` files.
+A current limitation in Astro is that `export` statements (other than `getStaticPaths()`) are not supported in `.astro` files.
 
 
 **Solution**: Move any `export` statements to a `.js` or `.ts` file.
--- a/src/pages/en/guides/troubleshooting.md
+++ b/src/pages/en/guides/troubleshooting.md
@@ -11,18 +11,6 @@ Astro provides several different tools to help you troubleshoot and debug your c
 
 Here are some common error messages you might see in the terminal, what they might mean, and what to do about them.
 
-### Transform failed with "Unexpected `export`"
-
-A current limitation in Astro is that `export` statements (other than `getStaticPaths()`) are not supported in `.astro` files.
-
-
-**Solution**: Move any `export` statements to a `.js` or `.ts` file.
-
-**Status**: Current limitation; a fix may be explored in the future.
-
-**Not sure that this is your problem?**  
-Check to see if anyone else has reported [this issue](https://github.com/withastro/astro/issues?q=is%3Aissue+is%3Aopen+Unexpected+export)!
-
 ### Cannot use import statement outside a module
 
 In Astro components, `<script>` tags are hoisted and loaded as [JS modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) by default. If you have included the [`is:inline` directive](/en/reference/directives-reference/#isinline) or any other attribute in your tag, this default behavior is removed.
--- a/src/pages/en/guides/troubleshooting.md
+++ b/src/pages/en/guides/troubleshooting.md
@@ -11,16 +11,17 @@ Astro provides several different tools to help you troubleshoot and debug your c
 
 Here are some common error messages you might see in the terminal, what they might mean, and what to do about them.
 
-### Transform failed with X error
+### Transform failed with "Unexpected `export`"
 
-This message often appears due to a current limitation in Astro requiring your import and export statements to be at the top of your `.astro` file.
+A current limitation in Astro is that `export` statements (other than `getStaticPaths`) are not supported in `.astro` files.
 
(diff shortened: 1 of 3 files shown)


hippotastic:
Does this change also affect the `layout` property in Markdown files? When translating https://docs.astro.build/en/core-concepts/layouts/#markdown-layouts I was wondering why it said this:

> Markdown files can use a special layout property *at the top of* the frontmatter to specify which `.astro` component to use as a page layout.

delucis:
> Does this change also affect the `layout` property in Markdown files?

I don’t think that was ever impacted by order in the frontmatter? Or if it was it isn’t anymore.

src/pages/en/guides/troubleshooting.md
```diff
@@ -11,16 +11,17 @@ Astro provides several different tools to help you troubleshoot and debug your c
 
 Here are some common error messages you might see in the terminal, what they might mean, and what to do about them.
 
-### Transform failed with X error
+### Transform failed with "Unexpected `export`"
 
-This message often appears due to a current limitation in Astro requiring your import and export statements to be at the top of your `.astro` file.
+A current limitation in Astro is that `export` statements (other than `getStaticPaths`) are not supported in `.astro` files.
```
delucis, line 16: Can’t we also `export interface Props`?

sarah11918:
> I don’t think that was ever impacted by order in the frontmatter? Or if it was, it hasn’t been like that for quite a while.

It absolutely affected me, not that long ago, tbh, fwiw!

delucis:
> It absolutely affected me, not that long ago, tbh, fwiw!

Oh, interesting! Just tried it and definitely doesn’t seem to be the case here: https://stackblitz.com/edit/github-cjkpls?file=src/pages/index.md 

Or was this with `import` inside `setup` in frontmatter maybe?

src/pages/en/guides/troubleshooting.md
```diff
@@ -11,16 +11,17 @@ Astro provides several different tools to help you troubleshoot and debug your c
 
 Here are some common error messages you might see in the terminal, what they might mean, and what to do about them.
 
-### Transform failed with X error
+### Transform failed with "Unexpected `export`"
```
sarah11918, line 14: Just checking, is this exactly how the error message is presented? Quotation marks and code ticks? I think that's what we had before, so that the message would read the way it would be seen by the user.

src/pages/en/guides/troubleshooting.md
sarah11918:
```suggestion
A current limitation in Astro is that `export` statements (other than `getStaticPaths()`) are not supported in `.astro` files.
```
I noticed that we are not *entirely* consistent in how we refer to `getStaticPaths()` in regular sentence text. I know I do typically include the parenthesis, but this is not 100% standard throughout the Docs.

Can we pick one way, whichever way, and whatever we use here, I'll make sure it's then reflected throughout the site?

src/pages/en/guides/troubleshooting.md
```diff
@@ -11,16 +11,17 @@ Astro provides several different tools to help you troubleshoot and debug your c
...
 
-This message often appears due to a current limitation in Astro requiring your import and export statements to be at the top of your `.astro` file.
+A current limitation in Astro is that `export` statements (other than `getStaticPaths()`) are not supported in `.astro` files.
```
delucis, line 16: Noticing that this still doesn’t mention `Prop`:
```suggestion
A current limitation in Astro is that `export` statements (other than `getStaticPaths()` and `Props`) are not supported in `.astro` files.
```

src/pages/en/guides/troubleshooting.md
```diff
@@ -11,17 +11,6 @@ Astro provides several different tools to help you troubleshoot and debug your c
...
-
-This message often appears due to a current limitation in Astro requiring your import and export statements to be at the top of your `.astro` file.
-
-**Solution**: Write your imports and exports at the top of your component script.
-
-**Status**: Current limitation; fix is being worked on.
-
-**Not sure that this is your problem?**  
-Check to see if anyone else has reported [this issue](https://github.com/withastro/astro/issues?q=is%3Aissue+is%3Aopen+Transform+failed+with+*+error)!
-
```
natemoo-re, line 24: Removed this since it's no longer relevant, nor is the `Unexpected export` error.

https://github.com/withastro/astro/pull/4003 will provide helpful hints in-context if a user does run into an error due to `export` hoisting

src/pages/en/guides/troubleshooting.md
natemoo-re: I removed this entirely. This limitations was actually removed in https://github.com/withastro/compiler/pull/463. I opened https://github.com/withastro/astro/pull/4003 to provide a better error hint in context for `export` hoisting issues, since `export` _can_ cause reference errors. We'll keep an eye out for this and can put that error message here if users have trouble.

src/pages/en/guides/troubleshooting.md
sarah11918: Aww... bye bye TOP NUMBER ONE ERROR.  🥲

sarah11918:
I think this LGTM, Nate!

PR #1005, opened 2022-07-15
Document use of Astro’s JSX types

### LINKED ISSUE #984: Document using / extending the Astro JSX definitions
This is very advanced usage, nonetheless we see some people do it. Notably, our official image integration use them to provide `img` props completions and type checking in the components 

Using them is done through a triple slash directive, this works both in Astro files and in TypeScript files

```astro
---
/// <reference types="astro/astro-jsx" />

// Access them directly:
export type Props = astroHTML.JSX.ImgHTMLAttributes;
---

<img ...>
```

Extending them is done through a .d.ts file redeclaring the namespace:

```typescript
declare namespace astroHTML.JSX {
  interface HTMLAttributes {
    mycustomattribute?: any;
  }
}
```

People tend to extend them whenever they want non-standard attributes while still having typechecking. Another case is when we made a mistake and some attributes isn't included in our definitions 😅

delucis:

- New or updated content

#### Description

- Closes #984
- Adds a “Built-in attribute types” subheading to the Typescript page’s “Component Props” section that explains how to use and also extend Astro’s built-in JSX definitions.

@Princesseuh I tried these examples out in the docs repo and it seems to work without the need for the `///` directive. Would you mind reading through this and letting me know if you think this covers what you imagined? 💜

### DIFF
--- a/src/pages/en/guides/typescript.md
+++ b/src/pages/en/guides/typescript.md
@@ -78,9 +78,36 @@ export interface Props {
   name: string;
   greeting?: string;
 }
-const { greeting = 'Hello', name } = Astro.props as Props
+const { greeting = 'Hello', name } = Astro.props as Props;
 ---
 <h2>{greeting}, {name}!</h2>
+```
+
+### Built-in attribute types
+
+Astro provides JSX type definitions to check your markup is using valid HTML attributes. You can use these types to help build component props. For example, if you were building a `<Link>` component, you could do the following to mirror the default HTML attributes in your component’s prop types.
+
+```astro
+---
+export type Props = astroHTML.JSX.HTMLAttributes<HTMLAnchorElement>;
+const { href, ...attrs } = Astro.props as Props;
+---
+<a {href} {...attrs}>
+  <slot />
+</a>
+```
+
+It is also possible to extend the default JSX definitions to add non-standard attributes by redeclaring the `astroHTML.JSX` namespace in a `.d.ts` file.
+
+```ts
+// src/custom-attributes.d.ts
+
+declare namespace astroHTML.JSX {
+  interface HTMLAttributes {
+    'data-count'?: number;
+    'data-label'?: string;
+  }
+}
 ```
 
 ## Type checking
--- a/src/pages/en/guides/typescript.md
+++ b/src/pages/en/guides/typescript.md
@@ -110,6 +110,16 @@ declare namespace astroHTML.JSX {
 }
 ```
 
+:::note
+`astroHTML` is injected globally inside `.astro` components. To use it in TypeScript files, use a [triple-slash directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html):
+
+```ts
+/// <reference types="astro/astro-jsx" />
+
+type MyAttributes = astroHTML.JSX.ImgHTMLAttributes;
+```
+:::
+
 ## Type checking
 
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/typescript.md
```diff
@@ -78,11 +78,38 @@ export interface Props {
...
 <h2>{greeting}, {name}!</h2>
 ```
 
+### Built-in attribute types
+
+Astro provides JSX type definitions to check your markup is using valid HTML attributes. You can use these types to help build component props. For example, if you were building a `<Link>` component, you could do the following to mirror the default HTML attributes in your component’s prop types.
+
+```astro
+---
+export type Props = astroHTML.JSX.HTMLAttributes<HTMLAnchorElement>;
```
Princesseuh, line 92: This is using the old `astro-jsx.d.ts` format, the file has significantly changed about a week ago and now has element specific attributes interfaces and doesn't require any generics anymore. For a Link component, you'd now do:

```suggestion
export type Props = astroHTML.JSX.AnchorHTMLAttributes;
```

src/pages/en/guides/typescript.md
```diff
@@ -78,11 +78,38 @@ export interface Props {
...
+
+It is also possible to extend the default JSX definitions to add non-standard attributes by redeclaring the `astroHTML.JSX` namespace in a `.d.ts` file.
+
+```ts
+// src/custom-attributes.d.ts
+
+declare namespace astroHTML.JSX {
+  interface HTMLAttributes {
+    'data-count'?: number;
+    'data-label'?: string;
```
Princesseuh, line 108: This is fine to add type checking, but `data-` attributes don't need to be added to `HTMLAttributes` to avoid errors as attributes with a dash in them are always valid (similarly to how elements with a dash in them are always valid)

image

src/pages/en/guides/typescript.md
delucis: Ah right! I was testing it out in docs and we’re still on beta.64. Thanks for the review.

Princesseuh:
> @Princesseuh I tried these examples out in the docs repo and it seems to work without the need for the /// directive.

This is true in Astro files only because the language server inject those types into Astro files in the editor, however it's technically improper code 😄 Additionally, it is needed in other files, so definitely needs to be documented. [Our own image integration use it in a `.ts` file, for instance](https://github.com/withastro/astro/blob/d2f68345f97eb5b55bc323d633017069398727b2/packages/integrations/image/src/types.ts)

src/pages/en/guides/typescript.md
Princesseuh: Hmm, not necessarily. The specific use case I'm thinking about was [this user who needed a new global attribute that we didn't have yet](https://github.com/withastro/astro/pull/3801#issuecomment-1172889426). I'd think it's more common in general to add a new global attribute than an element specific one, not sure though

delucis: TIL `inert`!

OK, I’ll leave this example as is. I know `data-` attributes were already supported, so this just adds types, but I thought that seemed like a vaguely useful example (maybe you built a client-side script that uses specific `data-` attributes or something) instead of a totally abstract one.

Thank you!

delucis:
@Princesseuh Added a note to that effect: 
image

Princesseuh (approved):

PR #1011, opened 2022-07-15
Expanding Components with Renaming Properties

### LINKED ISSUE #995: Document how to apply a `class` to an Astro component
I've already seen multiple support issues where new Astro developers are struggling with getting syntax like this to work:

```html
<SomeAstroComponent class="some-css-class">
    ...
</SomeAstroComponent>
```

The challenge is how to get `class` to be rendered in the HTML output.

This is a typical use case and it's unfortunately also a bit special due to `class` being seen as a reserved name in JS.

This does not work inside the component frontmatter:
```js
const { class } = Astro.props
```

You need to rename the property to get it to work, and then use the renamed property in the root element of the component template:

```js
const { class: className } = Astro.props
```

We should probably document this in an example.

lostra01:

- New or updated content

#### Description

- Closes #995 

Added a complete `Renaming Properties` to the components page with an example for renaming reserved `class` which also solves issue#995.

Tagging @sarah11918 and @delucis

### DIFF
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -137,6 +137,27 @@ const name = "Astro";
 <MyComponent templateLiteralNameAttribute={`MyNameIs${name}`} />
 ```
 
+#### Renaming Properties
+
+Some names, such as `class`, are reserved in JavaScript. To pass such parameters into a component, you can rename them in your frontmatter with the `parameter: newName` sytnax:
+
+`MyComponent.astro`
+```astro
+---
+const {class: className} = Astro.props;
+---
+<h1 class={className}>A renamed property is supplying my class</h1>
+```
+`index.astro`
+```astro
+---
+import MyComponent from './MyComponent.astro';
+---
+<MyComponent class="Astro" />
+```
+
+In astro, the class attribute can also take a special expression which is turned into a class string. You can read more about it in our [common directives](https://docs.astro.build/en/reference/directives-reference/#classlist) section.
+
 #### Dynamic HTML
 
 Local variables can be used in JSX-like functions to produce dynamically-generated HTML elements:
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/astro-components.md
```diff
@@ -137,6 +137,27 @@ const name = "Astro";
 <MyComponent templateLiteralNameAttribute={`MyNameIs${name}`} />
 ```
 
+#### Renaming Properties
+
+Some names, such as `class`, are reserved in JavaScript. To pass such parameters into a component, you can rename them in your frontmatter with the `parameter: newName` sytnax:
```
Princesseuh, line 142:
```suggestion
Some names, such as `class`, are reserved in JavaScript. To pass such parameters into a component, you can rename them in your frontmatter with the `parameter: newName` syntax:
```

sarah11918:
To clarify, is the goal behind this content to document *renaming properties when you import them* or to address people's specific difficulty with properly styling components with class names? The issue this is linked to is about applying a class to an Astro component.

I think if I were on this page and having trouble getting my styles recognized, I probably would not look for "renaming variables" as my solution. I would probably look under the heading for CSS styles, which right now only documents the style tag, then links to our styling guide page.

And, if I went immediately to the styling guide page, I'd never think to look for this information here. So, I just want to be sure I know what problem we're trying to solve, to know exactly where this info will be most useful (and visible)! 

That might mean putting it in the style guide, and making a section there on "Styling imported components in your template" (worded less badly than that). Because that's really what this example is addressing: not stying the component you're *in*, but rather styling the component you're *rendering inside* this component, right?

mayank99:
One of my most used ways of propagating props is like this:
```astro
---
export type Props = { [k: string]: unknown };
---
<div {...Astro.props} />
```

Works really well for small components and when the class names don't need to be merged.

Not sure if it's worth documenting but thought I'd mention it, since this PR currently talks about more than just class names. But if it gets reworked into a styling section then maybe it won't fit.

PR #1018, opened 2022-07-16
Minor wording change: TypeScript page

sarah11918:
I didn't catch this in time for its first merge. After sitting on it, I really would like the word "that" in there!


sarah11918:
FWIW, I have an LGTM from @delucis on this.

Issue #1021, opened 2022-07-17
Add system to provide captions for videos


delucis:
Any video in docs with an audio track [should include captions](https://www.a11yproject.com/checklist/#confirm-the-presence-of-captions) for people who can’t hear the audio. We currently have one example of this on [the Tailwind integration page](https://docs.astro.build/en/guides/integrations-guide/tailwind/#installation).

#### Considerations

We need to think about:

- Translating captions so they are available in multiple languages
- Handling video from GitHub (like the Tailwind example) as well as video files stored in the docs repo

#### Suggested approach

1. Store caption files in `/public/captions/[lang]/[video-filename].vtt`

2. Create a `<Video>` component that takes a video `src` and looks up matching filenames in `/public/captions/**/*.vtt`.

    - It should create a `<track>` for each matching caption file.

    - If it finds a caption file matching the current page’s language, that `<track>` should be marked with the `default` attribute.

    - If not, the `en` file (if present) should be marked as `default`.

    - `label` attributes can be looked up from `src/i18n/languages.ts`.

    Example syntax:
    
    ```astro
    <video controls>
       <source src="video-filename.mp4" type="video/mp4">
       <track label="English" kind="subtitles" srclang="en" src="/captions/en/video-filename.vtt" default>
       <track label="Deutsch" kind="subtitles" srclang="de" src="/captions/de/video-filename.vtt">
    </video>
    ```

3. In our integrations page generator instead of generating HTML markup for GitHub video URLs, we should use the `<Video>` component instead. That would allow us to keep those GitHub-hosted videos but caption them on the docs site.

#### Reference

Captions can be loaded from [WebVTT](https://en.wikipedia.org/wiki/WebVTT) files in a [`<track>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track) inside a `<video>` element.

Here’s a `.vtt` captions transcription for our existing video:

```vtt
WEBVTT

00:00.570 --> 00:06.520
Hey everyone, I’m Dan, and today I’m going to show you how to use Tailwind CSS with your Astro project.

00:06.815 --> 00:12.415
You can see I’ve got a basic Astro site here and it looks pretty boring, so I want to add some styles to it.

00:12.500 --> 00:15.500
I’ll click this button to open up a new terminal

00:15.511 --> 00:19.871
where I can run the command “npx astro add tailwind”.

00:20.107 --> 00:25.500
This command can be found in the integration docs which I will be sure to link.

00:26.327 --> 00:34.383
[...]

PR #1022, opened 2022-07-17
`<Video>` component with captions support

Yan-Thomas:

#### Description
- Related to #1021 

This PR adds a new `<Video>` component with captions support. It gets captions from `public/captions/[lang]/[video-file].vtt` and includes the corresponding captions in the video, defaulting to the current page captions (in case there aren't captions for the current page's language, it defaults to English).

I haven't added the docgen changes to the Markdown in the PR to avoid possible conflicts, check out this branch locally and run `npm run docgen:integrations` in your terminal to better see the changes. Here are a few gifs:

image

This second gif shows how the English caption is marked as default when you don't have captions for your language:

image


src/components/Video.astro
```diff
@@ -0,0 +1,32 @@
...
+
+export interface Props {
+	src: string;
+	type: string;
+}
+
+const { src, type } = Astro.props as Props;
+const url = new URL(Astro.request.url);
+const lang = getLanguageFromURL(url.pathname);
+const filename = src.substring(src.lastIndexOf('/') + 1, src.lastIndexOf('.'));
```
delucis, line 13: What do you think of using [Node’s path parser](https://nodejs.org/dist/latest-v14.x/docs/api/path.html#path_path_parse_path) here? Should work on URLs too in my testing.

```suggestion
const filename = parse(src).name;
```

src/components/Video.astro
```diff
@@ -0,0 +1,32 @@
+---
```
delucis, line 1: If you agree with using `path.parse`, we’ll also need to import it:

```suggestion
---
import { parse } from 'node:path';
```

src/components/Video.astro
```diff
@@ -0,0 +1,32 @@
...
+const fileCaptions = Object.keys(allCaptions).map((caption) => {
+	const captionPath = caption.substring(caption.indexOf('captions') - 1)
+	const captionLang = captionPath.substring(10, captionPath.indexOf(filename) - 1);
+	return {
+		path: captionPath,
+		lang: captionLang
+	}
+}).filter(({path}) => path.includes(filename))
+
+const defaultLang = fileCaptions.find(({lang: cLang}) => cLang == lang) ? lang : 'en';
```
delucis, line 25: This is totally subjective, but in these scenarios where you have two variables with the same name (`lang`), I usually choose to avoid the destructured rename and access one as an object property instead.

```suggestion
const defaultLang = fileCaptions.find((caption) => caption.lang == lang) ? lang : 'en';
```

src/components/Video.astro
```diff
@@ -0,0 +1,32 @@
...
+
+const allCaptions = import.meta.glob("../../public/captions/**/*.vtt", { as:'raw' });
+const fileCaptions = Object.keys(allCaptions).map((caption) => {
+	const captionPath = caption.substring(caption.indexOf('captions') - 1)
+	const captionLang = captionPath.substring(10, captionPath.indexOf(filename) - 1);
+	return {
+		path: captionPath,
+		lang: captionLang
+	}
+}).filter(({path}) => path.includes(filename))
```
delucis, line 23: Still in the realm of slightly subjective, but it took me a minute to read through this and understand, so I’ve tried pulling the directory into a named variable so we can use its length directly. I also renamed the other variables to simplify the returned object. What do you think?

```suggestion
const fileCaptions = Object.keys(allCaptions)
	.map((caption) => {
		const dir = '/captions/';
		const path = caption.substring(caption.indexOf(dir));
		const lang = path.substring(dir.length, path.indexOf(filename) - 1);
		return { path, lang };
	})
	.filter(({ path }) => path.includes(filename));
```

delucis (commented):
Looks great @Yan-Thomas! Thanks so much for working on this — it’s a really exciting addition and so in line with your a11y work 💜 

I took the liberty of running the docgen script and pushing the result up to this branch so we can see it working in all its glory.

src/components/Video.astro
Yan-Thomas: That's so nice! Didn't know Node had this method, thanks so much! 💜

delucis (approved):
Alright! As soon as checks pass, this LGTM! Thanks again @Yan-Thomas 🚀

PR #1023, opened 2022-07-18
Usage note for VFile.data

naiyerasif:

- New or updated content

#### Description

- Adds a note on how to access `VFile.data` through Astro with an example


sarah11918:
Hi @naiyerasif ! I can see that this depends on https://github.com/withastro/astro/pull/3952 so I'm going to mark this as a "Draft" here. This is our normal process so that docs never get merged before features do!  :)

PR #1028, opened 2022-07-18
Add react to optional peer deps

bholmesdev:

- New or updated content
- Translated content
- [X] Something else!

#### Description

Adds `react`, `react-dom`, and `@types/react` to optional peer deps. This should fix install errors after the react -> preact compat refactor!

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,5 +71,16 @@
     "node": "^14.15.0 || >=16.0.0",
     "pnpm": ">=7.0.0"
   },
-  "packageManager": "pnpm@7.0.0"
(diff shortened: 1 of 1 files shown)


delucis (commented):
I haven’t used `peerDependenciesMeta` before but wouldn’t the package that lists `react` etc. as `peerDependencies` need this config? That’s what [npm’s documentation example](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta) looks like. Don’t mind merging if you want to try it though as it seems harmless.

PR #1031, opened 2022-07-18
Set strict-peer-dependencies=false to fix `react` dep errors

bholmesdev:

- New or updated content
- Translated content
- [X] Something else!

#### Description

Adds `strict-peer-dependencies=false` to our `.npmrc`. This prevents install failures for `react` and `react-dom`. Note: we tried to explicitly silence failures for React in this PR #1028, but it seems this wont affect nested package peer dependencies 😢

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 shamefully-hoist=true
+strict-peer-dependencies=false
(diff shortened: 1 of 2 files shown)


delucis (approved):
🤞 take two!

PR #1032, opened 2022-07-18
fix: Reformat code samples on UI frameworks page

### LINKED ISSUE #1025: Reformat code samples on UI Frameworks page
On this page - https://docs.astro.build/en/core-concepts/framework-components/#passing-children-to-framework-components

The code samples don't conform to our standard syntax, so the Astro code highlighting isn't taking effect.

Please reformat these samples so that the file name is located **inside, and directly following the first `---`*** (to match the examples e.g. on https://docs.astro.build/en/core-concepts/astro-components/#slots) instead of written before the code fence.

Happy to receive a PR for this from any of our community members! 🚀

kevinzunigacuellar:

#### Description

- Closes #1025
- Reformat codeblocks in `core-concepts/framework-components.md` for every available language

### DIFF
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -125,8 +125,8 @@ Inside of an Astro component, you **can** pass children to framework components.
 
 
 ```astro
-// src/pages/MyAstroPage.astro
 ---
+// src/pages/MyAstroPage.astro
 import MyReactSidebar from '../components/MyReactSidebar.jsx';
 ---
 <MyReactSidebar>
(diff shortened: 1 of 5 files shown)


sarah11918:
Thanks @kevinzunigacuellar ! As you know, normally we wouldn't touch all the language files at once, but since this doesn't even change any text, and since it's YOU, I think we're safe. Just shhhh... don't let everyone else know I let you do this!  ;)

delucis:
Just a note to say we should have included `minor` in this commit message as this marked all all of these languages as up-to-date — maybe they already were so it’s not an issue, but wanted to flag as a general reminder to try to include that for this kind of thing (until Hippo’s translation manager becomes fully sentient and can work it out for itself 😅).

PR #1034, opened 2022-07-18
chore: REVERT npmrc change

bholmesdev:

- New or updated content
- Translated content
- [X] Something else!

#### Description

Turns out, adding a flag to ignore peer deps did _not_ resolve our CI issues we set out to fix. It also sounds like devs don't need this change when dev-ing locally. So... let's revert!

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
 shamefully-hoist=true
-strict-peer-dependencies=false


delucis (approved):
Pleasure doing business with you Mr. Holmes.

PR #1041, opened 2022-07-19
Navigation accessibility improvements

Yan-Thomas:

#### Description

I did an accessibility audit on The Docs and I found a few issues related to the navigation. Here's a list of what I changed, why I changed, and a few info sources I used:

### Removed the title attribute from the header.
The title attribute isn't meant to be used for labeling landmarks, also, there are a few accessibility concerns regarding `title`:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns
https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/

<hr/>

### Removed h1 tags from the header and added them as part of a `<ul>`.
The header is a fixed part of the page, and the header has a proper landmark (easily accessible using shortcuts in screen readers) that makes the use of `<h1>` tags unnecessary, also it is a good practice to avoid having more than one `<h1>`.

The header links weren't part of a list, so a screen reader user couldn't know how many links there was.

<hr/>

### Reworked the navigation landmarks' labels.
When you have more than one navigation/complementary landmark you have to specify them with labels, what happens is that our current labels had a few problems:
1) They contained 'navigation', screen readers already say 'navigation', so, when you have navigation as part of the label, the result is: "top navigation navigation landmark"
2) They weren't clear.

**"Site Navigation" to "Primary":** It's the docs' most important links, after all, people often open a lot of docs pages and are mainly navigating using the left sidebar. Also, all the navigations on the site could be considered a "site navigation".

**"Table of Contents" to "Secondary":** "Table of Contents" doesn't make sense when the content of the landmark is not only the Table of Contents but also the "Contribute" and "Community" menus. 

**"Top Navigation" to "Top":**: No reason to say "navigation" two times.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role
https://www.smashingmagazine.com/2011/06/planning-and-implementing-website-navigation/
https://www.scottohara.me/blog/2018/03/03/landmarks.html

<hr/>

### Added missing focus style in the `LeftSidebar`'s current page.
All focusable elements should have a clear focus style:

| Deploy Preview  |  docs.astro.build  |
| image |  image |

<hr/>

### Other Known Issues
Our old Tabs component has a few a11y problems, the controls don't behave as expected for keyboard/screen reader users and we use it in an inaccessible way to navigate between pages in the Installation pages. I will refactor the old Tabs component into a list of links and use the new, accessible, and shiny Tabs component for the `LeftSidebar` in a future PR after we become sure about what we really want to do with the `LeftSidebar`'s tabs.


delucis (approved):
Love it! Ship it! Brilliant!

Thanks @Yan-Thomas — always doing so much great work to keep our a11y in check 💜

PR #1045, opened 2022-07-20
Remove references to the <Markdown> component

matthewp:

- New or updated content

#### Description

- https://github.com/withastro/astro/pull/3986 removes the Markdown component from core, moves it to its own package.
- This removes documentation on the component. Docs are now in the package readme. We don't want to encourage people to use it anymore.


hippotastic:
The more I think about the removal of the section `### <Markdown />` causing hard to resolve link issues in translated versions, the more I like your idea of leaving that section in there for now, @matthewp, but instead of hiding it, replacing the warning content with something along the lines of "This component is no longer built into Astro".

As soon as the translated pages that were previously referring to this component have been updated by the translators, we can then remove the section entirely in a second step.

delucis (commented):
Love this fix and the green CI!

One note: we’ve just added an RC section in the migration guide, so we should probably also include a short note there saying something like

```md
### `<Markdown />` Component Removed

If you were previously using Astro’s built-in `<Markdown />` component, this has been moved to a separate package. You will now need to install `@astrojs/markdown` and update your imports accordingly. For more details, see [the `@astrojs/markdown` README](...readme-link).
```

sarah11918:
@matthewp I just updated the branch , so now this branch should have the latest migration guide with the RC changes section Chris is talking about. 

There are already some items in there, so something like @delucis 's suggestion can be added to match the style and then... did we decide this is good to go now??  (LGTM re: proofreading)

matthewp:
Updated with the suggested Migration note.

src/pages/en/migrate.md
```diff
@@ -42,6 +42,10 @@ Astro no longer supports components or JSX expressions in Markdown pages by defa
 
 To make migrating easier, a new [legacy flag](/en/reference/configuration-reference/#legacyastroflavoredmarkdown) can be used to re-enable previous Markdown features.
 
+### `<Markdown />` Component Removed
+
+If you were previously using Astro’s built-in `<Markdown />` component, this has been moved to a separate package. You will now need to install `@astrojs/markdown` and update your imports accordingly. For more details, see [the `@astrojs/markdown` README](https://github.com/withastro/astro/tree/main/packages/markdown/component).
```
sarah11918, line 47: Thanks @matthewp ! 

@delucis , I'm assuming that this `@astrojs/markdown` component does NOT get automatically added to our Integrations page?

- Do we *want* a legacy package on our integrations page, so that we don't *need* to send people externally to the README but can instead have a page in Docs? Or, do we specifically NOT want this showing up in Docs, because we don't want people using it unless they really have to?

- Do we *want* to take this opportunity here in the migration guide to encourage people to transition to MDX, or simply say, "Here's where the new package is?"

matthewp: I think we don't want people using it unless they really have to.

Agree that it's a good opportunity to push people towards MDX, I'll add that.

src/pages/en/migrate.md
```diff
@@ -42,6 +42,12 @@ Astro no longer supports components or JSX expressions in Markdown pages by defa
 
 To make migrating easier, a new [legacy flag](/en/reference/configuration-reference/#legacyastroflavoredmarkdown) can be used to re-enable previous Markdown features.
 
+### `<Markdown />` Component Removed
+
+If you were previously using Astro's built-in `<Markdown />` component, this has been moved to a separate package. You will now need to install `@astrojs/markdown` and update your imports accordingly. For more details, see [the `@astrojs/markdown` README](https://github.com/withastro/astro/tree/main/packages/markdown/component).
+
+Astro now has support for [MDX](https://mdxjs.com/) through our [MDX integration](https://github.com/withastro/astro/tree/main/packages/integrations/mdx). MDX gives you the ability to create components that can be imported into an Astro page that include both Markdown and component usage. MDX make a good alternative for the Markdown component due to it's large community and stable APIs.
```
sarah11918, line 49:
```suggestion
:::tip
Astro now has support for [MDX](https://mdxjs.com/) through our [MDX integration](https://github.com/withastro/astro/tree/main/packages/integrations/mdx). MDX gives you the ability to include both Markdown and imported components in the same file. MDX can be good alternative for the `<Markdown />` component due to its large community and stable APIs.
:::
```

Wondering whether this might be a good "aside"?

I did also condense this a little for readablility, which admittedly loses a bit of the nuance but might be good enough for a note in the migration guide?

src/pages/en/migrate.md
```diff
@@ -42,6 +42,12 @@ Astro no longer supports components or JSX expressions in Markdown pages by defa
 
 To make migrating easier, a new [legacy flag](/en/reference/configuration-reference/#legacyastroflavoredmarkdown) can be used to re-enable previous Markdown features.
 
+### `<Markdown />` Component Removed
+
+If you were previously using Astro's built-in `<Markdown />` component, this has been moved to a separate package. You will now need to install `@astrojs/markdown` and update your imports accordingly. For more details, see [the `@astrojs/markdown` README](https://github.com/withastro/astro/tree/main/packages/markdown/component).
```
sarah11918, line 47:
```suggestion
Astro's built-in `<Markdown />` component has been moved to a separate package. To continue using this component, you will now need to install `@astrojs/markdown` and update your imports accordingly. For more details, see [the `@astrojs/markdown` README](https://github.com/withastro/astro/tree/main/packages/markdown/component).
```

Just a little more direct, if you like.

PR #1056, opened 2022-07-20
Add guides on using `<Badge />` and `<Since />`

### LINKED ISSUE #1053: Document Badge and Since components in WRITING.md
Document usage of Badge and Since components alongside our Custom Aside component in the Writing Guide.

delucis:

- Something else!

#### Description

- Closes #1053
- Update WRITING.md with guidance on the `<Badge />` and `<Since />` components

### DIFF
--- a/WRITING.md
+++ b/WRITING.md
@@ -103,7 +103,7 @@ By default, the badge uses a muted colour scheme to blend in. It also has an acc
 
 As features are added to Astro, it can be helpful to document _when_ they were added. This allows users to easily see if the version of Astro they are running supports a specific feature as described in the docs.
 
-You can use the `<Since />` component to display this information in a standardised way. This component takes a single `v` prop, which indicates the Astro version the feature was added in.
+You can use the `<Since />` component to display this information in a standardized way. This component takes a single `v` prop, which indicates the version of Astro in which the feature was added.
 
 ```md
 ---
@@ -123,7 +123,7 @@ The advantages of using the component include:
 
 #### Examples
 
-The standard usage of this component is on its own line either immediately following the heading for the feature it is referring to, for example:
+The standard usage of this component is on its own line, immediately following the feature's heading, for example:
 
 ```md
 ## `Astro.clientAddress`
@@ -132,7 +132,7 @@ The standard usage of this component is on its own line either immediately follo
 Specifies the IP address of the request. This property is only available when building for SSR (server-side rendering) and should not be used for static sites.
 ```
 
- Or it can be used in a short block of information, for example:
+ Or, it can be used in a short block of information, for example:
 
 ```md
 ### `server.host`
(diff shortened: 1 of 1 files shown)


WRITING.md
```diff
@@ -71,6 +73,75 @@ Using `<blockquote>` for notes is deprecated.
...
+#### Examples
+
+- The [Editor Setup](https://docs.astro.build/en/editor-setup/#other-code-editors) page uses `<Badge />` to distinguish community and official editor support.
+- The [Deployment guides navigation component](https://docs.astro.build/en/guides/deploy/) uses `<Badge />` internally to label the deployment types each service supports.
+
+### Since Component
+
+As features are added to Astro, it can be helpful to document _when_ they were added. This allows users to easily see if the version of Astro they are running supports a specific feature as described in the docs.
+
+You can use the `<Since />` component to display this information in a standardised way. This component takes a single `v` prop, which indicates the Astro version the feature was added in.
```
sarah11918, line 106:
```suggestion
You can use the `<Since />` component to display this information in a standardized way. This component takes a single `v` prop, which indicates the version of Astro in which the feature was added.
```
Yeah, the "in which" is a judgment call to be correct vs common. I'd do this personally, but I won't object to ending the sentence with a proposition if you feel it sounds better.

BUT, I did explicitly choose "version of Astro" to get the "v is for version" connection more closely aligned.

WRITING.md
```diff
@@ -71,6 +73,75 @@ Using `<blockquote>` for notes is deprecated.
...
+This will render the text “**Added in:** v1.0.0”.
+
+The advantages of using the component include:
+
+- “Added in” is automatically translated on pages in other languages.
+- The passed version is checked against the current Astro version and a “NEW” badge will be added automatically as long as the version is relatively recent.
+
+#### Examples
+
+The standard usage of this component is on its own line either immediately following the heading for the feature it is referring to, for example:
```
sarah11918, line 126:
```suggestion
The standard usage of this component is on its own line, immediately following the feature's heading, for example:
```
Maybe we can get away with just this? (I took out the either to avoid an either / or spanning a whole whack of stuff. I think the "or" can still work on its own like this.)

WRITING.md
```diff
@@ -71,6 +73,75 @@ Using `<blockquote>` for notes is deprecated.
...
+
+```md
+## `Astro.clientAddress`
+<Since v="1.0.0-rc" />
+
+Specifies the IP address of the request. This property is only available when building for SSR (server-side rendering) and should not be used for static sites.
+```
+
+ Or it can be used in a short block of information, for example:
```
sarah11918, line 135:
```suggestion
 Or, it can be used in a short block of information, for example:
```

sarah11918 (approved):
A few suggestions to consider!

PR #1064, opened 2022-07-21
Fix invisible 404 page

### LINKED ISSUE #1050: 404 page content is hidden
The docs' 404 page has a CSS conflict that is hiding its content, as you can see in the screenshot below:
image

Yan-Thomas:

#### Description

- Closes #1050 

The 404 page content was being hidden because of CSS conflicts, this PR simplifies the SplashLayout code so that the 404 page is visible again. We can discuss if we want the 404 page content aligned to the middle or left, I just went with what was more similar to what we already had.

| docs.astro.build  |  Deploy Preview  |
| image |  image |

### DIFF
--- a/src/layouts/SplashLayout.astro
+++ b/src/layouts/SplashLayout.astro
@@ -16,26 +16,13 @@ const t = useTranslations(Astro);
 		<HeadCommon />
 		<title>{`${title} 🚀 ${t('site.title')}`}</title>
 		<style>
-			body {
-				width: 100%;
-				display: grid;
-				grid-template-rows: var(--theme-navbar-height) 1fr;
-			}
-			.layout {
-				display: grid;
-				grid-auto-flow: column;
-				grid-template-columns:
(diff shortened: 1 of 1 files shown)


sarah11918:
I'll let @Yan-Thomas and @delucis make the final call! This works!

If it's not *horribly* difficult, I think it would be great if we could show the sidebars too, even on the 404 page. (I know "on this page" and "contribute" might be problematic, so I get this may be non-trivial.) At least the left sidebar wouldn't FORCE people to go home.

Otherwise, it's already better TWWHB!

delucis:
Did we have a left sidebar on this page before? I suspect not given @Yan-Thomas is only changing CSS here. I’d say merge this for a quick fix and open an issue/PR to add the left sidebar too.

delucis (approved):

Yan-Thomas:
Thanks for the review @sarah11918 & @delucis ❤️! I'm going to work on the PR adding the `LeftSidebar` tomorrow 🙌

PR #1067, opened 2022-07-21
Remove translate link in non-`i18nReady`, fallback and translated pages

Yan-Thomas:

#### Description

This PR removes the "translate this page" link on the `<ContributeMenu>` component for translated, non-i18nReady, and fallback pages (if they aren't `i18nReady`).

| `i18nReady: false`  |  Fallback  |  Translated  |
| ------------------- | ------------------- | ------------------- |
|  image |  image | image |

### DIFF
--- a/src/components/RightSidebar/ContributeMenu.astro
+++ b/src/components/RightSidebar/ContributeMenu.astro
@@ -2,7 +2,12 @@
 import UIString from '../UIString.astro';
 import { getLanguageFromURL } from '../../util';
 
-const { i18nReady, editHref } = Astro.props;
+export interface Props {
+  i18nReady: boolean | undefined;
+  editHref: string;
+}
+
+const { i18nReady, editHref } = Astro.props as Props;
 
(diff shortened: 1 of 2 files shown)


src/components/RightSidebar/ContributeMenu.astro
```diff
@@ -1,7 +1,13 @@
 ---
 import UIString from '../UIString.astro';
+import { getLanguageFromURL } from '../../util';
 
-const { editHref } = Astro.props;
+const { i18nReady, editHref } = Astro.props;
```
delucis, line 5: While we’re here, do we want to add types for the props?

```suggestion
export interface Props {
  i18nReady: boolean | undefined;
  editHref: string;
}

const { i18nReady, editHref } = Astro.props as Props;
```

delucis (approved):
Great thinking on this one @Yan-Thomas! I’ve made a suggestion but this LGTM with or without it.

Yan-Thomas:
Thanks for the review @delucis 🙌!

PR #1078, opened 2022-07-21
[i18n README]: Adds extra warning re: do not translate

sarah11918:
- New or updated content

Starts to address issues of i18n guidance for translators.

- Adds more explicit warning re: when NOT translate
- Inclues link to CONTRIBUTING to see more about when things are/aren't marked for translation (TODO: which will include description of using phrases like "minor" and "typo" there, since it's really the English contributors who need to know about this).
- TODO: move this file, probably just to `src` like the other READMES until we decide on a better location for all the README guides?

### DIFF
--- a/src/i18n/README.md
+++ b/src/i18n/README.md
@@ -11,7 +11,8 @@ Translations all live in this GitHub repository. You can add and update them by 
 See our automated [Translation Status Overview issue](https://github.com/withastro/docs/issues/438) for a quick list of which pages are missing or need updating.
 
 
-> **Please do not translate any pages without first checking the status on our Tracker**! 
+> **Warning**
+> Please do not translate any pages without first checking their status in our [Overview Issue](https://github.com/withastro/docs/issues/438)! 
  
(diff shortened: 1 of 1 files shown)


src/i18n/README.md
```diff
@@ -10,6 +10,13 @@ Translations all live in this GitHub repository. You can add and update them by
 ### How can I find out what needs translating?
 See our automated [Translation Status Overview issue](https://github.com/withastro/docs/issues/438) for a quick list of which pages are missing or need updating.
 
+
+> **Please do not translate any pages without first checking the status on our Tracker**!
```
delucis, line 14: We say “Translation Status Overview issue” above and “Tracker” here — maybe it’s not obvious those are the same thing? I’d suggest repeating similar language and another link wouldn’t hurt.

We could also use [GitHub’s syntax for a warning block](https://github.com/github-community/community/discussions/16925) here:

```suggestion
> **Warning**
> Please do not translate any pages without first checking their status in our [Overview Issue](https://github.com/withastro/docs/issues/438)! 
```

delucis (commented):
Thanks Sarah! One suggestion but otherwise LGTM 🙌

PR #1083, opened 2022-07-22
add i18n labeler

kevinzunigacuellar:

- Something else!

#### Description

- Github action that labels i18n to all `.md` files in pages except for pages in `/en/`

**Note**: This is not a YOLO PR. This was tested in a copy of the docs repo. I couldn't get the action to include the i18n folder without failing but "not worse than what we had before" 😅

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,9 +4,4 @@
 - .github/workflows/**
 
 i18n:
-- any:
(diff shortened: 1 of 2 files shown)


delucis (approved):
Thanks for testing @kevinzunigacuellar! NWTWWHB (hopefully! 😅)

PR #1084, opened 2022-07-22
Move integrations version out of docgen

### LINKED ISSUE #1080: Move integration version number out of CI generation task
Currently we show a version number for each integration in its header, e.g. v0.1.3 here

Header of the @astrojs/image integration page including the version tag

This is added as a frontmatter variable in the pages by the auto-generation script:

https://github.com/withastro/docs/blob/eb3cc827dc856cd4e3660520d24cff74b1c199a8/scripts/generate-integration-pages.ts#L89

This has the side effect that the pages get updated when the version bumps **even if no other content has changed**. Once we move to translating these pages this will be a particularly annoying thing to need to go and update manually.

I’d suggest we move this data out of frontmatter and use a custom component that looks something like `<Version package="@astrojs/image" />` in the integrations layout instead:

https://github.com/withastro/docs/blob/eb3cc827dc856cd4e3660520d24cff74b1c199a8/src/layouts/IntegrationLayout.astro#L11

This component could then retrieve the latest version for us — either by having the integration page generator write version numbers to a data file (would still need updating each bump, but could be used across languages) or by reading straight from the npm registry like the `<Since />` component does.

Yan-Thomas:

#### Description

- Closes #1080 

This PR adds a `<Version>` component that fetches an integration (or any other desired package) latest version number.

**NOTE FOR FUTURE YAN:** Rerun docgen script again before merge to avoid conflicts


delucis (approved):
This LGTM — thanks for picking it up @Yan-Thomas!

Future Yan, I believe there’s a message waiting for you. 🧑‍🚀

PR #1087, opened 2022-07-22
[migration guide] - RSS has changed

### LINKED ISSUE #1030: Remove reference to the old rss() function
Since we now have `@astrojs/rss`, we no longer use the old method for generating rss feeds using `rss()` described here: https://docs.astro.build/en/reference/api-reference/#rss

Happy to receive a PR that does 2 things:

1. Remove this section entirely.
2. Add a note in the [1.0 beta section of the migration guilde](https://docs.astro.build/en/migrate/#astro-10-beta) saying that the process for generating RSS feeds has changed, and users should now consult the [RSS page](https://docs.astro.build/en/guides/rss/) in the docs site to update their projects.

sarah11918:
- New or updated content

Adds a note to the migration guide RC section that the process for generating RSS feeds has changed.

Closes #1030


Yan-Thomas:
LGTM. Thanks for this @sarah11918! 🙌

delucis:
I didn’t add this because Ben said this has been gone for quite a while! So not actually an RC change.

sarah11918:
It has, and it does... but there's nothing else there. Is it bad to fudge??  It should go *somewhere.*

delucis:
I’d say move it down into the beta section. If someone’s already on beta this would be misleading if they’re just looking for what to do to get on RC. And that way it’s there for anyone upgrading from <1 to beta/latest.

sarah11918:
@delucis What do you think? I also removed the stuff about "We don't expect any breaking changes, but if we do, we'll put them here" as well as outdated launch date.

PR #1096, opened 2022-07-22
Fixed grammatical errors

tejapaturu:

#### Description

- The verb "choose" should be in the to-infinitive form.
- The subjects "Hugo and Astro" is plural, hence the verb should be "offer".


sarah11918:
These page is scheduled for deletion.

PR #1100, opened 2022-07-22
Fixed grammar

tejapaturu:

#### Description

- The introductory phrase "however" is missing a comma.
- The use of noun phrase "most configuration" is missing a determiner.
- The introductory phrase "For example" is missing a comma after it.
- "hopefully for advanced users will enjoy" -> The preposition "for" is incorrectly used.
- "strategy" was misspelled.
- "with" is required to use the preposition correctly.
- "reference" doesn't require an "s".
- A comma is not needed when a dependent clause follows an independent clause.
- Fixed "separately" typo


src/pages/en/migrate.md
```diff
@@ -23,7 +23,7 @@ npm install astro@next--rc
 
 ### Updated: Vite 3
 
-Astro v1.0 RC has upgraded from Vite 2 to [Vite 3](https://vitejs.dev/). We've handled most of the upgrade for you inside of Astro, however some subtle Vite behaviors may still change between versions. Refer to the official [Vite Migration Guide](https://vitejs.dev/guide/migration.html#general-changes) if you run into trouble.
+Astro v1.0 RC has upgraded from Vite 2 to [Vite 3](https://vitejs.dev/). We've handled most of the upgrade for you inside of Astro, however, some subtle Vite behaviors may still change between versions. Refer to the official [Vite Migration Guide](https://vitejs.dev/guide/migration.html#general-changes) if you run into trouble.
```
sarah11918, line 26:
```suggestion
Astro v1.0 RC has upgraded from Vite 2 to [Vite 3](https://vitejs.dev/). We've handled most of the upgrade for you inside of Astro; however, some subtle Vite behaviors may still change between versions. Refer to the official [Vite Migration Guide](https://vitejs.dev/guide/migration.html#general-changes) if you run into trouble.
```
Thanks for spotting that there was a punctuation error here. The comma is needed, and additionally, either a semicolon or a period. I think either works, so I'm happy to let you choose the fix.

src/pages/en/migrate.md
```diff
@@ -51,7 +51,7 @@ RSS feeds should now be generated using the `@astrojs/rss` package, as described
 ## Migrate to v0.26
 ### New Configuration API
 
-Our Configuration API has been redesigned to solve a few glaring points of confusion that had built up over the last year. Most configuration has just been moved or renamed, which will hopefully be a quick update for most users. A few options have been refactored more heavily, and may require a few additional changes:
+Our Configuration API has been redesigned to solve a few glaring points of confusion that had built up over the last year. Most of the configuration has just been moved or renamed, which will hopefully be a quick update for most users. A few options have been refactored more heavily, and may require a few additional changes:
```
sarah11918, line 54:
```suggestion
Our Configuration API has been redesigned to solve a few glaring points of confusion that had built up over the last year. Most of the configuration options have just been moved or renamed, which will hopefully be a quick update for most users. A few options have been refactored more heavily, and may require a few additional changes:
```
I would probably choose to be a bit more explicit about mentioning the options here, if we're changing it anyway.

sarah11918 (commented):
Thanks again for spotting these. I've made a couple of suggestions for improvements!

PR #1105, opened 2022-07-23
Add new CSS specificity behavior

sarah11918:
- New or updated content

Per @natemoo-re's request, documents the new behaviour from https://github.com/withastro/rfcs/blob/main/proposals/0012-scoped-css-with-preserved-specificity.md regarding the change to specificity in scoped styles.

- Migration Guide: Describes the change (specificity no longer increased) , including warning to visually inspect when upgrading and correcting with a manual specificity increase if necessay

- CSS Styling page in Docs: Updates code sample description of  `<style>` tag CSS output, now showing the `where:()` CSS psuedo-class and adds a note about preserving authored specificity, thereby making it easy to work with other CSS files and libraries.


src/pages/en/guides/styling.md
```diff
@@ -26,15 +26,18 @@ Astro `<style>` CSS rules are automatically **scoped by default**. Scoped styles
...
-+  h1.astro-HHNQFKH6 { color: red; }
++  h1:where(.astro-HHNQFKH6) { color: red; }
 -  .text { color: blue; }
-+  .text.astro-HHNQFKH6 { color: blue; }
++  .text:where(.astro-HHNQFKH6) { color: blue; }
 </style>
 ```
 
-Scopes styles don't leak and won't impact the rest of your site. In Astro, it is okay to use low-specificity selectors like `h1 {}` or `p {}` because they will be compiled with scopes in the final output.
+Scopes styles don't leak and won't impact the rest of your site. In Astro, it is okay to use low-specificity selectors like `h1 {}` or `p {}` because they will be compiled with scopes in the final output.
```
Yan-Thomas, line 35:
```suggestion
Scoped styles don't leak and won't impact the rest of your site. In Astro, it is okay to use low-specificity selectors like `h1 {}` or `p {}` because they will be compiled with scopes in the final output.
```

Yan-Thomas (commented):
Nice work @sarah11918! I found a small typo, but after this, LGTM.

src/pages/en/guides/styling.md
sarah11918: You found a typo that's been there the whole time!  😬 💜

sarah11918:
Just note to us all that I just got this started for @natemoo-re who can run with this when he's back! Waiting for him to review!

src/pages/en/migrate.md
```diff
@@ -28,6 +28,29 @@ const canonicalURL = Astro.canonicalURL;
 const canonicalURL = new URL(Astro.url.pathname, Astro.site);
 ```
 
+## Changed: Scoped CSS specificity
```
sarah11918, line 31:
```suggestion
### Changed: Scoped CSS specificity
```

src/pages/en/migrate.md
```diff
@@ -28,6 +28,29 @@ const canonicalURL = Astro.canonicalURL;
...
+- div { color: red } // (0,0,1 specificity) 
++ div.astro-XXXXXX { color: red } // (0,1,1 specificity).
+</style>
+
+// After: appends :where(.astro-XXXXXX), maintaining the authored specificity. 
+<style>
+- div { color: red } // (0,0,1 specificity) 
++ div:where(.astro-XXXXXX) // (0,0,1 specificity).
+</style>
+```
```
delucis, line 47: I’ve tried to break up and rewrite this code sample to explain slightly more clearly exactly what is going on here — I found using a diff for input/output not super obvious in this case.

I’ve also updated the code samples to use the correct format for CSS comments and added a link to docs for the `:where` pseudo-class as its on the newer side of CSS syntax.

````suggestion
Technically, this is accomplished using [the `:where()` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) instead of using classes directly in Astro’s CSS output.

Let’s use the following style block in an Astro component as an example:

```astro
<style>
div { color: red; } /* 0-0-1 specificity */
</style>
```

Previously, Astro would transform this into the following CSS, which has a specificity of `0-1-1` — a higher specificity than the source CSS:

```css
div.astro-XXXXXX { color: red; } /* 0-1-1 specificity */
```

Now, Astro wraps the class selector with `:where()`, maintaining the authored specificity:

```css
div:where(.astro-XXXXXX) { color: red; } /* 0-0-1 specificity */
```
````

**Review by delucis — COMMENTED** [review]
Thanks so much for picking this up @sarah11918! Excellent work as usual. Made two suggestions for the migration guide, but otherwise LGTM!

File: src/pages/en/migrate.md [hunk 4]
```diff
@@ -28,6 +28,29 @@ const canonicalURL = Astro.canonicalURL;
 const canonicalURL = new URL(Astro.url.pathname, Astro.site);
 ```
 
+### Changed: Scoped CSS specificity
+
+Specificity will now be preserved in scoped CSS styles. This change will cause most scoped styles to _happen_ to take precedence over global styles. But, this behavior is longer explicitly guaranteed.
```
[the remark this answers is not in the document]
sarah11918: You read my mind! I was going to hunt this down for the next draft! Thanks!

PR #1106, opened 2022-07-23
Clarify linking between pages

### LINKED ISSUE #992: Mention that navigation uses the standard `a` element in Astro
**Why?** A user was unclear how to link to another page in Astro

- Usually because React/next.js make this confusing with `<Link>` element
- Could use a small callout that you can use `<a>` elements directly

   SvelteKit has this (https://kit.svelte.dev/docs/routing#pages):

   > Note that SvelteKit uses `<a>` elements to navigate between routes, rather than a framework-specific `<Link>` component.

#### To discuss

Where should this go? SvelteKit mentions this in the “Pages” section of their Routing guide. Our routing guide is much more concerned with the logic behind file-based routing, `getStaticPaths` etc. Maybe on “Pages”? On first glance I don’t see the best spot for it. Maybe we need a separate subheading although it would be a fairly short section.

```md
## Navigating between pages

Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) to navigate between routes, rather than a framework-specific `<Link>` component.
```

sarah11918:
- New or updated content

Explicitly mentions using a standard HTML `<a>` tag for linking to other pages on your site. There is no provided `<Link>` component.

To close #992

### DIFF
--- a/src/pages/en/core-concepts/astro-pages.md
+++ b/src/pages/en/core-concepts/astro-pages.md
@@ -11,6 +11,9 @@ i18nReady: true
 
 Astro leverages a routing strategy called **file-based routing.** Every `.astro` file in your `src/pages` directory becomes a page or an endpoint on your site based on its file path.
 
+Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your component template to link between pages.
+
+
 📚 Read more about [Routing in Astro](/en/core-concepts/routing/).
 
 ### Page HTML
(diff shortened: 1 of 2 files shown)


Yan-Thomas:
LGTM.

sarah11918:
Care to weigh in @delucis ? @kevinzunigacuellar ? Is this good enough for now? Would you change anything?

kevinzunigacuellar:
LGTM!

PR #1115, opened 2022-07-25
fixed non-working code example for draft-posts

ChristianK43:

#### Description

Added missing `.then()` to the `await` to make the example work correctly


hippotastic:
Hello @ChristianK43, thank you for bringing this up and your contribution! I just tested it and the example actually doesn't work, so good catch!

However, adding a `.then()` is actually not required in this case. The true reason for the error is that the order of operations is wrong - `await` must be executed **before** trying to call `.filter()`.

If we add brackets to ensure the desired order of operations, the code works:

```js
const posts = (await Astro.glob('../pages/post/*.md'))
  .filter((post) => !post.frontmatter.draft);
```

Alternatively, we could also use a new variable to avoid the brackets and make the code more explicit:

```js
const posts = await Astro.glob('../pages/post/*.md');
const nonDraftPosts = posts.filter((post) => !post.frontmatter.draft);
```

I'd be fine with any of those alternatives. @delucis @sarah11918 what do you prefer?

Once we've found a consensus, you could update your PR accordingly and we could get this moving forward! :)

sarah11918:
Thanks @ChristianK43 for the catch, and @hippotastic for reviewing ! 

I think my preference is for the two seperate statements. As you say, it more explicitly shows the filtering step.

ChristianK43:
Thanks @hippotastic  for the review and correction! I'll be happy to update my PR once you've agreed on a solution. :)

kevinzunigacuellar:
I like the second option as well 😇, very clean

delucis:
One more vote here for option 2! Thanks @ChristianK43!

hippotastic:
Ok @ChristianK43 - feel free to update your PR with the second option then and we're ready to merge! :)

ChristianK43:
done! :)

sarah11918:
Thanks for your first contribution to the Docs, @ChristianK43 !  🥳  Welcome to the Face Pile!

PR #1127, opened 2022-07-27
Migrate to `v1.0-rc`

Yan-Thomas:

#### Description

Docs migration to v1.0-rc! 🥳

~~Note: The `<TableOfContents>` component is throwing console errors and it isn't being updated properly, this is a bug/regression introduced unrelated to the migration (only now discovered but was already happening in the docs).~~ Thanks Hippo for the fix!


hippotastic (approved):
I just performed an automated visual regression test on multiple pages using BackstopJS, as well as a comparison of the entire build output using Araxis Merge and various filters to remove the noise coming from file name and scoped CSS hashes. I also checked the page myself in a browser for changes and console errors and couldn't find any.

LGTM!

hippotastic:
Whoops, after all this testing, we still found something - a strange issue on mobile Safari on iOS!

Algolia search doesn't work there. It works fine in `astro dev --host`, but stops working when doing `astro build` and `astro preview --host`. Same goes for the Netlify preview.

hippotastic:
Issue is fixed! LGTM again! 😂

Yan-Thomas:
Thank you @hippotastic for all the amazing testing and bug fixing! Everyone, please sit in your seats because we are about to lift off!

PR #1120, opened 2022-07-26
Add `cachedFetch()` function to speed up builds

hippotastic:
#### What kind of changes does this PR include?

- Changes to the docs site code

#### Description

- Adds a `cachedFetch()` function to `utils.ts` which uses `@11ty/eleventy-fetch` to cache `fetch()` calls.
- Mimics the `Response` object returned by the `fetch()` API.
- Hopefully makes Astro Docs builds and Netlify deploys more resilient.

PR #1133, opened 2022-07-28
Fix hydration by moving code into server-side file

hippotastic:

#### Description

- Fixes a hydration issue preventing the TOC from updating.
- In PR #1120, I added an import from `@11ty/eleventy-fetch` to `util.ts`, which is both being used in the frontmatter scripts of `.astro` files, but also in our TOC preact component. Apparently tree-shaking isn't being applied properly in this case, so hydration of the TOC component failed.
- I fixed the issue by moving the import and the code using it into a new `util-server.ts` file that is only referenced by server-side code.

### DIFF
--- a/src/components/ContributorList.astro
+++ b/src/components/ContributorList.astro
@@ -1,5 +1,5 @@
 ---
-import { cachedFetch } from '../util';
+import { cachedFetch } from '../util-server';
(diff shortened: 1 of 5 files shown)


Yan-Thomas:
Thanks for this @hippotastic! LGTM.

PR #1138, opened 2022-07-28
Add code snippet titles to `astro-components.md`

hippotastic:

- New or updated content

#### Description

- Used our new title feature to add file tabs to the code snippets on this page.


Yan-Thomas (commented):
LGTM. Note that the title is only `GreetingHeadline.astro` on line 197 (GitHub doesn't let me comment inside lines that weren't modified) rather than having the full file path, I don't have strong feelings about this but maybe we want to follow a pattern of always including the path?

hippotastic:
> LGTM. Note that the title is only `GreetingHeadline.astro` on line 197 (GitHub doesn't let me comment inside lines that weren't modified) rather than having the full file path, I don't have strong feelings about this but maybe we want to follow a pattern of always including the path?

Great catch! I'll add the missing path! :)

PR #1149, opened 2022-07-28
[migration guide] Moving to MDX does NOT require changing layout

sarah11918:
- New or updated content

relies on https://github.com/withastro/astro/pull/4088

Since @bholmesdev gave us `layout:` in MDX, we no longer have to tell people to use a Layout component when converting their `.md` files to `.mdx`

This removes that step, and updates the MDX code sample to show no `<Layout>` component being used, in favour of `layout:` in the frontmatter (which should be what they had before).

NOTE: While there is more exciting MDX layout stuff possible now, for the purposes of just explicitly explaining how to convert an *existing* Astro `.md` file to a working `.mdx` file, we don't need to mention that.

### DIFF
--- a/src/pages/en/migrate.md
+++ b/src/pages/en/migrate.md
@@ -40,33 +40,28 @@ If you're not familiar with MDX, here are some steps you can follow to quickly c
 
 1. Install the [`@astrojs/mdx`](/en/guides/integrations-guide/mdx/) integration.
 
-1. Change your existing `.md` file extensions to `.mdx`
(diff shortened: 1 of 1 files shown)


hippotastic:
LGTM! We should of course wait until the feature has actually been released. :)

delucis (approved):
Merging as the `layout` feature went out last night in [`@astrojs/mdx` v0.4.0](https://github.com/withastro/astro/blob/main/packages/integrations/mdx/CHANGELOG.md#040).

PR #1158, opened 2022-08-01
Sprinkling in some MDX into Markdown page content

sarah11918:
- New or updated content

While we want the MDX content to primarily live on the integrations page/README, this adds sprinkles of "and MDX!" throughout with some links to the integrations docs.

Also, changed the "Basic Example" which suggested creating an `index.md` page to a more generic `page-1.md`. When I tested this in a project that already had an `index.astro` file, the Astro file took prececence and the new Markdown file wasn't actually visible. So, figured this was safer.


src/pages/en/guides/markdown-content.md
```diff
@@ -103,14 +106,67 @@ An example blog post `content` object might look like:
...
+const { title } = Astro.props || Astro.props.content
+---
+<html>
+  <head></head>
+  <body>
+    <h1>{title}</h1>
+    <slot />
+  </body>
+</html>
+```
```
delucis, line 151: My preference would be for the second of you two examples just because it’s a slightly cleaner and easier to extend format (you can easily add `description` or whatever additional frontmatter stuff just by adding to the destructuring).

One note: we should switch `props` and `props.content` otherwise `props.content` will never get used:

````suggestion
```astro
---
// src/components/MyLayout.astro
const { title } = Astro.props.content || Astro.props;
---
<html>
  <head></head>
  <body>
    <h1>{title}</h1>
    <slot />
  </body>
</html>
```
````

There is some nuance here but I think if the aim is to grab stuff from frontmatter, this pattern works well 👍

File: src/pages/en/guides/markdown-content.md [hunk 2]
```diff
@@ -103,14 +106,67 @@ An example blog post `content` object might look like:
...
 
-Any Astro component (not just layouts!) can receive the values defined in your Markdown frontmatter as props. You can specify several types of data using YAML frontmatter, and capture even richer meta information from each blog post to use throughout your Astro site.
+Any Astro component (not just layouts!) can receive the values defined in your Markdown or MDX frontmatter as props. You can specify several types of data using YAML frontmatter, and capture even richer meta information from each blog post to use throughout your Astro site.
+
+Access and use these values in any `.astro` file as you would in a layout.
 
-Access these values in any `.astro` file as you would in a layout, as described above.
+```astro
+---
+src/components/MyAstroComponent.astro
```
delucis, line 161: Missing comment syntax

```suggestion
// src/components/MyAstroComponent.astro
```

src/pages/en/guides/markdown-content.md
```diff
@@ -103,14 +106,67 @@ An example blog post `content` object might look like:
...
+<html>
+  <head></head>
+  <body>
+    <h1>{title}</h1>
+    <slot />
+  </body>
+</html>
+```
+
 ### Frontmatter as Props
```
delucis, line 153: Now that I’m seeing this section, it’s not super clear _how_ these props get where they are in the example. Basically the pattern is grab `content` from a Markdown file (e.g. in a layout) then pass the object to another component.

I’m actually not 100% sure this section adds that much and is kind of confusing as-is. Do we want to consider removing it? Otherwise, I think we need an example where you see a layout render `<MyAstroComponent content={content} />`. Also given the md/mdx confusion do we need `content={Astro.props.content || Astro.props}`?

delucis (commented):
Thanks for continuing your drive to educate the masses about the wonders of MDX @sarah11918! Made a couple of comments 🙌

src/pages/en/guides/markdown-content.md
sarah11918: To address the last point first: If the heading is "Frontmatter as props" then this is how the frontmatter will be used in that Astro component, both md and mdx. Yes, this Astro component *could* also be receiving regular Astro props, in which case it would need to use the trick shown in layouts. But, the heading is for passing frontmatter. I don't know if this tells us that we ALWAYS want to refer to props coming from md/mdx as `Astro.props.content` ?? 

As to whether to keep it, originally there was no code example. It just said "look up at how a layout does it." It's always an option to go back to that, and to be able to keep the nice heading that probably is what many people will be looking for.  :)

The intention here was to provide a small example within the section itself, and being smaller, could show something like a mapping of tags instead of focusing on the entire concept of a layout (which needs to be a larger code snippet if we want to keep things truly copy/pastable). Additionally, we know many readers won't read more than the section they are in, so I thought having something here might be helpful?

src/pages/en/guides/markdown-content.md
```diff
@@ -103,14 +106,31 @@ An example blog post `content` object might look like:
 ```
 
 :::note
-`astro`, `file`, and `url` are the only guaranteed properties provided by Astro in the `content` prop. The rest of the object is defined by your frontmatter variables.
+`astro`, `file`, and `url` are the only guaranteed properties provided by Astro in the Markdown `content` prop. The rest of the object is defined by your frontmatter variables.
+
+📚 Read more about the [MDX content prop](/en/guides/integrations-guide/mdx/).
```
sarah11918, line 111:
```suggestion
📚 Read more about the [MDX `content` prop](/en/guides/integrations-guide/mdx/) and how it compares to the Markdown `content` object.
```

src/pages/en/guides/markdown-content.md
delucis: Yeah, I think the issue is that there’s nothing “special” about “frontmatter as props”, which I think having a dedicated section might imply. I was reading it thinking “Ooh, a component can magically get `content`!” whereas actually the advice we’re trying to give is: “`content` is just another (not special) JavaScript object — you can pass it to a component if you like”.

delucis (commented):
LGTM! Thanks @sarah11918 🙌

PR #1160, opened 2022-08-01
Add code block inline marking, including ins & del marker types

hippotastic:

#### Description

- Adds support for line marking definitions with an optional marker type prefix. Examples:
  - `{4-5,10}` (if no marker type prefix is given, it defaults to `mark`)
  - `mark={4-5,10}`
  - `del={4-5,10}` (new, marks a deleted line, alias name: `rem`)
  - `ins={4-5,10}` (new, marks an added line, alias name: `add`)
- Adds support for inline marking definitions inside forward slashes, with an optional marker type prefix. Examples:
  - `/sidebar/` (if no marker type prefix is given, it defaults to `mark`)
  - `mark=/sidebar/` (all common regular expression features are supported)
  - `mark=/slot="(.*?)"/` (if capture groups are contained, these will be marked)
  - `del=/src\/pages\/.*\.astro/` (escaping special chars with a backslash works, too)
  - `ins=/this|that/`
- Fixes various escaping issues by replacing our faulty `escapeHtml`/`unescapeHtml` functions with the battle-tested `html-escaper` package.
- Adds marker examples to `framework-components.md`:
  - Line & inline markers of `add`/`ins` type: 
image
  - Marking regular expression capture groups:
image


hippotastic:
Looks like Netlify's Node.js version doesn't support the RegExp flag `d` which I'm using to get the indices of capture groups. Will need to find an alternative for that. :)

hippotastic:
I added a fallback that seems to work well. https://deploy-preview-1160--astro-docs-2.netlify.app/en/core-concepts/framework-components/ :)

PR #1163, opened 2022-08-02
Allow inline marking of plaintext strings

hippotastic:

- New or updated content

#### Description

- Improved the inline marking authoring experience by adding support for plaintext matches in addition to the existing regular expressions.

- Here's an overview of the new plaintext matching syntax options:
  - `"Astro.props"`
    (if no marker type prefix is given, it defaults to `mark`)
  - `ins="<Button />"`
    (matches will be marked with "inserted" style)
  - `del="<p class=\"hi\">"`
    (special chars in the search string can be escaped by `\`)
  - `del='<p class="hi">'`
    (use single quotes to make it easier to match double quotes)

- Marked `migrate.md` using all available syntax options as an example.

### DIFF
--- a/integrations/astro-code-snippets.ts
+++ b/integrations/astro-code-snippets.ts
@@ -140,10 +140,13 @@ export function remarkCodeSnippets(): Plugin<[], Root> {
  * Meta information is the string after the opening code fence and language name.
  */
 function parseMeta(meta: string) {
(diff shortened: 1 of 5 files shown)


delucis (approved):
For the record following Discord chat: LGTM!

PR #1166, opened 2022-08-03
Mark code in `astro-components.md`

hippotastic:

- New or updated content

#### Description

- Uses our new code snippet inline marking feature to highlight important parts of the code blocks.

### DIFF
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -113,7 +113,7 @@ You can can define local JavaScript variables inside of the frontmatter componen
 
 Local variables can be added into the HTML using the curly braces syntax:
 
-```astro title="src/components/Variables.astro"
+```astro title="src/components/Variables.astro" "{name}"
 ---
 const name = "Astro";
 ---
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/astro-components.md
```diff
@@ -194,18 +194,18 @@ An Astro component can define and accept props. These props then become availabl
 
 Here is an example of a component that receives a `greeting` prop and a `name` prop. Notice that the props to be received are destructured from the global `Astro.props` object.
 
-```astro
+```astro /(?<!//.*)({?(greeting|name)}?)/ "Astro.props"
```
kevinzunigacuellar, line 197: I think this might be a bit overkill

hippotastic: Thank you for the feedback! What part do you consider to be an overkill and which alternative would you prefer here?

hippotastic: Ok, I experimented with it a bit and think I've found a better version: I removed the variable name marking and just left the main source of all data (`Astro.props`) marked, because this is what's being introduced in that code snippet.

What do you think @kevinzunigacuellar?

kevinzunigacuellar: I was thinking about removing all the highlights but I like the new version better! 

Good to merge!

PR #1171, opened 2022-08-03
Fix Shiki tokens with styles other than `color`

hippotastic:

#### Description

- Fixes cases where Shiki tokens contain additional styles like bold or italics. Before this fix, such tokens were not recognized and removed from the output in the code snippet marking phase.
  - Before:
image

  - After:
image


sarah11918:
LGTM!

PR #1172, opened 2022-08-03
Mark code in `markdown-content.md`

hippotastic:

- New or updated content

#### Description

- Uses our new code snippet inline marking feature to highlight important parts of the code blocks.

### DIFF
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -366,37 +366,37 @@ By default, Astro comes with [GitHub-flavored Markdown](https://github.com/remar
 
 2. Update `remarkPlugins` or `rehypePlugins` inside the `markdown` options:
 
-   ```js
-   // astro.config.mjs
-   export default {
-     markdown: {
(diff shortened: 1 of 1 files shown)


kevinzunigacuellar:
I was thinking we could add highlights to the [second](https://deploy-preview-1172--astro-docs-2.netlify.app/en/guides/markdown-content/#how-to-add-a-markdown-plugin-in-astro) `astro.config.mjs`. Maybe the import line and the remarkplugin line. What do you think?

image

hippotastic:
> I was thinking we could add highlights [...]. Maybe the import line and the remarkplugin line. What do you think?

I agree! Added green insert markings for both lines as suggested. :) Did anything else catch your eye or are we good to go?

kevinzunigacuellar:
> > I was thinking we could add highlights [...]. Maybe the import line and the remarkplugin line. What do you think?
> 
> I agree! Added green insert markings for both lines as suggested. :) Did anything else catch your eye or are we good to go?

Looks good to me!

PR #1189, opened 2022-08-03
rss: change post.frontmatter.slug to post.url

### LINKED ISSUE #1082: reference to `slug` in RSS docs
The docs for generating RSS make reference to a frontmatter field called "slug" which is not auto-generated. Trying to use the code given [here](https://docs.astro.build/en/guides/rss/#2-list-of-rss-feed-objects) as-is results in errors. All references to `slug` should be changed to `url`, which works correctly and is what's actually used if you call the glob version of `rss()` instead of the feed objects version.

kevinzunigacuellar:

#### Description

- Closes #1082 
- Update RSS codeblock example to use `post.url` variable


sarah11918:
Yay! Thanks @kevinzunigacuellar !  🥳

PR #1195, opened 2022-08-04
Fix wrong Node.js version numbers

strager:
Installing Astro with Node.js version 16.5.0 fails:

> error @astrojs/telemetry@0.4.1: The engine "node" is incompatible with this
> module. Expected version "^14.18.0 || >=16.12.0". Got "16.5.0"

Update the docs to reflect the minimum required Node.js versions.

- New or updated content

#### Description

Update the docs to reflect the minimum required Node.js versions.

### DIFF
--- a/src/pages/en/install/auto.md
+++ b/src/pages/en/install/auto.md
@@ -9,7 +9,7 @@ Ready to install Astro? Follow our automatic or manual set-up guide to get start
 
 #### Prerequisites
 
-- **Node.js** - `14.15.0`, `v16.0.0`, or higher.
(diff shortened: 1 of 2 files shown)


sarah11918:
Thanks for this! We absolutely do want the Docs to match the error message.

I'm just going to quickly confirm with engineering that this isn't going to change again on us anytime quickly, but then we'll get this added to Docs!

sarah11918:
Alright, confirmed we're good to go with these versions!  🥳 

Thanks so much for this super-helpful contribution, and welcome to Docs! 🚀

PR #1199, opened 2022-08-04
Add the Astro Prettier Plugin

Princesseuh:

- Something else!

#### Description

This adds our official Prettier plugin to the formatting setup the docs use and format Astro and TS files. However, this PR by itself doesn't format Markdown files as I fear it would absolutely wreak the i18n process


jasikpark:
shouldn't this also format `astro` codeblocks in the markdown docs?

Princesseuh:
> shouldn't this also format `astro` codeblocks in the markdown docs?

Markdown files are currently excluded from formatting to avoid breaking the i18n process!

.github/workflows/format.yml
```diff
@@ -0,0 +1,37 @@
...
+          token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
+      - name: Setup PNPM
+        uses: pnpm/action-setup@v2.2.1
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: 'pnpm'
+      - name: Install dependencies
+        run: pnpm install
```
delucis, line 28: We have a local action for these steps that we use to keep it consistent between workflows:

```suggestion
      - name: Install Tools & Dependencies
        uses: ./.github/actions/install
```

delucis (commented):
Thanks for this @Princesseuh — looking at a lot of these changes makes me happy :-D

One small improvement for the CI workflow to tailor it to docs but then LGTM 🙌

delucis (approved):
LGTM! Thank you 🙌

PR #1203, opened 2022-08-05
Explain when you might want to use import.meta.glob

Jutanium:
This PR adds on to the existing note that `Astro.glob` wraps `import.meta.glob`. It makes it clear that you can still use the latter, and explains why you'd want to.


sarah11918:
Thanks for this @Jutanium! 

So, gut-check: the original note WAS ACTUALLY ABOUT `Astro.glob()`. It explained itself by *mentioning* `import.meta.glob()` behavior, but I'm not sure we'd say that this note is ABOUT Vite's `import.meta.glob()`. Does that resonate with you?

I'm wondering whether we actually have two different notes (kind of do not like!), or whether your content is otherwise "un-noteworth" as a... replacement for? workaround to? `Astro.glob()`. If we view your addition as a piece of content separate from the existing note (which again, gut feeling that it is?)... any other thoughts on presenting this? 

I still do see it in `Astro.glob()` territory, and do think you have the right SPOT for it. But I have this little nagging worry about the original intent of pointing people to troubleshooting kind of getting lost here? I'm open to feedback on this!

Jutanium:
Right, it depends how you look at it: I looked at the original note as saying:

"Astro.glob extends import.meta.glob, and therefore X." I added "therefore Y." 

But you can view the original note as saying "X, because Astro.glob extends import.meta.glob." In that case it doesn't make sense to add this other thing in the same note.

I'm cool to break it off into another note

sarah11918:
@Jutanium Maybe only because I know how the note originated, it def feels like "X because..." to me. Let's try it with a subheading and see how it looks! I do see it as separate, and outside of the note will give it some more formatting flexibility, too.

(And, let me know if this is a "Dan is done with this" thing, or whether you want to continue with this. Happy either way!)

Jutanium:
@sarah11918 Here's one idea! I'll let you take it from here as I'm ambivalent

src/pages/en/reference/api-reference.md
```diff
@@ -34,8 +34,15 @@ const posts = await Astro.glob('../pages/post/*.md'); // returns an array of pos
 
 `.glob()` only takes one parameter: a relative URL glob of which local files you'd like to import. It’s asynchronous, and returns an array of the exports from matching files.
 
-:::note
-`Astro.glob()` is a wrapper of Vite's [`import.meta.glob()`](https://vitejs.dev/guide/features.html#glob-import), so it cannot accept variables as they are not statically analyzable. See [the troubleshooting guide](/en/guides/troubleshooting/#supported-values) for a workaround.
+`.glob()` can't take variables or strings that interpolate them, as they aren't statically analyzable. (See [the troubleshooting guide](/en/guides/troubleshooting/#supported-values) for a workaround.) This is because `Astro.glob()` is a wrapper of Vite's [`import.meta.glob()`](https://vitejs.dev/guide/features.html#glob-import).
+
+::: Note
```
delucis, line 39: Swinging by to pick up my co-author credit 😅 

```suggestion
:::note
```

Thanks for tackling this Dan!

sarah11918:
Another one for @bholmesdev! Ready when you are!

sarah11918:
Thanks, Dan! Let's get that money!

PR #1200, opened 2022-08-05
[add] Astro.props added to config

sarah11918:
- New or updated content

Addresses #632 

Just getting something super simple started!

PR #1208, opened 2022-08-06
Example with default values for props

kylebutts:

- New or updated content

#### Description

Following discussion in #1200, I added an example on how to use default values for props. I also moved the typescript example down so that defining and using a component were coupled. Happy to move that back if folks think it reads better the other way!

Before:
CleanShot 2022-08-06 at 08 26 29@2x

After:
CleanShot 2022-08-06 at 08 17 54@2x


src/pages/en/core-concepts/astro-components.md
```diff
@@ -245,17 +259,14 @@ const { greeting = "Hello", name } = Astro.props as Props;
 <h2>{greeting}, {name}!</h2>
 ```
 
-This component, when imported and rendered in other Astro components, layouts or pages, can be passed these props as attributes:
+Component props can be given default values to use when the consumer does not provide them.
```
sarah11918, line 262:
```suggestion
Component props can be given default values to use when none are provided.
```
I think I might prefer something shorter like this? What do you think?

sarah11918 (commented):
I like the example, thanks @kylebutts!

I think the reordering reads fine, so let's keep it. I think I'd prefer not to introduce the word "consumer" without having used it anywhere before, so it's not just dropped in there on people without having first established it. So I've suggested something shorter I think could introduce the example. Let me know what you think.

sarah11918:
Hey @kylebutts! I'm just commiting the minor wording change I suggested, then we'll get this merged, live on the site and helping people! 🥳  Thanks again for this!

kylebutts:
Thanks @sarah11918! I was at Rocky Mountain NP all weekend, hence my missingness :-)

sarah11918:
Aww.... sorry, I wouldn't normally have just merged on you so quickly, @kylebutts! I do like to let people at leave give their OKs. But, we're launching v1 on Tuesday, and I'm trying to get through as much as possible!  😅  

Your contribution is so helpful, I just wanted it IN THE DOCS, not sitting here.  :)

Hope you had a fabulous weekend in nature! ⛰️ 🌳

PR #1209, opened 2022-08-06
[add] Headless CMS content on data fetching page

sarah11918:
- New or updated content

We don't currently have any content re: adding a Headless CMS. These aren't "Guides" but a first-pass at illustrations of "what it looks like" with links to some external tutorials.

Added to the Data Fetching page to illustrate "fetching content from your Headless CMS" and also adjusted headings so this content is clearly visible in the right sidebar / on this page TOC.

Example from new page:
Screenshot 2022-08-06 13 52 13


src/pages/en/guides/data-fetching.md
```diff
@@ -72,33 +104,69 @@ const weather = json.data;
...
-// Output the result to the page
-  return <div>{JSON.stringify(data)}</div>;
-};
+### Example: WordPress + GraphQL
 
-export default Movies;
+```astro
+---
+// src/pages/blog/[slug].astro
+// Render pages dynamically from your Wordpress content
```
Yan-Thomas, line 148: This is the only case where we are using `Wordpress` rather than `WordPress`
```suggestion
// Render pages dynamically from your WordPress content
```

Yan-Thomas (commented):
Nice work as always @sarah11918! 🙌 I added a few suggestions but after these LGTM!

src/pages/en/guides/data-fetching.md
```diff
@@ -72,33 +104,69 @@ const weather = json.data;
...
+  body: `
+    {
+      page(id:"${slug}", idType:URI) {
+        title 
+        content 
+      }
+    }
+  `
+});
+ const data = await response.json();
```
sarah11918, line 164:
```suggestion
// src/pages/blog/[slug].astro
// Render pages dynamically from your WordPress content

import BaseLayout from '../../layouts/BaseLayout.astro';
// export default getStaticPaths(){...}
// const slug = Astro.params.slug 
const response = await fetch(import.meta.env.WORDPRESS_API_URL, {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: `
    {
      page(id:"${slug}", idType:URI) {
        title 
        content 
      }
    }
  `
});
const data = await response.json();
```

src/pages/en/guides/data-fetching.md
```diff
@@ -72,33 +104,70 @@ const weather = json.data;
...
-);
+### Example: StoryBlok API
 
-// Components that are build-time rendered also log to the CLI.
-// When rendered with a client:* directive, they also log to the browser console.
-console.log(data);
+```astro
+---
+// src/pages/[slug].astro
+// Fetch a list of your StoryBlok page links using @storyblok/js and axios
```
delucis, line 120: As we don’t _show_ `axios` in use (it’s just internal to Storyblok’s API client), we probably shouldn’t mention it here. (I know I mentioned it, but I think it would only need to be included if we were showing install instructions.)

```suggestion
// Fetch a list of your StoryBlok page links using @storyblok/js
```

src/pages/en/guides/data-fetching.md
```diff
@@ -72,33 +104,70 @@ const weather = json.data;
...
 
-export default Movies;
+```astro
+---
+// src/pages/blog/[slug].astro
+// Render pages dynamically from your WordPress content
+
+import BaseLayout from '../../layouts/BaseLayout.astro';
+// export default getStaticPaths(){...}
+// const slug = Astro.params.slug
```
delucis, line 152: I don’t love this style of elision. How about we just get rid of `getStaticPaths` and set a constant slug for simplicity’s sake?

```suggestion

const slug = 'about';
```

delucis (commented):
Thanks so much for stepping up and tackling this this weekend, @sarah11918!!!

I’ve made a couple of comments, but otherwise LGTM!

src/pages/en/guides/data-fetching.md
```diff
@@ -72,33 +104,70 @@ const weather = json.data;
...
-Remember, all data in Astro components is fetched when a component is rendered.
-
-Your deployed Astro site will fetch data **once, at build time**. In dev, you will see data fetches on component refreshes. If you need to re-fetch data multiple times client-side, use a [framework component](/en/core-concepts/framework-components/) or a [client-side script](/en/core-concepts/astro-components/#client-side-scripts) in an Astro component.
-:::
+## Fetch from a Headless CMS
 
+Fetch remote content from your favorite CMS like StoryBlok or WordPress!
 
-## `fetch()` in Framework Components
+With an API key, some helper functions, and `getStaticPaths()` to generate dynamic page routes, an `.astro` component can fetch and render data from your CMS to generate your page content.
```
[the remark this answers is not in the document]
sarah11918: Once again, wielding the power I should not have to update your commit suggestion to take out the deploy part of the link!

src/pages/en/guides/data-fetching.md
sarah11918: Ah! Good point! I didn't notice that the example would work in the "singular" with no trace of `getStaticPaths()`, but then should we change the file name to NOT be the dynamic `[slug].astro`?

delucis: Oh yes! We should. Missed that detail.

src/pages/en/guides/data-fetching.md
```diff
@@ -72,33 +104,70 @@ const weather = json.data;
...
-const Movies: FunctionalComponent = () => {
-// Output the result to the page
-  return <div>{JSON.stringify(data)}</div>;
-};
+### Example: WordPress + GraphQL
 
-export default Movies;
+```astro
+---
+// src/pages/blog/[slug].astro
```
delucis, line 147: Updated filename to match change in sample content — well spotted Sarah!

```suggestion
// src/pages/about.astro
```

src/pages/en/guides/data-fetching.md
```diff
@@ -72,33 +104,70 @@ const weather = json.data;
...
+// Render pages dynamically from your WordPress content
```
delucis, line 148: And I guess this should be amended too

```suggestion
// Fetch your about page content from the WordPress API
```

PR #1211, opened 2022-08-06
Update migrate.md

Hopelezz:
First instance of ESM acronym expressed fully.

#### Description

- Based on question in discord discussion.
image

### DIFF
--- a/src/pages/en/migrate.md
+++ b/src/pages/en/migrate.md
@@ -73,7 +73,7 @@ If you're not familiar with MDX, here are some steps you can follow to quickly c
 
 2. Change your existing `.md` file extensions to `.mdx`
 
-3. Remove the `setup:` property from your frontmatter, and write its [ECMAScript Module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) import statements below the frontmatter.
+3. Remove any `setup:` properties from your frontmatter, and write any import statements below the frontmatter instead.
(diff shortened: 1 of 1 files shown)


src/pages/en/migrate.md
```diff
@@ -73,7 +73,7 @@ If you're not familiar with MDX, here are some steps you can follow to quickly c
 
 2. Change your existing `.md` file extensions to `.mdx`
 
-3. Remove the `setup:` property from your frontmatter, and write its ESM import statements below the frontmatter.
+3. Remove the `setup:` property from your frontmatter, and write its [ECMAScript Module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) import statements below the frontmatter.
```
sarah11918, line 76:
```suggestion
3. Remove any `setup:` properties from your frontmatter, and write any import statements below the frontmatter instead.
```
I don't find the longer name particularly useful, when the link doesn't refer to it or explain what that phrase means. And, I don't think the link to import is helpful because we don't have to tell people what an import is. In the context of these instructions, you are simply writing the exact same statement in a different place. The link is only distracting, I think? You typically don't want to take readers away from a page unless it necessary.

I did change the text to be more clear about taking the import statement that is already there and writing it below the code fence instead.

sarah11918:
Thanks for the help making this clearer, @Hopelezz !

Issue #1154, opened 2022-07-30
Content audit for Accessibility guidance


sarah11918:
Let's review our docs content for 

- areas where it makes sense to mention (especially Astro-specific) tips for designing with a11y in mind.
- our code examples to make sure our own suggestions suggest good practices
- ... something else?

Building off of @Yan-Thomas's comments in Discord:

"We could sprinkle some accessibility tips and tricks in some pages, I think we can easily explain some of Astro's features using accessibility. For example, we can showcase how we can easily change an page lang atrribute using layouts and briefly explain why this is helpful. Also worth mentioning that UI framework-specific accessibility patterns should be the same (if the component is properly hydrated)."

delucis:
We could also consider this in our MPA vs SPA comparison: one benefit of MPAs is not needing to rebuild the a11y structure around navigation (focus management, route change announcements, etc). I remember Marcy Sutton had to do HEAPS of work to make Gatsby more accessible out of the box and these concerns are just solved for you by the browser vendors when you choose an MPA.

Yan-Thomas:
Based on what I suggested before, Chris' comment, and some thoughts I had while reading the docs looking for things we could improve, I would tackle this like this:

### Layouts + Pages
We know layouts are just Astro components, but I think people may think that there is some limitation, and we currently don't talk about using props to change bits of the page layout, this is a perfect opportunity to change our example to use props and show how we can use layouts to simplify internationalization. In the end, there would be a small tip (or note?) about the importance of the lang attribute for accessibility and how Astro makes this kind of stuff easy. In a bigger codebase, people will prefer to move this i18n logic to some route-checking script rather than manually adding a lang attribute, so I'm not sure if this example/use-case is good.

Also, in the Astro Pages' [Page HTML](https://docs.astro.build/en/core-concepts/astro-pages/#page-html) section, we show a `<html></html>` without the `lang` attribute, I think it's worth adding these small a11y improvements to examples as a way to influence and prompt people to ask and search things like "why use the lang attribute?"

### UI Frameworks
As mentioned before, I think we could add a small accessibility note at the end of "[Hydrating Interactive Components](https://docs.astro.build/en/core-concepts/framework-components/#hydrating-interactive-components)" explaining that framework-specific accessibility patterns should be the same if the component is properly hydrated.

### Static Assets
[...]

PR #1213, opened 2022-08-06
Added alt attribute to <img> tag and added note

Njong392:

- New or updated content

#### Description

- ~~Closes #~~ (Should not close the issue yet, but references #1154)
- The PR makes changes to the [other assets](https://docs.astro.build/en/guides/imports/#other-assets) section of the docs. In order to improve on the promotion of accessibility, I added an alt attribute to the image tag and then added a small note underneath to encourage its use

### DIFF
--- a/src/pages/en/guides/imports.md
+++ b/src/pages/en/guides/imports.md
@@ -134,7 +134,7 @@ All other assets not explicitly mentioned above can be imported via ESM `import`
 It can also be useful to place images in the `public/` folder as explained on the [project-structure page](/en/core-concepts/project-structure/#public).
 
 :::note
-Adding **ALT text** to `img` tags is awesome practice! Don't forget to add `alt` text to your images, or just leave the attribute empty if the image is purely decorative.
+Adding **alt text** to `<img>` tags is encouraged for accessibility! Don't forget to add an `alt="a helpful description"` attribute to your image elements. You can just leave the attribute empty if the image is purely decorative.
 :::
 
 ## WASM
--- a/src/pages/en/guides/imports.md
+++ b/src/pages/en/guides/imports.md
@@ -126,12 +126,16 @@ import svgReference from './image.svg'; // svg === '/src/image.svg'
 import txtReference from './words.txt'; // txt === '/src/words.txt'
 
 // This example uses JSX, but you can use import references with any framework.
-<img src={imgReference} />;
+<img src={imgReference} alt="image description" />;
 ```
 
 All other assets not explicitly mentioned above can be imported via ESM `import` and will return a URL reference to the final built asset. This can be useful for referencing non-JS assets by URL, like creating an image element with a `src` attribute pointing to that image.
 
 It can also be useful to place images in the `public/` folder as explained on the [project-structure page](/en/core-concepts/project-structure/#public).
+
+:::note
+Adding **ALT text** to `img` tags is awesome practice! Don't forget to add `alt` text to your images, or just leave the attribute empty if the image is purely decorative.
+:::
 
 ## WASM
 


src/pages/en/guides/imports.md
```diff
@@ -126,13 +126,17 @@ import svgReference from './image.svg'; // svg === '/src/image.svg'
...
-<img src={imgReference} />;
+<img src={imgReference} alt="image description" />;
 ```
 
 All other assets not explicitly mentioned above can be imported via ESM `import` and will return a URL reference to the final built asset. This can be useful for referencing non-JS assets by URL, like creating an image element with a `src` attribute pointing to that image.
 
 It can also be useful to place images in the `public/` folder as explained on the [project-structure page](/en/core-concepts/project-structure/#public).
 
+:::note
+Adding **ALT text** to `img` tags is awesome practice! Don't forget to add `alt` text to your images, or just leave the attribute empty if the image is purely decorative.
```
sarah11918, line 137:
```suggestion
Adding **alt text** to `<img>` tags is encouraged for accessibility! Don't forget to add an `alt="a helpful description"` attribute to your image elements. You can just leave the attribute empty if the image is purely decorative.
```
Thanks for this helpful (and, encouraged!) addition @Njong392! 🥳 

I've made some suggestions to consider, and this looks like a lot  😅 ... it's not that I'm super picky (ok... maybe I HAVE BEEN KNOWN to be picky, lol) it's just that I do like to explain the thoughts I have as I go through text so that people know why I might suggest changing even just a single word sometimes... that it's not just to pick at things.

So, here's what went through my mind with this. For your consideration; let me know what you think!

- it might seem redundant, but we do typically use the tag symbols in addition to the word when we reference elements. So I've added those for consistency.

- I checked a handful of websites, and most of the places I've seen use "alt text" in all lower case. Some use "Alt Text" but the lower case seems more common? If there's a canonical source you're aware of, please let me know! Otherwise, I think with the bold text already, we should probably just go with "alt text" lower case.

- It **is** AWESOME practice, but I think I like the idea of specifically stating why, and reinforcing that we shoudld care about accessibility, even beyond just alt text!

- I also similarly wanted to remove "or", because the suggestion should be include the tag no matter what, right? So, let's show them the attribute, and then separately say, "No, srsly, write the attribute! If it's purely decorative, we can just leave the alt text blank!"

Njong392: Ohh... you made the PR look a lot better :) I do understand your changes, and I think it reads a lot better with what you added. As per whether it's "alt text" or "Alt Text" or "ALT Text", I admit that I have never really paid much attention haha

Also, I just wanted to mention this. I did see someone mention before that alt texts could also be added to SVGs. Stuff like videos and GIFs too. Would it be too much to not only mention img tags in this situation? I get that the context of the note refers to the `` tag mentioned in the docs, but is it worth mentioning the others?

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :question: I have a question

sarah11918: (It's my responsibility to pay ALL the attention to ALL the things.....) ;)

OK, with all the other elements, too.... in that case what about something like:

> Don't forget to add an `alt="a helpful description"` attribute to all of your media elements

(Basically replacing "image" with "media"?) Or, what other generic term could we use there? We could even tack a list on the end like:

> Don't forget to add an `alt="a helpful description"` attribute to all of your media elements, including SVGs, videos and GIF too.

What about doing something like that?

Yan-Thomas: > Also, I just wanted to mention this. I did see someone mention before that alt texts could also be added to SVGs. Stuff like videos and GIFs too. Would it be too much to not only mention img tags in this situation? I get that the context of the note refers to the  tag mentioned in the docs, but is it worth mentioning the others?

Hi @Njong392 👋, that's a great question!

GIFs also use the `img` tag so I think we don't need to explicitly mention them. About videos and SVGs, they shouldn't be described with `alt` because their respective elements don't accept it, according to [W3Schools](https://www.w3schools.com/tags/att_img_alt.asp) and [MDN](https://developer.mozilla.org/pt-BR/docs/Web/HTML/Attributes).

To give more context, they have their own accessibility recommendations:
* Non-decorative SVGs should be described with `aria-label` or `aria-labelledby`.
* Videos need captions (if they contain audio) and/or be described in the following paragraph.

Njong392: Oh... I see where you are coming from. That is true! I now wonder how the note could be altered to fit every media type (if we should include every media type.)

@sarah11918 how about

> Don't forget to add a text-based description to all of your media elements (from images to SVGs, to GIFs, and videos)

Yan-Thomas: > I now wonder how the note could be altered to fit every media type (if we should include every media type.)

I personally think we don't need to mention every media type here, just image elements is fine I think. The accessibility patterns and gotchas vary a lot depending on the media and how it's used, so properly explaining this would be a guide of its own. What I think is our big objective here is to make our docs' readers aware of accessibility and incentivize them to know more about it rather than being an in-depth resource about it.

sarah11918: Yes, I agree with @Yan-Thomas here. It's a BIG topic that is not really in our scope, and if Yan thinks it's fine to keep it to image elements (since, that's what we're talking about here!) then I'm 100 on board with that! 

So, I would say all my original suggestions are still my takes on this! @Njong392, if you like those, you can click to "commit" the suggested comment as is. Or, rewrite to suit your taste and I'll look over it again.

Njong392: > So, I would say all my original suggestions are still my takes on this! @Njong392, if you like those, you can click to "commit" the suggested comment as is. Or, rewrite to suit your taste and I'll look over it again.

Awesome! I initially already agreed to your original suggestions, so I'll just go ahead and commit those :)


__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix

Yan-Thomas (approved):
LGTM. Thanks for this @Njong392!

sarah11918:
Thanks so much for stepping up for this, @Njong392! Welcome to Team Docs and the Face Pile!  🥳

PR #1223, opened 2022-08-07
Add note on dynamic HTML tags

### LINKED ISSUE #1180: Document dynamic tag name pattern
Splitting this off from #632. We should document the ability to have dynamic tag names on the Astro Components page!

> About dynamic tag rendering: there's nothing “special” about a tag prop so I'm not sure this would go there. You could name your prop is or as for example and it would also work. All that needs to happen is for you to use a variable that is a string, so that could come from a prop or something else:
>
> ```astro
> ---
> const El = 'div'
> ---
> <El>Hello!</El>
> ```

AsyncBanana:

- New or updated content

#### Description

- Closes #1180

This just adds a note to the Astro components page for dynamic HTML tags. I wasn't sure about the positioning of it, but it seemed like it made the most sense in the dynamic HTML section.

### DIFF
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -152,6 +152,17 @@ const items = ["Dog", "Cat", "Platypus"];
 </ul>
 ```
 
+:::tip
+You can also set tags dynamically:
+
+```astro
+---
+const El = 'div'
+---
+<El>Hello!</El> <!-- El will be transformed into a div -->
+```
+:::
+
 ### Fragments & Multiple Elements
 
 An Astro component template can render multiple elements with no need to wrap everything in a single `<div>` or `<>`, unlike JavaScript or JSX.
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -155,11 +155,12 @@ const items = ["Dog", "Cat", "Platypus"];
 :::tip
 You can also set tags dynamically:
 
-```astro
+```astro "El"
 ---
+// src/pages/index.astro
 const El = 'div'
 ---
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/astro-components.md
```diff
@@ -152,6 +152,17 @@ const items = ["Dog", "Cat", "Platypus"];
...
 
+:::tip
+You can also set tags dynamically:
+
+```astro
+---
+const El = 'div'
+---
+<El>Hello!</El> <!-- El will be transformed into a div -->
+```
```
sarah11918, line 163:
``````suggestion
```astro "El"
---
// src/pages/index.astro
const El = 'div'
---
<El>Hello!</El> <!-- renders as <div>Hello!</div> -->
```
``````
Making the code formatting changes to show a file name title, and (I think!) highlight the word "El." Committing to the branch so that I can verify in the deploy preview that this does what I think it does!

sarah11918 (commented):
Thanks for this, @AsyncBanana!

I know we can fancy-up the code snippet with some of our special syntax, so I'm committing a suggestion to your branch now to verify the appearance in the preview. We can still keep discussing it, but making sure it will render how we want to the page is helpful!

sarah11918:
I think the tip looks smashing with the highlighting! 

Can I get some guidance on the exact wording for the code comment: "renders as" vs "renders"? Something else? @AsyncBanana @Yan-Thomas @delucis

Yan-Thomas:
Nice work @AsyncBanana, thanks for this! LGTM 🙌 

> Can I get some guidance on the exact wording for the code comment: "renders as" vs "renders"?

I think "renders as" is fine!

sarah11918:
Woo! Thanks @AsyncBanana! I think this is good to go! 🥳

PR #1232, opened 2022-08-08
Markdown `content` prop -> `frontmatter` prop

bholmesdev:

- New or updated content

#### Description

Change all references to the Markdown layout `Astro.props.content` property to `Astro.props.frontmatter`. Both will continue to work, but `frontmatter` should be the new recommendation. Seemed a _bit_ much to add deprecation notices everywhere, so I'm voting for a clean sweep for 1.0 stable here!

### DIFF
--- a/src/pages/en/core-concepts/layouts.md
+++ b/src/pages/en/core-concepts/layouts.md
@@ -68,19 +68,19 @@ description: My first blog post!
 This is a post written in Markdown.
 ```
 
-When a Markdown file includes a layout, it passes a `content` property to the `.astro` component which includes the frontmatter properties and the final HTML output of the page.
+When a Markdown file includes a layout, it passes a `frontmatter` property to the `.astro` component which includes the frontmatter properties and the final HTML output of the page.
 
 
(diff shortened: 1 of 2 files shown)


Yan-Thomas:
Did a scan through the docs Markdown files and I think you covered all the `content` occurrences, LGTM! Thanks for this @bholmesdev 🙌

sarah11918:
@bholmesdev These look good, BUT, the content from `integrations-guide/integrations/mdx` is mirrored from https://github.com/withastro/astro/tree/main/packages/integrations/mdx/README.md so the changes have to be made there.

Otherwise, I think this is fine!

bholmesdev:
@sarah11918 MDX README updates here!
https://github.com/withastro/astro/pull/4237

sarah11918:
@bholmesdev Got it! (See my comment in PTAL re: one other link you *could* fix on that page for me, while you're there. Or, I can do it before you merge.)

ARE YOU OTHERWISE OK HAPPY WITH THIS PR, and the changes to the other 2 pages?

PR #1249, opened 2022-08-10
Make the tabs fancy

zadeviggers:

- Updated `Tabs` component, adding a fancy animation when the user switches tabs.

#### Description

- Adds fancy tab switching animation to... one of the many different tabs components. This one seemed the most general, so it was the one I modified.

Before:
https://user-images.githubusercontent.com/74938858/183867083-b67d32b2-2864-4c15-ba9d-ba9000c71a60.mov

After:
https://user-images.githubusercontent.com/74938858/183866697-3f469246-7f1a-4b7c-bde0-bfe4b030afb8.mov

Some of the CSS I had to change seemed kinda sketchy, so LMK if something's broken. I also changed the focus styles from applying on `:focus` to on `:focus-visible` (which has slightly lower browser support - [around 89% according to caniuse](https://caniuse.com/css-focus-visible) because it was running the effect).

Fun fact: this was the first time I've ever used `useLayoutEffect`!

### DIFF
--- a/src/components/tabs/Tabs.module.css
+++ b/src/components/tabs/Tabs.module.css
@@ -7,19 +7,43 @@
 	margin-inline-start: -1rem;
 	overflow-x: scroll;
 	/* Hide scrollbar on firefox */
-	scrollbar-width: none; 
+	scrollbar-width: none;
 }
 
 .tab-scroll-overflow::-webkit-scrollbar {
 	/* Hide scrollbar on chrome / safari */
-  display: none;
+	display: none;
 }
 
-.tab:focus {
+.tab,
+.tab:focus,
+.tab:active {
+	border-bottom: 0 !important;
+}
+
+.tab:focus-visible {
 	outline: 2px solid var(--theme-accent);
 	outline-offset: -2px;
 }
 
 .tabpanel {
 	padding-block-start: 1rem;
+}
+
+.tablist {
+	position: relative;
+}
+
+.selectedIndicator {
+	position: absolute;
+	bottom: -2px;
+	height: 4px;
+	background-color: var(--theme-accent);
+	transition-duration: 200ms;
+}
+
+@media (prefers-reduced-motion) {
+	.selectedIndicator {
+		transition-duration: 0;
+	}
 }
--- a/src/components/tabs/Tabs.module.css
+++ b/src/components/tabs/Tabs.module.css
@@ -42,6 +42,12 @@
 	transition-duration: 200ms;
 }
 
+@media (forced-colors: active) {
+	.selectedIndicator {
+		background-color: ButtonText;
+	}
+}
+
 @media (prefers-reduced-motion) {
 	.selectedIndicator {
 		transition-duration: 0;
--- a/src/components/tabs/Tabs.module.css
+++ b/src/components/tabs/Tabs.module.css
@@ -15,15 +15,13 @@
 	display: none;
 }
 
-.tab,
-.tab:focus,
-.tab:active {
-	border-bottom: 0 !important;
-}
-
-.tab:focus-visible {
+.tab:focus {
 	outline: 2px solid var(--theme-accent);
 	outline-offset: -2px;
+}
+
+.tab:not(:focus-visible) {
+  outline: 0;
 }
 
 .tabpanel {
--- a/src/components/tabs/Tabs.module.css
+++ b/src/components/tabs/Tabs.module.css
@@ -47,8 +47,8 @@
 	}
 }
 
-@media (prefers-reduced-motion) {
+@media (prefers-reduced-motion: no-preference) {
 	.selectedIndicator {
-		transition-duration: 0;
+		transition-duration: 200ms;
 	}
 }
--- a/src/components/tabs/Tabs.module.css
+++ b/src/components/tabs/Tabs.module.css
@@ -34,6 +34,8 @@
 
 .selectedIndicator {
 	position: absolute;
+	transform-origin: left;
+	left: 0;
 	bottom: -4px;
 	height: 4px;
 	background-color: var(--theme-accent);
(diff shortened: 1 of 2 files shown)


sarah11918:
Alright we've given @zadeviggers the silent treatment long enough! Welcome back to `/pulls/`! It's as if you never left....  😄 

The preview looks good! Nice touch! I'm sure @Yan-Thomas and @delucis will have feedback for you soon.

Yan-Thomas (commented):
Thanks for this amazing and stylish 💅 contribution @zadeviggers, loved it! I added a few suggestions, let me know what you think! 🙌 💜

zadeviggers:
@Yan-Thomas thanks for those great sugegstions! I honestly just didn't notice the border thing (it was late and I was very tired when I implemented it), and had no idea that there was a way to detect high contrast mode, so TIL.

src/components/tabs/Tabs.tsx
```diff
@@ -22,71 +22,85 @@ function isPanelSlotEntry(entry: [string, ComponentChildren]): entry is [PanelSl
...
+
+	useLayoutEffect(() => {
+		// Fancy indicator animation
+		const activeTab = tabButtonRefs?.current[`tab.${curr}`];
+		if (activeTabIndicatorRef.current && tabButtonContainerRef.current && activeTab) {
+			console.log('A');
+			const tabBoundingRect = activeTab.getBoundingClientRect();
+			const containerBoundingRect = tabButtonContainerRef.current.getBoundingClientRect();
+			activeTabIndicatorRef.current.style.width = tabBoundingRect.width + 'px';
+			activeTabIndicatorRef.current.style.left = tabBoundingRect.left - containerBoundingRect.left + 'px';
```
delucis, line 74: I was worried that animating `left` and `width` had the potential to cause performance issues. This is because changing both of these properties triggers changes to the element geometry and therefore layout calculations. ([CSS Triggers](https://csstriggers.com/) is a useful page for checking this kind of thing.)

After some performance testing, I could see a fair number of frames that are slightly over their time budget during the animation:

image

It’s not hugely slow (0.1ms over budget), but could be worse on lower powered devices or if something else were happening simultaneously maybe.

We can fix this by switching to using a `transform` to modify the indicator’s position instead:

```suggestion
			if (!activeTabIndicatorRef.current.style.width) activeTabIndicatorRef.current.style.width = '1px';
			activeTabIndicatorRef.current.style.transform = `translateX(${tabBoundingRect.left - containerBoundingRect.left}px) scaleX(${tabBoundingRect.width})`;
```

This relies on a bit of extra CSS I’ll post in my next comment, but basically we set `width` to a static value of `1px`, and then control the actual width using a scale — still the same maths as before but more performant! Similarly `translateX` replaces our `left` property.

src/components/tabs/Tabs.module.css
```diff
@@ -7,19 +7,49 @@
...
 	padding-block-start: 1rem;
 }
+
+.tablist {
+	position: relative;
+}
+
+.selectedIndicator {
+	position: absolute;
+	bottom: -4px;
```
delucis, line 39: We need a couple of extra CSS properties to prepare our indicator for the `transform` shown in the previous comment.

```suggestion
	transform-origin: left;
	left: 0;
	bottom: -4px;
```

src/components/tabs/Tabs.module.css
```diff
@@ -7,19 +7,49 @@
...
+	height: 4px;
+	background-color: var(--theme-accent);
+	transition-duration: 200ms;
```
delucis, line 42: It would be better to move this behind a `prefers-reduced-motion: no-preference` query. That way we default to no transition and turn on the transition for users who are OK with it.

```suggestion
```

src/components/tabs/Tabs.module.css
```diff
@@ -7,19 +7,49 @@
...
+	.selectedIndicator {
+		background-color: ButtonText;
+	}
+}
+
+@media (prefers-reduced-motion) {
+	.selectedIndicator {
+		transition-duration: 0;
+	}
+}
```
delucis, line 55: And here’s that change:

```suggestion
@media (prefers-reduced-motion: no-preference) {
	.selectedIndicator {
		transition-duration: 200ms;
	}
}
```

delucis (changes requested):
Hey @zadeviggers! Thanks for your patience with this one. Had to handle a bunch of big bugs in our RTL layouts at the end of the week and just couldn’t get to this one in time.

I did see some minor performance issues with this implementation so I’ve suggested some fixes for that. Thanks for this fun addition! 😀

src/components/tabs/Tabs.tsx
zadeviggers: Good catch! I completely forgot about trqnsforms here!

src/components/tabs/Tabs.module.css
```diff
@@ -7,19 +7,49 @@
...
+.tab,
+.tab:focus,
+.tab:active {
+	border-bottom: 0 !important;
+}
+
+.tab:focus-visible {
 	outline: 2px solid var(--theme-accent);
 	outline-offset: -2px;
 }
```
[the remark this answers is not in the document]
zadeviggers: Fair enough. I generally tend to be happy with anything around ~90% on caniuse, but this is a good call for things that are important for accessibility.

src/components/tabs/Tabs.module.css
zadeviggers: Alright, this seems fine, but I'm not sure that this really changes anything. The end result will always end up being the same, no matter which way we structure the CSS, right?

zadeviggers:
Alright @delucis, thanks for those changes. They all seemed fine so I've gone ahead and apllied them all.

delucis (approved):

PR #1251, opened 2022-08-10
Note baseUrl is required for aliases

bluwy:

- New or updated content

#### Description

Closes https://github.com/withastro/astro/issues/3729

Added a note block to show `baseUrl` is required to be set for import aliases to work. Astro would skip aliases if not set based on https://github.com/withastro/astro/blob/e918b3883e156a0de2148517b619a2cf451917d2/packages/astro/src/vite-plugin-config-alias/index.ts#L39.

### DIFF
--- a/src/pages/en/guides/aliases.md
+++ b/src/pages/en/guides/aliases.md
@@ -32,6 +32,10 @@ You can add import aliases from either `tsconfig.json` or `jsconfig.json`.
 }
 ```
 
+:::note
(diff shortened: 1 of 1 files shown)


Yan-Thomas:
LGTM, nice work @bluwy. This is your first contribution to the docs, right? Welcome :partying_face: :blue_heart:!

bluwy:
Thanks for the info! I thought to wait for a green approval before doing so, but I'll merge from now on if there's a LGTM.

PR #1256, opened 2022-08-10
How to create an index.html at root level via getStaticPaths()

jablonski:
See issue here: https://github.com/withastro/astro/issues/3990

- New or updated content

#### Description

Issue is discussed here: https://github.com/withastro/astro/issues/3990

### DIFF
--- a/src/pages/en/core-concepts/routing.md
+++ b/src/pages/en/core-concepts/routing.md
@@ -123,6 +123,10 @@ Matched parameters will be passed as a query parameter (`slug` in the example) t
 Rest parameters are optional by default, so `pages/post/[...slug].astro` could match `/post/` as well.
 :::
 
+:::tip
+If you want to create the index.html at root level with rest parameters (e.g. when using data from a CMS), you have to set the parameter to `undefined`.
+:::
+
 #### Example: Rest parameters
 
 For a real-world example, you can implement GitHub’s file viewer with the following named and rest paramenters:


delucis:
Thanks for this @jablonski. Was there a reason you closed this PR? I'm seeing it documented the recommended technique from https://github.com/withastro/astro/issues/3990

jablonski:
I've closed #3990, because I'm totally fine with _undefined_ as value instead of an empty string to create a top level index.html route, so nothing needs to be fixed. It was just a lack of documentation.

jablonski:
Now I understand the problem. This PR got closed automagically by Github because I've closed #3990. I cannot reopen it, but IMHO it should be merged to the docs.

sarah11918:
Thanks so much for this @jablonski! I followed your thread in Discord, and the Issue you created at withastro/astro for this. I think this is a great addtion to the docs, but maybe this could be expanded to be a bit more than a "tip" to really make sense in context. (And, it means we don't need two "tips" right next to each other.

I like how you presented your discovery of "undefined" in https://github.com/withastro/astro/issues/3990 . Would you be prepared to change this content from a tip to an actual, simple, small code example to illustrate creating the index.html route? I think that would be more clear for the reader of Docs. I think *just* the `export default getStaticPaths()` function that shows setting a parameter to "undefined" alongside some other params?

Let me know what you think! Don't be afraid to just write and suggest something, and we can all discuss further from there!

jablonski:
@sarah11918 I totally agree. 

Just a general thought: maybe it would help people to see self-contained examples in the docs which they can copy/paste to StackBlitz to try out things?

I've created an example, which should show how it should work in general. I'm not so sure about the wording (or the best place to put this example in the docs, because the example shows how to create the index.html at root level, but demonstrates also how to render content with dynamic paths (e.g. from a headless CMS).

Maybe this kind of content should be stored under a "recipes category" and linked from the docs?


```astro
---
  const pages = [
    {
      slug: undefined,
      title: "Astro Store",
      text: "Welcome to the Astro store!",
    },
    {
      slug: "products",
      title: "Astro products",
      text: "We have lots of products for you",
    },
    {
      slug: "products/astro-handbook",
      title: "The ultimative Astro handbook",
      text: "If you want to learn Astro, you must read this book.",
    },
  ];
  return pages.map(({ slug, title, text }) => {
    return {
      params: { slug },
      props: { title, text },
    };
  });
}

const { title, text } = Astro.props;
---

<html>
  <head>
    <title>{title}</title>
  </head>
  <body>
    <h1>{title}</h1>
    <p>{text}</p>
  </body>
</html>
```

jablonski:
I've created a new PR because I've deleted my fork already, so couldn't update it anymore:

https://github.com/withastro/docs/pull/1389

Sorry for the mess...

PR #1264, opened 2022-08-11
Adding an "unmaintained content" aside

sarah11918:
- New or updated content

This adds an "unmaintained content" aside to the top of a page that exists in Docs but is not currently linked to, as it is still being discovered by search engines. Since we haven't yet decided what to do with this page, this notice informs readers that it is simply unmaintained, not hidden for any nefarious reasons!
Screenshot 2022-08-11 09 13 40


Yan-Thomas:
LGTM! Thanks @sarah11918 🙌

PR #1273, opened 2022-08-12
With remarkReadingTime, extend remarkPlugins (not remove defaults)

mattgeowild:

Very minor fix. This simply makes it so the documentation will instruct the reader to *extend* their remark plugins, rather than overwrite the defaults.

Before:
```
import { remarkReadingTime } from './remark-reading-time.mjs';

export default {
  markdown: {
    remarkPlugins: [remarkReadingTime],
  },
};
```

After:
```
import { remarkReadingTime } from './remark-reading-time.mjs';

export default {
  markdown: {
    remarkPlugins: { extends: [remarkReadingTime] },
  },
};
```

#### Description

I ran into an issue with my tables not rendering, did some research, found that markdown doesn't include tables by default, but people tend to use GitHub-flavored markdown in order to add some extra features -- and my config didn't enable that, since I had directly followed the documentation on adding reading time. It just says to "apply this plugin to your config," with the code. If it'd be better to expand on that by describing extending/replacing, feel free to add that to this PR :)

### DIFF
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -459,7 +459,7 @@ import { remarkReadingTime } from './remark-reading-time.mjs';
 
 export default {
   markdown: {
-    remarkPlugins: [remarkReadingTime],
+    remarkPlugins: { extends: [remarkReadingTime] },
   },
 };
 


mattgeowild:
Just realized this is COMPLETELY wrong. `remarkPlugins` takes an array and I gave it an object. I was looking at @astrojs/mdx documentation when I made this. 💀 Not sure how to fix it then though, probably should've opened an issue, my bad y'all.

PR #1279, opened 2022-08-13
Replace all instances of `pnpx` with `pnpm dlx`

skra-pa-pa:

#### Description

In places when the docs show examples using `npx` and `yarn`, they showed `pnpx` for PNPM, which is [deprecated](https://pnpm.io/6.x/pnpx-cli) and replaced by `pnpm dlx` (or `pnpm exec` if the package is already installed--maybe the PR should use that instead?)


delucis:
Hi @skra-pa-pa — thanks so much for your contribution! We’re actually about to migrate away from recommending `npx`/`pnpx` entirely in favour of `pnpm run astro add ...`, so I’m going to close this, but I really appreciate you taking the time nevertheless! 🙌

PR #1280, opened 2022-08-13
Update why-astro.md

codymjarrett:
Maybe this isn't what you were going for but it's better than the grammatical error.

#### Description

Fixes weird grammar.


sarah11918:
Hi @codymjarrett, thanks for taking the time to submit a PR to the docs!

In this particular case, there's no grammatical error here and your suggestion does slightly change the meaning. So I'm going to close this PR now without applying a change. But, I am making a note of the feedback for our next review of this page that, although technically correct, it perhaps could read better.

I hope you'll keep contributing, and I hope you're enjoying building with Astro! 🚀

PR #1300, opened 2022-08-15
Use docs integration guides instead of GitHub links

delucis:

#### Description

Noticed we were still pointing to the adapter packages on GitHub from the SSR page. This updates those links to use the docs versions of those READMEs.


sarah11918:
❤️  LGTM, @delucis!

PR #1303, opened 2022-08-15
Add retry logic to fetch

delucis:

- Something else!

#### Description

Adds retry logic to our `githubGet` helper to try and mitigate the regular CI failures we get for “Invalid response body” when generating integration pages. Here’s a recent example: <https://github.com/withastro/docs/runs/7837529942?check_suite_focus=true#step:4:18>

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "htmlparser2": "^7.2.0",
     "kleur": "^4.1.5",
     "node-fetch": "^3.2.10",
+    "p-retry": "^5.1.1",
     "parse-numeric-range": "^1.3.0",
     "preact": "^10.10.1",
     "prettier": "^2.7.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -33,6 +33,7 @@ specifiers:
   kleur: ^4.1.5
   nanostores: ^0.5.13
   node-fetch: ^3.2.10
+  p-retry: ^5.1.1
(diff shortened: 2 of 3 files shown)


hippotastic (changes requested):
Hey, thanks a lot for working on improving this! The approach you're suggesting would result in hammering the API with consecutive requests without any timeout or randomization though.

I would prefer if we could use one of the popular battle-tested retry packages like https://www.npmjs.com/package/p-retry instead of implementing our own retry logic. What do you think?

delucis:
Thanks @hippotastic — yeah, this was quick and dirty, sorry! I’ll take a look at that suggestion. I’m also wondering if there’s a nice way to generalise this. We sometimes use `fetch` directly, other times `githubGet`.

natemoo-re (approved):

natemoo-re:
Ha I retract my approval! @hippotastic's suggestion is much better!

hippotastic:
I used `p-retry` now as discussed and tested it locally by simulating a connection error. It's retrying in exponentially increasing intervals for up to 5 times now. PTAL @delucis :)

hippotastic (approved):

hippotastic:
(Please disregard my self-approval. I tried to declare my old review with requested changes as outdated and this was the result. 😆)

natemoo-re (approved):
😎 Lookin' good!

PR #1305, opened 2022-08-15
Fix example usage for typing frontmatter

kezzughayyar-ardanlabs:

#### Description

interface `MarkdownInstance` has a property `frontmatter`. The interface doesn't embed the generic type into it's properties, so we need to call `frontmatter` attribute


delucis (approved):
Thanks for catching that @kezzughayyar-ardanlabs! 🙌

PR #1312, opened 2022-08-16
i18n(zh-cn): reverse the unnecessary translation

LiHowe:

- reverse the unnecessary code translation

#### Description

- The code block translation is not necessary, and it is strange to translate `slug` to `蛞蝓`.
- Link: https://docs.astro.build/zh-cn/guides/markdown-content/#%E5%AF%BC%E5%85%A5-markdown


delucis (approved):
Good catch! Agree the type the should be kept in English 👍

PR #1319, opened 2022-08-16
Migrating to Fathom Analytics

tony-sull:

- Something else!

#### Description

- Migrates from GTM to Fathom Analytics.  This is the same GDPR-compliant analytics tool used on the new astro.build site

Bonus: using the same Fathom project allows us to properly track users moving between the landing page and docs site all while respecting user privacy and international requirements 🚀

### DIFF
--- a/src/components/HeadCommon.astro
+++ b/src/components/HeadCommon.astro
@@ -45,14 +45,5 @@ import '@fontsource/ibm-plex-mono/400-italic.css';
 		document.getSelection().addRange(range);
 	});
 </script>
-<!-- Global site tag (gtag.js) - Google Analytics -->
-<script async src="https://www.googletagmanager.com/gtag/js?id=UA-130280175-15"></script>
-<script>
-	/* globals dataLayer */
-	window.dataLayer = window.dataLayer || [];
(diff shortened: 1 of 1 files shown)


delucis:
I’m seeing this error in Chrome Dev Tools: 
image
They link to these docs: <https://web.dev/samesite-cookies-explained/>

Do we need to move the script within docs?

matthewp:
Is that maybe a cookie you have cached? There shouldn't be a `_ga` cookie, that's google analytics. Unless this PR isn't completely removing it.

tony-sull:
Yeah that'd be an existing cookie from the current Google Analytics script, Fathom won't set a `_ga` cookie

Not sure the right way to handle that transition, would it make sense to add a custom header to the Netlify config to delete the cookie?

delucis (approved):
Ahhh, yup, my bad — testing in an incognito window shows no issues. Not sure we need do anything — those cookies will expire and this was only an issue because Lighthouse was flagging it, not actually causing a problem going forward I don’t think.

LGTM! Thanks so much @tony-sull!

PR #1332, opened 2022-08-17
Add example of conditional JSX expressions

### LINKED ISSUE #1267: Add conditional (if/then/else) HTML example to JSX expressions
Coming from a Vue world we often use `v-if` and `v-else` when deciding what to show/not show in HTML templates.  As I'm not super familiar with JSX expressions I was looking into the docs to see how I could do something similar with Astro.

I believe having some examples like the following in the documents would be helpful for those of us not familiar with JSX.  It feels like it could work well under https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html as an additional example.

Longhand
```js
---
const items = Astro.glob('./pages/events/*.md')
---
{() => {
  if(items.length > 0) (
    return <p>...</p>
  }
}}

```

Shorthand
```js
---
const items = Astro.glob('./pages/events/*.md')
---
{items.length > 0 && (
  <p>...</p>
)}
```

drewtownchi:
What kind of changes does this PR include?

- New or updated content
- Resolves #1267 

Description

- Documents how to conditionally show HTML in an Astro component


src/pages/en/core-concepts/astro-components.md
```diff
@@ -164,6 +164,32 @@ const El = 'div'
...
+---
+const item = true;
+---
+{item === true ? (
+    <p>Show me!<p>
+  ) : (
+    <p>Else show me if not true</p>
+  )
+}
+```
```
sarah11918, line 191:
``````suggestion
Astro can conditionally display HTML using JSX logical operators and ternary expressions.

```astro title="src/components/ConditionalHtml.astro" "item"
---
const visible = true;
---
{visible && <p>Show me!<p>}

{visible ? <p>Show me!<p> : <p>Else show me!</p>}
```
``````
Thanks for adding these great patterns, @drewtownchi! Very useful (I use these myself!) and a valuable addition.

I think these can be simplified for Docs purposes, since we try to display concise examples that point our readers towards "Astro-y" patterns. I think that showing the ternary pattern, but not a full "if/else" example, is probably sufficient enough for users to realize that they CAN display conditionally like this, and if they happen to be more comfortable with a full if/else, or want to account for more complicated logic, they would probably infer that they can write it that way.

So my suggestion would be to model this more something like the above, and we'll get some feedback from others, too!

drewtownchi:
Thanks Sarah! Those changes look good to me and are much more concise than my examples.

sarah11918:
Snazzy use of our highlighting feature, @drewtownchi!  💅  I'm just looking into why the Netlify display preview syntax highlighting doesn't look the way I'd expect it to... but if you're happy with these examples, then I'm confident we can get this into docs just as quickly as we figure out what's going on with that single white left brace!

sarah11918:
OK, looks like old "left brace" is a bigger problem than I'd hoped, so let's not let that stop us here! I'll make a separate issue to look into the formatting of these kinds of code samples.

Thanks for your first (of many?) contributions to the Docs, and welcome to the Face Pile!  🥳

PR #1335, opened 2022-08-17
Mention Response.redirect

ghostdevv:

- New or updated content

#### Description

Added a note about `Response.redirect` since `Astro.redirect` is not available in api routes and the property is not commonly known.

image


src/pages/en/guides/server-side-rendering.md
```diff
@@ -198,3 +198,25 @@ export async function post({ request }) {
...
+  const link = await getLinkUrl(id);
+
+  if (!link) {
+    return new Response(null, {
+      status: 404,
+      statusText: 'Not found'
+    });
+  }
+
+  return Response.rediret(link, 307);
```
sarah11918, line 220:
```suggestion
  return Response.redirect(link, 307);
```

sarah11918 (commented):
Thanks so much for this @ghostdevv! I remember the conversation from the Docs channel on this, and appreciate you putting this together so quickly!

I just fixed a quickie typo, and I'm going to ask @matthewp or @delucis for their review comments, here too. But, looking forward to getting you into the Docs soon!

delucis (commented):
This looks great! Thank you @ghostdevv!

I’m noticing you can’t do `Response.redirect('/page', 307)` because it expects a full URL. This makes things a little fiddlier to use for internal site redirects like `if (!loggedIn) Response.redirect('/login', 200)`. You need to add the full URL and potentially switch it to localhost during dev (and you don’t have access to `Astro.site` either).

Do we want to mention any of that? Maybe a note that `link` must be a fully qualified URL?

(Here’s my very simple test if someone wants to play: https://stackblitz.com/edit/github-rphchk?file=src/pages/api.ts)

ghostdevv:
Damn that's unfortunate :/ the places I have used this before must be patching `Response.redirect` I guess! Here is an updated copy

image

sarah11918:
Sorry, @ghostdevv! Normally we can get these merged more quickly, but we've had three Docs maintainers traveling and not entirely available for PRs this week. I just wanted to ping for freshness, and say we haven't forgotten about this, and @delucis will be back next week. Thanks for your patience!

src/pages/en/guides/server-side-rendering.md
```diff
@@ -198,3 +198,34 @@ export async function post({ request }) {
...
+```
+
+To do a local redirect you have to include the full URL, for example:
+
+```js title="src/pages/redirect.js"
+export async function get({ request }) {
+  const url = new URL(request.url);
+  return Response.redirect(`${url}/home`, 307);
+}
+```
```
delucis, line 231:
```suggestion
`Response.redirect` requires that you pass a full URL. For local redirects, you can use `request.url` as the base with [the `URL` constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) to build an absolute URL:

```js title="src/pages/redirect.js"
export async function get({ request }) {
  const url = new URL('/home', request.url);
  return Response.redirect(url, 307);
}
```
```

delucis (commented):
Whelp — I made it back, but missed this, sorry @ghostdevv!

I’ve made a small suggestion to build the url with the `URL` constructor directly (avoids worrying about where what slashes are etc.), but nice work on this one!

delucis (approved):
Thanks again for your patience — didn’t mean to _ghost_ you!

ghostdevv:
All good!

PR #1336, opened 2022-08-18
Fix heading level in Cloudflare deploy guide

timfee:
Troubleshooting section wasn't intended to be for SSR only, so hoisting the heading level!


delucis (approved):
Thanks for the quick follow-up @timfee! 🙌

PR #1348, opened 2022-08-19
Remove outdated translations

delucis:

- Deleted content 😢 

#### Description

This PR removes our translation/i18n support for Bulgarian, Bengali, Danish, Finnish, Hungarian, Italian, Korean, Dutch, and Polish. These have become quite out of date since we stopped accepting PRs outside our core language set.

- move the outdated Getting Started translations to our `old-translations` directory
- remove the relevant `ui`/`nav` files for these languages completely (all only contain 1 or 2 entries)
- remove these languages from `src/i18n/languages.ts` so they no longer show up in the language picker

Russian (`ru`) and Traditional Chinese (`zh-TW`) had both received more recent updates so I left those untouched but we can discuss if that’s the best decision or not.

---

In case anyone is disappointed to see this PR: please don’t think we don’t want to support more languages eventually! We just need to be intentional about how we scale up our translation activity to match community contributions and the amount of work maintainers have to put in to manage those. See [our i18n guide](https://github.com/withastro/docs/blob/0cac448a54ded79f0f696cf49d37fb2d0cea3077/TRANSLATING.md#why-can-astro-only-support-a-few-languages) for more details.


Yan-Thomas:
LGTM, thanks for tackling this @delucis 🙌

hippotastic:
LGTM! Sad to see the language selector shrink, but given the amount of changes Astro went through since the creation of most of these translations, I think it's better to archive them rather than showing outdated content to readers.

sarah11918:
LGTM! This isn't an end... it's a beginning! Laying the groundwork for these languages to eventually come back better than ever!

PR #1353, opened 2022-08-19
Improve support for RTL languages

delucis:

#### Description

This PR improves site layout & styling for RTL languages (currently only Arabic on our site).

The major thing is making sure the primary and secondary sidebars switch position so the main navigation is on the right as RTL users would expect.

Many more minor tweaks involved padding or margins in some components that needed to flip side depending on writing direction, or the chevrons indicating collapsed/open state in the nav menu.

One more QOL improvement: we can now mark a language as RTL in `src/i18n/languages.ts`. Up until now Arabic translators had to include `dir: rtl` in each page’s frontmatter. This is now a global setting, so all Arabic pages can use an RTL layout even if displaying fallback content which is LTR.


sarah11918:
Thanks for this, @delucis!

I tested some pages navigating to/from/between translated and non-translated pages. The only thing that I noticed that looked out of place to me, but maybe it's expected/OK in this context, is in the attached screenshot where English headings on non-translated pages that end in question marks have the punctuation leading the text.

Otherwise, I didn't spot anything out of place on desktop, and will look on my phone now!

Screenshot 2022-08-19 12 50 46

sarah11918:
Confirming phone view is fine, but medium-screen tablet view is wonky.

Screenshot 2022-08-19 12 57 16
Screenshot 2022-08-19 12 56 15

sarah11918:
LGTM! Both smaller Chrome window and tablet mobile view show the pages properly now, and question marks are where I'd expect them to be!

Yan-Thomas:
Thanks for this @delucis, great job! :raised_hands: :purple_heart:

There are a few issues I found, now the community menu links' SVGs are being inverted as well now, making the selector more specific should fix this. Also, the buttons on the Deploy page are occupying the same space:
image

Yan-Thomas:
LGTM!

PR #1359, opened 2022-08-20
Add note re: dynamic vs reactive

### LINKED ISSUE #864: `.jsx` vs `.astro` comparison should include the conceptual differences between the two
Currently, we have a [`.jsx` vs `.astro` comparsion](https://docs.astro.build/en/comparing-astro-vs-other-tools/#astro-vs-jsx) that covers the syntactic differences between the two languages.

This is helpful, but there's another difference that I want to point out.

JSX is used in app frameworks like React and Solid. There, JSX is used to include dynamically updated values. For example, in React you might see:

```jsx
<div>
  Your name is {name}
  <input type="text" onChange={updateName}/>
</div>
```

Here, `{name}` means: whenever the value of name changes, make sure it shows up here.
In Astro, `.astro` looks like JSX, and shares many of its features, but is not used for this key purpose.

I would like to make that clear in this section by explaining the purpose of the `.astro` syntax, which is to include dynamic values (_not_ reactive values).

Dynamic: we don't know what the value is, let's include it via a variable! It's not going to change, though: we run this _once_ (on the server).

Reactive: we don't know what the value is _and_ it's going to change.

sarah11918:
- New or updated content

Provides a quick note in the "Dynamic HTML" section stating that Astro allows you to include *dynamic* values, but not *reactive* values via varibles, then links to "More differences between Astro and JSX" further down the page.

Closes #864


src/pages/en/core-concepts/astro-components.md
```diff
@@ -111,6 +111,10 @@ const myFavoritePokemon = [/* ... */];
 
 You can define local JavaScript variables inside of the frontmatter component script within an Astro component. You can then inject these variables into the component's HTML template using JSX-like expressions!
 
+:::dynamic vs reactive
+Astro allows you to include include **dynamic values** (set as a variable's value once, as defined at build time), but **not reactive values** (regularly updated in response to a change in a variable's value). See below for more examples of [differences between Astro and JSX](#differences-between-astro-and-jsx).
+:::
```
Yan-Thomas, line 116:
```suggestion
:::note[dynamic vs reactive]
Astro allows you to include **dynamic values** (set as a variable's value once, as defined at build time), but **not reactive values** (regularly updated in response to a change in a variable's value). See below for more examples of [differences between Astro and JSX](#differences-between-astro-and-jsx).
:::
```

Yan-Thomas (commented):
Thanks for this @sarah11918, I found a typo and added `note [...]` so the note is correctly displayed, after these LGTM!

src/pages/en/core-concepts/astro-components.md
sarah11918: Thanks, as always, @Yan-Thomas!

src/pages/en/core-concepts/astro-components.md
```diff
@@ -111,6 +111,10 @@ const myFavoritePokemon = [/* ... */];
 
 You can define local JavaScript variables inside of the frontmatter component script within an Astro component. You can then inject these variables into the component's HTML template using JSX-like expressions!
 
+:::note[dynamic vs reactive]
+Astro allows you to include **dynamic values** (set as a variable's value once, as defined at build time), but **not reactive values** (regularly updated in response to a change in a variable's value). See below for more examples of [differences between Astro and JSX](#differences-between-astro-and-jsx).
+:::
```
Jutanium, line 116:
```suggestion
Using this approach, you can include dynamic values that are calculated in the frontmatter.  But once included, the values will never change, because Astro components are templates that run once. 

See below for more examples of [differences between Astro and JSX](#differences-between-astro-and-jsx).
:::
```

Jutanium:
Suggested a change to make the sentences smaller and take out "reactive" because the term isn't necessary if we're only doing a quick aside.

But I'm not too attached to the wording - any note along these lines is great!

src/pages/en/core-concepts/astro-components.md
```diff
@@ -111,6 +111,12 @@ const myFavoritePokemon = [/* ... */];
 
 You can define local JavaScript variables inside of the frontmatter component script within an Astro component. You can then inject these variables into the component's HTML template using JSX-like expressions!
 
+:::note[dynamic vs reactive]
+Using this approach, you can include dynamic values that are calculated in the frontmatter.  But once included, these values will never change. Astro components are templates that only run once, at build time.
+
+See below for more examples of [differences between Astro and JSX](#differences-between-astro-and-jsx).
```
sarah11918, line 117:
```suggestion
:::note[dynamic vs reactive]
Using this approach, you can include **dynamic** values that are calculated in the frontmatter.  But once included, these values are not **reactive** and will never change. Astro components are templates that only run once, at build time.

See below for more examples of [differences between Astro and JSX](#differences-between-astro-and-jsx).
```

PR #1360, opened 2022-08-20
Expand the Dynamic HTML section

### LINKED ISSUE #1276: Add: Explain how to render components via variables
I just spent a lot of time trying to figure all this stuff out. Maybe it's worth adding it to the docs?

## 1. First letter upper case convention

Variable names need to be capitalized correctly. (I understand it's a React/JSX convention) If they're not, component doesn't render. As such, this piece of code just creates a `<stored/>` element instead of the component stored in the variable.

```astro
---
// Store component into a variable
let stored = Component;
---
<!-- Render -->
<stored/>
```

The "s" in "stored" needs to be upper case. Would it be worth it adding that warning somewhere near the [Dynamic HTML](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) section?

## 2. Storing and rendering

My intuitive understanding of storing a component was this.

```astro
---
// Store component into a variable
let Stored = <Component/>;
---
<!-- Render -->
<Stored/>
```

Of course that doesn't work, because JSX isn't allowed in frontmatter scripts. Instead this is correct.

```astro
---
// Store component into a variable
let Stored = Component;
---
<!-- Render -->
<Stored/>
```

But again, as far as I can tell, the docs have no mention of this.

---

Would it be worth adding all this info?
Maybe provide an expanded example of storing components into variables and rendering them?
Let me know what you think. Thanks!

swift502:

- New or updated content

#### Description

Closes #1276

Adding two brief explanations:

- Dynamic tag variable names must be capitalized in order for them to work
- An example of rendering an imported component that's stored in a variable first

First iteration of this. Few more changes I considered:

- maybe the TIP window isn't necessary anymore? It seems like a reasonably substantive example at this point.
- change the window title to "src/components/DynamicTags.astro" to fit better with the examples above and below this one?
- change the "El" variable name to something more fitting. Maybe something like "DynamicTag"?

Let me know what you think!
\- Jan


Jutanium:
I love this suggestion; I adapted it a bit using your notes above.

> maybe the TIP window isn't necessary anymore? It seems like a reasonably substantive example at this point.
Agreed, and I think `Dynamic Tags` is now enough to deserve its own section!

> change the window title to "src/components/DynamicTags.astro" to fit better with the examples above and below this one?
Done!

> change the "El" variable name to something more fitting. Maybe something like "DynamicTag"?
I changed the example to be similar to the previous `.map` one. This:
- removes the triviality of renaming an import just to use a dynamic tag
- Suggests a natural  "Tag" variable name within the `.map`

Yan-Thomas:
Thanks for this @swift502, very good job! Also thanks to @Jutanium for the improvements, LGTM!

src/pages/en/core-concepts/astro-components.md
```diff
@@ -175,6 +163,34 @@ const visible = true;
...
+
+This also works with components:
+```astro title="src/components/DynamicTags.astro" /BigRedHeader|Tag|<Tag>|</Tag>/
+---
+import BigRedHeader from "./BigRedHeader.astro";
+const tags = ["h1", "h2", "h3", BigRedHeader];
+---
+
+<div>
+    Three sizes of headers:
```
sarah11918, line 189:
```suggestion
    Four different headers:
```
?? I'm honestly not sure this example works, because I have no idea what size the BigRedHeader is and whether it's the same size as one of the existing ones. But, this is going to give you four headers, is it not? And without knowing what the BigRedHeader actuually looks like, maybe they're not all in fact different?

The example kind of feels forced, and I'd almost prefer to see something entirely different that doesn't include any tags, but maybe just maps over an array of related components to make the point.

src/pages/en/core-concepts/astro-components.md
```diff
@@ -175,6 +163,34 @@ const visible = true;
 {visible ? <p>Show me!<p> : <p>Else show me!</p>}
 ```
 
+### Dynamic Tags
+
+Local variables can also represent dynamic HTML _tags_:
+
+```astro title="src/components/DynamicTags.astro" /Tag|<Tag>|</Tag>/
+---
+import MySpecialHeader from "../MySpecialHeader.astro"
```
sarah11918, line 172:
```suggestion
```
There doesn't seem to be a reason to include this import here, as it's never used in the example? But I do like the rest of the example!

sarah11918 (commented):
Thanks @Jutanium! Reviewed! First example good! Second example... I'm not convinced? Do you want to convince me?  :)

src/pages/en/core-concepts/astro-components.md
Jutanium: Totally was supposed to say `Four different headers` and that first example had a import that wasn't supposed to be there 🙈. And I've got a less forced example up my sleeve!

Jutanium:
@sarah11918 agree with your comments. The simplest _practical_ example I can think of is rendering a tag based on a prop.

That implies that the reader knows what props are. Even though that comes later in the document, I feel justified using it here because this is a reference document, but I want to hear your thoughts on that.

Jutanium:
Also: I [confirmed](https://discord.com/channels/830184174198718474/1012523253086048306) that it's okay to use `any` here.

sarah11918:
@Jutanium What was wrong with the original proposal?  😅  

I think now we miss the simplicity of the existing example, showing the HTML render, and the corresponding way it works with components.

Jutanium:
I wanted to motivate the feature more. In the original example, it would have been better in practice to use a div directly, and rename the component when you import it, so I read it as a “bad practice” example.

But it’s a good example in that it very clearly shows the mechanism we’re describing! So I’m fine to go back to it.

swift502:
Thanks @Jutanium, however I think you accidentally removed the only two things I was trying to convey. 😄

Namely that:
- variables must be capitalized in order to work. It might seem obvious to people experienced with JSX/React, but it took me (and 2 of my coworkers) hours of asking on Discord, googling and trial and error to finally figure it out. My goal was to have any mention of that implicit rule here.
- an example of how to store a component __into a variable__
In my project I needed to do that, and I intuitively did this:
```astro
---
const component = <Component/>
---
```
That's of course wrong, and throws a generic syntax error message, and I was once again asking on Discord and googling. So my goal was to have an example of how to do this.

And I must agree with @sarah11918. The previous example, while maybe bad practice, showed very, very clearly what Astro is _capable_ of, and the correct syntax for it.

---

That being said, something along the lines of `<Pokemon Link={CustomLinkComponent} />` example might be a good addition too, I don't see it anywhere on the rest of the page. But I'd again be in favor of making it as concise as possible. Not necessarily show a complicated real-world use case, but just that Astro is capable of passing components via props.

swift502:
At this point I propose to discard the second new example, bring back an example of storing components as variables:

```astro
---
import Pokemon from '../components/Pokemon.astro'
const StoredComponent = Pokemon;
---
<StoredComponent /> <!-- Renders as <Pokemon /> -->
```


And add a note along the lines of:

:::note Variables must be capitalized to work as dynamic tags! Lowercase variables will only render as regular HTML tags, and ignore the data you assigned to them. :::

Jutanium:
@swift502 You're totally right - I accidentally removed that disclaimer. And thanks for your patience and elaboration, I have a better understanding of what you were showcasing!

I agree with both of you that your original proposal is the simplest option to convey the mechanics of both of the behaviors. I pushed a commit that goes back to that, and still tries to incorporate your last three points.

There's an argument for showing an example where we pass props through the dynamic tag, and also an argument for showing how to take the tag as a prop. But that would be more "how-to" content, whereas this is intended as a reference, so I'll be stashing my examples for another day!

swift502:
Yep, this is what I was trying to achieve. 🙂
As long as the section states (at least) these things in some form I'm fine with merging.

swift502:
> There's an argument for showing an example where we pass props through the dynamic tag

I'm not against that. Though it does combine a lot of things. It feels like it could be in the Dynamic Attributes section, this new Dynamic Tag section, and even the Component Props section. So I'm not at all sure where it should end up.

Jutanium:
Right - if we wanted to add it, I think the section it would end up in depends on whether this doc assumes knowledge of the previous sections as it goes along. We've been thinking a lot about [content types](https://diataxis.fr/) and restructuring, which would influence that decision, so I think we should keep this PR to this simple example for now.

sarah11918:
I LIKE IT! Do we have both explicit LGTMs from @swift502 and @Jutanium? I approve merging, so Dan, please feel free in the event that I'm in the air when we have explicit confirmation from everyone.  :)

swift502:
Yep, LGTM.

PR #1368, opened 2022-08-20
Missing frontmatter

phocks:
post.title doesn't work has to be post.frontmatter.title


Yan-Thomas:
Thanks for this @phocks, good job. Welcome to the docs! 💜

PR #1371, opened 2022-08-21
Improve configuration reference warning and edit this page link

Yan-Thomas:

#### Description

This PR changes the configuration reference "edit this page" link to the proper Astro type definitions file instead of the docs generated file that shouldn't be changed. Also, a dev-only warning as added so none of us forget about this:

image

Note: The comments we have in the file don't show up in GitHub's Markdown Preview, so it's easy to don't see them. Now, having the weird `<DontEditWarning/>` we have an extra security/warning layer.

Observations:
- Before, we added `README.md` to pages' "edit this page" link with the custom `githubURL` prop by default. Instead of doing this, I opted for adding the `hasREADME` frontmatter prop so we can opt-in/out of this behavior as we please, we can sure discuss what could be the best approach here.
- I tried a few methods (slots, `<a>` tag directly, Markdown, etc.) to have a link to the correct file inside the aside, but all of these I tried caused errors, so I opted for just saying that you should click the edit this page link instead.

### DIFF
--- a/scripts/docgen.mjs
+++ b/scripts/docgen.mjs
@@ -8,13 +8,18 @@ const STUB = ``; // fs.readFileSync('/PATH/TO/MONOREPO/astro/packages/astro/src/
 const HEADER = `---
 # NOTE: This file is auto-generated from 'scripts/docgen.mjs'
 # Do not make edits to it directly, they will be overwritten.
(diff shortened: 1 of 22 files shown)


delucis (approved):
LGTM! Thanks, Yan. Let's see who still falls into the trap of editing one of these pages 😜

PR #1377, opened 2022-08-21
Remove CSS-in-JS reference

Jutanium:
Astro doesn't support CSS-in-JS for `.astro` files, and it only supports _some_ CSS-in-JS for framework components. This PR removes the mention of CSS-in-JS.

We may also want to outline the extent of its support in framework components, but for now this removes a misleading statement.


sarah11918:
I agree. I think the statement ends up hurting more than helping. LGTM!

PR #1396, opened 2022-08-23
Mention that directives are supported in .mdx

calebeby:

#### Description

The `.mdx` page already mentions that this is supported, but I figured it was worth mentioning it in the directives page since it currently just says directives are supported in `.astro` files which misled me.

### DIFF
--- a/src/pages/en/reference/directives-reference.md
+++ b/src/pages/en/reference/directives-reference.md
@@ -4,7 +4,7 @@ title: Template Directives Reference
 i18nReady: true
 ---
 
-**Template directives** are a special kind of HTML attribute available inside of any Astro component template (`.astro` files).
+**Template directives** are a special kind of HTML attribute available inside of any Astro component template (`.astro` files), and in `.mdx` files.
 
 Template directives are used to control an element or component's behavior in some way. A template directive could enable some compiler feature that makes your life easier (like using `class:list` instead of `class`). Or, a directive could tell the Astro compiler to do something special with that component (like hydrating with `client:load`).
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @calebeby, thanks for this!  You're right, in order to use a UI framework component in MDX, you do need to use a client directive. And, *some* of the other template directives work, but not *all* of them. So, I'm hesitant to say at the top of the page such a general statement as "these work in MDX."

Was it a client directive on a framework component you ended up using in MDX? Because we could then instead try to mention *that* if not having it there is misleading.

As for the other directives, I think the intention is for them to be used in `.astro` files, and if some happen to work in MDX, that might be a coincidence. So, we also don't want to document something that isn't a planned, supported feature.

Given all that, and your experience using this page while working with Astro and what you were looking for, do you have an idea for a more specfic, true statement we could use here?

calebeby:
Yeah it was a client directive. I can adjust the change to make it more specific.

src/pages/en/reference/directives-reference.md
```diff
@@ -4,7 +4,7 @@ title: Template Directives Reference
 i18nReady: true
 ---
 
-**Template directives** are a special kind of HTML attribute available inside of any Astro component template (`.astro` files).
+**Template directives** are a special kind of HTML attribute available inside of any Astro component template (`.astro` files), and in `.mdx` files.
```
sarah11918, line 7:
```suggestion
**Template directives** are a special kind of HTML attribute available inside of any Astro component template (`.astro` files), and some can also be used in `.mdx` files.
```
Do you think it works if we just say this @calebeby?

calebeby: Yeah 👍

sarah11918:
Woo! And we're good! Thanks for this helpful first-time contribution, @calebeby! Welcome to Team Docs and the Face Pile!  🥳

PR #1397, opened 2022-08-24
Note Astro.request.url search parameters

bluwy:

- New or updated content

#### Description

Close https://github.com/withastro/astro/issues/4361

Note that search parameters are not available in SSG, but available in SSR

### DIFF
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -142,7 +142,7 @@ const data = await Astro.glob<CustomDataFile>('../data/**/*.js');
 See also: [`Astro.url`](#astrourl)
 
 :::note
-With the default `output: 'static'` configuration, `Astro.request.url` does not contain search parameters, like `?foo=bar`, as it's not possible to determine them ahead of time during static builds. However in `output: 'server'` mode, `Astro.request.url` does contain search parameters as it can be determined from a server request.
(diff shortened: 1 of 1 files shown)


src/pages/en/reference/api-reference.md
```diff
@@ -140,6 +140,11 @@ const data = await Astro.glob<CustomDataFile>('../data/**/*.js');
 ```
 
 See also: [`Astro.url`](#astrourl)
+
+:::note
+With the default `output: 'static'` configuration, `Astro.request.url` does not contain search parameters, like `?foo=bar`, as it's not possible to determine them ahead of time during static builds. However in `output: 'server'` mode, `Astro.request.url` does contain search parameters as it can be determined from a server request.
```
Yan-Thomas, line 145: Nitpick: In the Configuration Reference we call `output` an option, rather than a configuration, so I think we should change this for consistency. Though, I would love to know if there's any argument to choose configuration over option here.

```suggestion
With the default `output: 'static'` option, `Astro.request.url` does not contain search parameters, like `?foo=bar`, as it's not possible to determine them ahead of time during static builds. However in `output: 'server'` mode, `Astro.request.url` does contain search parameters as it can be determined from a server request.
```

Yan-Thomas (commented):
Good job @bluwy :purple_heart:, I added a small nitpick suggestion.

src/pages/en/reference/api-reference.md
bluwy: `option` sounds good to me!

Yan-Thomas (approved):
LGTM!

PR #1404, opened 2022-08-25
Add remark rehype options to markdown content docs

gtnbssn:

- New or updated content

#### Description

It looks like https://github.com/withastro/astro/pull/4138 is going to make it in the next minor release, so i'm reopening this PR in order to document the new options!

That Astro PR adds the possibility to use remark rehype options in the Astro config, for instance if a user wants to translate automatically generated elements related to the footnotes. This Astro docs PR adds the relevant docs.

### DIFF
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -556,3 +556,50 @@ When using Prism, you'll need to add a stylesheet to your project for syntax hig
 3. Loading this [into your page's `<head>`](/en/core-concepts/astro-pages/#page-html) via a `<link>` tag.
 
 You can also visit the [list of languages supported by Prism](https://prismjs.com/#supported-languages) for options and usage.
+
+### Remark-rehype options and footnotes accessibility
+
+[GitHub-flavored Markdown](https://github.com/remarkjs/remark-gfm) allows you to author footnotes. For accessibility reasons, this will add an `h2` element with the word "Footnotes", as well as `aria-label`s in the back links that say "Back to content". You might want to translate this to the language of your site. The Markdown is transformed into HTML through remark-rehype which has [a number of options](https://github.com/remarkjs/remark-rehype#options), including some to translate this content.
+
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks for taking the initiative on this @gtnbssn! I think @bholmesdev has added documentation for this (so many Ben PRs, I can't keep track!) so I'm going to ask him to look at this to see if there's any key information here that didn't get captured in his other documentation.

As a more general comment, I think this might a bit large of an example for docs the way they are right now, but it would make a great personal blog post! We have been discussing internally how best to include content like these examples which are explaining how to implement a specific feature. If you have a blog, and can write a post about this yourself there, then you don't have to worry about conforming to our current docs format and structure. And then, as we make space in the docs for this kind of community how-to-guide content, we could look at including it at that time!

bholmesdev:
Thanks @sarah11918! It looks like this is documenting our new `remarkRehype` config option, which my docs PRs do _not_ address. I agree this content is a bit more detail that our general "Markdown" page should cover though. I definitely learned something from that `.sr-only` callout though 🙃 

I'll leave the call to you, but the content is well written as it stands!

gtnbssn:
Hi @sarah11918 ! reading through this again, i admit the CSS part might not be at its place here. It's related to this feature, but it doesn't document how to use it. I'll remove it!

Unfortunately i do not have a blog yet. I definitely learned something about the `sr-only` class while doing this. It seems it is such standard practice that it is a utility class in tailwind.

gtnbssn:
Here's a trimmed down version!

sarah11918:
Thanks @gtnbssn! Admittedly, this looks a *lot* more manageable now!  😅 

I'm thinking maybe it even goes higher, up as the last example in the Markdown Plugins section? Would you agree? That would mean making it the third `####` heading in that section. Any objection to that?

bholmesdev:
@sarah11918 That makes sense to me!

gtnbssn:
No objection!

PR #1409, opened 2022-08-25
New `remark` and `rehype` plugin explainer with `extends`

bholmesdev:

- New or updated content

#### Description

We have a new way to preserver Astro's default remark plugins when you apply your own: `extends`! https://github.com/withastro/astro/pull/4474.

This PR addresses a couple concerns:
- Add short explainers on remark and rehype based on [the MDX docs](https://docs.astro.build/en/guides/integrations-guide/mdx/#remarkplugins). New markdown users _probably_ don't know what these plugins can do or where to find a list of popular options (I definitely didn't a month ago). This should make plugins easier to understand.
- Move tip on Astro's defaults to in-line explainer
- Replace `remark-autolink-headings` example with `remark-toc`. Autolink headings is the one plugin we _shouldn't_ recommend, since it conflicts with Astro's own heading ID plugin 😅 
- Add new code examples with and without `extends`

### DIFF
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -366,54 +366,29 @@ The instructions below are for configuring standard Markdown. To configure MDX p
 
 ### Markdown Plugins
 
-Astro supports third-party [remark](https://github.com/remarkjs/remark) and [rehype](https://github.com/rehypejs/rehype) plugins for Markdown. These plugins allow you to extend your Markdown with new capabilities, like [auto-generating a table of contents](https://github.com/remarkjs/remark-toc), [applying accessible emoji labels](https://github.com/florianeckerstorfer/remark-a11y-emoji), and more. We encourage you to browse [awesome-remark](https://github.com/remarkjs/awesome-remark) and [awesome-rehype](https://github.com/rehypejs/awesome-rehype) for some full curated lists!
+Astro supports third-party [remark](https://github.com/remarkjs/remark) and [rehype](https://github.com/rehypejs/rehype) plugins for Markdown. These plugins allow you to extend your Markdown with new capabilities, like [auto-generating a table of contents](https://github.com/remarkjs/remark-toc), [applying accessible emoji labels](https://github.com/florianeckerstorfer/remark-a11y-emoji), and more. We encourage you to browse [awesome-remark](https://github.com/remarkjs/awesome-remark) and [awesome-rehype](https://github.com/rehypejs/awesome-rehype) for popular plugins!
 
-You can apply these plugins in your Astro config:
+This example applies the [remark-toc](https://github.com/remarkjs/remark-toc) and [rehype-minify](https://github.com/rehypejs/rehype-minify) plugins. See each project's README for installation instructions.
 
-<Tabs client:visible>
-<Fragment slot="tab.1.withDefaults">With defaults</Fragment>
-<Fragment slot="tab.2.withoutDefaults">Without defaults</Fragment>
-<Fragment slot="panel.1.withDefaults">
+:::tip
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/markdown-content.md
```diff
@@ -364,51 +366,54 @@ The instructions below are for configuring standard Markdown. To configure MDX p
...
+You can apply these plugins in your Astro config:
 
-By default, Astro comes with [GitHub-flavored Markdown](https://github.com/remarkjs/remark-gfm) and [remark-smartypants](https://github.com/silvenon/remark-smartypants) pre-enabled.
-:::
+<Tabs client:visible>
+<Fragment slot="tab.1.withDefaults">With defaults</Fragment>
+<Fragment slot="tab.2.withoutDefaults">Without defaults</Fragment>
+<Fragment slot="panel.1.withDefaults">
+
+We apply the [GitHub-flavored Markdown](https://github.com/remarkjs/remark-gfm) and [Smartypants](https://github.com/silvenon/remark-smartypants) remark plugins by default. This brings some niceties like auto-generating clickable links from text (ex. `https://example.com`) and formatting quotes for readability. To apply your own plugins while preserving these defaults, use a nested `extends` object like so:
```
sarah11918, line 378: Just checking, is `https://example.com` supposed to be clickable when rendered to the Docs page as an example of what actually happens? Because it's not, and that seems... weird? since we're talking about clickable links?

Maybe an example of a link showing up as clickable, and using https://astro.build might make more sense? Or, I'd honestly just lose the example, I think.

Also, I might make it explicit that the example is adding the rehymeMinifyHTML plugin, otherwise it's at first glance a little unintuitive that it is highlighted, vs just the "extends." Maybe something like:

``` 
To apply your own plugins while preserving these defaults, use a nested `extends` object as shown below when adding `rehypeMinifyHtml`:
```

You also have the option of highlighting JUST the extends object, if you like. (I'll let you decide.) Instead of highlighting that whole line, using just quotation marks around the text you'd like highlighted. It might show the object off nicely!

src/pages/en/guides/markdown-content.md
```diff
@@ -364,51 +366,29 @@ The instructions below are for configuring standard Markdown. To configure MDX p
...
 
-:::note
-Enabling custom `remarkPlugins` or `rehypePlugins` will remove these built-in plugins and you need to explicitly add these plugins if desired.
+This example applies the [remark-toc](https://github.com/remarkjs/remark-toc) and [rehype-minify](https://github.com/rehypejs/rehype-minify) plugins. See each project's README for installation instructions.
 
-By default, Astro comes with [GitHub-flavored Markdown](https://github.com/remarkjs/remark-gfm) and [remark-smartypants](https://github.com/silvenon/remark-smartypants) pre-enabled.
+:::tip
+We apply the [GitHub-flavored Markdown](https://github.com/remarkjs/remark-gfm) and [Smartypants](https://github.com/silvenon/remark-smartypants) plugins by default. This brings some niceties like generating clickable links from text and formatting quotes for readability. When adding your own plugins, you can preserve these defaults with the `extendDefaultPlugins` flag.
```
sarah11918, line 374:
```suggestion
Astro applies the [GitHub-flavored Markdown](https://github.com/remarkjs/remark-gfm) and [Smartypants](https://github.com/silvenon/remark-smartypants) plugins by default. This brings some niceties like generating clickable links from text and formatting quotes for readability. When adding your own plugins, you can preserve these defaults with the `extendDefaultPlugins` flag.
```
I think this just makes it SUPER clear that it's not "we" like, "in this example we" and is instead Astro defaults.

sarah11918:
LGTM, Ben!

Yan-Thomas:
Good job @bholmesdev, LGTM!

PR #1416, opened 2022-08-26
Update UI Frameworks docs to include `alpineJS` integration

### LINKED ISSUE #1415: Update UI Framework Components page for AlpineJS integration
In the Installing Integrations section of this this page, we still say that no integration is needed for AlpineJS, and because there was no integration, we don't include it with all the others in the config code example.

1. Remove the statement about not needing an integration.
2. Add the integration in both relevant places in the code sample. (Write an import statement, and include it in the integrations array.)

https://docs.astro.build/en/core-concepts/framework-components/#installing-integrations

Screenshot 2022-08-26 11 24 54

mohammed-elhaouari:

- New or updated content

#### Description

- Closes #1415

### DIFF
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -11,7 +11,7 @@ Astro supports a variety of popular frameworks including [React](https://reactjs
 
 ## Installing Integrations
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Looks perfect! Thanks for being so quick to take this on, @mohammed-elhaouari 🥳

PR #1421, opened 2022-08-27
docs(image): add note about error if alt is missing

DerYeger:

- New or updated content

#### Description

- Adds docs for changes of https://github.com/withastro/astro/pull/4511
- Also see https://github.com/withastro/astro/issues/4291

### DIFF
--- a/src/pages/en/guides/images.md
+++ b/src/pages/en/guides/images.md
@@ -119,10 +119,10 @@ Astro's `<Image />` and `<Picture />` components cannot be used with images in y
 :::
 
 :::caution
-Astro's `<Image />` and `<Picture />` components will throw an error if no `alt` text is provided.
-If no an accessible label is required, set `alt=""`.
-:::
+Astro's `<Image />` and `<Picture />` components will throw an error if no `alt` text is provided. 
 
+If the image is merely decorative (i.e. doesn't contribute to the understanding of the page), set `alt=""` so it is ignored by screen readers.
+:::
 ### `<Image />`
 
 Astro's [`<Image />` component](/en/guides/integrations-guide/image/#image-) allows you to optimize a single image and specify width, height, and/or aspect ratio. You can even transform your image to a particular output format, which can be used to avoid checking the file type of remote images. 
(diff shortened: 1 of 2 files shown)


src/pages/en/guides/images.md
```diff
@@ -118,6 +118,11 @@ After [installing the integration](/en/guides/integrations-guide/image/#installa
 Astro's `<Image />` and `<Picture />` components cannot be used with images in your `public/` folder. Use standard HTML or Markdown image syntax instead.
 :::
 
+:::caution
+Astro's `<Image />` and `<Picture />` components will throw an error if no `alt` text is provided.
+If no an accessible label is required, set `alt=""`.
+:::
+
```
Yan-Thomas, line 125: Readers may be confused about when no accessible label is required, so, what do you think of making this a bit more explanatory?

```suggestion
:::caution
Astro's `<Image />` and `<Picture />` components will throw an error if no `alt` text is provided. 

If the image is merely decorative (i.e. doesn't contribute to the understanding of the page), set `alt=""` so it is ignored by screen readers.
:::
```

Yan-Thomas (commented):
Thanks for this @DerYeger, good job! I added a suggestion, let me know what you think. :purple_heart: 

Also, you unfortunately ended up changing the `integrations-guide/image.md` file which is generated by a script that picks up changes from the integrations' READMEs. These changes should be made in the [`@astrojs/image`'s README file](https://github.com/withastro/astro/blob/main/packages/integrations/image/README.md) instead, could you please remove them from this PR (leaving only the `images.md` changes here) and make a new PR in the main Astro repo? I'll be sure to review the PR there as well!

DerYeger:
Thank you for the feedback @Yan-Thomas! I like you're more detailed explanation much more.

I will update my feature PR with the README change.

sarah11918:
Great, glad it meets with your approval, @Yan-Thomas!  💜 

@DerYeger I will merge this, and I'm going to look at the one in `withastro/astro` now. I'm less concerned about this text being identical there, but it also would not bother me if THAT integration README were updated to match. (We are planning on recombining some text anyway, so whatever happens there might change anyway...)  It's HERE ON THIS PAGE that I'm mostly concerned about, and I think this is great now!

PR #1422, opened 2022-08-27
fix(component-template-doc): add closing tag to ul

juansemprun:

- Minor content fixes

#### Description
Add closing tag to **ul** on the component template example in the files that was available 
https://docs.astro.build/en/core-concepts/astro-components/#the-component-template

- en
- de
- pt-br
- es
- zh-cn
- fr
- ja

#### Before
image

#### After
image

### DIFF
--- a/public/theme.css
+++ b/public/theme.css
@@ -195,6 +195,7 @@
 	--docsearch-modal-shadow: 0 0 var(--theme-glow-blur) var(--theme-glow-diffuse), inset 0 0 0 1px var(--theme-glow-highlight),
 		inset 0 0 var(--theme-glow-blur) var(--theme-glow-diffuse);
 	--docsearch-container-background: hsla(var(--color-gray-10), 0.8);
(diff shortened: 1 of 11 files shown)


Yan-Thomas:
Thanks for this @juansemprun, good catch! Just a thing, could you please remove the changes to the non-EN pages? We have a system in place that indicate which pages are outdated and the proper commit history of changes so translators can easily update them! If we did merge with the changes to the non-EN pages, our system would consider them as updated, which is not true if there are other changes that need to be made to be 100% up-to-date with the English docs.

Also, welcome to Astro Docs! :partying_face: :heart:

juansemprun:
hello @Yan-Thomas, would that be ok?

Also, thank you for the welcome! 🎉

Yan-Thomas:
> hello @Yan-Thomas, would that be ok?

Yes, that's perfect! Thank you again. 🙌 💜

Obs: We always welcome first-time contributors on our Discord server with funny little puns, so If you'd like to hang out there, please join us! https://astro.build/chat

PR #1426, opened 2022-08-29
Add custom components to mdx integration guide

### LINKED ISSUE #1234: [md/mdx] components export for `mdx` in integration guide
One thing I really like about MDX is that you can pass custom components for default html elements (h1/6, img, a, code, pre, etc.) with export const components = {}. The names comes from the [Table of Components](https://mdxjs.com/table-of-components/) for MDX. Much easier than writing remark/rehype plugins IMO. 

There's an example of it here, but could be nice to highlight in the docs: https://stackblitz.com/github/withastro/astro/tree/latest/examples/with-mdx?file=src%2Fpages%2Findex.mdx

@bholmesdev, Would you welcome an additional section in the MDX integration guide? I was told that you were working on those docs, so I don't want to interfere or do work you may already have done!

kylebutts:

- New or updated content

#### Description

- Closes #1234 by adding custom components description to MDX integration guide

### DIFF
--- a/src/pages/en/guides/integrations-guide/mdx.md
+++ b/src/pages/en/guides/integrations-guide/mdx.md
@@ -257,6 +257,44 @@ const { title, fancyJsHelper } = Astro.props;
 <!-- -->
 ```
 
+### Custom components
+
+Under the hood, MDX will convert markdown into html components. For example, 
+
+```md
+> A blockquote with *some* emphasis.
+```
+
+will be converted into
+
+```html
+<blockquote>
+  <p>A blockquote with <em>some</em> emphasis.</p>
+</blockquote>
+```
+
+MDX provides a way to tap into this rendering and use your own custom components. In the above example, you could create a custom Blockquote component (in any language) that has either a `<slot />`  component or accepts a `children` prop. Then in the MDX file you import the component and export it to the `components` export. 
+
+```mdx
+import Blockquote from '../components/Blockquote.astro';
+export const components = { blockquote: Blockquote };
+```
+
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/integrations-guide/mdx.md
```diff
@@ -257,6 +257,44 @@ const { title, fancyJsHelper } = Astro.props;
...
+```
+
+MDX provides a way to tap into this rendering and use your own custom components. In the above example, you could create a custom Blockquote component (in any language) that has either a `<slot />`  component or accepts a `children` prop. Then in the MDX file you import the component and export it to the `components` export. 
+
+```mdx
+import Blockquote from '../components/Blockquote.astro';
+export const components = { blockquote: Blockquote };
+```
+
+The advantage of this is it allows for the simplicity of writing in markdown without having to write the custom component or writing a remark/rehype plugin. A full list of components that can have custom components is on the [MDX website]https://mdxjs.com/table-of-components/).
```
Yan-Thomas, line 283: Just a small syntax issue here
```suggestion
The advantage of this is it allows for the simplicity of writing in markdown without having to write the custom component or writing a remark/rehype plugin. A full list of components that can have custom components is on the [MDX website](https://mdxjs.com/table-of-components/).
```

Yan-Thomas (commented):
Thanks so much for this @kylebutts, great work! Though, the `integrations-guide` files are generated from their READMEs in the main Astro repo, so you will have to make the PR there instead: https://github.com/withastro/astro/blob/main/packages/integrations/mdx/README.md

PR #1432, opened 2022-08-29
Add warning about SSR features inside layouts

Yan-Thomas:

- New or updated content

#### Description

Recently discovered that layouts and components can't use SSR features. This wasn't previously mentioned in the docs, so I'm adding a little warning for this.

Matthew's comment on the subject ("Is not being able to use SSR features on layouts expected behavior?"): 
> It's because of streaming. You can only send headers / status code once and by the time we render a layout the headers are already sent. so it is intentional, but unfortunate.

NOTE: I tested all three features listed below and all of them didn't seem to work. I did expect that `Astro.request.headers.get` or `Astro.redirect` would work, but that doesn't seem to be the case. I appreciate someone double-checking this so we are sure I haven't made a mistake here.

### DIFF
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -47,6 +47,10 @@ You can find instructions at the individual adapter links above to complete the 
 
 Astro will remain a static-site generator by default. But once you enable a server-side rendering adapter, **every route in your pages directory becomes a server-rendered route** and a few new features become available to you.
 
+:::warning
+The features below are not available inside layouts nor components. Because of streaming, things like headers and status codes can only be sent once, and by the time a layout or component is rendered, this had already happened.
+:::
+
 ### `Astro.request.headers`
 
 The headers for the request are available on `Astro.request.headers`. It is a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) object, a Map-like object where you can retrieve headers such as the cookie.
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -64,7 +64,7 @@ const cookie = Astro.request.headers.get('cookie');
 :::caution
 The features below are only available at the page level. (You can't use them inside of components, including layout components.)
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks, @Yan-Thomas! I'm assigning this one to @Jutanium  to take a look at the specific wording. Dan, tag! You're it!

Jutanium:
@Yan-Thomas I went to make a repro for `Astro.request.headers.get` and [it does seem to work!](https://stackblitz.com/edit/github-brqemp?file=src/components/TestRequest.astro). Let me know what you tried!

Yan-Thomas:
> I went to make a repro for Astro.request.headers.get and [it does seem to work!](https://stackblitz.com/edit/github-brqemp?file=src/components/TestRequest.astro) Let me know what you tried!

Thanks for double-checking @Jutanium, I actually had tried the same thing but the result wasn't being rendered to me, so I believe it was Stackblitz not updating my changes somehow. 

All the wording changes you did are also great, so I believe we are good to merge this!

src/pages/en/guides/server-side-rendering.md
```diff
@@ -61,6 +61,12 @@ const cookie = Astro.request.headers.get('cookie');
 </html>
 ```
 
+:::caution
+The features below are only available at the page level. (You can't use them inside of components, including layout components.)
+
+This is because these features modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header), which can't be modified after they're sent to the browser. In SSR mode, Astro uses HTML streaming to send each component to the browser as it renders them. This makes sure the user sees your HTML as fast as possible, but it means that by the time Astro runs your component code, it has already sent the Response headers.
```
sarah11918, line 67:
```suggestion
This is because these features modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header), which can't be modified after they're sent to the browser. In SSR mode, Astro uses HTML streaming to send each component to the browser as it renders them. This makes sure the user sees your HTML as fast as possible, but it also means that by the time Astro runs your component code, it has already sent the Response headers.
```
I promise I'm not just looking for some co-author cred! I *really* wanted to hear this word there, and I don't know why. Agree? Disagree? You make the call. I'm good with your final verdict either way.

Yan-Thomas: haha, don't worry! I agree with adding "also" here.

Issue #1412, opened 2022-08-26
Vercel deployment docs conflict


alex-grover:
The default Vercel deploy docs state that you don't need any additional configuration to deploy static sites, while the `@astrojs/vercel` docs state that you need a `vercel.json`.

https://docs.astro.build/en/guides/deploy/vercel/#static-site
https://docs.astro.build/en/guides/integrations-guide/vercel/#why-astro-vercel

I'm happy to make a PR, but I'm not sure which docs are correct.

sarah11918:
Thanks for this, @alex-grover! This is a documentation issue, but not the one you're thinking of!

In fact, both docs are correct because the Vercel integration is only for SSR, and is not necessary for deploying static sites. 

Someone added a note for the adapter usage about the `vercel.json` file in the wrong place on that adapter page, and it's misleading right now because it looks like it's related to static site deployment. I have a note to go and change that information's location to be further down the page, where it's clear that it is a note about this SSR adapter's usage. I'm sorry you hit this before I could get that change made!

I'm assuming the original wording I intend to revert to at the top of the page would not be confusing?

> If you’re using Astro as a static site builder — its behavior out of the box — you don’t need an adapter. 
>
> If you wish to [use server-side rendering (SSR)](https://docs.astro.build/en/guides/server-side-rendering/), Astro requires an adapter that matches your deployment runtime.

alex-grover:
thanks for the clarification! your proposed updates make sense to me 👍🏽

PR #1437, opened 2022-08-30
[vercel deploy] Note: re: 404 page issue

sarah11918:
- New or updated content

Moved content added to the SSR integration README in this PR https://github.com/withastro/astro/pull/4299 which attempted to document a workaround to this Vercel bug: https://github.com/withastro/astro/issues/4164 but did so in the wrong place and led to confusion.

Removed the content from the SSR adapter integration in this PR: https://github.com/withastro/astro/pull/4551

Closes https://github.com/withastro/docs/pull/1412

Totally open to edits! (Too Vercel blamey?) Just trying to quickly capture the info here.


delucis:
I’m going to run some tests on this one. This workaround has some notable side effects by disabling `trailingSlash` so want to double check this is definitely required.

src/pages/en/guides/deploy/vercel.md
```diff
@@ -17,6 +17,17 @@ Your Astro project can be deployed to Vercel as a static site, or as a server-si
 
 Your Astro project is a static site by default. You don’t need any extra configuration to deploy a static Astro site to Vercel. 
 
+:::note
+There is currently a Vercel issue displaying a 404 page on Astro websites. Until this is fixed, you can add the following config file at the root of your project:
+
+```json title="vercel.json"
+{
+  "cleanUrls": true,
+  "trailingSlash": false
```
delucis, line 26: In my testing, just setting `"cleanUrls": true` was enough to get the 404 page working, so I think we can narrow this down to that:

```suggestion
```

This is also less of a serious side effect than the `trailingSlash` option, which is more likely to conflict with desired behaviour. It’s still _definitely_ a bug on Vercel’s end that this is necessary, so curious to try to keep tabs on this so we know when this is fixed.

sarah11918:
@delucis With the code update, good to go?

delucis (approved):
LGTM!

PR #1440, opened 2022-08-30
Navigation Reorder Battle Royale

sarah11918:
- Something else!

Dealing with the Astro DX Week realization that alphabetical order is not gonna cut it for our side navigation...

**Immediate concern**: UI Frameworks was too low to be easily found, given how many people are going to reach for it early.

**Injection of Chaos**: Sarah went rogue and reordered some *other* menu items, too! (No, Ben, Sharing State is *still* not a core concept, but you were upgraded to premium economy...)

Are there any strenuous objections to any reordered items here that would not satisfy #NWTWWHB? Will also entertain reasoned pleas for personal pet pages, but the goal is immediate DX improvement!


Jutanium:
Definitely better than what we had before.
I imagine this is _not_ the PR for me to suggest renaming "Guides" to something else?

sarah11918:
You can try, @Jutanium. Do you have a clearly better suggestion that won't distract from the task at hand?  😅

Jutanium:
@sarah11918 I'M GOING IN

src/i18n/en/nav.ts
```diff
@@ -29,22 +29,22 @@ export default [
 	{ text: 'Static Assets', slug: 'guides/imports', key: 'guides/imports' },
 	{ text: 'Troubleshooting', slug: 'guides/troubleshooting', key: 'guides/troubleshooting' },
 	{ text: 'Deploy', slug: 'guides/deploy', key: 'guides/deploy' },
+	{ text: 'Server-side Rendering (SSR)', slug: 'guides/server-side-rendering', key: 'guides/server-side-rendering' },
 
 
 	{ text: 'Guides', header: true, type: 'learn', key: 'features' },
```
Jutanium, line 35: The difference between the `Basics` section and the `Guides` section isn't that `Guides` are Guides. It's that `Guides` are "taking Astro beyond basics", incorporating other tools from _beyond_ Astro _into_ Astro: TypeScript, integrations, frameworks, styling solutions, fonts.

Now is a good opportunity to rename Guides to reflect that. Something as simple as this would be NWTWWHB:
```suggestion
	{ text: 'Next Steps', header: true, type: 'learn', key: 'features' },
```
```suggestion
	{ text: 'More', header: true, type: 'learn', key: 'features' },
```
It could have some flair:
```suggestion
	{ text: 'Beyond Basics', header: true, type: 'learn', key: 'features' },
```

src/i18n/en/nav.ts
```diff
@@ -29,22 +29,22 @@ export default [
...
 	{ text: 'CSS & Styling', slug: 'guides/styling', key: 'guides/styling' },
+	{ text: 'Images', slug: 'guides/images', key: 'guides/images' },
 	{ text: 'Fonts', slug: 'guides/fonts', key: 'guides/fonts' },
 	{ text: 'Data Fetching', slug: 'guides/data-fetching', key: 'guides/data-fetching' },
-	{ text: 'Environment Variables', slug: 'guides/environment-variables', key: 'guides/environment-variables' },
-	{ text: 'Images', slug: 'guides/images', key: 'guides/images' },
 	{ text: 'Import Aliases', slug: 'guides/aliases', key: 'guides/aliases' },
-	{ text: 'Integrations', slug: 'guides/integrations-guide', key: 'guides/integrations-guide' },
+	{ text: 'Environment Variables', slug: 'guides/environment-variables', key: 'guides/environment-variables' },
 	{ text: 'RSS', slug: 'guides/rss', key: 'guides/rss' },
```
Jutanium, line 47: I like the idea of ordering these loosely by:
- How common it is to use a feature
- "order of operations": if one item is a prereq to use another, move it forward
I think Sharing State is less common than CSS, Images, and Fonts, so I'd move that down.
Tweaking TypeScript is less common of a goal than using UI Frameworks, so I'd move that down.
Configuring Astro goes deep into the weeds of config. It's more likely that you'd want to include a UI Framework or an integration than to configure something that _isn't_ covered by those pages. So I'd also move that down.

Here's _one_ possible version of incorporating those three changes:

```suggestion
	{ text: 'Integrations', slug: 'guides/integrations-guide', key: 'guides/integrations-guide' },
	{ text: 'UI Frameworks', slug: 'core-concepts/framework-components', key: 'core-concepts/framework-components' },
	{ text: 'Configuring Astro', slug: 'guides/configuring-astro', key: 'guides/configuring-astro' },
	{ text: 'TypeScript', slug: 'guides/typescript', key: 'guides/typescript' },
	{ text: 'CSS & Styling', slug: 'guides/styling', key: 'guides/styling' },
	{ text: 'Images', slug: 'guides/images', key: 'guides/images' },
	{ text: 'Fonts', slug: 'guides/fonts', key: 'guides/fonts' },
	{ text: 'Data Fetching', slug: 'guides/data-fetching', key: 'guides/data-fetching' },
	{ text: 'Sharing State', slug: 'core-concepts/sharing-state', key: 'core-concepts/sharing-state' },
	{ text: 'RSS', slug: 'guides/rss', key: 'guides/rss' },
  { text: 'Import Aliases', slug: 'guides/aliases', key: 'guides/aliases' },
	{ text: 'Environment Variables', slug: 'guides/environment-variables', key: 'guides/environment-variables' },
```

I could also see moving "Configuring Astro" and "Typescript" down to form a little "config files" group:
- Configuring Astro
- TypeScript
- Import Aliases
- Environmental Variables

Jutanium (commented):
Please do not be afraid to discard all of my thoughts here.

src/i18n/en/nav.ts
sarah11918: So, I'm not sure any of these are immediate no brainers, though I get the idea. And, I know these are just illustrative, but my criterion to you was "Do you have a no-brainer winning suggestion so we don't get bogged down with this" and I'm not sure we're quite at that level yet.  😄  If I put on Critical Cap:

"Next Steps" seems weird if I come into docs searching and I'm "already there." So, if I think about a variety of readers, not ones progressing in order, this doesn't resonate with me. I get what it means, but I think it could be worse than "Fake Guides."

"More" may be more accurate than Guides-that-aren't-really-Guides, but it totally feels like phoning it in. Looks lazy and sloppy.

"Beyond Basics" also feels a bit lazy. Basics! Beyond Basics! Even Less Basic?! I also think this hits a little worse.

I think "Guides" is a section heading that the average person is expecting. I also think the average person won't really notice that many of them aren't technically "guides." (Some of them sort of are!) I think people will see Core Concepts and get that those are a bit more conceptial. I think people will see Basics and think, "This is what I gotta know to wrap my head around Astro." And, I think people will see Guides and think, "These are natural extensions of the basic, core concepts of Astro. These are things I might want to figure out how to do." (e.g. add an RSS feed, set up TypeScript) So, I think Guides works? For now? 

I think coming up with something that tops these built-in advantages of the "strategic lie" of "Guides" is not as simple a task as it may seem. So, I think we should hold off on this for this PR to prioritize highlighting UI Frameworks, and maybe a couple of other things, more urgently.

src/i18n/en/nav.ts
sarah11918: I could get behind a little Config group. Would you even add its own heading? Part of me really kind of wants configuring Astro to be higher, because it seems like configuring might be a thing you'd want to do, but I think you're right. That's a bit more advanced and not what people are jumping to immediately. 

Let me try a reorder based on this and see what I come up with, and how closely it matches this.

src/i18n/en/nav.ts
Jutanium: Right, I couldn't come up with a no-brainer suggestion. I just don't like Guides when both sections are equal amounts Guides.

Yes, let's hold it off. I _had_ to bring it up, and I'm glad you wrote out this analysis.

src/i18n/en/nav.ts
sarah11918: Pushed another option, including a new section... waiting for checks, but let's see how this looks @Jutanium

Jutanium: It looks great! Little Config Group is all grown up with its own header

sarah11918:
OK! Want to wait to give @delucis @hippotastic @kevinzunigacuellar a chance to object before we merge this!

Also, time to speak up if you hate this new sidebar, @agustinmulet @kecrily @X7md @kyosuke @Chrissdroid!

kevinzunigacuellar:
maybe this is a bit nitpicking but maybe we can put fonts and CSS Styling next to each other?

sarah11918:
@kevinzunigacuellar I was thinking that CSS is probably more used than anything else, and probably Images more used than Fonts. Do you have a gut feeling whether you'd drop Images lower, or move it above CSS (in order to keep CSS and Fonts together)? Which feels better to you?

kyosuke:
@sarah11918 I like the new side navigation order!
I especially like how it's organized under "CONFIGURATION."

Not directly related to the order, but my concern about the sidebar is that the "UI Frameworks" link is to "Framework Components".

Could the navigation label be changed to "Framework Components" or could the page title be changed to "UI Frameworks"?

Similarly, "CSS & Styling" is now "Styling & CSS", so I think it would be better to unify either.

agustinmulet:
Since I was summoned I will give my input 😝 
Love the new configuration part!
On the other hand, I personally would not consider MDX as something that belong on basics since it requires an integration, and also wouldn't consider SSR as something basic. Also, in SSR we talk about fetching data and SSR is positioned before `Data Fetching`.

BUT that's just my opinion, the battle royale continues! 😆

kevinzunigacuellar:
> @kevinzunigacuellar I was thinking that CSS is probably more used than anything else, and probably Images more used than Fonts. Do you have a gut feeling whether you'd drop Images lower, or move it above CSS (in order to keep CSS and Fonts together)? Which feels better to you?

I see the logic now, it's a hard call. I would be okay with dropping images.
Perhaps, analytics can tell us more about which page is more popular 😇 and helps us decide.

sarah11918:
> @sarah11918 I like the new side navigation order! I especially like how it's organized under "CONFIGURATION."

Yay! This seems to be a popular idea! I think we'll keep it!

> Could the navigation label be changed to "Framework Components" or could the page title be changed to "UI Frameworks"?
> 
> Similarly, "CSS & Styling" is now "Styling & CSS", so I think it would be better to unify either.

Cries in redirects and Hippo link checker... 😅 

We will be looking at doing a site overhaul that will involve a lot of page redirects. So I'd rather not get into that YET which means changing only the text content in the nav bar (not file names) at this point.

"Framework Components" makes sense to me, and in fact I just had to answer a question today in Discord where my answer was, "No you can't have `.vue` *pages*, only *components*. So, I don't think emphasizing components is a bad idea...

As for the other, I'm honestly trying to remember if we stragegically wrote "CSS & Styling" *because* things were in alphabetical order, and we figured CSS was kind of a big deal that people would want to know about, so we wanted it higher in the list!  😂 I think CSS is just so MUCH MORE the thing that people will be looking for and catch people's eye, that I'm hesitant to change that one, though.

Anyone else have thoughts re: Framework Components vs UI Frameworks and CSS & Styling vs Styling & CSS??

sarah11918:
> On the other hand, I personally would not consider MDX as something that belong on basics since it requires an integration, and also wouldn't consider SSR as something basic. Also, in SSR we talk about fetching data and SSR is positioned before `Data Fetching`.

Gah! These are great points!

MD and MDX are now *so closely* aligned in how they're used (and, we're really pushing MDX usage because that's how you get what you *used* to get with Astro components) that it honestly seemed more trouble NOT to have them together. We're still working on this "story" (and, Ben is still writing more features!  😅 ) so there will absolutely be more changes to come in that space.

I threw SSR up into Basics for this draft to make it closer to Deploy, but strategically *right after* deploy so it looked kind of like "another deploy option" maybe? One thing we have noticed about our SSR doc coverage in general is it still *feels* like an experimental add-on, where we just removed the experimental warning. We are seeing a LOT of people using/interested in it, and I think it *is* going to end up being as basic as "deploy your site." So this was an attempt to give it more prominence, since we are not only advertising this feature, but also, the SSR guides (like the Integration Guides) are some of our nicer looking pages (THAN YOU CHRIS) and the only multi-level navigation items where you have to click AGAIN to get to info. So we don't want them lost down at the bottom of the site where, out of nowhere, it's CONTENT EXPLOSION! That's why Integrations are also so high up now.

This does make me wonder if we'd consider Data Fetching "Basics" now? Especially as one of our focuses over the next few weeks is adding CMS integration guides.

kyosuke:
In my opinion, I agree with Sarah that it would be better to change the label to "Framework Conponents".
However, if we do that, it might be a bit confusing since "Componets" is also under "BASICS".
How about changing the label of "Components" to "Astro Components”?

As for "CSS & Styling"... maybe CSS should come first. Because if I'm looking for Styling, I look for the word "CSS".

image

sarah11918:
I did just draft another branch, but yes, I'll change it to Astro components!

sarah11918:
Random thought: I just saw a couple of tweets about Astro where people referred to **JS Framework Components** to describe React, Vue, Svelte etc....

Is there a nice "Astro Component" vs "JS Framework Component" story there? Is there any way in which "JS Framework Components" is better than "UI Framework Components"?  (Or, no, because we want to *downplay* JS in general??)

delucis (approved):

src/i18n/en/nav.ts
```diff
@@ -21,30 +21,33 @@ export default [
...
-	{ text: 'Troubleshooting', slug: 'guides/troubleshooting', key: 'guides/troubleshooting' },
+	{ text: 'Data Fetching', slug: 'guides/data-fetching', key: 'guides/data-fetching' },
 	{ text: 'Deploy', slug: 'guides/deploy', key: 'guides/deploy' },
+	{ text: 'Troubleshooting', slug: 'guides/troubleshooting', key: 'guides/troubleshooting' },
 
 
 	{ text: 'Guides', header: true, type: 'learn', key: 'features' },
-	{ text: 'Configuring Astro', slug: 'guides/configuring-astro', key: 'guides/configuring-astro' },
+	{ text: 'Integrations & Adapters', slug: 'guides/integrations-guide', key: 'guides/integrations-guide' },
+	{ text: 'UI Framework Components', slug: 'core-concepts/framework-components', key: 'core-concepts/framework-components' },
```
sarah11918, line 37:
```suggestion
	{ text: 'UI Frameworks', slug: 'core-concepts/framework-components', key: 'core-concepts/framework-components' },
```

src/i18n/en/nav.ts
```diff
@@ -21,30 +21,33 @@ export default [
...
 	{ text: 'Static Assets', slug: 'guides/imports', key: 'guides/imports' },
-	{ text: 'Troubleshooting', slug: 'guides/troubleshooting', key: 'guides/troubleshooting' },
+	{ text: 'Data Fetching', slug: 'guides/data-fetching', key: 'guides/data-fetching' },
 	{ text: 'Deploy', slug: 'guides/deploy', key: 'guides/deploy' },
+	{ text: 'Troubleshooting', slug: 'guides/troubleshooting', key: 'guides/troubleshooting' },
 
 
 	{ text: 'Guides', header: true, type: 'learn', key: 'features' },
-	{ text: 'Configuring Astro', slug: 'guides/configuring-astro', key: 'guides/configuring-astro' },
+	{ text: 'Integrations & Adapters', slug: 'guides/integrations-guide', key: 'guides/integrations-guide' },
```
sarah11918, line 36:
```suggestion
	{ text: 'Integrations', slug: 'guides/integrations-guide', key: 'guides/integrations-guide' },
```

PR #1443, opened 2022-08-31
Update mpa-vs-spa case studies

nniinnoo:
Description:
This PR contains another link for case study of MPA vs. SPA between Next.js and Astro which measure several performance keys


sarah11918:
This is wonderful, thank you @nniinnoo! And welcome to the Docs as a first-time contributor!  🥳

PR #1446, opened 2022-08-31
Recommend using `dependencies` over `devDependencies` in package.json

### LINKED ISSUE #1435: Clarify when to use dependencies vs dev-dependencies
As far as I know, it does not really matter whether you install as dev-dependency or dependency. I think we are supposed to install everything as dev-dependency due to SSG? How does this change with SSR?

I think it should be mentioned somewhere. I did not find anything in the reference, not sure about the best place.

There are regularly some support discussions on discord that are related or solved by changing the dependencies:
[Discord: devDependencies over dependencies?](https://discordapp.com/channels/830184174198718474/956662458343366687)
[Discord: Tiny TypeScript question](https://discordapp.com/channels/830184174198718474/1004687477866496030)
[Discord: dev dependency and/or peer dependency](https://discordapp.com/channels/830184174198718474/830184175176122389/971050625162960906)
...

Jutanium:

- New or updated content

#### Description

- Closes #1435 
- Explains that there are two options for specifying dependencies, and recommends `dependencies` as per #4544
- Note that our [package publishing docs](https://docs.astro.build/en/guides/publish-to-npm/) may have different advice. That page needs a complete overhaul, so I'm leaving it for now.

### DIFF
--- a/src/pages/en/core-concepts/project-structure.md
+++ b/src/pages/en/core-concepts/project-structure.md
@@ -101,6 +101,8 @@ As a general rule, any CSS or JavaScript that you write yourself should live in 
 
 This is a file used by JavaScript package managers to manage your dependencies. It also defines the scripts that are commonly used to run Astro (ex: `npm start`, `npm run build`).
 
+There are two kinds of dependencies you can specify in a `package.json`: `dependencies` and `devDependencies` ([see the NPM docs](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file)). In most cases, these work the same: Astro needs all dependencies at build time, and your package manager will install both. We recommend putting all of your dependencies in `dependencies` to start.
+
 For help creating a new `package.json` file for your project, check out the [manual setup](/en/install/manual/) instructions.
 
 ### `astro.config.mjs`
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/project-structure.md
```diff
@@ -101,6 +101,8 @@ As a general rule, any CSS or JavaScript that you write yourself should live in
 
 This is a file used by JavaScript package managers to manage your dependencies. It also defines the scripts that are commonly used to run Astro (ex: `npm start`, `npm run build`).
 
+There are two kinds of dependencies you can specify in a `package.json`: `dependencies` and `devDependencies` ([see the NPM docs](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file)). In most cases, these work the same: Astro needs all dependencies at build time, and your package manager will install both. We recommend putting all of your dependencies in `dependencies` to start.
```
sarah11918, line 104:
```suggestion
There are [two kinds of dependencies](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file) you can specify in a `package.json`: `dependencies` and `devDependencies`. In most cases, these work the same: Astro needs all dependencies at build time, and your package manager will install both. We recommend putting all of your dependencies in `dependencies` to start, and only use `devDependencies` if you find a specific need to do so.
```
What do you think about moving the reference link here?

delucis (commented):
Worth noting that this is also the _opposite_ of what our starter examples do: they use `devDependencies` for everything. So it might be confusing to start recommending something different without also updating those.

src/pages/en/core-concepts/project-structure.md
aFuzzyBear: I like this, reads better

PR #1447, opened 2022-08-31
minor: Add dev-only warning to integrations pages

Yan-Thomas:

#### Description

This PR adds the dev-only "don't edit this page directly" warning to integrations pages as well. I have changed the wording a bit to make it more clear that this is not true for translations. Also added a comment asking for the comment and component to be deleted from translations so we avoid confusing i18n contributors.


sarah11918:
Amaing, thank you @Yan-Thomas! LGTM! 🙌

Yan-Thomas:
Thanks for the review @sarah11918! 💜

PR #1449, opened 2022-08-31
Add frontmatter props and generated pages guidance on `TRANSLATING.md`

Yan-Thomas:

- New or updated content
- Something else!

#### Description

This PR explains more about which frontmatter props should be translated and about generated pages and why you can safely ignore (and even remove) them from your translations.

I'm not good at writing, feel free to completely change this!


TRANSLATING.md
```diff
@@ -133,3 +133,39 @@ pnpm run add-language
...
+---
+
+// Rest of the file's content is here...
+```
+
+### Generated pages and dev-only warnings
+
+Some of our pages are generated from outside sources, therefore, as a way to avoid us and other contributors from changing them, we show dev-only warnings that are not relevant for translations.
+
+For these generated pages (like Configuration Reference), we recommend ignoring and removing the note and component (including its import) from the file, thus avoiding other translators from possibly thinking that this warning applies to translations as well.
```
sarah11918, line 171:
```suggestion
### What to translate: Frontmatter

Our pages are generated from Markdown files which have frontmatter properties. These are variables that hold information about the page (values) that we later use to specify the page's title, description, and other special data. 

Here's an example file showing the **properties** of `layout`, `title`, `description` and `i18nReady` along with their corresponding values for this page.

```
---
layout: ~/layouts/MainLayout.astro
title: Data Fetching
description: Learn how to fetch remote data with Astro using the fetch API.
i18nReady: true
---

// Rest of the file's content is here...
```
**tl/dr: Translate only some values, never translate properties!**

The frontmatter **properties** themselves, like `title` and `description` should not be translated, as doing so would cause a runtime error and break our CI.

The frontmatter **values** that should be translated are `title` and `description`. 

Other frontmatter properties that aren't mentioned here should be ignored and not translated, as we use them for handling our "Edit this page" links or for specifying in which category an integration belongs, etc. 

Here is the above example correctly translated:

```
---
layout: ~/layouts/MainLayout.astro
title: Fetching de datos
description: Aprenda como obtener datos remotos con Astro utilizando la API de fetch.
i18nReady: true
---

// Rest of the file's content is here...
```

### Generated pages and dev-only warnings

Some of our English page content is generated from outside sources, and must not be edited directly in this repository. We need to show dev-only warnings to prevent contributors from changing that English content here, and instead guide them towards the proper source location of the English content.

However, these pages are translated directly here, and **these warnings are not meant for translations**.

For these generated pages (like `configuration-reference`), we recommend **ignoring and removing the note and component (including its import) from the file**, thus avoiding confusion for other translators thinking that this warning applies to translations as well.
```
Amazing, @Yan-Thomas! Took a quick editing pass. What do you think?

Yan-Thomas: Great as always @sarah11918!

kevinzunigacuellar:
👀❤️✅

TRANSLATING.md
```diff
@@ -133,3 +133,48 @@ pnpm run add-language
...
+i18nReady: true
+---
+
+// Rest of the file's content is here...
+```
+**tl/dr: Translate only some values, never translate properties!**
+
+The frontmatter **properties** themselves, like `title` and `description` should not be translated, as doing so would cause a runtime error and break our CI.
+
+The frontmatter **values** that should be translated are `title` and `description`.
```
sarah11918, line 157:
```suggestion
The only frontmatter **values** that should be translated are those corresponding to the `title` and `layout` properties: "*Data Fetching*" and "*Learn how to fetch remote data with Astro using the fetch API*." 
```
I realized this could be confusing - don't translate title and description! Only translate title and description! lol

Yan-Thomas: Agree! I'll go with your idea. Just a thing, you mistook `description` with `layout`.
```suggestion
The only frontmatter **values** that should be translated are those corresponding to the `title` and `description` properties: "*Data Fetching*" and "*Learn how to fetch remote data with Astro using the fetch API*."
```

sarah11918:
LGTM, @Yan-Thomas!  💜

PR #1459, opened 2022-09-02
Include upgrade commands on migrate page

### LINKED ISSUE #1419: Upgrading Astro
Document how to upgrade Astro within your project using your package manager (npm, pnpm, yarn).

Jutanium:

- New or updated content

#### Description

- Closes #1419
- The migration page documents what you might run into while upgrading Astro. Now it also describes _how_ to upgrade Astro!

### DIFF
--- a/src/pages/en/migrate.md
+++ b/src/pages/en/migrate.md
@@ -1,4 +1,6 @@
 ---
+setup: |
+    import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 layout: ~/layouts/MainLayout.astro
 title: Migration Guide
 description: How to migrate your project to latest version of Astro.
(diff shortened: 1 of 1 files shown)


sarah11918:
Big, if true.

(I had no idea this command existed.  😅 )  LGTM, @Jutanium!

Just note for our future selves (or present Dan): if we could somehow automatically show the current version where we say "the latest version" I think that would be awesome.

Yan-Thomas:
> Just note for our future selves (or present Dan): if we could somehow automatically show the current version where we say "the latest version" I think that would be awesome.

We have the `Version.astro` component that shows the latest version of a package (works with any package in the npm registry). We can use it in union with `Badge.astro` to have this.

PR #1462, opened 2022-09-02
Change && to ??

mzaini30:
Sometimes, && not return false but null. Example: https://github.com/mzaini30/app/blob/master/src/layout/base.astro#L27

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 


delucis:
Thanks for this @mzaini30! I think in this case we want to stick to the more common use of `&&` as the example is showing usage with a boolean value but thanks for the suggestion nonetheless 🙌

PR #1491, opened 2022-09-07
SSR Clarifications: `build`, `preview`, and ability to enable without an adapter

### LINKED ISSUE #1410: Document `astro preview`
Specifically, we should mention that it doesn't work to test SSR.

We also need to explain on the SSR page that, during development, you _don't_ need an adapter.

Jutanium:

- New or updated content

#### Description

- Closes #1410
- Adds the SSR-mode behavior of `astro build` and `astro preview` (previously undocumented)
- Adds a note that `astro preview` doesn't work if SSR is enabled
- Makes it clear that you only need `output: server` (and not an adapter) to enable SSR during _development_
- Simplifies the explanation of an adapter

### DIFF
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -23,7 +23,7 @@ During development, you can enable SSR by adding the `output: server` configurat
     });
     ```
 
-To deploy an SSR project, you also need to add an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. Each adapter corresponds to a runtime and allows Astro to output a script that runs your project on that kind of server.
+To deploy an SSR project, you also need to add an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. Each adapter allows Astro to output a script that runs your project on a specific runtime.
 
 The following adapters are available today with more to come in the future:
 
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -12,7 +12,7 @@ i18nReady: true
 
 ## Enabling SSR in Your Project
 
-To enable SSR features during development, add the `output: server` configuration option:
+To get started, enable SSR features during development with the `output: server` configuration option:
 
     ```js ins={5}
     // astro.config.mjs
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -12,7 +12,18 @@ i18nReady: true
 
 ## Enabling SSR in Your Project
 
-To enable SSR you need to use an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. This runtime provides an API that your server-side code can use.
+During development, you can enable SSR by adding the `output: server` configuration option:
+
+    ```js ins={5}
+    // astro.config.mjs
+    import { defineConfig } from 'astro/config';
+
+    export default defineConfig({
+      output: 'server'
+    });
+    ```
+
+To deploy an SSR project, you also need to add an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. This runtime provides an API that your server-side code can use.
 
 Installing an adapter gives Astro access to the corresponding API, and allows Astro to output a script that runs your project on that kind of server.
 
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -12,7 +12,7 @@ i18nReady: true
 
 ## Enabling SSR in Your Project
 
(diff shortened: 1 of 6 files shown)


src/pages/en/reference/cli-reference.md
```diff
@@ -24,13 +24,17 @@ Sets which network IP addresses the dev server should listen on (i.e. non-localh
 
 ### `astro build`
 
-Builds your site for production.
+Builds your site for deployment. By default, this will generate static files in to a `dist/` directory. If [SSR is enabled](/en/guides/server-side-rendering/), this will generate the necessary server files to serve your site.
```
sarah11918, line 27:
```suggestion
Builds your site for deployment. By default, this will generate static files and place them in a `dist/` directory. If [SSR is enabled](/en/guides/server-side-rendering/), this will generate the necessary server files to serve your site. 
```

src/pages/en/guides/server-side-rendering.md
```diff
@@ -12,9 +12,18 @@ i18nReady: true
...
+    ```js ins={5}
+    // astro.config.mjs
+    import { defineConfig } from 'astro/config';
+
+    export default defineConfig({
+      output: 'server'
+    });
+    ```
+
+To deploy an SSR project, you also need to add an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. Each adapter allows Astro to output a script that runs your project on a specific runtime.
```
sarah11918, line 26: I'm not entirely sure about moving this comment to the bottom of the example, and having the example only show the `output: 'server'` line. Do you think it might pre-empt some problems to mention needing an adapter above the example, then *also* showing the adapter installed in the config file?

It's not terrible, and I get that the server part is *really* the generic "this is what makes it SSR" thing. But if no one is ever going to have that without *also* having an adapter, then I think I lean towards having the example show both together, even if we use a placeholder or example adapter in there. (vs this, which is to not show an adapter at all so that we don't have to make the call which one to show).

src/pages/en/reference/cli-reference.md
```diff
@@ -24,13 +24,17 @@ Sets which network IP addresses the dev server should listen on (i.e. non-localh
...
 
-Starts a local static file server to serve your built `dist/` directory. Useful for previewing your static build locally, before deploying it.
+Starts a local file server to serve your static `dist/` directory.
 
-This command is meant for local testing only, and is not designed to be run in production. For help with production hosting, check out our guide on [Deploying an Astro Website](/en/guides/deploy/).
+This command is useful for previewing your static build locally, before deploying it. It is not designed to be run in production. For help with production hosting, check out our guide on [Deploying an Astro Website](/en/guides/deploy/).
+
+:::note
+This command does not work for SSR builds, which require the server runtime that corresponds to your adapter.
+:::
```
sarah11918, line 37:
```suggestion
:::caution
This command does not work for SSR builds, which require the server runtime that corresponds to your adapter.
:::
```
You could even choose to make this a caution and/or give it a specific title. Up to you.

sarah11918 (commented):
Thanks Dan! Some comments below!  🚀

src/pages/en/guides/server-side-rendering.md
Jutanium: What I wanted was to show that you don't need an adapter during development, which wasn't clear before.

The idea is that there _is_ a use case of enabling `output: server` without an adapter: when you know you want SSR, but you don't yet know where you'll deploy.

I think I can find a cleaner way to say this, though

sarah11918, line 26:
```suggestion
When it's time to deploy an SSR project, you also need to add an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. Each adapter allows Astro to output a script that runs your project on a specific runtime.
```
What about this?

src/pages/en/guides/server-side-rendering.md
```diff
@@ -12,9 +12,18 @@ i18nReady: true
 
 ## Enabling SSR in Your Project
 
-To enable SSR you need to use an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. This runtime provides an API that your server-side code can use.
+To get started, enable SSR features during development with the `output: server` configuration option:
```
sarah11918, line 15:
```suggestion
To get started, enable SSR features in development mode with the `output: server` configuration option:
```
I don't know why, but "during development" isn't working for me.

To get started developing in SSR mode, add the `output: server` configuration option:  ???

Jutanium: I like your first suggestion! `development mode` is definitely more clear than what I had

sarah11918:
🥳  LGTM, @Jutanium !

PR #1495, opened 2022-09-08
add section on remote markdown

FredKSchott:

- New or updated content

#### Description

- Responds to some user feedback [discussed in Discord](https://discord.com/channels/830184174198718474/1017086651014578257)
- Answers the user question of whether or not "remote markdown to HTML" is supported in Astro. Makes it clear that it is possible to implement yourself, but not provided by Astro in any way. Does not commit to any new level of support.

cc @Jutanium who had thoughts when we discussed on Discord.


sarah11918:
I like this addition, but I wonder about its placement on the page.

We currently have an "Importing Markdown" section, which seems a little too jargonly-named because it refers to "importing Markdown in an Astro file" (not "fetching" Markdown from an external source, which is described on our Data Fetching page). I think that importing could be confused to mean "fetching." 

I also worry about the label of "Remote Markdown" only addressing run-time fetching of remote Markdown, which I believe is the problem.

I'm not opposed to merging this to get it *in* the docs, but I think then we should look to maybe restructure more like:

MD/MDX Pages (as existing; what if your `.md` / `.mdx` file is meant to be an entire page, use a layout etc.

FredKSchott:
@sarah11918 yea, I actually first thought about whether this belonged in that "Working with Content in Astro" guide that we've talked about, where a generic "Fetching Remote Markdown" guide isn't too different from a generic "Fetching from a CMS" guide (both use fetch). 

Not at all opposed to restructuring this info to live somewhere else, my main goal with this PR is to help the user who is coming to our site with the question (via search or just reading the markdown page): "Does Astro support fetching Markdown from a remote URL?"

src/pages/en/guides/markdown-content.md
```diff
@@ -357,6 +357,24 @@ const { Content } = Astro.props.post
   <Content/>
 </article>
 ```
+## Fetching Remote Markdown
+
+Astro was designed to render Markdown to HTML as a part of building your website. However, there may be certain cases where you need to fetch Markdown from a remote source at runtime, after the site has been built. For example, you may want to fetch and render Markdown from a remote API every time a user makes a request to your website.
+
+**Astro does not include a runtime Markdown parser!** To render remote Markdown to HTML at runtime, you will need to install and configure your own Markdown parser. This **will not** inherit from any of Astro's built-in Markdown and MDX settings that you have configured. Be sure that you understand these limitations before implementing this in your project. 
+
+```astro
```
sarah11918, line 366:
```suggestion
```astro title="src/pages/markdown-cheatsheet.astro"
```
@FredKSchott took the liberty of simply moving this up near "Importing Markdown" and renaming it "Fetching Remote Markdown" to a) be consistent with the -ing of the other nearby headings and b) bring it closer to the other section about injecting Markdown onto a page (vs the `.md` file being its own page).

Otherwise, I think I'd probably just suggest a file name title for this, so that it's clear this is happening in an `.astro` component, but good to have mention of this in the docs!

src/pages/en/guides/markdown-content.md
```diff
@@ -357,6 +357,24 @@ const { Content } = Astro.props.post
   <Content/>
 </article>
 ```
+## Fetching Remote Markdown
+
+Astro was designed to render Markdown to HTML as a part of building your website. However, there may be certain cases where you need to fetch Markdown from a remote source at runtime, after the site has been built. For example, you may want to fetch and render Markdown from a remote API every time a user makes a request to your website.
+
+**Astro does not include a runtime Markdown parser!** To render remote Markdown to HTML at runtime, you will need to install and configure your own Markdown parser. This **will not** inherit from any of Astro's built-in Markdown and MDX settings that you have configured. Be sure that you understand these limitations before implementing this in your project.
```
delucis, line 364: This description relies heavily on the concept of a “runtime” as a way to flag _when_ something happens. Looking through docs we use this in a few different contexts but I don’t think we ever really explain what this jargon means.

For example, on Getting Started we use it to mean both client-side and server-side JS:

> - **Zero JS, by default:** No JavaScript **_runtime_** overhead to slow you down.
> - **Edge-ready:** Deploy anywhere, even a global edge **_runtime_** like Deno or Cloudflare.

I think “runtime” is a bit vague because you also _run_ your build, which I guess you could call “Astro runtime”.

Additionally, although currently this section is focused on SSR, this is equally valid/helpful advice for remote Markdown sources a user might want to compile at build time, so we can probably generalise this.

```suggestion
Astro was designed to render local Markdown files to HTML when building your website. However, there may be certain cases where you need to fetch Markdown from a remote source. For example, you may want to fetch and render Markdown from a remote API at build time or, if using [SSR](/en/guides/server-side-rendering/), every time a user makes a request to your website.

To render remote Markdown to HTML, you will need to install and configure your own Markdown parser. This **will not** inherit from any of Astro's built-in Markdown and MDX settings that you have configured. Be sure that you understand these limitations before implementing this in your project. 
```

How does this sound?

Jutanium:
I'm happy with this (with the proposed changes) from a docs perspective! The description of the limitations and the disclaimers (_Be sure that you understand these limitations before implementing this in your project._) satisfy my concerns about this not being an elegant solution. We should revisit this as a product decision down the line.

FredKSchott:
Ah, you're right @delucis, "runtime" is an overloaded term. I like your proposed rewording to avoid the term. Will do another round of edits on this PR!

FredKSchott:
@sarah11918 I ended up moving it back down for this reason: it felt odd that this section about "how fetching remote markdown doesn't support the built-in Markdown parser" was in between two sections about how the built-in markdown parser works, and then how you configure it. I found that confusing (and a bit jarring) as a reader switching back and forth between contexts, so I moved to back to the bottom. 

I'm going to assuming that you're okay with this reasoning and will merge as-is (along with Chris's suggestions above) but if you'd still rather see it moved up, just let me know and I'll create a follow-up PR.

FredKSchott:
Thanks everyone for the reviews!

PR #1504, opened 2022-09-08
Small sidebar design refresh

delucis:

#### Description

- This PR tweaks the design of our sidebars following some recent discussion

##### Desktop view

<table>
<tr>
    <th>Before
    <th>After
<tr>
	<td>
	<td>
</table>

##### Mobile menu

<table>
<tr>
    <th>Before
    <th>After
<tr>
	<td>
	<td>
</table>

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -483,16 +483,14 @@ button {
 }
 
 h2.heading {
-	font-size: 1rem;
-	font-weight: 700;
-	text-transform: uppercase;
+	font-size: 1em;
+	font-weight: 600;
 	margin-bottom: 0.5rem;
 	margin-top: 1.5rem;
 }
 
 .header-link {
-	font-size: 1rem;
-	padding: 0.1rem 0 0.1rem 0;
+	font-size: 1em;
 	transition: border-inline-start-color 100ms ease-out, background-color 200ms ease-out;
 }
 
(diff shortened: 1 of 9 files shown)


hippotastic:
I‘m loving basically everything about this PR except the smaller font sizes. On mobile Safari both the TOC and the main navigation feel way to small now and there is a stark contrast between the native text size of the browser UI, language dropdown etc. (which have the expected size) and the new small navigation texts.

Would it be possible to just revert the font size change at least on mobile and keep everything else?

sarah11918:
Some initial thoughts:

Desktop: fabulous, no notes!
Tablet view (admittedly, by resizing my desktop screen... my tablet is too large): A little small if I turn my phone sideways, not too small when my destop browser is resized.

Mobile: Joe finds the "on this page" menu items a little too small. The Hamburger menu is maybe not too small? But def registers as small. When I asked, "Is it too small?" It was no, but a *considered* no, after much thought.

For reference: I immediately went to NextJS's docs, and their mobile menu font seems to be the same size/weight as this PR.
Tailwind: same size, *slightly* bolder maybe, but appears to be double spaced nav items. A LOT more white space. And, I suspect a bit wider kerning (letter spacing) of the letters within the words? (Or a font with those properties.)

So, I don't necessarily think it's the font SIZE that's the issue (even though it might feel like it?) on mobile. I think we can address this with more spacing? (I do think the bold sections work WELL.)

hippotastic (approved):
Looks great on my devices (iOS Safari, Chrome on Windows 10).

Yan-Thomas:
Thanks @delucis, just checked here and everything looks great! 🙌

sarah11918:
All looks good here! Lets Get that Menu!

PR #1507, opened 2022-09-08
Update: sharing-state.md.  Fix ref(s) for "cartStore" in code example

mikekidder:

- Minor content fixes:  path label change

#### Description

- change from "clientStore" -> "src/cartStore" to match source code

Updated all language md file found.

### DIFF
--- a/src/pages/en/core-concepts/sharing-state.md
+++ b/src/pages/en/core-concepts/sharing-state.md
@@ -321,7 +321,7 @@ Let's add a `cartItem` store to our `cartStore.js` from earlier. You can also sw
 <Fragment slot="tab.ts">TypeScript</Fragment>
 <Fragment slot="panel.js">
 ```js
-// clientStore.js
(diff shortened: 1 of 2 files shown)


delucis (commented):
Thanks for this @mikekidder!

Now that you’ve drawn my attention to it, I’m also noticing that the `src/components/AddToCartForm` components all import from `'./cart/store'`, which we should probably also update to `'../cartStore'` right? Would you like to add those changes to this PR?

mikekidder:
> Now that you’ve drawn my attention to it, I’m also noticing that the `src/components/AddToCartForm` components all import from `'./cart/store'`, which we should probably also update to `'../cartStore'` right? Would you like to add those changes to this PR?

sure thing!

delucis (approved):
Perfect! Thanks for fixing this @mikekidder and welcome to our contributors 🥳

Issue #1482, opened 2022-09-07
Clarify that Astro always treats your code as TypeScript


Jutanium:
Coming out of a conversation with @Princesseuh and @bholmesdev, we want to make two things clear in the docs:
- Even if you try to opt-out of TypeScript by removing the `tsconfig`, Astro and the language tools treats your code as TypeScript
- A "why TypeScript" note: _why_ does TypeScript help you as you use Astro?

Will draft the PR

PR #1509, opened 2022-09-08
Add information about TypeScript templates

Princesseuh:

- New or updated content

#### Description

Add some information about our new tsconfig.json bases that users should extend from instead of writing their own

### DIFF
--- a/src/pages/en/guides/typescript.md
+++ b/src/pages/en/guides/typescript.md
@@ -13,7 +13,7 @@ Astro doesn't perform any type checking itself. Type checking should be taken ca
 
 It is **strongly recommended** that you create a `tsconfig.json` file in your project, so that tools like Astro and VSCode know how to understand your project. Some features (like npm package imports) aren't fully supported in TypeScript without a `tsconfig.json` file.
 
-Template `tsconfig.json` are included in Astro that you can extend from to customize your project using [the `extends` setting](https://www.typescriptlang.org/tsconfig#extends). They can be found here: [astro/tsconfigs/base](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/base.json), [astro/tsconfigs/strict](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strict.json) and [astro/tsconfigs/strictest](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strictest.json).
+`tsconfig.json` templates are included in Astro that you can extend from using [the `extends` setting](https://www.typescriptlang.org/tsconfig#extends). They can be found here: [astro/tsconfigs/base](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/base.json), [astro/tsconfigs/strict](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strict.json) and [astro/tsconfigs/strictest](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strictest.json).
 
 ```json title="tsconfig.json"
 {
--- a/src/pages/en/guides/typescript.md
+++ b/src/pages/en/guides/typescript.md
@@ -13,24 +13,11 @@ Astro doesn't perform any type checking itself. Type checking should be taken ca
 
 It is **strongly recommended** that you create a `tsconfig.json` file in your project, so that tools like Astro and VSCode know how to understand your project. Some features (like npm package imports) aren't fully supported in TypeScript without a `tsconfig.json` file.
 
-Some TypeScript configuration options require special attention in Astro. Below is our recommended starter `tsconfig.json` file, which you can copy-and-paste into your own project. Every [astro.new template](https://astro.new/) includes this `tsconfig.json` file by default.
+Template `tsconfig.json` are included in Astro that you can extend from to customize your project using [the `extends` setting](https://www.typescriptlang.org/tsconfig#extends). They can be found here: [astro/tsconfigs/base](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/base.json), [astro/tsconfigs/strict](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strict.json) and [astro/tsconfigs/strictest](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strictest.json).
 
 ```json title="tsconfig.json"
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/typescript.md
```diff
@@ -13,24 +13,11 @@ Astro doesn't perform any type checking itself. Type checking should be taken ca
 
 It is **strongly recommended** that you create a `tsconfig.json` file in your project, so that tools like Astro and VSCode know how to understand your project. Some features (like npm package imports) aren't fully supported in TypeScript without a `tsconfig.json` file.
 
-Some TypeScript configuration options require special attention in Astro. Below is our recommended starter `tsconfig.json` file, which you can copy-and-paste into your own project. Every [astro.new template](https://astro.new/) includes this `tsconfig.json` file by default.
+`tsconfig.json` templates are included in Astro that you can extend from using [the `extends` setting](https://www.typescriptlang.org/tsconfig#extends). They can be found here: [astro/tsconfigs/base](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/base.json), [astro/tsconfigs/strict](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strict.json) and [astro/tsconfigs/strictest](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strictest.json).
```
sarah11918, line 16:
```suggestion
Three extensible `tsconfig.json` templates are included in Astro: `base`, `strict`, and `strictest`. The `base` template powers our VSCode extension and enables several modern ESM features. We recommend using `strict` or `strictest` if you plan to write TypeScript in your project. You can view and compare the three template configurations at [astro/tsconfigs/](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/).

To add or change a template, use [the `extends` setting](https://www.typescriptlang.org/tsconfig#extends):
```
What do you think about a structure like this? Of course, change anything as necessary to make sure it reads the way someone configuring TypeScript would expect.

src/pages/en/guides/typescript.md
```diff
@@ -63,10 +50,9 @@ import type { SomeType } from './script';
 ```
 This way, you avoid edge cases where Astro's bundler may try to incorrectly bundle your imported types as if they were JavaScript.
 
-In your `.tsconfig` file, you can instruct TypeScript to help with this. The [`importsNotUsedAsValues` setting](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) can be set to `error`. Then, TypeScript will check your imports and tell you when  `import type` should be used.
+In your `.tsconfig` file, you can instruct TypeScript to help with this. The [`importsNotUsedAsValues` setting](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) can be set to `error`. Then, TypeScript will check your imports and tell you when  `import type` should be used. This setting is included by default in our strict and strictest bases.
```
sarah11918, line 53:
```suggestion
In your `.tsconfig` file, you can instruct TypeScript to help with this. The [`importsNotUsedAsValues` setting](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) can be set to `error`. Then, TypeScript will check your imports and tell you when  `import type` should be used. This setting is included by default in our strict and strictest templates.
```
I'm wondering whether this should be "templates" since we call them templates above.

sarah11918 (commented):
This is great, @Princesseuh, thanks for this!

While we're in this file, I also asked you for thoughts about updating the intro paragraph for this page which I think might fit better now?

I also added a bit of info and explicit recommendation to the part about the templates, in case people reading it felt like they wanted a bit more info to know which one to choose. See what you think!

src/pages/en/guides/typescript.md
Princesseuh: I like the structure but this sentence isn't true:

> The `base` template powers our VSCode extension and enables several modern ESM features.

What powers the VS Code extension is ultimately the template the user has chosen. As for features, the `base` template is used as a basis for the `strict` and `strictest` template (but is also usable by itself, strictest extends from strict extends from base), so every templates has the same "features" (apart from strictness related ones)

I'll take the structure here and `extends` from it, ba dum tssh

src/pages/en/guides/typescript.md
Princesseuh: Yep, accidental leftover because I initially called them bases (that's a term that TypeScript uses, but it's used for a specific repo so not valid here)

src/pages/en/guides/typescript.md
```diff
@@ -13,24 +13,11 @@ Astro doesn't perform any type checking itself. Type checking should be taken ca
 
 It is **strongly recommended** that you create a `tsconfig.json` file in your project, so that tools like Astro and VSCode know how to understand your project. Some features (like npm package imports) aren't fully supported in TypeScript without a `tsconfig.json` file.
```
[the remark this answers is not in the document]
Princesseuh: If you don't mind, I think I'll leave this part for @Jutanium as part of https://github.com/withastro/docs/issues/1482!

Explaining why TypeScript is important (as a whole) is a super extensive topic, and even in the context of Astro projects there's a bit of nuance ultimately that I believe is important as even if you don't plan on writing any TypeScript, TypeScript is still what powers completions, diagnostics and other stuff so it's important that it "understands" your project

src/pages/en/guides/typescript.md
sarah11918: > I'll take the structure here and `extends` from it

Careful! We're gonna start getting you to write Docs Congrats messages!  ;)

Yes, please do make the statement accurate! I noticed in the template that `strict` extends from `base`, so I assumed that the properties mostly extended each other in order (or at least, both were extended from base). So while I didn't explicitly say that `strict = base plus a strict setting`, I thought maybe that would come across, while emphasizing that the features in base gave you stuff that was useful even if you didn't write any TS in your project. You're right though, it should not sound like base and ONLY base gives you that!

src/pages/en/guides/typescript.md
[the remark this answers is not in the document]
Princesseuh: Aah, I actually changed my mind on this right after sending the message, nevermind I'll do this

sarah11918:
This looks great @Princesseuh! Thanks so much for even tackling the first paragraph. I'm happy for you to merge when you're happy!

sarah11918 (approved):

PR #1521, opened 2022-09-09
Replace `minor` with `en-only` in i18n status ignore list

delucis:

- Something else!

#### Description

We have been using `minor` as a keyword for PRs that should not trigger updates to #438.
@sarah11918 pointed out that this is vague and often conflicts with people thinking a “minor” change is just a “small” change, but for us we’ve overloaded that word with an additional meaning.

This PR _removes_ `minor` from the i18n status script and adds a new `en-only` keyword we can use. Worth noting that we’ve also been using `minor` to mark fixes in other languages that shouldn’t mark a page as updated (typo fix that doesn’t update to match the English for example). We could also add `de-only`, `fr-only` keywords here if you think that is needed.


Yan-Thomas (approved):
Thanks @delucis, LGTM!

sarah11918 (approved):
I think this makes a lot of sense.

I can see that phrases like "broken link" or the word "typo" will *still* work, so I think that should catch especially one-off PRs from people who don't read this guide. 

Follow up: We should look at changing the PR template in GitHub to mention this new `en-only`.

PR #1534, opened 2022-09-11
Delete outdated mention of `--experimental-integrations`

delucis:

- New or updated content

#### Description

- Delete mention of the `--experimental-integrations` flag which is no longer needed.
- The Spanish translation also included this information and is otherwise completely up-to-date so I’ve removed this there too.
- `fr`, `pt-BR` and `zh-CN` also still include this, but those pages also need other updates, so I’ll leave that to our translators


sarah11918:
LGTM!  💜

PR #1538, opened 2022-09-12
Document HTML Components and Clarify Page Types

### LINKED ISSUE #1418: Document HTML Support
Specifically, mention _why_ you might want to use HTML. @natemoo-re, thoughts?

Documenting: https://github.com/withastro/astro/pull/3867 (which had Docs: TODO)

Jutanium:

- New or updated content

#### Description

- Closes #1418
- Adds a section in `Astro Components `for HTML Components
- Changes the overview of `Pages` to explain that there are different page types (thanks @delucis!)
- Adds an `Astro Pages` section in `Pages`, moving `Page HTML` and `Leveraging Page Layouts` under that. These subsections don't describe all pages, just `.astro` pages, so it doesn't make sense for them to be subsections of the overview.


src/pages/en/core-concepts/astro-components.md
```diff
@@ -486,12 +486,22 @@ Note that this approach skips the JavaScript processing, bundling and optimizati
 Astro detects these JavaScript client-side imports and then builds, optimizes, and adds the JS to the page automatically.
 
 ```astro
+
```
delucis, line 489: Nit: this’ll get stripped out anyway, but no need for the extra newline.
```suggestion
```

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -5,17 +5,30 @@ description: An introduction to Astro pages
 i18nReady: true
 ---
 
-**Pages** are a special type of [Astro component](/en/core-concepts/astro-components/) that live in the `src/pages/` subdirectory. They are responsible for handling routing, data loading, and overall page layout for every HTML page in your website.
+**Pages** are special files that live in the `src/pages/` subdirectory. They are responsible for handling routing, data loading, and overall page layout for every HTML page in your website.
```
delucis, line 8: Question: are the files “special”? Or is it that `src/pages/` is special? I wonder if “special” here suggests you need to do something different to these files but actually putting them in `pages` was already all that was needed.

```suggestion
**Pages** are files that live in the `src/pages/` subdirectory of your Astro project. They are responsible for handling routing, data loading, and overall page layout for every HTML page in your website.
```

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -5,17 +5,30 @@ description: An introduction to Astro pages
...
 
 ### File-based routing
 
-Astro leverages a routing strategy called **file-based routing.** Every `.astro` file in your `src/pages` directory becomes a page or an endpoint on your site based on its file path.
+Astro leverages a routing strategy called **file-based routing.**. Each page file in your `src/pages` directory becomes an endpoint on your site based on its file path.
```
delucis, line 12: Nit: I think “file” is enough here maybe? Also do we have the convention of using a trailing slash on directories like in the opening paragraph?

```suggestion
Astro leverages a routing strategy called **file-based routing.**. Each file in your `src/pages/` directory becomes an endpoint on your site based on its file path.
```

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -5,17 +5,30 @@ description: An introduction to Astro pages
...
 
 Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your component template to link between pages.
 
-
 📚 Read more about [Routing in Astro](/en/core-concepts/routing/).
 
+### Supported page files 
+
+Astro supports the following file types in the `src/pages` directory:
```
delucis, line 20: Not sure if this is convention but @sarah11918 can confirm!

```suggestion
Astro supports the following file types in the `src/pages/` directory:
```

src/pages/en/core-concepts/astro-components.md
```diff
@@ -486,12 +486,22 @@ Note that this approach skips the JavaScript processing, bundling and optimizati
...
 
+## HTML Components
+
+Astro supports using `.html` files instead of `.astro` files for components. These can be imported like Astro components, but don't support Astro features. As such:
+- They don't support frontmatter, server-side imports, or dynamic expressions.
+- Any `<script>` tags are left unbundled, treated as if they had `is:inline`. 
+- They can only [reference assets that are in the `public/` folder.](/en/guides/images/#public)
+- [Slots](/en/core-concepts/astro-components/#slots) work as they do in Astro components. If you'd like to use [`Web Components slots feature`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) instead, add `is:inline` to your `<slot>` element.
+
+You may want to use HTML Components if you're migrating from a site built without a framework, or if you want to ensure that your component has no dynamic features.
```
sarah11918, line 504:
```suggestion
## HTML Components

Astro supports importing and using `.html` files as components or placing these files within the `src/pages` subdirectory as pages. You may want to use HTML components if you're reusing code from an existing site built without a framework, or if you want to ensure that your component has no dynamic features.

HTML components must contain only valid HTML, and therefore lack key Astro component features:
- They don't support frontmatter, server-side imports, or dynamic expressions.
- Any `<script>` tags are left unbundled, treated as if they had `is:inline`. 
- They can only [reference assets that are in the `public/` folder.](/en/guides/images/#public)

:::note
A [`<slot />` element](/en/core-concepts/astro-components/#slots) inside an HTML component will work as it does in Astro component. In order to use the [HTML Web Component Slot](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) element instead, add `is:inline` to your `<slot>` element.
:::
```

Not sure all of the above is entirely the correct nuance, but here's my thinking behind some changes above that I'm not married to, but are illustrative of what came to mind:

- took out "instead" of astro files because presumably you can use both? (If not, then maybe that should be a bigger deal?)
- spent a LOT of time thinking about this "migrating an older site" and am leaning towards the specific "and therefore wanting to reuse some existing code" angle of it. Migrating an older site doesn't necessarily mean keeping the same code, and therefore doesn't necessarily mean needing an HTML component (so, want to avoid the confusion of someone thinking they need to/should use them?). I think these HTML components become useful *if* you are bringing that old code over *and wanting as little of a rewrite as possible*, just not simply by virtue of the fact you're migrating an older, html-based site?
- put the reason you might want to use them up front, so people know why they're even reading this or might need this
- assuming this is true, I'd preface the "things you can't do that you can in Astro" with the brief reminder that these are just straight HTML, therefore...
- OMG SLOTS LOL. I *think* the distinction to make here is that there's an "Astro Slot" and a regular HTML slot element, and the cautionary note needs to be if you're using a slot, it's gonna be treated like an Astro slot by default. BUT, if you want it to be treated like the regular web slot element, that's cool: add `is:inline`. I also thought about making this a note since it's maybe worth a call out that esp if you're bringing old code over that has a slot, you should make sure you know it's NOT going to be that default web component? Not sure, but wanted to illustrate the option.

UPDATED after reading the Pages edit to include mention of these being pages, too. I did not cram in "must contain a full proper document of HTML" but am not opposed if you want to find a way to do that.

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -5,17 +5,30 @@ description: An introduction to Astro pages
...
+- [`.astro`](#astro-pages)
+- [`.md`](#markdown-pages)
+- `.mdx` (with the [MDX Integration installed](/en/guides/markdown-content/#mdx-features))
+- [`.js`/`.ts`](/en/core-concepts/astro-pages/#non-html-pages)
+- `.html` (following the same restrictions as [HTML Components](/en/core-concepts/astro-components/#html-components))
```
sarah11918, line 25:
```suggestion
- [`.html`](/en/core-concepts/astro-components/#html-components)
```
I don't think anything from the HTML components section describes anything that would limit it being a page? Yeah, the HTML component section says it's not an Astro component, but... that's kind of what this entire list is about: other things that are not astro files that are pages. I'd be tempted just to remove it. And NOW I'm going to revist that section and add a quick nod to being a full page.

sarah11918 (commented):
Thanks @Jutanium! I love this additional content and have THOUGHTS for your consideration below.  😄

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -5,17 +5,30 @@ description: An introduction to Astro pages
...
+
+
+## Astro Pages
+
+Astro pages use the `.astro` file extension and support the same features as [Astro components](/en/core-concepts/astro-components/).
+
 ### Page HTML
```
delucis, line 32: Reading through this I’m struck that the heading “Page HTML” follows not long after we mention `.html` pages. Do you think that matters? Would renaming that heading help clarity at all or is it not an issue?

delucis (commented):
Yay! Glad we’re finally documenting this. I think you’ve done a great job scoping out where to place this too. Couple of nits and questions but otherwise looks really helpful 🙌

src/pages/en/core-concepts/astro-pages.md
sarah11918: I agree. I'd remove it entirely. It made more sense when we were really ONLY talking about Astro pages here.

Looking at the right sidebar now, I'm also considering whether that layouts subheading deserves to be there, too... will report back!

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -5,17 +5,30 @@ description: An introduction to Astro pages
 i18nReady: true
 ---
 
-**Pages** are a special type of [Astro component](/en/core-concepts/astro-components/) that live in the `src/pages/` subdirectory. They are responsible for handling routing, data loading, and overall page layout for every HTML page in your website.
+**Pages** are special files that live in the `src/pages/` subdirectory. They are responsible for handling routing, data loading, and overall page layout for every HTML page in your website.
 
 ### File-based routing
```
sarah11918, line 10:
```suggestion
## File-based routing
```
Suggesting this is now ##-level, since it's no longer a ### subsection of (implied) Astro pages

src/pages/en/core-concepts/astro-pages.md
sarah11918, line 32:
```suggestion
```
Suggest deleting this entirely. Not needed as a heading anymore, and potentially more confusing now that we have actual HTML pages discussed on this page.

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -5,17 +5,30 @@ description: An introduction to Astro pages
...
 
-Astro leverages a routing strategy called **file-based routing.** Every `.astro` file in your `src/pages` directory becomes a page or an endpoint on your site based on its file path.
+Astro leverages a routing strategy called **file-based routing.**. Each page file in your `src/pages` directory becomes an endpoint on your site based on its file path.
 
 Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your component template to link between pages.
 
-
 📚 Read more about [Routing in Astro](/en/core-concepts/routing/).
 
+### Supported page files
```
sarah11918, line 18:
```suggestion
## Supported page files 
```
In keeping with the page restruture, I think I'd pull this up to a ## heading, too

src/pages/en/core-concepts/astro-pages.md
sarah11918: Sarah can, and does, confirm!

src/pages/en/core-concepts/astro-components.md
Jutanium: I agree with every point. The slot disclaimer does make sense as a note rather than part of the limitations list!

src/pages/en/core-concepts/astro-pages.md
Jutanium: Yep, `special` was a holdover from "special kind of Astro component", but even there it might've been unnecessary!

src/pages/en/core-concepts/astro-pages.md
delucis: I don’t think Astro likes when you put non-page files in `src/pages/` IIRC? So in practice _unless_ using the `_` prefix trick, you should have _only_ pages in `src/pages/`?

I guess similar to “special”, I’m not sure sure how I make my file a _page_ file? So wanted to focus on location rather than file type.

src/pages/en/core-concepts/astro-pages.md
delucis: “Required Page Content”?

delucis: (ah, just saw the recommendation to delete — that works too!)

src/pages/en/core-concepts/astro-pages.md
sarah11918: Chaos monkey! What if we swap positions of supported page file types and file-based routing, so that when we get to this line in routing, people already have "supported file types" in their head? Does this then become easier to write/reference?

sarah11918:
I still want to delete "Leveraging Page Layouts" heading too. I don't think we need it, and I think it clutters the sidebar. Did you intentionally leave it in?

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -145,6 +150,11 @@ export const get: APIRoute = ({ params, request }) => {
 
 ```
 
+## HTML Pages
+
+You can place a `.html` file in the `/src/pages/` directory. This will be treated as an [HTML Component](/en/core-concepts/astro-components/#html-components) and turned into a page.
```
sarah11918, line 155: Do we need a "provided it is a full HTML document" kind of clarification here?

sarah11918: Are you saying you'd like to come up with one phrase, somewhere between "nothing" and the "existing technical-sounding phrase" and use it in both places? I think that would be fine.

OR if this will still make a page even if it's not a full html page, I'm OK to just leave as is!

Jutanium: Turns out this is just out of date. In my testing, if you don't add `<html>` in either an Astro page or an `html` page, it will get added for you, just like `DOCTYPE`.

sarah11918: OK, then as agreed, let's leave this as is!

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -50,12 +57,6 @@ import MySiteLayout from '../layouts/MySiteLayout.astro';
...
 
-#### Modifying `<head>`
-
-Note that using a `<head>` tag works like any other HTML tag: it does not get moved to the top of the page. We recommend writing `<head>` and its contents in your top-level layout.
-
-
-
 ## Markdown Pages
 
 Astro also treats any Markdown (`.md`) files inside of `/src/pages/` as pages in your final website. These are commonly used for text-heavy pages like blog posts and documentation.
```
sarah11918, line 62:
```suggestion
## Markdown/MDX Pages

Astro also treats any Markdown (`.md`) and (`.mdx`) files inside of `/src/pages/` as pages in your final website. These are commonly used for text-heavy pages like blog posts and documentation.
```

Jutanium: In the table of contents, I listed `.md` and `.mdx` as separate and linked out to the `MDX` integrations page. Should I combine them and link them both to this?

src/pages/en/core-concepts/astro-pages.md
Jutanium: Swapping! 🙊

src/pages/en/core-concepts/astro-pages.md
sarah11918: I would leave the links as-is above (since this doesn't really say much!) 😅  BUT include them both here together, since they work the same way. It would be weird to have one without the other.

src/pages/en/core-concepts/astro-components.md
```diff
@@ -492,6 +492,18 @@ Astro detects these JavaScript client-side imports and then builds, optimizes, a
...
 ```
 
+## HTML Components
+
+Astro supports importing and using `.html` files as components or placing these files within the `src/pages` subdirectory as pages. You may want to use HTML components if you're reusing code from an existing site built without a framework, or if you want to ensure that your component has no dynamic features.
+
+HTML components must contain only valid HTML, and therefore lack key Astro component features:
+- They don't support frontmatter, server-side imports, or dynamic expressions.
+- Any `<script>` tags are left unbundled, treated as if they had `is:inline`. 
+- They can only [reference assets that are in the `public/` folder.](/en/guides/images/#public)
```
sarah11918, line 502:
```suggestion
- They can only [reference assets that are in the `public/` folder](/en/guides/images/#public).
```

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -5,20 +5,28 @@ description: An introduction to Astro pages
...
+- [`.astro`](#astro-pages)
+- [`.md`](#markdownmdx-pages)
+- `.mdx` (with the [MDX Integration installed](/en/guides/integrations-guide/mdx/#installation))
+- [`.js`/`.ts`](#file-routes)
+- [`.html`](#html-pages)
 
-Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your component template to link between pages.
+## File-based routing
+
+Astro leverages a routing strategy called **file-based routing**. Each file in your `src/pages` directory becomes an endpoint on your site based on its file path.
```
sarah11918, line 21:
```suggestion
Astro leverages a routing strategy called **file-based routing**. Each file in your `src/pages/` directory becomes an endpoint on your site based on its file path.
```

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -50,17 +56,11 @@ import MySiteLayout from '../layouts/MySiteLayout.astro';
...
+## Markdown/MDX Pages
 
+Astro also treats any Markdown (`.md`) files inside of `src/pages/` as pages in your final website. If you have the [MDX Integration installed](/en/guides/integrations-guide/mdx/#installation), it also treats MDX (`.mdx`) files the same way. These are commonly used for text-heavy pages like blog posts and documentation.
 
-## Markdown Pages
-
-Astro also treats any Markdown (`.md`) files inside of `/src/pages/` as pages in your final website. These are commonly used for text-heavy pages like blog posts and documentation.
-
-Page layouts are especially useful for [Markdown files.](#markdown-pages) Markdown files can use the special `layout` front matter property to specify a [layout component](/en/core-concepts/layouts/) that will wrap their Markdown content in a full `<html>...</html>` page document.
+Page layouts are especially useful for [Markdown files.](#markdownmdx-pages) Markdown files can use the special `layout` front matter property to specify a [layout component](/en/core-concepts/layouts/) that will wrap their Markdown content in a full `<html>...</html>` page document.
```
sarah11918, line 63:
```suggestion
Page layouts are especially useful for [Markdown files](#markdownmdx-pages). Markdown files can use the special `layout` front matter property to specify a [layout component](/en/core-concepts/layouts/) that will wrap their Markdown content in a full `<html>...</html>` page document.
```

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -145,6 +141,10 @@ export const get: APIRoute = ({ params, request }) => {
 
 ```
 
+## HTML Pages
+
+You can place a `.html` file in the `/src/pages/` directory. This will be treated as an [HTML Component](/en/core-concepts/astro-components/#html-components) and turned into a page.
```
sarah11918, line 146:
```suggestion
Files with the `.html` file extension can be placed in the `src/pages/` and used directly as pages on your site. Note that some key Astro features are not supported in [HTML Component](/en/core-concepts/astro-components/#html-components).
```

Kind of combining how we describe Astro pages with the link to seeing HTML components for the things Astro won't do with these files.

src/pages/en/core-concepts/astro-pages.md
```diff
@@ -145,6 +141,10 @@ export const get: APIRoute = ({ params, request }) => {
 
 ```
 
+## HTML Pages
+
+Files with the `.html` file extension can be placed in the `src/pages/` and used directly as pages on your site. Note that some key Astro features are not supported in [HTML Component](/en/core-concepts/astro-components/#html-components).
```
sarah11918, line 146:
```suggestion
Files with the `.html` file extension can be placed in the `src/pages/` and used directly as pages on your site. Note that some key Astro features are not supported in [HTML Components](/en/core-concepts/astro-components/#html-components).
```

sarah11918:
LGTM DAN

Issue #1399, opened 2022-08-24
[TypeScript update] "as props" no longer needed


sarah11918:
With recent updates, you no longer need to write `as Props` when using Typescript.

Everything else in this code example should stay the same, and no changes to the text are necessary. Just remove "as Props" from the code example AND be sure to also update the hightlighting rule (first line of the code sample) accordingly!

https://docs.astro.build/en/guides/typescript/#component-props

![Screenshot 2022-08-24 11 45 49](https://user-images.githubusercontent.com/5098874/186463145-0c18e05a-5881-4504-84aa-c8cd55d33843.png)

PR #1543, opened 2022-09-13
Update typescript.md

ElianCodes:

- adds `as Props`

#### Description

Don't we need to use the `as Props` to actually get the types?


sarah11918:
Not anymore @ElianCodes!  Just changed recently! https://github.com/withastro/docs/issues/1399

So, closing this one, but thanks for having a watchful eye over the docs!  💜

PR #291, opened 2022-03-31
Update API Reference

natemoo-re:
Updates API reference page to bring it inline with v1.0.0 Beta changes.

PR #1552, opened 2022-09-13
`Astro.glob()`: `draft:true` warning

mrienstra:
There is a warning here: https://docs.astro.build/en/guides/markdown-content/#markdown-drafts

Seems wise to have it here as well: https://docs.astro.build/en/reference/api-reference/#astroglob

... Which is what this PR does. Happy to open an issue instead if it doesn't flow right, I was a little uncertain where exactly to put it.

This warning could instead be in the existing ["Markdown Files" subsection](https://docs.astro.build/en/reference/api-reference/#markdown-files), but I feel like it might seem a little random -- that subsection already feels a little random, as it was a bit unclear to me initially why it was there at all. Looks like it was added in https://github.com/withastro/docs/pull/291. Might open an issue about making it a little more clear, like tweaking the title or adding a short intro paragraph about where it's going.

### DIFF
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -36,6 +36,10 @@ const posts = await Astro.glob('../pages/post/*.md'); // returns an array of pos
 
 `.glob()` can't take variables or strings that interpolate them, as they aren't statically analyzable. (See [the troubleshooting guide](/en/guides/troubleshooting/#supported-values) for a workaround.) This is because `Astro.glob()` is a wrapper of Vite's [`import.meta.glob()`](https://vitejs.dev/guide/features.html#glob-import).
 
+:::caution[Astro.glob() and Markdown Drafts]
+Although [`draft: true`](/en/guides/markdown-content/#markdown-drafts) will prevent a page from being built on your site at that page route, `Astro.glob()` currently returns **all your Markdown files**.
+:::
+
 :::note
 You can also use `import.meta.glob()` itself in your Astro project. You may want to do this when:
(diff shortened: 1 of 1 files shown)


mrienstra:
Deploy preview link leading directly to changed section:
https://deploy-preview-1552--astro-docs-2.netlify.app/en/reference/api-reference/#astroglob

sarah11918:
Hi @mrienstra! 

I'm glad you were able to find the warning about draft posts in the Markdown content page! This is where we are choosing to put that info for now, although we do have plans for making sure that page in particular is reviewed again in the coming weeks to make it even more helpful. We have had a lot of product changes under the hool to our MDX and Markdown features, so there have been some content "add-ons" throughout the site for these topics in particular. Making everything easily discoverable is on our radar!

We would not typically include this kind of content in our Reference section. That is intentionally kept concise, and is more of a lookup kind of resource. We typically do not include things like edge cases, warnings (other than deprecation notices), or troubleshooting advice on these pages. They would quickly get cluttered and it would be more difficult for those trying to quickly properties, types etc. We do also have to balance not duplicating material when possible to reduce maintainance costs when features are updated. So, when the material exists in one logical place, we are unlikely to add it again elsewhere.

Thank you for taking the time to help improve the docs, and although I'm closing this PR, I don't think an issue for further exploration is needed. The content is available on the site in an apppropriate place right now, and we will be reorganizing our Markdown/MDX content in the coming weeks. Part of that plan will include highlighting this particular draft feature in a place where people needing that information will find it, including targeted guides for things like enabling draft blog posts, but not adding it to our reference section.

mrienstra:
Really appreciate you taking the time to explain the moving pieces & norms, I'll try to soak that up for future reference! 🪴🚰
_(no response needed)_

PR #785, opened 2022-06-16
Update configuring-astro.md

natemoo-re:
#### What kind of changes does this PR include?

- [ ] Minor content fixes (broken links, typos, etc.)
- [x] New or updated content
- [ ] Translated content
- [ ] Changes to the docs site code
- [ ] Something else!

#### Description

- Adds a new section to `configuring-astro.md` about referencing environment variables.
- Also mentions Astro's ability to load config files from the `config/` directory, a hidden feature that I personally love and hope more people start using!
- Paired with https://github.com/withastro/astro/pull/3617

PR #1567, opened 2022-09-14
Environment variables can't be used in config files

Jutanium:

- New or updated content

#### Description

We closed #785 as the corresponding feature was closed, but the note about Astro being unable to load env variables is still important information, even if we don't have a workaround. This PR restores that note and adds a corresponding one to the environmental variables change.

I could see us only wanting one of those two messages, or rather wanting this to be in Troubleshooting. Let me know!

@natemoo-re, if there's currently a workaround to access env variables in Astro, let me know.

### DIFF
--- a/src/pages/en/guides/environment-variables.md
+++ b/src/pages/en/guides/environment-variables.md
@@ -18,7 +18,7 @@ PUBLIC_ANYBODY=there
 In this example, `PUBLIC_ANYBODY` (accessible via `import.meta.env.PUBLIC_ANYBODY`) will be available in server or client code, while `SECRET_PASSWORD` (accessible via `import.meta.env.SECRET_PASSWORD`) will be server-side only.
 
 :::caution
-`import.meta.env` and `.env` files are not available inside [configuration files](/en/guides/configuring-astro#environment-variables). 
+`import.meta.env` and `.env` files are not available inside [configuration files](/en/guides/configuring-astro/#environment-variables). 
 :::
 
 ## Default environment Variables
--- a/src/pages/en/guides/environment-variables.md
+++ b/src/pages/en/guides/environment-variables.md
@@ -2,6 +2,7 @@
 layout: ~/layouts/MainLayout.astro
 title: Using environment variables
 description: Learn how to use environment variables in an Astro project.
+setup: import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 i18nReady: true
 ---
 
@@ -28,6 +29,7 @@ Astro includes a few environment variables out-of-the-box:
 
 ## Setting environment variables
 
+### `.env` files
 Environment variables can be loaded from `.env` files in your project directory.
 
 You can also attach a mode (either `production` or `development`) to the filename, like `.env.production` or `.env.development`, which makes the environment variables only take effect in that mode.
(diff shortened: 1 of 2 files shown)


delucis (commented):
You can access environment variables using `process.env` in the config file, with the caveat that `.env` files are NOT loaded yet. An environment variable set by a CI service or on the CLI _will_ be accessible though. For example, you could use `process.env.NETLIFY` to toggle a feature only in Netlify builds or use `TOKEN=secrets npm run dev` to expose `process.env.TOKEN`.

I think I've also heard people are using `dotenv` to load their `.env` files, so that could be a workaround if it doesn't have drawbacks — Nate would know best!

delucis:
No, `import.meta.env` is a Vite thing and the config is evaluated in a standard Node environment before Vite is running.

Jutanium:
Added a section on setting variables with the CLI.

I moved the caution in Environment Variables to the overview. It really is an overview, not an introduction, and `import.meta.env` is already brought up by that point.

I wasn't sure whether to document `process.env` in Environment Variables or Configuring Astro. I chose the later because Configuring Astro seems to want to be comprehensive, but I could see other possibilities:
- I could do the opposite where I just have a section in Environment Variables that has all the info, and don't change Configuring Astro. 
- Or I could have a caution in Configuring Astro that links to Environment Variables, where the info actually lies.

src/pages/en/guides/environment-variables.md
```diff
@@ -16,6 +17,10 @@ PUBLIC_ANYBODY=there
 
 In this example, `PUBLIC_ANYBODY` (accessible via `import.meta.env.PUBLIC_ANYBODY`) will be available in server or client code, while `SECRET_PASSWORD` (accessible via `import.meta.env.SECRET_PASSWORD`) will be server-side only.
 
+:::caution
+`import.meta.env` and `.env` files are not available inside [configuration files](/en/guides/configuring-astro/#environment-variables). 
+:::
+
 ## Default environment Variables
```
delucis, line 24: Consistent casing:

```suggestion
## Default environment variables
```

delucis (commented):
I think these are good additions, Dan — thank you!

It’s tricky deciding where to document `process.env` but I think I agree with where you ended up — the most common use case is in the astro config file so makes sense to document it there.

I could leave a bunch more comments on the _rest_ of the environment variables page but I’ll hold off until you say that would be helpful.

delucis:
More comments as requested by @Jutanium (to be added to this PR or tackled later at his discretion).

1. Opening paragraph:

   > Astro uses Vite for environment variables

   I’d probably rewrite to something like
   
   > Astro relies on Vite’s built-in environment variable support
   
   Or even rewrite the whole opening sentence:
   
   > Astro lets you [use any of Vite’s methods](https://vitejs.dev/guide/env-and-mode.html) to get and set environment variables.

2. “Default environment variables” section
   
   Totally a nit but in the first list item I’d write either “`development | production`” or “`development` or `production`” but not use the `|` outside of `<code>`.

2. “Supported file names” code snippet

   Could probably considered a nit, but might reformat as a table or list. At first glance I read it as a continuation of the `.env` file example above it.

4. “Getting environment variables” section

   The browser support tip seems excessively noisy in this context. I’d roll that content into the opening paragraph.

Jutanium:
Thanks @delucis! This page could use a broader reworking (including explaining what Environment Variables actually _are_), but I made minimal changes for now to address your points.

I removed the .env files snippet entirely as it was copy-pasted from Vite docs - I linked to that instead. I also removed the the tip rather than moving it, as we made this point clear as soon as we talked about client-side code in the intro.

delucis (approved):
Thanks Dan! LGTM 🙌

Issue #1407, opened 2022-08-25
Add: Explain how to run a function inside an inline script after the dom is loaded  


paolotiu:
When using an `is:inline` or `define:vars` on a script it becomes inlined in the served code 1:1. So, when someone wants to target an element further down the html (ex. `document.querySelector("#footer")`), they might be inclined to use `defer` or `async` since that's what is taught in vanilla sites. It won't work though since the `defer` or `async` attributes don't actually apply. 

I think we should add in the docs how to achieve this behavior using `document.addEventListener("DOMContentLoaded, fn)`

Willing to make a pr btw

sarah11918:
Thanks for this comment @paolotiu, and for being willing to make a PR! Would you be willing to just type in here what you would PR, so we can take a look at it, and think about where (on which page, in which section) it might actually best go?

sarah11918:
Hi @paolotiu - just following up here: are you still willing to give an example here of what you would PR, and where you think it would go in the docs?

paolotiu:
Hi @sarah11918! Sorry for the late reply. Here's a draft of what I would write.


is:inline doesn't get processed by the astro compiler, so it will load as is where it is placed. This means some dom contents may not have loaded when the script runs. If you want to target dom nodes, it is recommended to use `document.addEventListener("DOMContentLoaded, fn)` . This waits for the whole dom to load before the script runs.

> Note: defer and async attributes currently **do not** affect the script, so using `document.addEventListener("DOMContentLoaded, fn)` is the only way to target dom nodes below the script.

Jutanium:
I'm not sure that documenting how to use `DOMContentLoaded` (and its alternatives) fit within Astro docs (but might be a great addition to a future restructuring where we have room for guides or recipes). Trying to access elements too early is really an HTML/JS concern, and an involved one to explain thoroughly!

As for `defer` and `async` - these do not work on inline script tags in normal HTML, either, as far as I know. That said, guarding the user from the potential gotcha of using `defer` or `async` could be part of the task of documenting `is:inline` and explaining what it means for the script to be inline. I added a bit of clarification in #1568.

Thanks for bringing this to our attention!

paolotiu:
Agree! I think clarifying `is:inline` would be a better since its in the scope of Astro. Thanks for the work dan!

PR #1568, opened 2022-09-14
Script is:inline clarifications

Jutanium:

- New or updated content

#### Description

- To address #1407, explain that `defer` and similar attributes don't work on inlined scripts
- Explain _why_ `define:vars` implies `is:inline`.

### DIFF
--- a/src/pages/en/reference/directives-reference.md
+++ b/src/pages/en/reference/directives-reference.md
@@ -213,7 +213,9 @@ const message = "Astro is awesome!";
 ```
 
 :::caution
-Using `define:vars` on a `<script>` or `<style>` tag implies the [`is:inline` directive](#isinline), which means your scripts or styles won't be bundled and will be inlined directly into the HTML.
+Using `define:vars` on a `<script>` or `<style>` tag implies the [`is:inline` directive](#isinline), which means your scripts or styles won't be bundled and will be inlined directly into the HTML. 
+
+This is because when Astro bundles scripts, it creates one bundle for all instances of the script on the page and runs that once. `define:vars` requires a script to rerun with each set of values, so Astro creates an inline script instead.
 :::
 
 ## Advanced Directives
--- a/src/pages/en/reference/directives-reference.md
+++ b/src/pages/en/reference/directives-reference.md
@@ -164,7 +164,7 @@ By default, Astro will process, optimize, and bundle any `<script>` and `<style>
 
 The `is:inline` directive means that `<style>` and `<script>` tags:
 
-- Will not be bundled into an external file.
+- Will not be bundled into an external file. This means that attributes like `defer` which control the loading of an external file will have no effect.
 - Will not be deduplicated—the element will appear as many times as it is rendered.
 - Will not have its `import`/`@import`/`url()` references resolved relative to the `.astro` file.
 - Will be pre-processed, for example a `<style lang="sass">` attribute will still generate plain CSS.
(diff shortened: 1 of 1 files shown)


Yan-Thomas (approved):
LGTM! Nice job 🙌

sarah11918 (approved):

src/pages/en/reference/directives-reference.md
```diff
@@ -213,7 +213,9 @@ const message = "Astro is awesome!";
 ```
 
 :::caution
-Using `define:vars` on a `<script>` or `<style>` tag implies the [`is:inline` directive](#isinline), which means your scripts or styles won't be bundled and will be inlined directly into the HTML.
+Using `define:vars` on a `<script>` or `<style>` tag implies the [`is:inline` directive](#isinline), which means your scripts or styles won't be bundled and will be inlined directly into the HTML. 
+
+This is because when Astro bundles scripts, it creates one bundle for all instances of the script on the page and runs that once. `define:vars` requires a script to rerun with each set of values, so Astro creates an inline script instead.
```
delucis, line 218: Thanks for this Dan! I wonder if we can make this explanation a little more concrete (bundles, instances, etc. are a bit vague in my mind). Not entirely happy, but maybe we this could be a starting point?

```suggestion
This is because when Astro bundles a script, it includes and runs the script once even if you include the component containing the script multiple times on one page. `define:vars` requires a script to rerun with each set of values, so Astro creates an inline script instead.
```

Jutanium: Thanks @delucis, definitely prefer this way of explaining it - will see if I can make it more concise

Jutanium:
Note to self: clarify that `type=module` protects the scope of a script, so you need to use `is:inline` to add something to the global scope

src/pages/en/reference/directives-reference.md
```diff
@@ -164,7 +164,7 @@ By default, Astro will process, optimize, and bundle any `<script>` and `<style>
 
 The `is:inline` directive means that `<style>` and `<script>` tags:
 
-- Will not be bundled into an external file.
+- Will not be bundled into an external file. This means that attributes like `defer` which control the loading of an external file will have no effect.
```
delucis, line 167: What do you think about linking to some more extensive external documentation here?

```suggestion
- Will not be bundled into an external file. This means that [attributes like `defer`](https://javascript.info/script-async-defer) which control the loading of an external file will have no effect.
```

Link I’m proposing: <https://javascript.info/script-async-defer>

delucis (commented):
LGTM Dan — just left a suggestion.

> Note to self: clarify that `type=module` protects the scope of a script, so you need to use `is:inline` to add something to the global scope

On this, it is possible to add to the global scope explicity from a `type=module` script, but it’s generally considered bad practice because of the async nature of modules (if you can’t control when each script runs, depending on globals across scripts is recipe for trouble).

```astro
<!-- Astro bundles this into a type="module" script -->
<script>
window.foo = 'some global value';
</script>

<!-- equivalent inline version -->
<script is:inline>
var foo = 'some global value';
</script>
```

Basically both of these should be avoided 😅

tony-sull (approved):
lgtm!  Nicely done explaining this clearly without going deep into the weeds :+1:

src/pages/en/reference/directives-reference.md
Jutanium: I like this, it's better than the official MDN docs for explaining this point!

Jutanium:
Everyone reviewed and approved this one, so I'm going to merge and bring any additional changes (regarding the `module` limitations) to a future PR!

@delucis I'm okay with the wordiness of the bundling explanation for now as it's in an advanced aside, but I may revisit it in that PR.

PR #1572, opened 2022-09-15
Adding docs for the new astro:build:generated hook

tony-sull:

- New or updated content

#### Description

- Adding docs for the new hook from [PR #4772](https://github.com/withastro/astro/pull/4772)
- Updates the `astro:build:ssr` hook docs, it is never fired for SSG builds

### DIFF
--- a/src/pages/en/reference/integrations-reference.md
+++ b/src/pages/en/reference/integrations-reference.md
@@ -40,6 +40,7 @@ interface AstroIntegration {
           pages: Map<string, PageBuildData>;
           target: 'client' | 'server';
         }) => void | Promise<void>;
+        'astro:build:generated'?: (options: { dir: URL }) => void | Promise<void>;
         'astro:build:ssr'?: (options: { manifest: SerializedSSRManifest }) => void | Promise<void>;
         'astro:build:done'?: (options: { pages: { pathname: string }[]; dir: URL; routes: RouteData[] }) => void | Promise<void>;
     };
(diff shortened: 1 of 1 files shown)


Yan-Thomas:
LGTM!

src/pages/en/reference/integrations-reference.md
```diff
@@ -273,11 +274,23 @@ The address, family and port number supplied by the [NodeJS Net module](https://
 
 ```
 
+### `astro:build:generated`
+
+**Previous hook:** [`astro:build:setup`](#astrobuildsetup)
+
+**When:** After a production SSG build has completed.
+
+**Why:** To access generated routes and assets **before** build artifacts are cleaned up. This is a very uncommon use case, we recommend using [`astro:build:done`](#astrobuilddone) unless you really need to access the generated files before cleanup.
```
sarah11918, line 283:
```suggestion
**Why:** To access generated routes and assets **before** build artifacts are cleaned up. This is a very uncommon use case. We recommend using [`astro:build:done`](#astrobuilddone) unless you really need to access the generated files before cleanup.
```

src/pages/en/reference/integrations-reference.md
```diff
@@ -273,11 +274,23 @@ The address, family and port number supplied by the [NodeJS Net module](https://
 
 ```
 
+### `astro:build:generated`
+
+**Previous hook:** [`astro:build:setup`](#astrobuildsetup)
+
+**When:** After a production SSG build has completed.
```
delucis, line 281: We usually prefer the term “static” over SSG in docs as the two SSR + SSG acronyms are a bit too easy to confuse.

```suggestion
**When:** After a static production build has completed.
```

delucis: Also, would it be more accurate to say something like this?

> After a static production build has finished generating routes and assets.

I’m not familiar with the hook, but sounds like the build has not totally completed at this point?

delucis (commented):
Seems solid @tony-sull!

Couple of suggestions and a query for you nonetheless 🙌

src/pages/en/reference/integrations-reference.md
tony-sull: Yep, I like that wording better!  Static builds do a bit of cleanup *before* `astro:build:done` is triggered, the only difference with this new hook is that it runs before the cleanup step

sarah11918:
Note to all of us that astro PR #4775 has been "approved" but not merged. This shouldn't merge before the hook is active, but also, SHOULD merge WHEN the hook is active. Everyone keep their eyes open!  😄

tony-sull:
Updated note:  PR #4775 was just merged in and will go out in tomorrow's `minor` release 🎉 We'll be ready to merge this in once the release is cut

PR #1584, opened 2022-09-19
Creating a package: fix commands (minor)

mrienstra:

#### Description

Looks like commands were copied from https://github.com/withastro/docs/blob/62470c5/src/pages/en/guides/publish-to-npm.md?plain=1#L22-L29 & only the first command (for npm) was changed from `my-new-component-directory` to `demo`, all 3 should say `demo`.


delucis (approved):
Thanks for catching that! 🙌

PR #1593, opened 2022-09-21
Add redirects for some old paths still being visited

delucis:

#### Description

This PR adds some entries to our Netlify redirects config to redirect on some pages we’re seeing 404s for:

- pages starting with `/docs/...` will now redirect to the same URL minus the `/docs/` prefix
- an old “quick start” URL will now redirect to the install guide


sarah11918:
LGTM, thanks for this @delucis!

PR #1594, opened 2022-09-21
Only Strings are accepted as parameters.

### LINKED ISSUE #1591: Param can be a numer or not?
From the [documentation](https://docs.astro.build/en/reference/api-reference/#params):
```
For example, suppose that you have a page at src/pages/posts/[id].astro. If you export getStaticPaths from this page and return the following for paths:

---
  return [
    { params: { id: '1' } },
    { params: { id: '2' } },
    { params: { id:  3  } }  // Can be a number too!
  ];
}

const { id } = Astro.params;
---
<h1>{id}</h1>
Then Astro will statically generate posts/1, posts/2, and posts/3 at build time.
```
So the comment is clear: **``{ params: { id:  3  } }  // Can be a number too!``**

but if executing this code, I get the following error, so looks like it cannot be a number!:
```
07:09:57 [getStaticPaths] invalid path param: id. A string value was expected, but got `3`.
```

What is wrong? The documentation or Astro engine? I'm using **astro 1.2.3**

angelcervera:

- Minor content fixes

#### Description

`params` only accepts strings, but the documentation says that numbers are allowed as well.

Closes #1591

### DIFF
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -360,7 +360,7 @@ The `getStaticPaths()` function executes in its own isolated scope once, before 
 
 The `params` key of every returned object tells Astro what routes to build. The returned params must map back to the dynamic parameters and rest parameters defined in your component filepath.
 
-`params` are encoded into the URL, so only strings and numbers are supported as values. The value for each `params` object must match the parameters used in the page name.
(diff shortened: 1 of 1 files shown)


delucis (approved):
Perfect, thank you!

PR #1596, opened 2022-09-21
Add clarification to dynamic tags about hydration directives

delucis:

- New or updated content

#### Description

Matthew clarified in [a support thread](https://discord.com/channels/830184174198718474/1021446382982266910/1021468769102598206) the other day that the dynamic tag pattern we document won’t work for components using `client:*` hydration directives.

This PR updates that section in the component docs to note this limitation.

### DIFF
--- a/src/pages/en/core-concepts/astro-components.md
+++ b/src/pages/en/core-concepts/astro-components.md
@@ -183,11 +183,11 @@ const Component = MyComponent;
 <Component /> <!-- renders as <MyComponent /> -->
 ```
 
-There are a some things to bear in mind when using dynamic tags:
+When using dynamic tags:
 
-- **Variable names must be capitalized**, for example use `Element`, not `element`. Otherwise, Astro will try to render your variable name as a literal HTML tag.
+- **Variable names must be capitalized.** For example, use `Element`, not `element`. Otherwise, Astro will try to render your variable name as a literal HTML tag.
 
-- **Hydration directives are not supported.** When using [`client:*` hydration directives](/en/core-concepts/framework-components/#hydrating-interactive-components), Astro needs to know what components to bundle for production, and the dynamic tag pattern prevents this from working.
+- **Hydration directives are not supported.** When using [`client:*` hydration directives](/en/core-concepts/framework-components/#hydrating-interactive-components), Astro needs to know which components to bundle for production, and the dynamic tag pattern prevents this from working.
 
 ### Fragments & Multiple Elements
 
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/astro-components.md
```diff
@@ -183,9 +183,11 @@ const Component = MyComponent;
...
 ```
 
-:::note
-Variable names must be capitalized (`Element`, not `element`), for this to work. Otherwise, Astro will try to render your variable name as a literal HTML tag.
-:::
+There are a some things to bear in mind when using dynamic tags:
+
+- **Variable names must be capitalized**, for example use `Element`, not `element`. Otherwise, Astro will try to render your variable name as a literal HTML tag.
+
+- **Hydration directives are not supported.** When using [`client:*` hydration directives](/en/core-concepts/framework-components/#hydrating-interactive-components), Astro needs to know what components to bundle for production, and the dynamic tag pattern prevents this from working.
```
sarah11918, line 190:
```suggestion
When using dynamic tags:

- **Variable names must be capitalized**. For example, use `Element`, not `element`. Otherwise, Astro will try to render your variable name as a literal HTML tag.

- **Hydration directives are not supported.** When using [`client:*` hydration directives](/en/core-concepts/framework-components/#hydrating-interactive-components) with UI framework components, Astro needs to know which components to bundle for production, and the dynamic tag pattern prevents this from working.
```
Making at least this little suggestion to add UI framework components (as a way to again make it super clear these only work with non-Astro compoents)

Still mulling over the 2nd point in general. Since, it's not only "when using a client directive" that Astro needs to know stuff. It always needs to know stuff! So, trying to envision maybe reversing the clause to be something closer to the form:

Astro needs to know... therefore...

Jutanium: I think in this case, it's _only_ when using a client directive that "Astro needs to know" what to bundle for the browser. Astro needs to know this extra info ahead of time. If you're not using a client directive, Astro can just run your code and see what happens.

sarah11918:
Thanks for this @delucis! Left a couple of quick suggestions, but LGTM!

src/pages/en/core-concepts/astro-components.md
delucis: Thanks! Going to propose this then as I like most of these changes but don’t think we need to add “with UI framework components” because you can’t use `client:` directives with any other kind of components.

```suggestion
When using dynamic tags:

- **Variable names must be capitalized**. For example, use `Element`, not `element`. Otherwise, Astro will try to render your variable name as a literal HTML tag.

- **Hydration directives are not supported.** When using [`client:*` hydration directives](/en/core-concepts/framework-components/#hydrating-interactive-components), Astro needs to know which components to bundle for production, and the dynamic tag pattern prevents this from working.
```

PR #1599, opened 2022-09-21
Document new types for set:html

matthewp:

- New or updated content

#### Description

- This pairs with a new feature in https://github.com/withastro/astro/pull/4832

### DIFF
--- a/src/pages/en/reference/directives-reference.md
+++ b/src/pages/en/reference/directives-reference.md
@@ -80,7 +80,7 @@ import api from '../db/api.js';
 
 `set:html={Promise<Response>}` injects a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) into an element.
 
-This is most useful for use with `fetch()`. One use case is fetching old posts from a previous static-site generator.
+This is most helpful when using `fetch()`. For example, fetching old posts from a previous static-site generator.
 
 ```astro
 <article set:html={fetch('http://example/old-posts/making-soup.html')}></article>
--- a/src/pages/en/reference/directives-reference.md
+++ b/src/pages/en/reference/directives-reference.md
@@ -80,7 +80,7 @@ import api from '../db/api.js';
 
 `set:html={Promise<Response>}` injects a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) into an element.
 
-This is most useful for use with `fetch()`. One use-case is fetching old posts from a previous static-site generator.
+This is most useful for use with `fetch()`. One use case is fetching old posts from a previous static-site generator.
 
 ```astro
 <article set:html={fetch('http://example/old-posts/making-soup.html')}></article>
--- a/src/pages/en/reference/directives-reference.md
+++ b/src/pages/en/reference/directives-reference.md
@@ -67,6 +67,25 @@ const cmsContent = await fetchHTMLFromMyCMS();
 <Fragment set:html={cmsContent}>
 ```
 
+`set:html={Promise<string>}` injects an HTML string into an element that is wrapped in a promise.
+
+This can be used to inject HTML stored externally, such as a database.
+
+```astro
+---
+import api from '../db/api.js';
+---
+<article set:html={api.getArticle(Astro.props.id)}></article>
+```
+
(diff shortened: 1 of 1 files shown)


src/pages/en/reference/directives-reference.md
```diff
@@ -67,6 +67,25 @@ const cmsContent = await fetchHTMLFromMyCMS();
 <Fragment set:html={cmsContent}>
 ```
 
+`set:html={Promise<string>}` injects an HTML string into an element that is wrapped in a promise.
```
Yan-Thomas, line 70: For terms like Response, Request, and others, we generally capitalize the word, so I think it's nice to follow the same pattern here. It also makes it easier for people from other languages to recognize we are not saying "promise" but rather the JavaScript object "Promise"
```suggestion
`set:html={Promise<string>}` injects an HTML string into an element that is wrapped in a Promise.
```

src/pages/en/reference/directives-reference.md
```diff
@@ -67,6 +67,25 @@ const cmsContent = await fetchHTMLFromMyCMS();
...
+```astro
+---
+import api from '../db/api.js';
+---
+<article set:html={api.getArticle(Astro.props.id)}></article>
+```
+
+`set:html={Promise<Response>}` injects a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) into an element.
+
+This is most useful for use with `fetch()`. One use-case is fetching old posts from a previous static-site generator.
```
Yan-Thomas, line 83: Not completely sure, but looks like the correct spelling is "use case" (without a hyphen), though we do have both "use-case" and "use case" in the docs, the latter having more appearances.
```suggestion
This is most useful for use with `fetch()`. One use case is fetching old posts from a previous static-site generator.
```

Yan-Thomas (commented):
Thanks @matthewp 🙌, I added some small suggestions, let me know what you think!

src/pages/en/reference/directives-reference.md
```diff
@@ -67,6 +67,25 @@ const cmsContent = await fetchHTMLFromMyCMS();
 <Fragment set:html={cmsContent}>
 ```
 
+`set:html={Promise<string>}` injects an HTML string into an element that is wrapped in a Promise.
+
+This can be used to inject HTML stored externally, such as a database.
```
sarah11918, line 72:
```suggestion
This can be used to inject HTML stored externally, such as in a database.
```
I think the HTML would be *in* a database, not the database itself?

src/pages/en/reference/directives-reference.md
```diff
@@ -67,6 +67,25 @@ const cmsContent = await fetchHTMLFromMyCMS();
...
+```astro
+---
+import api from '../db/api.js';
+---
+<article set:html={api.getArticle(Astro.props.id)}></article>
+```
+
+`set:html={Promise<Response>}` injects a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) into an element.
+
+This is most useful for use with `fetch()`. One use case is fetching old posts from a previous static-site generator.
```
delucis, line 83: A small nit, but there’s “useful”, “use” and “use case” in quick succession here so we can probably make this a little clearer. Does this look OK?

```suggestion
This is most helpful when using `fetch()`. For example, fetching old posts from a previous static-site generator.
```

src/pages/en/reference/directives-reference.md
```diff
@@ -67,6 +67,25 @@ const cmsContent = await fetchHTMLFromMyCMS();
...
+This can be used to inject HTML stored externally, such as a database.
+
+```astro
+---
+import api from '../db/api.js';
+---
+<article set:html={api.getArticle(Astro.props.id)}></article>
+```
+
+`set:html={Promise<Response>}` injects a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) into an element.
```
delucis, line 81: What exactly gets injected here? The full contents of `await response.text()`? Or do we do something more to that content, like only what’s in `<body>` or something for fetched HTML?

delucis (commented):
Thanks @matthewp! Couple of small notes here.

src/pages/en/reference/directives-reference.md
matthewp: The full contents. It's not `await response.text()` though, it streams in the contents.

PR #1611, opened 2022-09-22
Document `create-astro --template`

mayank99:

- New or updated content

#### Description

Added a brief mention about `--template` to the auto install page, and a shell example to the hidden `integrations.md` page.

Not sure if that's the best place to do it, but I didn't want to clutter the install page, and at least it will show up in search now.

screen recording showing that searching for template takes you to the newly added paragraph


delucis (approved):
Thanks @mayank99! I wish we had a better place for this but this LGTM 👍

PR #1620, opened 2022-09-26
Overhaul CLI reference

delucis:

- New or updated content

#### Description

I noticed https://github.com/withastro/astro/issues/4870 this morning and it triggered me to look through our CLI reference.

I’ve tried to improve it by

- adding an introductory section that explains how to use the CLI more generally, including passing flags
- correcting some errors (removing mention of `--reload` which doesn’t seem to exist anymore, moving `--drafts` up to the `astro build` section as it only works there)
- breaking up “global” flags into “common flags” (to be combined with commands) and “global flags” (standalone, e.g. `astro --version`)
- adding links where relevant
- adding a warning not to use `dev` in production
- explaining why `--host` can be useful

I’ve also deleted an old Spanish translation of the CLI reference as we now have a newer version.


src/pages/en/reference/cli-reference.md
```diff
@@ -2,41 +2,106 @@
...
-**Flags**
+Sets which network IP addresses the dev server should listen on (i.e. non-localhost IPs). This can be useful for testing your project on local devices like a mobile phone during development.
 
-#### `--port`
+- `--host` — listen on all addresses, including LAN and public addresses
+- `--host [custom-address]` — expose on a network IP address at `[custom-address]`
 
-Specifies which port to run on. Defaults to `3000`.
+:::caution
+Do not use the `--host` flag to expose the dev server in a production environment. It is designed for local use while developing your site only.
```
Jutanium, line 77:
```suggestion
Do not use the `--host` flag to expose the dev server in a production environment. The dev server is designed for local use while developing your site only.
```

src/pages/en/reference/cli-reference.md
```diff
@@ -48,15 +113,15 @@ This command only checks types within `.astro` files.
...
 
-### `astro docs`
+## `astro docs`
 
 Launches the Astro Docs website directly from the terminal.
 
-### `astro telemetry`
+## `astro telemetry`
 
 Sets telemetry configuration for the current user. Telemetry is anonymous data that provides insights into which features are most often used.
```
Jutanium, line 126: Taking the opportunity to try and make it clear that Astro telemetry monitors _you_ as the CLI user, not your end user, and that the insights are for "us" (Astro) not something you can view
```suggestion
Sets telemetry configuration for the current CLI user. Telemetry is anonymous data that provides us insights into which Astro features are most often used.
```

src/pages/en/reference/cli-reference.md
```diff
@@ -2,41 +2,106 @@
...
 
-Runs  Astro's `dev` server. It starts an HTTP server which responds to requests for routes or pages that are specified within `src/pages` directory (unless overridden by your `pages` option set in the project [configuration](/en/reference/configuration-reference/)).
+#### `--host [optional host address]`
 
-**Flags**
+Sets which network IP addresses the dev server should listen on (i.e. non-localhost IPs). This can be useful for testing your project on local devices like a mobile phone during development.
 
-#### `--port`
+- `--host` — listen on all addresses, including LAN and public addresses
+- `--host [custom-address]` — expose on a network IP address at `[custom-address]`
```
Jutanium, line 74: Should this be `<custom-address>`, or is the idea that angle brackets are for less specific types

src/pages/en/reference/cli-reference.md
```diff
@@ -82,30 +147,26 @@ astro telemetry clear
...
 astro --config config/astro.config.mjs dev
 ```
 
-### `--root path`
+### `--root <path>`
 
-Specifies the path to the project root. If not specified the current working directory is assumed to be the root.
+Specifies the path to the project root. If not specified, the current working directory is assumed to be the root.
 
 The root is used for finding the Astro configuration file.
```
Jutanium, line 164: We could change the order so that `--root` appears before `--config`, and then we can say that `--config` "specifies the path to the config file _relative to the root_". This line could sound like root is _only_ used for finding the configuration file, and isn't precise about what that means

src/pages/en/reference/cli-reference.md
```diff
@@ -82,30 +147,26 @@ astro telemetry clear
...
 astro --root myRootFolder/myProjectFolder dev
 ```
 
-### `--reload`
-
-Clears the cache (dependencies are built within Astro apps).
-
 ### `--verbose`
 
 Enables verbose logging, which is helpful when debugging an issue.
```
Jutanium, line 172: Could use some clarification on what verbose logging does

src/pages/en/reference/cli-reference.md
```diff
@@ -114,13 +175,13 @@ Enables verbose logging, which is helpful when debugging an issue.
 
 Enables silent logging, which is helpful when you don't want to see Astro logs.
```
Jutanium, line 176:
```suggestion
Enables silent logging, which will run the server without any console output.
```

Jutanium (commented):
This is a much-needed improvement! I left some comments, mostly using this as an opportunity to clarify things that were there before.

src/pages/en/reference/cli-reference.md
```diff
@@ -2,41 +2,106 @@
...
+  </Fragment>
+</PackageManagerTabs>
+
+If you started your project using [the `create astro` wizard](/en/install/auto/#1-run-the-setup-wizard), you can also use the scripts in `package.json` for a shorter version of these commands. See the `README.md` in your project for details of which commands are available.
+
+<PackageManagerTabs>
+  <Fragment slot="npm">
+  ```shell
+  # run the dev server on port 8080 using the `start` script in `package.json`
+  npm start -- --port 8080
```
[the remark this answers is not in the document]
delucis: Fair — I went for this form to match the other package managers but I agree that this makes it easier for folks to switch out to a different script.
```suggestion
  npm run start -- --port 8080
```

src/pages/en/reference/cli-reference.md
```diff
@@ -2,41 +2,106 @@
...
+  ```shell
+  # run the dev server on port 8080 using the `start` script in `package.json`
+  yarn start --port 8080
+  ```
+  </Fragment>
+</PackageManagerTabs>
+
+## `astro dev`
+
+Runs  Astro's development server. It starts an HTTP server which responds to requests for routes or pages that are specified within `src/pages` directory (unless overridden by your `pages` option set in the project [configuration](/en/reference/configuration-reference/)).
```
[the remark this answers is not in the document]
delucis: Makes sense. How’s this read?

```suggestion
Runs  Astro's development server. This is a local HTTP server optimized for a quick development workflow, serving unbundled assets, with support for Hot Module Replacement (HMR), which updates your browser as you make changes in your code editor.
```

src/pages/en/reference/cli-reference.md
delucis: Agree on this. Do you think “Astro team” rather than “us” is clearer?

```suggestion
Sets telemetry configuration for the current CLI user. Telemetry is anonymous data that provides the Astro team insights into which Astro features are most often used.
```

src/pages/en/reference/cli-reference.md
delucis: I _think_ the convention is `[]` for optional parts of the command, `<>` for required, but I’m not wedded to it, as who knows that??

delucis: Also, in this context we’ve just shown the version _without_  an address so I think I’ll go with your suggestion 🙌 

```suggestion
- `--host <custom-address>` — expose on a network IP address at `<custom-address>`
```

src/pages/en/reference/cli-reference.md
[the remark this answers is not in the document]
Jutanium: oh sorry, I didn't mean to just refer to `npm`, but for all three!

src/pages/en/reference/cli-reference.md
Jutanium: I was trying to run `--verbose` myself and couldn't figure out the difference! An example or something to make that clear would be good, but this is definitely NWTWWHB!

src/pages/en/reference/cli-reference.md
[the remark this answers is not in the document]
Jutanium: We could probably cut the "optimized" part, but I'm happy either way. Nit about HMR, which doesn't necessarily refresh the page:

```suggestion
Runs  Astro's development server. This is a local HTTP server that doesn't bundle assets. It uses Hot Module Replacement (HMR) to update your browser as you save changes in your editor.
```

src/pages/en/reference/cli-reference.md
delucis: Seeing this too! Posted on Discord to gather feedback.

src/pages/en/reference/cli-reference.md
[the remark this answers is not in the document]
delucis: Micro-nit to also stamp on a double space 😅 

```suggestion
Runs Astro's development server. This is a local HTTP server that doesn't bundle assets. It uses Hot Module Replacement (HMR) to update your browser as you save changes in your editor.
```

delucis:
Going to merge on the advice of @Jutanium. We still need to figure out what to do with the `--verbose` CLI flag, but the rest is an improvement so we can tackle that separately.

PR #1624, opened 2022-09-26
Add docs for how to type Props of astro components that don't want props

jasikpark:

- New or updated content

#### Description

This is good for when you don't want to accept props for a component and want `astro check` to deny props to a component, so inputs don't go unused.

It can be misleading otherwise when you might expect props for an astro component to get passed through to the top-level wrapper element in that component

### DIFF
--- a/src/pages/en/guides/typescript.md
+++ b/src/pages/en/guides/typescript.md
@@ -103,6 +103,12 @@ const { greeting = 'Hello', name } = Astro.props;
 <h2>{greeting}, {name}!</h2>
 ```
 
+:::tip
+
+To type a component that denies props, you can pass `type Props = Record<string, never>;` for zero props and `type Props = { children: any; };` for no props except for the default slot.
+
+:::
+
 ### Built-in attribute types
 
 Astro provides JSX type definitions to check that your markup is using valid HTML attributes. You can use these types to help build component props. For example, if you were building a `<Link>` component, you could do the following to mirror the default HTML attributes in your component’s prop types.


src/pages/en/guides/typescript.md
```diff
@@ -103,6 +103,12 @@ const { greeting = 'Hello', name } = Astro.props;
 <h2>{greeting}, {name}!</h2>
 ```
 
+:::tip
+
+To type a component that denies props, you can pass `type Props = Record<string, never>;` for zero props and `type Props = { children: any; };` for no props except for the default slot.
```
Princesseuh, line 108: It's a bit annoying that using `type Props = Record<string, never>;` means you have to redefine `children`. This is not the case when doing `type Props = {};`

In my opinion, the ESLint rule is wrong here (and in general, for a JSX definition) 🤷‍♀️

jasikpark: I think it makes sense that you need to be specific here... in a function based JSX context like react function components, you just wouldn't accept a props argument to the function.

There's an argument for always requiring the user to specify the type of children rather than injecting it

react 18 removed implicit children: https://solverfox.dev/writing/no-implicit-children/

jasikpark: Where currently astro works like `React.FC<Props>`, while I'd like it to function like `Props`, where no props or children are accepted unless specified in the file, hopefully with a nice astro type to import for the type of the children/slots

jasikpark: though i see how this is an incremental thing, as we move people onto the idea 👍

Princesseuh: I agree with you that if we want to be very strict, it's much better to not have any implicit props. It's annoying in Astro context where we both want to support being very strict, but also not strict at all

jasikpark: It'd be great if that was configurable somehow maybe?? i dunno 😔

src/pages/en/guides/typescript.md
```diff
@@ -103,6 +103,12 @@ const { greeting = 'Hello', name } = Astro.props;
...
+
+:::
```
delucis, line 110: I wonder if this is important enough to not be a `tip`. Maybe we could collect these different scenarios like this:

```suggestion
### Common prop type patterns

- If your component takes no props or slotted content, you can use `type Props = Record<string, never>`.

- If your component must be passed children to its default slot, you can enforce this by using `type Props = { children: any; };`.
```

jasikpark: that's probably better

delucis (approved):
Wait… did I stomp entirely on your changes, @jasikpark?

Sorry! I guess this… LGTM? 😂

PR #1630, opened 2022-09-27
Docs: Nanostores/Clarification on supported frameworks

whotookzakum:
It can be a bit confusing as to why React and Solid were not mentioned--is sharing state between these two not as seamless as it is between the others? Hope that this slight change in wording can help.
Alternatives could be ".. sharing state between the supported frameworks ..", " .. Astro's supported frameworks .. "

#### Description

- Small wording change in the Nanostores docs


delucis (approved):
Thanks @whotookzakum 👍

PR #1632, opened 2022-09-27
Edit: remove `export` requirement for `Props`

bholmesdev:

- [X] Minor content fixes (broken links, typos, etc.)

#### Description

- Removes `export` requirement for the `Props` interface. You no longer need it for types to work!


delucis (approved):
Props for fixing this 🙌 

Aside: one thing I did notice in this repo is that ESLint complains about `Props` being unused now that it’s neither exported nor used via `as Props`. Haven’t dug deeper but probably something to tweak in that config.

Issue #1618, opened 2022-09-25
Document `all`, `del`, and other possible API route functions


Jutanium:
Right now, we only show  _get_ and _post_ in our [API routes docs](https://docs.astro.build/en/guides/server-side-rendering/#api-routes), but there are other possible functions.

This should most likely be part of the [endpoints guide](https://github.com/withastro/docs/issues/760) we're planning.

Issue #768, opened 2022-06-13
Document `getStaticPaths` in non-HTML pages


delucis:
Non-HTML pages support exporting a `getStaticPaths` function just like `.astro` dynamic routes do, but we don’t currently document it. This should at least be mentioned in the [non-HTML section](https://docs.astro.build/en/core-concepts/astro-pages/#non-html-pages) of the **Pages** guide and probably deserves a comment in the [`getStaticPaths` reference](https://docs.astro.build/en/reference/api-reference/#getstaticpaths) [and **Dynamic routes** docs](https://docs.astro.build/en/core-concepts/routing/#dynamic-routes).

Particular gotcha to mention:

If the file route is `src/pages/[api].json.ts`, the `api` param returned by `getStaticPaths` need to _include_ `.json`. Not sure if that’s intended or a bug — we should check with core. Here’s an example:

```js
import type { APIRoute } from 'astro';

export const get: APIRoute = async ({ params }}) => {
  return {
    body: JSON.stringify({
      path: params.api,
    }),
  };
};

export function getStaticPaths() {
  return [
    { params: { api: 'a.json' } }, // ✅ works => outputs '/a.json'
    { params: { api: 'b' } },      // ❌ breaks => outputs '/b'
  ];
}
```

(I guess under the hood this is essentially working as if the `.json` part of the filename is not there.)

delucis:
Closing in favour of tackling everything together in #760

PR #1638, opened 2022-09-28
Endpoints Guide

### LINKED ISSUE #760: New: Add "endpoints" guide
Our guides for creating non-HTML pages (aka "endpoints") are pretty tough to scan as a new user. Right now, we have:
- ["non-HTML pages" listed on our **pages** docs](https://docs.astro.build/en/core-concepts/astro-pages/#non-html-pages), which I can't find by searching for "endpoint" 😕 
- [the "response" section at the bottom of our **SSR** docs](https://docs.astro.build/en/guides/server-side-rendering/#response ); again, not find-able searching for "endpoint" 

☝️ We also have different advice for endpoint return values for SSG vs. SSR. In SSR, we recommend a formatted `Response` object with a defined content type. In SSG, you must use a `{ body }` object _only_ (`Response` will break during builds) to output a static file. This speaks to a more fundamental issue with SSR, where it's more tacked-on than interwoven with our existing docs.

**Solution:** a general "endpoints" docs page would be an excellent landing spot to compare SSG and SSR approaches to non-HTML files. We are beginning to encourage `.js` endpoints for any clientside data fetching as well, so a dedicated page for the concept should aid in onboarding SSR users.

Jutanium:

- New or updated content
#### Description

- Closes #760 and #1618 
- Moves the File Routes content from Astro Pages and the API Routes content from SSR into one page
- Renames "file routes" to "static file endpoints" and "api routes" to "server endpoints"
- Documents that you can match any HTTP method, that you must use `del` to match DELETE, and that you can use `all` to as a catch-all
- Fleshes out the buffer PNG example so that it works if you copy-paste it
- Show how you can use `params` and `getStaticPaths` for static file endpoints (#768), explain that `props` don't do anything here, and that you can access the dynamic parameter without `getStaticPaths` in SSR mode
- Labels the CAPTCHA example as an example, and change the order of the files so that the endpoint is shown before the code that calls it


src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
+---
+layout: ~/layouts/MainLayout.astro
+title: Endpoints
```
matthewp, line 3: I'm totally fine with calling these Endpoints, but just for reference I've also seen them called **API Routes**. I'm assuming the docs team discussed it and decided on Endpoints, but just in case not, there's another name we've used.

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+export function getStaticPaths () {
+    return [ 
+        { params: { id: "0"} },
+        { params: { id: "1"} },
+        { params: { id: "2"} },
+    ]
+}
+```
+
+This will generate three JSON endpoints at build time: `/api/1.json`, `/api/2.json`, `/api/3.json`. Dynamic routing with endpoints works the same as it does with pages, but because the endpoint is a function and not a component, [props](/en/reference/api-reference/#data-passing-with-props) aren't supported.
```
matthewp, line 82:
```suggestion
This will generate three JSON files at build time: `/api/1.json`, `/api/2.json`, `/api/3.json`. Dynamic routing with endpoints works the same as it does with pages, but because the endpoint is a function and not a component, [props](/en/reference/api-reference/#data-passing-with-props) aren't supported.
```

matthewp: Also, for SSG mode we probably could support `props`, at least I don't see a reason why not. I don't think we excluded that on purpose.

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+
+### `request`
+All endpoints receive a `request` property, but only `request.url` works in static mode. This returns the full URL of the current endpoint, using your [`site`](/en/reference/configuration-reference/#site) config option as the base.
```
matthewp, line 85: Is this true in dev mode? I would expect that the url is the localhost url during dev.

I would not expect a difference here between endpoints and pages, if there is one it's probably a bug.

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+    })
+  };
+}
+```
+
+## Server Endpoints (API Routes)
+Everything in the previous section can be used in SSR mode, but the endpoints will be built when they are requested. This unlocks new features that are unavailable at build time, and allows you to build API routes that listen for requests and securely execute code on the server at runtime.
+
+:::note
+Be sure to [enable SSR](http://localhost:3001/en/guides/server-side-rendering/#enabling-ssr-in-your-project) before trying these examples.
```
matthewp, line 103: this is pointing to localhost

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+:::
+
+Endpoints can now access `params` without exporting `getStaticPaths`, and they can return a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object, allowing you to set status codes and headers:
```
matthewp, line 106:
```suggestion
Endpoints can access `params` without exporting `getStaticPaths`, and they can return a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object, allowing you to set status codes and headers:
```

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+  }
+}
+```
+
+### `request` 
+In SSR mode, the `request` property returns a fully usable [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) object that refers to the current request. This allows you to accept data and check headers:
+
+```ts title="src/pages/test-post.json.ts"
+export const post: APIRoute = async ({ request }) => {
+  if (request.headers.get("Content-Type") === "application/json") {
```
matthewp, line 181: good example

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+    }), {
+      status: 200
+    })
+  }
+  return new Response(null, { status: 400 });
+}
+```
+
+### Redirects
+Since `Astro.redirect` is not available in API Routes you can use [`Response.redirect`](https://developer.mozilla.org/en-US/docs/Web/API/Response/redirect) to redirect:
```
matthewp, line 195: This is a bug and something we'll probably be changing, should we document it anyways since it's useful info for now and just remove later?

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+// Return an object with `body` to save the file contents in your final build.
+export async function get({params, request}) {
+  return {
+    body: JSON.stringify({
+      name: 'Astro',
+      url: 'https://astro.build/',
+    }),
+  };
+}
+```
```
sarah11918, line 30:
``````suggestion
File routes export a `get` function (optionally `async`) which gets called to generate the file. It returns an object with a `body` to save the file contents in your final build. In static mode, Astro will call this at build time to generate the corresponding file.

```js
// Example: src/pages/builtwith.json.ts
// Outputs: /builtwith.json

export async function get({params, request}) {
  return {
    body: JSON.stringify({
      name: 'Astro',
      url: 'https://astro.build/',
    }),
  };
}
```
There's some duplication there to be edited, but I just moved the code comment out into the paragraph itself. I feel like both might have been redundant? Can that literally come out and become part of the lead-in paragraph?

Or was your thinking, "This is for the people who aren't going to read the paragraph, so I'd better put the exact same thing in the code sample?" (I get that! Just does seem to say the same thing twice?)
``````

File: src/pages/en/core-concepts/endpoints.md [hunk 9]
```diff
@@ -0,0 +1,275 @@
...
+
+The return object can also have an `encoding` property. It can be any valid [`BufferEncoding`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bdd02508ddb5eebcf701fdb8ffd6e84eabf47885/types/node/buffer.d.ts#L169) accepted by node.js' `fs.writeFile` method. For example, to produce a binary png image:
```
sarah11918, line 32:
```suggestion
The return object can also have an `encoding` property. It can be any valid [`BufferEncoding`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bdd02508ddb5eebcf701fdb8ffd6e84eabf47885/types/node/buffer.d.ts#L169) accepted by Node.js' `fs.writeFile` method. For example, to produce a binary png image:
```

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+export async function get({ params, request }) => {
+  const response = await fetch("https://astro.build/assets/press/full-logo-light.png");
+  const buffer = Buffer.from(await response.arrayBuffer());
+  return {
+    body: buffer,
+    encoding: 'binary',
+  };
+```
+
+You can type your endpoint functions using the `APIRoute` type:
```
sarah11918, line 44: You can "also"? "instead"?

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+
+```js
+import type { APIRoute } from 'astro';
+
+export const get: APIRoute = async function get ({params, request}) {
+...
+```
+
+As their only argument, endpoint functions recieve an object with two properties: `params` and `request`.
```
sarah11918, line 53: I feel like maybe this should go higher, even before all the examples? This isn't just about the `APIRoute` type, right? This is for all of these?

Like up in line 14 maybe:

> File routes export a `get` function (optionally `async`) that receives an object with two properties (`params` and `request`) as its only argument...

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+
+### `params` and Dynamic routing
+
+Endpoints support the same [dynamic routing](/en/core-concepts/routing/#dynamic-routes) features that pages do. Name your file with a bracketed parameter name and export a [`getStaticPaths()` function](/en/reference/api-reference/#getstaticpaths). Then, you can access the parameter using the `params` property passed to the endpoint function:
+
+```ts title="src/pages/[id].json.ts"
+import type { APIRoute } from 'astro';
+
+const usernames = ["Sarah", "Chris", "Dan"]
```
sarah11918, line 62: 😍

src/pages/en/core-concepts/endpoints.md
sarah11918, line 85:
```suggestion
All endpoints receive a `request` property, but in static mode, only `request.url` is available. This returns the full URL of the current endpoint, using your [`site`](/en/reference/configuration-reference/#site) config option as the base.
```
I worry about somone scanning and reading "`request.url` only works in static mode" by mistake. So, I've swapped the phrase around to hit BUT IN STATIC MODE (even though I don't *love* the two commas) and I think we can ebe more precise than "works" ... what's correct here, "available"? "have access to"?

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+  return {
+    body: JSON.stringify({
+      path: new URL(request.url).pathname
+    })
+  };
+}
+```
+
+## Server Endpoints (API Routes)
+Everything in the previous section can be used in SSR mode, but the endpoints will be built when they are requested. This unlocks new features that are unavailable at build time, and allows you to build API routes that listen for requests and securely execute code on the server at runtime.
```
sarah11918, line 100:
```suggestion
Everything described in the previous static files section can also be used in SSR mode: files export a `get` function and return a `body`, dynamic routing is available, and all endpoints receive a request property. But, unlike in `static` mode, when you configure `server` mode, the endpoints will be built when they are requested. This unlocks new features that are unavailable at build time, and allows you to build API routes that listen for requests and securely execute code on the server at runtime.
```
Love the second sentence here especially!

The concern I have is that some people are going to jump to this section immediately, not having read the from the top of the page ("I'm doing SSR!"). So, I think this section needs a bit more of an explicit lead in and comparison with what came before. Not totally married to the suggestion above, but something like that I think would make that hit home?

src/pages/en/core-concepts/endpoints.md
sarah11918, line 195:
```suggestion
Since `Astro.redirect` is currently not yet available in API Routes you can use [`Response.redirect`](https://developer.mozilla.org/en-US/docs/Web/API/Response/redirect) to redirect:
```
Per Matthew's comment, we could imply that this may yet change, if you wanted.

sarah11918 (commented):
Love this page @Jutanium! Made some suggestions for your consideration!  ❤️

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
...
+      name: 'Astro',
+      url: 'https://astro.build/',
+    }),
+  };
+}
+```
+
+The return object can also have an `encoding` property. It can be any valid [`BufferEncoding`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bdd02508ddb5eebcf701fdb8ffd6e84eabf47885/types/node/buffer.d.ts#L169) accepted by node.js' `fs.writeFile` method. For example, to produce a binary png image:
+
+```ts title="src/pages/astro-logo.png.ts" {7}
```
kylebutts, line 34: Small highlighting issue, should be {6}
CleanShot 2022-09-28 at 18 52 14@2x

src/pages/en/core-concepts/endpoints.md
Jutanium: Yeah, let's document it for now. Is it just `Astro.redirect` that should be available, or other functions on the Astro global?

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
+---
+layout: ~/layouts/MainLayout.astro
+title: Endpoints
+description: Learn how to create endpoints that serve any kind of data
+i18nReady: true
+---
+Astro lets you create custom endpoints to serve any kind of data. You can use this to generate images, expose an RSS document, or build a full API for your site.
+
+If your project is in [static](/en/reference/configuration-reference/#output) mode, custom endpoints are called at build time to produce static files. In [SSR](/en/guides/server-side-rendering/) mode, custom endpoints turn into live server endpoints that are called on request. Static and SSR endpoints are defined similarly, but SSR endpoints support additional features.
```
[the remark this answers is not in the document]
Jutanium: I like `statically-generated sites`, and I rephrased the second sentence to make it clear that static is the default: 
> `In statically-generated sites, your custom endpoints are called at build time to produce static files. If you opt in to [SSR](/en/guides/server-side-rendering/) mode...`

src/pages/en/core-concepts/endpoints.md
```diff
@@ -0,0 +1,275 @@
+---
+layout: ~/layouts/MainLayout.astro
+title: Endpoints
+description: Learn how to create endpoints that serve any kind of data
+i18nReady: true
+---
+Astro lets you create custom endpoints to serve any kind of data. You can use this to generate images, expose an RSS document, or build a full API for your site.
```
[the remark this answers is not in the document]
Jutanium: Do you think we should do it here, making "API Routes" a synonym for all custom endpoints, and remove the `(API Routes)` next to `Server Endpoints`? I was thinking that it was specifically the server version that was previously called API Routes, and the static version we called non-HTML pages  (and then file routes).

Or maybe something like `Astro lets you create custom endpoints to serve any kind of data. You can use this to generate images, expose an RSS document, or use them as API Routes to build a full API for your site.` What do you think?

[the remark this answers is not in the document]
sarah11918: I like that!

src/pages/en/core-concepts/endpoints.md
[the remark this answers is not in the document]
sarah11918: Yes, I like that!

src/pages/en/core-concepts/endpoints.md
Jutanium: I think we can just document it as-is, and change it when needed

Jutanium:
Thanks for the reviews, I've made changes to address the comments!

sarah11918:
LGTM, Dan! (I'll update branch to make my co-author cred easy. 😉   )

PR #1646, opened 2022-09-29
Update CLI Reference with `--site` & `--base` flags

natemoo-re:

- New or updated content

#### Description

- Adds missing `--site` flag to CLI reference, which has been supported for a long time.
- Adds new `--base` flag to CLI reference, which is implemented in https://github.com/withastro/astro/pull/4917


sarah11918:
LGTM, @natemoo-re! Just to make sure docs are coordinated with the feature release properly, because `--site` already exists, but `--base` is new, and not yet released...

Either this PR needs to wait on the upstream PR https://github.com/withastro/astro/pull/4917 to merge, and we can apply the `merge-on-release` label to remind us to wait. OR

We can use the `<Since /> component to specify the version of Astro that `--base` works with (even if it's a future version) and it's less of a concern if Docs are merged before the feature is released. This would look like:

```
---
setup: |
 ---
### `--base`

<Since v="1.4.0" />
```
(Might we want to do this anyway, in case people on older versions try to use `--base` ?)

src/pages/en/reference/cli-reference.md
```diff
@@ -167,6 +168,16 @@ Specifies the path to the config file relative to the project root. Defaults to
 astro --config config/astro.config.mjs dev
 ```
 
+### `--site`
```
delucis, line 171: Do these need to indicate the arguments they get passed like we do for `--root` and `--config`?

```suggestion
### `--site <url>`
```

src/pages/en/reference/cli-reference.md
```diff
@@ -167,6 +168,16 @@ Specifies the path to the config file relative to the project root. Defaults to
...
+
+Configures the [`site`](/en/reference/configuration-reference/#site) for your project. Passing this flag will override the `site` value in your `astro.config.mjs` file, if one exists.
+
+### `--base`
```
delucis, line 175:
```suggestion
### `--base <path>`
```

delucis (commented):
Looks good Nate! One comment with how we show these headings. Not entirely happy with my proposed argument names, happy for you to change those if you think something else would be better.

delucis:
(added the flags to the PR title so we can test #1653)

PR #1648, opened 2022-09-30
Documenting Cascading Order: Introduction and Examples

### LINKED ISSUE #1642: CSS specificity example
Make sure the order of CSS styles applied is clearly documented, as per the description here:  https://github.com/withastro/astro/issues/3357#issuecomment-1262247873

I propose adding this right before CSS Integrations as an `<h2>` ~~maybe `## CSS Specificity` or `## Style Specificity`~~ (See attached screenshot for placement) since it builds off of both the Astro `<style>` tag and importing external spreadsheets described in the two sections above.

Further to Matthew's comment, I propose using `## Cascading Order` to incorporate both specificity and order of appearance when we give an example of "what should happen." 

This is now how it's intended to work, and should be documented as true. If people find it doesn't work that way, it's a bug and should be reported as an Issue in `withastro/astro`, not changed in documentation.
Screenshot 2022-09-29

Jutanium:

- New or updated content

#### Description

- Closes #1642
- Introduces the two relevant cascading order factors: specificity and order of appearance
- Describes the precedence of scoped styles, imported styles, and link tag styles relative to each other
- Shows examples for each feature
- Explains that the later imported style has higher precedence
- Explains [as per this comment](https://github.com/withastro/docs/issues/1642#issuecomment-1262917210) that when a component imports styles, they're imported at the same time as the component, and shows an example of this being relevant

### DIFF
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -315,7 +315,7 @@ import PurpleComponent from "./PurpleComponent.astro";
 ```
 
 ### Link Tags
-Style sheets loaded via [link tags](#load-a-static-stylesheet-via-link-tags), before any other styles in an Astro component. Therefore, these styles will have lower precedence than imported stylesheets and scoped styles:
+Style sheets loaded via [link tags](#load-a-static-stylesheet-via-link-tags) are evaluated in order, before any other styles in an Astro file. Therefore, these styles will have lower precedence than imported stylesheets and scoped styles:
 
 ```astro title="index.astro"
 ---
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -290,7 +290,8 @@ import "./make-it-purple.css"
 </div>
 ```
 
-Importing a component applies any CSS it imports, even if the component is never used:
+While `<style>` tags are scoped and only apply to the component that declares them, _imported_ CSS can "leak". Importing a component applies any CSS it imports, even if the component is never used:
+
 ```astro title="PurpleComponent.astro"
 ---
 import "./make-it-purple.css"
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -197,7 +197,7 @@ When evaluating multiple CSS rules that apply to the same element, the browser u
 If two rules have the same specificity, the _order of appearance_ is evaluated, and the last rule will take precedence:
 ```astro title="MyComponent.astro"
 <style>
-  h1 { color: purple}
+  h1 { color: purple }
   h1 { color: red }
 </style>
 <div>
@@ -207,7 +207,7 @@ If two rules have the same specificity, the _order of appearance_ is evaluated, 
 </div>
 ```
 
-Among Astro features, if the specificity of a rule is the same, **scoped styles** take precedence, then **imported styles**, then **`<link>` tags in the head**.
+Among Astro features, if the specificity of a rule is the same, **scoped styles** take precedence over **imported styles**. Both of these have higher precedence than **`<link>` tags in the head**.
 
 ### Scoped Styles 
 
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -182,7 +182,7 @@ Astro components will sometimes have to evaluate multiple sources of CSS. For ex
 
 When conflicting CSS rules apply to the same element, browsers first use _specificity_ and then _order of appearance_ to determine which value to show.
 
-If one rule is more _specific_ than another, no matter where the CSS rule appears, it will take precedence:
+If one rule is more _specific_ than another, no matter where the CSS rule appears, its value will take precedence:
 
 ```astro title="MyComponent.astro"
 <style>
@@ -198,7 +198,7 @@ If one rule is more _specific_ than another, no matter where the CSS rule appear
 </div>
 ```
 
-If two rules have the same specificity, then the _order of appearance_ is evaluated, and the last rule will take precedence:
+If two rules have the same specificity, then the _order of appearance_ is evaluated, and the last rule's value will take precedence:
 ```astro title="MyComponent.astro"
 <style>
   h1 { color: purple }
@@ -219,7 +219,7 @@ Astro CSS rules are evaluated in this order of appearance:
 
 ### Scoped Styles 
 
-Using [scoped styles](#scoped-styles) does not increase the _specificity_ of your CSS, but they will always come last in the _order of appearance_. They will therefore take precedence over other styles of the same specificity. For example, if you import a stylesheet with the same rule as a scoped style, the scoped style will apply:
+Using [scoped styles](#scoped-styles) does not increase the _specificity_ of your CSS, but they will always come last in the _order of appearance_. They will therefore take precedence over other styles of the same specificity. For example, if you import a stylesheet that conflicts with a scoped style, the scoped style's value will apply:
 
 ```css title="make-it-purple.css"
 h1 {
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -287,18 +287,18 @@ import "./make-it-purple.css"
 ```
 
 If a component imports CSS, that CSS will be imported at the same time, even if the component is never used:
-```astro title="PurpleParent.astro"
+```astro title="PurpleComponent.astro"
 ---
 import "./make-it-purple.css"
 ---
 <div>
-  <slot/>
+  <p>I import purple CSS.</p>
 </div>
 ```
 ```astro title="MyComponent.astro"
 ---
 import "./make-it-green.css"
-import PurpleParent from "./PurpleParent.astro";
+import PurpleComponent from "./PurpleComponent.astro";
 ---
 <style>
   h1 { color: red }
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -315,7 +315,7 @@ import PurpleComponent from "./PurpleComponent.astro";
 ```
 
 ### Link Tags
-Style sheets loaded via [link tags](#load-a-static-stylesheet-via-link-tags) in the head are evaluated before any other styles in an Astro component, in the order in which they appear. Therefore, these styles will always have lower precedence than imported stylesheets and scoped styles:
+Style sheets loaded via [link tags](#load-a-static-stylesheet-via-link-tags), before any other styles in an Astro component. Therefore, these styles will have lower precedence than imported stylesheets and scoped styles:
 
 ```astro title="index.astro"
 ---
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -263,7 +263,7 @@ import "./make-it-purple.css"
 
 ### Import Order
 
-When importing multiple stylesheets, if any styles have the same specificity, the _last one imported_ wins:
+When importing multiple stylesheets in an Astro component, the CSS rules are evaluated in the order that they are imported. A higher specificity will always determine which styles to show, no matter when the CSS is evaluated. But, when conflicting styles have the same specificity, the _last one imported_ wins:
 
 ```css title="make-it-purple.css"
 div > h1 {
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/styling.md
```diff
@@ -176,6 +176,166 @@ You can also use the `<link>` element to load a stylesheet on the page. This sho
...
+<div>
+  <h1 class="red">
+    This header will be red!
+  </h1>
+</div>
+```
+
+Among Astro features, if the specificity of a rule is the same, the resulting styles will appear in this order:
+
+- **`<link>` tags in the head** (lowest precedence)
```
sarah11918, line 212: This is not Astro-specific, so I think it might make sense to provide an MDN link to Cascading Order and make it a bit more clear that you're "teaching CSS" here, not Astro. And, I *might* even remove these initial examples altogether.

OR, I'd keep the examples, but then make the text more super-Astro specific:

``````suggestion
Astro components will sometimes have to evaluate multiple CSS "instructions" for the same element. For example, your component might import a CSS stylesheet, include its own `<style>` tag, *and* be rendered inside a layout.

When multiple CSS rules apply to the same element, browsers first use _specificity_ and then _order of appearance_ to determine which style to show.

If one rule is more _specific_ than another, no matter when the CSS rule is evaluated, it will take precedence:

```astro title="MyComponent.astro"
<style>
  h1 { color: red }
  div > h1 {
    color: purple
  }
</style>
<div>
  <h1>
    This header will be purple!
  </h1>
</div>
```

If two rules have the same specificity, then the _order of appearance_ is evaluated, and the last rule will take precedence:
```astro title="MyComponent.astro"
<style>
  h1 { color: purple }
  h1 { color: red }
</style>
<div>
  <h1>
    This header will be red!
  </h1>
</div>
```

Astro CSS rules are evaluated in this order of appearance:

- **`<link>` tags in the head** (lowest precedence)
``````
Also, I'm not sure about using `class="red"` because we never define what the red class is? Are we supposed to infer that the class has a styling rule to make *something* red? All text? just h1? The `<style>` tag in the component isn't styling the class, it's styling the element. 

It seems these examples work without the class applied to them, and if we're not actually showing where the class comes from, or that it's even being applied (we don't import a global stylesheet, nor do we show that this component is in a layout that does) then I don't think it's helping explain why the style tag is making the h1 red.

File: src/pages/en/guides/styling.md [hunk 2]
```diff
@@ -176,6 +176,166 @@ You can also use the `<link>` element to load a stylesheet on the page. This sho
...
+<div>
+  <h1>
+    This header will be purple!
+  </h1>
+</div>
+```
+
+### Import Order
+
+When importing multiple stylesheets, if any styles have the same specificity, the _last one imported_ wins:
```
sarah11918, line 262:
```suggestion
When importing multiple stylesheets in an Astro component, the CSS rules are evaluated in the order that they are imported. A higher specificity will always determine which styles to show, no matter when the CSS is evaluated. But, when conflicting styles have the same specificity, the _last one imported_ wins:
```
Something like this. I wanted to (a) give this a bit more (even if repeated) context in case anyone jumps to this section right away. And (b) explain what's happening in terms of the order of evaluation (since that's the key factor here). If they _have_ been reading diligently top to bottom, then they KNOW order of evaluation matters when specificity is the same. What they need to know in this section is "In which order does ASTRO evaluate multiple stylesheets?" and I prefer to say that explicitly, matching the context we've set above.

src/pages/en/guides/styling.md
```diff
@@ -176,6 +176,166 @@ You can also use the `<link>` element to load a stylesheet on the page. This sho
...
+  <h1>
+    This header will be purple!
+  </h1>
+</div>
+```
+
+### Link Tags
+During compilation, scoped styles and imported stylesheets are added to the _bottom of the head_ as `<link>` tags, after any other link tags.
+
+This means that if you load a style sheet via a [link tag](#load-a-static-stylesheet-via-link-tags) in the head, it will always have lower precedence:
```
sarah11918, line 316: And, I do find this a bit confusing too. If this is my `.astro` component as written, there is nothing being added to the head _of my original source component_ that I didn't explicitly write? That's what happens to create the HTML that is sent to the browser, but that doesn't change what's in my actual component file? So, I don't really like that we immediately start talking about turning things into HTML link elements when we really need to tell people when these links as written in their project will be evaluated in comparison to other styling options in their project.

And again, I'd think about rephrasing this in terms of the two factors, specificity and order of appearance, that you've established at the beginning. It's like if we give those rules to follow (Specificity first, then order of appearance), then every explanation only needs speak in those terms. Assuming they know what specificity means, then the real job of this section is to explain Astro's ORDER of evaluating everything. That's what they mostly need to know.

I've rewritten something, which I'm not sure is what you intended, but is what the example seems to want to be explaining.

```suggestion
Style sheets loaded via [link tags](#load-a-static-stylesheet-via-link-tags) in the head are evaluated before any other styles in an Astro component, in the order in which they appear. Therefore, these styles will always have lower precedence than imported styleshees and scoped styles:
```

The other stuff feels more like a "fun fact: ALL styles eventually become stylesheet links in the browser HTML!" if you want to try to include it. But, I don't like it as the description of the *reason* for the styling of this example.

> When Astro bundles and processes your CSS, scoped styles and imported stylesheets are added to the _bottom of the head_ as `<link>` tags, after any other link tags. Your browser then evaluates these link tags in order of their appearance.

src/pages/en/guides/styling.md
```diff
@@ -176,6 +176,166 @@ You can also use the `<link>` element to load a stylesheet on the page. This sho
...
+
+- **`<link>` tags in the head** (lowest precedence)
+- **imported styles**
+- **scoped styles** (highest precedence)
+
+### Scoped Styles 
+
+Using [scoped styles](#scoped-styles) does not increase the _specificity_ of your CSS, but it will always come last in the _order of appearance_. They will therefore take precedence over other styles of the same specificity. For example, if you import a stylesheet with the same rule as a scoped style, the scoped style will apply:
+
+```css title="make-it-purple.css"
```
sarah11918, line 220: I love these scoped style examples!! `make-it-so`  😂

sarah11918 (commented):
Hey @Jutanium! This is SUPER helpful, thanks for taking this on!

I made some suggestions for your consideration. See what you think!

src/pages/en/guides/styling.md
```diff
@@ -176,6 +176,166 @@ You can also use the `<link>` element to load a stylesheet on the page. This sho
...
+---
+<style>
+  h1 { color: red }
+</style>
+<div>
+  <h1>
+    This header will be purple!
+  </h1>
+</div>
+```
```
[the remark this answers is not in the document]
Jutanium: Totally agree, thanks! I originally wanted to use the component as a parent, but then I realized I needed to show that it even brings its CSS when _not_ a parent.

src/pages/en/guides/styling.md
Jutanium: Oh no, I accidentally left this in when I copied it from an experimentation - will remove! 🤦

src/pages/en/guides/styling.md
Jutanium: I like this, thanks!

src/pages/en/guides/styling.md
Jutanium: Gotcha, I can see how referring to the head of the site introduces more confusion than it explains. I took your suggestion and just swapped the clauses to make it slightly more clear, and removed my unnecessary "always":

> Style sheets loaded via [link tags](#load-a-static-stylesheet-via-link-tags) in the head are evaluated in order, before any other styles in an Astro component. Therefore, these styles will have lower precedence than imported stylesheets and scoped styles:

src/pages/en/guides/styling.md
```diff
@@ -176,6 +176,168 @@ You can also use the `<link>` element to load a stylesheet on the page. This sho
 
 Because this approach uses the `public/` directory, it skips the normal CSS processing, bundling and optimizations that are provided by Astro. If you need these transformations, use the [Import a Stylesheet](#import-a-local-stylesheet) method above.
 
+## Cascading Order
+
+Astro components will sometimes have to evaluate multiple CSS "instructions" for the same element. For example, your component might import a CSS stylesheet, include its own `<style>` tag, *and* be rendered inside a layout.
```
sarah11918, line 181: Ooh, I couldn't think of what word I wanted to use instead of "rules" in case that was going to be too repetitive, so I used "instructions" but put in scare quotes to make it a placeholder. Is there a word you like better? Or, if instructions is fine, maybe it should be without the quotation marks?

Jutanium: Or just leave it out:

> `.astro` files will sometimes have to evaluate multiple styles that apply to the same element. For example, your component might import a CSS stylesheet, include its own `<style>` tag, *and* be rendered inside a layout.

sarah11918: Yup, that looks good. Ship it!

src/pages/en/guides/styling.md
```diff
@@ -176,6 +176,168 @@ You can also use the `<link>` element to load a stylesheet on the page. This sho
...
+  h1 { color: red }
+</style>
+<div>
+  <h1>
+    This header will be purple!
+  </h1>
+</div>
+```
+
+If a component imports CSS, that CSS will be imported at the same time, even if the component is never used:
```
sarah11918, line 293:
```suggestion
Importing a component also imports its CSS style rules along with it, even if the component is never used:
```
Realized this didn't actually mention the part about a component importing another component. Could have said, "If an **imported** component imports CSS..." but that starts to get mucky. Tried to rearrange to make it as succinct as possible, to cut down on repeated "component" and "import" as much as possible, but you might have something better.

src/pages/en/guides/styling.md
```diff
@@ -176,6 +176,168 @@ You can also use the `<link>` element to load a stylesheet on the page. This sho
...
+</style>
+<div>
+  <h1>
+    This header will be purple!
+  </h1>
+</div>
+```
+
+### Link Tags
+Style sheets loaded via [link tags](#load-a-static-stylesheet-via-link-tags), before any other styles in an Astro component. Therefore, these styles will have lower precedence than imported stylesheets and scoped styles:
```
sarah11918, line 318:
```suggestion
Style sheets loaded via [link tags](#load-a-static-stylesheet-via-link-tags) are evaluated before any other styles in an Astro component. Therefore, these styles will have lower precedence than imported stylesheets and scoped styles:
```

src/pages/en/guides/styling.md
kevinzunigacuellar: This could get tricky to understand if previously we said that CSS doesn't leak from other components. I was thinking maybe we can include some good practices to go around it. what do you think Sarah and Dan?

Jutanium: We said that _scoped_ styles don't leak - maybe the best practice here is to use Scoped Styles instead?

Jutanium: Oh, and to be sure to import your layout first! Adding

sarah11918: Yes, this "not leaking" tripped me up at first... I almost wrote, WAIT IS THIS TRUE??
 
So to be clear, it's only the IMPORTED styles that carry over with the component, not its `<style>` tag styles?

So, we *could* try to rephrase like "imported styles are transitive" kind of thing... if a style finds its way into your component via importing (whether imported directly via a statment, or imported indirectly via a *component* import), then it's gonna end up in the list of styles that will be evaluated. And, its precedence will depend on the order of the import statement that brought it in: whether you were importing a stylesheet directly, or a component that imported its own stylesheet.

Jutanium: I think "leaking" might be the best way to describe this... tell me what you think of my rephrasing and tip in the latest commit

src/pages/en/guides/styling.md
Jutanium: I think wherever we use "Astro components", there's a potential for confusion that it doesn't apply to pages, and this is equally true in an intro

Jutanium: Following our convo, I'm going to leave it as "Astro components"!

src/pages/en/guides/styling.md
sarah11918: I like it!

delucis (commented):
I love this Dan! Hats off for having the determination to chase down all the nuances here and lay them out so clearly 🙌

PR #1653, opened 2022-09-30
Changes breaking text in congrats-bot.mjs script

### LINKED ISSUE #1615: Escape `--flags` in Discord CongratsBot
We generate a Discord post congratulating commit authors using [a workflow](https://github.com/withastro/docs/blob/main/.github/workflows/discord-congrats.yml) and [custom script](https://github.com/withastro/docs/blob/main/scripts/congrats-bot.mjs).

A couple of times now, commits containing something like a shell flag, e.g. `--template` or `--experimental-integrations`, has broken the formatting, with the flag and everything following it omitted from the Discord output. Here’s the result of [a recent example run](https://github.com/withastro/docs/actions/runs/3111970690/jobs/5044872096):

Screenshot of Congrats Bot Discord message reading “🚀 Merged! Mayank: [`Document create-astro”

We probably need to escape flags starting with two hyphens somehow.

Lofty-Brambles:

- Something else!
- Replaces the message breaking `--` with `––` instead, as to not break discord markdown.

#### Description

- Closes #1615  

### DIFF
--- a/scripts/congrats-bot.mjs
+++ b/scripts/congrats-bot.mjs
@@ -9,7 +9,7 @@ setDiscordMessage(COMMIT_AUTHOR, COMMIT_ID, COMMIT_MESSAGE);
  * @param {string} commitMsg A full commit message
  */
 function setDiscordMessage(author, id, commitMsg) {
-	const commitMessage = commitMsg.split('\n').shift().replaceAll('`', '').replaceAll('--', '––');
+	const commitMessage = commitMsg.split('\n').shift().replaceAll('`', '').replaceAll('--', '-​-');
 
 	const coAuthors = commitMsg
 		.split('\n')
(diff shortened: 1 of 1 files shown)


scripts/congrats-bot.mjs
```diff
@@ -9,7 +9,7 @@ setDiscordMessage(COMMIT_AUTHOR, COMMIT_ID, COMMIT_MESSAGE);
  * @param {string} commitMsg A full commit message
  */
 function setDiscordMessage(author, id, commitMsg) {
-	const commitMessage = commitMsg.split('\n').shift().replaceAll('`', '');
+	const commitMessage = commitMsg.split('\n').shift().replaceAll('`', '').replaceAll('--', '––');
```
delucis, line 12: Interesting. Didn’t think of this approach!

I think we can even go one better:

```suggestion
	const commitMessage = commitMsg.split('\n').shift().replaceAll('`', '').replaceAll('--', '-​-');
```

Very hard to see, but I’ve added a zero-width space between the two hyphens. What do you think?

Lofty-Brambles: Sounds like an even smoother approach! Added it!
The only worry is discord might not read it as a character, but from my test, it seems to.

delucis (approved):
Thanks! Lucky we have a PR open at the moment that we can test this with when it merges, so I'll get this merged now and we can revisit if there any problems.

Welcome to Team Docs @Lofty-Brambles!

PR #1677, opened 2022-10-02
Quickstart button: launch in new window

sarah11918:
In PRs no one asked for....

Experimenting with having the Quickstart button open in a new tab. (I never really liked that it took you away from Docs, especially since sometimes phrases like "Quick start" actually do take you to more docs, not *away* from docs, and launching an entirely different site.)

Therefore, also changed the wording of the button to be more in line with making it clear that the link was going to open our basic template. And, I think it's a more explicit comparison with "viewing all other templates." 

The only downside, being slightly longer, is that on mobile the button and the text are not always going to fit all on one line anymore. In tablet view, (except for the very smallest version of tablet view) it is still all on one line. But, I still like the more explicit description of what the button is going to do, and that it no longer takes you away from docs.  Open to thoughts on this!

### DIFF
--- a/src/components/Button.astro
+++ b/src/components/Button.astro
@@ -6,7 +6,7 @@ const { variant = 'primary' } = Astro.props;
 ---
 
 <span class:list={[`link pixel variant-${variant}`, className]} {style}>
-	<a {href}>
+	<a {href} target="_blank">
 		<span><slot /></span>
 	</a>
 </span>
--- a/src/pages/en/getting-started.md
+++ b/src/pages/en/getting-started.md
@@ -35,7 +35,7 @@ Check out our detailed [Why Astro](/en/concepts/why-astro/) breakdown to learn m
 Visit [astro.new](https://astro.new/) and choose from a variety of templates to get started. Play around with a full, working version of Astro right in your browser!
 
 <div style="display: flex; flex-wrap: wrap; gap: 0.5rem;">
-    <Button href="https://astro.new/basics?on=stackblitz">Quickstart!</Button>
+    <Button href="https://astro.new/basics?on=stackblitz">Launch basic template!</Button>
     <Button variant="outline" href="https://astro.new/">View all templates →</Button>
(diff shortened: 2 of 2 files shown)


Jutanium:
Any instance of the Button will now open in a new tab, and this change makes a lot of sense. The Button component is a fancy link component designed for external links, which should open in a new tab.

I like the more explicit copyA I think "Basic template" works just as well if we want a shorter version. Since the header already says "Try Astro in your browser" I don't think we _need_ the "launch".

sarah11918:
Thanks, @Jutanium! I'm leaning towards keeping "Launch" for now to make it clear that the template will open and run, not just take you to a GitHub repo.

src/components/Button.astro
```diff
@@ -6,7 +6,7 @@ const { variant = 'primary' } = Astro.props;
 ---
 
 <span class:list={[`link pixel variant-${variant}`, className]} {style}>
-	<a {href}>
+	<a {href} target="_blank">
```
delucis, line 9: In general I’d advise against opening in new windows. It enforces behaviour on users which they can’t easily opt out of and may not expect, whereas it is relatively simple to opt _in_ to opening in a new tab/window if you want it.

This article is helpful in listing some more reasons to avoid this 😅 
https://css-tricks.com/use-target_blank/

delucis: Right, I think the overall advice in that article is don't change default behaviour. A user can use cmd/ctrl + click to open in a new tab (or right click etc) if that's what they want but they probably won't (and can't) preemptively make sure to open in the current tab if _that's_ what they want.

Jutanium: Gotcha, this makes sense.

sarah11918, line 9:
```suggestion
	<a {href}>
```

sarah11918:
Merging with the explicit wording. Reversion PRs are freeee.....

PR #1693, opened 2022-10-03
Improve deploy/gitlab.md documentation

ralacerda:
- New or updated content

#### Description

As noted on #1382:
- The documentation had an unclear explanation to why we need to set `dist` to `public` and `public` to another value.
- The given `astro.config.mjs` example did not reflect the steps in the guide

This PR include my attempt to better explain that `public` is required to be the output folder for GitLab Pages, with links to the documentation to the `public` directory and an example of a `astro.config.mjs` after setting `public` to `static` and `build` to `public`. Some of the phrases are based on the example repo files, as is the decision to use `public` as `static` in the example.

It also includes useful comments that were on ".gitlab-ci.yml" from the example repo.

Finally, it fixes a small typo (removing the dot from `.site`)

I'm taking part on the hacktoberfest! :smile:

### DIFF
--- a/src/pages/en/guides/deploy/gitlab.md
+++ b/src/pages/en/guides/deploy/gitlab.md
@@ -42,7 +42,7 @@ For example, here are the correct `astro.config.mjs` settings when the `public/`
        paths:
          - node_modules/
      script:
-         # Specify the steps involved to build your app here
+       # Specify the steps involved to build your app here
        - npm install
        - npm run build
 
@@ -51,7 +51,7 @@ For example, here are the correct `astro.config.mjs` settings when the `public/`
          # The folder that contains the built files to be published.
          # This must be called "public".
          - public
-         - 
+
      only:
        # Trigger a new build and deploy only when there is a push to the
        # branch(es) below
--- a/src/pages/en/guides/deploy/gitlab.md
+++ b/src/pages/en/guides/deploy/gitlab.md
@@ -14,9 +14,9 @@ Check out [the official GitLab Pages Astro example project](https://gitlab.com/p
 ## How to deploy
 
 1. Set the correct `site` in `astro.config.mjs`.
-2. Set `dist` in `astro.config.mjs` to `public`. GitLab Pages requires exposed files to be located in a folder called "public". This instructs Astro to put the static build output in a folder called `public`.
+2. Set `outDir` in `astro.config.mjs` to `public`. GitLab Pages requires exposed files to be located in a folder called "public". This instructs Astro to put the static build output in a folder called `public`.
 
-If you were using the [`public/` directory](/en/reference/configuration-reference/#publicdir) as a source of static files in your Astro project, rename it and use that new folder name in `astro.config.mjs` for the value of `public`.
+If you were using the [`public/` directory](/en/reference/configuration-reference/#publicdir) as a source of static files in your Astro project, rename it and use that new folder name in `astro.config.mjs` for the value of `publicDir`.
 
 For example, here are the correct `astro.config.mjs` settings when the `public/` directory is renamed to `static/`:
 
--- a/src/pages/en/guides/deploy/gitlab.md
+++ b/src/pages/en/guides/deploy/gitlab.md
@@ -21,11 +21,13 @@ If you were using the [`public/` directory](/en/reference/configuration-referenc
 For example, here are the correct `astro.config.mjs` settings when the `public/` directory is renamed to `static/`:
 
    ```js
+   import { defineConfig } from 'astro/config';
+   
    export default defineConfig({
      sitemap: true,
      site: 'https://astro.build/',
-     dist: 'public',
-     public: 'static',
+     outDir: 'public',
+     publicDir: 'static',
    });
    ```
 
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/deploy/gitlab.md
```diff
@@ -13,30 +13,44 @@ Check out [the official GitLab Pages Astro example project](https://gitlab.com/p
...
+
+Here is an example of a `astro.config.mjs` where the `public/` directory is set to `static`.
 
    ```js
    export default defineConfig({
      sitemap: true,
      site: 'https://astro.build/',
+     dist: 'public',
+     public: 'static',
    });
```
sarah11918, line 28:
`````suggestion
2. Set `dist` in `astro.config.mjs` to `public`. GitLab Pages requires exposed files to be located in a folder called "public". This instructs Astro to put the static build output in a folder called `public`.

If you were using the [`public/` directory](/en/guides/images/#public) as a source of static files in your Astro project, rename it and use that new folder name in `astro.config.mjs` for the value of `public`.

For example, here are the correct `astro.config.mjs` settings when the `public/` directory is renamed to `static/`:

   ```js
   export default defineConfig({
     sitemap: true,
     site: 'https://astro.build/',
     dist: 'public',
     public: 'static',
   });
`````
Check the preview to make sure this is properly indented, but since there is only one code sample shown for both sets of instructions (`dist` and `public`), I think it makes sense to keep them together both as step 2. (And then, the step below Step 3 doesn't need to renamed.)

File: src/pages/en/guides/deploy/gitlab.md [hunk 2]
```diff
@@ -13,30 +13,44 @@ Check out [the official GitLab Pages Astro example project](https://gitlab.com/p
...
          - node_modules/
      script:
+         # Specify the steps involved to build your app here
        - npm install
        - npm run build
+
      artifacts:
        paths:
+         # The folder that contains the built files to be published. This
+         # must be called "public".
```
sarah11918, line 49:
```suggestion
         # The folder that contains the built files to be published.
         # This must be called "public".
```
See how it looks in preview, but might make sense to give each sentence its own line?

src/pages/en/guides/deploy/gitlab.md
```diff
@@ -13,30 +13,44 @@ Check out [the official GitLab Pages Astro example project](https://gitlab.com/p
...
          - public
+         - 
      only:
+       # Trigger a new build and deploy only when there is a push to the
+       # below branch(es)
```
sarah11918, line 54:
```suggestion
       # Trigger a new build and deploy only when there is a push to the
       # branch(es) below
```

sarah11918 (commented):
This looks great @ralacerda ! Thank you for this contribution! 🥳 

I've added the Hacktoberfest label and made some suggestions below. See what you think!

ralacerda:
Thanks for the review @sarah11918 :partying_face: 

I think all your suggestions made sense and I committed them.

I also fixed the link to the `public` documentation to point to the configuration reference (instead of the picture reference).
I updated the `astro.config.mjs` file to use `outputDir` and `publicDir`. I tested it locally to make sure it worked.

src/pages/en/guides/deploy/gitlab.md
```diff
@@ -13,30 +13,47 @@ Check out [the official GitLab Pages Astro example project](https://gitlab.com/p
 
 ## How to deploy
 
-1. Set the correct `.site` in `astro.config.mjs`.
-2. Set `dist` in `astro.config.mjs` to `public` and `public` in `astro.config.mjs` to a newly named folder that is holding everything currently in `public`. The reasoning is because `public` is a second source folder in astro, so if you would like to output to `public` you'll need to pull public assets from a different folder. Your `astro.config.mjs` might end up looking like this:
+1. Set the correct `site` in `astro.config.mjs`.
+2. Set `outDir` in `astro.config.mjs` to `public`. GitLab Pages requires exposed files to be located in a folder called "public". This instructs Astro to put the static build output in a folder called `public`.
+
+If you were using the [`public/` directory](/en/reference/configuration-reference/#publicdir) as a source of static files in your Astro project, rename it and use that new folder name in `astro.config.mjs` for the value of `publicDir`.
```
sarah11918, line 19:
```suggestion
2. Set `outDir:public` in `astro.config.mjs`. GitLab Pages requires exposed files to be located in a folder called "public". This setting instructs Astro to put the static build output in a folder called `public`.

If you were using the [`public/` directory](/en/core-concepts/project-structure/#public) as a source of static files in your Astro project, rename it and use that new folder name in `astro.config.mjs` for the value of `publicDir`.
```
Good call on changing the link, and I think we can even do better! I think I like the link to the project structure page where we talk about each main folder/file in a basic Astro project setup.

I also changed the beginning of the instruction to make it less cluttered with all the styled words. Maybe this looks a little cleaner now?

ralacerda: I agree that is an even better link. :smile: 

I think the beginning is cleaner now.
But I just now noticed there is two phrases ending with "in a folder called `public`", so maybe if we bring them together the paragraph would look even better.

Maybe:

> This setting instructs Astro to put the static build output in a folder called `public`, which is the folder required by GitLab Pages for exposed files.

Since english is not my first language, I'm not sure if it sounds natural, so I'm totally open for suggestions.

sarah11918: I like it a lot! Let's do that, and then I think this PR is good to go!  🥳

sarah11918:
Welcome to Team Docs as a first-time contributor! 🚀 🧑‍🚀

PR #1698, opened 2022-10-03
[error msg fix] change "txt" lang to "js" in Github deploy instructions [en-only]

sarah11918:

This should avoid the error message about "txt" not being a supported language for Markdown code. I've changed every lanaguage, so am using the `en-only` keyword to prevent a status update trigger.


Jutanium:
LGTM - we don't need highlighting there, so this works great to make the error go away.

PR #1699, opened 2022-10-03
Revise and elaborate TypeScript intro

### LINKED ISSUE #1482: Clarify that Astro always treats your code as TypeScript

Jutanium:

- New or updated content

#### Description

- Closes #1482
- Explains that TypeScript helps prevent errors, motivating the Component Props section later on
- Clarifies that the VSCode extension gives you type hints whether or not you write TypeScript
- Clarifies that Astro always treats your component code as TypeScript, even. if you leave out a `tsconfig.json`

### DIFF
--- a/src/pages/en/guides/typescript.md
+++ b/src/pages/en/guides/typescript.md
@@ -14,8 +14,7 @@ You don't need to write TypeScript code in your Astro projects to benefit from i
 The Astro dev server won't perform any type checking, but you can use a [separate script](#type-checking) to check for type errors from the command line.
 
 ## Setup
-
-Astro starter projects include a `tsconfig.json` file in your project. Even if you don't write TypeScript code, this file is important so that tools like Astro and VS Code know how to understand your project. Some features (like npm package imports) aren't fully supported in the editor without a `tsconfig.json` file. If you install Astro manually, be sure to create this file yourself.
+Astro always treats your component code as TypeScript, but some features (like npm package imports) aren't fully supported in the editor without a `tsconfig.json` file. Even if you don't write TypeScript code, this file is important so that tools like Astro and VS Code know how to understand your project. Astro starter projects include this file, but if you install Astro manually, be sure to create it yourself.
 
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/typescript.md
```diff
@@ -7,11 +7,14 @@ i18nReady: true
...
 
-## Setup
+You don't need to write TypeScript code in your Astro projects to benefit from it. The [Astro VSCode Extension](/en/editor-setup/) uses TypeScript to provide autocompletion, hints, and errors in your editor.
+
+The Astro dev server won't perform any type checking, but you can use a [separate script](#type-checking) to check for type errors from the command line.
 
-Astro starter projects include a `tsconfig.json` file in your project. Even if you don't write TypeScript code, this file is important so that tools like Astro and VS Code know how to understand your project. Some features (like npm package imports) aren't fully supported in the editor without a `tsconfig.json` file. If you install Astro manually, be sure to create this file yourself.
+## Setup
+Astro always treats your component code as TypeScript, but some features (like npm package imports) aren't fully supported in the editor without a `tsconfig.json` file. Even if you don't write TypeScript code, this file is important so that tools like Astro and VS Code know how to understand your project. Astro starter projects include this file, but if you install Astro manually, be sure to create it yourself.
```
sarah11918, line 17:
```suggestion
Astro starter projects include a `tsconfig.json` file in your project. Even if you don't write TypeScript code, this file is important so that tools like Astro and VS Code know how to understand your project. Some features (like npm package imports) aren't fully supported in the editor without a `tsconfig.json` file. If you install Astro manually, be sure to create this file yourself.
```
I'm not sold on this change at all. (Above suggestion reverts to original.) 

The "but" doesn't seem to be the right relationship between these two statements here. I also prefer starting out by mentioning the `tsconfig.json` file; seems to be more direct, especially when the heading is "Setup."  (Here's the setup file!)

- You'll see this file
- Even if you don't think you need it, you do
- if you installed manually, make it yourself.

I also don't immediately see the best place to stick in "Astro always treats your component code as TypeScript", which is the new idea you wanted to get in there. Maybe that's a clue that it doesn't belong in the "setup" section? What about throwing this idea up in the "you'll get benefits even if you don't use it" section?

Jutanium: Agreed, I think I overthought this one!

sarah11918:
I like it up there much better, @Jutanium! LGTM!

PR #1702, opened 2022-10-03
Remove Invalid Hook troubleshooting entry now that it's fixed

Jutanium:

- New or updated content

#### Description

- Since [this fix](https://github.com/withastro/astro/pull/4831) was merged, we no longer need to document this gotcha.

That said, maybe we should keep it and say that it was fixed in 1.3.0?


sarah11918:
Yay! How exciting! LGTM, @Jutanium

PR #1708, opened 2022-10-03
Workaround for `import.meta.glob` perf issue

### LINKED ISSUE #1695: Dev server slower boot time as of today
Hi all- I've been playing with this docs repo recently and noticed that today it was far slower to boot up. I _think_ the issue began after the [og-image PR](https://github.com/withastro/docs/commit/6b78cffd72d2989af88ac160c2ac62e9a287656d) was merged but not 100% sure. The terminal is also showing this error about `The language "txt" doesn't exist, falling back to plaintext.`: 

Screen Shot 2022-10-02 at 8 25 13 PM

delucis:

#### Description

- Closes #1695 

- Removes an `import.meta.glob` call that was tanking dev server performance on start up

- The workaround is not particularly pretty. Basically I hand-roll a minimal version of Astro’s glob + parse Markdown pipeline that does just what we need in this context: globs all the `.md` pages, parses their frontmatter with `gray-matter` and gives us `title` and `description`. At least in my local tests, this resolves our issues although obviously it’s not battle tested, so a Windows test might throw up some path weirdness?

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "eslint": "^8.21.0",
     "eslint-plugin-astro": "^0.14.0",
     "fast-glob": "^3.2.11",
+    "gray-matter": "^4.0.3",
     "hastscript": "^7.0.2",
     "html-escaper": "^3.0.3",
     "htmlparser2": "^7.2.0",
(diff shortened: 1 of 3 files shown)


Jutanium:
Fixes it for me!

sarah11918:
Faster here!

delucis:
Thanks for testing! This solution still feels slightly gross but maybe it can go away once we migrate to MDX.

PR #1709, opened 2022-10-03
Dynamic routes revision

### LINKED ISSUE #1631: Dynamic routes section assumes SSG mode
In our [Dynamic Routes](https://docs.astro.build/en/core-concepts/routing/#dynamic-routes) documentation, we assume that `getStaticPaths` is required, and don't mention SSR mode, where it is ignored.

We do print a warning when `getStaticPaths` is used in SSR mode, but there's still a potential gotcha here of a user copy-and-pasting an example when they're in SSR mode and wondering why it isn't working.

We might need to divide this section between SSR and SSG mode, because it isn't as simple as leaving out `getStaticPaths` in SSR mode - you have to replace it with the analogous logic, but inside the frontmatter. So for our **Example: Dynamically create a top level root page** to work on SSR, it would need to look something like this:

```astro
---
const pages = {
  "products": {
    title: "Astro products",
    text: "We have lots of products for you",
  },
  "products/astro-handbook": {
    title: "The ultimative Astro handbook",
    text: "If you want to learn Astro, you must read this book.",
  },
}

const defaultPage = {
    title: "Astro Store",
    text: "Welcome to the Astro store!",
  };

const slug = Astro.params.slug;
const { title, text } = pages[slug] || defaultPage;
---
<html>
  <head>
    <title>{title}</title>
  </head>
  <body>
    <h1>{title}</h1>
    <p>{text}</p>
  </body>
</html>
```

Jutanium:

- New or updated content

#### Description

- Closes #1631
- Introduces the different setups and behavior for dynamic routes in SSR vs SSG
- Introduces `Astro.params` alongside `getStaticPaths`, rather than in a separate section. This matches the tutorial, but also avoids having to repeat examples to show the full picture.
- Removes incomplete "examples" that don't show the implementation
- Reimplements the final example so that it works in SSR

Right now, much of the "guide"-style content for `getStaticPaths` is in the references section. A future PR would restructure and bring that information here.

### DIFF
--- a/src/pages/en/core-concepts/routing.md
+++ b/src/pages/en/core-concepts/routing.md
@@ -34,7 +34,9 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
 
 ## Dynamic routes
 
-A single Astro page file can specify dynamic route parameters in its filename to generate multiple routes that match. For example, you might create an `[author].astro` file that generate a bio page for every author on your blog. `author` becomes a _parameter_ that you can access from inside the page.
+A single Astro page file can specify dynamic route parameters in its filename to generate matching pages. For example, you might create an `authors/[author].astro` file that generates a bio page for every author on your blog. `author` becomes a _parameter_ that you can access from inside the page.
+
+In the default static mode, these pages are generated at build time, and so you must predetermine the list of `author`s that get a corresponding file. In SSR mode, the pages will be created on request for any route that matches.
 
 <!-- Named parameters allow you to specify values for "named" levels of these route paths, and rest parameters allow for more flexible "catch-all" routes. -->
 
@@ -79,16 +81,17 @@ export function getStaticPaths () {
 }
 
 const { lang, version } = Astro.params;
-...
 ---
 ...
 ```
 
 This will generate `/en-v1/info` and `/fr-v2/info`.
 
+Parameters can also be part of enclosing folder names, so we could use `src/pages/[lang]/[version]/info.astro` with the same `getStaticPaths` to generate `/en/v1/info` and `/fr/v2/info`.
+
 📚 Learn more about [`getStaticPaths()`](/en/reference/api-reference/#getstaticpaths).
 
-### Rest parameters
+#### Rest parameters
 
 If you need more flexibility in your URL routing, you can use a rest parameter (`[...param]`) in your `.astro` filename to match file paths of any depth:
 
@@ -103,7 +106,6 @@ export function getStaticPaths() {
 }
 
 const { path } = Astro.params;
-...
 ---
 ...
 ```
@@ -114,7 +116,7 @@ This will generate `/sequences/one/two/three`, `/sequences/four`, and `/sequence
 
 Here, we use a rest parameter (`[...slug]`) and the [props](https://docs.astro.build/en/reference/api-reference/#data-passing-with-props) feature of `getStaticPaths()` to generate pages for slugs of different depths.
 
-```astro title="src/pages/[...slug].astro" "slug: undefined"
+```astro title="src/pages/[...slug].astro"
 ---
 export async function getStaticPaths() {
   const pages = [
@@ -154,6 +156,55 @@ const { title, text } = Astro.props;
   </body>
 </html>
 ```
+
+### Server (SSR) Mode
+In [SSR mode](en/guides/server-side-rendering/), dynamic routes are defined the same way: include `[param]` or `[...path]` brackets in your file names to match arbitrary strings or paths. But because the routes are no longer built ahead of time, the page will be served to any matching route, and `getStaticPaths` has no effect. 
+
+```astro title="src/pages/resources/[resource]/[id].astro"
+---
+const { resource, id } = Astro.params;
+---
+<h1>{resource}: {id}<h1>
+```
+This page will be served for any value of `resource` and `id`: `resources/users/1`, `resources/colors/blue`, etc.
+
+#### Modifying the `[...slug]` example for SSR
+
+Because SSR pages can't use `getStaticPaths`, they can't receive props. Here, we modify our [earlier example](#example-dynamic-pages-at-multiple-levels) to work in SSR by looking up the value of the `slug` param in an object.
+
+```astro title="src/pages/[...slug].astro"
+---
+const pages = {
+  "products": {
+    title: "Astro products",
+    text: "We have lots of products for you",
+  },
+  "products/astro-handbook": {
+    title: "The ultimative Astro handbook",
+    text: "If you want to learn Astro, you must read this book.",
+  },
+}
+
+const defaultPage = {
+    title: "Astro Store",
+    text: "Welcome to the Astro store!",
+  };
+
+const slug = Astro.params.slug;
+const { title, text } = pages[slug] || defaultPage;
+
+---
+<html>
+  <head>
+    <title>{title}</title>
+  </head>
+  <body>
+    <h1>{title}</h1>
+    <p>{text}</p>
+  </body>
+</html>
+```
+
 
 ## Route Priority Order
 
--- a/src/pages/en/core-concepts/routing.md
+++ b/src/pages/en/core-concepts/routing.md
@@ -34,117 +34,85 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
 
 ## Dynamic routes
 
-A single Astro Page component can also specify dynamic route parameters in its filename to generate multiple routes that match a given criteria. You can create several related pages at once, such as author pages, or a page for each blog tag. Named parameters allow you to specify values for "named" levels of these route paths, and rest parameters allow for more flexible "catch-all" routes.
+A single Astro page file can specify dynamic route parameters in its filename to generate multiple routes that match. For example, you might create an `[author].astro` file that generate a bio page for every author on your blog. `author` becomes a _parameter_ that you can access from inside the page.
 
-:::note
-Even dynamically-created pages and routes are generated at build time.
-:::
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/routing.md
```diff
@@ -34,119 +34,81 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
 
 ## Dynamic routes
 
-A single Astro Page component can also specify dynamic route parameters in its filename to generate multiple routes that match a given criteria. You can create several related pages at once, such as author pages, or a page for each blog tag. Named parameters allow you to specify values for "named" levels of these route paths, and rest parameters allow for more flexible "catch-all" routes.
+A single Astro page file can specify dynamic route parameters in its filename to generate matching pages. For example, you might create an `authors/[author].astro` file that generates a bio page for every author on your blog. `author` becomes a _parameter_ that you can access from inside the page.
```
delucis, line 37: Are we relying on "single" doing any work here? If so it's not clear and if not I think we can remove:

```suggestion
An Astro page file can specify dynamic route parameters in its filename to generate matching pages. For example, you might create an `authors/[author].astro` file that generates a bio page for every author on your blog. `author` becomes a _parameter_ that you can access from inside the page.
```

src/pages/en/core-concepts/routing.md
```diff
@@ -34,119 +34,81 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
...
-- `pages/[username]/settings.astro` → (`/fred/settings`, `/drew/settings`, etc.)
-- `pages/[lang]-[version]/info.astro` → (`/en-v1/info`, `/fr-v2/info`, etc.)
+The filename can include multiple parameters, which must all be included in the `params` objects in `getStaticPaths()`:
 
-#### The `Astro.params` object
-
-Astro components that generate routes dynamically have access to an `Astro.params` object for each route. This allows you to use those generated parts of the URL in your component script and template.
-
-```astro / (id) |{id}/ /(?<!//.*)Astro.params/
+```astro title="src/pages/[lang]-[version]/info.astro"
```
delucis, line 66: Ooh! I'd always assumed these had to be separated by slashes. Cool that you can do this too!

src/pages/en/core-concepts/routing.md
```diff
@@ -34,119 +34,81 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
...
-:::tip
-Rest parameters are optional by default, so `pages/post/[...slug].astro` could match `/post/` as well.
-:::
+#### Rest parameters
 
-#### Example: Rest parameters
+If you need more flexibility in your URL routing, you can use a rest parameter (`[...param]`) in your `.astro` filename to match file paths of any depth:
 
-For a real-world example, you can implement GitHub’s file viewer with the following named and rest paramenters:
+```astro title="src/pages/sequences/[...path]"
```
delucis, line 90: We need an extension here right?
````suggestion
```astro title="src/pages/sequences/[...path].astro"
````

File: src/pages/en/core-concepts/routing.md [hunk 4]
```diff
@@ -187,6 +149,55 @@ const { title, text } = Astro.props;
...
+---
+const { resource, id } = Astro.params;
+---
+<h1>{resource}: {id}<h1>
+```
+This page will be served for any value of `resource` and `id`: `resources/users/1`, `resources/colors/blue`, etc.
+
+#### Modifying the `[...slug]` example for SSR
+
+Because SSR pages can't use `getStaticPaths`, they can't receive props. Here, we modify our [earlier example](#example-dynamic-pages-at-multiple-levels) to work in SSR by looking up the value of the `slug` param in an object.
```
delucis, line 165: Given folks likely want either static or SSR docs, I wonder if linking these is ideal? But certainly NWTWWHB!

src/pages/en/core-concepts/routing.md
```diff
@@ -187,6 +149,55 @@ const { title, text } = Astro.props;
...
+    title: "Astro products",
+    text: "We have lots of products for you",
+  },
+  "products/astro-handbook": {
+    title: "The ultimative Astro handbook",
+    text: "If you want to learn Astro, you must read this book.",
+  },
+}
+
+const defaultPage = {
```
delucis, line 180: Instead of having a default, would it make sense to show a 404 example here if the page isn't found?

delucis (commented):
Awesome work @Jutanium! Getting to the pointy end of this end points work 👏

src/pages/en/core-concepts/routing.md
Jutanium: Ah thanks for pointing this out, this example actually isn't equivalent to the SSG one because of this - changed!

src/pages/en/core-concepts/routing.md
Jutanium: I was thinking that the differences between the two would be instructive, so I wanted to link them so you could compare them. 

Do you think it's useful to have this equivalent SSR example to match the SSG one in the first place?

Jutanium:
@delucis and @kevinzunigacuellar both liked the GitHub file routes example, and even though it isn't "a real example", it illustrates the combination of rest parameters with other named parameters really well. Added back!

src/pages/en/core-concepts/routing.md
```diff
@@ -34,104 +34,82 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
...
-    // Generates: /dogs/rover
     {params: {dog: 'rover'}},
-    // Generates: /dogs/spot
     {params: {dog: 'spot'}},
   ];
 }
+
+const { dog } = Astro.params;
 ---
+<div>Good dog, {dog}!</div>
```
sarah11918, line 59: 😍

src/pages/en/core-concepts/routing.md
```diff
@@ -34,104 +34,82 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
...
 ```
 
-📚 Learn more about [`getStaticPaths()`](/en/reference/api-reference/#getstaticpaths).
-
-Routes can be generated from multiple named parameters, at any level of the filepath:
-
-- `pages/blog/[slug].astro` → (`/blog/hello-world`, `/blog/post-2`, etc.)
-- `pages/[username]/settings.astro` → (`/fred/settings`, `/drew/settings`, etc.)
-- `pages/[lang]-[version]/info.astro` → (`/en-v1/info`, `/fr-v2/info`, etc.)
+This will generate three pages: `dogs/clifford`, `dogs/rover`, and `dogs/spot`, each displaying the corresponding dog name.
```
sarah11918, line 62:
```suggestion
This will generate three pages: `/dogs/clifford`, `/dogs/rover`, and `/dogs/spot`, each displaying the corresponding dog name.
```

src/pages/en/core-concepts/routing.md
```diff
@@ -34,104 +34,82 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
...
 
-#### The `Astro.params` object
+The filename can include multiple parameters, which must all be included in the `params` objects in `getStaticPaths()`:
```
sarah11918, line 64: Love this, and the example below!

src/pages/en/core-concepts/routing.md
```diff
@@ -142,11 +120,11 @@ In this example, a request for `/withastro/astro/tree/main/docs/public/favicon.s
 }
 ```
 
-#### Example: Dynamically create a top level root page
+#### Example: Dynamic pages at multiple levels
 
-To dynamically create an index.html at root level (e.g. for content fetched from a headless CMS), include an object with `slug: undefined` in your `getStaticPaths()` function.
+Here, we use a rest parameter (`[...slug]`) and the [props](/en/reference/api-reference/#data-passing-with-props) feature of `getStaticPaths()` to generate pages for slugs of different depths.
```
sarah11918, line 125:
```suggestion
Here, we use a rest parameter (`[...slug]`) and the [`props`](/en/reference/api-reference/#data-passing-with-props) feature of `getStaticPaths()` to generate pages for slugs of different depths.
```
Do we want props code-syntax'ed?

src/pages/en/core-concepts/routing.md
```diff
@@ -187,6 +165,55 @@ const { title, text } = Astro.props;
 </html>
 ```
 
+### Server (SSR) Mode
+In [SSR mode](/en/guides/server-side-rendering/), dynamic routes are defined the same way: include `[param]` or `[...path]` brackets in your file names to match arbitrary strings or paths. But because the routes are no longer built ahead of time, the page will be served to any matching route, and `getStaticPaths` has no effect.
```
sarah11918, line 169:
```suggestion
In [SSR mode](/en/guides/server-side-rendering/), dynamic routes are defined the same way: include `[param]` or `[...path]` brackets in your file names to match arbitrary strings or paths. But because the routes are no longer built ahead of time, the page will be served to any matching route. Since there are no "static" routes, `getStaticPaths` has no effect and is not necessary.
```
Not married to this wording, but it seems to me that there are two nuances I might choose to address here:

- "gSP has no effect" sounds a little weirder and less explicit than "so you don't NEED/USE gSP to make your dynamic routes"
- you could totally use the name itself to emphasize the difference/lack of need to use it! 

I leave it to you to decide! (But I did think "has no effect" was a bit weird sounding.)

sarah11918 (commented):
Looks amazing, @Jutanium! Just left some quick comments for your consideration!

src/pages/en/core-concepts/routing.md
Jutanium: Agreed, love the idea of referencing the name! I wanted to make it clear that you _shouldn't_ use GSP:

> Since these are not _static_ routes, `getStaticPaths` should not be used.

sarah11918:
LGTM!

PR #1720, opened 2022-10-04
Caution against passing functions to HTML elements, framework components

### LINKED ISSUE #1588: Gotcha: passing functions as props
- You can pass functions as props to Astro components, and they can use them in frontmatter / to determine what to render. 
-  But you can't pass functions as events to elements. There's a common "gotcha" moment we see when a user tries:
```astro
---
function handleClick () {
    console.log("button clicked!");
}
---
<button onClick={handleClick}/>
```
- You can pass functions from Astro components to framework components, but _only_ if they don't have a client directive. This is both a gotcha and just info missing from the Framework Components page.

Jutanium:

- New or updated content

#### Description

- Closes #1588
- Caution for passing functions to elements in the Dynamic Attributes section of Astro Components
- Adds a "passing props to framework components" section (to match "passing children to framework components")
- Adds a caution in that section for passing functions as props to hydrated framework components

This was my best guess for placement; the "passing event handlers in Astro components" could deserve a spot in the troubleshooting section, too.


delucis (commented):
Thanks, Dan! Looks good 🙌

src/pages/en/core-concepts/astro-components.md
```diff
@@ -143,6 +143,23 @@ const name = "Astro";
 <MyComponent templateLiteralNameAttribute={`MyNameIs${name}`} />
 ```
 
+:::caution
+HTML element attributes will be converted to strings, so functions and objects don't work predictably when passed as HTML attributes.
+Notably, you can't pass an event handler from an Astro component to an HTML element:
```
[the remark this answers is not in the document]
Jutanium: Ah, if "from" refers to the "passing", in my mind this happens in the component template where the attribute assignment is. But in my wording "from" sounds like it's referring to where the event handler function originates, which would be the frontmatter. Changed to:
> Notably, you can't assign an event handler to an HTML element in an Astro component:

src/pages/en/core-concepts/astro-components.md
```diff
@@ -143,6 +143,23 @@ const name = "Astro";
...
+
+```astro
+---
+function handleClick () {
+    console.log("button clicked!");
+}
+// This doesn't work!
+---
+<button onClick={handleClick}>Nothing will happen when you click me!</button>
```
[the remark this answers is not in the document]
Jutanium: Do you think it isn't necessary to illustrate this gotcha? I felt that it might not be clear enough without an example. I like the idea of showing an alternative, and maybe that will reduce the chance of people trying the first code example if we keep it:


> Notably, you can't assign an event handler to an HTML element in an Astro component:

```astro
---
function handleClick () {
    console.log("button clicked!");
}
// This doesn't work!
---
<button onClick={handleClick}>Nothing will happen when you click me!</button>
```

Instead, use a client-side script to add the event handler:

```astro
<button id="button">Click Me</div>
<script>
function handleClick () {
  console.log("button clicked!");
}
document.getElementById("button").addEventListener("click", handleClick);
</script>
```

Let me know what you think.

Jutanium:
I added a working example alongside the broken one. I agree that there's a danger in showing bad examples, but I don't think it's clear what the gotcha is without one.

src/pages/en/core-concepts/astro-components.md
```diff
@@ -143,6 +143,35 @@ const name = "Astro";
...
+```
```
delucis, line 158: A bit nit-picky, but wondering if we can do more to highlight that this is the BROKEN example:

````suggestion
```astro
---
// dont-do-this.astro
function handleClick () {
    console.log("button clicked!");
}
---
<!-- ❌ This doesn't work! ❌ -->
<button onClick={handleClick}>Nothing will happen when you click me!</button>
```
````

File: src/pages/en/core-concepts/astro-components.md [hunk 4]
```diff
@@ -143,6 +143,35 @@ const name = "Astro";
...
+
+Instead, use a client-side script to add the event handler:
+
+```astro
+---
```
delucis, line 163: We could add a title here too

```suggestion
---
// do-this-instead.astro
```

delucis (commented):
Looks good, Dan! Made a suggestion to try and highlight the non-working example, but also OK with leaving this as-is if you’re happy.

src/pages/en/core-concepts/astro-components.md
```diff
@@ -143,6 +143,37 @@ const name = "Astro";
...
+// dont-do-this.astro
+function handleClick () {
+    console.log("button clicked!");
+}
+---
+<!-- ❌ This doesn't work! ❌ -->
+<button onClick={handleClick}>Nothing will happen when you click me!</button>
+```
+
+Instead, use a client-side script to add the event handler:
```
sarah11918, line 161: Either here, or in the earlier text, is it worth a tiny call out to "think about how you'd do this in HTML/JS"? This is a great example, both of what people will try to do and what they should do instead. 

The two approaches seem different enough from each other -- one is very "framework-y" where your component is going to render your button and the other is like, just make an html button! One passes props, and the other does not at all. I do wonder if a call out re: "thinking differently" might be helpful.

sarah11918 (commented):
Love this! Just left a few comments to also consider, but looks like you and Chris have got it down. 🙌

src/pages/en/core-concepts/astro-components.md
Jutanium: We only want them to think differently in _this_ case (passing an event handler), as a workaround. Other dynamic props still work, so I don't want to overstate the case.

Would you prefer something like this, to invoke the similarity?

> Instead, use a client-side script to add the event handler, like you would in vanilla JavaScript:

sarah11918: Yup, I think something like that is perfect!

src/pages/en/core-concepts/framework-components.md
```diff
@@ -119,6 +119,26 @@ import MyVueComponent from '../components/MyVueComponent.vue';
...
+import Counter from '../components/Counter.svelte';
+---
+<div>
+  <TodoList initialTodos={["learn Astro", "review PRs"]} />
+  <Counter startingCount={1} />
+</div>
+```
+
+:::caution
+A function can be passed as a prop to a framework component, but the function can only be called on the server. This is because functions can't be _serialized_ (transferred from the server to the client) by Astro. If the component is hydrated, an error will occur when it tries to call the function.
```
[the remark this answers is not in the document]
sarah11918: Is this a case of "You probably don't want to do this, because it doesn't work the way you might think it does/the way it does in SPAs?" I hesitate to start the note with "You CAN pass it... " when the result is very likely to be "... but it's not gonna work."

If you can't even describe a good use case, what if we phrase this to advise against doing this? I still don't get a good sense from reading this when it actually would be a working thing for me to do.

So, what is the caution trying to caution us ABOUT - it doesn't work the way you think it will? It's so rarely going to be the thing you want that you probably should try a different pattern in your code? I think that could still be clearer here.

sarah11918 (commented):
More comments!

src/pages/en/core-concepts/framework-components.md
```diff
@@ -119,6 +119,28 @@ import MyVueComponent from '../components/MyVueComponent.vue';
...
+import TodoList from '../components/TodoList.jsx';
+import Counter from '../components/Counter.svelte';
+---
+<div>
+  <TodoList initialTodos={["learn Astro", "review PRs"]} />
+  <Counter startingCount={1} />
+</div>
+```
+
+:::caution[passing functions as props]
```
sarah11918, line 137: I like this better! I think the structure is more helpful here!

sarah11918:
LGTM!

PR #1726, opened 2022-10-05
Add a section on passing a class to a child component

### LINKED ISSUE #995: Document how to apply a `class` to an Astro component

Jutanium:

- New or updated content
#### Description

- Closes #995 
- Adds a new section on the Styling page, per [this comment](https://github.com/withastro/docs/pull/1011#issuecomment-1185737025)
- I focused on passing `class`, but made it clear that all HTML attributes don't automatically pass through
- Shows two examples: one where you don't destructure props at all, and one where you destructure but rename

### DIFF
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -114,7 +114,7 @@ interface Props {
 </div>
 ```
 
- If you'd like to destructure a props object that contains `class`, you must rename it. This is because `class` is a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript.
+ If you'd like to destructure a props object that contains a `class` prop, you must rename it. This is because `class` is a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript.
 
 ```astro title="src/components/MyComponent.astro" {6,8}
 ---
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -97,10 +97,13 @@ const backgroundColor = "rgb(24 121 78)";
 In Astro, HTML attributes like `class` do not automatically pass through to child components.
 
 
-Instead, accept a `class` prop in the child component and apply it to the root element:
+Instead, accept a `class` prop in the child component and apply it to the root element. When destructuring, you must rename it, because `class` is a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript.
 
-```astro title="src/components/MyComponent.astro" {1}
-<div class={Astro.props.class}>
+```astro title="src/components/MyComponent.astro" {2,4}
+---
+const { class: className } = Astro.props;
+---
+<div class={className}>
   <slot/>
 </div>
 ```
@@ -117,16 +120,7 @@ import MyComponent from "../components/MyComponent.astro"
 <MyComponent class="red">This will be red!</MyComponent>
 ```
 
- If you'd like to destructure a props object that contains a `class` prop, you must rename it. This is because `class` is a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript.
 
-```astro title="src/components/MyComponent.astro" {2,4}
----
-const { class: className } = Astro.props;
----
-<div class={className}>
-  <slot/>
-</div>
-```
 
 ## External Styles
 
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -92,6 +92,43 @@ const backgroundColor = "rgb(24 121 78)";
 
 📚 See our [directives reference](/en/reference/directives-reference/#definevars) page to learn more about `define:vars`.
 
+### Passing a `class` to a child component
+
+In Astro, HTML attributes like `class` do not pass through to a child component unless they are used as props.
+
+```astro title="src/pages/index.astro"
+---
+import MyComponent from "../components/MyComponent.astro"
+---
+<MyComponent class="red">Color me!</MyComponent>
+```
+
+```astro title="src/components/MyComponent.astro" {6}
+---
+interface Props {
+  class: string
+}
+---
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/styling.md
```diff
@@ -92,6 +92,43 @@ const backgroundColor = "rgb(24 121 78)";
...
 
+### Passing a `class` to a child component
+
+In Astro, HTML attributes like `class` do not pass through to a child component unless they are used as props.
+
+```astro title="src/pages/index.astro"
+---
+import MyComponent from "../components/MyComponent.astro"
+---
+<MyComponent class="red">Color me!</MyComponent>
```
kevinzunigacuellar, line 103: I was thinking maybe we can add the style tag to make it more obvious.

```suggestion
<style>
  .red {
    color: red;
  }
</style>
<!-- this won't color the text to red -->
<MyComponent class="red">Color me!</MyComponent>
```

src/pages/en/guides/styling.md
```diff
@@ -92,6 +92,43 @@ const backgroundColor = "rgb(24 121 78)";
...
+```
+
```
kevinzunigacuellar, line 105:
```suggestion
Instead, use props to apply styles to child components
```

kevinzunigacuellar (commented):
just a random idea I got while reading this, feel free to dismiss it

src/pages/en/guides/styling.md
Jutanium: This makes sense, thanks!

src/pages/en/guides/styling.md
```diff
@@ -92,6 +92,51 @@ const backgroundColor = "rgb(24 121 78)";
 
 📚 See our [directives reference](/en/reference/directives-reference/#definevars) page to learn more about `define:vars`.
 
+### Passing a `class` to a child component
+
+In Astro, HTML attributes like `class` do not automatically pass through to child components.
```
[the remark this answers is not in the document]
Jutanium: Agreed - I removed the bad example, and added `index.astro` to the working example.

src/pages/en/guides/styling.md
```diff
@@ -92,6 +92,51 @@ const backgroundColor = "rgb(24 121 78)";
...
+```astro title="src/pages/index.astro"
+---
+import MyComponent from "../components/MyComponent.astro"
+---
+<style>
+  .red {
+    color: red;
+  }
+</style>
+<MyComponent class="red">This will be red!</MyComponent>
```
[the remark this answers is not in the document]
Jutanium: It does work ([Stackblitz](https://stackblitz.com/edit/github-epdtsu?file=src/pages/index.astro)) - it looks like Astro passes the scoped class name down to its children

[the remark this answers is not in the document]
Jutanium: @delucis Ha, this was opened today: https://github.com/withastro/astro/issues/5112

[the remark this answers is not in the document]
delucis: Ooh… that is WEIRD behaviour! And it only passes the scoped class when you pass a `class`… 🤔 

I’d still be tempted for the docs example here to be more tailored to a global CSS situation either like a Tailwind example or using `is:global` or a global CSS file. At least for now until that bug report is triaged and we know if that’s expected or not.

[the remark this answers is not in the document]
Jutanium: I went and changed it to `is:global`, as it doesn't require another import or explaining a Tailwind dependency for the example.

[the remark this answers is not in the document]
delucis: Thanks!

delucis:
That is a fair point Sarah — I guess the examples would also be more succinct and focused without the `Props` interfaces.

Jutanium:
@sarah11918 Removed the TS, should be good to go!

src/pages/en/guides/styling.md
```diff
@@ -92,6 +92,42 @@ const backgroundColor = "rgb(24 121 78)";
...
 
+### Passing a `class` to a child component
+
+In Astro, HTML attributes like `class` do not automatically pass through to child components.
+
+
+Instead, accept a `class` prop in the child component and apply it to the root element:
+
+```astro title="src/components/MyComponent.astro" {1}
+<div class={Astro.props.class}>
```
sarah11918, line 103: Happier now with these examples! Just pointing out that I don't think we show `Astro.props` anywhere else in Docs as an attribute directly. We *always* have a `const { class } = Astro..props` destructuring statement in frontmatter, as far as I can remember.

I'm not opposed to this, and I know it makes the example more concise. But I worry about having the only time we show this be in one of these "you can't do this, you have to do this!" kind of situations. If we had introduced in a more general passing props intro-y example, then this wouldn't stand out as, "Oh, maybe for exactly this situation, I need to do this?"

Thoughts?

Jutanium: It _is_ something you need for just this situation! It's one of two workarounds for `class` being a reserved word: you either destructure and rename (second example) or leave it on the `props` object (this first example).

Jutanium: If it's confusing, I could just remove it and keep the second workaround.

sarah11918: Ah, I totally see that now, that they are two versions of the same thing! I kind of... want to just keep the second example? It seems more in keeping with the rest of the Docs? We still never show applying something from props directly like that.

OR, I'd show the second one first, since that is very Astro-Docs-like. Then, show this as an example of avoiding needing to rename, and THIS becomes the, "Hey, you can actually do this, too, skipping the frontmatter" example that doesn't otherwise show up in Docs.

sarah11918:
LOOKS GOOD TO MERGE! @Jutanium

sarah11918:
LGTM @Jutanium

PR #1728, opened 2022-10-05
Revise hydration section

Jutanium:

- New or updated content

#### Description

- Replace a Caution which made it seem like the component framework JavaScript was sent on page load
- Add a client:only example to the code snippet, because we mention that "most directives" render the component on the server
- Make the intro more concise

### DIFF
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -65,7 +65,7 @@ By default, your framework components will only render on the server, as static 
 
 A framework component can be made interactive (hydrated) using a [`client:*` directive](/en/reference/directives-reference/#client-directives). These are component attributes that determine when your component's JavaScript should be sent to the browser.
 
-With client: `idle`, `load` or `visible`, your component will first render on the server to generate static HTML. Component JavaScript will be sent to the browser according to the directive you chose. The component will then hydrate and become interactive.
+With all client directives except `client:only`, your component will first render on the server to generate static HTML. Component JavaScript will be sent to the browser according to the directive you chose. The component will then hydrate and become interactive.
 
 ```astro title="src/pages/interactive-components.astro" /client:\S+/
 ---
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -82,7 +82,7 @@ the user scrolls down and the component is visible on the page -->
 <InteractiveCounter client:visible />
 
 <!-- This component won't render on the server, but will render on the client when the page loads -->
-<InteractiveModal client:only="svelte">
+<InteractiveModal client:only="svelte" />
 ```
 
 The JavaScript framework (React, Svelte, etc) needed to render the component will be sent to the browser along with the component's own JavaScript. If two or more components use the same framework, the framework will only be sent once.
--- a/src/pages/en/core-concepts/framework-components.md
+++ b/src/pages/en/core-concepts/framework-components.md
@@ -63,17 +63,16 @@ By default, your framework components will only render on the server, as static 
 
 ## Hydrating Interactive Components
 
-A framework component can be made interactive (hydrated) using one of the `client:*` directives. This is a component attribute to define how your component should be **rendered** and **hydrated**.
+A framework component can be made interactive (hydrated) using a [`client:*` directive](/en/reference/directives-reference/#client-directives). These are component attributes that determine when your component's JavaScript should be sent to the browser.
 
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/framework-components.md
```diff
@@ -59,33 +59,33 @@ import MyReactComponent from '../components/MyReactComponent.jsx';
...
 <InteractiveCounter client:visible />
+
+<!-- This component won't render on the server, but will render on the client when the page loads -->
+<InteractiveModal client:only="svelte" />
 ```
 
-:::caution
-Any renderer JS necessary for the component's framework (e.g. React, Svelte) is downloaded with the page. The `client:*` directives only dictate when the _component JS_ is imported and when the _component_ is hydrated.
-:::
+The JavaScript framework (React, Svelte, etc) needed to render the component will be sent to the browser along with the component's own JavaScript. If two or more components use the same framework, the framework will only be sent once.
```
sarah11918, line 88:
```suggestion
The JavaScript framework (React, Svelte, etc) needed to render the component will be sent to the browser along with the component's own JavaScript. If two or more components on a page use the same framework, the framework will only be sent once.
```
Have we confirmed whether the framework renderer JS is only sent once... per page? per... ever?

"on a page"... "in your project"...

Jutanium: Yes, Astro will only ever send the framework once per page! Pinging @delucis for a gut-check

delucis: “Once per page” seems fair 👍 

I guess technically the data transfer ends up being less because after the first page, you probably have the file containing the framework cached, so on subsequent pages that request hits your browser’s cache, but “once per page” seems a good bet to describe this.

src/pages/en/core-concepts/framework-components.md
```diff
@@ -59,33 +59,33 @@ import MyReactComponent from '../components/MyReactComponent.jsx';
...
+import InteractiveModal from "../components/InteractiveModal.svelte"
 ---
 <!-- This component's JS will begin importing when the page loads -->
 <InteractiveButton client:load />
 
 <!-- This component's JS will not be sent to the client until
 the user scrolls down and the component is visible on the page -->
 <InteractiveCounter client:visible />
+
+<!-- This component won't render on the server, but will render on the client when the page loads -->
```
kevinzunigacuellar, line 84: I think this information is too important to be left in a code-block comment. Maybe we can separate them into client and server-rendered components?

src/pages/en/core-concepts/framework-components.md
```diff
@@ -59,33 +59,33 @@ import MyReactComponent from '../components/MyReactComponent.jsx';
...
 
 ## Hydrating Interactive Components
 
-A framework component can be made interactive (hydrated) using one of the `client:*` directives. This is a component attribute to define how your component should be **rendered** and **hydrated**.
+A framework component can be made interactive (hydrated) using a [`client:*` directive](/en/reference/directives-reference/#client-directives). These are component attributes that determine when your component's JavaScript should be sent to the browser.
 
-This [client directive](/en/reference/directives-reference/#client-directives) describes whether or not your component should be rendered at build-time, and when your component's JavaScript should be loaded by the browser, client-side.
-
-Most directives will render the component on the server at build time. Component JS will be sent to the client according to the specific directive. The component will hydrate when its JS has finished importing.
+With most directives, your component will first render on the server to generate static HTML. Component JavaScript will be sent to the browser according to the directive you chose. The component will then hydrate and become interactive.
```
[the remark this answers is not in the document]
Jutanium: I agree that the "most" was awkward... I changed it to "with all client directives except `client:only`" for now,  because of `client:media`, but I agree with you that we should split the sections

src/pages/en/core-concepts/framework-components.md
Jutanium: Yeah, as I was working on this PR, I felt the awkwardness of explaining `client:only` along with the other client directives. I agree that ideally we should split this up, but I'm not sure if it should happen in this PR.

sarah11918:
Just replying to both comments, which are basically the same comment, that this PR as is makes things clearer, and I think it's fine to merge without figuring out how to deal with the pesky `client:only`. Remember, the heavy lifting of these descriptions is currently done on the Template Directives reference page. So the real mandate of this page is just to make people know that these exist, and to point them on to reference for more. (We may not always choose to keep this structure, but that's what we have right now, and it's OK to fall back on that.)

So, LGTM when you both are happy!

kevinzunigacuellar:
Lets Get This Merged! 🚀

PR #1767, opened 2022-10-10
i18n(es): Update `deploy/deno.md`

agustinmulet:

- New or updated content

#### Description

- Updates `deploy/deno.md` guide in spanish.

Please add hacktoberfest label 🎃

### DIFF
--- a/src/pages/es/guides/deploy/deno.md
+++ b/src/pages/es/guides/deploy/deno.md
@@ -157,7 +157,7 @@ Si tu proyecto está almacenado en GitHub, la [página web de Deno Deploy](https
         "dev": "astro dev",
         "start": "astro dev",
         "build": "astro build",
-        "preview": "deno run --allow-net --allow-read --allow-env ./dist/server/entry.mjs"
+        "preview": "deno run --allow-net --allow-read --allow-env ./dist/server/entry.mjs",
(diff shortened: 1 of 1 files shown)


AbrahamX3 (approved):

src/pages/es/guides/deploy/deno.md
```diff
@@ -158,7 +158,7 @@ Si tu proyecto está almacenado en GitHub, la [página web de Deno Deploy](https
         "start": "astro dev",
         "build": "astro build",
         "preview": "deno run --allow-net --allow-read --allow-env ./dist/server/entry.mjs"
```
fredoist, line 160: It's missing a comma here, same in English version, should it be added?

```suggestion
        "preview": "deno run --allow-net --allow-read --allow-env ./dist/server/entry.mjs",
```

breadadams: Yup @fredoist is right, looks like there's an invalid JSON here, in all locales. Maybe it can be tracked as a separate issue?

kevinzunigacuellar: @breadadams and @fredoist thank you for bringing this up. Yes, like @breadadams suggested I think the best way to go about this is to open a separate issue for invalid JSON and correct the original ("en") docs. This way all locales can correct this too. We can merge this PR after this issue is resolved.

PR #1770, opened 2022-10-10
Upgrade to Astro 1.4.7

zadeviggers:
- Updates Astro to `1.4.7` (latest release)

#### Description
This is needed for work on offline support since `vite-plugin-pwa` needs Vite `3.1.0` or greater, and Astro `1.2.1` (what the docs site is currently on) uses Vite `3.0.9`. I figured it would be much easier to get this done quickly in a separate pr to keep things tidy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@types/node": "^18.6.4",
     "@typescript-eslint/eslint-plugin": "^5.32.0",
     "@typescript-eslint/parser": "^5.32.0",
-    "astro": "1.2.1",
+    "astro": "1.4.7",
     "astro-eslint-parser": "^0.4.5",
     "astro-og-canvas": "^0.1.4",
     "bcp-47-normalize": "^2.1.0",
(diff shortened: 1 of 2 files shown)


aFuzzyBear (approved):
getting the docs offline as a pwa would be huge, great start on this @zadeviggers, 👊
I'll let @delucis sign off on this when he gets better,

delucis (approved):
LGTM — thanks Zade!

One thing: I was out last week so may have missed some discussion around adding PWA support. Could you point me to that discussion or open up an issue we could use to discuss before jumping into working on that? PWAs can introduce some dev complexities I’m not sure we necessarily need to take on right now, but would love to see a few different perspectives before we make a decision.

zadeviggers:
Hey @delucis, thanks for the comment. It's just been brought up a couple of times, and I thought I would take a shot at implementing it. It would be an opt-in system, which means it wouldn't affect development most of the time. 

I've requested talking about it be added to the agenda for the next community call so that everyone can add their thoughts in one place.

delucis:
Awesome, thanks @zadeviggers!

PR #1773, opened 2022-10-10
ci: make linkcheck 2 times faster

wtchnm:

- Something else!

#### Description

It makes the `lint:linkcheck` command run 2 times faster by disabling OG image generation.

##### Local

Hyperfine

##### CI

| Before | After |
| ------ | ----- |
| Before | After |

I'm participating in Hacktoberfest, so if PR is ok, I'd like it to be labeled as hacktoberfest-accepted.

### DIFF
--- a/src/pages/open-graph/[...path].ts
+++ b/src/pages/open-graph/[...path].ts
@@ -14,8 +14,7 @@ import { readFile } from 'fs/promises';
 // combo to extract title & description from each page.
 
 /** Paths for all of our Markdown content we want to generate OG images for. */
-// @ts-expect-error 2339 - env does not exist in import.meta as this file is not included in tsconfig.json
-const paths = !import.meta.env.SKIP_OG ? await glob('src/pages/**/*.md') : [];
+const paths = process.env.SKIP_OG ? [] : await glob('src/pages/**/*.md');
 
 /**
  * An object mapping file paths to a file loader method, mimicking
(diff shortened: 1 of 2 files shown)


sarah11918:
Thanks for submitting this @wtchnm! I've applied the Hacktoberfest label 🎃 so you should be good for credit there.

Unfortunately this PR will have to wait until og-Chris is back at the keyboard, so I just wanted to give you a quick response and let you know we've got this, but it might not move very quickly.  😄

wtchnm:
Hi @sarah11918, no problem. Thanks for adding the label and feedback!

src/pages/open-graph/[...path].ts
```diff
@@ -14,7 +14,8 @@ import { readFile } from 'fs/promises';
 // combo to extract title & description from each page.
 
 /** Paths for all of our Markdown content we want to generate OG images for. */
-const paths = await glob('src/pages/**/*.md');
+// @ts-expect-error 2339 - env does not exist in import.meta as this file is not included in tsconfig.json
+const paths = !import.meta.env.SKIP_OG ? await glob('src/pages/**/*.md') : [];
```
Comment (delucis) [inline src/pages/open-graph/[...path].ts:18]: We should be able to use `process.env` instead of `import.meta.env` and avoid the need for the `@ts-expect-error` comment.

```suggestion
const paths = process.env.SKIP_OG ? [] : await glob('src/pages/**/*.md');
```

delucis (commented):
This is awesome @wtchnm — thank you! Actually crossed my mind to add this when I was adding the OG images but didn’t get around to it. Super helpful. Made one nit-picky suggestion, but this is great!

wtchnm:
Thanks, @delucis. Suggestion accepted!

delucis (approved):
That was quick! Thanks again @wtchnm 🙌

PR #1777, opened 2022-10-11
Use `extendDefaultPlugins` in more code samples

mrienstra:

- New or updated content

#### Description

1. Use `extendDefaultPlugins: true` in more code samples. There were a few that I didn't add it to.

   My thought process when evaluating each one was basically: Would I expect this code sample to drop the default plugins as a side effect of adding a new plugin? Or, will I likely be carefully considering the behavior of the code sample, such that I am likely to understand that it drops the default plugins?

   Another motivation for these changes is to reduce the likelihood of new users blinding copying these code samples, then being confused as to why they can no longer use GitHub-flavored Markdown.

2. Plus more extendDefaultPlugins cross-linking (minor). One indentation fix.

---

Preview links:
- [/en/guides/integrations-guide/mdx/](https://deploy-preview-1777--astro-docs-2.netlify.app/en/guides/integrations-guide/mdx/)
- [/en/guides/markdown-content/](https://deploy-preview-1777--astro-docs-2.netlify.app/en/guides/markdown-content/)
- [/en/reference/configuration-reference/](https://deploy-preview-1777--astro-docs-2.netlify.app/en/reference/configuration-reference/)

### DIFF
--- a/src/pages/en/guides/integrations-guide/mdx.md
+++ b/src/pages/en/guides/integrations-guide/mdx.md
@@ -407,7 +407,7 @@ import shikiTwoslash from 'remark-shiki-twoslash';
 
 export default {
   markdown: {
-  syntaxHighlight: false,
+    syntaxHighlight: false,
   },
   integrations: [mdx({
     remarkPlugins: [shikiTwoslash, { /* Shiki Twoslash config */ }],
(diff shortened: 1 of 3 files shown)


delucis (commented):
Thanks for this @mrienstra! For the most part I think these are helpful changes, but one I think we probably want to revert to avoid over complicating an example where `extendDefaultPlugins` isn’t strictly speaking relevant.

delucis (approved):
@sarah11918 suggested I just go ahead and finish this off to get this merged — thanks again @mrienstra! 🙌

PR #1783, opened 2022-10-12
fix: missing trailing commas

breadadams:

#### Description

As discussed in https://github.com/withastro/docs/pull/1767#discussion_r991704745, inserted some missing trailing commas in `package.json` example fragments.

(I'm taking part in Hacktoberfest 🎃)

### DIFF
--- a/src/pages/en/guides/deploy/deno.md
+++ b/src/pages/en/guides/deploy/deno.md
@@ -58,8 +58,8 @@ Next, Update your `preview` script in `package.json` with the change below.
         "dev": "astro dev",
         "start": "astro dev",
         "build": "astro build",
-        "preview": "astro preview"
-        "preview": "deno run --allow-net --allow-read --allow-env ./dist/server/entry.mjs"
+        "preview": "astro preview",
+        "preview": "deno run --allow-net --allow-read --allow-env ./dist/server/entry.mjs",
       }
     }
     ```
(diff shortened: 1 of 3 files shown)


kevinzunigacuellar:
Thank you for the PR @breadadams! 🚀

I would like to follow our translating guidelines and only update our English version of the docs. This way all locales would be aware of the change and can update accordingly.

breadadams:
Ah good point @kevinzunigacuellar, I've reverted the change to the Spanish file and updated the PR.

sarah11918:
Hi @breadadams! Nice to see your name around in all these docs PR notifications!  🥳 

I'm looking at this now, but I think our code examples are correct as-is. Trailing commas are disallowed in `.json` objects, so it's proper without them.

For future reference (I know, we have so many rules to follow here because of our complicated setup!), in addition to only changing `en` files for things like this, some of our files are imported from source content in a different repository and simply mirrored here. So always check at the top of a file when you're editing, and there should be a note to tell you when NOT to edit a file directly and where the source file is located. A couple of these files here are actually like that, so we wouldn't edit them in this repo anyway.

So, I'm closing this PR, but I'm sure I'll be seeing you and your keen eyes in new ones! 🦅 👀

breadadams:
Hey @sarah11918, thanks 😄 

Yep, my bad, I was supposed to just fix one case where there's actually a missing comma and I went overboard 😅. I've adjusted the fix again, if you'd consider reopening the PR.

Thanks!

sarah11918:
Hi @breadadams, thanks! And yeah, sorry it's all a bit complicated. 😄  But the real problem is that it's not a correction to add the comma at all. There should be no trailing comma in json. In JavaScript it's common, and acceptable. But, not in json.

breadadams:
Yep no worries @sarah11918, I know. The issue lies [here](https://docs.astro.build/en/guides/deploy/deno/), in the Deno deploy guide:

image

Where a new line is being added to the package.json scripts. I've updated my branch but it doesn't reflect in the PR right now since it's closed.

sarah11918:
OK, gotcha! Yes that one (but only the 2nd last line) needs a comma. (And, in the earlier example, which is a diff, no comma needed because the deleted line wouldn't have had a comma, either.)

As long as we're on the same page!  😄

sarah11918:
I'm not able to reopen this one, there seems to be a problem. So, submit a different PR I guess? I'll apply the Hacktoberfest label there, too. (And, I think you might get credit for both, as long as the label here isn't removed.)

breadadams:
Right, yeah the fact that those code snippets are diffs is what threw me off.

Ok I'll open a fresh pull request. Thanks and sorry for creating all this noise!

PR #1801, opened 2022-10-17
Combine 0.0 and 0.1

Jutanium:
As discussed! I removed the bolded "concepts you'll learn along the way", because I felt that the intro was getting too heavy with those and they suffered from the "wait, I haven't heard of that yet" problem. But I'm okay if you want to add them back.


sarah11918:
LGTM Dan!

PR #1809, opened 2022-10-19
Fix OG image generation for pages like `index.md`

delucis:

#### Description

After merging #1089 I noticed Open Graph images weren’t showing correctly for some pages in the tutorial because they use `/directory/index.md` naming for some pages and this was not handled correctly (generating `/directory/index.png` images instead of the expected `/directory.png`).

This PR fixes that by tweaking some logic in our endpoint but mainly by upgrading `astro-og-canvas` where I fixed the core logic (https://github.com/delucis/astro-og-canvas/commit/98be21325b1ea6c176ad563e96af0ba333b899c2).

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@typescript-eslint/parser": "^5.32.0",
     "astro": "1.4.7",
     "astro-eslint-parser": "^0.4.5",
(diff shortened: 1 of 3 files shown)


Jutanium:
LGTM!

PR #1811, opened 2022-10-19
[tutorial] updated page titles and descriptions for unit check ins

sarah11918:
Choosing a structure for the x.0 "Check In" pages that works both in the navigation sidebar and for social images.


Jutanium:
LGTM!

PR #1812, opened 2022-10-19
Tutorial: add step to install astro extension

Jutanium:
Adding a step to install the Astro extension when you first open the project in VSCode


sarah11918:
LGTM, @Jutanium! Thanks for getting this in so quickly!

PR #1821, opened 2022-10-20
Use `import type` for `APIContext` imports in reference

mrienstra:

#### Description

Minor, `import { APIContext }` --> `import type { APIContext }`


delucis (approved):
Thanks @mrienstra — agree this helps clarify slightly what you’re importing even if not strictly speaking required.

mrienstra:
> Thanks @mrienstra — agree this helps clarify slightly what you’re importing even if not strictly speaking required.

@delucis -- you probably know this already -- it will show an error when extending `astro/tsconfigs/strict` or `astro/tsconfigs/strictest`, due to [tsconfigs/strict.json#L6-L7](https://github.com/withastro/astro/blob/863391f/packages/astro/tsconfigs/strict.json#L6-L7):
```json
    // Error when a value import is only used as a type.
    "importsNotUsedAsValues": "error"
```

PR #1823, opened 2022-10-20
Capitalize bullet list items

mrienstra:

#### Description

- Capitalize bullet list items
- Normalize spacing around `<PreCheck>` tags


delucis (commented):
This all LGTM _assuming_ we want these to be capitalised. I’ll leave that ultimate decision to @sarah11918 as keeper of our (implicit) style guide! 🗝️

sarah11918:
I'm OK with this for consistency. I actually don't like the capitalization visually in our `<PreCheck />` components, I much prefer the lower case there. Sarah's Personal Style Guide has always followed the "if the bullet point doesn't start a sentence but finishes one off, it shouldn't have a capital letter." But, there's a common style guide that DOES capitalize in this situation. And I know some of our Docs peeps have commented that it looks weird to them to not capitalize.

So I'll allow it because the most important thing is that people aren't distracted. And, it's an easy consistency rule to follow that doesn't depend on whether bullet points are 2nd half of sentences are not. I'll allow it, but I don't like it!

PR #1825, opened 2022-10-20
Link first mention of browser dev tools to external docs

mrienstra:

- New or updated content

#### Description

- Given that this is suppose to be a friendly intro, let's make it easy to figure out how to use browser dev tools, for those new to them.

### DIFF
--- a/src/pages/en/tutorial/2-pages/2.md
+++ b/src/pages/en/tutorial/2-pages/2.md
@@ -67,7 +67,7 @@ Now that you have built pages using `.astro` files, let's make some blog posts u
 
 2. Check your browser preview again at `localhost:3000/posts/post-1`. You should now see content on this page.
 
-3. Use your browser's Dev Tools to inspect this page. Notice that although you have not typed any HTML elements, your Markdown has been converted to HTML. You can see elements such as headings, paragraphs and list items.
+3. Use your [browser's Dev Tools](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_are_browser_developer_tools) to inspect this page. Notice that although you have not typed any HTML elements, your Markdown has been converted to HTML. You can see elements such as headings, paragraphs and list items.
 
 ## Link to your posts
 


delucis (commented):
I’d like @sarah11918 to weigh in here. I know we intentionally avoided adding a lot of inline links as they can pull a reader out of the flow and we want to keep them focused on the tutorial tasks as much as possible. Instead we opted for “Resources” sections at the bottom of pages where helpful.

sarah11918:
Yes, Chris is right. In the tutorial we really do want to keep external links at the bottom of the page, with a `<Badge />` component to indicate that they take you out of the tutorial. We occasionally link within text to a previous lesson (and I'm actually considering moving that one out) but never externally in favour of a link at the bottom of the page.

When people see links, they assume there's info they need to know. Instead, we want to establish the pattern of some resources at the bottom for whatever might have good external resources. This could go nicely at the bottom of the page, though! Can you add it there, following the pattern from another page that has an external resource link at the bottom?

mrienstra:
Thanks @delucis & @sarah11918, good call, makes sense, done!

src/pages/en/tutorial/2-pages/2.md
```diff
@@ -167,4 +167,6 @@ To create content for pages built with `.astro` files, I write text within stand
 
 - [Markdown Cheat Sheet from The Markdown Guide](https://www.markdownguide.org/cheat-sheet/)  <Badge>external</Badge>
 
+- [What are browser developer tools? MDN](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_are_browser_developer_tools)  <Badge>external</Badge> - Learn the basics of using browser dev tools.
```
sarah11918, line 170:
```suggestion
- [What are browser developer tools? MDN](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_are_browser_developer_tools)  <Badge>external</Badge> 
```
Thanks for adding this! The other ones on the page here don't have any description, and the title is pretty self-explanatory, so I'm taking that out.

PR #1828, opened 2022-10-20
[tutorial] removed references to JSX

sarah11918:
Replaced "JSX-like expressions" with "JavaScript expressions" in the two places it showed up.

Also, removed an internal link.


Jutanium:
LGTM!

PR #1831, opened 2022-10-20
[tutorial] less confusing self-quiz instruction

sarah11918:
Instructs the user to "guess" rather than "write" a response. Also gives a hint that the answer might be "nothing."


Jutanium:
LGTM!

PR #1832, opened 2022-10-20
[tutorial] Replace StackBlitz embeds with links

Jutanium:
Added a StackblitzIntro component so we can tweak this later on.


sarah11918:
LGTM @Jutanium

PR #1833, opened 2022-10-20
[tutorial] simplified About This Tutorial page

sarah11918:
Makes the About page less wall of text Q and A

image

### DIFF
--- a/src/pages/en/tutorial/0-introduction/1.md
+++ b/src/pages/en/tutorial/0-introduction/1.md
@@ -7,55 +7,54 @@ setup: |
   import Checklist from '~/components/Checklist.astro';
   import Box from '~/components/tutorial/Box.astro';
 ---
-## What Background knowledge do I need?
+## What do I need to know to get started?
 
-If you have some basic familiarity with **HTML**, **Markdown**, **CSS**, and a little **JavaScript**, then you're totally good to go! But no worries, you'll be able to complete the entire tutorial just by following the instructions.
+If you have some basic familiarity with **HTML**, **Markdown**, **CSS**, and a little **JavaScript**, then you're totally good to go! You'll be able to complete the entire tutorial just by following the instructions. Astro is for everyone! 🧑‍🚀 👩‍🚀 👨‍🚀 
 
-## Is this tutorial for me?
+You'll find answers to more common questions below.
 
-Yes! This tutorial is for everyone!
+<details>
+<summary>How do I use the checklists at the bottom of each page?</summary>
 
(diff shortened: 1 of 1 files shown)


Jutanium:
Love this as collapsed. 
One of our action items is to make the choice between Unit 1 and Unit 2 clear on this page. Instead of the "Unit 1 looks pretty basic" section (they won't know what Unit 1 looks like) I think we should pull out the choice between the two to the introduction. We can do that in a separate PR, though!

src/pages/en/tutorial/0-introduction/1.md
```diff
@@ -7,55 +7,54 @@ setup: |
   import Checklist from '~/components/Checklist.astro';
   import Box from '~/components/tutorial/Box.astro';
 ---
-## What Background knowledge do I need?
+## What do I need to know to get started?
 
-If you have some basic familiarity with **HTML**, **Markdown**, **CSS**, and a little **JavaScript**, then you're totally good to go! But no worries, you'll be able to complete the entire tutorial just by following the instructions.
+If you have some basic familiarity with **HTML**, **Markdown**, **CSS**, and a little **JavaScript**, then you're totally good to go! You'll be able to complete the entire tutorial just by following the instructions. Astro is for everyone! 🧑‍🚀 👩‍🚀 👨‍🚀
```
Jutanium, line 12: Something like
```suggestion
If you have some basic familiarity with **HTML**, **Markdown**, **CSS**, and a little **JavaScript**, then you're totally good to go! You'll be able to complete the entire tutorial just by following the instructions. Astro is for everyone! 🧑‍🚀 👩‍🚀 👨‍🚀 

If you already have a working development environment set up, skip to [**Unit 2**](en/tutorial/2-pages/).
Otherwise, continue to the next section below!
```

sarah11918: It's not just a working dev environment though. It's being on GitHub and Nelify deployed (as we discussed re other things). So, cool to put it up higher, but then it needs to be what all's in that Ch 1 question's answer.

sarah11918: Also, for "quick win" purposes, the nav below will take a reader to Unit 1... we don't have another path yet (infrastructure-wise), so prhasing it ABOUT what Unit 1 is LIKE is in keeping with "the page is sending me to unit 1."  

Even though they haven't SEEN unit 1 yet, so it seems like a premature question, this primes them for "it's gonna seem super basic" and gives them permission to skip ahead manually, at any point in that chapter, or right now! Because our existing nav is not going to direct them there.

That was my thinking, anyway

PR #1836, opened 2022-10-20
Experiment with a more basic tutorial navigation component

delucis:

#### Description

- WIP on improving the tutorial navigation


sarah11918:
I really like this, @delucis! 

Will we put the "on this page" navigation from tablet/mobile mode on it for smaller screen sizes, too? If so, do we *only* use what's on the current page like in Docs, or do we use the entire tutorial nav? Only the other tutorial pages?

I think we'd have to think about what makes the most sense for navigating the tutorial in tablet view, since it's maybe not even so much on the exact page as the OTHER tutorial pages that might need navigating?

delucis:
Going to close this one for now: don‘t think we’ll end up going with exactly this either way, so will create a new branch/PR for any future explorations.

PR #1853, opened 2022-10-21
[tutorial] Add iframe to homepage

Jutanium:
This adds a preview of the completed project to the homepage. (We'd previously used the Stackblitz preview, but its call-to-action button distracted some readers and it doesn't work on all browsers.)


sarah11918:
LET'S GO! @Jutanium

PR #1855, opened 2022-10-21
Ch1 multiple choice rework

sarah11918:
Reworks some Q's in Ch1 to be multiple choice, for better UX.


delucis (commented):
LGTM! Only query would be why you went for “Terminal” over “Netlify” in the rewrite but I’m sure you have your reasons!

We _might_ want something more specific than “application for running commands” as I guess that applies to VS Code too, but happy to merge as-is if you need this shipped.

PR #1857, opened 2022-10-21
Add section on `astro/types`

natemoo-re:

- New or updated content

#### Description

- Paired with https://github.com/withastro/astro/pull/5147
- Updates TypeScript reference with outline of `astro/types` exports


src/pages/en/guides/typescript.md
```diff
@@ -106,19 +106,42 @@ const { greeting = 'Hello', name } = Astro.props;
 <h2>{greeting}, {name}!</h2>
 ```
 
-### Common prop type patterns
+## Type Utilities
```
delucis, line 110: Could we add a `<Since v="" />` component here to flag when these were added in case someone on an older version tries to use these?

src/pages/en/guides/typescript.md
```diff
@@ -106,19 +106,42 @@ const { greeting = 'Hello', name } = Astro.props;
...
+```
+
+### `WithChildren`
+
+If your component must be passed children to its default slot, you can enforce this by using `WithChildren`.
+
+```astro ins="WithChildren"
+---
+import { WithChildren } from 'astro/type-utils'
+type Props = WithChildren
```
delucis, line 132: How should a user extend this with props?

Is it a case of `WithChildren & { prop: string }` or `interface Props extends WithChildren {}`? (Or I guess even `WithChildren<{ prop: string }>` if it’s generic.)

delucis (commented):
Nice QoL improvement — thanks @natemoo-re! Left a couple of queries.

src/pages/en/guides/typescript.md
natemoo-re: I actually was thinking about this utility and decided against it since we might implement the `Component` interface RFC and this wouldn't work anymore.

src/pages/en/guides/typescript.md
natemoo-re: Yep, great idea!

src/pages/en/guides/typescript.md
```diff
@@ -112,13 +114,25 @@ const { greeting = 'Hello', name } = Astro.props;
...
 
-Astro provides JSX type definitions to check that your markup is using valid HTML attributes. You can use these types to help build component props. For example, if you were building a `<Link>` component, you could do the following to mirror the default HTML attributes in your component’s prop types.
+<Since v="1.6.0" />
 
-```astro title="src/components/Link.astro" ins={2}
+Astro comes with some built-in utility types for common prop type patterns. These are available under the `astro/types` entrypoint.
+
+### Built-in HTML attributes
+
+Astro provides the `HTMLAttributes` type to check that your markup is using valid HTML attributes. You can use these types to help build component props. For example, if you were building a `<Link>` component, you could do the following to mirror the default HTML attributes for `<a>` tags in your component’s prop types.
```
sarah11918, line 125:
```suggestion
Astro provides the `HTMLAttributes` type to check that your markup is using valid HTML attributes. You can use these types to help build component props.

For example, if you were building a `<Link>` component, you could do the following to mirror the default HTML attributes for `<a>` tags in your component’s prop types.
```

sarah11918: super-nit: I'd consider breaking up the explanation from the "for example" since they each are about the same length.

Otherwise, LGTM! I'll consider this ready to `merge-on-release`! 🥳

PR #1894, opened 2022-10-23
i18n(de): Add common issues in translation guide

hippotastic:

- Something else!

#### Description

- Adds more common translation issues based on the experience from past PR reviews.


delucis (approved):
LGTM! Thanks for keeping this updated @hippotastic.

PR #1895, opened 2022-10-24
Remove shamefully-hoist

bluwy:

- Something else!

#### Description

- Remove `shamefully-hoist` in `.npmrc`
- Add missing dependencies
- All scripts should work as before
- This removes the need for `--shamefully-hoist` in Astro's smoke test

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -1,7 +1,6 @@
 import { defineConfig } from 'astro/config';
 import preact from '@astrojs/preact';
 import sitemap from '@astrojs/sitemap';
-import vitePreact from '@preact/preset-vite';
 
 import { toString } from 'hast-util-to-string';
 import { h } from 'hastscript';
@@ -16,6 +15,8 @@ import { rehypeTasklistEnhancer } from './plugins/rehype-tasklist-enhancer';
 
 import languages from './src/i18n/languages';
 import { normalizeLangTag } from './src/i18n/bcp-normalize';
+
+const isBuild = process.argv[2] === 'build';
 
 const AnchorLinkIcon = h(
 	'svg',
@@ -60,7 +61,10 @@ export default defineConfig({
 		astroCodeSnippets(),
 	],
 	vite: {
-		plugins: [vitePreact()],
+		ssr: {
+			// Externalize deps of `astro-og-canvas`
+			external: isBuild ? [] : ['canvaskit-wasm', 'entities'],
+		},
 	},
 	markdown: {
 		syntaxHighlight: 'shiki',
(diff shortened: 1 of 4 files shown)


astro.config.ts
```diff
@@ -60,7 +61,10 @@ export default defineConfig({
 		astroCodeSnippets(),
 	],
 	vite: {
-		plugins: [vitePreact()],
+		ssr: {
+			// Externalize deps of `astro-og-canvas`
+			external: isBuild ? [] : ['canvaskit-wasm', 'entities'],
+		},
```
bluwy, line 67: Currently Astro's package crawler only `noExternal` third-party Astro packages, we'd also need to `external` their sub dependencies, and in dev only speficially to avoid `entry.mjs` build errors.

This workarounds the issue for now, but we can implement the feature upstream soon.

astro.config.ts
```diff
@@ -60,7 +61,10 @@ export default defineConfig({
 		astroCodeSnippets(),
 	],
 	vite: {
-		plugins: [vitePreact()],
```
bluwy, line 63: I'm actually not sure why `@preact/preset-vite` is added here, but `pnpm dev` and HMR seems to still work well after removing this.

delucis: For some reason I needed this to get `preact/compat` working when we first migrated to it (the search component we depend on is a React component), looks like the search is still working after removing so I guess we’ve fixed whatever this was patching back then 🥳

delucis (commented):
Thanks @bluwy! Did some quick manual tests of the deploy preview and everything seems to be working as expected still. Same in a local pull of this PR, so should be good to go although may be good to clear up what we should do about `astro-og-canvas` if possible.

astro.config.ts
bluwy: It would need to be fixed in Astro's [dependency walker](https://github.com/withastro/astro/blob/bf5ad72e81da36c68f96c2431ec547e356538b80/packages/astro/src/core/create-vite.ts#L191) to auto external the packages sub-dependencies too. I'm planning to unify this code so it can be shared between Astro, vite-plugin-svelte and vite-plugin-solid this week, so everything should work OOTB then!

astro.config.ts
```diff
@@ -60,7 +61,10 @@ export default defineConfig({
...
+		ssr: {
+			// Externalize deps of `astro-og-canvas`
+			external: isBuild ? [] : ['canvaskit-wasm', 'entities'],
```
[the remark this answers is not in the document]
bluwy: That is definitely better than what I had 😄 Will switch to this.

delucis (approved):
LGTM! Thanks again @bluwy

Issue #1892, opened 2022-10-23
Images: `format` prop is required for remote images


Jutanium:
Following up on [this support thread](https://discord.com/channels/830184174198718474/1032660889318461482), the `format` attribute is _required_ for remote images. Let's note this when we [introduce remote images](https://discord.com/channels/830184174198718474/1032660889318461482).

sarah11918:
Closing this, as we just updated the images guide and the integration README. It's now listed as required for `<Image />` and explains which default is used for `<Picture />` (so, it's not required there).

PR #1897, opened 2022-10-24
Images guide fixes

Jutanium:
- New or updated content

#### Description

- Moves alt text from an image integration caution into its own section
- Addresses #1892 from the docs side (we still need to update the integration documentation)
- Fixes the code samples so that they're copy-pastable and include alt text

### DIFF
--- a/src/pages/en/guides/images.md
+++ b/src/pages/en/guides/images.md
@@ -114,15 +114,6 @@ Astro's official image integration provides two different Astro components for r
 
 After [installing the integration](/en/guides/integrations-guide/image/#installation), you can import and use these two components wherever you can use Astro components, including `.mdx` files!
 
-:::note
-Astro's `<Image />` and `<Picture />` components cannot be used with images in your `public/` folder. Use standard HTML or Markdown image syntax instead.
-:::
-
-:::caution
-Astro's `<Image />` and `<Picture />` components require the `alt` attribute which provides descriptive text for images. These components will throw an error if no `alt` text is provided.
-
-If the image is merely decorative (i.e. doesn't contribute to the understanding of the page), set `alt=""` so that the image is properly understood and ignored by screen readers.
-:::
 ### `<Image />`
 
 Astro's [`<Image />` component](/en/guides/integrations-guide/image/#image-) allows you to optimize a single image and specify width, height, and/or aspect ratio. You can even transform your image to a particular output format, which can be used to avoid checking the file type of remote images. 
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/images.md
```diff
@@ -135,35 +126,39 @@ Image files in your project's source directory can be imported in frontmatter an
 
 #### Remote Images
 
-Remote images require a full URL as the image `src`. Also, you must either provide `width` and `height`, or one of the dimensions plus the required `aspectRatio` to the `<Image />` component.
+To use a remote image, pass a full URL to the `<Image />`'s src attribute. `<Image />` won't infer dimensions and format from this remote file. You must provide the `format` to render the image with, and you must either provide `width` and `height` or one of the two dimensions plus an `aspectRatio`.
```
sarah11918, line 129:
```suggestion
To use a remote image, pass a full URL to the `<Image />`'s src attribute. `<Image />` won't infer dimensions and format from this remote file. You must provide the `format` to render the image with, and you must either provide both `width` and `height` or one of the two dimensions plus an `aspectRatio`. The `alt` attribute is also required.
```
Maybe the "both" is helpful for a longer sentence explanation? (I don't think we have to bullet point the two options, but we could. An explicit "both" could be a good compromise.)

Also, it feels like a good place to remind re: `alt` while we're explicitly mentioning the attributes that are required.

Jutanium: If we mention `alt` here, we'd need to reference it in all of the other sections (local/remote for Image/Picture), otherwise it will seem like alt is only required here. I'd rather state it in the intro section if we want a reminder

src/pages/en/guides/images.md
```diff
@@ -135,35 +126,42 @@ Image files in your project's source directory can be imported in frontmatter an
...
+<!-- cropping to a specific width and height -->
+<Image src={localImage} width={300} height={600} alt={localAlt}/>
+<Image src={remoteImage} width={544} height={184} format="png" alt={remoteAlt}/>
 
-// cropping to a specific aspect ratio and converting to an avif format
-<Image src={localImage} aspectRatio="16:9" format="avif" />
-<Image src={imageUrl} height={200} aspectRatio="16:9" format="avif" />
+<!-- cropping to a specific aspect ratio and converting to an avif format-->
+<Image src={localImage} aspectRatio="16:9" format="avif" alt={localAlt}/>
+<Image src={remoteImage} height={200} aspectRatio="16:9" format="avif" alt={localAlt}/>
```
sarah11918, line 158:
```suggestion
<Image src={remoteImage} height={200} aspectRatio="16:9" format="avif" alt={remoteAlt}/>
```

src/pages/en/guides/images.md
```diff
@@ -135,35 +126,42 @@ Image files in your project's source directory can be imported in frontmatter an
...
 
-// local image imports can also be inlined directly
-<Image src={import('../assets/local.png')} />
+<!-- local image imports can also be inlined directly-->
+<Image src={import('../assets/logo.png')} alt={localAlt}/>
+
+<!-- If an image is stored in the `/public` folder, use its path relative to `/public`-->
+<Image src="penguin.jpg" width="300" aspectRatio={1} format="png" alt="A happy penguin"/>
```
sarah11918, line 164:
```suggestion
<Image src="/penguin.jpg" width="300" aspectRatio={1} format="png" alt="A happy penguin"/>
```
Do we need this?

Jutanium: It works either way. Let's add it in to match the code sample in an earlier section

sarah11918: Interesting! If it works either way, then I guess we decide how we want to handle it. There's something to be said for NOT using the / if it's not required, to distinguish it from something in `src`.

src/pages/en/guides/images.md
Jutanium: I decided to take your approach because the other sections call out required attributes!

sarah11918:
LGTM, @Jutanium!

PR #1906, opened 2022-10-24
Update `astro preview` docs after 1.5

Jutanium:
- New or updated content

As of 1.5, `astro preview` doesn't just work in static mode. This PR removes the "static" qualifier from our description of `astro preview`, and updates the caution to explain when it works in SSR.

### DIFF
--- a/src/pages/en/reference/cli-reference.md
+++ b/src/pages/en/reference/cli-reference.md
@@ -92,14 +92,14 @@ Includes [Markdown draft pages](/en/guides/markdown-content/#markdown-drafts) in
 
 ## `astro preview`
 
-Starts a local file server to serve your static `dist/` directory.
+Starts a local server to serve your `dist/` directory.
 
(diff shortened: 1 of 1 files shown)


delucis (commented):
Looks good! Do you think we need to flag that this is since 1.5.0 somewhere here?

Jutanium:
> Looks good! Do you think we need to flag that this is since 1.5.0 somewhere here?
I could go back to the original "static" phrasing of the first description, and rephrase the caution:

Since Astro 1.5.0, `astro preview` also works for SSR builds if you use an adapter that supports it. Currently, only the [Node adapter](/en/guides/integrations-guide/node/) supports `astro preview`.

delucis:
That reads well to me! 👍

sarah11918:
LGTM!

sarah11918:
I think the comment looks good. Doesn't need to be in a caution now, right?

delucis (approved):
Thanks Dan!

PR #1908, opened 2022-10-25
Tutorial: better multiple choice styles & remove content

delucis:
#### What kind of changes does this PR include?

- Minor content fixes (broken links, typos, etc.)
- Changes to the docs site code

#### Description

- Fix indentation on a few last `<MultipleChoice>` instances that weren’t nested under their numbered question.

- Remove the “What’s all the interactive quizzes?” `<details>` in 0.1 as discussed.

- Flip the `box-shadow` direction on the submit button in `<MultipleChoice>` so it matches the checklist buttons (and the `<Box>`) direction.

- Update the letters in the `<MultipleChoice>` options so that they look more like the checklist buttons (including bounce! 🏀💨)
    image

PR #1914, opened 2022-10-25
More `<MultipleChoice>` polish

delucis:

#### Description

- I forgot to make the `<MultipleChoice>` submit button “pop” in the right direction as part of #1908 — this PR fixes that.
- I’ve also switched up the transition style so it uses the same “bounce” easing as our other buttons for consistency.
- These transitions were active even for users who prefer reduced motion, so I’ve fixed that, disabling animations when appropriate.


sarah11918:
Looks bouncy to me!  @delucis

PR #1916, opened 2022-10-25
Image guide updates

sarah11918:
- New or updated content

This starts to incorporate some of the Image Integration's README content, and correct and clarify some changes that have occured since writing. 

- removes warnings/notes that are no longer necessary; de-prioritizes ones that don't have to be UP FRONT, IN YOUR FACE
- address not having any way to specify defaults, offers a component-wrapping solution (INVOLVES TAILWIND, LET'S CHANGE THAT!)
- clarifies that images in public ARE ACTUALLY able to be optimized by the integration, even though generally this does not happen
- moves Alt Text section lower, to come after all the "types" of images, since it relates to all of them
- general clarity/edits


src/pages/en/guides/images.md
```diff
@@ -104,29 +114,35 @@ The [`public/` directory](/en/core-concepts/project-structure/#public) is for fi
...
 ## Astro's Image Integration
 
-:::caution
-When you install the `@astrojs/image` integration, `.astro` files will no longer be able to use standard HTML `<img>` tags for images located in your project `src`. All local images must use the integration components instead.
-
-See the [image integration guide](/en/guides/integrations-guide/image/) for more notes about this new, experimental feature!
-:::
-
-Astro's official image integration provides two different Astro components for rendering optimized images: `<Image />` and `<Picture />`.
+Astro's official image integration provides two different Astro components for rendering optimized images: `<Image />` and `<Picture />` as well as a basic image transformer, with full support for static sites and server-side rendering.
```
sarah11918, line 117:
```suggestion
Astro's official image integration provides two different Astro components for rendering optimized images, `<Image />` and `<Picture />`, as well as a basic image transformer, with full support for static sites and server-side rendering.
```

src/pages/en/guides/images.md
```diff
@@ -97,24 +97,23 @@ import logo from '../images/logo.png';
 
 ### `public/`
 
-Your images stored in `public/` can be used by components (`.astro`, `.mdx` and other UI frameworks) and also Markdown files (`.md`, `.markdown`).
+Your images stored in `public/` can be used by components (`.astro`, `.mdx` and other UI frameworks) and also Markdown files.
 
-The `src` attribute is **relative to the public folder**. These images can also be used in Markdown files using the `![]()` notation.
 
-Files in the `/public` directory are always served or copied as-is, with no processing. We recommend that local images are kept in `src/` when possible so that Astro can transform, optimize and bundle them.
+However, files in the `/public` directory are always served or copied as-is, with no processing. We recommend that local images are kept in `src/` when possible so that Astro can transform, optimize and bundle them.
```
delucis, line 103: Doing this means you _can’t_ use the images in Markdown right? Do we want to call that out? If you have a Markdown-powered blog, we specifically would _not_ recommend images in `src/`, I guess.

```suggestion
However, files in the `/public` directory are always served or copied as-is, with no processing. If you are using images outside of Markdown files, we recommend that local images are kept in `src/` when possible so that Astro can transform, optimize and bundle them.
```

src/pages/en/guides/images.md
```diff
@@ -142,7 +141,9 @@ You must also either provide `width` and `height`, or one of the dimensions plus
 
 #### Local Images in `public/`
 
-The `<Image />` component can also be used with images stored in the `public/` directory and the `src` attribute is relative to the public folder. It will be treated as a remote image, which requires either both `width` and `height`, or one dimension and an `aspectRatio` attribute.
+The `<Image />` component can also be used with images stored in the `public/` directory and the `src` attribute is relative to the public folder. It will be treated as a remote image, which requires either both `width` and `height`, or one dimension and an `aspectRatio` attribute.
```
delucis, line 144: Ooh, this is weird! 😂 TIL.

src/pages/en/guides/images.md
```diff
@@ -142,7 +141,9 @@ You must also either provide `width` and `height`, or one of the dimensions plus
 
 #### Local Images in `public/`
 
-The `<Image />` component can also be used with images stored in the `public/` directory and the `src` attribute is relative to the public folder. It will be treated as a remote image, which requires either both `width` and `height`, or one dimension and an `aspectRatio` attribute.
+The `<Image />` component can also be used with images stored in the `public/` directory and the `src` attribute is relative to the public folder. It will be treated as a remote image, which requires either both `width` and `height`, or one dimension and an `aspectRatio` attribute. 
+
+Note that your original image will be copied unprocessed to the build folder, like all files located in `public/`. However, Astro's image integration is still able to return optimized versions of your images stored in `public/`.
```
delucis, line 146: Style nit + is it copied BUT or copied AND?

```suggestion
Your original image will be copied unprocessed to the build folder, like all files located in `public/`, and Astro's image integration will also return optimized versions of the image.
```

src/pages/en/guides/images.md
```diff
@@ -243,21 +244,43 @@ export const galaxy = 'https://astro.build/assets/galaxy.jpg';
...
+---
+import { Picture } from '@astrojs/image/components';
 
+const {src, ...attrs} = Astro.props;
+---
+<Picture src={src} widths={[400, 800, 1500]} sizes="(max-width: 767px) 100vw, 736px" {...attrs} />
+
+<style>
+  img, picture :global(img), svg {
+    @apply rounded-xl shadow-lg my-10;
```
delucis, line 263: Torn on using Tailwind syntax here. Do we hate a slightly longer, vanilla CSS version?

```suggestion
    margin-block: 2.5rem;
    border-radius: 0.75rem;
```

src/pages/en/guides/images.md
```diff
@@ -243,21 +244,43 @@ export const galaxy = 'https://astro.build/assets/galaxy.jpg';
...
+
+These are treated as remote images by the `<Image />` and `<Picture />` components, so they will require you to specify dimension and format values.
 
 Alternatively, if the CDN provides a Node.js SDK, you can use that in your project. For example, [Cloudinary’s SDK](https://cloudinary.com/documentation/node_integration) can generate the `<img>` tag with the appropriate `src` for you.
 
+## Alt Text
+
+Not all users can see images in the same way, so accessibility is an especially important concern when using images. Use the `alt` attribute to provide [descriptive alt text](https://www.w3.org/WAI/tutorials/images/) for images. 
+
+This attribute is required for the Image integration's `<Image />` and `<Picture />` components. These components will throw an error if no alt text is provided.
```
delucis, line 280: Elsewhere we say “image integration” all lowercase — should this be the same?

```suggestion
This attribute is required for the image integration's `<Image />` and `<Picture />` components. These components will throw an error if no alt text is provided. 
```

src/pages/en/guides/images.md
```diff
@@ -243,21 +244,43 @@ export const galaxy = 'https://astro.build/assets/galaxy.jpg';
...
 
 Alternatively, if the CDN provides a Node.js SDK, you can use that in your project. For example, [Cloudinary’s SDK](https://cloudinary.com/documentation/node_integration) can generate the `<img>` tag with the appropriate `src` for you.
 
+## Alt Text
+
+Not all users can see images in the same way, so accessibility is an especially important concern when using images. Use the `alt` attribute to provide [descriptive alt text](https://www.w3.org/WAI/tutorials/images/) for images. 
+
+This attribute is required for the Image integration's `<Image />` and `<Picture />` components. These components will throw an error if no alt text is provided. 
+
+If the image is merely decorative (i.e. doesn’t contribute to the understanding of the page), set `alt=""` so that the image is properly understood and ignored by screen readers.
```
delucis, line 282: I think we can simplify here maybe?

```suggestion
If the image is merely decorative (i.e. doesn’t contribute to the understanding of the page), set `alt=""` so that screen readers know to ignore the image.
```

src/pages/en/guides/images.md
```diff
@@ -10,11 +10,12 @@ Astro provides several ways for you to use images on your site, whether they are
 
 ### In `.astro` files
 
-Astro uses standard HTML `<img>` or `<img />` elements to display images within your `.astro` files. All HTML image attributes are supported.
+Astro uses standard HTML `<img>` or `<img />` elements to display images, and all HTML image attributes are supported.
```
delucis, line 13: I read “`` or ``” as a typo at first — like trying to work out what the second was supposed to be instead. Do you think we can get away without showing the self-closing version?

```suggestion
Astro uses standard HTML `<img>` elements to display images, and all HTML image attributes are supported. 
```

src/pages/en/guides/images.md
```diff
@@ -27,10 +28,11 @@ import rocket from '../images/rocket.svg';
 <img src={rocket} alt="A rocketship in space."/>
 ```
 
-### In `.md` files
+### In Markdown files
 
 You can use standard Markdown `![]()` syntax or standard HTML `<img>` tags in your `.md` files for images located in your `public/` folder, or remote images on another server.
 
+If you cannot keep your images in `public/`, we recommend using the Astro MDX integration to allow you to import and include components in Markdown.
```
delucis, line 35: My suggestion here could probably do with some further massaging, but it might be worth being specific that the integration doesn’t actually let you do extra stuff **IN** Markdown — you still need to switch to another format.

```suggestion
If you cannot keep your images in `public/`, we recommend using the `.mdx` file format, which allows you to combine imported components with Markdown-like syntax. Use the [MDX integration](/en/guides/integrations-guide/mdx/) to add support for MDX to Astro.
```

(I could also see an argument for moving this line _after_ the code sample? But I do like the consistent pattern of: heading, text, code sample.)

src/pages/en/guides/images.md
delucis: Or — if we’re really concerned that `` doesn’t work in MDX — we could write:

```suggestion
Astro uses standard HTML `<img />` elements to display images, and all HTML image attributes are supported. 
```

Although I think this might be technically inaccurate in that the “standard” HTML element is specifically `` without the closing slash. 🤷

delucis (commented):
Amazing work figuring this all out and putting this together @sarah11918 — really appreciate it!

src/pages/en/guides/images.md
sarah11918: I like what you wrote better, but there's a difference between SSG and SSR  😅 , so I just want to make sure this still is accurate.

In SSR, only the original image is copied over, but the fetch call returns an optimized image (without creating a second image in your build folder). In SSG, you end up with TWO copies of your image copied into the build folder: the original, and an optimized one.

I think that's why I said "is able to return" which admittedly sounds clunky. I think "will also return" is legit here, eh?

src/pages/en/guides/images.md
sarah11918: Oh not torn at all! I explicitly wrote in my comment at the beginning, PLZ UN-TAILWIND the example!  😅

src/pages/en/guides/images.md
sarah11918: I was wondering if the "when possible" was strong enough on its own, since the corresponding section on in `src` says that you can NOT use Markdown with your images there.

src/pages/en/guides/images.md
sarah11918: This is not the weirdest thing about our image integration... 😅

src/pages/en/guides/images.md
sarah11918: I think we only really need to worry about what works in Astro (`.astro`) files here in this section. And, I think both work? Do you think it's fine just to say "HTML ``" (we drop the "standard" and the self-closing one)?  Wait until someone complains?

src/pages/en/guides/images.md
delucis: “We recommend” is pretty strong. Do we think a reader has a clear picture of what “when possible” means? I’d be worried they miss the `src` comment — unless it’s perhaps really highlighted in a caution? But I like the goal here of getting rid of those!

src/pages/en/guides/images.md
delucis: Yeah, I think “will also return” is ok! There’s nothing in here that really explains any of that SSR/SSG distinction and “return” is sufficiently vague as to not mean “copy to `dist`”, I think, so would apply in all scenarios.

src/pages/en/guides/images.md
```diff
@@ -91,42 +97,53 @@ import logo from '../images/logo.png';
...
-
-After [installing the integration](/en/guides/integrations-guide/image/#installation), you can import and use these two components wherever you can use Astro components, including `.mdx` files!
+After [installing the integration](/en/guides/integrations-guide/image/#installation), you can import and use these two components wherever you can use Astro components: `.astro` and `.mdx` files.
 
 ### `<Image />`
 
 Astro's [`<Image />` component](/en/guides/integrations-guide/image/#image-) allows you to optimize a single image and specify width, height, and/or aspect ratio. You can even transform your image to a particular output format.
 
-This component is useful for images where you want to keep a consistent size across displays, or closely control the quality of an image (e.g. logos).
+This component is useful for images where you want to keep a consistent size across displays, or closely control the quality of an image (e.g. logos).
```
sarah11918, line 124:
```suggestion
This component is useful for images where you want to keep a consistent size across displays, or closely control the quality of an image (e.g. logos).
```

src/pages/en/guides/images.md
delucis: “HTML ``” works! And both work in `.astro` yup.

sarah11918, line 13:
```suggestion
Astro uses HTML `<img>` elements to display images, and all HTML image attributes are supported. 
```

src/pages/en/guides/images.md
```diff
@@ -91,42 +97,53 @@ import logo from '../images/logo.png';
...
 ```
 
 ## Astro's Image Integration
 
-:::caution
-When you install the `@astrojs/image` integration, `.astro` files will no longer be able to use standard HTML `<img>` tags for images located in your project `src`. All local images must use the integration components instead.
-
-See the [image integration guide](/en/guides/integrations-guide/image/) for more notes about this new, experimental feature!
-:::
+Astro's official image integration provides two different Astro components for rendering optimized images, `<Image />` and `<Picture />`, as well as a basic image transformer. There is full support for both static sites and server-side rendering.
```
Jutanium, line 116: Because Image and Picture sound like synonyms, I like differentiating them. But I'm not sure what I wrote here is any good.
If we don't document `getImage`, it might be confusing to mention here
```suggestion
Astro's official image integration provides two different Astro components for rendering optimized images, `<Image />` and `<Picture />`. `<Image/>` lets you optimize a single image, and `<Picture/>` lets you send a set of images for the most optimal one be chosen in the browser. There is full support for both static sites and server-side rendering.
```

sarah11918: I agree with ditching the transformer, but I was keeping this simple here, and under each one I do say what they are for, and for Image, when they might prefer to use Picture. I *think* that's enough, and doesn't front-load the info?

sarah11918, line 116:
```suggestion
Astro's official image integration provides two different Astro components for rendering optimized images, `<Image />` and `<Picture />`. There is full support for both static sites and server-side rendering.
```

Jutanium: My thought was that we'd want to explain the difference upfront so they can understand why there are two without reading both sections - but I don't think it's totally necessary!

PR #1886, opened 2022-10-22
ci: update integration docs

fredkbot:
This PR is auto-generated by a nightly GitHub action to update the individual integration pages from the integration READMEs in withastro/astro.

PR #1919, opened 2022-10-25
`docgen:integrations`: don't strip hashes from links

mrienstra:

- Changes to the docs site code (specifically to `scripts/generate-integration-pages.ts`, which is called by `pnpm run docgen:integrations`)

#### Description

- [I noticed](https://github.com/withastro/docs/pull/1886#issuecomment-1290997156) that a link hash was disappearing somewhere between `withastro/astro` and `withastro`docs`:
    ```diff
    - https://docs.astro.build/en/reference/configuration-reference/#markdownextenddefaultplugins
    +                         /en/reference/configuration-reference/
    ```
- Tracked down the source of the issue, this PR fixes it (confirmed locally).
- Also made a very minor change to an unrelated regex in the same script, to reduce the chance of it erroneously matching in the future.
- Tempting to figure out how to test this script, but I gather it will be retired sometime in the next few months.

### DIFF
--- a/scripts/generate-integration-pages.ts
+++ b/scripts/generate-integration-pages.ts
@@ -92,7 +92,7 @@ class IntegrationPagesBuilder {
 	 */
 	async #processReadme({ name, readme, srcdir, category }: IntegrationData): Promise<string> {
 		// Remove title from body
-		readme = readme.replace(/# (.+)/, '');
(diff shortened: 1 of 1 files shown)


delucis (approved):
Nice catch @mrienstra! I hadn’t realised these fragment links were getting stripped out — thanks for fixing.

PR #1920, opened 2022-10-25
Initial work overhauling styles for the tutorial project

delucis:
#### What kind of changes does this PR include?

- New or updated content

#### Description

- Updates the styling used in the tutorial project up to and including lesson 3.2
- Main aim is to make things look a little more minimal while still preserving the same basic learning flow
- I have tweaked a few bits of markup too where it seemed we could do better: use `<footer>` in the footer component for example.

I have been updating the example repo as I go, so you can get a glimpse by looking at the `unit-4/start` branch on StackBlitz: https://stackblitz.com/github/withastro/blog-tutorial-demo/tree/unit-4/start (N.B. this does still include the hamburger/nav styles, which I haven’t tackled yet)

Here’s a taste:

image

PR #1921, opened 2022-10-26
typo in tutorials

Rackar:

#### Description

- Closes # 
    typo

### DIFF
--- a/src/pages/en/tutorial/2-pages/5.md
+++ b/src/pages/en/tutorial/2-pages/5.md
@@ -45,7 +45,6 @@ There are a few ways to define styles **globally** in Astro, but in this tutoria
     h1 {
       margin: 0;
       padding: 1.25rem 0;
-      color: blue;
     }
     ```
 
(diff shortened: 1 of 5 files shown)


src/pages/en/tutorial/2-pages/5.md
```diff
@@ -45,6 +45,7 @@ There are a few ways to define styles **globally** in Astro, but in this tutoria
     h1 {
       margin: 0;
       padding: 1.25rem 0;
+      color: blue;
```
delucis, line 48: Ah, I removed this intentionally in #1920 but am now seeing the mention of “blue” later in the page:

> Are there any conflicting styles, and if so, which are applied?
>
> Yes, `h1` is defined as blue globally, but purple locally in the `<style>` tag. The purple color is applied.

Let’s remove this fix here and I’ll have a think about how to make sure we still have a some conflicting styles we can ask readers about.

```suggestion
```

delucis (commented):
Thank you for spotting these details — super helpful!

One is going to need a little more work, so I’ve suggested removing the fix here and I’ll tackle that as I work on updating some of these pages today.

delucis (approved):
Thanks again @Rackar! And welcome as a contributor to Astro’s docs 🎉

PR #1922, opened 2022-10-26
docs(en): mention playwright in testing guide

mikelgo:
As the example is describing a test with Playwright, I guess it would be nice to mention it also in the testing-libs at the beginning.


delucis (approved):
Makes sense!

PR #1925, opened 2022-10-26
Add more missing dependencies

bluwy:

- Something else!

#### Description

I missed some dependencies that should be installed (used in the local integrations).

I also moved the deps specified in `.npmrc` into `package.json` so Astro's smoke test should now pass. Should unblock https://github.com/withastro/astro/pull/5193

Added the notes for these special dependencies in `MAINTAINERS.md`

Notes:

It might be good in the future to split out local integrations and markdown plugins as separate packages monorepo-style to keep the root `package.json` clean.


delucis (commented):
Thanks @bluwy! Will we be able to remove the indirect dependencies once your work on the improved dependency walker is done? Feels odd to need to include them like this, but happy to unblock your work on the `astro` repo!

bluwy:
The `github-slugger` might have to wait till we move off Astro-flavored markdown as I can't find a quick way to prevent it (and it's also deprecated).

For `canvaskit-wasm`, it might be a bit hard since it's not bundle-friendly (uses `__dirname`), so it shouldn't be bundled. But because we need to bundle `astro-og-canvas`, it creates a problem where the final bundle would try to import `canvaskit-wasm` directly (not possible in strict dep installation). So the workaround is needed.

My work on the dependency walker (coming soon™️) wouldn't help with this case though. It'll only help removing the custom `viteSsrFix` integration we added in https://github.com/withastro/docs/pull/1895

delucis:
Oh man, I hadn’t actually looked at the glue code in `canvaskit-wasm` — ironically I picked it with the idea that wasm might be more friendly to different environments. I guess you _could_ technically use the compiled `.wasm` directly, but not exactly sure.

Anyway thanks for the summary! Feel free to merge this when you need to 🙌

bluwy:
In some Vite discussion we also toyed of the idea of resolving the `canvaskit-wasm` to a nested path so it works with `pnpm` too even if it's externalized. I feel that it's a bit dangerous but it's definitely another potential path.

Merging this one for now. Thanks for the review!

PR #1928, opened 2022-10-26
remove pre-process line

### LINKED ISSUE #1918: Incorrect/confusing documentation for processing of is:inline
Hey all! The documentation for template directives says the following:

> is:inline tells Astro to leave the <script> or <style> tag as-is in the final output HTML. _The contents will not be processed_, optimized, or bundled. _This limits some Astro features, like_ importing an npm package or _using a compile-to-CSS language like Sass_. (emphasis added)

https://github.com/withastro/docs/blame/main/src/pages/en/reference/directives-reference.md#L182

Based on my reading, that means that  a `<style lang="scss" is:inline>` will not be converted to CSS on build. However, later we read this:

> _Will be pre-processed_, for example a `<style lang="sass">` attribute _will still generate plain CSS_. (emphasis added)

https://github.com/withastro/docs/blame/main/src/pages/en/reference/directives-reference.md#L189

This might be a simple difference in definition for process _v. pre-process, but as is I believe the documentation is confusion.

For the record, `<style lang="scss" is:inline>` **will not** output CSS.

Jutanium:

- New or updated content

#### Description

- Closes #1918 

Removes misleading line about SCSS in `is:inline` styles. Waiting for confirmation that there isn't other "pre-processing" happening.

### DIFF
--- a/src/pages/en/reference/directives-reference.md
+++ b/src/pages/en/reference/directives-reference.md
@@ -186,7 +186,6 @@ The `is:inline` directive means that `<style>` and `<script>` tags:
 - Will not be bundled into an external file. This means that [attributes like `defer`](https://javascript.info/script-async-defer) which control the loading of an external file will have no effect.
 - Will not be deduplicated—the element will appear as many times as it is rendered.
 - Will not have its `import`/`@import`/`url()` references resolved relative to the `.astro` file.
-- Will be pre-processed, for example a `<style lang="sass">` attribute will still generate plain CSS.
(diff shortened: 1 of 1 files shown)


Jutanium:
We [got our confirmation](https://discord.com/channels/830184174198718474/1034844447403425812/1034850676683055234):

> `is:inline` means not processed by Astro at all
that's true when used on a style or script or slot
it means "leave this alone"
so no, sass will not work on it

delucis (approved):
LGTM! Thanks for figuring this out Dan 💜

PR #1931, opened 2022-10-26
Tutorial: Ensure style conflict for a teachable moment!

delucis:

- New or updated content

#### Description

#1921 reported we were still talking about `<h1>`s being blue in the tutorial and I’d missed this “Analyze the pattern” where we show that a local `<style>` tag on the About page overrides `global.css` in #1920. This PR patches that up by setting an extra big font-size on the about page and pointing out to the reader that _that_ overrides the smaller font-size in the global styles.

I have the equivalent changes to https://github.com/withastro/blog-tutorial-demo cued up ready to go.


sarah11918:
"Ensures conflict for a teachable moment" is my new band name. LGTM!

PR #1940, opened 2022-10-27
links to full list of Image/Picture component properties

sarah11918:
- New or updated content

Wanted to make the links back to the full list of properties that the `<Image />` and `<Picture />` components can take clearer. Current guide focuses on the minimum required attribures for basic usage, but it should be easy for a reader to get to the full list of attributes available.


delucis (approved):
LGTM!

PR #1958, opened 2022-10-31
Remove Vite SSR fix

bluwy:

- Something else!

#### Description

This removes a configuration fix added in #1895, and upgrades to Astro 1.6.0 which automates the fix by default.


delucis (approved):
Looks great! Thanks @bluwy.

PR #1960, opened 2022-10-31
Update .stackblitzrc

ggdaltoso:
As we are Astro fans at StackBlitz, we noticed that the `.stackblitzrc` still calls for `npm`. So, I'm updating it!

#### Description

And it now works in Web Publisher!
https://stackblitz.com/~/github.com/ggdaltoso/docs/edit/ggdaltoso/fix-stackblitz-startCommand/src/pages/en/concepts/mpa-vs-spa.md

#### Screenshot

image

### DIFF
--- a/.stackblitzrc
+++ b/.stackblitzrc
@@ -1,5 +1,5 @@
 {
-  "startCommand": "npm start",
+  "startCommand": "pnpm start",
   "env": {
     "ENABLE_CJS_IMPORTS": true
   }


Jutanium:
Thank you for this, we're Stackblitz fans at Astro!

One question: our starter templates ([like this one](https://github.com/withastro/astro/tree/main/examples/blog)) don't have `.stackblitzrc` files at all. Is this something they'll need to work properly?

delucis (approved):
Makes sense seeing as we’re using `pnpm` in this repo! Thanks @ggdaltoso.

ggdaltoso:
> One question: our starter templates ([like this one](https://github.com/withastro/astro/tree/main/examples/blog)) don't have `.stackblitzrc` files at all. Is this something they'll need to work properly?

It isn't strict necessarily to have a `.stackblitzrz` file. 
We'll assume which runner we should use based on the `lock` file. If there is no `lock` file, we'll default to `npm`. In addition to that, the lock file will speed up the installation phase considerably, so it is always good to ship with a lock file if it is possible.

PR #1961, opened 2022-10-31
Note static endpoint can return a response object

bluwy:

- New or updated content

#### Description

Based on https://github.com/withastro/astro/pull/5258, clarify that you can return a response object from a static endpoint as long as it's a successful response.

### DIFF
--- a/src/pages/en/core-concepts/endpoints.md
+++ b/src/pages/en/core-concepts/endpoints.md
@@ -39,7 +39,20 @@ export async function get({ params, request }) {
 }
 ```
 
-You can also type your endpoint functions using the `APIRoute` type:
+You can also return a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object as long as it's a [successful response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses) (e.g. status code 200).
+
+```ts title="src/pages/name.json.ts"
+export async function get({ params, request }) {
+  return new Response(
+    JSON.stringify({ name: "Astro" }),
+    {
+      status: 200
+    }
+  )
+}
+```
+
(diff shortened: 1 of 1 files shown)


delucis (commented):
I’m wondering if we want to document this? I understand supporting it for consistency, but if I understand correctly you can’t use any of the features of `Response` really, right? Headers etc. aren’t supported. Are there any circumstances where returning a `Response` would be preferable to `{ body: '...' }`?

delucis:
I’d like another opinion from @Jutanium or @sarah11918 before I stomp on your PR! 😅 

But I definitely think it’s OK to merge the Astro feature without this PR merged yet and we can take a bit more time to decide how best to document this.

Jutanium:
I discovered when first working on this guide that it worked in dev and not build, so thanks @bluwy for fixing that! In terms of docs, I like first introducing the a full-fledged `Response` in the SSR section, where it fully works. We don't want to imply that `statusText` or `headers` work in SSG mode, and introducing it there with those caveats begs the question of why you'd use it at all. 

The flip side is that someone might _try_ to use it in SSG mode, notice that it partially works, and then get confused when most of it gets ignored. In the spirit of "don't include code samples that we don't want copied", we could replace the code sample with a note or caution:


> :::caution
> You can return a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object, but it must return a [successful](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses) `status` code. The response body will be built as the endpoint, but other properties (like `header` and `statusText`) will be ignored.
> :::

I'm also happy to not include it at all.

bluwy:
Thanks for the comments @Jutanium. I think it sounds better to not document this until we have better support for it. I agree with all the points you mentioned! Closing this one for now.

PR #1965, opened 2022-10-31
Update `Astro.glob` docs: `url` property

fuzzykiller:
`url` property is not available on `frontmatter` in this context.

- New or updated content

#### Description

The current documentation about the result of `Astro.glob()` is incorrect. The `.frontmatter` property does _not_ contain a `.url` property. Instead, this property is _next to_ `.frontmatter`.

Only _inside_ the page itself is `.url` available on `Astro.props.frontmatter`.

The relevant change in Astro itself is here:

https://github.com/withastro/astro/commit/f7afdb889fe4e97177958c8ec92f80c5f6e5cb51#diff-d8b56c1e9ed82985e955e08eaa9b7c9900b8f71172be47b76ca72dab68f22a44

### DIFF
--- a/.github/hacktoberfest.md
+++ b/.github/hacktoberfest.md
@@ -32,3 +32,9 @@ These awesome people reviewed i18n PRs for us. Thank you! 💜
 - @AbrahamX3 reviewed #1705
 - @breadadams reviewed #1767
 - @breadadams reviewed #1930
+- @moustaphadev reviewed #1656
+- @moustaphadev reviewed #1890
+- @moustaphadev reviewed #1900
+- @moustaphadev reviewed #1902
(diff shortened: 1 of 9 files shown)


delucis (commented):
Thanks! I can confirm this is the current behaviour: https://stackblitz.com/edit/github-9jfton?file=src/pages/index.astro

Going to check in with @bholmesdev that this is expected (I think it probably is).

Holben888 (approved):
U r eLevating our docs one PR at a time 👍

>

delucis (approved):
Thanks again @fuzzykiller! Removed the changes to the non `/en/` pages as we leave those to our translators to update.

PR #1976, opened 2022-11-02
Set up redirects based on 404 data & filter sitemap

delucis:

#### Description

- Add a Netlify redirect for an old `renderer-reference` URL

- Add a route with a `<meta>` refresh redirect for `/[lang]/tutorial`

    We may one day _have_ an actual `/tutorial` landing page potentially, so using a `.astro` route with `<meta>` seemed safer than configuring a Netlify redirect we’d be more likely to forget to remove.

#### Testing

These URLs were 404s but should now work:

- https://deploy-preview-1976--astro-docs-2.netlify.app/reference/renderer-reference/
- https://deploy-preview-1976--astro-docs-2.netlify.app/en/tutorial/

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -1,20 +1,17 @@
-import { defineConfig } from 'astro/config';
 import preact from '@astrojs/preact';
-import sitemap from '@astrojs/sitemap';
+import { defineConfig } from 'astro/config';
 
 import { toString } from 'hast-util-to-string';
 import { h } from 'hastscript';
 import { escape } from 'html-escaper';
 
-import { tokens, foregroundPrimary, backgroundPrimary } from './syntax-highlighting-theme';
 import { astroAsides } from './integrations/astro-asides';
-import { astroSpoilers } from './integrations/astro-spoilers';
 import { astroCodeSnippets } from './integrations/astro-code-snippets';
-import { remarkFallbackLang } from './plugins/remark-fallback-lang';
+import { astroSpoilers } from './integrations/astro-spoilers';
+import { sitemap } from './integrations/sitemap';
 import { rehypeTasklistEnhancer } from './plugins/rehype-tasklist-enhancer';
-
-import languages from './src/i18n/languages';
-import { normalizeLangTag } from './src/i18n/bcp-normalize';
+import { remarkFallbackLang } from './plugins/remark-fallback-lang';
+import { backgroundPrimary, foregroundPrimary, tokens } from './syntax-highlighting-theme';
 
 const AnchorLinkIcon = h(
 	'svg',
@@ -38,12 +35,6 @@ const createSROnlyLabel = (text: string) => {
 	return node;
 };
 
-const sitemapBlocklist = new Set([
-	...Object.keys(languages).map((lang) => `/${lang}/`),
-	...Object.keys(languages).map((lang) => `/${lang}/install/`),
-	...Object.keys(languages).map((lang) => `/${lang}/tutorial/`),
-]);
-
 // https://astro.build/config
 export default defineConfig({
 	site: 'https://docs.astro.build/',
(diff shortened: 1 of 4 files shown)


src/pages/[lang]/tutorial.astro
```diff
@@ -0,0 +1,9 @@
+---
+import languages from '../../i18n/languages';
+
+export const getStaticPaths = () => Object.keys(languages).map((lang) => ({ params: { lang } }));
+
+const { lang } = Astro.params;
+---
+
+<meta http-equiv="refresh" content={`0;url=/${lang}/tutorial/0-introduction/`} />
```
Comment (tony-sull) [inline src/pages/[lang]/tutorial.astro:9]: Will this show up in the sitemap?

Comment (delucis) [reply src/pages/[lang]/tutorial.astro:9]: Excellent question. The answer is: yes!

https://deploy-preview-1976--astro-docs-2.netlify.app/sitemap-0.xml

We can filter sitemap URLs I believe. Let me look at that — we should probably exclude our other similar redirects too.

sarah11918:
No objections here, once you're happy @delucis !

Just putting it on the radar that since we've discussed adding additional content in the "Tutorials" section, maybe it's time for @Jutanium to start discussing/sketching out a generic tutorial landing page anyway? I think this could fit in next sprint, actually!

src/pages/[lang]/tutorial.astro
Comment (delucis) [reply src/pages/[lang]/tutorial.astro:9]: OK, question back time: does it matter if it is in sitemap?

Comment (tony-sull) [reply src/pages/[lang]/tutorial.astro:9]: Its usually best to keep redirects out of the sitemap, but that can be tackled in a separate PR :+1:

Its really not a huge deal in general, I think it's pretty common for redirects to linger in a sitemap for a while (or forever on some sites)

tony-sull (approved):
lgtm!

src/pages/[lang]/tutorial.astro
Comment (delucis) [reply src/pages/[lang]/tutorial.astro:9]: OK — pushed a fix for the low-hanging fruit here. Our `[lang]/something.astro` redirects like the one added in this PR are easy to filter out. Other dynamic redirects in the root of `src/pages/` will be trickier and can be tackled later I guess!

astro.config.ts
```diff
@@ -47,6 +53,7 @@ export default defineConfig({
 	integrations: [
 		preact({ compat: true }),
 		sitemap({
+			filter: (page) => sitemapBlocklist.has(new URL(page).pathname),
```
delucis, line 56: Would help if I got the logic the right way round 😅 

```suggestion
			filter: (page) => !sitemapBlocklist.has(new URL(page).pathname),
```

src/pages/[lang]/tutorial.astro
Comment (delucis) [reply src/pages/[lang]/tutorial.astro:9]: Engaged morning brain and implemented a much better filtering set-up so now the sitemap shouldn’t include any of our redirect routes!

natemoo-re (approved):

integrations/sitemap.ts
```diff
@@ -0,0 +1,30 @@
...
+/** Match a pathname starting with “lighthouse” or one of our language tags. */
+const ValidRouteRE = new RegExp(`^/(lighthouse|${langTags.join('|')})/`);
+
+/** Test a pathname is not in our blocklist and starts with a valid prefix. */
+const isValidPath = (path: string) => !blocklist.has(path) && ValidRouteRE.test(path);
+
+/** Get a preconfigured instance of the `@astrojs/sitemap` integration. */
+export function sitemap(): AstroIntegration {
+	return AstroSitemap({
+		filter: (page) => isValidPath(new URL(page).pathname),
```
tony-sull, line 24: Nice!

tony-sull (approved):

PR #1977, opened 2022-11-02
Add feedback widget to tutorial pages

Jutanium:


src/components/tutorial/TutorialNav.astro
```diff
@@ -68,6 +68,16 @@ const isCurrentUnit = (unit: typeof units[number]) =>
...
 		}
 	</TabbedContent>
+
+	<div class="feedback">
+		<a href="javascript:;" 
+			data-feedback-fish
+			data-feedback-fish-path={currentUrl}
+			>
+			Give us feedback
+		</a>
```
delucis, line 78: We should probably use a button here as the functionality is not really that of a link:

```suggestion
		<button data-feedback-fish data-feedback-fish-path={currentUrl}>
			Give us feedback
		</button>
```

delucis (commented):
Thanks for looking into this @Jutanium! I’m excited to see what we can learn from users with this.

Quick initial design feedback, I’d lean towards this being outside of the tabbed UI as it’s always shown. Very rough example of what I mean after a quick play with dev tools:

image

src/components/tutorial/TutorialNav.astro
sarah11918: Agreed with Chris on both counts. I like a button outside of the nav.

Excited to try this!

delucis:
Tweaked the button styling a little.

| light | dark |
|-------|------|
| image | image |

sarah11918:
I noticed that now that it's out of the navigation component, it does not appear at all on any screen size that uses the drop down "on this page" navigation. It *might* be less of a concern for mobile, but since we decided tablet/not full screen view was important, I think we should try to make it visible there.

delucis:
Now with mobile feedback button 🎉 

| dark | light |
|---|---|
| image | image |

sarah11918:
What's the smallest screensize we should prepare for? At 321px, the two buttons will start to superimpose:

image

delucis:
Sarah notices everything 😓 

I knew this, but made sure it worked down to 325px which is the narrowest I’ve heard reported from users (had some issues at this width on astro.build).

NWTWWHB? 🤞

sarah11918:
Let's do it! 🥳

PR #1986, opened 2022-11-04
Improve `Astro.slots` reference

### LINKED ISSUE #595: `Astro.slots` example is not working

Jutanium:

- New or updated content

#### Description

- Closes #595
- Adds back an example for `Astro.slots.render` with the second argument
- Changes the example to use a string parameter rather than an array, so we don't run into array-inside-array confusion
- Removes TypeScript from the example (which only used it partially in the first place)


delucis:
Chris Update!

- Removed the table showing `has()` and `render()` in favour of subheadings
- Added dedicated explanation & example for `has()`
- Added a note _not_ to use `render()` if you really just need `<slot />`
- Tried to rework the `<Shout /> example a tiny bit but mostly gave up: stumbling block is we’ve never explained that you even _can_ pass a function as a child (I think this is mentioned nowhere else?) so this is like “advanced use of an advanced use we haven’t explained”. Trying to fix this _here_ felt bloated and out of place for reference.

sarah11918:
LGTM, @delucis! As you say, maybe an advanced usage, but seems clear enough.

PR #1987, opened 2022-11-04
i18n(fr): add new word to glossary

MoustaphaDev:

- New or updated content

#### Description

Add a french word for "Re-render"


delucis (approved):
Thanks @MoustaphaDev! Double checked React’s docs and they agree:

> [EN](https://reactjs.org/docs/react-component.html#shouldcomponentupdate): The default behavior is to re-render on every state change

> [FR](https://fr.reactjs.org/docs/react-component.html#shouldcomponentupdate): Le comportement par défaut consiste à rafraîchir à chaque modification

PR #1997, opened 2022-11-05
[tutorial] add a diff in 4-1 for clarity

morinokami:

#### Description

Added a diff to make it easier to see where changes have been made.

### DIFF
--- a/src/pages/en/tutorial/4-layouts/1.md
+++ b/src/pages/en/tutorial/4-layouts/1.md
@@ -101,11 +101,12 @@ You still have some Astro components repeatedly rendered on every page. Let's re
 
 1. Pass the page title to your layout component from `index.astro` using a component attribute: 
 
-    ```astro title="src/pages/index.astro" 'title={pageTitle}'
+    ```astro title="src/pages/index.astro" 'title={pageTitle}' del={5} ins={6}
     ---
     import BaseLayout from '../layouts/BaseLayout.astro';
     const pageTitle = "Home Page";
     ---
+    <BaseLayout>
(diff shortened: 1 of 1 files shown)


src/pages/en/tutorial/4-layouts/1.md
```diff
@@ -101,11 +101,12 @@ You still have some Astro components repeatedly rendered on every page. Let's re
 
 1. Pass the page title to your layout component from `index.astro` using a component attribute: 
 
-    ```astro title="src/pages/index.astro" 'title={pageTitle}'
+    ```astro title="src/pages/index.astro" 'title={pageTitle}' del={5} ins={6}
     ---
     import BaseLayout from '../layouts/BaseLayout.astro';
     const pageTitle = "Home Page";
     ---
+    <BaseLayout>
```
sarah11918, line 109:
``````suggestion
    ```astro title="src/pages/index.astro" 'pageTitle={pageTitle}'
    ---
    import BaseLayout from '../layouts/BaseLayout.astro';
    const pageTitle = "Home Page";
    ---
``````
I think I see the problem with the original highlighting as written!

This should be all that's required, and I think this should make only the attribute highlight. We don't really need to remove then rewrite the opening tag if we can get just the new part to highlight, and I think this should fix it!

morinokami: Ah, you're right! Fixed the code so that the added part is highlighted✍️

Thanks for your help!

sarah11918:
Woo! Then.... I think we're done??? Amazing work! Thank you so much for contributing to the tutorial!  💜 💪

PR #1998, opened 2022-11-05
[tutorial] move the first test in 2-2 to 2-3

morinokami:

#### Description

The first test in 2-2 is about `.astro` files, which has not yet been discussed. So I moved this test to 2-3 where it seems more appropriate.


sarah11918:
I noticed this myself last night, too!  😅 

Part of what happened is we rearranged some pages late in the process, to change the order that material was introduced. Originally, this did fit. But, we clearly missed this one when we tried to update all the pages for the new order!  💜

PR #2003, opened 2022-11-05
Change quiz question for Markdown tutorial page

iainsimmons:

- New or updated content

#### Description

- Change the [Test your knowledge quiz question and answers on the Markdown tutorial page](https://docs.astro.build/en/tutorial/2-pages/2/#test-your-knowledge)
- It was previously about YAML frontmatter, which had not been introduced
- It was previously the exact same as in the following page's [Test your knowledge section](https://docs.astro.build/en/tutorial/2-pages/3/#test-your-knowledge)

Just wanted to add that this tutorial is fantastic and just what I need to get more colleagues on board with Astro!

### DIFF
--- a/src/pages/en/tutorial/2-pages/2.md
+++ b/src/pages/en/tutorial/2-pages/2.md
@@ -164,10 +164,10 @@ Now that you have built pages using `.astro` files, let's make some blog posts u
 <Box icon="question-mark">
 ### Test your knowledge
 
-1. A `.md` file's frontmatter is written in:
+1. Content in a Markdown (`.md`) file is converted to:
     <MultipleChoice>
-        <Option>HTML</Option>
-        <Option isCorrect>YAML</Option>
+        <Option isCorrect>HTML</Option>
+        <Option>CSS</Option>
         <Option>JavaScript</Option>
     </MultipleChoice>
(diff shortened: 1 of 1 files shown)


iainsimmons:
Ahh sorry, just realised the page about dynamic HTML asks about `.astro` file frontmatter, instead of `.md`.

But there is still no mention of YAML frontmatter on the Markdown page.

sarah11918:
Hey @iainsimmons, this is a fantastic question; thank you!  🥳 

At not quite the *last* minute, we reordered and split up a bunch of pages for a better flow and as you have found, there still might be the odd thing that slipped past us regarding what has/hasn't come before.

I so appreciate you taking the time to contribute, coming up with a darn good question, and also for your very kind words about the tutorial! Welcome to Team Docs, and look for yourself in the Face Pile soon!  🙌

iainsimmons:
> At not quite the _last_ minute, we reordered and split up a bunch of pages for a better flow and as you have found, there still might be the odd thing that slipped past us regarding what has/hasn't come before.

This makes a lot of sense. Whatever you and the team did, it works so well! I'm still going through the tutorial, but it already shows such a level of care, detail and understanding of your potential users.

PR #2009, opened 2022-11-07
Add component usage example in the MDX docs

MadLittleMods:

- New or updated content

#### Description

Add Astro and React component usage example in the MDX docs. This was pretty hard to figure out even though the docs say it's possible. And there is confusion with the old `setup: |` syntax that is [spread](https://github.com/snowpackjs/astro/issues/491) [across](https://craigglennie.com/posts/2021/rewriting-in-astro) the internet already.

The only component usage example I could find in the existing docs comes from the migration guide converting an old React `.md` to `.mdx` in https://docs.astro.build/en/migrate/#converting-existing-md-files-to-mdx

Relevant docs:

 - https://docs.astro.build/en/guides/markdown-content/#using-components-in-mdx
    - This PR adds an example here
 - https://docs.astro.build/en/guides/integrations-guide/mdx/#components
    - It would be nice to cross-link or have an example here too (suggestion welcome)
    - Update: it looks like we sorta have an example in https://docs.astro.build/en/guides/integrations-guide/mdx/#custom-components but this is a little different for overriding existing components. And I completely missed this when Googling and trying to figure this out.

Update: Or perhaps link to https://github.com/withastro/astro/blob/latest/examples/with-mdx/src/pages/index.mdx?plain=1 which I just found from https://docs.astro.build/en/guides/integrations-guide/mdx/#examples

### DIFF
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -273,6 +273,22 @@ With the `@astrojs/mdx` integration, you can use Astro or UI framework component
 
 Don't forget to include a `client:directive` if necessary!
 
+```astro title="src/pages/about.mdx"
+---
+layout: ../layouts/BaseLayout.astro
+title: About me
+---
+import Button from '../components/Button.astro';
+import ReactCounter from '../components/ReactCounter.jsx';
+
+I live on **Mars** but feel free to <Button title="Contact me" />.
+
+Here is my counter component, working in MDX:
+
+<ReactCounter client:load />
+```
+
+
 ## Importing Markdown
 
 You can import Markdown and MDX files directly into your Astro files! You can import one specific page with `import` or multiple pages with `Astro.glob()`.
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -273,7 +273,7 @@ With the `@astrojs/mdx` integration, you can use Astro or UI framework component
 
 Don't forget to include a `client:directive` if necessary!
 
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/markdown-content.md
```diff
@@ -273,6 +273,22 @@ With the `@astrojs/mdx` integration, you can use Astro or UI framework component
 
 Don't forget to include a `client:directive` if necessary!
 
+```astro title="src/pages/about.mdx"
```
MadLittleMods, line 276: The syntax here might need updating. I wasn't sure what to put for the `{3,6}` stuff, etc

delucis: I’d suggest this here:

````suggestion
```astro title="src/pages/about.mdx" {5-6} /<.+\/>/
````

- `{5-6}` will highlight lines 5 to 6 where the import statements are.
- `/<.+\/>/` is regular expression that will match both `<Button />` and `<ReactCounter />` and highlight the usage.

delucis (commented):
Thanks @MadLittleMods! I’ve left a suggestion for the code block syntax.

src/pages/en/guides/markdown-content.md
MadLittleMods: Thanks for the review and explanation @delucis ❤️

delucis (approved):
Thanks again @MadLittleMods! Welcome to team docs 🥳

MadLittleMods:
Thanks for the review and merge @delucis 👯

PR #2010, opened 2022-11-07
Update server-side-rendering.md

connected-tcaceres:

#### Description
- minor doc update (positioning of periods)

### DIFF
--- a/src/pages/en/guides/server-side-rendering.md
+++ b/src/pages/en/guides/server-side-rendering.md
@@ -61,7 +61,7 @@ You can add any of the official adapters with the following `astro add` command.
 
 #### Manual Install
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @connected-tcaceres, thanks for taking the time to contribute to the docs! We absolutely want to get our periods correct.  ✅  

However, in this case, the periods you've indicated are correct as written. When an entire sentence is contained within parentheses, the parentheses themselves aren't a part of the larger sentence: they completely surround it. When only the last bit of a sentence is parenthetical, then absolutely yes, the period goes outside, belonging to the larger sentence. 

So, I'm closing this PR, but we'd love it if you'd PR any of those in future if you find them! We can always use extra eyes like yours to spot the things we don't catch in our editing. 👀

connected-tcaceres:
Ah, sorry. For some reason I totally missed that the whole sentence was in parentheses.

sarah11918:
No worries! We'd rather see it just in case! Thanks again for contributing, and don't let this stop you from trying again. I'm sure you'll find *something* we missed eventually!  😅

PR #2012, opened 2022-11-07
[tutorial] Added step: checking for tags

sarah11918:
Makes a clear instruction to ensure that every Markdown blot post written includes tags by adding a new Step 1 to the instructions for generating tag pages. 

In order to keep the code as simple as possible for generating tag pages, I did not include any conditional (if no tag) logic. Some readers are encountering errors in this line, that is likely caused by having created their own content for blog posts that does not include a `tags` array. (I suspect this is why only *some* people are encountering this error.)

There is a line in the previous section that says "make sure to include all these front matter properties" but that might not be strong enough, and I'll consider revisiting that page, too.

### DIFF
--- a/src/pages/en/tutorial/5-astro-api/2.md
+++ b/src/pages/en/tutorial/5-astro-api/2.md
@@ -158,9 +158,11 @@ The following example shows how to replace your code on this page with code that
 Even if it looks challenging, you can try following along with the steps to build this function yourself! If you don't want to walk through the JavaScript required right now, you can skip ahead to the [finished version of the code](#final-code-sample) and use it directly in your project, replacing the existing content.
 :::
 
+### 1. Check that all your blog posts contain tags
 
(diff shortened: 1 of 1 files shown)


delucis (commented):
Nice! Definitely good to get this reminder in to check all your posts have tags.

I don’t think it would be a _bad_ idea to have a more future-proof version like:

```js
    const filteredPosts = allPosts.filter((post) => {
      if (!post.frontmatter.tags) return false;
      return post.frontmatter.tags.includes(tag));
    }
```

While we’re here, I guess `filteredPosts` could also be `postsWithTag` or something to be more explicit, but I’ll leave this up to you. LGTM!

sarah11918:
Thanks Chris!

I'll merge this quickly now, as it is a quick fix to the immediate problem. Changing the code in the sample here would also mean changing the project's GitHub repo code, so I'll create a task to do all of that together when I can meticulously check all the moving pieces.

PR #2014, opened 2022-11-07
[tutorial] use PackageManagerTabs in 1-2 and 5-4

morinokami:

#### Description

I noticed that PackageManagerTabs component is used for the first command in 1-2, but not where the other npm commands are used. So, for consistency, I've updated 1-2 and 5-4 to use this component.


sarah11918:
This looks helpful! Thanks for the thoughtful addition! (I'm still trying to keep up with you...)  😄

PR #2015, opened 2022-11-07
add remark plugin to replace asides

### LINKED ISSUE #967: Integration page asides
Here’s something that I didn’t get to while working on #954 that would be cool to add to the integration READMEs workflow.

The READMEs currently use a mix of styles for tips and warnings using blockquotes as is common on GitHub. I would like to see these standardised to [GitHub’s new syntax for notes and warnings](https://github.com/github-community/community/discussions/16925):

```md
> **Note**
> This is a note

> **Warning**
> This is a warning
```

> **Note**
> This is a note

> **Warning**
> This is a warning

**THEN** I would like to add a remark plugin to our processing pipeline here:

https://github.com/withastro/docs/blob/2c93d8ff0615ac11a6ddcc0b009636c166530fe0/scripts/generate-integration-pages.ts#L73-L77

That finds that syntax and converts it to:

```md
:::note
This is a note
:::

:::caution
This is a warning
:::
```

kevinzunigacuellar:

#### Description

- Closes #967 
- Probably needs a lot of refactoring but the main idea is there 😇

### DIFF
--- a/scripts/generate-integration-pages.ts
+++ b/scripts/generate-integration-pages.ts
@@ -186,14 +186,14 @@ function replaceAsides() {
 		visit(tree, 'blockquote', (node) => {
 			// first child (<blockquote><p>...</p></blockquote>)
 			const [firstChild, trailingText] = node.children[0].children;
-			
+
 			if (firstChild.type === 'strong') {
 				if (
 					// check for **Note** or **Warning**
 					firstChild.children[0].value.includes('Note') ||
 					firstChild.children[0].value.includes('Warning')
 				) {
-					let AsideType =
+					const AsideType =
 						firstChild.children[0].value.toLowerCase() === 'note' ? 'note' : 'caution';
 
 					// remove blockquotes `>`
--- a/scripts/generate-integration-pages.ts
+++ b/scripts/generate-integration-pages.ts
@@ -182,37 +182,33 @@ function absoluteLinks({ base }: { base: string }) {
 /** Remark plugin to replace GitHub . */
 function replaceAsides() {
 	return function transform(tree: Root) {
-		// visit blockquote
 		visit(tree, 'blockquote', (node) => {
-			// first child (<blockquote><p>...</p></blockquote>)
 			const openingParagraph = node.children[0];
 			const [firstChild, trailingText] = openingParagraph.children;
 
-			if (firstChild.type === 'strong') {
-				if (
-					// check for **Note** or **Warning**
-					/Note|Warning/.test(firstChild.children[0].value)
-				) {
-					const AsideType =
-						firstChild.children[0].value.toLowerCase() === 'warning' ? 'caution' : 'note';
+			// check for **Note:** or **Warning:** at the beginning of the first paragraph
+			if (firstChild.type === 'strong' && /Note|Warning/.test(firstChild.children[0].value)) {
 
-					// remove blockquotes `>`
-					node.type = 'paragraph';
+				// assign aside type
+				const AsideType =
+					firstChild.children[0].value.toLowerCase() === 'warning' ? 'caution' : 'note';
 
-					// if starts with `: ` remove it
-					trailingText.value = trailingText.value.replace(/^: /, '\n');
+				// remove blockquotes `>`
+				node.type = 'paragraph';
 
-					// replace **strong** for :::aside
-					firstChild.type = 'text';
-					firstChild.value = `:::${AsideType}`;
+				// replace **strong** for :::aside
+				firstChild.type = 'text';
+				firstChild.value = `:::${AsideType}`;
 
-					// append ::: at end of the paragraph
-					const lastChild = {
-						type: 'text',
-						value: '\n:::',
-					};
-					node.children.push(lastChild);
-				}
+				// if trailingText starts with `: ` replace it with a newline
+				trailingText.value = trailingText.value.replace(/^: /, '\n');
+
+				// append ::: at end of the paragraph
+				const lastChild = {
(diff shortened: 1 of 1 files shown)


scripts/generate-integration-pages.ts
```diff
@@ -178,6 +179,47 @@ function absoluteLinks({ base }: { base: string }) {
 	};
 }
 
+/** Remark plugin to replace GitHub . */
+function replaceAsides() {
+	return function transform(tree: Root) {
+		// visit blockquote
+		visit(tree, 'blockquote', (node) => {
+			// first child (<blockquote><p>...</p></blockquote>)
+			const [firstChild, trailingText] = node.children[0].children;
```
delucis, line 188: It can be a bit hard to follow all the nested `children` in remark plug-ins, so I personally find it can help declaring variables so we get some more explicit names, like:

```suggestion
			const openingParagraph = node.children[0];
			const [firstChild, trailingText] = openingParagraph.children;
```

scripts/generate-integration-pages.ts
```diff
@@ -178,6 +179,47 @@ function absoluteLinks({ base }: { base: string }) {
...
+
+			if (firstChild.type === 'strong') {
+				if (
+					// check for **Note** or **Warning**
+					firstChild.children[0].value.includes('Note') ||
+					firstChild.children[0].value.includes('Warning')
```
delucis, line 194: I might be tempted to reach for a regular expression here:

```suggestion
					/Note|Warning/.test(firstChild.children[0].value)
```

scripts/generate-integration-pages.ts
```diff
@@ -178,6 +179,47 @@ function absoluteLinks({ base }: { base: string }) {
...
+				) {
+					const AsideType =
+						firstChild.children[0].value.toLowerCase() === 'note' ? 'note' : 'caution';
```
delucis, line 197: This technically makes “caution” our default note type. Not sure we’d ever actually hit this, but just in case I think defaulting to “note” probably makes more sense:

```suggestion
						firstChild.children[0].value.toLowerCase() === 'warning' ? 'caution' : 'note';
```

scripts/generate-integration-pages.ts
```diff
@@ -178,6 +179,47 @@ function absoluteLinks({ base }: { base: string }) {
...
+
+					// remove blockquotes `>`
+					node.type = 'paragraph';
+
+					// if starts with `: ` remove it
+					if (trailingText.value.startsWith(': ')) {
+						trailingText.value = trailingText.value.replace(/^: /, '\n');
+					}
```
delucis, line 205: Should be safe to skip the `if` here I think: if `/^: /` doesn’t match, nothing will be replaced anyway.

```suggestion
					trailingText.value = trailingText.value.replace(/^: /, '\n');
```

delucis: Question: do we need to add the additional `\n`? If so, I’d love the comment in the line above to mention why.

delucis (commented):
This is awesome — so happy we (YOU!) finally did this!

Left some comments, but seems pretty great to me 😀

scripts/generate-integration-pages.ts
kevinzunigacuellar: caution this may explode CI 😆🔥😇

kevinzunigacuellar:
'we are Groot' 💜

scripts/generate-integration-pages.ts
```diff
@@ -178,6 +179,41 @@ function absoluteLinks({ base }: { base: string }) {
 	};
 }
 
+/** Remark plugin to replace GitHub . */
```
delucis, line 182: Nit: this comment is complete!

```suggestion
/** Remark plugin to replace GitHub note/warning syntax with docs-style asides. */
```

scripts/generate-integration-pages.ts
```diff
@@ -178,6 +179,41 @@ function absoluteLinks({ base }: { base: string }) {
...
+function replaceAsides() {
+	return function transform(tree: Root) {
+		visit(tree, 'blockquote', (node) => {
+			const openingParagraph = node.children[0];
+			const [firstChild, trailingText] = openingParagraph.children;
+
+			// check for **Note:** or **Warning:** at the beginning of the first paragraph
+			if (firstChild.type === 'strong' && /Note|Warning/.test(firstChild.children[0].value)) {
```
delucis, line 190: Style nit you can skip if you prefer: this could be an opportunity for an early return:

```js
if (
	firstChild.type !== 'strong') ||
	!/Note|Warning/.test(firstChild.children[0].value)
) {
	return;
}
```

That would avoid nesting everything that follows inside the `if` statement. But it’s fine as-is too.

delucis (commented):
Couple of small nits here, but once you’ve accepted or dismissed these — LGTM!

scripts/generate-integration-pages.ts
kevinzunigacuellar: I think I saw early returns as common practice in remark plugins. I will add this 😇

delucis (approved):
LGTM! Awesome work on this one Kevin!

Fun fact about the early returns, a lot of the remark utilities like `visit` we’re using here, secretly support special constants to control them:

```js
import { visit, CONTINUE, EXIT, SKIP } from 'unist-util-visit';

function transform(tree) {
  visit(tree, node => {
    if (node.notThisOne) return CONTINUE;
    if (node.notThisOrItsChildren) return SKIP;
    if (node.nothingAfterThis) return EXIT;
  });
}
```

PR #2020, opened 2022-11-09
Update 5.3 tags styles & markup

delucis:

- New or updated content

#### Description

Updates HTML markup and styles used in the code challenge in tutorial unit 5.3:
- It’s now just a paragraph containing links
- Styles updated to look like this:
    image

- Also checked with later changes and seems to play well when we add dark mode:
   image

I _might_ have opted for an unordered list for this, but list styles complicate things a bit, so this markup keeps things simple.

I have the equivalent changes cued up to push to https://github.com/withastro/blog-tutorial-demo when we’re ready (@sarah11918 noticed we missed this bit of tags code entirely there).

### DIFF
--- a/src/pages/en/tutorial/5-astro-api/3.md
+++ b/src/pages/en/tutorial/5-astro-api/3.md
@@ -262,49 +262,46 @@ Follow the steps below, then check your work by comparing it to the [final code 
 
 1. Copy the `<div class="tags">`... `</div>` and `<style>...</style>` from `src/pages/tags/index.astro` and reuse it inside `MarkdownLayoutPost.astro`: 
 
-  ```astro title="src/layouts/MarkdownPostLayout.astro" ins={14-18, 22-41}
-  ---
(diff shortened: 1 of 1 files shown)


sarah11918:
This all looks great, @delucis! Just checking, because originally we had duplicated what was on `tags/index.astro` with what was in `MarkdownPostLayout.astro`. So, your changes here change each individual post (by modifying the layout).

Is it your intention to leave the `/tags/` page unchanged? If not, then we should use these styles earlier on the page too, to also style the way the tags show up on that page.

delucis:
It was intentional that they be different, but maybe that doesn’t make sense? I still need to revisit the tags page (would be a separate PR), but because there the tags are more like THE thing, I thought different styles could make sense. If you think it’s safer to make them match, I can fix that in this PR though.

delucis:
Going to close this — will take another pass at this when there’s time.

PR #2023, opened 2022-11-09
[tutorial] 3-2 Introduce props

sarah11918:
- New or updated content

We received a Feedback Fish that in 3-2, we didn't really introduce or explain props before using them.

This adds a minimal introduction to the Social Media component where before there was none, only instructions starting at step 1.


delucis (approved):
I like this! It’s minimal, but I think that’s in keeping with the “do, experience, learn” vibe we have going.

PR #2024, opened 2022-11-09
Add an error reference page

Princesseuh:
Docs part of https://github.com/withastro/astro/pull/5355. Require that PR to be merged in before we merge this

- New or updated content

#### Description

Adds an automatically generated reference page for all errors that can happen in Astro. This needs a PR to compiler and core to actually add the documentation before being merged, but at least the mechanism is there!


scripts/error-docgen.mjs
```diff
@@ -0,0 +1,242 @@
...
+<DontEditWarning />
+
+The following reference covers all errors that can be emitted by Astro. To learn more about common pitfalls, see our guide on [Troubleshooting](/en/guides/troubleshooting/).
+
+`;
+
+const FOOTER = ``;
+
+export async function run() {
+	const compilerErrorData = { errors: [], jsdoc: [] }; // await getCompilerErrors();
```
Princesseuh, line 45: Compiler errors are unintentionally not covered in this PR. I'll tackle it separately since it potentially requires some work in the error shapes from the compiler

scripts/error-docgen.mjs
```diff
@@ -0,0 +1,242 @@
...
+function extractStringFromFunction(func) {
+	const arrowIndex = func.indexOf('=>') + '=>'.length;
+
+	// eslint-disable-next-line no-useless-escape
+	return escapeHtml(func.slice(arrowIndex).trim().slice(1, -1));
+
+	function escapeHtml(unsafe) {
+		return unsafe
+			.replaceAll('${', '`')
+			.replaceAll('}', '`')
```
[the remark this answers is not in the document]
delucis: Related: do we want these blockquotes to look more like they will in the terminal/overlay? I could add some custom styles for this page if so, e.g.

image

scripts/error-docgen.mjs
```diff
@@ -0,0 +1,242 @@
...
+			.replaceAll(':', '\\:')
```
delucis, line 219: Is this addressing the issue we had with colons like `client:only`?

Looks like we also need it for headings:

image

Full text should be “Missing hint on client:only directive”.

delucis (commented):
Left some initial comments. Great work!

One stylistic thing I wonder about: in docs we usually wrap any bit of code in `<code>` backticks, including in headings. Is it impractical to include these in the JSDoc comments too or would it mess with use of these strings elsewhere.

Example of what I mean:

```diff
- Invalid value returned by getStaticPaths
+ Invalid value returned by `getStaticPaths`
```

scripts/error-docgen.mjs
Princesseuh: Ah yep, it was to work around the issue. Good catch!

src/i18n/en/nav.ts
```diff
@@ -88,6 +88,11 @@ export default [
 		slug: 'reference/configuration-reference',
 		key: 'reference/configuration-reference',
 	},
+	{
+		text: 'Errors',
+		slug: 'reference/error-reference',
+		key: 'reference/error-reference',
+	},
```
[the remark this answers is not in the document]
sarah11918: Yes, agree @delucis. I would also suggest a slightly more descriptiove name like "Error Reference" or "Error Glossary." Side note: I tried to quickly search to see what OTHER frameworks call this page, and I couldn't find anyone else with one!  💅

src/pages/en/reference/error-reference.md
```diff
@@ -0,0 +1,387 @@
...
+i18nReady: true
+githubURL: https://github.com/withastro/astro/blob/main/packages/astro/src/core/errors/errors-data.ts
+setup: |
+  import Since from '../../../components/Since.astro';
+  import DontEditWarning from '../../../components/DontEditWarning.astro';
+---
+
+<DontEditWarning />
+
+The following reference covers all errors that can be emitted by Astro. To learn more about common pitfalls, see our guide on [Troubleshooting](/en/guides/troubleshooting/).
```
sarah11918, line 18: Sorry, I couldn't find where to edit this statement. (I could find the messages themselves, but not this line.)

Can I suggest something like:

> The following reference is a complete list of the errors you may encounter while using Astro. For additional assistance, including common pitfalls, please also see our [Troubleshooting Guide](/en/guides/troubleshooting/).

sarah11918:
I'm so excited for this page! I left a comment re: maybe tweaking the opening paragraph on the page (before the error messages), and I think I'd like the sidebar entry to say more than just "Errors"... but otherwise, I'm happy to let @Princesseuh and @delucis make the call re: specific visuals and formatting. I'll jump in if asked, but don't feel the need to have any input with you both on the case!

src/i18n/en/nav.ts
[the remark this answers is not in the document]
Princesseuh: Moved to the bottom and called it "Error Reference" 👍

scripts/error-docgen.mjs
```diff
@@ -0,0 +1,256 @@
+/* eslint-disable no-mixed-spaces-and-tabs */
+import fs from 'fs';
+import jsdoc from 'jsdoc-api';
+import fetch from 'node-fetch';
+import ts from 'typescript';
+
+import '../../astro/packages/astro/dist/core/errors/errors-data.js';
```
delucis, line 7: This is a left-over import that needs removing right?

```suggestion
```

scripts/error-docgen.mjs
```diff
@@ -0,0 +1,256 @@
...
+			comment.description && `#### What went wrong?`,
+			comment.description && comment.description.trim(),
+			``,
+			comment.see
+				? `**See Also:**\n${comment.see.map((s) => `- ${s.replace('-', '')}`.trim()).join('\n')}`
+				: undefined,
+			`\n\n`,
+		]
+			.filter((l) => l !== undefined)
+			.join('\n');
```
delucis, line 113: Doesn’t need to block this PR, but this is… fun to read 😅 

If there’s time, I’d love to see a refactored version that breaks this logic up a bit, but understand if that’s not possible.

scripts/error-docgen.mjs
```diff
@@ -0,0 +1,256 @@
...
+
+	const jsDocComments = jsdoc
+		.explainSync({ source: compiledResult })
+		.filter(
+			(data) =>
+				data.kind !== 'package' &&
+				!data.undocumented &&
+				data.tags &&
+				data.tags.some((tag) => tag.title === 'docs')
+		);
```
delucis, line 167: This logic is used here and in L191 and I’m not really able to follow what’s going on just by reading it. Might be good to either extract to a reused function with a descriptive name or just add a comment to each of them that explains exactly what’s being filtered for and why.

delucis (commented):
I think this is almost good to go! First thing I think might _need_ fixing. Second & third comments are more subjective things we can let slide if necessary. Excited to welcome this page into the happy docs family!

scripts/error-docgen.mjs
Princesseuh: This is definitely possible, a lot of the structure is based on the other docgen scripts (which are also a tiny bit hard to read). Will do my best to make things clearer!

delucis (approved):
Awesome! Thanks so much for making those changes @Princesseuh 💜 

LGTM 🎉

PR #2025, opened 2022-11-09
Fix use of `rehype-minify` in Markdown plugins example

### LINKED ISSUE #1722: Update Markdown example showing rehype-minify plugin
Our example for Markdown plugins (https://docs.astro.build/en/guides/markdown-content/#markdown-plugins) shows a rehype-minify plugin being used. But, this is a collection of plugins, not a plugin itself. 

The plugin used in the example code should be replaced. Maybe the preset [`rehype-preset-minify`](https://github.com/rehypejs/rehype-minify/tree/main/packages/rehype-preset-minify) should be used instead? Maybe @bholmesdev has an opinion?

delucis:

#### Description

- Fixes #1722
- Updates our Markdown plugin example to use `rehype-preset-minify` instead of `rehype-minify` which is a collection of utilities and can’t be used directly.

### DIFF
--- a/src/pages/en/guides/markdown-content.md
+++ b/src/pages/en/guides/markdown-content.md
@@ -421,7 +421,7 @@ The instructions below are for configuring standard Markdown. To configure MDX p
 
 Astro supports third-party [remark](https://github.com/remarkjs/remark) and [rehype](https://github.com/rehypejs/rehype) plugins for Markdown. These plugins allow you to extend your Markdown with new capabilities, like [auto-generating a table of contents](https://github.com/remarkjs/remark-toc), [applying accessible emoji labels](https://github.com/florianeckerstorfer/remark-a11y-emoji), and more. We encourage you to browse [awesome-remark](https://github.com/remarkjs/awesome-remark) and [awesome-rehype](https://github.com/rehypejs/awesome-rehype) for popular plugins!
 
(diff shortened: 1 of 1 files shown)


sarah11918:
LGTM, @delucis!

src/pages/en/guides/markdown-content.md
```diff
@@ -430,12 +430,12 @@ Astro applies the [GitHub-flavored Markdown](https://github.com/remarkjs/remark-
 ```js title="astro.config.mjs" ins={2,3,7,8,11}
 import remarkToc from 'remark-toc';
-import rehypeMinifyHtml from 'rehype-minify';
+import { rehypeAccessibleEmojis } from 'rehype-accessible-emojis';
```
delucis, line 433: I ran some tests and Astro doesn’t actually support `rehype-preset-minify` either. The preset is an object like `{ settings, plugins }` which is apparently supported by `rehype.use(...)` but not when passed to Astro’s `rehypePlugins` array (maybe it should be supported, not sure?)

So I’ve just switched this out completely for something that does work and is cool: a plugin that wraps emojis in a span with an accessible label.

sarah11918: Oooh, way cooler!  ❄️ 🚀  LGTM!

PR #2029, opened 2022-11-10
Update `AstroInstance` interface reference

MoustaphaDev:

- New or updated content

#### Description

The update the [`AstroInstance`](https://github.com/withastro/astro/blob/main/packages/astro/src/@types/astro.ts#L983-L986) interface

### DIFF
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -86,6 +86,7 @@ Astro files have the following interface:
 
 ```ts
 export interface AstroInstance {
+  /* The file path of this file */
   file: string;
 	url: string | undefined;
 	default: AstroComponent;
--- a/src/pages/en/reference/api-reference.md
+++ b/src/pages/en/reference/api-reference.md
@@ -86,6 +86,8 @@ Astro files have the following interface:
 
 ```ts
 export interface AstroInstance {
+  file: string;
+	url: string | undefined;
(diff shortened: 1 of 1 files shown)


src/pages/en/reference/api-reference.md
```diff
@@ -86,6 +86,8 @@ Astro files have the following interface:
 
 ```ts
 export interface AstroInstance {
+  file: string;
```
delucis, line 89: Do we want to add some comments like we do for `MarkdownInstance`?

```suggestion
  /* The file path of this file */
  file: string;
```

src/pages/en/reference/api-reference.md
```diff
@@ -86,6 +86,8 @@ Astro files have the following interface:
...
+	url: string | undefined;
```
delucis, line 90: Maybe we should also hint at _why_ this is sometimes `undefined`?

```suggestion
  /* The URL for this file (if it is in the pages directory) */
	url: string | undefined;
```

delucis (commented):
Thanks @MoustaphaDev, this is helpful! Left some comments.

src/pages/en/reference/api-reference.md
MoustaphaDev: Thanks for the suggestion!

delucis (approved):
Thanks! Great improvement 🙌

PR #2030, opened 2022-11-10
Update typescript.md

jessegood:
There are tons of examples of `export` being used on  `interface Props` out in the wild so I think it would be good to make clear this is not necessary in the docs.

#### Description
Added a note stating you don't need `export` on `interface Props`.

### DIFF
--- a/src/pages/en/guides/typescript.md
+++ b/src/pages/en/guides/typescript.md
@@ -85,11 +85,7 @@ import Layout from '@layouts/Layout.astro';
 
 ## Component Props
 
-Astro supports typing your component props via TypeScript. To enable, add a TypeScript `Props` interface to your component frontmatter. The [Astro VSCode Extension](/en/editor-setup/) will automatically look for the `Props` interface and give you proper TS support when you use that component inside another template.
-
-**Note**\
-Many of the themes and examples you will find still use `export` on `interface Props` as originally our docs said you needed to. However, this is *not* necessary unless you really want to import this into another file! (Which is probably not the case).
-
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/typescript.md
```diff
@@ -87,6 +87,9 @@ import Layout from '@layouts/Layout.astro';
 
 Astro supports typing your component props via TypeScript. To enable, add a TypeScript `Props` interface to your component frontmatter. The [Astro VSCode Extension](/en/editor-setup/) will automatically look for the `Props` interface and give you proper TS support when you use that component inside another template.
 
+**Note**\
+Many of the themes and examples you will find still use `export` on `interface Props` as originally our docs said you needed to. However, this is *not* necessary unless you really want to import this into another file! (Which is probably not the case).
+
```
sarah11918, line 92:
```suggestion
Astro supports typing your component props via TypeScript. To enable, add a TypeScript `Props` interface to your component frontmatter. An `export` statement maybe be used, but is not necessary. The [Astro VSCode Extension](/en/editor-setup/) will automatically look for the `Props` interface and give you proper TS support when you use that component inside another template.
```
Hi @jessegood! Thanks for taking the time to contribute the docs!

As you've noticed, in more recent versions of Astro the `export` is no longer necessary. Fortunately, it's not a breaking change since everything still works even if you do copy an older example out there with export.

We try to reserve notes (which are big, and visually take up a lot of screen room and the reader's attention) for items that really need special attention drawn to them, or breaking changes. Otherwise, we try not to distract the reader if we can, either visually or with extra text that makes reading harder to get through.

I think a little note in here just casually mentioning that the statements are optional serves your purpose of letting the reader know they're optional. It also shows up immediately when we instruct you to add the Props interface, so whenever we can put info EXACTLY where people are going to be using it, that's a win!

What do you think about my suggestion above?

jessegood: LGTM (y)

sarah11918:
Yay! Excited to welcome you to Team Docs, @jessegood!  🥳 Here's hoping that fewer people are confused because of you!

PR #2038, opened 2022-11-11
Update dependencies and format the code

zihan-ch:

#### Description

Update dependencies and format the code


delucis:
Hi @zihan-ch — thanks for trying this! In general it’s better to upgrade dependencies one by one (or in small related groups) as it can be harder to debug issues when updating everything in one go. I’ll close this PR, but let me know if there’s specific dependencies you think there’s a reason we need to update urgently and we can discuss updating those!

PR #2040, opened 2022-11-11
Add troubleshooting guidance for CSPs without unsafe-inline

delucis:

- New or updated content

#### Description

Add troubleshooting guidance based on @Princesseuh’s suggestion for users who have a Content Security Policy that disallows the use of inline scripts.

### DIFF
--- a/src/pages/en/guides/troubleshooting.md
+++ b/src/pages/en/guides/troubleshooting.md
@@ -52,7 +52,7 @@ You may see the following error logged in the browser console:
 
 > Refused to execute inline script because it violates the following Content Security Policy directive: …
 
-This means that your site’s [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP) disallows inline scripts. Because Astro uses inline `<script>` tags in its output by default, interactive components may break triggering this error.
+This means that your site’s [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP) disallows running inline `<script>` tags, which Astro outputs by default.
 
 **Solution:** To work around this, you can force Astro to bundle all styles and scripts into external assets using the [`vite.build.assetsInlineLimit`](https://vitejs.dev/config/build-options.html#build-assetsinlinelimit) setting in `astro.config.mjs`.
 
--- a/src/pages/en/guides/troubleshooting.md
+++ b/src/pages/en/guides/troubleshooting.md
@@ -127,6 +127,25 @@ nodeLinker: "node-modules"
 
 In Astro, using a `<head>` tag works like any other HTML tag: it does not get moved to the top of the page or merged with the existing `<head>`. Because of this, you usually only want to include one `<head>` tag throughout a page. We recommend writing that single `<head>` and its contents in a [layout component](/en/core-concepts/layouts/).
 
+### My Content Security Policy (CSP) disallows inline scripts
+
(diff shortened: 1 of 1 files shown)


Princesseuh:
Perhaps it would make more sense for this to be in the "Common errors" section? I wouldn't be surprised if some people were blindly copying CSP recommendations and would then look for the error message specifically instead

delucis:
Could be. Do we know what the error message looks like?

Princesseuh:
> Could be. Do we know what the error message looks like?

It looks like `Refused to execute inline script because it violates the following Content Security Policy directive: ${directiveUsed}`

delucis:
Thank you! This makes for a veerrrry long heading 😅 
image
Going to think about how best to handle it.

sarah11918:
> It looks like `Refused to execute inline script because it violates the following Content Security Policy directive: ${directiveUsed}`

Are there any *other* reasons why a user might get an error that starts with `Refused to execute inline script`? I could see *that* being a reasonable heading, and then explain this error in detail in the section. And, if it's not the only one, then CSP becomes one possible reason that the inline script didn't execute.

Princesseuh:
> > It looks like `Refused to execute inline script because it violates the following Content Security Policy directive: ${directiveUsed}`
> 
> Are there any _other_ reasons why a user might get an error that starts with `Refused to execute inline script`? I could see _that_ being a reasonable heading, and then explain this error in detail in the section. And, if it's not the only one, then CSP becomes one possible reason that the inline script didn't execute.

As far as I know, there's no other reasons you'd get `Refused to execute inline script`. You'd get similar messages for other CSPs rules, but they start and are worded differently

delucis:
How do we feel about this? Is truncating the error message an issue? It would be kind of cool to have CSP in the heading for search results (some people searched “csp” in the past month), but the full error as title seems over kill.

image

I took out the explanation of what CSP is/does in this version because it fit less well in this context.

src/pages/en/guides/troubleshooting.md
```diff
@@ -46,6 +46,31 @@ This error can be thrown when trying to import or render an invalid component, o
 
 **Status**: Expected Astro behavior, as intended.
 
+### Refused to execute inline script
+
+You may see the following error logged in the browser console:
+
+> Refused to execute inline script because it violates the following Content Security Policy directive: …
+
+This means that your site’s [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP) disallows inline scripts. Because Astro uses inline `<script>` tags in its output by default, interactive components may break triggering this error.
```
sarah11918, line 55:
```suggestion
This means that your site’s [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP) disallows inline scripts, which are Astro's default `<script>` output.
```
Maybe this is all that needs to be said?

delucis: Maybe this? Because it could also be a framework component rather than a `<script>` tag that causes the `<script>` in the build output. Only hesitancy is that this clashes with “zero JS by default”, but I think it’s OK in this context.

```suggestion
This means that your site’s [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP) disallows running inline `<script>` tags, which Astro outputs by default.
```

sarah11918: Looks great! (to me)

PR #2047, opened 2022-11-13
Fix strict type error in routing example

Astisme:
I was trying this out and got the title's error.
Because Astro uses `undefined` for the root, the first object could not be accessed using pages[slug]. If Astro were to use `"undefined"` instead, no problem would have occurred with this piece of code.

To fix this problem I had to use an if/else to check whether the type of slug was undefined and get the root page from the pages object. Then I was still getting this error for pages[slug]
```
Element implicitly has an 'any' type because expression of type 'string | number' can't be used to index type '{ undefined: { title: string; text: string; }; products: { title: string; text: string; }; "products/astro-handbook": { title: string; text: string; }; }'.
  No index signature with a parameter of type 'string' was found on type '{ undefined: { title: string; text: string; }; products: { title: string; text: string; }; "products/astro-handbook": { title: string; text: string; };
```
which could only be solved by instantiating the types for both page and pages.

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 


Astisme:
My bad. I didn't see the text below my description.
## What kind of changes does this PR include?

- Minor content fixes (error in code example)

## Description

- Closes nothing.
- What does this PR change? Read above
- Did you change something visual? No.

Astisme:
@sarah11918 I see the PR is waiting approval for the execution of a check. 

What should I do?

sarah11918:
Hi @Astisme! You just have to wait! We have not yet reviewed this contribution. When we review it, then we'll leave comments and be sure to let you know if there's anything further that needs to be done, or if we need to ask you questions. Thanks for your patience!

delucis (commented):
Hi @Astisme! I’m not sure how important it is to document this here. This type error will only occur with strict type checking (which Astro does not set by default).

If we really do want to make this compatible with script TS, I’d probably suggest a one-line fix:

```ts
const page = pages[slug as keyof typeof pages];
```

But in general we try to avoid this kind of syntax as a lot of people are not familiar with TypeScript.

An alternative would be to switch `pages` to use an array although this is less performant:

```ts
const pages = [
	{
		slug: undefined,
		title: 'Astro Store',
		text: 'Welcome to the Astro store!',
	},
	{
		slug: 'products',
		title: 'Astro products',
		text: 'We have lots of products for you',
	},
	{
		slug: 'products/astro-handbook',
		title: 'The ultimate Astro handbook',
		text: 'If you want to learn Astro, you must read this book.',
	}
];

const page = pages.find((page) => page.slug === slug);
```

delucis:
Let’s go with the array version — doesn’t require any TS trickery and like you say, is equivalent to the example above. Thanks @Astisme!

src/pages/en/core-concepts/routing.md
```diff
@@ -182,24 +182,26 @@ Because SSR pages can't use `getStaticPaths`, they can't receive props. Here, we
...
+		slug: 'products/astro-handbook',
+		title: 'The ultimate Astro handbook',
+		text: 'If you want to learn Astro, you must read this book.',
+	}
+];
 
 const { slug } = Astro.params;
-const page = pages[slug];
-if (!page) return Astro.redirect("/404");
+const page = pages.find( page => page.slug === slug);
```
delucis, line 204: Thanks @Astisme!

I think we will want to keep the 404 guard here, but otherwise looks great!

```suggestion
const page = pages.find((page) => page.slug === slug);
if (!page) return Astro.redirect("/404");
```

delucis (approved):
LGTM! Thanks again @Astisme 🙌

PR #2050, opened 2022-11-14
[tutorial] 4-3 update multiple choice answers

sarah11918:
- New or updated content

Improved multiple choice questions for section 4-3. One was previously questionable, but all were a hold-over from conversion from fill-in-the-blank format. So, I took the opportunity to provide better answers for all the questions in the process.


delucis (approved):
A) NWTWWHB
B) LGTM
C) both of the above?

✅🙌

PR #2061, opened 2022-11-16
Address typo

ted-dino:

#### Description

- What does this PR change? 
              - I noticed a typo while reading through documentation. 

**Before**
image

**After**
image

### DIFF
--- a/src/pages/en/guides/integrations-guide/image.md
+++ b/src/pages/en/guides/integrations-guide/image.md
@@ -87,7 +87,7 @@ export default {
 
 ### Installing `sharp` (optional)
 
-First, install the `sharp` package using your package manger. If you're using npm or aren't sure, run this in the terminal:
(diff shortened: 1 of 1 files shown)


delucis:
Thanks for this @ted-dino! Unfortunately, this is a page that gets automatically generated from this integration’s README, which lives here: https://github.com/withastro/astro/blob/main/packages/integrations/image/README.md

We’ll need to update the file in that repo instead of this one. I can still merge this, but it’ll get overwritten tomorrow 😅 

Would you be up for making this same fix in that file instead?

ted-dino:
> Thanks for this @ted-dino! Unfortunately, this is a page that gets automatically generated from this integration’s README, which lives here: https://github.com/withastro/astro/blob/main/packages/integrations/image/README.md
> 
> We’ll need to update the file in that repo instead of this one. I can still merge this, but it’ll get overwritten tomorrow sweat_smile
> 
> Would you be up for making this same fix in that file instead?

meaning i have to redo my PR?

Issue #618, opened 2022-05-20
Improve search experience


delucis:
### Have experience with or interested in learning more about DocSearch? Help us out and take a look at one of these tasks!

- [x] Increase the number of results returned by DocSearch (currently maxes out at 5). Maybe with [the `pagination` parameter](https://www.algolia.com/doc/api-reference/search-api-parameters/) passed to [`searchParameters`](https://docsearch.algolia.com/docs/api/#searchparameters-1)?

   What’s the number of results we want? Should it paginate to include as many as possible (if that’s possible)?

- [ ] Improve the snippet for each result so there’s clearer context around what you might be clicking on. (Awaiting more feedback from @sarah11918 on what we’d like there.)

- [x] Break up search results into categories: docs vs tutorial
- [x] Look into duplicate results. Here’s an example where results show up twice, one always linking to the page’s `#article` ID.

   <img width="597" alt="image" src="https://user-images.githubusercontent.com/357379/169560114-73322bf0-0413-4faa-b781-6042d019edef.png">

    22/11/3 — This doesn’t seem to be happening anymore?

- [x] Look into items in the integration nav component ranking in search, e.g. https://github.com/withastro/docs/issues/2042#issuecomment-1312723775
    
    **Update (18 Nov 2022):** This should be resolved when the Algolia crawler next runs — I updated the crawler to strip the navigation component from the page before extracting records.

sarah11918:
I owe you comments here, and I will provide!

1. My MAIN gripe is the number of items shown (which may just be a configuration setting?) I'm never convinced it's showing me all the hits for a search term, and I've seen it prioritize multiple hits on the same page rather than surfacing another page where something might be.

That almost makes it hard to evaluate further, because I don't know what I'm missing! So, I think even just your three items above would go QUITE a long way to improving this.

sarah11918:
Let's see... to throw some wild ideas out there


First set of results: the word is a hit for page title/file name
(So, if you just want to hit a major page for your result, you don't have to read anything in a snippet, just go.)

Second group of results: the word is in a heading
(Similarly, if you see a relevant heading, you probably don't care much about extra context displayed, just go.)

Otherwise, hits for the word in text display:
- Page name (file/route or title?)
- Section heading in which the text occurs (might restrict this to a L1 to L3 heading, otherwise it *might* be too specific to give a better context?)
- 2 - 3 lines surrounding the word (word in sentence context)
-  # of times the word appears on that page? (this would give a kind of "heat map")

[...]

PR #2071, opened 2022-11-18
Initial set-up for DocSearch categorisation [en-only]

delucis:

#### Description

- Working towards #618 
- Adds a hidden element to each page like `<span id="docsearch-lvl0" hidden>Learn</span>` that Algolia’s crawler can use to extract the page category
- Adds a small helper that works out the category (i.e. “Learn”) from the page URL
- Moves our one reference page that was still in `guides/` to `reference/` and sets up redirects for it (marking this PR as `en-only` so this doesn’t update i18n statuses)


src/layouts/BaseLayout.astro
```diff
@@ -139,6 +140,7 @@ if (isFallback) canonicalURL.pathname = canonicalURL.pathname.replace(`/${lang}/
 			</aside>
 			<div id="main-content" class="main-column">
 				<slot />
+				<span id="docsearch-lvl0" hidden>{getPageCategory(url)}</span>
```
delucis, line 143: I wanted to test out if the crawler picks this up as expected but it’s only able to crawl on our domain, so doesn’t work on the Netlify preview. I think it should, but if it doesn’t I’ll remove this and try a different approach.

delucis (commented):
I should mention that this is preliminary work — impact will be seen next time Algolia crawls. I’ll run a few tests before triggering that to make sure all my assumptions were correct.

src/util/getPageCategory.ts
```diff
@@ -0,0 +1,20 @@
+// TODO: Move this data to our i18n system to support localized category labels.
+const defaultCategory = 'Learn';
+const categories = [
+	['tutorial/', 'Tutorials'],
+	['reference/', 'Reference'],
+	['guides/deploy/', 'Deploy Guides'],
+	['guides/integrations-guide/', 'Integration Reference'],
```
sarah11918, line 7:
```suggestion
	['guides/integrations-guide/', 'Integration Guides'],
```

Amazing, @delucis! SO excited for this search categorization!  🙌 

I think I'd prefer "Integration Guides" here, to be more in keeping with the file path/URL and the tab they'll show up under? But LGTM!

PR #2072, opened 2022-11-18
Use unique class name in tutorial nav tabs

delucis:

#### Description

We use a `.content` class on our main page content and instruct DocSearch to use this to find content. We don’t want to include the tutorial navigation tabs in search results and also don’t want styles from the global `.content` bleeding into them, so this renames `.content` to `.panels`.


sarah11918:
LGTM! Thanks for being so diligent and attentive to this, @delucis!

PR #2074, opened 2022-11-19
fix pnpm installs in storyblok guide

kevinzunigacuellar:

#### Description

- Adds `axios` and `vite` to pnpm installs


delucis (approved):
This makes sense. Thanks @kevinzunigacuellar 🙌

PR #2075, opened 2022-11-20
Fix search result links by reintroducing URL hash

delucis:

#### Description

#2071 introduced a bug where the hashes in links to search results were lost, meaning you’d always just jump to the top of pages. This fixes that!


delucis (commented):
Merging without review as I think this is pretty urgent for weekend users.

sarah11918:
Thanks for this! We'll keep an eye on it throughout the day and update as needed!

PR #2079, opened 2022-11-21
Correct en-only typo in Contentful CMS Guide

reub:
Correct en-only typo in Contentful CMS Guide from "Use can use" to "You can use"

#### Description

This changes a one word typo from "Use" to "You" in the Contentful CMS guide en version.


delucis (approved):
Thanks for spotting that and fixing it @reub!

PR #2088, opened 2022-11-22
Add `class:list` example to styling docs

delucis:

- New or updated content

#### Description

After seeing [this tweet](https://twitter.com/simonswiss/status/1594986667747659777), I was thinking about how we don’t really surface `class:list` outside of reference. There’s one line in [an example in the component docs](https://docs.astro.build/en/core-concepts/astro-components/#the-component-template) but nothing in the styling docs.

Tried adding an example to “Styling & CSS”:

image

Maybe feels a bit like duplication of our reference entry of this and there‘s also an argument this belongs to the component docs rather than styling docs but we’re already showing `is:global` and `define:vars` on this page, so I think this is a reasonable fit.

### DIFF
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -70,6 +70,25 @@ This is a great way to style things like blog posts, or documents with CMS-power
 
 Scoped styles should be used as often as possible. Global styles should be used only as-needed.
 
+### Combining classes with `class:list`
+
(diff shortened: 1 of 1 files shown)


sarah11918:
I think this looks like a helpful addition to the StylES page! We might want to revisit splitting more content out of "Components" in the future anyway. I say let's get this into docs, even if a bit of duplication for now.

PR #2089, opened 2022-11-22
Strategic content tweaks for better search results

delucis:

#### Description

Went through the top 60 search terms for the past 30 days to try to see what might be improved through some strategic content tweaks.

Most of this boils down to:

1. Use terms we want to match as high as possible in the page hierarchy for results we want to show. For example, if a user searches for “style”, make sure “style” is in the page title or `<h2>` for results we want them to see.
2. _Avoid_ terms in the same places if we don’t want that content to rank highly in search. For example, if a user searches for “style” and we don’t want our tutorial page “Add a Gangnam Style translation blog post” to match, we could rename it to “Add a lyrics translation blog post”.

Here are all my tweaks:

- **Avoid “global” in tutorial page title**
    Ranks high for searches for “glob” which we’d like to have better results for.

- **Remove “styled” from tutorial page title**
    Ranks highly for “style” searches and isn’t as helpful as our CSS guide.

- **Use “Styles” instead of “Styling“ in title to match “style” searches**
    People often search for “style” which is not an exact match to “styling”. Changing this should help ranking slightly.

- **Use “styling” instead of “styles” to avoid matching “style” searches**
    Apply the inverse logic to help reduce the ranking of this tutorial page.

- **Avoid “frontmatter” in tutorial `<h2>`**
    “frontmatter” in other pages is always in an `<h3>` so is ranking lower, avoding this here to ensure better results rank higher.

- **Include “SCSS” in Sass heading**
    “SCSS” and “Sass” are often use interchangeably, including both in the subheading should produce a better search result.

- **Avoid “HTML” in tutorial heading**
    Searches for “html” ranked this tutorial page ahead of more helpful guide content

- **Avoid “JavaScript” in tutorial page title**
    More than 100 people searched this recently and our imports guide content is probably the best result.

Happy to tweak any of these further or revert if they seem like overreach.

### DIFF
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -1,7 +1,7 @@
 ---
 layout: ~/layouts/MainLayout.astro
-title: Styling & CSS
-description: Learn how to style components with Astro.
+title: Styles & CSS
+description: Learn how to style components in Astro with scoped styles, external CSS, and tooling like Sass and PostCSS.
 i18nReady: true
 setup: |
   import Since from '../../../components/Since.astro';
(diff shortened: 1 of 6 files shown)


sarah11918:
These all look great! Ship it!

PR #2091, opened 2022-11-22
Tweaks to improve “link” search results

delucis:

#### Description

- Remove “link” from page titles in tutorial lessons
- Add a “Link between pages” subheading to our “Pages” page to help that content stand out

Probably not a silver bullet because “Link between pages” is `<h3>` and there are some `<h2>`s containing “link” in the tutorial, but should be a start. We can re-evaluate if we want to do more after re-crawling.

### DIFF
--- a/src/pages/en/core-concepts/astro-pages.md
+++ b/src/pages/en/core-concepts/astro-pages.md
@@ -24,7 +24,7 @@ Astro leverages a routing strategy called **file-based routing**. Each file in y
 
 ### Link between pages
 
-Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your component template to link between pages.
+Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your Astro components to link to other pages on your site.
 
 ## Astro Pages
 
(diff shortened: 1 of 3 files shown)


src/pages/en/core-concepts/astro-pages.md
```diff
@@ -20,10 +20,12 @@ Astro supports the following file types in the `src/pages/` directory:
 
 Astro leverages a routing strategy called **file-based routing**. Each file in your `src/pages/` directory becomes an endpoint on your site based on its file path.
 
-Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your component template to link between pages.
-
 📚 Read more about [Routing in Astro](/en/core-concepts/routing/).
 
+### Link between pages
+
+Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your Astro components to link to other pages on your site.
```
delucis, line 27: Should this be “pages”? You’d also use `<a>` in e.g. a React component and these are the page docs after all.

```suggestion
Write standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) in your Astro pages to link to other pages on your site.
```

sarah11918: Ooh, good call. I was thinking about that...

PR #2092, opened 2022-11-22
Add Feedback Fish across whole docs site

delucis:

#### Description

- Add the Feedback Fish button to the right sidebar across the whole site
- Make the “Give us feedback” label translatable

Example of location:

image

Worth noting that this sidebar is not visible at narrower viewport widths. We don’t show any of the “extra” (non-ToC) stuff on smaller screens and the approach I took for small screens in the tutorial isn’t an option because that space is used by the current heading text next to “On this page”.

My take is that gathering feedback outside of the tutorial is slightly less crucial, so maybe only showing on wider screens is OK and NWTWWHB? Otherwise, we may need to spend more time thinking. One option could be to move the entire “extra” sidebar stuff to the bottom of each page.

### DIFF
--- a/src/components/RightSidebar/RightSidebar.astro
+++ b/src/components/RightSidebar/RightSidebar.astro
@@ -3,6 +3,7 @@ import TableOfContents from './TableOfContents';
 import ContributeMenu from './ContributeMenu.astro';
 import CommunityMenu from './CommunityMenu.astro';
 import { useTranslations } from '../../i18n/util';
+import FeedbackButton from '../tutorial/FeedbackButton.astro';
 
 const t = useTranslations(Astro);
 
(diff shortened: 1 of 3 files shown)


sarah11918:
I can get behind this! One thing to note is that in "tablet" (medium screen view) we *do* import the "Community" section over into the left sidebar. So in that mode we *could* also consider moving the feedback widget. We don't show the "contribute" items in that view, like Edit this page, because that might be difficult to do at that screen size anyway. But simply leaving feedback I think *could* be an option.

But, non-blocking! I'd be happy to merge this in as-is!

tony-sull:
Yep I think hiding it on mobile makes sense, we can always add that late.  I did try a quick super hacky DOM reorder and the feedback modal looks good on mobile so the question should just be if it makes sense on mobile and where to put the button 👍

delucis:
Thanks for the reviews! I also noticed that due to our pattern of duplicating content for different page areas, on tutorial pages the Feedback Fish resources seem to get loaded twice: 
image
Don’t think that should mean double downloading or even double parsing, but I guess will mean they _run_ twice.

I like Sarah’s suggestion of adding the widget to the left sidebar, so will open a follow-up PR for that that also tweaks the feedback component to avoid the multiple `<script>` tags.

PR #2094, opened 2022-11-22
Add final code check link & confetti to tutorial completion page

delucis:

- New or updated content

#### Description

Updates to the tutorial completion page:

- Add a link to the final version of the tutorial code on GitHub and StackBlitz

- Add confetti on page load to celebrate 🎊

### DIFF
--- a/src/components/tutorial/CompletionConfetti.astro
+++ b/src/components/tutorial/CompletionConfetti.astro
@@ -1,4 +1,6 @@
 <script>
+	import { ProgressStore } from './ProgressStore';
+
 	// Guard code because thanks to `Astro.glob()`, this script will end up on
 	// pages it is not intended to run on.
 	if (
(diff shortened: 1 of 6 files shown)


src/components/tutorial/CompletionConfetti.astro
```diff
@@ -0,0 +1,24 @@
+<script>
+	if (
+		window.location.pathname.endsWith('/tutorial/6-islands/3/') &&
```
tony-sull, line 3: is there an easy way to compute the last tutorial step in case more steps are added later?

tony-sull, line 3: Can the pathname check be left out since it's added as a component directly to the page?

delucis: Ah, I should probably have added a comment explaining this slightly eccentric code — will do that now.

We glob these pages so this script will end up everywhere. Added the check to target a specific page and avoid downloading the confetti lib unnecessarily (hence the dynamic import).

sarah11918:
Obviously I love this! Though, not to be the no fun, stickler here.... but I think we discussed the confetti flying when you completed the final checklist?  😅 

More seriously, for continuity, it maybe seems a little off to be reading, "There's only one edit left to make.." and before you even get that far, confetti explodes in your face!  😄   How difficult would it be to tie it to the last checkmark, or at least wait until you've scrolled to the bottom of the page?

delucis:
Thanks for the idea @sarah11918! The confetti now only shows when you check the final checkbox off. Added bonus: you can uncheck and recheck as much as you want for extra confetti 🎉

sarah11918:
Preview looks great! I think this is an amazing add to this page!

PR #2095, opened 2022-11-22
Show Feedback Fish in left sidebar on narrower screens

delucis:

#### Description

Realised this was easier than I expected!

- Moved the feedback button into the community menu so it moves just like the community menu does as @sarah11918 suggested.

- Tweaked how we load the Feedback Fish script so we only load it once.

One thing to consider: because we’re using the left sidebar, on tutorial pages we now have TWO feedback buttons on smaller screens:

image

Is that an issue? The left sidebar is usually scrolled up so most likely you wouldn’t see both at once.


sarah11918:
Agreed! I'm not bothered by the feedback widget being in two places, for the advantage of getting it in the rest of the docs!  LGTM!

PR #2096, opened 2022-11-23
i18n(ja): Translate CMS guides

mantaroh:

- Translated content

#### Description

- Translate CMS sub pages.


delucis (changes requested):
Just realised I never commented here with the conclusion of our discussions on Discord.

We decided it would make sense to remove the stub translations but keep the longer translations in this PR. This was so that if one of the English stubs gets turned into a full guide, Japanese readers can see that instead of an outdated translated stub. (Great thinking @kyosuke!)

Sorry for the extra work you did @mantaroh! We can remove the stubs from this PR and review the other pages. Thanks again 💜

sarah11918:
Hi @mantaroh! Just following up on this PR to keep it fresh!

Can you please remove the stub translations from this PR as Chris requested above? We will also shout this out for review on Discord! There's no reason why people can't be reviewing the longer, full guide translations here.  :)

mantaroh:
> Hi @mantaroh! Just following up on this PR to keep it fresh!

Yep.
Sorry for bothering with you so long 🙏🏼 . 


> Can you please remove the stub translations from this PR as Chris requested above? We will also shout this out for review on Discord! There's no reason why people can't be reviewing the longer, full guide translations here. :)

You mean that remove the translated page which have `stub:true` like https://github.com/withastro/docs/pull/2096/files#diff-1a9d3abb1a4a5f7a8feb0870165da04f324819311c519ffca8c5b2eda5cfbd94R5 . right?

delucis:
> You mean that remove the translated page which have `stub:true` like [#2096 (files)](https://github.com/withastro/docs/pull/2096/files#diff-1a9d3abb1a4a5f7a8feb0870165da04f324819311c519ffca8c5b2eda5cfbd94R5) . right?

Yes, exactly! Thank you.

delucis (approved):
These look good to me. Thanks again @mantaroh!

Issue #2099, opened 2022-11-23
No hierarchy in table of contents


chenxsan:
I noticed that there's no hierarchy in the list of table of contents.

<img width="502" alt="" src="https://user-images.githubusercontent.com/1091472/203541940-479ff6e8-5e88-43ab-b434-0e84c64fae25.png">

While I expect it to be something like:

```
<ul>
  <li>h2 title</li>
  <li>another h2 title
    <ul>
      <li>h3 title</li>
      <li>another h3 title</li>
    </ul>
  </li>
</ul>
```
Let me know if you're interested in a pull request to update it now that I've finish the code for my own Astro site's table of contents.

delucis:
That sounds helpful, thanks @chenxsan! We’ll need to make sure the code highlighting the current item still works.

PR #2101, opened 2022-11-23
Add heading level hierarchy to table of contents

chenxsan:

1. Refactor the HTML structure of table of contents
2. Since the HTML structure is changed, some changes to css is needed so we can keep the UI same as before

#### Description

- Closes https://github.com/withastro/docs/issues/2099

### DIFF
--- a/src/components/RightSidebar/RightSidebar.astro
+++ b/src/components/RightSidebar/RightSidebar.astro
@@ -3,11 +3,13 @@ import TableOfContents from './TableOfContents';
 import ContributeMenu from './ContributeMenu.astro';
 import CommunityMenu from './CommunityMenu.astro';
 import { useTranslations } from '../../i18n/util';
+import generateToc from '~/util/generateToc';
 
 const t = useTranslations(Astro);
 
 const { content, githubEditUrl } = Astro.props;
 const headings = content.astro?.headings;
+const overview = t('rightSidebar.overview');
 ---
 
 <nav aria-label={t('rightSidebar.a11yTitle')}>
@@ -15,8 +17,8 @@ const headings = content.astro?.headings;
 		headings && (
 			<TableOfContents
 				client:media="(min-width: 50em)"
-				headings={headings}
-				labels={{ onThisPage: t('rightSidebar.onThisPage'), overview: t('rightSidebar.overview') }}
+				toc={generateToc(headings, overview)}
+				labels={{ onThisPage: t('rightSidebar.onThisPage'), overview: overview }}
 				isMobile={false}
 			/>
 		)
--- a/src/components/RightSidebar/RightSidebar.astro
+++ b/src/components/RightSidebar/RightSidebar.astro
@@ -3,7 +3,7 @@ import TableOfContents from './TableOfContents';
 import ContributeMenu from './ContributeMenu.astro';
 import CommunityMenu from './CommunityMenu.astro';
 import { useTranslations } from '../../i18n/util';
-import generateToc from '~/util/generateToc';
+import generateToc from '../../util/generateToc';
 
 const t = useTranslations(Astro);
 
@@ -18,7 +18,7 @@ const overview = t('rightSidebar.overview');
 			<TableOfContents
 				client:media="(min-width: 50em)"
 				toc={generateToc(headings, overview)}
-				labels={{ onThisPage: t('rightSidebar.onThisPage'), overview: overview }}
+				labels={{ onThisPage: t('rightSidebar.onThisPage') }}
 				isMobile={false}
 			/>
 		)
--- a/src/components/RightSidebar/TableOfContents.tsx
+++ b/src/components/RightSidebar/TableOfContents.tsx
@@ -1,15 +1,11 @@
 import { unescape } from 'html-escaper';
 import type { FunctionalComponent } from 'preact';
-import { useState, useEffect, useRef } from 'preact/hooks';
+import type { MarkdownHeading } from 'astro';
+import { useState, useEffect } from 'preact/hooks';
 import './TableOfContents.css';
 
-interface Heading {
-	depth: number;
-	slug: string;
-	text: string;
-}
 interface Props {
-	headings: Heading[];
+	headings: MarkdownHeading[];
 	labels: {
 		onThisPage: string;
 		overview: string;
@@ -17,7 +13,7 @@ interface Props {
 	isMobile?: boolean;
 }
 
-interface TocItem extends Heading {
+interface TocItem extends MarkdownHeading {
 	children: TocItem[];
 }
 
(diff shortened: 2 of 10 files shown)


chenxsan:
Here's the test file I had in my project which can be run with vitest:

```ts
import generateToc from './generateToc'
import { describe, expect, test } from 'vitest'
describe('generateToc', () => {
  test('should throw when heading 3 comes before heading 2', () => {
    expect(() =>
      generateToc([
        {
          slug: 'heading 3',
          text: 'heading 3',
          depth: 3,
        },
        {
          slug: 'heading 2',
          text: 'heading 2',
          depth: 2,
        },
      ])
    ).toThrowError(/^Orphan heading found: heading 2.$/)
  })
  test('should pass when heading 1 comes before heading 2', () => {
    const toc = generateToc([
      {
        slug: 'heading 1',
        text: 'heading 1',
        depth: 1,
      },
      {
        slug: 'heading 2',
        text: 'heading 2',
        depth: 2,
      },
    ])
    expect(toc).toEqual([
      {
        slug: 'heading 1',
        text: 'heading 1',
        depth: 1,
        children: [
          {
            slug: 'heading 2',
            text: 'heading 2',
            depth: 2,
            children: [],
          },
        ],
      },
    ])
  })

  test('should pass when heading 2 comes before heading 2', () => {
    const toc = generateToc([
      {
        slug: 'heading 2',
        text: 'heading 2',
        depth: 2,
      },
      {
        slug: 'heading 2',
        text: 'heading 2',
        depth: 2,
      },
      {
        slug: 'heading 3',
        text: 'heading 3',
        depth: 3,
      },
      {
        slug: 'heading 2',
        text: 'heading 2',
        depth: 2,
      },
    ])
    expect(toc).toEqual([
      {
        slug: 'heading 2',
        text: 'heading 2',
        depth: 2,
        children: [],
      },
      {
        slug: 'heading 2',
        text: 'heading 2',
        depth: 2,
        children: [
          {
            slug: 'heading 3',
            text: 'heading 3',
            depth: 3,
            children: [],
          },
        ],
      },
      {
        slug: 'heading 2',
        text: 'heading 2',
        depth: 2,
        children: [],
      },
    ])
  })
})
```

src/components/RightSidebar/TableOfContents.tsx
```diff
@@ -3,15 +3,67 @@ import type { FunctionalComponent } from 'preact';
...
+	for (let heading of headings) {
+		if (toc.length === 0) {
+			toc.push({
+				...heading,
+				children: [],
+			});
+		} else {
+			const lastItemInToc = toc[toc.length - 1];
+			if (heading.depth < lastItemInToc.depth) {
+				throw new Error(`Orphan heading found: ${heading.text}.`);
```
delucis, line 40: Because this Preact code will run in the browser, it would be good to avoid throwing errors that could break there. One option would be to move all this `generateToc` logic into an Astro component that wraps this one. That way we could safely throw at build time. It would also have the advantage of having all this code run only at build time and never be shipped to users. What do you think?

delucis (commented):
Thanks for this @chenxsan! I’ve left one suggestion.

There’s also one small styling difference in this PR:

| Production site | This PR preview |
|-----------------|-----------------|
| image | image |

The PR preview is missing the `.current-header-link a` styling that makes the text a different shade. Looks like we’ll need to change that in `index.css` to `a.current-header-link`.

src/components/RightSidebar/TableOfContents.tsx
chenxsan: Makes sense. However I have no idea how to make it with another Astro component since we can't use Astro component inside tsx. Wrapping `TableOfContents.tsx` inside Astro component would result some warnings since we pass `client:media` to an Astro component.

Anyway, I've refactor the `TableOfContents.tsx` component to receive a `toc` prop received at build time. Note that since there's no more easy way to get the `currentHeading` from `toc` data, I had to refactor `currentID` state into `currentHeading`.

src/components/RightSidebar/RightSidebar.astro
```diff
@@ -3,20 +3,22 @@ import TableOfContents from './TableOfContents';
 import ContributeMenu from './ContributeMenu.astro';
 import CommunityMenu from './CommunityMenu.astro';
 import { useTranslations } from '../../i18n/util';
+import generateToc from '~/util/generateToc';
```
delucis, line 6: Let’s keep this a relative import path for consistency with the other imports:

```suggestion
import generateToc from '../../util/generateToc';
```

src/components/RightSidebar/TableOfContents.tsx
```diff
@@ -1,22 +1,27 @@
...
+import { useState, useEffect } from 'preact/hooks';
 import './TableOfContents.css';
+import type { TocItem } from '../../util/generateToc';
 
 interface Props {
-	headings: { depth: number; slug: string; text: string }[];
+	toc: TocItem[];
 	labels: {
 		onThisPage: string;
 		overview: string;
```
delucis, line 11: Looks like this label is no longer used right because we’re including it in the generated headings? We can remove it here in that case and also stop passing it in the places `<TableOfContents />` gets used.

```suggestion
```

src/components/RightSidebar/RightSidebar.astro
```diff
@@ -3,20 +3,22 @@ import TableOfContents from './TableOfContents';
...
 
 <nav aria-label={t('rightSidebar.a11yTitle')}>
 	{
 		headings && (
 			<TableOfContents
 				client:media="(min-width: 50em)"
-				headings={headings}
-				labels={{ onThisPage: t('rightSidebar.onThisPage'), overview: t('rightSidebar.overview') }}
+				toc={generateToc(headings, overview)}
+				labels={{ onThisPage: t('rightSidebar.onThisPage'), overview: overview }}
```
delucis, line 21:
```suggestion
				labels={{ onThisPage: t('rightSidebar.onThisPage') }}
```

src/layouts/MainLayout.astro
```diff
@@ -23,10 +25,10 @@ const githubEditUrl = getGithubEditUrl(Astro);
 					<nav>
 						<TableOfContents
 							client:media="(max-width: 72em)"
-							headings={headings}
+							toc={generateToc(headings, overview)}
 							labels={{
 								onThisPage: t('rightSidebar.onThisPage'),
-								overview: t('rightSidebar.overview'),
+								overview: overview,
```
delucis, line 31:
```suggestion
```

delucis (commented):
Awesome! Thanks so much. Just left a few tidy-up suggestions, but then this should be good to go 🚀

delucis (approved):
Thanks for this high quality contribution @chenxsan! Really great work.

By the way, you mentioned you had a test for this ready to go. Is setting up some testing for the docs site something you’d be interested in? We don’t have anything currently but some tests that run in our CI would certainly be helpful. Let me know if you’d be interested in that!

chenxsan:
> Thanks for this high quality contribution @chenxsan! Really great work.
> 
> By the way, you mentioned you had a test for this ready to go. Is setting up some testing for the docs site something you’d be interested in? We don’t have anything currently but some tests that run in our CI would certainly be helpful. Let me know if you’d be interested in that!

Yeah, I had vitest to run that unit test. Since Astro is built on vite, it's sort of seamless to run unit tests with vitest. I can file a pull request if you'd like one.

delucis:
> I can file a pull request if you'd like one.

Sure! That sounds really helpful.

PR #2102, opened 2022-11-23
Add new “Scripts and Event Handling” guide

### LINKED ISSUE #2100: No documentation about Web Components
Though it's straightforward to set up and use web components with Astro, it can be confusing for newcomers to figure out the preferred way to enable and use them.

Thanks to @delucis help on Discord I managed to create a [sample repo](https://github.com/fernandofranca/astro-web-components) with web components, however I think it's an important subject to be added to the docs.

delucis:

- New or updated content

#### Description

- Simplifies the [“Client-side scripts” section](https://deploy-preview-2102--astro-docs-2.netlify.app/en/core-concepts/astro-components/#client-side-scripts) of the Components guide
- Adds a new guide to writing `<scripts>` — https://deploy-preview-2102--astro-docs-2.netlify.app/en/guides/client-side-scripts/

    Includes guidance on:
  - Adding event listeners from a `<script>`
  - Writing a custom element
  - Passing variables from component frontmatter to a script (_without_ `define:vars`)
- Moves detailed content on script bundling and loading out of the Components guide to the new page
- Closes #2100

#### Useful things to answer

How does this feel? It’s a slightly 50-50 page with a reference-y first half and guide-y second half. 

Are the examples clear and helpful? I tried to pick things I could distil down but also chose not to dive too deep into custom element syntax etc. as I think that probably is out of scope for the Astro docs.

Given that, are the _links_ helpful? Can people find what they need to know by following them?

Nav bar position? I feel grouping this with both styling & UI frameworks makes sense. Does SSR need to move up/down?

### DIFF
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -74,7 +74,7 @@ Astro will build, optimize, and add these scripts to the page for you, following
 
 **When to use this:** If your JavaScript file lives inside of `public/` or on a CDN.
 
-This approach skips the JavaScript processing, bundling, and optimizations that are provided by Astro when you import scripts as described above.
+To load scripts outside of your project's `src/` folder, include the `is:inline` directive. This approach skips the JavaScript processing, bundling, and optimizations that are provided by Astro when you import scripts as described above.
 
 ```astro title="src/components/ExternalScripts.astro" "is:inline"
 <!-- absolute path to a script at `public/my-script.js` in your project -->
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -43,6 +43,7 @@ To avoid bundling the script, you can use the `is:inline` directive.
 <script is:inline>
   // Will be rendered into the HTML exactly as written!
   // Local imports are not resolved and will not work.
+  // If in a component, will repeat each time the component is used.
 </script>
 ```
 
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -88,7 +88,9 @@ To load scripts outside of your project's `src/` folder, include the `is:inline`
 
 ### Handle `onclick` and other events
 
-Some UI frameworks use custom syntax for event handling like `onclick={...}` (React/Preact) or `@click="..."` (Vue). Astro components follow web standards and recommend using [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) in a `<script>` tag to handle user interactions.
+Some UI frameworks use custom syntax for event handling like `onClick={...}` (React/Preact) or `@click="..."` (Vue). This is not possible in Astro components, which follow standard HTML syntax more closely.
+
+Instead, you can use [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) in a `<script>` tag to handle user interactions.
 
 ```astro title="src/components/AlertButton.astro"
 <button class="alert">Click me!</button>
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -109,7 +109,7 @@ Instead, you can use [`addEventListener`](https://developer.mozilla.org/en-US/do
 ```
 
 :::note
-If you have multiple `<AlertButton />` components on a page, Astro will not run the component's script multiple times. Scripts are bundled and run only once. Using `querySelectorAll` ensures that this script attaches the event listener to every button with the `alert` class found on the page.
+If you have multiple `<AlertButton />` components on a page, Astro will not run the script multiple times. Scripts are bundled and only included once per page. Using `querySelectorAll` ensures that this script attaches the event listener to every button with the `alert` class found on the page.
 :::
 
 ### Web components with custom elements
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -37,7 +37,7 @@ By default, `<script>` tags are processed by Astro.
 </script>
 ```
 
-To avoid bundling the script, you can use the `is:inline` attribute.
+To avoid bundling the script, you can use the `is:inline` directive.
 
 ```astro title="src/components/InlineScript.astro" "is:inline"
 <script is:inline>
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -1,0 +1,73 @@
+---
+title: Scripts and Event Handling
+description: How to add client-side interactivity to Astro components using native browser JavaScript APIs.
+layout: ~/layouts/MainLayout.astro
+i18nReady: false
+---
+
+You can add client interactivity to your Astro components without [using a UI framework](/en/core-concepts/framework-components/) like React, Svelte, Vue, etc. using standard HTML `<script>` tags. This allows you to send JavaScript to the browser to add functionality to your components.
+
+## Script bundling in Astro
+
+By default, `<script>` tags are processed by Astro.
+
+- Any imports will be bundled, allowing you to import local files or Node modules.
+- The processed script will be injected into your page’s `<head>` with [`type="module"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).
+- TypeScript is fully supported, including importing TypeScript files
+- If your component is used several times on a page, the script tag will only be included once.
+
+```astro title="src/components/Example.astro"
+<script>
+  // Processed! Bundled! TypeScript-supported!
+  // ESM imports work, even to npm packages.
+</script>
+```
+
+To avoid bundling the script, you can use the `is:inline` attribute.
+
+```astro title="src/components/InlineScript.astro" "is:inline"
+<script is:inline>
+  // Will be rendered into the HTML exactly as written!
+  // ESM imports will not be resolved relative to the file.
+</script>
+```
+
+Multiple `<script>` tags can be used in the same `.astro` file using any combination of the methods above.
+
+:::note
+Adding `type="module"` or any other attribute to a `<script>` tag will disable Astro's default bundling behavior, treating the tag as if it had an `is:inline` directive.
+:::
+
+📚 See our [directives reference](/en/reference/directives-reference/#script--style-directives) page for more information about the directives available on `<script>` tags.
+
+## Loading scripts
+
+Sometimes it’s useful to write your scripts as separate `.js`/`.ts` file or you may need to reference an external script on another server. You can do this by referencing these in a `<script>` tag’s `src` attribute.
+
+### Import local scripts
+
+**When to use this:** If your script lives inside of `src/`.
+
+Astro will build, optimize, and add these scripts to the page for you, following its [script bundling rules](#script-bundling-in-astro).
+
+```astro title="src/components/LocalScripts.astro"
+<!-- relative path to script in `src/` -->
+<script src="./local-script.js"></script>
+
+<!-- also works for local TypeScript files -->
+<script src="./script-with-types.ts"></script>
+```
+
+### Load external scripts
+
+**When to use this:** If your JavaScript file lives inside of `public/` or on a CDN.
+
+This approach skips the JavaScript processing, bundling, and optimizations that are provided by Astro when you import scripts as described above.
+
+```astro title="src/components/ExternalScripts.astro"
+<!-- absolute path to a script at `public/my-script.js` in your project -->
+<script is:inline src="/my-script.js"></script>
+
+<!-- full URL to a script on a remote server -->
+<script is:inline src="https://analytics.example.com/script.js"></script>
+```
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -88,7 +88,7 @@ To load scripts outside of your project's `src/` folder, include the `is:inline`
 
 ### Handle `onclick` and other events
 
-Some UI frameworks use custom syntax for event handling like `onClick={...}` (React/Preact) or `@click="..."` (Vue). This is not possible in Astro components, which follow standard HTML syntax more closely.
+Some UI frameworks use custom syntax for event handling like `onClick={...}` (React/Preact) or `@click="..."` (Vue). Astro follows standard HTML more closely and does not use custom syntax for events.
 
 Instead, you can use [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) in a `<script>` tag to handle user interactions.
 
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -147,7 +147,11 @@ In this example, we define a new `<astro-heart>` HTML element that tracks how ma
 </script>
 ```
 
-This is similar to how you might handle user input without a custom element. But, instead of using `document.querySelector()`, we used `this.querySelector()` to get DOM elements. `this.querySelector()` only searches within the current custom element instance, so it’s easier to work with the children of one component instance at a time. Because the browser understands that this code is for a custom element, it will automatically run it for you each time it finds `<astro-heart>` on the page.
+There are two advantages to using a custom element here:
+
+1. You can use `this.querySelector()` to get DOM elements instead of searching the whole page using `document.querySelector()`. `this.querySelector()` only searches within the current custom element instance, making it easier to work with the children of one component instance at a time.
+
+2. The browser will run our custom element’s `constructor()` method each time it finds `<astro-heart>` on the page. This means your code only needs to handle one component at a time instead of setting up all components on the page at once.
 
 📚 You can learn more about custom elements in [web.dev’s Reusable Web Components guide](https://web.dev/custom-elements-v1/) and [MDN’s introduction to custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements).
 
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -149,7 +149,7 @@ In this example, we define a new `<astro-heart>` HTML element that tracks how ma
 
 There are two advantages to using a custom element here:
 
-1. You can use `this.querySelector()` to get DOM elements instead of searching the whole page using `document.querySelector()`. `this.querySelector()` only searches within the current custom element instance, making it easier to work with the children of one component instance at a time.
+1. Instead of searching the whole page using `document.querySelector()`, you can use `this.querySelector()`, which only searches within the current custom element instance. This makes it easier to work with only the children of one component instance at a time.
 
 2. Although a `<script>` only runs once, the browser will run our custom element’s `constructor()` method each time it finds `<astro-heart>` on the page. This means you can safely write code for one component at a time, even if you intend to use this component multiple times on a page.
 
--- a/src/pages/en/guides/client-side-scripts.md
+++ b/src/pages/en/guides/client-side-scripts.md
@@ -5,16 +5,30 @@ layout: ~/layouts/MainLayout.astro
 i18nReady: false
 ---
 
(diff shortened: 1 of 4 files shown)


src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Script loading
+
+You may want to write your scripts as separate `.js`/`.ts` files or need to reference an external script on another server. You can do this by referencing these in a `<script>` tag’s `src` attribute.
+
+#### Import local scripts
+
+**When to use this:** If your script lives inside of `src/`.
+
+Astro will build, optimize, and add these scripts to the page for you, following its [script bundling rules](#script-bundling-in-astro).
```
delucis, line 63:
```suggestion
Astro will build, optimize, and add these scripts to the page for you, following its [script bundling rules](#script-bundling).
```

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+  buttons.forEach((button) => {
+    button.addEventListener('click', () => {
+      alert('Button was clicked!');
+    });
+  });
+</script>
+```
+
+:::note
+Because this script will run once even if we add many instances of `<AlertButton />` to a page, we use `querySelectorAll` to get all instances of our component and add listeners to each one instead of using `querySelector`, which would only return the first button it found.
```
kevinzunigacuellar, line 110: small suggestion, feel free to ignore
```suggestion
Because this script will be shipped only once even if we add many instances of `<AlertButton />` to a page, we use `querySelectorAll` to get all instances of our component and add listeners to each one instead of using `querySelector`, which would only return the first button it found.
```

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+		}
+  }
+
+  // Tell the browser to use our AstroHeart class for <astro-heart> elements.
+  customElements.define('astro-heart', AstroHeart);
+</script>
+```
+
+This is similar to how you might handle user input without a custom element. But, instead of using `document.querySelector()`, we used `this.querySelector()` to get DOM elements. `this.querySelector()` only searches within the current custom element instance, so it’s easier to work with the children of one component instance at a time. Because the browser understands that this code is for a custom element, it will automatically run it for you each time it finds `<astro-heart>` on the page.
+
```
kevinzunigacuellar, line 149: if we were to use this component multiple times, will the javascript be shipped only once, or once per instance?

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+📚 You can learn more about custom elements in [web.dev’s Reusable Web Components guide](https://web.dev/custom-elements-v1/) and [MDN’s introduction to custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements).
+
+### Pass frontmatter variables to scripts
```
kevinzunigacuellar, line 152: can we include the `define:vars` example here too 🙏
```suggestion
### Pass frontmatter variables to scripts with web components
```

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+      });
+		}
+  }
+
+  // Tell the browser to use our AstroHeart class for <astro-heart> elements.
+  customElements.define('astro-heart', AstroHeart);
+</script>
+```
+
+This is similar to how you might handle user input without a custom element. But, instead of using `document.querySelector()`, we used `this.querySelector()` to get DOM elements. `this.querySelector()` only searches within the current custom element instance, so it’s easier to work with the children of one component instance at a time. Because the browser understands that this code is for a custom element, it will automatically run it for you each time it finds `<astro-heart>` on the page.
```
sarah11918, line 148: Flagging as needs work!  😅 

I'm not sure *I* get why this is different than the previous case, other than "your browser understands that this code is for a custom element"... so in that spirit, I'm going to try something (that you will correct later) like...
```suggestion
 Unlike scripts that are bundled and only run once, the script for `<astro-heart>` will be run by the browser each time it finds this custom element on the page. The browser understands that `this.querySelector()` only searches within the current custom element instance to get DOM elements. So, even if you have multiple `<AstroHeart />` components, its script does not have to query for/add event listenters to all instances  of its child elements.
```

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+  const buttons = document.querySelectorAll('button[data-alert]');
+
+  // Handle clicks on each button.
+  buttons.forEach((button) => {
+    button.addEventListener('click', () => {
+      alert('Button was clicked!');
+    });
+  });
+</script>
+```
```
sarah11918, line 107: If this example works with a css `class=""` instead of a `data-` attribute, I'm *tempted* to show that here instead. (Not opposed to what's here, just thinking...)

My thinking: 
- CSS classes might be more familiar, and might already be in use for an element anyway.
- You have a nice introduction/explanation to `data-`  attributes in the "using frontmatter variables" section. So, it seems weird to be using them here, then introducing them at the bottom of the page.

We *could* move that introduction up here,  but I think this is a simple enough (and often searched!) example that we don't need to clutter what's not very complicated here.

In which case I might even add to the line above with something like:

> You can add [`addEventListeners`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) to elements identified by class, id, attribute  etc. in a `<script>` tag to handle user interactions.

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+
+- Any imports will be bundled, allowing you to import local files or Node modules.
+- The processed script will be injected into your page’s `<head>` with [`type="module"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).
+- TypeScript is fully supported, including importing TypeScript files
+- If your component is used several times on a page, the script will only be included once.
+
+```astro title="src/components/Example.astro"
+<script>
+  // Processed! Bundled! TypeScript-supported!
+  // ESM imports work, even to npm packages.
```
sarah11918, line 36: Flagging that this, in conjunction with/as a contrast to the inline script, needs some work.

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+</script>
+```
+
+To avoid bundling the script, you can use the `is:inline` attribute.
+
+```astro title="src/components/InlineScript.astro" "is:inline"
+<script is:inline>
+  // Will be rendered into the HTML exactly as written!
+  // ESM imports will not be resolved relative to the file.
```
sarah11918, line 45: Reminding us that we are not sold on this!

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+
+📚 You can learn more about custom elements in [web.dev’s Reusable Web Components guide](https://web.dev/custom-elements-v1/) and [MDN’s introduction to custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements).
+
+### Pass frontmatter variables to scripts
+
+In Astro components, the code in [the frontmatter](/en/core-concepts/astro-components/#the-component-script) between the `---` fences runs on the server and is not available in the browser. To send variables from the server to the client, we need a way to store our variables and then read them when JavaScript runs in the browser.
+
+One way to do this is to use [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) to store the value of variables in your HTML output. Scripts, including custom elements, can then read these once your HTML loads in the browser.
+
+In this example component, a `message` prop is stored in a `data-message` attribute. Data attributes are available in JavaScript using an element’s `dataset` property, so the custom element logic can read `this.dataset.message` and get the value of the prop in the browser.
```
sarah11918, line 158:
```suggestion
One way to do this is to use [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) to store the value of variables in your HTML output. Scripts, including custom elements, can then read `this.dataset.*` once your HTML loads in the browser.

In this example component, a `message` prop is stored in a `data-message` attribute and is available in the browser at `this.dataset.message`.
```
Is this true? It's more concise.  😄

sarah11918 (commented):
This is an amazing page, @delucis!  🚀 I can't wait for it to go live!

I think there might only be 1 or 2 spots to really revisit and nail the explanations, but I've made some possible editing suggestions that are potentially better, if true! You'll see this before I will tomorrow, so some time for you to digest, correct me, and then we can tackle it tomorrow!

src/pages/en/guides/client-side-scripts.md
delucis:
```suggestion
  // Importing local scripts and Node modules works.
```

src/pages/en/guides/client-side-scripts.md
delucis:
```suggestion
  // Local imports are not resolved and will not work.
```

delucis: Another option could even be just to remove this line wholesale.

We’ve said “rendered as written” and that the script isn’t bundled, so might be sufficient.

```suggestion
```

src/pages/en/guides/client-side-scripts.md
delucis: Agree — `data-*` attributes are often recommended for this because they help you separate styling and behaviour, but `class` examples are more common, and showing the data attribute later means we can cover both bases.

````suggestion
```astro title="src/components/AlertButton.astro"
<button class="alert">Click me!</button>

<script>
  // Find all buttons with the class `alert` attribute on the page.
  const buttons = document.querySelectorAll('button.alert');

  // Handle clicks on each button.
  buttons.forEach((button) => {
    button.addEventListener('click', () => {
      alert('Button was clicked!');
    });
  });
</script>
```
````

File: src/pages/en/guides/client-side-scripts.md (hunk 2 above)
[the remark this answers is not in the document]
Comment (delucis) [reply src/pages/en/guides/client-side-scripts.md:110]: Not sure how different these are in messaging, but I do like your wording better. Tweaked ever so slightly:

```suggestion
If you have multiple `<AlertButton />` components on a page, Astro will not run the component's script multiple times. Scripts are bundled and run only once. Using `querySelectorAll` ensures that this script attaches the event listener to every button with the `alert` class found on the page.
```

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+</script>
+```
+
+:::note
+Because this script will run once even if we add many instances of `<AlertButton />` to a page, we use `querySelectorAll` to get all instances of our component and add listeners to each one instead of using `querySelector`, which would only return the first button it found.
+:::
+
+### Web components with custom elements
+
+[Custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) provide a standard API that allows you to define your own HTML elements with custom behavior. Defining a custom element in a `.astro` component can be a good way to build interactivity without reaching for a UI framework library.
```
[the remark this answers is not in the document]
delucis: 👍 just removing a stray space:

```suggestion
You can create your own HTML elements with custom behavior using the Web Components standard. Defining a [custom element](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) in a `.astro` component allows you to build interactive components without needing a UI framework library.
```

src/pages/en/guides/client-side-scripts.md
delucis: It’s not really true, no 😆 

An element’s data attributes are available in the `dataset` object _for that element_.

So, for example:

```html
<p data-marquee>Thanksgiving Offer: chickens going cheep!</p>

<script>
const p = document.querySelector('p');
// p.dataset has a marquee property
p.dataset.marquee;
</script>
```

In our case, we can use `this.dataset` because `this` is the root node in a custom element, but `this.dataset` is not simply “available in the browser.”

Would this be an improvement do you think?

```suggestion
One way to do this is to use [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) to store the value of variables in your HTML output. Scripts, including custom elements, can then read these using an element’s `dataset` property once your HTML loads in the browser.

In this example component, a `message` prop is stored in a `data-message` attribute, so the custom element can read `this.dataset.message` and get the value of the prop in the browser.
```

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+
+```astro title="src/components/InlineScript.astro" "is:inline"
+<script is:inline>
+  // Will be rendered into the HTML exactly as written!
+  // Local imports are not resolved and will not work.
+</script>
+```
+
+:::note
+Adding `type="module"` or any other attribute to a `<script>` tag will disable Astro's default bundling behavior, treating the tag as if it had an `is:inline` directive.
```
delucis, line 50: You _can_ add `src` and take advantage of bundling (as we show below) so tweaking this:

```suggestion
Adding `type="module"` or any attribute other than `src` to a `<script>` tag will disable Astro's default bundling behavior, treating the tag as if it had an `is:inline` directive.
```

src/pages/en/guides/client-side-scripts.md
delucis: Yeah — I wasn’t happy with this either. Here’s a second attempt.

```suggestion
There are two advantages to using a custom element here:

1. You can use `this.querySelector()` to get DOM elements instead of searching the whole page using `document.querySelector()`. `this.querySelector()` only searches within the current custom element instance, making it easier to work with the children of one component instance at a time.

2. The browser will run our custom element’s `constructor()` method each time it finds `<astro-heart>` on the page. This means your code only needs to handle one component at a time instead of setting up all components on the page at once.
```

Will commit this so I can see it better in context and await feedback.

src/pages/en/guides/client-side-scripts.md
delucis: Oops, this got overridden by Sarah’s rewrite here. The new wording is “Scripts are bundled and run only once.” Do you think “bundled” is enough to convey the same concept as “shipped” here?

src/pages/en/guides/client-side-scripts.md
delucis: Selfishly going to commit my variant so I can reread the full thing.

src/pages/en/guides/client-side-scripts.md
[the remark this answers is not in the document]
sarah11918: Haha... my space bar was stiiiiiicking last night! It was a struggle to fight against either no spaces, or spaces appearing before my eyes. Thanks for the catch!

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,205 @@
...
+
+<!-- full URL to a script on a remote server -->
+<script is:inline src="https://analytics.example.com/script.js"></script>
+```
+
+## Common script patterns
+
+### Handle `onclick` and other events
+
+Some UI frameworks use custom syntax for event handling like `onclick={...}` (React/Preact) or `@click="..."` (Vue). Astro components follow web standards and recommend using [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) in a `<script>` tag to handle user interactions.
```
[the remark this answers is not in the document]
delucis: Here’s what I went with for now:

> Some UI frameworks use custom syntax for event handling like `onClick={...}` (React/Preact) or `@click="..."` (Vue). This is not possible in Astro components, which follow standard HTML syntax more closely.
> 
> Instead, you can use [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) in a `<script>` tag to handle user interactions.

Do you think that’s OK?

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  // Tell the browser to use our AstroHeart class for <astro-heart> elements.
+  customElements.define('astro-heart', AstroHeart);
+</script>
+```
+
+There are two advantages to using a custom element here:
+
+1. You can use `this.querySelector()` to get DOM elements instead of searching the whole page using `document.querySelector()`. `this.querySelector()` only searches within the current custom element instance, making it easier to work with the children of one component instance at a time.
```
sarah11918, line 152:
```suggestion
1. You can use `this.querySelector()` to get DOM elements instead of searching the whole page using `document.querySelector()`. `this.querySelector()` only searches within the current custom element instance, making it easier to target the children of one component instance at a time.
```
?? I feel like this might relate back to "querying" a bit better? Just a thought.

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,211 @@
...
+
+2. The browser will run our custom element’s `constructor()` method each time it finds `<astro-heart>` on the page. This means your code only needs to handle one component at a time instead of setting up all components on the page at once.
```
sarah11918, line 154:
```suggestion
2. Although a `<script>` is only run once, the browser will run our custom element’s `constructor()` method each time it finds `<astro-heart>` on the page. This means you can safely write code for one component at a time, even if you intend to use this components multiple times on a page.
```
?? I do like setting up a callback/contrast to before when we said you had to write for multiple instances, so whatever the correct opening clause is, I think one would work here.

Another possible take...
> Unlike a script that is bundled and only run once, any custom element's `constructor()` method will be run ...

Also, since this is framed as an "advantage", we could consider something a little more direct re: the code author "YOU CAN DO THIS..."

For your consideration!

delucis: Excellent! Tiny tweaks before committing:

```suggestion
2. Although a `<script>` only runs once, the browser will run our custom element’s `constructor()` method each time it finds `<astro-heart>` on the page. This means you can safely write code for one component at a time, even if you intend to use this component multiple times on a page.
```

src/pages/en/guides/client-side-scripts.md
delucis: Is “target“ verging on jargon? I could go either way but leaning “work with” as descriptive of the fact you want to _do_ things with them.

sarah11918: I was updating this comment as you were replying! I can go back to "work with"

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,211 @@
...
+<script is:inline src="https://my-analytics.com/script.js"></script>
+```
+
+## Common script patterns
+
+### Handle `onclick` and other events
+
+Some UI frameworks use custom syntax for event handling like `onClick={...}` (React/Preact) or `@click="..."` (Vue). This is not possible in Astro components, which follow standard HTML syntax more closely.
+
+Instead, you can use [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) in a `<script>` tag to handle user interactions.
```
delucis, line 93:
```suggestion
Some UI frameworks use custom syntax for event handling like `onClick={...}` (React/Preact) or `@click="..."` (Vue). Astro follows standard HTML more closely and does not use custom syntax for events.

Instead, you can use [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) in a `<script>` tag to handle user interactions.
```

sarah11918: I can get behind this! LGTM!

src/pages/en/guides/client-side-scripts.md
```diff
@@ -0,0 +1,211 @@
...
+  buttons.forEach((button) => {
+    button.addEventListener('click', () => {
+      alert('Button was clicked!');
+    });
+  });
+</script>
+```
+
+:::note
+If you have multiple `<AlertButton />` components on a page, Astro will not run the component's script multiple times. Scripts are bundled and run only once. Using `querySelectorAll` ensures that this script attaches the event listener to every button with the `alert` class found on the page.
```
delucis, line 112: Slight rework to try to address Kevin’s comment.

```suggestion
If you have multiple `<AlertButton />` components on a page, Astro will not run the script multiple times. Scripts are bundled and only included once per page. Using `querySelectorAll` ensures that this script attaches the event listener to every button with the `alert` class found on the page.
```

sarah11918: Approve!

sarah11918:
LGTM!

And so begins the downfall of "Components"... (*everything's a component*)

PR #2105, opened 2022-11-24
Storyblok guide: only use preview token

kevinzunigacuellar:

- New or updated content

#### Description

- There have been some problems loading environmental variables on the config file in this guide ([Discord](https://discord.com/channels/830184174198718474/1002802280267001858/1044989082876510300)) This PR removes the use of `import.meta.env` in the config file and only uses one variable `STORYBLOK_TOKEN` for data fetching.
- Adds vite and axios to npm and yarn commands
- Minimal reproduction [Stackblitz](https://stackblitz.com/edit/github-d3xnpe?file=README.md)


src/pages/en/guides/cms/storyblok.md
```diff
@@ -22,17 +22,14 @@ To get started, you will need to have the following:
 
 2. **A Storyblok account and space** - If you don't have an account yet, [sign up for free](https://app.storyblok.com/#/signup) and create a new space.
 
-3. **Storyblok API tokens** - You can find and generate your API tokens in the Access Tokens tab of your Storyblok space settings. 
-    - **Preview token** - This token will be used in development to fetch drafts or unpublished versions of your content.
-    - **Public token** - This token will be used in production builds to fetch only published content.
+3. **Storyblok Preview token** - This token will be used to fetch drafts and published versions of your content. You can find and generate your API token in the Access Tokens tab of your Storyblok space settings.
```
delucis, line 25: If we always use the preview token, can users somehow filter out their drafts in production when they fetch that data?

delucis (commented):
Thanks @kevinzunigacuellar! Left one comment about how we might want to adjust this guide if we’re switching this up. Also posted a hackier solution on GitHub, but that’s probably something better addressed on Storyblok’s end.

src/pages/en/guides/cms/storyblok.md
kevinzunigacuellar: Yes, the preview token allows filtering for published and draft content by adding the query params `version=draft/published` which can be used in conjunction with `import.meta.env.DEV` outside of the config file 🎉 ([docs](https://www.storyblok.com/docs/api/content-delivery#topics/introduction)).

delucis: Sorry I missed this @kevinzunigacuellar. That’s great!

Do you think we should add a little mention of that somewhere in that case?

> To hide draft posts in production add the `version=published` query param to this request: …example…

kevinzunigacuellar: @delucis Yes, I have included this `query` param in the code snippets for both the integration setup and the tutorial. I mentioned what it does in the tutorial but not in the integration set up. I think it would be okay to repeat this.

delucis: D’oh! Didn’t actually check you hadn’t already done this, sorry @kevinzunigacuellar!

In that case, I think this PR LGTM!

delucis (approved):

src/pages/en/guides/cms/storyblok.md
kevinzunigacuellar: To be quite honest, I have no idea why I included it. I seem like a good idea at the time. Happy it came in handy.

PR #2109, opened 2022-11-25
add env.d.ts to project directory tree in manual install

sarah11918:
- New or updated content

The example file structure shown as the last step in the manual install instructions was missing `env.d.ts`. This adds that.


delucis (commented):
Good catch! We should also update the tree formatting to match I think.

(On my phone so sorry for the line-by-line suggestions.)

PR #2111, opened 2022-11-27
Prevent TOC IntersectionObserver from being created on every render, closes #2110

### LINKED ISSUE #2110: SIGSEGV while recording performance with 4x CPU slowdown
[Screencast from 2022-11-27 02-27-12.webm](https://user-images.githubusercontent.com/37726261/204120757-65b7b17d-3054-4810-9962-1d59a7ae0ef8.webm)

This issue is probably caused by Preact firing the useEffect that creates the TOC IntersectionObserver in every render: https://github.com/withastro/docs/blob/e96c0ca152cf93f49ed46e2be16c6867a61781cb/src/components/RightSidebar/TableOfContents.tsx#L85
Removing the ref from the dependency array should fix the issue.

wtchnm:

#### Description

- Closes #2110


delucis (approved):
Thanks for fixing this @wtchnm! This change makes sense to me 🙌

PR #2115, opened 2022-11-29
Note scoped classes passed to child components

bluwy:

- New or updated content

#### Description

Close https://github.com/withastro/astro/issues/5112

Clarify that the issue is expected behaviour and how to prevent it if it happens.

Question: Should we show the user how to filter out the scoped class? It might involve regex replace that maybe even make sense to export out from Astro as an API, but it also feels like a bandaid API to the problem. I'm slightly leaning to not for now unless more people hit this edge case.

### DIFF
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -143,7 +143,7 @@ import MyComponent from "../components/MyComponent.astro"
 This pattern lets you style child components directly. Astro will pass the parent’s scoped class name (e.g. `astro-HHNQFKH6`) through the `class` prop automatically, including the child in its parent’s scope.
 
 :::note[Scoped classes from parent components]
-As the `class` prop contains scoped classes from the parent, it's recommended to create unique class names in the child component to prevent unintended style cascading from the parent component.
+Because the `class` prop includes the child in its parent’s scope, it is possible for styles to cascade from parent to child. To avoid this having unintended side effects, ensure you use unique class names in the child component.
 :::
 
 
--- a/src/pages/en/guides/styling.md
+++ b/src/pages/en/guides/styling.md
@@ -132,7 +132,7 @@ const { class: className } = Astro.props;
 ---
 import MyComponent from "../components/MyComponent.astro"
 ---
-<style is:global>
+<style>
   .red {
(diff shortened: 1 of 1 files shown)


delucis:
Thanks @bluwy! I’m wondering if we can flip the framing to view this as a feature rather than an unhelpful behaviour you need to work around.

From my understanding, a parent component passes its `astro-????` class to a child component _if_ you set the `class` attribute on the child component. So in this example `div:where(astro-XXXX) > :where(astro-XXXX)` will also apply to the contents of `<Child />` because it will have `class="foo astro-XXXX"` (assuming `<Child />` is wired up to pass props to whichever elements need them):

```astro
---
import Child from './Child.astro';
---
<div>
  <Child class="foo" />
</div>

<style>
  div > * {
    margin: 1rem;
  }
</style>
```

This could be seen as an alternative to writing `div > :global(*)` to style child components that could be more selective. What’s the most minimal set up that would support this. Is `<Child class="" />` enough? Or even `<Child class />`?

bluwy:
> From my understanding, a parent component passes its `astro-????` class to a child component _if_ you set the `class` attribute on the child component.

I believe `astro-???` is passed regardless of if the component has the class attribute, according to the linked issue's repro. I kinda prefer `div > :global(*)` though because it's explicit that you're doing non-scope-safe styling, but also maybe that's because I've used Svelte for so long 😄 

Thinking about this more, it seems more complex that I thought. I'll write up some notes on the maintainers chat to confirm how we should handle this in the future and potentially for v2.

bluwy:
Based on yesterday's discussion, I think we could frame this feature in a way. We could say `div > *` works because Astro attaches the hashed class to the Component's `class` attribute. In preparation for the potential of `Astro.hash`, we can update this to say "if you relied on this feature, it now needs to be opt-in using `Astro.hash`".

I also think we should have a small warning still that if the Component passes down the hashed class, it needs to be careful to keep the selectors unique.

I'll make some changes but interested in what you think of the plan too!

sarah11918:
Just adding my approval here! Great work!

delucis (approved):
LGTM! (Let’s get those scoped classes!)

PR #2120, opened 2022-11-29
Generate custom anchors for errors so we can refer to them statically

Princesseuh:

#### Description

Use a remark plugin to generate custom ids for the headings in the error reference so we can refer to them by their name instead of the title. 

This has the following benefits:
- Links will never change
- Which mean, we can link to errors from other Astro projects without worrying about links breaking if the title change
- Links to specific errors are independent on the language, so it'll be easier to link to them in the user's language in the future (well the path still changes, but we could hardcode in other projects that one if needed)


delucis (approved):
Anchors away! ⚓ 

Looks great, thanks @Princesseuh!

PR #2122, opened 2022-11-30
Add vitest unit tests

chenxsan:

- Add Vitest to run unit tests
- Include a unit test for [`generateToc`](https://github.com/withastro/docs/blob/main/src/util/generateToc.ts) introduced in https://github.com/withastro/docs/pull/2101
- Enable CI for unit tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,3 +48,15 @@ jobs:
 
       - name: Run Check
         run: pnpm run lint:linkcheck
+  unitTest:
+    name: Check for code issues with Vitest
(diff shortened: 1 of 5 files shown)


dsomel21:
Neat! What's so cool about vi-test?

chenxsan:
> Neat! What's so cool about vi-test?

It's built by the same core members of Vite, which Astro is built upon, and out of the box TypeScript support. That's why I choose it over other test frameworks.

delucis (approved):
Thanks @chenxsan — this looks great 🙌

PR #2128, opened 2022-11-30
 Add linter check that fails for nondescript link labels

### LINKED ISSUE #2123: Descriptive text link checker
Just like we have link checkers for the url's of links in PRs submitted, after watching a stream on Docs Accessibility, I'd like to see if we can add another check for link *text* to at least catch the low-hanging fruit of the most common non-descriptive link text phrases.

I'd like to extend our link checks to at least fail if any link has the text "here" or "read more." Maybe others have some good generic phrases to include?

I know docs aren't in "danger" because link text is always evaluated and offending text is caught during reviews. But, I think it's a great model for docs, and anytime we can point to something we do that's *thoughtful*, I think that's a huge win. And we do get so many new contributors all the time, it may end up being helpful in gently onboarding new docs peeps!

delucis:

#### Description

- Closes #2123 
- Adds a check to our link linter that fails if a link’s text is `read more`, `here` or `more`
- Fixes two links in the Buddy deploy guide that were just labelled “here”
- Up for including other tests here if we can think of any, but this has already caught a couple of issues!

### DIFF
--- a/scripts/lib/linkcheck/checks/good-link-label.ts
+++ b/scripts/lib/linkcheck/checks/good-link-label.ts
@@ -1,0 +1,38 @@
+import { DomUtils } from 'htmlparser2';
+import kleur from 'kleur';
+import { dedentMd } from '../../output.mjs';
+import { CheckBase, CheckHtmlPageContext } from '../base/check';
+import { IssueType } from '../base/issue';
+
+/** List of labels that are insufficiently descriptive for a link. */
+const blocklist = new Set(['read more', 'here', 'more']);
+
+export class GoodLabels extends CheckBase {
+	private static readonly BadLabel = new IssueType({
+		title: 'link(s) with vague or nondescript labels',
+		prefix: kleur.gray(`[${kleur.yellow().bold('lbl')}]`),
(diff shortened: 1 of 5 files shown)


sarah11918:
I am so excited by this!!  And it's already working, so.. LGTM!

scripts/lib/linkcheck/checks/good-link-label.ts
```diff
@@ -0,0 +1,38 @@
+import { DomUtils } from 'htmlparser2';
+import kleur from 'kleur';
+import { dedentMd } from '../../output.mjs';
+import { CheckBase, CheckHtmlPageContext } from '../base/check';
+import { IssueType } from '../base/issue';
+
+/** List of labels that are insufficiently descriptive for a link. */
+const blocklist = new Set(['read more', 'here', 'more']);
```
BenDMyers, line 8: **suggestion:** Add `click here` as an option? That's a pretty trope-y generic link text.

delucis: Great suggestion! Thanks for taking the time to review @BenDMyers 💜

PR #2130, opened 2022-12-01
Separate error reference in multiple pages

Princesseuh:

- Something else!

#### Description

@FredKSchott thought the error reference page was too heavy to read with all the errors and as such requested that I separate it in multiple page. This will also allow for analytics to be collected regarding which pages are the most visited without needing any code changes


sarah11918:
Gah! Github ate my comment... guess I get to make it shorter now  😅 

This does look way more readable! I like the error name, code and description all on this page, since someone could be searching any one of those. I think this is the perfect division of what's all on the main page, vs what goes on individual pages.

1. I'll leave it to Chris to weigh in whether something like a "Back to all Issues" link, or any other layout on the individual error pages makes sense. I kind of like it, but I do recognize users have a back button. Anything else?

2. I'm leaning towards (and willing to be told why I'm wrong) the links being the error **codes** and not the description. I can list all the reasons why  (again, thanks Github!), but instead, I'll just float it and y'all can let me know what you think.

3. I'm less sure, and wondering aloud, whether having the slug also be the error code makes sense. I might be the only one who tends to type in exact URLs if I know them 😅 , but there's someting *very* satisfying to me about that URL pattern.

Princesseuh:
> 3. I'm less sure, and wondering aloud, whether having the slug also be the error code makes sense. I might be the only one who tends to type in exact URLs if I know them 😅 , but there's someting _very_ satisfying to me about that URL pattern.

Something to note about the URLs is that we need to be able to link to them from other projects, by making them the error slug, it's a very easy 1:1 link that works in all languages. Anything else will require significantly more work to do (but is of course possible if absolutely needed)

FredKSchott:
Thanks @Princesseuh! This meets all of my goals, so LGTM

sarah11918:
> Something to note about the URLs is that we need to be able to link to them from other projects, by making them the error slug, it's a very easy 1:1 link that works in all languages. Anything else will require significantly more work to do (but is of course possible if absolutely needed)

Good enough for me on point 3, then!

Still leaning toward the error code itself be the thing that is linked, I think, rather than the description of the error. I think that's better visually, and for screen readers? Would be happy to hear thoughts for/against.

And, the first point, about what each individual error page looks like, is non-blocking and can always be revisited if we decide this page should have anything like a "Back to all errors" link, or "Visit the troubleshooting guide if you're still having problems."

delucis (commented):
This is great! I like how clean the index and individual pages are now and the analytics data will definitely be useful.

I’m a bit worried about terms in page titles ranking high in search, but we can tackle that when we get there if we see it’s an issue.

I think I agree with Sarah re the URL, e.g. `/en/reference/errors/03001/` instead of `/en/reference/errors/staticredirectnotavailable/`. Usually I’d say a URL with meaningful words in makes sense, but in this case where it’s long and loses its PascalCase, I’m less convinced. OR I think I’d be happy with converting to kebab case too, e.g. `/en/reference/errors/static-redirect-not-available/`.

I like the idea of a “See all errors” link in the footer but don’t think it’s essential to getting this PR merged.

I do see one more issue: we may need to tweak our left sidebar logic as it currently doesn’t know the individual error pages are part of reference so defaults to the “Learn” tab for those, but should probably default to the “Reference” tab. Let me know if you need a hand in tracking down where that’s happening.

sarah11918:
And @delucis - separate from the href being e.g. `/errors/03005`, what would you think about the link text *itself* being the Error code: e.g.

```
OnlyResponseCanBeReturned ([E03005](/en/reference/errors/03005))
Invalid type returned by Astro page.
```

I think this will look much cleaner than having lines of text linked. And, something discussed in the accessibility stream, if you are navigating by voice, you have to read out the exact link to the device in order to navigate there. So, I think it's easier to say "03005". (There are other things I like about this, but those are two good ones I think.) What do you think?

delucis:
Ah, I’m not sure about that one. Some assistive tech will give you a list of on-page links, so that one would give you a list that contains “E03001, E03002, E03003…” which does not seem very helpful. I think descriptive link text is probably preferable here, but wish I knew more to say that more conclusively.

FredKSchott:
> I’m a bit worried about terms in page titles ranking high in search, but we can tackle that when we get there if we see it’s an issue.

Do I remember correctly and Algolia would let us have an entire "Errors" section of the search results? Because I could definitely see the value in that. Not a blocking comment, agreed that we can play with this more post merge.

sarah11918:
See, I was thinking the error code WOULD be more helpful for exactly that reason, because if you were on this page, it's because you probably got a particular error code. And, rather than keeping the exact error message in your head (and contrasting it with similarly worded things), you could just remember the error code itself. The long descriptions might be more difficult to remember/distinguish than "Oh yeah, it's error code 03005"?

delucis:
> Do I remember correctly and Algolia would let us have an entire "Errors" section of the search results?

Yes. And we can use `pageRank` to nudge results. Tutorial pages are currently ranked `-100` which Algolia combines with other signals to decide what to display. Could try some more extreme values to see if it would always force pages in the error category to the bottom of results. I’d like to evaluate once this is merged and indexed before jumping to conclusions though.

Princesseuh:
> See, I was thinking the error code WOULD be more helpful for exactly that reason, because if you were on this page, it's because you probably got a particular error code. And, rather than keeping the exact error message in your head (and contrasting it with similarly worded things), you could just remember the error code itself. The long descriptions might be more difficult to remember/distinguish than "Oh yeah, it's error code 03005"?

It's best not to put too much emphasis on error codes, because they're actually not shown in the overlay. Right now the only place you can actually see error codes is inside the editor, and even then we're not 100% sure about keeping them

In the new overlay, the static information is the error's name, ex: `FailedToLoadModuleSSR`

image

sarah11918:
Ah, if error codes aren't prominent in the error overlay, then I totally agree - they're not going to be the thing that sticks in your head!

I still do think that linking the description is visually a little much? What does everyone think, then about linking the error keyword itself? Does this look nicer? (And, in this case the link is actually the slug of the URL it leads to, so there is something nice about that.)

```
[OnlyResponseCanBeReturned](/en/reference/errors/onlyresponsecanbereturned) (E03005)
Invalid type returned by Astro page.
```

Otherwise, I think the remaining issues are:
- Categorizing so that Error messages show up in their own category in Algolia search (I think this would be helpful.)
- Making sure that when you're on an individual error message page, the "reference" section of the sidebar is shown, not "Learn" so that you "know where you are" and can get back to the Error messages page easily.

Other thoughts?

src/util/getPageCategory.ts
```diff
@@ -5,6 +5,7 @@ const categories = [
 	['reference/', 'Reference'],
 	['guides/deploy/', 'Deploy Guides'],
 	['guides/integrations-guide/', 'Integration Guides'],
+	['reference/errors/', 'Reference'],
```
delucis, line 8: I might suggest making this its own contained category. This is only for search purposes and I think there it could be helpful to separate errors from general reference material:
```suggestion
	['reference/errors/', 'Error Reference'],
```

delucis (commented):
Thanks for the response @Princesseuh!

I do still think kebab-case URLs could be nice, especially as it avoids the casing difference between PascalCase filename and lowercase URL — but I’m generally happy to defer to you.

I’m personally not convinced by making the links the less-readable error name instead of the message — and don’t think you’d gain much visually as the difference in length between name and message is not enormous:

| Currently | Alternative |
|---|---|
| image | image |

sarah11918:
Of those two images, I do prefer the "alternative" fwiw, but non-blocking!

It's a go from me when Chris is happy with the structure of everything!

delucis:
I think the last outstanding point for discussion is the if URLs should be kebab case or not?

I’m a bit nervous about the file names (and links) being a different case from the URLs (and Netlify forces everything to lowercase even though we’re not authoring that way), so kebab case would solve that, but happy to hear arguments against.

Princesseuh:
I moved the URLs to kebab-case, the Netlify problem convinced me! I also moved the links to the name like @sarah11918 suggested, after thinking about it I think it makes more sense than being on the title below

sarah11918:
Both updates look amazing to me! (I really do like the links this way much better.  😅 )

Sarah approves!

delucis:
Nice — thanks Erika! I'm also remembering now where this bit us before: we had `pt-BR`,`zh-CN` etc. as our locales (which is the standard format for BCP language tags) which meant sitemap, canonicals etc. all included the uppercase BUT because Netlify lowercases everything, that meant none of the pages Google and Algolia crawled were ever indexed — because the

PR #2131, opened 2022-12-01
Review Storyblok page

### LINKED ISSUE #2127: Storyblok tutorial: not having a 404 page for SSR causes infinite loop
In the Storyblok [blog turorial for SSR](https://docs.astro.build/en/guides/cms/storyblok/#server-side-rendering). If the page request fails the user will be redirected to a 404. However, the guide doesn't suggest creating one and not having one causes an infinite loop.
Screenshot 2022-11-30 at 13 41 36

kevinzunigacuellar:

- New or updated content

#### Description

- Complete review of the page ✨
- Closes #2127

### DIFF
--- a/src/pages/en/guides/cms/storyblok.md
+++ b/src/pages/en/guides/cms/storyblok.md
@@ -8,11 +8,11 @@ setup: |
 stub: false
 ---
 
-[Storyblok](https://www.storyblok.com/) is a component based headless CMS that allows you to manage your content by creating reusable components called Bloks.
+[Storyblok](https://www.storyblok.com/) is a component based headless CMS that allows you to manage your content using reusable components called Bloks.
 
 ## Integrating with Astro
 
-In this section, we will use the [Storyblok integration](https://github.com/storyblok/storyblok-astro) to connect Storyblok to Astro.
+In this section, you will use the [Storyblok integration](https://github.com/storyblok/storyblok-astro) to connect Storyblok to Astro.
 
 ### Prerequisites
 
@@ -45,7 +45,7 @@ Your root directory should now include this new file:
 
 ### Installing dependencies
 
-To connect Astro with your Storyblok space, install the official [Storyblok integration](https://github.com/storyblok/storyblok-astro) using the single command below for your preferred package manager:
+To connect Astro with your Storyblok space, install the official [Storyblok integration](https://github.com/storyblok/storyblok-astro) using command below for your preferred package manager:
 
 <PackageManagerTabs>
   <Fragment slot="npm">
@@ -94,29 +94,29 @@ export default defineConfig({
 
 The Storyblok integration requires an object with the following properties:
 
-1. `accessToken` - This references the Storyblok API tokens that we added in the previous step.
+1. `accessToken` - This references the Storyblok API token that you added in the previous step.
 
     :::tip
     Since the Astro config file does not normally support environment variables, use the `loadEnv` function from Vite to load them.
     :::
 
-2. `components` - An object that maps Storyblok component names to paths to your local components. This is required to render your Storyblok components in Astro.
+2. `components` - An object that maps Storyblok component names to paths to your local components. This is required to render your Storyblok Bloks in Astro.
 
     :::note
-    The component paths are relative to the `src` directory.
+    The component paths are relative to the `src` directory. For example, if your component is located at `src/storyblok/MyComponent.astro`, the path would be `storyblok/MyComponent` (without the `.astro` extension).
     :::
 
 3. `apiOptions` - An object containing [Storyblok API options](https://github.com/storyblok/storyblok-astro#options). 
 
     :::caution
-    The default region is `eu`. If your Storyblok space was created in the US region, you will need to set the `region: 'us'`.
+    By default, the region is `eu`. If your Storyblok space was created in the US region, you will need to set the region to `us`.
     :::
 
 ### Connecting Bloks to Astro components
 
 To connect your Bloks to Astro, create a new folder named `storyblok` in the `src` directory. This folder will contain all the Astro components that will match your Bloks in your Storyblok Blok library.
 
-In this example, we have a `blogPost` Blok content type in our Storyblok library with the following fields:
+In this example, you have a `blogPost` Blok content type in our Storyblok library with the following fields:
 
 - `title` - A text field
 - `description` - A text field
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/cms/storyblok.md
```diff
@@ -45,7 +45,7 @@ Your root directory should now include this new file:
 
 ### Installing dependencies
 
-To connect Astro with your Storyblok space, install the official [Storyblok integration](https://github.com/storyblok/storyblok-astro) using the single command below for your preferred package manager:
+To connect Astro with your Storyblok space, install the official [Storyblok integration](https://github.com/storyblok/storyblok-astro) using command below for your preferred package manager:
```
sarah11918, line 48:
```suggestion
To connect Astro with your Storyblok space, install the official [Storyblok integration](https://github.com/storyblok/storyblok-astro) using the command below for your preferred package manager:
```
took out one word too many!  😄

src/pages/en/guides/cms/storyblok.md
```diff
@@ -94,29 +94,29 @@ export default defineConfig({
...
     :::note
-    The component paths are relative to the `src` directory.
+    The component paths are relative to the `src` directory. For example, if your component is located at `src/storyblok/MyComponent.astro`, the path would be `storyblok/MyComponent` (without the `.astro` extension).
     :::
 
 3. `apiOptions` - An object containing [Storyblok API options](https://github.com/storyblok/storyblok-astro#options). 
 
     :::caution
-    The default region is `eu`. If your Storyblok space was created in the US region, you will need to set the `region: 'us'`.
+    By default, the region is `eu`. If your Storyblok space was created in the US region, you will need to set the region to `us`.
```
sarah11918, line 112:
```suggestion
    By default, the region is `eu`. Other available regions are ...
```
I'll let you suggest a wording based on what's actually true, but I know other services let you choose an EU or US instance, but it's not actually *required* based on where you live. E.g., I think some services let me choose which servers I want my data stored in, and I might choose that based on privacy laws.

So, is this a thing where you HAVE to set `us` or your service doesn't work? And, what if I'm in Canada?  😅 Am I supposed to choose `us` because it's closer to me? Is `us` the only other option besides `eu`? If there are other options available, then it makes more sense to list what they are and say that you can choose a different one if... you want to?

src/pages/en/guides/cms/storyblok.md
```diff
@@ -94,29 +94,29 @@ export default defineConfig({
...
     :::
 
 ### Connecting Bloks to Astro components
 
 To connect your Bloks to Astro, create a new folder named `storyblok` in the `src` directory. This folder will contain all the Astro components that will match your Bloks in your Storyblok Blok library.
 
-In this example, we have a `blogPost` Blok content type in our Storyblok library with the following fields:
+In this example, you have a `blogPost` Blok content type in our Storyblok library with the following fields:
```
sarah11918, line 119:
```suggestion
In this example, you have a `blogPost` Blok content type in your Storyblok library with the following fields:
```
😄

src/pages/en/guides/cms/storyblok.md
```diff
@@ -203,13 +206,13 @@ const content = data.story.content;
 <StoryblokComponent blok={content} />
 ```
 
-To query our data, we will use the `useStoryblokApi` hook. This will initialize a new client instance using our integration configuration.
+To query your data, use the `useStoryblokApi` hook. This will initialize a new client instance using our integration configuration.
```
sarah11918, line 209:
```suggestion
To query your data, use the `useStoryblokApi` hook. This will initialize a new client instance using your integration configuration.
```

src/pages/en/guides/cms/storyblok.md
```diff
@@ -250,7 +253,7 @@ To add content, go to the content section by clicking the **Content** tab. Using
     content: Hi there! This blog post was build with Astro.
     ```
 
-Now that we have our content ready, we can switch to our Astro project and start building our blog.
+Now that you have your content ready, you can switch to Astro and start building your blog.
```
sarah11918, line 256:
```suggestion
Now that you have your content ready, return to your Astro project and start building your blog.
```
Is this what you mean? Leave Storyblok and go work on your Astro project?

src/pages/en/guides/cms/storyblok.md
```diff
@@ -325,9 +328,9 @@ const posts = data.stories.map(story => {
 </ul>
 ```
 
-Finally, add your components to the `components` property of the `astro.config.mjs` file. Each key is the name of the Blok in Storyblok, and the value is the path to the component relative to `src`.
+Finally, add your components to the `components` property of the `storyblok` config object. The key is the name of the Blok in Storyblok, and the value is the path to the component relative to `src`.
```
sarah11918, line 331:
```suggestion
Finally, add your components to the `components` property of the `storyblok` config object in `astro.config.mjs`. The key is the name of the Blok in Storyblok, and the value is the path to the component relative to `src`.
```
I'm still tempted to identify the file in which the object exists.

sarah11918 (commented):
Looks great, Kevin! Left some tidying up suggestions, but otherwise, good to go!

src/pages/en/guides/cms/storyblok.md
kevinzunigacuellar: yes, thank you!

src/pages/en/guides/cms/storyblok.md
kevinzunigacuellar: I see your point thanks for bringing this up. In this case, this setting depends on where the user created its content space, not where the user is located. There is only two options "us" and "eu".

I decided to include this since the first time I used the integration, I had no idea that this option existed. So I created my content space in the US and set up my project, and it didn't work. I was stuck on this one for some time, the error only suggested that the fetch request failed. So I tried everything: getting a new key, rewriting the fetch request, and checking the component mapping. I don't remember how I found this option but I decided to try and it worked.

Hence, my thinking was if I got stuck on this one, there is a high chance someone else might get stuck on this one too, so lets mention it.

sarah11918: Yes, absolutely let's metion it then! So, I wonder what happens if you're not in either of those locations? I wonder how it chooses which one you have to use?  Is it ONLY if you're in the US? Like, what do you think happens if you're in Canada?

Let's keep it like this for now, because it's a good warning, but we might want to revisit this. I'm not sure at this exact moment how to delicately say "STORYBLOK KNOWS WHERE YOU LIVE" lol.

kevinzunigacuellar: Sorry, I forgot to mention that whenever you create a new content space, you can select a region from the dropdown. 
image

sarah11918: I think the instruction is good as is then! I'm happy with this whenever you want to merge! It is FRIDAY afterall!  🥳

PR #2133, opened 2022-12-02
Migrate docs to MDX (en-only)

delucis:

#### Description

- Migrate the docs to use the MDX integration, switching to `.mdx` files for content.
- Renames all content from `.md` to `.mdx` moving any imports out of `setup` in frontmatter and down the MDX body.  
    Here’s [a gist of the script I used to automate AFMD to MDX migration](https://gist.github.com/delucis/20ce3f4981235d968ae3440eadb11808) in case it’s useful to anyone.
- Updates layouts, scripts, components, etc. to expect a _mixed_ MD/MDX codebase. **Should I just pull the plug and make it MDX-only?** (N.B. Spoiler syntax `||text||` now only supports MDX as I couldn’t figure out how to make a micromark syntax-extension handle both MD & MDX.) 
- Turns off AFMD 😢👋

#### To-do

- [x] Render `<Content />` that contains a fragment — 🚫 Currently blocked by https://github.com/withastro/astro/issues/5519
- [x] Solve out-of-memory error in Netlify & GitHub Action builds  
    For now, we’re setting the `NODE_OPTIONS` environment variable to `--max_old_space_size=4096` to increase memory limits in both CI environments.
- [x] Fix `<Aside />` usage/custom syntax
- [x] Probably also fix `||spoiler syntax||`
- [x] Fix differences in heading IDs between AFMD & MDX


sarah11918:
What an amazing job @delucis :rocket: :muscle: :raised_hands: 

Admittedly, I haven't clicked through *every* single page, but I tested a number of our more "interesting" pages, and I haven't yet found anything that looks off! Lots of time for people to click around and verify, and no problem to fix smaller things after the fact. So, I can't see any reason not to approve this! New year, new Docs tech!  :partying_face:

sarah11918 (approved):
End of an Astro-Flavored Markdown era!

PR #2144, opened 2022-12-06
Document disabling telemetry with an environment variable

delucis:

- New or updated content

#### Description

- Came up in https://github.com/withastro/rfcs/discussions/405#discussioncomment-4323822 that you can disable telemetry by setting an environment variable as well as running `astro telemetry disable`. This notes that in the CLI reference. Didn’t go into any detail as it’s already kind of out of place in the CLI reference, but we don’t have anywhere better for this. Let me know if you think this deserves more.

### DIFF
--- a/src/pages/en/reference/cli-reference.md
+++ b/src/pages/en/reference/cli-reference.md
@@ -143,7 +143,7 @@ astro telemetry clear
 ```
 
 :::tip[Want to disable telemetry in CI environments?]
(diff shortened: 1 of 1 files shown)


sarah11918:
Agreed that it feels a *little* out of place (maybe eventually in Installation or some other kind of "setup-y" page would make sense?) but NWTWWHB! I'll approve it!

PR #2151, opened 2022-12-08
Add cookies.delete options description

alexpdraper:
Update with context from https://github.com/withastro/astro/pull/5341

- New or updated content

#### Description

`Astro.cookies.delete` is gaining the ability to specify a `domain` (see [this PR in the Astro repo](https://github.com/withastro/astro/pull/5341)) and already has the ability to specify a `path`. This PR is to provide that info in the docs.


src/pages/en/reference/api-reference.md
```diff
@@ -206,7 +206,7 @@ Astro.response.headers.set('Set-Cookie', 'a=b; Path=/;');
...
 | `has`          | `(key: string) => boolean`                       | Whether this cookie exists. If the cookie has been set via `Astro.cookies.set()` this will return true, otherwise it will check cookies in the `Astro.request`.          |
 | `set`       | `(key: string, value: string \| number \| boolean \| object, options?: CookieOptions) => void` | Sets the cookie `key` to the given value. This will attempt to convert the cookie value to a string. Options provide ways to set [cookie features](https://www.npmjs.com/package/cookie#options-1), such as the `maxAge` or `httpOnly`.   |
-| `delete`       | `(key: string) => void` | Marks the cookie as deleted. Once a cookie is deleted `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`.   |
+| `delete`       | `(key: string, options?: CookieDeleteOptions) => void` | Marks the cookie as deleted. Once a cookie is deleted `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options allow setting the `domain` and `path` of the cookie being deleted. |
```
sarah11918, line 209:
```suggestion
| `delete`       | `(key: string, options?: CookieDeleteOptions) => void` | Marks the cookie as deleted. Once a cookie is deleted `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options allow setting the `domain` and `path` of the cookie to delete. |
```
Any objections to this change in wording?

sarah11918 (commented):
This looks so helpful, @alexpdraper!  🥳  What a great first contribution to docs!

Not gonna lie, we in Docs already kind of hate this table because it's not doing us any favours visually  😅 ... but we'll add your row to the pile and when we do figure out how to present the data better, you'll be right in there.

Made one small verb tense suggestion that is a little cleaner, as long as you feel it represents what's happening. If you and @matthewp are happy with that, then this PR is good to go!  🚀

PR #2163, opened 2022-12-11
fix:redirect to the new URL

molychn:

#### Description

- Changed the link to the new URL due to vitejs/vite has moved framework plugins out of core.


molychn:
> Thanks for this link fix @molychn!
> 
> Unfortunately this page is generated automatically from the package’s README in a different repo, so we need to make the change there: https://github.com/withastro/astro/blob/main/packages/integrations/vue/README.md
> 
> Would you be up for making a PR with these same changes in that repo?

@delucis Okay, I have make a [PR](https://github.com/withastro/astro/pull/5585) to fix in that repo, this PR I would close.

PR #2165, opened 2022-12-11
Migration guides from

crutchcorn:
Per prior discussions with @sarah11918, these are my edits to her work on Astro migration guides from Gatsby. These edits have left me feeling relatively happy with the Gatsby guide, and I think it should be ready for another pass of reviews.

A few questions I had along the way:

1. Is the code highlighting/naming correct in all code blocks?
2. Are we happy with the current iteration of the `<Image>` component? It's one of the only instances without a before/after
3. Are we happy with the length? It got _long_ quick.

This PR also adds in a few comments regarding the NextJS migration guide that I will be tackling next.


sarah11918:
Merging! It's the new thing!

PR #2114, opened 2022-11-28
[migration-guide] from-gatsby initial page draft

sarah11918:
This is a first draft of a Migrate from Gatsby to Astro guide.

It is *not* a step-by-step walk-through, because it would be difficult to predict what someone's existing project looks like. So it's more, "things to think about, probably roughly in this order" with varying degrees of explicit instruction. 

It may need significant work, and is open to any and all suggestions! 

It is intentionally longer than shorter, because it's easier to cut out whatever we decide is unnecessary. For the sake of starting this draft, some of it is a little hand-wavy with the knowledge that if things aren't explicit enough or don't capture the right nuance, someone's gonna speak up. Not all links to related docs sections are yet in place, because I could be linking forever. So, a comprehensive "linking enough but not too much" audit is yet to come.

**The goals of this page are:**
- To highlight the key differences and similarities **to prepare the reader for switching frameworks**
- To **guide a little more explicitly through initial project creation**, offering two suggestions for getting started:
	- create-astro, then bring existing Gatsby files over 
	- update dependencies in package.json, install Astro manually, and edit files in place.
- To explain `astro.config` in relation to Gatsby's configuration files, and **to help the reader extract any existing configuration** and put it into the appropriate places
- To **provide general but comprehensive guidance on converting existing `.js(x)` Gatsby files into `.astro` components**, including:
	- a before (Gatsby)/after (Astro) example of a component 
	- SEVERAL examples of walk-through conversions
	- a list of several key syntax differences (e.g. `className`, `{children}`, `<Link to="" />` `style={{}}`) and their Astro equivalents
- **To suggest a migration path** starting with converting Layouts (so that pages will be complete HTML documents when using a Layout), then importing existing pages and posts.

**The challenges of this page are:**
- **To fulfill the goals without overwhelming the reader**. I'm less concerned about page length itself and more concerned that the sidebar doesn't make people say, "Yikes!"
- **To write section headings that will not confuse our Algolia search results** and send people to this page for "normal docs things" (e.g. Image Integration, Site Configuration). This is why the headings are now intentionally long-ish, because shortening them runs the risk of Algolia mistaking the section for regular docs.
- **Gatsby components are HUGE!**  😅 It's admittedly visually overwhelming to have SUCH LARGE code examples as "before" models, when they include one (or more!) functions, and GraphQL queries that take so many lines to format etc. It's making this page look long, but the actual process for converting the files is entirely manageable, and I'd love to be able to show that here.

[...]

PR #2170, opened 2022-12-12
Migration Guides from Gatsby and NextJS to Astro

crutchcorn:
> This is a continuation of https://github.com/withastro/docs/pull/2114, migrated to my fork per @sarah11918's guidance in order to make it easier for me to make larger contributions. The following is the original PR description as written by @sarah11918 

This is a first draft of a Migrate from Gatsby to Astro guide.

It is *not* a step-by-step walk-through, because it would be difficult to predict what someone's existing project looks  like. So it's more, "things to think about, probably roughly in this  order" with varying degrees of explicit instruction.

It may need significant work, and is open to any and all suggestions!

It is intentionally longer than shorter, because it's  easier to cut out whatever we decide is unnecessary. For the sake of  starting this draft, some of it is a little hand-wavy with the knowledge that if things aren't explicit enough or don't capture the right  nuance, someone's gonna speak up. Not all links to related docs sections are yet in place, because I could be linking forever. So, a  comprehensive "linking enough but not too much" audit is yet to come.

**The goals of this page are:**

- To highlight the key differences and similarities **to prepare the reader for switching frameworks**

- To **guide a little more explicitly through initial project creation**, offering two suggestions for getting started:

  - create-astro, then bring existing Gatsby files over
  - update dependencies in package.json, install Astro manually, and edit files in place.

- To explain `astro.config` in relation to Gatsby's configuration files, and **to help the reader extract any existing configuration** and put it into the appropriate places

- To **provide general but comprehensive guidance on converting existing `.js(x)` Gatsby files into `.astro` components**, including:

  - a before (Gatsby)/after (Astro) example of a component
  - SEVERAL examples of walk-through conversions
  - a list of several key syntax differences (e.g. `className`, `{children}`, `<Link to="" />` `style={{}}`) and their Astro equivalents

- **To suggest a migration path** starting with  converting Layouts (so that pages will be complete HTML documents when  using a Layout), then importing existing pages and posts.

**The challenges of this page are:**

- **To fulfill the goals without overwhelming the reader**. I'm less concerned about page length itself and more concerned that the sidebar doesn't make people say, "Yikes!"
- **To write section headings that will not confuse our Algolia search results** and send people to this page for "normal docs things" (e.g. Image  Integration, Site Configuration). This is why the headings are now  intentionally long-ish, because shortening them runs the risk of Algolia mistaking the section for regular docs.
- **Gatsby components are HUGE!**  😅 It's admittedly visually overwhelming to have SUCH LARGE code examples  as "before" models, when they include one (or more!) functions, and  GraphQL queries that take so many lines to format etc. It's making this  page look long, but the actual process for converting the files is  entirely manageable, and I'd love to be able to show that here.

**Site Infrastructure Note:** Uses the CMS  Guides templating for now. By the time we have 3 guides ready for  publishing, we will have also created a proper Migrations Guide landing  page and stubs as placeholders for other guides.


sarah11918:
I will make sure I "update branch" at least once, for my co-authored-by cred!  😂

src/pages/en/guides/migrate-to-astro.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Migration Guides
+
+<CMSGuidesNav />
+
+Note that many of these pages are **stubs**: they're collections of resources waiting for your contribution!
+
+## Why migrate your site to Astro?
+
+Astro provides many benefits: performance, simplicity and several built-in and available integrations!
```
crutchcorn, line 22: The following suggestion was originally written by @sarah11918:

```suggestion
Astro provides many benefits: performance, simplicity, and many of the features you want built right in to the framework. When you do need to extend your site, Astro provides several official and 3rd-party community integrations!
```

Not entirely sold on this, but if there's a way to say "some of the stuff you have to install separately (and you're almost certainly going to anyway) is just included in Astro. Plus, there's more"... this is what I was trying to get at.

src/pages/en/guides/migrate-to-astro.md
```diff
@@ -0,0 +1,72 @@
...
+
+Depending on your existing project, you may be able to bring some, or much of your existing files. However, Astro's project structure is likely to be different than the one you're using. So, there are some key concepts that will help you make the transition to Astro.
```
crutchcorn, line 24: The following suggestion was originally written by @sarah11918:

```suggestion
Migrating may be less work than you think! Depending on your existing project, you may be able to fit some, or many of your files directly into Astro's project structure. Much of your existing content may be reusable when converted to `.astro` components.
```

What do we think about something like this?

src/pages/en/guides/migrate-to-astro.md
```diff
@@ -0,0 +1,72 @@
...
+
+Astro allows you to choose between a statically-generated site and server-side rendering (SSR), making it a great replacement for SSGs or for sites that need to fetch some page data on the fly.
+
+Because Astro's focus is shipping as little JavaScript as necessary [via its island architecture](/en/concepts/islands/#what-is-an-astro-island), it is well-suited to sites where client-side interactivity is localized to particular components.
+
+## What changes will I encounter when moving to an island architecture?
+
+Migrating your site to Astro means thinking about your site in "islands" or isolated sections: which parts are static? which are dynamic? which are interactive? which ones need to communicate with others?
+
+Thinking about your site as "islands" may be a shift of thinking, which brings with it some additional considerations:
```
crutchcorn, line 39: The following suggestion was originally written by @sarah11918:

```suggestion
As you shift your thinking away from a default shared app-level context and towards "islands by default" in Astro, you'll design differently:
```

Totally not sold on how I worded the "away from", especially since they could be moving "Away from" any number of things!" and honestly could even drop it. BUT, if there's a clear, concise drop in for "away from...." that is representative of all non-island architecture thinking, I think I'd like to include it in a sentence like this.

Using a simpler verb tense here, I'm also now less sold on the line above being in the "Migrating..." verb tense. ("When you migrate your site to Astro, you'll now think about...") So I'm totally open to changing that, too!

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
+layout: ~/layouts/CMSLayout.astro
+stub: false
+framework: Gatsby
+---
+
+Here are some tips for converting a Gatsby project to Astro. This is not a full, step-by-step walkthrough, but it will guide you through some changes you will have to make. 
+
+## Key Similarities
+
+While we'll touch on the differences between Gatsby and Astro shortly, it's important to note the ways in which the two overlap.
```
crutchcorn, line 13: The following suggestion was originally written by @sarah11918:

```suggestion
Gatsby and Astro share some similarities that will help you migrate your project:
```

Not opposed to an introductory sentence here, but also not super convinced that we need one. (Good for SEO?) Also would consider `## Key Similarities between Gatsby and Astro` in case we think that is better heading/Algolia/SEO usage.

A little note: I'm less-inclined to narrate the structure of the page (e.g. tell them what's coming up). I think this page is already gonna be quite long, there's a TOC in the sidebar, and people will get from this heading that this section is `similarities` (which is why I also think we might not need any sentence at all? Esp if we put the two names in the heading).

Thoughts?

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
+
+- Astro also has [official integrations for using React components](/en/guides/integrations-guide/react/). Note that in Astro, React files **must** have a `.jsx` or `.tsx` extension.
+
+- Astro projects can also be SSG or SSR. (Support for per-page rendering strategy is planned.)
+
+- Astro has support for NPM package usage, including several for React. You may be able to keep some or all of your existing React components and dependencies.
+
+## Key Differences
+
+Now that you understand the ways that Astro and Gatsby align, how do they diverge?
```
crutchcorn, line 29: The following suggestion was originally written by @sarah11918:

```suggestion
When you rebuild your Gatsby site in Astro, you will notice some important differences, especially in how your project data is queried and used to create your site.
```

rebuild? recreate? port? migrate? convert? I'm torn between intentionally using a new word that might bring us new hits, vs a word that maybe feels a bit more natural/appropriate in context? Thoughts?

Also, again here I'm not entirely convinced that any sentence is necessary (is having a preamble just a wordier way to say, "differences!") BUT if we do use it, I'm thinking maybe to make it useful, a nod to an actual key difference like the GraphQL data layer, and `createPates`? Again, open to thoughts here!

(Continued point: leaning away from self-narration of the page)

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
+
+Astro is a multi-page site, and `index.astro` is your home page. It generates HTML pages for each configured route but, in contrast to Gatsby, only initializes JavaScript on the interactive elements on-screen.
+
+### Page routing
+
+Because of the differences between React apps and Astro MPA apps, you don't need to know any domain-specific knowledge to link between different pages. Instead of Gatsby's custom `<Link to="/path">` component, you'll use the HTML standard `<a href="/path">` tag.
+
+In addition, while Gatsby primarily uses the `createPages` function to create dynamic pages, [Astro opts for page-based routing](/en/core-concepts/routing/#dynamic-routes) that better represents the individual HTML files that are output in a build.
+
+Finally, should you need to access information about the current route within an Astro component, [you're able to use `Astro.url`](/en/reference/api-reference/#astrourl), which exposes a [web-standard `URL` class](https://developer.mozilla.org/en-US/docs/Web/API/URL).
```
crutchcorn, line 43: The following suggestion was originally written by @sarah11918:

```suggestion
### Page routing

You don't need to know any domain-specific knowledge to link between different pages in Astro. Instead of Gatsby's custom `<Link to="/path">` component, use the HTML standard `<a href="/path">` tag.

In addition, while Gatsby primarily uses the `createPages` function to create dynamic pages, [Astro opts for page-based routing](/en/core-concepts/routing/#dynamic-routes) that better represents the individual HTML files that are output in a build.

To access information about the current route, use [`Astro.url`](/en/reference/api-reference/#astrourl) to expose a [web-standard `URL` class](https://developer.mozilla.org/en-US/docs/Web/API/URL).
```

Fixed the links to make them relative inside the project, and made the first and third points a little more direct. (Use/ vs. you'll use; not annotating "additionally/finally")

I grok the middle point, but I think the wording needs some work. The points to me are:

- Gatsby needs `createPages` AT ALL to... create any pages! Astro CAN create pages (non-dynamically) on file-placement alone. Dynamic page routes are not in fact necessary (although, totally possible!)
- `createPages` is like a config setup thing to create ALL your pages. Astro's `getStaticPaths` is right inside `src/pages/`, collocated with the set of pages you want to build dynamically. Maybe that's the way in which Astro's dynamic page routing is "closer to the platform of HTML"?

Is there any other nuance about that I'm not getting that should be represented? Can we keep working on the wording of point 2?

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
+
+Your Astro project will not use any of these `gatsby-*.js` files, but there may be some content that you can reuse:
+
+- `gatsby-config.js`: Move your `siteMetadata: {}` into `src/data/siteMetadata.js` (or `siteMetadata.json`) to import data about your site (title, description, social accounts etc.) into page layouts.
+- `gatsby-browser.js`: Consider adding anything used here directly into your main layout's `<head>` tag.
+- `gatsby-node.js`: This file handles multiple server-side operations that have built-in support in Astro:
+     - GraphQL node customization: You do not need to customize the schema in Astro, but viewing the Gatsby schema may help you with defining types in your Astro projects.
+     - Page creation: Astro supports both [async data loading](/en/reference/api-reference/#getstaticpaths) and [pagination](/en/core-concepts/routing/#pagination) right out of the box. You can cross-reference your `createPages` function to migrate the pages to Astro's methodologies.
```
crutchcorn, line 90: The following suggestion was originally written by @sarah11918:

```suggestion
     - Page creation: Any page file located within `src/pages/` will become a page route in Astro with no `createPages` function needed. However, for dynamic page generation, Astro supports both [async data loading](/en/reference/api-reference/#getstaticpaths) and [pagination](/en/core-concepts/routing/#pagination) right out of the box. You can cross-reference your `createPages` function to migrate to Astro's routing methods.
```

Would you say this is still accurate? I do think it's worth mentioning that you don't necessarily NEED to generate your pages dynamically.. like, at all!

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
+</html>
+```
+
+```css title="src/layouts/global.css"
+p {
+    font-weight: bold;
+}
+```
+
+#### CSS Preprocessors
```
crutchcorn, line 480: The following comment was originally written by @sarah11918:

Just calling out that this section doesn't feel very "Here's what to do to replicate your Gatsby setup in Astro."

Unless the example below is wildly different from how this is done in Gatsby, I'd consider just a sentence or two saying you CAN use them and so e.g. "You can use your existing setup" or a very quick explanation of what do DO when converting a site.

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
+   &:hover {
+       color: red;
+   }
+}
+</style>
+```
+
+### Gatsby Code Comments to Astro
+
+An Astro file uses JavaScript code comments in the frontmatter `//` but HTML code comments in the template body. (`<!-- -->`)
```
crutchcorn, line 508: The following suggestion was originally written by @sarah11918:

Should we include something like the code comment sample that's here? https://docs.astro.build/en/core-concepts/astro-components/#component-structure

src/pages/en/guides/migration-guide/from-nextjs.md
```diff
@@ -0,0 +1,771 @@
...
+                textAlign: `center`,
+                marginBottom: `1em`
+            }}>Astro has {stars} 🧑‍🚀</p>
+            <Footer />
+        </>
+    )
+}
+
+export default Component;
+```
```
crutchcorn, line 128: The NextJS section is 100% not ready for review, but I wanted to get your early thoughts on this code sample @sarah11918 

I couldn't find anything that quite utilized the same amount of code principles we had in the Gatsby site. What's worse, Gatsby doesn't have a great single strong example; instead [they opt for many many smaller examples](https://nextjs.org/examples).

As a result, I took an existing example: https://github.com/vercel/next.js/blob/canary/examples/data-fetch/pages/preact-stars.tsx

And modified it for our use.

The modified version shows how you would migrate:

- Styling migrations
- Data fetching
- Conditional rendering

All in one semi-small code sample. What do you think? Do we like this or should I replace it with something else?

sarah11918: Early returns look good! I think you've done a great job of finding an example that even somewhat mirrors what we were able to do with the Gatsby example. 

I *love* showing a conversion with `getStaticProps()`. I think with Gatsby, leaning into the GraphQL and `createPages` made sense. With Next.js, the main differences are going to be different, and certainly `getStaticProps()` *is* something that was on my radar... since `getStaticPaths()` is a shared feature, but the props, not so much.

Let's absolutely roll with this!

src/pages/en/guides/migrate-to-astro.md
crutchcorn: I really like the "You'll design differently" line, but think the "default shared app-level context" is a bit terminology-heavy.

Maybe we could try:

```suggestion
As you shift your thinking away from "all JavaScript all the time" and towards "islands by default" in Astro, you'll design differently:
```

It's a little oversimplified, but it may speak better to engineers like me who struggle with terminology overload?

What do you think @sarah11918

src/pages/en/guides/migration-guide/from-gatsby.md
crutchcorn: Yeah, totally agree with you here @sarah11918 - the "narrate the structure" wasn't my favorite either. I was really struggling to throw this sentence together, but your edit works nicely.

I did want to keep an introductory sentence here to make the glanceability a bit nicer for the end user.

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
+title: Migrating from Gatsby
+description: Tips for migrating an existing Gatsby project to Astro
+layout: ~/layouts/CMSLayout.astro
+stub: false
+framework: Gatsby
+---
+
+Here are some tips for converting a Gatsby project to Astro. This is not a full, step-by-step walkthrough, but it will guide you through some changes you will have to make. 
+
+## Key Similarities
```
crutchcorn, line 11:
```suggestion
## Key Similarities between Gatsby and Astro
```

src/pages/en/guides/migration-guide/from-gatsby.md
crutchcorn: I think any of the words you suggested here feel pretty appropriate. In addition to bringing us new potential SEO, it also helps break up the "monotony" of reading in-depth content to have our terminology shift a _little_ to mix things up.

As for pointing towards a key difference, I'm not sure I'm entirely sold simply because I feel like people might glance over this sentence and assume it's the _primary_ difference between Astro and Gatsby, despite there being a couple (which is the point of this section)

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
+
+Gatsby is a React app, and uses `index.js` as your project's root. While Gatsby is able to generate an HTML page for each of the routes configured, it also re-initializes React for all of the content on-screen, including static portions.
+
+Astro is a multi-page site, and `index.astro` is your home page. It generates HTML pages for each configured route but, in contrast to Gatsby, only initializes JavaScript on the interactive elements on-screen.
+
+### Page routing
+
+You don't need to know any domain-specific knowledge to link between different pages in Astro. Instead of Gatsby's custom `<Link to="/path">` component, use the HTML standard `<a href="/path">` tag.
+
+In addition, while Gatsby primarily uses the `createPages` function to create dynamic pages, [Astro opts for page-based routing](/en/core-concepts/routing/#dynamic-routes) that better represents the individual HTML files that are output in a build.
```
crutchcorn, line 41: Re-posting @sarah11918's comment about this point here:

I grok the middle point, but I think the wording needs some work. The points to me are:

- Gatsby needs `createPages` AT ALL  to... create any pages! Astro CAN create pages (non-dynamically) on  file-placement alone. Dynamic page routes are not in fact necessary  (although, totally possible!)
- `createPages` is like a config setup thing to create ALL your pages. Astro's `getStaticPaths` is right inside `src/pages/`, collocated with the set of pages you want to build dynamically. Maybe  that's the way in which Astro's dynamic page routing is "closer to the  platform of HTML"?

Is there any other nuance about that I'm not getting that should be represented? Can we keep working on the wording of point 2?

crutchcorn: We can definitely keep working on the wording of point 2, but there is a bit of nuance that might not be captured otherwise:

- [Gatsby _can_ use file-based routing without `createPages`](https://www.gatsbyjs.com/docs/reference/routing/file-system-route-api#collection-routes), but per their own admission:
    > If you want more control over the page creation you should use the `createPages` API.
- Gatsby is able to use `createPages` _alongside_ page-based file routing, they are not mutually exclusive.

Would love your thoughts on how we can improve the middle sentence without implying that the user _must_ utilize `createPages`.

src/pages/en/guides/migrate-to-astro.md
sarah11918: Approved!

src/pages/en/guides/migration-guide/from-gatsby.md
```diff
@@ -0,0 +1,1199 @@
...
```
sarah11918, line 41:
```suggestion
Gatsby uses the `createPages` function to configure creating dynamic pages, whereas [Astro opts for page-based dynamic routing](/en/core-concepts/routing/#dynamic-routes). In Astro, both static pages and pages that generate multiple pages are located within `src/pages/` and share many similarities. 
```
So, what if we focus on the fact that `getStaticPaths()` *in a page file itself* creates pages, whereas Gatsby's page creation is kind of closer to a "config/setup" type thing? Would this be true? (Not sold on this wording; wondering whether this qualifies as an important diff to hit, and whether it's meaningful/helpful to point this out to those converting.)

PR #2172, opened 2022-12-12
docs: fix Astro.site docs on undefined configs

okikio:

- New or updated content

#### Description

- Clarifies docs on `Astro.site` when `site` is undefined in the Astro config,


delucis (approved):
Thanks for catching that @okikio 🙌

PR #2175, opened 2022-12-13
Use type import in “Built-in HTML attributes” example

simbas:

#### Description

- What does this PR change?
The PR adds 'type' to the `import { HTMLAttributes } from 'astro/types'` from `Built-in HTML attributes` section.


delucis (approved):
Thanks for adding that detail @simbas!

PR #2185, opened 2022-12-16
Add font guide for Tailwind CSS

surjithctly:
Add guide for adding custom fonts with tailwindcss.

- New or updated content


sarah11918:
Thanks for contributing this, @surjithctly! 

I'm not sure about putting this in the docs, since it seems to be aabout Tailwind configuration. We don't typically document things that aren't a part of Astro, or specifics about integrating with Astro, because these things are usually described in their own docs. And, we can't easily know when it has changed or needs updating, so it's an extra maintenance cost for us to try to keep our info current.

But, we do want to be helpful to Astro users, so it's a balancing act!

I just want to make sure that I understand which content you're proposing here is very specific to Astro, and which is content that is found at the original link https://tailwindcss.com/docs/font-family#using-custom-values . Was the original link not helpful? Does it not explain how to do this? Thanks for helping us understand better what problem this fixes!

surjithctly:
Because tailwind is now popular choice. The current docs is confusing. While it shows how to use fontsource (3rd party, their code can also change) we can definitely show tailwind as well. I’m only thinking about users confusion and potentially avoid dead end. 

So it doesn’t need to worry about 3rd party updates because it’s something we can always update later. and its not breaking like code changes.

Finally, astro/image is still throws error on node 18 (third party) and the team is yet to fix.

So, my vote is to add it. ✅

sarah11918:
Those sound like good reasons to me, thanks! Just wanted some clarification. I'll take a look at this one more closely right now!  🚀

src/pages/en/guides/fonts.md
```diff
@@ -103,11 +103,45 @@ The [Fontsource](https://fontsource.org/) project simplifies using Google Fonts
...
+            sans: ["InterVariable", "Inter", ...defaultTheme.fontFamily.sans],
+          },
+        },
+      },
+      plugins: [],
+    };
+
+    ```
+    
+4. Now, every sans (default) text in your project will use the above font. Learn more on [Tailwind’s docs on adding custom font families](https://tailwindcss.com/docs/font-family#using-custom-values).
```
sarah11918, line 142:
``````suggestion
### Using Fontsource with Tailwind CSS

To add custom fonts in to your project using Fontsource when you have the [Tailwind integration](/en/guides/integrations-guide/tailwind/) installed, 

1. [Install the font you want from Fontsource] (#using-fontsource) using the method described above.

2.  In `tailwind.config.cjs`,  import the font into your project. This example imports `Inter Variable` and `Inter` with default fallback fonts from Tailwind CSS.
    ```js
    const defaultTheme = require("tailwindcss/defaultTheme");
    module.exports = {
      content: ["..."],
      theme: {
        extend: {
          fontFamily: {
            sans: ["InterVariable", "Inter", ...defaultTheme.fontFamily.sans],
          },
        },
      },
      plugins: [],
    };
    ```
  3. Import the font package in the component where you want to use the font. Usually, you will want to do this in a common layout component to make sure the font is available across your site.

    The following example imports the `Inter Variable` font.  The necessary `@font-face` rules needed to set up the font and a fallback with default `Inter` font are automatically added.

    ```astro title="src/layouts/Layout.astro
    ---
    import '@fontsource/inter/variable.css';
    ---
    ```
  
Now, every sans (default) text in your project will use the the installed font. See [Tailwind’s docs on adding custom font families](https://tailwindcss.com/docs/font-family#using-custom-values) for more information.  
``````

sarah11918 (commented):
OK! Here's some initial thoughts on this.

I know Tailwind is very popular, and popular with Astro users, so thank you for providing this! We don't document *much* about Tailwind, but I think you're right and we can make this easier for our readers by having this here.  🚀 

I made just a few changes below, so please have a look and let me know if you think this agrees with your intentions. If not, let me know why not! In particular:

- I made this a level 3 heading under using Font source. I don't want to pretend that this is big, comprehensive Tailwind advice on this page. And, since it relies on a step from the earlier section, I see this as, "How to use Font Source if you're also using Tailwind." I could be wrong! But it feels like it could be a subheading here.
- I reordered the steps, because updating the tailwind config files feels like it should be shown (so you can see what is configured) before importing the font into your layout component.
- I just did some small text edits to make the style fit more with what's existing in docs.

This suggestion might not be entirely perfect! But, it is closer to what I think I see fitting in with the rest of docs on this page. Using what you see here, and your judgment, what edits do you think you'd like to make?

surjithctly:
Thank you. This is perfect for me 👍

You can merge this once ready.

kevinzunigacuellar:
I've just tested the steps in my own site, everything works as expected 💜

surjithctly:
Thanks so much @kevinzunigacuellar for trying and glad it worked out well. it proves many people needed this :)

delucis (approved):
Thanks for this contribution @surjithctly! Made some tweaks so that it’s clear you can do this both with Fontsource and with local font files.

PR #2188, opened 2022-12-17
Add missing backtick

chenxsan:

#### Description

Add the missing backtick for inline code.


delucis (approved):
Good catch! Thanks for fixing that 🚀

PR #2189, opened 2022-12-17
Fix example code in guide

chenxsan:

- Fix code in example


delucis (approved):
Perfect. Thank you!

PR #2204, opened 2022-12-18
chore: bump actions/checkout in GitHub deploy guide 

swift502:

- New or updated content

#### Description

- Primarily, update the script so it reflects that users should now use actions/checkout@v3 instead of v2, which was recently merged in https://github.com/withastro/action/pull/17
- Script is now a 1:1 match with the one found in [withastro/action](https://github.com/withastro/action) repository

Thanks!

### DIFF
--- a/src/pages/en/guides/deploy/github.md
+++ b/src/pages/en/guides/deploy/github.md
@@ -44,8 +44,8 @@ Astro maintains the official `withastro/action` to deploy your project with very
 2. Create a new file in your project at `.github/workflows/deploy.yml` and paste in the YAML below.
 
     ```yaml title="deploy.yml"
-    name: Github Pages Astro CI
-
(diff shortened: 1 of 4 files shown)


lino-levan:
LGTM! Not sure if the wording changes were necessary but it does improve clarity.

swift502:
> Not sure if the wording changes were necessary

@lino-levan  I just matched the script from github, so now it should be 1:1. It seemed like the right thing to do.

Thanks for the Deno guide update, I didn't know this concerned other guides as well.

lino-levan:
I'm not super sure why but a lot of the guides seem to be using `actions/checkout@v2`. It seems to work the same with v3 so I'm not entirely sure why this is the case (other than no one has bothered to update it yet).

delucis (approved):
Thanks @swift502 🙌

PR #2212, opened 2022-12-19
Fix highlight regex for `CollectionEntry` code example

zadeviggers:

Old: 
image
New:
image

#### Description

Fixes a broken code highlight regex.

I'm not sure what exactly the cause of the issue was - it might have been a typo (the square brackets), or depending on how the code that tests the regex works, it might have been the use of a capturing group instead of a non-capturing group (a group that starts with `?:`).


delucis (approved):
Nice! Thanks for catching that @zadeviggers 🙌

PR #2216, opened 2022-12-20
Document rehype heading IDs plugin

delucis:

- New or updated content

#### Description

- Accompanying docs for https://github.com/withastro/astro/pull/5654
- May be overkill as these are less commonly used options, but we can cut if necessary


src/pages/en/guides/markdown-content.md
```diff
@@ -459,6 +459,28 @@ export default {
 }
 ```
 
+#### Markdown heading IDs
+
+Astro injects an `id` attribute into all heading elements (`<h1>` to `<h6>`) in Markdown and MDX files and provides a `getHeadings()` utility for retrieving these IDs in [Markdown exported properties](#exported-properties).
+
+You can customize heading IDs by adding an alternative plugin that injects `id` attributes (e.g. `rehype-slug`). Your custom IDs will be reflected in HTML output and the items returned by `getHeadings()`.
```
sarah11918, line 466:
```suggestion
You can customize these heading IDs by adding a rehype plugin that injects `id` attributes (e.g. `rehype-slug`). Your custom IDs, instead of Astro's defaults, will be reflected in HTML output and the items returned by `getHeadings()`.
```
"an alternate" plugin maybe sounds like there is one, official sanctioned alternative? Are they all rehype plugins that do this? If so, what about saying that?

src/pages/en/guides/markdown-content.md
```diff
@@ -459,6 +459,28 @@ export default {
...
+
+By default, Astro injects `id` attributes after your rehype plugins have run. If you need to access the IDs from one of your custom rehype plugins, you can import and use Astro’s `rehypeHeadingIds` plugin directly:
```
sarah11918, line 468:
```suggestion
By default, Astro injects `id` attributes only after your rehype plugins have run. If one of your custom rehype plugins needs to access the IDs injected by Astro, you can import and use Astro’s `rehypeHeadingIds` plugin directly. Be sure to list it before any plugins that rely on it:
```
I made up the list them in order thing, but if it's true, it should probably be explicitly said.

And, just checking: if Astro injects the id attribute after your rehype plugins have run, and one of your plugins has already injected IDs, then Astro does NOT change/alter the id? So, would it be correct to say that Astro injects an id iff there's not already one there? Otherwise, how does it not change these? I had this question after reading that line, so others might too?

sarah11918 (approved):
Thanks, @delucis! This is so helpful and looks great! I made a few tiny tweaks to see what you think, and left a question I was still left with after reading the text, so maybe others would have it, too!

src/pages/en/guides/markdown-content.md
delucis: Fair question. The answer is kind of what I was hinting at with:

> Your custom IDs will be reflected in HTML output and the items returned by `getHeadings()`.

But you’re right and maybe we should be more explicit that we only inject if there isn’t an `id` already.

src/pages/en/guides/markdown-content.md
delucis: Right it would be a rehype plugin. `rehype-slug` is the most common (? citation needed) but there are others I think. Or you could write your own if you were feeling adventurous.

src/pages/en/guides/markdown-content.md
delucis: I like this suggestion, but making some small edits:

```suggestion
By default, Astro injects `id` attributes after your rehype plugins have run. If one of your custom rehype plugins needs to access the IDs injected by Astro, you can import and use Astro’s `rehypeHeadingIds` plugin directly. Be sure to add `rehypeHeadingIds` before any plugins that rely on it:
```

src/pages/en/guides/markdown-content.md
```diff
@@ -459,6 +459,28 @@ export default {
 }
 ```
 
+#### Markdown heading IDs
+
+Astro injects an `id` attribute into all heading elements (`<h1>` to `<h6>`) in Markdown and MDX files and provides a `getHeadings()` utility for retrieving these IDs in [Markdown exported properties](#exported-properties).
+
+You can customize these heading IDs by adding a rehype plugin that injects `id` attributes (e.g. `rehype-slug`). Your custom IDs, instead of Astro's defaults, will be reflected in HTML output and the items returned by `getHeadings()`.
+
+By default, Astro injects `id` attributes only after your rehype plugins have run. If one of your custom rehype plugins needs to access the IDs injected by Astro, you can import and use Astro’s `rehypeHeadingIds` plugin directly. Be sure to list it before any plugins that rely on it:
```
delucis, line 468: Remaking this suggestion so I can commit it:

```suggestion
By default, Astro injects `id` attributes after your rehype plugins have run. If one of your custom rehype plugins needs to access the IDs injected by Astro, you can import and use Astro’s `rehypeHeadingIds` plugin directly. Be sure to add `rehypeHeadingIds` before any plugins that rely on it:
```

delucis: (Turns out I’d already committed this suggestion last week 🤦 — I still like my tweaks, but seems pettier now 😅)

src/pages/en/guides/markdown-content.md
```diff
@@ -459,6 +460,28 @@ export default {
 }
 ```
 
+#### Heading IDs and plugins
+
+Astro injects an `id` attribute into all heading elements (`<h1>` to `<h6>`) in Markdown and MDX files and provides a `getHeadings()` utility for retrieving these IDs in [Markdown exported properties](#exported-properties).
+
+You can customize these heading IDs by adding a rehype plugin that injects `id` attributes (e.g. `rehype-slug`). Your custom IDs, instead of Astro's defaults, will be reflected in HTML output and the items returned by `getHeadings()`.
```
sarah11918, line 467:
```suggestion
You can customize these heading IDs by adding a rehype plugin that injects `id` attributes (e.g. `rehype-slug`). Your custom IDs, instead of Astro's defaults, will be reflected in the HTML output and the items returned by `getHeadings()`.
```

Having a brain freeze, but do we need a "the" here?

sarah11918 (approved):
Looks great to me!

src/pages/en/guides/markdown-content.md
delucis: I _think_ it’s one of those situations where it doesn’t really matter (?) but sure, why not 👍

PR #2228, opened 2022-12-22
Edit/storyblok guide error fixes

### LINKED ISSUE #2183: [cms/storyblok] update set up guide
**Description** 
A new version of the storyblok integration just came out 🎉 [release](https://github.com/storyblok/storyblok-astro/releases). A few keypoints:

- Migration from JS to TS
- Axios is no longer a peer dep

### LINKED ISSUE #2227: Astro/Storyblok CMS guide build error when following instructions
## What's the problem?
In the Astro/Storyblok docs, there is a typo in the setup instructions that causes an error across every page where `components` are used. As the code happens to be focused on a component that auto-generates components, it's not clear what's actually causing the problem.

Here's the error that's shown to the user

```
Cannot destructure property 'id' of '(intermediate value)' as it is null.
 error   Cannot destructure property 'id' of '(intermediate value)' as it is null.
  File:
    /Users/dan/Code/storyblok-astro-starter/node_modules/@storyblok/astro/dist/storyblok-astro.mjs:55:23
  Code:
    54 |         for await (const [n, a] of Object.entries(t)) {
    > 55 |           const { id: l } = await this.resolve("/src/" + a + ".astro");
         |                       ^
      56 |           s.push(`import ${g(n)} from "${l}"`);
      57 |         }
      58 |         return `${s.join(";")};export default {${Object.keys(t).map((n) => g(n)).join(",")}}`;
  Stacktrace:
TypeError: Cannot destructure property 'id' of '(intermediate value)' as it is null.
    at Context.load (file:///Users/dan/Code/storyblok-astro-starter/node_modules/@storyblok/astro/dist/storyblok-astro.mjs:55:23)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.load (file:///Users/dan/Code/storyblok-astro-starter/node_modules/astro/node_modules/vite/dist/node/chunks/dep-5605cfa4.js:40507:32)
    at async loadAndTransform (file:///Users/dan/Code/storyblok-astro-starter/node_modules/astro/node_modules/vite/dist/node/chunks/dep-5605cfa4.js:36859:24)

/Users/dan/Code/storyblok-astro-starter/node_modules/shiki/dist/index.js:2213
      throw Error(`No language registration for ${lang}`);
            ^

Error: No language registration for mjs
    at getGrammar (/Users/dan/Code/storyblok-astro-starter/node_modules/shiki/dist/index.js:2213:13)
    at codeToThemedTokens (/Users/dan/Code/storyblok-astro-starter/node_modules/shiki/dist/index.js:2222:26)
    at Object.codeToHtml (/Users/dan/Code/storyblok-astro-starter/node_modules/shiki/dist/index.js:2236:20)
    at getViteErrorPayload (file:///Users/dan/Code/storyblok-astro-starter/node_modules/astro/dist/core/errors/dev/vite.js:90:54)
    at async Timeout._onTimeout (file:///Users/dan/Code/storyblok-astro-starter/node_modules/astro/dist/vite-plugin-astro-server/response.js:18:61)
error Command failed with exit code 1.
```

## Where can it be found or how can it be replicated?
https://docs.astro.build/en/guides/cms/storyblok/#connecting-bloks-to-components
Inside the codeblock -> `components: { blogPostList: ... }`. If you compare this to the codeblock directly above you'll notice that the file locations are different, which is what causes the error.

## How can this problem be resolved?
There are 2 potential fixes

1. is to remove the line of code which simply fixes the error
2. is a more complete solution as other parts of the guide talk about blogPostList as a component that can be _used_ within a page, so updating the docs to account for a blog page which uses a blogPostList component would make the most sense.

danspratling:

#### Description

- Closes #2227
- EDIT: Also Closes #2183 
- **What does this PR change?** Fixes some incorrect code examples in Astro/Storyblok guide
- **Did you change something visual?** Nope! Just copy & code blocks


src/pages/en/guides/cms/storyblok.md
```diff
@@ -163,7 +163,7 @@ Finally, to connect the `blogPost` Blok to the `BlogPost` component, add a new p
 - The key is the name of the Blok. In this case, it is `blogPost`.
```
kevinzunigacuellar, line 163:
```suggestion
- The key is the name of the Blok in Storyblok. In this case, it is `blogPost`.
```

src/pages/en/guides/cms/storyblok.md
```diff
@@ -85,8 +85,8 @@ export default defineConfig({
         // Add your components here
       },
       apiOptions: {
-        // Choose your Storyblok space region 
-        region: 'us',
+        // Choose your Storyblok space region
+        region: 'us', // or 'eu'
```
edv-ns, line 89: [optional] `// or 'eu' (default)`

src/pages/en/guides/cms/storyblok.md
```diff
@@ -185,6 +185,11 @@ export default defineConfig({
 });
 ```
 
+:::caution
+  Your components should exactly match your Storyblok Blok library "technical name" field. If these don't match, or you're trying to reference a component that doesn't exist in Storyblok you'll get an error.
+  :::
```
kevinzunigacuellar, line 190: I think we can move this to line 165. Right after the docs explain what they `key` and `value` are.
```suggestion
:::caution
The `key` should exactly match your Blok name in Storyblok to be referenced correctly. If these don't match, or you're trying to reference a component that doesn't exist in Storyblok you'll get an error.
:::
```

edv-ns (commented):
Additionally, since you are touching this file, maybe you could remove the mention of `axios@^0.27.2`, because starting `storyblok/astro@2.0.0` the axios is no longer a dependency https://github.com/storyblok/storyblok-astro#installation

edv-ns:
Additionally, since you are touching this file, maybe you could remove the mention of `axios@^0.27.2`, because starting `storyblok/astro@2.0.0` the axios is no longer a dependency. See https://github.com/storyblok/storyblok-astro#installation

src/pages/en/guides/cms/storyblok.md
```diff
@@ -379,20 +390,24 @@ export async function getStaticPaths() {
...
+
+const { story } = Astro.props
 ---
 <html lang="en">
   <head>
     <title>Storyblok & Astro</title>
   </head>
   <body>
-    <StoryblokComponent blok={content} />
+    {story.full_slug === 'home' && (
```
kevinzunigacuellar, line 407: For the reader's benefit, I think we should keep the slug page as simple as possible. In this case, what do you think about removing the h1 element but keeping the story to be passed as props? Then the reader can use all the metadata stored in the story to build their page. 🎉

src/pages/en/guides/cms/storyblok.md
```diff
@@ -424,6 +439,9 @@ try {
     <title>Storyblok & Astro</title>
   </head>
   <body>
+    {story.full_slug === 'home' && (
```
kevinzunigacuellar, line 442: Referencing my previous comment. I think we can replace the content variable with a story variable. So the `<StoryblokComponent blok={story.content} />` will match in both examples and the reader have access to the story metadata.

kevinzunigacuellar (commented):
Thanks a lot of the PR @danspratling! I left a few comments about removing the title `(<h1>)`. My intention for `blogPostList` was to illustrate to the reader that the Page blok can contain other bloks such as `blogPostList`

Also let me know what do you think about replacing the content prop for the story prop.

src/pages/en/guides/cms/storyblok.md
kevinzunigacuellar: The region is mentioned in a callout before but it doesn't hurt to be repeated. I spent a good 2 hours debugging this small option when I first used Storyblok and that's the reason It is included here.

kevinzunigacuellar:
> Additionally, since you are touching this file, maybe you could remove the mention of `axios@^0.27.2`, because starting `storyblok/astro@2.0.0` the axios is no longer a dependency. See https://github.com/storyblok/storyblok-astro#installation

Actually, I created an issue to update the setup section (to match their new release). I haven't had the opportunity to try their newest version yet but if that's the only change we can include this in this PR assuming @danspratling is using v2.

danspratling:
Thanks for reviewing so quickly!

Yep I am using v2 @kevinzunigacuellar and didn't notice any other issues with the guide (Always love a smooth upgrade!)

It's late for me but I'm happy to review the comments again tomorrow and make the suggested changes/amends

edv-ns:
> > Additionally, since you are touching this file, maybe you could remove the mention of `axios@^0.27.2`, because starting `storyblok/astro@2.0.0` the axios is no longer a dependency. See https://github.com/storyblok/storyblok-astro#installation
> 
> Actually, I created an issue #2183 to update the setup section (to match their new release). I haven't had the opportunity to try their newest version yet but if that's the only change we can include this in this PR assuming @danspratling is using v2.

Yup, this PR may potentially also closes #2183 💪 We may mention it in description.

src/pages/en/guides/cms/storyblok.md
danspratling: Agreed. I think it's good to mention both because following the storybook docs, they only suggest adding 'us' if you need it (leave blank for eu).

If you're skimming the article it'd be very easy for an EU user to miss that this has been flipped, even with the which is why I added the comment to help draw attention to this area even if you're quickly copying and pasting.

src/pages/en/guides/cms/storyblok.md
danspratling: When I was building this tutorial, and debugging some of the issues above, I was struggling to see the difference between the homepage and the blog posts

As they both use the same file, it wasn't clear which I was looking at at any given time, or which was causing errors.

This isn't a problem with storyblok or astro really, the usage of `[...slug]` is correct, but I think that being a little redundant here can actually help people getting started if they have any problems.

If we want to handhold their mistakes, I think the title is helpful, but if we don't then we could let them figure out this debugging step themselves.

danspratling:
@kevinzunigacuellar @edv-ns Updated per all comments! Let me know if there's anything else 😄

edv-ns:
LGTM ✅ Good job!

kevinzunigacuellar:
I will let @sarah11918 give the final touches, thanks to everyone for their help 💜

src/pages/en/guides/cms/storyblok.md
```diff
@@ -160,10 +160,14 @@ Your root directory should include this new file:
 
 Finally, to connect the `blogPost` Blok to the `BlogPost` component, add a new property to your components object in your Astro config file. 
 
-- The key is the name of the Blok. In this case, it is `blogPost`.
+- The key is the name of the Blok in Storyblok. In this case, it is `blogPost`.
 - The value is the path to the component. In this case, it is `storyblok/BlogPost`.
 
-```js title="astro.config.mjs" ins={15}
+:::caution
+  The `key` should exactly match your Blok name in Storyblok to be referenced correctly. If these don't match, or you're trying to reference a component that doesn't exist in Storyblok you'll get an error.
```
sarah11918, line 167:
```suggestion
  The `key` should exactly match your Blok name in Storyblok to be referenced correctly. If these don't match, or you're trying to reference a component that doesn't exist in Storyblok, you'll get an error.
```

src/pages/en/guides/cms/storyblok.md
```diff
@@ -299,15 +303,19 @@ const content = renderRichText(blok.content)
 
 It uses the `useStoryblokApi` hook to fetch all the stories with the content type of `blogPost`. It uses the `version` query parameter to fetch the draft versions of the stories when in development mode and the published versions when building for production.
 
-```astro title="src/pages/blogPostList.astro"
+We aren't passing any props just yet so `Astro.props` is just used to set up the editor in Storyblok, but we could choose to pass additional props if we wanted.
```
sarah11918, line 306:
```suggestion
`Astro.props` is used to set up the editor in Storyblok. Additional props can also be passed to your component here, if needed.
```

sarah11918:
Thanks and Welcome to Team Docs, @danspratling! 🥳  

Way to step up, @edv-ns and @kevinzunigacuellar!

Merry Christ`merge`, every one! 🎄 🎁

PR #2242, opened 2022-12-26
fix: update WordPress cms guide formatting and examples

lukemcdonald:

- Updated code examples to not throw an error on single dino pages
- Improved semantics of code examples
- Remove extra spacing and line breaks for consistency

#### Description

The code example to display post content on single dino pages was throwing an error. The content returned from the API is an array of post objects. The change made destructures the first post from the response.

Additionally, some markup was updated to improve semantics. In some cases, a paragraph was being used to `set:html` content. In this case, the content returned included additional markup including p tags and other block level elements. In other places, tags were not being properly closed.

Lastly, a few minor edits were made to add and remove line breaks and spacing for consistency.


kevinzunigacuellar:
I was thinking maybe we should switch to the wp graphql endpoint to make the markup cleaner. I've heard that's the approach most people take to make WP sites static. or we can have both 🎉

sarah11918:
@kevinzunigacuellar That's not a bad idea! I think you're right, and GraphQL is popular for this. Let's consider that an addition to make after the site is converted to MDX? 

We're now in this stage where we need to think about what's easy to do before/after (while!) merging the MDX PR, so let's keep everything currently PR'd as simple as possible and wait to make new ones until the site is converted over to MDX!

src/pages/en/guides/cms/wordpress.md
```diff
@@ -124,62 +123,55 @@ import Layout from '../../layouts/Layout.astro';
...
 ```astro title="/src/pages/dinos/[slug].astro" {3}
 <Layout title={post.title.rendered}>
-  <section >
+  <article>
     <img src={post._embedded['wp:featuredmedia']['0'].source_url} />
-    <h2 set:html={post.title.rendered}>
-    <article set:html={post.content.rendered} />
-  </section>
+    <h1 set:html={post.title.rendered} />
+    <div set:html={post.content.rendered} />
```
kevinzunigacuellar, line 168: I am not sure what the response is but perhaps this could be a good use for our <Fragment> component 
```suggestion
    <Fragment set:html={post.content.rendered} />
```

kevinzunigacuellar (commented):
Everything looks good! We can get the ball rolling for graphql in a separate PR. Thanks a lot @lukemcdonald

sarah11918:
I think @kevinzunigacuellar thought this was good to go, so I'm going to release it, and we can PR any corrections if necessary!  :partying_face:

PR #2249, opened 2022-12-28
Add i18n for screen reader section labels

### LINKED ISSUE #1586: Make screen reader section labels translatable
We use custom configuration `rehype-autolink-headings` to wrap our headings and add the clickable link icon. This includes a screenreader label which we inject but this is currently hardcoded in English:

https://github.com/withastro/docs/blob/41ea6653d78cfa90d96cd41019dbf1440a07f2af/astro.config.ts#L37

We should make this translatable like our other language strings.

Yan-Thomas:

- Translated content

#### Description

- Closes #1586 

This PR adds a new rehype plugin to translate the screen reader anchor labels injected by `rehype-autolink-headings`. If the current language doesn't have a translation yet, it will fallback to English.

| PT-BR (translated)  | ES (fallback) |
| image | image |

### DIFF
--- a/src/i18n/util.ts
+++ b/src/i18n/util.ts
@@ -98,14 +98,10 @@ export function useTranslations(
 	Astro: Readonly<AstroGlobal>
 ): (key: UIDictionaryKeys) => string | undefined {
 	const lang = getLanguageFromURL(Astro.url.pathname) || 'en';
-	return function getTranslation(key: UIDictionaryKeys) {
-		const str = translations[lang]?.[key] || translations[fallbackLang][key];
-		if (str === undefined) console.error(`Missing translation for “${key}” in “${lang}”.`);
-		return str;
-	};
+	return useTranslationsForLang(lang as UILanguageKeys);
 }
 
-export function useTranslationsFromString(
+export function useTranslationsForLang(
 	lang: UILanguageKeys
 ): (key: UIDictionaryKeys) => string | undefined {
 	return function getTranslation(key: UIDictionaryKeys) {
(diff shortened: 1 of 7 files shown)


src/i18n/util.ts
```diff
@@ -104,3 +104,13 @@ export function useTranslations(
 		return str;
 	};
 }
+
+export function useTranslationsFromString(
```
delucis, line 108: I might suggest renaming this `useTranslationsForLang`. What do you think?

src/i18n/util.ts
```diff
@@ -104,3 +104,13 @@ export function useTranslations(
 		return str;
```
delucis, line 104: Could we refactor `useTranslations` to use the new function and avoid duplication?

```js
export function useTranslations(
	Astro: Readonly<AstroGlobal>
): (key: UIDictionaryKeys) => string | undefined {
	const lang = getLanguageFromURL(Astro.url.pathname) || 'en';
	return useTranslationsForLang(lang);
}
```

delucis (commented):
Clever solution! Shame `rehype-autolink-headings` doesn’t provide `file` etc. as context to its different methods so we could just plug this in there. Made a small naming suggestion & refactor recommendation.

LOVE that this is your first PR back — i18n + a11y in one go 🙌

src/i18n/util.ts
Yan-Thomas: Agree, smart suggestions!

delucis (approved):
LGTM! Welcome back @Yan-Thomas 🚀

PR #2250, opened 2022-12-28
Injected frontmatter change

bholmesdev:

- New or updated content

#### Description

- Bump injected frontmatter to an L3 heading, and rename to "modifying frontmatter programmatically." Users have typically asked how to "modify" frontmatter in support forums. Adding a TOC entry with "modifying" instead of "injection" may help discoverability!
- Add tip that `data.astro.frontmatter` now includes all document frontmatter properties


src/pages/en/reference/api-reference.mdx
```diff
@@ -830,7 +830,7 @@ A function to compile a given Markdown or MDX document for rendering. This retur
 
 - `<Content />` - A component used to render the document's contents in an Astro file.
 - `headings` - A generated list of headings, [mirroring Astro's `getHeadings()` utility](/en/guides/markdown-content/#exported-properties) on Markdown and MDX imports.
-- `injectedFrontmatter ` - An object of frontmatter [injected via remark or rehype plugins](/en/guides/markdown-content/#example-injecting-frontmatter). Set to type `any`.
+- `injectedFrontmatter ` - An object of frontmatter [modified via remark or rehype plugins](/en/guides/markdown-content/#modifying-frontmatter-programmatically). Set to type `any`.
```
sarah11918: Just checking that you still want to name this `injectedFrontmatter` even though all our descriptions now refer to it as "modified frontmatter"?  (I understand that changing the code is probably a bigger deal than changing words in the docs. Just pointing it out!)

bholmesdev: I'm aware, and I'm debating a good name for it 😓 
`pluginFrontmatter` is the rough name I'm going with?

src/pages/en/guides/markdown-content.mdx
```diff
@@ -497,6 +497,12 @@ You can add frontmatter properties to all of your Markdown and MDX files by usin
     }
     ```
 
+    :::tip
+    <Since v="2.0.0" />
```
delucis, line 501: Looks like we need to import this component to fix the build. This at the top of the file should do it:
```js
import Since from '~/components/Since.astro'
```

src/pages/en/reference/api-reference.mdx
bholmesdev: Update: decided on `remarkPluginFrontmatter`! Remark is the umbrella for "remark and rehype plugins" according to the unified spec. Also want to clarify this is the frontmatter from remark _only,_ and won't reflect any schemas you have applied.

src/pages/en/guides/markdown-content.mdx
bholmesdev: Ah yep, was modifying this file mid-MDX migration and missed it 👍

src/pages/de/guides/markdown-content.mdx
```diff
@@ -4,6 +4,8 @@ title: Markdown
 description: Verwendung von Markdown mit Astro
 ---
 
+import Since from '~/components/Since.astro'
```
delucis, line 7: Wrong language, this is the German file 😅

bholmesdev: Verdammt

src/pages/en/guides/markdown-content.mdx
```diff
@@ -497,6 +498,12 @@ You can add frontmatter properties to all of your Markdown and MDX files by usin
...
+
+    `data.astro.frontmatter` contains all properties from a given Markdown or MDX document. This allows you to modify existing frontmatter properties, or compute new properties based on other properties.
```
sarah11918, line 504: nit: is there a better description of "other" properties?

Do you mean exactly the file's existing frontmatter properties?

If so, I could see:
> based on these existing properties
> based on these current properties
> compute new properties programmatically from the existing properties

If you truly mean, from "all other random things in the universe" then I might be tempted to just say "or compute new properties." and leave it at that, to avoid saying properties yet again.  😅

src/pages/en/reference/api-reference.mdx
```diff
@@ -830,14 +830,14 @@ A function to compile a given Markdown or MDX document for rendering. This retur
 
 - `<Content />` - A component used to render the document's contents in an Astro file.
 - `headings` - A generated list of headings, [mirroring Astro's `getHeadings()` utility](/en/guides/markdown-content/#exported-properties) on Markdown and MDX imports.
-- `injectedFrontmatter ` - An object of frontmatter [injected via remark or rehype plugins](/en/guides/markdown-content/#example-injecting-frontmatter). Set to type `any`.
+- `remarkPluginFrontmatter ` - The [modified frontmatter object](/en/guides/markdown-content/#modifying-frontmatter-programmatically) applied via remark or rehype plugins. Set to type `any`.
```
sarah11918, line 833:
```suggestion
- `remarkPluginFrontmatter ` - The modified frontmatter object after any [remark or rehype plugins have been applied](/en/guides/markdown-content/#modifying-frontmatter-programmatically). Set to type `any`.
```
Maybe?

sarah11918 (approved):
Looks great, @bholmesdev! I think the change is so great for usablility!  🚀 

Some non-blocking comments for thought!

src/pages/en/guides/markdown-content.mdx
bholmesdev: How many properties could a prop shop prop if a prop shop could prop properties?
...okay "or compute new properties" is fine 😆

PR #2254, opened 2022-12-29
Document `vite.ssr.noExternal`

### LINKED ISSUE #1455: Documenting `vite.ssr.noExternal`
Vite externalizes SSR dependencies by default: https://vitejs.dev/guide/ssr.html#ssr-externals

A lot of npm packages are not ESM-ready so it leads to errors when using ESM imports.

This can be fixed by forcing those packages to be transpiled by vite using [`vite.ssr.noExternal`](https://vitejs.dev/config/ssr-options.html#ssr-noexternal) in astro.config

```js
vite: {
  ssr: {
    noExternal: ['package-name']
  }
}
```

Since it's not obvious how to arrive at this solution (and vite's docs are a bit sparse), maybe this should be added somewhere in the astro docs, e.g. the troubleshooting section?

Also there are probably other cases for using `vite.ssr.noExternal` that I can't think of right now.

mayank99:

New or updated content

#### Description

Closes #1455 

1. In configuration-reference: mentioned `vite.ssr.noExternal` linking to vite docs, and added example.
2. In imports guide: linked back to `vite` section in configuration-reference
3. ~~In troubleshooting: added more explanation of ESM vs CJS and repeated same example of `vite.ssr.noExternal`.~~ removed


src/pages/en/guides/troubleshooting.mdx
```diff
@@ -71,6 +71,31 @@ export default defineConfig({
 
 Alternatively, you can update your CSP to include `script-src: 'unsafe-inline'` to allow inline scripts to run.
 
+### exports is not defined
```
mayank99, line 74: I used one of the error messages I've seen before, but I think this issue can manifest in many different kinds of errors, so I'm not really sure if this is the right move.

src/pages/en/guides/imports.mdx
```diff
@@ -79,16 +79,17 @@ While Astro understands JSX syntax out-of-the-box, you will need to include a fr
...
 import { Icon } from 'astro-icon';
 ---
 ```
-If you've installed an NPM package, you can import it in Astro. Even if a package was published using a legacy format, Astro will convert the package to ESM so that `import` statements work.
+If a package was published using a legacy format, Astro will try to convert the package to ESM so that `import` statements work. In some cases, you may need to adjust your [`vite` config](/en/reference/configuration-reference/#vite) for it to work.
 
-:::warning
 Some packages rely on a browser environment. Astro components runs on the server, so importing these packages in the frontmatter may [lead to errors](/en/guides/troubleshooting/#document-or-window-is-not-defined).
-:::
+
```
mayank99, line 92: I noticed that this `:::warning` admonition thing was not doing anything in the live version so i removed it.

src/pages/en/guides/troubleshooting.mdx
```diff
@@ -71,6 +71,31 @@ export default defineConfig({
...
+
+This error can occur when you try to use a CommonJS module (CJS) in an ES module (ESM) environment.
+
+Astro uses Vite which generates [ES modules during SSR](https://vitejs.dev/guide/ssr.html#ssr-format) and [externalizes SSR dependencies](https://vitejs.dev/guide/ssr.html#ssr-externals) by default. This means that npm packages need to use the ESM format to work out-of-the-box.
+
+A common scenario is when you're trying to use an npm package that only supports the CJS format, or one of its dependencies only supports CJS.
```
bluwy, line 80: Vite does support CJS dependencies in SSR and during build, so npm packages doesn't need to use ESM. I think the common issue is that the build output could be:

```js
import { named } from 'cjs-pkg'
```

And this may sometimes not work in nodejs, which it recommends doing:

```js
import pkg from 'cjs-pkg'
const { named } = pkg
```

instead. `noExternal` does fix this case too. Perhaps we want to re-phrase the title and description about this error instead? I don't think I've seen `exports is not defined` error a lot.

src/pages/en/guides/troubleshooting.mdx
```diff
@@ -71,6 +71,31 @@ export default defineConfig({
...
+
+**Solution:** Ideally, you should try to work with the package author to support ESM, or use a different package. In some cases, you may be able to work around it by forcing Vite to tranform the broken package using `vite.ssr.noExternal` in your astro config.
```
bluwy, line 82:
```suggestion
**Solution:** Ideally, you should try to work with the package author to support ESM, or use a different package. In some cases, you may be able to work around it by forcing Vite to tranform the broken package using `vite.ssr.noExternal` in your Astro config.
```

src/pages/en/guides/troubleshooting.mdx
mayank99: i've removed this section entirely. i'm not really sure how best to document it

src/pages/en/guides/imports.mdx
delucis: Ah, this is because it should have been `:::caution` — mind adding it back? Might as well fix it while we’re here.

mayank99: done

delucis: Wow! Lightspeed dev 🚀

src/pages/en/reference/configuration-reference.mdx
```diff
@@ -546,8 +546,21 @@ Pass additional configuration options to Vite. Useful when Astro doesn't support
 
 View the full `vite` configuration object documentation on [vitejs.dev](https://vitejs.dev/config/).
 
+A common use case is to pass a package name to [`vite.ssr.noExternal`](https://vitejs.dev/config/ssr-options.html#ssr-noexternal) to prevent it from being externalized.
```
delucis, line 549: Do we have a better way to explain/describe what “externalized” means? The code comment about packages being “transformed for SSR” would already be better I think.

mayank99: i figured the code comment plus the link to vite docs would be sufficient

what about this:
```suggestion
A common use case is to pass a package name to [`vite.ssr.noExternal`](https://vitejs.dev/config/ssr-options.html#ssr-noexternal) to prevent it from being externalized from Vite's SSR transform system.
```
or another link
```suggestion
A common use case is to pass a package name to [`vite.ssr.noExternal`](https://vitejs.dev/config/ssr-options.html#ssr-noexternal) to prevent it from being [externalized](https://vitejs.dev/guide/ssr.html#ssr-externals).
```

bluwy: +1 for the second choice. The "externalized" concept is a bit hard to explain in brief so an external link seems to be the best alternative for me.

I also think that maybe we can remove this paragraph, and instead put the link and brief reason in the example's comment instead?

delucis: Links in code comments aren’t clickable so my preference would be to keep them in a paragraph. You both seem to like the second option, so let’s go with that. Thanks!

delucis (approved):
Thanks for your work on this @mayank99 & @bluwy — I think this is good to go 🙌

PR #2255, opened 2022-12-29
[troubleshooting] remove suggested workaround re: executing inline scripts

### LINKED ISSUE #2150: Astro inlines script and style for astro islands even with assetsInlineLimit set to 0
Astro's docs here: https://docs.astro.build/en/guides/troubleshooting/#refused-to-execute-inline-script suggest the following config to force script and styles to be not inlined, so that an astro generated site is comptible with Content-Security-Policy that doesn't allow unsafe-inline:

```
export default defineConfig({
  vite: {
    build: {
      assetsInlineLimit: 0,
    },
  },
});
```

This works for a site that uses simple Astro components. It stops working once you start using the Astro Islands architecture to hydrate e.g. svelte components. Astro always inlines the island hydration script and related style, even wiht the above configuration.

It would be nice if the documentation could be updated to reflect this. Hopefully with an alternative way to achieve the island hydration script and style not to be inlined in the html..

sarah11918:
- New or updated content

Closes #2150 (The documentation issue; the greater conversation is ongoing elsewhere.)

As per @delucis suggestion and @matthewp approval, removes a problematic workaround that does not work in every case, and instead just leaves the suggestion to update your CSP.

(Note: eventually error messages will make their way to the Error Reference page. But, at least this removes potentially harmful advice for now.)

### DIFF
--- a/src/pages/en/guides/troubleshooting.mdx
+++ b/src/pages/en/guides/troubleshooting.mdx
@@ -54,22 +54,7 @@ You may see the following error logged in the browser console:
 
 This means that your site’s [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP) disallows running inline `<script>` tags, which Astro outputs by default.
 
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM — long term we do need a proper answer to this, but this will do for now.

(I was kind of unsure about removing as this workaround does work for some people so isn’t totally unhelpful, but probably not right for docs.)

PR #2256, opened 2022-12-29
Refactor markFallbackNavEntries dir

bluwy:

#### Description

Refactor `markFallbackNavEntries`'s way of deriving the lang directory. The assumption of the function being compiled into `dist/entry.mjs` isn't guaranteed, as it could also be chunked into the `dist/chunks/` directory, making the path incorrect. The latter happens for Rollup 3.

This fixes to derive from `process.cwd()` instead. This unblocks https://github.com/withastro/astro/pull/5685


delucis (approved):
Tested the non-English sections of the preview deploy and fallback indicators are appearing as expected, so LGTM 👍

Issue #1732, opened 2022-10-06
Cookies: Needs guide + examples


Jutanium:
Now that the [Cookies API](https://docs.astro.build/en/reference/api-reference/#astrocookies) is documented on our references page, we need to provide some instructional content and examples. 

This probably makes the most sense on the SSR page, as cookies are specific to SSR mode.

wassfila:
I have an example I'd be happy to contribute it to the Astro examples if needed
https://github.com/MicroWebStacks/astro-examples#13_client-persistent-counter

sarah11918:
Sounds great, @wassfila! 🙌 

Looking at the SSR page now, it seems like something alongside and approx the same length/depth as `Astro.request.headers` and `Astro.redirect` would fit. That is, a very basic, minimal usage example to show `Astro.cookies` in use, and then link to the reference page for the details. This page isn't the place for a full-featured example, but it would be nice to have a visual of it in use.

pilcrowOnPaper:
I think this should be closed?

delucis:
Oh yes, well spotted @pilcrowOnPaper! Thanks again @wassfila 🙌

PR #2259, opened 2022-12-30
Cookies SSR Page Guide + example

wassfila:

- New content for cookies doc in SSR page

#### Description

- Closes # https://github.com/withastro/docs/issues/1732
- has code snippet like other examples in the page
- I have tested this code, but please review


src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -121,6 +121,32 @@ The features below are only available at the page level. (You can't use them ins
 This is because these features modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header), which can't be modified after they're sent to the browser. In SSR mode, Astro uses HTML streaming to send each component to the browser as it renders them. This makes sure the user sees your HTML as fast as possible, but it also means that by the time Astro runs your component code, it has already sent the Response headers.
 :::
 
+### `Astro.cookies`
+
+On the `Astro` global, this interface allows to manipulate cookies. It allows to directly get them from requests and set them in responses. Whereas `Astro.request.headers.get('cookie')` gets all cookies and still need to be parsed, `Astro.cookies.get` gets a single cookie in an object.
+
+More details about the API in the reference [Astro.cookies](/en/reference/api-reference/#astrocookies)
```
sarah11918, line 128:
```suggestion
This is a utility to read and modify a single cookie. It allows you to check, set, get and delete a cookie.

Use `Astro.request.headers.get('cookie')` to get all cookies at once.

See more details about [`Astro.cookies` and the `AstroCookie` type](/en/reference/api-reference/#astrocookies) in the API reference.
```

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -121,6 +121,32 @@ The features below are only available at the page level. (You can't use them ins
...
+
+The example below shows how to update the value of a counter cookie.
```
sarah11918, line 130:
```suggestion
The example below updates the value of a cookie that updates a page view counter.
```
(I'm assuming that's what this counter is? Is worth mentioning I think because so often we use counters with + and - buttons as examples. That's probably what people think of when they first see "counters" in docs!)  :smile:

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -121,6 +121,32 @@ The features below are only available at the page level. (You can't use them ins
...
+
+```astro title="src/pages/index.astro" {4,10}
+---
+let counter = 0
+
+const cookie = Astro.cookies.get("counter")
+
+if(cookie.value != undefined){
```
sarah11918, line 138: Just checking here, because admittedly I'm not familiar with using this, but would this be an opportunity to write something using `Astro.cookies.has()` so we can showcase more of the API?

sarah11918 (commented):
Thanks so much for this contribution @wassfila :raised_hands: It's nice to see an example of this for docs!

I tried to shorten this into more "Learn" style content vs "Reference" style content, especially when there are links back to the API. 

I'm also tempted to put this 3rd, *after* `.request.headers` and `.redirect` because they both  use `Astro.request.headers.get('cookie')` in their examples. This one illustrates just a single cookie, and so maybe feels like something that would come after examples that get all cookies on the page.

Let me know what you think!

src/pages/en/guides/server-side-rendering.mdx
wassfila: yes, sure, I even hesitated about that and mentioned it. Now I added it and pushed, also updated highlights.

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -121,6 +121,33 @@ The features below are only available at the page level. (You can't use them ins
...
 
+### `Astro.cookies`
+
+This is a utility to read and modify a single cookie. It allows you to check, set, get and delete a cookie.
+
+Use `Astro.request.headers.get('cookie')` to get all cookies at once.
+
+See more details about [`Astro.cookies` and the `AstroCookie` type](/en/reference/api-reference/#astrocookies) in the API reference.
+
+The example below updates the value of a cookie that updates a page view counter.
```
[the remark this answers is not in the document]
sarah11918: I was a little wordy there, eh?  😅

wassfila:
@sarah11918 as a sum up, yes, I agree with all your suggestions, you have a better visibility how the end result should look like. And it all makes it somehow clearer.

sarah11918:
Just waiting on a new deploy preview to check it visually! What do you think about moving `Astro.cookie` section to below `.redirect`?

wassfila:
> Just waiting on a new deploy preview to check it visually! What do you think about moving `Astro.cookie` section to below `.redirect`?

I'm not sure if that is necessary, it feels to me closer to the Astro.request as Astro.cookies is nothing but an Astro.response manipulation in the .astro frontmatter. The redirect and Response sections are then a bit different and more advanced because they no longer rely on the Astro taking care of the request/response.

Also the Caution right before the Astro.cookies does fit well the Astro.cookies `This is because these features modify the Response headers` 

This being said, I also think the dependency is pretty minimal and after redirect would be just fine if you have some other logic in mind even if it's just a fit intuition, you can proceed swapping them, I actually thought it was done but your commtis during the suggestions, or just let me know and it would be a matter of seconds (but only when I'm at the big PC :) ) and not mobile.

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -121,6 +121,33 @@ The features below are only available at the page level. (You can't use them ins
 This is because these features modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header), which can't be modified after they're sent to the browser. In SSR mode, Astro uses HTML streaming to send each component to the browser as it renders them. This makes sure the user sees your HTML as fast as possible, but it also means that by the time Astro runs your component code, it has already sent the Response headers.
 :::
 
+### `Astro.cookies`
+
+This is a utility to read and modify a single cookie. It allows you to check, set, get and delete a cookie.
+
+Use `Astro.request.headers.get('cookie')` to get all cookies at once.
```
matthewp, line 128: This gets the cookie string header. Given there isn't an explanation for this or why you might want it, should this sentence be here?

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -121,6 +121,33 @@ The features below are only available at the page level. (You can't use them ins
...
+
+The example below updates the value of a cookie for a page view counter.
+
+```astro title="src/pages/index.astro" {4,5,9}
+---
+let counter = 0
+
+if(Astro.cookies.has("counter")){
+  const cookie = Astro.cookies.get("counter")
+	counter = parseInt(cookie.value) + 1
```
matthewp, line 140: It's a little easier to do `cookie.number()` which parses it for you.

src/pages/en/guides/server-side-rendering.mdx
wassfila: But it's from the example right before, without any mention people might be confused what the difference between them are. If that is not a concern, then can be dropped for simplicity

src/pages/en/guides/server-side-rendering.mdx
wassfila: Yeah, you meant `cookie.value.number()`, why not.

src/pages/en/guides/server-side-rendering.mdx
sarah11918: I think the fact that we've emphasized a *single* cookie makes this clear enough without the extra statement?

src/pages/en/guides/server-side-rendering.mdx
matthewp: It should just be `cookie.number()`. `cookie.value` is the raw string value. The definition of the return value of get() is here: https://github.com/withastro/astro/blob/9ef0c50427e658e0a6a4f20b14ff39c67fed76a6/packages/astro/src/core/cookies/cookies.ts#L37

wassfila: right, so it's an object API, did not realise that, then it's good to add it to the example

wassfila:
So, updated :

- .number()
- removed Astro.headers statement
- Let me know of your final preference with regard to the order, where it is now or below .redirect ?

sarah11918:
Thanks, this looks good to me! Let's ship it!

PR #2262, opened 2022-12-30
Consistent installation title

Tc-001:

#### Description

Both CLI and manual installation now have the same title to help give the illusion of tabs

Discussion for reference: https://discord.com/channels/830184174198718474/830184175176122389/1058459575789027338

I chose "Installation" because it is consistent with the sidebar as well, as it does in most other pages.
Can be  changed if there is a better name, I guess "Installation (automatic)" could work as well

### DIFF
--- a/src/pages/en/install/auto.mdx
+++ b/src/pages/en/install/auto.mdx
@@ -1,5 +1,5 @@
 ---
-title: Install Astro with the Automatic CLI
+title: Installation
 description: 'How to install Astro with NPM, PNPM, or Yarn via the create-astro CLI tool.'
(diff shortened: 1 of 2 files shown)


sarah11918:
LGTM! @Tc-001

delucis:
Ah, sorry I missed this discussion! I’m not sure this was a good idea: different pages _should_ have different titles to distinguish them, not just for users on those pages, but also in search results in and outside of docs. I think we should probably revert this. The real issue is with our “fake” tabs — we can probably solve this with a better UI here instead.

PR #2265, opened 2022-12-31
Revert "Consistent installation title"

delucis:
Reverts withastro/docs#2262 — see https://github.com/withastro/docs/pull/2262#issuecomment-1368137847

### DIFF
--- a/src/pages/en/install/auto.mdx
+++ b/src/pages/en/install/auto.mdx
@@ -1,5 +1,5 @@
 ---
-title: Installation
+title: Install Astro with the Automatic CLI
 description: 'How to install Astro with NPM, PNPM, or Yarn via the create-astro CLI tool.'
 layout: ~/layouts/MainLayout.astro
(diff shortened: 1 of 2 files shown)


Tc-001:
Fair
This *could* also be solved with similar length titles so that there isn't a line wrap difference, so no vertical shift when navigating between them.

I will probably be asleep for most of the discussion, so those are my 2 cents.
Good luck with the new year and/or wiki stuff!

delucis:
Yes, that’s true! Maybe we can think of a good way to still have a difference while being short enough to avoid the line break.

sarah11918:
D'oh! Sorry @delucis - I took a quick peek to make sure this wouldn't *break* anything, but didn't fully account for other goings-on on the site. I thought a unified title might be better because it would match how we present it like one page in the sidebar, but obviously for the reasons you mentioned, it's not that simple.

LGTR (looks good to revert!)

PR #2267, opened 2023-01-01
Align left edge of Tabs and TabGroup

morinokami:

#### Description

- Tweaks the design of the Tabs and TabGroup components to align their left edge.
  - Sorry if this was intentional😅, but only the left side of the Tabs component was outside of the container, which was a bit odd to me, so I created this PR to fix that.
  - (Update) Found that TabGroup had the same style, so I updated it as well for consistency.
  - This may be a PR based on my personal preference, so feel free to close it if it is inappropriate🙏

##### Mobile view

Tabs

| Before | After |
| --- | --- |
| mobile-before | mobile-after |

TabGroup

| Before | After |
| --- | --- |
| mobile-before | mobile-after |

##### Desktop view

Tabs

| Before | After |
| --- | --- |
| desktop-before | desktop-after |

TabGroup

| Before | After |
| --- | --- |
| desktop-before | desktop-after |

### DIFF
--- a/src/components/TabGroup/InstallGuideTabGroup.astro
+++ b/src/components/TabGroup/InstallGuideTabGroup.astro
@@ -8,7 +8,7 @@ const lang = getLanguageFromURL(currentPage);
 const segments = currentPage.split('/');
 ---
 
-<div class="TabGroup no-flex pull-wide">
+<div class="TabGroup no-flex">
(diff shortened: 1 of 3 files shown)


sarah11918:
Thanks, @morinokami! This does look nicer to me. I'll let @delucis handle this one officially!  :raised_hands:

delucis (approved):
Thanks @morinokami! This _was_ intentional (by me in the past) but I think I agree with you that this looks good (now I am in the present).

Issue #2042, opened 2022-11-11
Docs "search stats" investigation


sarah11918:
Hey Team Docs!

We have our first "data dump" of the search terms people have been using on the Docs site. Since we know these are all popular searches, we want to make sure that people are being pointed to the best places in docs to answer their questions.

So, **we're looking for interested people to a) try searching some of these terms and b) report back here in a reply re: whether you think our search widget is currently helpful to our readers**, returning good results etc. 

Of particular interest, things like:

1. How many *different pages* were represented by these 5 results?
2. Was the **top result** the "best" suggestion (in your mind - the most helpful/general page on the topic)? If not, was the first search term result at least a "good" suggestion?
3. Was a page entirely missing from the results that probably *should* have been included? (e.g. You searched for Markdown and did *not* see "Markdown & MDX" as one of the 5 results.)
4. Do you think more results would be helpful for this search term?
5. Do you feel that someone looking at the 5 results gets enough info about each search result to know the best one to choose?

And any other observations are welcome!
 
This is the list of our most-searched for terms in Docs. It's fine if multiple people report back their observations for the same term, but in the interest of getting as many covered as possible, I'll come back periodically and check ones off this list if at least one person has replied with their observations.  (And, I might update and add to the questions above too, as I think of new things, so keep checking back!)

_🔍 = carried out before search overhaul; 🔎 = carried out after search overhaul_

- [x] [image — 🔍](https://github.com/withastro/docs/issues/2042#issuecomment-1312513289)
- [x] [react — 🔍](https://github.com/withastro/docs/issues/2042#issuecomment-1312723775)
- [x] [tailwind — 🔍](https://github.com/withastro/docs/issues/2042#issuecomment-1320485371)
- [x] [props — 🔎](https://github.com/withastro/docs/issues/2042#issuecomment-1338835317)
- [x] [script — 🔍](https://github.com/withastro/docs/issues/2042#issuecomment-1312213968)
- [x] [slot — 🔎](https://github.com/withastro/docs/issues/2042#issuecomment-1369334301)
- [ ] client
- [x] [glob — 🔎](https://github.com/withastro/docs/issues/2042#issuecomment-1426952111)
- [ ] ssr
- [ ] mdx
- [x] [css — 🔎](https://github.com/withastro/docs/issues/2042#issuecomment-1548229593)
- [ ] 404
- [ ] env
- [ ] vite
- [ ] build
[...]

PR #2275, opened 2023-01-03
Fix categorisation of error reference pages for Algolia

delucis:

#### Description

Addresses some recent poor search results reported in #2042 

We add a hidden element to pages that Algolia picks up and uses to extract that page’s category, e.g.

```html
<span id="docsearch-lvl0" hidden>Learn</span>
```

Due to a bug in our category utility, we were categorising error reference pages as “Reference”. This PR fixes that so that these pages now contain:

```html
<span id="docsearch-lvl0" hidden>Error Reference</span>
```

Will still need to reevaluate these search terms (`getStaticPaths` and `slots` are current known issues) once this is live to see if the “Error Reference” category needs demoting in Algolia’s crawler logic.

### DIFF
--- a/src/util/getPageCategory.ts
+++ b/src/util/getPageCategory.ts
@@ -1,11 +1,15 @@
 // TODO: Move this data to our i18n system to support localized category labels.
 const defaultCategory = 'Learn';
+
+// Order is important here. Pages are tested to see if they *start* with one of
+// these paths and will return early when one matches. This means more specific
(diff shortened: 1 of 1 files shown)


Princesseuh (approved):
Make sense to me!

aFuzzyBear (approved):
Nice, 🤖

delucis:
Just a borked run. Re-running now.

PR #2276, opened 2023-01-03
[i18nIgnore] Remove Vue TypeScript troubleshooting section

bluwy:

- New or updated content

#### Description

Reflects the upcoming breaking change in https://github.com/withastro/astro/pull/5724. The troubleshooting section isn't an issue anymore as it has been fixed by Volar.

I also removed it for the translated docs.

### DIFF
--- a/src/pages/en/guides/typescript.mdx
+++ b/src/pages/en/guides/typescript.mdx
@@ -190,11 +190,3 @@ For the default setting of `jsxImportSource: react`, you would use:
 // For Solid
 /** @jsxImportSource solid-js */
 ```
-
-### Vue components are mistakenly typed by the `@types/react` package when installed
-
-The types definitions from the `@types/react` package are declared globally and therefore will be mistakenly used to typecheck `.vue` files when using [Volar](https://github.com/johnsoncodehk/volar).
-
-**Status**: Expected behavior.
-
-**Solution**: There's currently no reliable way to fix this, however a few solutions and more discussion can be found in [this GitHub discussion](https://github.com/johnsoncodehk/volar/discussions/592).
--- a/src/pages/es/guides/typescript.mdx
+++ b/src/pages/es/guides/typescript.mdx
@@ -189,11 +189,3 @@ Para la configuración por defecto `jsxImportSource: react`, deberías hacer lo 
 // Para Solid
 /** @jsxImportSource solid-js */
 ```
-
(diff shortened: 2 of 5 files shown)


delucis (changes requested):
Thanks @bluwy! Going to “request changes” just to make some red UI scare us away from merging accidentally before v2 is released. No actual changes needed.

Will also add an `en-only` keyword to the PR name — this means our i18n tracker (https://github.com/withastro/docs/issues/438) will ignore this commit and leave the translations in their current “needs updating” state instead of marking them as up-to-date.

sarah11918:
@delucis I think this can be merged now? Upstream PR is merged and released.

delucis:
Hmm. Interesting question. At the moment _most_ users will still be on a 1.x version, so I wonder in this case if it makes sense to only remove once we actually release 2.0 to stable not just beta.

General principle I’m thinking:
- For new _features_ available in v2 beta: merge early and flag they require v2
- For general housekeeping like this where our advice is currently valid, I’d lean towards merging _late_ once v2 proper is out.


What do you think?

sarah11918:
Makes sense! And, if a problem simply isn't going to come up in v2, then there's less danger having "advice for when this happens" there, because no one will read it!  ;)

We *could* update (rather than remove the thing entirely) with a note saying that upgrading to v2 (even beta) is a solution, as this problem is fixed in v2?

bluwy:
I think the only thing tying this change is that Astro 1.0 would have a hardcoded link to this section of the docs. In Astro 2.0, the link is removed so the docs can be removed as well. The fix for the section is to upgrade Volar (VSCode extension), and given most extension auto-updates, it could also be safe to remove this section early if needed 🤔

delucis:
So currently if a user runs into this, they would be shown a link to this section of the docs? That would make me lean towards remove later rather than earlier.

Unless you’re saying that most people will _already_ have upgraded Volar so this error is probably already rare?

bluwy:
Ah I miss this notification. Yeah they would be shown the old link, but only if they try to integrate Vue and and a JSX framework. Plus yes that Volar should already be upgraded to latest for most users so they shouldn't have an issue with it today. So I think this is a pretty rare case.

delucis:
Cool, sounds reasonably safe then to merge when you want.

sarah11918 (approved):

PR #2278, opened 2023-01-03
[tutorial] quick fixes from Hippo feedback

sarah11918:

This applies the "quick wins" feedback from @hippotastic after taking the tutorial.

Specifically:
- Fixes `localhost:3000` to no longer be `https:` 
- Gives a little more context re: WHY you might want to use the checklists/tutorial tracker.
- Correctly states the cause/effect of the GitHub/Netlify deployment relationship
- Minimally refers to "lessons" inside of "Unit 1" since, as Hippo points out, we use numbers for both Units and pages inside units. This provides at least a vocab early on to distinguish. (*Not sure I'm entirely happy yet, since Unit 0 is not named as such, so Hippo's totally right, when you're IN Unit 0, you only see a 0 and a 1 for those current lesson pages. You don't know that you aren't IN unit 1 when you're on Unit 0-lesson 1.*)

- Does NOT address the "on the first page, I completed the checkbox and "Build a blog" was marked completed. So, where's my blog??" totally valid comment!  :sweat_smile: This turned out to ***not*** be a "quick win" because with our infrastruture, the title of the page is the item that's checked off. As our main landing/index page, we WANT a generic "Build a blog!" statement. But, agreed that it might look strange to see "build a blog" marked as completed.

### DIFF
--- a/src/pages/en/tutorial/0-introduction/1.mdx
+++ b/src/pages/en/tutorial/0-introduction/1.mdx
@@ -21,13 +21,15 @@ You check them off!
 
 At the end of each page, you'll find a clickable checklist of tasks you should now be able to do. Check these items off to see your progress in the Tutorial Tracker.
 
+Using the tracker is optional, but it can help you remember your place if you complete the tutorial over multiple visits. You can also leave some checklists blank as a reminder of units that you want to revisit later.
+
 (This data is only saved to your browser's local storage, and is not available elsewhere. No data is sent to, nor stored by Astro.)
 </details>
 
 <details>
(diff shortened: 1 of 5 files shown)


hippotastic:
Thank you for these quick fixes! I agree that the "unit 1" reference isn't completely solved here, but everything is BTWWHB!

sarah11918:
I'm going to take this as an LGTM, Hippo?  :smile:

hippotastic:
Yes, that's what I meant! BTWWHB & LGTM! 😄

PR #2285, opened 2023-01-04
Fix integrations absolute links and changelog generation errors

Yan-Thomas:

#### Description

Lots of links to other docs pages and local links (those starting with `#`) were being wrongly prepended by the `absoluteLinks` remark plugin. I couldn't find a reason as to why the `absoluteLinks` remark plugin exists in the first place, taking a look at #954, I guess it was fixing something related to Markdown footnotes (maybe we can remove it?). Changed a line to make it ignore the links I didn't want it to change, but perhaps this can cause another problem. - I did some tests and couldn't find any broken links from the change in the remark plugin.

Additionally, I found out the Alpine.js integration changelog wasn't being deleted, so I made a small change to ensure its deletion as well.

### DIFF
--- a/scripts/generate-integration-pages.ts
+++ b/scripts/generate-integration-pages.ts
@@ -170,7 +170,11 @@ function absoluteLinks({ base }: { base: string }) {
 		function visitor(node: Link | Definition) {
 			// Sanitize URL by removing leading `/`
 			const relativeUrl = node.url.replace(/^.?\//, '');
-			node.url = new URL(relativeUrl, base).href;
+			// Don't add absolute path to local and docs links.
+			node.url =
+				node.url.startsWith('#') || node.url.startsWith('/')
+					? node.url
+					: new URL(relativeUrl, base).href;
 		}
 		visit(tree, 'link', visitor);
 		visit(tree, 'definition', visitor);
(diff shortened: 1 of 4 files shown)


scripts/generate-integration-pages.ts
```diff
@@ -170,7 +170,11 @@ function absoluteLinks({ base }: { base: string }) {
 		function visitor(node: Link | Definition) {
 			// Sanitize URL by removing leading `/`
 			const relativeUrl = node.url.replace(/^.?\//, '');
-			node.url = new URL(relativeUrl, base).href;
+			// Don't add absolute path to local and docs links.
+			node.url =
+				node.url.startsWith('#') || node.url.startsWith('/')
+					? node.url
+					: new URL(relativeUrl, base).href;
```
delucis, line 177: I think we only want to do this for `#` fragment links. Links starting with `/` are relative to `https://github.com/` in the README, so should be made absolute.

There do appear to be some bare links to docs in the MDX integration README, but these should be updated to use absolute links as those links currently won’t work for someone reading the README on GitHub or npm.

```suggestion
			node.url = node.url.startsWith('#') ? node.url : new URL(relativeUrl, base).href;
```

Will try to mop the various issues up:

- [ ] Update the README at source in the astro repo
- [ ] Update this PR with this suggestion & re-run
- [ ] Fix the anchor links that are incorrect currently:
  - [ ] astro PR for the error page
  - [ ] docs PR for translations

delucis (approved):
LGTM! I’ll address the broken translation links in a follow-up PR and the error page broken link will be resolved when we re-run the error docgen.

PR #2287, opened 2023-01-04
Ignore commits containing `i18nIgnore` in i18n status tracker

delucis:

- Something else!

#### Description

Adds a new keyword for ignoring commits in our i18n status tracker: commits containing `i18nIgnore` will be now also be ignored. Allows us to stop abusing other keywords like `typo` or `en-only`.


sarah11918 (approved):
Thanks! Maybe I'll remember this one!

PR #2280, opened 2023-01-03
Update Markdown content docs for config rework

bholmesdev:
#### What kind of changes does this PR include?

- New or updated content

#### Description

- Remove `smartypants` from Astro defaults
- Update MDX extend behavior

PR #2288, opened 2023-01-04
[i18nIgnore] Delete outdated content with broken link in es/ja

delucis:

#### Description

Spanish & Japanese translations of the Markdown content include text linking to a MDX subheading which no longer exists.

These paragraphs were wholesale removed here: https://github.com/withastro/docs/pull/2280/files#diff-5aa3990791d164a679ef35f083c2cc0e6afb3d4b007445fba1ee0d5adf9df89dL434-L442

So I think we can do the same here, fixing the broken link, but using the new `i18nIgnore` keyword to avoid marking these pages as updated as they do still need some work.


sarah11918 (approved):
Thanks! LGTM

PR #2300, opened 2023-01-05
[rss] Add to package on GitHub

sarah11918:
Adds a link to https://github.com/withastro/astro/tree/main/packages/astro-rss which was not previously on this page!


delucis (approved):
LGTM 👍

PR #2302, opened 2023-01-06
[tutorial] remove backticks inside spoiler text

sarah11918:
Quick fix because `code styling` doesn't seem to work inside our spoiler text for Analyze the Pattern in our tutorial.

spoiler

Without backticks:
spoilerfix


lloydjatkinson:
LGTM

PR #2303, opened 2023-01-06
i18n(pt-BR): Update `markdown-content.mdx` and `layouts.mdx` translations

Yan-Thomas:

- Translated content

#### Description

Updated `markdown-content.mdx` & `layouts.mdx` translations


delucis (commented):
Heads up that https://github.com/withastro/docs/commit/d7be2b759d34de14c4b1c06df0fab927150ebc25 made a small addition to `markdown-content`.

Overall this seems to look good to me, so if we can’t find a language reviewer, I think it’s OK if you want to merge without one after updating `markdown-content`. That way we can avoid needing to update this PR too much each time an English change comes in.

PR #2305, opened 2023-01-06
[i18nIgnore] Prep to remove `astroFlavoredMarkdown` in reference

delucis:

#### Description

- Removes links to the AFMD config entry in the `/en/` migration guide.
- Mark the migration guide as `i18nReady: false` — we’ll change this once the v1 -> v2 guide replaces this
- Removes links to the AFMD config entry in the `/ru/` Markdown Content guide.
- Uninstalls a direct dependency `MAINTAINERS.md` said was only needed for AFMD and updated that note

Does _not_ change Spanish, Portuguese, French, and Chinese pages as these should not cause any broken links. I think it’s safe to leave these as is until they get a proper refresh.

### DIFF
--- a/src/pages/en/migrate.mdx
+++ b/src/pages/en/migrate.mdx
@@ -2,7 +2,7 @@
 layout: ~/layouts/MainLayout.astro
 title: Migration Guide
 description: How to migrate your project to latest version of Astro.
-i18nReady: true
+i18nReady: false
 ---
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
--- a/src/pages/en/migrate.mdx
+++ b/src/pages/en/migrate.mdx
@@ -153,7 +153,7 @@ If you're not familiar with MDX, here are some steps you can follow to quickly c
     ```
 
 :::tip
-While you are transitioning to MDX, you may wish to enable the `legacy.astroFlavoredMarkdown` flag and include both **`.md` and `.mdx`** files, so that your site continues to function normally even before all your files have been converted. Here is one way you can do that:
+While you are transitioning to MDX, you may wish to enable the `legacy.astroFlavoredMarkdown` flag (removed in v2.0) and include both **`.md` and `.mdx`** files, so that your site continues to function normally even before all your files have been converted. Here is one way you can do that:
 
 ```astro
 ---
--- a/src/pages/en/migrate.mdx
+++ b/src/pages/en/migrate.mdx
@@ -95,7 +95,7 @@ When upgrading, please visually inspect your site output to make sure everything
 
 Astro no longer supports components or JSX expressions in Markdown pages by default. For long-term support you should migrate to the [`@astrojs/mdx`](/en/guides/integrations-guide/mdx/) integration.
 
-To make migrating easier, a new [legacy flag](/en/reference/configuration-reference/#legacyastroflavoredmarkdown) can be used to re-enable previous Markdown features.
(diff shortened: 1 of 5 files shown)


Yan-Thomas (approved):
LGTM!

src/pages/en/migrate.mdx
```diff
@@ -95,7 +95,7 @@ When upgrading, please visually inspect your site output to make sure everything
 
 Astro no longer supports components or JSX expressions in Markdown pages by default. For long-term support you should migrate to the [`@astrojs/mdx`](/en/guides/integrations-guide/mdx/) integration.
 
-To make migrating easier, a new [legacy flag](/en/reference/configuration-reference/#legacyastroflavoredmarkdown) can be used to re-enable previous Markdown features.
+To make migrating easier, a new `legacy.astroFlavoredMarkdown` flag can be used to re-enable previous Markdown features.
```
sarah11918, line 98:
```suggestion
To make migrating easier, a new `legacy.astroFlavoredMarkdown` flag can be used to re-enable previous Markdown features. (Removed in v2.0)
```
I was thinking of some kind of note here. Either as above, or 

> a new `legacy.astroFlavoredMarkdown` flag (removed in v2.0)

Thoughs?

src/pages/en/migrate.mdx
```diff
@@ -153,7 +153,7 @@ If you're not familiar with MDX, here are some steps you can follow to quickly c
     ```
 
 :::tip
-While you are transitioning to MDX, you may wish to [enable the legacy flag](/en/reference/configuration-reference/#legacyastroflavoredmarkdown) and include both **`.md` and `.mdx`** files, so that your site continues to function normally even before all your files have been converted. Here is one way you can do that:
+While you are transitioning to MDX, you may wish to enable the `legacy.astroFlavoredMarkdown` flag and include both **`.md` and `.mdx`** files, so that your site continues to function normally even before all your files have been converted. Here is one way you can do that:
```
sarah11918, line 156: I would add whichever note previous here too. Keep the historical accuracy, but note that this no longer exists.

src/pages/en/migrate.mdx
delucis: Even though we’re going to delete this all? I’m not opposed though — doesn’t do any harm.

sarah11918: Yes, still my preference! (FWIW)

delucis: Edited your suggestion to use your second suggestion 👍

sarah11918:
Looks great! The only other thing I am tempted to do is to delete all the other language's `migrate.mdx` files at this point? Not opposed to keeping them if people feel strongly, but I'd nix them now.

src/pages/en/migrate.mdx
delucis:
```suggestion
While you are transitioning to MDX, you may wish to enable the `legacy.astroFlavoredMarkdown` flag (removed in v2.0) and include both **`.md` and `.mdx`** files, so that your site continues to function normally even before all your files have been converted. Here is one way you can do that:
```

delucis:
I think my preference would be to remove the translated migration guides in the same PR that adds the new English migration guide. What do you think?

sarah11918 (approved):

delucis:
We can discuss then. If we’re now thinking of “migration == coming to Astro from elsewhere” and talking more about “upgrading Astro versions”, we might want to reflect that in the URL/filename too and add a redirect for links to the old location.

PR #2312, opened 2023-01-06
i18n(pt-BR): Update `astro-components.mdx`, `astro-pages.mdx` and a few more!

Yan-Thomas:

- Translated content

#### Description

Updated `astro-components.mdx` and `astro-pages.mdx` translations plus a few pages that needed small (1-10 LOC) changes.


delucis (commented):
Looks like we already have a conflict with this PR. I’d say let’s get this merged when you have time to fix that conflict @Yan-Thomas — with all the docs upheaval we’re expecting in the run up to v2 it’s probably best to merge earlier especially on big PRs like this. If someone spots mistakes later, we can fix things then!

PR #2326, opened 2023-01-09
Re-revert "Consistent installation title"

Tc-001:

#### Description

My second shot at this PR, for context see #2262 and #2265 

The titles now are:
- Installation (Automatic CLI)
- Installation (Manual Setup)

I am using the tab names, so it should still be consistent, they also have around the same breakpoint so the line amount should match on most screen sizes

cc @delucis, let me know if this works for you

### DIFF
--- a/src/pages/en/install/auto.mdx
+++ b/src/pages/en/install/auto.mdx
@@ -1,5 +1,5 @@
 ---
-title: Install Astro with the Automatic CLI
+title: Installation (Automatic CLI)
 description: 'How to install Astro with NPM, PNPM, or Yarn via the create-astro CLI tool.'
 layout: ~/layouts/MainLayout.astro
 i18nReady: true
--- a/src/pages/en/install/auto.mdx
+++ b/src/pages/en/install/auto.mdx
@@ -1,5 +1,5 @@
 ---
-title: Installation (Automatic CLI)
+title: Install Astro with the Automatic CLI
 description: 'How to install Astro with NPM, PNPM, or Yarn via the create-astro CLI tool.'
 layout: ~/layouts/MainLayout.astro
 i18nReady: true
--- a/src/pages/en/install/manual.mdx
+++ b/src/pages/en/install/manual.mdx
@@ -1,5 +1,5 @@
 ---
-title: Installation (Manual Setup)
+title: Install Astro with manual setup
 description: 'How to install Astro manually with NPM, PNPM, or Yarn.'
 layout: ~/layouts/MainLayout.astro
 i18nReady: true
--- a/src/pages/en/install/manual.mdx
+++ b/src/pages/en/install/manual.mdx
@@ -1,5 +1,5 @@
 ---
-title: Install Astro manually
+title: Installation (Manual Setup)
 description: 'How to install Astro manually with NPM, PNPM, or Yarn.'
 layout: ~/layouts/MainLayout.astro
 i18nReady: true


src/pages/en/install/auto.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Install Astro with the Automatic CLI
+title: Installation (Automatic CLI)
```
delucis, line 2:
```suggestion
title: Install Astro (Automatic CLI)
```

src/pages/en/install/manual.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Install Astro manually
+title: Installation (Manual Setup)
```
delucis, line 2:
```suggestion
title: Install Astro (Manual Setup)
```

delucis (commented):
Thanks for circling back to this @Tc-001!

What do you think of this suggestion preserving “Install Astro” instead of “Installation”? I kind of like having “Astro” in there for clarity in search engine results and this is only very slightly longer than your suggestion.

Tc-001:
I like it!
Will make the change in a few hrs

sarah11918:
@Tc-001 - just pinging for freshness! Make Chris' changes, and it's good to merge, I think!

Tc-001:
They are a bit too different for the effect, but the following works:

- Install Astro with manual setup
- Install Astro with the Automatic CLI

sarah11918:
What does it look like if you compare:

> Install Astro Wizard

> Install Astro manually

I actually *like* how Install Auto manually looks as a title. So, can we tighten up the Auto? Other suggestions for phrasing?

Tc-001:
|a|b|c|
|---|---|---|
|image| image | image|

"Install Astro automatically" just barely is smaller than the word wrap point, but unsure how it looks on other browsers
Wizard works but I think the wording is a bit weird? Will see if I can come up with anything

sarah11918:
Install Astro via Wizard  ??

Install Astro with Wizard ??

Tc-001:
Maybe? Still a bit strange for my taste, but they do fit!

Tc-001:
I have found some, but needing them to be around the same length is a bit limiting, I don't think there is a perfect one

Tc-001:
For me personally [this](https://github.com/withastro/docs/pull/2326#issuecomment-1377828690) makes the most sense, as it is a line and a bit in length in the largest size

But maybe it would be better to remove the tabs and make them distinctively 2 different pages? They could link to each other with something like:
> "The wizard will now run some commands, if you want to know what they are or run them yourself, click here. "

Tc-001:
I think I will sleep on it, is is getting quite late
Also will probably move to discord, a lot easier to paste screenshots there

Best of lucks to your projects!

kevinzunigacuellar:
Is there any reason to keep the automatic in "automatic cli"?
 I think "with the cli" is simpler

Tc-001:
It could work and fits in one line!
Although it wraps a lot sooner so the effect would break if the browser is not maximized, but it would do it's job ~80% of the time

sarah11918:
Coming back to this.... 😅 

What about some distinction more like:

"Automatic Setup" / "Install Manually"

- They *both* involve the CLI, just in one case you're installing and running the wizard, and in the other, you're just installing Astro. So, I don't think using "CLI" for one vs the other is exactly correct?
- It's not the "installing" that's automatic, it's the SETUP with all the dependencies, eh?

sarah11918:
OK! So we have a different fix for this issue.... we're going to tackle this in #2390 and kill off the fake tabs entirely!  :hocho: 

This is the only time we use tabs to "fake" page navigation, and it's a pattern that is clearly problematic and that we shouldn't enable. Plus, having two separate pages allows us to customize the top of the page so we can still nudge people to `create astro`.  

Since the whole reason for this pattern was to avoid duplicating content, but now we actually want that content to be different anyway, I'm closing this issue in favour of the solution to making two entirely different pages. There's no need for this content to be this closely aligned anymore.

PR #2329, opened 2023-01-09
Fix dynamic redirect route

delucis:

#### Description

While updating an `Astro.glob()` call as part of #2133, I messed up one of the patterns so that a dynamic route we use to redirect from routes like `/reference/configuration-reference` to `/en/reference/configuration-reference` wasn’t working properly. Noticed it today after spotting an increase in docs 404s since #2133 merged. This PR fixes the glob pattern.


Yan-Thomas (approved):
Nice catch, LGTM!

PR #2332, opened 2023-01-09
Update recommended Node version to 16

delucis:

- Something else!

#### Description

The `.nvmrc` file in the repo can be used by environments to pick which Node version to use while working on the docs site. Astro v2 will drop support for Node 14, so this PR updates this file to recommend Node 16 instead. We already run GitHub CI with Node 16 but this’ll be first time building with Node 16 on Netlify which respects `.nvmrc`.


Yan-Thomas (approved):
LGTM

PR #2339, opened 2023-01-10
update create astro and template flag instructions

sarah11918:
This fixes the instructions for running create astro with a template flag in the migration guides.

It also adds the example we have for starting a new project from any GitHub template from the hidden `integrations/integrations` page to Auto install, to better surface these instructions.

Lastly, it changes *most* instances of `create-astro` to `create astro`, since we are usually referring to the command to run, not the name of the package itself. (This has caused confusion in the past, so doing this while I'm here, but can be talked out of that change.) Admittedly, some of these occurances (both changed, and not changed) were judgment calls. Mostly based on what I thought might be least confusing to the reader. e.g. If the name of the wizard shows up JUST BEFORE they need to run the command, then I removed the hyphen so the reader is not seeing two conflicting things together.

### DIFF
--- a/src/pages/en/core-concepts/project-structure.mdx
+++ b/src/pages/en/core-concepts/project-structure.mdx
@@ -6,7 +6,7 @@ i18nReady: true
 ---
 import FileTree from '~/components/FileTree.astro'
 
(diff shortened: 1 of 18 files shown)


delucis (approved):
LGTM!

PR #2313, opened 2023-01-06
[Content collections] Add `.astro` directory setup

bholmesdev:
#### What kind of changes does this PR include?

- New or updated content

#### Description

- Add `.astro` directory explanation and setup
  - tip to add to `gitignore`
  - note on `reference` type setup

PR #2345, opened 2023-01-10
[Content collections] Clarify type generation step

bholmesdev:

- New or updated content

#### Description

- Add type generation to "Getting Started"
- Add tips where the `astro:content` module is used


delucis (commented):
I’m wondering if this might be overkill. Especially as there are some similar tips added in #2313. I’d suggest we wait for the other content collections PRs to merge and also for the PR removing the note about the experimental flag and then regroup here.

PR #2349, opened 2023-01-11
i18n(zh-cn): fix punctuation and bold typos

liruifengv:


delucis (approved):
Thanks for catching those @liruifengv! 🙌

PR #2354, opened 2023-01-11
Remove experimental from content collections!

bholmesdev:

- New or updated content

#### Description

- It's baseline time.


delucis (commented):
Exciting! This should be ready to merge when the next beta goes out 🥳 

One note: I’d suggest changing the `<Since />` component on line 16 to `v="2.0.0"` now that we’re removing the note about experimental config as it won’t really be available in any obvious way in 1.7.0.

~~Maybe worth double checking we’re not saying 1.7.0 anywhere else for this?~~ Ah, yup a few other places: API & CLI Reference pages.

Otherwise LGTM!

PR #2365, opened 2023-01-12
remove scary warning

sarah11918:
Warning is scary, not helpful!

- `astro add` is more reliable/familiar to users than at the time of this warning
- Houston walks a user through the changes to their file
- the integration READMEs show manual installation and what a properly installed and configured integration should look like, in case a user needs to compare.

Removing the scary warning as a quick fix not to scare off new users. Will revisit this page with more intention in the future.


delucis (approved):
Looks less scary to me!

Yan-Thomas (approved):
LGTM!

PR #2371, opened 2023-01-14
Fred's edit pass on content collections guide

FredKSchott:
- updated content

#### Description

As a part of my v2.0 readiness docs pass, I took a look at the content collections guide and had some suggestions for how to improve the guide. My edits focused on three main areas:

1. **Improve readability and ordering:** Since this is meant to be a guide, I took a serious look at how the information is presented and in what order. Instead of jumping the user immediately into TypeScript configuration, the guide now introduces the concepts more slowly, building as it goes. The goal is to define terms early, introduce new concepts one at a time, and not overwhelm.
2. **Collect concepts into a single place:** I took a look at places where concepts were explained across multiple sections, and did my best to bring them into a single section. This is feedback we've gotten before so I'm extra sensitive to it!
3. **Remove reference material:** Remove anything that is really reference. This was mainly directed at detailed type definitions, which belong in reference and not guides. Adding to reference is left as TODO.

#### TODO

1. Add back Content Entry type definition somewhere within reference.
2. Resolve `getEntry()` SSR issue

### DIFF
--- a/src/pages/en/guides/content-collections.mdx
+++ b/src/pages/en/guides/content-collections.mdx
@@ -74,7 +74,7 @@ If you find yourself working with different types of content, you should create 
 
 A content collection is always a top-level folder inside of the `src/content/` directory. You cannot nest one collection inside of another. However, you can use subdirectories to organize your content within a collection.
 
-For example, you can use the following directory structure to organize i18n translations within a single `docs` collection. When you go to query this collection, you'll be able to filter the result by language using the file path. 
+For example, you can use the following directory structure to organize i18n translations within a single `docs` collection. When you query this collection, you'll be able to filter the result by language using the file path. 
 
 <FileTree>
 - src/content/
--- a/src/pages/en/guides/content-collections.mdx
+++ b/src/pages/en/guides/content-collections.mdx
@@ -23,7 +23,7 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
 
 A **content collection** is any directory inside the reserved `src/content` project directory, such as `src/content/newsletter` and `src/content/blog`. Only content collections are allowed inside the `src/content` directory. This directory cannot be used for anything else.
 
-A **content entry** is any piece of content stored inside of your content collection. Content entries are stored as either Markdown (`.md`) or MDX (`.mdx`) files, saved inside of content collection directory. Your content entries can be named anything you want, but we recommend using a consistent naming scheme to make it easier to find and organize your content. 
+A **content entry** is any piece of content stored inside of your content collection. Content entries are stored as either Markdown (`.md`) or MDX (`.mdx`) files, saved inside of the content collection directory. Your content entries can be named anything you want, but we recommend using a consistent naming scheme to make it easier to find and organize your content. 
 
 <FileTree>
 - src/content/
--- a/src/pages/en/guides/content-collections.mdx
+++ b/src/pages/en/guides/content-collections.mdx
@@ -16,31 +16,46 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
   <Since v="2.0.0" />
 </p>
 
-**Content collections** are a feature to help manage your Markdown and MDX files in Astro. A collection will organize your content, validate your frontmatter, and provide automatic TypeScript type-safety of fetched content. 
+**Content collections** are the best way to work with Markdown and MDX in any Astro project. Content collections are a feature of Astro that help manage your content files in a project. Collections help to organize your content, validate your frontmatter, and provide automatic TypeScript type-safety for all of your content.
 
-Content collections are the best way to work with Markdown and MDX in your project. For smaller projects, you can also [save your Markdown directly inside of the `src/pages/` directory.](/en/core-concepts/astro-pages/#markdownmdx-pages)
 
 ## What are Content Collections?
 
 A **content collection** is any directory inside the reserved `src/content` project directory, such as `src/content/newsletter` and `src/content/blog`. Only content collections are allowed inside the `src/content` directory. This directory cannot be used for anything else.
 
-Content collections help you organize the content files in your project. A content collection can contain any number of related Markdown (`.md`) or MDX (`.mdx`) files. These files within a collection are called **content entries.**
+A **content entry** is any piece of content stored inside of your content collection. Content entries are stored as either Markdown (`.md`) or MDX (`.mdx`) files, saved inside of content collection directory. Your content entries can be named anything you want, but we recommend using a consistent naming scheme to make it easier to find and organize your content. 
 
 <FileTree>
 - src/content/
-  - **newsletter/** 
-    - week-1.md
-    - week-2.md
-    - week-3.md
+  - **newsletter/** the "newsletter" collection
+    - week-1.md a collection entry
+    - week-2.md a collection entry
+    - week-3.md a collection entry
 </FileTree>
 
-Once you have a content collection in place with at least one content entry, you can query your content using Astro's built-in APIs.
+Once you have a collection, you can start [querying your content](#querying-collections) using Astro's built-in content APIs.
 
-### Creating a new content collection
+### The ".astro" Directory
 
-To create a new content collection, add a new top-level directory to the `src/content/` folder of your project. If a `src/content` directory doesn't already exist in your project, create it.
+Astro stores important metadata for content collections in an `.astro` directory in your project. No action is needed on your part to maintain or update this directory. You are encouraged to ignore it entirely while working in your project.
 
-Fill your collection directory with related Markdown and/or MDX files. Content collections work best when all files in the collection follow a similar frontmatter structure. If two files represent two different kinds of content (a blog post and an author profile, for example) they most likely belong in two different collections. 
+The `.astro` directory will be updated for you automatically anytime you run the [`astro dev`](/en/reference/cli-reference/#astro-dev), [`astro build`](/en/reference/cli-reference/#astro-build) commands. You can run [`astro sync`](/en/reference/cli-reference/#astro-sync) at any time to update the `.astro` directory manually.
+
+:::tip
+If you're using Git for version control, we recommend ignoring the `.astro` directory by adding `.astro` to your `.gitignore`. This tells Git to ignore this directory and any files inside of it.
+
+```bash
+echo ".astro" >> .gitignore
+```
+:::
+
+
+
+### Organizing with multiple collections 
+
+If two files represent two different kinds of content (e.g. a blog post and an author profile), they most likely belong in two different collections. This is an important concept to remember because important features like frontmatter validation and automatic TypeScript typings require that all entries share a similar frontmatter structure.
+
+If you find yourself working with different types of content, you should create multiple collection to represent each type. You can create as many different collections in your project as you'd like.
 
 <FileTree>
 - src/content/
@@ -48,47 +63,79 @@ Fill your collection directory with related Markdown and/or MDX files. Content c
     - week-1.md
     - week-2.md
     - week-3.md
-  - **authors/** diffrent content types belong in different collections
+  - **authors/** split different content types into new collections
     - grace-hopper.md
     - alan-turing.md
     - batman.md
 </FileTree>
 
-If your content frontmatter varies too wildly between files in a single collection, features like frontmatter validation and automatic TypeScript typings become less powerful. You can optionally define and enforce a required [frontmatter schema](#creating-a-collection-schema) across the entire collection.
 
-### Organizing your collections with subdirectories
+### Organizing with subdirectories
 
-A content collection is *always* a top-level folder inside of the `src/content/` directory. You cannot nest one collection inside of another. However, you can use a nested subdirectory to organize your content within a collection. 
+A content collection is always a top-level folder inside of the `src/content/` directory. You cannot nest one collection inside of another. However, you can use subdirectories to organize your content within a collection.
 
-For example, you can use the following subdirectory structure to add international translations within a collection:
+For example, you can use the following directory structure to organize i18n translations within a single `docs` collection. When you go to query this collection, you'll be able to filter the result by language using the file path. 
 
 <FileTree>
 - src/content/
-  - docs/
-    - en/
-    - es/
-    - ...
+  - docs/ this collection uses subdirectories to organize by language
+    - **en/**
+    - **es/**
+    - **de/**
 </FileTree>
 
-## Configuration
+## Defining Collections
 
-Astro allows you to customize and configure your project's content collections through an optional `src/content/config.ts` file (`.js` and `.mjs` extensions are also supported). 
+:::caution
+The `src/content/config.ts` file is optional. However, choosing not to define your collections will disable some of their best features like frontmatter schema validation or automatic TypeScript typings. 
+:::
 
-Currently, this file supports [creating a collection schema](#creating-a-collection-schema) and [defining custom unique slugs](#defining-custom-slugs) for your collections.
+To get the most out of your content collections, create a `src/content/config.ts` file in your project (`.js` and `.mjs` extensions are also supported). This is a special file that Astro will automatically load and use to configure your content collections.
 
-### Creating a collection schema
-
-Schemas are an optional way to enforce consistent frontmatter within a collection. Without a schema, you have no way to guarantee that a frontmatter property actually exists when you go to reference it inside a page or component template.
-
-To create a content collection schema, first create a `src/content/config.ts` file if one does not already exist (`.js` and `.mjs` extensions are also supported). This file should:
-
-1. Import the proper utilities from `astro:content`. 
-2. Define a schema for each collection you'd like to validate.
-3. Export a single `collections` object to apply your schemas.
 
 ```ts
 // src/content/config.ts
-// 1. Import the proper utilities from `astro:content`
+// 1. Import utilities from `astro:content`
+import { defineCollection } from 'astro:content';
+// 2. Define your collection(s)
+const blogCollection = defineCollection({ /* ... */ });
+// 3. Export a single `collections` object to register your collection(s)
+//    This key should match your collection directory name in "src/content"
+export const collections = {
+  'blog': blogCollection,
+};
+```
+
+
+### Setting up TypeScript
+
+To get the most out of content collections in Astro, you may need to update your `tsconfig.json` to enable `strictNullChecks`. This change is only required if you **do not** already extend Astro's `strict` or `strictest` recommended TypeScript settings in your `tsconfig.json` file.
+
+```json title="tsconfig.json" ins={5}
+{
+  // Note: No change needed if you use "astro/tsconfigs/strict" or "astro/tsconfigs/strictest"
+  "extends": "astro/tsconfigs/base",
+  "compilerOptions": {
+    "strictNullChecks": true
+  }
+}
+```
+
+### Defining a collection schema
+
+Schemas enforce consistent frontmatter within a collection. A schema **guarantees** that your frontmatter exists in a predictable form when you need to reference or query it. If any file violates its collection schema, Astro will provide a helpful error to let you know.
+
+Schemas also power Astro's automatic TypeScript typings for your content. When you define a schema for your collection, Astro will automatically generate and apply a TypeScript interface to it. The result is full TypeScript support when you query your collection, including proprety auto-completion and type-checking. 
+
+To create your first content schema, create a `src/content/config.ts` file if one does not already exist (`.js` and `.mjs` extensions are also supported). This file should:
+
+1. Import the proper utilities from `astro:content`. 
+2. Define each collection that you'd like to validate with a schema.
+3. Export a single `collections` object to register your collections.
+
+```ts
+// src/content/config.ts
+// 1. Import utilities from `astro:content`
 import { z, defineCollection } from 'astro:content';
 // 2. Define a schema for each collection you'd like to validate.
 const blogCollection = defineCollection({
@@ -106,25 +153,22 @@ export const collections = {
 
 ### Defining multiple collections
 
-Astro supports working with multiple collections in a single project. 
+You can define as many collections in your project as you would like. As long as they are exported from the `src/content/config.ts` file from inside the `collections` object, they will be loaded by Astro.
 
 ```ts
 // src/content/config.ts
-// ...
-
-// Export multiple collections to register them
 export const collections = {
-  'blog': blogCollection,
-  'newsletter': newsletterCollection,
-  'profile': profileCollection,
+  'blog': defineCollection({ /* ... */ }),
+  'newsletter': defineCollection({ /* ... */ }),
+  'profile-authors': defineCollection({ /* ... */ }),
 };
 ```
 
-As your project grows, you are free to reorganize your codebase and move your schemas and collections out of the `src/content/config.ts` file. Your collections and schemas can live anywhere in your project, as long as they are imported into your `src/content/config.ts` file and and then re-exported from the `collections` object.
+As your project grows, you are also free to reorganize your codebase and move logic out of the `src/content/config.ts` file. Defining your schemas seperately can be useful for reusing schemas across multiple collections and sharing schemas with other parts of your project.
 
 ```ts
 // src/content/config.ts
-// 1. Import your schemas and utilities
+// 1. Import your utilities and schemas
 import { defineCollection } from 'astro:content';
 import {blogSchema, newsletterSchema} from '../schemas';
 // 2. Define your collections
@@ -139,7 +183,7 @@ export const collections = {
 
 ### Using third-party collection schemas
 
-You are free to import schemas and collections from anywhere, including external npm packages. This can be useful when working with themes and libraries that provide their own collection schemas for you to use.
+You can import collection schemas from anywhere, including external npm packages. This can be useful when working with themes and libraries that provide their own collection schemas for you to use.
 
 
 ```ts
@@ -155,9 +199,9 @@ export const collections = {
 
 ### Defining datatypes with Zod
 
-Astro leverages [Zod](https://github.com/colinhacks/zod) to create frontmatter schemas in the form of [Zod objects](https://github.com/colinhacks/zod#objects). With Zod, Astro is able to validate every file's frontmatter within a collection *and* provide automatic TypeScript types when you go to query content from inside your project.
+Astro uses [Zod](https://github.com/colinhacks/zod) to power its content schemas. With Zod, Astro is able to validate every file's frontmatter within a collection *and* provide automatic TypeScript types when you go to query content from inside your project.
 
-To use Zod in Astro, import the `z` utility from `"astro:content"`. This is a re-export of the Zod library, and you are able to use all Zod features inside of Astro. See [Zod’s README](https://github.com/colinhacks/zod) for complete documentation on how Zod works and what features are available.
+To use Zod in Astro, import the `z` utility from `"astro:content"`. This is a re-export of the Zod library, and it supports all of the features of Zod. See [Zod’s README](https://github.com/colinhacks/zod) for complete documentation on how Zod works and what features are available.
 
 
 ```ts
@@ -205,46 +249,10 @@ const blog = defineCollection({
 ```
 
 
-### Setting up TypeScript for content collections
-
-To benefit from the full TypeScript and autocompletion features of [using schemas with your collections](#creating-a-collection-schema), you may need to update your `tsconfig.json` to enable `strictNullChecks`.
-
-```json title="tsconfig.json" ins={3}
-{
-  "compilerOptions": {
-    "strictNullChecks": true
-  }
-}
-```
-
-Note that if you are currently extending Astro's `strict` or `strictest` recommended settings, no changes is needed.
-
-```json title="tsconfig.json" "strict" "strictest"
-{
-  // No change needed!
-  "extends": "astro/tsconfigs/strict"
-}
-```
-
-### The ".astro" Directory
-
-Astro generates type information and metadata from your content collections in a new `.astro` directory. No action is needed on your part to maintain or update this directory. It is safe to ignore it in your project.
-
-:::tip
-If you're using Git for version control, we recommend ignoring this generated directory by adding `.astro` to your `.gitignore`.
-:::
-
-The `.astro` directory will be updated for you automatically anytime you run the [`astro dev`](/en/reference/cli-reference/#astro-dev), [`astro build`](/en/reference/cli-reference/#astro-build) commands. You can run [`astro sync`](/en/reference/cli-reference/#astro-sync) yourself to update `.astro` manually.
-
-
-## The "src/content" Directory
-
-Astro treats the `src/content/` directory as special. This is where **collections** (folders) of Markdown/MDX **entries** (files) can be stored, with a single configuration file to define each collection's **schema** (frontmatter data types and shape). Files other than your `.md`/`.mdx` content are not permitted inside `src/content/`.
-
 
 ## Querying Collections
 
-Astro provides two functions to query a collection and return one (or more) content entries from it: [`getCollection()`](/en/reference/api-reference/#getcollection) and [`getEntryBySlug()`](/en/reference/api-reference/#getentrybyslug).
+Astro provides two functions to query a collection and return one (or more) content entries: [`getCollection()`](/en/reference/api-reference/#getcollection) and [`getEntryBySlug()`](/en/reference/api-reference/#getentrybyslug).
 
 ```js
 import { getCollection, getEntryBySlug } from 'astro:content';
@@ -254,9 +262,11 @@ const allBlogPosts = await getCollection('blog');
 const oneBlogPost = await getEntryBySlug('blog', 'enterprise');
 ```
 
+Both functions return content entries as defined by the [`CollectionEntry`](/en/reference/api-reference/#collection-entry-type) type.
+
 #### Filtering collection queries
 
-`getCollection` takes an optional "filter" callback that allows you to filter your query based on an entry's `id`, `slug`, or `data` (frontmatter) properties.
+`getCollection()` takes an optional "filter" callback that allows you to filter your query based on an entry's `id`, `slug`, or `data` (frontmatter) properties.
 
 You can use this to filter by any content criteria you like. For example, you can filter by frontmatter properties like `draft` to prevent any draft blog posts from publishing to your blog:
 
@@ -280,7 +290,7 @@ const englishDocsEntries = await getCollection('docs', ({ id }) => {
 
 ### Using content in Astro templates
 
-Once you have queried your collection entries, you can use those entries to render information about your content. You can access each entry directly inside of your Astro component template. This lets you to render HTML for things like links to your content (using the content `slug`) or information about your content (using the `data` property).
+Once you have queried your collection entries, access each entry directly inside of your Astro component template. This lets you to render HTML for things like links to your content (using the content `slug`) or information about your content (using the `data` property).
 
 For information about rendering your content to HTML, see [Rendering Content to HTML](/en/guides/content-collections/#rendering-content-to-html) below.
 
@@ -306,7 +316,7 @@ const blogEntries = await getCollection('blog');
 
 A component can also pass an entire content entry as a prop. 
 
-If you do this, you can use the `CollectionEntry` utility to correctly type your components props using TypeScript.  This utility takes a string argument that matches the name of your collection schema, and will inherit all of the properties of that collection's schema.
+If you do this, you can use the [`CollectionEntry`](/en/reference/api-reference/#collection-entry-type) utility to correctly type your components props using TypeScript.  This utility takes a string argument that matches the name of your collection schema, and will inherit all of the properties of that collection's schema.
 
 ```astro /CollectionEntry(?:<.+>)?/
 ---
@@ -321,7 +331,7 @@ const { post } = Astro.props;
 ---
 ```
 
-## Rendering Content to HTML
+### Rendering content to HTML
 
 Once queried, you can render a collection entry to HTML using the entry `render()` function property. Calling this function gives you access to rendered content and metadata, including both a `<Content />` component and a list of all rendered headings.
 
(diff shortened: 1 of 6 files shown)


src/pages/en/guides/content-collections.mdx
```diff
@@ -320,157 +232,73 @@ const blogPosts = await getCollection('blog', ({ data }) => {
...
-	}
-	---
-	```
-
 This will generate page routes for every entry in the `blog` collection, mapping each entry’s slug to a URL. For example, an entry at `src/content/blog/hello-world.md` will have a slug of `hello-world` and the collection entry `src/content/blog/en/intro.md` will have a slug of `en/intro`.
 
 Because this dynamic route is in `src/pages/posts/`, the final URLs will be `/posts/hello-world/` and `/posts/en/intro/`.
 
-### Rendering post contents
+#### Generating pages in SSR mode
```
FredKSchott, line 296: @bholmesdev I hate to be the bearer of bad news, but I think I found an issue with Content Collections when we switch to SSR. How does fetching an entry work in an SSR site, where you call `getEntry()` based on the `Astro.request.pathname` value?

`slug` is the wrinkle: if the pathname is my `slug`, but `getEntry()` is by `id`, you have no way to call `getEntry()`. `getEntry('blog', slugFromPathname)` wouldn't match the correct entry, as I understand it.

Lets chat on Tuesday, I have thoughts!

bholmesdev: No worries, very aware of this! We recommend using `getCollection()` with a filter today:

```ts
---
// taken from SSR tests in core

const matches = await getCollection('blog', post => post.slug === slug);
const post = matches[0];
if (!post) {
	throw new Error(`Not found: ${slug}`);
}
const { Content } = await post.render();
---
```

I agree `getEntry()` could be changed to address this. Discussing with a user in our `#general` chat, turning `getEntry()` into a "find first match" function would be helpful here:

```diff
---
import { getEntry } from 'astro:content';

const { slug } = Astro.params;
- const matches = await getCollection('blog', post => post.slug === slug);
- const post = matches[0];
+ const post = await getEntry('blog', post => post.slug === slug);
...
```

So you can pass the ID string directly if known for an O(1) lookup time, or pass a function to lookup by other properties. This also unlocks lookup by `data` properties without a `matches` array.

We can also discuss ensuring all `slug`s are unique to allow direct lookup by slug:
```ts
const post = await getEntry('blog', { slug: Astro.params.slug });
```
...though this has more breaking implications, and wouldn't be as flexible. Since we bundle a map of the entire collection into your SSR build, regardless of whether you use `getEntry()` or `getCollection()`, I lean towards the first option. Happy to discuss further next week!

FredKSchott: Yes, let's definitely discuss early next week. But, I'll throw this idea out now so that its not a surprise on Tuesday, and hopefully having the weekend to sit with it helps:

- `slug` doubling as both nice-to-have meta-property _and_ an essential ID lookup is beginning to smell (for lack of a better term), especially since `id` is already meant to be that essential ID lookup property.
- `slug` may be redundant, given that we already have `id`
- `slug` can be removed, in favor of `id`

This was less of an issue up to this point when you consider `slug` as a nice-to-have built-in helper to help users create their URLs easier. Nothing wrong with that! But, given that "get by request path" will actually be an essential use-case for any `[slug].astro` page (SSG too, not just SSR), the confusion and redundancy between `id` and `slug` becomes more of a critical issue to resolve.

I see what you mean where this can be solved in userland by just using `getCollection()` and filtering. But I'd argue that getting the entire collection just to filter & grab a single entry is not a very friend API, and not what users would expect from such a basic use-case (we have feedback from users today to back this up, who are upset with having to do this with `import.meta.glob` and `Astro.glob`).

I also see what you mean where this could be added as an _additional_ feature to `getEntry()`, but that's where I'll ask us to take a step back and reconsider _why_ we need two different lookup methods, when really one should suffice?


### Alternative: Remove `slug` in favor of `id`

https://github.com/withastro/astro/compare/content-slug-to-id?expand=1

I spun up a quick prototype, merging `slug` and `id` together to create a more slug-like, unique `id`. Fun fact: this is actually something that I pushed for originally, back when `id` was the full file path. I think this presents an opportunity to solve a problem by revisiting this idea.


### Example

```diff
- <a href={'/blog/' + post.slug}>{post.data.title}</a>
+ <a href={'/blog/' + post.id}>{post.data.title}</a>
```

```diff
// src/pages/blog/index.astro
export async function getStaticPaths() {
  const posts = await getCollection('blog');
  return posts.map((post) => ({
-   params: { slug: post.slug },
+   params: { slug: post.id },
  }));
}
```

User code would be updated to use the entry ID instead of its slug. `id` becomes more slug-like to accommodate this. The main change to `id` is that it no longer has the file extension included. It also no longer uses github-slugger internally. Instead, it's expected that the user should use slug-friendly filenames in content collections.

```
src/content/blog/first-post.md         id: "first-post"
src/content/blog/foo/second-post.md    id: "foo/second-post"
src/content/blog/third post.md         warning/error
```

### Still to decide: can we support custom slugs/permalinks?

With slug gone, so is the current `slug()` customizer hook in `defineCollection`. I do love this idea of id/slug/permalink customization, and I'd still love to bring it back.

However, I'll reiterate here that this feature was never granted as an essential/required goal of the design doc, and has always been considered okay to skip for an initial release to be added later as a fast-follow after. IIRC, it was mainly seen as okay to add to initial release scope because it was seen as easy. I wouldn't want us to sacrifice on the critical APIs in an attempt to save a non-critical API that can be added later. 

One potentially easy answer would be to just take all the work already done on `slug()`, and rename it to `id()`. Dynamically changing IDs feels a bit awkward, but since that's already what the current proposal is doing with slugs I don't its a step backwards from where we are today and an acceptable compromise. Given that this is an advanced use-case, we would probably be safe to deprecate this later if we came up with something better.

I'd also like to explore an alternative proposal to support a static `id:` frontmatter property in any entry file, that manually sets the `id` (defaulting to the file name, if no `id:` frontmatter is given). I think the fact that this is what users already do and expect to work with the `permalink` property gives us a good datapoint that this is a pattern that users would enjoy. It's less powerful than the dynamic `slug()`/`id()` callback idea, but I think I could argue that the requirement of a `id:` frontmatter to be static and visible at the top of any file is easier to reason about.

### To be revisited Tuesday!

I'll be offline for the rest of the weekend dealing with house repair stuff, so I won't be able to revisit this or say much more until Tuesday. Given all the options that we have available, I'm confident that we can come up with a solution here next week without sacrificing our 2.0 launch timeline.

bholmesdev: I... disagree that custom `slug`s can be removed, since the RFC process and discussing with users both exposed custom `slug`s as important. I also don't see why reconciling `id` and `slug` _requires_ that custom `slug` mapping needs to be removed, since it fairly straightforward to ensure all generated slugs are unique at build time. I do agree that reconciling `id` and `slug` makes sense though. We can discuss then, yes!

matthewp: I ran into the SSR issue myself. My first thought was that `getEntry()` should return `| undefined` and that it would be your responsibility as a user to to convert your URL slug to a valid id. If you pass it a known static id it could omit the possible undefined. Not saying this is the solution, but it was my instinct.

----

I think we should be careful not to leak page concepts into content. Content should just be content. You can generate pages with the content in any number of ways.

That's why I don't like the "slug-like id" idea, @FredKSchott. This means you can't have `post.md` and `post.mdx` in the same collection. To me that violates the idea that it's content and we've just created another pages folder.

matthewp: Here's an idea that I think no one will like (probably even me!) but let me through it out anyways:

- Get rid of `getEntry()` entirely. How it's shown being used in the docs is probably not how anyone is actually going to use it. There might be times where you hard-code a known piece of content to get, but the confusion the API causes the minimal value created through the API.
- Get rid of the concept of `slug`. The user can add slug to their frontmatter themselves and use that as a key in this scenario, if they want.
  - This prevents the "duplicate id" problem.
  - This prevents leaking page concepts into content.
- Rename `id` to `filename` to more clearly identify what it is and to signal that the user is going to have to figure out a way to translate that to a slug (if they need to do so) or to use a frontmatter property of their own choosing.

matthewp: I think essential problem here is that we don't have the concept of a "page". A collection is generic. Which is good but makes this use-case harder.

I'd like to make a comparison to Wordpress, which does have the concept of a page. When you create a page it:

- Creates a permalink from the title
- Allows you to override the (permalink)

5082c15022ac6e42b6b523e2f8acf0b1db9ecd0a

So maybe this is the missing thing we need to add. Rather than making all collections of pages, make it so that you can signal which collections actually *are* used as pages.



One idea going off of this, in `defineCollection` add an optional `type` property whose value can be `'page'`. For collections of type pages you are *required* to provide a `slug`. You can then use a new querying API:

```astro
---
import { getPageBySlug } from 'astro:content';

const entry = getPageBySlug(Astro.params.slug);

if(!entry) return new Response(null, { status: 404 })
---
<!-- ... -->
```

delucis: Hmm. What user issue does this extra complexity solve?

I want to make sure I understand the issue. The issue is that `getEntry` looks up by `id` but for “page” content, you more likely want to look up by `slug` as that’s what will be in `Astro.params` in SSR.

Main option today:

```js
const [entry] = await getCollection('blog', (post) => post.slug === slug);
if (!entry) return new Response(null, { status: 404 });
```

A userland (or official) `getEntryBySlug(collection, slug)` abstraction around that is already possible today so is this really as big an issue as we think?

sarah11918:
Just a reminder that we need to incorporate https://github.com/withastro/astro/pull/5893 here, because `getEntry()` has now been renamed to `getEntryBySlug()`

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,106 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
     - week-2.md
     - week-3.md
+  - **authors/** diffrent content types belong in different collections
+    - grace-hopper.md
+    - alan-turing.md
+    - batman.md
 </FileTree>
 
-### Collections with nested directories
+If your content frontmatter varries too wildly between files in a single collection, features like frontmatter validation and automatic TypeScript typings become less powerful. You can optionally define and enforce a required [frontmatter schema](/en/guides/content-collections/#defining-a-collection-schema) across the entire collection.
```
sarah11918, line 57:
```suggestion
If your content frontmatter varries too wildly between files in a single collection, features like frontmatter validation and automatic TypeScript typings become less powerful. You can optionally define and enforce a required [frontmatter schema](#creating-a-collection-schema) across the entire collection.
```

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,131 +127,145 @@ defineCollection({
...
+});
 ```
 
-📚 See [Zod’s documentation](https://github.com/colinhacks/zod) for a complete list of features.
 
-### Custom entry slugs
+### Updating TypeScript configuration
 
-By default, Astro will generate a `slug` for each content entry based on its `id` parsed by [github-slugger](https://github.com/Flet/github-slugger).
+To benefit from the full TypeScript and autocompletion features of [using schemas with your collections](#defining-a-collection-schema), you may need to update your `tsconfig.json` to enable `strictNullChecks`.
```
sarah11918, line 164:
```suggestion
To benefit from the full TypeScript and autocompletion features of [using schemas with your collections](#creating-a-collection-schema), you may need to update your `tsconfig.json` to enable `strictNullChecks`.
```

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,106 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
-## Configuring collections
-
-You can configure your content collections with an optional `src/content/config.ts` file (`.js` and `.mjs` extensions are also supported). 
+## Configuration
 
-This file currently allows you to [define a collection schema](#defining-a-collection-schema), and to [create custom slugs](#custom-entry-slugs) for your collections.
+Astro allows you to customize and configure your project's content collections through an optional `src/content/config.ts` file (`.js` and `.mjs` extensions are also supported). 
 
-### Defining a collection schema
+Currently, this file supports [defining a collection schema](#defining-a-collection-schema) and [creating custom unique slugs](#custom-entry-slugs) for your collections.
```
sarah11918, line 77:
```suggestion
Currently, this file supports [creating a collection schema](#creating-a-collection-schema) and [defining custom unique slugs](#defining-custom-slugs) for your collections.
```

sarah11918:
Note: I added a necessary corresponding change to a link reference from `api-reference` here.

I know this PR might not be in its final form, but at least now the other page that will need changes is here, and visible for editing. (We are also going to have to update this page by renaming `getEntry()` for example.)

sarah11918:
Current status:

- [ ]  all links fixed (in both `content-collections` and `api-reference`) to correspond to edits to headings
- [ ] `getEntry` renamed to `getEntryBySlug` including any corresponding definitions (`id` -> `slug`) and code samples (`post-1.md` -> `post-1`)
- [ ] obvious typos etc. fixed
- [ ] Conflicts resolved
- [ ] Editing for content, style etc.

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,131 +127,145 @@ defineCollection({
...
-### Data returned from a collection query
+### Using content in templates
 
-`getCollection()` and `getEntry()` will return entries that include:
- - `id` - a unique ID using the file path relative to `src/content/[collection]`.
- - `slug` - a URL-ready slug. Defaults to the `id` parsed by [github-slugger](https://github.com/Flet/github-slugger).
- - `data` - an object of frontmatter properties inferred from your collection schema. Defaults to `any` if no schema is configured.
- - `body` - a string containing the raw, uncompiled body of the Markdown or MDX document.
- - `render()` - a function that returns the compiled body of the Markdown or MDX document via the `<Content />` component ([See complete documentation](#rendering-entry-content)).
+Once you have queried your content, you can use the returned entries inside of your Astro pages and components. For example, you can include frontmatter in a component template, or pass an entire content entry object as a prop.
```
sarah11918, line 251:
```suggestion
Once you have queried your content, you can use the returned entries inside of your Astro pages and components. For example, you can include frontmatter in a component template from a `data` object, or pass an entire content entry object as a prop.

See the [entire list of properties returned](/en/reference/api-reference/#getcollection) from queries.
```

Since we've lost the list of returned data properties on this page, I'd argue for being just a tiny bit more explicit somehow about what is returned and available for rendering here if we want to offload the bulk of the info to another page. I think it is appropriate here to at least say that you get a `data` object of your frontmatter properties, and `slug`, for example. 

And, somewhere (I'm just making all these notes here, but I really mean incorporated around this section), I would explicitly link back to the properties on `api-reference` (assuming that's where they'll go).

sarah11918 (commented):
@FredKSchott I have done some tidying, and left some higher level comments here. Mostly, this is great! There are a couple of things I feel are worth mentioning, to make sure we're intentionally choosing against some things that were here before. Otherwise, this is a great revamp in general, and not too many nits!  :partying_face:

src/pages/en/guides/content-collections.mdx
FredKSchott: Resolved in Discord, no change needed to `id`/`slug` but `getEntry` will be updated to `getEntryBySlug` to account for this much more common usecase.

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,131 +127,145 @@ defineCollection({
...
 import { getCollection } from 'astro:content';
 const enDocs = await getCollection('docs', ({ id }) => {
-  // Return all entries in `src/content/docs/en/`
   return id.startsWith('en/');
 });
----
 ```
 
-### `getEntry()`
+### `getEntryBySlug()`
```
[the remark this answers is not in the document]
FredKSchott: yes! I removed but didn't add back to reference, will do that in this PR.

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,131 +127,145 @@ defineCollection({
...
- - `data` - an object of frontmatter properties inferred from your collection schema. Defaults to `any` if no schema is configured.
- - `body` - a string containing the raw, uncompiled body of the Markdown or MDX document.
- - `render()` - a function that returns the compiled body of the Markdown or MDX document via the `<Content />` component ([See complete documentation](#rendering-entry-content)).
+Once you have queried your content, you can use the returned entries inside of your Astro pages and components. For example, you can include frontmatter in a component template, or pass an entire content entry object as a prop.
 
-Querying your content files with `getCollection()`or `getEntry()` allows you to use frontmatter properties from an entry's `data` object in [JSX-like expressions](/en/core-concepts/astro-components/#jsx-like-expressions) or pass props to other components, such as a layout. You can optionally add type safety with a built-in utility.
+In the following example, we bring together everything we've learned to display a filtered list of links to your blog:
```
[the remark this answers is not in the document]
FredKSchott: thank you! I love to see some hard voice/grammar rules that I can lean on now as a contributor!

src/pages/en/guides/content-collections.mdx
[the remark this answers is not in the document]
FredKSchott: actually, it already exists in `api-reference`! Updating this section to point to that, and then focus on more guide-like, use-case driven discussion.

src/pages/en/guides/content-collections.mdx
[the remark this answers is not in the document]
FredKSchott: resolved!

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,106 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
+To create a content collection schema, first create a `src/content/config.ts` file if one does not already exist (`.js` and `.mjs` extensions are also supported). This file should:
 
-For example, say you maintain two collections: one for release announcements and one for blog content. Your entries at `src/content/releases/` should include a `title` and `version`.  Your `src/content/engineering-blog/` collection entries should have a `title`, list of `tags`, and an optional `image` URL. 
-
-You can specify the expected shape of your frontmatter with the `schema` field of `defineCollection`:
+1. Import the proper utilities from `astro:content`. 
+2. Define a schema for each collection you'd like to validate.
+3. Export a single `collections` object to apply your schemas.
 
 ```ts
```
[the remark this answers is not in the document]
FredKSchott: done! tried a new approach to this that also lets us highlight using external schemas. Let me know what you think!

src/pages/en/guides/content-collections.mdx
FredKSchott: done!

src/pages/en/guides/content-collections.mdx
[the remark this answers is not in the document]
FredKSchott: done!

FredKSchott:
@sarah11918 second pass complete!

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +173,250 @@ defineCollection({
...
-It can also take a second, optional parameter: a filter function based on schema properties. This allows you to query for only some items in a collection based on `id`, `slug`, or frontmatter values via the `data` object.
+Astro treats the `src/content/` directory as special. This is where **collections** (folders) of Markdown/MDX **entries** (files) can be stored, with a single configuration file to define each collection's **schema** (frontmatter data types and shape). Files other than your `.md`/`.mdx` content are not permitted inside `src/content/`.
 
-```astro
----
-import { getCollection } from 'astro:content';
 
-// Get all `src/content/blog/` entries
-const allBlogPosts = await getCollection('blog');
+## Querying Collections
```
sarah11918, line 245: From this section, to Generating Routes, I'm not sure the heading levels/right sidebar organization is quite right.

Content-wise, Rendering Content doesn't seem like it should be a `##` as it seems more a peer with "Using Content in Astro Templates". But, I'm not sure I'd throw *everything* under the title of "Querying Collections"? And, they alternate between "script stuff" and "templating stuff".

These related items seem to be:

Script - Making the queries/filtering
Template- Templating the metadata (but not using `render()`)
Script - Passing an entire Content as props
Template - Rendering the content itself with `render()`
Script/Template of modified metadata

I think these could all be `###` peers under one `##`? (Which we'd need a name for.)

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +173,250 @@ defineCollection({
...
 
 </details>
 
-## Generating pages from content collections
 
-You can create pages based on your content collections using [dynamic routes](/en/core-concepts/routing/#dynamic-routes). 
+## Generating Routes from Content
 
-Use `getCollection()` inside a [`getStaticPaths()`](/en/reference/api-reference/#getstaticpaths) function to query your content entries and provide the `slug` parameter for each page. 
+Because content collections live outside of the `src/pages` directory, its up to you to create your [dynamic routes](/en/core-concepts/routing/#dynamic-routes) that will render your content and any collection entries to HTML.
```
sarah11918, line 372:
```suggestion
Because content collections live outside of the `src/pages/` directory, its up to you to create [dynamic routes](/en/core-concepts/routing/#dynamic-routes) that will render your content and any collection entries to HTML.
```

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +173,250 @@ defineCollection({
...
-	  return blog.map(entry => ({
-	    params: { slug: entry.slug },
-	  }));
-	}
-	---
-	```
+### Building for static output (default)
 
-This will generate page routes for every entry in the `blog` collection, mapping each entry’s slug to a URL. For example, an entry at `src/content/blog/hello-world.md` will have a slug of `hello-world` and the collection entry `src/content/blog/en/intro.md` will have a slug of `en/intro`.
+If you are building a static website (Astro's default behavior), you are expected use Astro's [`getStaticPaths()`](/en/reference/api-reference/#getstaticpaths) to create multiple routes from the same `src/pages` component at build-time.
```
sarah11918, line 378:
```suggestion
If you are building a static website (Astro's default behavior), you will need to create multiple routes from a single page component at build-time using Astro's [`getStaticPaths()`](/en/reference/api-reference/#getstaticpaths).
```

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +173,250 @@ defineCollection({
...
 
-Because this dynamic route is in `src/pages/posts/`, the final URLs will be `/posts/hello-world/` and `/posts/en/intro/`.
+Inside of `getStaticPaths()`, you would call [`getCollection()`](/en/reference/api-reference/#getcollection) to query your content entries and then create all of the static paths you need using the `slug` parameter for each page.
```
sarah11918, line 380:
```suggestion
Inside of `getStaticPaths()`, call [`getCollection()`](/en/reference/api-reference/#getcollection) to query your content entries and then create all of the static paths you need using the `slug` parameter for each page. 
```

sarah11918 (commented):
Fred!

Here are my review notes for your consideration. You're of course free to ingore any and all of them. But, these are the comments that come to mind that I'd at least bring to your attention. Some might seem like style preferences, but I've tried to only do that when I think I can make things tighter.

So, here ya go!

P.S. I have not dealt with failing links/conflicts.

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,152 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
 
-## Getting started
+## What are Content Collections?
 
-First, run the [`astro dev`](/en/reference/cli-reference/#astro-dev), [`astro build`](/en/reference/cli-reference/#astro-build), or [`astro sync`](/en/reference/cli-reference/#astro-sync) commands to generate types. This will make the `astro:content` module available for [querying](#querying-content-collections) and [configuring](#configuring-collections) your content collections.
+A **content collection** is any directory inside the reserved `src/content` project directory, such as `src/content/newsletter` and `src/content/blog`. Only content collections are allowed inside the `src/content` directory. This directory cannot be used for anything else.
 
-### Update TypeScript configuration
+Content collections help you organize the content files in your project. A content collection can contain any number of related Markdown (`.md`) or MDX (`.mdx`) files. These files within a collection are called **content entries.**
```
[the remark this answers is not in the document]
FredKSchott: reworded! But kept a similar focus on defining the term. This term becomes important later so it feels important to give it that focus here.

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,152 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
+<FileTree>
+- src/content/
+  - **newsletter/** 
+    - week-1.md
+    - week-2.md
+    - week-3.md
+</FileTree>
 
-## Collections
+Once you have a content collection in place with at least one content entry, you can query your content using Astro's built-in APIs.
```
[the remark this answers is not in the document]
FredKSchott: you're right. Leaving it in but making it link to the query APIs, in case someone wants to skip over the configuration section and go straight to usage.

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,152 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
 </FileTree>
 
-## Configuring collections
+## Configuration
 
-You can configure your content collections with an optional `src/content/config.ts` file (`.js` and `.mjs` extensions are also supported). 
+Astro allows you to customize and configure your project's content collections through an optional `src/content/config.ts` file (`.js` and `.mjs` extensions are also supported). 
 
-This file currently allows you to [define a collection schema](#defining-a-collection-schema), and to [create custom slugs](#custom-entry-slugs) for your collections.
+Currently, this file supports [creating a collection schema](#creating-a-collection-schema) and [defining custom unique slugs](#defining-custom-slugs) for your collections.
```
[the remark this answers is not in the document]
FredKSchott: I actually think this whole sentence can go! it doesn't add much and is slightly confusing ("wait, why are they calling out that its limited to just two features?")

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,152 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
+Astro leverages [Zod](https://github.com/colinhacks/zod) to create frontmatter schemas in the form of [Zod objects](https://github.com/colinhacks/zod#objects). With Zod, Astro is able to validate every file's frontmatter within a collection *and* provide automatic TypeScript types when you go to query content from inside your project.
 
-You can extend any of these types with `.optional()` if a frontmatter property is not always required or `.defaultValue(value)` to provide a value to use when the property is not set in frontmatter. If only a limited set of values is valid for a property, you can specify these using [the `.enum()` method](https://github.com/colinhacks/zod#zod-enums).
+To use Zod in Astro, import the `z` utility from `"astro:content"`. This is a re-export of the Zod library, and you are able to use all Zod features inside of Astro. See [Zod’s README](https://github.com/colinhacks/zod) for complete documentation on how Zod works and what features are available.
```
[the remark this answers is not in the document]
FredKSchott: the "this is a re-export" point is important so that people know it's not somehow different from Zod. Reworded to help it read better.

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +173,250 @@ defineCollection({
...
+If you're using Git for version control, we recommend ignoring this generated directory by adding `.astro` to your `.gitignore`.
+:::
 
-Astro provides two functions to query collections:
+The `.astro` directory will be updated for you automatically anytime you run the [`astro dev`](/en/reference/cli-reference/#astro-dev), [`astro build`](/en/reference/cli-reference/#astro-build) commands. You can run [`astro sync`](/en/reference/cli-reference/#astro-sync) yourself to update `.astro` manually.
 
-### `getCollection()`
 
-`getCollection()` returns multiple entries in a collection. It requires the name of a `collection` as a parameter.  By default, it returns all items in the collection.
+## The "src/content" Directory
```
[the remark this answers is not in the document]
FredKSchott: Yup agreed, reworked it!

src/pages/en/guides/content-collections.mdx
FredKSchott: I'm not sure I entirely follow this, but I agree this feels a little incorrectly placed. Just reorganized it to be a party of "querying", lmk what you think.

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +173,250 @@ defineCollection({
...
-### Data returned from a collection query
+### Using content in Astro templates
 
-`getCollection()` and `getEntry()` will return entries that include:
- - `id` - a unique ID using the file path relative to `src/content/[collection]`.
- - `slug` - a URL-ready slug. Defaults to the `id` parsed by [github-slugger](https://github.com/Flet/github-slugger).
- - `data` - an object of frontmatter properties inferred from your collection schema. Defaults to `any` if no schema is configured.
- - `body` - a string containing the raw, uncompiled body of the Markdown or MDX document.
- - `render()` - a function that returns the compiled body of the Markdown or MDX document via the `<Content />` component ([See complete documentation](#rendering-entry-content)).
+Once you have queried your collection entries, you can use those entries to render information about your content. You can access each entry directly inside of your Astro component template. This lets you to render HTML for things like links to your content (using the content `slug`) or information about your content (using the `data` property).
```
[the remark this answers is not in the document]
FredKSchott: this kind of edit is the best

FredKSchott:
@sarah11918 thanks for the amazing review! Made some great improvements based on your feedback. It sounds like you're either "LGTM" or close to it, so I'll leave this open until Monday or until I see your approval come through. 

I'm feeling pretty good about where this is now, but don't hesitate to give this another review if you have the time and feel the need.

sarah11918:
Just a quick reminder that merging this will remove the experimental note, so we might want to wait until Tuesday anyway! We have a "merge.on.release" label we can add for extra safety against forgetting.

I'm sure it will be great, and I'll just find time to scan for typos etc. just to be safe.

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,196 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
 
-## Getting started
+## What are Content Collections?
 
-First, run the [`astro dev`](/en/reference/cli-reference/#astro-dev), [`astro build`](/en/reference/cli-reference/#astro-build), or [`astro sync`](/en/reference/cli-reference/#astro-sync) commands to generate types. This will make the `astro:content` module available for [querying](#querying-content-collections) and [configuring](#configuring-collections) your content collections.
+A **content collection** is any directory inside the reserved `src/content` project directory, such as `src/content/newsletter` and `src/content/blog`. Only content collections are allowed inside the `src/content` directory. This directory cannot be used for anything else.
 
-### Update TypeScript configuration
+A **content entry** is any piece of content stored inside of your content collection. Content entries are stored as either Markdown (`.md`) or MDX (`.mdx`) files, saved inside of content collection directory. Your content entries can be named anything you want, but we recommend using a consistent naming scheme to make it easier to find and organize your content.
```
sarah11918, line 26:
```suggestion
A **content entry** is any piece of content stored inside of your content collection. Content entries are stored as either Markdown (`.md`) or MDX (`.mdx`) files, saved inside of the content collection directory. Your content entries can be named anything you want, but we recommend using a consistent naming scheme to make it easier to find and organize your content. 
```
At a minimum, there's a missing word, so that's what I've corrected.

Additionally: I feel like this got wordier.  :smile:

I would personally hesitate to put the word "any" in the reader's mind when we're about to restrict usage to Markdown and MDX files. This would be my suggestion:

> A **content entry** is a Markdown (`.md`) or MDX (`.mdx`) file stored inside of the content collection directory. 

I also don't feel like the advice on naming is particularly helpful? Are there *absolutely no* restrictions on file names? (And if so, I'd phrase it that way because that could be helpful info worth noting, and a little more in keeping with the docs tone.) It can start with a special character? An underscore? Contain emojis? I also don't think it's needed to suggest consistent naming, but if phrased in terms of letting them know that it's a "flat structure" or something like that, maybe?

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,196 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
 
-## Collections
 
-A collection is a directory in `src/content/` containing Markdown or MDX files. Every Markdown or MDX file in `src/content/` **must** belong to a collection directory, since Astro [provides built-in functions](#querying-content-collections) for querying your content by the collection directory name. 
+### Organizing with multiple collections 
 
-Content within a collection should share the same frontmatter shape and types. You can optionally enforce these types [by configuring a schema](/en/guides/content-collections/#defining-a-collection-schema).
+If two files represent two different kinds of content (e.g. a blog post and an author profile), they most likely belong in two different collections. This is an important concept to remember because important features like frontmatter validation and automatic TypeScript typings require that all entries share a similar frontmatter structure.
```
sarah11918, line 56:
```suggestion
If two files represent different kinds of content (e.g. a blog post and an author profile), they most likely belong in different collections. Helpful features like frontmatter validation and automatic TypeScript typings require that all entries in the same collection share a similar frontmatter structure.
```
I came for the double "important" and I stayed to avoid narrating the remembering process.

I chose "helpful" features, but could also see "powerful" features. I think those are more helpful to the reader than "important" (since these are not actually required).

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,196 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
 
-### Collections with nested directories
 
-Collections are **top-level folders** within `src/content/`. You cannot nest collections, but you may use nested directories within a collection to better organize a collection's content. All nested directories will share the same schema defined for the top-level collection.
+### Organizing with subdirectories
 
-For example, you can use this structure for internationalization:
+A content collection is always a top-level folder inside of the `src/content/` directory. You cannot nest one collection inside of another. However, you can use subdirectories to organize your content within a collection.
+
+For example, you can use the following directory structure to organize i18n translations within a single `docs` collection. When you go to query this collection, you'll be able to filter the result by language using the file path.
```
sarah11918, line 77:
```suggestion
For example, you can use the following directory structure to organize i18n translations within a single `docs` collection. When you query this collection, you'll be able to filter the result by language using the file path. 
```
Helps to keep the sentence moving, and also helps to not have extraneous verbs for the translators so they can more easily focus on the needed action verbs.

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,196 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
+    - **en/**
+    - **es/**
+    - **de/**
 </FileTree>
 
-## Configuring collections
+## Defining Collections
 
-You can configure your content collections with an optional `src/content/config.ts` file (`.js` and `.mjs` extensions are also supported). 
+:::caution
```
sarah11918, line 89: We know from exprience, it's a scary look in docs to have a caution right under a heading! Especially a "normal" heading like "Defining Collections." -> "Oh, let's see how to define my collection.... FLAMING CAUTION"

This caution should be moved to the bottom of the section, and I'd argue a caution is perhaps overkill for this and the text can stand on its own. The sentiment is fine, but at best this seems like a "note" or a "tip", which is at least a little more playful.

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,196 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
+
+```json title="tsconfig.json" ins={5}
+{
+  // Note: No change needed if you use "astro/tsconfigs/strict" or "astro/tsconfigs/strictest"
+  "extends": "astro/tsconfigs/base",
+  "compilerOptions": {
+    "strictNullChecks": true
+  }
+}
+```
```
sarah11918, line 122:
``````suggestion
To get the most out of content collections in Astro, you may need to update your `tsconfig.json` to enable `strictNullChecks`. This change is only required if currently extend Astro's `base` TypeScript settings in your `tsconfig.json` file.

```json title="tsconfig.json" ins={5} "base"
{
  // Note: No change needed if you use "strict" or "strictest"
  "extends": "astro/tsconfigs/base",
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```
``````
I think we can tighten this up by avoiding the negative framing.

File: src/pages/en/guides/content-collections.mdx [hunk 24]
```diff
@@ -16,166 +16,196 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
+
+### Defining a collection schema
 
-To configure schemas, create a `src/content/config.ts` file (`.js` and `.mjs` extensions are also supported). This file should:
+Schemas enforce consistent frontmatter within a collection. A schema **guarantees** that your frontmatter exists in a predictable form when you need to reference or query it. If any file violates its collection schema, Astro will provide a helpful error to let you know.
 
-1. Import the `defineCollection` and `z` utilities from `astro:content`. 
-2. Define a `schema` for each collection.
-2. Export a single `collections` object, with each object key corresponding to the collection's folder name.
+Schemas also power Astro's automatic TypeScript typings for your content. When you define a schema for your collection, Astro will automatically generate and apply a TypeScript interface to it. The result is full TypeScript support when you query your collection, including proprety auto-completion and type-checking.
```
sarah11918, line 128:
```suggestion
Schemas also power Astro's automatic TypeScript typings for your content. When you define a schema for your collection, Astro will automatically generate and apply a TypeScript interface to it. The result is full TypeScript support when you query your collection, including proprety autocompletion and type-checking. 
```
No hyphen

src/pages/en/guides/content-collections.mdx
```diff
@@ -16,166 +16,196 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
 });
+// 3. Export a single `collections` object to register your collection(s)
+export const collections = {
+  'blog': blogCollection,
+};
+```
+
+### Defining multiple collections
 
+You can define as many collections in your project as you would like. As long as they are exported from the `src/content/config.ts` file from inside the `collections` object, they will be loaded by Astro.
```
sarah11918, line 156:
```suggestion
You can use `defineCollection()` as many times as you want to create multiple schemas.  All collections must be exported from inside the single `collections` object.
```
I think this might be clearer, more explicitly hitting home "defineCollections many times, export a collections object once."

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +217,211 @@ defineCollection({
...
-### Data returned from a collection query
+### Using content in Astro templates
 
-`getCollection()` and `getEntry()` will return entries that include:
- - `id` - a unique ID using the file path relative to `src/content/[collection]`.
- - `slug` - a URL-ready slug. Defaults to the `id` parsed by [github-slugger](https://github.com/Flet/github-slugger).
- - `data` - an object of frontmatter properties inferred from your collection schema. Defaults to `any` if no schema is configured.
- - `body` - a string containing the raw, uncompiled body of the Markdown or MDX document.
- - `render()` - a function that returns the compiled body of the Markdown or MDX document via the `<Content />` component ([See complete documentation](#rendering-entry-content)).
+Once you have queried your collection entries, access each entry directly inside of your Astro component template. This lets you to render HTML for things like links to your content (using the content `slug`) or information about your content (using the `data` property).
```
sarah11918, line 293:
```suggestion
Once you have queried your collection entries, you can access each entry directly inside of your Astro component template. This lets you to render HTML for things like links to your content (using the content `slug`) or information about your content (using the `data` property).
```
Not wrong! But feels like this fits the sentence, so I just wanted to make sure this was a deliberate omission. Ignore if intended!

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +217,211 @@ defineCollection({
...
+<Content />
 ```
 
-### Access the `<Content />` component from `render()`
 
-To render the content of a Markdown or MDX entry, use the `<Content />` component returned by its `render()` function. This allows you to generate pages from your content entries (see [Generating pages from content collections](#generating-pages-from-content-collections)), add post previews to your homepage, or display your content elsewhere on your site.
+## Generating Routes from Content
 
-For example, this page renders the contents of `content/announcements/welcome.md` and uses some of its frontmatter properties:
+Because content collections live outside of the `src/pages/` directory, its up to you to create [dynamic routes](/en/core-concepts/routing/#dynamic-routes) that will render your content and any collection entries to HTML.
```
sarah11918, line 352:
```suggestion
Because content collections live outside of the `src/pages/` directory, it's up to you to create [dynamic routes](/en/core-concepts/routing/#dynamic-routes) that will render your content and any collection entries to HTML.
```
nit: apostrohe

src/pages/en/guides/content-collections.mdx
```diff
@@ -187,290 +217,211 @@ defineCollection({
...
-{blogPosts.map(async (post) => {
-  const { remarkPluginFrontmatter } = await post.render();
-  return <p>{post.data.title} — {remarkPluginFrontmatter.readingTime}</p>
-})}
+<h1>{entry.data.title}</h1>
+<Content />
 ```
 
-Assuming `readingTime` was injected ([see our reading time example](/en/guides/markdown-content/#example-calculate-reading-time)), it will be available on the `remarkPluginFrontmatter` object.
+## Modifying Frontmatter with Remark
```
sarah11918, line 404: This is an appropriate use of the caution! Something we don't actually want anyone to do!  :smile:

sarah11918 (commented):
OK, a small handful of corrections that *must* be changed (typos etc.) and a few more things for your consideration!

I am now confident that whatever you choose to make of this, it'll be great!  🥳

src/pages/en/guides/content-collections.mdx
FredKSchott: This was something we went back and forth on last week, but settled on the implemented behavior that filenames can be literally anything! I'd still rather be much more strict personally (only lower-case, dashes, etc) but this felt too late to change. 

Post-launch, I would love for us to pick this back up as a docs "recommendation", so that even if we keep supporting anything, we could add a recommendation here for a certain format.

src/pages/en/guides/content-collections.mdx
FredKSchott: no worries, I'll switch to note!

src/pages/en/guides/content-collections.mdx
FredKSchott: This is no longer true though. There are situations where you can be extending something other than base (or not extending anything, and defining everything yourself) that you still need to make this change.

FredKSchott:
I'm going to merge this sooner actually, because I realized that the current docs have gotten out of date with the API. There will likely be one more PR to this page before release, so that will give us a chance for any final review before Tuesday's launch.

PR #2373, opened 2023-01-14
Update Preview Deployments URL

morinokami:

#### Description

- I noticed that the page to which the original link pointed, https://vercel.com/docs/concepts/deployments/environments, did not have a heading corresponding to `#preview`, so I updated the url so that it points to the Preview Deployments docs page.


kevinzunigacuellar:
nice catch @morinokami, thanks a lot

PR #2382, opened 2023-01-14
[tutorial] Update outdated image src

sarah11918:
Updates the image used in the code snippet for the fourth blog post. The original image no longer exists, and this updates with a new one.

PR #2388, opened 2023-01-16
Use more reliable image source for tutorial

delucis:

#### Description

- Follow up to #2382 
- The tutorial is using an image from astro.build but because these are optimized with `@astrojs/image` they are liable to change filename, breaking the tutorial. This PR switches to using [a static image](https://docs.astro.build/default-og-image.png) within the docs site instead.


Yan-Thomas (approved):
LGTM!

PR #2389, opened 2023-01-16
feat(tsconfig): Use Astro's tsconfig preset

Princesseuh:

- Something else!

#### Description

Update the tsconfig.json to Astro's presets instead of manually writing things

image

### DIFF
--- a/src/components/tutorial/TutorialNav.astro
+++ b/src/components/tutorial/TutorialNav.astro
@@ -54,7 +54,8 @@ const isCurrentUnit = (unit: typeof units[number]) =>
 									</div>
 									{currentUrl.endsWith(lesson.slug) && (
 										<ul class="on-this-page">
-											{(await lesson.getHeadings())
+											{lesson
+												.getHeadings()
 												.filter(({ depth }) => depth <= 2)
 												.map((h) => (
 													<li>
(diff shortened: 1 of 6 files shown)


src/components/tutorial/TutorialNav.astro
```diff
@@ -54,7 +54,8 @@ const isCurrentUnit = (unit: typeof units[number]) =>
 									</div>
 									{currentUrl.endsWith(lesson.slug) && (
 										<ul class="on-this-page">
-											{(await lesson.getHeadings())
```
Princesseuh, line 57: Unrelated, but this made an hint in `astro check`, since the await is unneeded (`getHeadings()` is apparently sync)

delucis: The types say so but I seem to recall I actually needed the `await` when building that, probably worth double checking.

Princesseuh: Seems to work: 
image

delucis: Nice — thanks for checking!

tsconfig.json
```diff
@@ -1,21 +1,13 @@
 {
+  "extends": "astro/tsconfigs/strict",
```
delucis, line 2: LGTM!

One question: how far are we from being able to use `strictest`?

Princesseuh: 38 errors away, 36 of which are caused by `strictNullChecks: true` 😄

delucis (approved):

PR #2390, opened 2023-01-16
Remove fake tabs from install guides

delucis:

- New or updated content

#### Description

- Supersedes #2326 

The “fake” tab pattern in the install guides has been tripping us up a bit. These are two separate pages, so let’s treat them that way properly! This also gives us some leeway to stress preferences a bit and encourage use of the `create astro` guide which is generally the better DX.

### DIFF
--- a/src/pages/en/install/manual.mdx
+++ b/src/pages/en/install/manual.mdx
@@ -11,7 +11,7 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
 This guide will walk you through the steps to manually install and configure a new Astro project if you prefer not to use [the automatic CLI tool](/en/install/auto/).
 
-#### Want a quick way to get started?
+#### Prefer a quicker way to get started?
 
 <Button href="/en/install/auto/">Try the create astro CLI wizard →</Button>
 
--- a/src/pages/en/install/manual.mdx
+++ b/src/pages/en/install/manual.mdx
@@ -4,20 +4,22 @@ description: 'How to install Astro manually with NPM, PNPM, or Yarn.'
 layout: ~/layouts/MainLayout.astro
 i18nReady: true
 ---
+import Button from '~/components/Button.astro'
 import FileTree from '~/components/FileTree.astro';
-import InstallGuideTabGroup from '~/components/TabGroup/InstallGuideTabGroup.astro';
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
 
-Ready to install Astro? Follow our automatic or manual set-up guide to get started.
+Ready to install Astro? Follow this manual set-up guide to get started.
+
+#### Want a quick way to get started?
+
+<Button href="/en/install/auto/">Try the create astro CLI wizard →</Button>
 
 #### Prerequisites
 
 - **Node.js** - `v16.12.0` or higher.
 - **Text editor** - We recommend [VS Code](https://code.visualstudio.com/) with our [Official Astro extension](https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode).
 - **Terminal** - Astro is accessed through its command-line interface (CLI).
-
-<InstallGuideTabGroup />
 
 #### Installation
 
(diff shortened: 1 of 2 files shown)


src/pages/en/install/manual.mdx
```diff
@@ -4,21 +4,23 @@ description: 'How to install Astro manually with NPM, PNPM, or Yarn.'
...
 i18nReady: true
 ---
+import Button from '~/components/Button.astro'
 import FileTree from '~/components/FileTree.astro';
-import InstallGuideTabGroup from '~/components/TabGroup/InstallGuideTabGroup.astro';
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
 
-Ready to install Astro? Follow our automatic or manual set-up guide to get started.
+Ready to install Astro? Follow this manual set-up guide to get started.
```
sarah11918, line 12:
```suggestion
This guide will walk you through the steps to manually install and configure a new Astro project.
```
I'm thinking something less "encouraging" here, to nudge people towards the wizard unless they're darn sure they want to go through the manual process.

src/pages/en/install/manual.mdx
```diff
@@ -4,21 +4,23 @@ description: 'How to install Astro manually with NPM, PNPM, or Yarn.'
...
+
+#### Want a quick way to get started?
```
sarah11918, line 14:
```suggestion
#### Want to let Houston do the work for you instead?
```
Too cheesy?  :sweat_smile:

sarah11918 (commented):
OK, not married to either of these suggestions as written, but my goals are:
- nudge towards CLI wizard (in case people end up here not by choice, thinking this IS how you install Astro and not realizing that the other is way more common, and preferred.)
- Present the "Try the CLI!" in a *comparative* light: it's... easiER? fastER? less work? I want there to be an explicit reason to try the CLI. (Hence, my attempt at "let Houston do the work.")

src/pages/en/install/manual.mdx
delucis: Would this be overkill? I’m cool with your version if so.

```suggestion
This guide will walk you through the steps to manually install and configure a new Astro project if you prefer not to use [the automatic CLI tool](/en/install/auto/).
```

src/pages/en/install/manual.mdx
sarah11918: > Want to let our automated assistant Houston do the work for you instead?

As I wrote that, I was like, "ugh, I hate automated assistants" so NOT that.

But maybe something like: Want a guided CLI experience?

delucis: If we want comparative maybe just?
```suggestion
#### Want a quicker way to get started?
```

delucis: I don’t think the CLI part is really the important thing, it’s more the speed/convenience angle that’s attractive. Plus “CLI wizard” is in the button text. I thought “quick and easy” initially but usually avoid “easy” if I can.

sarah11918: Yeah, my objection was that I'm not sure this clearly enough says that it's a REPLACEMENT for doing the steps below. For someone who doesn't know any better, this could be a way to start THE STEPS BELOW.

sarah11918: > Prefer a guided installation experience?

delucis: Ahhh, missed that nuance. How about…

```suggestion
#### Want a quicker way to get started instead?
```

delucis: Ooh, I do like “prefer” to help communicate this.

delucis: Maybe?

```suggestion
#### Prefer a quicker way to get started?
```

delucis:
> I'm noticing we have no pagination at the bottom of the manual installation page.

Yeah, only pages in the left sidebar know what their before/after pages are, so anything not included there won’t have pagination links.

sarah11918 (approved):

Tc-001:
Yay looks good!
Glad this could be solved nicely

PR #2391, opened 2023-01-16
Improve subpages handling and add missing pages

Yan-Thomas:

#### Description

This PR improves our sidebar subpages handling, now making it work for pages with technically infinite segments, so `en/example/[subpage-1]/[subpage-2]/[subpage-3]/[subpage-4]` will still be considered a subpage of `en/example/`, not only `en/example/[subpage]`.

This improves the navigation experience by highlighting the current main page in the sidebar and also collapsing the page section on mobile. I also added the missing "CMS" and "Migrate to Astro" to the subpages regex.

| Before | After |
| 5zcVqgU | UEpLS0S |

FIRST SIDEBAR PR OF THE YEAR BTW! 🥳

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -17,9 +17,10 @@ export function removeTrailingSlash(path: string) {
 export function removeSubpageSegment(path: string) {
 	// Include new pages with subpages as part of this regex.
 	const regex = /(?:install|deploy|integrations-guide|tutorial|migrate-to-astro|cms)\//;
+	const matches = regex.exec(path);
 
-	if (regex.test(path)) {
-		const matchIndex = regex.exec(path)!.index;
+	if (matches) {
+		const matchIndex = matches.index;
 		// Get the first slash index after the main page path segment.
 		const slashIndex = path.slice(matchIndex).indexOf('/') + matchIndex;
 		return path.slice(0, slashIndex);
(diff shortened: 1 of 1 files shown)


src/util.ts
```diff
@@ -15,9 +15,14 @@ export function removeTrailingSlash(path: string) {
...
 /** Remove the subpage segment of a URL string */
 export function removeSubpageSegment(path: string) {
-	// Include new pages with subpages as part of this if statement.
-	if (/(?:install|deploy|integrations-guide|tutorial)\//.test(path)) {
-		return path.slice(0, path.lastIndexOf('/'));
+	// Include new pages with subpages as part of this regex.
+	const regex = /(?:install|deploy|integrations-guide|tutorial|migrate-to-astro|cms)\//;
+
+	if (regex.test(path)) {
+		const matchIndex = regex.exec(path)!.index;
```
delucis, line 22: We can probably avoid the non-null assertion and running the regex twice like this, I think? `regex.exec()` returns `null` if it doesn’t match.

```suggestion
	const matches = regex.exec(path);
	if (matches) {
		const matchIndex = matches.index;
```

delucis (commented):
Thanks, Yan! Looks good, just one code style suggestion.

PR #2397, opened 2023-01-17
Minor update on Astro component docs

pilcrowOnPaper:

- New or updated content

#### Description

Closes (none)

This PR was inspired by a discussion I had when reviewing a translation for the docs on Astro components. Here's the code block in question:

```html
<!-- Include other components with a `client:` directive to hydrate: -->
<ReactPokemonComponent client:visible />
```

Personally, I found the comment to be confusing. I understood what it meant but felt it could be improved. Since it's showcases 2 things (using imported components and the client directive), I felt the comment should be split into two:

```html
<!-- Use imported components like any other UI frameworks -->
<!-- and hydrate them by adding a `client` directive`: -->
<ReactPokemonComponent client:visible />
```

I feel this makes it clear the intention behind the code example.

### DIFF
--- a/src/pages/en/core-concepts/astro-components.mdx
+++ b/src/pages/en/core-concepts/astro-components.mdx
@@ -95,7 +95,7 @@ const myFavoritePokemon = [/* ... */];
 <!-- Use props and other variables from the component script: -->
 <p>My favorite pokemon is: {Astro.props.title}</p>
 
-<!-- Use imported components like any other UI frameworks  -->
+<!-- Use imported components like any other UI frameworks -->
 <!-- and hydrate them by adding a `client` directive`: -->
 <ReactPokemonComponent client:visible />
 
--- a/src/pages/en/core-concepts/astro-components.mdx
+++ b/src/pages/en/core-concepts/astro-components.mdx
@@ -95,7 +95,8 @@ const myFavoritePokemon = [/* ... */];
 <!-- Use props and other variables from the component script: -->
 <p>My favorite pokemon is: {Astro.props.title}</p>
 
-<!-- Include other components with a `client:` directive to hydrate: -->
+<!-- Use imported components like any other UI frameworks  -->
+<!-- and hydrate them by adding a `client` directive`: -->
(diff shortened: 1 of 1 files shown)


src/pages/en/core-concepts/astro-components.mdx
```diff
@@ -95,7 +95,8 @@ const myFavoritePokemon = [/* ... */];
 <!-- Use props and other variables from the component script: -->
 <p>My favorite pokemon is: {Astro.props.title}</p>
 
-<!-- Include other components with a `client:` directive to hydrate: -->
+<!-- Use imported components like any other UI frameworks -->
+<!-- and hydrate them by adding a `client` directive`: -->
```
sarah11918, line 99:
```suggestion
<!-- Include components from other UI frameworks -->
<!-- Hydrate them with a `client` directive`: for interactivity-->
```
Thanks for picking up the discussion with a PR!  Let's try to make this better! :partying_face: 

What do you think about this slight change? Does this express your two ideas cleanly?

The problem with using "like" is that in English, it can mean so many different things. It's just not precise enough:

> Use components, *for example, use those from* other UI frameworks

> Use components, *just as / the same way you use them* in other UI frameworks

Both of these are entirely possible interpretations here.

pilcrowOnPaper: Yeah, I realized it wasn't clear enough just as I was going to bed and I was just in the middle of editing it 😅. My original intention was the latter ("the same way you use..."), but I think the word "use" is too vague here as well. In all other frameworks, components get hydrated automatically but that's not the case in Astro. 

I also thought it would be better to explain the syntax for including external components, but I now think it's pretty self-explanatory so just having the code example should be good enough.

Going back to hydration though, this is something that should be clearly explained in the page now I think about it. There's a comment about the client directive and hydration, but it doesn't explain why those are necessary in the first place. This usually isn't an issue since people will search for stuff they don't know (like the directives), but for UI components, there's some preconceptions regarding it that should be clearly addressed. I'll close this PR and try to create a new one with a few more tweaks.

PR #2309, opened 2023-01-06
NuxtJS Migration guides to Astro

crutchcorn:
> This PR is sibling work to #2211, but does not depend on it. They can be merged and reviewed seperately.

This is a first draft of the "Migrate to Astro from NuxtJS" article.

**The goals of this page are:**

- To highlight the key differences and similarities **to prepare the reader for switching frameworks**
- To **guide a little more explicitly through initial project creation**
- To **provide general but comprehensive guidance on converting existing framework templating files (like `.jsx`) into `.astro` components**, including:

  - A before/after example of a component
  - SEVERAL examples of walk-through conversions
  - A list of several key syntax differences (e.g. `<NextLink>`, `:class`, `{{variable}}`, `v-if`) and their Astro equivalents
- **To suggest a migration path** starting with  converting Layouts (so that pages will be complete HTML documents when  using a Layout), then importing existing pages and posts.

**The challenges of this page are:**

- **To fulfill the goals without overwhelming the reader**. I'm less concerned about page length itself and more concerned that the sidebar doesn't make people say, "Yikes!"

- **To write section headings that will not confuse our Algolia search results** and send people to this page for "normal docs things" (e.g. Image  Integration, Site Configuration). This is why the headings are now  intentionally long-ish, because shortening them runs the risk of Algolia mistaking the section for regular docs.

PR #2211, opened 2022-12-19
Next.js/Gatsby Migration guides to Astro

crutchcorn:
> This is a continuation of https://github.com/withastro/docs/pull/2170, migrated back to Astro's docs repo per @sarah11918's guidance. Parts of this PR description was written by @sarah11918

Edited to add: 
Discord thread for even more discussion at: https://discord.com/channels/830184174198718474/1052270358721470484

This is a first draft of a series of "Migrate to Astro from" articles. The initial migration guides include:

- [x] Migrating from Gatsby
- [x] Migrating from Next.js
- [x] ~~Migrating from Nuxt.js~~ (to follow in another PR)

As well as stubs for many other migrations guides that we're hoping to have community contributions towards.

Because we're unable to predict exactly what will be in someone's existing project, these are less step-by-step walkthroughs and more "general guidance to help you migrate". That said, we do have a few step-by-step sections simply to demonstrate what an example project _could_ look like.

These sections may be un-necessarily long in the current draft stage; that's intentional. We'll cut as we do our editing and boil it down to what we feel is right.

**The goals of this page are:**

- To highlight the key differences and similarities **to prepare the reader for switching frameworks**

- To **guide a little more explicitly through initial project creation**

- To explain `astro.config` in relation to a Framework's configuration files, and **to help the reader extract any existing configuration** and put it into the appropriate places

- To **provide general but comprehensive guidance on converting existing framework templating files (like `.jsx`) into `.astro` components**, including:

  - A before/after example of a component
  - SEVERAL examples of walk-through conversions
  - A list of several key syntax differences (e.g. `className`, `{children}`, `<Link to="" />` `style={{}}`) and their Astro equivalents

- **To suggest a migration path** starting with  converting Layouts (so that pages will be complete HTML documents when  using a Layout), then importing existing pages and posts.

**The challenges of this page are:**

- **To fulfill the goals without overwhelming the reader**. I'm less concerned about page length itself and more concerned that the sidebar doesn't make people say, "Yikes!"
- **To write section headings that will not confuse our Algolia search results** and send people to this page for "normal docs things" (e.g. Image  Integration, Site Configuration). This is why the headings are now  intentionally long-ish, because shortening them runs the risk of Algolia mistaking the section for regular docs.
[...]

PR #2400, opened 2023-01-17
Add a link to blog post about migration from Gridsome

fyodorio:

- New or updated content

#### Description

- Extend community resources list for "Migrating from Gridsome" page with new community content

### DIFF
--- a/src/pages/en/guides/migrate-to-astro/from-gridsome.mdx
+++ b/src/pages/en/guides/migrate-to-astro/from-gridsome.mdx
@@ -66,4 +66,4 @@ To convert other types of sites, such as a portfolio or documentation site, see 
 
 ## Community Resources
 
-- Add your own!
+- Blog Post: [Migration from Gridsome to Astro](https://fyodor.io/migration-from-gridsome-to-astro/).


sarah11918:
This looks great, @fyodorio! I read through your article and I loved your thoroughness there!  :partying_face: 

As you noted in your post, our existing migration guides are currently "stubs" with just some helpful pointers so that we can draw attention to some key differences between platforms to set you in the right direction.

We are currently polishing and very close to releasing some initial beefier guides (Gatsby, Next and Nuxt) that are a little closer to what you've written. Of course, they can't be quite as specific because we can't assume what people are already using/doing in their projects. But they do go into a little more detail. No pressure, but if you'd consider leaving a comment regarding overall structure/helpfulness as we explore this model, we'd absolutely appreciate a peek at what we've got going on in #2211 and #2309. 

I'm thrilled to link to your post here on this page, and if you ever wanted to write something a little more detailed for this page itself, to match our fuller (general advice) guides that are coming, please keep us in mind!

In the mean time, welcome to Team Docs!  :raised_hands:

PR #2404, opened 2023-01-18
Update Astro component docs

pilcrowOnPaper:

- New or updated content

#### Description

Closes (none)

1. Address SSR when talking about rendering
2. Moved the contents of "next steps" section
3. Remove the "next steps" section entirely as the pages are already in order
4. Replace React component in the code example to Astro component since framework components should be explained in a different page
5. Minor changes

### DIFF
--- a/src/pages/en/core-concepts/astro-components.mdx
+++ b/src/pages/en/core-concepts/astro-components.mdx
@@ -13,7 +13,7 @@ Astro component syntax is a superset of HTML. The syntax was [designed to feel f
 
 Astro components are extremely flexible. Often, an Astro component will contain some **reusable UI on the page**, like a header or a profile card. At other times, an Astro component may contain a smaller snippet of HTML, like a collection of common `<meta>` tags that make SEO easy to work with. It can even contain an entire page layout.
 
-The most important thing to know about Astro components is that they **render to HTML ahead of time**. It could be during the build step or on the server on demand, depending on how you have configured your project. Even if you run JavaScript code inside of your components, all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default. 
+The most important thing to know about Astro components is that they **render to HTML**, be it ahead of time in static builds or on-demand in [server-side rendering (SSR)](/en/guides/server-side-rendering/). Even if you run JavaScript code inside of your components, all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
 
 
 ## Component Structure
--- a/src/pages/en/core-concepts/astro-components.mdx
+++ b/src/pages/en/core-concepts/astro-components.mdx
@@ -116,7 +116,7 @@ You're not limited to just Astro components though. You can also include **JavaS
 You can define local JavaScript variables inside of the frontmatter component script within an Astro component. You can then inject these variables into the component's HTML template using JSX-like expressions!
 
 :::note[dynamic vs reactive]
-Using this approach, you can include **dynamic** values that are calculated in the frontmatter. But once included, these values are not **reactive** and will never change. Astro components are templates that only run during the rendering step.
+Using this approach, you can include **dynamic** values that are calculated in the frontmatter. But once included, these values are not **reactive** and will never change. Astro components are templates that only run once, at build time.
 
 See below for more examples of [differences between Astro and JSX](#differences-between-astro-and-jsx).
 :::
--- a/src/pages/en/core-concepts/astro-components.mdx
+++ b/src/pages/en/core-concepts/astro-components.mdx
@@ -13,7 +13,7 @@ Astro component syntax is a superset of HTML. The syntax was [designed to feel f
 
 Astro components are extremely flexible. Often, an Astro component will contain some **reusable UI on the page**, like a header or a profile card. At other times, an Astro component may contain a smaller snippet of HTML, like a collection of common `<meta>` tags that make SEO easy to work with. Astro components can even contain an entire page layout.
 
-The most important thing to know about Astro components is that they **render to HTML**, be it ahead of time in static builds or on-demand in [server-side rendering (SSR)](/en/guides/server-side-rendering/). Even if you run JavaScript code inside of your components, all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
+The most important thing to know about Astro components is that they **don't render on the client**. They render either at build-time or on-demand using [server-side rendering (SSR)](/en/guides/server-side-rendering/). You can include JavaScript code inside of your component frontmatter, and all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
 
 
 ## Component Structure
--- a/src/pages/en/core-concepts/astro-components.mdx
+++ b/src/pages/en/core-concepts/astro-components.mdx
@@ -13,7 +13,7 @@ Astro component syntax is a superset of HTML. The syntax was [designed to feel f
 
 Astro components are extremely flexible. Often, an Astro component will contain some **reusable UI on the page**, like a header or a profile card. At other times, an Astro component may contain a smaller snippet of HTML, like a collection of common `<meta>` tags that make SEO easy to work with. Astro components can even contain an entire page layout.
 
-The most important thing to know about Astro components is that they **don't render on the client**. They render either at build-time or on-demand using [server-side rendering (SSR)](/en/guides/server-side-rendering/). You can include JavaScript code inside of your component frontmatter, and all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
+The most important thing to know about Astro components is that they **don't render on the client**. They render to HTML either at build-time or on-demand using [server-side rendering (SSR)](/en/guides/server-side-rendering/). You can include JavaScript code inside of your component frontmatter, and all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
 
 
 ## Component Structure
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks for this, @pilcrowOnPaper! It has been on my radar to update docs in general with more emphasis on including the SSR option inside and all around docs, not just as its own page somewhere. This is a great start to that! It's absolutely the right idea, and a very helpful contribution.

I'm going to make sure we get more eyes on this, too, as I'd like us to really dig into this one and make sure we're doing SSR justice!

src/pages/en/core-concepts/astro-components.mdx
```diff
@@ -11,9 +11,9 @@ i18nReady: true
...
+Astro components are extremely flexible. Often, an Astro component will contain some **reusable UI on the page**, like a header or a profile card. At other times, an Astro component may contain a smaller snippet of HTML, like a collection of common `<meta>` tags that make SEO easy to work with. It can even contain an entire page layout.
 
-The most important thing to know about Astro components is that they **render to HTML during your build.** Even if you run JavaScript code inside of your components, it will all run ahead of time, stripped from the final page that you send to your users. The result is a faster site, with zero JavaScript footprint added by default.
+The most important thing to know about Astro components is that they **render to HTML ahead of time**. It could be during the build step or on the server on demand, depending on how you have configured your project. Even if you run JavaScript code inside of your components, all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
```
Yan-Thomas, line 16: Not sure about the wording here, saying "Astro components render to HTML ahead of time [....] on demand in the server" feels kinda contradictory to me? How about changing this a bit?

```suggestion
The most important thing to know about Astro components is that they **render to HTML**, be it ahead of time in static builds or on-demand in SSR. Even if you run JavaScript code inside of your components, all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
```

Yan-Thomas (commented):
Thanks @pilcrowOnPaper, making the docs more SSR inclusive are a great idea. I have some thoughts on the wording though, let me know what you think!

src/pages/en/core-concepts/astro-components.mdx
pilcrowOnPaper: That sounds better 👍 

I guess we're allowed to use the word "SSR" here?

Yan-Thomas: Great question! I feel like SSR is such a common buzzword nowadays that most people will understand we are talking about server-side rendering, although I can see the argument for using the full term instead, I'll let @sarah11918 weight on this.

sarah11918: We're allowed to say it! Often if it's at the very top of a page like this, we might spell it out in full and then also put (SSR) in brackets after. And, I would probably choose to link to the SSR page from it.

Yan-Thomas: Agree!

```suggestion
The most important thing to know about Astro components is that they **render to HTML**, be it ahead of time in static builds or on-demand in [server-side rendering (SSR)](/en/guides/server-side-rendering/). Even if you run JavaScript code inside of your components, all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
```

pilcrowOnPaper: Sounds good!

src/pages/en/core-concepts/astro-components.mdx
```diff
@@ -11,9 +11,9 @@ i18nReady: true
...
+Astro components are extremely flexible. Often, an Astro component will contain some **reusable UI on the page**, like a header or a profile card. At other times, an Astro component may contain a smaller snippet of HTML, like a collection of common `<meta>` tags that make SEO easy to work with. It can even contain an entire page layout.
 
-The most important thing to know about Astro components is that they **render to HTML during your build.** Even if you run JavaScript code inside of your components, it will all run ahead of time, stripped from the final page that you send to your users. The result is a faster site, with zero JavaScript footprint added by default.
+The most important thing to know about Astro components is that they **render to HTML**, be it ahead of time in static builds or on-demand in [server-side rendering (SSR)](/en/guides/server-side-rendering/). Even if you run JavaScript code inside of your components, all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
```
Jutanium, line 16:
```suggestion
The most important thing to know about Astro components is that they **render to HTML**, be it ahead of time in static builds or on-demand in [server-side rendering (SSR)](/en/guides/server-side-rendering/). You can include JavaScript code inside of your component frontmatter, and all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
```

It isn't true that all of the JavaScript code inside of your components will be stripped; `<script>` tags are sent!

Jutanium, line 16: I would rephrase this paragraph to focus on **not rendering on the client**, rather than **rendering to HTML**. React components also render to HTML. Something like:

```suggestion
The most important thing to know about Astro components is that they **don't render on the client**. They render either at build-time or on-demand using [server-side rendering (SSR)](/en/guides/server-side-rendering/). You can include JavaScript code inside of your component frontmatter, and all of it will be stripped from the final page sent to your users' browsers. The result is a faster site, with zero JavaScript footprint added by default.
```

src/pages/en/core-concepts/astro-components.mdx
```diff
@@ -11,9 +11,9 @@ i18nReady: true
 
 Astro component syntax is a superset of HTML. The syntax was [designed to feel familiar to anyone with experience writing HTML or JSX](#differences-between-astro-and-jsx), and adds support for including components and JavaScript expressions. You can spot an Astro component by its file extension: `.astro`.
 
-Astro components are extremely flexible. Often, an Astro component will contain some **reusable UI on the page**, like a header or a profile card. At other times, an Astro component may contain a smaller snippet of HTML, like a collection of common `<meta>` tags that make SEO easy to work with. Astro components can even contain an entire page layout.
+Astro components are extremely flexible. Often, an Astro component will contain some **reusable UI on the page**, like a header or a profile card. At other times, an Astro component may contain a smaller snippet of HTML, like a collection of common `<meta>` tags that make SEO easy to work with. It can even contain an entire page layout.
```
sarah11918, line 14:
```suggestion
Astro components are extremely flexible. Often, an Astro component will contain some **reusable UI on the page**, like a header or a profile card. At other times, an Astro component may contain a smaller snippet of HTML, like a collection of common `<meta>` tags that make SEO easy to work with. Astro components can even contain an entire page layout.
```

src/pages/en/core-concepts/astro-components.mdx
```diff
@@ -107,12 +107,16 @@ const myFavoritePokemon = [/* ... */];
...
+
+📚 Learn about using [JavaScript framework components](/en/core-concepts/framework-components/) in your Astro project.
+
 ## JSX-like Expressions
 
 You can define local JavaScript variables inside of the frontmatter component script within an Astro component. You can then inject these variables into the component's HTML template using JSX-like expressions!
 
 :::note[dynamic vs reactive]
-Using this approach, you can include **dynamic** values that are calculated in the frontmatter.  But once included, these values are not **reactive** and will never change. Astro components are templates that only run once, at build time.
+Using this approach, you can include **dynamic** values that are calculated in the frontmatter. But once included, these values are not **reactive** and will never change. Astro components are templates that only run during the rendering step.
```
sarah11918, line 119:
```suggestion
Using this approach, you can include **dynamic** values that are calculated in the frontmatter. But once included, these values are not **reactive** and will never change. Astro components are templates that only run once, at build time.
```
Reverting this change, since in the new PR, this part of the text is in a new file. I've copied the suggestion to include in the new Astro Syntax page. This way, all edits in this PR only discuss content staying on the components page.

pilcrowOnPaper:
Totally forgot about this PR since I was busy last week, my bad!

sarah11918:
Why are all 3 of us suddenly in this PR at once?  :sweat_smile:

src/pages/en/core-concepts/astro-components.mdx
```diff
@@ -107,12 +107,16 @@ const myFavoritePokemon = [/* ... */];
...
+
+📚 Learn about using [JavaScript framework components](/en/core-concepts/framework-components/) in your Astro project.
+
 ## JSX-like Expressions
 
 You can define local JavaScript variables inside of the frontmatter component script within an Astro component. You can then inject these variables into the component's HTML template using JSX-like expressions!
 
 :::note[dynamic vs reactive]
-Using this approach, you can include **dynamic** values that are calculated in the frontmatter.  But once included, these values are not **reactive** and will never change. Astro components are templates that only run once, at build time.
+Using this approach, you can include **dynamic** values that are calculated in the frontmatter. But once included, these values are not **reactive** and will never change. Astro components are templates that only run once, at build time.
```
sarah11918, line 119:
```suggestion
Using this approach, you can include **dynamic** values that are calculated in the frontmatter.  But once included, these values are not **reactive** and will never change. Astro components are templates that only run once, at build time.
```

sarah11918:
Hi @pilcrowOnPaper ! As I mentioned to you, restructuring this page *was* on our radar, and I have an open PR to chop up this page a bit that's going to make some of this obsolete.

I'm taking the appropriate suggestions from here and simply updating *that* PR so your changes can be reflected there, and won't get undone/overwritten.

I think the easiest thing will be for me to close this one once I've pulled out the content we want, and I'll make the changes we want in #2550 instead, since this page won't exist like this very soon.  :)  Sorry it took so long to get to this one, but we knew changes were coming and I think this is the best way to handle it now!

pilcrowOnPaper:
yeah that's fine, thanks for reviewing @sarah11918  @Yan-Thomas  @Jutanium !

PR #2408, opened 2023-01-18
Rename getEntry to getEntryBySlug

matthewp:

- New or updated content

#### Description

- Based on conversations in core it was decided to rename `getEntry` to `getEntryBySlug` to reduce the overlap between id and slug. The hope is that in the future we provide better utilities around types of collections; for example some collections are used to generate pages and should have a slug (or permalink) and some are not, in which case they should not have that.
- Here's the PR in core: https://github.com/withastro/astro/pull/5893


src/pages/en/guides/content-collections.mdx
```diff
@@ -274,28 +274,28 @@ const enDocs = await getCollection('docs', ({ id }) => {
 ---
 ```
 
-### `getEntry()`
+### `getEntryBySlug()`
 
-`getEntry()` is function that returns a specific entry in a collection by entry ID (file path relative to the collection). Both of these are required parameters.
+`getEntryBySlug()` is function that returns a specific entry in a collection by its slug (a slug that can be used in a URL). Both of these are required parameters.
```
sarah11918, line 279:
```suggestion
`getEntryBySlug()` is function that returns a specific entry in a collection by its slug (filename without the extension). Both of these are required parameters.
```
Is this what it is? Is the slug exactly the file name (minus extension), or is it user-defined?

Saying "*a* slug" makes it sound like it could be up for grabs. If  something defined in frontmatter, then maybe we could say something like:

> (a value defined in the frontmatter)

Otherwise, the rest of the changes look good to me!

sarah11918: Yes, I think linking to the term below is good. Why try to figure out how to say it twice?  😄 

I guess the closest heading is `#data-returned-from-a-collection-query`

If you'd rather, what about something like:

> its slug (URL-ready)

> its URL-ready slug

Does that convey it well enough?

matthewp: Yeah, I think either of those work well enough.

matthewp: Went with `its slug (URL-ready)`

matthewp:
Closing in favor of @FredKSchott's PR which has more changes.

PR #2409, opened 2023-01-18
content collections in styling, markdown pages

sarah11918:
This PR starts to sprinkle in content collections in appropriate places throughout the docs.

1. Markdown/MDX: Content collections is now the recommended suggestion for creating pages, so it is now listed under the `## Markdown and MDX Pages` heading.

1. Pages: Content collections is now mentioned in Pages

1. Styling of content collections came up in discussion, which is really no different than styling your Markdown content anywhere it occurs. So, in order to support new collections users this PR adds a section on Markdown styling to the Styling page. And, the Markdown/MDX page links TO it, rather than duplicating content.

TODO: Link to this new styling markdown section FROM the content collections page itself. This page already has a complicated PR in the works, so this will be added later to this PR or in a future one.

### DIFF
--- a/src/pages/en/guides/styling.mdx
+++ b/src/pages/en/guides/styling.mdx
@@ -12,7 +12,7 @@ import Since from '../../../components/Since.astro';
 
 Astro was designed to make styling and writing CSS a breeze. Write your own CSS directly inside of an Astro component or import your favorite CSS library like [Tailwind][tailwind]. Advanced styling languages like [Sass][sass] and [Less][less] are also supported.
 
-## Styling in Astro
+## Styling in Astro Components
 
 Styling an Astro component is as easy as adding a `<style>` tag to your component or page template. When you place a `<style>` tag inside of an Astro component, Astro will detect the CSS and handle your styles for you, automatically.
 
@@ -481,6 +481,13 @@ Vue in Astro supports the same methods as `vue-loader` does:
 
 Svelte in Astro also works exactly as expected: [Svelte Styling Docs][svelte-style].
 
+## Markdown Styling
+
+Any Astro styling methods are available to a [Markdown layout component](/en/core-concepts/layouts/#markdownmdx-layouts).
+
+You can apply global styles to your Markdown content by adding [`<style>` tags](#global-styles) and [imported stylesheets](#external-styles) to the layout that wraps your page content. You can also add [CSS integrations](#css-integrations) including [Tailwind](/en/guides/integrations-guide/tailwind).
+
+Additionally, you can install and use [Markdown plugins](/en/guides/markdown-content/#markdown-plugins) such as the [Tailwind/typography plugin](https://tailwindcss.com/docs/typography-plugin) for styling Markdown.
 
 ## Advanced
 
--- a/src/pages/en/guides/styling.mdx
+++ b/src/pages/en/guides/styling.mdx
@@ -487,7 +487,7 @@ Any Astro styling methods are available to a [Markdown layout component](/en/cor
 
 You can apply global styles to your Markdown content by adding [`<style>` tags](#global-styles) and [imported stylesheets](#external-styles) to the layout that wraps your page content. You can also add [CSS integrations](#css-integrations) including [Tailwind](/en/guides/integrations-guide/tailwind/).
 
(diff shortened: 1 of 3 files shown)


src/pages/en/guides/styling.mdx
```diff
@@ -12,7 +12,7 @@ import Since from '../../../components/Since.astro';
 
 Astro was designed to make styling and writing CSS a breeze. Write your own CSS directly inside of an Astro component or import your favorite CSS library like [Tailwind][tailwind]. Advanced styling languages like [Sass][sass] and [Less][less] are also supported.
 
-## Styling in Astro
+## Styling in Astro Components
```
sarah11918, line 15:
```suggestion
## Styling in Astro
```

src/pages/en/guides/styling.mdx
```diff
@@ -481,6 +481,13 @@ Vue in Astro supports the same methods as `vue-loader` does:
 
 Svelte in Astro also works exactly as expected: [Svelte Styling Docs][svelte-style].
 
+## Markdown Styling
+
+Any Astro styling methods are available to a [Markdown layout component](/en/core-concepts/layouts/#markdownmdx-layouts).
+
+You can apply global styles to your Markdown content by adding [`<style>` tags](#global-styles) and [imported stylesheets](#external-styles) to the layout that wraps your page content. You can also add [CSS integrations](#css-integrations) including [Tailwind](/en/guides/integrations-guide/tailwind).
```
sarah11918, line 488:
```suggestion
You can apply global styles to your Markdown content by adding [`<style>` tags](#global-styles) and [imported stylesheets](#external-styles) to the layout that wraps your page content. You can also add [CSS integrations](#css-integrations) including [Tailwind](/en/guides/integrations-guide/tailwind/).
```

src/pages/en/guides/styling.mdx
```diff
@@ -481,6 +481,13 @@ Vue in Astro supports the same methods as `vue-loader` does:
 
 Svelte in Astro also works exactly as expected: [Svelte Styling Docs][svelte-style].
 
+## Markdown Styling
+
+Any Astro styling methods are available to a [Markdown layout component](/en/core-concepts/layouts/#markdownmdx-layouts).
+
+You can apply global styles to your Markdown content by adding [`<style>` tags](#global-styles) and [imported stylesheets](#external-styles) to the layout that wraps your page content. You can also add [CSS integrations](#css-integrations) including [Tailwind](/en/guides/integrations-guide/tailwind/).
```
delucis, line 488: Do we need to say all these things again if they're already on this page and we're already saying "you can do anything on this page"?

src/pages/en/guides/styling.mdx
```diff
@@ -481,6 +481,13 @@ Vue in Astro supports the same methods as `vue-loader` does:
...
+
+Additionally, you can install and use [Markdown plugins](/en/guides/markdown-content/#markdown-plugins) such as the [Tailwind/typography plugin](https://tailwindcss.com/docs/typography-plugin) for styling Markdown.
```
delucis, line 490: The tailwind plug-in is a smart thing to call out here but it's not the same kind of plug-in as we document at that link so maybe we only mention tailwind here?

```suggestion
If you are using Tailwind, the [typography plugin](https://tailwindcss.com/docs/typography-plugin) can be useful for styling Markdown.
```

delucis (commented):
Thanks for going through and doing this @sarah11918! Hopefully these comments show up ok and if they don't... I'll blame it on my phone.

src/pages/en/guides/styling.mdx
sarah11918: So, I *get* that... but I struggled. This is our canonical, "How to style Markdown" and it seems like the answer is.... don't? :sweat_smile: Style in the layout component instead?

I'm also accounting for people who search and end up down here at the bottom of the page, so wanted to call out the things from above that are useful to them.

delucis: Ok, so sounds like the answer is "Yes, we do need to say these things"

I'm cool with that! 👍

delucis (approved):
Thanks, Sarah! Let's get those sprinkles 💖

PR #2413, opened 2023-01-19
Document using `--template` with other branch names

### LINKED ISSUE #2372: Wrong starter template description (and typo)
## What? 
The following is a rather ambiguous (slightly misleading) sentence as it gives the impression you can take any public Github astro project and start a new project based on it (i.e., copy the files as a "template") 

CleanShot 2023-01-14 at 08 51 57@2x

> You can also start a new astro project based on an existing GitHub repostory by passing a --template argument to the create astro command.

In reality, this does not seem to be the case. 

## How? 

CleanShot 2023-01-14 at 08 39 04@2x

it appears (as I suspected), the creation from an "existing Github repository" command/description only works with existing starter templates 

e.g., 

✅ this works: 

```bash
npm create astro@latest -- --template satnaing/astro-paper
```

❌ this does not work: 

```bash
npm create astro@latest -- --template ohansemmanuel/astrojs-ditch-the-runtime-react 
```

## Suggestion 
If the assumption above is correct i.e., this works for existing templates, then it's **perhaps** worth reconsidering rewording the sentence to: 

> You can also start a new astro project based on an existing **starter template** GitHub ~repostory~ **repository** by passing a --template argument to the create astro command.

CleanShot 2023-01-14 at 08 58 12@2x

delucis:

- New or updated content

#### Description

- Closes #2372

- #2372 revealed that using `create astro --template` fails if the repository being used does not have a `main` branch because it will try to copy that branch by default.

- This PR adds a note explaining how to set a specific branch. I made this advice fairly generic, because you might want to use a specific branch for a number of reasons, but if this seems _too_ generic, we could explain further.


src/pages/en/install/auto.mdx
```diff
@@ -110,6 +110,8 @@ You can also start a new astro project based on an existing GitHub repository by
   </Fragment>
 </PackageManagerTabs>
 
+By default this command will use the template repository’s `main` branch. To use a different branch name, pass it as part of the `--template` argument: `<github-username>/<github-repo>#<branch>`.
```
ohansemmanuel, line 113: nitpick: a comma after "By default ..." i.e., 

> By default,  this command will use the template repository’s ...

sarah11918:
Thanks for tackling this Chris, and trying to figure out what I'd prefer here:

- update the command example to include `#branch` so there's only ONE long command, but mention that branch is only "other than main"
- also have an example of the ASTRO templates themselves, which you can do just with `--template blog` vs `--template withastro/astro#blog` as these instructions would suggest (even though that's not the nice, all-purpose pattern).
- instead of a paragraph, show multiple commands depending on what you want to open? An Astro template? Any repository you want? (and if so, maybe the "branch" becomes *less* important to point out, because it ends up being important for *our* repository, but we have a nice way of doing it?)

I honestly don't know what I want here! But, non-blocking, and def NWTWWHB! Will approve to get this in if we can't immediately think of something better. But I don't *think* this is what I'd like long-term as our approach here.

sarah11918 (approved):

PR #2419, opened 2023-01-20
Use lists to improve legibility for images docs

pilcrowOnPaper:

- New or updated content

#### Description

- Closes (none)
- Made use of lists where appropriate (most content/sentences remain unchanged)


sarah11918:
Hi @pilcrowOnPaper! 

I do like that you've thought about how to make the content look better, but I'm quite against changing the initial lines with all the attributes in one line of text into a bulleted list. 

The point is that it's very easy for someone to see at a glance what the required properties are for that component, and I think stretching it out into a visually expanded list, with more info like instructions for use, loses that. As we always have to consider, this is going to be a page people come back to for a quick reference, and I don't think making that part more drawn-out and mixing the info together is helpful.

What could be interested in seeing, if you wanted to experiment, is something where the initial list of items stays the same, linked to their config options in the image integration guide, but the paragraphs below become bullet points? Or take a more appealing visual form?

I agree that the paragraphs themselves aren't very attractive, and perhaps *that* info could be better presented. But, I would really have to be convinced and outvoted that making a bulleted list for the attributes is better!  ;)

sarah11918:
I'll also add that changes like this that aren't *corrections* or additions of content to the docs, we usually discuss them first in Issues or in a thread in the docs channel on Discord so that we can all agree on something before a PR shows up. It would be great if we could start doing that for improvements you'd like to see in the docs, since especially as we're trying to get the v2 docs launched and out the door, these PRs are our "inbox" of stuff we have to get done! 

Improvements like this, we often take a lot more time discussing them and figuring out what we want, and Issues or Discord are just an easier place for the kind of back-and-forth we need.  :raised_hands:

pilcrowOnPaper:
Yeah sure, I'll continue the discussion on Discord. I'll remember that for next time. 

Personally, having the required attributes and the whys/value type next to it is much more helpful. And, I think it's still easy to see the required attributes in a glance, but that could be just me.

pilcrowOnPaper:
Hm it might be better if I rework this PR from scratch, since there're some things that could be improved. Do you mind if I close this one for now, and maybe open a new PR maybe next month or so (since I'm going to be busy in the next few weeks)?

sarah11918:
@pilcrowOnPaper Don't mind at all! And, there's work being done on the image component itself that will necessitate lots of changes to this page soon(ish?) anyway, so we should both be keeping an eye on that!  🙌

PR #2421, opened 2023-01-20
Document version of RSS package needed for new features

delucis:

- New or updated content

#### Description

- Adds a `<Since />` component to the new RSS docs to indicate the feature is new (in beta!)
- Adds a minimal note to the same docs showing the old behaviour for people using older versions. Maybe not needed? Excessively polite?
- Updates `<Since />` so we can use it in cases like this where we’re documenting changes to packages other than `astro`: `<Since v="2.0.0" pkg="@astrojs/rss" />`
- Updates `WRITING.md` to document these changes

This does slightly change the output of `<Since />` as it could now be talking about different packages:

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/WRITING.md
+++ b/WRITING.md
@@ -88,18 +88,15 @@ Sometimes it is helpful to add a small badge to some content to label or highlig
 
 Badges work best when they only contain a single word or — at a push — two words. Think of them as a tag or label for something, not a way to highlight longer passages of text.
 
-```md
----
-setup: |
-  import Badge from '~/components/Badge.astro';
----
+```mdx
+import Badge from '~/components/Badge.astro';
 
 <Badge>Nice</Badge>
 ```
 
 By default, the badge uses a muted colour scheme to blend in. It also has an accented variant that can be used if you need it to stand out more from the surrounding context:
 
-```md
+```mdx
 <Badge variant="accent">Wow!</Badge>
 ```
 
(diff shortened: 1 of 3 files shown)


src/pages/en/guides/rss.mdx
```diff
@@ -126,6 +128,13 @@ export async function get(context) {
 }
 ```
 
+:::note[Using an older version?]
+In versions `@astrojs/rss` before v2.1.0, pass your glob result straight to `items` without the `pagesGlobToRssItems` wrapper:
```
Yan-Thomas, line 132: I think you accidentally changed the order here.
```suggestion
In `@astrojs/rss` versions before v2.1.0, pass your glob result straight to `items` without the `pagesGlobToRssItems` wrapper:
```

Yan-Thomas (commented):
Overall LGTM! Just added a small typo fix suggestion.

src/pages/en/guides/rss.mdx
sarah11918: Above, we include the `()` for `pagesGlobToRssItems()`. We should choose one and be consistent throughout.  :smile:

delucis: Good catch — thank you! I meant to include an “of” in between there. Updated with the function parentheses too:

```suggestion
In versions of `@astrojs/rss` before v2.1.0, pass your glob result straight to `items` without the `pagesGlobToRssItems()` wrapper:
```

sarah11918:
Looks great, @delucis! Excellent work as always, especially in handling WRITING.md! Exactly what I would have wanted to see there.  :raised_hands:

delucis:
> Just checking that the example at the top of the page will also work in older versions? since it's relying on `get(context)` whereas the previous example given didn't?

Good question! I think that’s fine — `get` handlers have always had `context` I think — but will double check.

WRITING.md
```diff
@@ -110,47 +107,56 @@ By default, the badge uses a muted colour scheme to blend in. It also has an acc
...
 ```
 
-This will render the text “**Added in:** v1.0.0”.
+This will render the text “**Added in:** `astro@1.0.0`”.
 
 The advantages of using the component include:
 
 - “Added in” is automatically translated on pages in other languages.
-- The passed version is checked against the current Astro version and a “NEW” badge will be added automatically as long as the version is relatively recent.
+- The passed version is checked against the current Astro version and “NEW” or “BETA” badges will be added automatically based on data from npm.
```
delucis, line 130: Missed this mention of Astro where it could now be any package:

```suggestion
- The passed version is checked against the current package version and “NEW” or “BETA” badges will be added automatically based on data from npm.
```

sarah11918 (approved):

PR #2431, opened 2023-01-23
docs(svelte): Fix vitePreprocess import

vargasmesh:

#### Description

I was integrating Astro with Svelte by following https://docs.astro.build/pt-br/guides/integrations-guide/svelte/.

But when I used the code snippet from the `Intellisense for TypeScript`, it broke my application.

The problem that the docs say to import `vitePreprocess` from `@astrojs/svelte`, but the right import is:

```js
import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';
```

### DIFF
--- a/src/pages/en/guides/integrations-guide/svelte.mdx
+++ b/src/pages/en/guides/integrations-guide/svelte.mdx
@@ -140,7 +140,7 @@ If you're using a preprocessor like TypeScript or SCSS in your Svelte files, you
 
 ```js
 // svelte.config.js
-import { vitePreprocess } from '@astrojs/svelte';
+import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';
 
 export default {
 	preprocess: vitePreprocess(),


bluwy:
The docs seems to be updated for the beta, so it should work when you upgrade to `@astrojs/svelte` beta. Not quite sure how to handle this for now 😬

delucis:
Thanks for this @vargasmesh! `@astrojs/svelte` v2 is now published so this should be fixed. We _could_ add a disclaimer to [the Svelte README in the main `astro` repo](https://github.com/withastro/astro/blob/main/packages/integrations/svelte/README.md#intellisense-for-typescript) to help indicate this was added in v2. Something like:

```md
**Added in:** `@astrojs/svelte@2.0.0`
```

Would happily merge a PR like that! Will close this one now that the v2 release is out 🙌

vargasmesh:
Here we go @delucis: https://github.com/withastro/docs/pull/2451

I used the `Since` component as it is used in other pages. Thank you.

PR #2330, opened 2023-01-09
[upgrade guide] [i18nIgnore] Add upgrade to v2 guide + moving pieces

sarah11918:
Creates an "Upgrade v1 -> v2" guide, and makes all the necessary infrastructure changes:
- adds "Upgrade to v2" in the Start Here section of the nav bar
- creates a new `/guides/upgrade-to/v2.mdx` page
- renames the previous `migrate.mdx` file to `/guides/upgrade-to/v1.mdx` (this can eventually be deleted if we decide. "migrating to older versions?" link will go to a fixed GitHub commit version of this file so that it can be removed from docs)
- redirects `/en/migrate/` and older existing `/en/migrate/0.21.0` redirect to `/en/guides/upgrade-to/v2` in `netlify.toml` instead of an Astro page file.
- references this guide on the Getting Started docs landing page

Thank you @delucis for suggestions re: finer strategy points!

- [x ] TODO: Write the content!  :sweat_smile: 
(Will start by mostly faithfully reproducing Changelog content; update to better "guide" content incrementally.)

ALSO TODO:
- [x] Convert code samples to Docs (not GitHub) format
- [x] Decide/consitenc-ize how we refer to Astro versions
- ~~[x] reference new guide on main Getting Started page ~~ (Will make separate PR that's not ignored by translators)
- [x] decide for sure if this belongs in root, or under "guides"
- [x] if under guides, then update `netlify.toml` redirects
- [x] decide whether to link to old v1 upgrade guide on GitHub or in docs (currently GitHub)

PR #2436, opened 2023-01-23
Add v2 callout to docs landing page

delucis:

- New or updated content

#### Description

- Add a call out to the Getting Started page with links to the v2 announcement blog post (not published yet) and the upgrade guide (added in #2330)

- Should not be merged before those two are! And we’ll want to update branch at that point so that the link checker passes

Astro version 2 is here! docs callout


sarah11918 (approved):
Just what we need!  🙌

dreyfus92 (approved):
LGTM !

PR #2451, opened 2023-01-24
docs(svelte): Add Since tag to Intellisense for TypeScript

vargasmesh:

#### Description

Prior v2, vitePreprocess was imported from @sveltejs/vite-plugin-svelte.

See https://github.com/withastro/docs/pull/2431 for more information


delucis:
Ah, thank you @vargasmesh! Unfortunately I wasn’t clear in my comment in #2431. This page is automatically generated from [the Svelte integration README in the `astro` repo](https://github.com/withastro/astro/blob/main/packages/integrations/svelte/README.md?rgh-link-date=2023-01-24T18%3A05%3A44Z#intellisense-for-typescript), so we need to make this change there (and can’t use `<Since />` because of this, because the README is shown in docs but also on GitHub, npm etc.)

Happy to merge this so we can welcome you as a docs contributor, but we’ll need to update that source README so this doesn’t get overwritten by our nightly CI.

vargasmesh:
Now it is clear @delucis.  Can you check if it is something like this https://github.com/withastro/astro/pull/5966?

PR #2465, opened 2023-01-25
Use const instead of let

morinokami:

#### Description

- I thought it would be better to use `const` here, since it is unlikely that you will reassign other values to `allPosts`.


delucis (approved):
Thanks @morinokami! I think that's correct 👍

PR #2467, opened 2023-01-25

the-dijkstra:
This PR Fixes a small issue with the command `echo ".astro" >> .gitignore` as it will append ".astro" to ".gitignore"
file without accounting for a new line character. which might break the `.gitignore` file.

example running the command on a `.gitignore` file like this
```gitignore
.vscode
```

will endup with something like this:
```gitignore
.vscode.astro
```


sarah11918:
Looks great! Thank you @the-dijkstra! :raised_hands:

PR #2470, opened 2023-01-25
fix pagesGlobToRssItems example in v2 upgrade guide

defeated:

#### Description

- What does this PR change? The pagesGlobToRssItems function returns a Promise, which causes an error ("[RSS] Invalid or missing options") unless the promise is resolved.


Jutanium:
Thanks for the quick fix! I double-checked and made sure the tutorial [doesn't have the same issue.](https://docs.astro.build/en/tutorial/5-astro-api/4/)

PR #2485, opened 2023-01-27
Clarify Markdown layout props in Markdown guide

### LINKED ISSUE #2429: Reference to headings being in Astro.props not evident in docs
Happy to be proved wrong! But I was unable to find this information, until I happened upon the comment below:

___
So to confirm everything here:
- Today, you can access headings from MD and MDX using `Astro.props.headings`. Update to `astro@latest` and `@astrojs/mdx@latest` if that isn't the case!
- We should update `@examples/docs` to use `headings` properly

@danilopolani Let me know if that's correct!

_Originally posted by @bholmesdev in https://github.com/withastro/astro/issues/4218#issuecomment-1213552093_

Jutanium:

- New or updated content

#### Description

- Closes #2429 
- Change wording so it's clear that `Astro.props.frontmatter` is not the only prop passed to the layout
- Link directly to the list of props in a CTA at the end of the section

Note: having three link-outs might feel clunky. Open to suggestions for reformatting the end of this section. Three 📚? Replace the current 📚 Learn More link with the link to the properties? Okay as is? Let me know.


src/pages/en/guides/markdown-content.mdx
```diff
@@ -133,7 +133,7 @@ description: Find out what makes Astro awesome!
 This is a post written in Markdown.
 ```
 
-Frontmatter properties are then available to the layout through `Astro.props`:
+Metadata are then available to the layout through `Astro.props`. For example, you can access frontmatter properties through `Astro.props.frontmatter`:
```
sarah11918, line 136:
```suggestion
[Specific properties are then available to the layout component](/en/core-concepts/layouts/#markdown-layout-props)  through `Astro.props`. For example, you can access frontmatter properties through `Astro.props.frontmatter`:
```
What about adding a link here instead of another callout, so we can remove the bottom line?

Not entirely opposed to metadata, though in this sentence we'd have some people arguing singular, some plural.  :sweat_smile: I think something like "Specific properties" might be even more useful, since the properties available to a layout component *are* different than properties exported when you're importing a Markdown component otherwise. (We just had this come up re: switching to content collections, which means changing from using a layout property to explicitly importing.)

Jutanium: Love this change, thanks!

src/pages/en/guides/markdown-content.mdx
```diff
@@ -151,8 +151,9 @@ const {frontmatter} = Astro.props;
 </html>
 ```
 
-You can also [style your Markdown](/en/guides/styling/#markdown-styling) in your layout component.
+[See all properties](/en/core-concepts/layouts/#markdown-layout-props) available through `Astro.props` in a Markdown layout.
```
Jutanium, line 154:
```suggestion

```

sarah11918 (approved):
It's Dan's Friday Merge Day, and we're all just living in it!  Thanks Dan!

PR #2487, opened 2023-01-27
Update Render deploy guide to account for Astro 2.0 Node.js version requirement

HiDeoo:

- New or updated content

#### Description

This PR is a follow-up of https://github.com/withastro/astro/issues/6004 and this [Discord support thread](https://discord.com/channels/830184174198718474/1068446497353498666).

It is currently not possible to deploy a project using Astro 2.0 to Render by blindly following the guide due to the fact that Astro 2.0 requires at least Node.js 16.12.0 and Render [using by default Node.js version 14](https://render.com/docs/node-version).

The proposed change is *heavily* inspired by the existing [Cloudflare Pages deploy guide](https://docs.astro.build/en/guides/deploy/cloudflare/) which has the same issue that is solved using an environment variable specifying the Node.js version added after clicking on an `Advanced` button in the UI, which is the exact same process on Render. Other solutions are also suggested with links to the relevant Render documentation describing how to use them.

### DIFF
--- a/src/pages/en/guides/deploy/render.mdx
+++ b/src/pages/en/guides/deploy/render.mdx
@@ -15,4 +15,5 @@ You can deploy your Astro project to [Render](https://render.com/), a service to
 4. Give your website a name, select the branch and specify the build command and publish directory
    - **build command:** `npm run build`
    - **publish directory:** `dist`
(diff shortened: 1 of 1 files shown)


Jutanium:
Thanks for this - this is a necessary disclaimer, and thanks for using our existing format. (Interesting that both platforms hide this behind an **advanced** button!)

PR #2492, opened 2023-01-27
Updates Node Version requirements in GitLab deploy

Firq-ow:

- Minor content fix (wrong version in gitlab yaml)

#### Description

As of Astro Version 2.0.0, a node version >= 16 is required, which means that `node:lts` is better suited than `node:14` in the deploy file `.gitlab-ci.yml` for GitLab. This ensures that people who copy the file from the docs won't end up with a broken CI/CD pipeline.


Jutanium:
Thank you! And welcome to team docs!

PR #2495, opened 2023-01-28
Fix error in Image example

BryceRussell:

#### Description

- Fix `<Image>` example: `aspectRatio` prop requires a `width` or `height` prop

### DIFF
--- a/src/pages/en/guides/integrations-guide/image.mdx
+++ b/src/pages/en/guides/integrations-guide/image.mdx
@@ -549,7 +549,7 @@ import heroImage from '../assets/hero.png';
 <Image src={heroImage} width={300} height={600} alt="descriptive text" />
 
 // cropping to a specific aspect ratio and converting to an avif format
-<Image src={heroImage} aspectRatio="16:9" format="avif" alt="descriptive text" />
+<Image src={heroImage} width={300} aspectRatio="16:9" format="avif" alt="descriptive text" />
 
 // image imports can also be inlined directly
 <Image src={import('../assets/hero.png')} alt="descriptive text" />


MoustaphaDev (approved):

MoustaphaDev:
Hey @BryceRussell, thanks for wanting to tackle this!

However I believe this might not be needed, as it seems to me that the `Image` component used with an `aspectRatio` property only requires to have `height` or `width` when the image is remote. [See](https://github.com/withastro/astro/blob/main/packages/integrations/image/src/lib/get-image.ts#L65)

If local, the metadata of the image are fetched from disk. [See](https://github.com/withastro/astro/blob/main/packages/integrations/image/src/vite-plugin-astro-image.ts#L50)

I think changing that example could make more people unaware of the possibility of omitting dimensions, although the docs clearly state that *dimensions are optional for local images, the original image size will be used if not provided*

BryceRussell:
I tested this locally with a 500x500px image

If I only specify an `aspectRatio` it does not resize the image to fit ratio as mentioned in the example
```jsx
<Image src={image} aspectRatio="16:9" alt="" />
```
```html
<img ... width="500" height="500">
```

However If I add a `width` or `height` then it will resize the image to fit the ratio
```jsx
<Image src={image} width={300} aspectRatio="16:9" alt=""/>
```
```html
<img ... width="300" height="169">
```

The [docs for `aspectRatio`](https://docs.astro.build/en/guides/integrations-guide/image/#aspectratio) also mentions this: 

> Combine with either `width` or `height` to automatically calculate and crop the other dimension.

Jutanium:
@MoustaphaDev According to the code you cited, if `src` is provided:
- If both `width` and `height` are provided, `resolveSize` returns and `aspectRatio` is ignored
- If neither `width` and `height` are provided, `resolveSize` throws and `aspectRatio` is ignored
So this change does seem necessary

MoustaphaDev:
Thanks for testing this @BryceRussell, I'll take time to reevaluate my initial assumption

Princesseuh:
> @Princesseuh Is this intended behavior for the image component: whether the `src` is local or remote, `aspectRatio` should do nothing unless a width or height is manually provided?

Yep, `aspectRatio` is only used to calculate `width` (if `height` is missing or `height` (if `width` is missing). If both are provided already, it's unnecessary.

`aspectRatio` by itself does not control cropping / resizing.

MoustaphaDev (approved):

Jutanium:
@BryceRussell, thanks so much for this correction - this is an autogenerated file based on [this integration Readme](https://github.com/withastro/astro/blob/main/packages/integrations/image/README.md). Would you be willing to open a PR there?

BryceRussell:
I opened a PR here: https://github.com/withastro/astro/pull/6019

PR #2500, opened 2023-01-29
fix: delete `(` that should not exist in mjs config intellisense

kecrily:

#### Description

delete `(` that should not exist in mjs config intelliSense


Jutanium:
Great catch, thank you!

PR #2501, opened 2023-01-29
Content Collections Migration Guide

Jutanium:

- New or updated content

#### Description

- Adds a step-by-step guide for moving from file-based MD/MDX routing to content collections, using the tutorial project as an example.
- This would be a good candidate for a Recipe once we set up that section!

[Netlify Preview](https://deploy-preview-2501--astro-docs-2.netlify.app/en/guides/content-collections/#migrating-from-file-based-routing)


src/pages/en/guides/content-collections.mdx
```diff
@@ -396,6 +396,122 @@ const { Content } = await entry.render();
...
+```md title="src/content/post-1.md" del={2}
+---
+layout: ../../layouts/MarkdownPostLayout.astro
+title: 'My First Blog Post'
+pubDate: 2022-07-01
+...
+---
+```
+
+5. To fetch your content, use [`getCollection`](/en/reference/api-reference/#getcollection) or [`getEntryBySlug`](/en/reference/api-reference/#getentrybyslug) instead of `Astro.glob` or `import` statements. For example, our unique tags logic becomes:
```
sarah11918, line 466:
```suggestion
5. Replace `Astro.glob()` with [`getCollection`](/en/reference/api-reference/#getcollection) to fetch content and metadata from your Markdown files. You will also need to update references to the returned data object, since you will now find your frontmatter values in `data.frontmatter`.

For example, our unique tags logic becomes:
```
1. I think this puts the instruction clearer, up front. I might also hit that this is on "other pages/components or in other files" you might have around your project. e.g. a blog index page, a dynamic page for generating a page for each blog post tag, rss feed. This sets us up nicely for where we're headed re: updating a few different files.

2. I'd also include some kind of reference to `data.frontmatter` here, above the example. (Not in love with wording, just spewed quickly to get it out!)

3. I also think they might not even be using individual file imports anywhere, so I'd hit getCollection (and maybe include getEntryBySlug? but if so, not interwoven with the one we KNOW they will need).

> Similarly, any individual Markdown or MDX file imports should be replaced with [`getEntryBySlug`](/en/reference/api-reference/#getentrybyslug).

THIS could become the "after the example" note, instead of the one about `data.frontmatter`

sarah11918 (commented):
Excellent kick off to this, Dan! Love so much about it!  :partying_face: 

I left some notes to get started!  :raised_hands:

src/pages/en/guides/content-collections.mdx
Jutanium: Thanks for all of your comments - should be good to go!

Yan-Thomas (approved):
Nice work @Jutanium, LGTM!

src/pages/en/guides/content-collections.mdx
```diff
@@ -396,6 +396,127 @@ const { Content } = await entry.render();
...
+export const collections = {
+  posts: postsCollection,
+};
+```
+
+:::tip
+If your editor doesn't recognize `astro:content`, make sure you're on the latest version of Astro and try restarting the dev server. 
+:::
+
+5. To [generate routes](/en/guides/content-collections/#generating-routes-from-content), create a `.astro` page with a [dynamic route](/en/core-concepts/routing/#dynamic-routes) in `src/pages/posts` to [query and render your content](#querying-collections). Our tutorial blog uses the static output mode, so we'll use `getStaticPaths()` to create a route for each post:
```
sarah11918, line 435:
```suggestion

:::note
Inside a collection, Markdown and MDX files no longer automatically create pages using Astro's [file-based routing inside `src/pages/`](/en/guides/markdown-content/#file-based-routing). You will need to [generate routes from your collections](/en/guides/content-collections/#generating-routes-from-content) yourself.
:::

5. Create a `src/pages/posts/[...slug].astro`. This page will use [dynamic routing](/en/core-concepts/routing/#dynamic-routes) to generate a page for each collection entry. 

This page will also need to [query your collection](#querying-collections) to fetch page slugs and make the page content available to each route.

Render your post `<Content />` within the layout for your Markdown or MDX pages, since they can no longer rely on the `layout` frontmatter property to specify their own layout.
```
I'm not sold on this edit, but the original felt way too dense for one step. We felt very much in "draw the rest of the owl" territory here!  :sweat_smile: 

So, I tried to emphasize routing as a special thing, and make the other points as distinct follow-ups.

src/pages/en/guides/content-collections.mdx
```diff
@@ -396,6 +396,127 @@ const { Content } = await entry.render();
...
+
+const { entry } = Astro.props;
+const { Content } = await entry.render();
+---
+<MarkdownPostLayout frontmatter={entry.data}>
+  <Content />
+</MarkdownPostLayout>
+```
+
+6. Since we wrapped the content in a layout in the dynamic route, remove the `layout` definition in each post's frontmatter.
```
sarah11918, line 457:
```suggestion
6. Remove the `layout` definition in each individual post's frontmatter. This `layout` property is only available to Markdown and MDX files within `src/pages/`. Instead, a layout is specified for all of your generated routes at once.
```

src/pages/en/guides/content-collections.mdx
```diff
@@ -396,6 +396,127 @@ const { Content } = await entry.render();
...
+layout: ../../layouts/MarkdownPostLayout.astro
+title: 'My First Blog Post'
+pubDate: 2022-07-01
+...
+---
+```
+
+7. Replace `Astro.glob()` with [`getCollection()`](/en/reference/api-reference/#getcollection) to fetch content and metadata from your Markdown files. You will also need to update references to the returned post object, since you will now find your frontmatter values on the `data` property.
+
+For example, our unique tags logic becomes:
```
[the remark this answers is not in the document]
sarah11918: Since this example is very close to the dynamic route page from before, maybe before this we could show the "blog index" page converted? Something where we just use `getCollection()` in order to create the linked list of posts? I don't think we need to remove this one, since it feels like an extra "Hey, think of all the places you might have used this!" But, I think maybe starting with an example of blog index makese sense, both for variety and for an example of something different.

sarah11918 (commented):
Would love it if you could take a look at these suggestions Dan!

src/pages/en/guides/content-collections.mdx
```diff
@@ -396,6 +396,127 @@ const { Content } = await entry.render();
 <Content />
 ```
 
+## Migrating from File-Based Routing 
+If your existing site uses [file-based routing](/en/core-concepts/astro-pages/#markdownmdx-pages) with Markdown files in the `src/pages` folder, you may want to migrate to content collections. This gives you the benefit of type-safe frontmatter, but requires you to set up routing yourself. This guide uses the result of the [Build a Blog tutorial](/en/tutorial/0-introduction/) as an example starting point.
```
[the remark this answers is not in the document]
Jutanium: If we leave this guide here, I'm onboard with this change, as we discussed the benefits of content collections already. But if we want to move this into a self-contained recipe, providing some motivation seems worthwhile.

src/pages/en/guides/content-collections.mdx
Jutanium: It isn't true that the `layout` property functionality isn't _available_ with content collections. It will still work: when you call `entry.render()`, if the entry has a layout property, it will render it with a layout.

So it's not that you _can't_ do it, but since all of the posts specify the same layout, we might as well "pull it out" and wrap it ourselves.

Jutanium: Made an edit that hopefully reflects your comments!

src/pages/en/guides/content-collections.mdx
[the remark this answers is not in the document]
Jutanium: Added all the pages!

src/pages/en/guides/content-collections.mdx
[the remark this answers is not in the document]
sarah11918: Agreed! For here, though, I say let's remove this info, (in part because I want to avoid thinking about exactly how to position this and what kind of caveat to include in this introduction) and we'll work it back in when it goes somewhere else!  :raised_hands:

src/pages/en/guides/content-collections.mdx
```diff
@@ -396,6 +396,173 @@ const { Content } = await entry.render();
...
+
+    const { entry } = Astro.props;
+    const { Content } = await entry.render();
+    ---
+    <MarkdownPostLayout frontmatter={entry.data}>
+      <Content />
+    </MarkdownPostLayout>
+    ```
+
+6. Now that you wrapped your content in a layout, remove the `layout` definition in each individual post's frontmatter.
```
sarah11918, line 467:
```suggestion
6. Remove the `layout` definition in each individual post's frontmatter. Your content is now wrapped in a layout when rendered, and this property is no longer needed.
```
Agreed with the edits to this section Dan! Tidy this up a bit if you need to, but I did rephrase to start with the action verb/instruction. (I didn't think too hard here, nor did I refer to the tutorial blog project. This might be intuitive enough to not need to say that here.)

PR #2503, opened 2023-01-30
Note Code component Shiki edge case

bluwy:

- New or updated content

#### Description

Close https://github.com/withastro/astro/issues/5912

We may fix this in the future, but it's slightly complex at the moment, so noting a workaround for now.

### DIFF
--- a/src/pages/en/reference/api-reference.mdx
+++ b/src/pages/en/reference/api-reference.mdx
@@ -873,7 +873,13 @@ import { Code } from 'astro/components';
 
 This component provides syntax highlighting for code blocks at build time (no client-side JavaScript included). The component is powered internally by Shiki and it supports all popular [themes](https://github.com/shikijs/shiki/blob/main/docs/themes.md) and [languages](https://github.com/shikijs/shiki/blob/main/docs/languages.md). Plus, you can add your custom themes and languages by passing them to `theme` and `lang` respectively.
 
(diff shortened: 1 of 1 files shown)


src/pages/en/reference/api-reference.mdx
```diff
@@ -873,6 +873,8 @@ import { Code } from 'astro/components';
 
 This component provides syntax highlighting for code blocks at build time (no client-side JavaScript included). The component is powered internally by Shiki and it supports all popular [themes](https://github.com/shikijs/shiki/blob/main/docs/themes.md) and [languages](https://github.com/shikijs/shiki/blob/main/docs/languages.md). Plus, you can add your custom themes and languages by passing them to `theme` and `lang` respectively.
 
+When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like `pnpm`, you may also need to install Shiki so it works during build: `pnpm install shiki`.
```
Yan-Thomas, line 876: Thanks, @bluwy! I feel like we tend to separate even small commands into their own line in docs (with the additional advantage of showing a terminal-like UI for it). Also, since this is an edge case, I'd put it inside a note aside.

````suggestion
:::note
When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like `pnpm`, you may also need to install Shiki so it works during build:

```bash
pnpm install shiki
```
:::
````

bluwy: Good to know! The changes looks good too.

Yan-Thomas (approved):

PR #2504, opened 2023-01-30
Use double quotes to enclose html attributes for consistency

morinokami:

#### Description

- Although this is very trivial, while translating testing.mdx, I found that one of html attributes in the code sample was enclosed in single quotes. Since there seems to be no particular reason to use single quotes here, I have changed them to double quotes for consistency.


Yan-Thomas:
Nice catch @morinokami, thanks for spotting and fixing this! 🙌

PR #2507, opened 2023-01-30
i18n(zh-cn): Add translations for `docsearch`

manchan4869:

- Translated content

#### Description

- Add translations


delucis (approved):
Thanks @manchan4869!

PR #2510, opened 2023-01-30
Included common gotcha on script and style import

pollomarzo:
It can be hard to track down the source of a `<script>` or  `<style>` tag if it's included from an unused import. Also, it's important to point it out, as it might be expected that "non-explicitly used" JS isn't in the final build.

- New or updated content

#### Description

- Added new section in "Common Gotchas" in the "Troubleshooting" page, to highlight that `<script>` and `<style>` tags are bundled depending on component import, not component use.


Jutanium:
Thanks for this addition! I want to pin down the behavior as precisely as possible: does the script/style inclusion happen when the component is _imported_ or when it is _used_ within a template? [In this test](https://stackblitz.com/edit/github-vlpypp?file=src/pages/index.astro) the script and style aren't included from the import alone. If that's the case, we can adjust the wording, something like:

> Astro's build process works on the module graph: this means that once a component is ~~imported~~ used within a template, its `<script>` and `<style>` tags are processed, optimized, and bundled, whether it is ultimately rendered or not. ~~(or used at all!)~~

pollomarzo:
> Thanks for this addition! I want to pin down the behavior as precisely as possible: does the script/style inclusion happen when the component is _imported_ or when it is _used_ within a template? [In this test](https://stackblitz.com/edit/github-vlpypp?file=src/pages/index.astro) the script and style aren't included from the import alone. If that's the case, we can adjust the wording, something like:
> 
> > Astro's build process works on the module graph: this means that once a component is ~imported~ used within a template, its `<script>` and `<style>` tags are processed, optimized, and bundled, whether it is ultimately rendered or not. ~(or used at all!)~

You're right! I should've checked more thoroughly. I encountered this because I was trying to conditionally render on a component: do you think i should include the "conditional rendering" phrase in there, so that it's easier to find? 

BTW, does this mean astro removes unused imports before the bundling step?

pollomarzo:
As i mentioned in the commit, I'm unsure about the wording of "rendered", but i don't know what the wording was in other parts of the docs.
should i also change

> whether it ~is rendered~ appears in the output or not

Jutanium:
@pollomarzo I like that change. And maybe "unexpected" captures the behavior better than "unwanted"?

src/pages/en/guides/troubleshooting.mdx
```diff
@@ -137,6 +137,10 @@ nodeLinker: "node-modules"
 
 In Astro, using a `<head>` tag works like any other HTML tag: it does not get moved to the top of the page or merged with the existing `<head>`. Because of this, you usually only want to include one `<head>` tag throughout a page. We recommend writing that single `<head>` and its contents in a [layout component](/en/core-concepts/layouts/).
 
+### An unwanted `<script>` or `<style>` is included
+
```
Jutanium, line 141:
```suggestion
When working with [conditionally displayed](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) components, you may notice a component's `<script>` or `<style>` tags being included even if that component doesn't appear in the final output.
```

Adding a sentence to help describe the error before explaining it

Jutanium: We use the both the term ["conditionally displayed"](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) and ["conditionally rendered"](https://docs.astro.build/en/tutorial/2-pages/3/#conditionally-render-elements) elsewhere in the docs, so I chose the term that matches where I wanted to link!

sarah11918: This might make it sound like the problem occurs *only* when conditionally rendered (and not, for example, just any unused import, like something you never bothered to remove when you decided not to render the component.)

I might choose something like:

> You may notice an imported component's `<script>` or `<style>` tags included in your HTML source even if that component doesn't appear in the final output. For example, this will occur with [conditionally rendered](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) components that are not displayed.

src/pages/en/guides/troubleshooting.mdx
```diff
@@ -137,6 +137,10 @@ nodeLinker: "node-modules"
...
+Astro's build process works on the module graph: this means that once a component is used, its `<script>` and `<style>` tags are processed, optimized, and bundled, whether it is rendered or not. This does not apply if the `is:inline` directive is applied.
```
sarah11918, line 142:
```suggestion
Astro's build process works on the module graph: once a component is evaluated, its `<script>` and `<style>` tags are processed, optimized, and bundled, whether it is rendered or not. (Note: This does not apply to scripts when the `is:inline` directive is applied.)
```
1. the colon takes the place of needing to say "this means that" so we can shorten this!
2. I think "used" isn't quite descriptive enough. And, I think "used" could be interpreted as another way of saying "rendered" (e.g. I didn't *use* the component, so why do I see it's stuff??) so I'd like to avoid any chance of confusion. Can we think of a better word? I threw in "evaluated" as a placeholder, but what do we think is the best, descriptive verb here? "read" ? ... ?

sarah11918 (approved):
Thanks for this super helpful contribution @pollomarzo! :rocket:

I agree with Dan's suggestion, and gave my own interpretation of it. What do you think of the suggestions here?

sarah11918 (changes requested):
Do not approve! lol. Hit the wrong button!  :woman_facepalming:

src/pages/en/guides/troubleshooting.mdx
Jutanium: See my earlier comment - I don’t think the problem happens with any unused import, you have to use it in some way to trigger the issue!

sarah11918: Has this been fixed then, because that for sure *used* to be the case??

src/pages/en/guides/troubleshooting.mdx
pollomarzo: 1. agreed!
2. since the wording is not 100% obvious to you guys either, how about going for "evaluated" adding a simple `(i.e. appears after the code fence)`

pollomarzo:
sorry for taking this out of the comment threads, but multiple threads on the same two lines were starting to get confusing (how do you guys deal with it all the time? making my head spin!)
i ended up taking the middle road:

- moved from "unwanted" to "unexpected"
- went for "evaluated" the first time and "appears in the final output" the second time. I'm still not sure about these but i can't really find any better ways, always taking suggestions!
- included "conditionally rendered" as an example since i want to help someone else who might be trying to find it the same way i looked for it

```
Astro's build process works on the module graph: once a component is evaluated, its `<script>` and `<style>` tags are processed, optimized, and bundled, whether it appears in the final output or not (e.g. when [conditionally rendered](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) components are not displayed). This does not apply to scripts when the `is:inline` directive is applied.
```

Jutanium:
I like this, except for the word "evaluated" because of its meaning with expressions:
```{ false && <MyComponent/> } ```
Here, `<MyComponent/>` is never _evaluated_ - the `&&` is "short-circuited" by `false`. But because it is included in the template, it triggers the script/style inclusions.

I agree that "used" isn't descriptive enough - how about `included in the template`?

**Comment by pollomarzo** [comment]
> I like this, except for the word "evaluated" because of its meaning with expressions: `{ false && <MyComponent/> } ` Here, `<MyComponent/>` is never _evaluated_ - the `&&` is "short-circuited" by `false`. But because it is included in the template, it triggers the script/style inclusions.

You're so right! I should've thought about it

> I agree that "used" isn't descriptive enough - how about `included in the template`?

I think it's better than my initial attempt `appears after the code fence`!

**Comment by pollomarzo** [comment]
@sarah11918 do we have your blessing? :)
i hope it's ok to @ people! first PR!

**Comment by Jutanium** [comment]
Thank you so much for all of your work on this! One more thing and we should be good to go - both Sarah and I wanted some sort of intro sentence to explain the thing that you'd _notice_. Do you mind including Sarah's intro sentence, and then removing the later "conditionally rendered" parenthetical in the second paragraph?

> You may notice an imported component's <script> or <style> tags included in your HTML source even if that component doesn't appear in the final output. For example, this will occur with [conditionally rendered](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) components that are not displayed.

File: src/pages/en/guides/troubleshooting.mdx [hunk 3]
```diff
@@ -137,6 +137,10 @@ nodeLinker: "node-modules"
 
 In Astro, using a `<head>` tag works like any other HTML tag: it does not get moved to the top of the page or merged with the existing `<head>`. Because of this, you usually only want to include one `<head>` tag throughout a page. We recommend writing that single `<head>` and its contents in a [layout component](/en/core-concepts/layouts/).
 
+### An unexpected `<script>` or `<style>` is included
+
+You may notice an imported component's `<script>` or `<style>` tags included in your HTML source even if that component doesn't appear in the final output. For example, this will occur with [conditionally rendered](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) components that are not displayed. Astro's build process works on the module graph: once a component is included in the template, its `<script>` and `<style>` tags are processed, optimized, and bundled, whether it appears in the final output or not. This does not apply to scripts when the `is:inline` directive is applied.
```
Jutanium, line 142:
```suggestion
You may notice an imported component's `<script>` or `<style>` tags included in your HTML source even if that component doesn't appear in the final output. For example, this will occur with [conditionally rendered](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) components that are not displayed. 

Astro's build process works on the module graph: once a component is included in the template, its `<script>` and `<style>` tags are processed, optimized, and bundled, whether it appears in the final output or not. This does not apply to scripts when the `is:inline` directive is applied.
```
Just added a paragraph break! LGTM!

Jutanium (approved):

src/pages/en/guides/troubleshooting.mdx
sarah11918, line 142:
```suggestion
You may notice an imported component's `<script>` or `<style>` tags included in your HTML source even if that component doesn't appear in the final output. For example, this will occur with [conditionally rendered](/en/core-concepts/astro-components/#dynamic-html) components that are not displayed. Astro's build process works on the module graph: once a component is included in the template, its `<script>` and `<style>` tags are processed, optimized, and bundled, whether it appears in the final output or not. This does not apply to scripts when the `is:inline` directive is applied.
```

src/pages/en/guides/troubleshooting.mdx
```diff
@@ -137,6 +137,12 @@ nodeLinker: "node-modules"
 
 In Astro, using a `<head>` tag works like any other HTML tag: it does not get moved to the top of the page or merged with the existing `<head>`. Because of this, you usually only want to include one `<head>` tag throughout a page. We recommend writing that single `<head>` and its contents in a [layout component](/en/core-concepts/layouts/).
 
+### An unexpected `<script>` or `<style>` is included
+
+You may notice an imported component's `<script>` or `<style>` tags included in your HTML source even if that component doesn't appear in the final output. For example, this will occur with [conditionally rendered](https://docs.astro.build/en/core-concepts/astro-components/#dynamic-html) components that are not displayed.
```
sarah11918:
```suggestion
You may notice an imported component's `<script>` or `<style>` tags included in your HTML source even if that component doesn't appear in the final output. For example, this will occur with [conditionally rendered](/en/core-concepts/astro-components/#dynamic-html) components that are not displayed. 
```
pass link check

sarah11918 (approved):
Yay! Thanks for sticking with us, and welcome to Team Docs @pollomarzo!  🥳

sarah11918:
PS - It's great to @ people! I'm sorry, I was just on a call. Made it in here as soon as I could!

PR #2511, opened 2023-01-30
Add `id` attribute explanation.

jazzypants1989:
I was curious about what the id property was, because I was wondering why I would want to over-write it with my own custom slugs.

To clarify, I think using the filename makes a lot of sense and it was my guess, but the attribute's name "id" doesn't make that immediately clear.

Sorry if I did anything wrong. :)

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 


src/pages/en/guides/content-collections.mdx
```diff
@@ -234,7 +234,7 @@ defineCollection({
 
 ### Defining custom slugs
 
-Every content entry generates a URL-friendly `slug` property from its file `id`. The slug is used to query the entry directly from your collection. It is also useful when creating new pages and URLs from your content.
+Every content entry generates a URL-friendly `slug` property from its file `id`. The file `id` is simply the filename including its extension (`.md` or `.mdx`), but the extension is stripped for the slug. The slug is used to query the entry directly from your collection. It is also useful when creating new pages and URLs from your content.
```
sarah11918, line 237:
```suggestion
Every content entry generates a URL-friendly `slug` property from its [file `id`](/en/reference/api-reference/#id). The slug is used to query the entry directly from your collection. It is also useful when creating new pages and URLs from your content.
```
Rather than lengthen this paragraph, I think a good immediate measure is to link this to the definition of the propety itself. This reference page I think gives the clearest illustration of the similarity/difference between `slug` and `id`, and avoids extra words for those who don't need it (e.g. are revisiting the page to quickly remind themselves how to do something). Also, when we have a good description of something elsewhere, we do try to link so that if something changes, we only have to update in one place.

Do you think this solves the immediate problem of making it easier to understand what the file `id` is?

jazzypants1989: Yes, this makes perfect sense! I didn't know that API reference page existed or I would have done the same myself. Thank you so much for being awesome. :)

PR #2516, opened 2023-01-31
Remove "Save" button reference in GitHub Pages deploy guide

HiDeoo:

- New or updated content

#### Description

While setting up a repro for an issue and following the [GitHub Pages deploy guide](https://docs.astro.build/en/guides/deploy/github/), I noticed that step **4** reference a **Save** button that I couldn't find for a few seconds.

It turns out GitHub now automatically saves this change when selecting **GitHub Actions** as the **Source** of the site. Here is a video demonstrating the new flow:

https://user-images.githubusercontent.com/494699/215743743-a406eeaf-7351-4b44-b43f-2104b6b56b73.mp4


sarah11918:
Thanks for updating, @HiDeoo, and this video was super-helpful! Someone's in the running for Docs PR writer of the week... :partying_face:

PR #2521, opened 2023-01-31
Correct usage of `getImage` function 

MoustaphaDev:

- New or updated content

#### Description

When providing a string to the `getImage` function, it is taken as a remote image. To import local images we need to import them.

`height` and `width` are required too (or `aspectRatio` if one is missing)

### DIFF
--- a/src/pages/en/guides/integrations-guide/image.mdx
+++ b/src/pages/en/guides/integrations-guide/image.mdx
@@ -451,7 +451,11 @@ This can be helpful if you need to add preload links to a page's `<head>`.
 ---
 import { getImage } from '@astrojs/image';
 
-const { src } = await getImage({src: '../assets/hero.png'});
+const { src } = await getImage({
+    src: import('../assets/hero.png'),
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks @MoustaphaDev! (Firstly, this PR will have to be made to the package README at withastro/astro, but while I'm here...) 

So, I'm also trying to reconcile this with what is further on the page:

We say:

> This helper takes in an object with the same properties as the `<Image />` component

... which makes me think that the properties we are required to provide are the same ones as we'd be required to pass into an `<Image />` component for a local image in your project `src/`... which would only be `src` and `alt`. So, I think this means there's some change to the surrounding description necessary, too?

Also, we should absolutely be including an `alt` property here!

So, this PR needs to happen to the main repo, but I also think it's more than just the code change here?

MoustaphaDev:
Thanks for the heads up @sarah11918, you're absolutely right, `src` and `alt` are the only required props for local images. Will make those changes in the superseding PR

PR #2530, opened 2023-02-01
i18n(zh-cn): fix wrong translation

liruifengv:


delucis (approved):
Thanks for fixing this @liruifengv!

PR #2533, opened 2023-02-01
Fix `importsNotUsedAsValues` error in src/i18n

morinokami:

#### Description

- Fix importsNotUsedAsValues error in i18n/translation-checkers.ts and i18n/util.ts.

### DIFF
--- a/src/i18n/translation-checkers.ts
+++ b/src/i18n/translation-checkers.ts
@@ -1,7 +1,7 @@
 import type { ModalTranslations } from '@docsearch/react';
 import enNav from './en/nav';
 import type enUI from './en/ui';
-import languages from './languages';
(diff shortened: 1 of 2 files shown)


delucis (commented):
Thanks for this @morinokami! One thing I’m wondering is why this wasn’t caught by our CI running `astro check` 🤔

morinokami:
@delucis Maybe it's because .ts files are not checked by `astro check`?

Screen Shot 2023-02-01 at 23 13 34

delucis:
Ahhh, yes! We only run `astro check` not also `tsc`. Makes sense. I’ll merge this then. It would be good at some point to also run `tsc` to type check other files if we don’t have too many errors to clean up.

PR #2535, opened 2023-02-01
Check types with `tsc` in CI

delucis:

#### Description

- Runs `tsc` ([TypeScript’s compilation CLI](https://www.typescriptlang.org/docs/handbook/compiler-options.html)) in CI. This adds type-checking in `.ts` files. Currently we only type check `.astro` files with `astro check`.
- Fixes issues currently thrown up by `tsc`
- Most of these were minimal (`type` imports not marked as such, orphaned keys in translation string dictionaries). A couple of remark plugins needed more work.
- Two errors seemed insurmountable so I slapped a `@ts-expect-error` comment on them:
  - `@11ty/eleventy-fetch` doesn’t provide types and nor does Definitely Typed
  - One remark plugin was triggering “potentially infinite” type instantiation. Tried a few things but couldn’t fix it. (Also only seems to be the case with `tsc` CLI. Editor tooling is slow but doesn’t hit that depth limit.)


integrations/astro-asides.ts
```diff
@@ -37,8 +37,9 @@ function remarkAsides(): unified.Plugin<[], mdast.Root> {
 	const variants = new Set(['note', 'tip', 'caution', 'danger']);
 
 	const transformer: unified.Transformer<mdast.Root> = (tree, file) => {
+		// @ts-expect-error Possibly infinite type instantiation we can’t do anything about.
```
Princesseuh, line 40: Remark types are very scary, it makes the editor so slow, ha!

src/util-server.ts
```diff
@@ -3,6 +3,7 @@
 	Do not import this file from a hydrated client-side component.
 */
 
+// @ts-expect-error Package without types we can’t do anything about.
```
Princesseuh, line 6: It's usually seen as better to declare an empty module for the package than to expect error it, but both are okay

`declare module "@11ty/eleventy-fetch";`

Princesseuh (approved):
Just a non-blocking comment, but otherwise it looks good to me!

src/util-server.ts
delucis: My thinking was to limit this to the one file where we’re using eleventy-fetch. Declaring a module has to be done in a separate `.d.ts` otherwise, right?

Declaring a module in this file throws:

> Invalid module name in augmentation. Module '@11ty/eleventy-fetch' resolves to an untyped module, which cannot be augmented.

Princesseuh: Yeah, it would have to be in a separate file. `@ts-expect-error` is totally fine too

PR #2539, opened 2023-02-01
Add note about prerendering MDX files

RobertAKARobin:

- New or updated content

#### Description

    - Adds an example of how to enable prerendering on MDX pages when using SSR.

### DIFF
--- a/src/pages/en/guides/server-side-rendering.mdx
+++ b/src/pages/en/guides/server-side-rendering.mdx
@@ -129,7 +129,7 @@ export const prerender = true;
 # This is my page
 ```
 
-Plain `.md` Markdown pages do not support exporting variables, and so must be converted to `.mdx` before they can be prerendered.
+Markdown (`.md`) pages do not support prerendering.
 
 ### `Astro.request.headers`
 
--- a/src/pages/en/guides/server-side-rendering.mdx
+++ b/src/pages/en/guides/server-side-rendering.mdx
@@ -105,7 +105,7 @@ Astro will remain a static-site generator by default. But once you enable a serv
 
 <Since v="2.0.0" />
 
-To enable **hybrid rendering**, any [`.astro` page](/en/core-concepts/astro-pages/) or [server endpoint](/en/core-concepts/endpoints/#server-endpoints-api-routes) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
+To enable **hybrid rendering**, [pages](/en/core-concepts/astro-pages/) and [server endpoints](/en/core-concepts/endpoints/#server-endpoints-api-routes) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
 
 ```astro title="src/pages/index.astro" {2}
 ---
--- a/src/pages/en/guides/server-side-rendering.mdx
+++ b/src/pages/en/guides/server-side-rendering.mdx
@@ -74,7 +74,7 @@ You can also add an adapter manually by installing the package and updating `ast
      </Fragment>
      <Fragment slot="pnpm">
      ```shell
-     pnpm install @astrojs/my-adapter 
+     pnpm install @astrojs/my-adapter
      ```
      </Fragment>
      <Fragment slot="yarn">
@@ -105,9 +105,11 @@ Astro will remain a static-site generator by default. But once you enable a serv
 
 <Since v="2.0.0" />
 
-To enable **hybrid rendering**, [pages](/en/core-concepts/astro-pages/) and [server endpoints](/en/core-concepts/endpoints/#server-endpoints-api-routes) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
+To enable **hybrid rendering**, any [page](/en/core-concepts/astro-pages/) or [server endpoint](/en/core-concepts/endpoints/#server-endpoints-api-routes) that supports exporting variables (`.astro`, `.mdx`, `.ts`, or `.js`) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
 
-```astro title="src/pages/index.astro" {2}
+Simply add `export const prerender = true`:
+
+```astro title="src/pages/mypage.astro" {2}
 ---
 export const prerender = true;
 // ...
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -105,7 +105,7 @@ Astro will remain a static-site generator by default. But once you enable a serv
 
 <Since v="2.0.0" />
 
-To enable **hybrid rendering**, any [page](/en/core-concepts/astro-pages/) or [server endpoint](/en/core-concepts/endpoints/#server-endpoints-api-routes) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
+To enable **hybrid rendering**, any [`.astro` page](/en/core-concepts/astro-pages/) or [server endpoint](/en/core-concepts/endpoints/#server-endpoints-api-routes) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
```
Jutanium, line 108:
```suggestion
To enable **hybrid rendering**, [pages](/en/core-concepts/astro-pages/) and [server endpoints](/en/core-concepts/endpoints/#server-endpoints-api-routes) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
```

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -117,6 +117,20 @@ export const prerender = true;
...
+---
+layout: '../layouts/markdown.astro'
+title: 'My page'
+---
+export const prerender = true;
+
+# This is my page
+```
+
+Plain `.md` Markdown pages do not support exporting variables, and so must be converted to `.mdx` before they can be prerendered.
```
Jutanium, line 132:
```suggestion
Markdown (`.md`) pages do not support prerendering.
```
We don't need to explicitly mention converting - it's clear that they must choose one of the other options in order to use prerendering.

Jutanium (changes requested):
Thanks so much for getting this started! Left a few points to edit, and we should be good to go.

src/pages/en/guides/server-side-rendering.mdx
sarah11918: Yup, just want to say I agree with Dan here. I never like to tell people to change their stuff. (We have some VERY strong fans of Markdown :sweat_smile:  and you don't necessarily want to make it sound like THEY made a poor choice or something.) This gives them the information they need so they can decide whether it's worth the tradeoff for themselves!

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -117,6 +117,20 @@ export const prerender = true;
 </html>
 ```
 
+[MDX pages](/en/guides/markdown-content/#using-exported-variables-in-mdx) can also be prerendered:
```
sarah11918, line 120: I like @Jutanium's suggestion to remove "any" pages (since clearly, that's not true). I think if we do that, then BOTH  the `.astro` and the `.mdx` examples need an introductory line, because now MDX is not an "also" thing.

Maybe something like:

> Add `export const prereder = true` to any `.astro` or `.mdx` page in the appropriate place to render it statically:

And then show the two code snippets? Something like that? What do we think?

RobertAKARobin: Yeah, just saying "pages" implies that _all_ pages can be prerendered, which is not the case. The Pages docs say:

> Astro supports the following file types in the src/pages/ directory:
> 
> [.astro](https://docs.astro.build/en/core-concepts/astro-pages/#astro-pages)
> [.md](https://docs.astro.build/en/core-concepts/astro-pages/#markdownmdx-pages)
> .mdx (with the [MDX Integration installed](https://docs.astro.build/en/guides/integrations-guide/mdx/#installation))
> [.html](https://docs.astro.build/en/core-concepts/astro-pages/#html-pages)
> .js/.ts (as [endpoints](https://docs.astro.build/en/core-concepts/endpoints/))

So I think it makes sense to be explicit that only `.astro`, `.mdx`, and endpoints can be prerendered. I'll add an example for `.js/.ts`.

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -105,9 +105,11 @@ Astro will remain a static-site generator by default. But once you enable a serv
 
 <Since v="2.0.0" />
 
-To enable **hybrid rendering**, any [page](/en/core-concepts/astro-pages/) or [server endpoint](/en/core-concepts/endpoints/#server-endpoints-api-routes) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
+To enable **hybrid rendering**, any [page](/en/core-concepts/astro-pages/) or [server endpoint](/en/core-concepts/endpoints/#server-endpoints-api-routes) that supports exporting variables (`.astro`, `.mdx`, `.ts`, or `.js`) can opt-in to prerendering. These files will be statically rendered at build-time, similar to the default static [`output`](/en/reference/configuration-reference/#output) mode.
```
sarah11918, line 108: I like this change a lot! :rocket:

RobertAKARobin: Pardon my flurry of force-pushes — whenever I make a commit I suddenly want to change it in 42 ways. :)

src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -117,6 +119,36 @@ export const prerender = true;
...
+
+```astro title="src/pages/myendpoint.ts" {2}
+export const prerender = true;
+
+export async function get() {
+  return {
+    body: JSON.stringify({ message: `This is my endpoint` }),
+  };
+}
+```
```
sarah11918, line 140:
```suggestion
```
I suggest we don't include *both* a `.js` and `.ts` example, especially since they are so similar. In the docs, we typically only include `.js` and it's assumed that people will apply these examples to `.ts`.

src/pages/en/guides/server-side-rendering.mdx
sarah11918: Haha, no worries! Totally the same! Just trying to keep up here!

Jutanium (approved):

sarah11918:
Looking great now, @RobertAKARobin! Thanks for this helpful PR, and for rolling with the changes as we get it massaged into Docs form. And, welcome to Team Docs!  :partying_face:

RobertAKARobin:
I did a dumb: https://github.com/withastro/docs/pull/2565/files

PR #2545, opened 2023-02-02
Add `html` to RSS guide

bholmesdev:

- New or updated content

#### Description

- Recommend `html` helper for rendering Markdown contents https://github.com/withastro/astro/pull/6097
- Add MDX caution banner to make instructions more explicit. Felt like we were hiding this caveat before

### DIFF
--- a/src/pages/en/guides/rss.mdx
+++ b/src/pages/en/guides/rss.mdx
@@ -141,6 +141,10 @@ items: import.meta.glob('./blog/*.{md,mdx}'),
 
 The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
 
+:::caution
+These examples focus on rendering Markdown contents. If you are using MDX, you will need to compile MDX contents to HTML manually. We recommend standard Markdown parsers like [Markdown-it](https://github.com/markdown-it/markdown-it) to get started, though this will not resolve JSX expressions or components.
+:::
+
 When using content collections, you can render post contents by calling the post's `render()` function and using the `html` property. Note the `html` property **is not present** for MDX entries. If you need to render MDX content, you can compile your MDX manually using the post's `body` property.
 
 :::tip
--- a/src/pages/en/guides/rss.mdx
+++ b/src/pages/en/guides/rss.mdx
@@ -151,7 +151,7 @@ When using content collections, you can render post contents by calling the post
 Whenever you're using HTML content in XML, we suggest using a package like [`sanitize-html`](https://www.npmjs.com/package/sanitize-html) in order to make sure that your content is properly sanitized, escaped, and encoded.
 :::
 
-```js title="src/pages/rss.xml.js" ins={2, 11, 14} "async (post)"
+```js title="src/pages/rss.xml.js" ins={2, 12, 15} "await Promise.all(" "async (post)"
 import rss from '@astrojs/rss';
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/rss.mdx
```diff
@@ -141,35 +141,41 @@ items: import.meta.glob('./blog/*.{md,mdx}'),
 
 The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
 
+:::caution
```
sarah11918, line 144: Gonna call a flag on the play for caution here! It's not a good look to basically have Section Title then immediately CAUTION as if the *thing itself* is risky. If it's only our examples/explanation that requires a warning, then we fix that with writing!  ;)

There are a couple of different ways we could go about doing this, but let's start by getting your gut reaction: if you were not allowed to use a caution, AND not allowed to put ANY kind of aside (note/tip) upfront like this, how/where would include this piece of information?

src/pages/en/guides/rss.mdx
```diff
@@ -141,35 +141,41 @@ items: import.meta.glob('./blog/*.{md,mdx}'),
...
+These examples focus on rendering Markdown contents. If you are using MDX, you will need to compile MDX contents to HTML manually. We recommend standard Markdown parsers like [Markdown-it](https://github.com/markdown-it/markdown-it) to get started, though this will not resolve JSX expressions or components.
+:::
+
+When using content collections, you can render post contents by calling the post's `render()` function and using the `html` property. Note the `html` property **is not present** for MDX entries. If you need to render MDX content, you can compile your MDX manually using the post's `body` property.
```
sarah11918, line 148: Using content collections isn't *required* for full-post content right? We had this before? (Since 1.6.14)

This seems like when this page was updated, it was updated to now *only* show an example with content collections. I think it's important to show both, especially since this our blog tutorial includes creating an RSS feed but not using collections. So while we're here and updating the example, let's find a way to re-incorporate an example that works with globbing files in `src/pages/`, assuming that still works.

This was the original example for full post content. Will this (or an updated version of this) still work for projects not using collections?

```js ins={2, 16} title={src/pages/rss.xml.js}
import rss from '@astrojs/rss';
import sanitizeHtml from 'sanitize-html';
// Works with Markdown files only!
const postImportResult = import.meta.glob('../posts/**/*.md', { eager: true }); 
const posts = Object.values(postImportResult);
export const get = () => rss({
  title: 'Buzz’s Blog',
  description: 'A humble Astronaut’s guide to the stars',
  site: import.meta.env.SITE,
  items: posts.map((post) => ({
    link: post.url,
    title: post.frontmatter.title,
    pubDate: post.frontmatter.pubDate,
    content: sanitizeHtml(post.compiledContent()),
  }))
});
```

sarah11918 (commented):
I've left a couple of larger issue questions to consider here so I can move forward with this, @bholmesdev!

I am also giving you permission to "tag out" at any time if you're uninterested in solving these issues and want to tell Team Docs to figure it out.  :sweat_smile:

sarah11918:
Hi @bholmesdev - I think this idea was scrapped, and so this PR can be closed now. Please reopen if that's not the case! (There were actually two open Docs PRs associated with this, one from you and one from @bluwy , but only that one was closed alongside the Astro PR.)

PR #2561, opened 2023-02-03
Tutorial 2.1 - Don't use the term "routes" before introducing it

Jutanium:

- New or updated content

#### Description

Addresses this Feedback Fish:
> Nice intro, very supportive of people new to web dev!
> 
> With that in mind has the concept of routes been introduced yet in 2.1? There's a note that says "Astro uses standard HTML <a> elements to navigate between routes", but I'm not sure people will know what a route is yet.

Changes "routes" there to pages, and adds additional context to the note - it's implied that we're contrasting Astro's links to other framework's link systems, but if you're new to frameworks in general this isn't apparent.

### DIFF
--- a/src/pages/en/tutorial/2-pages/1.mdx
+++ b/src/pages/en/tutorial/2-pages/1.mdx
@@ -70,7 +70,7 @@ To make it easier to preview all your pages, add HTML page navigation links befo
 Check that you can click these links to move back and forth between pages on your site.
 
 :::note
-Unlike many frameworks, Astro uses standard HTML `<a>` elements to navigate between pages, with traditional page refreshes.
+Unlike many frameworks, Astro uses standard HTML `<a>` elements to navigate between pages (also called _routes_), with traditional page refreshes.
 :::
 
 <Box icon="puzzle-piece">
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Approving this, Dan! Is it worth introducing the word routes some how here? "pages (or routes" or do you think not bother?  Your call!

src/pages/en/tutorial/2-pages/1.mdx
```diff
@@ -70,7 +70,7 @@ To make it easier to preview all your pages, add HTML page navigation links befo
 Check that you can click these links to move back and forth between pages on your site.
 
 :::note
-Astro uses standard HTML `<a>` elements to navigate between routes, with traditional page refreshes.
+Unlike many frameworks, Astro uses standard HTML `<a>` elements to navigate between pages, with traditional page refreshes.
```
Jutanium, line 73:
```suggestion
Unlike many frameworks, Astro uses standard HTML `<a>` elements to navigate between pages (also called _routes_), with traditional page refreshes.
```
@sarah11918 How about this, so it feels like a direct (rather than by-the-way) introduction?

sarah11918: Ship it!

PR #2563, opened 2023-02-03
Tutorial 2.2 - Explain Frontmatter

Jutanium:

- New or updated content

#### Description

Addresses this Feedback Fish:
> In those example, the images are not being rendered...

If a reader doesn't know what frontmatter is, they might expect the image to be rendered alongside the rest of the document. (This is especially true since the title _does_ show up, as we have it in both the frontmatter and the body.)

This change briefly introduces frontmatter and tells the reader not to expect the image to show up.

### DIFF
--- a/src/pages/en/tutorial/2-pages/2.mdx
+++ b/src/pages/en/tutorial/2-pages/2.mdx
@@ -74,7 +74,7 @@ Now that you have built pages using `.astro` files, let's make some blog posts u
 3. Use your browser's Dev Tools to inspect this page. Notice that although you have not typed any HTML elements, your Markdown has been converted to HTML. You can see elements such as headings, paragraphs, and list items.
 
 :::note
-The information at the top of the file, inside the code fences, is called frontmatter. This data—including tags and a post image—won't appear until we access it later in the tutorial. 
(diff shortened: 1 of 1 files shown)


src/pages/en/tutorial/2-pages/2.mdx
```diff
@@ -73,6 +73,10 @@ Now that you have built pages using `.astro` files, let's make some blog posts u
 
 3. Use your browser's Dev Tools to inspect this page. Notice that although you have not typed any HTML elements, your Markdown has been converted to HTML. You can see elements such as headings, paragraphs, and list items.
 
+:::note
+The information at the top of the file, inside the code fences, is called frontmatter. This data—including tags and a post image—won't appear until we access it later in the tutorial.
```
sarah11918, line 77:
```suggestion
The information at the top of the file, inside the code fences, is called frontmatter. This data—including tags and a post image—is information *about* your post that Astro can use. It does not appear directly on the page, but we will access it later in the tutorial to enhance your site. 
```
How about something like this?

Jutanium: Yes! One change:

> It does not appear on the page automatically, but we will access it later in the tutorial to enhance your site

(it will eventually appear directly on the page)

sarah11918: Amended my suggestion!

sarah11918 (approved):
Approving with the change we discussed! :partying_face:

PR #2566, opened 2023-02-04
Correct Storyblok SSG sample

Jutanium:

- New or updated content

#### Description

- Updates the `getStaticPaths` example to call the API _after_ the paths are generated, on each resulting page. This allows Storyblok Bridge to work. Thanks to Storyblok's @manuelschroederdev for this fix!
- Removes the `undefined` route generation. This no longer works, as we need a slug to be passed through to the page. The user can specify their own `index.astro` page if they want the `undefined`/home route, and this isn't integral to learning Storyblok.

### DIFF
--- a/src/pages/en/guides/cms/storyblok.mdx
+++ b/src/pages/en/guides/cms/storyblok.mdx
@@ -313,7 +313,7 @@ import { useStoryblokApi } from '@storyblok/astro'
 const storyblokApi = useStoryblokApi();
 
 const { data } = await storyblokApi.get('cdn/stories', {
-  version: import.meta.env.DEV ? "draft" : "published",
+  version: import.meta.env.DEV ? ersion: : "published",
   content_type: 'blogPost',
 })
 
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/cms/storyblok.mdx
```diff
@@ -313,7 +313,7 @@ import { useStoryblokApi } from '@storyblok/astro'
 const storyblokApi = useStoryblokApi();
 
 const { data } = await storyblokApi.get('cdn/stories', {
-  version: import.meta.env.DEV ? "draft" : "published",
+  version: import.meta.env.DEV ? ersion: : "published",
```
kevinzunigacuellar, line 316: typo?
```suggestion
  version: import.meta.env.DEV ? "draft" : "published",
```

Jutanium: thanks!

sarah11918 (approved):
LGTM, @Jutanium !

PR #2569, opened 2023-02-05
Add info about JetBrains Astro plugin

Yan-Thomas:

- Translated content

#### Description

This PRs changes a paragraph with more up-to-date info about the currently in Early Access JetBrains plugin for Astro, with guidance on how to install it and which IDEs can benefit from it.


sarah11918 (approved):
What would we do without you on the case, @Yan-Thomas? Thank you!

PR #2571, opened 2023-02-05
Update ghost.mdx

AsenCME:

#### Description
When running ghost locally and running Astro locally, "localhost" is resolved as ":::", which in turn crashes it. Changing it to "127.0.0.1" (provided that both are running locally) fixes the problem. Just ran into this now while integrating Ghost.


sarah11918:
Thank you for this helpful contribution, @AsenCME! We appreciate you taking the time to make this fix, and welcome to Team Docs! 🥳

PR #2574, opened 2023-02-05
Add missing CLI references.

jamesmurdza:

#### Description

The current page only includes one CLI option although there are several.


sarah11918:
Thanks for this, @jamesmurdza! This configuration page is auto-generated though, so we can't just make this change in this file. I'm going to have to send you on an adventure!  :smile: In order to make this change, you'll need to PR this file in the main Astro repo: https://github.com/withastro/astro/blob/main/packages/astro/src/%40types/astro.ts

Let us know if you have any trouble with that, and I'll close this PR in favour of the change being made in the other file! 🙌

PR #2579, opened 2023-02-06
Storyblok Revision: Home page and blogPost filtering

Jutanium:

- New or updated content

#### Description

- Adds an `index.astro` page for the homepage
- Storyblok API call now filters to only grab blog posts (so you don't render the homepage in two places)
- The dynamic route page is now at `pages/blog/[...slug].astro`

[Deploy Preview](https://deploy-preview-2579--astro-docs-2.netlify.app/en/guides/cms/storyblok/#generating-pages)


src/pages/en/guides/cms/storyblok.mdx
```diff
@@ -192,18 +192,16 @@ export default defineConfig({
...
 ---
 import { useStoryblokApi } from '@storyblok/astro'
 import StoryblokComponent from '@storyblok/astro/StoryblokComponent.astro'
 
 const storyblokApi = useStoryblokApi()
 
-const { data } = await storyblokApi.get("cdn/stories/test-post", {
+const { data } = await storyblokApi.get("cdn/stories/home", {
   version: import.meta.env.DEV ? "draft" : "published",
 });
```
kevinzunigacuellar, line 207: Maybe we need a disclaimer at the end stating that this fetch will only work if you have a story named `home`

src/pages/en/guides/cms/storyblok.mdx
```diff
@@ -368,29 +366,54 @@ export default defineConfig({
...
+const storyblokApi = useStoryblokApi();
+const { data } = await storyblokApi.get('cdn/stories/home');
+const content = data.story.content;
+---
+<html lang="en">
+  <head>
+    <title>Storyblok & Astro</title>
+  </head>
+  <body>
+    <Page blok={content} />
```
kevinzunigacuellar, line 386:
```suggestion
    <Page blok={content} />
```

I am not sure why we are not using the Storyblok component (from the storyblok library) in this case. It should work as it was but only fetching data from one story.

src/pages/en/guides/cms/storyblok.mdx
Jutanium: We do say to create `home` earlier

src/pages/en/guides/cms/storyblok.mdx
Jutanium: What is `Page` for then? I thought `Page` was for rendering the `page` type

src/pages/en/guides/cms/storyblok.mdx
```diff
@@ -192,18 +192,16 @@ export default defineConfig({
 
 ### Fetching data
 
-To test the setup, in Storyblok create a new story with the `blogPost` content type named `test-post`. 
+To test the setup, create a new page in the `src/pages` directory named `index.astro` with the following content:
 
-In Astro, create a new page in the `src/pages` directory named `test-post.astro` with the following content:
-
-```astro title="pages/test-post.astro"
+```astro title="pages/index.astro"
```
sarah11918, line 197:
``````suggestion
```astro title="src/pages/index.astro"
``````
Pretty sure we mostly start filepaths from `src/`.

File: src/pages/en/guides/cms/storyblok.mdx [hunk 4]
```diff
@@ -368,29 +366,54 @@ export default defineConfig({
 
 ### Generating pages
 
+To create a route for a specific `page`, you can fetch its directly from the Storyblok API and pass it to the `Page` component. Create an `index.astro` file in `src/pages` to render the `home` page:
```
sarah11918, line 369:
```suggestion
To create a route for a specific `page`, you can fetch its content directly from the Storyblok API and pass it to the `Page` component. Create an `index.astro` file in `src/pages/` to render the `home` page:
```
Fetch "it", or fetch "its....(missing word/property)"? (I'd lean towards the latter, because you have to pass an actual *thing* to the `<Page />` component.

(This suggestion only edits what's there, and does not take into account Kevin's comment that perhaps the step should itself be a different instruction. 😄 )

src/pages/en/guides/cms/storyblok.mdx
```diff
@@ -368,29 +366,54 @@ export default defineConfig({
...
+  <head>
+    <title>Storyblok & Astro</title>
+  </head>
+  <body>
+    <Page blok={content} />
+  </body>
+</html>
+```
+
+To dynamically render all of the blog posts, create a dynamic route. This approach varies depending on whether you're using **static site generation** (the default) or **server-side rendering**.
```
sarah11918, line 391:
```suggestion
To generate pages for all of your blog posts, create a `.astro` page that will create dynamic routes. This approach varies depending on whether you're using **static site generation** (the default) or **server-side rendering**.
```
Just thinking that the reader doesn't care that "blog posts are dynamically generated" so much as they care that "a page is made for each blog post"? Plus, this avoids using the word "dynamic" twice. :smile: (And, I'm not sure I'd say *a* dynamic route is created, but a page that creates a bunch of routes.) Tweak this as you like!

src/pages/en/guides/cms/storyblok.mdx
```diff
@@ -368,29 +366,54 @@ export default defineConfig({
...
+
 #### Static site generation
 
 If you are using Astro's default static site generation, you will use [dynamic routes](/en/core-concepts/routing/#dynamic-routes) and the `getStaticPaths` function to generate your project pages.
 
-Create a new file in the `src/pages` directory called `[...slug].astro` and add the following code:
+Create a new file in a `src/pages/blog` directory called `[...slug].astro` and add the following code:
```
sarah11918, line 397:
```suggestion
Create a new directory `src/pages/blog/` and add a new file called `[...slug].astro` with the following code:
```
If the directory doesn't exist, probably good practice to instruct the reader to create the directory.

src/pages/en/guides/cms/storyblok.mdx
```diff
@@ -420,12 +443,12 @@ This file will generate a page for each story, with the slug and content fetched
 
 If you’ve [opted into SSR mode](/en/guides/server-side-rendering/#enabling-ssr-in-your-project), you will use dynamic routes to fetch the page data from Storyblok.
```
sarah11918, line 444: Just noting here that the previous instruction to create the directory and the file is techically under "if you're using SSG" so you might want a similar instruction (vs just a code sample that exists in comparison to the previous section) in case someone just skimmed and came directly here.

src/pages/en/guides/cms/storyblok.mdx
Jutanium: My bad - I meant to only make this change in the **Making a blog** section. Reverting this one

src/pages/en/guides/cms/storyblok.mdx
kevinzunigacuellar: @Jutanium I see where this is coming from. Uhmm I am not sure which one should be used. Page is the container for multiple bloks in the page content-type story.

PR #2581, opened 2023-02-06
Client directive disclaimer: component must be imported

### LINKED ISSUE #2480: Use of client directives requires direct import
According to @matthewp in https://github.com/withastro/astro/issues/5853:

> The client: directives only work with direct imports. This is because Astro needs to track which components are being used in the client so that it can build them all out properly.

I feel like this could be in the docs somewhere, probably in [the rules section of `directives-reference`](https://github.com/withastro/docs/blob/main/src/pages/en/reference/directives-reference.mdx#rules).

I would write the PR myself, but I'm not sure of the scope of this issue — I assume components must be direct imports for _any_ directives to work, but I'm not sure.

Jutanium:

- New or updated content

#### Description

- Closes #2480 
- Clarifies under Client Directives that they must be placed on an imported component.

### DIFF
--- a/src/pages/en/reference/directives-reference.mdx
+++ b/src/pages/en/reference/directives-reference.mdx
@@ -97,6 +97,8 @@ These directives control how [UI Framework components](/en/core-concepts/framewo
 
 By default, a UI Framework component is not hydrated in the client. If no `client:*` directive is provided, its HTML is rendered onto the page without JavaScript.
 
+A client directive must be placed on an imported component (i.e., not a [dynamic](/en/core-concepts/astro-components/#dynamic-tags) or [passed](/en/guides/markdown-content/#custom-components-with-imported-mdx) component).
+
 ### `client:load`
 
 - **Priority:** High
--- a/src/pages/en/reference/directives-reference.mdx
+++ b/src/pages/en/reference/directives-reference.mdx
@@ -97,7 +97,7 @@ These directives control how [UI Framework components](/en/core-concepts/framewo
 
 By default, a UI Framework component is not hydrated in the client. If no `client:*` directive is provided, its HTML is rendered onto the page without JavaScript.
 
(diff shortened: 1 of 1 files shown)


Yan-Thomas (approved):
LGTM! Nice work @Jutanium 🙌

src/pages/en/reference/directives-reference.mdx
```diff
@@ -97,6 +97,8 @@ These directives control how [UI Framework components](/en/core-concepts/framewo
 
 By default, a UI Framework component is not hydrated in the client. If no `client:*` directive is provided, its HTML is rendered onto the page without JavaScript.
 
+A client directive must be placed on an imported component (i.e., not a [dynamic](/en/core-concepts/astro-components/#dynamic-tags) or [passed](/en/guides/markdown-content/#custom-components-with-imported-mdx) component).
```
sarah11918, line 100:
```suggestion
A client directive can only be used on a UI framework component that is directly imported into a `.astro` component.  Hydration directives are not supported when using [dynamic tags](/en/core-concepts/astro-components/#dynamic-tags) and [custom components passed via the `components` prop](/en/guides/markdown-content/#custom-components-with-imported-mdx).
```
Is it that they *must* be used on an imported component, or can *only* be used on this type of component?

I might go for something closer to this, and would give the exclusions a full, proper sentence. (I don't know if "directly imported" makes sense, but it was my first attempt at describing something that gets its own, explicit import statement.)

And, I noticed that both of those links go to places where our examples are *not* UI framework components. (e.g. it's an Astro component for Dynamic Tags). So, these types of components we're linking to aren't even guaranteed to be UI framework components in the first place, and I don't *think* that warrants any special reminder of that fact (i.e. if they're not UI framework, they totally can't have client directives)... but I did notice it so I'm mentioning it in case we think it's noteworth.

Jutanium: I like this rewording and I think it helps remediate any confusion with the examples not being framework components!

sarah11918 (approved):

PR #2591, opened 2023-02-07
Documents `loadEnv` workaround 

### LINKED ISSUE #2506: Adding a "caution" box to the integration API reference
I run into an issue with the integration API today. 

When you import lib functions within an custom integration which itself use the env variables as described [here](https://docs.astro.build/en/guides/environment-variables/) the env vars will be undefined. 

aFuzzyBear on the official Discord looked into this and there is a proper way also make them available in the config script. 
But at first glance it may not be obvious. 

**Possible solution:**
I am happily create an PR with a little "caution" box to hint to this behaviour and show the possible workaround aFuzzyBear provided to make env vars available also in the astro.config.mjs file.

The example code by FuzzyBear, i would try to make it a little more generic and less about the code we tested it with : 
```js
import { test } from './src/lib/test';
import { loadEnv } from 'vite';

const key = loadEnv('', process.cwd(), 'HELLO_WORLD');
// https://astro.build/config
export default defineConfig({
  integrations: [
    {
      name: 'test:id',
      hooks: {
        'astro:build:done': async ({ dir, routes }) => {
          const testVAR = await test();
          console.log('TESTVAR', testVAR);
          console.log(key);
        },
      },
    },
  ],
});
```

Affected page: 
https://docs.astro.build/en/reference/integrations-reference/

Jutanium:

- New or updated content

#### Description

- Closes #2506, but **doesn't** add a caution box to the integration API reference. That's because this doesn't have anything to do with integrations specifically, but rather the Astro config file.
- Documents the `loadEnv` workaround for accessing environment variables in the Astro config file. This comes up often in support, but is missing from our docs.
- Updates wording to implicate that `import.meta.env` **is** available - it just doesn't have `.env` variables on it. This is why we can use `import.meta.env.MODE` to tell `loadEnv` to grab the right `.env` files.

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -16,10 +16,7 @@ import { rehypei18nAutolinkHeadings } from './plugins/rehype-i18n-autolink-headi
 import { rehypeTasklistEnhancer } from './plugins/rehype-tasklist-enhancer';
 import { remarkFallbackLang } from './plugins/remark-fallback-lang';
 import { theme } from './syntax-highlighting-theme';
-import { loadEnv } from 'vite';
 
-const env = loadEnv(import.meta.env.MODE, process.cwd(), '');
-console.log(env);
 // https://astro.build/config
 export default defineConfig({
 	site: 'https://docs.astro.build/',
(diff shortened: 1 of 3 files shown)


sarah11918 (approved):
LGTM, @Jutanium!

PR #2593, opened 2023-02-07
docs: refactor nodejs installation steps for blog tutorial

### LINKED ISSUE #2592: [Enhancement] Make Node.js installation a separate step in the Build a blog tutorial
The [Prepare your dev environment](https://docs.astro.build/en/tutorial/1-setup/1/) page contains `Terminal` and `Code Editor` sections as prerequisites for you to get started.
However, the `Terminal` section **also contains** details on how to install `Node.js` and these instructions are messy:
1. First, it tells installing Node.js as the main requirement for Astro. OK. But then, it also tells checking the version of Node.js. This is confusing because the instructions don't take into account a negative flow, when the reader might not have Node.js installed yet. Why would they need to check the version of a tool that they haven't installed yet? Yes, this command is a 'de-facto' node installation checker as well, but this should be stated more clearly.
    * No terminal output is shown. It's not clear what the reader should expect to see after running the command.
3. > **If necessary**, follow the instructions to install Node.js via the command line.
    
    **If necessary** sounds very strange as it implies "If not installed or in case you have an older version". Anyway, not pretty obvious.

My suggestion is to refactor this part by separating Node.js installation from the `Terminal` and making steps more clear.

**PR is ready**

bandantonio:

- New or updated content

#### Description

- Closes #2592


Jutanium (approved):
This is a great improvement to the flow of that section. I suspect we'll want to remove the emoji, but I'll look towards @sarah11918's editorial voice for that!

src/pages/en/tutorial/1-setup/1.mdx
```diff
@@ -26,15 +26,22 @@ You will use a **command line (terminal)** to create your Astro project and to r
 
 You can access the command line through a local terminal program for your operating system. Common applications include **Terminal** (MacOS/Linux), **Command Prompt** (Windows), and **Termux** (Android). One of these will probably already be on your machine. 
 
-For Astro to run on your system, you will also need to have [**Node.js**](https://nodejs.org/en/) installed. To check to see whether you already have a compatible version (`v16.12.0` or higher) installed: 
+### Node.js
 
-1. Launch your terminal and type the following into the command line:
+For Astro to run on your system, you also need to have [**Node.js**](https://nodejs.org/en/) of version `v16.12.0` or later.
```
sarah11918, line 31:
```suggestion
For Astro to run on your system, you will also need to have [**Node.js**](https://nodejs.org/en/) installed. 
```
I agree that Node is significant enough that it can have its own heading! I'm just reworking to avoid the phrase "Node of version" which is a little unusual here in the context of the rest of our text.

I've also saved mention of the version number for the next line.

src/pages/en/tutorial/1-setup/1.mdx
```diff
@@ -26,15 +26,22 @@ You will use a **command line (terminal)** to create your Astro project and to r
...
+First, check whether you already have Node.js installed and check the installed version. To do this, run the following command in your terminal:
```
sarah11918, line 32:
```suggestion

To check to see whether you already have a compatible version (`v16.12.0` or higher) installed, run the following command in your terminal:
```
Looking at the preview, these lines run together. So, I think an extra line space is needed here.

And, I'm not sold on the proposed change. It seems overly wordy (check and check; then refering back to it again with "this") when we currently say "check that you have a compatible version installed" which covers both cases directly.

src/pages/en/tutorial/1-setup/1.mdx
```diff
@@ -26,15 +26,22 @@ You will use a **command line (terminal)** to create your Astro project and to r
...
 
-2. If necessary, follow the instructions to [install Node.js](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm) via the command line.
+// Example output
+v16.14.0
+```
+
+* If the command returns a version number higher than `v16.12.0`, you're good to go! 🎉 If not, you need to install the compatible Node.js version.
+* If the command returns an error message like `Command 'node' not found`, Node.js is not installed in your system.
+
+To install Node.js (or the compatible version), follow [this guide](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm).
```
sarah11918, line 44:
```suggestion
```
As mentioned above, we don't need this anymore.

src/pages/en/tutorial/1-setup/1.mdx
```diff
@@ -26,15 +26,22 @@ You will use a **command line (terminal)** to create your Astro project and to r
...
+```sh
+node -v
 
-2. If necessary, follow the instructions to [install Node.js](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm) via the command line.
+// Example output
+v16.14.0
+```
+
+* If the command returns a version number higher than `v16.12.0`, you're good to go! 🎉 If not, you need to install the compatible Node.js version.
+* If the command returns an error message like `Command 'node' not found`, Node.js is not installed in your system.
```
sarah11918, line 42:
```suggestion
If the command returns a version number higher than `v16.12.0`, you're good to go!

If the command returns an error message like `Command 'node' not found`, or a version number lower than `v16.12.0`, then you need to [install a compatible Node.js version](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm).
```
I think it makes more sense to split these into a success/failure case, vs either being good or not in the first point, then not being good in the second point. And in that case, we can bring the link up into this point with the solution, and avoid the next paragraph entirely!  🥳 

Also, bullet points don't seem quite right here? We typically only use them to finish sentences with multple options, or a distinct list of comparable items. So, I could see using them if we did something like:

> If the command returns: 
> - a version number higher than...
> - an error message or a version number lower than...

You could try that and see if you think it looks good!

sarah11918 (commented):
Thanks for this contribution to the tutoral, @bandantonio! I agree that Node is pretty important, so I can see giving it its own heading here.

I've left some suggestions for how I can see this working and fitting into our existing docs. There are some parts of the original that I still think work, and I've made some suggestions to incorporate your ideas! See what you think!

src/pages/en/tutorial/1-setup/1.mdx
bandantonio: I disagree. Node is basically the only technical prerequisite to run Astro (terminal and code editor are just tools of choice), so making readers aware of the minimum compatible version is **crucial**, and this information should be provided as early as possible. We can further think (and agree, for sure) about rewording, but I strongly believe that details about the version must be there.

src/pages/en/tutorial/1-setup/1.mdx
bandantonio: > It seems overly wordy

[npm docs](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#checking-your-version-of-npm-and-nodejs) use almost the same wording.

> "check that you have a compatible version installed" which covers both cases directly.

No, it doesn't. Again, the current version doesn't take into acount the negative flow, when the reader doesn't have Node at all.

Let's read the current version: `To check to see whether you already have a compatible version ... installed` What outcomes are expected with this phrase?
The reader has either a compatible version installed, or an incompatible one.
But there is a third possible outcome, which is not covered by the current version: when the reader **gets an error** after performing `node -v` because they have no Node installed in their system. 

Yes, it is a very unlikely scenario considering the target audience experience and skills, but it is still a possible outcome.

The first Technical Writer's dogma is: "Don't expect your readers to be as smart as you are".

src/pages/en/tutorial/1-setup/1.mdx
bandantonio: > I think it makes more sense to split these into a success/failure case, vs either being good or not in the first point, then not being good in the second point. And in that case, we can bring the link up into this point with the solution, and avoid the next paragraph entirely! 🥳

Well, yes, agreed. As long as the solution is the same for both cases, we can do that trick :)

src/pages/en/tutorial/1-setup/1.mdx
Jutanium: > "To check to see whether you already have a compatible version (`v16.12.0` or higher) installed" 
> What outcomes are expected with this phrase?

The possible outcomes are:
- You already have a compatible version installed
- You don't have a compatible version installed
    - _Within_ this second outcome, you could either have no version installed at all, or you could have an incompatible version.

I agree that Sarah's text here is less clunky and achieves the same result, especially given your revisions below to explicitly anticipate the error case in the next section.

bandantonio: Yes, It kinda makes sense. I just wanted to be a bit more clear. If you feel like the readers won't have troubles understanding this, I'm totally OK with that.
Anyway, we can always handle this based on feedback

bandantonio:
@sarah11918 @Jutanium fixed all your mentions, please have a look.

src/pages/en/tutorial/1-setup/1.mdx
```diff
@@ -26,15 +26,22 @@ You will use a **command line (terminal)** to create your Astro project and to r
...
-For Astro to run on your system, you will also need to have [**Node.js**](https://nodejs.org/en/) installed. To check to see whether you already have a compatible version (`v16.12.0` or higher) installed: 
+### Node.js
 
-1. Launch your terminal and type the following into the command line:
+For Astro to run on your system, you will also need to have [**Node.js**](https://nodejs.org/en/) installed, version `v16.12.0` or later.
 
-    ```sh
-    node -v
-    ```
+To check to see whether you already have a compatible version (`v16.12.0` or higher) installed, run the following command in your terminal:
```
Jutanium, line 33: Thanks for this! One nit and we should be good to go!
Now that we provide the version in the first sentence, we don't need to repeat it: 

```suggestion
To check to see whether you already have a compatible version installed, run the following command in your terminal:
```

Jutanium (approved):

PR #2415, opened 2023-01-19
Corrected the way to use imported images in a regular img tag

mandar1jn:
As pointed out in [this support thread](https://discord.com/channels/830184174198718474/1065656313784307722/1065656313784307722) the way that the docs say you need to import images into a normal  tag is inaccurate. This corrects that

#### What kind of changes does this PR include?

- New or updated content

#### Description

- Closes `none` 
- What does this PR change? Give us a brief description.
Changes `src={IMAGE}` to `src={IMAGE.dev}` in the image integration guide
- Did you change something visual? A before/after screenshot can be helpful.
Not really

PR #2594, opened 2023-02-07
Warning for using `<img>` with the image integration installed

### LINKED ISSUE #2586: Document that `@astrojs/image` changes the shape of imported images
Adding `@astrojs/image` to a project changes the shape returned by images imports from a simple `string` to the following:

```ts
{
  src: string;
  width: number;
  height: number;
  format: 'avif' | 'gif' | 'heic' | 'heif' | 'jpeg' | 'jpg' | 'png' | 'tiff' | 'webp';
}
```

This is valid project-wise, including inside React and other frameworks components. `image.src` is equivalent to the previous string, and allows using the image directly, as such it's simply a matter of appending `.src`. Nonetheless, this is currently undocumented and (rightfully) confuses users

Jutanium:

- New or updated content

#### Description

- Replaces https://github.com/withastro/docs/pull/2415
- Closes #2586 
- Documents the `?url` workaround for using image imports with `` when the image integration is installed
- We could also document the `rocket.src` workaround, but I chose to go with one recommended solution for this PR.

### DIFF
--- a/src/pages/en/guides/images.mdx
+++ b/src/pages/en/guides/images.mdx
@@ -34,7 +34,7 @@ The official [image integration](#astros-image-integration) will change image im
 If you have the image integration installed, add `?url` to your imports when using `<img>`.
 ```astro "?url"
 ---
-import rocket from '../images/rocket.svg';
+import rocket from '../images/rocket.svg?url';
 ---
 <img src={rocket} alt="A rocketship in space."/>
 ```
--- a/src/pages/en/guides/images.mdx
+++ b/src/pages/en/guides/images.mdx
@@ -139,9 +139,6 @@ This component is useful for images where you want to keep a consistent size acr
 
 For responsive images, or art direction, use the `<Picture />` component instead.
 
-:::note
-Installing the image integration
-:::
 #### Local Images in `src/`
 
 (required attributes: [`src`](/en/guides/integrations-guide/image/#src), and [`alt`](/en/guides/integrations-guide/image/#alt))
--- a/src/pages/en/guides/images.mdx
+++ b/src/pages/en/guides/images.mdx
@@ -104,7 +104,7 @@ Import them from a **relative file path** or [import alias](/en/guides/aliases/)
 // src/pages/index.astro
 
 // Access images in `src/images/`
-import logo from '../images/logo.png?url';
+import logo from '../images/logo.png';
 ---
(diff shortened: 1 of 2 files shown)


src/pages/en/guides/images.mdx
```diff
@@ -29,6 +29,17 @@ import rocket from '../images/rocket.svg';
 <img src={rocket} alt="A rocketship in space."/>
 ```
 
+:::caution
+The official [image integration](#astros-image-integration) will change image imports to return an object.
```
sarah11918, line 33:
```suggestion
When installed, the [official image integration](#astros-image-integration) will change all image imports to return an object, even when you don't use the `<Image />` component.
```
I don't know whether this is a little much, but I do think it's important to say something like "When installed..." behaviour is changed, especially since we're so early on this page!

Jutanium: At this point, they haven't been exposed to the `<Image />` component yet, and there's a risk of causing confusion between it and ``. How about this?

```suggestion
When installed, the [official image integration](#astros-image-integration) will change all image imports to return an object, even when you don't use the integration's `<Image />` component.
```

sarah11918: Yes, that would be a better version! Still not sure myself if that's toooo much, but I would 100 add "when installed".  Open to thoughts on this!

sarah11918: And, still wondering "note/tip" vs caution. Caution seems VERY warn-y, and this is literally the most basic usage of images!  😅 

Could we downplay the aside and _maybe_ add something in Troubleshooting that's optimized for OMG USING BOTH?? and will be a good search hit? Is that a reasonable compromise?

sarah11918: I like that! Since it's only a problem when the image integration is installed, that's where the flames go. Perfect.

But, it should be towards the end of the section, not right up front, so it doesn't look overwhelming when you first get there. Explain how the integration works, warning at the end re: normal img tag. 👍

Jutanium: Thanks for this discussion - check out the new section!

sarah11918:
I think this looks good, @Jutanium! I really like where we ended up with this. Approved when you're happy!

sarah11918 (approved):

PR #2602, opened 2023-02-08
fix: code example for built-in html attributes

mhstrkmp:

- Fix code example in docs

#### Description

- Code example for [Built-in HTML attributes](https://docs.astro.build/en/guides/typescript/#built-in-html-attributes) throws error via linting and `astro check`
- Error: `Spread types may only be created from object types.`

**Current example:**

```astro
---
import type { HTMLAttributes } from 'astro/types'
// use a `type`
type Props = HTMLAttributes<'a'>;
// or extend with an `interface`
interface Props extends HTMLAttributes<'a'> {
  myProp?: boolean;
}
const { href, ...attrs } = Astro.props;
---
<a {href} {...attrs}>
  <slot />
</a>
```

**Change added to docs via this PR in files where section is present:**

```astro
---
import type { HTMLAttributes } from 'astro/types'
// use a `type`
type Props = HTMLAttributes<'a'>;
// or extend with an `interface`
interface Props extends HTMLAttributes<'a'> {
  myProp?: boolean;
}
const { href, ...attrs } = Astro.props;
---
<a href={href} {...attrs}>
  <slot />
</a>
```

### DIFF
--- a/src/pages/en/guides/typescript.mdx
+++ b/src/pages/en/guides/typescript.mdx
@@ -127,7 +127,7 @@ interface Props extends HTMLAttributes<'a'> {
 }
 const { href, ...attrs } = Astro.props;
 ---
-<a {href} {...attrs}>
+<a href={href} {...attrs}>
   <slot />
 </a>
 ```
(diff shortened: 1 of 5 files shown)


Yan-Thomas:
Thanks for this @mhstrkmp, nice catch! And welcome to Team Docs 🙌

PR #2605, opened 2023-02-09
Add hint for rest parameter subdirectory necessity

chimp1nski:
Highlight the necessity to use a rest parameter when using subdirectories to organize collections.

- New or updated content

#### Description

I updated the docs to include a hint to highlight the necessity of a rest parameter when using subdirectories to organize collections. 

This wasn't really obvious to me since the error that was thrown (`Expected "slug" to match "[^\/#\?]+?", but got "foo/bar"`) when starting `dev` and trying to `build` only gave some kind of hint to what went wrong.

The blog starter template helped me clarify. In hindsight it was clear that this was necessary but I thought it would be a good addition to the docs since others might run into the same issue.

Thanks btw. for the amazingly wonderful great work you guys are doing 🥳

### DIFF
--- a/src/pages/en/guides/content-collections.mdx
+++ b/src/pages/en/guides/content-collections.mdx
@@ -345,7 +345,7 @@ const { Content, headings } = await entry.render();
 
 ## Generating Routes from Content
 
-Content collections are stored outside of the `src/pages/` directory. You will need to create a new [dynamic route](/en/core-concepts/routing/#dynamic-routes) to generate HTML pages from your collection entries. Your dynamic route will map the incoming request param (ex: `Astro.params.slug` in `src/pages/blog/[slug].astro`) to fetch the correct entry inside a collection.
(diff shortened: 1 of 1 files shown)


delucis:
Thanks @chimp1nski! I wonder if it actually would make sense to use the rest parameter for the main example instead of in a separate call-out. It's probably a common scenario and I don't think there are major drawbacks.

Paging @bholmesdev @sarah11918 to see what they think.

src/pages/en/guides/content-collections.mdx
```diff
@@ -376,6 +376,10 @@ const { Content } = await entry.render();
 
 This will generate a new page for every entry in the `blog` collection. For example, an entry at `src/content/blog/hello-world.md` will have a slug of `hello-world`, and therefore its final URL will be `/posts/hello-world/`.
 
+:::caution
```
bholmesdev, line 379: Instead of a scary caution, I'd suggest making the basic example use a `[...rest]` parameter as well. Less to worry about, and futureproofs your code for when subdirectories are added!

sarah11918:
Completely agree with everyone here! A wonderful addition, @chimp1nski . Thank you and welcome to Team Docs! 🥳

PR #2613, opened 2023-02-09
Correct `Response` -> "endpoint context" in SSR endpoint introduction

### LINKED ISSUE #2540: [suggestion] Amplify meaning of 'context' object in the case of Server Endpoints
First, note that this issue is solely focused on _docs_ - it is not a remark on the technical implementation of the `context` object used by Astro. In fact, `context` is great, but I found it subtly confusing when coming from the perspective of "other" frameworks.

Screen Shot 2023-02-01 at 7 15 13 PM

The above shot is excellent; clear and terse.

Screen Shot 2023-02-01 at 7 16 07 PM

This one caused confusion for me. To me, it too strongly implies that a server endpoint takes a `Request` - as in the standard Web Api sense. Which is true but...

The point is that server endpoints _do_ take a request but it's been recast into the form of the `context` object which is super convenient, but it's different from the Web Api `Request`.  (And endpoints do return a standard `Response` object.)

**Maybe consider adding a note** in the case of the second screen shot (and perhaps everywhere server endpoints are discussed) that amplifies the point that, in Astro server endpoints,  the Web Api `Request` object has been cast into a form more convenient to use called `context`.

Jutanium:

- New or updated content

#### Description

- Closes #2540

### DIFF
--- a/src/pages/en/guides/server-side-rendering.mdx
+++ b/src/pages/en/guides/server-side-rendering.mdx
@@ -230,5 +230,5 @@ if (!product) {
 
 ### Server Endpoints
 
-A server endpoint, also known as an **API route**, is a special function that lives in a `.js` or `.ts` file within the `src/pages` folder. 
+A server endpoint, also known as an **API route**, is a special function exported from a `.js` or `.ts` file within the `src/pages` folder. 
 The function takes an [endpoint context](/en/reference/api-reference/#endpoint-context) and returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response). A powerful feature of SSR, API routes are able to securely execute code on the server side. To learn more, see our [Endpoints Guide](/en/core-concepts/endpoints/#server-endpoints-api-routes).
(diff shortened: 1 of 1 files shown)


src/pages/en/guides/server-side-rendering.mdx
```diff
@@ -230,4 +230,5 @@ if (!product) {
 
 ### Server Endpoints
 
-A server endpoint, also known as an **API route**, is a `.js` or `.ts` file within the `src/pages` folder that takes a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) and returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response). A powerful feature of SSR, API routes are able to securely execute code on the server side. To learn more, see our [Endpoints Guide](/en/core-concepts/endpoints/#server-endpoints-api-routes).
+A server endpoint, also known as an **API route**, is a special function exported from a `.js` or `.ts` file within the `src/pages` folder.
```
sarah11918, line 233:
```suggestion
A server endpoint, also known as an **API route**, is a special function exported from a `.js` or `.ts` file within the `src/pages/` folder. 
```

sarah11918 (approved):
This is a nice piece of writing, @Jutanium 💅

delucis (approved):
LGTM!

PR #2621, opened 2023-02-10
Feature backend as a service in docs

codercatdev:

- New or updated content
- Translated content

#### Description

- This PR adds a new section to the doc guides allowing additions of popular Backend as a Service examples.

image


Jutanium:
This is an incredible PR, thank you! We'll be working next week on reviewing and revising. That will include working through the AppWrite tutorial and testing the guide, as well as working through copy, organization, and titles. We're super excited to work with you!

sarah11918:
Just wanted to quickly chime in with my own thanks for this amazingly thorough contribution, and to let you know that I've done a "pre-review review" behind the scenes that I'll share with @Jutanium rather than clog up this PR. 

(Mostly it's stuff that relates to specific docs usage that we don't want to end up being death by a thousand paper cuts in this PR, which we'll save to address content etc. For example, we've never used `BaaS` in our docs before, so just generating some rules on when to use the acronym, when to spell out, how to use it as a noun, how to pluralize it... no need for that to happen in here!)

But, stuff *is* happening behind the scenes and just wanted to make sure everyone knew that!  🚀

src/pages/en/guides/baas/appwriteio.mdx
```diff
@@ -0,0 +1,427 @@
...
+```astro
+---
+import { appwriteDatabases } from "../lib/appwrite";
+
+interface BlogPost {
+    title: string,
+    content: string
+}
+
+const entries = await databases.createDocument('[DATABASE_ID]', '[COLLECTION_ID]', '[DOCUMENT_ID]', {});
```
Jutanium, line 150: Won't this just give us empty entries? Shouldn't this be `listDocuments`?

codercatdev: This was just an example of one type of use. Maybe it is misleading though and we should switch it to match the rest of the example more closely. We can keep chatting in the discord channel as well :D

codercatdev:
Using #2702 instead

PR #2646, opened 2023-02-13
Add note about the `allowJs` tsconfig compiler option in `content-collections.mdx`

MoustaphaDev:

- New or updated content

#### Description

To be merged after https://github.com/withastro/astro/pull/6229

Add note about the `allowJs` tsconfig compiler option and how it is necessary to set it to `true` when using a javascript file as the Content Collection's configuration file.


delucis (commented):
My instinct is we might be able to get away with a less prominent sentence for this but I’ll defer to @sarah11918’s better judgement.

i.e. if we “recommend” `.ts`, we can just mention `allowJs` at the same time as mentioning you can also use `.mjs`/`.js`, without needing a full note.

MoustaphaDev:
Yeah I wasn't sure where to put this really, but I felt like putting this note anywhere else would be a good fit. I'll see how I can move it there, in the mean time, I'm happy to take your suggestions 🙂

src/pages/en/guides/content-collections.mdx
```diff
@@ -121,6 +121,22 @@ If you **do not** already extend Astro's `strict` or `strictest` recommended Typ
...
+{
+  // Note: No change needed if you use "astro/tsconfigs/strict" or "astro/tsconfigs/strictest"
+  "extends": "astro/tsconfigs/base",
+  "compilerOptions": {
+    "strictNullChecks": true,
+    "allowJs": true
+  }
+}
+```
+:::
```
delucis, line 138: Thanks again for this @MoustaphaDev! I read through the page thinking about this in context and I actually think this is maybe the right place for this. What do you think about these tweaks?

My goals here were:
- slim this down a bit to be clear and compact
- frame positively: doing X? => do Y

```suggestion
:::note[Not using TypeScript?]
If you use `.js` or `.mjs` files in an Astro project, you can enable IntelliSense and type checking in your editor using a [`jsconfig.json`](https://code.visualstudio.com/docs/languages/jsconfig) or by enabling `allowJs` in your `tsconfig.json`:

```json title="tsconfig.json" ins={5}
{
  "extends": "astro/tsconfigs/base",
  "compilerOptions": {
    "strictNullChecks": true,
    "allowJs": true
  }
}
```
:::
```

Comment (Princesseuh) [reply src/pages/en/guides/content-collections.mdx:138]: I would recommend `allowJs` first over `jsconfig.json`, since the latter isn't standard and not every tool load it / load it properly

Comment (delucis) [reply src/pages/en/guides/content-collections.mdx:138]: Thanks @Princesseuh! In that case I might be tempted to cut it down to just this. No extra action needed if you have `jsconfig.json` but we’re not recommending that path anyway.

````suggestion
:::note[Not using TypeScript?]
If you use `.js` or `.mjs` files in an Astro project, you can enable IntelliSense and type checking in your editor by enabling `allowJs` in your `tsconfig.json`:
```json title="tsconfig.json" ins={5}
{
  "extends": "astro/tsconfigs/base",
  "compilerOptions": {
    "strictNullChecks": true,
    "allowJs": true
  }
}
```
:::
````

Comment (MoustaphaDev) [reply src/pages/en/guides/content-collections.mdx:138]: Sounds good

File: src/pages/en/guides/content-collections.mdx [hunk 2]
```diff
@@ -121,6 +121,19 @@ If you **do not** already extend Astro's `strict` or `strictest` recommended Typ
 }
 ```
 
+:::note[Not using TypeScript?]
```
sarah11918, line 124: I think this note looks good!

The only thing I'm wondering about is how explicitly to make sure people know that the section above (that's not in a note) ALSO (very likely) applies to peeps not using TypeScript.  someone's eyes might just glance here and see the ONE new line, when really, if they're not using TS, they won't have `strict` or `strictest`, so in fact, they'll need EVERYTHING in the code sample in the note.

This *could* be as simple as "If you're not using typescript, you'll ALSO have to..." (not sure whether in fact ALL the lines should be `+` diffs). Again, the worry is just that the "note" format is so visually strong, and says, "Not
 using TypeScript?" that I can totally see this pulling people into skipping the part about strictNullChecks above. So, this "note" might have to cover that, too.

(Which also means, should this in fact be a note, or just a "If you're not writing any TypeScript at all..." paragraph?)

delucis: The code sample here does show the full `tsconfig` you need — `allowJs` and `strictNullChecks` — so maybe it’s enough to simply highlight _both_ of those as new/inserted lines?

Definitely think there’s an argument for paragraph/subhead over note though — happy either way.

src/pages/en/guides/content-collections.mdx
```diff
@@ -121,6 +121,19 @@ If you **do not** already extend Astro's `strict` or `strictest` recommended Typ
...
+If you use `.js` or `.mjs` files in an Astro project, you can enable IntelliSense and type checking in your editor by enabling `allowJs` in your `tsconfig.json`:
+```json title="tsconfig.json" ins={5}
```
sarah11918, line 126:
````suggestion
If you use `.js` or `.mjs` files in an Astro project, you can enable IntelliSense and type checking in your editor by enabling `allowJs` in your `tsconfig.json` in addition to adding `strictNullChecks` as described above:
```json title="tsconfig.json" ins={4-5}
````
Is this overkill?

src/pages/en/guides/content-collections.mdx
MoustaphaDev: @sarah11918 Yeah I think making this note a paragraph will make the previous paragraph look less specific to TypeScript users, we won't even need to highlight both lines, we'll just add to the previous code block

src/pages/en/guides/content-collections.mdx
MoustaphaDev: Sorry, I just saw this suggestion. Do you think this still fits in new the commit I made?

sarah11918: Ooh, I like that! I think this looks good!

MoustaphaDev: Awesome! Just waiting for the ✅s then.

sarah11918:
Just a ping @MoustaphaDev that I think we were all happy with this! There now seems to be a conflict to resolve, but I think as soon as all the checks pass, this is yours to merge! 🥳

MoustaphaDev:
I completely missed your comment @sarah11918 , I'm going to get this merged soon

MoustaphaDev:
Thanks for your reviews @sarah11918 and @delucis!

PR #2664, opened 2023-02-15
SST deploy: no need for create-sst@rc anymore

brettneese:
- New or updated content

#### Description
create-sst@2.0 and sst@2.0  were released last week, so there's no need to use create-sst@rc anymore.

image


Yan-Thomas:
Nice catch @brettneese, thanks!

PR #2671, opened 2023-02-17
docs: use `z.coerce.date()`

tordans:
Assuming a newish version of ZOD is used, this short hand is a lot easier to understand.


tordans:
Update: This is not true. It will not work with my suggested change => closing.

Wondering though… why :-)

delucis:
I think it depends on the format your date is in. I have used `z.coerce.date()` successfully in some projects — but in others even a straight `z.date()` can work as YAML parses certain things as dates.

PR #2672, opened 2023-02-17
Bump vite to 4.1

bluwy:

#### Description

Upgrade vite to 4.1 which should make build less likely to OOM.

Also upgraded vitest so that internally it uses the same vite version.


delucis (approved):
Thanks! Should be good to go once tests pass.

PR #2679, opened 2023-02-18
i18n(de): Move translated file to correct location

sarasate:
- Translated content

#### Description
Move a translated file from the old doc path to the new one.


delucis (approved):
LGTM! Thanks for fixing that @sarasate 💜

PR #2684, opened 2023-02-18
Contentful: Correct code sample

Jutanium:

#### Description

Addresses Feedback Fish:
> https://docs.astro.build/en/guides/cms/contentful/#fetching-data
The code block has two errors in it. The `BlogPost` interface has a `title` and `content` key but towards the bottom `item.fields.name` is used instead of `items.field.title`. Also, where it sets the html to the article, `items.fields.content` is used instead of `item.fields.content`.


sarah11918 (approved):
Thanks for snagging this, @Jutanium !  :rocket:

PR #2685, opened 2023-02-18
Remove "private pages" in Excluding Pages

### LINKED ISSUE #2674: Unclear description of excluded pages
From the docs:

>You can exclude pages, or even whole directories from being built by prefixing their names with an underscore (`_`).
>
>This allows you to create private pages, and also to co-locate tests, utilities, and components with their related pages, preventing them from being built into `.html` files and placed into the `dist/` directory.

Due to the mention of private pages, I assumed this would be a good option to have a dev environment only style guide or components page. However, a quick test of that idea and a subsequent reread of the docs and source shows that this is not as I assumed and `_` prefixed files are completely ignored: 

https://github.com/withastro/astro/blob/54166393be16bc12956d34a78ace1b1c02d69419/packages/astro/src/core/routing/manifest/create.ts#L245-L247

So my question is in what sense are 'private pages' meant here? The example includes `_hidden-page.astro` but to view this page wouldn't it need to be run through the compiler/renderer? If I'm missing some configuration option that enables this I'd love to know.

Jutanium:

- New or updated content

#### Description

- Closes #2674 by replacing "create private pages" (which is misleading, as the pages aren't shown at all) to "temporarily disable pages"


src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -376,7 +376,7 @@ const params = Astro.params;
 
 You can exclude pages, or even whole directories from being built by prefixing their names with an underscore (`_`).
 
-This allows you to create private pages, and also to co-locate tests, utilities, and components with their related pages, preventing them from being built into `.html` files and placed into the `dist/` directory.
+This allows you to temporarily disable pages, and also to co-locate tests, utilities, and components with their related pages, preventing them from being built into `.html` files and placed into the `dist/` directory.
```
sarah11918, line 379:
```suggestion
This allows you to disable page building, and also to co-locate tests, utilities, and components with their related pages, preventing them from being built into `.html` files and placed into the `dist/` directory.
```
I'm not sure temporary is the key thing here? (Although that is one particular use case.) But more accurately, it's that it will be ignored/not built. What about something like "disable page building"?

Jutanium: We've said that in the previous sentence. I was thinking of a use case for preventing building, like "commenting out pages." I'm happy to remove this:
> This allows you to co-locate tests, utilities, and components with their related pages...

sarah11918: Maybe even 
```suggestion
This allows you to do things like temporarily disable pages, and also to co-locate tests, utilities, and components with their related pages, preventing them from being built into `.html` files and placed into the `dist/` directory.
```
I think it felt a little too much like "the only reason why"and not "a reason why"? Don't know why it didn't sit well with me. What do you think about just the modification above?

Jutanium: I don't love the "do things," and we do list another reason why right after the first. It doesn't sound like a list because of the "and also". How does this sound:

```suggestion
You can use this to temporarily disable pages or to co-locate tests, utilities, and components with their related pages, preventing them from being built into `.html` files and placed into the `dist/` directory.
```

sarah11918 (approved):
Thanks, Dan! LGTM!

PR #2686, opened 2023-02-18
i18n(de): Translate imports.mdx

sarasate:

- Translated content

#### Description

- Translate imports.mdx to german


Dezaku:
Habe es mir mal aufgrund der Uhrzeit nur ganz schnell durchgelesen und konnte keine falsch verwendeten Wörter oder Rechtschreibfehler finden.  Ich würde mir die Übersetzung morgen nochmal genauer anschauen falls ich es nicht vergesse, ansonsten könnte ja jemand anderes nochmal drüber schauen.

src/content/docs/de/guides/imports.mdx
```diff
@@ -0,0 +1,267 @@
+---
+title: Statische Inhalte
+description: Erfahre, wie du verschiedene Inhalte in Astro importierst.
+i18nReady: true
+---
+
+Astro unterstützt die meisten statischen Elemente, ohne dass eine Konfiguration erforderlich ist. Du kannst die Anweisung `import` an beliebiger Stelle in deinem Projekt-JavaScript verwenden (einschließlich Astro-Frontmatter) und Astro wird eine optimierte Kopie dieses statischen Assets in deinem endgültigen Build einfügen. `@import` wird auch innerhalb von CSS & `<style>` Tags unterstützt.
```
delucis, line 7: Noticing here we have “statischen Elemente” in the first sentence and then “statischen Assets” in the second — both are “static assets” in the original text. Which of these do you think is better? We should probably be consistent and pick one.

src/content/docs/de/guides/imports.mdx
```diff
@@ -0,0 +1,267 @@
+---
+title: Statische Inhalte
```
delucis, line 2: In English the title is just “Imports” which highlights the action of importing rather than specifically the kind of thing you might import. Maybe “Imports” or “Importieren” can work in German too?

src/content/docs/de/guides/imports.mdx
```diff
@@ -0,0 +1,267 @@
...
+description: Erfahre, wie du verschiedene Inhalte in Astro importierst.
```
delucis, line 3: I think “Inhalte” is OK here but wanted to throw out “Dateitypen” or “Dateiformaten” as alternatives to think about. The English is “content types”.

delucis (commented):
Thank you @sarasate! Did my best to look out for details — let me know if some of my suggestions are off as German isn’t my first language.

src/content/docs/de/guides/imports.mdx
```diff
@@ -0,0 +1,267 @@
...
+---
+<!-- Alle Komponenten anzeigen -->
+{components.map((component) => (
+  <div>
+    <component.default size={24} />
+  </div>
+))}
+```
+
+### Glob Patterns
```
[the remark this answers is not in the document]
sarasate: Thanks @delucis for reviewing. Actually there is no good translation for that term in german. There are 2 potential translations, 'Dateimuster' oder 'Suchmuster', but I would never use any of those as a dev. We usually stick to `Pattern` as the germanized version. Anyway, I would propose to use `Suchmuster` as the heading, immediately followed by `(Glob Pattern)` in parenthesis, and then refer to 'Suchmuster' in the rest of the text. What do you think?

src/content/docs/de/guides/imports.mdx
sarasate: Same problem as with `glob pattern`. Usually we use `asset` in german too, as there is no one-to-one translation with the exact same meaning in development. Nonetheless, in the context here, `Elemente` is fine, so we can go with it.

src/content/docs/de/guides/imports.mdx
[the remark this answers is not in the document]
delucis: If you usually say “Glob Pattern” I think it’s OK to stick to that! I did find this page, but could be an exception: https://docs.oracle.com/de-de/iaas/Content/devops/using/glob-patterns.htm

Do you think it should be hyphenated? “Glob-Pattern”?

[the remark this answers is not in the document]
sarasate: Ok, then we'll stick to Glob-Pattern. Usually there's a hyphen in compound words in german, but it's not mandatory. Guess we can use it, i'll update and use uppercase everywhere.

delucis (approved):
Thanks @sarasate! Appreciate the the time you took to fix this up 🙌

PR #2687, opened 2023-02-19
i18n(zh-cn): Update nav.ts based on reorg-incl-recipes

liruifengv:

- New or updated content
- Translated content

#### Description

- Sorting them
- I would like to translate `recipes` into `方案`.


sarah11918:
Thanks so much for this!

PR #2689, opened 2023-02-19
docs: add testing guide for cypress

### LINKED ISSUE #2168: Expand Testing Guides to include Cypress
Currently, [our testing guide page](https://docs.astro.build/en/guides/testing/) states the following:

> Testing helps you write and maintain working Astro code. Astro supports many popular tools for unit tests, component tests, and end-to-end tests including Jest, Mocha, Jasmine, Cypress and Playwright.

However, the page only has a mention for how to integrate Playwright into E2E testing for Astro apps.

It would be awesome to see similar content for Cypress, since it's explicitly mentioned as well.

bandantonio:

- New or updated content

#### Description

- Closes #2168

### DIFF
--- a/src/content/docs/en/guides/testing.mdx
+++ b/src/content/docs/en/guides/testing.mdx
@@ -111,7 +111,7 @@ npx cypress open
 
 Once the Cypress App is launched, choose **E2E Testing**, then select the browser to be used to run tests.
 
-Once the test run is finished, you should see green marks in the output confirming that your test passed:
+Once the test run is finished, you should see green check marks in the output confirming that your test passed:
 
 ```shell title="Output from npx cypress run"
 Running:  index.cy.ts                                                                     (1 of 1)
@@ -132,7 +132,7 @@ To check that your test really does work, you can change the following line in t
   </body>
 ```
 
-and then run the test again. You should see a red mark in the output confirming that your test failed.
+Then run the test again. You should see a red mark in the output confirming that your test failed.
 :::
 
 ### Next steps
--- a/src/content/docs/en/guides/testing.mdx
+++ b/src/content/docs/en/guides/testing.mdx
@@ -54,7 +54,7 @@ import { defineConfig } from 'cypress'
 export default defineConfig({
   e2e: {
     supportFile: false
-  },
+  }
 })
 ```
 
--- a/src/content/docs/en/guides/testing.mdx
+++ b/src/content/docs/en/guides/testing.mdx
@@ -157,33 +157,33 @@ Alternatively, you can install Playwright within your Astro project using the pa
 
 1. Choose a page to test. We will use the example `index.astro` page below.
 
-```html title="src/pages/index.astro"
----
----
-<html lang="en">
-  <head>
-    <title>Astro is awesome!</title>
-    <meta name="description" content="Pull content from anywhere and serve it fast with Astro's next-gen island architecture." />
-  </head>
-  <body></body>
-</html>
-```
+    ```html title="src/pages/index.astro"
+    ---
+    ---
+    <html lang="en">
+      <head>
+        <title>Astro is awesome!</title>
+        <meta name="description" content="Pull content from anywhere and serve it fast with Astro's next-gen island architecture." />
+      </head>
+      <body></body>
+    </html>
+    ```
 
-2. Create a new folder and add the following test file in `src/test`. Copy and paste the following test into the file to verify that the page meta information is correct. Update the value of the page `<title>` to match the page you are testing. 
+1. Create a new folder and add the following test file in `src/test`. Copy and paste the following test into the file to verify that the page meta information is correct. Update the value of the page `<title>` to match the page you are testing. 
 
-```jsx title="src/test/index.spec.ts" "Astro is awesome!"
-import { test, expect } from '@playwright/test';
+    ```jsx title="src/test/index.spec.ts" "Astro is awesome!"
+    import { test, expect } from '@playwright/test';
 
-test('meta is correct', async ({ page }) => {
-  await page.goto("http://localhost:3000/");
+    test('meta is correct', async ({ page }) => {
+      await page.goto("http://localhost:3000/");
 
-  await expect(page).toHaveTitle('Astro is awesome!');
-});
-```
+      await expect(page).toHaveTitle('Astro is awesome!');
+    });
+    ```
 
-:::tip[set a baseurl]
-You can set [`"baseURL": "http://localhost:3000"`](https://playwright.dev/docs/api/class-testoptions#test-options-base-url) in the `playwright.config.ts` configuration file to use `page.goto("/")` instead of `page.goto("http://localhost:3000/")` for a more convenient URL.
-:::
+    :::tip[set a baseurl]
+    You can set [`"baseURL": "http://localhost:3000"`](https://playwright.dev/docs/api/class-testoptions#test-options-base-url) in the `playwright.config.ts` configuration file to use `page.goto("/")` instead of `page.goto("http://localhost:3000/")` for a more convenient URL.
+    :::
 
 ### Running your Playwright tests
 
@@ -191,14 +191,15 @@ You can run a single test or several tests at once, testing one or multiple brow
 
 1. To run our test from the previous example using the command line, use the `test` command. Optionally, include the file name to run just the single test:
 
-```sh
-npx playwright test index.spec.ts
-```
+    ```sh
+    npx playwright test index.spec.ts
+    ```
 
-2. To see the full HTML Test Report, open it using the following command:
-```sh
-npx playwright show-report
-```
+1. To see the full HTML Test Report, open it using the following command:
+
+    ```sh
+    npx playwright show-report
+    ```
 
 :::tip
 Run your tests against your production code to more closely resemble your live, deployed site. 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,115 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+  <Fragment slot="yarn">
+  ```shell
+  yarn add cypress --dev
+  ```
+  </Fragment>
+</PackageManagerTabs>
+
+### Configuration
+
+In the root of your project, create a `cypress.config.ts` file with the following content:
```
Jutanium, line 56: I get `Unknown file extension ".ts"` with this config setup. It's worth changing to `.js` to avoid this potential error in a getting started guide

```suggestion
In the root of your project, create a `cypress.config.js` file with the following content:
```

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,115 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+        <title>Astro is awesome!</title>
+        <meta name="description" content="Pull content from anywhere and serve it fast with Astro's next-gen island architecture." />
+      </head>
+      <body>
+      <h1>Hello world from Astro</h1>
+      </body>
+    </html>
+    ```
+
+2. Create an `index.spec.js` file in the `cypress/e2e` folder. Copy and paste the following test into the file to verify that the page title is correct. Update the value of the page `<title>` to match the page you are testing.
```
Jutanium, line 86:
```suggestion
2. Create an `index.spec.js` file in the `cypress/e2e` folder. Use the following test to verify that the page title and header are correct.
```

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,115 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+    You can set [`"baseUrl": "http://localhost:3000"`](https://docs.cypress.io/guides/end-to-end-testing/testing-your-app#Step-3-Configure-Cypress) in the `cypress.config.ts` configuration file to use `cy.visit("/")` instead of `page.goto("http://localhost:3000/")` for a more convenient URL.
+    :::
+
+### Running your Cypress tests
+
+You can run a single test or several tests at once, testing one or multiple browsers. By default, Cypress uses a built-in instance of [Electron](https://www.electronjs.org/) to run your tests, and it also [supports](https://docs.cypress.io/guides/guides/launching-browsers) Firefox and Chrome-family browsers (including Edge).
+
+Cypress can be run from the command line or from the Cypress App. The App provides a visual interface for running and debugging your tests.
+
+To run our test from the previous example using the command line, execute the following command:
```
Jutanium, line 107:
```suggestion
First, start the dev server so Cypress can access your live site. To run our test from the previous example using the command line, execute the following command:
```

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,115 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+    });
+    ```
+
+    :::tip[set a baseUrl]
+    You can set [`"baseUrl": "http://localhost:3000"`](https://docs.cypress.io/guides/end-to-end-testing/testing-your-app#Step-3-Configure-Cypress) in the `cypress.config.ts` configuration file to use `cy.visit("/")` instead of `page.goto("http://localhost:3000/")` for a more convenient URL.
+    :::
+
+### Running your Cypress tests
+
+You can run a single test or several tests at once, testing one or multiple browsers. By default, Cypress uses a built-in instance of [Electron](https://www.electronjs.org/) to run your tests, and it also [supports](https://docs.cypress.io/guides/guides/launching-browsers) Firefox and Chrome-family browsers (including Edge).
```
Jutanium, line 103: If we're not explaining _how_ to run several tests at once and how to test in multiple browsers, let's remove this paragraph

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,115 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+npx cypress run
+```
+
+To run the test using the Cypress App, execute the following command:
+
+```shell
+npx cypress open
+```
+
+Once the Cypress App is launched, choose **E2E Testing**, and the browser that will be used to run tests.
```
Jutanium, line 119:
```suggestion
Once the Cypress App is launched, choose **E2E Testing**, then select the browser to be used to run tests.
```

Jutanium (commented):
Thanks for this PR! Super glad to add Cypress to these docs. Added a few edits/comments - let me know if you have any questions!

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,115 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
 See the [Astro + Vitest starter template](https://github.com/withastro/astro/tree/latest/examples/with-vitest) on GitHub.
 
+## Cypress
+
+Cypress is a next generation front end testing tool built for the modern web. Cypress enables you to write all types of tests:
+
+* End-to-end tests
+* Component tests
+* Integration tests
+* Unit tests
```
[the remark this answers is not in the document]
bandantonio: > I can't find either in their docs.

Well, it is written in the [Overview](https://docs.cypress.io/guides/overview/why-cypress#Who-uses-Cypress). Anyway, I don't mind being less specific, this information doesn't change the main purpose of the tool

src/content/docs/en/guides/testing.mdx
bandantonio: > I get `Unknown file extension ".ts"` with this config setup. It's worth changing to `.js` to avoid this potential error in a getting started guide

It's strange, because I've tested my setup multiple times without any issues.

cypress

So, in your case, changing the extension solved the error? I believe there should be an error related to using `import` in `.js` files and [according to the docs](https://docs.cypress.io/guides/references/configuration#Intelligent-Code-Completion), the config should be changed for js:

```js
const { defineConfig } = require('cypress')

module.exports = defineConfig({ ... })
```

Jutanium: Yep, I get this error: 
image
And all I had to do was change to `.js`, no `require` necessary

sarah11918:
Thanks! I will wait for an explicit go-ahead from you before I even START my read-through!

bandantonio:
@sarah11918 Fell free to proceed with your review

src/content/docs/en/guides/testing.mdx
```diff
@@ -6,7 +6,7 @@ i18nReady: true
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
 
-Testing helps you write and maintain working Astro code. Astro supports many popular tools for unit tests, component tests, and end-to-end tests including Jest, Mocha, Jasmine, Cypress and Playwright. You can even install framework-specific testing libraries such as React Testing Library to test your UI framework components.
+Testing helps you write and maintain working Astro code. Astro supports many popular tools for unit tests, component tests, and end-to-end tests including Jest, Mocha, Jasmine, [Cypress](#cypress) and [Playwright](#playwright). You can even install framework-specific testing libraries such as React Testing Library to test your UI framework components.
```
sarah11918, line 9:
```suggestion
Testing helps you write and maintain working Astro code. Astro supports many popular tools for unit tests, component tests, and end-to-end tests including Jest, Mocha, Jasmine, [Cypress](https://cypress.io) and [Playwright](https://playwright.dev). You can even install framework-specific testing libraries such as React Testing Library to test your UI framework components.
```
I think it's more our convention to link to the 3rd party services themselves when we first introduce them. What do we think about this?

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,106 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+export default defineConfig({
+  e2e: {
+    supportFile: false
+  }
+})
+```
+
+### Create your first Cypress test
+
+1. Choose a page to test. We will use the example `index.astro` page below.
```
sarah11918, line 63:
```suggestion
1. Choose a page to test. This example will test the example page `index.astro` below.
```
We don't use "we" in our instructions as per our style guide. When necessary, "you" is OK. Otherwise, we rewrite to avoid addressing any particular person and try to keep things as neutral as possible.

(And yes, I know the example below uses "we".. .and that should also be corrected. So, can't fault you for following what you see in front of you on the same page! Baby improvement steps! Let's at least catch *this* one while we're here... 😅 )

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,106 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+        <title>Astro is awesome!</title>
+        <meta name="description" content="Pull content from anywhere and serve it fast with Astro's next-gen island architecture." />
+      </head>
+      <body>
+      <h1>Hello world from Astro</h1>
+      </body>
+    </html>
+    ```
+
+2. Create an `index.spec.js` file in the `cypress/e2e` folder. Use the following test into the file to verify that the page title and header are correct.
```
sarah11918, line 79:
```suggestion
2. Create an `index.spec.js` file in the `cypress/e2e` folder. Use the following test in the file to verify that the page title and header are correct.
```

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,106 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+```
+
+To run the test using the Cypress App, execute the following command:
+
+```shell
+npx cypress open
+```
+
+Once the Cypress App is launched, choose **E2E Testing**, then select the browser to be used to run tests.
+
```
sarah11918, line 111:
```suggestion
First, start the dev server so Cypress can access your live site. 

To run our test from the previous example using the command line, execute the following command:

```shell
npx cypress run
```

Alternatively, to run the test using the Cypress App, execute the following command:

```shell
npx cypress open
```

Once the Cypress App is launched, choose **E2E Testing**, then select the browser to be used to run tests.

```
Let's separate the generic instructions (start the dev server) from the pair of options for running Cypress.

This *may* even be a candidate for our "Tabs" component, if you felt particularly adventurous! One set of instructions for CLI and one for the App. But, not necessary, and non-blocking!

File: src/content/docs/en/guides/testing.mdx [hunk 11]
```diff
@@ -16,6 +16,106 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+### Next steps
```
sarah11918, line 112: Before next steps, it might be nice to briefly describe what kind of test results/feedback (and where) one should expect to receive.

I'll again put this as non-blocking, but the point of testing is to be able to respond to passing/failing tests! So, I think it would be important for this guide to let someone know what should happen if they've done this correctly, how to see their test results etc. In the example above, what happens when the test passes? Let's tell the reader so they know they did everything correctly!

sarah11918 (commented):
Looking pretty good, @bandantonio and @Jutanium!!

I've left some suggestions (many non-blocking!) after a read through finally of this one myself. Dan, I trust you to merge based on my comments when you both are happy!  🥳

src/content/docs/en/guides/testing.mdx
bandantonio: It's really tempting to make tabs, but currently this is the only place where cli/app separation exists. I believe we can back to restructuring and optimizing this page once we add all the usage examples for test frameworks mentioned at the beginning of the page.

Jutanium (commented):
LGTM with these two changes!

src/content/docs/en/guides/testing.mdx
```diff
@@ -16,6 +16,132 @@ A Vite-native unit test framework with ESM, TypeScript and JSX support powered b
...
+To check that your test really does work, you can change the following line in the `index.astro` file:
+
+ ```astro title="src/pages/index.astro" del={2} ins={3}
+  <body>
+    <h1>Hello world from Astro</h1>
+    <h1>Hello from Astro</h1>
+  </body>
+```
+
+Then run the test again. You should see a red mark in the output confirming that your test failed.
```
sarah11918, line 135:
```suggestion
Then run the test again. You should see a red "X" in the output confirming that your test failed.
```
I'm not sure whether this is actually an X, but I think the very last thing here is to also be explicit about what kind of mark a failing test produces. It's helpful to describe the mark with more than just its colour, considering colourblindness etc.

bandantonio: Yes, that's a good point, I haven't considered it yet.

PR #2690, opened 2023-02-19
Extending `window` and `globalThis`

### LINKED ISSUE #2662: Document how to extend `window`
Sometimes, you want to add a property to the `window` interface. TypeScript doesn't really document how to do this properly, and this situation is more likely to arise in Astro since we support TypeScript directly in script tags.

The solution is to add to `env.d.ts` (or any other `.d.ts` that's in the project scope) a Window interface, like such:

```ts
interface Window {
  helloChris: string;
}
```

The property will then appear in completions and by type-checked with no issues:

image

### Side note

A similar situation happens for `globalThis`, but the way to add property to it is completely different, as you need to do the following:

```ts
export {}
declare global {
  function myFunction(): boolean;
  var myVariable: number;
}
```

image

Jutanium:

- New or updated content

#### Description

- Closes #2662

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -83,6 +83,29 @@ import Layout from '@layouts/Layout.astro';
 }
 ```
 
+## Extending `window` and `globalThis`
+
+You may want to add a property to the global object. You can do this by adding a global declaration to your `env.d.ts` file:
+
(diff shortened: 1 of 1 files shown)


Princesseuh (approved):
Content looks great to me!

sarah11918 (approved):
Another one for the books, Dan!  🥳

PR #2698, opened 2023-02-21
Fix font rendering problem with og images for Korean pages

jcha0713:

#### Description

This pull request resolves a font rendering issue with Open Graph images that are generated for Korean pages.
I discovered that the matching Korean fonts do not exist which caused this error:

Screenshot_2023-02-21_at_10 23 58

I added Noto Sans KR fonts and the result is:

getting-started


Yan-Thomas (approved):
Nice catch @jcha0713, thanks for the fix! I'll let @delucis have a final look, but LGTM!

delucis (approved):
Awesome — thank you for fixing this @jcha0713!

PR #2699, opened 2023-02-21
Specify that WOFF fonts are only accepted for now as a local font

idiglove:

- New or updated content

#### Description

> After tinkering around with Astro for the first time, I tried rendering a local font, and only WOFF formats worked. 
image

### DIFF
--- a/src/content/docs/en/guides/fonts.mdx
+++ b/src/content/docs/en/guides/fonts.mdx
@@ -19,6 +19,10 @@ If you want to add font files to your project, we recommend adding them to your 
 
 Let’s imagine you have a `DistantGalaxy.woff` font file.
 
+:::note
+WOFF formats only are accepted for now. 
+:::
+
 1. Add your font file to `public/fonts/`.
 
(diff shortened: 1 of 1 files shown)


BryceRussell:
I tested this locally with a `.ttf` font from google and it worked for me

A few reasons your font may not have worked are:
- The file is corrupted / does not work as a font
- The `format()` in the `src` property of your `@font-face` does not [match your font file's format](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src#font_formats)
- The `url()` in the `src` property of your `@font-face` does not point to your font file in `/public`
- The `font-family` of your `@font-face` does not match with `font-family` property declared on selectors in your CSS

delucis:
Thanks for this @idiglove! But like @BryceRussell said, other font formats are (or at least should be!) supported. Happy to help figure out what might be going wrong. If it's something on Astro's end, an issue in https://github.com/withastro/astro would be helpful.

Will close this for now — thanks again!

idiglove:
> * match your font file's format

I see, I tried those before. That's strange, I did download the font from google too.
I will have to try again. Thank you :)

PR #2705, opened 2023-02-21
Remark-toc heading disclaimer

Jutanium:

- New or updated content
- 
#### Description

- Addresses Feedback Fish:
> remark-toc  ... other than adding to the config, what else is there??? I do that and it does not show up on my page. This is not clear.

- Adds a note that remark-toc expects a "toc" or "table of contents" heading by default in order to show the table of contents.
- Adds a remark-toc customization to the Customizing a plugin example

### DIFF
--- a/src/content/docs/en/guides/markdown-content.mdx
+++ b/src/content/docs/en/guides/markdown-content.mdx
@@ -497,6 +497,7 @@ export default defineConfig({
 #### Customizing a plugin
 
 In order to customize a plugin, provide an options object after it in a nested array. 
+
 The example below adds the [heading option to the `remarkToc` plugin](https://github.com/remarkjs/remark-toc#optionsheading) to change where the table of contents is placed, and the [`behavior` option to the `rehype-autolink-headings` plugin](https://github.com/rehypejs/rehype-autolink-headings#optionsbehavior) in order to add the anchor tag after the headline text.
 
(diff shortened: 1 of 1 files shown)


Yan-Thomas (approved):
Nice work @Jutanium, LGTM!

src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -494,16 +496,18 @@ export default defineConfig({
 
 #### Customizing a plugin
 
-In order to customize a plugin, provide an options object after it in a nested array. The example below adds the [`behavior` option to the `rehype-autolink-headings` plugin](https://github.com/rehypejs/rehype-autolink-headings#optionsbehavior) in order to add the anchor tag after the headline text.
+In order to customize a plugin, provide an options object after it in a nested array.
```
sarah11918, line 499:
```suggestion
In order to customize a plugin, provide an options object after it in a nested array. 

```
I tend to like a line space between a general instruction/description and showing a specific example. Up to you!

(Aside: if you really did not want a line space, we should discuss/decide then add to our writing guide guidance re: putting one sentence per line with the intention that they still show up in the same paragraph. I'm not actually a fan, as I find it distracting to read/edit the code, despite knowing that it's a "thing" that people do for referencing lines. But, this can be up for discussion!)

Otherwise, LGTM!

PR #2707, opened 2023-02-22
i18n(fr): fix missing translated title in side nav

matdexir:

#### Description
- Added missing french title to the side bar of a page that is already translated.
Basically the page about using CMS content is already translated in french however the side nav bar is still in english.


matdexir:
I think that you could add a CI step to check for this type of minor mistake in the future

sarah11918:
Nicely spotted! Thank you for this @matdexir!

PR #2723, opened 2023-02-23
Update example for Generating Pages

WeaveMN:
Using this example I ran into an issue with the static site build (astro build) where the directly imported Page component was in the wrong position in the generated bundle causing a "Cannot access '$$Page' before initialization" error.  Switching to use the StoryblokComponent resolved the issue.  See example on the storyblok/stpryblok astro Readme: https://github.com/storyblok/storyblok-astro#2-getting-storyblok-stories-and-using-the-storyblok-bridge

- New or updated content

#### Description

- Closes # 
  -  Updates single page code example.

### DIFF
--- a/src/content/docs/en/guides/cms/storyblok.mdx
+++ b/src/content/docs/en/guides/cms/storyblok.mdx
@@ -367,16 +367,20 @@ export default defineConfig({
 
 ### Generating pages
 
-To create a route for a specific `page`, you can fetch its content directly from the Storyblok API and pass it to the `Page` component. Create an `index.astro` file in `src/pages/` to render the `home` page:
+To create a route for a specific `page`, you can fetch its content directly from the Storyblok API and pass it to the `StoryblokComponent` component.  Remember to make sure you have added the `Page` component to your astro.config.mjs.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/cms/storyblok.mdx
```diff
@@ -367,24 +367,28 @@ export default defineConfig({
 
 ### Generating pages
 
-To create a route for a specific `page`, you can fetch its content directly from the Storyblok API and pass it to the `Page` component. Create an `index.astro` file in `src/pages/` to render the `home` page:
+To create a route for a specific `page`, you can fetch its content directly from the Storyblok API and pass it to the `StoryblokComponent` component.  Remember to make sure you have added the `Page` component to your astro.config.mjs.
```
Jutanium, line 370:
```suggestion
To create a route for a specific `page`, you can fetch its content directly from the Storyblok API and pass it to the `StoryblokComponent` component.
```
This reverts a change I made that in retrospect was problematic, so I'm happy with this PR! I don't think we need this line, though, as registering the page component happened in the previous step.

WeaveMN: Looks good!

PR #2740, opened 2023-02-25
i18n(de): navigation route name

sarasate:

- Translated content

#### Description

- Fixed the route name for a renamed translated guide page.


delucis (approved):
Thanks for the follow up @sarasate 🙌

PR #2742, opened 2023-02-26
docs: add missing props for Astro.props

### LINKED ISSUE #2739: Suggested change to docs to improve onboarding
This is a suggestion to improve the TS type checking for users just starting out with Astro (this includes myself).

I found that I was having hidden name referencing issue when trying to rename some of the props in a component.

It was specifically around forgetting the update the name inside the destructured Astro.props object.

I was using this snippet as my base

```
export interface Props {
  to: string;
  bodyText: string;
  class?: string;
}

const { to, bodyText, class: className } = Astro.props as Props;
```

But during changes I was forgetting to update the destructured object - and was only updating the interface object.

After scratching my head as to why no errors were throwing I realised the interface was never attached to the Astro.prop in the tutorial.

Tutorial link: https://docs.astro.build/en/core-concepts/astro-components/

Current code example:
```
---
// src/components/GreetingHeadline.astro
interface Props {
  name: string;
  greeting?: string;
}

const { greeting = "Hello", name } = Astro.props;
---
<h2>{greeting}, {name}!</h2>
```

Suggested code example:
```
---
// src/components/GreetingHeadline.astro
interface Props {
  name: string;
  greeting?: string;
}

const { greeting = "Hello", name } = Astro.props as Props;
---
<h2>{greeting}, {name}!</h2>
```
Saving new users the headache of missing TS errors.

bandantonio:

#### Description

- Closes #2739

### DIFF
--- a/src/content/docs/en/core-concepts/astro-components.mdx
+++ b/src/content/docs/en/core-concepts/astro-components.mdx
@@ -324,7 +324,7 @@ interface Props {
   greeting?: string;
 }
 
-const { greeting = "Hello", name } = Astro.props;
+const { greeting = "Hello", name } = Astro.props as Props;
 ---
 <h2>{greeting}, {name}!</h2>
 ```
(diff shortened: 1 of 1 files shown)


delucis:
Thanks again for this @bandantonio. I haven't heard back on #2739 so I'll close this for now as the `as Props` assertion shouldn't be required when using Astro's language tools.

PR #2743, opened 2023-02-26
Make collection conversion more clear against the existing tutorial

snide:

#### Description

Coming from the blog tutorial, there are some slight changes to the conversion doc to make it 100% clear what needs to be done. Specifically [the new branch references the publish date](https://github.com/withastro/blog-tutorial-demo/blob/content-collections/src/content/config.ts#L6) as an actual `z.date()` object. If users are coming over form the tutorial, they'll have written their dates as `publishDate: 2022-02-03` not `publishDate: "2022-02-03"`. Copying the code as is with the transformation will actually cause a build error.

I left the old transformation in there as a comment, because it's difficult to know what a user would actually use, and seeing the transformation is helpful to show that types can be transformed.

With the RSS feed, it's important to note that the function itself now needs to be async, so I've added a highlight to that line and added to the comment.

FANTASTIC documentation. One of the best I've seen on the web. Thank you for being awesome.

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -233,10 +233,10 @@ defineCollection({
     tags: z.array(z.string()),
     // An optional frontmatter property. Very common!
     footnote: z.string().optional(),
-    // Dates in frontmatter, written without "" around them are read in as actual date objects
-    // Alternatively, you could transform a string date to a real date with
-    // z.string().transform((str) => new Date(str))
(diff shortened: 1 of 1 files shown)


Jutanium (commented):
Thanks for this - left some comments!

snide:
@Jutanium Your reasoning made sense to me. Noticed some typos you added so added another commit on top. Thanks for the review.

Jutanium:
@snide It wasn't a typo, but I like what you did with it better! Let's get this merged!

PR #2744, opened 2023-02-26
docs: fix unescaped string in title

### LINKED ISSUE #2735: Use an escape character in the `markdown-content` example
### Describe the problem :
It's not really a problem, but, I prefer to warn.
The problem is on [this page](https://docs.astro.build/en/guides/markdown-content/#importing-markdown). It is not really visible to users.
But, if you pay attention, you can see that the following string does not have an escape character to avoid closing the string. Nevertheless, the UI is not impacted :
```md
title: 'The best post I've ever written'
```

### What this issue changes for users :
Absolutely nothing, except a loss of a few seconds of users copying the code.

### Screenshots :

You can see the problem in image : 
Missing escape char
Missing escape char

### A proposed solution :
1. This can be solve by changing the single quotation marks to double quotation marks :
```md
title: "The best post I've ever written"
```
2. Or by changing the single quotation mark to an escape character :
```md
title: 'The best post I\'ve ever written'
```

### The exact links to this "problem" in the file :
- https://github.com/withastro/docs/blob/main/src/content/docs/en/guides/markdown-content.mdx?plain=1#L283
- https://github.com/withastro/docs/blob/main/src/content/docs/en/guides/markdown-content.mdx?plain=1#L310

bandantonio:

#### Description

- Closes #2735

### DIFF
--- a/src/content/docs/en/guides/markdown-content.mdx
+++ b/src/content/docs/en/guides/markdown-content.mdx
@@ -280,7 +280,7 @@ When you import Markdown and MDX files in an Astro component, you get an object 
 
 ```md title="/src/pages/posts/great-post.md"
 ---
-title: 'The best post I've ever written'
+title: 'The greatest post of all time'
 author: 'Ben'
 ---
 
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks for fixing that!

PR #2758, opened 2023-02-28
[i18nIgnore] Add new subpage handling logic

Yan-Thomas:

#### Description

Adds a new subpage handling logic to the sidebar, based on the frontmatter `type` field. I added two new types to make this work properly (`install` and `error`) to assure the main Install and Error pages are being properly highlighted.

Note: I found out the `description` field for the sharing-state translations was missing, so I added the original one in English to avoid errors for now. After the merge I will warn translators about this.

### DIFF
--- a/src/util/getPageCategory.ts
+++ b/src/util/getPageCategory.ts
@@ -16,7 +16,7 @@ const categories = [
  * @param url URL for the current page.
  * @returns The category for the current page as used by Algolia DocSearch to group search results.
  */
-export function getPageCategory(url: URL) {
+export function getPageCategory(url: { pathname: string }) {
 	const langAgnosticPath = url.pathname.replace(/\/\w\w(-\w\w)?\//, '');
 	for (const [path, label] of categories) {
 		if (langAgnosticPath.startsWith(path)) return label;
--- a/scripts/error-docgen.mjs
+++ b/scripts/error-docgen.mjs
@@ -205,7 +205,6 @@ function getErrorReferenceEntryHeader(errorTitle) {
 title: ${errorTitle}
 i18nReady: true
 githubURL: https://github.com/withastro/astro/blob/main/packages/astro/src/core/errors/errors-data.ts
-type: error
 ---
 import DontEditWarning from '~/components/DontEditWarning.astro'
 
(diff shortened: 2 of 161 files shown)


delucis:
Reverted your changes and made a refactor with smaller scope. Thanks again for looking at this @Yan-Thomas, I’d love a double check that what I’ve done makes sense.

One thing that might still need doing is making sure all the non-English recipes have `type: recipe` and a `description` in frontmatter.

src/util/getPageCategory.ts
```diff
@@ -16,7 +16,7 @@ const categories = [
  * @param url URL for the current page.
  * @returns The category for the current page as used by Algolia DocSearch to group search results.
  */
-export function getPageCategory(url: URL) {
+export function getPageCategory(url: { pathname: string }) {
 	const langAgnosticPath = url.pathname.replace(/\/\w\w(-\w\w)?\//, '');
 	for (const [path, label] of categories) {
 		if (langAgnosticPath.startsWith(path)) return label;
```
Yan-Thomas, line 22: I'm not sure if this is a dev-only behavior, but the switch to the Reference tab isn't working for the Errors pages. The "langAgnosticPath" is starting with `C:/` for me, that's why it is unable to match with `path`. 

Considering we add the final `/` to the categories I guess our chances of ending up causing a route conflict are small, so a fast fix for this would be to change the string method to `includes` instead:

```suggestion
		if (langAgnosticPath.includes(path)) return label;
```

Yan-Thomas (commented):
Thanks @delucis, I added the missing fields to translations and also found a small issue: when you're in an individual error page, the actual highlighted tab is "Learn". Added another comment to the specific code with more context and a possible fix.

src/util/getPageCategory.ts
delucis: Oh I think this is because of one place that uses `import.meta.url`!

delucis: Pushed a small change that I think should fix this. (Not 100% sure because I think it was a Windows-only thing maybe.)

Yan-Thomas:
Thanks @delucis, this fixed the issue! Let me know when you think this is good to be merged into Sarah's PR.

delucis (approved):

PR #2762, opened 2023-03-01
Minor fix to file extensions in cypress section

adrian-burkhart:

#### Description

- `npm run cypress` failed for me, because the spec file the guide had me create was named `index.spec.ts`, but cypress was looking for `index.cy.ts`. Therefore, I updated the guide to state that you should create an `index.cy.ts` file instead.
- The guide was using `.ts` extensions in some places, and `.js` extensions in others. Therefore, I changed the `.ts` extensions to `.js` to make it more consistent.

### DIFF
--- a/src/content/docs/en/guides/testing.mdx
+++ b/src/content/docs/en/guides/testing.mdx
@@ -76,9 +76,9 @@ export default defineConfig({
     </html>
     ```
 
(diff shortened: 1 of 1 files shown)


Jutanium:
Great catch, thank you!

PR #2765, opened 2023-03-01
Document new getStaticPaths type helpers

matthewp:

- New or updated content

#### Description

- Documents https://github.com/withastro/astro/pull/6150
- This is a 2.1 feature


src/content/docs/en/guides/typescript.mdx
```diff
@@ -179,6 +179,56 @@ type MyAttributes = astroHTML.JSX.ImgHTMLAttributes;
...
+        propB: 22
+      }
+    }
+  });
+}
+
+type Props = InferGetStaticPropsType<typeof getStaticPaths>;
+// ^? { propA: boolean; propB: number; }
+---
+```
```
delucis, line 230: Do you think it would make sense to combine these two examples? They share a lot and you’ll often use both together.

I’ve also tweaked the example to use the real-world `getCollection()` call instead of a hypothetical `fetchPosts()` and moved to show the types on `Astro.props`/`params` to help make it clear that setting types here allows Astro to infer the type of those.

````suggestion
You can get the type of [`Astro.params`](/en/reference/api-reference/#astroparams) with `InferGetStaticParamsType` and the type of [`Astro.props`](/en/reference/api-reference/#astroprops) with `InferGetStaticPropsType`:

```astro title="src/pages/posts/[...slug].astro" {2,14-15}
---
import { InferGetStaticParamsType, InferGetStaticPropsType } from 'astro';

  return posts.map((post) => {
    return {
      params: { slug: post.slug },
      props: { draft: post.data.draft, title: post.data.title },
    };
  });
}

type Params = InferGetStaticParamsType<typeof getStaticPaths>;
type Props = InferGetStaticPropsType<typeof getStaticPaths>;

//               ^? { slug: string; }
//                ^? { draft: boolean; title: string; }
---
```
````

File: src/content/docs/en/guides/typescript.mdx [hunk 2]
```diff
@@ -179,6 +179,56 @@ type MyAttributes = astroHTML.JSX.ImgHTMLAttributes;
 ```
 :::
 
+### getStaticPaths
```
delucis, line 182: Not quite happy with this but two aims here:
1) Small style nit for `getStaticPaths()` being in backticks and including parens.
2) I’m trying to communicate that this is not the main `getStaticPaths()` docs in the heading — important for search results — while not making this too verbose in context on the page

```suggestion
### Infer `getStaticPaths()` types
```

delucis (commented):
Thanks @matthewp! Left a couple of suggestions.

src/content/docs/en/guides/typescript.mdx
matthewp: yep! like this.

delucis (approved):
I think this is ready when 2.1 is released!

Issue #2726, opened 2023-02-24
Flicking issue while scrolling on some pages


ansidev:
Steps to reproduce:
1. Prerequisites:
- Chrome (110.0.5481.177 (Official Build) (x86_64)) on macOS (13.2.1 (22D68)).
- No extension is enabled (recommended).
2. Open an example test URL: https://docs.astro.build/en/guides/content-collections/ (I guess this issue occurs on long pages)
3. Scroll up and down.

The address bar or bookmark bar (if enabled) will be flicking.

delucis:
Thanks for the report @ansidev! Any chance you could share a screen capture of this behaviour? [Kap](https://getkap.co/) or QuickTime Player are both good options for this on macOS.

I’m testing on the same Chrome 110.0.5481.177 and macOS 13.2.1 combo but don’t see anything wrong 🤔

ansidev:
Other browsers (macOS Ventura 13.2.1 (22D68))
- MS Edge Stable 110.0.1587.50 (Official build) (x86_64): OK.
- Firefox 110.0 64-bit: OK.

delucis:
Hi @ansidev! Would you be able to test if this deploy preview is any better for you?

https://deploy-preview-2769--astro-docs-2.netlify.app/en/guides/content-collections

ansidev:
> https://deploy-preview-2769--astro-docs-2.netlify.app/en/guides/content-collections

@delucis LGTM!

delucis:
Hooray! Happy that fixed it 🥳

Thanks again for the issue.

PR #2769, opened 2023-03-01
Replace `drop-shadow` with `box-shadow` in `<CodeSnippet>`

### LINKED ISSUE #1790: Safari on macOS: Poor rendering performance
**Problem**
Browsing and scrolling through the docs is very slow and laggy. High CPU usage. Especially when code blocks come into view.

- Intel MacBook Pro, 8 GB RAM
- macOS 12.4 
- Safari 
 
**Screen recording**
You can see poor scrolling performance, invisible content, flickering content.

- excessive CPU usage
- excessive memory usage, up to 300 MB graphics memory when scrolling

https://user-images.githubusercontent.com/183989/195606103-14cebfb3-9c73-41f6-b34e-347eb41aa435.mov

**Troubleshooting**
- Removing the filter seems to resolve the issue

image

You probably won’t see this on a M1 or M2. The issue is running the site on an older Macbook on a 5K display. But it’s a sign that there’s a potential issue for anyone with older hardware. With my setup I can spot performance issues in any three.js site, Figma, etc. 😄 But it doesn’t mean that everything is slow. It’s just a sign that there might be excessive graphics memory used (e.g. massive sized layers)

delucis:

#### Description

- Closes #1790 (and maybe #2726?)

- Changes how the glow around our code samples is rendered. Previously this used the `drop-shadow` CSS filter which has poor performance in some browsers. I’ve replaced this with several `box-shadow`s tweaked to try and replicate roughly the same effect. This isn’t quite possible but hopefully folks agree the difference is small enough to be OK for the performance gain.

| Before | After |
|---|---|
| image | image |
| image | image |

### DIFF
--- a/src/components/CodeSnippet/CodeSnippet.astro
+++ b/src/components/CodeSnippet/CodeSnippet.astro
@@ -143,7 +143,6 @@ function parseMarkingDefinition(serializedArr: string, parts: RegExp, parseError
 <style lang="scss" is:global>
 	.code-snippet {
 		--glow-border: 1px solid var(--theme-glow-highlight);
-		filter: drop-shadow(0 0 0.3rem var(--theme-glow-diffuse));
 
 		.header,
 		pre {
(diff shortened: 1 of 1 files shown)


sarah11918:
Will let Yan weigh in on the code, but the effect looks fine on Sarah's Evil Devices!

Yan-Thomas (commented):
Can't confirm if fixes the Safari performance issue, but LGTM!

PR #2773, opened 2023-03-02
content collection SSR code snippet update

60fov:
- New or updated content

#### Description
- change the content collection guide's code snippet on SSR to handle TS errors.

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -396,13 +396,20 @@ If you are building a dynamic website (using Astro's SSR support), you are not e
 
 ```astro
 ---
-// src/pages/posts/[...slug].astro
 import { getEntryBySlug } from 'astro:content';
 // 1. Get the slug from the incoming server request
 const { slug } = Astro.params;
-// 2. Query for the entry directly using the request slug
+// 2. check if the slug identifier matches the filename
+if (slug === undefined) {
+    throw new Error("slug/filename mismatch");
+}
+// 3. Query for the entry directly using the request slug
 const entry = await getEntryBySlug('blog', slug);
-// 3. (Optional) render the entry to HTML in the template
+// 4. redirect if the entry does not exist
+if (entry === undefined) {
+    return Astro.redirect("/404");
+}
+// 5. (Optional) render the entry to HTML in the template
 const { Content } = await entry.render();
 ---
 <h1>{entry.data.title}</h1>


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -396,13 +396,20 @@ If you are building a dynamic website (using Astro's SSR support), you are not e
...
+// 3. Query for the entry directly using the request slug
 const entry = await getEntryBySlug('blog', slug);
-// 3. (Optional) render the entry to HTML in the template
+// 4. redirect if the entry does not exist
+if (entry === undefined) {
+    return Astro.redirect("/404");
+}
+// 5. (Optional) render the entry to HTML in the template
 const { Content } = await entry.render();
 ---
```
Jutanium, line 414: It's not that there's a "mismatch"; the slug simply wasn't provided. They accessed `/posts/` instead of `/posts/slug`. Let's make that clear and add the filename back:

```suggestion
---
// src/pages/posts/[...slug].astro
import { getEntryBySlug } from "astro:content";
// 1. Get the slug from the incoming server request
const { slug } = Astro.params;
if (slug === undefined) {
	throw new Error("Slug is required");
}
// 2. Query for the entry directly using the request slug
const entry = await getEntryBySlug("blog", slug);
// 3. Redirect if the entry does not exist
if (entry === undefined) {
	return Astro.redirect("/404");
}
// 4. (Optional) render the entry to HTML in the template
const { Content } = await entry.render();
---
```

60fov: the identifier for the variable from astro.params does not have to be `slug` but does have to match the filename
e.g.
```js
/* filename: [foo].astro */
const { foo } = Astro.params;
```
but maybe I misunderstood you since "_They accessed `/posts/` instead of `/posts/slug`_" doesn't make sense to me in this context.
regardless the difference is negligible imo. the only improvements beyond this would be a better error message.

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -397,16 +397,21 @@ If you are building a dynamic website (using Astro's SSR support), you are not e
...
+}
 // 2. Query for the entry directly using the request slug
-const entry = await getEntryBySlug('blog', slug);
-// 3. (Optional) render the entry to HTML in the template
+const entry = await getEntryBySlug("blog", slug);
+// 3. Redirect if the entry does not exist
+if (entry === undefined) {
+	return Astro.redirect("/404");
+}
+// 4. (Optional) render the entry to HTML in the template
```
Jutanium, line 412:
```suggestion
// 4. (Optional) Render the entry to HTML in the template
```

Yan-Thomas:
Thanks for this @60fov, LGTM! @Jutanium, feel free to merge it when you think it's ready 🙌

PR #2783, opened 2023-03-03
Upgrade astro to 2.0.17

bluwy:

#### Description

Upgrade astro to 2.0.17 to improve build speed (I hope this works)

EDIT: Looks like this bring down the build times from 355s to 343s only (on Netlify) 😦 Maybe better than nothing


delucis (approved):
LGTM! Clicked around the deploy preview and things seem to be working as expected.

PR #2785, opened 2023-03-03
Remove `--max_old_space_size=4096` from CI

delucis:

- Something else!

#### Description

- Removes our 4GiB memory limit when building the site in CI
- Mainly a test run to see if @bluwy’s performance improvements are enough to allow this — and if so, if there’s any performance benefit to removing the manual limit


delucis:
Confirmed this flag is still needed in CI for now 👍

PR #2550, opened 2023-02-02
[reorg] new Recipes section; Basics streamlined

sarah11918:
Initial reorganization of the sidebar! :partying_face: 

tl/dr: 
- introduces a new Recipes section in sidebar for "guide landing pages"
- new "All Recipes" page as a landing page for some existing examples/walkthroughs
- shuffles pages in sidebar: fewer pages in basic (moved to guides); fewer pages in guides (moved to recipes)

This PR attempts to make the content of the Learn sidebar sections "a good length" so as not to be overwhelming.

___Getting Started:___
~~The CLI reference page has been renamed and moved here. Also, the introduction has been re-written for a "getting started" audience to introduce the `astro` commands they will use right from the beginning, and explains the `package.json` scripts that are included after running`create-astro` and how (and why) to use them.~~
- **OUT OF SCOPE**: CLI Reference page will remain where it is, but we know we want to have "friendlier CLI content earlier, in Learn somewhere, while still being able to leave the techical descriptions of each command for Reference.

___Basics:___
- Only a few pages kept here as truly "basics" you "need to know"
- Part of the idea is to make this section look "easy to manage/get through"
- **Big Change**: Astro components page - about half of the content has been removed:
   - New page "Astro Syntax" for the entire JSX-like expressions material
   - Script and style sections removed entirely as these are covered in their own pages
- **OUT OF SCOPE** : rewriting Markdown/MDX to remove MDX content and KISS; re-editing Routing page to incorporate more of SSR basics/examples

___Guides:___
- Some entire pages removed from sidebar (file not changed) and now linked to from "All Recipes"
- Some pages had their long, specific examples removed and into new pages in `/recipes/` for better discoverability

___Recipes:___
- New section header added (also cc'd i18n-gang to ask them to contribute translations for "recipes" and "all recipes" in their languages and to PR *this branch*. To date, two have done so.
- Moved all fancy/magic/landing pages HERE: CMS guides, Deploy Guides, Migrate to Astro guides, Integrations landing page
- Created an "All Recipes" page (is it really all, if those other pages aren't here? :sweat_smile: ) which will be a landing page for longer, walk-through, guided examples so they don't have to overwhelm the other content pages. In some cases, entire pages that seem to be "how to do something" (e.g. fonts) have been moved here.
- Edited our sidebar logic so that "All Recipes" in the sidebar is highlighted when viewing a recipe *that is located in `/recipes/`*.
[...]

PR #2795, opened 2023-03-06
Add a recipe for HTML forms and a recipe for JS forms

Jutanium:
To be merged after #2550

- New or updated content

#### Description

These are the two forms recipes from #2550, in their own PR. The first ex

### DIFF
--- a/src/content/docs/en/recipes/build-forms-api.mdx
+++ b/src/content/docs/en/recipes/build-forms-api.mdx
@@ -1,0 +1,401 @@
+---
+title: Build Forms With API Routes
+description: Learn how to use JavaScript to send form submissions to an API Route
+i18nReady: false
+type: recipe
+---
+import UIFrameworkTabs from "~/components/tabs/UIFrameworkTabs.astro";
+import PackageManagerTabs from "~/components/tabs/PackageManagerTabs.astro"
+
+An HTML form causes the browser to refresh the page or navigate to a new one. To send form data to an API endpoint instead, you must intercept the form submission using JavaScript. 
+
+This recipe shows you how to send form data to an API endpoint and handle that data.
+
+## Prerequisites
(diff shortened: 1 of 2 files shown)


src/content/docs/en/recipes/build-forms-api.mdx
```diff
@@ -0,0 +1,396 @@
...
+        ```
+      </Fragment>
+
+    </UIFrameworkTabs>
+
+  2. Create the `post` API endpoint that will receive the form data. Use `request.formData()` to process it. Be sure to validate the form values before you use them. 
+  
+        This example sends a JSON object with a message back to the client.
+
+      ```ts title="src/api/feedback.ts" "request.formData()" "post"
```
Yan-Thomas, line 140: The API route should be inside pages, right?
```suggestion
      ```ts title="src/pages/api/feedback.ts" "request.formData()" "post"
```

src/content/docs/en/recipes/build-forms-api.mdx
```diff
@@ -0,0 +1,396 @@
...
+        if (!name || !email || !message) {
+          return new Response(
+            JSON.stringify({
+              message: "Missing required fields",
+            }),
+            { status: 400 }
+          );
+        }
+        // Do something with the data, then return a success response
+        return new Response({ message: "Submitted!"}, { status: 200 });
```
kevinzunigacuellar, line 158: the object response needs to be a string 😉
```suggestion
        return new Response(
          JSON.stringify({
            message: "Success!"
          }),
          { status: 200 }
        );
```

The response needs to be stringified

kevinzunigacuellar (commented):
🧑‍🍳

src/content/docs/en/recipes/build-forms-api.mdx
Jutanium: Thanks!

PR #2799, opened 2023-03-07
Update GraphQL data fetching example

### LINKED ISSUE #2775: GraphQL data fetching example is broken
Got this report via Feedback Fish:

> Hey, just to let you know that the code example in the GraphQL queries section of the docs here: https://docs.astro.build/en/guides/data-fetching/#graphql-queries does not work.
>
> I wanted to reference it to play with GraphQL in Astro, but when I copy/pasted the code to test locally, it failed telling me the response wasn't valid json. So I went to the URL of the sample data at https://graphql-weather-api.herokuapp.com and it seems that's the problem. The app is down, I daresay because Heroku has axed all those lovely hobby tier apps and maybe this was a victim of that?

We should probably find an alternative dummy API we can use instead.

sasoria:

- New or updated content

#### Description

- Closes #2775

This PR updates the graphql example to a working graphql api, in this case [swapi-graphql](https://github.com/graphql/swapi-graphql) and tries to reuse the syntax from the old example (i.e variables and arguments). 

Sadly, I don't have the capacity to translate everything to other languages, so I've kept it in english. I'm using the name `Film.astro` as it relates to the query, but I've seen that `Movies` has been used in som examples above this one. Let me know if there is anything you would like me to change 😊

### DIFF
--- a/src/content/docs/de/guides/data-fetching.mdx
+++ b/src/content/docs/de/guides/data-fetching.mdx
@@ -95,6 +95,7 @@ const response = await fetch("https://graphql-weather-api.herokuapp.com",
   });
 const json = await response.json();
 const weather = json.data;
+
 ---
 <h1>Abruf des Wetters zur Erstellungszeit</h1>
 <h2>{weather.getCityByName.name}, {weather.getCityByName.country}</h2>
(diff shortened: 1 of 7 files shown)


delucis (changes requested):
@sasoria Thanks for picking this up! Appreciate it.

In general we prefer to [just update the English pages](https://github.com/withastro/docs/blob/main/CONTRIBUTING.md#making-prs-pull-requests) and let our translators take care of updating their languages as they’ll often catch nuances we miss. Could you revert the changes to non-English pages and then we can focus on getting the English changes spot on?

sasoria:
Thanks for pointing me in the right direction. I've reverted the changes on the other languages as requested.

Jutanium (approved):

delucis (approved):

PR #2809, opened 2023-03-08
Experimental assets: document how to use Sharp

Jutanium:

- New or updated content

#### Description

- Corrects the code sample in image-service-reference
- Adds a "Using sharp" section to the experimental assets guide that explains how to install and use sharp


sarah11918 (approved):
Thanks for this @Jutanium! Wonderful addition!

PR #2816, opened 2023-03-09
[DRAFT] `getCollection` perf improvements when called multiple times

bholmesdev:

#### Description

@delucis Found that calling `getCollection('docs')` in multiple places noticeably slowed builds over centralizing to a since `getCollection()` call. This branch experiments with a caching layer in core for prod builds to get the same benefits.


delucis (commented):
Ooh, very interesting! Looks like it’s working right?

Funnily enough, I’ve actually really enjoyed the `import { pages } from '~/content'` pattern we were pushed into, but happy to see we might be able to help ensure that performance for everyone, without them needing to hunt for that optimisation. 💜

PR #2838, opened 2023-03-13
Fix compiler issues url

liruifengv:

- New or updated content
#### Description

- Fix compiler issues url


Yan-Thomas:
Thanks for this @liruifengv, though the Error Pages are generated from our `errors-data.ts` file in the main astro repo, could you please make a PR to it fixing the link? Here's the file and line you will find what you need to change:
https://github.com/withastro/astro/blob/main/packages/astro/src/core/errors/errors-data.ts#L31

PR #2839, opened 2023-03-13
Feature: copy code blocks to clipboard

### LINKED ISSUE #2643: Copy to clipboard feature for CodeSnippets
I've recently gone through the tutorial and believe it will benefit to have a copy to clipboard feature for the `CodeSnippet` component.

e.g. On the [Write your first Markdown blog post](https://docs.astro.build/en/tutorial/2-pages/2/) part you have to copy 3 .md snippets.

I'm willing to create a PR for this feature! 😄

kyonlee:

- Something else!

#### Description

- Closes #2643
- Adds a copy button to code blocks, that when clicked copies the contents in the code blocks to the clipboard.
CopyButton


ElianCodes (commented):
Awesome feature!

Not a fan of outputting the copied result to an `alert` tho. I think it would be better to popup some kind of toast notification.

sarah11918:
What a great contribution @kyonlee .. especially a first time one!  :partying_face: So many people have been asking for this, and as @delucis said, we are awaiting a new package from one of our maintainers that will add this. But happy to get something in the docs sooner, if we can!  :raised_hands:

I have a couple of quick comments:
- I think we would probably prefer a universal "copy" icon instead of the word copy, since we do translate this site into so many languages. That would be one fewer translation for us to manage! :sweat_smile: 
- I noticed that the copy only appears on hover, in the upper right corner. This meansit also quickly disappears as I move the cursor away. And, if I'm just scrolling down the page in the middle of the codeblock, I don't even know that copying is a possibility at all because I never see it. I would think a static indicator that the block could be copied would be even more helpful!
- 
Would love for others to test this out and leave feedback, too!

kyonlee:
Thanks @sarah11918 and @ElianCodes!

Wasn't sure where the discussion should belong 😅
But as mentioned in the issue I opened.

The alerts currently in place were just for testing and will be removed.

Things to consider:

* The icon of the button
  * Any preferred icon library to be used?
* Have code blocks to opt into the copy feature - like how it's currently done for code markings

Any other feedback is welcomed!

sarah11918:
> The icon of the button - Any preferred icon library to be used?

I think this is a question for @delucis ! I don't know whether we use any particular icon set now.

> Have code blocks to opt into the copy feature - like how it's currently done for code markings

I think maybe for v1 of this, just make EVERY code sample have it? It's potentially a LOT of updating work to add this in after the fact to all our existing code samples. And, I think I'm OK if some snippets that don't make sense to be copied have a copy button? Will also ask @delucis  and @Yan-Thomas  what they think about that.

Jutanium:
@sarah11918 How about a way to opt _out_? We could have it as part of the ``` string

sarah11918:
I think opt OUT makes a lot of sense, if that's easy to implement, @kyonlee -- if that's harder (which it might be?) then I would say let's include it by default for now, but I do feel pretty strongly that it needs to be visible at all times, not just on hover.

delucis:
We’ve used Octicons in general so far but I don’t love their copy icon, so I’d suggest using this one from Material Icons:

content_copy_FILL0_wght400_GRAD0_opsz48

We will still need this to have an accessible label, so will want “copy” for that and make it possible to translate it for other languages.

I agree with @sarah11918 that we can add opt-out functionality later, especially as we have other plans for changes to our code block infrastructure.

Another note: currently you can’t navigate to the copy button with the tab key, we should make sure this is possible. Simplest way is probably just to have the copy button always visible (dimmed perhaps except on hover/focus?). That would also address Sarah’s concern about not having a hint that copying is possible (especially important for touch devices).

And yes, definitely better not to use `alert()` to confirm the copy. Maybe we can take inspiration from the copy button on https://astro.build/? Source code is here: https://github.com/withastro/astro.build/blob/a264601e73bd148df3edcf92c8d07fce78105622/src/components/CodeBlock.astro#L14-L38

kyonlee:
@delucis I've updated the button to use the icon that you have provided and have set an accessible label and title to the button.

@sarah11918 I have gone with the always showing the button dimmed except on hover/focus.

I've also removed the alerts in favor of the button's color change for success/error. (Added the "Copied" tooltip wasn't as trivial as I initially thought 😅)

Please check it out and let me know if you have further feedback.

src/components/CodeSnippet/CodeSnippet.astro
```diff
@@ -318,3 +352,26 @@ function parseMarkingDefinition(serializedArr: string, parts: RegExp, parseError
 	<figcaption class="header">{titleHtml && <span class="title" set:html={titleHtml} />}</figcaption>
 	<Fragment set:html={codeSnippetHtml} />
 </figure>
+<script>
+	// Find all buttons with the \`copy-button\` class on page.
+	const buttons = document.querySelectorAll('button.copy-button') as NodeListOf<HTMLButtonElement>;
+	
+	// Handle clicks on each button.
+	buttons.forEach((button) => {
```
delucis, line 360: The “copied” pop-up like in the implementation I linked to is important because it allows us to use an [ARIA Live Region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) to announce the successful action to assistive technologies like screen readers.

Here’s an example of how VoiceOver on macOS reads the interaction on the https://astro.build homepage (no sound but you can see what is read out in the box on the right):

https://user-images.githubusercontent.com/357379/227250731-ea585c0a-dccc-43f2-aa3b-dd65e6ec12b2.mp4

src/components/CodeSnippet/copy-button.ts
```diff
@@ -0,0 +1,45 @@
...
+
+		this.code = this.code.replace(/&amp;/g, "&")
+			.replace(/&lt;/g, "<")
+			.replace(/&gt;/g, ">")
+			.replace(/&quot;/g, "\"")
+			.replace(/&#39;/g, "'");  
+	}
+
+	renderToHtml() {
+		return `<button class="copy-button" title="Copy to clipboard" aria-label="Copy" value="${encodeURIComponent(this.code)}"><svg xmlns="http://www.w3.org/2000/svg" height="24" width="24"><g transform="scale(0.5 0.5)"><path d="M15 37.95q-1.25 0-2.125-.875T12 34.95v-28q0-1.25.875-2.125T15 3.95h22q1.25 0 2.125.875T40 6.95v28q0 1.25-.875 2.125T37 37.95Zm0-3h22v-28H15v28Zm-6 9q-1.25 0-2.125-.875T6 40.95V12.3q0-.65.425-1.075Q6.85 10.8 7.5 10.8q.65 0 1.075.425Q9 11.65 9 12.3v28.65h22.2q.65 0 1.075.425.425.425.425 1.075 0 .65-.425 1.075-.425.425-1.075.425Zm6-37v28-28Z" fill="currentcolor"/></g></svg></button>`;
```
delucis, line 43: We need to be able to translate the “Copy to clipboard” text so users of the site can see this text in the language they have selected. Generally we do this by adding a string to `src/i18n/en/ui.ts` to start with and using our `useTranslations()` helper to get the label for the current page. This would require a small refactor here probably to pass the label through `CodeSnippet.astro` and have it picked up by this class.

src/components/CodeSnippet/CodeSnippet.astro
```diff
@@ -262,6 +262,40 @@ function parseMarkingDefinition(serializedArr: string, parts: RegExp, parseError
...
+				position: absolute;
+				top: 0.5rem;
+				right: 0.5rem;
+				cursor: pointer;
+				border-radius: 0.4rem;
+				opacity: 0.2;
+				-webkit-transition: visibility 0s, opacity 0.3s linear;
+				-moz-transition: visibility 0s, opacity 0.3s linear;
+				-o-transition: visibility 0s, opacity 0.3s linear;
+				transition: visibility 0s, opacity 0.3s linear;
```
delucis, line 276: We don’t need prefixes for these properties and I think we don’t need the `visibility` property here?

```suggestion
				transition: opacity 0.3s linear;
```

src/components/CodeSnippet/CodeSnippet.astro
```diff
@@ -262,6 +262,40 @@ function parseMarkingDefinition(serializedArr: string, parts: RegExp, parseError
...
+				.copy-button {
+					opacity: 1;
+				}
+			}
+
+			&:focus {
+				.copy-button {
+					opacity: 1;
+				}
+			}
```
delucis, line 298: We can combine these styles:

```suggestion
			&:hover,
			&:focus {
				.copy-button {
					opacity: 1;
				}
			}
```

delucis (commented):
Thanks for this @kyonlee! I think we’ll still need a bit more work to get this ready for docs. Let me know if you have any questions. If something seems difficult to do, I can help out!

sarah11918:
Hi @kyonlee! There's no rush, I'm just checking in to see whether you need any help with Chris' feedback. Please just let us know!

sarah11918:
Hi @kyonlee - I'm just following up to see whether you're still interested in continuing with Chris' suggestions! If not, you can always close this PR for now, and come back to it again at a later time.

kyonlee:
Hi @sarah11918, I'm still very interested in completing this PR! I've just been a bit busy recently. Was planning to complete it over the easter break and ask Chris for help on the pop-up part!

src/components/CodeSnippet/copy-button.ts
kyonlee: I didn't want to go overboard so I've gone with a simple approach of just passing the label through the `ShikiBlock` and then to the `CopyButton` class. If you had something better in mind please let me know!

src/components/CodeSnippet/CodeSnippet.astro
kyonlee: I'll need a hand implementing this. 🙏 😅

Yan-Thomas:
Hey @kyonlee, thanks for this! Here is the implementation for the "copy to clipboard" functionality added to `astro.build`: https://github.com/withastro/astro.build/blob/main/src/components/CodeBlock.astro, I think this might help you implement this here. I'd also be happy to take over for you and commit the necessary changes. 🙌

src/components/CodeSnippet/copy-button.ts
delucis: That’s perfect — thanks!

kyonlee:
@delucis I've added the tooltip! Please have a look and let me know if anything else is needed.

Yan-Thomas (approved):
LGTM! Thanks for the amazing contribution @kyonlee (and thanks for the support and improvements here @delucis) 💜:raised_hands:

delucis (approved):
Thanks again for this contribution @kyonlee — super helpful! I spent some time today tweaking a few details so it should now be ready to merge. Here’s a quick summary of some key details I ended up changing:

- I added `aria-live="polite"` to the copy button wrapper and switched to showing and hiding the tooltip via `display: none/block` instead of `opacity`. That way the ARIA Live Region picks up a “new” element when the tooltip is revealed and announces its contents to screenreader users

- I tweaked some of the styling so that it adapts for users reading in a right-to-left languages (Arabic on our site). This meant using [“logical properties”](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) — like `margin-inline-end` instead of `margin-right` or `inset-inline-start` instead of `left` — because these adapt automatically to the document’s writing direction.

- I switched to rendering the copy button _after_ the main code sample. This lets users reach the code first in the tab order before deciding to copy it. I also moved it _outside_ the main code flow so that if the code overflows causing horizontal scrolling, the copy button doesn’t scroll left/right with the code.

Other than that, I just tweaked a few colours and styles to make the button match the current site styles a little better.

This is going to be such a helpful and appreciated addition as the site, so thank you for taking the time to make it work!

PR #2842, opened 2023-03-13
Fix of the Nginx configuration file

guiguir68:

#### Description

This pull request is a fix for a configuration issue in the Nginx file. 

The current Nginx configuration does not automatically serve index.html files in subdirectories such as /dist/about/index.html, resulting in a "301 Moved Permanently" error when attempting to access the /about URL (for example). This error creates an issue when using a reverse proxy upstream, because Nginx redirects us to an URL with the container's port.


Jutanium:
Pinging @ElianCodes to review!

ElianCodes (approved):
100% valid fix! Error on my behalf, sorry 😅!

PR #2852, opened 2023-03-14
Tutorial Styling Improvements

Yan-Thomas:

#### Description

A few changes to get our tutorial ready for translations and to improve accessibility.

- Added RTL styles to `<Box/>` and `<Checklist/>` components:

| Before | After |
| image | image |

- Increased `line-height` for lists with `<kbd/>` elements

| Before | After |
| image | image |

- Improved Tutorial Tracker accessibility with better steps color contrast (now passing WCAG AAA) and added the missing focus style for the tab:

| Before | After |
| image | image |

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -95,10 +95,6 @@ nav ul {
 	margin-top: 0.5rem;
 }
 
-.content > section > :is(ul, ol) :has(kbd) {
-	line-height: 2.0;
-}
-
 .content > section > :first-child {
 	margin-top: 0;
 }
@@ -238,8 +234,8 @@ h5 {
 }
 
 p,
-.content ul {
-	line-height: 1.65;
+.content ul, .content ol {
+	line-height: 1.75;
 }
 
 p,
@@ -364,9 +360,9 @@ pre {
 	font-family: var(--font-body);
 	font-size: 0.9375rem;
 	border-radius: 0.25rem;
-	padding: 0.125rem 0.375rem;
+	padding: 0.0625rem 0.375rem;
 	border: 1px solid var(--theme-shade-subtle);
-	box-shadow: 0 3px var(--theme-shade-subtle);
+	box-shadow: 0 2px var(--theme-shade-subtle);
 	background-color: var(--theme-bg-offset);
 	white-space: nowrap;
 }
--- a/public/index.css
+++ b/public/index.css
@@ -95,6 +95,10 @@ nav ul {
 	margin-top: 0.5rem;
 }
 
+.content > section > :is(ul, ol) :has(kbd) {
+	line-height: 2.0;
+}
+
 .content > section > :first-child {
 	margin-top: 0;
 }
--- a/src/components/Checklist.astro
+++ b/src/components/Checklist.astro
@@ -39,7 +39,8 @@ const { key = 'default' } = Astro.props as Props;
 		position: relative;
 		margin: 0 -1rem;
 		border-radius: 0.5rem;
-		padding: 0.5rem 0.5rem 0.5rem 4rem;
+		padding-block: 0.5rem;
+		padding-inline: 4rem 0.5rem;
 		cursor: pointer;
 		/* Compensate for space between checkbox and text. */
 		text-indent: -0.4ch;
@@ -62,7 +63,7 @@ const { key = 'default' } = Astro.props as Props;
 	check-list :global(input[type='checkbox']::after) {
 		content: '✔︎';
 		position: absolute;
-		left: 1rem;
+		inset-inline-start: 1rem;
 		top: 50%;
 		transform: translateY(-50%) translate(3px, -3px);
 		transform-origin: center;
--- a/src/components/tutorial/TutorialNav.astro
+++ b/src/components/tutorial/TutorialNav.astro
@@ -106,6 +106,10 @@ const isCurrentUnit = (unit: typeof units[number]) =>
 		font-weight: bold;
 	}
 
+	:global(a:focus unit-progress-icon) {
+		outline: 2px solid var(--theme-accent-secondary);
+	}
+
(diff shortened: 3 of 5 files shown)


sarah11918:
Thanks for this, @Yan-Thomas! I'll ask @delucis to approve this, then we can move onto the i18nReady pages!

public/index.css
```diff
@@ -95,6 +95,10 @@ nav ul {
 	margin-top: 0.5rem;
 }
 
+.content > section > :is(ul, ol) :has(kbd) {
```
delucis, line 98: [`:has()` isn’t supported in Firefox](https://caniuse.com/css-has) just yet, so this is only a partial fix and this approach also results in this mix of line heights which I think doesn’t look amazing:

image

I think I might suggest a multi-part solution. This PR on the left, my suggestion on the right: 
image

How to do it:

1. Bump up line-height everywhere to `1.75`. This is a small change and I actually kind of like the extra breathing room it gives.

2. Tweak `kbd` styles to take up slightly less vertical space:

   - `box-shadow: 0 2px var(--theme-shade-subtle)` — 2px instead of the current 3px
   
   - `padding: 0.0625rem 0.375rem` — 1px instead of 2px vertical padding

src/components/Checklist.astro
delucis, line 1: Awesome — thanks for fixing this and the `<Box>` styles Yan! 💜

src/components/tutorial/UnitProgressIcon.astro
```diff
@@ -122,6 +122,14 @@ const t = useTranslations(Astro);
...
 	}
 
+	:global([aria-selected='true']) .segment:not(.done) {
+		stroke: hsl(var(--color-gray-20));
+	}
+
+	:global(.theme-dark [aria-selected='true']) .segment:not(.done) {
+		stroke: hsl(var(--color-gray-70));
+	}
+
```
delucis, line 132: While it’s true that now you can see the empty segments better, doesn’t this now mean the contrast between complete/incomplete segments is too low?

If we remove colours it’s very subtle that these are different:

image

I’d suggest making the empty segments only much more slightly lighter than they are currently (or leaving as-is — I think arguably it’s the complete segments that are the key information and you can see the empty space in contrast to those).

src/components/tutorial/TutorialNav.astro
```diff
@@ -106,6 +106,10 @@ const isCurrentUnit = (unit: typeof units[number]) =>
 		font-weight: bold;
 	}
 
+	:global(a:focus unit-progress-icon) {
+		outline: 2px solid var(--theme-accent-secondary);
```
delucis, line 110: I now see a _double_ outline in Firefox but if this is fixing a scenario where it’s completely missing in some browsers, NWTWWHB.

image

delucis (changes requested):
Thanks for tackling this, Yan! I’ve left some suggestions as I think we’ll still want to polish this a bit more.

Leaving a note for myself that I also want to make sure i18n progress is synced across languages. That can happen in a later PR (but ideally before merging #2856).

public/index.css
Yan-Thomas: I considered "everywhere" as in, content paragraphs & lists, which I liked, but bumping everywhere to 1.75 simply didn't feel right for some reason.

src/components/tutorial/UnitProgressIcon.astro
Yan-Thomas: Tweaked these a bit. I guess this works as a compromise, but I'm fine with leaving it as-is if you still think it isn't enough contrast. I reduced the saturation of the images to zero to try emulate what you did in your screenshot.

Dark:
image
image


Light:
image
image

src/components/tutorial/TutorialNav.astro
Yan-Thomas: Nice catch! I'm not entirely sure why this happens, but the `<a/>` in Firefox inherits the default focus style for links, but not Chrome (and probably all the other Chromium based browsers). I believe it might be related to `role="tab"` changing how the element is interpreted? In any case, changing `outline-offset` to match the inherited outline fixes this.

Yan-Thomas:
Thanks @delucis, made changes according to your suggestions. FYI, I can see the tutorial being properly tracked between translations, I probably thought it didn't because my local storage is constantly deleted.

src/components/tutorial/UnitProgressIcon.astro
delucis: I think this is good!

I took my screenshot using Firefox’s “Simulate Achromatopsia” option in the Accessibility tab in dev tools, which is useful for testing this and other color vision variations:

image

src/components/tutorial/TutorialNav.astro
delucis: Smart fix — thanks @Yan-Thomas!

delucis (approved):
LGTM! Thanks for taking care of this.

One thing still to think about in the future is if we need to handle RTL layout in the keyboard navigation of `<TutorialNav>`. Currently right arrow moves to the next tab, which is to the _left_ in RTL. I’d need to research whether an RTL keyboard maps that already or if we need to flip that ourselves. Either way that’s non-blocking, and I don’t think we’re likely to get an Arabic tutorial translation anytime soon.

Yan-Thomas:
> I’d need to research whether an RTL keyboard maps that already or if we need to flip that ourselves. Either way that’s non-blocking, and I don’t think we’re likely to get an Arabic tutorial translation anytime soon.

That's something I haven't considered, good point! Just saw they do this manually in the Material UI repo and the Material Design guidelines [kinda implies this has to be done manually](https://m2.material.io/design/usability/bidirectionality.html#mirroring-layout).

PR #2854, opened 2023-03-14
Update data-fetching.mdx

ffxsam:
Added clarity around fetch for static vs SSR

#### Description

Clarified a point about when `fetch()` gets executed.

### DIFF
--- a/src/content/docs/en/guides/data-fetching.mdx
+++ b/src/content/docs/en/guides/data-fetching.mdx
@@ -8,7 +8,7 @@ i18nReady: true
 
 ## `fetch()` in Astro
 
-All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs. This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/data-fetching.mdx
```diff
@@ -8,7 +8,7 @@ i18nReady: true
 
 ## `fetch()` in Astro
 
-All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs. This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML.
+All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs. This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at run time.
```
Yan-Thomas, line 11: Nit: for consistency with other docs pages, I'd change this to `runtime` instead
```suggestion
All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs. This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at runtime.
```

Yan-Thomas (commented):
Thanks @ffxsam, I have a small nit, after this LGTM!

delucis (approved):
Thanks for this update @ffxsam!

PR #2855, opened 2023-03-14
Make more pages ready for translation

Yan-Thomas:

- New or updated content

#### Description

New 69 pages for our translators to enjoy! In this case, all the integrations, tutorial, CMS, and migration pages are now marked as `i18nReady: true`.

Preferably, we should merge #2852 before this one.

### DIFF
--- a/scripts/generate-integration-pages.ts
+++ b/scripts/generate-integration-pages.ts
@@ -24,6 +24,7 @@ interface IntegrationData {
 	category: 'renderer' | 'adapter' | 'other';
 	readme: string;
 	srcdir: string;
+	i18nReady: string;
 }
 
 const prettyCategoryDescription: Record<string, unknown> = {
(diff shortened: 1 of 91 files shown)


scripts/generate-integration-pages.ts
```diff
@@ -31,6 +31,10 @@ const prettyCategoryDescription: Record<string, unknown> = {
 	adapter: 'SSR adapter to deploy your Astro project',
 	other: 'integration in your Astro project',
 };
+
+// Add pages titles that should be marked as `i18nReady: false` as part of this array.
+const translationBlacklist: string[] = [];
```
delucis, line 36: I’d suggest using “blocklist” instead of “blacklist” or a use-case specific name like `i18nNotReadyIntegrations` or something. We can also follow the same pattern as for `#deprecatedIntegrations` setting this inside `IntegrationsPagesBuilder`.

Maybe we want to block translation of Markdoc for now given it’s still very experimental?

delucis (commented):
Impressive work here, @Yan-Thomas, thank you! Left one suggestion for the integration script.

scripts/generate-integration-pages.ts
Yan-Thomas: Fair points, I'll make these changes after arriving from college 🙌

Yan-Thomas:
Done @delucis! Let me know if you have any other comments/suggestions 🙌

dreyfus92 (approved):
LGTM! ✅ We need to start shipping new translations.

PR #2859, opened 2023-03-15
feat(image_services): Add mention of new validateOptions hook

Princesseuh:

- New or updated content

#### Description

https://github.com/withastro/astro/pull/6555 adds a new hook to the Image Service API that can be used to validate and augment the passed options. The built-in services uses this hook to set the default format as `webp` if the user hasn't passed a format, for example.

It takes the same `options` as the other hooks, so an `ImageTransform` and it returns an ImageTransform as well. Returning an `ImageTransform` is mandatory, so the implementation should probably always end with `return options;`

This PR additionally contain some misc fix to the rest of the file


sarah11918 (approved):
LGTM, @Princesseuh!  :tada:

Yan-Thomas (commented):
LGTM! But I do agree with @Jutanium here, might be helpful to have an example.

Princesseuh:
Core PR has been merged, will merge this when it's released

sarah11918:
Merging-on-release, but we have an example ready to add in a separate PR.

PR #2866, opened 2023-03-16
Add "needs reviews" section to i18n.docs.astro.build

### LINKED ISSUE #1059: Can we have an “in progress” section? So each language would have missing/needs updating/in progress.

TheOtterlord:

- Something else!

#### Description

- Closes #1059
- What does this PR change? Adds a "Translations that need reviews" section to [i18n.docs.astro.build](https://i18n.docs.astro.build)
- Did you change something visual? Adds the following section

image

### DIFF
--- a/scripts/translation-status.ts
+++ b/scripts/translation-status.ts
@@ -11,6 +11,14 @@ import output from './lib/output.mjs';
 import type { PageData, PageIndex, PageTranslationStatus } from './lib/translation-status/types';
 import { toUtcString, tryGetFrontMatterBlock } from './lib/translation-status/utils.js';
 
+interface PullRequest {
+	html_url: string;
+	title: string;
+	labels: {
+		name: string;
+	}[];
+}
+
 /**
  * Uses the git commit history to build an HTML-based overview of
  * the current Astro Docs translation status.
@@ -80,9 +88,12 @@ class TranslationStatusBuilder {
 		// Determine translation status by source page
 		const statusByPage = this.getTranslationStatusByPage(pages);
 
+		// Fetch all pull requests
+		const pullRequests = await this.getPullRequests();
+
 		// Render a human-friendly summary
 		output.debug(`- Building HTML file...`);
-		const html = this.renderHtmlStatusPage(statusByPage);
+		const html = this.renderHtmlStatusPage(statusByPage, pullRequests);
 
 		// Write HTML output to file
 		fs.writeFileSync(this.htmlOutputFilePath, html);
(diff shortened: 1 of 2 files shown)


scripts/translation-status.ts
```diff
@@ -92,6 +103,15 @@ class TranslationStatusBuilder {
 		output.debug('');
 	}
 
+	/** Get all pull requests with the `i18n` tag */
+	async getPullRequests() {
+		const pullRequests = (await fetch(
+			`https://api.github.com/repos/${this.githubRepo}/pulls?state=open&per_page=50`
+		).then((res) => res.json())) as PullRequest[];
```
Yan-Thomas, line 110: Pointing out we have a function for getting data from GitHub, you can find it in [scripts/lib/github-get.mjs](https://github.com/withastro/docs/blob/main/scripts/lib/github-get.mjs). It has retry, headers and makes easier to add our GitHub token to avoid the request being limited, could you make the changes to use it in here? Here's how we're using it in the [generate-docgen-integration-pages.ts](https://github.com/withastro/docs/blob/main/scripts/generate-integration-pages.ts#L73) script if that helps!

Also heads up that adding an env variable possibly will block the PR/deploy for security reasons, if this happens we will take action to unblock it, so no need to worry! 🙌

Yan-Thomas (commented):
Thanks @TheOtterlord! This is such a nice feature to add in the Translation Status. Also super happy to see you making your docs repo debut, let's get that Astro Badge you deserve! 💜 

I've made a comment regarding fetching the GitHub API, let me know if you end up needing any help 🙌

scripts/translation-status.ts
TheOtterlord: That's a handy util. Thanks for linking the usage too! That made it a lot quicker to add :smile:

scripts/translation-status.ts
```diff
@@ -332,6 +358,21 @@ class TranslationStatusBuilder {
 		return lines.join('\n');
 	}
 
+	renderTranslationNeedsReview(prs: PullRequest[]) {
+		const lines: string[] = [];
+
+		if (prs.length > 0) {
+			lines.push(`<ul>`);
+			lines.push(
+				...prs.map((pr) => `<li>` + `${this.renderLink(pr.html_url, pr.title)} ` + `</li>`)
```
delucis, line 367: What do you think of stripping backticks out of the PR titles?

```suggestion
				...prs.map((pr) => {
					const title = pr.title.replaceAll('`', '');
					return `<li>` + this.renderLink(pr.html_url, title) + `</li>`;
				})
```

There are a few in your screenshot and it’s a bit of a nit pick, but I guess we can remove them? 
PR titles rendered including backticks

delucis (commented):
This is fantastic, thank you @TheOtterlord! I’ve left a couple of minor comments, but then should be good to go.

I can think of more fun improvements now we have this, like maybe auto-labelling each language differently and then using those labels to group PRs by language, etc., or fetching the comments for each PR so we can show if there’s been any discussion or not, but that’s for a future enhancement!

scripts/translation-status.ts
TheOtterlord: Nice. yeah, that'll look much better

TheOtterlord:
Pushed the suggestions :+1:

> I can think of more fun improvements now we have this, like maybe auto-labelling each language differently and then using those labels to group PRs by language, etc., or fetching the comments for each PR so we can show if there’s been any discussion or not, but that’s for a future enhancement!

+1 to this

delucis (approved):
Let’s get this into docs! Excited to see it live. Thanks @TheOtterlord 🥳

PR #2869, opened 2023-03-16
[TEST] Test `astro-og-canvas` upgrade against #2856

delucis:
Test this upgrade to `astro-og-canvas` does the trick


delucis:
Links are failing as expected but looks like Netlify’s building OK! 🥳 

Think we can close and #2870 should be good to merge.

PR #2870, opened 2023-03-16
Bump `astro-og-canvas` to 0.1.7

delucis:

- Something else!

#### Description

Upgrades to the latest version of `astro-og-canvas` which manually frees memory after generating each image. #2856 builds were erroring and my working hypothesis is that we were hitting an internal memory limit in canvaskit-wasm.

I’m testing this update against the changes in #2856 in #2869.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "astro": "^2.0.17",
     "astro-auto-import": "^0.2.1",
     "astro-eslint-parser": "^0.9.2",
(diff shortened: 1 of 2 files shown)


delucis:
Tests passed in #2869, so I think this is the fix. We can then update branch in #2856 and hopefully everything works! 🤞

Yan-Thomas:
LGTM @delucis, 🚀!

PR #2872, opened 2023-03-16
Remove dev dependency flag from CSS preprocessor installation

Jutanium:

- New or updated content

#### Description

Addresses feedback fish:

> Earlier in the docs, we're told to import dependencies as plain dependencies, not dev dependencies. Here we're shown the -D flag being passed... quick explanation as to why this case warrants adding preprocessors as dev dependencies would be helpful!


dreyfus92 (approved):
LGTM! ✅

Yan-Thomas (approved):
LGTM!

delucis (approved):
I guess? I definitely think of these as dev dependencies — you're not shipping Sass to your frontend — but in most cases this probably doesn't matter.

PR #2875, opened 2023-03-18
Correcting script command in package.json

WintrCodes:

#### Description
Removing erroneous 'npm', 'pnpm', and 'yarn' word from the 'dev' script in package.json


WintrCodes:
Tested using 'npx' instead of 'npm' at request of @Yan-Thomas - this also worked. 

I had already submitted the PR prior to testing this, so someone can amend this to use 'npx' if that's preferred.

Yan-Thomas:
According to issues in the TinaCMS repo, the script should be package manager agnostic and avoid npx, so there isn't a mismatch between your global and local versions of Tina. Thanks for catching this @WintrCodes, LGTM!

PR #2876, opened 2023-03-18
Remove old `extendDefaultPlugins` from examples

BryceRussell:

#### Description

Remove old `extendDefaultPlugins` option from config examples for remark plugins

#### Note

With the changes to markdown plugin configuration in 2.0 the second MDX  example in step 2 of [this section](https://docs.astro.build/en/guides/markdown-content/#modifying-frontmatter-programmatically) may be redundant since the first example will now work with MDX by default, should it be removed too?


sarah11918:
Great catch! Thanks @BryceRussell :raised_hands:

PR #2884, opened 2023-03-19
Translation Tuesday

### LINKED ISSUE #1057: Ping relevant roles in `#docs-i18n` on Discord every “Translation Tuesday”.
If there are new additions to the out-of-date pages list, ping the relevant roles in `#docs-i18n` on Discord. Suggestion to do this every “Translation Tuesday”.

TheOtterlord:

- Something else!

#### Description

- Closes #1057
- What does this PR change? Implements "Translation Tuesday". A weekly ping in the discord alerting people to new/updated content in need of translation.

> **Note**
> It currently uses the Docsbot webhook environment secret. It might be worth adding a separate one to ping in `#docs-i18n` instead of `#docs`. 

image


ElianCodes:
I love this idea!

sarah11918:
This is so cool, @TheOtterlord! I agree that it's maybe better placed in the i18n channel, but I have no problems with it starting out in the main docs channel, especially to start where it might get more visibility!

I'll let @delucis and @Yan-Thomas approve the implementation, but I'm super happy with any initial version of this we get:

- ping just the `i18n-gang` main role for simplicity's sake?
- ping each individual language role on each line??
- no ping at all to start?

I would say, do NOT ping `@team-docs` with these, but I'm happy with whatever else y'all choose to use for the ping.

TheOtterlord:
Glad you like it!
The current version pings the `@i18n-gang` role, but that can be remove. I don't think I'd recommend pinging each language role, just because that eats into the 2000 character limit in a Discord webhook.

hippotastic:
Hey, thanks a lot for tackling this! The screenshot of the resulting message looks great!

I'm wondering if you can reuse existing functionality from my translation-status.ts script to avoid having two places where checks for translation freshness, i18nignore keywords etc. are which could get out of sync.

If I see it correctly, it should already be enough to add an `export` statement in front of the class `TranslationStatusBuilder` and then you can import, instantiate and use it in your script as well.

Instead of calling its `run` method, you would then call the more granular individual methods to get the status information you need (you can see what `run` does for reference). Everything should already be there.

An alternative approach I would see is to refactor the TranslationStatusBuilder so that the shared logic used by your new script and my old one are extracted to a separate lib file that both scripts can use.

TheOtterlord:
@hippotastic thanks! Using it via an export seemed to work fine. Also separated out the regex into a const so I can still use it against pull request info

hippotastic:
Great progress, thank you! :)

I wonder if it's possible to also replace the block where you build an index of all pages and then get the last commit per page and translation with calls to `builder.createPageIndex` and `builder.getTranslationStatusPerPage`, because then the scripts would also share the same logic there. It feels to me like the existing methods in the builder should return a structure that contains everything you need at this point in your code.

TheOtterlord:
Sorry it took so long. I had to rework the order in which things were done a little, but everything works now. I've also added a runtime check to the translation-status method because it would occasionally run that script too (I eventually gave up on finding the reason it only ran sometimes, and so added the check).

hippotastic:
Thank you for continuing to work on this!

The reason why the translation status script got executed as well is because it contains top-level code that gets executed on import (the code you added the check to). I'm sorry about that, didn't think of this detail before!

You could solve this by moving the TranslationStatusBuilder class declaration to its own file below the `lib` folder and importing that class both in the translation status script and your new translation tuesday script. That way, no code gets executed upon import and you wouldn't need the check you added.

One more question regarding your code: Could you elaborate on why you are fetching pull request information? If the goal of the script is to show which translations of a page still need to be updated, the PRs should not be required.

Please note that the call to `getTranslationStatusByPage` returns an array of the following type:

```ts
export type PageTranslationStatus = {
	subpath: string;
	sourcePage: PageData;
	githubUrl: string;
	translations: {
		[language: string]: {
			page: PageData;
			isMissing: boolean;
			isOutdated: boolean;
			githubUrl: string;
			sourceHistoryUrl: string;
		};
	};
};
```

As you can see, each source page already has an accompanying map of matching translations per language, including per-language flags like `isOutdated` or `isMissing`. You should be able to build the entire Discord message just from this array without requiring any further GitHub API calls.

TheOtterlord:
> You could solve this by moving the TranslationStatusBuilder class declaration to its own file in the lib/translation-status folder and importing that class both in the translation status script and your new translation Tuesday script. That way, no code gets executed upon import and you wouldn't need the check you added.

Sounds good, I'll move that to a file.

> One more question regarding your code: Could you elaborate on why you are fetching pull request information? If the goal of the script is to show which translations of a page still need to be updated, the PRs should not be required.

I'm still using `getTranslationStatusByPage`, but because sometimes the translation teams are so quick, sometimes a page gets PRed pretty quickly, but might not be merged by the time of the script running on Tuesdays. The solution in place at the moment combines both of these to get a more complete result for what translations are still needed (essentially it eliminates any it's found to already be translated in a PR).

This isn't perfect, e.g. if someone translates a page that has had a major update, but doesn't take that update into account at the time of the last major commit (which I'd assume is really rare because then the translation is outdated). It also might not be worth the effort of fetching each open PR and comparing them against the main branch, because it does make it more complex to debug etc.

hippotastic:
> I'm still using `getTranslationStatusByPage`, but because sometimes the translation teams are so quick, sometimes a page gets PRed pretty quickly, but might not be merged by the time of the script running on Tuesdays. The solution in place at the moment combines both of these to get a more complete result for what translations are still needed (essentially it eliminates any it's found to already be translated in a PR).
> 
> This isn't perfect, e.g. if someone translates a page that has had a major update, but doesn't take that update into account at the time of the last major commit (which I'd assume is really rare because then the translation is outdated). It also might not be worth the effort of fetching each open PR and comparing them against the main branch, because it does make it more complex to debug etc.

Gotcha! Thank you for the explanation. We also discussed this topic in the past (partially in the related issue), including the "race condition" when open PRs may have become outdated by a source page update that happened after the last commit to the PR, but decided to keep PRs out of scope for now in order not to overcomplicate this topic.

Therefore I'd also recommend keeping the PR check out of this script for now, which should enable you to remove all the extra API calls and use the existing array as described to build the Discord message. This will also ensure that both scripts are perfectly in sync at the time of posting the Discord message and share the same logic.

However, the discussion brought me to the idea that it may be useful to make the link to https://i18n.docs.astro.build a bit more prominent. Thanks to your great work adding the "Translations that need reviews" section, this page always has all the up to date information on PRs, too.

So rather than just adding `And [more](https://i18n.docs.astro.build)...` to the end of the Discord message, you could change this to: `See our [Translation Status page](https://i18n.docs.astro.build) for more, including open PRs.`

TheOtterlord:
Migrated the class to a lib file & removed the PR part (it runs so much quicker without all those requests xD)

.github/workflows/discord-i18n-ping.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Install Tools & Dependencies
+        uses: ./.github/actions/install
+
+      - id: message
+        name: Format Discord message
+        env:
+          COMMIT_AUTHOR: ${{ github.event.commits[0].author.name }}
+          COMMIT_MESSAGE: ${{ github.event.commits[0].message }}
+          COMMIT_ID: ${{ github.event.commits[0].id }}
+        run: node ./scripts/tuesday-bot.mjs
```
hippotastic, line 28:
```suggestion
        run: pnpm tsm --require=./scripts/lib/filter-warnings.cjs ./scripts/tuesday-bot.ts
```
We need to use this command as the script is using TypeScript.

.github/workflows/discord-i18n-ping.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Install Tools & Dependencies
+        uses: ./.github/actions/install
+
+      - id: message
+        name: Format Discord message
+        env:
+          COMMIT_AUTHOR: ${{ github.event.commits[0].author.name }}
+          COMMIT_MESSAGE: ${{ github.event.commits[0].message }}
+          COMMIT_ID: ${{ github.event.commits[0].id }}
```
hippotastic, line 27:
```suggestion
```
These lines should not be required for the script as it doesn't access these ENV variables.

.github/workflows/discord-i18n-ping.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  schedule:
+    - cron:  '0 0 * * TUE'
+  workflow_dispatch:
+
+jobs:
+  translation:
+    name: translation_tuesday
+    if: ${{ github.repository_owner == 'withastro' && github.event.commits[0].message != '[ci] format' }}
```
hippotastic, line 11:
```suggestion
    if: ${{ github.repository_owner == 'withastro' }}
```
As the script is not being run on a commit, but by a cron job or manually, we shouldn't try to reference a commit here.

.stackblitzrc
hippotastic, line 1: Was this file removed by accident? It shouldn't have anything to do with the PR, so I'd avise to restore it.

.vscode/extensions.json
hippotastic, line 1: Same here

.vscode/settings.json
hippotastic, line 1: And here :)

scripts/tuesday-bot.ts
```diff
@@ -0,0 +1,52 @@
...
+		Object.keys(toTranslate).length
+	} pages with major changes. Please help us translate these pages to your language!\n\n${list}`;
+
+	// message length limit is 2000 characters
+	while (message.length > 1902) {
+		const lastNewline = message.lastIndexOf('\n', 2000);
+		message = message.slice(0, lastNewline);
+	}
+
+	message += '\n\nSee our [Translation Status page](https://i18n.docs.astro.build) for more, including open PRs.';
```
hippotastic, line 49:
```suggestion
	const suffix = '\n\nSee our [Translation Status page](https://i18n.docs.astro.build) for more, including open PRs.';

	// Keep the entire message including the suffix within Discord's limits
	const maxLengthWithoutSuffix = 2000 - suffix.length;
	while (message.length > maxLengthWithoutSuffix) {
		const lastNewline = message.lastIndexOf('\n', maxLengthWithoutSuffix);
		message = message.slice(0, lastNewline);
	}

	message += suffix;
```

I tried to get rid of the "magic number" 1902 here by making sure that it gets calculated from the message suffix length. This way, if we ever decide to change the suffix, we don't need to think about updating this value.

hippotastic (commented):
Again, thanks a lot for your continued work on this! Your changes look great to me.

Here are some final suggested changes from my end, then we should be good to go!

.github/workflows/discord-i18n-ping.yml
TheOtterlord: Ah yep, this was from before I switched to typescript. Good catch!

.stackblitzrc
TheOtterlord: Oh wow, I had no idea this had been removed, weirdly enough it's also in the gitignore. I've managed to add them all back

TheOtterlord:
Changed made (and files restored :sweat_smile:)

scripts/tuesday-bot.ts
```diff
@@ -0,0 +1,55 @@
...
+    )}) (${langs[0] === 'all' ? 'all' : langs.filter(lang => {
+      if (lang === 'en') return false;
+      const { isMissing, isOutdated } = s.translations[lang];
+      return isMissing || isOutdated;
+    }).join(', ')})`;
+  }).join('\n');
+
+	let message = `**Translation Tuesday!** <@951985780828545095>\n\nWe have ${
+		Object.keys(toTranslate).length
+	} pages with major changes. Please help us translate these pages to your language!\n\n${list}`;
```
hippotastic, line 41:
```suggestion
	} pages with major changes since last week. Please help us translate these pages to your language!\n\n${list}`;
```
I ran this script and compared its output to the i18n status page and noticed that the message contains only a limited set of pages (and rightfully so, because it's limited to major changes since last week here). However, this was not reflected in the message in any way, so I suggest adding this.

hippotastic (commented):
I lied! Here are two more changes. 😆

I discovered them by checking out your PR locally and running the scripts. These should really be the last ones now.

scripts/tuesday-bot.ts
TheOtterlord: Yeah, this sounds good

hippotastic (approved):
LGTM!

MoustaphaDev:
Late to the party but I have to say this is so cool! Amazing work @TheOtterlord

PR #2890, opened 2023-03-20
Update assets.mdx (en-only)

boogerbuttcheeks:

#### Description

- Change "an" to "a".

### DIFF
--- a/src/content/docs/en/guides/assets.mdx
+++ b/src/content/docs/en/guides/assets.mdx
@@ -47,7 +47,7 @@ When you next run Astro, it will update your `src/env.d.ts` file to configure ty
 
 ### Move your images to `src/assets/`
 
-Create a `src/assets/` folder, which Astro will recognize as your new assets folder.
+Create `src/assets/`, which Astro will recognize as your new assets folder.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/assets.mdx
```diff
@@ -47,7 +47,7 @@ When you next run Astro, it will update your `src/env.d.ts` file to configure ty
 
 ### Move your images to `src/assets/`
 
-Create an `src/assets/` folder, which Astro will recognize as your new assets folder.
+Create a `src/assets/` folder, which Astro will recognize as your new assets folder.
```
sarah11918, line 50:
```suggestion
Create `src/assets/`, which Astro will recognize as your new assets folder.
```
I'm going to remove the problem entirely! We make our choices based on pronunciation, and if someone reads the line as "an S-R-C" folder, they would use "an". If they read "a SOURCE" folder, then it would be "a", and we can't know which way people are saying it in their heads when they read.

What about this compromise?

sarah11918: Yay! OK, then let's make it so!  :rocket:

PR #2892, opened 2023-03-20
improve performance, fix flash of content on load

FredKSchott:

- Improve UX while browsing the docs site!

#### Description

- Add local cache headers for all statically generated (fingerprinted!) assets
- Inline a few very small scripts that were important to be inlined for reducing thrash on load
- ~~Move `index.css` and `theme.css` out of `public/` and into `src/` so that they get automatic fingerprinting for free.~~ CSS Ordering issues, had to revert this.
- Replace `scrollIntoView` sidebar scroll position method with a LocalStorage approach which has zero flash of unscrolled content (Note: I can still see some flash of unscrolled content on "Fast 3G" while the sidebar loads in, since this script runs inline but only after its fully populated. Maybe there's some hack where we could give it a min-height that can be scrolled _before_ the content loads in? I'll leave this out of scope for now).

#### Before

Screen Shot 2023-03-20 at 12 51 15 PM

#### After

Screen Shot 2023-03-20 at 12 50 56 PM

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,8 @@
+[[headers]]
+  for = "/_astro/*"
+  [headers.values]
+    Cache-Control = "public, max-age=604800, immutable"
+
 [build]
   command = "NODE_OPTIONS=--max_old_space_size=4096 pnpm netlify:build"
   ignore = "git diff --quiet $COMMIT_REF $CACHED_COMMIT_REF -- /"
--- a/src/components/LeftSidebar/LeftSidebar.astro
+++ b/src/components/LeftSidebar/LeftSidebar.astro
@@ -84,11 +84,18 @@ if (isReference) {
 	</ul>
 </nav>
 
-<script>
-	window.addEventListener('DOMContentLoaded', () => {
-		var target = document.querySelector('[data-current-parent="true"]');
-		target?.scrollIntoView({ block: 'center' });
-	});
+<!-- Preserve sidebar scroll across page loads -->
+<script is:inline>
+	{
+		const leftSidebar = document.querySelector('.nav-groups');
+		const leftSidebarScroll = localStorage.getItem('sidebar-scroll');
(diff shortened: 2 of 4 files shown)


Yan-Thomas (approved):
Nice work @FredKSchott, LGTM!

src/components/LeftSidebar/LeftSidebar.astro
```diff
@@ -84,11 +84,18 @@ if (isReference) {
...
+	{
+		const leftSidebar = document.querySelector('.nav-groups');
+		const leftSidebarScroll = localStorage.getItem('sidebar-scroll');
+		if (leftSidebarScroll !== null) {
+			leftSidebar.scrollTop = parseInt(leftSidebarScroll, 10);
+		}
+		window.addEventListener('beforeunload', () => {
+			localStorage.setItem('sidebar-scroll', leftSidebar.scrollTop);
+		});
+	}
```
natemoo-re, line 98: They said it couldn't be done!

natemoo-re (approved):
Looks great, nice catch!

Yan-Thomas (changes requested):
Blocking this! Not sure what's happening but in the latest deploy I can now see some CSS styling issues happening:
image

FredKSchott:
gar, more CSS ordering fun. I moved those public `.css` files into into `src/` to give them fingerprinting, but it looks like the `.astro` style tags no longer apply in the right order and are overridden by `index.css`

netlify.toml
```diff
@@ -1,3 +1,8 @@
+[[headers]]
+  for = "/_astro/*"
+  [headers.values]
+    Cache-Control = "public, max-age=604800, immutable"
+
```
delucis, line 5: Thank you! Been meaning to do that for ages.

delucis (approved):
LGTM! 🚀

PR #2895, opened 2023-03-20
Add i18n Recipe

Jutanium:

- New or updated content

#### Description

- Adds an i18n recipe that explains how to use content collections and dynamic routes to add i18n features
-


src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+        // src/pages/index.astro
+        return Astro.redirect('/en/');
+        ---
+        ```
+      </Fragment>
+    </StaticSsrTabs>
+
+### Use collections for translated content
+
+1. Create a folder in `src/content/` for each type of content you want to include, and create subdirectories for each supported language. For example, to support English and French blog posts:
```
Yan-Thomas, line 59: The comma here is unnecessary, though removing it will expose the double "create". Considering this, I believe that just changing the second one to "add" solves the problem.
```suggestion
1. Create a folder in `src/content/` for each type of content you want to include and add subdirectories for each supported language. For example, to support English and French blog posts:
```

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+        const { Content, headings } = await page.render();
+        ---
+        <h1>{page.data.title}</h1>
+        <p>by {page.data.author} at {formattedDate}</p>
+        <Content/>
+        ```
+      </Fragment>
+    </StaticSsrTabs>
+
+    📚 [Read more about dynamic routing](/en/core-concepts/routing/#dynamic-routes)
```
Yan-Thomas, line 153: Missing comma
```suggestion
    📚 [Read more about dynamic routing](/en/core-concepts/routing/#dynamic-routes).
```

Yan-Thomas (commented):
Nice work peeps, I added a few small suggestions, overall LGTM!

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+        date: z.date()
+      })
+    });
+
+    export const collections = {
+      'blog': blogCollection
+    };
+
+    ```
+    📚 [Read more about Content Collections.](/en/guides/content-collections/)
```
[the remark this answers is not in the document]
delucis: Actually looking closely we also leave out the generic link text too, which is probably better.

```suggestion
    📚 Read more about [Content Collections.](/en/guides/content-collections/)
```

src/content/docs/en/recipes/i18n.mdx
delucis:
```suggestion
    📚 Read more about [dynamic routing](/en/core-concepts/routing/#dynamic-routes).
```

Jutanium:
Thank you @delucis, it's awesome!

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
+---
+title: Add internationalization features
```
lorenzolewis, line 2:
```suggestion
title: Add internationalization (i18n) features 
```
Was just looking at this on the site and my eye always looks for "i18n" specifically

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+title: Add internationalization features 
+description: Use dynamic routing and content collections to add i18n support to your Astro site.
+type: recipe
+i18nReady: false
+---
+import FileTree from '~/components/FileTree.astro'
+import StaticSsrTabs from '~/components/tabs/StaticSsrTabs.astro'
+
+Astro doesn't have built-in internationalization (i18n) support, but you can build your own i18n solution. In this recipe, you'll learn how to use content collections and dynamic routing to serve content for different languages.
+
```
lorenzolewis, line 11: I'd start off with was sort of URL scheme you're going for (what's the output of following this guide). For example `astro.build/[lang]/blog/foo-bar`

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+            const [lang, ...slug] = page.slug.split('/');
+            return { params: { lang, slug: slug.join('/') || undefined }, props: page }
+          })
+
+          return paths;
+        }
+
+        const { lang, slug } = Astro.params;
+        const page = Astro.props;
+        const formattedDate = page.data.date.toLocaleString(lang);
```
lorenzolewis, line 118: `toLocaleString()` could potentially be confusing for those not familiar with it. Maybe a callout of what that is and link to some nice MDN docs?

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+
+Create dictionaries of terms to translate the labels for UI elements around your site. This allows your visitors to experience your site fully in their language.
+
+1. Create a `src/i18n/ui.ts` file to store your translation strings in:
+
+    ```ts
+    // src/i18n/ui.ts
+    export const languages = {
+      en: 'English',
+      fr: 'français',
```
lorenzolewis, line 165:
```suggestion
      fr: 'Français',
```
Same caps as English

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+      </Fragment>
+    </StaticSsrTabs>
+
+    📚 Read more about [dynamic routing](/en/core-concepts/routing/#dynamic-routes).
+
+### Translate UI strings
+
+Create dictionaries of terms to translate the labels for UI elements around your site. This allows your visitors to experience your site fully in their language.
+
+1. Create a `src/i18n/ui.ts` file to store your translation strings in:
```
lorenzolewis, line 159: I wonder if `ui.ts` is the best filename? Maybe `strings.ts` or something similar? 🤔

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+    ```ts
+    // src/i18n/ui.ts
+    export const languages = {
+      en: 'English',
+      fr: 'français',
+    };
+
+    export const defaultLang = 'en';
+    
+    export const ui = {
```
lorenzolewis, line 170: If changing this then you could shorten this exported const to `s` (for string) or `t` (for translation) to make it that much easier. I think `t` is what other i18n libraries tend to use.

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+    ---
+    // src/components/Nav.astro
+    import { getLangFromUrl, useTranslations } from '../i18n/utils';
+    
+    const lang = getLangFromUrl(Astro.url);
+    const t = useTranslations(lang);
+    ---
+    <ul>
+        <li>
+            <a href={`/${lang}/home/`}>
```
lorenzolewis, line 217: astro-i18n uses a function to generate routes which I like, similar DX as strings so it makes more sense in my mind than getting into string interpolation https://github.com/alexandre-fernandez/astro-i18n#translated-routing-function

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+    // src/pages/en/about.astro
+    import Base from "../../layouts/Base.astro"
+    ---
+    <Base>
+        <h1>About me</h1>
+        ...
+    </Base>
+    ```
+
+### Let users switch between languages
```
lorenzolewis, line 270: If your use case assumes you have the same pages across each language then it can be a bit of confusing UX to switch languages but then be taken to the index page. I like it where it just swaps out the locale slug but keeps you on the respective page.

lorenzolewis (commented):
I absolutely love this recipe! A lot of the comments are probably more opinion so feel free to do with them what you'd like 🎉

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,322 @@
...
+    ```
+    
+2. Create a helper function to detect the page language based on the current URL and a helper to let us get translations strings for different parts of our UI in `src/i18n/utils.ts`:
+
+    ```js
+    // src/i18n/utils.ts
+    import { ui, defaultLang } from './ui';
+    
+    export function getLangFromUrl(url: URL) {
+      const [, lang] = Astro.url.pathname.split('/');
```
BryceRussell, line 190:
```suggestion
      const [, lang] = url.pathname.split('/');
```
I think this is suppose use the `url` argument instead of the Astro global

delucis: Oh yeah, good catch!

src/content/docs/en/recipes/i18n.mdx
delucis: Yeah, I agree, but felt maybe that that assumption was one too many for the recipe? We enforce this on the docs site — all pages must use the English slugs — but you might also want to use translated slugs and then implementing that is a little more complex. Curious to hear whether other think this is essential to include in this recipe.

src/content/docs/en/recipes/i18n.mdx
delucis: The `useTranslations` helper returns a `t` function, but you’re not intended to use this variable directly, so I think this OK? I could get behind a _more_ verbose name though, like `uiStrings`.

lorenzolewis: Keep in mind you'll be using that function A LOT so short is sweet (even with nice autocomplete)

lorenzolewis: Ah nevermind, I mis-read that and didn't realise at first you don't use it directly. So yes, if that's the case a more verbose name makes sense.

src/content/docs/en/recipes/i18n.mdx
delucis: Good thinking. Also just confirmed this is much more frequently what people search for. I’m going to switch this up so we do include “internationalization” in the description.

src/content/docs/en/recipes/i18n.mdx
delucis: Nice! Yeah, we decided to keep this recipe real simple and avoid a specific library but we’ve linked astro-i18n at the end of the page for people to explore.

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,327 @@
...
+    // src/i18n/utils.ts
+    import { ui, defaultLang } from './ui';
+    
+    export function getLangFromUrl(url: URL) {
+      const [, lang] = url.pathname.split('/');
+      return lang || defaultLang;
+    }
+    
+    export function useTranslations(lang: keyof typeof ui) {
+      return function t(key: keyof typeof ui[defaultLang]) {
```
BryceRussell, line 200:
```suggestion
      return function t(key: keyof typeof ui[typeof defaultLang]) {
```
Fix unresolved type error

delucis: Oh, yeah. Ugly 😅 

But correct — thank you!

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,327 @@
...
+    
+2. Create a helper function to detect the page language based on the current URL and a helper to let us get translations strings for different parts of our UI in `src/i18n/utils.ts`:
+
+    ```js
+    // src/i18n/utils.ts
+    import { ui, defaultLang } from './ui';
+    
+    export function getLangFromUrl(url: URL) {
+      const [, lang] = url.pathname.split('/');
+      return lang || defaultLang;
```
BryceRussell, line 196:
```suggestion
      return (lang as keyof typeof ui) || defaultLang;
```
Fix type mismatch in `Nav.astro` example where `getLangFromUrl()` returns type `string` but `useTranslations()` expects type `keyof typeof ui`

delucis: Hmmm. In this case I think we might want to fix it in the `useTranslations()` function. It would be lying to say `getLangFromUrl` always returns a key of the UI dictionary. 🤔

delucis: Actually, perhaps the best here _is_ to guarantee this is always a valid language in this function, but that requires an extra if statement. I’ll update the full function.

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,328 @@
...
+        <p>by {page.data.author} • {formattedDate}</p>
+        <Content/>
+        ```
+      </Fragment>
+    </StaticSsrTabs>
+
+    📚 Read more about [dynamic routing](/en/core-concepts/routing/#dynamic-routes).
+
+    :::tip[Date formatting]
+    The example above uses the built-in [`toLocaleString()` date-formatting method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) to create a human-readable string from our frontmatter date.
```
sarah11918, line 156:
```suggestion
    The example above uses the built-in [`toLocaleString()` date-formatting method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) to create a human-readable string from the frontmatter date.
```
frontmatter's ?

from the date in the frontmatter?

Whatever lets us avoid "our".

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,328 @@
...
+    This ensures the date and time are formatted to match the user’s language.
+    :::
+
+### Translate UI strings
+
+Create dictionaries of terms to translate the labels for UI elements around your site. This allows your visitors to experience your site fully in their language.
+
+1. Create a `src/i18n/ui.ts` file to store your translation strings in:
```
sarah11918, line 164:
```suggestion
1. Create a `src/i18n/ui.ts` file to store your translation strings:
```
I think we can get away without the ending preposition? I think this still works? Non-blocking.

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,328 @@
...
+        'nav.twitter': 'Twitter',
+      },
+      fr: {
+        'nav.home': 'Acceuil',
+        'nav.about': 'À propos',
+      },
+    } as const;
+    ```
+    
+2. Create a helper function to detect the page language based on the current URL and a helper to let us get translations strings for different parts of our UI in `src/i18n/utils.ts`:
```
sarah11918, line 188:
```suggestion
2. Create two helper functions: one to detect the page language based on the current URL, and one helper to get translations strings for different parts of the UI in `src/i18n/utils.ts`:
```

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,328 @@
...
+    
+    export function useTranslations(lang: keyof typeof ui) {
+      return function t(key: keyof typeof ui[typeof defaultLang]) {
+        return ui[lang][key] || ui[defaultLang][key];
+      }
+    }
+    ```
+
+    :::note[Did you notice?]
+    In step 1, the `nav.twitter` string was not translated to French. The `useTranslations` helper here will return the default language’s value if a key is not translated so that French users will still see “Twitter”.
```
sarah11918, line 208:
```suggestion
    In step 1, the `nav.twitter` string was not translated to French. You may not want every term translated, such as proper names or common industry terms. The `useTranslations` helper will return the default language’s value if a key is not translated. In this example, French users will also see “Twitter” in the nav bar.
```
Something like this? I think the explicit call out, "Hey! You don't have to translate everything!" is potentially helpful here.

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,328 @@
...
+    :::
+    
+3. Use our helpers to choose the UI string that corresponds to the current language. For example, a nav component might look like:
```
sarah11918, line 211:
```suggestion
3. Import the helpers where needed and use them to choose the UI string that corresponds to the current language. For example, a nav component might look like:
```

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,328 @@
...
+        </li>
+        <li>
+            <a href="https://twitter.com/astrodotbuild">
+              {t('nav.twitter')}
+            </a>
+        </li>
+    </ul>
+    ```
+
+4. Each page should have a `lang` attribute on the `<html>` element that matches the language on the page. In this example, a [reusable layout](/en/core-concepts/layouts/) grabs the language from the route:
```
sarah11918, line 240:
```suggestion
4. Each page must have a `lang` attribute on the `<html>` element that matches the language on the page. In this example, a [reusable layout](/en/core-concepts/layouts/) gets the language from the current route:
```
is this accurate enough?

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -0,0 +1,328 @@
...
+            <meta name="viewport" content="width=device-width" />
+            <title>Astro</title>
+        </head>
+        <body>
+            <slot />
+        </body>
+    </html>
+    ```
+
+    Using this base layout will make sure our pages use the correct `lang` attribute automatically.
```
sarah11918, line 263:
```suggestion
    You can then use this base layout to ensure that pages use the correct `lang` attribute automatically.
```

sarah11918 (commented):
Looks good @Jutanium and @delucis! I am SO EXCITED for this! The top requested recipe!  :rocket:

Some tiny suggestions to consider, of course, but I look forward to people using this for their weekend hacking!  :raised_hands:

elvis-pereira:
Really amazing recipe! I just discovered Astro and started a project using it, this recipe couldn't have come at a better time as i18n is a requirement for this specific project. That said, I really think something like this should show how to handle untranslated pages, ideally in these cases serving a fallback page in another language is much better than returning a 404 page.

src/content/docs/en/recipes/i18n.mdx
delucis:
```suggestion
4. Each page must have a `lang` attribute on the `<html>` element that matches the language on the page. In this example, a [reusable layout](/en/core-concepts/layouts/) extracts the language from the current route:
```

delucis:
>  I really think something like this should show how to handle untranslated pages, ideally in these cases serving a fallback page in another language is much better than returning a 404 page.

Thanks for the suggestion @elvis-pereira! I can imagine adding a “Connecting pages between languages” sections for this perhaps, which maybe could also cover [Lorenzo’s comment about the language picker](https://github.com/withastro/docs/pull/2895#discussion_r1146795333). The challenge is how to tell Astro that two pages are connected, so that we can know if one is missing or not. If you use the same slugs across languages (e.g. `/en/about`, `/fr/about`, `/zh/about` etc.) it would work one way, but otherwise you’d need extra frontmatter to wire this up. I’d suggest we tackle this either as an additional section or a new recipe once we have this one published.

sarah11918 (approved):
The perfect way to end a quarter.... on to Q2, with this recipe under our belts!

PR #2905, opened 2023-03-22
Fix error docgen script for updated error data

delucis:

#### Description

- Get our error docgen script working again.
- Now that `error-data.ts` in the Astro repo uses the `satisfies` keyword we needed to update `typescript` and `tsm`.
- Changing to an object literal in `error-data.ts` instead of `defineErrors({...})` in https://github.com/withastro/astro/commit/1e7d6af41219644516757a1bf21d36cdb8795f50 changed how JSDoc parsed `comment.longname` so I’ve switched to `comment.name` to preserve the previous behaviour we had.
- Updating to TS 5.0 uncovered a few type issues in some of our custom remark plugins. Simplified a few spots where we were still theoretically handling AFMD to fix this.


pnpm-lock.yaml
```diff
@@ -380,11 +380,34 @@ packages:
       - supports-color
     dev: true
 
-  /@astrojs/mdx/0.17.2_astro@2.0.17:
+  /@astrojs/markdown-remark/2.1.1_astro@2.1.4:
```
Princesseuh, line 383: Hmm, how come there's two versions of `@astrojs/markdown-remark`?

delucis: Hmm. Maybe we bumped in `astro` but we’re not on the latest of the MDX integration? 🤔

delucis: Yup, bumping MDX to latest moves us back to just one copy of markdown-remark.

Princesseuh (approved):

Issue #2871, opened 2023-03-16
Avoid hydration mismatch error in tutorial


delucis:
Our [`Greeting.jsx` component in the tutorial](https://docs.astro.build/en/tutorial/6-islands/1/#include-a-preact-greeting-banner) randomises its initially displayed greeting opening the door to a hydration mismatch on the client. We should avoid this by making the initially displayed greeting predictable, e.g.
```diff
-  const [greeting, setGreeting] = useState(randomMessage());
+  const [greeting, setGreeting] = useState(messages[0]);
```

Discord report: https://discord.com/channels/830184174198718474/830184175176122389/1086023239723130971

sarah11918:
Fixed by @BryceRussell ! Not the first time I've written that today! :partying_face:

PR #2908, opened 2023-03-22
Fix hydration mismatch error in tutorial example

BryceRussell:

#### Description

**Closes**: #2871 

Avoid hydration mismatch error in tutorial example, initialize greeting state using the first message in messages instead of a random message


delucis (approved):
Thanks @BryceRussell!

PR #2910, opened 2023-03-23
Update assets guide to mention that images in the frontmatter are relative to the current file

Princesseuh:

- New or updated content

#### Description

https://github.com/withastro/astro/pull/6627 changes how frontmatter images work so they're now relative to the current file instead of `src/assets`. People can still use `src/assets` through the provided alias `~/assets/` or through pathing to it relatively.


sarah11918 (approved):
:partying_face:

PR #2922, opened 2023-03-24
Redirect `/` to `/en/getting-started/` in Netlify config

delucis:

- Something else!

#### Description

I noticed we don’t currently have a server-side redirect for folks visiting https://docs.astro.build

Instead, the page has to load, then the `<meta>` redirect is used to take us to the English landing page. This can be a little slow, so I’ve updated the Netlify config to do this redirect on the server instead of in the browser.


sarah11918 (approved):
Thanks Chris! This feels like a Friday in docs... :partying_face:

PR #2926, opened 2023-03-25
Feat (Rehype ExternalLink): Add the external link system on the Rehype function

pioupia:

#### Description

This PR allows users to open external links in a new tab.

### DIFF
--- a/plugins/rehype-i18n-autolink-headings.ts
+++ b/plugins/rehype-i18n-autolink-headings.ts
@@ -21,15 +21,25 @@ export function rehypei18nAutolinkHeadings() {
 
 		// Find anchor links
 		visit(tree, 'element', (node) => {
-			if (node.tagName === 'a' && node.properties?.class === 'anchor-link') {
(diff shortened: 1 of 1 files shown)


delucis:
Thanks for this @pioupia but we intentionally avoid opening links in a new tab/window in general. Most browsers make it easy to choose to open a link in a new tab, but if we force this behaviour on someone, they can’t easily choose to open a link in the _same_ tab. Because opening in a new window is mainly a user preference, in this case it’s preferable to stick to the default behaviour. Hope that makes sense!

pioupia:
> Thanks for this @pioupia but we intentionally avoid opening links in a new tab/window in general. Most browsers make it easy to choose to open a link in a new tab, but if we force this behaviour on someone, they can’t easily choose to open a link in the same tab. Because opening in a new window is mainly a user preference, in this case it’s preferable to stick to the default behaviour. Hope that makes sense!

No problem, I understand perfectly! And it didn't take me long, so it's ok.

PR #2944, opened 2023-03-28
fix: Fix InferGetStaticPropsType example

Princesseuh:

#### Description

Our tooling does not automatically uses `type Params` to type `Astro.params`. It never has, actually. Additionally, the two slash commands were in the wrong spot

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -35,7 +35,7 @@ Additionally, our templates include an `env.d.ts` file inside the `src` folder t
 /// <reference types="astro/client" />
 ```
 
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM! _Will_ we ever auto-type `Astro.params`? I guess it’s less important and also using params in a component would still need the `as` so maybe best just to leave it.

Yan-Thomas (approved):

Princesseuh:
> LGTM! _Will_ we ever auto-type `Astro.params`? I guess it’s less important and also using params in a component would still need the `as` so maybe best just to leave it.

Maybe! We still have this old accepted RFC to implement eventually https://github.com/withastro/roadmap/blob/main/proposals/0026-component-interface.md

sarah11918 (approved):
Thanks, @Princesseuh! LGTM!

PR #2952, opened 2023-03-28
i18n(de): translate environment-variables.mdx

sarasate:

- Translated content

#### Description

- Translate environment-variables.mdx


src/content/docs/de/guides/environment-variables.mdx
```diff
@@ -0,0 +1,111 @@
...
+Erstelle einfach eine `.env` Datei im Projektverzeichnis und füge ihr einige Variablen hinzu.
+
+```ini title=".env"
+# Diese Funktion ist nur verfügbar, wenn sie auf dem Server ausgeführt wird!
+DB_PASSWORD="foobar"
+# Das wird überall verfügbar sein!
+PUBLIC_POKEAPI="https://pokeapi.co/api/v2"
+```
+
+Mehr über `.env`-Dateien findest du in der Vite-Dokumentation (https://vitejs.dev/guide/env-and-mode.html#env-files).
```
Yan-Thomas, line 49: Pointing out that this link here is broken! Could you change it so the link wraps the equivalent text in German? It's below this heading in English if that helps! 🙌 
https://docs.astro.build/en/guides/environment-variables/#env-files

Yan-Thomas:
Sorry for the back-and-forth, we recently had this added to this page, could you add to the PR? After this will be good to be merged! https://github.com/withastro/docs/commit/5f678d6b3fc60d40abab6089c4e55a67ac093b6b

sarasate:
@Yan-Thomas No prob, will check that later today..

sarasate:
@Yan-Thomas There is a problem with the anchor link in that paragraph, it's missing completely in the german version of the `configuration-reference.mdx` file because they are out of sync (by a big margin). What to do in that case?

sarasate:
Ok, I'll remove the anchor for now. There is a difference of about 250 lines between the en/de versions of that file, seems to be a bigger task.

Yan-Thomas (approved):
LGTM, thanks @sarasate!

PR #2961, opened 2023-03-29
fix: use new shape for image()

Princesseuh:

- New or updated content

#### Description

https://github.com/withastro/astro/pull/6703 changes the shape of `image()` to come from `schema` instead. The underlying technical reasons are highlighted in the PR


Yan-Thomas (approved):
Love the change, nice work @Princesseuh, Astro PR not merged yet but LGTM!

sarah11918:
Just adding that since our (stable) content collections page already doesn't show a schema using assets, I don't think this change requires us to update anywhere else. The only other page to keep consistent with this one is the Image Service API reference page. So if there's no change also required there, then this is fine to docs!

PR #2963, opened 2023-03-30
fix(typescript): Update getStaticPaths option to be fully typed

Princesseuh:

- New or updated content

#### Description

This updates the `getStaticPaths` example in the TypeScript guide to be fully typed. The syntax does look weird because of `satisfies`, but it's important to show that this is the best way to type `getStaticPaths`, because typing it normally (using `: GetStaticPaths`) breaks `InferGetStaticParamsType` (and its Props equivalent) since it works by inferring the return value (which typing it directly removes the ability to)

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -244,9 +244,9 @@ You can get the type of [`Astro.params`](/en/reference/api-reference/#astroparam
 
 ```astro title="src/pages/posts/[...slug].astro" {2,14-15}
 ---
-import { InferGetStaticParamsType, InferGetStaticPropsType } from 'astro';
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM!

Can you use `satisfies` in a return type? (Couldn’t find it at all in the handbook… only in the announcement blog post.) Would look less weird if we were only annotating the result, but obviously that’s not helpful if you also want the parameters typed.

Princesseuh:
> LGTM!
> 
> Can you use `satisfies` in a return type? (Couldn’t find it at all in the handbook… only in the announcement blog post.) Would look less weird if we were only annotating the result, but obviously that’s not helpful if you also want the parameters typed.

I believe you can, but in the case of `getStaticPaths`, you do want the params typed (for `paginate`)

Yan-Thomas (approved):
LGTM!

PR #2964, opened 2023-03-30
update pnpx to pnpm dlx

ElianCodes:

#### Description

Switches `pnpx` for `pnpm dlx` since `pnpx` is deprecated [documented here](https://pnpm.io/cli/dlx)

### DIFF
--- a/src/content/docs/en/guides/cms/tina-cms.mdx
+++ b/src/content/docs/en/guides/cms/tina-cms.mdx
@@ -23,7 +23,7 @@ To get started, you'll need an existing Astro project.
       </Fragment>
       <Fragment slot="pnpm">
       ```shell
-      pnpx @tinacms/cli@latest init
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks, @ElianCodes! Just checking: does this require someone to have upgraded to the latest version of pnpm? And if so, is it reasonable to assume that they have? (You know how sometimes we have different instructions for different versions of things?)

Just want to make sure that this is now-universal advice, then it's good to go!

ElianCodes:
Hey @sarah11918, nope! The `pnpm dlx` command has been there since some time now! So even on 'outdated' pnpm versions, the `pnpx` command is also backwards compatible and still works, but is not recommended to use anymore. So we should be good to merge this!

(the `pnpx` was actually deprecated since version 6.x, most users are on 7.x or 8.x)

PR #2969, opened 2023-03-31
Fixed InferGetStaticPropsType example to use Props type

marceloverdijk:

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -261,7 +261,7 @@ type Props = InferGetStaticPropsType<typeof getStaticPaths>;
 
 const { slug } = Astro.params as Params;
 //               			 ^? { slug: string; }
-const { title } = Astro.props;
+const { title } = Astro.props as Props;
 //                			^? { draft: boolean; title: string; }
 ---
 ```


delucis:
Thanks for this @marceloverdijk. Astro's language server actually automatically knows that Astro.props has the type of a local `Props` type so this shouldn't be necessary. Are you seeing a situation where that doesn't work?

marceloverdijk:
@delucis yes I noticed an issue in VS code:

When having just:

```
  const drivers = await prisma.driver.findMany({
    select: {
      id: true,
    },
  })
  return drivers.map((driver) => {
    return {
      params: { id: driver.id },
    }
  })
}

type Params = InferGetStaticParamsType<typeof getStaticPaths>

const { id } = Astro.params // so without as Params
```

VS code will state that the `id` is of type: `const id: string | undefined`
and it gives me other type issues using that `id`.

When changing it to:

```
const { id } = Astro.params  as Params
```

then the type inferring works as expected and the `id` has the correct type.

Note that Astro [docs](https://docs.astro.build/en/guides/typescript/) also do that:

```
---
import { InferGetStaticParamsType, InferGetStaticPropsType } from 'astro';

  return posts.map((post) => {
    return {
      params: { slug: post.slug },
      props: { draft: post.data.draft, title: post.data.title },
    };
  });
}

type Params = InferGetStaticParamsType<typeof getStaticPaths>;
type Props = InferGetStaticPropsType<typeof getStaticPaths>;

const { slug } = Astro.params as Params;
//                      ^? { slug: string; }
//                      ^? { draft: boolean; title: string; }
---
```

Note that the docs only use `as Params` for the `Astro.params` but not `as Props` for the `Astro.props` (https://github.com/withastro/docs/pull/2969)

marceloverdijk:
PS I initially created this PR which is related: https://github.com/withastro/astro.build/pull/640

delucis:
Right, we don’t currently auto-type `Astro.params`. But we _do_ for `Astro.props` which is what this PR is proposing changing, right?

marceloverdijk:
Ah I see, then you are absolutely right. Sorry for confusion.

I've updated https://github.com/withastro/astro.build/pull/640 as well to only include the `as Params` part in the blog post.

PR #2970, opened 2023-03-31
New recipe: Add icons to external links

TheOtterlord:

- New or updated content

#### Description

- What does this PR change? Adds a new recipe for adding icons to external links present in Markdown. (Recipe suggestion from https://github.com/withastro/docs/discussions/2894#discussioncomment-5465122)


src/content/docs/en/recipes/external-links.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: Add icons to external links
```
Jutanium, line 2:
```suggestion
title: Add icons to external links
```

src/content/docs/en/recipes/external-links.mdx
```diff
@@ -0,0 +1,45 @@
...
+description: Learn how to add icons to external links in your Markdown files
+i18nReady: false
+type: recipe
+---
+
+Using a Rehype plugin, you can add icons to external links in your Markdown files, as well as open those links in a new tab.
```
Jutanium, line 8:
```suggestion
Using a rehype plugin, you can identify links in your markdown files that point to external sites and modify them. This includes adding icons to the end of each link.
```

src/content/docs/en/recipes/external-links.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+## Recipe
+
+1. Install the `rehype-external-links` plugin.
+
+    ```shell
+    npm install rehype-external-links
+    ```
+
+2. Add the plugin to your `astro.config.mjs` file. Pass the plugin options by passing an array with the plugin as the first item and the options as the second item. Set the content type to `text`, or `raw` to use HTML as the content.
```
Jutanium, line 21:
```suggestion
2. Import the plugin into your `astro.config.mjs` file. Pass it to the `rehypePlugins` array, along with an options object. Add a content property and set its `type` to `text` if you want to add plain text to the end of the link, or `raw` to add HTML to the end of the link.
```

src/content/docs/en/recipes/external-links.mdx
```diff
@@ -0,0 +1,45 @@
...
+    export default defineConfig({
+      // ...
+      markdown: {
+        rehypePlugins: [
+          [
+            rehypeExternalLinks,
+            {
+              target: '_blank',
+              rel: ['noopener', 'noreferrer'],
+              content: { type: 'text', value: ' 🔗' }
```
Jutanium, line 36:
```suggestion
              content: { type: 'text', value: ' 🔗' }
```

For the purposes of this example, we don't need to set these attributes to add an icon.

Additionally, the package warns us to not set `target`, and setting `rel` to these values is only appropriate if `target` is `_blank`.

src/content/docs/en/recipes/external-links.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: Add icons to external links
+description: Learn how to add icons to external links in your Markdown files
```
Jutanium, line 3:
```suggestion
description: Learn how to install a rehype plugin to add icons to external links in your Markdown files
```

Jutanium (commented):
Hey, thanks for the recipe!

I do think this task is rather niche - I think it is more likely that someone would want to add an icon to ALL links, and this recipe can't help with that.

But, that could just be me!

I'm happy to include this recipe, and we can link to it from the rehype plugin docs to show a concrete example of installing a rehype plugin. To me, _that_ is the most valuable part of this recipe.

Yan-Thomas:
> Accessibility concerns: Adding an icon using content, without alt text, may not be the most screenreader-friendly solution.
There is an alt argument for content but it does not have full browser support.

From what I know, some screen readers are able to distinguish the external by default like JAWS, and making a distinction on internal/external links is not really an accessibility recommendation/guideline, but sure, an welcome additional info.

The main accessibility issue here is how descriptive your links are. External links and their surrounding text should clarify they're going to take you to another domain, for example: 
> Read more about [useEffect() on React.js documentation](https://react.dev/reference/react/useEffect).

My course of action here would probably just add an Accessibility warning, explaining the [content property's accessibility issues](https://developer.mozilla.org/en-US/docs/Web/CSS/content#accessibility_concerns) and that your content itself should help make it clear if a link is external. 

@TheOtterlord, feel free to ping me if you need help adding this!

sarah11918 (commented):
Thanks @TheOtterlord for this recipe! I'm so excited that people are getting involved and contributing!

I think I noticed "Yan is typing" so we'll see how much our comments conflict!  :laughing: Will ask both @Jutanium and @Yan-Thomas to take a peek at my suggestions!

sarah11918:
Haha @TheOtterlord , we were both adding a note at the same time!  :laughing:

TheOtterlord:
@Yan-Thomas I've added an accessibility note which I think best explains it, as well as linking to the MDN section you shared. I wasn't sure whether to go beyond this explanation, mostly because I didn't want it to seem too lengthy

> The value of the content property is [not represented in the accessibility tree](https://developer.mozilla.org/en-US/docs/Web/CSS/content#accessibility_concerns). As such, it’s best to make clear that the link is external in the surrounding content, rather than relying on the icon alone.

sarah11918:
I'm going to let @Yan-Thomas have the final say in the note, @TheOtterlord , and merge this PR when he's happy with it.

I think my preference is to have the note AFTER the example though, so it doesn't split up the instruction from the code example.

sarah11918 (approved):
Once @Yan-Thomas makes the final decision with the accessibility note, it's his to merge!

TheOtterlord:
> I think my preference is to have the note AFTER the example though, so it doesn't split up the instruction from the code example.

Yeah, that would probably read better when following the recipe. 

> Once @Yan-Thomas makes the final decision with the accessibility note, it's his to merge!

Awesome :+1:

Yan-Thomas:
Agree with @sarah11918 here, let's up the note after the example @TheOtterlord 🙌, after this ready to merge

TheOtterlord:
Done :+1:

<details>
<summary>Preview</summary>

</details>

Yan-Thomas (approved):
LGTM!

PR #2972, opened 2023-03-31
[WRITING.md] adds recipe writing examples

sarah11918:
Adds some initial recipe writing guidance to our existing writing guide.

This is to be added to over time, but captures some recent discussion/decisions in our official guide!

### DIFF
--- a/WRITING.md
+++ b/WRITING.md
@@ -49,6 +49,57 @@ As a general guide for writing tone, you can follow the [Google Developers Guide
 
 Also see tips on how to [write inclusive documentation](https://developers.google.com/style/inclusive-documentation) and [write accessible documentation](https://developers.google.com/style/accessibility)
 
+## Recipe Writing
+
+Here are some tips and examples to match Astro Docs' style for instructional content, like a recipe:
+
+### Imperative (instruction, command) tense
+
+Whenever possible, give the reader a direct instruction:
+
+e.g. Type the command...
+
+Do not use _"Let's..."_ or _"Next, we will..."_. You are not sitting there with your reader! (This also means you will never need the words _we_, _us_, and _our_. If you see them, rephrase!)
+
+If you must address the reader, it is OK to use "you" and "your". Especially for emphasis in very important steps where something could go wrong:
+
+e.g. Before continuing, you must check your... or else...
+
+### Avoid "story telling"
+
+Recipes should be commands to follow, as concisely and directly as possible. Do not tell a story around what is happening, but do provide context or clarity if needed as a brief goal/benefit/reason.
+
+Here is an example edit we made to one of our own recipes to illustrate the difference:
+
+**Example:** Introduce a set of instructions with a "Big Step" (e.g. overall goal) + reason before listing specific steps to follow
+
+**Before: narrative style**
+"As well as needing your content in different languages, you will often need to translate labels for UI elements around your site. We can do this by creating dictionaries of terms instead of hard-coding text in one langugae in our templates."
+
+1.
+2. ...
+
+**After: imperative style**
+"Create dictionaries of terms to translate the labels for UI elements around your site. This allows your visitors to experience your site fully in their language."
+
+1.
+2. ...
+
+
+### Opinionated Instructions
+
+Sometimes you will need to give an instruction where the reader could make a different choice than you, and the recipe would still work. You will need to illustrate the example with the decision you made, but you want to make it clear that other decisions are acceptable. In this case, try to separate the instruction from the opinion, then state the choice that your example uses. Your reader will be able to first process what you are doing, and then will see the choice you have made separately and can make a similar or different choice in their own project. 
+
+This is more reassuring than a vauge "it could go anywhere!" statement. Additionally now you do not need to further explain the filename of a code snippet, that this file will not necessarily exist in their project, etc.
+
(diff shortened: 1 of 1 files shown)


ElianCodes (approved):
I really like this! LGTM!

WRITING.md
```diff
@@ -49,6 +49,60 @@ As a general guide for writing tone, you can follow the [Google Developers Guide
...
+
+### Imperative (instruction, command) tense
+
+Whenever possible, give the reader a direct instruction:
+
+e.g. Type the command...
+
+Do not use _"Let's..."_ or _"Next, we will..."_. You are not sitting there with your reader! (This also means you will never need the words _we_, _us_, and _our_. If you see them, rephrase!)
+
+If you must address the reader, it is OK to use "you" and "your". Especially for emphasis in very important steps where something could go wrong:
```
Yan-Thomas, line 64: I think here the preposition should be "on"? Feel free to disregard.
```suggestion
If you must address the reader, it is OK to use "you" and "your". Especially for emphasis on very important steps where something could go wrong:
```

WRITING.md
```diff
@@ -49,6 +49,60 @@ As a general guide for writing tone, you can follow the [Google Developers Guide
...
+
+e.g. Before continuing, you must check your... or else...
+
+### Avoid "story telling"
```
Yan-Thomas, line 68: I think that "storytelling" is only one word.
```suggestion
### Avoid "storytelling"
```

WRITING.md
```diff
@@ -49,6 +49,60 @@ As a general guide for writing tone, you can follow the [Google Developers Guide
...
+>
+> 1.
+> 2. ...
+
+
+### Opinionated Instructions
+
+Sometimes you will need to give an instruction where the reader has options. You want to illustrate the example with the specific choice you made, but you also want to make it clear that other decisions are acceptable. 
+
+In this case, try to separate the instruction from the opinion. First give the action to take with a more general description. Then state the opinionated choice that your example uses. Your reader will be able to first process what you are doing, and then will see the choice you have made. They can still follow your instruction, making a choice that works in their own project.
```
Yan-Thomas, line 93: Comma and grammar nitpicking
```suggestion
In this case, try to separate the instruction from the opinion. First, give the action to take with a more general description. Then state the opinionated choice that your example uses. Your reader will be able to first process what you are doing and then will see the choice you have made. They can still follow your instruction, making a choice that works for their own project. 
```

WRITING.md
```diff
@@ -49,6 +49,60 @@ As a general guide for writing tone, you can follow the [Google Developers Guide
...
+> 2. ...
+
+
+### Opinionated Instructions
+
+Sometimes you will need to give an instruction where the reader has options. You want to illustrate the example with the specific choice you made, but you also want to make it clear that other decisions are acceptable. 
+
+In this case, try to separate the instruction from the opinion. First give the action to take with a more general description. Then state the opinionated choice that your example uses. Your reader will be able to first process what you are doing, and then will see the choice you have made. They can still follow your instruction, making a choice that works in their own project. 
+
+This can be easier to follow (and perhaps more reassuring!) than a statement that gives multiple options. This is also usually easier to write, since you do not have to remind the reader that this file is from *your* example, and may not appear in their project.
```
Yan-Thomas, line 95: Comma nitpick again 😅 
```suggestion
This can be easier to follow (and perhaps more reassuring!) than a statement that gives multiple options. This is also usually easier to write since you do not have to remind the reader that this file is from *your* example, and may not appear in their project.
```

Yan-Thomas (approved):
Love it @sarah11918, I added a few grammar/comma suggestions, content LGTM! 🙌

dreyfus92 (commented):
What do you think of changing this?

WRITING.md
```diff
@@ -49,6 +49,60 @@ As a general guide for writing tone, you can follow the [Google Developers Guide
...
+
+Do not use _"Let's..."_ or _"Next, we will..."_. You are not sitting there with your reader! (This also means you will never need the words _we_, _us_, and _our_. If you see them, rephrase!)
+
+If you must address the reader, it is OK to use "you" and "your". Especially for emphasis in very important steps where something could go wrong:
+
+e.g. Before continuing, you must check your... or else...
+
+### Avoid "story telling"
+
+Recipes should be commands to follow, as concisely and directly as possible. Do not tell a story around what is happening, but do provide context or clarity if needed as a brief goal/benefit/reason.
```
dreyfus92, line 70:
```suggestion
Recipes should be a set of instructions to follow, as concisely and directly as possible. Do not tell a story around what is happening, but do provide context or clarity if needed as a brief goal/benefit/reason.
```

WRITING.md
sarah11918: I think "in a step"... the same way we'd say "Do this in step 3" (At least, that's how I was envisioning the sentence, as being "inside a particular step" while you are writing it as opposed to "putting empasis ON step 3")

WRITING.md
sarah11918: I think both in and for are good here (since like above, I'm taking the "being inside a step" or "working inside the files of your project" view when writing recipes) so either works for me! But, I cannot argue with your commas.  :smile:

WRITING.md
sarah11918: Thanks for this!

PR #2981, opened 2023-04-04
grammar: an spa => a spa

d-sanderson:

#### Description

~~- Closes # ~~
Fixes a small grammatical error in the MPAs vs. SPAs section. Changed "an SPA" to "a SPA"

~~- Did you change something visual? A before/after screenshot can be helpful.~~


delucis:
“An SPA” is correct here — the rules around when “an” or “a” should be used can vary but cases like this where the word _sounds_ like it starts with a vowel, e.g. “Ess-P-A”, can use “an”.

Thank you for the PR but we’ll stick with the current version.

PR #3005, opened 2023-04-06
Add RecipeLink component and link to recipes

Jutanium:

- New or updated content

#### Description

- Adds a RecipeLink component that accepts recipe slugs and renders link(s) to the recipe(s)
- Waiting for @ximenavf92's recipe design to undraft this PR


src/components/RecipeLinks.astro
```diff
@@ -0,0 +1,42 @@
...
+	}
+	if (slug.endsWith('/')) {
+		corrected = corrected.slice(0, -1);
+	}
+	return corrected;
+}
+
+const slugs = Astro.props.slugs.map(convertSlug);
+
+const entries = await Promise.all(slugs.map((slug) => getEntryBySlug('docs', slug)));
```
TheOtterlord, line 21: We could throw a custom error to make it clear what's wrong when you mistype a slug. The current error thrown is `Cannot read properties of undefined (reading 'slug')`
```suggestion
const entries = await Promise.all(slugs.map(async (slug, i) => {
  let entry = await getEntryBySlug('docs', slug)
  if (!entry) throw Error(`Could not find entry for slug "${Astro.props.slugs[i]}"`)
  return entry
}));
```
<br/>
<details>
<summary>Before</summary>
<br/>

</details>
<details>
<summary>After</summary>
<br/>

</details>

src/components/RecipeLinks.astro
```diff
@@ -0,0 +1,42 @@
+---
+import { getEntryBySlug } from 'astro:content';
+
+interface Props {
```
TheOtterlord, line 4: Fix eslint error
```suggestion
export interface Props {
```

TheOtterlord (commented):
Leaving my 2 cents here. Can't wait to have this component in the docs! The ownership on recipes is amazing

src/components/RecipeLinks.astro
Jutanium: Thanks, @TheOtterlord!

sarah11918:
Love you jumping in here, @TheOtterlord!

@Jutanium, let's use this draft PR for adding recipe links to the other existing recipes, then consider this one good (even if it doesn't incorporate any design from Ximena yet.) We can make a separate PR to update the `RecipeLinks` component itself.

src/components/RecipeLinks.astro
```diff
@@ -0,0 +1,76 @@
...
+const entries = await Promise.all(
+	slugs.map(async (slug, i) => {
+		let entry = await getEntryBySlug('docs', slug);
+		if (!entry) throw Error(`Could not find entry for slug "${Astro.props.slugs[i]}"`);
+		return entry;
+	})
+);
+
+const isList = entries.length > 1;
+const noun = isList ? 'Recipes' : 'Recipe';
```
sarah11918, line 30:
```suggestion
const noun = isList ? 'recipes' : 'recipe';
```
I saw you changed this intentionally, but are we sure about the case here? Recipe isn't a proper noun for us, so I think lowercase is more consistent.

sarah11918 (commented):
Looking good @Jutanium ! I noticed that if there's only one recipe, then everything is all on one line in the component. I think I like that! One small point re: capitalization, but checking through, this all looks good!

Did you want to document using the recipe component in a separate PR?

src/components/RecipeLinks.astro
```diff
@@ -0,0 +1,76 @@
...
+	})
+);
+
+const isList = entries.length > 1;
+const noun = isList ? 'recipes' : 'recipe';
+---
+
+<div class="root">
+	<div class="flex">
+		<img src="/houston_chef.png" width="26" alt="Chef Houston" />
```
delucis, line 35: Would love to hear what @Yan-Thomas thinks, but I’m not sure we want “Chef Houston” to be read out here. Imagine a screen reader: “Chef Houston Related recipes: ...” I think we can treat the image as purely decorative:

```suggestion
		<img src="/houston_chef.png" width="26" alt="" />
```

src/components/RecipeLinks.astro
```diff
@@ -0,0 +1,76 @@
...
+
+<style>
+	.title {
+		font-weight: 600;
+		font-size: 1rem;
+		color: #c7c5d3;
+	}
+	.flex {
+		display: flex;
+		align-items: center;
```
delucis, line 60: We should let items wrap here to avoid this effect on narrow viewports:
image


```suggestion
		align-items: center;
		flex-wrap: wrap;
```

src/components/RecipeLinks.astro
```diff
@@ -0,0 +1,76 @@
...
+	}
+	.flex > * {
+		margin-right: 0.25rem;
+	}
+	.flex > img {
+		margin-right: 0.5rem;
+	}
```
delucis, line 67: Let’s do the spacing with `gap` here:
```suggestion
		gap: 0.25rem;
	}
	.flex > img {
		margin-right: 0.25rem;
	}
```

src/components/RecipeLinks.astro
```diff
@@ -0,0 +1,76 @@
...
+	ul {
+		margin-left: 0.5rem;
+	}
+	div.root {
```
delucis, line 71: Nit, but no reason to specify the element here:
```suggestion
	.root {
```

public/houston_chef.png
delucis, line 1: Can we compress this? I’d drop it in https://squoosh.app/, make it 50% smaller (so it’s ~52px wide), and optimize it.

delucis (changes requested):
Thanks for building this out, Dan! Chef Houston is definitely a fun touch to be making an appearance in docs. Made a bunch of suggestions but mostly just cleaning up code and fixing bugs — love the idea in general 🙌

src/components/RecipeLinks.astro
```diff
@@ -0,0 +1,64 @@
...
+	}
+	ul {
+		margin-left: 0.5rem;
+	}
+	.root {
+		background-color: var(--theme-bg-offset);
+		border-radius: 8px;
+		padding: 24px;
+		width: max-content;
+		max-with: 100%;
```
delucis, line 62: Oops, I typo’d in my suggestion, sorry!
```suggestion
		max-width: 100%;
```

delucis (approved):
LGTM! 👨‍🍳🤖

Issue #2975, opened 2023-04-01
Broken images url for `Build a Blog` tutorial


dreamorosi:
Following the tutorial "Build a Blog", at the step 2.2 - Write your first Markdown blog post, and section [Write Markdown Content](https://docs.astro.build/en/tutorial/2-pages/2/#write-markdown-content) the reader creates their first blog post.

The post contains an image in the front matter:
```yaml
image:
    url: 'https://astro.build/assets/blog/astro-1-release-update/cover.jpeg' 
    alt: 'The Astro logo with the word One.'
```

The url for the image returns a 404 and so the layout of the page at [step 4.2](https://docs.astro.build/en/tutorial/4-layouts/2/#try-it-yourself---customize-your-blog-post-layout) is incomplete.

The same happens for the other two images in the other two posts of the tutorial blog.

If one of the maintainers could indicate a replacement images to use I'd be happy to contribute a tiny PR with the fix.

TheOtterlord:
#2865 related

sarah11918:
Thanks for this! We merged a PR to fix. :partying_face:

Issue #2865, opened 2023-03-16
Update all broken assets


Jutanium:
Search for `astro.build/assets` (e.g. tutorial, [buffer example](https://docs.astro.build/en/core-concepts/endpoints/#static-file-endpoints) and replace them with live assets, hosted in the docs repo.

sarah11918:
completed!

PR #3016, opened 2023-04-08
Fix broken `astro.build/asset` links

BryceRussell:

#### Description

**Closes** https://github.com/withastro/docs/issues/2865, https://github.com/withastro/docs/issues/2975

- Replace broken `https://astro.build/assets` links with new working links
- Update alt texts

### DIFF
--- a/src/content/docs/en/guides/assets.mdx
+++ b/src/content/docs/en/guides/assets.mdx
@@ -97,8 +97,8 @@ Your existing images in `public/` and remote images are still valid but are not 
 <img src="/images/stars.png" alt="A starry night sky.">
 
 <!-- Remote image on another server -->
-![Astro logo](https://astro.build/assets/logo.png)
-<img src="https://astro.build/assets/logo.png" width="25" alt="Astro logo">
+![Astro logo](https://astro.build/assets/press/logomark-light.png)
+<img src="https://astro.build/assets/press/logomark-light.png" width="25" alt="Astro logo">
 ```
 
 ### Convert from `@astrojs/image`
--- a/src/content/docs/en/guides/images.mdx
+++ b/src/content/docs/en/guides/images.mdx
@@ -25,7 +25,7 @@ The `src` attribute is required, and its format will depend on where your images
 import rocket from '../images/rocket.svg';
 ---
 <!-- Remote image on another server -->
-<img src="https://astro.build/assets/logo.png" width="25" alt="Astro">
+<img src="https://astro.build/assets/press/logomark-light.png" width="25" alt="Astro">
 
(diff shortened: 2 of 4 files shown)


BryceRussell:
I saw that the issue mentioned hosting these assets inside the docs repo itself but that would make the links a bit long and messy like: `https://raw.githubusercontent.com/withastro/docs/main/public/default-og-image.png`. I updated all the broken links to live `/assets` links for now but I can add images to repo and use those instead if desired, let me know.

sarah11918:
Hey @BryceRussell this is so helpful! But I looked through, and all the updated links (except arc and rays) on the docs pages are *not* pointing to the new images you added and they 404? Can you please double the link content on these?

And, if we change the tutorial content, then we also need to change the tutorial repo (and all the branches properly) that mirrors the project that the tutorial builds. So we'll need to loop @delucis on that once these link examples here are figured out!

BryceRussell:
Oops, thanks @sarah11918. I fixed the file names, so the images are now visible.

For the changes to the tutorial page, should I make a PR to `withastro/blog-tutorial-demo` with the changes?

BryceRussell:
How important are the tutorial blog images? Can they be repeated?

If they can't be repeated the tutorial repo has an extra 4th post not mentioned in docs that would require another image

I think maybe it would be better to delete the 'rays' and 'arc' images entirely and only repeat the full logo image for all 4 blog post frontmatters in the tutorial

delucis:
Let’s also sync with @ximenavf92 on this one as I know she was working on some assets for this!

BryceRussell:
Ahh I see, awesome! I initially misread the issue when I submitted the PR so I tried shoehorning the public assets in to save it 😅

sarah11918:
In interest of getting this fixed so that people will stop telling us the tutorial image links won't work, I'd like to merge this!  :sweat_smile: 

@BryceRussell can you make an issue in the actual tutorial blog example repo ( https://github.com/withastro/blog-tutorial-demo ) saying that we'll need to properly apply these changes throughout the various branches (or wait until we have permanent assets and do it then)? Then, ask either @delucis or @Jutanium how these changes need to be worked back in there and you could PR that repo, too!  :raised_hands: 

(I'm OK if the tutorial content works, even if there's a bit of a delay in catching up the example repo.)

BryceRussell:
I think it would be best to wait for dedicated assets before updating the tutorial repo to avoid any complications. (Unless there is a reason they cant be out of sync for a bit)

sarah11918:
Yeah, I personally think being out of sync is the lesser evil than the tutorial code not working when people copy it, so let's merge this now, and we'll update the repo (and these pages again) when we get resources!

PR #3034, opened 2023-04-12
docs: front matter -> frontmatter

stormynight9:
Minor content fixes: all of Astro's docs use "frontmatter" instead of "front matter"


delucis (approved):
Well spotted @stormynight9! Thanks and welcome to the Astro docs contributors 🥳

PR #3096, opened 2023-04-25
docs: update assets related pages for new configuration shape

Princesseuh:

- New or updated content

#### Description

https://github.com/withastro/astro/pull/6848 updates the configuration setting for the image service to be an object of the following shape : `service: {entrypoint: string, config: Record<string, any>}` instead of just being a string. This allow user to pass configuration to the image service. For instance, Vercel uses this to provide a list of allowed widths.

This PR updates the Assets and Image Service pages to include this new information and all the associated utilities it added. Additionally, this PR adds an example for `validateOptions`, as it was previously missing.

### DIFF
--- a/src/content/docs/en/reference/image-service-reference.mdx
+++ b/src/content/docs/en/reference/image-service-reference.mdx
@@ -34,10 +34,19 @@ An external service points to a remote URL to be used as the `src` attribute of 
 import type { ExternalImageService } from "astro";
 
 const service: ExternalImageService = {
-  getURL(options: ImageTransform) {
+  validateOptions(options: ImageTransform, serviceConfig: Record<string, any>) {
+    // Enforce the user set max width.
+    if (options.width > serviceConfig.maxWidth) {
+      console.warn(`Image width ${options.width} exceeds max width ${serviceConfig.maxWidth}. Falling back to max width.`);
+      options.width = serviceConfig.maxWidth;
+    }
+
+    return options;
+  },
+  getURL(options, serviceConfig) {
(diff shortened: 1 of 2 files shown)


src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -116,20 +125,20 @@ At build time for static sites and pre-rendered routes, both `<Image />` and `ge
 
 In dev mode and SSR mode, Astro doesn't know ahead of time which images need to be optimized. Astro uses a GET endpoint (by default, `/_image`) to process the images at runtime. `<Image />` and `getImage()` pass their options to `getURL()`, which will return the endpoint URL. Then, the endpoint calls `parseURL()` and passes the resulting properties to `transform()`.
 
-#### getConfiguredImageService
+#### getConfiguredImageService & imageServiceConfig
```
sarah11918, line 128: @Jutanium - How do you feel about these two being combined into one heading? Ideally, I think we'd normally split them up, and define/explain each one separately.

Since you went through these examples extensively the first time around, do you have a suggestion for a sensible way to devote space to each one separately that doesn't force us to duplicate examples and make this long and terrible?

Not blocking, because the example below is entirely reasonable. But I do think it would be preferable if each were treated separately, if we had a good way of doing it.

sarah11918 (commented):
I'll ask @Jutanium to also review this, since he has the most experience with the original. I've left a comment that's more for him than for you, @Princesseuh and otherwise, I'm happy when Dan's happy!  :partying_face:

PR #3101, opened 2023-04-26
Backend as a Service Stubs

Jutanium:

- New or updated content
- 
#### Description

- Adds a new recipe section called Backend as a Service
- Adds stubs for Supabase, Firebase, Appwrite and Tigris

@sarah11918: **I'd like to change this to either "add a backend" or "backend services".** Thoughts?


src/i18n/en/nav.ts
```diff
@@ -78,6 +78,22 @@ export default [
...
+	{ text: 'Connecting a CMS', slug: 'guides/cms', key: 'guides/cms' },
+	{ text: 'Backend as a Service', slug: 'guides/baas', key: 'guides/baas' },
+	{ text: 'Images', slug: 'guides/images', key: 'guides/images' },
+	{ text: 'Fonts', slug: 'guides/fonts', key: 'guides/fonts' },
+	{
+		text: 'Sharing State',
+		slug: 'core-concepts/sharing-state',
+		key: 'core-concepts/sharing-state',
+	},
+	{ text: 'RSS', slug: 'guides/rss', key: 'guides/rss' },
```
sarah11918, line 96: This whole edit is messed up, since it was probably from the original PR before we had recipes.

Make sure it's only adding the backend section to recipes, and nothing else is changed.

sarah11918:
Just checking in on the priorities of this one here:

- @Jutanium - priority is to go back and rename all the things so that everything uses some form of "backend" instead of "baas".
- @kevinzunigacuellar - I can see you've started your Firebase recipe here, but I think more strategically, we should focus on getting this merged with three stubs first. And, I think Dan should be in a position to merge this today.


So, I'm going to suggest we strip out the Firebase content from this PR here, and make that a separate PR so that writing and reviewing it doesn't have to hold up the infra merge. This would also be consistent with the guidance we've given AppWrite, since content will take longer than the structure. Thanks for all your speedy work so far Kevin! Let's make it its own PR!  :partying_face:

src/content/docs/en/guides/backend.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: Use a backend service with Astro
+description: How to use a backend service to add authentication, storage and data
+i18nReady: true
+---
+import BackendGuidesNav from '~/components/BackendGuidesNav.astro';
+
+**Ready to add a backend to your Astro project?** Follow one of our guides to integrate a backend service.
```
sarah11918, line 8:
```suggestion
**Ready to add auth, storage or data to your Astro project?** Follow one of our guides.
```

PR #3102, opened 2023-04-26
Add PUBLIC_TWO_LANG flag to build two translations only

bluwy:

#### Description

- Add PUBLIC_TWO_LANG flag to build two translations only.
- Also upgraded Astro to 2.3.1, which should bring slightly faster dev server startups.

The new flag is added so that Astro's smoke test can test building without taking a lot of CI time


src/content.ts
```diff
@@ -1,7 +1,17 @@
 import { getCollection } from 'astro:content';
 import { isEnglishEntry, isRecipeEntry, isTutorialEntry } from './content/config';
 
-export const allPages = await getCollection('docs');
+export const allPages = await getAllPages();
```
delucis, line 4: What do you think of doing this instead of the extra function?

```suggestion
export const allPages = await getCollection('docs', (doc) => (
	import.meta.env.PUBLIC_ONLY_EN ? isEnglishEntry(doc) : true
));
```

src/content.ts
```diff
@@ -1,7 +1,17 @@
...
+
+async function getAllPages() {
+	const pages = await getCollection('docs');
+
+	if (import.meta.env.PUBLIC_ONLY_EN) {
+		return pages.filter(isEnglishEntry);
+	} else {
+		return pages;
+	}
+}
```
delucis, line 17: In which case… 🥳 
```suggestion
```

src/i18n/languages.ts
```diff
@@ -17,4 +17,13 @@ export default {
 	ru: 'Русский',
 } as const;
 
+if (import.meta.env.PUBLIC_ONLY_EN) {
+	// @ts-expect-error This conflicts with the const type above, but it's fine when building
```
delucis, line 21: This feels weird, but I guess it’s better than `languages = import.meta.env.PUBLIC_ONLY_EN ? { ... } : { ... }`?

delucis (commented):
Mostly looks OK — we could also add one additional language alongside English to be a little more realistic while saving time (a newer one like Korean or Polish could be a good choice) as discussed on Discord.

src/content.ts
bluwy: Ah that looks a lot better!

src/i18n/languages.ts
bluwy: I didn't want the types to be affected, but I can try the ternary to see if it builds fine

bluwy: I had to use a `Record<string, string>` to make `pnpm tsc` work. Using `as const` would make the types having `[key]?: string` (undefined-able), which makes typechecking fail. Happy to walk back to the previous hack too if you prefer to keep the `as const` type.

delucis (approved):
Looks good to me! Thanks @bluwy.

bluwy:
Made a small typo 😬 Thanks for the review!

delucis (approved):
I approve even more now that that’s fixed! 😅

PR #3111, opened 2023-04-28
document behavior of script tags in JSX expressions

MoustaphaDev:

- New or updated content

#### Description

Closes https://github.com/withastro/astro/issues/6905

Scripts inside JSX expressions are treated as inline scripts tags. I believe this is because the compiler doesn't extract these scripts because we wouldn't be able to know whether or not to include them in the head until the component template is run.

All these scripts will be inline scripts 
```tsx
{
    <script>
        alert('Hello Astro!');
    </script>
}
{
    false && <script>
       alert(''Unreachable")
    </script>
}
{
    true && <script>
        // will throw since inline scripts aren't processed to compile TS -> JS
        const name = document.getElementById('name') as HTMLInputElement;
    </script>
}
```


Princesseuh:
I'm not sure why my specific review was asked for, but this looks good to me 👍

I wonder if there would be value in documenting workarounds, though those also come with caveats

MoustaphaDev:
I'm sorry if the ping was too broad, but glad to hear it looks good

sarah11918:
Give me some time to think about this, @MoustaphaDev , because you've changed the focus of the note now away from the idea of adding an attribute to a script tag (and them what happens when you do that), and it now is starting to list different things that happen to have the same effect / end result, putting the focus on the things that make scripts not bundled.

It's a subtle difference, but it really does change the original note's purpose. So, I've been thinking about how to best include this new information.

MoustaphaDev:
I understand, I've seen the use of two consecutive admonitions in some pages of the docs. We can document this in a “warning” block maybe?

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -73,7 +73,7 @@ To avoid bundling the script, you can use the `is:inline` directive.
 ```
 
 :::note
-Adding `type="module"` or any attribute other than `src` to a `<script>` tag will disable Astro's default bundling behavior, treating the tag as if it had an `is:inline` directive.
+Adding `type="module"` or any attribute other than `src` to a `<script>` tag will disable Astro's default bundling behavior, treating the tag as if it had an `is:inline` directive. The same will be true when the script is written in a JSX expression.
```
sarah11918, line 76:
```suggestion
Astro's default bundling behavior will be disabled in some situations. In particular, adding `type="module"` or any attribute other than `src` to a `<script>` tag will cause Astro to treat the tag as if it had an `is:inline` directive. The same will be true when the script is written in a JSX expression.
```
What do you think about something like this, @MoustaphaDev ?

MoustaphaDev: Looks good!

PR #3134, opened 2023-05-01
Fix the example code errors in the Endpoints section

Ekarmore:

#### Description

- I found some issues in the example code in the [params and Dynamic routing](https://docs.astro.build/en/core-concepts/endpoints/#params-and-dynamic-routing) section while browsing through the documentation. The expected result mentioned in the documentation for the file path `src/pages/[id].json.ts` and the routes generated by the `getStaticPaths()` function is _**“This will generate three JSON endpoints at build time: `/api/1.json, /api/2.json, /api/3.json`.”**_  However, the actual result generated by the example code is **_`/0.json, /1.json, /2.json`._**

- So I modified the file path in the sample code in the document to `scr/pages/api/[id].json.ts`, and changed the expected generated JSON endpoints in the document to the actual ones that will be produced, namely /api/0.json, /api/1.json, /api/2.json.

- I found that this part has the same issue in other language versions as well.


delucis (approved):
Thanks @Ekarmore and for the review @ElianCodes!

PR #3141, opened 2023-05-02
Document Code component new inline prop

bluwy:

- New or updated content

#### Description

Related astro PR: https://github.com/withastro/astro/pull/6959

Document `<Code inline` to output inline code instead of a code block.


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -880,10 +880,12 @@ import { Code } from 'astro/components';
 ---
 <!-- Syntax highlight some JavaScript code. -->
 <Code code={`const foo = 'bar';`} lang="js" />
-<!-- Optional: customize your theme. -->
+<!-- Optional: Customize your theme. -->
 <Code code={`const foo = 'bar';`} lang="js" theme="dark-plus" />
 <!-- Optional: Enable word wrapping. -->
 <Code code={`const foo = 'bar';`} lang="js" wrap />
+<!-- Optional: Output inline code. -->
+<Code code={`const foo = 'bar';`} lang="js" inline />
```
sarah11918, line 888:
```suggestion
Here is <Code code={`const foo = 'bar';`} lang="js" inline /> available inline.
```
I'm not sold on the wording, but I think demonstrating that the component is meant to be used inline, as per the [original, original illustration of the toggle](https://github.com/withastro/astro/issues/5595#issuecomment-1397731722) is more helpful. If you can think of a better sentence, please do!  :sweat_smile:

bluwy: I kinda miss the alignment though with this 😄 maybe if we want to show it's inline, how about:
```suggestion
<span>
  Here is an inline example:
  <Code code={`const foo = 'bar';`} lang="js" inline />
</span>
```

delucis: Wouldn’t a block-level wrapper element like `<p>` make more sense than `<span>` here?

```suggestion
<p>
  <Code code={`const foo = 'bar';`} lang="js" inline />
  will be rendered inline.
</p>
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -880,10 +880,12 @@ import { Code } from 'astro/components';
 ---
```
delucis, line 880: I can’t comment on the appropriate line above, but this code example would probably be easier to understand if we highlighted the extra attributes in the “optional” examples by adding metadata to the opening of the code block: https://github.com/withastro/docs/blob/main/WRITING.md#text-highlighting

src/content/docs/en/reference/api-reference.mdx
bluwy: Ah you're right. I thought the issue was nesting block elements inside inline elements, which `span` easily identifies as an inline element, but the issue seems to be specific to `<p>` with it's spec strictly preventing block elements while `<span>` is more permissive.

src/content/docs/en/reference/api-reference.mdx
bluwy: Done! I'm actually surprised how flexible the highlighter syntax is to handle regexes 😄 

image

delucis: We have @hippotastic to thank for that 💜

delucis (approved):
Looks good to me! Thanks @bluwy 🚀

ElianCodes (approved):
LGTM 🚀🥳

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -874,16 +874,21 @@ The Markdown component is no longer built into Astro. See how to [import Markdow
...
-<!-- Optional: customize your theme. -->
+<!-- Optional: Customize your theme. -->
 <Code code={`const foo = 'bar';`} lang="js" theme="dark-plus" />
 <!-- Optional: Enable word wrapping. -->
 <Code code={`const foo = 'bar';`} lang="js" wrap />
+<!-- Optional: Output inline code. -->
+<p>
+  <Code code={`const foo = 'bar';`} lang="js" inline />
+  will be rendered inline.
+</p>
```
sarah11918, line 891:
```suggestion
<p> Render <Code code={`const foo = 'bar';`} lang="js" inline /> inline.</p>
```
I kind of want to push back here.  :sweat_smile: I think the real reason to show any kind of example is to see it actually inline inside a sentence. It doesn't *look* obviously inline in the example above (even though it is).

I would suggest something like above, or if we *really* value consistency of visuals/formatting above all, then go back to the original `<Code code={`const foo = 'bar';`} lang="js" inline />` since this is a reference page and it can be more condensed.

bluwy: I think the `<p>` tag already signals that it's an inline content, so I'm not sure if we need to wrap with words to further emphasize it. Otherwise I'd also opt for the original example for consistency, but I do like the `<p>` tag there as an example. But these are also only my thoughts though.

sarah11918: Good enough! Everyone else LGTM'd this, so let's release it as is!

PR #3154, opened 2023-05-03
docs(endpoints): added new example for returning buffers from endpoints

haasal:

New content/code-snippet in the docs

#### Description

The PR adresses [this closed issue](https://github.com/withastro/astro/issues/5903).
When using the encoding: 'binary' providers like vercel send the buffer as  plain text instead of a binary png. This can be prevented by using the Response object (just like in other frameworks like svelte)


sarah11918 (commented):
Thanks for contributing this! I followed your and @Princesseuh's conversation trail, and I tried to make the note more direct.

Erika, can you please fact-check this? And, I know you mentioned something about adding documentation about this. Would you consider this note sufficient, or do you think we should consider replacing the existing example with one that returns a `Response`?  Is that something that will work in all situations, so is maybe more universal to show?

haasal:
In my opinion the example I added should be the default way to return Buffers in general. Maybe we should replace the 'binary' example completely (as it doesn't work)

sarah11918:
I'm going to ask @delucis what he thinks about this!

delucis:
Thanks for your patience on this this @haasal! Looking at [the original `astro` issue](https://github.com/withastro/astro/issues/5903), it looks like you were experiencing this with an SSR adapter, right?

I believe the current example is for static scenarios (where Astro needs to write the endpoint to the filesystem while building). I’m not sure if we actually support returning `Response` in static mode or not. We should double check `Response` works in a static scenario and if it doesn’t document clearly these two situations: 1) object with `body` and `encoding` in static mode and 2) `Response` in SSR.

haasal:
Yes. Now I see it too. However it mentions nowhere in the docs how to do it with SSR on, which is really important (I.e. s3 buckets). So I would simply add my paragraph/example slightly modified to the SSR part of the endpoint guide

sarah11918:
@haasal Would you be willing to update this PR to modify the paragraph and example slightly, as you suggested, and move it to the SSR section?

haasal:
Sure. I'll add the example to the SSR section.

sarah11918:
Hi @haasal - Just checking in here to see whether you're still up for this! I'm sorry, I've been so swamped getting next weeks' big launch docs together for our new features, I've been a little slow to follow up on community PRs.

But we would be so happy to have your contribution, if you're still interested!

haasal:
I think this should be ready. I messed up something with git along the way. Please check if everything is correct before merging.

src/content/docs/en/core-concepts/endpoints.mdx
```diff
@@ -127,8 +126,20 @@ export async function get({ params }) {
   });
 }
 ```
+
 This will respond to any request that matches the dynamic route. For example, if we navigate to `/helmet.json`, `params.id` will be set to `helmet`. If `helmet` exists in the mock product database, the endpoint will use create a `Response` object to respond with JSON and return a successful [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/API/Response/status). If not, it will use a `Response` object to respond with a `404`.
 
+In **SSR** mode certain providers require the `Content-Type` header to return an image. In this case, use a `Response` object to specify a `headers` property like in the example above.
```
sarah11918, line 132:
```suggestion
In SSR mode, certain providers require the `Content-Type` header to return an image. In this case, use a `Response` object to specify a `headers` property. For example, to produce a binary `.png` image:
```
This looks good, and well-placed, thank you! I've made some slight adjustments because now this is in the SSR section, I don't think SSR needs to be bolded here. And, someone may not have read the above section on static endpoints, so I think we can keep the description self-contained.

I'm just going to commit this so we can go ahead and merge it!  :raised_hands:

PR #3044, opened 2023-04-13
i18n(pt-BR): Small translation updates

Yan-Thomas:
#### What kind of changes does this PR include?

- Translated content

#### Description

PT-BR translation updates for:
- `astro-pages.mdx`
- `endpoints.mdx`
- `framework-components.mdx`
- `layouts.mdx`
- `routing.mdx`
- `editor-setup.mdx`
- `client-side-scripts.mdx`
- `data-fetching.mdx`

PR #3162, opened 2023-05-04
Refactor `<RecipeLink>` to handle i18n

delucis:

#### Description

- Updates `<RecipeLink>` to work better with our i18n system
  - Slugs can be set using their translation path, e.g. `pt-br/recipes/foo`, even if the translation doesn’t exist yet — we’ll grab the English title in this case.
  - When a recipe hasn’t been translated yet, the link will be shown with an `(EN)` suffix on non-English pages like we do for other links on the site.

Here’s an example of the component on the Polish endpoints guide, which hasn’t been translated yet: 
image

This PR should unblock proper `<RecipeLink>` usage in #3044.

### DIFF
--- a/src/components/RecipeLinks.astro
+++ b/src/components/RecipeLinks.astro
@@ -1,39 +1,56 @@
 ---
-import { getEntryBySlug } from 'astro:content';
-import { removeLeadingSlash, removeTrailingSlash } from '~/util';
+import { recipePages } from '~/content';
+import { getLanguageFromURL, stripLangFromSlug } from '~/util';
 
 export interface Props {
 	slugs: string[];
 }
 
-const slugs = Astro.props.slugs.map((slug) => removeLeadingSlash(removeTrailingSlash(slug)));
+const lang = getLanguageFromURL(Astro.url.pathname);
+const fallbackSlug = (slug: string) =>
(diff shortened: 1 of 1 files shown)


Yan-Thomas (approved):
Tested locally and works great, LGTM!

src/components/RecipeLinks.astro
```diff
@@ -1,39 +1,56 @@
...
+const firstRecipe = recipes[0];
+if (!firstRecipe) {
+	throw new Error('No slugs passed to `<RecipeLinks>` component. Make sure you pass at least one.');
+}
 ---
 
 <div class="root">
 	<div class="flex">
 		<img src="/houston_chef.webp" width="26" alt="" />
 		<strong>Related {noun}:</strong>
```
Yan-Thomas, line 37: Non-blocking, but I'm wondering if we can do something to allow translating these? I know singular/plural translation stuff can get tricky in some languages, though we could just use something like "Related recipe(s)" and allow translators to add the entire thing even if it's not perfect.

delucis: Oh yeah! Good point. Let’s do it in a separate PR so we can at least get your translation PRs unblocked.

Simplest here is probably to have two translation keys, one for singular, one for plural — “Related recipe” and “Related recipes” in English — to accommodate whatever plurals look like in each language.

PR #3184, opened 2023-05-05
Optimize static mdx

bluwy:

#### Description

Optimize static subtrees of the MDX intermediate hast as `set:html` prop to reduce final AST size


delucis (approved):
This code looks great to me, the build times do seem to be improved, and I wasn’t able to spot anything broken in the site, so great work @bluwy!

PR #3187, opened 2023-05-06
Sort tutorial pages by directory name only

morinokami:

#### Description

Changed the sorting of tutorial pages in `getTutorialPages` to use the name of the directory containing the content, rather than the path to the directory. This prevents pages from being out of order when translating some of the tutorial pagess into languages other than en.

For example, consider the case where you have created the file `src/content/docs/ja/tutorial/0-introduction/index.mdx` to start translating the tutorial into Japanese. When retrieving the list of pages with `getTutorialPages`, the original implementation compares the string `"ja/tutorial/0-introduction"` to the string starting with `"en"` (e.g., `"en/tutorial/1-setup"`). As a result, `"ja/tutorial/0-introduction"` always comes after the other en pages, which is clearly wrong. To correct this behavior, we need to change the sorting to use only directory names.


sarah11918:
Thanks for this, @morinokami !  :raised_hands: I'm going to ask @delucis to review this one!

delucis (approved):
Absolutely makes sense. Thanks for this improvement @morinokami!

PR #3189, opened 2023-05-06
Properly decode html entities in copy button

### LINKED ISSUE #3186: Entity code copied from tutorial snippets instead of `<` HTML symbol
In the first few units of tutorial, if you try to copy the snippet, it pastes to VS Code with entity code for `<` symbols.

Here is example site: [https://docs.astro.build/en/tutorial/1-setup/3/](https://docs.astro.build/en/tutorial/1-setup/3/)

Copied and pasted code looks like this:
```
---
---

&#x3C;html lang="en">
  &#x3C;head>
    &#x3C;meta charset="utf-8" />
    &#x3C;link rel="icon" type ="image/svg+xml" href="/favicon.ico" />
    &#x3C;meta name="viewport" content="width=device-width" />
    &#x3C;meta name="generator" content = {Astro.generator} >
    &#x3C;title>Astro&#x3C;/title>
  &#x3C;/head>
  &#x3C;body>
    &#x3C;h1>Astro&#x3C;/h1>
  &#x3C;/body>
&#x3C;/html>
```

MoustaphaDev:

#### Description

- Closes #3186
Use the well tested [html-entities](https://www.npmjs.com/package/html-entities) npm package to handle HTML entities decoding for the copy button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "@fontsource/ibm-plex-mono": "^4.5.10",
     "@nanostores/preact": "^0.1.3",
     "canvas-confetti": "^1.6.0",
+    "html-entities": "^2.3.3",
     "jsdoc-api": "^7.1.1",
     "nanostores": "^0.5.13",
     "rehype-autolink-headings": "^6.1.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10,6 +10,9 @@ dependencies:
   canvas-confetti:
     specifier: ^1.6.0
     version: 1.6.0
+  html-entities:
+    specifier: ^2.3.3
+    version: 2.3.3
   jsdoc-api:
     specifier: ^7.1.1
     version: 7.1.1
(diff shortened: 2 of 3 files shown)


sarah11918:
Thanks for this, @MoustaphaDev! I'm guessing from the error message that we'll also need to install `html-entities`? It's not currently in the docs `package.json`.

sarah11918:
Ah, we might have an ignore rule to prevent people PRing extra packages etc. We might need to edit that file manually, first. So if you run into trouble, try that!

MoustaphaDev:
I didn't push the `package.json` modifications, should be good now!

delucis:
Haven't looked closely but can the `html-escaper` package we already use do the same thing?

MoustaphaDev:
`html-entities` can handle a wider range of special characters than `html-escaper` (because of that, it's heavier)

MoustaphaDev:
Just realized it might be way too heavy to add it for a simple copy button
See [`html-entities`](https://bundlephobia.com/package/html-entities@2.3.3) and [`html-escaper`](https://bundlephobia.com/package/html-escaper@3.0.3) on Bundlephobia

MoustaphaDev:
Closing this for now as completed in #3206, could be an inspiration the current fix actually doesn't cover all special characters

PR #3198, opened 2023-05-07
how about add korean in translation languages

WooDaeHyun:
there's some pages already translated in korean.
so please add korean in README.md 🤗
it can be make more passion about astro to korean developers! 
thanks for read my propose! 

propose to add korean about astro translation languages

### DIFF
--- a/TRANSLATING.md
+++ b/TRANSLATING.md
@@ -47,7 +47,7 @@ Currently we are aiming to translate the Astro documentation into the following 
 - German
 - Japanese
 - Spanish
-- korean
+- Korean
 
 ### Why can Astro only support a few languages?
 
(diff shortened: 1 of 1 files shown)


TRANSLATING.md
```diff
@@ -47,6 +47,7 @@ Currently we are aiming to translate the Astro documentation into the following
 - German
 - Japanese
 - Spanish
+- Korean
```
delucis, line 50: Good point! We should also add Polish while we’re here:

```suggestion
- Korean
- Polish
```

WooDaeHyun: yes that's also good point! bro! 👍

delucis (approved):
Thanks for this update. Welcome to the Astro docs contributors @WooDaeHyun! 🚀

PR #3201, opened 2023-05-07
Fix HTML entity encoding issue in copy button

0xWerz:

- Fixing the doc **copy code** button.

#### Description

- Update copy button functionality to properly decode HTML entities using the html-entities library, allowing users to copy unencoded HTML code to the clipboard


0xWerz:
duplicated PR #3189

PR #3202, opened 2023-05-07
New recipe: Call endpoints from the server

### LINKED ISSUE #2043: No documentation about fetching Astro endpoints on the server
On the page [data fetching](https://docs.astro.build/en/guides/data-fetching/) there is no mention how to fetch data from endpoints made using Astro. This caused me some problems with fetching while in production, as fetching a `/api/example` doesn't work.

I finaly stumbled upon the [Astro.url ](https://docs.astro.build/en/reference/api-reference/#astrourl) Api reference. We can use the origin to assemble the url. The docs could provide a simillar snippet like this.

```astro
---
const books = await fetch(new URL("/api/books", Astro.url.origin)).then((res) => res.json());
---

<!--Display content--!>
```

TheOtterlord:

- New or updated content

#### Description

- Closes #2043
- What does this PR change? Adds a recipe to explain calling endpoints from the server, without using fetch or needing to separate out code.


Yan-Thomas (approved):
LGTM, thanks for adding this @TheOtterlord! I believe it might be a good moment for us to add the RecipeLinks component to the Data Fetching page with a link for this recipe? I'll let @sarah11918 decide on this!

sarah11918:
Yes! I'm looking at this now @TheOtterlord and wondering whether there might be a good place for this on the Endpoints page too, actually?

Where do you think it would be good to link this from?

src/content/docs/en/recipes/call-endpoints.mdx
```diff
@@ -0,0 +1,37 @@
+---
+title: Call endpoints from the server
+description: Learn how to call endpoints from the server in Astro.
+i18nReady: true
+type: recipe
+---
+
+## Recipe
```
sarah11918, line 8:
```suggestion
Calling an endpoint in Astro allows you to... This recipe uses a server endpoint to ... (VERY SPECIFICALLY WHAT EXACTLY THIS EXAMPLE DOES)

## Prerequisites

- A project with [SSR](/en/guides/server-side-rendering/) (output: 'server') enabled

## Recipe
```
Just an example of an introductary paragraph and of a Prerequisites section. If there are no prerequisites, let's still include the section, and just say `- none`!

Note that this recipe should be something very specific, that someone can follow through entirely and add something new to their page. So, it shouldn't be "here's what adding an endpoint is like..." It should be: "Follow these steps and your project will now do this." So, we do have to tell them what exactly is going to happen. It's then on the reader to adjust as necessary for their purposes to do something that is useful to them.  :)

src/content/docs/en/recipes/call-endpoints.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+1. Create an endpoint (e.g. `src/pages/api/hello.ts`):
```
sarah11918, line 10:
```suggestion
1. Create an endpoint in a new file `src/pages/api/hello.ts` that returns some data:
```
We should be specific about what this code is doing

src/content/docs/en/recipes/call-endpoints.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+    ```ts title="src/pages/api/hello.ts"
+    import type { APIRoute } from 'astro'
+
+    export const get: APIRoute = () => {
+      return {
+        body: JSON.stringify({
+          hello: 'world',
```
sarah11918, line 18:
```suggestion
          greeting: 'Hello',
```
So, I didn't like that "world" is not a reasonable value for "hello"  ??  :sweat_smile: Plus, since you use this to recreate the text "Hello world" it's not great to have duplicating words because it's not as clear tracing through that things are entirely separate.

src/content/docs/en/recipes/call-endpoints.mdx
```diff
@@ -0,0 +1,37 @@
...
+        }),
+      }
+    }
+    ```
+
+2. Import the method from the endpoint & call it from a page (e.g. `src/pages/index.astro`). Await the result in the case it's a promise, and convert it to a `Response` if it's not already. Then, you can use the `Response` methods to get the data you need:
```
sarah11918, line 24:
```suggestion
2. On any Astro page  (e.g. `src/pages/index.astro`), import the `get()` method from the endpoint. Then, you can use the `Response` methods to get the data you need:

  ```astro title="src/pages/index.astro"
  ---
  import { get } from './api/hello.ts'
  let response = await get(Astro)
  const data = await response.json()
  ---
  <h1>{data.greeting} world!</h1>
  ```

```
OK, there was a lot going on in that instruction. AND, since this is YOUR recipe, you should know whether or not it's a promise/response? So, I'm going to suggest cutting that, making the example more simple. 

You could include another step that showing the full example saying that depending on how much is known about the data you're fetching, you could also add a line to convert to a response, if it's not already one.

Also, we haven't explained what "Astro" is in that example above?

src/content/docs/en/recipes/call-endpoints.mdx
```diff
@@ -0,0 +1,37 @@
...
+    import { get } from './api/hello.ts'
+
+    let res = await get(Astro)
+    if (!(res instanceof Response)) res = new Response(res.body, res)
+
+    const data = await res.json()
+    ---
+
+    <h1>Hello {data.hello}</h1>
+    ```
```
sarah11918, line 37: If you are using this method to fetch data from an API and are unsure about the type of data you'll receive, you can add a line to convert it to a `Response` if it's not already. 

``````suggestion {5}
```astro title="src/pages/index.astro" 
---
import { get } from './api/hello.ts'

let response = await get(Astro)
if (!(response instanceof Response)) res = new Response(response.body, response)

const data = await response.json()
---

<h1>{data.greeting} world!</h1>
```
``````
You could do something like this to show adding the new line

sarah11918 (commented):
Hi @TheOtterlord! Sorry I'm just now getting around to reviewing some of the older community contributions, after getting all the launch docs reviewed and ready!

Here are some comments below so that we can whip this one into shape. Have a read through and see what you can come up with!  :raised_hands:

TheOtterlord:
Sorry it took so long to get back :sweat_smile: 
Added the recipe link to the bottom

sarah11918:
OK, back to this now, and the example looks great @TheOtterlord 

I'm going to suggest that on the endpoints page itself, though, that this recipe go much higher. Right now, it looks like it's intended to relate to "redirects" and I think this is a much more straight ahead example than that.

Maybe put it right before the note in the introduction? What do you think about that?

TheOtterlord:
The note in the introduction to [server endpoints](https://deploy-preview-3202--astro-docs-2.netlify.app/en/core-concepts/endpoints/#server-endpoints-api-routes)?

sarah11918:
Yes, I'm thinking in here, right before the note where it starts talking all about params and stuff: 
image

sarah11918 (approved):
LGTM!

Issue #3115, opened 2023-04-29
Issues with z.date() in config.ts


thdoan:
Can the documentation provide more details here?

https://docs.astro.build/en/guides/content-collections/#defining-datatypes-with-zod

```
    // In frontmatter, dates written without quotes around them are interpreted as Date objects
    publishDate: z.date(),
```

I tried various "dates written without quotes" in my .md files and they all resulted in _Expected type `"date"`, received "string"_ errors when building. Example:

**post-1.md**
```
---
publishDate: 4/27/2023
---
```

If there's an exact format we need to use, please specify it in the documentation. I eventually resolved it by updating `config.ts` to this:

```
    publishDate: z.coerce.date(),
```

delucis:
Ah yes, this needs to be something that YAML considers is a date/timestamp, like `2023-04-27`. This is because Markdown frontmatter is usually written in YAML. The [YAML spec](https://yaml.org/spec/1.2.2/#24-tags) has some more examples of valid timestamps in Example 2.22.

thdoan:
@delucis this is great info. Would it be possible to add this to the documentation as I think it would be helpful for those new to Astro/zod.

sarah11918:
The line directly under the one you referenced does show a date string, so I do think that there's a visual example right in that section for what a date in YAML frontmatter should look like.

What I think actually addresses this problem better is something that specifically introduces YAML on the Markdown/MDX page. We really only teach/introduce what the frontmatter *is* in the tutorial. I do think our `markdown-content` page could include mention of the top section of a Markdown file being written in YAML and should adhere to that language.

So I'm going to close this issue, but make a task to add some YAML content to our Markdown page. I think something there can helpful for more than just this one specific case. Thanks for pointing this out!

PR #3203, opened 2023-05-08
mention YAML in Markdown

sarah11918:
In response to #3115 , someone was having trouble identifying the correct date format to use in their blog posts. They were not using a date format recognized by YAML, and suggested we define a proper date format to use in docs.

To address this issue more generally, I think adding mention that frontmatter *is* YAML (which we only do in the tutorial) is more broadly helpful. This defines a Markdown file's frontmatter as being YAML right at the top of the Markdown/MDX page, so that people will have a resource for knowing the language of the frontmatter, and a reference link for usage.


Yan-Thomas (approved):
Nice idea, it's something a lot of people don't know about Markdown in Astro that is useful to add, LGTM!

PR #3219, opened 2023-05-09
add recipes as search category

sarah11918:
Updates our search category definitions:

- All "special" guides, including new ones that never got special categories, as well as individual recipes, are now "Recipes" (instead of Deploy Guide, Integration Guide, e.g.)
- Also categorizes our Upgrade Guides (to v1, v2)

Will need to re-run Algolia to pick up the changes. 

Currently:
("Deploy Guide" too specific, "Learn" too general. Both will now show up under "Recipes")
firebase


delucis (commented):
Nice work @sarah11918! I tested this by looking at the source code on some of the impacted pages and finding the element with `id="docsearch-lvl0"` and making sure its contents were the expected category (Algolia then looks for this hidden element when indexing to learn what the category is).

One suggestion, but then LGTM!

PR #3220, opened 2023-05-09
fix: handle inline highlight due to rehype-optimize-static

### LINKED ISSUE #3215: Some code highlighting ranges are incorrect
I found that some code highlighting ranges are incorrect, for example:

* https://docs.astro.build/en/guides/aliases/
  image
  image
* https://docs.astro.build/en/guides/styling/#frameworks-and-libraries
  image
* https://docs.astro.build/en/guides/testing/#create-your-first-playwright-test
  image
* https://docs.astro.build/en/guides/typescript/#import-aliases
  image
* https://docs.astro.build/en/guides/typescript/#type-checking
  image

To find the cause I checked the source code like

https://github.com/withastro/docs/blob/496517384b52ed6265ed319e641d26034ee2cfbb/src/content/docs/en/guides/aliases.mdx?plain=1#L11
https://github.com/withastro/docs/blob/496517384b52ed6265ed319e641d26034ee2cfbb/src/content/docs/en/guides/typescript.mdx?plain=1#L274

but the code itself that specifies the highlighting target seems to be correct 🤔

After that, since I could confirm that the same code working fine in a slightly earlier version, I searched for the commit where the problem occurred by bisection search and found that the code highlighting was shifted after the merge of the following PR:

https://github.com/withastro/docs/pull/3184 (b4532a9b)

Unfortunately, it was not clear to me why the PR above caused the problem, but I hope this information will help to resolve the issue.

bluwy:

#### Description

Fix #3215
Closes #3217 

Fix shiki inline highlight through `CodeSnippet.astro`. The code before used to anticipate rehype to encode the HTML entities, now with `rehype-optimize-static` it doesn't as it uses `set:html` instead, which skips through rehype rendering, but through Astro instead.

Before
image

After

image

NOTE: If one day we remove `rehype-optimize-static`, the highlighting offset problem will return. I can't seem to find a way to fix both at once. For now I added comments, to link the two cases together.

### DIFF
--- a/src/components/CodeSnippet/shiki-line.ts
+++ b/src/components/CodeSnippet/shiki-line.ts
@@ -1,5 +1,5 @@
 import chroma from 'chroma-js';
-import { escape, unescape } from '~/util/html-entities';
+import { unescape } from '~/util/html-entities';
 import { ensureTextContrast } from './color-contrast';
 import {
 	InlineMarkingDefinition,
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM! Sounds like @hippotastic has a long-term fix coming very soon, but I’ll merge this for now.

PR #3221, opened 2023-05-09
Improve `<RecipeLinks>` i18n support

Yan-Thomas:

- Translated content

#### Description

Adds two new UI labels for singular/plural "related recipe(s)" in `<RecipeLinks/>` component.

Translated example, from `/pt-br/guides/content/#showcasing-your-content`
image


sarah11918:
Thanks @Yan-Thomas ! I love how this takes singular and plural into account, too!

I noticed when I visited https://deploy-preview-3221--astro-docs-2.netlify.app/pt-br/guides/content/ that the link (as shown in your screenshot above) displays the English site, but the page itself DOES seem to be translated? Is it because the component will only show the exact URL passed as props, and maybe the original site uses the English link? Can you take a closer look at that?

Yan-Thomas:
>  Is it because the component will only show the exact URL passed as props, and maybe the original site uses the English link? Can you take a closer look at that?

Yes, considering this is a fallback page, the expected behavior is that it will link to the English page since it's the one in the component props, which for me looks to be alright to happen. I mean, it might be confusing that "related recipes" is translated, but that's the same behavior we have for everything that uses the `<UIString />` component in a fallback page.

sarah11918:
So, just to confirm, the *site* doesn't know that the page has been translated. The component will only render what's actually fed as props into the component, so it's up to the author to put either a translated link, if it exists or to put the English link.

One thing that comes to mind, as in the case here, if the recipe gets translated *after* the component is written, then even though a translated page exists, the reader will not be taken to it. I wonder if we have a good way of  reminding/ensuring that components should be checked to see whether they need to be updated because the page they link to has now been translated?

Yan-Thomas:
> One thing that comes to mind, as in the case here, if the recipe gets translated after the component is written, then even though a translated page exists, the reader will not be taken to it. I wonder if we have a good way of reminding/ensuring that components should be checked to see whether they need to be updated because the page they link to has now been translated?

From what I tested before in my translation PR, the link checker will cause an error if we're linking to an `en/` page inside a non-English page. So any translated page will have the links in their respective language anyway, redirecting to the fallback if they haven't been translated yet.

sarah11918:
Sounds perfect, @Yan-Thomas! Do we need @delucis to take a look at this?

sarah11918 (approved):

sarah11918:
Just pinging this one, @Yan-Thomas! Are you waiting to merge this?

Yan-Thomas:
>Just pinging this one, @Yan-Thomas! Are you waiting to merge this?

Should be fine, let's merge already!

PR #3231, opened 2023-05-11
Fix some formatting errors

delucis:

#### Description

We’ve been seeing some CI errors due to a bug in how Prettier is formatting a file. This PR refactors that file to avoid the bug (code behaviour is unchanged).

I also updated Prettier and the Prettier Astro plugin to the latest versions and re-ran the formatter which changed formatting in minor ways in a few spots.

### DIFF
--- a/src/components/tabs/AstroJSXTabs.astro
+++ b/src/components/tabs/AstroJSXTabs.astro
@@ -6,12 +6,12 @@ import Tabs from './Tabs';
 	<Fragment slot="tab.1.jsx">JSX</Fragment>
 	<Fragment slot="tab.2.astro">Astro</Fragment>
 
-	<Fragment slot="panel.1.jsx"
-		><slot name="1.jsx" />
+	<Fragment slot="panel.1.jsx">
+		<slot name="1.jsx" />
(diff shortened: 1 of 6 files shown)


src/components/tabs/AstroJSXTabs.astro
```diff
@@ -6,12 +6,12 @@ import Tabs from './Tabs';
...
-	<Fragment slot="panel.1.jsx"
-		><slot name="1.jsx" />
+	<Fragment slot="panel.1.jsx">
+		<slot name="1.jsx" />
 		<slot name="jsx" />
 	</Fragment>
-	<Fragment slot="panel.2.astro"
-		><slot name="2.astro" />
+	<Fragment slot="panel.2.astro">
+		<slot name="2.astro" />
```
Princesseuh, line 14: Hmm, that one can cause visual changes, I'm not sure why the formatter changed it automatically, afaik our hugging logic shouldn't change this unless you changed it manually

delucis:
Merging this to unblock CI. I checked the component @Princesseuh called out and the visual output seems to be OK. Let me know if anyone sees something odd!

PR #3233, opened 2023-05-11
Data collections and collection references

bholmesdev:

- New or updated content

#### Description

Document the new "data collection" and "collection reference" features. See the RFC for the original feature proposal: https://github.com/withastro/roadmap/pull/569

- Update preambles to mention both content and data collections.
- **Add new sections on defining and resolving references.** Each are added as subsections to maintain the overall flow of the guide.
- **Add the new `type` field to all `defineCollection()` examples.** Yes, `type` _will_ default to `'content'` to make this change non-breaking for existing content collections. Though to keep these docs future-proof (in case `type` becomes a required field), `type` has been added for both content and data collections. Open to adding the line "`type` defaults to `'content'` but you can specify manually" if we prefer!
- Update language to be consistent when explaining content _or_ data collections (ex. frontmatter -> "frontmatter or data file")
- Update `getEntryBySlug()` recommendations to use `getEntry()`. This allows you to query content _and_ data collections,  without needing to explain how `getEntryBySlug()` is for content collections only. `getEntryBySlug()` is still around in the API reference, with a note to use the new `getEntry()` function for ease-of-use.
- Add new sections to the API reference:
  - `getEntry()`
  - `getEntries()`
  - `defineCollection() -> type`
  - `reference()`
  - Update "Collection Entry Type"


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -61,11 +61,12 @@ If you find yourself working with different types of content, you should create
...
-  - **authors/** split different content types into new collections
-    - grace-hopper.md
-    - alan-turing.md
-    - batman.md
+  - **blog/**
+    - post-1.md
+    - post-2.md
+  - **authors/**
+    - grace-hopper.json
+    - alan-turing.json
```
bholmesdev, line 69: Originally had two collections here, one content and one data. But I realized this _might_ make it confusing what "different types of content" means; does "type" mean content vs. data or something about the structure? Three collections should hopefully make that clearer!

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -391,20 +494,20 @@ This will generate a new page for every entry in the `blog` collection. For exam
 
 ### Building for server output (SSR)
 
-If you are building a dynamic website (using Astro's SSR support), you are not expected to generate any paths ahead-of-time during the build. Instead, your page should examine the request (using `Astro.request` or `Astro.params`) to find the `slug` on-demand, and then fetch it using [`getEntryBySlug()`](/en/reference/api-reference/#getentrybyslug). 
+If you are building a dynamic website (using Astro's SSR support), you are not expected to generate any paths ahead-of-time during the build. Instead, your page should examine the request (using `Astro.request` or `Astro.params`) to find the `slug` on-demand, and then fetch it using [`getEntry()`](/en/reference/api-reference/#getentry).
```
Yan-Thomas, line 497: Not something from this PR, but since we're already here, it should be "ahead of time" instead to be consistent with other pages.
```suggestion
If you are building a dynamic website (using Astro's SSR support), you are not expected to generate any paths ahead of time during the build. Instead, your page should examine the request (using `Astro.request` or `Astro.params`) to find the `slug` on-demand, and then fetch it using [`getEntry()`](/en/reference/api-reference/#getentry).
```

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -20,9 +20,9 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
+A **content collection** is any directory inside the reserved `src/content` project directory, such as `src/content/newsletter` and `src/content/authors`. Only content collections are allowed inside the `src/content` directory. This directory cannot be used for anything else.
 
-A **content entry** is any piece of content stored inside of your content collection directory. Content entries are stored as either Markdown (`.md`) or MDX (`.mdx`) files. You can use any filename you want, but we recommend using a consistent naming scheme (lower-case, dashes instead of spaces) to make it easier to find and organize your content.
+A **collection entry** is any piece of content stored inside of your content collection directory. Entries can use content authoring formats including Markdown (`.md`) and MDX (`.mdx`), or as data formats including YAML (`.yaml`) and JSON (`.json`). You can use any filename you want, but we recommend using a consistent naming scheme (lower-case, dashes instead of spaces) to make it easier to find and organize your content.
```
Yan-Thomas, line 25: Nit: unnecessary comma
```suggestion
A **collection entry** is any piece of content stored inside of your content collection directory. Entries can use content authoring formats including Markdown (`.md`) and MDX (`.mdx`) or as data formats including YAML (`.yaml`) and JSON (`.json`). You can use any filename you want, but we recommend using a consistent naming scheme (lower-case, dashes instead of spaces) to make it easier to find and organize your content.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -754,6 +771,37 @@ This function accepts the following properties:
 
 [See the `Content Collection` guide](/en/guides/content-collections/#defining-a-collection-schema) for example usage.
 
+### `reference()`
+
+**Type:** `(collection: string) => ZodEffects<ZodString, { collection, id: string } | { collection, slug: string }>`
+
+The `reference()` function is used in the content config to define a relationship, or "reference," from one collection to another. This accepts a collection name and validates the entry identifier(s) specified in your content frontmatter or data file.
+
+This example defines references from a blog author to the `authors` collection, and an array of related posts to the same `blog` collection:
```
Yan-Thomas, line 780: Nit: unnecessary comma
```suggestion
This example defines references from a blog author to the `authors` collection and an array of related posts to the same `blog` collection:
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -778,10 +826,79 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
...
+---
+```
+
+`getEntry()` can also be used to get referenced entries to access the `data`, `body`, or `render()` properties:
+
+```astro
+---
+import { getEntry } from 'astro:content';
+
+const enterprisePost = await getEntry('blog', 'enterprise);
```
Yan-Thomas, line 861: Highlighting fix
```suggestion
const enterprisePost = await getEntry('blog', 'enterprise');
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -778,10 +826,79 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
...
+- `(Array<{ collection: string, id: string }>) => CollectionEntry<collection>`
+- `(Array<{ collection: string, slug: string }>) => CollectionEntry<collection>`
+
+`getEntries()` is a function that retrieves multiple collection entries with the same collection name. This is useful for resolving an array of [references](en/guides/content-collections/#defining-collection-references) to their associated `data`, `body`, and `render()` properties.
+
+```astro
+---
+import { getEntries } from 'astro:content';
+
+const enterprisePost = await getEntry('blog', 'enterprise);
```
Yan-Thomas, line 887: Highlighting fix part 2
```suggestion
const enterprisePost = await getEntry('blog', 'enterprise');
```

Yan-Thomas (commented):
Thanks @bholmesdev, amazing work on data collections! Added a few initial suggestions for you 🙌

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -169,9 +173,18 @@ You can use `defineCollection()` as many times as you want to create multiple sc
...
+    schema: z.object({ /* ... */ })
+  }),
+  'newsletter': defineCollection({
+    type: 'content',
+    schema: z.object({ /* ... */ })
+  }),
+  'authors': defineCollection({
+    type: 'data',
+    schema: z.object({ /* ... */ })
+  }),
```
bholmesdev, line 187: Debatable if this is needed. Happy to revert

bholmesdev:
Hm, link linter is raising an error on a new link to the `guide/#defining-collection-references` section from the API reference page. But testing it manually, the link works fine? Maybe it can't handle cross-page heading links if that heading link is new 🤔

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -778,10 +826,79 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
...
+import { getEntries } from 'astro:content';
+
+const enterprisePost = await getEntry('blog', 'enterprise');
+
+// Get related posts referenced by `data.relatedPosts`
+const captainProfile = await getEntries(enterprise.data.relatedPosts);
+---
+```
+
 ### `getEntryBySlug()`
```
sarah11918, line 894: Can we... deprecate `getEntryBySlug`? :sweat_smile:  I mean, *say that* it's deprecated/legacy? (Even if we don't actually deprecate for a loooong time.) I feel like it's just messy to keep this around, and define/explain it if another function entirely replaces it. It would be cleaner to just say update, because this will eventually go away.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -778,10 +826,79 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
...
+<p>
+  <Since v="2.5" />
+</p>
+
+**Types:**
+
+- `(Array<{ collection: string, id: string }>) => CollectionEntry<collection>`
+- `(Array<{ collection: string, slug: string }>) => CollectionEntry<collection>`
+
+`getEntries()` is a function that retrieves multiple collection entries with the same collection name. This is useful for resolving an array of [references](en/guides/content-collections/#defining-collection-references) to their associated `data`, `body`, and `render()` properties.
```
sarah11918, line 881:
```suggestion
`getEntries()` is a function that retrieves multiple collection entries from the same collection. This is useful for [returning an array of referenced entries](/en/guides/content-collections/#defining-collection-references) to access their associated `data`, `body`, and `render()` properties.
```
Fixed the link syntax. 

Also, "references" felt like a strange link text. Out of context, it doesn't mean very much, so it should be a slightly longer phrase, I think.

AND we don't really define/use "a reference, n." elsewhere -- we talk about the `reference()` utility. I'm seeing how "a referenced entry" feels as the noun that we use.

non-blocking nit: "retrieve/return for access to properties" feels a little more our existing style vs. "resolve to."

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -778,10 +826,79 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
...
+  <Since v="2.5.0" />
+</p>
+
+**Types:**
+
+- `(collection: string, contentSlugOrDataId: string) => CollectionEntry<collection>`
+- `({ collection: string, id: string }) => CollectionEntry<collection>`
+- `({ collection: string, slug: string }) => CollectionEntry<collection>`
+
+`getEntry()` is a function that retrieves a single collection entry by collection name and either the entry `id` (for JSON or YAML collections) or entry `slug` (for Markdown or MDX collections).
```
sarah11918, line 841: Quick question: reading the content-collections page, it made it seem more like `id` existed for both, but `slug` is only available for `type: content`. This makes it sound like you choose `id` for data, and slug for content.

Also, this says Markdown or MDX but doesn't mention Markdoc. So, we should maybe future-proof (and make things shorter!) by refering to the `type` type (right now, `data` and `content`) when we need to differentiate, vs file type?

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -778,10 +826,79 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
...
+---
+import { getEntry } from 'astro:content';
+
+// Get `src/content/blog/enterprise.md`
+const enterprisePost = await getEntry('blog', 'enterprise');
+
+// Get `src/content/captains/captain-kirk.yaml`
+const captainProfile = await getEntry('captains', 'captain-kirk');
+---
+```
```
sarah11918, line 853:
```suggestion
```
I'm not sure I'd include this example at all, but rather continue right into getting the `data` etc., then if necessary, show all things in one sample. I think it's cleaner to state what this function can all be used for, then show the examples in one sample snippet.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -778,10 +826,79 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
...
+import { getEntry } from 'astro:content';
+
+const enterprisePost = await getEntry('blog', 'enterprise');
+
+// Get the profile referenced by `data.captain`
+const captainProfile = await getEntry(enterprise.data.captain);
+---
+```
+
+See the `Content Collection` guide for examples of [querying](/en/guides/content-collections/#querying-collections) and [references](/en/guides/content-collections/#resolving-references).
```
sarah11918, line 868:
```suggestion
See the `Content Collections` guide for examples of [querying collection entries](/en/guides/content-collections/#querying-collections).
```
References is included in the querying section, and in fact, I have another note coming that I'm not entirely sure `resolving-references` needs to be its own section in the querying section anyway.  :smile:

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -797,7 +914,7 @@ const enterprise = await getEntryBySlug('blog', 'enterprise');
 
 ### Collection Entry Type
 
-The [`getCollection()`](#getcollection) and [`getEntryBySlug()`](#getentrybyslug) functions each return entries with the `CollectionEntry` type. This type is available as a utility from `astro:content`:
+Query functions including [`getCollection()`](#getcollection) and [`getEntry()`](#getentry) each return entries with the `CollectionEntry` type. This type is available as a utility from `astro:content`:
```
sarah11918, line 917: I know you've said "including.." so this doesn't need to be a comprehensive list, but I'll also note that `getEntries()` is not mentioned here.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -807,29 +924,40 @@ The `CollectionEntry<TCollectionName>` type is an object with the following valu
...
+A unique ID using the file path relative to `src/content/[collection]`. Enumerates all possible string values based on the collection entry file paths. Note that collections [defined as `type: 'content'`](#type) include the file extension in their ID, while collections defined as `type: 'data'` do not.
 
-#### `slug`
+#### `collection`
 
-**Example Type:** `'entry-1' | 'entry-2' | ...`
+**Example Type:** `'blog' | 'authors' | ...`
 
-A URL-ready slug. Defaults to the `id` without the file extension, but can be overridden by setting [the `slug` property](/en/guides/content-collections/#defining-custom-slugs) in a file's frontmatter.
+The name of the entry's collection.
```
sarah11918, line 936:
```suggestion
The name of a top-level folder under `src/content/` in which entries are located. This is the name used to reference the collection in your schema, and in querying functions.
```
I think we do need to connect this to the directory/folder name, as well as mention that this is what we consider to be the collection's "name" and that it's used in the schema, functions etc. I just blurted this suggestion above out quickly, but that's kind of the idea I'm thinking.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -807,29 +924,40 @@ The `CollectionEntry<TCollectionName>` type is an object with the following valu
...
 
 #### `data`
 
 **Type:** `CollectionSchema<TCollectionName>`
 
 An object of frontmatter properties inferred from your collection schema ([see `defineCollection()` reference](#definecollection)). Defaults to `any` if no schema is configured.
 
-#### `body`
+#### Content-specific properties
```
sarah11918, line 944: Not entirely sure what to do here, but it's kind of weird that `slug`, `body` and `render` now end up a heading size smaller than the other properties. I wonder if there's a way to identify these as only available for `type:content` without another heading. Is it too terribly messy/ugly to include for each one of these, maybe under **Type**, "available for every collection `type`" or "available for `type: content` collections only"  ??

sarah11918 (commented):
Woo, @bholmesdev !  :tada:

For Reasons., I'm splitting this review up. I have made some suggestions for "Querying Collections" (line 262) to the end of this PR (including the API reference page).

I have not looked super closely yet at anything earlier (intro, defining collections etc.) but that's coming!

In addition to comments below, some other thoughts that came up about the querying collections section in general:

1. line 264 "Astro provides TWO functions to query..." but now we have `getCollection()`, `getEntry()` and `getEntries()` I'm not getting a fundamentally different vibe from `getEntries`, so I want to make sure I'm not missing anything.
2. Related, I really don't feel like "Resolving References" is its own thing inside of querying data. Firstly, I don't think anyone's going to come to docs wondering, "Hmm... I wonder how I resolve my references..." so I don't think that's a heading I'd choose. I feel like it's still part of querying.. it just happens to be doing some querying of another collection from the collection "you're currently on". It also ALSO uses `getEntry()`, so I feel like there's a better flow/narrative path to just deal with them all together as querying collections. You really *are* just retrieving data from an entry... just from a different collection from the entry you're otherwise working with, is the mental model I'm formulating.
3. Also related and following on that, I think I have a suggestion below that removes an aside note re: slug only being available for `type: content`. I think right up top in this querying collections section (without a separate heading for references), we can be more explicit about using `getEntry` and what arguments you'll pass it based on the collection type.

So, I know that I've written this as 3 things, but it's all just about the very first section of Querying Collections and the approaches I'd tend to take there. The stuff below in the rest of this review is the more typical "about this line here" stuff.

More to come, but this is enough for you to consider right now!

src/content/docs/en/reference/api-reference.mdx
bholmesdev: I spent _so long_ trying this, and it was much more confusing with the "for type whatever" callouts everywhere. I agree the heading levels are annoying though this made it much clearer which were content-specific 😓 If you have a better refactor, push it up! This is the best I could come up with

src/content/docs/en/reference/api-reference.mdx
sarah11918: I'm trying to remember/hunt down what we did for deprecation in the past. I feel like we actually used the `<Badge /> component or something like that for emphasis. I think updating the note for now should be fine, though!

src/content/docs/en/reference/api-reference.mdx
sarah11918: OK, for now let's embrace the weirdness! No need to get this right on the first try. I have no issues with the content, so maybe the proper formatting will just reveal itself to us, eventually.

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -245,9 +267,52 @@ defineCollection({
 })
 ```
 
+### Defining collection references
+
+Collection entries can also "reference" other related entries. A common example is a blog post that references reusable author profiles stored as JSON, or related post URLs stored in the same collection.
+
+You can define these relationships using the `reference()` utility from your collection schema. This accepts a collection name and validates the entry identifier(s) in your frontmatter or data file.
```
sarah11918, line 274:
```suggestion
### Defining collection references

Collection entries can also "reference" other related entries. 

With the `reference()` function from the Collections API, you can define a property in a collection schema as an entry from another collection. For example, you can require that every `space-shuttle` entry includes a `pilot` property which uses the `pilot` collection's own schema for type checking, autocomplete and validation.

A common example is a blog post that references reusable author profiles stored as JSON, or related post URLs stored in the same collection:
```
This is how I'm thinking of it in my head, so I've spewed an example that is maybe a little more user-oriented re: "defining your property with another entry!"

Just throwing some alt wording at you to see whether you think a slight tweak in nuance is helpful here.

sarah11918 (commented):
Yay Ben! Left some more thoughts on the earlier part of this. Really liking it!

I'm tagging out for the night, and I've commented on the entire thing here, so some stuff for you to consider, and think about whether it's helpful!

src/content/docs/en/reference/api-reference.mdx
bholmesdev: Looks like we use `:::caution[Deprecated]` for `Astro.canonicalURL`. Think that's a good way forward!

src/content/docs/en/reference/api-reference.mdx
bholmesdev: Took a stab at refactoring to "Available for" callouts instead. Let me know if that's any better!

image

src/content/docs/en/reference/api-reference.mdx
bholmesdev: Correct, `id` _does_ exist for both. But in practice, we've found that users rarely use the _Content_ `id` since it isn't URL-friendly. This means it isn't used for `getStaticPaths()`, isn't used for SSR lookups, isn't used for... much of anything. The goal is to treat `slug` as the de-facto identifier now that we know how people use it, and come up with a much nicer solution come v3!

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -20,9 +20,9 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
...
+A **content collection** is any directory inside the reserved `src/content` project directory, such as `src/content/newsletter` and `src/content/authors`. Only content collections are allowed inside the `src/content` directory. This directory cannot be used for anything else.
 
-A **content entry** is any piece of content stored inside of your content collection directory. Content entries are stored as either Markdown (`.md`) or MDX (`.mdx`) files. You can use any filename you want, but we recommend using a consistent naming scheme (lower-case, dashes instead of spaces) to make it easier to find and organize your content.
+A **collection entry** is any piece of content stored inside of your content collection directory. Entries can use content authoring formats including Markdown (`.md`) and MDX (`.mdx`) or as data formats including YAML (`.yaml`) and JSON (`.json`). You can use any filename you want, but we recommend using a consistent naming scheme (lower-case, dashes instead of spaces) to make it easier to find and organize your content.
```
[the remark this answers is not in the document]
bholmesdev: Good point! I agree documenting the underscore pattern is useful, but this isn't _quite_ right. Underscore files will still be included in the build if you're using them for, say, relative image paths or MDX component imports. They're just excluded _from collection queries._ Wondering if this is the best place to explain that given it's the intro?

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -134,28 +135,31 @@ If you use `.js` or `.mjs` files in an Astro project, you can enable IntelliSens
...
+Schemas enforce consistent frontmatter or entry data within a collection. A schema **guarantees** that this data exists in a predictable form when you need to reference or query it. If any file violates its collection schema, Astro will provide a helpful error to let you know.
 
 Schemas also power Astro's automatic TypeScript typings for your content. When you define a schema for your collection, Astro will automatically generate and apply a TypeScript interface to it. The result is full TypeScript support when you query your collection, including property autocompletion and type-checking. 
 
-To create your first content schema, create a `src/content/config.ts` file if one does not already exist (`.js` and `.mjs` extensions are also supported). This file should:
+To define your first collection, create a `src/content/config.ts` file if one does not already exist (`.js` and `.mjs` extensions are also supported). This file should:
```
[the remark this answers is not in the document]
bholmesdev: Okay hang on, super confused why punctuation would go inside the parens?

src/content/docs/en/guides/content-collections.mdx
bholmesdev: LOVE this example :clap

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -20,9 +20,9 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
```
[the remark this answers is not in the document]
bholmesdev: Reading this section I can _immediately_ tell who wrote it 😅 Guess sell-y marketing speak was inevitable. What if we rewrite the first sentence to simply:

> **Content collections** are the best way to author content in any Astro project.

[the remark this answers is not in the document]
bholmesdev: Update: I also feel like the first and second sentence say _almost_ the same thing. What if we cut out the second sentence and do this instead:

> **Content collections** are the best way to manage and author content in any Astro project. Collections help to organize your documents, validate your frontmatter, and provide automatic TypeScript type-safety for all of your content.

bholmesdev:
@sarah11918 Re "Resolving References" section: I definitely divided with a heading to split up the examples, not for SEO. Though I _do_ expect people to wonder "ah crap, I just queried. Now how do I get the data for these references?" Resolving references was my stab at a short title to capture that need. Happy to either remove the title or find a better name if you prefer!

src/content/docs/en/guides/content-collections.mdx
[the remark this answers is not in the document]
sarah11918: Yup, I like that! I'll commit that change for you.

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -15,14 +15,13 @@ import TypeScriptSettingTabs from '~/components/tabs/TypeScriptSettingTabs.astro
   <Since v="2.0.0" />
 </p>
 
-**Content collections** are the best way to work with Markdown and MDX in any Astro project. Content collections are a feature of Astro that help manage your content files in a project. Collections help to organize your content, validate your frontmatter, and provide automatic TypeScript type-safety for all of your content.
-
+**Content collections** are the best way to author content in any Astro project. Content collections are a feature of Astro that help manage your content files in a project. Collections help to organize your content, validate your frontmatter, and provide automatic TypeScript type-safety for all of your content.
```
sarah11918, line 18:
```suggestion
**Content collections** are the best way to manage and author content in any Astro project. Collections help to organize your documents, validate your frontmatter, and provide automatic TypeScript type-safety for all of your content.
```

src/content/docs/en/guides/content-collections.mdx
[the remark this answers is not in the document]
sarah11918: Ok, so rather than get into more detail here, what about dropping that you can name them whatever you want entirely:

> A **collection entry** is any piece of content stored inside of your content collection directory. Entries can use content authoring formats including Markdown (`.md`) and MDX (`.mdx`) or as data formats including YAML (`.yaml`) and JSON (`.json`). We recommend using a consistent naming scheme (lower-case, dashes instead of spaces) for your files to make it easier to find and organize your content, but this is not required.

src/content/docs/en/reference/api-reference.mdx
sarah11918: I do personally like that better, because I like having all the properties in one place (rather than people needing to know where to go for certain types of properties before even being able to read about them). As the list of `type`s grows (?) we can always consider using "all types" for some properties, and "content only" for others.

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -245,9 +270,54 @@ defineCollection({
 })
 ```
 
+### Defining collection references
+
+Collection entries can also "reference" other related entries. 
+
+With the `reference()` function from the Collections API, you can define a property in a collection schema as an entry from another collection. For example, you can require that every `space-shuttle` entry includes a `pilot` property which uses the `pilot` collection's own schema for type checking, autocomplete and validation.
```
sarah11918, line 277:
```suggestion
With the `reference()` function from the Collections API, you can define a property in a collection schema as an entry from another collection. For example, you can require that every `space-shuttle` entry includes a `pilot` property which uses the `pilot` collection's own schema for type checking, autocomplete, and validation.
```
Adding the comma before Yan does!  :P

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -261,26 +331,64 @@ Your blog post content here.
...
+// Example: retrieve `src/content/authors/grace-hopper.json`
+const graceHopperProfile = await getEntry('authors', 'grace-hopper');
 ```
 
 Both functions return content entries as defined by the [`CollectionEntry`](/en/reference/api-reference/#collection-entry-type) type.
 
-#### Filtering collection queries
+### Resolving references
+
+When using [collection references](#defining-collection-references), you can use the `getEntry()` and `getEntries()` functions to retrieve the referenced entry `data`.
```
sarah11918, line 355:
```suggestion
### Using an entry's referenced data

Any [references defined in your schema](#defining-collection-references) must be queried separately after first querying your collection entry. You can use the `getEntry()` function again, or `getEntries()`, to retrieve the referenced entry from the returned `data` object. 
```
Ok, what do you think about something more along the lines of this? I feel like being more explicit about a "2-step query approach" might be helpful? Happy for you to adjust as necessary!

Possible headings? (Note, not just for SEO, but for someone scanning the page TOC and thinking of words that resonate: "Yes! That's what I'm trying to do!")
> Using an entry's referenced data
> Querying an entry's references
> Accessing referenced data

sarah11918 (commented):
I think this is it, Ben!  :tada:

Very tiny thoughts left for you to look at, and one or two nits. Thank you, this page is even more amazing now!

src/content/docs/en/guides/content-collections.mdx
bholmesdev: I like "Accessing referenced data" the most 🏆

sarah11918 (approved):
It is hereby my pleasure to approve this PR, @bholmesdev! 🥳

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -631,5 +740,5 @@ const { remarkPluginFrontmatter } = await blogPost.render();
 <p>{blogPost.data.title} — {remarkPluginFrontmatter.readingTime}</p>
 ```
 
-The remark and rehype pipelines only runs when your content is rendered, which explains why `remarkPluginFrontmatter` is only available after you call `render()` on your content entry.  In contrast, `getCollection()` and `getEntryBySlug()` cannot return these values directly because they do not render your content.
+The remark and rehype pipelines only runs when your content is rendered, which explains why `remarkPluginFrontmatter` is only available after you call `render()` on your content entry.  In contrast, `getCollection()` and `getEntry()` cannot return these values directly because they do not render your content.
```
Yan-Thomas, line 743: Nit: runs -> run so the verb agrees with the subject
```suggestion
The remark and rehype pipelines only run when your content is rendered, which explains why `remarkPluginFrontmatter` is only available after you call `render()` on your content entry. In contrast, `getCollection()` and `getEntry()` cannot return these values directly because they do not render your content.
```

Yan-Thomas (approved):
It is hereby my pleasure to find one more nit! But that's the last one, get my approval as well 🙌

PR #3244, opened 2023-05-13
Update .eslintrc.js

TKDev7:
This is a setting I have seen many projects use. However I am not sure what it does exactly lol. I just set it in case someone in the team does


sarah11918:
The best PRs are ones that we... have no idea what they do?  :sweat_smile: 

I'm going to defer to @delucis here. We very recently had some issues in this area and recently added a fix, so I would want to make sure this doesn't interfere with the intentional work Chris recently did!

PR #3254, opened 2023-05-15
Add Note to Install Locally

### LINKED ISSUE #3045: No mention to not install Astro globally
Installing Astro globally is not recommended or supported, and often leads to weird issues that can be hard to debug. 

I think it would be judicious to add a section somewhere perhaps in the installation or getting started area to avoid this situation. 

Oftentimes, people install Astro globally by accident because they see a command being "astro dev" instead of "npm run dev", run that, see an error and Google tell them to install the package globally

Jothsa:

- New or updated content

#### Description

- Closes #3045 
- What does this PR change? I added a note to the getting started page and manual install guide warning users to not install Astro globally.

### DIFF
--- a/src/content/docs/en/install/manual.mdx
+++ b/src/content/docs/en/install/manual.mdx
@@ -59,6 +59,10 @@ Once you are in your new directory, create your project `package.json` file. Thi
 
 First, install the Astro project dependencies inside your project.
 
+:::note[Important]
+Astro must be installed locally, not globally. Make sure you are *not* running `npm install -g astro` `pnpm add -g astro` or `yarn add global astro`.
+:::
+
 <PackageManagerTabs>
   <Fragment slot="npm">
   ```shell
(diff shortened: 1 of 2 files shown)


src/content/docs/en/install/manual.mdx
```diff
@@ -59,6 +59,10 @@ Once you are in your new directory, create your project `package.json` file. Thi
 
 First, install the Astro project dependencies inside your project.
 
+:::note[Important]
```
sarah11918, line 62: I do like the note here. I think this is where people are installing manually, and are most likely to make this "mistake"!

sarah11918 (commented):
This note looks good, @Jothsa! :raised_hands: Let's keep it to the manual installation page only, I think!

Jothsa:
Thanks, @sarah11918! I think I accepted your changes correctly

sarah11918:
Everything looks great to me! Once the tests pass, we're good to go!  :partying_face:

PR #3261, opened 2023-05-15
Update the entire hybrid rendering section

sarah11918:
Annoyingly, here's a PR to your branch, @matthewp :sweat_smile: 

See what you think about updating the entire hybrid rendering section to make the two flow better together.


matthewp:
Ah yes, this is a good call. I think moving away from the term "hybrid rendering", or just only using it in the case of `output: 'hybrid'` clears things up a lot.

PR #3264, opened 2023-05-16
Move Styles to Component from Global in Tutorial

### LINKED ISSUE #2929: ⚠️ Issue Why tutorial recommends the usage of global styling?
These two pages recommend to use global styles:
- https://docs.astro.build/en/tutorial/3-components/3/
- https://docs.astro.build/en/tutorial/3-components/4/

It contradicts a piece of advice given on one of the previous pages - use global styling only when it's necessary. In the linked cases, I don't think it really is necessary.

Jothsa:

- New or updated content

#### Description

- Closes #2929 
- What does this PR change? In the tutorial, I moved component styles from a global file to the specific component. I did not change the styles at all, I could remove the class names to illustrate Astro's scoping, but imo it's still best practice to use class names and the like.

### DIFF
--- a/src/content/docs/en/tutorial/3-components/3.mdx
+++ b/src/content/docs/en/tutorial/3-components/3.mdx
@@ -105,7 +105,7 @@ Since your site will be viewed on different devices, let's create a page navigat
     </div>
     ```
 
-2. Copy the CSS styles below into `global.css`. These styles:
+2. Copy the CSS styles below into `<style></style>` tags in `Navigation.astro`. These styles:
 
     - Style and position the navigation links for mobile
     - Include an `expanded` class that can be toggled to display or hide the links on mobile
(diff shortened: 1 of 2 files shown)


sarah11918:
Hi Jothsa! Thanks for going ahead and grabbing an issue, but reminder that just because someone files an issue doesn't mean that we've decided it needs changing in docs. (You can also tell because not every item has a "help wanted" tag, which means we haven't decided that we're ready for someone to fix it yet.)

I'm not saying that this one is a bad idea, just that it was not on our priority list to discuss yet. And, it's a particularly busy time right now, which is why we hadn't yet discussed in the issue how best to proceed with it.

I won't close this PR right now, instead I'll see how the next two weeks shake out and whether we can get to this. 

But since it's not something we were looking to do, and since any changes to the tutorial code require a significant effort to change the corresponding code project repo, in multple branches with proper rebasing, we are limited in when we have the resources to change the tutorial. When code is working, it's not as high on the priority chain.

I just wanted to let you know what's happening, and maybe help guide you towards issues that are easier for us to review, and fix more urgent problems. We love how involved you've been getting in the docs PRs and we feel bad when we can't keep up with you!  :sweat_smile: :raised_hands:

sarah11918:
Hey @Jothsa !

Just wanted to let you know that we've had some discussion around the tutorial code, and we've actually decided to do a larger audit of the Hamburger component and mobile nav so we can tackle some issues (like this one! based on feedback like yours!) all at once.

In the next few weeks @ElianCodes will be taking ownership of the issue described here, which will eventually then include documenting whatever changes we decide to make. If you'd like to get involved there, and then contribute to the resulting documentation, we'd be happy to have you! https://github.com/withastro/blog-tutorial-demo/issues/11

So, I'm going to close this PR for now, since while it seems like just a tiny text change, we can't change the code snippets without also diving in and changing all the branches of the tutorial demo code, too, and that's a bit more work. Given that we're going to treat this as a larger, more comprehensive task, we're moving the work over there first.

Thanks again for your patience and understanding, and hope to see you involved in more docs/tutorial PRs!

Jothsa:
No prob! I've been swamped with school lately, but hopefully I'll be able to contribute once my schedule stabilizes later this year. Sorry I just disappeared w/o saying anything 😅

PR #3268, opened 2023-05-16
change where required response is mentioned

sarah11918:
@ematipico I think it's now easier to say right when we define `onRequest()` that a response is required, and perhaps it fits better up there because it is required. 

What do you think about a change in where the information goes like this?


ematipico (approved):
Looks good to me!

PR #3270, opened 2023-05-17
Markdown Recipe

### LINKED ISSUE #2864: Pull out Markdown plugin examples into a recipe
[This section](https://docs.astro.build/en/guides/markdown-content/#modifying-frontmatter-programmatically) should be a recipe!

Jothsa:

- New or updated content

#### Description

- Closes #2864  
- What does this PR change? Give us a brief description. I moved the add reading time example from the Markdown & MDX guide to a new recipe. 

Let me know if I should do any other examples from the page.


sarah11918:
Hey this looks great @Jothsa! I'm really excited for this!

Just a couple of general changes I think we can make to improve the page, then I'll edit more closely:

- There aren't any prerequisites really, because Astro has remark support built in. You don't have to first install remark and THEN choose a remark plugin. So, I would say more like Prerequisites are none, saying that Astro has built-in support for adding remark plugins (with the same lin). 
- I know you've copied the existing text, but now that I see it on its own like this, do you think that it makes more sense to show adding to config BEFORE using it in a file? (That's the beauty of pulling these things out and making them feature content on their own recipe page! Rather than getting buried on a long page, we can really focus on them!) It seems to me the logical order of actions should be:
	- install
	- configure
	- use

Can you try recrafting the existing page based on the above, then I'll do a bit of editing, too?

Jothsa:
Thanks for the great feedback, @sarah11918! I'll get on it later this morning

Jothsa:
@sarah11918, it's my understanding that the plugin is created in a file created by the user. Personally, I think it makes more sense to create the file/plugin and then add it to the config. Otherwise, wouldn't you be saying, create a blank file, add it to your config, and then add the plugin code to the file? I'm fine either way, but just wanted to mention it.

sarah11918:
Oh, I get you now. Agreed! Good call! Let's leave that as is, and just make sure prerequisites makes sense.  :raised_hands:

Jothsa:
This needs edits as well 😅 Should have it done by Monday

src/content/docs/en/recipes/reading-time.mdx
```diff
@@ -0,0 +1,88 @@
...
+  <Fragment slot="yarn">
+  ```shell
+yarn add reading-time mdast-util-to-string
+  ```
+  </Fragment>
+</PackageManagerTabs>
+
+2. Create a remark plugin.
+
+This plugin uses the mdast-util-to-string package to get the markdown file's text. This text is then passed to the reading-time package to calculate the reading time in minutes.
```
sarah11918, line 40:
```suggestion
This plugin uses the `mdast-util-to-string` package to get the Markdown file's text. This text is then passed to the `reading-time` package to calculate the reading time in minutes. 
```

src/content/docs/en/recipes/reading-time.mdx
```diff
@@ -0,0 +1,88 @@
...
+    remarkPlugins: [remarkReadingTime],
+  },
+});
+```
+
+Now all Markdown documents will have a calculated `minutesRead` property in their frontmatter. 
+
+4. Display Reading Time
+
+Use the minutesRead property to display the reading time in a [markdown layout.](/en/guides/markdown-content/#markdown-and-mdx-pages) For example:
```
sarah11918, line 74:
```suggestion
Use the `minutesRead` frontmatter property from `Astro.props` to add the reading time to a [Markdown layout](/en/guides/markdown-content/#markdown-and-mdx-pages). The example layout below displays the reading time above the page content:
```

sarah11918 (commented):
This is looking amazing, @Jothsa! It feels like such a clean read!

I added a couple of suggestions, mostly for standardizing things: packages/properties etc should all use the code formatting when referring to them by name (instead of e.g. talking "about" the title of your page in a paragraph), and Markdown is always capitalized in prose.  :)

I also thought we could add a little more context around step 4. See what you think!

Jothsa:
Thanks!

src/content/docs/en/recipes/reading-time.mdx
```diff
@@ -0,0 +1,88 @@
+---
+title: Add Reading Time
+description: Build a remark plugin to add reading time to your Markdown or MDX files.
+i18nReady: false
+type: recipe
+---
+
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
+
+Create a [remark plugin](https://github.com/remarkjs/remark) which adds a reading time property to your Markdown or MDX files's frontmatter. Use this property to display the reading time for each page.
```
sarah11918, line 10:
```suggestion
Create a [remark plugin](https://github.com/remarkjs/remark) which adds a reading time property to the frontmatter of your Markdown or MDX files. Use this property to display the reading time for each page.
```

sarah11918:
Looking good! I just checked the Netlify deploy preview, and can you see that the content isn't indented inside all the steps like the other recipes are? e.g. if you check another docs page like https://deploy-preview-3270--astro-docs-2.netlify.app/en/recipes/build-forms-api/ you'll see the content is indented so that the numbered steps create a new left margin.

This should just be a matter of indenting content within each step, but check another recipe page to be sure. Once we get the formatting to match the others, I think this is good to go! :partying_face:

Jothsa:
Whoops. Thanks! Should be fixed now

sarah11918:
Let's do this!

PR #3292, opened 2023-05-18
Update integration docs

### LINKED ISSUE #3257: astro:build:done hook missing parameter
The docs for the `astro:build:done` hook do not describe the `pages` parameter: https://docs.astro.build/en/reference/integrations-reference/#astrobuilddone This seems to be the only place to access a list of built pages (where values for static route params have been realized into values), so it seems important.

TheOtterlord:

- New or updated content

#### Description

- Closes #3257
- What does this PR change?
  - Updates the integration docs to include the `pages` param. Seems this was removed previously as `pages` was due to be deprecated, but never was. `routes` also currently has some limitations, so `pages` is still useful for the time being.
  - This also removes a note about `routes` metadata being empty in SSR because that doesn't appear to be true ([example](https://stackblitz.com/edit/github-dpkcu2?file=integration.ts&on=stackblitz)).


TheOtterlord:
Removed the types section. `Check links` failed it and it's honestly not necessary. The type doesn't actually exist, let alone is anything exported to the user, and with one property it feels like a separate section just adds more scroll height more than anything.

dreyfus92 (approved):
LGTM! ✅

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -330,7 +330,7 @@ The address, family and port number supplied by the [NodeJS Net module](https://
 **Why:** To access generated routes and assets for extension (ex. copy content into the generated `/assets` directory). If you plan to transform generated assets, we recommend exploring the [Vite Plugin API](https://vitejs.dev/guide/api-plugin.html) and [configuring via `astro:config:setup`](#updateconfig-option) instead.
 
 ```js
-'astro:build:done'?: (options: { dir: URL; routes: RouteData[] }) => void | Promise<void>;
+'astro:build:done'?: (options: { dir: URL; routes: RouteData[], pages: PageData[] }) => void | Promise<void>;
```
tony-sull, line 333: `pages` here should probably be typed as `{ pathname: string }>[]` here to match Astro's types, but it'd be a great idea for us to add a `PageData` type into Astro's type definitions to make this more future proof!

TheOtterlord: Ah yep, I missed this when removing the types. I'll fix this in a sec!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -406,6 +406,12 @@ interface RouteData {
 }
 ```
 
+#### `pages` option
```
sarah11918, line 409:
```suggestion
#### `pages` option

**Type:**
```
If this is supposed to be analogous with `dir` option and `route` option, should it not also have the Type sub-heading?

TheOtterlord: Yep. Added it with the type from line 333

sarah11918:
Alright, then let's do this thing!  :partying_face:

Issue #3269, opened 2023-05-17
Potential update to rendering multiple elements 


ohansemmanuel:
## what? 
> ... you should wrap these elements inside a fragment as you would in JavaScript or JSX.

![CleanShot 2023-05-17 at 16 29 40](https://github.com/withastro/docs/assets/10930234/ec8328b2-1bd4-4759-9df2-71de4d319457)

## why? 
It's worth asking, why's this important? 

For example, not doing this in JSX throws an error. However, it works just fine in the latest versions of Astro. 

See https://stackblitz.com/edit/github-o8fjnb?file=src/pages/index.astro 

If we're telling people they **should** do this, it's worth mentioning "why" as this seems to work just fine i.e., without error as in the JSX comparison. 

[See docs.](https://docs.astro.build/en/core-concepts/astro-syntax/#fragments--multiple-elements)

sarah11918:
Hi @ohansemmanuel - I can tell you that when these docs were written, wrapping the elements was required. It would not work otherwise.

I'm checking in with the dev team to see whether something has changed and we can update this advice!  :raised_hands:

dreyfus92:
@sarah11918 I just verified that there's no need to have a fragment to perform this action. I can handle the update if you want me to.

ohansemmanuel:
Thanks @sarah11918

PR #3296, opened 2023-05-19
Update: `astro-syntax.mdx`

dreyfus92:

- New or updated content

#### Description

- Closes https://github.com/withastro/docs/issues/3269
Removed the example where you should wrap elements inside in a fragment in order to avoid errors with the old compiler, there's no need to use fragments to use JSX expressions.

### DIFF
--- a/src/content/docs/en/core-concepts/astro-syntax.mdx
+++ b/src/content/docs/en/core-concepts/astro-syntax.mdx
@@ -122,49 +122,11 @@ When using dynamic tags:
 
 - **Hydration directives are not supported.** When using [`client:*` hydration directives](/en/core-concepts/framework-components/#hydrating-interactive-components), Astro needs to know which components to bundle for production, and the dynamic tag pattern prevents this from working.
 
-### Fragments & Multiple Elements
+### Fragments
 
-An Astro component template can render multiple elements with no need to wrap everything in a single `<div>` or `<>`, unlike JavaScript or JSX.
+Astro supports using either `<Fragment> </Fragment>` or the shorthand `<> </>`.
 
-```astro title="src/components/RootElements.astro"
----
-// Template with multiple elements
----
-<p>No need to wrap elements in a single containing element.</p>
-<p>Astro supports multiple root elements in a template.</p>
-```
-
-However, you can use fragments if you wish to do it. Astro supports using either `<Fragment> </Fragment>` or the shorthand `<> </>`.
-
-Here you have some examples where fragments can be useful:
-
-Fragments can be handy when fetching HTML from a CMS(e.g. Ghost or WordPress).
-
-```astro title="/src/pages/index.astro"
----
-import Layout from "../layouts/Layout.astro"
-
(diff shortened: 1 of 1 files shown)


ohansemmanuel:
Thanks for getting on this @dreyfus92 🔥 

If we wanna do this right, I reckon we also have to update the "[Astro syntax vs JSX](https://docs.astro.build/en/core-concepts/astro-syntax/#differences-between-astro-and-jsx)" section to highlight we don't need this. Happy to contribute as well, but it seems rather straightforward. 

CleanShot 2023-05-19 at 07 18 54


We'd likely need @sarah11918 's input on the new addition as well. 

Cheers!

src/content/docs/en/core-concepts/astro-syntax.mdx
```diff
@@ -134,22 +134,7 @@ An Astro component template can render multiple elements with no need to wrap ev
...
-  {items.map((item) => (
-    <>
-      <li>Red {item}</li>
-      <li>Blue {item}</li>
-      <li>Green {item}</li>
-    </>
-  ))}
-</ul>
-```
+Astro supports using either `<Fragment> </Fragment>` or the shorthand `<> </>`.
```
sarah11918, line 137:
```suggestion
However, you can use fragments to ... Astro supports using either `<Fragment> </Fragment>` or the shorthand `<> </>`.
```
This sentence on its own feels a little like, "Wait, you just said I don't need fragments?!"  :smile:

I think one use of Fragment we show elsewhere is avoiding extra wrapping HTML elements, especially when using `set:html`. What about adding something here so that this section says both that we DO support them, and when you might use them?

sarah11918 (commented):
Thanks for running with this @dreyfus92!  :runner: 

See my comment below re: making that sentence feel like it has a bit more of a purpose.

And thanks for the helpful feedback, @ohansemmanuel . My gut reaction is that since we literally have fragments and multiple elements in the sidebar already, and it's immediately above "differences between JSX and Astro", I don't think we need to mention this twice. 

I see two potential options, and I'm open to exploring both:

1. Just stick with this change for now that corrects a currently incorrect section in docs, and we can always revisit if we think it needs improvement later.
2. Consider moving the entire "Fragements and multiple elements" *into* the differences section.

What do y'all think?

dreyfus92:
I will try to make some changes based on the feedback. Thanks @ohansemmanuel & @sarah11918.

sarah11918:
Alright, I think I'm happy with this one! Sorry, @dreyfus92 , I sent you on a bit of a wild path when I said "shouldn't we say when you'd use it?" without noticing that DIRECTLY BELOW we did give an example of that!  :sweat_smile: 

So, I just incorporated that into the statment, and I kept "Fragments" in the Astro part, but moved "multiple elements" down into differences between JSX and Astro.

I'm happy with this, so I'm merging it and getting it off our plates as it addresses the initial issue!  :partying_face: If someone finds something horribly wrong with this, then you are welcome to make a new PR.

Thanks everyone for the work on this one!  :raised_hands:

PR #3297, opened 2023-05-19
fix: removes wrong inline style info

### LINKED ISSUE #3165: Potentially misleading documented behaviour of define:vars
Hi there, 

The following section of the doc seems rather misleading: 

https://docs.astro.build/en/reference/directives-reference/#definevars

(1) above describes `define:vars` with respect to both `<style>` and `<script>` 
(2) describes only the behaviour with `<script>` 

The problem here is with (1) 

**Correct me if I'm wrong** but using `define:vars` in a `<style>` within a component markup **still bundles the component styles i.e., contrary to what's documented here.** 

Consider an example with a single component rendered three times and `define:vars` used on both `<style>` and `<script>` elements. 

CleanShot 2023-05-04 at 11 54 29

- 1, 2,3 in the screenshot above: the script is inlined as expected. Everything else holds true for the script. 

However, this isn't the case for the styles. 

- the variable is locally declared on the parent element in the markup
- the style isn't added multiple times with multiple component use (as with a script inline directive) 
- the component style is still bundled and locally scoped (not global) 

Does the doc here need to be adjusted to exclude styles? i.e., (1)

ohansemmanuel:
- New or updated content

#### Description

- Closes #3165

> Using define:vars in a <style> within a component markup still bundles the component styles, i.e., contrary to what's documented. Please refer to the #3165 for more details.

### DIFF
--- a/src/content/docs/en/reference/directives-reference.mdx
+++ b/src/content/docs/en/reference/directives-reference.mdx
@@ -252,7 +252,7 @@ const message = "Astro is awesome!";
 ```
 
 :::caution
-Using `define:vars` on a `<script>` or `<style>` tag implies the [`is:inline` directive](#isinline), which means your scripts or styles won't be bundled and will be inlined directly into the HTML. 
+Using `define:vars` on a `<script>` tag implies the [`is:inline` directive](#isinline), which means your scripts won't be bundled and will be inlined directly into the HTML. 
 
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Thanks for this @ohansemmanuel , and for confirming in the issue @Jothsa ! 

It's always helpful when an issue is discussed and verified before a PR comes in. Great team work!  :raised_hands:

PR #3306, opened 2023-05-21
Add more clarifications for custom slugs that has multiple level for content collections

margani:

- New or updated content

#### Description

- Makes docs more clear about defining custom slugs for content collections
- I added a paragraph to explains that the custom slugs that contains `/` character must be used within the pages that have rest parameters.


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -321,6 +321,8 @@ When using `type: 'content'`, every content entry generates a URL-friendly `slug
 
 You can override an entry's generated slug by adding your own `slug` property to the file frontmatter. This is similar to the "permalink" feature of other web frameworks. `"slug"` is a special, reserved property name that is not allowed in your custom collection `schema` and will not appear in your entry's `data` property. 
 
+If the custom slug for your content contains `/` character, meaning it has multiple levels, you must use [rest parameters](/en/core-concepts/routing/#rest-parameters) for the dynamic route that generates it. Refer to the example code [here](/en/guides/content-collections/#building-for-static-output-default).
```
sarah11918, line 324:
```suggestion
:::note
If your custom slugs contain the `/` character to produce URLs with multiple path segments, you must use a [rest parameter (`[...path]`)](/en/core-concepts/routing/#rest-parameters) in the `.astro` filename for this dynamic routing page.
:::
```
Hi @margani. Thanks for this contribution!

I edited it a little, and I actually think it makes more sense to show this down around line 501-ish where you link to the code sample, directly after it. if we're saying you need to be looking "here", then that's usually a sign that the text should also be "there"!  :smile:

Can you update this to turn this into a note like above, and put it around line 502? (I think line 501 is the empty line just after the code example). That also means we don't need a second link, and we can refer directly to the code sample that the reader would have just seen!  :raised_hands:

sarah11918 (commented):
Thanks for adding this, @margani!  :partying_face:  I left some instructions below!

src/content/docs/en/guides/content-collections.mdx
margani: Thank you @sarah11918 , I updated it.

sarah11918:
Thanks, @margani, looks great! Unfortunately we're having a bit of our trouble with our Netlify builds right now, so we're not merging PRs until that's resolved. But don't worry, we'll get you into Team Docs as soon as we can!  :partying_face:

margani:
> Thanks, @margani, looks great! Unfortunately we're having a bit of our trouble with our Netlify builds right now, so we're not merging PRs until that's resolved. But don't worry, we'll get you into Team Docs as soon as we can!  :partying_face: 

Awesome,  thank you for helping me out.

Yan-Thomas (approved):
LGTM! Feel free to merge when you see fit @sarah11918 🙌

PR #3315, opened 2023-05-22
Fix Create Page link

lorenzolewis:

#### Description

Currently when you click the "Create page +" button on https://i18n.docs.astro.build it will build a URL for GitHub that includes the lang part twice. An example of this can be seen for the page `concepts/islands.mdx` for Arabic:

Screenshot 2023-05-22 at 17 07 08

This should hopefully fix that so the extra part in the slug isn't there (and instead is relying on the parameters).


delucis (approved):
Oh nice! For ages GitHub had a bug where it ignored the first segment of the filename requiring us to include it twice. Looks like they finally fixed that.

Thanks for catching that :raised_hands:

PR #3316, opened 2023-05-22
Hotfix for failing Fontsource API

delucis:

#### Description

- We download font files from the [Fontsource API](https://api.fontsource.org/) as part of Open Graph image generation. The API is currently down, causing our builds to fail. This PR is a temporary fix to unblock stuff quickly.
- This PR adds the fonts to the Astro repo and updates the URL pointing at Fontsource to download from GitHub instead.
- It adds the fonts in a dedicated branch — https://github.com/withastro/docs/tree/temp-fonts-cdn — to avoid adding them to `main` and general git history.
- The long-term fix would be to use these files in the repo but ALSO avoid fetching them and just load them from the file system. That requires changes to how [`astro-og-canvas`](https://github.com/delucis/astro-og-canvas) works though, so I opted for this quicker fix for now.


sarah11918:
Thanks so much for this `/raw/chris/`  :smile:

sarah11918 (approved):

Yan-Thomas (approved):
LGTM!

delucis:
> Thanks so much for this `/raw/chris/` 😄

🤣

Took the dog out and decided I didn’t want these font files ending up in `main` and in its git history, so just switched this to use a separate `temp-fonts-cdn` branch for the files with this PR only updating the URLs in the OpenGraph endpoint.

PR #3324, opened 2023-05-23
Fix typos in guides/cms/contentful.mdx

morinokami:

#### Description

- In some parts of the code in contentful.mdx, the`BlogPost` type was written as `blogPost`, so these are corrected.


Yan-Thomas (approved):
Nice catch, thanks for this @morinokami!

PR #3326, opened 2023-05-23
Moving off of custom domains for Fathom analytics

tony-sull:

- Something else!

#### Description

Updates how the Fathom Analytics script is loaded.

**Why?** Fathom recently ran into an issue with some events being lost when the script is loaded via a custom subdomain.  The bug is specifically related to SSL certificates sometimes not auto-renewing on time, ours hasn't been impacted yet so we shouldn't have lost any analytics data to date.


delucis:
Thanks for this! Sorry not to review quickly but here’s a belated LGTM 🚀

PR #3168, opened 2023-05-04
Display note on outdated translations

TheOtterlord:
#### What kind of changes does this PR include?

- New or updated content
- Changes to the docs site code

#### Description

- Closes #1481 
- What does this PR change? Adds an aside to the top of pages with outdated translations.
- Did you change something visual? A before/after screenshot can be helpful.

<details>
<summary>Preview</summary>

</details>

PR #3329, opened 2023-05-24
Remove outdated translation note

delucis:

#### Description

- We noticed #3168 increased build times on Netlify by about 50%, most likely from my investigations because it introduces many `git log` calls.
- This PR removes the outdated content notice in case we decide we want to speed builds back up. I tried various optimisation paths but without much success.


delucis:
I think we’re going to make the call to merge this and remove the outdated content notice. It’s super nice to have but has been slowing down maintainers when merging. I think we can maybe revisit an alternative approach to doing this — maybe there’s a cheaper way to calculate staleness, or we’re able to do it in some async secondary action that doesn’t slow down builds. Not sure yet, but it’s definitely a cool thing to have if we can figure it out!

I’ll leave this to @Yan-Thomas or @sarah11918 to merge when it’s a good time for them as they’re probably the people this impacts most!

Yan-Thomas (approved):
Thanks for finding this out @delucis! Merging now 🚀

PR #3331, opened 2023-05-25
update: remove experimental wording about middleware

ematipico:

- New or updated content

#### Description

This PR updates the middleware in preparation for the 2.6 release.

This PR must be merged after https://github.com/withastro/astro/pull/7109 is merged.

I am not really sure if the information I added is under the correct category; I would greatly appreciate some guidance! 

Here's the list of the **additional** changes:
- heading about type safety, which also talks about the utility `defineMiddleware`, which wasn't mentioned before;
- updated the `api-reference.mdx` to add information about `context.locals` and `Astro.locals`.


sarah11918 (commented):
Thanks for this @ematipico !  Here are some early thoughts. 

I think think we should also look at the link on this page the RFC that I had used somewhere, I think, so we didn't have to add more here while this was experimental. I *think* that still exists here? If so, we'll have make sure that all the good stuff we want to document ends up directly on the middleware page!

src/content/docs/en/guides/middleware.mdx
```diff
@@ -11,60 +11,61 @@ This also allows you to set and share request-specific information across endpoi
...
-const data = Astro.locals;
----
-<h1>{data.title}</h1>
-<p>This {data.property} is from middleware.</p>
+```js tilte="src/middleware.js"
+
+/**
+ * @type import("astro").MiddlewareResponseHandler
+ */
+export const onRequest = (context, next) => { // `context` and `next` are automatically typed
```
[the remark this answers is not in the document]
ematipico: There **shouldn't** be problems

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -426,6 +426,31 @@ And would render HTML like this:
 </ul>
 ```
 
+
+### `Astro.locals`
+
+`Astro.locals` is an API available via [Astro middleware](/en/guides/middleware/). It's an object that can contain any arbitrary information.
```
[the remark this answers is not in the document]
ematipico: Sure!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -570,6 +595,24 @@ export function get({ redirect }: APIContext) {
...
+```ts title="src/middleware.ts"
+import type { MiddlewareResponseHandler } from 'astro';
+
+export const onRequest: MiddlewareResponseHandler = ({ locals }, next) => {
+    locals.foo = "bar"
+    return next();
+}
+```
+
+See also: [`Astro.redirect()`](#astroredirect)
```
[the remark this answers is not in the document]
ematipico: Oh yeah, thank you for catching that!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -426,6 +426,25 @@ And would render HTML like this:
...
+
+`Astro.locals` is an API available via [Astro middleware](/en/guides/middleware/). It's an object that can contain any arbitrary information.
+
+```astro title="src/pages/Orders.astro"
+---
+const title = Astro.locals.welcomeTitle();
+const orders = Array.from(Astro.locals.orders.entries());
+---
+<h1>{title}</h1>
+<p>This {data.property} is from middleware.</p>
```
sarah11918, line 440:
```suggestion
```
I'm going to suggest we remove this, since we don't show `data` anywhere. I think the two matching examples are enough!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -426,6 +426,25 @@ And would render HTML like this:
 </ul>
 ```
 
+
+### `Astro.locals`
+
+`Astro.locals` is an API available via [Astro middleware](/en/guides/middleware/). It's an object that can contain any arbitrary information.
```
sarah11918, line 432:
```suggestion
`Astro.locals` is an object containing any values from the [`context.locals`](#contextlocals) object from a middleware. Use this to access data returned by middleware in your `.astro` files.
```
OK, PLOT TWIST!  :sweat_smile: 

So I looked at the definition for `Astro.props` (because, this is like props for middleware). What do you think about something like this instead?

sarah11918 (commented):
Sorry, had another crazy idea. What do you think?

src/content/docs/en/reference/api-reference.mdx
ematipico: I missed that, good catch!

src/content/docs/en/reference/api-reference.mdx
ematipico: That works very well! Thank you!

sarah11918 (approved):

wassfila:
as discussed "Middleware (experimental)" still in the heading, in the left menu.

PR #3352, opened 2023-05-26
Configurable paths expect a string, not an object

jose8a:
The current example for configurable paths returns an error.The value for each of these configurable paths that are used by Astro - they expect a string, not an object. Wrapping the currently used example values within string literal notation, such as `new URL(...)` fixes this.

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 


sarah11918:
Thanks for this @jose8a ! We've confirmed this works, but we're discussing whether there's a different option we'd suggest here. We might even be able to do something so that Astro can accept a URL here instead of a string, so just sit tight while we discuss this one!  Be back to this soon! :raised_hands:

jose8a:
No worries. Do what y'all need to do. I am using Astro as a dependency in a project where Astro's files are not at the top-level, so I had to modify the paths for those folders, and that's how I came upon this issue and its fix.

src/content/docs/en/guides/configuring-astro.mdx
```diff
@@ -108,9 +108,9 @@ import { defineConfig } from 'astro/config'
 
 export default defineConfig({
   // Resolves to the "./foo" directory, relative to this config file
-  root: new URL("./foo", import.meta.url),
+  root: `${new URL("./foo", import.meta.url)}`,
```
delucis, line 111: I think it might be nicer here to explicitly call `toString()` instead of using the template string to do that implicitly:

```suggestion
  root: new URL("./foo", import.meta.url).toString(),
```

src/content/docs/en/guides/configuring-astro.mdx
```diff
@@ -108,9 +108,9 @@ import { defineConfig } from 'astro/config'
...
   // Resolves to the "./public" directory, relative to this config file
-  publicDir: new URL("./public", import.meta.url),
+  publicDir: `${new URL("./public", import.meta.url)}`,
```
delucis, line 113:
```suggestion
  publicDir: new URL("./public", import.meta.url).toString(),
```

delucis (commented):
Thanks for this fix @jose8a and for your patience as we got to this! Suggesting a small syntax change and then this should be good to go.

sarah11918 (commented):
Looks great! Thanks for this @jose8a and @delucis! I do like the `toString` as a solution that clearly indicates a string is required. Let's add this to the FriYAY merge day celebrations! :partying_face:

sarah11918 (approved):

Yan-Thomas (approved):
LGTM!

PR #3353, opened 2023-05-26
Unflag hybrid rendering docs

matthewp:

#### Description

- This is for the release of hybrid rendering as unflagged, tentatively in 2.6.
- So far the only change is removing the experimental flag. I'm not sure if it's better to try and combine the "hybrid rendering" section above. Happy to take a pass at that if people think it would be better to explain at one time.

### DIFF
--- a/src/content/docs/en/guides/server-side-rendering.mdx
+++ b/src/content/docs/en/guides/server-side-rendering.mdx
@@ -151,9 +151,6 @@ export { defineConfig } from 'astro/config';
 
 export default defineConfig({
   output: 'hybrid',
-  experimental: {
-    hybridOutput: true,
-  },
 });
 ```
 


sarah11918:
Sure! Go for it!

I see maybe a possible logical flow as:
- hybrid rendering means SOME pages pre-rendered, and some not
- New in Astro 2.6, there are two modes to determine what the default is, and therefore which pages need to be marked to do the non-default thing
- For mostly static sites with some SSR  (non-prerendered) pages, choose hybrid (new in Astro 2.6)
- For mostly SSR pages, with the option to pre-render only some select pages, choose server (the only option available prior to Astro 2.6)

sarah11918:
(And, if you don't get around to it, just uncommenting the experimental flag would be fine, too!)  :smile:

matthewp:
Ok, I think it's worth trying to combine too. I'll make a pass and if you don't like it the commit can just be removed.

matthewp:
@sarah11918 Took a pass at it, let me know what you think.

sarah11918:
Hi @matthewp - I made #3378 as a PR to your branch since I found I wanted to edit some things outside of your changes here. For example we still had the bold line about **every page being server-rendered** once you enable SSR, which isn't true anymore. 

I took the take that *both* modes are a way to get some (or all) SSR into your project, and defined/described them as such. There's a lot that's parallel between the two so thinking of them, and structuring the page that way is quite clean. We might decide we don't like that, but I think there's ultimately some value in that clarity, so I wanted to consider it. (Framing it that way vs hybrid rendering being it's own weird thing *does* make me think that the name `hybrid` feels a little off, so that's one reason not to do it. *Both* `server` and `hybrid` allow for a combination, or hybrid, of rendering... :sweat_smile: )

See what you think  in the other PR, and like I wrote there, it doesn't have to be the final call. But if we like that structure/framing, then we can take this from there.

sarah11918:
Also, just a reminder for us that there should be a corresponding `withastro/astro/` PR that removes this option entirely from `configuration-reference` like the PR for custom client directives here: https://github.com/withastro/astro/pull/7237  I think that's all that needs to happen in the other repo for docs.

matthewp:
Happy with how this landed!

sarah11918 (approved):
Happy, too!  :partying_face:

PR #3354, opened 2023-05-26
content collections `type` only v2.5.0+

sarah11918:
Someone was getting an error setting up their content collections with `type: content` because they hadn't yet upgraded to 2.5.0.

This quick change explicitly says that the `type` property was added in 2.5.0. We shouldn't need this for very long, and we can figure out a better way to handle changes like this. But there will still be some people on other pre 2.5 versions of Astro who may run into this issue if they just try setting up content collections from scratch with our existing docs, not realizing this isn't just how it always was.


LushawnDev:
Yep this is perfect and very clear, thank you @sarah11918 !

delucis (approved):
LGTM! We do have the `<Since v="2.5.0" />` component we use elsewhere, but I don’t see any obvious place we could use that format here, so this works!

PR #3374, opened 2023-05-29
Elaborate more on the route priority order

MoustaphaDev:

- New or updated content

#### Description

The current route priority order section lacks some info. For example, endpoints always take precedence in their route category, and prerendered routes always take precedence over server routes. This PR adds more about all that.


src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -233,18 +233,36 @@ It's possible for multiple routes to match the same URL path. For example each o
...
+- Server endpoints using named parameters take precedence over server routes using named parameters
+- Server routes using named parameters take precedence over other routes using rest parameters
+
+3- Dynamic routes with rest parameters
 - Rest parameters have the lowest priority
-- Ties are resolved alphabetically
+- Static endpoints using rest parameters take precedence over other static routes using rest parameters
+- Static routes using rest parameters take precedence over server endpoints using rest parameters
+- Server endpoints using rest parameters take precedence over server routes using rest parameters
+- Server routes using rest parameters take precedence over other routes using rest parameters
```
MoustaphaDev, line 251: I wonder if it would be a good idea to create a [priority queue](https://www.google.com/search?q=Priority+Queue&tbm=isch) or a pyramid component to represent these

sarah11918:
Sorry, I'm still not at the stage of really editing this yet, but clarifying question:

This is now a lot more words to read, and I think really all it's saying new is that in every case, endpoints take priority over routes? Maybe we just need a sentence that says that somewhere, and it applies everywhere?

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -239,6 +239,8 @@ Astro needs to know which route should be used to build the page. To do so, it s
 - Rest parameters have the lowest priority
 - Ties are resolved alphabetically
 
+sMoreover, endpoints always take precedence over pages. Also, if you have dynamic routes (using rest parameters or named parameters) that are prerendered, they will take precedence too over dynamic routes that aren't.
+
```
sarah11918, line 243:
```suggestion
- Pre-rendered dynamic routes take precedence over static dynamic routes
- Rest parameters have the lowest priority
- Endpoints always take precedence over pages
- Ties are resolved alphabetically
```
What do you think about just adding to our list, like this? (view it with the points in the lines just above)

MoustaphaDev: That works for me, thanks for suggesting that!
There's just a minor change to do, in the first bullet point in “over static dynamic routes” it should be “**server**” instead of “**static**”.

sarah11918: Perfect, thanks!

sarah11918 (approved):
Looks great to me! Thanks for this!  :rocket:

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -236,9 +236,10 @@ Astro needs to know which route should be used to build the page. To do so, it s
 
 - Static routes without path parameters will take precedence over all other routes
 - Dynamic routes using named parameters take precedence over rest parameters
+- Pre-rendered dynamic routes take precedence over server dynamic routes
 - Rest parameters have the lowest priority
+- Endpoints always take precedence over pages
 - Ties are resolved alphabetically
```
MoustaphaDev, line 242: Just bringing back the line break
```suggestion
- Ties are resolved alphabetically

```

PR #3375, opened 2023-05-29
New recipe: Minify HTML responses

### LINKED ISSUE #3327: Missing docs: How to use `nanostores` with Astro components
Our [Share State Between Islands](https://docs.astro.build/en/core-concepts/sharing-state) guide explains how to use `nanostores` with a variety of UI frameworks. However, not everyone wants to include a UI framework, and might prefer to do it in pure JavaScript. Since a `script` in an Astro file isn't an island it makes sense that it's not documented here, but I feel it does deserve a place somewhere. The way I see it, there are a couple of ways to solve this:

1. "Rebrand" the guide as "Share State Between Components"
2. Create a new guide elsewhere specifically for Astro (probably a recipe?) and link to it using the related recipes component.

This present most recently in #3181 and in the Discord [`#feedback-ideas`](https://discord.com/channels/830184174198718474/872579324446928896/1110884462281097256) channel.

Happy to make a PR once we decide what we want.

TheOtterlord:

- New or updated content

#### Description

- What does this PR change? Adds a new recipe for using middleware to minify HTML responses.


ElianCodes (approved):
I love this!! Thank you @TheOtterlord ❤️🚀

src/content/docs/en/recipes/minification.mdx
```diff
@@ -0,0 +1,65 @@
...
+---
+
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
+
+Astro supports built-in [HTML minification](/en/reference/configuration-reference/#compresshtml) which works for `.astro` components, but not for [islands](/en/concepts/islands/). This recipe shows how to use Astro middleware to minify complete HTML responses.
+
+## Prerequisites
+
+- An Astro project with [middleware enabled](/en/guides/middleware/#enabling-middleware-in-your-project)
+
```
TheOtterlord, line 15: Middleware is no-longer experimental.
```suggestion
```

Yan-Thomas (commented):
Great work @TheOtterlord! After Sarah has reviewed it as well, LGTM!

src/content/docs/en/recipes/minification.mdx
```diff
@@ -0,0 +1,61 @@
...
+    ```
+    </Fragment>
+    <Fragment slot="yarn">
+    ```shell
+    yarn add html-minifier
+    ```
+    </Fragment>
+  </PackageManagerTabs>
+
+2. Create a middleware file `src/middleware.ts`:
```
sarah11918, line 34:
```suggestion
2. Create a middleware file `src/middleware.ts`. Add the following code to intercept an HTML response and returned a minified version:
```
I think a *bit* of explanation here, just at this level, is helpful! Nothing more than would comfortably fit in a sentence. Adjust so that it's accurate!

src/content/docs/en/recipes/minification.mdx
```diff
@@ -0,0 +1,61 @@
+---
+title: Minify HTML responses with middleware
+description: Use Astro middleware to minify HTML responses
+i18nReady: false
+type: recipe
+---
+
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
+
+Astro supports built-in [HTML minification](/en/reference/configuration-reference/#compresshtml) which works for `.astro` components, but not for [islands](/en/concepts/islands/). This recipe shows how to use Astro middleware to minify complete HTML responses.
```
sarah11918, line 10: I think this could be a *little* more connect the dots.

The lede is that Astro supports built-in minification... but then we're installing something else to minify? Can you make the connection stronger here between "not working for islands" and "a complete HTML response?"

I'm assuming that the idea is something like Astro will minify on a *per-component* basis, but only for `.astro` components. Since your islands are NOT getting minified by Astro's own compression, your HTML could get more mini! This external package can minify a complete page's HTML Response, and is how you can get HTML minification site-wide, at the page level, which would therefore include your islands.

I think the link missing is from "islands not minified" to "use middleware to minimize HTML Responses". It might not be *immediately* clear that this is the way to include islands. And, I think starting with the statement about what *sounds* like the thing they are going to do in this recipe (minify things!) being built in doesn't set the immediate expectation of installing another package.

src/content/docs/en/recipes/minification.mdx
```diff
@@ -0,0 +1,61 @@
...
+
```
sarah11918, line 11:
```suggestion

## Prerequisites

Any existing Astro project.
```
I still like having a prerequisites section, and being explicit that there are no constraints on how your project works! SSG/SSR/content collections etc. Or, if there *is* anything required, stick it there. Middleware isn't experimental anymore, so that's not a requirement.

src/content/docs/en/recipes/minification.mdx
```diff
@@ -0,0 +1,61 @@
...
+    const html = await response.text();
+    const minifiedHtml = minify(html)
+    
+    return new Response(minifiedHtml, {
+      status: 200,
+      headers: response.headers
+    });
+  }
+  ```
+
```
sarah11918, line 57:
```suggestion

4. Check / Inspect using dev tools...
```
I like some kind of guidance here, even if it's not to compare from before, but give them something they can look at. Can they see the minified HTML somewhere? Can they see a smaller bundle? Give them some kind of conclusion here!

src/content/docs/en/recipes/minification.mdx
```diff
@@ -0,0 +1,61 @@
+---
+title: Minify HTML responses with middleware
+description: Use Astro middleware to minify HTML responses
+i18nReady: false
```
sarah11918, line 4:
```suggestion
i18nReady: true
```
I believe in you!

sarah11918 (commented):
I don't know when I'll get around to all of these @TheOtterlord , but I tried to sneak at least one in this week!  :raised_hands:

TheOtterlord:
Thanks for the review!

I've addressed a few of your comments, but looking at this again after a few months, I'm not sure if this is needed. `compressHTML` does seem to compress framework components that aren't hydrated, since they're part of the `.astro` page, and therefore the output of a `.astro` component. Also, islands js bundles look to already be minified in the bundle. There are still uses for this type of minification, given how customisable packages like `html-minifier` are, but that's not really suitable for a recipe. I should have revisited this PR sooner :sweat_smile:

sarah11918:
To be fair *I* should have revisited this PR sooner... :sweat_smile: 

We'll have our chance again; there are so many more for me to get to!  :raised_hands:

PR #3378, opened 2023-05-29
restructure SSR options on page

sarah11918:
Hey @matthewp, I took a shot at a larger reorg, since there was some stuff not within the areas you PR'd that I felt could benefit from being a part of "our new outlook" here.

Not saying this is necessarily "it" yet, but what do you think about something like this?

### DIFF
--- a/src/content/docs/en/guides/server-side-rendering.mdx
+++ b/src/content/docs/en/guides/server-side-rendering.mdx
@@ -13,7 +13,10 @@ import Since from '~/components/Since.astro';
 
 ## Enabling SSR in Your Project
 
-To get started, enable SSR features in development mode with the `output: server` configuration option:
+To enable SSR features in development mode, add one of the two `output` configuration options: `server` or (introduced in v2.6.0) `hybrid`. Both of these modes allow all or just some of your [pages](/en/core-concepts/astro-pages/) or [server endpoints](/en/core-concepts/endpoints/#server-endpoints-api-routes) to be server-rendered. Each configuration option has a different default behavior, and allows indidual routes to opt-out of the default accordingly:
+
+- __`output: 'server'`__: Server-rendered by default. Any individual page or endpoint can be pre-rendered. Use this when most or all of your site should be server-rendered.
+- __`output: 'hybrid'`__: Pre-rendered to HTML by default. Any individual page or endpoint can *opt-out* of pre-rendering. Use this when most of your site should be static.
 
 ```js ins={5}
 // astro.config.mjs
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -13,7 +13,10 @@ import Since from '~/components/Since.astro';
 
 ## Enabling SSR in Your Project
 
-To get started, enable SSR features in development mode with the `output: server` configuration option:
+To enable SSR features in development mode, add one of the two `output` configuration options: `server` or (introduced in v2.6.0) `hybrid`. Both of these modes allow all or just some of your [pages](/en/core-concepts/astro-pages/) or [server endpoints](/en/core-concepts/endpoints/#server-endpoints-api-routes) to be server-rendered. Each configuration option has a different default behavior, and allows indidual routes to opt-out of the default accordingly:
```
natemoo-re, line 16:
```suggestion
To enable SSR features for production deployments, update your `output` configuration to `'server'` or `'hybrid'` (introduced in **v2.6.0**). Both of these modes control which [pages](/en/core-concepts/astro-pages/) or [server endpoints](/en/core-concepts/endpoints/#server-endpoints-api-routes) should be server-rendered. Each configuration option has a different default behavior, and allows individual routes to opt-out of the default accordingly:
```

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -13,7 +13,10 @@ import Since from '~/components/Since.astro';
...
+
+- __`output: 'server'`__: Server-rendered by default. Any individual page or endpoint can be pre-rendered. Use this when most or all of your site should be server-rendered.
```
natemoo-re, line 18:
```suggestion
- __`output: 'server'`__: Server-rendered by default. Use this when most or all of your site should be server-rendered. Any individual page or endpoint can *opt-in* to pre-rendering.
```

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -13,7 +13,10 @@ import Since from '~/components/Since.astro';
...
+- __`output: 'hybrid'`__: Pre-rendered to HTML by default. Any individual page or endpoint can *opt-out* of pre-rendering. Use this when most of your site should be static.
```
natemoo-re, line 19:
```suggestion
- __`output: 'hybrid'`__: Pre-rendered to HTML by default. Use this when most of your site should be static. Any individual page or endpoint can *opt-out* of pre-rendering.
```

natemoo-re (commented):
Overall structure looks great! Had a few suggestions for ordering / clarity

PR #3395, opened 2023-06-01
Document new Redirects feature

### LINKED ISSUE #3311: Redirects
- Implementation PR: https://github.com/withastro/astro/pull/7067/
- Stage 3 proposal: https://github.com/withastro/roadmap/pull/587
- Stage 2 proposal: https://github.com/withastro/roadmap/issues/466 

## Needs

- API reference docs
- ? - This feature is most useful in SSG, so maybe it should go there? This feature does work in SSR as well, but you can already do redirects in SSR (manually) so it's not adding a new capability, just making one easier.

## How it works

You can add a top-level `redirects` config with keys that are from routes and the values are the to route. For example:

```js

export default defineConfig({
  redirects: {
    '/api': '/docs/api'
  }
})
```

matthewp:

- New or updated content

#### Description

- Closes #3311
- Removes redirects docs from the `server-side-rendering` page.
- Updates `api-reference` to remove note about Astro.redirect being an SSR feature.
- Adds a section to `routing` explaining the redirects configuration and how it is prioritized.


src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -219,6 +220,74 @@ const { title, text } = page;
...
+
+These redirects follow the same rules as file-based routes. Dynamic routes are allowed as long as both the new and old route contain the same parameters, for ex:
+
+```js
+{
+  "/blog/[...slug]": "/articles/[...slug]"
+}
+```
+
+You can also specify the status code by providing an object as the value:
```
sarah11918, line 250:
```suggestion
Using SSR or a static adapter, you can also provide an object as the value, allowing you to specify the `status` code in addition to the new `destination`:
```

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -219,6 +220,74 @@ const { title, text } = page;
 </html>
 ```
 
+## Redirects (experimental)
+
```
sarah11918, line 224:
```suggestion
## Redirects

Sometimes you will need to redirect your readers to a new page, either permanently because your site structure has changed, or in response to an action such as logging in to an authrenticated route.

You can define rules to [redirect users to permanently-moved pages](#configured-redirects-experimental) in your Astro config. Or, [redirect users dynamically](#dynamic-redirects) as they use your site.

### Configured Redirects (experimental)
```
Especially since only the new (permanent, configured) redirects are actually experimental, I think for now this sectioning makes sense? Both types of redirects are sub-headings of redirects?

sarah11918 (commented):
Looking great! What do you think about the tweaking of making both these configured and dynamic redirects sub-headings of a Redirects section?

src/content/docs/en/guides/cms/contentful.mdx
```diff
@@ -442,7 +442,7 @@ const data = await contentfulClient.getEntries<BlogPost>({
 ---
 ```
 
-If the entry is not found, you can redirect the user to the 404 page using [`Astro.redirect`](/en/guides/server-side-rendering/#astroredirect).
+If the entry is not found, you can redirect the user to the 404 page using [`Astro.redirect`](/en/core-concepts/routing#dynamic-redirects).
```
sarah11918, line 445:
```suggestion
If the entry is not found, you can redirect the user to the 404 page using [`Astro.redirect`](/en/core-concepts/routing/#dynamic-redirects).
```

src/content/docs/en/core-concepts/routing.mdx
matthewp: Yeah, good point, I like this.

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -219,6 +220,79 @@ const { title, text } = page;
 </html>
 ```
 
+## Redirects
+
+Sometimes you will need to redirect your readers to a new page, either permanently because your site structure has changed, or in response to an action such as logging in to an authrenticated route.
```
sarah11918, line 225:
```suggestion
Sometimes you will need to redirect your readers to a new page, either permanently because your site structure has changed, or in response to an action such as logging in to an authenticated route.
```

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -219,6 +220,79 @@ const { title, text } = page;
...
+  redirects: {
+    '/old-page': {
+      status: 302,
+      destination: '/new-page'
+    }
+  }
+});
+```
+
+When running `astro build` Astro will write out HTML files with the [meta refresh](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#examples) tag by default. Supported adapters will instead write out the host's configuration file with the redirects.
```
sarah11918, line 273:
```suggestion
When running `astro build`, Astro will output HTML files with the [meta refresh](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#examples) tag by default. Supported adapters will instead write out the host's configuration file with the redirects.
```

sarah11918:
I think we're good!! It's time for Yan to find everything we missed on all the "merge on release" docs, so dev is off the clock!

sarah11918 (approved):

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -219,6 +220,79 @@ const { title, text } = page;
 </html>
 ```
 
+## Redirects
+
+Sometimes you will need to redirect your readers to a new page, either permanently because your site structure has changed, or in response to an action such as logging in to an authenticated route.
```
Yan-Thomas, line 225: Unnecessary comma
```suggestion
Sometimes you will need to redirect your readers to a new page, either permanently because your site structure has changed or in response to an action such as logging in to an authenticated route.
```

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -219,6 +220,79 @@ const { title, text } = page;
...
+  experimental: {
+    redirects: true
+  },
+  redirects: {
+    '/old-page': '/new-page'
+  }
+});
+```
+
+These redirects follow the same rules as file-based routes. Dynamic routes are allowed as long as both the new and old route contain the same parameters, for example:
```
Yan-Thomas, line 247: Not sure how this works in English, but "new and old" is an implicit plural, and supposedly "route" should be changed to agree in number.
```suggestion
These redirects follow the same rules as file-based routes. Dynamic routes are allowed as long as both the new and old routes contain the same parameters, for example:
```

Yan-Thomas (commented):
Overall LGTM, added two small nit suggestions!

src/content/docs/en/core-concepts/routing.mdx
sarah11918: yes, this absolutely correct, Yan!

PR #3414, opened 2023-06-03
Fixing content collections guide "Accessing referenced data" example

leomp12:

#### Description

- What does this PR change? It fixes the buggy example at https://docs.astro.build/en/guides/content-collections/#accessing-referenced-data , currently `<h1>{blogPost.data.title}</h1>` would fire fatal error with undefined `blogPost`.
- Did you change something visual? Nope.

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -358,7 +358,7 @@ Any [references defined in your schema](#defining-collection-references) must be
 ---
 import { getEntry, getEntries } from 'astro:content';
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Great catch, @leomp12! :raised_hands: 

I guess this fix could have gone either way: update the script or update the `<h1>`, but I think having `blogPost` vs just plain old `data` in this example is nice to distinguish it from the `author`, so let's do this thing!  Welcome to Team Docs!  :partying_face:

leomp12:
> but I think having blogPost vs just plain old data in this example is nice to distinguish it from the author

yup, that's exactly what I thought...

PR #3416, opened 2023-06-04
Update content-collections.mdx: true drafts

colmanhumphrey:

#### Description

- Replaces filter back to what it was previous to https://github.com/withastro/docs/pull/2371. Filtering to non-drafts makes sense too, but currently either the variable name (and the copy above) or the filter should change for consistency.
- @FredKSchott, were you thinking of maybe changing it to be non-draft entries?

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -390,7 +390,7 @@ You can use this to filter by any content criteria you like. For example, you ca
 ```js
 // Example: Filter out content entries with `draft: true`
 import { getCollection } from 'astro:content';
-const blogEntries = await getCollection('blog', ({ data }) => {
+const publishedBlogEntries = await getCollection('blog', ({ data }) => {
   return data.draft !== true;
 });
 ```
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -391,7 +391,7 @@ You can use this to filter by any content criteria you like. For example, you ca
 // Example: Filter content entries with `draft: true`
 import { getCollection } from 'astro:content';
 const draftBlogEntries = await getCollection('blog', ({ data }) => {
-  return data.draft !== true;
+  return data.draft === true;
 });
 ```
 
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -388,10 +388,10 @@ The `slug` property is specific to content collections, and will not be availabl
 You can use this to filter by any content criteria you like. For example, you can filter by properties like `draft` to prevent any draft blog posts from publishing to your blog:
 
 ```js
-// Example: Filter content entries with `draft: true`
+// Example: Filter out content entries with `draft: true`
 import { getCollection } from 'astro:content';
-const draftBlogEntries = await getCollection('blog', ({ data }) => {
-  return data.draft === true;
+const blogEntries = await getCollection('blog', ({ data }) => {
+  return data.draft !== true;
 });
 ```
 


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -391,7 +391,7 @@ You can use this to filter by any content criteria you like. For example, you ca
 // Example: Filter content entries with `draft: true`
 import { getCollection } from 'astro:content';
 const draftBlogEntries = await getCollection('blog', ({ data }) => {
-  return data.draft !== true;
+  return data.draft === true;
```
sarah11918, line 394: Since filtering is probably mostly used to NOT include drafts, let's revert this to: 

```suggestion
  return data.draft !== true;
```
But update the copy in the code comment. (Stupid GitHub won't let me select all these lines together in one comment, so that one'll come next.)

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -391,7 +391,7 @@ You can use this to filter by any content criteria you like. For example, you ca
 // Example: Filter content entries with `draft: true`
```
sarah11918, line 391:
```suggestion
// Example: Filter out content entries with `draft: true`
```
Something like this works?

sarah11918 (commented):
Hey @colmanhumphrey, thanks for spotting this! 

You're right that it's inconsistent now, and I think it's probably more helpful to filter OUT draft posts, since that's what people are more likely to be doing. So, I think Fred's code had the write idea, but probably slipped up when updating the text.

Left some suggestions that I think would work, for your consideration!

src/content/docs/en/guides/content-collections.mdx
colmanhumphrey: Hey thanks @sarah11918! We'll also need to change the variable name, as otherwise the conflict remains. I assume `blogEntries` is good enough?

sarah11918: D'oh! Yup, you're right! Got a good variable name for NON-draft posts? maybe something like publishedPosts ?

(The example is so much easier to write for drafts, but most people are going to want NOT drafts!  :sweat_smile:  So, I think we should optimize for code people can use. We'll jump on the grenade and make it nice!)

colmanhumphrey: Oh sorry, I made an edit there and we crossed wires. I'm fairly easy here, `publishedPosts` or `publishedBlogEntries` both seem fine, do you prefer the former?

sarah11918: No preference! We'll go with yours!

sarah11918: I think I *do* like the explicitness of `published` though, rather than just `blogEntries`, for illustrating what we're doing. 

So, my slight vote is for `publishedBlogEntries` over `blogEntries`. What do you think? Am I being toooo pedantic?

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -388,9 +388,9 @@ The `slug` property is specific to content collections, and will not be availabl
 You can use this to filter by any content criteria you like. For example, you can filter by properties like `draft` to prevent any draft blog posts from publishing to your blog:
 
 ```js
-// Example: Filter content entries with `draft: true`
+// Example: Filter out content entries with `draft: true`
-const draftBlogEntries = await getCollection('blog', ({ data }) => {
+const blogEntries = await getCollection('blog', ({ data }) => {
```
sarah11918, line 393:
```suggestion
const publishedBlogEntries = await getCollection('blog', ({ data }) => {
```
Ok, I'm suggesting it @colmanhumphrey! Docs are for teaching/learning, so let's be super explicit!  :smile:

colmanhumphrey: I think it makes sense — it's easy to understand and when people use this in their actual code, they know they can just change it

PR #3418, opened 2023-06-04
Update endpoints.mdx

hvlxh:
i change the code(endpoint functions using `APIRoute`) into better looking code

### DIFF
--- a/src/content/docs/en/core-concepts/endpoints.mdx
+++ b/src/content/docs/en/core-concepts/endpoints.mdx
@@ -45,7 +45,7 @@ You can also type your endpoint functions using the `APIRoute` type:
 ```ts
 import type { APIRoute } from 'astro';
 
-export const get: APIRoute = async function get ({params, request}) {
+export const get: APIRoute = ({ params, request }) => {
 ...
 ```
 


src/content/docs/en/core-concepts/endpoints.mdx
```diff
@@ -45,7 +45,7 @@ You can also type your endpoint functions using the `APIRoute` type:
 ```ts
 import type { APIRoute } from 'astro';
 
-export const get: APIRoute = async function get ({params, request}) {
+export const get: APIRoute = ({ params, request }) => {
```
delucis, line 48: Thanks @hvlxh!

This is OK, but we should keep the `async` keyword here otherwise things like the `await fetch()` in the example above wouldn’t work:
```suggestion
export const get: APIRoute = async ({ params, request }) => {
```

sarah11918:
Thank you for taking the time to contribute to Astro docs, @hvlx!  But I'm looking at the earlier example, which does include `async function` and I do think that's the better option for examples on this page, so that they are all consistent.

The code samples we use in docs are meant to give as much information to the readers as possible, who might not be very experienced coders. We do tend to prefer more explicit code, and I think keeping `async function` is helpful for some of our less-experience readers, even if it's not the code our more experienced readers would write themselves.

So, we won't be changing this sample, but I hope you'll make another PR when you see something that needs to be fixed, or to help us make our docs better!  :raised_hands:

PR #3423, opened 2023-06-05
Achieve consistency with indefinite article in MPA vs SPA concept en-only

finnala:

#### Description

This PR aims to achieve consistency in the wording of "a SPA" and "an SPA" in the en [SPA vs MPA](https://github.com/withastro/docs/blob/main/src/content/docs/en/concepts/mpa-vs-spa.mdx) concept page. In some instances the wording "an SPA" is used and in others "a SPA" is used. "an SPA" seems to appear more commonly.

This PR changes the two instances of "a SPA" to "an SPA" to achieve consistency.


sarah11918 (approved):
Thanks for this, @finnala! We may never know what is right, but we can be wrong together, always! :sweat_smile: Welcome to Team Docs!  :partying_face:

PR #3432, opened 2023-06-06
[migrate-to] update docs suggestions to starlight

sarah11918:
Updates four of our "migrate to" guides that previously recommended starting with our docs template.
- refers (and links to) Starlight instead of "docs template"
- updates `-- template` install instructions for Starlight


delucis (approved):
This looks great! Thank you for going through these @sarah11918 🙌

PR #3471, opened 2023-06-10
Proposed change: Fixing a typo in the documentation.

ShiniDev:
The previous sentence contained a typo where the phrase 'when clicking the button' was omitted. I have added it to provide clarity and improve the sentence's accuracy. The said changes are based on the code that is provided on the tutorial, the greeting is now not random when the page is loaded due to previous fix.

### DIFF
--- a/src/content/docs/en/tutorial/6-islands/1.mdx
+++ b/src/content/docs/en/tutorial/6-islands/1.mdx
@@ -80,7 +80,7 @@ This component will take an array of greeting messages as a prop and randomly se
     </BaseLayout>
     ```
 
-    Check the preview in your browser: you should see a random greeting, but the button won't work!
+    Check the preview in your browser: you should see a random greeting when clicking the button, but the button won't work!
 
 
 4. Add a second `<Greeting />` component with the `client:load` directive.


sarah11918:
Hi @ShiniDev! Thank you for taking the time to contribute, but your fix isn't exactly what's needed here. The use of the word "should" here is to allow the reader to know they've performed the action correctly, this is what they should see: a random greeting. We're not using should to mean "what one would expect to happen" because one should not assume they know what to expect when they are learning something, and it's not our intention to describe intended behaviour at this point.

The behaviour you are describing re: the Greeting not being random *after* first load is correct! But, the purpose of this sentence is to mention that you WILL see a random greeting appear (nothing to do with the button) when you visit the page preview in dev mode, and yet the button doesn't work. Insights like yours are what we hope the reader will have, but we explicitly choose not to explain all the details of everything going on so that the reader can discover things through doing, and not have to read through a lot of explanation. (That's what the full docs are for!) So, we won't be changing this sentence at this time, but we appreciate that when you believed you'd found an error, you took the time to help and fix it!  🙌  We hope you'll do that again!

ShiniDev:
Thanks for the reply! appreciate it.

PR #3481, opened 2023-06-12
Fix formatting on an error page

delucis:

#### Description

This error has been removed from Astro, so won’t update with changes made there. I still think we should keep the page around in case someone running an old version of Astro runs into the error, so fixing the message formatting in docs.

### DIFF
--- a/src/content/docs/en/reference/errors/locals-not-serializable.mdx
+++ b/src/content/docs/en/reference/errors/locals-not-serializable.mdx
@@ -13,9 +13,7 @@ import DontEditWarning from '~/components/DontEditWarning.astro'
 <DontEditWarning />
 
 
-> **LocalsNotSerializable**: 
-            return `The information stored in `Astro.locals` for the path "HREF" is not serializable.\nMake sure you store only serializable data.`;
-         (E03034)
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Makes sense!

src/content/docs/en/reference/errors/locals-not-serializable.mdx
```diff
@@ -13,9 +13,7 @@ import DontEditWarning from '~/components/DontEditWarning.astro'
 <DontEditWarning />
 
 
-> **LocalsNotSerializable**: 
-            return `The information stored in `Astro.locals` for the path "HREF" is not serializable.\nMake sure you store only serializable data.`;
-         (E03034)
+> **LocalsNotSerializable**: The information stored in `Astro.locals` for the path "HREF" is not serializable. Make sure you store only serializable data. (E03034)
```
delucis, line 16: Small tweak. Remembered we use code backticks for variables in these message examples:
```suggestion
> **LocalsNotSerializable**: The information stored in `Astro.locals` for the path "`HREF`" is not serializable. Make sure you store only serializable data. (E03034)
```

PR #3505, opened 2023-06-15
fix: correct way to type `Locals`

ematipico:

#### Description

The correct way to type the `Locals` object was changed, and we forgot to update doc site. In fact, the `examples/` folder in the main repo is different:

https://github.com/withastro/astro/blob/3f1cb6b1a001fb03419a313f72c9f4846b890fe0/examples/middleware/src/env.d.ts#L2-L9


Yan-Thomas (approved):
LGTM!

PR #3519, opened 2023-06-17
API reference: make `AstroCookie.value` optional

wackbyte:

- New or updated content

#### Description

Makes `AstroCookie.value`'s type `string | undefined` rather than `string` as this is its actual type.

(Note: that table formatting may need to be fixed :sweat_smile:)


wackbyte:
It's written as `string | undefined` in the code [here](https://github.com/withastro/astro/blob/f87680f6df2e8525d9b8667e2ae7e600543e1c73/packages/astro/src/core/cookies/cookies.ts#L18)

Yan-Thomas (approved):
LGTM! Thanks for catching this @wackbyte 💜

PR #3527, opened 2023-06-19
guide: ISR | Incremental Static Regeneration

lilnasy:
- New or updated content

#### Rendered Preview
https://docs-git-fork-lilnasy-main-astrodotbuild.vercel.app/en/guides/incremental-static-regeneration/


Alynva:
I think that you should remove the `else` block (keep it's content), so when you remove the stale cache, it still runs that code

lilnasy:
@Alynva which `else` do you mean? Could you make a comment in the diff?

src/content/docs/en/guides/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,199 @@
...
+    
+    else {
+        const response = await next()
+        cache[request.url] = {
+            expires: Date.now() + 60 * 1000,
+            response: response.clone()
+        }
+        // return fresh response
+        return response
+    }
```
Alynva, line 59: I mean like this, so when the request is cached but it is stale, after it being deleted the request is treated as non-cached. Otherwise, the function would return undefined when the cache is stale.

```suggestion
    const response = await next()
    cache[request.url] = {
        expires: Date.now() + 60 * 1000,
        response: response.clone()
    }
    // return fresh response
    return response

```

src/content/docs/en/guides/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,199 @@
...
+    else {
+        const response = await next()
+        
+        // set expiry
+        response.headers.set('X-expires', String(Date.now() + 60 * 1000))
+
+        await cache.put(request.url, response.clone())
+        // return fresh response
+        return response
+    }
```
Alynva, line 100: ...and the same in this code:
```suggestion
    const response = await next()
    
    // set expiry
    response.headers.set('X-expires', String(Date.now() + 60 * 1000))

    await cache.put(request.url, response.clone())
    // return fresh response
    return response
```

Alynva (changes requested):
@lilnasy Sorry, I was on mobile. I've added the suggestions now

src/content/docs/en/guides/incremental-static-regeneration.mdx
lilnasy: You're right, my example did not work, thanks for this!

I would probably simplify it even more later. Fallthrough can be really tricky to think about, especially for new developers, which astro gets often.

src/content/docs/en/guides/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,218 @@
...
+```
+Vercel only supports `s-maxage` and `stale-while-revalidate` in the Cache-Control header. See [vercel documentation](https://vercel.com/docs/concepts/edge-network/caching) for current information.
+
+### Netlify
+Netlify supports caching via two methods: `Cache-Control` headers, which works with edge and serverless functions; and, On-demand Builders, a type of function dedicated for optimized caching.
+
+#### On-demand Builders
+On-demand Builders are serverless functions used to generate web content as needed that’s automatically cached on Netlify’s Edge CDN. They enable you to build 
+To enable them, set the `builders` option to `true` in your config file.
+```ts title="stro.config.js" ins={6}
```
dreyfus92, line 136:
```suggestion
```ts title="astro.config.js" ins={6}

```

src/content/docs/en/guides/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,218 @@
...
+import { defineConfig } from 'astro/config'
+import netlify from '@astrojs/netlify/functions'
+
+export default defineConfig({
+  output: 'server',
+  adapter: netlify({ builders: true })
+})
+```
+By default, pages for your site will be built when a user visits them for the first time and then cached at the edge for subsequent visits. To set a revalidation time, call the `runtime.setBuildersTtl(ttl)` local with the duration (in seconds). For example, to set a revalidation time of 60 seconds:
+```ts title="src/pages/index.astro"
```
dreyfus92, line 146:
```suggestion
```astro title="src/pages/index.astro"

```

dreyfus92 (changes requested):
Maybe this fixes might help @lilnasy

src/content/docs/en/guides/incremental-static-regeneration.mdx
lilnasy: Oh, thanks for the catch! I was looking at the wrong code block.

src/content/docs/en/guides/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,276 @@
...
+```ts title="src/middleware.ts" ins={9}
+import { defineMiddleware } from "astro:middleware"
+
+const cache = new Map
+
+export const onRequest = defineMiddleware(async ({ request }, next) => {
+    
+    let expiresIn = 60
+    
+    locals.revalidate = seconds => { expiresIn = seconds }
```
artmsv, line 96: I don't think `locals` is globally exposed, is it part of `onRequest` ctx argument?
```suggestion
export const onRequest = defineMiddleware(async ({ request, locals }, next) => {
    
    let expiresIn = 60
    
    locals.revalidate = seconds => { expiresIn = seconds }

```

lilnasy: You're exactly right! Thanks

src/content/docs/en/guides/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,276 @@
...
+        const expires = Number(cachedResponse.headers.get('X-Expires'))
+        
+        if (expires > Date.now()) {
+            return cachedResponse
+        }
+        
+        else {
+            await cache.delete(request)
+        }
+    }
```
dreyfus92, line 163:
```suggestion
    if (cachedResponse) {
        const expires = Number(cachedResponse.headers.get('X-Expires'))
        if (expires > Date.now()) {
            return cachedResponse
        } else {
            await cache.delete(request)
        }
    }
```

dreyfus92 (changes requested):
Just wondering if having new lines in between is intentional to improve readibility. If not I leave this is minor suggestion @lilnasy.

lilnasy:
You're right, it was for readability - in my editor. I will change that to align with the convention, if we have one.

src/content/docs/en/guides/incremental-static-regeneration.mdx
lilnasy: You're right, it was for readability - in my editor. I will change that to align with the convention, if we have one.

src/content/docs/en/guides/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,272 @@
...
+```ts
+const cache = await caches.open("astro")
+```
+You just opened a [`Cache`](https://developer.mozilla.org/en-US/docs/Web/API/Cache)! You can now call `await cache.put(request, response)` to save a response, and `await cache.match(request)` to retrieve it. See [web.dev's Cache API quick guide](https://web.dev/cache-api-quick-guide/) if you want to learn more.
+
+### Self-hosted Node.js
+ISR can be implemented in Node.js by writing a caching middleware. The following middleware caches every response in memory for 60 seconds:
+
+```ts title="src/middleware.ts"
+import { defineMiddleware } from "astro:middleware"
```
artmsv, line 50: I'm not sure if it is for a new API, but currently it gives me `package not found` with this identifier.

```suggestion
import { defineMiddleware } from "astro/middleware"

```

artmsv: I noticed there are more examples with this format `astro:middleware` in this PR

dreyfus92: This is how you import that function https://docs.astro.build/en/guides/middleware/.

artmsv: > This is how you import that function https://docs.astro.build/en/guides/middleware/.

https://stackblitz.com/edit/github-oiqdwp-dddxlh?file=src%2Fmiddleware.js,src%2Fpages%2Findex.astro

Although this is `"astro": "^2.10.14",` and here what it gives me:

```
> @example/basics@0.0.1 dev
> astro dev

  🚀  astro  v2.10.14 started in 94ms
  
  ┃ Local    http://localhost:3000/
  
Error: The following dependencies are imported but could not be resolved:

  astro:middleware (imported by /home/projects/github-oiqdwp-dddxlh/src/middleware.js)

Are they installed?
    at eval (/home/projects/github-oiqdwp-dddxlh/node_modules/vite/dist/node/chunks/dep-df561101.js:45757:23)
    at async eval (/home/projects/github-oiqdwp-dddxlh/node_modules/vite/dist/node/chunks/dep-df561101.js:45166:38)
5:37:55 PM [vite] Error when evaluating SSR module /home/projects/github-oiqdwp-dddxlh/src//middleware: failed to import "astro:middleware"
|- Error: Cannot find module 'astro:middleware' imported from '/home/projects/github-oiqdwp-dddxlh/src/middleware.js'
    at nodeImport (/home/projects/github-oiqdwp-dddxlh/node_modules/vite/dist/node/chunks/dep-df561101.js:56062:25)
    at ssrImport (/home/projects/github-oiqdwp-dddxlh/node_modules/vite/dist/node/chunks/dep-df561101.js:55964:30)
    at eval (/home/projects/github-oiqdwp-dddxlh/src/middleware.js:3:50)
    at instantiateModule (/home/projects/github-oiqdwp-dddxlh/node_modules/vite/dist/node/chunks/dep-df561101.js:56026:15)

Failed to load url astro:middleware (resolved id: astro:middleware) in /home/projects/github-oiqdwp-dddxlh/src/middleware.js. Does the file exist?
```

dreyfus92: It is weird, cause locally seems to work just fine, you can even find the module declaration in the latest version of Astro:

```ts
declare module 'astro:middleware' {
	export * from 'astro/middleware/namespace';
}
```

lilnasy: We introduced this alias in 3.0 for consistency with astro:assets and astro:content.

https://github.com/withastro/astro/blob/astro@3.0.0/packages/astro/CHANGELOG.md#minor-changes

TheOtterlord:
**Heads up!**
We've just migrated the docs site from Netlify to Vercel. With that, the Netlify app has been removed, and will no longer update deploy previews. Vercel should post a comment with the new deploy preview the next time this PR is updated.

src/content/docs/en/guides/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,272 @@
...
+- Your content is located inside your project (for example, as markdown files in `src/content`).
+- You don't want the hassle of running a server or using a serverless platform.
+- Your website offers logged-in experiences. If each user needs to be sent a slightly different HTML from others, caching the page may not be possible.
+- The structure of your site (layouts, and styling) changes frequently. Changes to the structure require a full rebuild.
+
+## Implementing ISR for Your Project
+
+ISR involves saving the response such that it can be reused for future requests. The exact implementation depends on where you are deploying.
+
+If you aren't using an adapter already, you are going to need to add one. See the [integrations guide](/en/guides/integrations-guide/) for specific instructions. In your config, make sure that the `output` is either `"server"` or `"hybrid"`. Note that when using "hybrid", the pages you want generated, cached, and invalidated will need to be opted-out of prerendering. Otherwise, they will only be rendered once during build. See the [server side rendering guide](/en/guides/server-side-rendering/#opting-out-of-pre-rendering) to see how.
```
sarah11918, line 30:
```suggestion
If you aren't using an adapter already, you are going to need to add one. See the [integrations guide](/en/guides/integrations-guide/) for specific instructions. In your config, make sure that the `output` is either `"server"` or `"hybrid"`. Note that when using "hybrid", the pages you want generated, cached, and invalidated will need to be opted-out of prerendering. Otherwise, they will only be rendered once during build. See the [server side rendering guide](/en/guides/server-side-rendering/#opting-out-of-pre-rendering-in-hybrid-mode) to see how.

```

sarah11918:
🫡

selrond:
Why was this closed?

lilnasy:
It was just sitting in the backlog for too long. We may still revisit this, we just didn't get around to sitting down to go over it.

Although, a well-edited version of this was published as a blog post on logsnag!

https://logsnag.com/blog/implementing-isr-in-astro

aretrace:
This information should be included in the docs. Astro is particularly attractive for projects that would also happen to greatly benefit from this technique.

oemer-aran:
I would also add this to the official docs. It includes deployment details for different hosting services, which are very valuable. Also reading from the official docs conveys more trust than a blog post. I read through the posts on Logsnag. They were really insightful and very well written. But the code is more like a POC (a very good one), which could potentially be extended with real cache providers.

On-Demand Revalidation would really be a killer feature with astro.js. It's the only real missing piece for me. Especially if you want to serve live data from a CMS which has a API request limit, like Sanity.

Raik410:
Please add this to the official documentation

PR #3560, opened 2023-06-23
Update congrats workflow to use new shared action

delucis:

- Something else!

#### Description

- Updates our CongratsBot workflow to use the new shared one in https://github.com/withastro/automation
- Deletes the script file from the docs repo as it’s no longer needed
- Adds some custom emoji to the possible emojis to pick from at the start of messages. All the possible options:
   image


sarah11918 (approved):
Emojis approved, though I'm laughing at someone providing a HUGE contribution and seeing NWTWWHB...  😂

PR #3572, opened 2023-06-23
Fix titles link

Waxer59:

#### Description

Fixed the icon linking the page title to the URL.

* Previous behavior:
image
* Now behavior: 
image


dreyfus92 (approved):
LGTM ✅

Yan-Thomas:
I can see removing this CSS rule causes the viewport to work less in smaller viewports (I don't think it's a big problem considering most people aren't going to see the site in a 230px wide screen), but here's what's happening:

| Before | After|
|--------|--------|
|image| image |

Waxer59:
I have already added a mediaquery that solves this problem.

Yan-Thomas (approved):
Thanks, perfect!

sarah11918 (approved):
I am happy with the discussion I've seen here! This seems to have been a well-thought-out improvement, and I'm happy to approve subject to Yan or Chris being satisfied!  :purple_heart:

delucis (changes requested):
In general we always use `min-width` media queries instead of `max-width`. Can we change this to keep `overflow-x: hidden` the default style and then disable it at `min-width: 50em` instead, please?

Waxer59:
Changes made!

src/components/PageContent/PageContent.astro
```diff
@@ -91,7 +97,7 @@ const direction = rtlLanguages.has(lang) ? 'rtl' : 'ltr';
 			height: 0;
 		}
 	}
-
+
```
Waxer59, line 100:
```suggestion
```

delucis (approved):
Thanks @Waxer59! Made a tweak to avoid using `!important` which is generally best avoid if it’s at all possible. Here we just needed to move the media query `.content` style below the `.content` without a media query. I took the opportunity to group up the different media queries too.

Yan-Thomas (approved):
LGTM!

PR #3578, opened 2023-06-26
Release the translators!

sarah11918:
This marks 27 more pages `i18nReady: true`!

Notably not transated:
- Content Colletions
- Middleware
- Astro Syntax
- Firebase
- All the Recipes!!

Also marks some stubs for translation, which will be great first time PRs for people! (Maybe figure out a way to highlight and guide people toward `stub: true` when searching for ones to translate?)

Notably NOT yet marked:
- Experimental Assets, as this page will need to change significantly for V3 anyway
- Builder.io, as there's an existing full-guide PR being edited right now.

PR #3581, opened 2023-06-26
Remove Markdoc from `i18nNotReady` in docgen script

delucis:

- Something else!

#### Description

Removes the Markdoc integration from the set of pages not ready for i18n in the generator script for integration READMEs.

#3578 changed this manually but we need this change to make sure the CI bot doesn’t try to flip it back every day.


sarah11918 (approved):
I didn't realize we had a whole line of code just for this file! Happy to have Markdoc removed from this list so we can start translating it.

Yan-Thomas (approved):
Funny when you build stuff and then you completely forgets it exists! Thanks 🙌

delucis:
Going to live on the edge and merge before checks pass because this script is not even being run in those checks. 😎

PR #3635, opened 2023-06-30
Move code block titles into comments (i18nIgnore)

hippotastic:

#### Description

- To prepare the upgrade from our old built-in code snippets component to the new Expressive Code package, we have to make some changes to our MDX contents.
- Our old code snippets supported defining code block titles by adding a bold formatted paragraph directly above the fenced code block. This syntax is no longer supported in Expressive Code because it may be unexpected that surrounding markdown code influences the rendering of a code block.
- This PR therefore replaces any occurrences of this pattern with file name comments inside the code blocks.
- It also uses the proper comment syntax based on the respective file type and contents:
  - `astro`: If a frontmatter block is present, a `// JS comment` is added at the beginning of the frontmatter block. Otherwise, an `<!-- HTML comment -->` is added at the beginning of the code.
  - `markdown`/`md`: If a frontmatter block is present, a `# YAML/TOML comment` is added at the beginning of the frontmatter block. Otherwise, an `<!-- HTML comment -->` is added at the beginning of the code.
  - `js(x)`/`ts(x)`: A `// Single-line comment` is added at the beginning of the code.
  - `xml`: A `title="..."` attribute is added to the opening code fence as no title comments are supported inside `xml` snippets yet.
- Note: This PR includes changes to generated files in `src/content/en/guides/integrations-guide`. These changes will also be submitted to the source package README files to prevent them from being overwritten later.

### DIFF
--- a/src/content/docs/de/core-concepts/layouts.mdx
+++ b/src/content/docs/de/core-concepts/layouts.mdx
@@ -15,10 +15,9 @@ Layouts werden normalerweise im Verzeichnis `src/layouts` deines Projekts abgele
 
 ## Beispiel-Layout
 
-**`src/layouts/MeinLayout.astro`**
-
 ```astro
 ---
+// src/layouts/MeinLayout.astro
 ---
 <html>
   <head>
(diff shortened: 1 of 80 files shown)


sarah11918 (approved):
Inspected every English page by hand, and can confirm no wonkiness to any code snippets!  :partying_face: 

(I did notice that some snippets that never did have titles maybe should... but that's a story for another day!)

Great work, Hippo! Thanks for being so thorough, and for also tackling these integrations READMES at the source!

PR #3639, opened 2023-06-30
Allow blank issues

Yan-Thomas:

- Something else!

#### Description

This allows blank issues so that the Feedback Fish issues does not lose its old behavior.


delucis (approved):
Thanks for the quick fix @Yan-Thomas!

sarah11918 (approved):
Thanks Yan! Very e`fish`ent quick fix!

PR #3643, opened 2023-07-01
fix form example validation

0ui:

#### Description

The problem with the existing form example is that it generates an error object shaped like

```js
{ username: '', email: '', password: '' }
```

and then allows the form to submit if

```js
!Object.keys(errors).length
```

except that will always be `false` because the object's keys aren't removed. It's simply unchanged if there were no errors. This update instead checks if any of the object's values are truthy, allowing it to submit if no errors are appended to the values. It could also be done a few other ways, this is just to make the example valid.


dreyfus92 (approved):
Nice catch @0ui, LGTM ✅

kevinzunigacuellar:
Nice catch! and very clean implementation to check for errors 💜

PR #3644, opened 2023-07-01
Migrate code blocks to Expressive Code

hippotastic:

#### Description

- Migrates Astro Docs from our built-in `CodeSnippets` component to the new Expressive Code package using the `astro-expressive-code` integration.
- Refactoring the built-in component into separate NPM packages is an important step to keep the Astro Docs repo lean. It also makes the code snippet features accessible to a wider audience, including Starlight, and improves extensibility through a plugin architecture.
- The new code snippet frames look slightly different from the old design, but I've tried to keep them as close to the Astro Docs style as possible to make everyone feel at home right away.


hippotastic:
Let's collect any issues or things we may want to change here!

I'll start: One issue I'm seeing in the deploy preview is that the copy button doesn't seem to be working on mobile Safari. I'll need to investigate why the API doesn't want to work as intended there.

Yan-Thomas:
Great to see Expressive Code in Astro Docs finally! It worked perfectly for me in Chrome (Windows 11) and Chrome & Samsung Internet (Android 11)

hippotastic:
The copy button is now working in mobile Safari, too. It was a simple focus CSS selector issue. Looking for any other issues now! Feel free to join :)

hippotastic:
I discovered some code blocks that used tabs for indentation and therefore introduced an option to normalize tabs to a given number of spaces. We're now using the default tab width of 2, matching our `.editorconfig` setting and the old Astro Docs code snippets behavior.

sarah11918:
I'm scanning all the pages Hippo and everything is looking great so far!

One weird code snippet I just came across (desktop, Linux, Chrome) is on the `astro-components` page, where the highlighted text is green, not blue:
astro-components

It's not a problem per se but it's the only one that seems green and not blue? Probably worth looking at!

sarah11918:
And, found a red one this time!
thistimered

hippotastic:
@sarah11918 Thank you very much for looking through!

That the highlights are green is actually correct there, as the code snippet is using the "inserted" (`ins`) style for the markings. The opening fence looks like this:

````md
```astro ins="= \"Hello\"" ins="= \"Astronaut\""
````

Here's the excerpt from the source file:

https://github.com/withastro/docs/blob/5ccc7e19bb836928e3839ba74a2dc6aa51fdd12c/src/content/docs/en/core-concepts/astro-components.mdx#L158-L164

So, this doesn't look like a bug to me, it's just following the instructions given by the docs page author.

**Comment by hippotastic** [comment]
Same applies for the red markings you found. This is the "deleted" style. It's all the same as in the old Astro Docs 😄

**Comment by sarah11918** [comment]
Perfect! I guess we use it so infrequently, I forgot you could `ins` and `del` strings!  :laughing:

**Comment by sarah11918** [comment]
Also pointing out that the RTL code appears right-justified, but our existing docs have the Arabic code *left*-justified, so I would assume that's the correct alignment? We'll check with our translators!

**Comment by hippotastic** [comment]
Code blocks in RTL languages are fixed now, too! Did you find anything else or are we ready to make the switch?

**Review by sarah11918 — APPROVED** [review]
I've just tested again on a handful of Sarah's evil devices, and it all looks good on my end! (visually)  :partying_face: Spectacular job, Hippo! :hippopotamus:

**Comment by hippotastic** [comment]
Glad to hear that, thank you for testing so thoroughly! I'll wait for @delucis before merging though. You've all had so much patience, now it's my turn. 😄

File: syntax-highlighting-theme.ts
Comment (delucis) [inline syntax-highlighting-theme.ts:1]: This is only used inside `integrations/expressive-code.ts` I think? I’d love to see it moved out of the root folder in that case and alongside the code using it.

File: public/index.css [hunk 1]
```diff
@@ -329,34 +329,11 @@ code {
...
-	direction: ltr;
-	unicode-bidi: isolate;
-}
-
 /*RTL Fix Arrows*/
 [dir='rtl'] .toc-mobile-container svg {
 	transform: rotateY(180deg);
 }
 
-table,
```
delucis, line 343: Looks like this style also applied to `<table>`s:

| Live site | This PR |
|---|---|
| Table displayed in Astro’s docs | Table displayed in Astro’s docs |

Not sure how much we care — the tables are rarely used — but we should probably only remove `pre` from this block unless we feel like investigating thoroughly how much of these styles are needed for tables?

delucis (commented):
Can’t believe this PR is here! Such a highly anticipated moment! This looks great @hippotastic and boy does it feel good seeing this: Git diff summary showing 205 additions and 1298 deletions  😀 

Spotted a couple of things mentioned below.

One additional one: on the live Astro site, the copy button is revealed when anywhere over the code snippet is hovered or focused, not only the button itself. This makes discovery much more likely than when we rely on people hovering over the relatively small area of just the copy button. Do you think that’s something we can fix up reasonably quickly?

ElianCodes (commented):
This is awesome!

public/index.css
hippotastic: First of all, thank you for looking through this so thoroughly!

The removal of `table` was actually intentional. I did check all occurrences in our live content (which are limited to exactly two tables - the one you've screenshotted, plus one in the legacy v1 migration guide), and the tables honestly looked better and more readable to me both on desktop and mobile in my new version without these styles.

Besides that, I also don't think that applying a default `pre`-like styling to every text in a table is a good idea. Tabular data isn't always automatically code. In the screenshot you provided, I'd actually find a `pre` styling of the "collection" column misleading as it distracts from the actual keywords on the right that authors may want to copy & paste into their `package.json` files.

Let me know what you think!

delucis: Thanks for explaining. I never loved these styles (and usually advocate against table usage when it comes up 😄) so as long as this was intentional, I’m OK with it 👍

delucis (approved):
Thanks for the updates @hippotastic! Confirming everything looks good on my end now 🙌

PR #3645, opened 2023-07-01
i18n(es): Updated graphql example

sasoria:

- New or updated content

#### Description

- Updated the graphql example from this [commit](https://github.com/withastro/docs/commit/6b4fc99ee1924298ab1904be42c715f27000a001).

### DIFF
--- a/src/content/docs/es/guides/data-fetching.mdx
+++ b/src/content/docs/es/guides/data-fetching.mdx
@@ -69,38 +69,33 @@ export default Movies;
 
 Astro también puede usar `fetch()` para consultar a un servidor GraphQL con cualquier query de GraphQL válida.
 
-```astro title="src/components/Weather.astro" "await fetch"
+```astro title="src/components/Film.astro" "await fetch"
 ---
-const response = await fetch("https://graphql-weather-api.herokuapp.com",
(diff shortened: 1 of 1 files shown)


dreyfus92:
Hey @sasoria, thanks for bringing this up, but I think that in order to make an update on the docs, you should start by updating the EN version of them which is the default one. Afterwards if your pr gets merged the script that we have built in in the repo will notify other if the EN version was updated and check if other langs require to make the update based on the changes that were applied to the EN version of the docs.

sasoria:
Thanks for the explanation, but I'm not sure I understand. I updated the EN docs in march and the changes never made it to the ES docs. Here's the original [issue](https://github.com/withastro/docs/issues/2775) I solved.

Do we still want the outdated example in the ES docs?

dreyfus92:
My bad I wasn't aware that this was previously updated, I just noticed that our script never notified us about this update. You can check here what notify us when something needs to be updated to other locales [The Translation Status Overview](https://i18n.docs.astro.build/)

dreyfus92 (approved):
LGTM ✅

Waxer59 (approved):
LGTM ✔

Yan-Thomas (approved):
Nice catch, thanks peeps!

PR #3653, opened 2023-07-05
feat: `astro info` command

ematipico:

- New or updated content

#### Description

Context: https://github.com/withastro/astro/pull/7432

The PR adds documentation of the new command.

### DIFF
--- a/src/content/docs/en/reference/cli-reference.mdx
+++ b/src/content/docs/en/reference/cli-reference.mdx
@@ -274,6 +274,10 @@ Adds an integration to your configuration. Read more in [the integrations guide]
 
 Launches the Astro Docs website directly from the terminal.
 
+## `astro info`
+
+Reports useful information about the Astro environment. Useful to providing information when opening an issue.
+
 ## `astro telemetry`
 
 Sets telemetry configuration for the current CLI user. Telemetry is anonymous data that provides the Astro team insights into which Astro features are most often used.


src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -274,6 +274,10 @@ Adds an integration to your configuration. Read more in [the integrations guide]
 
 Launches the Astro Docs website directly from the terminal.
 
+## `astro info`
```
ematipico, line 277: I thought that the commands were in alphabetic order.

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -274,6 +274,10 @@ Adds an integration to your configuration. Read more in [the integrations guide]
...
+
+Reports useful information about the Astro environment. Useful to providing information when opening an issue.
```
sarah11918, line 279:
``````suggestion
Reports useful information about your current Astro environment. Useful for providing information when opening an issue.

```shell
astro info
```

Example output:

```
Astro version            v2.6.6
Package manager          pnpm
Platform                 darwin
Architecture             arm64
Adapter                  @astrojs/vercel/serverless
Integrations             None
```
``````
Snagged this from the changeset!

**Review by sarah11918 — COMMENTED** [review]
Just pulled in from the Astro PR's changeset to show the sample output. Also, we need to add this above around like 70-ish where we list the output of the help command!

File: src/content/docs/en/reference/cli-reference.mdx (hunk 1 above)
Comment (sarah11918) [reply src/content/docs/en/reference/cli-reference.mdx:278]: Apparently not :joy:

File: src/content/docs/en/reference/cli-reference.mdx [hunk 3]
```diff
@@ -67,6 +67,7 @@ Commands
             check  Check your project for errors.
               dev  Start the development server.
              docs  Open documentation in your web browser.
+             info  Emit useful information about Astro current setup. Useful for debugging.
```
sarah11918, line 70:
```suggestion
             info  List info about your current Astro setup.
```
This is way longer than the others. Do you think something like this is good enough?

ematipico: No problem, I will update the code too

PR #3657, opened 2023-07-05
Update PR template

Waxer59:

- Something else!

#### Description

Update PR template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,13 @@
 <!-- Thank you for opening a PR! We really appreciate you taking the time to help out 🙌 -->
 
+<!-- 
+Do not delete this field, it is used to indicate in which future version this document should be published, if your PR should be merged with the current version or you are not sure simply replace "<version>" with "current".
+
+Minor content corrections, translations, new recipes/guides among others should set "<version>" to "current".
+-->
+
+### Release in Astro version `<version>`
+
 #### What kind of changes does this PR include?
 <!-- Delete any that don’t apply -->
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,12 +1,9 @@
 <!-- Thank you for opening a PR! We really appreciate you taking the time to help out 🙌 -->
 
 <!-- 
-Do not delete this field, it is used to indicate in which future version this document should be published, if your PR should be merged with the current version or you are not sure simply replace "<version>" with "current".
-
-Minor content corrections, translations, new recipes/guides among others should set "<version>" to "current".
+NOTE: Replace "current" with a future version of Astro if this PR is for an upcoming release (e.g. a feature not yet released) and include a link to the `withastro/astro` implementation PR.
 -->
-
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,5 +1,10 @@
 <!-- Thank you for opening a PR! We really appreciate you taking the time to help out 🙌 -->
 
+<!-- 
+NOTE: Replace "current" with a future version of Astro if this PR is for an upcoming release (e.g. a feature not yet released) and include a link to the `withastro/astro` implementation PR.
+-->
+### For Astro version: `current` <!-- [<`withastro/astro` implementation PR>](<url>) -->
```
sarah11918, line 6:
```suggestion
<!-- 
NOTE: If this PR is for an upcoming release (e.g. a feature not yet released), uncomment the line below. Replace `version` with a future version of Astro (e.g. `2.8.0`) and include a link to the `withastro/astro` implementation PR.
-->

<!-- ### For Astro version: `version`. See [Astro PR #](url). -->
```
I think it makes more sense to state the "if" first, so that if not, people just ignore it!

What do you think about this? I think maybe starting with the line entirely hidden, and just letting the feature dev's uncomment it is actually MORE helpful to me, because I won't see as many "mistakes" and will ONLY see this line if I have to worry about not merging immediatly.

sarah11918: I'm going to commit to see if we get the Netlify checks to pass...

Waxer59: I agree with both things. the user will simply ignore it as you mention as its PR is not of that type. and leaving the title commented by default will confuse the users even less and will be much more readable for you as you mentioned.

dreyfus92:
I'm not sure why would some one open a PR on upcoming versions of Astro, if `astro docs` are based on the current version of Astro. Wouldn't they open a PR in the `astro` repo instead of opening one here?

delucis (commented):
I don’t think this is a terrible idea! I do wonder if it needs to be right up top given it’s not relevant and somewhat confusing for most contributors?

Could be a simpler comment after the description bullet points?

```md
<!-- Are these docs for an upcoming Astro release? -->
<!-- Uncomment the line below and fill in the future Astro version and link the relevant `withastro/astro` PR.  -->
<!-- ### For Astro version: `version`. See [Astro PR #](url). -->
```

sarah11918:
Oooh, I like that Chris! And yes, I'd be fine with it at the bottom of the template.

Waxer59:
Personally I think that being a title and being at the beginning of the md would be more striking and would be a good reminder for the PR to put this extra information. But I would also agree to move it to the bottom of the document if you prefer.

delucis:
I think this particular question applies to maybe 5% of PRs — usually PRs from regular Astro contributors — so it’s probably not important enough to need the extra prominence of being at the beginning. It’s enough to be a reminder lower in the template.

Waxer59:
I agree, I have applied these changes

sarah11918 (approved):
This looks great! Thanks, team, for coming together and making something awesome! I love it!  :purple_heart: 

Will merge when the checks pass!  :partying_face:

PR #3659, opened 2023-07-06
Clarify that file-based routes supersede redirects per priority level

paulrudy:
Since the list of route priority levels immediately precedes this, this language might make it clearer that file-based routes take precedence over redirects at the same route priority level.

#### Description

A change to docs text for clarity relating to route priority levels and redirects


sarah11918 (approved):
Looks great, @paulrudy, thank you!  :partying_face:

Issue #3251, opened 2023-05-15
Document fetch API particular behavior on the server


MoustaphaDev:
On the server `fetch` needs to be passed the full url of the resource to retrieve

Wrong way
```astro
---
const res = await fetch("/api") // this will raise an error
---
```

Right way
```astro
---
const res = await fetch("http://localhost:3000/api")
---
```

nienow:
The following works and could be added as an example:
```
---
const res = await fetch(Astro.url + "/api");
---
```

sasoria:
This was solved in #3677 so I think it can be closed.

delucis:
Thanks for addressing this @sasoria!

PR #3677, opened 2023-07-08
Document url constraint in data fetching

sasoria:

- New or updated content

#### Description

- Part of #3251

This Pull request documents a constraint on the URL provided to `fectch()` in the section about Data fetching. I'm leaving #3251 open, since there are new hints there that could be included in the docs, if we want that.


sarah11918:
Thank you for snagging this issue, @sasoria !  :raised_hands: 

Whenever possible, we do like to work key details into existing text, vs adding more note-type things. So I looked at the intro as a whole, and I'm wondering about a solution something like:

> All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs using the full URL  (e.g. `https://example.com/api` or `Astro.url + "/api"`) . 
>
> This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](https://deploy-preview-3677--astro-docs-2.netlify.app/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at runtime.

What do you think of something in this style?

sasoria:
> What do you think of something in this style?

I love it, thank you for the suggestion!

src/content/docs/en/guides/data-fetching.mdx
```diff
@@ -8,7 +8,7 @@ i18nReady: true
 
 ## `fetch()` in Astro
 
-All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs. This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at runtime.
+All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs using the full URL (e.g. https://example.com/api or Astro.url + "/api"). This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at runtime.
```
ElianCodes, line 11: I think we should add code ticks here, to make it a little more obvious.

```suggestion
All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs using the full URL (e.g. https://example.com/api or `Astro.url + "/api"`). This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at runtime.
```

sasoria: Yes, thank you for bringing this up 🙌

src/content/docs/en/guides/data-fetching.mdx
```diff
@@ -8,7 +8,7 @@ i18nReady: true
 
 ## `fetch()` in Astro
 
-All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs. This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at runtime.
+All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs using the full URL (e.g. https://example.com/api or `Astro.url + "/api"`). This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at runtime.
```
sarah11918, line 11:
```suggestion
All [Astro components](/en/core-concepts/astro-components/) have access to the [global `fetch()` function](https://developer.mozilla.org/en-US/docs/Web/API/fetch) in their component script to make HTTP requests to APIs using the full URL (e.g. https://example.com/api or `Astro.url + "/api"`).

This fetch call will be executed at build time, and the data will be available to the component template for generating dynamic HTML. If [SSR](/en/guides/server-side-rendering/) mode is enabled, any fetch calls will be executed at runtime.
```
Looking at the preview, the paragraph was just getting a little long. So I split it up!

sarah11918 (approved):
I'm not sure why this one sat here so long :sweat_smile: All those CMS guides distracting us!

Thanks so much for this, and for your patience, @sasoria !  :raised_hands:

PR #3697, opened 2023-07-11
Clean up package.json

MrHBS:
I believe these fields are not needed for private projects so I removed them.


delucis (approved):
Thanks! Some of these were actually required, but we can remove `engines`.

PR #3700, opened 2023-07-11
Finally add retry logic to Version component

Yan-Thomas:

#### Description

After all those years in this vital industry, I finally added retry logic to the Version component. No more build errors, the world is saved!

Note: I noticed some layout shifting while opening the Integrations pages in dev mode, but building locally and previewing made this disappear, so I guess it's fine.

### DIFF
--- a/src/components/Version.astro
+++ b/src/components/Version.astro
@@ -1,4 +1,5 @@
 ---
+import pRetry, { AbortError } from 'p-retry';
 import { cachedFetch } from '../util-server';
 
 export interface Props {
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM @Yan-Thomas — thanks for finally adding this! Feels like today has been a long line of finallys 😄 

I don’t _think_ this could have introduced layout shift at all — wonder if it’s something else causing that 🤔

Yan-Thomas:
> I don’t think this could have introduced layout shift at all — wonder if it’s something else causing that 🤔

I guess we had a similar issue before with a script being tree-shaken incorrectly and that caused a layout shift? Or is it just my machine, who knows? It didn't happen in Netlify Preview so I think we're fine 🙌

PR #3701, opened 2023-07-12
chore: reduce characters to run commands

ematipico:

#### Description

`yarn` and `pnpm` allow to run arbitrary commands without the use `run`. I thought it could be nice to have this shortcuts:
- less characters to type
- we teach our users something cool

Feel free to reject the PR if you think that was intentional


Yan-Thomas (commented):
LGTM, I've tried a few of these using `pnpm` and it seems to work just fine. I know we had a couple of issues and back-and-forth about which commands to document, so I think it's worth checking if Sarah or Chris have any thoughts about this 🙌

delucis:
I know we've discussed always using `npm install` over `npm i` so that the command is more easily read but don't think this has ever come up.

There _is_ something to be said for consistency between what we document for the different package managers — and `run X` does help clarify you're running a package script — but I don't know what the right answer is! I personally do always use the short form without writing `run`.

ElianCodes:
I'm not sure about this. Personally, I always use the short form. `npm i -D` or `pnpm dev` ect. But I do think we should generalise and make a decision and apply that everywhere.

I do agree with @delucis that using `pnpm run ...` clarifies that we are running a script.

ematipico:
We can still prepare a section (getting started?) where we say "In yarn and pnpm you can omit `run` if you run a script".

I am going to be honest, once I found that it's possible to omit `run` on those packages, it was mind-blowing. I mean the shortcuts are there to make our lives easier, should we tell our users that? It's still a learning opportunity

delucis:
Thanks again for this @ematipico! We chatted a bit and felt that it's not really our role to document pnpm/yarn features, and there's an advantage to keeping things explicit with the `run` command because for less experienced users they can more easily understand that these are package scripts rather than built-in commands of the package manager.

PR #3715, opened 2023-07-14
Remove Unused Import "h" from Tutorial 6-1

nhtsai:
I found an unused import in the tutorial code in 6-1 that looks like could be removed. Thank you!

#### Description

- What does this PR change?
Removes unused import from tutorial code 6-1.


TheOtterlord (commented):
`h` looks to be a required import when you're using JSX in Preact (according to this [Stack Overflow answer](https://stackoverflow.com/a/53099664/14104186) and the [Preact docs](https://preactjs.com/guide/v8/getting-started/#import-what-you-need))

PR #3736, opened 2023-07-17
Improve tutorial 4-3

morinokami:

- New or updated content

#### Description

- Updated the explanation at the beginning since it is slightly out of sync with the content of Tutorial 4-2.
- I thought it was a little difficult to understand what "achieve that" in line 28 meant, so replaced that with a more concrete description of what we want to do next.

### DIFF
--- a/src/content/docs/en/tutorial/4-layouts/3.mdx
+++ b/src/content/docs/en/tutorial/4-layouts/3.mdx
@@ -15,7 +15,7 @@ import PreCheck from '~/components/tutorial/PreCheck.astro';
 
 
 
-Now that you have your first `.md` blog post written, let's make it look like the rest of the pages on your website!
+Now that you have added a layout to each blog post, let's make it look like the rest of the pages on your website!
 
 <PreCheck>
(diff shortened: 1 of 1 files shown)


src/content/docs/en/tutorial/4-layouts/3.mdx
```diff
@@ -15,7 +15,7 @@ import PreCheck from '~/components/tutorial/PreCheck.astro';
 
 
 
-Now that you have your first `.md` blog post written, let's make it look like the rest of the pages on your website!
+Now that you have added a layout to each blog post, let's make it look like the rest of the pages on your website!
```
sarah11918, line 18:
```suggestion
Now that you have added a layout to each blog post, let's make your posts look like the rest of the pages on your website!
```

sarah11918 (approved):
I like this change @morinokami , thank you!

Yes, the tutorial did have some topics in a slightly different order and we changed things around at the very last minute. This apparently was missed. Thanks for making it make better sense!  :raised_hands:

PR #3752, opened 2023-07-18
draft edits

sarah11918:
Just for @matthewp !  :smile:

### DIFF
--- a/src/content/docs/en/guides/client-side-routing.mdx
+++ b/src/content/docs/en/guides/client-side-routing.mdx
@@ -84,7 +84,7 @@ import { ViewTransitions } from 'astro:transitions';
 
 ## Transition Directives
 
-Use `transition: *` directives on page elements in your `.astro` components to control the transition animation for individual elements on your page during navigation.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/client-side-routing.mdx
```diff
@@ -58,11 +82,22 @@ import { ViewTransitions } from 'astro:transitions';
 <ViewTransitions />
 ```
 
-Just by doing that, now the next page that you navigate to will happen without full-page navigation. Next, you'll likely want to enable animations between common parts of your page.
+## Transition Directives
+
+Use `transition: *` directives on page elements in your `.astro` components to control the transition animation for individual elements on your page during navigation.
```
matthewp, line 87:
```suggestion
Use `transition:*` directives on page elements in your `.astro` components to control the transition animation for individual elements on your page during navigation.
```

the space made it seem like a CSS prop to me.

matthewp:
Just 1 very minor thing, like the reorg, flow makes sense now!

PR #3764, opened 2023-07-19
Remove abort when `<Version />` fetch fails

Yan-Thomas:

#### Description

Turns out adding retrying logic didn't really fix this for us, now I changed the `AbortError` to a normal `Error` so that when it fails it will actually retry. And with +5 retries, to be sure.

From what I understood from Eleventy Fetch docs, it won't cache a failed response, so I guess we're fine in this aspect!


delucis (approved):
LGTM — thanks Yan! I don't _think_ we should need the extra retries but they don't hurt.

PR #3770, opened 2023-07-20
Use plain text instead of `<StackblitzIntro>`

Yan-Thomas:

#### Description

This PR removes the `<StackblitzIntro>` component to use plain text instead, allowing it to be translated. I've kept the component for now so the current translations don't break, but we're free to remove once there's no translations using it.


delucis (commented):
This looks good to me — thanks for following through on this!

What do you think about doing the same as you’ve done here also for the Spanish tutorial so we can rip out the component entirely? We can ping the i18n crew for a quick update of those pages then I’m sure.

Yan-Thomas:
> What do you think about doing the same as you’ve done here also for the Spanish tutorial so we can rip out the component entirely? We can ping the i18n crew for a quick update of those pages then I’m sure.

Right, I think we can do this. The problem I was thinking was that updating branch would eventually cause a error for the Russian and Japanese PRs that use it, but I guess we can also change these in their respective PRs.

delucis:
> updating branch would eventually cause a error for the Russian and Japanese PRs that use it

I also thought that, but given we _want_ them to replace it with translated text, I actually think breaking there is a good thing: it’s the sign some action is needed.

Yan-Thomas:
> I also thought that, but given we want them to replace it with translated text, I actually think breaking there is a good thing: it’s the sign some action is needed.

Oh right, that's fair. Guess we're good to merge then!

PR #3771, opened 2023-07-20
swap client-side routing with view transitions

sarah11918:
Renames the main page and tweaks some wording throughout to hit "View Transitions" harder than "Client-side routing" for 2.9.

This also updates the config reference locally, which I'll also do at the source in a separate PR to match.


delucis (approved):
Looks great! Excellent quick work @sarah11918!

matthewp (approved):

PR #3806, opened 2023-07-24
Heading level that doesn't correspond

luoingly:

- Minor content fixes

#### Description

Compare to the next heading in the same page:

https://github.com/withastro/docs/blob/d5d197eadf58bba837de5a7ccba0381ccb4155e6/src/content/docs/en/tutorial/1-setup/index.mdx#L66

And the same heading in another file:

https://github.com/withastro/docs/blob/d5d197eadf58bba837de5a7ccba0381ccb4155e6/src/content/docs/en/tutorial/2-pages/index.mdx#L24

So here should also be `##` instead of `###`.

https://github.com/withastro/docs/blob/d5d197eadf58bba837de5a7ccba0381ccb4155e6/src/content/docs/en/tutorial/1-setup/index.mdx#L55


sarah11918 (approved):
Great spotting, @luoingly !  :rocket:

PR #3810, opened 2023-07-24
Document caveats with using ViewTransitions component

matthewp:

- New or updated content

#### Description

- Part of https://github.com/withastro/astro/pull/7786
- Documents how scripts run when using the `<ViewTransitions />` component.


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -217,3 +217,34 @@ import { ViewTransitions } from 'astro:transitions';
 :::note[known limitations]
 The `morph` animation cannot be simulated in traditional CSS. So any element using this animation will not be animated.
 :::
+
+## Caveats
+
+When using the `<ViewTransitions />` component page navigations do not behave exactly the same as they do when navigating directly to a page (such as typing the URL into the address bar).
```
sarah11918, line 223:
```suggestion
`<ViewTransitions />` is a routing component that affects how scripts are executed during page navigation.
```
Related to comment above, would something like this be an accurate introduction to the points below?

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -217,3 +217,34 @@ import { ViewTransitions } from 'astro:transitions';
...
+
+### Scripts
+
+When navigating between pages with the `<ViewTransitions />` component scripts are run in sequential order to match browser behavior. If you have code that sets up global state, this state will need to take into account that the script might execute more than once. You can usually do this by checking for the global state and conditionally executing your code.
```
sarah11918, line 227:
```suggestion
When navigating between pages with the `<ViewTransitions />` component, scripts are run in sequential order to match browser behavior. 

If you have code that sets up global state, this state will need to take into account that the script might execute more than once. Check for the global state in your `<script>` tag, and conditionally execute your code where possible:
```
Splitting up the headline to know vs. the solution.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -217,3 +217,34 @@ import { ViewTransitions } from 'astro:transitions';
...
+
+```astro
+<script is:inline>
+  if(!window.SomeGlobal) {
+    window.SomeGlobal = {} // ....
+  }
+</script>
+```
+
+Module scripts are only ever executed once, because the browser keeps track of which modules are already loaded. For these scripts you do not need to worry about re-execution.
```
sarah11918, line 237:
```suggestion
Module scripts are only ever executed once because the browser keeps track of which modules are already loaded. For these scripts, you do not need to worry about re-execution.
```
Commas before Yan gets here!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -217,3 +217,34 @@ import { ViewTransitions } from 'astro:transitions';
...
+
+#### astro:load
+
+Additionally, Astro triggers the `astro:load` event on the `document` when a new page loads. This happens both when the initial page loads and for any subsequent navigation. You can use this event to run code on every page navigation, or only once ever.
```
sarah11918, line 241:
```suggestion
An event listener to/that.... FILL THIS IN MATTHEW

When using the `<ViewTransitions />` component, Astro triggers the `astro:load` event on the `document` when a new page loads. This happens both when the initial page loads and for any subsequent navigation. 

Use this event to run code on every page navigation, or only once ever:
```
Since `astro:load` is new and undocumented, let's give it at least an intro explanatory sentence here. FILL THAT IN, MATTHEW!  :smile:

Also, just splitting up ideas to make the text more scannable.

sarah11918:
Also checking, @matthewp, is this waiting on a new release? 2.9.2 went out this morning, but your implementation PR is still open. Is this for 2.9.3?

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -217,3 +217,34 @@ import { ViewTransitions } from 'astro:transitions';
 :::note[known limitations]
 The `morph` animation cannot be simulated in traditional CSS. So any element using this animation will not be animated.
 :::
+
+## Caveats
```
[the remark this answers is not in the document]
matthewp: Yeah, after talking on Discord I think this is more about scripts and not general caveats.

matthewp:
@sarah11918 yeah, whichever the next patch release is. Hopefully 2.9.3

src/content/docs/en/guides/view-transitions.mdx
matthewp: Wasn't the sentence I had introductory?

matthewp: I'll try another pass at it.

sarah11918 (approved):
Looks great! I really like these changes. I'll keep an eye out for the next patch release to know when this should be merged.  :raised_hands:

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -217,3 +217,34 @@ import { ViewTransitions } from 'astro:transitions';
...
+<script is:inline>
+  if(!window.SomeGlobal) {
+    window.SomeGlobal = {} // ....
+  }
+</script>
+```
+
+Module scripts are only ever executed once because the browser keeps track of which modules are already loaded. For these scripts, you do not need to worry about re-execution.
+
+#### astro:load
```
sarah11918, line 237:
```suggestion
#### `astro:load`
```
inline code formatting, so that it shows up properly on the docs page

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -217,3 +217,34 @@ import { ViewTransitions } from 'astro:transitions';
...
+
+An event that fires on page navigation. You can listen to this event on the `document`. The `<ViewTransitions />` component fires this event both on initial page navigation (MPA) and any subsequent navigation, either forwards or backwards.
```
sarah11918, line 239:
```suggestion
An event that fires on page navigation. You can listen to this event on the `document`.

The `<ViewTransitions />` component fires this event both on initial page navigation (MPA) and any subsequent navigation, either forwards or backwards.
```
smaller paragraphs

Yan-Thomas (approved):
LGTM!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -203,6 +203,38 @@ import { ViewTransitions } from 'astro:transitions';
...
+
+#### `astro:load`
+
+An event that fires on page navigation. You can listen to this event on the `document`.
+
+The `<ViewTransitions />` component fires this event both on initial page navigation (MPA) and any subsequent navigation, either forwards or backwards.
+
+You can use this event to run code on every page navigation, or only once ever:
+
+```astro
```
delucis, line 230: This addition looks great!

One small suggestion to highlight the `once` argument in this code block to reinforce that it’s there. Given this is the only example we give of using `astro:load` I’m a little concerned some folks may just blindly copy without noticing the `once`:

````suggestion
```astro "{ once: true }"
````

PR #3813, opened 2023-07-25
Clarified Astro.glob() docs

putnikproj:

- New or updated content

#### Description

- What does this PR change? Adds a clarification about `import.meta.glob()` api, as it is a bit misleading now (I spent a lot of time figuring out what was wrong myself). There's even a GitHub issue describing this change: https://github.com/withastro/astro/issues/6328#issuecomment-1485008866.
- I even think that it is worth pointing out that Astro has a different api for loading local files into static site setup (I mean, Content Collections). If it is so, please, tell me how to mention it correctly.


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -39,7 +39,7 @@ const posts = await Astro.glob('../pages/post/*.md'); // returns an array of pos
 :::note
 You can also use `import.meta.glob()` itself in your Astro project. You may want to do this when:
 - You need this feature in a file that isn't `.astro`, like an API route. `Astro.glob()` is only available in `.astro` files, while `import.meta.glob()` is available anywhere in the project.
-- You don't want to load each file immediately. `import.meta.glob()` can return functions that import the file content, rather than returning the content itself.
+- You don't want to load each file immediately. `import.meta.glob()` can return functions that import the file content, rather than returning the content itself. (Regardless of using imported files on the page, all styles and scripts used in that files will still be bundled and added to the page, as this is decided by static analysis, not at runtime)
```
sarah11918, line 42:
```suggestion
- You don't want to load each file immediately. `import.meta.glob()` can return functions that import the file content, rather than returning the content itself. Note that this import includes all styles and scripts for any imported files. These will be bundled and added to the page whether or not a file is actually used, as this is decided by static analysis, not at runtime.
```
Just tried to do two things here:
- lead with the headline (styles and scripts always imported/bundled!) then with the conditional (even if...)
- Break this up into two sentences because it was a lot of thoughts to have in one.

What do you think? Are we quite there yet?

sarah11918 (commented):
Thanks for your addition here, @putnikproj ! I know this is something we do tell people when answering support questions, so makes perfect sense to add it to the docs.

See my suggestion below for changing the wording slightly to fit more with the overall style of our docs!  :raised_hands:

sarah11918:
Excellent, then let's do this thing!!  Welcome to Team Docs!  :partying_face:

PR #3814, opened 2023-07-25
Nudge towards View Transitions defaults

sarah11918:
This slightly tweaks and reorganizes the new View Transitions page based on early feedback:

- emphasizes the "happy path" - SPA mode and defaults!
- describes naming transitions before the animations (since that was found to be more useful)
- makes it clearer that you do not need any transition directives, so that maybe people will start without jumping into using them.
- removes the "on a single page" section as that will be less common, and reduces it to a mention

cc @Yan-Thomas for review based on his experience!


Yan-Thomas (approved):
LGTM! Thanks @sarah11918 🙌

PR #3828, opened 2023-07-25
Update WordPress.mdx to describe how to get smaller image sizes

rickgregory:

 
- New or updated content
 
#### Description
_Added a couple of sentences to the section on embeddable content with information on how get smaller sizes of the featured image (the code that is there gets the full size image which is often too large)._

### DIFF
--- a/src/content/docs/en/guides/cms/wordpress.mdx
+++ b/src/content/docs/en/guides/cms/wordpress.mdx
@@ -157,7 +157,11 @@ let [post] = await res.json();
 
 The `_embedded['wp:featuredmedia']['0'].source_url` property is returned, and can be used to display the featured image on each dinosuar page. 
 
-As written, this will return the full sized image which may be too large for your use. If you need one of the smaller sizes created by WordPress, you can retrieve it like this: `_embedded['wp:featuredmedia']['0'].media_details.sizes.[SIZE].source_url` where [SIZE] is one of the image sizes created by WordPress. For example, `_embedded['wp:featuredmedia']['0'].media_details.sizes.medium.source_url` will return the medium image size.
+As written, this will return the full sized image which may be too large for your use. If you need one of the smaller sizes created by WordPress, you can retrieve it like this:
+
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @rickgregory , thanks for this contribution! I think this is a super helpful addition, but I have another idea for how to get that across:

What if we just update the code line itself to include the medium size, and then slip in a mention of that in the text:

> The `_embedded['wp:featuredmedia']['0'].media_details.sizes.medium.source_url` property is returned, and can be used to display the featured image on each dinosuar page.

This reveals the pattern without needing to explicitly explain it. I think just having it in the URL is enough to signal to people, "Oh, I guess I could pick a different size if I wanted" and then they can read in Wordpress documentation about their sizes. We try to keep these guides as minimal on explanation as possible, and I like "showing" vs telling.

What do you think about that?

rickgregory:
@sarah11918 - that's a good idea. I just tested to make sure that syntax allows the full image size and it do, so there's no downside to using this syntax. However, while I like show vs tell, I think noting that the size name is there is important. Otherwise we're asking the reader to infer that .medium.source_url is only one of several possible sizes. A quick "Replace .medium. with your desired image size" solves that and doesn't feel too verbose to me.  So...

> The _embedded['wp:featuredmedia']['0'].media_details.sizes.medium.source_url property is returned, and can be used to display the featured image on each dinosuar page. Replace _.medium._ with your desired image size.

sarah11918:
@rickgregory Sure thing! For syntax, I would use:

```
(Replace `medium` with your desired image size.)
```

But otherwise, go ahead and make that change!  :raised_hands:

rickgregory:
All done, let me know if you need other edits.

sarah11918:
Ooh, just a reminder to also update line 163 (GitHub won't let me suggest a change here) so that the text matches the code sample, then I think we're good! Thank you!

rickgregory:
Doh. Done :)

sarah11918 (approved):
@rickgregory , this looks great now! Thank you for taking the time to contribute, and welcome to Team Docs!  :partying_face:

PR #3829, opened 2023-07-26
chore(en): describe more accepted `class:list` value kinds

duanwilliam:

- (very) minor content fixes (broken links, typos, etc.)

#### Description

Currently, the documentation on `class:list` doesn't mention anything about how it handles false values.
This was an issue when it stringified `false`, diverging in behavior from the `clsx` package the documentation references (https://github.com/withastro/astro/issues/3920);
while that since has since been fixed (https://github.com/withastro/astro/pull/3922), there's still no mention of how it handles such values in the documentation. This PR adds a note of how it handles `false` (as well as `null`) in the docs, as observed in the source code at https://github.com/withastro/astro/blob/main/packages/astro/src/runtime/server/util.ts#L19.

### DIFF
--- a/src/content/docs/en/reference/directives-reference.mdx
+++ b/src/content/docs/en/reference/directives-reference.mdx
@@ -31,6 +31,7 @@ A template directive is never included directly in the final HTML output of a co
 - `Object`: All truthy keys are added to the element `class`
 - `Array`: flattened
 - `Set`: flattened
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Love this, @duanwilliam! Thanks so much for this addition, and welcome to Team Docs! :partying_face:

PR #3837, opened 2023-07-26
Document the astro:beforeload event

matthewp:

- New or updated content

#### Description

- Documents the new `astro:beforeload` event


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,6 +235,34 @@ You can use this event to run code on every page navigation, or only once ever:
   }, { once: true });
 </script>
 ```
+
```
delucis, line 238: Given that we now have `astro:load` *and* `astro:beforeload`, we probably do need a line or two about the difference, rather than just relying on people reading the description of each and working it out.

I’d suggest an `h3` before `astro:load`, e.g. `### Useful events` or something with one or two sentences describing the difference.

May also be good to update the first sentence under `astro:load`? That says “An event that fires on page navigation” but how is that different from the `beforeload` “after the new page replaces the old page”?

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,6 +235,34 @@ You can use this event to run code on every page navigation, or only once ever:
...
+#### `astro:beforeload`
+
+An event that fires immediately after the new page replaces the old page. You can listen to this event on the `document`.
+
+This event is useful to restore any state on the DOM that needs to transfer over to the new page. If you are implementing dark mode support, this event can be used to restore state onto the new page like so:
```
delucis, line 243: Breaking up the generic “this is useful” and the specific example across paragraphs:

```suggestion
This event is useful to restore any state on the DOM that needs to transfer over to the new page.

For example, if you are implementing dark mode support, this event can be used to restore state across page loads:
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,6 +235,34 @@ You can use this event to run code on every page navigation, or only once ever:
...
+	toggle();
+  // ViewTransitions nav
+	document.addEventListener('astro:beforeload', toggle);
+</script>
+<style is:global>
+	html[data-dark] body {
+		background: black;
+		color: white;
+	}
+</style>
```
delucis, line 263: I think we can do some renaming here to clarify what’s happening, plus drop the `<style>` block which isn’t crucial to the example:

```suggestion
<script>
	const setDarkMode = () => {
		if (localStorage.darkMode) {
			document.documentElement.dataset.dark = '';
		}
	};

  // Runs on initial navigation
	setDarkMode();
  // Runs on view transitions navigation
	document.addEventListener('astro:beforeload', setDarkMode);
</script>
```

delucis (commented):
Thanks @matthewp! Making a couple of suggestions.

src/content/docs/en/guides/view-transitions.mdx
matthewp: Hm, is that common? It seems like it's just going to be a repeat of the sentences in each of the event descriptions. 

> May also be good to update the first sentence under astro:load?

yes, this one can be more specific.

matthewp: Updated the astro:load.

sarah11918 (approved):
Thanks, Matthew! I think the difference is clear, now that we're documenting both!

PR #3874, opened 2023-07-27
Update labels & remove default titles

### LINKED ISSUE #3872: Update issue template to not use default titles
### 📋 Explain your issue

Our issue templates currently pre-fill the `title` field with text, starting with a square bracket-wrapped type string, e.g. `[CONTENT]`, `[OTHER]` etc.

We should switch to leaving `title` _blank_ by default, to force issue openers to fill it in. And use the issue template’s _label_ config to apply a tag like “content”, “other” etc.

### LINKED ISSUE #3873: Remove `i18n` label from `CONTENT` issue template
### 📋 Explain your issue

Our CONTENT issue template currently automatically applies the `i18n` label to all issues created using it. We should remove this.

TheOtterlord:

- Something else!

#### Description

- Closes #3872 
- Closes #3873 
- What does this PR change? Updates the issue templates to favour labels over title prefixes


sarah11918 (approved):
Conditionally approving, except that I  just want to confirm that by *deleting* the title line entirely, it simply leaves title blank? Just checking we don't have to put an empty string or anything for a required field.

TheOtterlord:
Yep, I believe so. I couldn't find any documentation saying it's either required or optional, but as reference, the [bug report in withastro/docs](https://github.com/withastro/astro/blob/main/.github/ISSUE_TEMPLATE/---01-bug-report.yml?plain=1) does not use the `title` field.

PR #3884, opened 2023-07-27
Add note about pnpm and using vite

### LINKED ISSUE #3191: Environment variables
In the docs is mentioned that you can load environment variables with a helper function from vite package. This is the [link](https://docs.astro.build/en/guides/configuring-astro/#environment-variables) to the part where the docs explain how Astro evaluates config files.

This should be placed in the `astro.config.mjs`

```js 
import { loadEnv } from "vite";
const { SECRET_PASSWORD } = loadEnv(import.meta.env.MODE, process.cwd(), "");
```

But for some reason, when you import the helper funcion you get an error that can't find the `vite` package.

What I was trying to accomplish was to load env vars directly in the `astro.config.mjs` file but apparently it won't work.

TheOtterlord:

- New or updated content

#### Description

- Closes #3191
- What does this PR change? Adds a note about pnpm, and importing modules that are not installed in the project.


Yan-Thomas (approved):
Merging this so we don't miss it in the mergeaddon!

PR #3929, opened 2023-07-28
Fix union type check

Yan-Thomas:

#### Description

Fix to an `astro check` issue because we now have more pages than TypeScript can handle 💅 
image


TheOtterlord (approved):
LGTM!

PR #3936, opened 2023-07-29
Move filenames to the code snippet title

Alynva:

- New or updated content

#### Description

  - Moves the filenames in code snippets from the single-line comment to the `title` attribute of the code snippet. I've used this regex to find the ocorrences `/```(.*)(?! title=")(.*?)(\n---)?\n *// *(.*\..*)/g` and this one to replace them `/```$1$2 title="$4"$3/g`, which might not be the best one, but it helped.
  - The snippet title is currently automatically detected by [`astro-expressive-code`](https://www.npmjs.com/package/astro-expressive-code) with the following implementation, but I thought that might be good to padronize the docs and start to use only the `title` attribute https://github.com/expressive-code/expressive-code/blob/709b2929bcaea355b02b359c476b38570dbc70df/packages/%40expressive-code/plugin-frames/src/utils.ts#L87
  - **All files** in this commit 46b41414d5b4bce4522d1a06716b6d1dd18c88c4 **have** line highlighting (diff) that ~SHOLD BE FIXED BEFORE MERGING~ (fixed by 4e560e5416e685707956b7261f12ac413e537344).
  - **Some files** in this commit 97d16a51ae49d86e53d6fd6293ed0a9851092efc **have** line highlighting (diff) that ~SHOLD BE FIXED BEFORE MERGING~ (fixed by 4e560e5416e685707956b7261f12ac413e537344).
  - Some ocorrences might be discussable that I didn't replaced *yet*. I replaced exclusively when the title would be only the filename. In some places like this that the comment starts with "Example: " before the filename, **I did not** replaced, and any other that contains normal text in the comment. https://github.com/withastro/docs/blob/a2c1bc35197875d9f381080d55a7fd92d5cb39b9/src/content/docs/en/core-concepts/astro-pages.mdx#L34-L46
  - In some cases we could also move normal text from the comment to the title because sometimes it is the *title* of the code snippet. But that might be another discussion.


Alynva:
Marking as WIP, but I'll wait for feedback before fixing conflicts as the main branch is constantly being updated, wich might affect this PR multiple times...

delucis:
Thanks for the PR @Alynva! Curious to hear feedback from others about what they think about standardising to this approach. Personally, I generally prefer using the comment syntax over `title` when writing examples as these display when rendered with a regular Markdown renderer, but curious what other people think.

sarah11918:
So, this is not a *great* reason, but I personally tend to prefer our `title=""` syntax just because it's easier to scan and know that we have included a title, and it just *feels* more like what the code sample will look like, with the title separated from the code.

BUT, I acknowledge that the other format also works, and works in more places (like on GitHub in the READMEs).

Since this site is changing so much all the time, what do we think of maybe either: standardizing this for the Starlight conversion, or not trying to change ALL the things at once, but target specific sections of the site in smaller PRs. 

There are already a bunch of conflicts here, and I feel like we might never have a window where a huge PR like this can sneak in and easily merge.  😅

sarah11918:
Hi @Alynva , so we've been discussing with the DX team and we're not sure that standardizing is a task to undertake right now... especially when each of us seem to have a different preferred syntax.  :joy: 

There are good use cases for either syntax, and in some cases our hand is tied by choosing one over the other. So, we've decided that especially with other pressing decisions and work that needs to be done, having to make a decision and then make everything *just right* for a PR that touches almost every single page is a little more than we want to take on right now!  :sweat_smile: 

So we're going to close this one, and also a quick reminder that this might have been better filed in Issues before making you go through all this work! That's usually where we triage and discuss whether ideas are something we're willing and able to take on. A PR is more like throwing something in the inbox, and I'd say this fits in the category of not an obvious no brainer to do, so requires discussion first. (As evidenced by we... chose not to do it!)  :smile:

Appreciate you taking the time and thinking about the underlying code of the docs. You've been a super active contributor lately, and we love it, and want you to keep going!  :rocket:

Alynva:
It's not a problem, it didn't take me much time, it was just some find&replace regex. I'm glad to be able to contribute, and I'll use the issues a little more to keep things organized. You're awesome ❤️

PR #3941, opened 2023-07-29
Update docs theme link to Starlight

Yan-Thomas:

#### Description

I couldn't find the docs theme in Astro anymore, and I also believe we should mention Starlight rather than the old docs theme anyway.


sarah11918:
Great call, Yan! I know there are more of these in the "Migrate from" pages though, not just Vue Press. Off the top of my head, Docusaurus, Gitbook... anything that claims to be "docsy" will have the suggestion to use the docs template.

Would you mind finding all of those and updating to Starlight from mention of "generic docs template" and also, check the code sample because we might have the package manager tabs for starting with the docs template commands, and those should now be Starlight!  :star2:

delucis (commented):
Confirming that the old theme is also linked from the Docusaurus migration guide which should be updated.

That’s the only other one I could find though. Thanks for spotting this @Yan-Thomas!

delucis (commented):
Only now reading that after the link it says "this website was built with that template" 😅

I guess this statement will become true again, so OK to leave in and be briefly inaccurate?

src/content/docs/en/guides/migrate-to-astro/from-docusaurus.mdx
```diff
@@ -35,7 +35,7 @@ When you rebuild your Docusaurus site in Astro, you will notice some important d
...
 
-- Docusaurus was designed to build documentation websites and has some built-in, documentation-specific website features that you would have to build yourself in Astro. Instead, Astro offers some of these features through an [official docs theme](https://github.com/withastro/astro/tree/latest/examples/docs). This website was built using that template! You can also find more [community docs themes](https://astro.build/themes?search=&categories%5B%5D=docs) with built-in features in our Themes Showcase.
+- Docusaurus was designed to build documentation websites and has some built-in, documentation-specific website features that you would have to build yourself in Astro. Instead, Astro offers some of these features through an [official docs theme](https://starlight.astro.build). This website was built using that template! You can also find more [community docs themes](https://astro.build/themes?search=&categories%5B%5D=docs) with built-in features in our Themes Showcase.
```
sarah11918, line 38:
```suggestion
- Docusaurus was designed to build documentation websites and has some built-in, documentation-specific website features that you would have to build yourself in Astro. Instead, Astro offers some of these features through an [official docs theme](https://starlight.astro.build). This website was the inspiration for that template! You can also find more [community docs themes](https://astro.build/themes?search=&categories%5B%5D=docs) with built-in features in our Themes Showcase.
```
Gah! Depends ... how soon? I don't like the idea of saying it was built with "that template" when it wasn't! What about...?

PR #3945, opened 2023-07-29
WIP: Test Starlight migration basics

delucis:

#### Description

- Installs the Starlight integration and adds basic config for it.
- Opening this PR to see what happens on Netlify/GH Actions!

#### To-do

Collecting tasks that need doing as I come across them.

- [ ] Schema extensions. Starlight doesn’t currently make it simple to extend its docs schema. Added a quick package patch to this branch to make builds work, but need a built-in solution in Starlight proper to simplify complex frontmatter schema for folks.

- [ ] Sitemap exclusions. We currently [exclude redirect pages when setting up the sitemap integration](https://github.com/withastro/docs/blob/f8a04f15d7418c13b6b006e56c277086c6b87442/integrations/sitemap.ts#L8-L19). Starlight configures the integration for us, but no way to pass in a filter.

- [ ]  Heading links. Starlight doesn’t currently include a clickable link icon for each heading, like we have in docs. Should look to add it upstream. (Design input required?)


delucis:
Closing in favour of #4266

PR #3950, opened 2023-07-30
Changes endpoints.mdx

hvlxh:
It will not work because assigning function in variable should be an callback `() => {}` not but using function keyword `function example()`


hvlxh:
We are also able to do like this:
```ts
export async function get({ params, request }): APIRoute {
...
```

Alynva:
WDYM? That does work, look this

image

TheOtterlord:
Thanks for the PR!
The existing pattern is valid syntax known as a [function expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/function), and will work fine in an endpoint. Looking at the rest of the file, we are mixing this and arrow function syntax, so this might be an opportunity to make our examples on this page more consistent.

sarah11918 (approved):
Thanks for this, @hvlxh! As others have pointed out, it brings a consistency to the page that can only be helpful for readers. Welcome to Team Docs! :partying_face:

PR #3963, opened 2023-07-31
i18n(ar): Updates `ar/ui.ts`

g9wz:
#### What?

Rewrote `ar/ui.ts` from the ground up

#### Why?

- It Had a considerable amount of typos
- Some elements were missing
  - I assume because it was outdated 

#### Notes:
- I purposely left some untranslated vocabulary elements & marked them with a multi-line comment specifying when they should be translated (along which file/s).
  - They only show up in specific - untranslated - pages, translating one without the other will mix different languages / directionalities in the same page.


Yan-Thomas (approved):
LGTM! Thanks @g9wz 💜

PR #3966, opened 2023-07-31
🐛 fix(#3313): adds ResponseInit shape

### LINKED ISSUE #3313: Consider updating ResponseInit link
# What? 

The ResponseInit link in https://docs.astro.build/en/reference/api-reference/#astroresponse is dead. 

CleanShot 2023-05-22 at 17 33 02@2x

It points to MDN, but the hash link no longer exists. 

## Potential solution 
We could potentially point to the actual spec https://fetch.spec.whatwg.org/#responseinit

ohansemmanuel:

- New or updated content

#### Description

- Closes #3313 
- As recommended in #3313,  this PR adds the `ResponseInit` object shape. 

CleanShot 2023-08-07 at 16 38 03@2x


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -176,7 +176,19 @@ With the default `output: 'static'` option, `Astro.request.url` does not contain
...
+
+`Astro.response` is a standard `ResponseInit` object. It has the following structure. 
+
+| Name         | Type                                            | Description                                                  |
+| ------------ | ----------------------------------------------- | ------------------------------------------------------------ |
+| `status`     | `number \| undefined`                            | The status code of the response, e.g., `200`.                   |
+| `statusText` | `string`                                        | The status message associated with the status code, e.g., `OK`. |
+| `headers`    | `Headers \| string[][] \| Record<string, string>` | The HTTP headers of the response.                  |
+
+
```
delucis, line 188: Awesome, thanks for picking this up @ohansemmanuel! What do you think of using a code snippet like this to show this information?

````suggestion
```ts
interface ResponseInit {
  /**  The status code of the response, e.g., `200`. */
  status: number | undefined;
  /** The status message associated with the status code, e.g., `OK`. */
  statusText: string;
  /** The HTTP headers of the response. */
  headers: Headers | string[][] | Record<string, string>;
}
```
````

We tend to avoid tables as they don‘t work super well on mobile:
image

Comment (delucis) [reply src/content/docs/en/reference/api-reference.mdx:188]: Alternatively, we could use a list? This version doesn’t include the types, but perhaps that’s ok.

> - `status`: The status code of the response, e.g., `200`.
> - `statusText`: The status message associated with the status code, e.g., `'OK'`.
> - `headers`: The HTTP headers of the response.

```suggestion
- `status`: The status code of the response, e.g., `200`.
- `statusText`: The status message associated with the status code, e.g., `'OK'`.
- `headers`: The HTTP headers of the response.
```

ohansemmanuel: That makes sense @delucis The interface snippet captures the full information here. I'll update to use that! 

Is it worth changing other bits to use the same form for consistency e.g., the [AstroCookies reference](https://docs.astro.build/en/reference/api-reference/#astrocookies), [Pagination prop](https://docs.astro.build/en/reference/api-reference/#the-pagination-page-prop) use a similar table

delucis: Oh good shout! Let's consult with @sarah11918 here.

There's a bit more info in that table, so wonder if actually a list would be the best choice when we face these situations. That's also the more accessible solution as assistive technologies can step through the list items in a way they can't with a code snippet.

@sarah11918 any thoughts?

delucis: Thanks for the feedback @sarah11918! The distinction of code block for example and text content for docs makes sense to me. @ohansemmanuel maybe we can get more type context into my suggestion — what do you think of this?

> - `status`: The numeric status code of the response, e.g., `200`.
> - `statusText`: The status message associated with the status code, e.g., `'OK'`.
> - `headers`: A [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance that you can use to set the HTTP headers of the response.

I double checked and `headers` in this instance should always have this shape, so you can do `headers.set()` etc.

```suggestion
> - `status`: The numeric status code of the response, e.g., `200`.
> - `statusText`: The status message associated with the status code, e.g., `'OK'`.
> - `headers`: A [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance that you can use to set the HTTP headers of the response.
```

sarah11918: Hi @ohansemmanuel , just checking in to see whether you'd had a chance to consider @delucis most recent comment! I think we all agree this this would be a a great solution to that link issue, and Chris and I feel pretty confident about text and not a table. 

Would you like to discuss this further, or are you happy with Chris' suggestion?

delucis (approved):
Thanks again for pushing this through @ohansemmanuel — appreciate it! 💜

PR #3969, opened 2023-07-31
Add translation about stub in layout

imbant:
- Translated content

#### Description

Add translation about stub. Please see [this issue](https://github.com/withastro/docs/issues/3970) about why I create this PR


Alynva:
Looks like the `getTranslation` throws an error when the key is not found in that language:

https://github.com/withastro/docs/blob/main/src/i18n/util.ts#L59-L70

So we might need to add all translations before mergin!

/cc @sarah11918 @Yan-Thomas @delucis

Alynva (approved):

imbant:
> ~Looks like the `getTranslation` throws an error when the key is not found in that language:~
> 
> ~https://github.com/withastro/docs/blob/main/src/i18n/util.ts#L59-L70~
> 
> ~So we might need to add all translations before mergin!~
> 
> ~/cc @sarah11918 @Yan-Thomas @delucis~
> 
> My bad, didn't saw the ` || translations[fallbackLang][key]` x.x LGTM!

Yep, I tested that when there is no translation for that language, it will fall back to English :)

imbant:
Sorry, I just broke my main branch by mistake. So I force-pushed it to fix it. Everything is the same as when Alynva approved these changes

sarah11918 (approved):
Nice effort, team! I will wait for Yan or Chris for final approval on this one, but really appreciate this thoughtful contribution, @imbant ! :partying_face:

sarah11918:
Hi @imbant, we had a team meeting about this one (and, my apologies, I forgot we'd already a long time ago had a discussion months ago!) and our translators at the time decided that it's better NOT to translate this content. In fact, it was my mistake to mark the stubs as `i18n:true` in the first place!

Most of our stub pages are really quite small, with little text to translate, so there's not lot of extra benefit from translating them. And then, if we receive an English contribution to make a more full guide, there would be a big difference between the English language page and a not-yet-updated translated page. We think it's better to have those pages fall-back to English rather than have a time where we display a stub for non-English pages when there is actually good content on the English page.

Keeping the stubs untranslated means that they will default to English, and we will only mark them as `i18n:true` when there *is* a full guide. That way, this problem never happens.

Thank you for starting these translations, even though we can't use them.  :sweat_smile: We look forward to more contributions from you!  :raised_hands:

sarah11918:
Oops, apparently I missed a meeting!

I think we're going to fix the issue a different way, and so we CAN use this!  So I'm going to stay out of this and let @Yan-Thomas finish this off. Reopening at this point, since we think we can use it!

imbant:
OK. Thanks for the detailed explanation. Just let me know when there is something that I need to change😀

Yan-Thomas (approved):
LGTM! Thanks @imbant 🙌

PR #3992, opened 2023-07-31
Add title and adjust highlight in configured redirects code samples

AishaBlake:

#### Description

- No issue but addresses [request from @sarah11918](https://github.com/withastro/docs/issues/2042#issuecomment-1658196170)
- Adds a title missing from the third code snippet in the [Configured Redirects](https://docs.astro.build/en/core-concepts/routing/#configured-redirects) section of the Routing page. In addition, I adjusted the highlights in the first and third code snippets so that they mark the code being described in the text. In the current version, the first code snippet highlights the final line and two lines that don't exist (7-9).

Before:
current code samples in configured redirects section linked above

After:
same code sample, now both showing config.astro.mjs as the title and with only the redirect configuration code highlighted

### DIFF
--- a/src/content/docs/en/core-concepts/routing.mdx
+++ b/src/content/docs/en/core-concepts/routing.mdx
@@ -231,7 +231,7 @@ You can define rules to [redirect users to permanently-moved pages](#configured-
 
 You can specify a mapping of permanent redirects in your Astro config with the `redirects` value. For most redirects, this is a mapping of an old route to the new route:
 
-```js title="astro.config.mjs" {5, 7-9}
+```js title="astro.config.mjs" {5}
 import { defineConfig } from 'astro/config';
 
 export default defineConfig({
@@ -251,7 +251,7 @@ These redirects follow the same rules as file-based routes. Dynamic routes are a
 
 Using SSR or a static adapter, you can also provide an object as the value, allowing you to specify the `status` code in addition to the new `destination`:
 
-```js
+```js title="astro.config.mjs" {5-8}
 import { defineConfig } from 'astro/config';
 
 export default defineConfig({


src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -231,7 +231,7 @@ You can define rules to [redirect users to permanently-moved pages](#configured-
 
 You can specify a mapping of permanent redirects in your Astro config with the `redirects` value. For most redirects, this is a mapping of an old route to the new route:
 
-```js title="astro.config.mjs" {5, 7-9}
+```js title="astro.config.mjs" {5}
```
sarah11918, line 234:
```suggestion
```js title="astro.config.mjs" {4-6}
```
I think the original intention was to show also adding the `redirects: {}` config option, since it won't otherwise exist and this is the first time they're seeing this. So, I suggest highlighting all of this!

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -251,7 +251,7 @@ These redirects follow the same rules as file-based routes. Dynamic routes are a
 
 Using SSR or a static adapter, you can also provide an object as the value, allowing you to specify the `status` code in addition to the new `destination`:
 
-```js
+```js title="astro.config.mjs" {5-8}
```
sarah11918, line 254: Totally agree with not also highlighting `redirects` here, since we're focusing on the object that you can provide!  :raised_hands:

sarah11918 (commented):
Thanks so much for jumping on this @AishaBlake! I love me some consistent titles, and oh no, how did we never catch the previous wonky highlighting?  :sweat_smile: 

Just quick suggestion below to see what you think, and so thrilled to be welcoming you to Team Docs! :partying_face:

src/content/docs/en/core-concepts/routing.mdx
AishaBlake: I could see using `ins` for the second instance, comparing `'/new-page'` to the object. In this first case, though, we're not really comparing anything. Thoughts?

sarah11918 (approved):
Looking great to me! I'm approving to get this in our merge queue, since it's a constant "update branch" battle around here while Yan is merging translation PRs. But this will go live just as soon as can!

Thanks again for taking the time to contribute while I know you're so busy with other, non-tech endeavours. Does this mean... maybe we see an Astro Opal Grove Games website in the near future??? :eyes: :smile:

src/content/docs/en/core-concepts/routing.mdx
kevinzunigacuellar: I see, I was thinking that since this code snippet introduces a new configuration option to the config file. It would be easier to follow the green highlight. Sort of "Add this to the config file to opt-in for redirects"

AishaBlake:
Actually, yes! I migrated my personal site as a test and now I'm working on moving Opal Grove's over. Starting with the one Astro Shopify theme but I've got a bunch of things I want to add 🤩

I'll share more on Twitter but excited about this first little PR!

PR #4008, opened 2023-08-01
i18n(zh-cn): Translate `guides/deploy/space.mdx`

100gle:

- New or updated content
- Translated content

#### Description

Translate `guides/deploy/space.mdx` content to Chinese.


Yan-Thomas:
Thanks @100gle, nice catch with the English links, but could you make it a separate PR only with the English file changes? This way, other translators won't be confused that there is a Chinese Translation PR in the "source change history" for the original page, after this I'll merge it right away and merge this translation next!

100gle:
> Thanks @100gle, nice catch with the English links, but could you make it a separate PR only with the English file changes? This way, other translators won't be confused that there is a Chinese Translation PR in the "source change history" for the original page, after this I'll merge it right away and merge this translation next!

OK. I will separate it into a new PR tomorrow.

Yan-Thomas (approved):
LGTM!

PR #4009, opened 2023-08-01
Storyblok folder and API behavior note

thejackshelton:
…slugs in an API request.

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 


src/content/docs/en/guides/cms/storyblok.mdx
```diff
@@ -443,7 +443,9 @@ const story = data.story;
 </html>
 ```
 
-This file will generate a page for each story, with the slug and content fetched from the Storyblok API. 
+This file will generate a page for each story, with the slug and content fetched from the Storyblok API.
+
+> When adding folders inside of storyblok, you'll want to include them when interacting with the Storyblok API. For example, in the get request above we could use **cdn/stories/blog**, with a blog folder inside rather than using them at the root.
```
ElianCodes, line 448: Don't you think this would function better as a note component?

also improved the wording a bit.

```suggestion
:::note
When adding folders inside of Storyblok, you'll want to include them when interacting with the Storyblok API. For example, in the GET request above we could use **cdn/stories/blog**, with a blog folder inside rather than using them at the root.
:::
```

ElianCodes (commented):
Thanks for your PR @thejackshelton!

Left a little comment for you to review!

src/content/docs/en/guides/cms/storyblok.mdx
sarah11918, line 448: Thanks for this addition, @thejackshelton! In order to figure out how to best include this content, I do have some questions though. You wouldn't think one single paragraph sparks so much discussion, and yet, here we are!  That's my job!  :smile:

- @ElianCodes is right in that we don't use "block quote" syntax for notes. (We use it for.... quotations!) We have a specific note format for asides.
- BUT, we only use notes for truly tangential, "beside the point" information. And I'm not sure whether that's the content you're giving here.

In general the wording is a little unclear to me:
- The "when...., you will want to..." phrasing is a bit tricky and usually ends up with more words than necessary. We'd often go for a more direct "For" instead, e.g. "For content that has been edited.... do this" In fact, we try to avoid "You" at all, and just state the instructions / active verb directly, and under which circumstances when that's not universal advice.

- "could" is a very scary word in docs! "You *could* do this..." Well,... should I? Must I? This is typically unclear advice to the reader, and we always look closely when we see this word in the hopes of making it more explicit. This is telling me to "idk, maybe?" do something, and we try to avoid using the word "could" unless we are truly describing an uncertainty. (It "could" rain...) To let the reader know that they have a choice, we would typical tell them that they "can" do X or Y. This sounds more confident, reassuring, and definitive! 

(And, we are often strategic about when we let them know they have a choice, in hopes of keeping them on a happy path when following the instructions. We don't HAVE to tell them of all the things possible.)

So, we need to establish how key is this information to include so that a reader's project will work when they follow the steps?

- If key, then we do NOT use a note, and we would try to start a sentence with "For this particular situation.... ,(do this)" if that fits. 
- If a kind of "heads up!" aside, less common situation but they really should know, then we would use a note.

I would also prefer to avoid phrases like "you will want to" and "could"?  Would you be willing to take another shot at writing this paragraph, with more direct language that avoids addressing the reader and replaces "could" with more direct advice for the user?

ElianCodes:
Merging `main` in here to make sure it's rebased on v3

thejackshelton:
@ElianCodes @sarah11918 apologies for not responding sooner! 

I've been working on a couple open source projects, this week I also started getting more familiar with partytown, which the docs site is actually an Astro site on a much older version. Looking forward to getting it to v3.0 at some point with some cool view transitions! I saw some issues on there relating to the astrojs/partytown integration, gonna try to debug that later this week 👍 

I really appreciate the suggestions Sarah! I could definitely improve my language here. 

I initially created that note because I spent a good hour or two figuring this out (I was still new to storyblok). @ElianCodes I've committed your changes and I'm going to implement Sarah's feedback.

sarah11918 (approved):
Thanks so much for this @thejackshelton , and sorry it took so long for us to go back and forth. I'm trying to catch up after launch week.  :joy:

I love this in our note format instead of the block quote! So happy to (finally) welcome you to Team Docs!  🥳

PR #4027, opened 2023-08-03
chore: Fix the missing links for `guides/deploy/space.mdx`

100gle:

- New or updated content

#### Description

I found that the old Space document links are invalid When I translated Chinese version(#4008). So I fix it to make them all be right.


ElianCodes (approved):
Thanks @100gle for fixing these!

Let's ship it! 🚢🚀

PR #4028, opened 2023-08-03
Update explanatory title in the Deno adapter guide

sasoria:

#### Description

This pull request updates the explanatory title of the Deno adapter guide. It follows the standard set in #1787 that was introduced for the Node adapter.


TheOtterlord:
If I'm right, I believe this should be edited in the [package README](https://github.com/withastro/astro/tree/main/packages/integrations/deno/)? Since this file is pulled from the main repo. Otherwise, it will be overridden the next time a change is made over there.

PR #4029, opened 2023-08-03
Explain how to get environment variables in SSR mode

### LINKED ISSUE #3049: Explain how environment variables work in SSR apps
This is how environment variables are resolved in import.meta.env. Given a variable FOO what happens during the build is:

1. If `FOO` exists in the environment `import.meta.env.FOO` is replaced with `process.env.FOO`.
    - All of the adapters replace this code with the native environment calls (for ex. `Deno.env.get('FOO')`
2. If `FOO` does not exist in the environment `import.meta.env.FOO` is replaced statically. If it is in a `.env` file for example, that will be used as the static value.
    - If `FOO` is not in the environment and not in a `.env` file then it essentially will always be undefined at runtime.

sasoria:

- New or updated content

#### Description

- Closes #3049

This Pull request explains how to access environment variables in SSR mode. Feel free to add sugggestions, maybe we also should have a section on environment variables in the Deno adapter guide? 

Here's a complete list of how we can use environment variables in different adapters:

| Adapter | Usage |
| --- | --- |
| Node | `process.env` |
| Netlify | `process.env` |
| Vercel | `process.env` |
| Cloudflare | `import.meta.env` |
| Deno | `Deno.env.get` |

### DIFF
--- a/src/content/docs/en/guides/environment-variables.mdx
+++ b/src/content/docs/en/guides/environment-variables.mdx
@@ -93,7 +93,7 @@ Because Vite statically replaces `import.meta.env`, you cannot access it with dy
 :::
 
 In SSR mode the environment variables can be accessed at runtime based on the SSR adapter. For most adapters you can access environment variables with `process.env`. With the Deno adapter you can get them with `Deno.env.get()`.
-Cloudflare also has its [own way](en/guides/integrations-guide/cloudflare/#environment-variables) of handling environment variables. Astro will first check the server environment for variables and if they don't exist it will try to look for them in .env files.
+Cloudflare also has its [own way](/en/guides/integrations-guide/cloudflare/#environment-variables) of handling environment variables. Astro will first check the server environment for variables and if they don't exist it will try to look for them in .env files.
 
 ## IntelliSense for TypeScript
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -91,6 +92,9 @@ const data = fetch(`${import.meta.env.PUBLIC_POKEAPI}/pokemon/squirtle`);
 Because Vite statically replaces `import.meta.env`, you cannot access it with dynamic keys like `import.meta.env[key]`.
 :::
 
+In SSR mode the environment variables can be accessed at runtime based on the SSR adapter. For most adapters you can access environment variables with `process.env`. With the Deno adapter you can get them with `Deno.env.get()`.
+Cloudflare also has its [own way](/en/guides/integrations-guide/cloudflare/#environment-variables) of handling environment variables. Astro will first check the server environment for variables and if they don't exist it will try to look for them in .env files.
+
```
ElianCodes, line 97:
```suggestion
When using SSR, environment variables can be accessed at runtime based on the SSR adapter being used. With most adapters you can access environment variables with `process.env`, for the Deno adapter you have to use `Deno.env.get()`. Cloudflare has its [own way](/en/guides/integrations-guide/cloudflare/#environment-variables) of handling environment variables. 
Astro will check the server environment for variables, if they don't exist, Astro will look for them in .env files.

```

ElianCodes (commented):
Rewrote the wording a little! Everything else LGTM!

Big Thanks @sasoria! ❤️

src/content/docs/en/guides/environment-variables.mdx
sasoria: Thank you for the suggestions! 🙌

Issue #4034, opened 2023-08-03
Document new usecase for slot drilling


ElianCodes:
### 📚 Subject area/topic

named slots

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/core-concepts/astro-components/#named-slots

### 📋 General description or bullet points (if proposing new content)

Slot documentation can be enhanced with additional tips for nested layouts. See this example from the Discord AI bot that answered my question for me.

Yes, you can pass the "metadata" slot through Layout.astro to render it in BaseLayout.astro. To do this, you can use named slots in both Layout.astro and BaseLayout.astro. 

First, in BaseLayout.astro, define a named slot with the name "metadata":

```astro
<!-- BaseLayout.astro -->
<slot name="metadata" />
```

Next, in Layout.astro, import BaseLayout.astro and use a named slot with the same name "metadata":

```astro
<!-- Layout.astro -->
---
import BaseLayout from './BaseLayout.astro';
---
<BaseLayout>
  <slot name="metadata" slot="metadata" /> <-- This is the important bit
  <slot />
</BaseLayout>
```

Now, when you use Layout.astro in your page files, you can pass data to both the default slot and the named "metadata" slot, and they will be rendered correctly in BaseLayout.astro.

For example, in your page file:

```astro
<!-- src/pages/your-page.astro -->
---
import Layout from '../layouts/Layout.astro';
---
<Layout>
  <meta name="description" content="Your page description" slot="metadata" />
  <h1>Your page content</h1>
</Layout>
```

In this example, the meta tag with the "metadata" slot will be rendered in the "metadata" slot of BaseLayout.astro, and the h1 tag will be rendered in the default slot.

### 🖥️ Reproduction in StackBlitz (if reporting incorrect content or code samples)

/

PR #4045, opened 2023-08-03
transferring slots

BryceRussell:

- New or updated content

#### Description

Adds example on how to transfer slots from one component to another using a `<slot name="name" slot="name""/>` tag

- Closes: https://github.com/withastro/docs/issues/4034
- https://deploy-preview-4045--astro-docs-2.netlify.app/en/core-concepts/astro-components/#transferring-slots

Let me know if any of this could be worded better, I originally called this pattern "slot drilling" but I think "slot transferring" is a better word for explaining what is happening.

### DIFF
--- a/src/content/docs/en/core-concepts/astro-components.mdx
+++ b/src/content/docs/en/core-concepts/astro-components.mdx
@@ -281,6 +281,59 @@ const { title } = Astro.props
 </div>
 ```
 
+### Transferring slots
+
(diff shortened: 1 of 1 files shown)


ElianCodes (approved):
Looks good to me! Thanks @BryceRussell

ElianCodes:
I don't think there's any more real editing needed here, so let's ship it! 🚀🛳️

PR #4052, opened 2023-08-03
Add Bun package manager to docs

colinhacks:

- New or updated content

#### Description

Where relevant, this adds tabs for the `bun` package manager and `bunx` package runner alongside npm/yarn/pnpm. (Bun can be used a [standalone package manager](https://bun.sh/package-manager) in any project, regardless of whether it uses Node.js or Bun as a runtime.)

### DIFF
--- a/src/components/tabs/Tabs.tsx
+++ b/src/components/tabs/Tabs.tsx
@@ -36,10 +36,24 @@ type Props = {
 	sharedStore?: string;
 };
 
+const tabOrdering = {
+	npm: 0,
+	pnpm: 1,
+	yarn: 2,
+	bun: 3,
+};
+
 export default function Tabs({ sharedStore, ...slots }: Props) {
 	const tabId = genTabId();
-	const tabs = Object.entries(slots).filter(isTabSlotEntry);
-	const panels = Object.entries(slots).filter(isPanelSlotEntry);
+	// npm then pnpm then yarn then bun
+	const sortedSlots = Object.entries(slots).sort(([a], [b]) => {
+		const aName = a.split('.')[1];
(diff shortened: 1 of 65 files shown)


src/components/tabs/Tabs.tsx
```diff
@@ -36,10 +36,24 @@ type Props = {
...
+	yarn: 2,
+	bun: 3,
+};
+
 export default function Tabs({ sharedStore, ...slots }: Props) {
 	const tabId = genTabId();
-	const tabs = Object.entries(slots).filter(isTabSlotEntry);
-	const panels = Object.entries(slots).filter(isPanelSlotEntry);
+	// npm then pnpm then yarn then bun
+	const sortedSlots = Object.entries(slots).sort(([a], [b]) => {
```
colinhacks, line 49: Before this, the tabs were getting sorted alphabetically, so Bun was first. This sets a fixed order for the tabs based on how the current appear in the docs.

sarah11918:
Thanks for this, @colinhacks ! I know the Technical Steering Committee is evaluating the PR made to Astro core, and intends to let us know the outcome of that, and any documentation decisions, so we are taking our cue from that!

ElianCodes:
Hey @colinhacks!

After a discussion with the core team, we've decided to not update the package manager tabs at this time. But, we do have a Bun recipe. We'll gladly take another look at the recipe, to make sure that using Bun as a package manager is documented there.

So, for now, we'll keep all Bun content to one page and maybe you also want to take a second look at that one: https://docs.astro.build/en/recipes/bun/

Thanks for putting so much time, effort and love in this PR!

PR #4080, opened 2023-08-05
[migrate-to-astro] content-first wording updates

sarah11918:

- New or updated content
#### Description

Updates the existing Migrate To Astro From... guides that had a "content-focused" point in the Key Differences section. (Note, not every guide had this difference.)

Updates our wording from "content-focused" to "content-first" to match the new CRA guide's language. We may eventually (or now!) want to update these and the CRA again together, but for now, at least they will be much closer to each other.


dreyfus92 (approved):
LGTM ✅

PR #4096, opened 2023-08-06
Update astro-expressive-code dependency

hippotastic:

#### Description

- In response to a discussion on the [Astro Discord](https://discord.com/channels/830184174198718474/872579324446928896/1136320887956721704) and a subsequent [feature request](https://github.com/expressive-code/expressive-code/issues/41) for Expressive Code, I added a feature to the "Copy to clipboard" button of terminal window frames that removes comment lines starting with `#` from the copied text.

  This is useful to reduce the copied text to the actual commands users need to run, instead of also copying explanatory comments or instructions.
- This PR updates our `astro-expressive-code` dependency to enable this feature in Astro Docs.


hippotastic:
I just noticed that our previous version of Expressive Code was quite old, so this new version also brings other useful features.

Here's the most important one for our use case:

- Support for `diff`-like syntax and `lang` meta attribute.

  To mark lines as inserted or deleted, we can now use the widely supported `diff` language as an alternative to adding line numbers to the opening code fence.

  We can even specify a separate syntax highlighting language by adding a `lang="..."` attribute to the opening fence. See [README.md](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#combining-syntax-highlighting-with-diff-like-syntax) for more details.

  This is especially helpful for any content that is not only displayed on Astro Docs, but also on other platforms like NPM (e.g. integration READMEs). Other platforms will display the diff highlighting, while Astro Docs will also be able to display proper syntax highlighting on top of the diff.

delucis (approved):
This looks great — thanks for the improvements and the update @hippotastic!

I tested the terminal snippets and they worked as expected.

Also realised while testing that lines marked with `del` are included in the copied text. So I guess we have more to think about there 😄

PR #4149, opened 2023-08-09
i18n(it): Translated recipes page

csarnataro:

- Translated content

#### Description

- Closes # N/A
- Added the translation of the recipes page (src/content/docs/it/recipes.mdx)


src/content/docs/it/recipes.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+## Soluzioni dalla community
+
+Se hai una soluzione che vuoi condividere, consulta [le nostre linee guida per contribuire](https://github.com/withastro/docs/blob/main/contributor-guides/submitting-a-recipe.md)  
+
+- [Importare immagini utilizzando un nome file dinamico](https://vaihe.com/blog/astro/astro-dynamic-image-prop/)
+- [Aggiungere transizioni di pagina animate con Swup](https://navillus.dev/blog/astro-plus-swup)
+- [Utilizzare UnoCSS in Astro](https://www.elian.codes/blog/23-02-11-implementing-unocss-in-astro/)
+- [Costruire una tabella dei contenuti utilizzando le intestazioni dei file markdown di Astro](https://kld.dev/building-table-of-contents/)
+- [Creare un plugin Remark per rimuovere i "righini" dai tuoi file Markdown](https://eatmon.co/blog/remove-runts-markdown/)
```
csarnataro, line 23: "Righini" (a word completely unknown to me until today) is the translation of "runts".
See: 
https://denisesarrecchia.org/2020/02/20/orfane-vedove-e-righini-gli-errori-grafici-piu-comuni/ 
https://www.treccani.it/vocabolario/righino/

src/content/docs/it/recipes.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+## Soluzioni ufficiali
+
+<RecipesNav />
+
+## Soluzioni dalla community
+
+Se hai una soluzione che vuoi condividere, consulta [le nostre linee guida per contribuire](https://github.com/withastro/docs/blob/main/contributor-guides/submitting-a-recipe.md)  
+
+- [Importare immagini utilizzando un nome file dinamico](https://vaihe.com/blog/astro/astro-dynamic-image-prop/)
```
csarnataro, line 19: **question:** Since the original source of the recipe is in English, should we add a note (e.g. `<sup>EN</sup>` to let users know that the original recipe is not in Italian? Something similar to 

image

in https://docs.astro.build/it/getting-started/ ?

Other translations are not including this note (e.g. the Spanish one: https://docs.astro.build/es/recipes/ ), nevertheless we could add it, if it can be useful for Italian readers

milovangudelj: Those annotations are only used in the navbar and refer to pages that are present on the docs website. Adding them to external links would require a new component and it would probably impact the overall readability of the page.

It's not worth it in my opinion.

src/content/docs/it/recipes.mdx
```diff
@@ -0,0 +1,36 @@
...
+- [Aggiungere al tuo sito la funzionalità di ricerca con Pagefind](https://blog.otterlord.dev/post/astro-search/)
+- [Configurare VSCode, ESLint e Prettier in progetti Astro](https://patheticgeek.dev/blog/astro-prettier-eslint-vscode)
+- [Ottimizzare le build di progetti Astro: modificare i file HTML post-build](https://straffesites.com/en/blog/optimize-astro-html-post-build)
+- [Aggiungere la modalità scura in Astro con Tailwind CSS](https://www.kevinzunigacuellar.com/blog/dark-mode-in-astro/)
+- [Generare sitemaps localizzate nei tuoi progetti Astro con Storyblok](https://straffesites.com/en/blog/localized-sitemap-astro-storyblok)
+- [Integrare Prettier con Astro e Tailwind CSS](https://straffesites.com/en/blog/integrate-prettier-astro-tailwindcss)
+- [Aggiungere un feed RSS ad Astro utilizzando le storie di Storyblok](https://straffesites.com/en/blog/rss-feed-astro-storyblok)
+- [Utilizzare tRPC in Astro](https://www.thomasledoux.be/blog/using-trpc-astro-islands-react)
+- [Aggiungere il supporto multilingua al tuo sito Astro con Crowdin: una guida alla localizzazione in Astro](https://crowdin.com/blog/2023/06/21/astro-localization-and-i18n)
```
milovangudelj, line 32:
```suggestion
- [Come rendere il tuo sito Astro multilingue con Crowdin: una guida alla localizzazione in Astro](https://crowdin.com/blog/2023/06/21/astro-localization-and-i18n)
```

src/content/docs/it/recipes.mdx
```diff
@@ -0,0 +1,36 @@
...
+- [Aggiungere le immagini di un blog post al feed RSS in Astro](https://webreaper.dev/posts/astro-rss-feed-blog-post-images/)
```
milovangudelj, line 33:
```suggestion
- [Aggiungere le immagini di un blog post al feed RSS Astro](https://webreaper.dev/posts/astro-rss-feed-blog-post-images/)
```

Astro should be treated as a brand and therefore we shouldn't add possessive prepositions to it unless they're present in the original document. It's the same as in "pasta della Barilla vs pasta Barilla".

src/content/docs/it/recipes.mdx
```diff
@@ -0,0 +1,36 @@
...
+- [Adottare un'architettura micro-frontends con Astro](https://medium.com/@sergio.a.soria/setting-up-micro-frontends-with-astro-and-ecma-script-modules-137340d2c520)
+- [Aggiungere autenticazione basata su nome utente e password con Lucia](https://lucia-auth.com/guidebook/sign-in-with-username-and-password/astro)
+- [Aggiungere Github OAuth con Lucia](https://lucia-auth.com/guidebook/github-oauth/astro)
```
milovangudelj, line 36:
```suggestion
- [Aggiungere Github OAuth con Lucia](https://lucia-auth.com/guidebook/github-oauth/astro)
- [Integrare Sentry con Astro](https://akashrajpurohit.com/blog/seamless-error-tracking-integrating-sentry-with-astro/)
```

This line was added in #4182

milovangudelj (changes requested):
Added a few suggestions here and there.

src/content/docs/it/recipes.mdx
csarnataro: "Come rendere il tuo sito" doesn't sounds good to me. What about something like "come realizzare" or "come sviluppare" or even "come implementare" (which is rather more technical)

csarnataro: BTW, today I learnt that "multilingua" (with final "a") is not a word! It's so common that I used it for years, but it seems to be wrong.

https://www.treccani.it/vocabolario/multilingue/

milovangudelj: Yup, it's one of those cases where people commonly just use the wrong term and forget about the correct one. You never stop learnig new things :v:

milovangudelj: > "Come rendere il tuo sito" doesn't sounds good to me. What about something like "come realizzare" or "come sviluppare" or even "come implementare" (which is rather more technical)

Idk... "How to make ..." has nothing technical to it and it's a transformative action so "Come rendere ..." is just the simplest most straightforward way to translate it. And an even better solution would be to write it like this "Come rendere multilingue il tuo sito Astro ...". Just moving the word "multilingue" makes it easier to read.

csarnataro: Agree, "rendere multilingue..." makes everything clearer.

milovangudelj (approved):
LGTM! :rocket:

Yan-Thomas (approved):
LGTM! Thanks peeps 🙌

PR #4168, opened 2023-08-10
Update Environment Variables example

natemoo-re:

#### Description

- Closes https://github.com/withastro/astro/issues/6241
- The example in the [Environment Variables](https://docs.astro.build/en/guides/configuring-astro/#environment-variables) section was misleading because using `import.meta.env` is not supported in the config file. This value was always set to `'development'`, which was confusing users.


ElianCodes (approved):
Thanks for this fix Nate!

PR #4191, opened 2023-08-12
Adding multi-stage build to the Docker recipe

sentisso:

- New or updated content

#### Description

This PR extends the [Docker recipe](https://docs.astro.build/en/recipes/docker/#ssr) by adding a [multi-stage build](https://docs.docker.com/build/building/multi-stage/) to the present Dockerfile example, which makes sure that the npm dependencies do not get reinstalled on every code change.

I'm aware that right below the Dockerfile is a note saying:
> These are just examples of Dockerfiles. You can customize them to your needs.

and that the devs could implement the multi-stage build themselves, but I still think this new example is a good idea, since most of the devs will most likely just copy the Dockerfile.

E.g. in our team we blindly copied the previous Dockerfile example, it worked right out of the box and because of that, we didn't bother to modify it, even though we could've save ~3 minutes on each deploy by using this exact multi-stage build config _(our deploy time went from ~4 minutes to ~1 minute, because the dependencies do not need to be reinstalled on each source code change)_.


sentisso:
I was thinking that maybe the text above the Dockerfile example should briefly mention the used multi-stage build technique, it's advantages and maybe a link to the official Docker docs.

ElianCodes:
Hey @sentisso, thank you so much for putting time and effort in this PR!

While I do agree that a multi-stage build is a way better way of writing and deploying your project, this might be a little bit to complicated for 'newcomers' to both Astro and Docker.

Maybe it would be better to leave the old `Dockerfile` as is, and create a new section with a header especially for multi-stage builds. 

What do you think of that @sentisso

sentisso:
Hello @ElianCodes, yes I agree that it would be better to have those two sections.

The old one basically saying: _"these are the minimal steps, that will get the app running"_ and the new one: _"here's a more advanced setup, that takes the advantage of some ✨ docker magic ✨, which will make the deploys more efficient"_.

Now there's a question on how to phrase this... I was thinking that something like this could work:

## SSR
This minimal Dockerfile will build your site and serve it using Node.js on port 3000 and therefore requires the [Node adapter](https://docs.astro.build/en/guides/integrations-guide/node/) installed in your Astro project.
`insert old Dockerfile...`

## Multi-stage build (not sure about this header)
Here's an example of a more advanced Dockerfile that will, thanks to Docker's [multi-stage builds](https://docs.docker.com/build/building/multi-stage/), build your site in a more efficient way, reducing the build time sometimes even by minutes.
By taking advantage of the mutliple build stages we make sure that the npm dependencies do not get reinstalled unnecessarily when **only the source code** changes.
`insert new Dockerfile...`

ElianCodes:
Hey @sentisso, thanks for taking the time to discuss and think about this!

I can certainly see `multi-stage build` being a header. I think the location for this would be best under `### static`.
Everything else you mentioned; I agree with!

sentisso:
Hello @ElianCodes , I've updated the fork with the new multi-stage build section and reverted the old `Dockerfile`.
I've also modified the text a little bit, so now it sounds a bit better.

Lemme know, if any changes or corrections should made. Otherwise I'm quite satisfied with the current version.

Thanks!

src/content/docs/en/recipes/docker.mdx
```diff
@@ -31,7 +31,7 @@ WORKDIR /app
 COPY . .
 
 RUN npm install
-RUN npm run build
+RUN npm build
```
ElianCodes, line 34: Pretty sure this has to stay `npm run build`. `npm build` is an internal command an not what you specified in `package.json` scripts.

```suggestion
RUN npm run build
```

src/content/docs/en/recipes/docker.mdx
```diff
@@ -128,6 +128,38 @@ http {
 }
 ```
 
+#### Multi-stage build
+Here's an example of a more advanced Dockerfile that, thanks to Docker's [multi-stage builds](https://docs.docker.com/build/building/multi-stage/), optimizes the build process for your site, sometimes reducing the build time even by minutes.
```
ElianCodes, line 132: I think this title might be better, since we are showing an example using Node adapter
I also combined both alineas into one, which makes it a bit shorter!

```suggestion
### Multi-stage build (using SSR)

Here's an example of a more advanced Dockerfile that, thanks to Docker's [multi-stage builds](https://docs.docker.com/build/building/multi-stage/), optimizes the build process for your site by only reinstalling the npm dependencies when the source code changes. This can reduce the build time even by minutes, depending on the size of your dependencies.
```

src/content/docs/en/recipes/docker.mdx
```diff
@@ -128,6 +128,38 @@ http {
...
+By utilizing multiple build stages, we ensure that the npm dependencies are not unnecessarily reinstalled when **only the source code** changes.
```
ElianCodes, line 133:
```suggestion
```

src/content/docs/en/recipes/docker.mdx
```diff
@@ -128,6 +128,38 @@ http {
...
+
+```docker title="Dockerfile"
+FROM node:lts AS base
+WORKDIR /app
+
+# By copying only the package.json and package-lock.json here, we ensure that the following `-deps` steps are independent of the source code.
+# Therefore, the `-deps` steps will be skipped if only the source code changes.
+COPY package.json pacakge-lock.json ./
```
ElianCodes, line 141: small typo
```suggestion
COPY package.json package-lock.json ./
```

ElianCodes (commented):
Left a couple typo's and suggestions for you to review!

But looking good! Thanks for your work @sentisso!

sarah11918 (approved):
Hey @sentisso! It's been a few days since @ElianCodes last review. You did say you were satisfied, and these changes look pretty minor, so I'm going to suggest that Elian commit his own suggestions directly so we can get this helpful content merged, and you officially into Team Docs!  :partying_face: 

Feel free to come in and make the changes if you get here before Elian does, but otherwise, I'm suggesting that Elian merge and if you spot anything that needs changing, well... then you just get your SECOND Astro Docs PR, don't you?  :smile:

Thank you for this helpful addition!  :raised_hands:

PR #4226, opened 2023-08-15
docs: Update code highlight in layouts.mdx

codersjj:

#### Description

I think it might be better to keep the code highlighting format here consistent with previous sections.
Like this: `{frontmatter.title}`, rather than this: `frontmatter.title`


ElianCodes (approved):
Hey @codersjj! Thanks for taking the time to look into this!

Looks good to me!

PR #4166, opened 2023-08-10
Upgrade to Astro v3 guide

ElianCodes:
#### What kind of changes does this PR include?

- New or updated content

#### Description

- Closes #3756
- Adds the upgrade to v3 guide
- Will be updated as each 3.0 beta release changelog is published

**This is still very draft. Do not begin to translate yet.** (But, closer to launch week we WILL alert translators so that we can merge the upgrade guide on launch day in as many languages possible at once.)  :tada: 

#### Todo:

- [x] Change "Upgrade to Astro v2" -> "Upgrade to Astro v3" in sidebar
~~- [ ]Use custom aside to mention upgrade guide (can just change the text from the starlight one)~~
- [x] @sarah11918 should figure out if separate upgrade image guide is needed
- [x] All changes should be reflected in regular docs pages; e.g. Node version. (esp. check tutorial!)

#### References:

- [HackMD organization notes](https://hackmd.io/m4L-DcPDQ6eVDMrWkXu6Dg)
- [Old v2 upgrade guide model/format](https://docs.astro.build/en/guides/upgrade-to/v2/)

## Preview

https://deploy-preview-4166--astro-docs-2.netlify.app/en/guides/upgrade-to/v3/

Issue #3756, opened 2023-07-18
Astro 3.0 breaking changes tracker


matthewp:
### 📋 Explain your issue

This issue is to track 3.0 breaking changes to be part of the guide. Will add things to this through comments, using this template:

```markdown
### [changed/removed/added/deprecated] feature/name/title/description
In Astro v2.x, [ a statement in the past tense about what Astro did/used to do. e.g. Astro bundled all assets into several different folders....]

Astro v3.0 [ a statement in the present tense about how Astro works now. e.g. now bundles all assets into one single folder located in... ]

#### What should I do?

[Active sentences in the imperative (command) tense, with code examples showing diffs as appropriate]

e.g. Update... / Remove... / Replace... / Create...
```

Example upgrade guide: https://docs.astro.build/en/guides/upgrade-to/v2/

Princesseuh:
### Removed support for Node 16
In Astro v2.x, Node 16 was supported.

Astro v3.0 it's not the case anymore, the lowest supported version is now `v18.14.1`

#### What should I do?

https://docs.astro.build/en/guides/upgrade-to/v2/ has a section for when we dropped Node 14, so it'll be pretty much the same thing. This page also has a mention of the supported version: <https://docs.astro.build/en/install/auto/>

If a technical reason is wanted, Node 16 is getting deprecated in September, so not too long after Astro 3.0 releases. Additionally `v18.14.1` was chosen specifically because of bugs in previous versions of Node 18

ematipico:
### Deprecated lowercase function endpoints 

In Astro v2.x, endpoints are defined using lowercase function names: `get`, `post`, `put`, `all`, and `del`.

Astro v3.0 deprecates them in favour of uppercase function names.

#### What should I do?

Update the names as follow:

- `get` to `GET`
- `post` to `POST`
- `put` to `PUT`
- `all` to `ALL`
- `del` to `DELETE`

Princesseuh:
### [changed/removed/added/deprecated] feature/name/title/description
In Astro v2.x, we would magically flatten the result of `getStaticPaths` automatically for the user, so if you returned an array of array, it would still work

Astro v3.0 we removed this in favour of the user doing it themselves when they need to, in line with how we typically prefer for users to do explicit things.

#### What should I do?

If you're returning an array of arrays instead of an array of object (as you should), please flatten it using `.flat()`, or if you're using a `.map()`, use `.flatMap()` instead

(we also added an error message for this)

bluwy:
[...]

PR #4239, opened 2023-08-16
Use astro: namespace modules

matthewp:

- New or updated content

#### Description

- In 3.0 there are new aliases for middleware and components. This updates the docs to use those.


matthewp:
I didn't update https://github.com/withastro/docs/issues/3756 because this is technically *not* a breaking change, the old imports will still work. But we are recommending the `astro:` namespace now. So happy to add to that issue as well if it feels like it should be.

ElianCodes:
Hey Matthew, as this is a 3.0 change, can you send the PR to `v3-upgrade-guide` instead of `main`?

That way, we can merge that PR when we're ready for v3! (https://github.com/withastro/docs/pull/4166)

sarah11918 (approved):
Approving this PR for 3.0 Launch!

Notes for Docs:
- Not a breaking issue, and the old notation still works, so we'll not worry about adding to the migration guide and no, don't need an issue! We'll just keep our messaging consistent going forward.
- We might change the base branch for this before merging
- Let's also do our own find/replace to confirm we haven't missed any!

PR #4244, opened 2023-08-17
Update description in "Using one Layout for .md, .mdx, and .astro" section.

codersjj:

- New or updated content

#### Description

Considering the code first tries to get the `title` from `frontmatter`, the order in the paragraph should be:

the layout will display the page title either from a frontmatter YAML `title` property or from an Astro component passing a `title` attribute

This more accurately represents the actual execution flow of the code:

1. First try to get the `title` from `frontmatter`
2. If not exists, then use the `title` attribute from component

By swapping the order in the paragraph, it can be more consistent with the code behavior and improve readability.


sarah11918 (approved):
Approved for merge queue! Thanks for this thoughtful edit, @codersjj !

Yan-Thomas (approved):
LGTM! Thanks @codersjj 🙌

PR #4255, opened 2023-08-18
fix: show where client directives should be added

third774:

- New or updated content

#### Description

Adds information about where client directives should be registered. This wasn't clear to me when reading the docs, had to find the original RFC for reference. Hope this helps!


third774:
Er, hrm — I don't think I've got it right just yet. Leaving this open though to maybe serve as a starting point for discussion around how this could be better. Otherwise, once I get it working I'll come back and update this PR.

PR #4257, opened 2023-08-18
fix: show how client directives should be added

third774:

- New or updated content

#### Description

Adds a more thorough example of how to add a client directive.


ElianCodes (approved):
Hey @third774! Thanks so much for adding this!

Let's get it merged 🚀🛳️

PR #4261, opened 2023-08-18
Update nav styles code in `tutorial/3-components/3.mdx`

codersjj:

#### Description

Although the original CSS works fine at this stage of the tutorial, it could be written more precisely. This is because the desired effect is 

```css
.nav-links a:hover {
  background-color: #ff9776;
}
.nav-links a:focus {
  background-color: #ff9776;
}
```

but writing it as 

```css
.nav-links a:hover, a:focus {
  background-color: #ff9776;
}
```

makes it equivalent to 

```css
.nav-links a:hover {
  background-color: #ff9776;
}
a:focus {
  background-color: #ff9776;
}
```

which may not be the intended result.


codersjj:
Or, change it to this:

```css
.nav-links :is(a:hover, a:focus) {
  background-color: #ff9776;
}
```

sarah11918:
Hi @codersjj! Thanks for leaving some PRs to help improve our Tutorial!

As you can see, I've lableled and assigned them. Just a heads up that part of the reason these are taking a while to go through is that if we decide to make a change to any of the blog tutorial's code in the text/instructions, then we also need to update the blog demo code in another repo itself, and take care to properly adjust each appropriate branch, and that's just a bit more of a process. :smile:

We're preparing for Astro v3.0 launch week next week (!!) so we're trying to make sure that all new features are properly documented and that our docs are ready to help people with potentially breaking changes. So, I'm sure these PRs of yours look like quick fixes, but we not only need extra time and resource to get tutorial code changes just right, we also have full plates right now!

So thanks for your patience, and we'll try not to let these sit here for too long before we triage and tackle them!  Thank you again for contributing! :raised_hands:

ElianCodes (approved):
Hey @codersjj! Looking good! I'll take it from here to get our codesample repo updated too!

PR #4262, opened 2023-08-18
Adds new document about the user can't pass jsx-style node to props to Astro Syntax.

cihad:

- New content

##### Added 2 section to [Differences between Astro and JSX](https://docs.astro.build/en/core-concepts/astro-syntax/#differences-between-astro-and-jsx):
- You can't pass node to props
- You can't get node jsx-style nodes in script body.


src/content/docs/en/core-concepts/astro-syntax.mdx
```diff
@@ -148,6 +148,41 @@ In Astro, you use the standard `kebab-case` format for all HTML attributes inste
...
+]
+
+// 🚫 This doesn't work
+const renderedTodos = todos.map(todo => <TodoItem todo={todo} />) 
+---
+
+<div>
+  {renderedTodos}
+</div>
+```
```
bluwy, line 170: I think we should clarify this early at the top of the `JSX-like expressions`, maybe under:

> You can define local JavaScript variables inside of the frontmatter component script between the two code fences (---) of an Astro component.

We should clarify that anything between two code fences only support JS or TS, and not JSX.

src/content/docs/en/core-concepts/astro-syntax.mdx
```diff
@@ -148,6 +148,41 @@ In Astro, you use the standard `kebab-case` format for all HTML attributes inste
...
+
+#### Node as Property Input
+
+In Astro, you can't use component or html nodes as property input.
```
bluwy, line 174: Perhaps instead of "Property Input", we can refer to them as "Attribute Values", similar to the "Attributes" section above. Maybe this can be merged into the "Attributes" section as well, but I'm also fine if it's a new section.

ElianCodes:
Merging `main` in here to make sure it's rebased on v3

sarah11918:
Hi @cihad ! As others have commented, this is not how we would document the notes you are trying to add. But there is some good content that you have mentioned! If we add it in a slightly different format, I do think it would make our docs better!

 Are you still interested in working on this PR and using Bluwy's suggestions?

sarah11918:
Hi @cihad, since there hasn't been any recent activity on this PR I am closing it. You can reopen if you'd like to pick up working on this contribution with everyone's suggestions!

PR #4274, opened 2023-08-19
Update code in `tutorial/5-astro-api/2.mdx`

codersjj:

#### Description

Keep consistency with [previous code](https://github.com/codersjj/docs-1/blob/eab2a55219ecdaada2097eb9cbcb1997ad29d6db/src/content/docs/en/tutorial/5-astro-api/2.mdx#L95C11-L95C11).


ElianCodes:
Hey @codersjj, while I do understand what you're going for. I don't think we need to add this change, since we're always defining our tags in every post.

codersjj:
> Hey @codersjj, while I do understand what you're going for. I don't think we need to add this change, since we're always defining our tags in every post.

Alright, should I close this PR then?

PR #4297, opened 2023-08-21
[DRAFT] Exploring core concept changes for 3.0

FredKSchott:
> _Note: I'm opening this PR way earlier than I normally would so that I can get preview URLs as I work. This is all still in flux and not yet ready for feedback!_

## Summary of changes

- **Getting Started:** removes existing high-level explanation of Astro, new one still TODO
- **Why Astro:** removes existing detailed explanation of Astro, new one still TODO
- **MPA vs. SPA:** removed, no longer relevant distinction
- **Sidebar/Nav changes:** 
  - removed dedicated "Core Concepts" section
  - moved "Why Astro" into "Start Here" section, renamed to "Why Astro?"
  - moved "Astro Islands" into "Basics" section, renamed to "Islands" to match the rest
  - moved "Endpoints" into "Basics" section, renamed to "APIs" to match the rest

## Before

Screen Shot 2023-08-20 at 11 46 22 PM

## After

Screen Shot 2023-08-20 at 11 45 51 PM


sarah11918:
Notes: as long as pages are just moving around in the sidebar (do not change file names/paths, but the page display name `nav.ts` can be whatever), no bigger changes are necessary throughout the rest of docs.

If we remove MPA vs SPA, we need to:
- [ ] remove it from the sidebar
- [ ] move the page into `old-translations` folder in the other languages
- [ ] search the entire docs repo for any internal links that should be updated
- [ ] set up a redirect in `netlify.toml` and **decide where this should now go** (I suggest a section in "Why Astro?" that we can link to directly, that we want people to read when they might be following a link from an external source. See next point)
- [ ] **Specifically write content that replaces the idea of "MPA vs SPA"**, use a header that can be linked to for the Netlify redirect. (Unless you just want this to point to Why Astro page generically. BUT, I think if people have been led to this page, this is the perfect moment to send them to a very specfic, targeted message)

A separate comment: I suspect people will notice if this note is no longer in docs, given how much attention it regularly receives. That doesn't mean we can't ditch it, or that it can't change. BUT if we release revamped docs and this note is at all different (or removed), I *do* suspect people will notice. We just should be prepared for that response.

image

sarah11918 (changes requested):
Preemptively blocking!  :sweat_smile:

This PR should be merged *last* of the v3.0 PRs.

Issue #4318, opened 2023-08-22
[v3 checks] `markdown.draft`


sarah11918:
### 📚 Subject area/topic

markdown.draft

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/markdown-content/#draft-pages

### 📋 General description or bullet points (if proposing new content)

@bluwy to check this section of the Markdown page and make any relevant changes based on https://github.com/withastro/astro/pull/8099/ (`markdown.draft` deprecated). Notably (but not limited to, if you notice something else!!):

1. Is `mdx.draft` still fine?
2. Is the method of `Astro.glob()` shown for filtering based on `draft:true` still fine, BUT the config option to build by default is gone, or should the ENTIRE notion of draft Markdown pages when NOT using content collections be removed?

Thanks!  :raised_hands: 

### 🖥️ Reproduction in StackBlitz (if reporting incorrect content or code samples)

none

bluwy:
I'll send a PR, but gonna fill in the answers here for context first:

> 1. Is `mdx.draft` still fine?

Anything `draft` except via content collections should be removed. So MDX drafts should also be removed, but markdown would be the most common case especially that our option is named `markdown.draft` when referring it.




> 2\. Is the method of `Astro.glob()` shown for filtering based on `draft:true` still fine, BUT the config option to build by default is gone, or should the ENTIRE notion of draft Markdown pages when NOT using content collections be removed?

Yeah I think `Astro.glob` and filter by `draft: true` is fine. But now it's more like a "this is a trick you can use if you have draft pages", than a feature to cover. So I think it makes more sense to remove it entirely?

Could also be less confusing if someone searches for "draft pages" and we'd only have one with content collections only. 🤔

sarah11918:
Not sure why this wasn't closed automatically, but closed by #4354 ! :partying_face:

PR #4354, opened 2023-08-24
Remove markdown draft reference

bluwy:

- New or updated content

#### Description

- Close https://github.com/withastro/docs/issues/4318
- Remove references of markdown and MDX draft features.
- Add a new small paragraph about content collections draft filtering where it's available in dev, but not in prod.

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -395,6 +395,16 @@ const publishedBlogEntries = await getCollection('blog', ({ data }) => {
 });
 ```
 
+You can also create draft pages that are available when running the dev server, but not built in production:
+
+```js
+// Example: Filter out content entries with `draft: true` only when building for production
+import { getCollection } from 'astro:content';
+const blogEntries = await getCollection('blog', ({ data }) => {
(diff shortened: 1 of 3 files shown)


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -395,6 +395,16 @@ const publishedBlogEntries = await getCollection('blog', ({ data }) => {
...
 
+You can also create draft pages that are available when running the dev server, but not built in production:
+
+```js
+// Example: Filter out content entries with `draft: true` only when building for production
+import { getCollection } from 'astro:content';
+const blogEntries = await getCollection('blog', ({ data }) => {
+  return import.meta.env.PROD ? data.draft !== true : true;
+});
+```
```
bluwy, line 406: Note: this example is similar to the one above, but I decided to split it out so it specifically mentions the word "draft pages" which helps with searching, and because the one above has a nice example that doesn't mix well if I combine them 🤔 

I'm open however if you think we should combine them still.

sarah11918 (approved):
Good to go! Thanks for snagging that issue, @bluwy !

sarah11918:
tagging @liruifengv in case you want to see these changes also going into the upgrade guide!  :purple_heart:

PR #4357, opened 2023-08-25
add note on draft with content collection support

trent-howard:

- New or updated content

#### Description

Content collections don't support the [draft pages feature](https://docs.astro.build/en/guides/markdown-content/#draft-pages), and it looks like support will not be added https://github.com/withastro/astro/issues/7324#issuecomment-1682698482

This adds a note to the draft pages section directing the reader to use a collection filter instead.


ElianCodes (approved):
Hey @trent-howard, thanks for your PR!

Merge coming up!

PR #4366, opened 2023-08-25
Add PH tip to getting started page

delucis:

- New or updated content

#### Description

- [x] Adds a tip to the Getting Started page to trail our PH launch of Starlight
- [x] Need to update the link once we have the correct URL

image

Should be good to go but keeping this as a draft as it shouldn’t be merged before 29 August.


sarah11918 (approved):
And, best news is that when we're ready to remove it, then it becomes the v3.0 upgrade link! 🚀

ElianCodes (approved):
🔥🔥 Hype Hype 🚀🚀

PR #4419, opened 2023-08-29
chore: supplement `functionPerRoute` docs

ematipico:

- New content

#### Description

This PR adds more information to the `functionPerRoute` option, in regards to the serverless environments.

### DIFF
--- a/src/content/docs/en/reference/adapter-reference.mdx
+++ b/src/content/docs/en/reference/adapter-reference.mdx
@@ -358,6 +358,30 @@ export default function createIntegration() {
 The `entryFile` is of type `URL` and represents the physical path of the file in the file system. This means that the paths change based on the OS where the code runs. 
 :::
 
+#### Serverless environments
+
+If you use `functionPerRoute` in a serverless environment, it's important you understand how this feature works under the hood, so you can take an informative decision.
+
+When enabled, Astro (and its adapter), will generate a JavaScript file called "handler" for each route. 
+
+:::note
+A handler might have different names based on your hosting platform: lambda, function, page, etc.
+:::
+
+Your hosting platform will run the handler and eventually will render the `Response` returned by it. 
+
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -358,6 +358,30 @@ export default function createIntegration() {
...
+
+If you choose `functionPerRoute: true`, each route becomes a handler and each one of them will be subject to a cold start.
+
+Instead, if you choose `functionPerRoute: false`, you'll have one single handler in charge of rendering all your routes. This means that the first time you trigger the handler, you'll suffer the cold start, but then all the other routes won't subject to it. 
+
+On the other hand, by shipping one single handler, you won't benefit from the code splitting.
+
+:::note
+It's important that you understand your hosting platform and how it works.
+:::
```
sarah11918, line 383:
```suggestion
Setting `functionPerRoute:true` in a serverless environment creates a JavaScript file (handler) for each route. A handler might have different names based on your hosting platform: lambda, function, page, etc. 

Each of these routes is subject to a [cold start](https://azure.microsoft.com/en-us/blog/understanding-serverless-cold-start/) when the handler runs, which may cause some delay. This delay is influenced by different factors.

With `functionPerRoute: false`, there is only one single handler in charge of rendering all your routes. When this handler is first triggered, you will be subject to a cold start. Then, all other routes should function without delay. However, you will lose the benefit of code splitting that `functionPerRoute:true` provides.

:::note
It's important that you understand your hosting platform, and how it works, in order to choose the appropriate `functionPerRoute` configuration for your project.
:::
```

sarah11918 (commented):
What do you think about something like this? @ematipico  Can you make this true?  :smile:

ElianCodes (commented):
This is very good! I agree with Sarah's suggestions, so LGTM after applying

ElianCodes (approved):
Lgtm!

dreyfus92 (approved):
LGTM ✅

PR #4421, opened 2023-08-29
Fix number cannot be assigned to string issue in script example

sunapi386:
Fix type issue number cannot be assigned to string.

#### Description

- Fix minor type issue.


delucis (approved):
Thank you for the fix @sunapi386! 🚀

PR #4267, opened 2023-08-18
Add basic custom Translation Tracker syntax

Yan-Thomas:
#### What kind of changes does this PR include?

- Changes to the docs site code

#### Description

This PR adds a rudimentary version of the custom tracker syntax to help with the 3.0 PRs changing lots of translations to fix links (LOOKING AT YOU 3.0 IMAGE GUIDE).

This allows you to add a `@tracker-major:(list of file paths separated by ;)` to a final squashed commit description (the one you have to confirm before you merge) that makes all changes be considered minor except the ones you have explicitly marked as major.

🚨 **THE FILE PATH HAS TO INCLUDE EVERYTHING (SEE BELOW)** 🚨 

Example:

**PR updates the pages (br/1.mdx, br/2.mdx), but also updates the link to another page that isn't fully updated yet (br/3.mdx).**
Using: `@tracker-major:./src/content/docs/br/1.mdx;./src/content/docs/br/2.mdx`, `br/3.mdx` continues to be marked as outdated.

More info in the code comment!

**How to test it?**
1. Pull the branch
2. Run `pnpm translation-status` (or with your preferred package manager) 
3. Open the `dist/translation-status/index.html` file. (this is the current status in the branch!)
4. Now, make a new commit changing any pages you want, and add the custom syntax to the commit description.
5. Do steps 2 & 3 again.
6. See the changes done in the tracker. Might be helpful to have `i18n.docs.astro.build` side by side, but will be inconsistent the moment the branch isn't 100% updated in relation to main.

PR #4422, opened 2023-08-29
Add glob matching to `@tracker-major` directive

Yan-Thomas:

- Translated content

#### Description

Now the directive also accepts glob patterns! Which will come in handy to deal with the 3.0 docs.

See the comment added to code plus the base PR for this feature for me instructions on how to test: https://github.com/withastro/docs/pull/4267

Note that minimatch glob syntax is a bit different from some other libraries, so check their docs first to be sure!

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "@nanostores/preact": "^0.1.3",
     "canvas-confetti": "^1.6.0",
     "jsdoc-api": "^7.1.1",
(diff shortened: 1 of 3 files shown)


sarah11918 (approved):
I trust you, Yan!

ElianCodes (approved):
Looks good from what I understand of it!

PR #4432, opened 2023-08-30
Create deprecation notices in config reference docgen script

delucis:

- Something else!

#### Description

- Updates our docgen script that generates https://docs.astro.build/en/reference/configuration-reference/
- If an entry in https://github.com/withastro/astro/blob/main/packages/astro/src/%40types/astro.ts includes a `@deprecated` tag for an entry, the script now injects a `:::caution[Deprecated]` aside with the JSDoc tag’s contents.
- This PR doesn’t include the changes — will let CI generate those once we have all the 3.0 docs merged, but here’s a preview: 
   Documentation for two build config options showing yellow deprecation notices


sarah11918 (approved):
This looks great @delucis ! 

Just to clarify re: usage, do we just provide a version number e.g. `@deprecated: 3.0` or do we provide the string at the source?

delucis:
> Just to clarify re: usage, do we just provide a version number e.g. @deprecated: 3.0 or do we provide the string at the source?

You should provide a full sentence at source. For example, for the asides in the screenshot the source is `@deprecated Deprecated since version 3.0.`. But there’s also scope there for providing more details like the `markdown.drafts` one which is `@deprecated Deprecated since version 3.0. Use content collections instead.`.

(It’s technically also supported to just have a bare `@deprecated` tag with _no_ additional content, but preference is for a sentence or two with full context.

Issue #4456, opened 2023-08-30
Mention Appending `?url` to images


sarah11918:
### 📚 Subject area/topic

images

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/images/

### 📋 General description or bullet points (if proposing new content)

Add this to images/upgrade guide!

https://github.com/withastro/astro/issues/8305

### 🖥️ Reproduction in StackBlitz (if reporting incorrect content or code samples)

none

PR #4462, opened 2023-08-31
Added documentation about old import behavior in Astro v3.x

dreyfus92:

- New or updated content

#### Description

- Closes https://github.com/withastro/docs/issues/4456

As requested in the previously mentioned issue, I added a section about how to deal with the old import behavior for images in Astro v3.x


src/content/docs/en/guides/images.mdx
```diff
@@ -700,3 +700,24 @@ If you were using the image integration in Astro v2.x, complete the following st
 You can now declare an associated image for a content collections entry, such as a blog post's cover image, in your frontmatter using its path relative to the current folder.
 
 The new `image` helper for content collections lets you validate the image metadata using Zod. Learn more about [how to use images in content collections](/en/guides/images/#images-in-content-collections)
+
+### Navigating Image Imports in Astro v3.0
+
+In Astro v3.0, if you find yourself wanting to preserve the old import behavior for images and require a string representation of the image's URL, consider appending `?raw` or `?url` to the end of your image path when importing it. For example:
```
BryceRussell, line 706: The `?raw` syntax does not return a string representation of the image's URL, when used with image formats like `.png`, `.webp`, etc it will return binary
```suggestion
In Astro v3.0, if you find yourself wanting to preserve the old import behavior for images and require a string representation of the image's URL, consider appending `?url` to the end of your image path when importing it. For example: 
```

dreyfus92: woops, my bad. I actually both were meant for the same purpose, I will apply this change asap.

src/content/docs/en/guides/images.mdx
```diff
@@ -700,3 +700,24 @@ If you were using the image integration in Astro v2.x, complete the following st
...
+
+  <svg>
+    <use xlink:href={Sprite + '#cart'} />
+  </svg>
+
+```
+
+This approach ensures you obtain the URL string. Keep in mind that during development, Astro uses a `src/` path, but upon building, it generates hashed paths like `/_astro/cat.a6737dd3.png`. 
+
+If you prefer to work directly with the image object, you can access the `.src` property. This approach can be especially handy for tasks like managing image dimensions for Core Web Vitals metrics.
```
ElianCodes, line 721:
```suggestion
If you prefer to work directly with the image object itself, you can access the `.src` property. This approach is best for tasks like managing image dimensions for Core Web Vitals metrics and preventing CLS. 
```

src/content/docs/en/guides/images.mdx
```diff
@@ -700,3 +700,24 @@ If you were using the image integration in Astro v2.x, complete the following st
...
+    <use xlink:href={Sprite + '#cart'} />
+  </svg>
+
+```
+
+This approach ensures you obtain the URL string. Keep in mind that during development, Astro uses a `src/` path, but upon building, it generates hashed paths like `/_astro/cat.a6737dd3.png`. 
+
+If you prefer to work directly with the image object, you can access the `.src` property. This approach can be especially handy for tasks like managing image dimensions for Core Web Vitals metrics. 
+
+For those transitioning to the new import behavior, combining `?url` and `.src` methods might be the right spell for seamless image handling.
```
ElianCodes, line 723:
```suggestion
If you are transitioning into the new import behavior, combining `?url` and `.src` methods might be the right method for seamless image handling.
```

src/content/docs/en/guides/images.mdx
```diff
@@ -700,3 +700,24 @@ If you were using the image integration in Astro v2.x, complete the following st
 You can now declare an associated image for a content collections entry, such as a blog post's cover image, in your frontmatter using its path relative to the current folder.
 
 The new `image` helper for content collections lets you validate the image metadata using Zod. Learn more about [how to use images in content collections](/en/guides/images/#images-in-content-collections)
+
+### Navigating Image Imports in Astro v3.0
+
+In Astro v3.0, if you find yourself wanting to preserve the old import behavior for images and require a string representation of the image's URL, consider appending `?url` to the end of your image path when importing it. For example:
```
ElianCodes, line 706:
```suggestion
In Astro v3.0, if you have to preserve the old import behavior for images and require a string representation of the image's URL, append `?url` to the end of your image path when importing it. For example: 
```

ElianCodes (commented):
Hey @dreyfus92! Thanks for tackling this!

Left a couple suggestions for you to review! (had to make it somewhat more boring)

dreyfus92:
Thank you so much for your reviews @BryceRussell & @ElianCodes 😁

PR #4463, opened 2023-08-31
Remove reference to deprecated build.split

paulrudy:
- New or updated content

#### Description

- `build.split` has been moved to [`functionPerRoute`](https://docs.astro.build/en/reference/adapter-reference/#functionperroute) in SSR adapters in Astro v3. This PR removes the `build.split` section of the docs
- It also moves some useful information (file path visualization and info about `outDir`) from the deprecated section to the `functionPerRoute` section (note! Please double check that the information as added is correct!)


Alynva (changes requested):
You didn't imported `FileTree`. I think the import in `server-side-rendering.mdx` can be removed too.

paulrudy:
thanks! made both those changes

ematipico (approved):
Looks good! Thank you for catching this!

PR #4466, opened 2023-08-31
Update docker.mdx - Nginx 404 config

guiguir68:

- New or updated content

#### Description
Improved Nginx configuration to handle and display a 404 error gracefully. 

Previously, the existing Nginx configuration was causing a 500 error when we attempted to access a non-existing page. With this update, Nginx will serve a custom 404.html page from the Astro site to provide a better user experience.


ElianCodes (approved):
Super useful fix! Thanks @guiguir68!

Let's get it merged

PR #4477, opened 2023-08-31
Add notes about old-translations on i18n tracking dashboard.

Genteure:

- Something else!

#### Description

Added a button next to the "Create page" button when a matching file in `old-translations` is found.
Files in `old-translations` are still valuable to translators as they could contain a lot of re-usable paragraphs. A "View old translation" button reminds translators the existence of such files, which will help prevent duplicate work.

image


ElianCodes (approved):
This is amazing! Thank you so much for this @Genteure !

TheOtterlord (approved):
Thanks for adding this! :partying_face:

Yan-Thomas (approved):
This is a great addition, thanks so much for this @Genteure, LGTM!

PR #4497, opened 2023-09-01
chore: revert Node to 18

ElianCodes:
Revert Node back to 18 for Netlify


delucis:
Do we also want to update the GitHub actions? Or are we OK having them on Node 20 until we see issues?

PR #4504, opened 2023-09-01
Fixed grammar error at islands.mdx (en-only)

IfeOlulesi:

#### Description
- Corrected grammar error in indefinite article usage (an should be a)


dreyfus92 (approved):
Nice catch @IfeOlulesi, LGTM ✅

ElianCodes:
Actually, using "an" here, is fine, since the pronunciation starts with a vowel "an espeeay" not "a espeeay"

But also checking with @sarah11918, to make sure I'm correct

src/content/docs/en/concepts/islands.mdx
```diff
@@ -40,7 +40,7 @@ import MyReactComponent from '../components/MyReactComponent.jsx';
 <MyReactComponent />
 ```
 
-But sometimes, client-side JavaScript is required for creating interactive UI. Instead of forcing your entire page to become an SPA-like JavaScript application, Astro asks you to create an island.
+But sometimes, client-side JavaScript is required for creating interactive UI. Instead of forcing your entire page to become a SPA-like JavaScript application, Astro asks you to create an island.
```
sarah11918, line 43:
```suggestion
But sometimes, client-side JavaScript is required for creating interactive UI. Instead of forcing your entire page to become a single-page JavaScript application (SPA), Astro asks you to create an island.
```

sarah11918 (approved):
Yes, Elian is correct. The correct word depends on the pronunciation which will be different for each reader. Some read "spa" as an entire word, and some read out each individual letter.

Normally, therefore, we try to avoid the problem and rewrite the sentence entirely! :joy:

I've done that below to avoid the issue. Thanks for raising it, as it can be awkward for people who read it "the other way" than the way it is written. This lets us make it better for everyone!  🙌 

Welcome to Team Docs!  🥳

IfeOlulesi:
Oh wow!!

That's a really amazing way to solve the problem😁

Many thanks @ElianCodes & @sarah11918 

Glad to be part of the team!!

maiieul:
@ElianCodes Actually I like "an espeeay". Would improve the DX imho 😁

PR #4515, opened 2023-09-02
guides/import: Add `.src` to `src={imgReference.src} `

tordans:
Without this change, I get the following error:

```
Type 'ImageMetadata' is not assignable to type 'string'.ts(2322)
astro-jsx.d.ts(701, 3): The expected type comes from property 'src' which is declared here on type 'ImgHTMLAttributes'
```

It is also how it is documented in https://docs.astro.build/en/guides/images/#images-in-astro-files

### DIFF
--- a/src/content/docs/en/guides/imports.mdx
+++ b/src/content/docs/en/guides/imports.mdx
@@ -133,7 +133,7 @@ import svgReference from './image.svg'; // svgReference === '/src/image.svg'
 import txtReference from './words.txt'; // txtReference === '/src/words.txt'
 
 // This example uses JSX, but you can use import references with any framework.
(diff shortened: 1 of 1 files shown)


tordans:
On a separate note, I wonder why this text
> // This example uses JSX, but you can use import references with any framework.

references JFX explicity. The example it not valid JSX AFAIK, because there is no function wrapper and no return statement. Wouldn't it be to write it as a `.astro` component which would only require to add the `---` part…?

ElianCodes (approved):
Good catch! This indeed has to be updated. Big thanks @tordans!

On the JSX note; I think we should keep it this way, since this is just valid JavaScript, we can always revisit to make it a `.astro` file, but for now, I would keep it like it is.

PR #4522, opened 2023-09-02
Update server-side-rendering.mdx

kmelve:

#### Description

This commit makes the code example for adapters run when copy-pasted.

The current code example won't work as-is and throws a “Setting the 'mode' option is required.” error without any suggestions for resolution. Readers then must realize that they must go to docs for the nodejs adapter and figure out what `mode` is.


ElianCodes (approved):
Thanks for this addition @kmelve!

PR #4523, opened 2023-09-02
Fixing Docker multi-stage build description

sentisso:

#### Description

I previously created PR #4191, which has since been merged. However, we overlooked a minor error in the description, which states:
> multi-stage builds, optimizes the build process for your site by only reinstalling the npm dependencies when the source code changes

but this statement is not correct! The advanced Dockerfile optimizes the build process by **not reinstallling** the npm dependencies when only the source code changes.

This PR fixes that description.


ElianCodes (approved):
yes! This is indeed a more accurate description! Thanks @sentisso !

PR #4526, opened 2023-09-03
Update recipes.mdx

teinett:
Added link to video: Learn how to add fuzzy searching to an SSG Astro site using FuseJS and client-size JavaScript.


ElianCodes (approved):
This is indeed a good resource to add! Thanks @teinett !

PR #4538, opened 2023-09-04
Use Fragments

sakgoyal:

#### Description

* Changed the `Add integrations and dependencies` section to use Fragments like the previous section. 
* This should also automatically change all tabs/fragments to the same package manager


sarah11918 (approved):
Works perfectly, @sakgoyal !  🚀  Thank you, and welcome to Team Docs!  🥳

PR #4545, opened 2023-09-05
Fix the missing highlight for example codes

100gle:

#### Description

Replace `diff` suffix in code block to `js` and set right line number for highlight.


sarah11918 (approved):
Great fix! Thanks @100gle ! 🚀

PR #4568, opened 2023-09-06
deploy/github: Document workaround for astro 3 bug

tordans:
The docs don't work ATM, https://github.com/withastro/astro/issues/8352#issuecomment-1703768414 shows the fix.

### DIFF
--- a/src/content/docs/en/guides/deploy/github.mdx
+++ b/src/content/docs/en/guides/deploy/github.mdx
@@ -20,7 +20,7 @@ Astro maintains the official `withastro/action` to deploy your project with very
 
     export default defineConfig({
       site: 'https://astronaut.github.io',
-      base: '/my-repo',
+      base: import.meta.env.DEV ? '' : '/my-repo',
     })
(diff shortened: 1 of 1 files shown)


ElianCodes (approved):
Good catch! Thanks for adding this @tordans!

sarah11918:
Hi @tordans ! As you'll be able to see if you go back and check the issue you linked, thanks to your drawing attention to it we've triaged the issue as an important one to fix, so we're on it!

We don't document workarounds in docs, and we keep the docs to be what the devs expect and intend to happen. Instead, when Astro is "misbehaving" compared to the docs, that's a sign we have an issue to fix. That issue comment you linked is the best way to have workarounds documented! That way, people can find a temporary fix but also, they know it's part of a bug that will eventually be fixed so they don't get too attached to it.

So I'm going to close this as not something we document, but instead a bug to be fixed, but thank you for wanting to contribute to help others out! We appreciate your thinking of the community! 🙌

PR #4576, opened 2023-09-06
Update build-forms-api.mdx for React example

sunapi386:

#### Description
The onSubmit event handler expects a FormEvent<HTMLFormElement>. This changes the event parameter type to FormEvent<HTMLFormElement> to fix the issue. 

Also uses the good practices for `autoComplete`, and the `for` in label as detailed here https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete


sarah11918 (approved):
Thank you for these helpful additions to the recipe!

PR #4597, opened 2023-09-08
Change extension to .cjs to align with documentation

AitorMT:

- Translated content

#### Description

Changed from ``postcss.config.js`` to ``postcss.config.cjs`` extension for consistency with documentation in other languages.


dreyfus92 (commented):
Hello @AitorMT, nice catch. There are locales that have `.cjs` extension still, dunno if we should change this to `.js` just the en version since we have the tracker or open individual PRs to apply this change. Gonna tag @Yan-Thomas to see what he thinks about it.

AitorMT:
Okey, I'll wait and see, sorry for calling you to check again, I made a mistake with the button.

sarah11918:
Just reminding everyone here that the translations SHOULD match exactly what the English docs say!

We don't decide things like this in the translation PRs. If when translating you think the English is wrong, then it's time to PR the English docs!  ;)  Thanks!  🙌

PR #4594, opened 2023-09-08
Fix: Flash of Dark Theme

apz-gh:
**Note:** This PR is a temporary fix until the issue with `setTheme` is fixed. 

**Current Status:** 

- On first load or when refreshing the browser, `setTheme` isn't setting the `theme` value.
- This leads to `theme` being 'undefined', causing the `.theme-dark` class to be added. 
- This PR doesn't solve the `setTheme` problem directly. Instead, it checks to ensure the theme is set to either "dark" or "light". 

- CleanShot 2023-09-08 at 08 50 46@2x 
- CleanShot 2023-09-08 at 08 51 29@2x
 
#### What kind of changes does this PR include?

- Something else:
    - Minor tweak to prevent the flash of the dark theme. 
    
#### Description: 

- Added a check to ensure the theme value is 'light' or 'dark', providing a temporary solution to the `setTheme` issue in `useEffect`.

#### Before / After 

- Did you change something visual

**Before:**

https://github.com/withastro/docs/assets/20273871/9006143e-c79f-40f5-822d-8308e706551e

https://github.com/withastro/docs/assets/20273871/a77536e8-7661-48d4-bdde-1161aa5807fa

https://github.com/withastro/docs/assets/20273871/41b08978-ab47-49b0-a02a-a96e7cfdab54

https://github.com/withastro/docs/assets/20273871/26235901-6412-48e4-9b48-2d363d9f0a05

https://github.com/withastro/docs/assets/20273871/65f64f15-21a1-40ed-9bdc-2774907182a0

**After:**

https://github.com/withastro/docs/assets/20273871/627849e7-7d14-4511-aaab-7dcdb8181c1d

PR #4602, opened 2023-09-08
Avoiding undefined initial state in theme hook

tony-sull:

#### Description

Quick follow-up to #4594

Turns out TypeScript wasn't screaming about a type mismatch in the `useState` hook :eyes:

I would have expected an error here when a default value of `light` or `dark` isn't provided, I could see that being super tricky to properly type with generics though!

This updates the `useState` call to default to the light theme, avoiding the state where the theme value was `undefined` and accidentally triggering a flash of dark mode

### DIFF
--- a/src/components/Header/ThemeToggleButton.tsx
+++ b/src/components/Header/ThemeToggleButton.tsx
@@ -41,7 +41,7 @@ const icons = [
 ];
 
 const ThemeToggle = ({ labels, isInsideHeader }: Props) => {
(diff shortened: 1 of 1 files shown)


lilnasy (changes requested):
Flashes for dark mode users now.

https://github.com/withastro/docs/assets/69170106/913e781d-3a11-4096-9a0f-df46a040d43e

tony-sull:
Hah, ah shoot so it does.  I guess undefined is the "right" unknown state there, will close this out for now but may look for a different way to clean this up

PR #4608, opened 2023-09-09
fix: typo in port number

itsmatteomanf:

#### Description

Port is 4321, not 4322 😄


sarah11918 (approved):
Thank you! I bet I did this update by hand, not as part of find and replace... :joy:

itsmatteomanf:
> Thank you! I bet I did this update by hand, not as part of find and replace... 😂

It was so minor, but so weird in the same row as one that was correct 🤣

PR #4609, opened 2023-09-09
Add `i18n` label to translatable labels

Yan-Thomas:
- Translated content

#### Description

Adds new glob paths so that the files within `src/i18n/(lang)` are also given the `i18n` label.


sarah11918 (approved):
I ask, and I receive!  💜

PR #4614, opened 2023-09-10
docs/bun: Add link to external installation guide

tordans:
This makes it easy to look for the homebrew installation steps for example.


sarah11918:
Hey @tordans , see my comment from the other PR: we entirely rewrote this page today for the new Bun 1.0, so this PRs an old version of the page. You're absolutely welcome to check out the new version of this page live in the docs and see if this is still something you'd suggest. 🙌

PR #4615, opened 2023-09-10
docs/bun: Clarify why some commands need the `run`

tordans:
I stumbled on the fact that only the build was `bun run build` and everything else is just `bun`. I assume this page is the first time using bun for people like it is for me, therefore some basics are good to add.
The solution is simple: Some commands on the page use reserved keywords like `install`, others do not need the run because `bun` defaults to `bun run` whenever it is not a reserved keyword, which is why documenting `bun preview` works. So it is just the `bun run build` which is the outlier here… which is what I tried to clarify in the docs with this PR.

Wording might very well be something to improve.

---

Also ping https://github.com/oven-sh/bun/pull/4756

### DIFF
--- a/src/content/docs/en/recipes/bun.mdx
+++ b/src/content/docs/en/recipes/bun.mdx
@@ -74,12 +74,12 @@ Some integrations might not work as expected. If you find any issues, please [op
 To build and serve your site, Bun has familiar commands:
 
 ```bash
-bun run build
+bun run build # note: it is `bun run` here because `bun build` is an internal bun feature
 ```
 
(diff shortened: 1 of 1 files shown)


jacobthesheep:
These can be in the same pr https://github.com/withastro/docs/pull/4615

sarah11918:
Hi @tordans! We actually just redid this page entirely because it needed updating after Bun's 1.0 announcement!

Feel free to take a look through the page and see whether it addresses your concern, but I'm going to close this PR because it's to an older version of the page.  🙌

PR #4630, opened 2023-09-10
i18n(es) Redundant description

AitorMT:

- New or updated content

#### Description

The deleted text is because it is redundant information, in the following paragraph, which is the one that is maintained, it gives the same information and is more complete. If it is kept, it would be the same in the first paragraph as in the second.


dreyfus92 (approved):
Nice catch this wasn't updated/removed as the `en` version of the docs, thank you for this @AitorMT. LGTM ✅

nabu1615 (approved):
It looks good to me, the text is redundant and it is not necessary.

Waxer59 (approved):
LGTM ✔

Yan-Thomas (approved):
LGTM! Nice catch @AitorMT 🙌

PR #4634, opened 2023-09-10
fix Api call for retrieving stories

roberto-butti:
For retrieving stories, instead of `cdn/stories/blog`, you need to call `cdn/stories`

#### Description

Fix a typo for calling api


sarah11918 (approved):
Thanks for spotting this, @roberto-butti ! I know we recently made a change for a possible case using an extra `/blog/` folder, but you're right; it should not have snuck in here.  Welcome to Team Docs!  🥳

PR #4647, opened 2023-09-11
Added the missing `Since` component

100gle:

- New or updated content
- Translated content

#### Description

Use `<Since />` component to replace `Added in` text


sarah11918:
Hi @100gle - this is the same issue as the other PR that edited the Integration guides. We just can't use our own docs formatting in these files.  So, these have to stay as proper Markdown format, even if it's not as nice for us in Docs. Booooo.  :(

PR #4653, opened 2023-09-11
Add note about installing Sharp when using Bun install

markjaquith:

- New or updated content

#### Description

Bun cannot currently install sharp properly, so installing Sharp using `npm` is currently necessary. This adds a note documenting what you need to do to get Sharp working when using `bun install` to install your packages.

### DIFF
--- a/src/content/docs/en/reference/errors/missing-sharp.mdx
+++ b/src/content/docs/en/reference/errors/missing-sharp.mdx
@@ -18,6 +18,8 @@ import DontEditWarning from '~/components/DontEditWarning.astro'
 ## What went wrong?
 Sharp is the default image service used for `astro:assets`. When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like pnpm, Sharp must be installed manually into your project in order to use image processing.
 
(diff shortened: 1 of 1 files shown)


jacobthesheep:
https://github.com/oven-sh/bun/issues/4549

sarah11918:
Hi @markjaquith!  Thank you for wanting to inform users about complications with sharp when using bun!

These error references are not maintained in this docs repo, though. Our error messages are generated in the core astro repo and automatically pulled in here. So we do not edit these files directly. (When you edit one of these kinds of pages, you should see a warning at the top of the file directing you to the correct place to edit the source content.)

Also, we typically don't document workarounds or temporary fixes, because then we would need to keep track and revisiting comments like this. As this is a bun error, a bun user should expect to look for and find issues with bun in the bun repo (as has been linked here!)

This error currently documents our own usage restrictions, but we would not document another projects temporary errors. The error message as it stands still provides the correct guidance: sharp must be manually installed. Potentially, we could generalize this message to emphasize that root cause (whatever the reason) and give the pnpm as an example of a (common Astro case) where this manual install is necessary.

If you'd like to reopen a PR to the main Astro repo to edit the error message at source, we can consider more generic wording there, but we will not be updating this message to mention bun specifically. Thanks again for wanting to share your knowledge! 🙌

PR #4654, opened 2023-09-11
Fix labeler not labeling i18n PRs properly

hippotastic:

#### Description

- Fixes a labeler issue that prevented any i18n PRs from being labeled.
- Bug was accidentally introduced in https://github.com/withastro/docs/pull/4609 due to a misunderstanding of how the "any" logic works. I checked the action's source code to debug the issue and found this solution.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -3,5 +3,8 @@
 '🚨 action':
 - .github/workflows/**
 
+# Note: The "any" logic requires ANY file of the PR to match ALL (!) of the globs in the array,
+# so to match files in multiple different paths, we also need multiple "any" arrays.
 i18n:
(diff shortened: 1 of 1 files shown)


delucis (approved):
Ahhhh, so that’s how it’s supposed to work 😅 

Thanks for digging into the source code to figure that out @hippotastic!

Yan-Thomas:
> Ahhhh, so that’s how it’s supposed to work 😅

same, I guess everyone fell into this trap 😓

Yan-Thomas (approved):
LGTM! Thanks @hippotastic 🙌

hippotastic:
No worries! I simply couldn't believe that this is not possible. :)

PR #4657, opened 2023-09-11
[en-only] Clarify usage of npm script flags and `astro info` output

jacobthesheep:

- New or updated content

#### Description

- Rephrases the note about using '--' to pass flags to npm scripts.
- Displays the new expected `astro info` output that was updated in [3.0.5](https://github.com/withastro/astro/releases/tag/astro%403.0.5)


sarah11918:
Hey @jacobthesheep!

Heads up that we ask that translations not be included with English changes in the same PR, because then we can't merge them, until we also have a Spanish review... 😅 

This one also appears to make some changes in English that it doesn't in Spanish, so this would make it look to our system that the Spanish is fully up to date when doesn't seem like this would be the case?

It also makes it harder for the translators to KNOW if there is material to review in a PR that touches multiple languages, because it doesn't conform to our regular PR title system.

The normal process is, we take a PR for the English change (we need to edit them... in which case, now the translation isn't even useful anyway) and only once that's approved and merged, then our i18n system is prepared to notify translators there has been an (accepted, final) change, and a translated-only PR can be made that is obvious to other translators that a language review is needed.

So, please remove the changes to the corresponding Spanish file here so that we can make this follow our normal procedure! Thanks!

jacobthesheep:
> Hey @jacobthesheep!
> 
> Heads up that we ask that translations not be included with English changes in the same PR, because then we can't merge them, until we also have a Spanish review... 😅
> 
> This one also appears to make some changes in English that it doesn't in Spanish, so this would make it look to our system that the Spanish is fully up to date when doesn't seem like this would be the case?
> 
> It also makes it harder for the translators to KNOW if there is material to review in a PR that touches multiple languages, because it doesn't conform to our regular PR title system.
> 
> The normal process is, we take a PR for the English change (we need to edit them... in which case, now the translation isn't even useful anyway) and only once that's approved and merged, then our i18n system is prepared to notify translators there has been an (accepted, final) change, and a translated-only PR can be made that is obvious to other translators that a language review is needed.
> 
> So, please remove the changes to the corresponding Spanish file here so that we can make this follow our normal procedure! Thanks!

Done.

jacobthesheep:
> D'oh! Sorry @jacobthesheep that this one slipped my radar because it still contains a Spanish document, which means it still has the `i18n` label and gets flagged as a translation (which I normally don't review myself!)
> 
> Sorry to ask again, but can you remove the Spanish page please? 😅
> 
> Otherwise, this is looking good, and appreciate you catching the formatting glitches. I'll need to verify with our devs whether that output change is what they'd expect (whether in fact it's now just "None"), but otherwise we'll get this one merged in real soon! 🙌

Done!

sarah11918:
Alrighty! Thanks for staying on the line... devs have confirmed and we are all systems go!  🚀

sarah11918 (approved):
Thanks @jacobthesheep for updating the `astro info` output! This is now `astro`ut of this world!

PR #4658, opened 2023-09-11
i18b(es): Update guides/server-side-rendering.mdx

nabu1615:

- New or updated content

#### Description

Update translation for src/content/docs/es/guides/server-side-rendering.mdx

### DIFF
--- a/src/content/docs/es/guides/server-side-rendering.mdx
+++ b/src/content/docs/es/guides/server-side-rendering.mdx
@@ -147,7 +147,7 @@ Astro seguirá siendo un generador de sitios estáticos por defecto. Pero una ve
 
 ### Configuración de rutas individuales
 
-Ambos modos `server` e `hybrid` permiten páginas y endpoints renderizados en el servidor y renderizarán todas las páginas de acuerdo a su comportamiento por defecto. Sin embargo, ambos modos permiten marcar una página individual para optar por no utilizar este comportamiento por defecto.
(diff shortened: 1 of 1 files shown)


dreyfus92 (commented):
Hey @nabu1615 thanks for jumping in, but I don't think these changes are necessary since they match the English version of the docs.

nabu1615:
> Hey @nabu1615 thanks for jumping in, but I don't think these changes are necessary since they match the English version of the docs.

It looks confused for me, the idea about help with the translation tasks is more than just translate, the literal translation for this text does not match a logic idea that's becase it needs help from native speakers.

Waxer59:
Thank you very much for the PR, I agree with @dreyfus92 and think these changes are not necessary. However, if you are interested in continuing to contribute there is a page where you can see the [pages available for translation](https://i18n.docs.astro.build/). Where there is currently [one available](https://github.com/withastro/docs/blob/main/src/content/docs/en/recipes/add-yaml-support.mdx).

nabu1615:
> Thank you very much for the PR, I agree with @dreyfus92 and think these changes are not necessary. However, if you are interested in continuing to contribute there is a page where you can see the [pages available for translation](https://i18n.docs.astro.build/). Where there is currently [one available](https://github.com/withastro/docs/blob/main/src/content/docs/en/recipes/add-yaml-support.mdx).

Ok, thank you, I will take a look, I will close this PR.

PR #4663, opened 2023-09-12
fix: tutorial navigation in i18n pages

moneychien19:

#### Description

In tutorial pages which are not translated yet, tutorial tracker will not work as expected.
- This is because the navigation component determine whether it’s in current unit by `en` URL.
- This PR fix it by stripping out language factor.


Yan-Thomas (approved):
Nice fix @moneychien19, thanks for this! And sorry for the wait 🙌 

Welcome to Team Docs 🥳

PR #4667, opened 2023-09-12
[rendering-modes] New page!

sarah11918:
#### Description

This:
- [X] adds a new page called Rendering Modes **replace the intro content on the SSR page**
- [X] updates the sidebar nav to show Rendering Modes 
- [X] moves the existing content about improving page performance via streaming to a separate Recipe
- [x] updates links on other docs pages if/as necessary depending on whether we want to link to a conceptual page or adding an adapter (also, probably updated headings even if pointing to the same page)

For future follow up (because it will touch a lot of pages):
-  replace existing text "SSR" (e.g. SSR mode, SSR adapters) to "server", "server-rendered", "on-demand rendered" etc. per the context.
- check for older docs that only refer to `server` mode, because that's all there was when they were written!
- consider renaming this page

PURPOSE: 

- The new "basics" page describes Astro's rendering modes (`output: 'static'`, `output: 'hybrid'`, and `output: 'server'`) together all on a new page titled Rendering Modes. 
- This and a revamped SSR page (file-name to potentially change to on-demand-rendering but later, for less disruption) also leans into the phrases "pre-rendering" and "on-demand rendering" while downplaying SSG/SSR

Previously, we had 
- no real information at all about static, pre-rendering, because that's just what Astro "was".
- only an SSR page that was not significantly refreshed since it was an experimental feature, and had grown by patchwork additions rather than being a united page

Therefore, we often had to describe the difference between all these modes in support threads, including when and why you might choose to use them.

The emphasis here shifts from "a page about SSR" to a core concepts page describing and comparing the three `output` modes and then a page about enabling `server` and `hybrid` as an "add on"

Direct preview link to the new page: https://deploy-preview-4667--astro-docs-2.netlify.app/en/core-concepts/rendering-modes/ and revamped SSR page: https://deploy-preview-4667--astro-docs-2.netlify.app/en/guides/server-side-rendering/


src/content/docs/en/guides/rendering-modes.mdx
```diff
@@ -0,0 +1,396 @@
...
+
+When it's time to deploy an SSR project, you also need to add an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. Each adapter allows Astro to output a script that runs your project on a specific runtime.
+
+The following adapters are available today with more to come in the future:
+
+- [Cloudflare](/en/guides/integrations-guide/cloudflare/)
+- [Deno](/en/guides/integrations-guide/deno/)
+- [Netlify](/en/guides/integrations-guide/netlify/)
+- [Node.js](/en/guides/integrations-guide/node/)
+- [Vercel](/en/guides/integrations-guide/vercel/)
```
sarah11918, line 86:
```suggestion
You can find both [official and community SSR adapters in our integrations directory](https://astro.build/integrations/?search=&categories%5B%5D=adapters).
```

sarah11918:
FRIDAY CHAOS... splitting this into two pages to better suit the proposed new sidebar sections...

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -3,44 +3,64 @@ title: Server-side Rendering
...
 - Deploy your site to a host using an *adapter*.
 
-Consider enabling server-side rendering in your Astro project if you need the following:
+Consider enabling on-demand server rendering in your Astro project if you need the following:
 
-- **API endpoints**: SSR enables you to create specific pages that function as API endpoints for tasks like database access, authentication, and authorization while keeping sensitive data hidden from the client.
+- **API endpoints**: Create specific pages that function as API endpoints for tasks like database access, authentication, and authorization while keeping sensitive data hidden from the client.
 
-- **Protected pages**: If you need to restrict access to a page based on user privileges, you can enable SSR to handle user access on the server.
+- **Protected pages**: Handle user access on the server for restrict access to a page based on user privileges.
```
sarah11918, line 19:
```suggestion
- **Protected pages**: Restrict access to a page based on user privileges, by handling user access on the server.
```

src/content/docs/en/core-concepts/rendering-modes.mdx
```diff
@@ -0,0 +1,54 @@
+---
+title: "Rendering Modes"
+i18nReady: false
```
sarah11918, line 3:
```suggestion
i18nReady: true
```

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -99,137 +139,64 @@ You can add any of the [official Astro adapter integrations](/en/guides/integrat
...
----
-console.log(Astro.request.method) // GET (when navigated to in the browser)
----
-```
+<RecipeLinks slugs={["en/recipes/streaming-improve-page-performance"]}/>
 
-:::caution
-The features below are only available at the page level. (You can't use them inside of components, including layout components.)
+:::caution[HTML streaming]
+Features that modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) are only available at the **page level** when HTML streaming is enabled. (You can't use them inside of components, including layout components.) By the time Astro runs your component code, it has already sent the Response headers and they cannot be modified.
```
lilnasy, line 190: > when HTML streaming is enabled.

Is this the case. I don't know for sure, but I would be surprised if it is. Usually, the pattern has been that Astro maintains its streaming limitations even during static rendering.

sarah11918: Ah, so my takeaway is it's *correct* but not limited to SSR.

So just confirming we still think that the SSR page is the right place for this. (Even if not limited to SSR, is this a place where people will more likely run into that limitation with what they're trying to do with on-demand rendering?)

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -51,33 +71,53 @@ export const prerender = true;
...
-### Adding an Adapter
+### Add an Adapter
 
-When it's time to deploy an SSR project, you also need to add an adapter. This is because SSR requires a server _runtime_: the environment that runs your server-side code. Each adapter allows Astro to output a script that runs your project on a specific runtime.
+To deploy a project in `server` or `hybrid` mode, you also need to add an **adapter**. This is because both of these modes require a server _runtime_: the environment that runs code on the server to generate pages when they are requested. Each adapter allows Astro to output a script that runs your project on a specific runtime, such as Vercel, Netlify or Cloudflare.
 
-You can find both [official and community SSR adapters in our integrations directory](https://astro.build/integrations/?search=&categories%5B%5D=adapters).
+You can find both [official and community adapters in our integrations directory](https://astro.build/integrations/?search=&categories%5B%5D=adapters). Choose the one that corresponds to your [deployment environment](/en/guides/deploy/).
 
 #### `astro add` Install
```
ElianCodes, line 116: Should this be capitalized?

```suggestion
#### `astro add` install
```

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -270,108 +241,47 @@ if (!product) {
...
-    <p>{factPromise}</p>
-  </body>
-</html>
-```
 
-In this example, `A name` will render while `personPromise` and `factPromise` are loading.
-Once `personPromise` has resolved, `A fact` will appear and `factPromise` will render when it's finished loading.
+{/* 
+
+ */}
```
ElianCodes, line 287: I think we can delete this empty comment?

```suggestion
```

ElianCodes (approved):
Seems like all of the questions and notes I had about these pages during talking and doc'ing are taken care of!

Really good pages!

src/content/docs/en/guides/server-side-rendering.mdx
sarah11918: Another Sarah Dark Pattern 👻

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -99,137 +139,64 @@ You can add any of the [official Astro adapter integrations](/en/guides/integrat
...
+With HTML streaming, a document is broken up into chunks, sent over the network in order, and rendered on the page in that order. In `server` or `hybrid` mode, Astro uses HTML streaming to send each component to the browser as it renders them. This makes sure the user sees your HTML as fast as possible, although network conditions can cause large documents to be downloaded slowly, and waiting for data fetches can block page rendering.
 
-### Opting-out of server-rendering
+<RecipeLinks slugs={["en/recipes/streaming-improve-page-performance"]}/>
 
-For a mostly server-rendered app configured as `output: server`, add `export const prerender = true` to any page or route to pre-render a static page or endpoint:
+:::caution[HTML streaming]
+Features that modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) are only available at the **page level** when HTML streaming is enabled. (You can't use them inside of components, including layout components.) By the time Astro runs your component code, it has already sent the Response headers and they cannot be modified.
```
sarah11918, line 190:
```suggestion
Features that modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) are only available at the **page level**. (You can't use them inside of components, including layout components.) By the time Astro runs your component code, it has already sent the Response headers and they cannot be modified.
```

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -99,137 +139,64 @@ You can add any of the [official Astro adapter integrations](/en/guides/integrat
...
+### HTML streaming
 
-Both `server` and `hybrid` modes allow for server-rendered pages and endpoints and will render all pages according to their default behavior. However, both modes allow you to mark an individual page to opt-out of this default behavior.
+With HTML streaming, a document is broken up into chunks, sent over the network in order, and rendered on the page in that order. In `server` or `hybrid` mode, Astro uses HTML streaming to send each component to the browser as it renders them. This makes sure the user sees your HTML as fast as possible, although network conditions can cause large documents to be downloaded slowly, and waiting for data fetches can block page rendering.
 
-### Opting-out of server-rendering
+<RecipeLinks slugs={["en/recipes/streaming-improve-page-performance"]}/>
 
-For a mostly server-rendered app configured as `output: server`, add `export const prerender = true` to any page or route to pre-render a static page or endpoint:
+:::caution[HTML streaming]
```
sarah11918, line 189:
```suggestion
:::caution
```
now that there's a section heading, not needed on the aside itself

src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -99,137 +139,64 @@ You can add any of the [official Astro adapter integrations](/en/guides/integrat
...
----
-```
+<RecipeLinks slugs={["en/recipes/streaming-improve-page-performance"]}/>
 
 :::caution
-The features below are only available at the page level. (You can't use them inside of components, including layout components.)
+Features that modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) are only available at the **page level**. (You can't use them inside of components, including layout components.) By the time Astro runs your component code, it has already sent the Response headers and they cannot be modified.
 
-This is because these features modify the [Response headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header), which can't be modified after they're sent to the browser. In SSR mode, Astro uses HTML streaming to send each component to the browser as it renders them. This makes sure the user sees your HTML as fast as possible, but it also means that by the time Astro runs your component code, it has already sent the Response headers.
+Check your adapter's documentation to see whether HTML streaming is used. It may be the default setting, and may or may not be a configurable option.
```
sarah11918, line 192:
```suggestion
```

PR #4678, opened 2023-09-13
fix: generate rss.xml file by lowercasing get() function

Knorgias:

#### Description

After following the [rss-feed blog tutorial](https://docs.astro.build/en/tutorial/5-astro-api/4/#install-astros-rss-package) I came across an issue.

**Issue:**
When running `npm run build` then `npm run preview`, rss.xml file does not seem to be created, as /rss.xml on local returns a 404.

**Potential solution:**
This issue was solved after lowercasing GET()

from:
`export async function GET() {`
to:
`export async function get() {`

by finding and applying [koba-masa's suggestion on Astro discord channel](https://discord.com/channels/830184174198718474/1147825423594565752/1147825423594565752)

### DIFF
--- a/src/content/docs/en/tutorial/5-astro-api/4.mdx
+++ b/src/content/docs/en/tutorial/5-astro-api/4.mdx
@@ -78,7 +78,7 @@ Individuals can subscribe to your feed in a feed reader, and receive a notificat
 
     import rss, { pagesGlobToRssItems } from '@astrojs/rss';
 
-    export async function GET() {
+    export async function get() {
       return rss({
         title: 'Astro Learner | Blog',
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Thanks for this, @Knorgias  ! 

Just to let you know, the underlying problem is that the RSS generating method needs updating for v3.0, which WILL eventually use the capital `GET`. But you're right, while we continue to have the older instructions from v2.0, it needs to be the lower case `get`. So don't be surprised if you come back in a few weeks and we've updated this whole page again to use `GET`, because that is *eventually* our long term solution! 

But so that it works for people until then, this is a great help!  🙌  Welcome to Team Docs!  🥳

PR #4685, opened 2023-09-14
[en-only] Make cli-reference page cleaner

jacobthesheep:

- New or updated content

#### Description

- Updated all the command sections to display their available flags
- Replaced the common flags sections by turning them into a list
- Removed the global flags section as the description is already provided at the top of the page


sarah11918:
Hi! We're working on this page internally during our content audit/rewrite right now, so we're not accepting external contributions other than error fixes to this page right now!

PR #4709, opened 2023-09-15
fix ahref audit issues

FredKSchott:

#### Description
- Fix two ahref audit issues:
  - multiple `h1` elements on the page
  - bad link

### DIFF
--- a/src/components/tutorial/FeedbackButton.astro
+++ b/src/components/tutorial/FeedbackButton.astro
@@ -46,7 +46,7 @@ const instanceId = Math.round(Math.random() * 10e9).toString(36);
 				</svg>
 			</button>
 			<form>
-				<h1><UIString key="feedback.formTitle" /></h1>
+				<h4><UIString key="feedback.formTitle" /></h4>
 				<fieldset class="category-options">
 					<legend class="sr-only"><UIString key={'feedback.categoryGroupLabel'} /></legend>
 					{
--- a/src/components/tutorial/FeedbackButton.astro
+++ b/src/components/tutorial/FeedbackButton.astro
@@ -46,7 +46,7 @@ const instanceId = Math.round(Math.random() * 10e9).toString(36);
 				</svg>
 			</button>
 			<form>
-				<h4><UIString key="feedback.formTitle" /></h4>
+				<h1><UIString key="feedback.formTitle" /></h1>
 				<fieldset class="category-options">
 					<legend class="sr-only"><UIString key={'feedback.categoryGroupLabel'} /></legend>
 					{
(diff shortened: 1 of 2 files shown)


src/content/docs/ru/concepts/why-astro.mdx
```diff
@@ -122,7 +122,7 @@ Astro включает в себя синтаксис компонентов, р
 объединение в бандл, оптимизации, получение данных из вне и многое другое. 
 Вы можете создавать отличные веб-сайты, даже не выходя за рамки основного набора функций Astro. 
 
-Если вам нужно больше контроля, вы можете расширить Astro с более [100 интеграций](https://astro.build/en/integrations/) как
+Если вам нужно больше контроля, вы можете расширить Astro с более [100 интеграций](https://docs.astro.build/en/integrations/) как
```
ElianCodes, line 125: Seems like this is a broken link

```suggestion
Если вам нужно больше контроля, вы можете расширить Astro с более [100 интеграций](https://astro.build/integrations/) как
```

src/components/tutorial/FeedbackButton.astro
```diff
@@ -46,7 +46,7 @@ const instanceId = Math.round(Math.random() * 10e9).toString(36);
 				</svg>
 			</button>
 			<form>
-				<h1><UIString key="feedback.formTitle" /></h1>
+				<h4><UIString key="feedback.formTitle" /></h4>
```
ElianCodes, line 49: I think SEO-wise, the H1 should actually be valid, since it's inside a from in a dialog

ElianCodes (commented):
Left a couple suggestions for you to review!

src/components/tutorial/FeedbackButton.astro
FredKSchott: I see what you mean about the context of the modal it makes sense, but I believe headers are meant to be considered globally. I only know this because having multiple `h1` elements on the page confuses search engines like google, and is explicitly flagged as an error in audit tools like ahref.

ElianCodes: In that case; Keep the change you made!

sarah11918: I know Chris is OOO, but Chris' instinct was that this existing `<h1>` was actually fine too. I'm not the person to know, but just advising! Merge at your own risk!

sarah11918 (approved):
Approving with a `Chr`aveat! (a `Sw`arning?)

src/components/tutorial/FeedbackButton.astro
Yan-Thomas: > I only know this because having multiple h1 elements on the page confuses search engines like google

I am pretty sure this isn't the truth anymore. Here's [confirmation from Google](https://www.youtube.com/watch?v=WsgrSxCmMbM) and a [helpful article on this matter](https://www.greenlanemarketing.com/resources/articles/seo-101-seo-and-multiple-h1-tags/).

FredKSchott: TIL! That's fascinating, okay will revert this then, I like the semantic argument.

PR #4715, opened 2023-09-15
fix(errors): Fix error formatting not properly handling angle brackets

Princesseuh:

#### Description

With the new image component, we're adding more errors that have angle brackets in them (ex: `<Image />`). MDX doesn't like this, so we escaped them previously and it'd all work nicely, but then it seems like something changed and now it doesn't work. Not sure. This fixes everything, though.

This PR also updates the `packageManager` field, the old version of `pnpm` used didn't support Node 20.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "remark-smartypants": "^2.0.0",
     "sass": "^1.54.3"
   },
-  "packageManager": "pnpm@8.2.0",
+  "packageManager": "pnpm@8.6.12",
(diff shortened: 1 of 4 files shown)


sarah11918 (approved):
Thanks, @Princesseuh . The generated previews look great. I'll close https://github.com/withastro/astro/pull/8561 and let the new script take care of the formatting instead!

Yan-Thomas (approved):
LGTM!

PR #4726, opened 2023-09-16
docs: clarify difference in HTML and SVG attributes between Astro and JSX

yialo:

- New or updated content

#### Description


sarah11918:
The guidance we have in this section is correct, and what we would like to show here!

PR #4727, opened 2023-09-16
feat: Mention bun package manager support

ap-1:

- New or updated content

#### Description

- Since Bun was added as a supported package manager on the official Astro action (https://github.com/withastro/action/pull/26), specify `bun` as an additional detected package manager and `bun.lockb` as the respective lockfile to commit.


ElianCodes (approved):
Nice catch @ap-1! Thanks for adding this!

TheOtterlord (approved):
Thanks for adding this!

PR #4728, opened 2023-09-16
i18n(es): Update from-nuxtjs.mdx

AitorMT:

- New or updated content

#### Description

I have removed a block which is duplicated in the “Crea un nuevo proyecto Astro” section, as below is your updated version of the `PackageManagerTabs`.

Also fixed the example of importing a file, as it was wrong.


nabu1615 (approved):
Good catch, LGTM.

Yan-Thomas (approved):
Nice catch, thanks @AitorMT!

PR #4741, opened 2023-09-18
Fix integration script for non-core adapters

matthewp:

#### Description

- Removes `@astrojs/netlify` and `@astrojs/deno` from docs generation as those integrations are no longer controlled by core. They will be documented in their own repository, like with any other 3rd party package.

- Needed for https://github.com/withastro/astro/pull/8559
- Needed for https://github.com/withastro/astro/pull/8574


scripts/generate-integration-pages.ts
```diff
@@ -38,12 +38,21 @@ class IntegrationPagesBuilder {
...
 	readonly #i18nNotReadyIntegrations = new Set<string>([]);
 
 	constructor(opts: { githubToken?: string; sourceBranch: string; sourceRepo: string }) {
 		this.#githubToken = opts.githubToken;
 		this.#sourceBranch = opts.sourceBranch;
 		this.#sourceRepo = opts.sourceRepo;
+		this.#nonCoreIntegrations.set('netlify', {
+			packageName: '@astrojs/netlify',
+			repo: 'withastro/netlify-adapter'
+		});
```
delucis: Chatted a bit with @sarah11918 earlier today and agreed that any of these that remain “official” (i.e. under the `withastro` GitHub org and `@astrojs` npm scope) make sense to keep in docs. Not sure where all of this will land, but if some adapters end up with that status (like the Netlify one in its new location), then something like this change from the initial commit probably will make sense.

sarah11918:
Note to us to merge this when the affected repositories have been moved, and we'll make any necessary docs changes to content on our end at that time. 

(E.g. if there is no page remaining to pull in, our content will simply not update and continue to show what's current, so we'll have to either remove manually on our end, or update manually to show a redirect etc.)

src/content/docs/en/guides/integrations-guide/deno.mdx
```diff
@@ -1,188 +1,11 @@
...
-## Contributing
-
-This package is maintained by Astro's Core team. You're welcome to submit an issue or PR!
-
-## Changelog
-
-See [CHANGELOG.md](https://github.com/withastro/astro/tree/main/packages/integrations/deno/CHANGELOG.md) for a history of changes to this integration.
-
-[astro-integration]: /en/guides/integrations-guide/
+The Deno adapter allows Astro to deploy your SSR site to Deno targets including Deno Deploy. The Deno adapter was previously maintained by Astro but now is maintained by Deno directly. Usage is now documented [in the adapter repository](https://github.com/denoland/deno-astro-adapter).
```
sarah11918, line 11:
```suggestion
The Deno adapter allows Astro to deploy your SSR site to Deno targets including Deno Deploy. 

The Deno adapter was previously maintained by Astro but now is maintained by Deno directly. Usage is now documented [in the Deno adapter repository](https://github.com/denoland/deno-astro-adapter).
```

src/content/docs/en/guides/integrations-guide/deno.mdx
```diff
@@ -1,188 +1,13 @@
...
-#
-# TRANSLATORS: please remove this note and the <DontEditWarning/> component.
-
 type: integration
 title: '@astrojs/deno'
-description: Learn how to use the @astrojs/deno SSR adapter to deploy your Astro project.
-githubURL: 'https://github.com/withastro/astro/tree/main/packages/integrations/deno/'
+description: The Deno Astro adapter
+githubURL: https://github.com/withastro/docs/tree/main/src/content/docs/en/guides/integrations-guide/deno.mdx
 hasREADME: true
```
sarah11918, line 6: @delucis Should we set this to false if we are not pulling in a README but rather setting our own page manually?
```suggestion
hasREADME: false
```

scripts/generate-integration-pages.ts
```diff
@@ -37,13 +37,18 @@ class IntegrationPagesBuilder {
 	readonly #githubToken?: string;
 	readonly #sourceBranch: string;
 	readonly #sourceRepo: string;
-	readonly #deprecatedIntegrations = new Set(['turbolinks']);
+	readonly #deprecatedIntegrations = new Set(['turbolinks', 'deno']);
```
sarah11918, line 40: @delucis Does setting deprecated mean a page/integration icon won't be generated? Or is this safe to do given we still actually want it to show up?

sarah11918 (approved):
I'm approving the page text! This should wait for Chris to confirm some page generation questions I had.

scripts/generate-integration-pages.ts
delucis: This is only in the generation script, so tells the CI job to ignore future updates to the README.md in the Astro repo. I think that’s what we want given we’re now hand-writing the page in docs?

src/content/docs/en/guides/integrations-guide/deno.mdx
delucis: Hmm. We might need to tweak some site code here potentially.

We use `githubURL` to set the values for the **GitHub** and **Changelog** links here: 
Header for the @astrojs/deno page

So if we’re maintaining how those behave, we should probably make `githubURL` the Deno repo and keep `hasREADME: true` (`true` distinguishes integrations from the config reference that comes from the types file). BUT that will mean the edit link is wrong…

Alternatively, we remove `type: integration` so that this page looks like a standard page in docs without the integration-specific header, and remove `githubURL` and `hasREADME`. This would have the side effect of removing the Deno logo from the integrations nav component (and the integrations nav from the footer of this page — although we could easily add it back manually).

I was arguing on Discord for aiming to keep this around in the integrations nav for a bit, but if that requires some custom code just for this one page… maybe that’s not worth the effort 😅

src/content/docs/en/guides/integrations-guide/deno.mdx
```diff
@@ -1,188 +1,13 @@
...
-
-This package is maintained by Astro's Core team. You're welcome to submit an issue or PR!
-
-## Changelog
-
-See [CHANGELOG.md](https://github.com/withastro/astro/tree/main/packages/integrations/deno/CHANGELOG.md) for a history of changes to this integration.
+The Deno adapter allows Astro to deploy your SSR site to Deno targets including Deno Deploy. 
 
-[astro-integration]: /en/guides/integrations-guide/
+The Deno adapter was previously maintained by Astro but now is maintained by Deno directly. Usage is now documented [in the Deno adapter repository](https://github.com/denoland/deno-astro-adapter).
```
delucis, line 13: I’d be in favour of minimal migration docs here once we know the new package name. Something along the lines of:

1. Uninstall `@astrojs/deno`
2. Install `deno-astro-adapter` (or whatever it ends up being called)
3. Update the import statement in `astro.config.mjs` to use the new package:
   ```diff
   - import deno from '@astrojs/deno';
   + import deno from 'deno-astro-adapter';
   ```
   
But I can understand if we’re reticent to provide any guidance or want to keep that somewhere else.

sarah11918: I do like this! Let's do it!

scripts/generate-integration-pages.ts
sarah11918: sounds good then. Resolving!

src/content/docs/en/guides/integrations-guide/deno.mdx
sarah11918: OK, so yeah, it seems like removing `type: integration` is the path of least resistance! :sweat_smile: 

I propose we do that, it removes Deno from the list, and then the edit this page link should be correct, right? I'll ask you to propose the change that does this properly, and I'll handle adding migration instructions to the text.

src/content/docs/en/guides/integrations-guide/deno.mdx
sarah11918, line 13:
``````suggestion
The Deno adapter was previously maintained by Astro but now is maintained by Deno directly. Usage is now documented [in the Deno adapter repository](https://github.com/denoland/deno-astro-adapter).

## Migrate to the new adapter

If you are currently using this Astro adapter, you will need to migrate to the new Deno version.
 
1. Remove the existing adapter by [uninstalling the `@astrojs/deno` integration](/en/guides/integrations-guide/#removing-an-integration). 
2. Install `deno-astro-adapter` by following [Deno's installation instructions](https://github.com/denoland/deno-astro-adapter#installation).
3. Update the import statement in `astro.config.mjs` to use the new package:
   ```diff
   - import deno from '@astrojs/deno';
   + import deno from 'deno-astro-adapter';
   ```
``````
BIG NOTE: The instructions on their readme still refer to `@astrojs/deno` and include all the existing installation instructions. That needs to be updated over there!

Comment (matthewp) [reply src/content/docs/en/guides/integrations-guide/deno.mdx:13]: I think we're going to want to merge this *before* they have released a new adapter (which we don't know when is going to happen). So would it be ok to go with this for now and update the guidance once that's out?

**Comment by matthewp** [comment]
The home for Netlify might be changing, so holding off on this one for now, putting back as a draft.

**Comment by matthewp** [comment]
Ok, this is ready for review again. This is cleaned up so that only Deno is removed from generation. Netlify is now maintained in https://github.com/withastro/adapters/tree/main/packages/netlify and the plan is that other adapters will be moved over. I'll update the script again when those are moved over.

File: scripts/generate-integration-pages.ts [hunk 6]
```diff
@@ -62,38 +64,48 @@ class IntegrationPagesBuilder {
...
-						? 'renderer'
-						: keywords.includes('astro-adapter')
-						? 'adapter'
-						: 'other';
-					const i18nReady = (!this.#i18nNotReadyIntegrations.has(pkg.name)).toString();
-					const readme = await (await fetch(readmeURL)).text();
-					return { name, category, readme, srcdir: pkg.name, i18nReady };
 				})
 		);
 	}
```
delucis, line 111: `packages/integrations` needs replacing with the new `sourcePath` option in `#processReadme()` below (sorry GitHub won’t let me comment on that line).

src/content/docs/en/guides/integrations-guide/deno.mdx
```diff
@@ -1,188 +1,13 @@
 ---
-# NOTE: This file is auto-generated from 'scripts/generate-integration-pages.ts'
-#       and pulls content directly from the package’s README.
-#       DO NOT MAKE EDITS TO THIS FILE DIRECTLY, THEY WILL BE OVERWRITTEN!
-#       For corrections, please edit the package README at
-#       https://github.com/withastro/astro/tree/main/packages/integrations/deno/
-#
-# TRANSLATORS: please remove this note and the <DontEditWarning/> component.
-
 type: integration
```
delucis, line 2: We can remove this as proposed by @sarah11918 
```suggestion
```

src/content/docs/en/guides/integrations-guide/deno.mdx
```diff
@@ -1,188 +1,13 @@
...
 title: '@astrojs/deno'
-description: Learn how to use the @astrojs/deno SSR adapter to deploy your Astro project.
-githubURL: 'https://github.com/withastro/astro/tree/main/packages/integrations/deno/'
+description: The Deno Astro adapter
+githubURL: https://github.com/withastro/docs/tree/main/src/content/docs/en/guides/integrations-guide/deno.mdx
 hasREADME: true
 category: adapter
```
delucis, line 7: And we’ll also want to remove these:

```suggestion
```

delucis (commented):
Thanks @matthewp! Left a couple of notes.

Also mentioning that these changes mean Deno will no longer show up in our integrations nav:

SSR adapters navigation showing only Cloudflare, Netlify, Node, and Vercel adapters

I think that was clear from our discussion, but might be worth us checking where and how we want to highlight its existence in the future, so that people who want it can still find it.

sarah11918:
Following @delucis comment above re: now Deno doesn't show up on our list, I updated the SSR page with links to both the integrations directory on astro.build and to our official integrations guide page in docs.

src/content/docs/en/guides/integrations-guide/deno.mdx
sarah11918: @matthewp What about this as a compromise?

src/content/docs/en/guides/integrations-guide/deno.mdx
matthewp: Updated these, thanks!

src/content/docs/en/guides/integrations-guide/deno.mdx
matthewp: Removed

scripts/generate-integration-pages.ts
matthewp: Good catch, updated this.

delucis (approved):
Looks good — thanks @matthewp!

PR #4744, opened 2023-09-18
[i18nIgnore] Use bunx command instead of bun

Natetronn:

- minor typos

#### Description

When running `bun --bun astro [dev|build|preview]`, an infinite loop is created, where the command repeats itself over and over again. This requires CTRL+C to terminate.

`bunx`, on the other hand, works as expected.

Note: the bun guide for Astro uses `bunx`:
https://bun.sh/guides/ecosystem/astro

Aside: someone might consider updating the Astro/bun Starter Template package.json scripts accordingly.


Natetronn:
Update: this may be an issue with bun on my system. I updated bun and it still looped the command. I've removed bun and reinstalled and that seems to have _kind_ of fixed it, although I've been getting some mixed results. 

This issue may arise when running the scripts using VS Code's NPM Scripts running, since it first runs using `npm run [dev|start|build]` etc. before running the bun command before finally running `bunx`. But I haven't been able to confirm that or not, as I'm getting mixed results, like I said.

```
 *  Executing task in folder myproject: npm run dev 

> myproject@0.0.1 dev
> bun --bun astro dev

$ bunx astro dev <------------------------- note the extra command gets fired when using bun instead of bunx (see code below)
  🚀  astro  v3.1.0 started in 107ms
  
```

```
 *  Executing task in folder myproject: npm run dev 

> myproject@0.0.1 dev
> bunx --bun astro dev

  🚀  astro  v3.1.0 started in 100ms
  
```
I realize this isn't a trouble shooting thread, but I did want to be transparent about what I've found out since making this PR. Ultimately, from all I can tell, I believe using `bunx` over `bun` is the better option, since it runs once less conversion command (and was the command that was being repeated infinity.)

jacobthesheep:
> Update: this may be an issue with bun on my system. I updated bun and it still looped the command. I've removed bun and reinstalled and that seems to have _kind_ of fixed it, although I've been getting some mixed results.
> 
> This issue may arise when running the scripts using VS Code's NPM Scripts runner, since it first runs using `npm run [dev|start|build]` etc. before running the `bun` command before finally running `bunx`. But I haven't been able to confirm that or not, as I'm getting mixed results, like I said.
> 
> ```
>  *  Executing task in folder myproject: npm run dev 
> 
> > myproject@0.0.1 dev
> > bun --bun astro dev
> 
> $ bunx astro dev <------------- note the extra command gets fired when using bun instead of bunx, which sometimes repeats forever (note: the code below doesn't include this extra command, since it starts with bunx out the door.)
>   🚀  astro  v3.1.0 started in 107ms
>   
>   ┃ Local    http://localhost:4321/
>   ┃ Network  use --host to expose
> ```
> 
> ```
>  *  Executing task in folder myproject: npm run dev 
> 
> > myproject@0.0.1 dev
> > bunx --bun astro dev
> 
>   🚀  astro  v3.1.0 started in 100ms
>   
>   ┃ Local    http://localhost:4321/
>   ┃ Network  use --host to expose
> ```
> 
> I realize this isn't a trouble shooting thread, but I did want to be transparent about what I've found out since making this PR. Ultimately, from all I can tell, I believe using `bunx` over `bun` is the better option, since it runs one less conversion command; the command that was being repeated infinitely.

Please add `[i18nIgnore]` to the PR title.

jacobthesheep:
> #### What kind of changes does this PR include?
> * minor typos
> 
> #### Description
> When running `bun --bun astro [dev|build|preview]`, an infinite loop is created, where the command repeats itself over and over again. This requires CTRL+C to terminate.
> 
> `bunx`, on the other hand, works as expected.
> 
> Note: the bun guide for Astro uses `bunx`: https://bun.sh/guides/ecosystem/astro
> 
> Aside: someone might consider updating the Astro/bun Starter Template package.json scripts accordingly.

What do you mean by "Astro/bun Starter Template package.json"?

Natetronn:
> What do you mean by "Astro/bun Starter Template package.json"?
I mean when using this to create a site, the package.json scripts make no reference to bun commands:

`bunx create-astro@latest my-astro-project-using-bun`

So instead of `dev: astro dev`, as is the case now, perhaps it could `bunx --bun astro dev`. I realize create-astro@latest is going to be setup universally, however and that we'll need to change the commands on our own.

jacobthesheep:
> > What do you mean by "Astro/bun Starter Template package.json"?
> 
> I mean when using this to create a site, the package.json scripts make no reference to bun commands:
> 
> `bunx create-astro@latest my-astro-project-using-bun` - [source](https://docs.astro.build/en/recipes/bun/#create-a-new-astro-project-with-bun)
> 
> So instead of `"dev": "astro dev"` for example, as is the case now, perhaps it could be more bun specific: `"dev": "bunx --bun astro dev"`. I realize `create-astro@latest` is going to be setup universally, however and that we'll need to change the commands on our own.

I don't understand why you would update the script to be `"dev": "bunx --bun astro dev"` since you don't already do that with any of the other package managers. At least I don't because you need to use `bunx --bun` to run the `dev` script so It wouldn't make sense to include `bunx --bun` inside the script.

Natetronn:
> I don't understand why you would update the script to be `"dev": "bunx --bun astro dev"` since you don't already do that with any of the other package managers. At least I don't because you need to use `bunx --bun` to run the `dev` script so It wouldn't make sense to include `bunx --bun` inside the script.

VS Code's NPM Scripts (play button) feature doesn't currently support bun. The options are auto, npm or yarn. So it picks one (`npm run dev`, as seen in the code comments [here](https://github.com/withastro/docs/pull/4744#issuecomment-1724505425)) and then it runs the script command, which is why I had to add `bunx --bun astro dev` to the script as well. 

Even if it did support bun, it probably won't run bun with the `--bun` flag or use `bunx` instead of just `bun` and with the `--bun` flag, for that matter.

That said, the is a contextual issue, so I understand why it doesn't make sense to update it now. On that note, my initial aside was added before my update, where I discovered the root cause of the issue was related to running the script via VS Code NPM Scripts feature.

Does that make sense?

Ultimately, probably best to disregard that _aside_ at this point. Should we also disregard this whole PR as well? Or perhaps add a note about the issue in the docs for Astro/Bun or something (?) 

_I don't know, it's an issue that exists and was trying to help others avoid it, but I understand that it isn't exactly straight forward here._

jacobthesheep:
> > I don't understand why you would update the script to be `"dev": "bunx --bun astro dev"` since you don't already do that with any of the other package managers. At least I don't because you need to use `bunx --bun` to run the `dev` script so It wouldn't make sense to include `bunx --bun` inside the script.
> 
> VS Code's NPM Scripts (play button) feature doesn't currently support bun. The options are auto, npm or yarn. So it picks one (`npm run dev`, as seen in the code comments [here](https://github.com/withastro/docs/pull/4744#issuecomment-1724505425)) and then it runs the script command, which is why I had to add `bunx --bun astro dev` to the script as well, otherwise it would run `astro dev` using npm or yarn instead of `bunx --bun`.
> 
> bun-issue
> 
> Even if it did support bun, it probably won't run bun with the `--bun` flag or use `bunx` instead of just `bun` and with the `--bun` flag, for that matter.
> 
> That said, the is a contextual issue, so I understand why it doesn't make sense to update it now. On that note, my initial aside was added before my update, where I discovered the root cause of the issue was related to running the script via VS Code NPM Scripts feature.
> 
> Does that make sense?
> 
> Ultimately, probably best to disregard that _aside_ at this point. Should we also disregard this whole PR as well? Or perhaps add a note about the issue in the docs for Astro/Bun or something (?)
> 
> _I don't know, it's an issue that exists and my hope was to help others avoid it, if possible, but I understand that it isn't exactly straight forward here._

I understand now. But no yeah, keep the PR.

sarah11918:
D'doh! I'm now finding all these PRs that have been flagged as translations because multiple language files were changed. Sorry @Natetronn!

Normally, I'd ask that everything but the English page be removed, but I think we can sneak this one in since it's already got the magic title and the changes are quite contained. Even for code updates like this, we really do prefer each language crew make their own changes so that their code doesn't change from out from under their feet on them, and someone might be working on a page already locally... it's just more considerate for them!  😄  (And, as you've seen, it means that I actually notice that this exists!)

But, let's get this updated now that I've found it! Welcome to Team Docs!  🥳

PR #4756, opened 2023-09-19
Document triggering navigation

matthewp:

- New or updated content

#### Description

- Created a section "Router control" to explain how you can have control over the router.
- Added a section on manually triggering navigation.

- Closes Issue https://github.com/withastro/astro/pull/8571


martrapp (changes requested):
Excluding PDFs and images from transitions is no longer a good example, since now all non-html files are reloaded anyway as well as all html-files that do not have `<ViewTransitions />` in their head.

The way it is worded now, it might give the impression that people have to opt out.

We should also clearly state that *both* pages of the navigation must have `<ViewTransitions />` in their `<head>` to prevent reload. (*actually any element with a `name=astro-view-transitions-enabled` somewhere in the document*)

The use case for `astro-data-reload` would be a navigation to another html page, to which you would normally want transitions (why else would the target page have `<ViewTransitions />`?), but on this very link between those two pages there should be a full reload. Now I'm starting to run out of imagination. Are there still any other good examples of data-astro-reload?

In the above cases `data-astro-reload` might be a performance optimizer: The checks for `media-type` and `astro-view-transitions-enabled` can only be performed after the target document has been retrieved. If a reload is forced, it must be loaded again. In contrast, `data-astro-reload` is checked first when a link is clicked, and there is no double loading, which can be important if it takes a long time to generate and load the image that can not be cached.

matthewp:
@martrapp Ah yes, that example was added before the media type check was added, so it was the right way to do it at the time. I would probably still use the tag to prevent an unnecessary fetch + redirect, but I agree that there can be a better example. I'll update this.

martrapp (approved):
Love your examples!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,18 +235,45 @@ const myFade = {
...
 ```html
-<a href="/quarterly-earnings.pdf" data-astro-reload>
+<a href="/articles/emperor-penguins" data-astro-reload>
 ```
 
 These links will be ignored by the router and a full page navigation will occur.
 
+### Trigger navigation
+
+There may be occasions when you need to trigger a navigation imperatively, such as via an event that occurs within JavaScript. The `astro:transitions/client` module provides a method for doing this. In any client script or client component you can use the module like so:
```
sarah11918, line 261:
```suggestion
You can also trigger client-side navigation via events not normally listened to by the `<ViewTransitions />` router using `navigate`. This method from the `astro:transitions/client` module is available in any client script or client component.

The following example shows using an option selected by a site visitor to trigger navigation: 
```
Improve the description as needed, but we do always try to describe what the code is doing so that "like this" isn't left up to the reader!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,18 +235,45 @@ const myFade = {
...
+// Navigate to the selected option automatically.
+document.querySelector('select').onchange = (ev) => {
+  let href = ev.target.value;
+  navigate(href);
+};
+```
+
+The `navigate` method takes these arguments:
+
+- `href` - The path to the page you wish to navigate. The href must be to within the same origin. Any external links will trigger redirection.
```
sarah11918, line 275:
```suggestion
The `navigate` method takes one argument:

- `href` - The new page to navigate to. This must be a relative URL to a page within your own domain. Any external links will trigger redirection.
```
I can see that more options are planned, but if there's only one, we should really say one.  :smile: 

And, the example above doesn't make it clear what the format should be because the content of the link is actually in a selected value. If someone were writing a path directly, would it be a relative URL like a normal link? (e.g. `/about/`) I think more explanation is necessary for this.

sarah11918 (commented):
OK! I apologize that a Github review is a terrible way to do this. I've made some structural change suggestions, and pointed out a couple of things that will need some clarification and checking. 

I really wished this were nicer, but that one stupid deleted line over which I can't edit meant I had to leave several comments around it.

See what you think about incorporating some of my suggestions! And of course, make them accurate. :sweat_smile: @matthewp @martrapp

src/content/docs/en/guides/view-transitions.mdx
martrapp: The href does not have to be relative. It can be absolute, it can even be complete with protocol, host & post. 
And while it may be too complicated to explain here, I suppose it can even have a different origin without triggering a reload if the target page has the correct CORS headers.

Suggestion:
The `navigate` method takes one argument:
- `href` - The new page to navigate to.

If an error occurs while loading the target page, the navigate function falls silently back to a normal browser load with no view transitions. Typical errors are:
* The document cannot be loaded due to an invalid `href`
* The document is not an HTML document
* The target page does not define `<ViewTransitions/>`
* The target page and the current page do not have the same origin (protocol, host, port) and there are no other measures to support CORS (Cross-Origin Resource Sharing).

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,18 +235,45 @@ const myFade = {
...
+- Clicks on `<a>` elements.
+- Backwards and forwards navigation events.
+
+The following sections discuss how you can further control when navigation occurs within the router.
+
+### Preventing client-side navigation
+
+There are some cases where you might want to not navigate via CSR. A transition will only occur when both the old and new page has the `<ViewTransitions />` component. There may be cases where you want a nice animation coming from one page, such as going from an article listing page `/articles` to an article detail page `/articles/emperor-penguins`, but you don't want an animation when coming from somewhere else on the site.
 
 In these cases, you can opt out of client-side routing on a per-link basis using the `data-astro-reload` attribute like so:
```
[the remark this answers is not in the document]
martrapp: I agree that the need to have `<ViewTransitions>` at both ends is part of the top of the page.

Instead of 
> Astro provides a `<ViewTransitions>` routing component that can be added to the `<head>` of a single page to control page transitions when you navigate to another page. 

we could say:
> Astro provides a `<ViewTransitions />` routing component that can be added to the `<head>` of any page. The view will transition seamlessly when you navigate between two pages that both contain the <ViewTransitions /> component.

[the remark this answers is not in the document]
matthewp: Yeah, this is an improvement, can move this piece of information as part of a refactor.

src/content/docs/en/guides/view-transitions.mdx
matthewp: We are going to have other options, so I'll go ahead and add them here.

matthewp: These have been added now, and the wording on `href` updated.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,68 @@ const myFade = {
...
+    - `history` - Either `'push'`, `'replace'`, or `'auto'`. When the option is `'push'` the router will use `history.pushState` to create a new entry in the browser history. If the option is `'replace'` it will use `history.replaceState` which will update the URL without adding a new entry into navigation. The `'auto'` option will also do a `history.pushState` in most cases, but if the URL is not one that can be transitioned to, it will not throw and the current URL will remain with no changes to the browser history.
+
+## Replace entries in the browser history
+
+Normally, each time you navigate, a new entry is written to the browser's history. This allows navigation between pages using the browser's `back` and `forward` buttons. For view transitions, it is possible to suppress history entries by using the `data-astro-history` attribute with the `replace` value on normal anchors. This is useful for pages that should not be navigated back to, such as a confirmation page. 
+
+The following example navigates to the `/main` page, but does not add a new entry to the browsing history. Instead, it reuses the current entry in the history and overwrites it.
```
Yan-Thomas, line 291:
```suggestion
The following example navigates to the `/main` page but does not add a new entry to the browsing history. Instead, it reuses the current entry in the history and overwrites it.
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,68 @@ const myFade = {
...
+- Clicks on `<a>` elements.
+- Backwards and forwards navigation events.
 
-```html
-<a href="/quarterly-earnings.pdf" data-astro-reload>
+The following sections discuss how you can further control when navigation occurs within the router.
+
+### Preventing client-side navigation
+
+There are some cases where you cannot navigate via CSR, since both pages involved must use the `<ViewTransitions />` router to prevent a full-page reload. You may also not want client-side routing on every navigation change, and would prefer a traditional page navigation on select routes instead.
```
Yan-Thomas, line 249: I believe both the commas here can be removed
```suggestion
There are some cases where you cannot navigate via CSR since both pages involved must use the `<ViewTransitions />` router to prevent a full-page reload. You may also not want client-side routing on every navigation change and would prefer a traditional page navigation on select routes instead.
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,68 @@ const myFade = {
...
+```astro title="src/pages/index.astro"
+<a href="/articles/emperor-penguins" data-astro-reload>
+```
+
+```astro title="src/pages/articles.astro"
+<a href="/articles/emperor-penguins">
 ```
 
-These links will be ignored by the router and a full page navigation will occur.
+Links with the `data-astro-reload` attribute will be ignored by the router and a full page navigation will occur.
```
Yan-Thomas, line 263: We previously used "full-page" (with a hyphen), which I believe is the most correct, I would change this here so we are all on the same page (see what I did here? ~~I am sorry~~) about which wording to use.
```suggestion
Links with the `data-astro-reload` attribute will be ignored by the router and a full-page navigation will occur.
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,68 @@ const myFade = {
...
+  let href = ev.target.value;
+  navigate(href);
+};
+```
+
+The `navigate` method takes these arguments:
+
+- `href` - The new page to navigate to. This must be a URL to a page within your own domain. Any external links will trigger redirection.
+- `options` - An object with the following properties:
+    - `history` - Either `'push'`, `'replace'`, or `'auto'`. When the option is `'push'` the router will use `history.pushState` to create a new entry in the browser history. If the option is `'replace'` it will use `history.replaceState` which will update the URL without adding a new entry into navigation. The `'auto'` option will also do a `history.pushState` in most cases, but if the URL is not one that can be transitioned to, it will not throw and the current URL will remain with no changes to the browser history.
```
Yan-Thomas, line 285: We can make this sentence a bit less verbose.
```suggestion
    - `history` - Either `'push'`, `'replace'`, or `'auto'`. When the option is `'push'` the router will use `history.pushState` to create a new entry in the browser history. If the option is `'replace'` it will use `history.replaceState` which will update the URL without adding a new entry into navigation. The `'auto'` option will also do a `history.pushState` in most cases, but if the URL cannot be transitioned to, it will not throw and the current URL will remain with no changes to the browser history.
```

Yan-Thomas (commented):
Adding a few nits here!

src/content/docs/en/guides/view-transitions.mdx
sarah11918: Yan is correct! I'll take this comment out of my pending review comments.  :smile:

sarah11918 (commented):
Mini review, still working on the second part of the document, which seeing it now makes me want to make some structural changes... so publishing these to get them out of the way!

src/content/docs/en/guides/view-transitions.mdx
sarah11918: Resolving this, since I've rewritten for readability! Thanks for noticing this could be improved, Yan!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,71 @@ const myFade = {
...
-Once you have enabled client-side routing on a page by adding the `<ViewTransitions />` component, every anchor on the page that links to another page on your site will be navigated via client-side routing. There are some cases where you might want to not navigate via CSR. One example is if the link is to any non-page, such as a PDF within your `public/` folder. Or an API route that produces an image.
+The `<ViewTransitions />` router handles navigation by listening to:
 
-In these cases, you can opt out of client-side routing on a per-link basis using the `data-astro-reload` attribute like so:
+- Clicks on `<a>` elements.
+- Backwards and forwards navigation events.
 
-```html
-<a href="/quarterly-earnings.pdf" data-astro-reload>
+The following sections discuss how you can further control when navigation occurs within the router.
```
sarah11918, line 245:
```suggestion
The following options allow you to further control when navigation occurs within the router:

- `astro-data-reload`: an `<a>` tag attribute to force a full-page navigation
- `astro-data-history="auto | push | replace"`: an `<a>` tag attribute to control the browser's history
- `navigate(href, options)`: a method available to any client script or client component to trigger navigation
```
I think this is the introduction to the section I was wanting! Now, the sub sections below can be more usage based and we don't as easily lose these options inside specific user activities.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,75 @@ const myFade = {
...
+The `navigate` method takes these arguments:
+
+- `href` - The new page to navigate to.
+- `options` - An object with the following properties:
+    - `history`: `'push'` | `'replace'` | `'auto'`
+    	- `'push'` the router will use `history.pushState` to create a new entry in the browser history. 
+    	- `'replace'` it will use `history.replaceState` which will update the URL without adding a new entry into navigation.
+    	- `'auto'` option will also do a `history.pushState` in most cases, but if the URL is not one that can be transitioned to, it will not throw and the current URL will remain with no changes to the browser history.
+
+## Replace entries in the browser history
```
sarah11918, line 294:
```suggestion
### Replace entries in the browser history
```
This feels like it should be a sub-heading under Router Control

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,75 @@ const myFade = {
...
+
+Normally, each time you navigate, a new entry is written to the browser's history. This allows navigation between pages using the browser's `back` and `forward` buttons. For view transitions, it is possible to suppress history entries by using the `data-astro-history` attribute with the `replace` value on normal anchors. This is useful for pages that should not be navigated back to, such as a confirmation page.
```
sarah11918, line 296:
```suggestion
Normally, each time you navigate, a new entry is written to the browser's history. This allows navigation between pages using the browser's `back` and `forward` buttons. 

The <ViewTransitions /> router allows you to overwrite history entries by adding the `data-astro-history` attribute to any individual `<a>` tag.

The `data-astro-history` attribute can be set to the same three values as the [`history` option of the `navigate()` function](#trigger-navigation): 

`data-astro-history`: `'push'` | `'replace'` | `'auto'`
- `'push'`: the router will use `history.pushState` to create a new entry in the browser history. 
- `'replace'`: the router will use `history.replaceState` to update the URL without adding a new entry into navigation.
- `'auto'` (default): the router will attempt `history.pushState`, but if the URL is not one that can be transitioned to, the current URL will remain with no changes to the browser history.

The following example navigates to the `/main` page but does not add a new entry to the browsing history. Instead, it reuses the current entry in the history (`/confirmation`) and overwrites it.

```astro title="/confirmation.astro"
<a href="/main" data-astro-history="replace">
```

Now, when navigating back from the `/main` page, the browser will not display the `/confirmation' page, but the page before it.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,75 @@ const myFade = {
...
+```
+
+The `navigate` method takes these arguments:
+
+- `href` - The new page to navigate to.
+- `options` - An object with the following properties:
+    - `history`: `'push'` | `'replace'` | `'auto'`
+    	- `'push'` the router will use `history.pushState` to create a new entry in the browser history. 
+    	- `'replace'` it will use `history.replaceState` which will update the URL without adding a new entry into navigation.
+    	- `'auto'` option will also do a `history.pushState` in most cases, but if the URL is not one that can be transitioned to, it will not throw and the current URL will remain with no changes to the browser history.
```
sarah11918, line 292:
```suggestion
- `options` - An object with the following properties:
	- `history`: `'push'` | `'replace'` | `'auto'`
		- `'push'`: the router will use `history.pushState` to create a new entry in the browser history. 
		- `'replace'`: the router will use `history.replaceState` to update the URL without adding a new entry into navigation.
		- `'auto'` (default): the router will attempt `history.pushState`, but if the URL is not one that can be transitioned to, the current URL will remain with no changes to the browser history.
```

sarah11918: edited wording and hoping formatting works

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,89 @@ const myFade = {
...
+	- `history`: `'push'` | `'replace'` | `'auto'`
+		- `'push'`: the router will use `history.pushState` to create a new entry in the browser history. 
+		- `'replace'`: the router will use `history.replaceState` to update the URL without adding a new entry into navigation.
+		- `'auto'` (default): the router will attempt `history.pushState`, but if the URL is not one that can be transitioned to, the current URL will remain with no changes to the browser history.
+
+### Replace entries in the browser history
+
+Normally, each time you navigate, a new entry is written to the browser's history. This allows navigation between pages using the browser's `back` and `forward` buttons. 
+
+The <ViewTransitions /> router allows you to overwrite history entries by adding the `data-astro-history` attribute to any individual `<a>` tag.
```
sarah11918, line 298:
```suggestion
The `<ViewTransitions />` router allows you to overwrite history entries by adding the `data-astro-history` attribute to any individual `<a>` tag.
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,89 @@ const myFade = {
...
+
+The following example navigates to the `/main` page but does not add a new entry to the browsing history. Instead, it reuses the current entry in the history and overwrites it.
+
+```astro title="/confirmation.astro"
+<a href="/main" data-astro-history="replace">
+```
+
+This has the effect that if you go back from the `/main` page, the browser will not display the `/confirmation' page, but the page before it.
+
+The values (`auto`, `push`, `replace`) and behavior of the `data-astro-history` attribute are the same as for the [`history` option of the `navigate()` function](#trigger-navigation). `auto` is the default if the `data-astro-history` attribute is omitted.
```
sarah11918, line 320:
```suggestion
```

sarah11918:
Left lots of little comments above, and this new section can be revisited once we have these options all on a reference page, too. For now, I think what we have balances needing to surface the new options released while providing use-case based sections and examples.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,88 @@ const myFade = {
...
+- `'auto'` (default): the router will attempt `history.pushState`, but if the URL is not one that can be transitioned to, the current URL will remain with no changes to the browser history.
+
+The following example navigates to the `/main` page but does not add a new entry to the browsing history. Instead, it reuses the current entry in the history (`/confirmation`) and overwrites it.
+
+```astro title="/confirmation.astro"
+<a href="/main" data-astro-history="replace">
+
+The following example navigates to the `/main` page but does not add a new entry to the browsing history. Instead, it reuses the current entry in the history and overwrites it.
+
+```astro title="/confirmation.astro"
```
martrapp, line 314:
```suggestion
```

martrapp (changes requested):

martrapp (approved):
Looks Good! Tank you!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -235,17 +235,83 @@ const myFade = {
...
+document.querySelector('select').onchange = (ev) => {
+  let href = ev.target.value;
+  navigate(href);
+};
+```
+
+The `navigate` method takes these arguments:
+
+- `href` - The new page to navigate to.
+- `options` - An object with the following properties:
```
sarah11918, line 288:
```suggestion
- `href` (required) - The new page to navigate to.
- `options` - An optional object with the following properties:
```

PR #4766, opened 2023-09-20
Add Bun and Deno to create-astro guide

birkskyum:

- New or updated content

#### Description

Added Bun and Deno to getting started. Intentionally added after in the end of the list, as newcomers. See pictures.

Screenshot 2023-09-20 at 23 47 10
Screenshot 2023-09-20 at 23 47 14

### DIFF
--- a/src/components/tabs/PackageManagerTabs.astro
+++ b/src/components/tabs/PackageManagerTabs.astro
@@ -6,6 +6,8 @@ import Tabs from './Tabs';
 	<Fragment slot="tab.npm">npm</Fragment>
 	<Fragment slot="tab.pnpm">pnpm</Fragment>
 	<Fragment slot="tab.yarn">Yarn</Fragment>
+	<Fragment slot="tab.zdeno">Deno</Fragment>
+	<Fragment slot="tab.zbun">Bun</Fragment>
 
 	<Fragment slot="panel.npm">
 		<slot name="npm" />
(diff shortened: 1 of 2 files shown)


sarah11918:
Hi @birkskyum ! This is just to let you know that we're discussing this internally, because there's been a lot of bun activity recently, but it's still unsupported by *us* and so we are hesitant to document it. Given that we cannot offer any support with issues using bun, we are leaning against including this in our own documentation. If we do, it will only be to give us a clear place to say, "we're not responsible, file any issues with bun directly" and that's not exactly what we like to have in our own docs, especially on our Getting Started page! :sweat_smile: 

So this is just to let you know where this stands right now! We'll be discussing more and eventually deciding whether this is something for our own documentation. Thank you for making this contribution, as we really do appreciate the community jumping in and chipping in! This just might not be what we decide to go with. :raised_hands:

sarah11918:
OK, our official stance on this right now is *not* to update and show more package managers at this time.

So sorry, I'll be closing this PR @birkskyum, but I do appreciate that you took the time to contribute, and I hope you'll be back when you spot a typo or see an Issue you feel like fixing!  🙌

birkskyum:
Okay. Guess i just saw it a bit different. To me this wasn't really about package managers, but more about runtimes. The are guides for 3x Node workflows, and no bun or Deno.

PR #4786, opened 2023-09-22
Document the route announcer

matthewp:

- New or updated content

#### Description

- In Astro 3.2 there will be a route announcer (a way to tell AT that the page has changed and the new title). This documents the behavior.

- Part of https://github.com/withastro/astro/pull/8621


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -356,7 +356,25 @@ For example, if you are implementing dark mode support, this event can be used t
...
+
+Switching to client-side routing and animating the pages both come with accessibility challenges, and Astro aims to make sites opt-ing into View Transitions as accesible-by-default as possible.
+
+### Route announcement
+
+<Since v="3.2.0" />
+
+The `<ViewTransitions />` component includes a route announcer for page navigation. When using server-side routing assistive technologies will announce the page title after a page loads, which acts as a way to let the user know that the page has changed. When doing client-side routing you want the same thing to occur.
+
+The component achieves this by adding an element to the page with the `aria-live` attribute set to `assertive`. This tells AT (assistive technology) to announce immediately. This announcement text checks these in priority order:
```
sarah11918, line 369:
```suggestion
To add route announcement to client-side routing, the component adds an element to the new page with the `aria-live` attribute set to `assertive`. This tells AT (assistive technology) to announce immediately. The component also checks for the following, in priority order, to determine the announcement text:
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -356,7 +356,25 @@ For example, if you are implementing dark mode support, this event can be used t
...
+
+- The `<title>`, if it exists.
+- The first `<h1>` it finds.
+- The `pathname` of the page.
+
+No configuration or action is needed by you to enable this.
```
sarah11918, line 375:
```suggestion
We strongly recommend you always include a `<title>` in each page for accessibility.
```
Suggested to move this existing line up top so people know right away they don't have to read/do anything! Instead, this would be a great place for that note from the implementation PR saying that we suggest always having a title!

sarah11918 (commented):
Looks great @matthewp ! Made some suggestions for you to consider!

We'll address the broken link once we decide on exactly what that heading will be, then it will be fine!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -356,7 +356,25 @@ For example, if you are implementing dark mode support, this event can be used t
 </script>
 ```
 
-## `prefers-reduced-motion`
+## Accessibility
```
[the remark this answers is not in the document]
Yan-Thomas: That's fine! Astro by itself doesn't have any "accessibility gotchas" - we sprinkle a11y through our docs even if we aren't making a heading for it, so we're well covered!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -356,7 +356,27 @@ For example, if you are implementing dark mode support, this event can be used t
...
+
+Enabling client-side routing and animating page transitions both come with accessibility challenges, and Astro aims to make sites opting into View Transitions as accessible-by-default as possible.
```
sarah11918, line 361:
```suggestion
Enabling client-side routing and animating page transitions both come with accessibility challenges, and Astro aims to make sites opting in to View Transitions as accessible-by-default as possible.
```
We're going to get cards and letters, even from Yan, but this is 100% correct. :sweat_smile:

PR #4835, opened 2023-09-28
Fix premature tags array refactoring

ejiek:

#### Description

In a section 5.3 of **Building a Blog** there is a section "Create an array of tags".
The firts example in this section should have a harcoded array of tags for them to be dynamicly acquired later.
Before this PR the first example contained the final result, making the following refactoring unclear.


ejiek:
The fist step of "Create your list of tags" is
```
+     <div>{tags.map((tag) => <p>{tag}</p>)}</div>

sarah11918:
Hi @ejiek ! It looks like in the mean time, another fix to the tutorial code was made. You both had the right idea!!  :smile: 

We can still commit the PR though, to celebrate your contribution to docs! (sometimes it happens; people have the same fix!) Thank you again for spotting this, and for taking the time to make the fix! We really do appreciate it.

PR #4848, opened 2023-09-28
Docs - Tutorial Updates

richeklein:

- New or updated content

#### Description

Update the code example to match the question. The question asks "What would it look like to define an array of all your tags, then display them in a list on this page?", but the code example skips ahead to show the dynamic solution.


TheOtterlord (approved):
Good catch! Thanks so much for the fix @richeklein :raised_hands:

PR #4866, opened 2023-09-29
docs: Add mention of Picture and new `srcset`-related attributes

Princesseuh:

- New or updated content

#### Description

This is docs for https://github.com/withastro/roadmap/pull/715.


src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -173,6 +173,8 @@ export type ImageTransform = {
     src: ImageMetadata | string;
     width?: number;
     height?: number;
+    widths?: number[] | undefined;
+	  densities?: (number | `${number}x`)[] | undefined;
```
florian-lefebvre, line 177: Looks like it breaks syntax highlighting

Princesseuh: GitHub seems to be the only one having a problem with it, it works both on the docs and in VS Code

florian-lefebvre: Feel free to resolve the conversation then 👍

src/content/docs/en/guides/images.mdx
```diff
@@ -145,6 +142,22 @@ When using local images in their original aspect ratio, the `width` and `height`
 
 However, both of these properties are required for remote images and images stored in your `public/` folder as Astro is unable to analyze these files.
 
+##### densities
+
+<Since v="3.3.0" /> <Badge>Experimental</Badge>
+
+A list of pixel densities to generate for the image. By default, no other densities are generated. Densities that are equals to widths larger than the original image will be ignored, as to avoid upscaling the image.
+
+This value will be used to generate a `srcset` attribute on the `<img>` tag.
```
sarah11918, line 151:
``````suggestion
A list of pixel densities to generate for the image. 

If provided, this value will be used to generate a `srcset` attribute on the `<img>` tag. Do not provide a value for `widths` when using this value.

Densities that are equals to widths larger than the original image will be ignored to avoid upscaling the image.

```astro
---
import { Image } from 'astro:assets';
import myImage from "../assets/my_image.png";
---
<Image src={myImage} densities={[1x, 2x, 3x]} alt="A description of my image." />
```
``````
Can we add an astro/html output example here? Something like above, with the output?

File: src/content/docs/en/guides/images.mdx [hunk 3]
```diff
@@ -145,6 +142,22 @@ When using local images in their original aspect ratio, the `width` and `height`
...
+
+##### widths
+
+<Since v="3.3.0" /> <Badge>Experimental</Badge>
+
+A list of widths to generate for the image. By default, no other widths are generated. Widths that are larger than the original image will be ignored in order to avoid upscaling the image.
+
+This value will be used to generate a `srcset` attribute on the `<img>` tag. When using `widths`, a [`sizes` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/sizes) is also required for `srcset` to take effect.
```
sarah11918, line 159:
``````suggestion
##### widths

<Since v="3.3.0" /> <Badge>Experimental</Badge>

A list of widths to generate for the image. 

If provided, this value will be used to generate a `srcset` attribute on the `<img>` tag. A [`sizes` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/sizes) must also be provided.

Do not provide a value for `densities` when using this value. Only one of these two values can be used to generate a `srcset`.

Widths that are larger than the original image will be ignored to avoid upscaling the image.

```astro
---
import { Image } from 'astro:assets';
import myImage from "../assets/my_image.png";
---
<Image src={myImage} widths={[300, 600, 900]} alt="A description of my image." />
```
``````
Similarly, I think an astro/html example would be nice here!

File: src/content/docs/en/guides/images.mdx [hunk 4]
```diff
@@ -209,6 +222,54 @@ const {src, ...attrs} = Astro.props;
...
+### `<Picture />`
+
+<Since v="3.3.0" /> <Badge>Experimental</Badge>
+
+Use the built-in `<Picture />` Astro component to display a responsive image with multiple formats and/or sizes.
+
+```astro title="src/pages/index.astro"
+---
+import { Picture } from 'astro:assets';
+import myImage from "../assets/my_image.png";
```
sarah11918, line 234:
```suggestion
import myImage from "../assets/my_image.png"; // Image is 1600x900
```
This was on the Image component, and seems to match the output shown!

src/content/docs/en/guides/images.mdx
```diff
@@ -209,6 +222,54 @@ const {src, ...attrs} = Astro.props;
...
+</picture>
+```
+
+#### Properties
+
+`<Picture />` accepts all the properties of the `<Image />` component, plus the following:
+
+##### `formats`
+
+An array of image formats to use for the `<source>` tags. Every entry will be added as `<source>` elements in the order they were inputted. As such, your most modern format should go first. By default, this is set to `['webp']`.
```
sarah11918, line 263:
```suggestion
An array of image formats to use for the `<source>` tags. Entries will be added as `<source>` elements in the order they are listed, and this order determines which format is displayed. For the best performance, list the most modern format first (e.g. `webp` or `avif`). By default, this is set to `['webp']`.
```

src/content/docs/en/guides/images.mdx
```diff
@@ -209,6 +222,54 @@ const {src, ...attrs} = Astro.props;
...
+
+##### `fallbackFormat`
+
+Format to use as a fallback value for the `<img>` tag. Defaults to `.png`, unless the image is animated or a SVG, in which case it'll fallback to `.gif` and `.svg` respectively.
```
sarah11918, line 267:
```suggestion
Format to use as a fallback value for the `<img>` tag.

Defaults to `.png`for static images, `.gif` for animated images, and `.svg` for an SVG.
```

src/content/docs/en/guides/images.mdx
```diff
@@ -209,6 +222,54 @@ const {src, ...attrs} = Astro.props;
...
+
+##### `pictureAttributes`
```
sarah11918, line 269: I'll still want to edit this wording a tiny bit, but would love to see an example here before I do!

sarah11918 (commented):
Just finishing this "mostly done" review now! There are a couple of suggestions I've made for you to see if they make sense.

The only thing I haven't looked at really yet is the `pictureAttributes` which I think could benefit from an example on this page!

sarah11918 (commented):
Just finishing this "mostly done" review now! There are a couple of suggestions I've made for you to see if they make sense.

The only thing I haven't looked at really yet is the `pictureAttributes` which I think could benefit from an example on this page!

src/content/docs/en/guides/images.mdx
```diff
@@ -145,6 +142,75 @@ When using local images in their original aspect ratio, the `width` and `height`
...
+```html
+<img
+  src="/_astro/my_image.hash.webp"
+  srcset="
+    /_astro/my_image.hash.webp 240w,
+    /_astro/my_image.hash.webp 540w,
+    /_astro/my_image.hash.webp 720w
+  "
+  alt="A description of my image."
+  width="700"
```
sarah11918, line 207: Just checking, should this be 720?

src/content/docs/en/guides/images.mdx
```diff
@@ -209,6 +275,79 @@ const {src, ...attrs} = Astro.props;
...
+
+##### `fallbackFormat`
+
+Format to use as a fallback value for the `<img>` tag.
+
+Defaults to `.png` for static images, `.gif` for animated images, and `.svg` for SVG files.
+
+##### `pictureAttributes`
+
+An object of attributes to be added to the `<picture>` tag. Every other attributes, apart from the ones used for the image transformation, will be applied to the inner `<img>` element. This is notably useful for applying style and classes to your image.
```
sarah11918, line 326:
```suggestion
An object of attributes to be added to the `<picture>` tag. Use this property to apply attributes to the inner `<img>` element such as styles and classes.
```
Is this correct to say? Does it lose any important nuance or distinction that the reader needs to know to use the component if we just say this?

Princesseuh: Unfortunately no, the distinction that the attributes are applied to the `<picture>` element is actually important.

sarah11918: OH, I misread it at first! Now I think I get it! Let me try again..

sarah11918: @Princesseuh I updated the above. Is that correct? (If so, is it good?  :sweat_smile: )

Princesseuh: It is correct and good!

src/content/docs/en/guides/images.mdx
```diff
@@ -209,6 +276,81 @@ const {src, ...attrs} = Astro.props;
...
+
+Format to use as a fallback value for the `<img>` tag.
+
+Defaults to `.png` for static images, `.gif` for animated images, and `.svg` for SVG files.
+
+##### `pictureAttributes`
+
+An object of attributes to be added to the `<picture>` tag. 
+
+Use this property to apply attributes to the outer `<picture>`  element itself. Attributes applied to the `<Picture />` component directly will apply to the inner `<img>` element, except for those used for image transformation.
```
sarah11918, line 329:
```suggestion
Use this property to apply attributes to the outer `<picture>` element itself. Attributes applied to the `<Picture />` component directly will apply to the inner `<img>` element, except for those used for image transformation.
```
Just caught a double space. Will commit the change directly.

src/content/docs/en/guides/images.mdx
```diff
@@ -145,6 +142,76 @@ When using local images in their original aspect ratio, the `width` and `height`
...
 
+##### densities
+
+<Since v="3.3.0" /> <Badge>Experimental</Badge>
+
+A list of pixel densities to generate for the image.
+
+If provided, this value will be used to generate a `srcset` attribute on the `<img>` tag. Do not provide a value for `widths` when using this value.
+
+Densities that are equals to widths larger than the original image will be ignored to avoid upscaling the image.
```
Yan-Thomas, line 153: I guess here the correct usage would be "equal" instead? 
```suggestion
Densities that are equal to widths larger than the original image will be ignored to avoid upscaling the image.
```

Yan-Thomas (approved):
LGTM! Just a small nit but then we're good 🙌

PR #4878, opened 2023-09-29
Clarify use case of astro:after-swap event

brandon-pereira:

- New or updated content

#### Description

- This PR adds a clarification on the astro:after-swap logic. I have been struggling to implement some basic SPA-esque user experiences (scroll to X element after new page loads, if loaded via view transition) and was struggling. I believe this contribution will help future folks understand how to implement these types of interactions.


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -422,6 +422,9 @@ For example, if you are implementing dark mode support, this event can be used t
 </script>
 ```
 
+Keep in mind that this event is fired before the new pages scripts have started listening. As such, in the above example, any modifications to the dark mode
```
sarah11918, line 425:
```
An event that fires immediately after the new page replaces the old page. You can listen to this event on the `document` and trigger actions that will occur before the new page's scripts run.

This event, when listened to on the outgoing page, is useful to pass along and restore any state on the DOM that needs to transfer over to the new page.

For example, if you are implementing dark mode support, this event can be used to restore state across page loads:
```

Thanks for this thoughtful contribution, @brandon-pereira! 

I know you've given a nice "Keep in mind..." addition, but I never like to tell people to "keep something in mind" when I could have just told them properly in the first place!  😄 

So, I took a shot at redrafting the text *above* the code sample to see whether we could incorporate your extra nuance and just get it right the first time. Let me know what you think!

brandon-pereira: @sarah11918 I like it :) I did make `outgoing page` bold because I think it's import and easy to miss that key phrase when reading the docs. I can remove the bold if you disagree.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -401,9 +401,9 @@ You can use this event to run code on every page navigation, or only once ever:
 
 ### `astro:after-swap`
 
-An event that fires immediately after the new page replaces the old page. You can listen to this event on the `document`.
+An event that fires immediately after the new page replaces the old page. You can listen to this event on the `document` and trigger actions that will occur before the new page's scripts run.
```
sarah11918, line 404:
```suggestion
An event that fires immediately after the new page replaces the old page. You can listen to this event on the `document` and trigger actions that will occur before the new page's DOM elements render and scripts run. 
```
What do you think about adding this too, while we're here? I've put this advice in the new tiny view transitions tutorial to explain why we're using it to check for dark mode to avoid a flash of light mode first. I think maybe it's even more helpful to say not just scripts? What do you think?

brandon-pereira: Sure, that works, adjusted @sarah11918

PR #4888, opened 2023-09-30
i18n(fr) Update aliases file

McFlyPartages:

- Updated translated content


McFlyPartages:
I accidentally added the setup-editor file, if someone could delete it because it already has its PR #4887 and I can't do it.

Thanks

at-the-vr:
Hey @McFlyPartages since you don't want to update `setup-editor`, you can follow this [instructions from stackoverflow](https://stackoverflow.com/a/38744242). This should prevent any possible merge conflicts as well 😉

McFlyPartages:
Idem une petite relecture pour valider ?

PR #4899, opened 2023-10-01
Updating outdated information for astro check documentation for Astro 3 

### LINKED ISSUE #4789: `astro check` documentation is outdated for Astro 3
### 📚 Subject area/topic

TypeScript

https://docs.astro.build/en/guides/typescript/#type-checking

This is actually mostly my fault.

The typechecking section mentions using both `tsc --noEmit` and `astro check`, but in Astro 3, those commands do essentially the same thing, because `astro check` now checks all the files included in your TypeScript project, which in most case will include your `.(t|j)s(x)`.

This text: 
> astro check only checks types within .astro files, and tsc --noEmit only checks types within .ts and .tsx files.
should be removed

and the package.json should be updated to remove the call to `tsc --noEmit`

N/A

jdwilkin4:

- New or updated content

#### Description

This PR removes the mention of `tsc --noEmit` in both the `package.json` example and note for astro check docs in the Type Checking section for astro 3. 

[Link to page with the fix](https://deploy-preview-4899--astro-docs-2.netlify.app/en/guides/typescript/#type-checking)

#### Before
Screenshot 2023-09-30 at 9 28 39 PM

#### After
Screenshot 2023-09-30 at 9 29 43 PM

- Closes #4789

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -270,15 +270,15 @@ const { title } = Astro.props;
 
 To see type errors in your editor, please make sure that you have the [Astro VS Code extension](/en/editor-setup/) installed. Please note that the `astro start` and `astro build` commands will transpile the code with esbuild, but will not run any type checking. To prevent your code from building if it contains TypeScript errors, change your "build" script in `package.json` to the following:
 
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Thanks so much for snagging this error and making the fix, @jdwilkin4 ! This looks great, and is so helpful. :raised_hands:

PR #4910, opened 2023-10-02
Styling child components with Astro3 defaults

arty-name:

- New or updated content

#### Description

This PR updates the section of the docs about passing classes to child components to make it match the new default approach introduced in Astro 3. The previous version of documentation was written for [scopedStyleStrategy](https://docs.astro.build/en/reference/configuration-reference/#scopedstylestrategy) being `class` or `where`, but `attribute` is used by default in Astro 3, so the old approach does not work anymore.

https://github.com/withastro/astro/issues/8309#issuecomment-1701771500

[The preview of the changed section](https://deploy-preview-4910--astro-docs-2.netlify.app/en/guides/styling/#passing-a-class-to-a-child-component)


src/content/docs/en/guides/styling.mdx
```diff
@@ -136,14 +136,13 @@ const backgroundColor = "rgb(24 121 78)";
 
 In Astro, HTML attributes like `class` do not automatically pass through to child components.
 
-
-Instead, accept a `class` prop in the child component and apply it to the root element. When destructuring, you must rename it, because `class` is a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript.
+Instead, accept a `class` prop in the child component and apply it to the root element. When destructuring, you must rename it, because `class` is a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript. The default styles scoping strategy of Astro also requires you to pass the `data-astro-cid-*` attribute, and the simplest way to do this is by taking the `...rest` of the props and forwarding it to the component.
```
sarah11918, line 139:
```suggestion
Instead, accept a `class` prop in the child component and apply it to the root element. When destructuring, you must rename it, because `class` is a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript. 

When using the default `scopedStyleStrategy: 'attribute'`, you must also pass the `data-astro-cid-*` attribute. You can do this by also passing the `...rest` of the props to the component.
```
Just making some shorter, more active sentences!

src/content/docs/en/guides/styling.mdx
```diff
@@ -160,9 +159,10 @@ import MyComponent from "../components/MyComponent.astro"
 <MyComponent class="red">This will be red!</MyComponent>
 ```
 
-This pattern lets you style child components directly. Astro will pass the parent’s scoped class name (e.g. `astro-hhnqfkh6`) through the `class` prop automatically, including the child in its parent’s scope. Note this pattern only works when your [`scopedStyleStrategy` option](/en/reference/configuration-reference/#scopedstylestrategy) is either `'where'` or `'class'`.
-:::note[Scoped classes from parent components]
-Because the `class` prop includes the child in its parent’s scope, it is possible for styles to cascade from parent to child. To avoid this having unintended side effects, ensure you use unique class names in the child component.
+This pattern lets you style child components directly. If your [`scopedStyleStrategy` option](/en/reference/configuration-reference/#scopedstylestrategy) is set to either `'where'` or `'class'` you can remove the `...rest` props to make it simpler.
```
sarah11918, line 162:
```suggestion
This pattern lets you style child components directly. If your [`scopedStyleStrategy` option](/en/reference/configuration-reference/#scopedstylestrategy) is set to either `'where'` or `'class'`, the `...rest` props are not necessary.
```

sarah11918 (commented):
Thank you for taking the initiative and updating this example to be more helpful with our new default scoped styles strategy!  :raised_hands: 

I just make some tiny edits to make the language more in line with Astro docs style (shorter, more active instructions). Make sure I didn't lose or change any meaning here, and let me know if you think this works!

src/content/docs/en/guides/styling.mdx
arty-name: Right, I see what the goal is.

After looking at the suggested wording, I wrote a response and deleted it because I have not noticed the word "default". Is this also an indication of it not standing out enough? What do you think about this alternative?

> You must also pass the `data-astro-cid-*` attribute, unless you have changed `scopedStyleStrategy` to `class` or `where`. You can do this by passing the `...rest` of the props to the component.

sarah11918: Ooh, that makes a lot of sense, yes! Even better than the original, I think!

arty-name: Thank you! I have updated the PR to use this sentence, could you please review it again?

src/content/docs/en/guides/styling.mdx
```diff
@@ -136,14 +136,15 @@ const backgroundColor = "rgb(24 121 78)";
 
 In Astro, HTML attributes like `class` do not automatically pass through to child components.
 
-
 Instead, accept a `class` prop in the child component and apply it to the root element. When destructuring, you must rename it, because `class` is a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript.
 
+You must also pass the `data-astro-cid-*` attribute, unless you have changed `scopedStyleStrategy` to `'class'` or `'where'`. You can do this by passing the `...rest` of the props to the component.
```
sarah11918, line 141:
```suggestion
Using the default scoped style strategy, you must also pass the `data-astro-cid-*` attribute. You can do this by passing the `...rest` of the props to the component. If you have changed `scopedStyleStrategy` to `'class'` or `'where'`, the `...rest` prop is not necessary.
```
Just separating more clearly the two cases here, instead of mixing the "unless" in the middle of the need and the way to achieve it.

(This also means we can remove line 164, I think?)

src/content/docs/en/guides/styling.mdx
```diff
@@ -160,9 +161,10 @@ import MyComponent from "../components/MyComponent.astro"
 <MyComponent class="red">This will be red!</MyComponent>
 ```
 
-This pattern lets you style child components directly. Astro will pass the parent’s scoped class name (e.g. `astro-hhnqfkh6`) through the `class` prop automatically, including the child in its parent’s scope. Note this pattern only works when your [`scopedStyleStrategy` option](/en/reference/configuration-reference/#scopedstylestrategy) is either `'where'` or `'class'`.
-:::note[Scoped classes from parent components]
-Because the `class` prop includes the child in its parent’s scope, it is possible for styles to cascade from parent to child. To avoid this having unintended side effects, ensure you use unique class names in the child component.
+This pattern lets you style child components directly. If your [`scopedStyleStrategy` option](/en/reference/configuration-reference/#scopedstylestrategy) is set to either `'where'` or `'class'`, the `...rest` props are not necessary.
```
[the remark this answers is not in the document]
arty-name: Yeah, I noticed this duplication earlier but you found a way to avoid it altogether :+1:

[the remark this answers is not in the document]
sarah11918: Avoiding writing is what I do best! lol. Let's ship this!

arty-name:
I’ve updated the text, guess you will be happy with it now :)

src/content/docs/en/guides/styling.mdx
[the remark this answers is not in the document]
arty-name: This is the way :joy:

PR #4911, opened 2023-10-02
Use status 404 instead of redirect

arty-name:

- New or updated content

#### Description

This PR changes multiple places in the documentation to return the Response object with 404 status instead of redirecting to the `/404` URL. The previous approach has downsides of losing the information the user has provided, among others, making it impossible for the user to correct a typo in the URL.

### DIFF
--- a/src/content/docs/en/core-concepts/routing.mdx
+++ b/src/content/docs/en/core-concepts/routing.mdx
@@ -206,7 +206,7 @@ const pages = [
 
 const { slug } = Astro.params;
 const page = pages.find((page) => page.slug === slug);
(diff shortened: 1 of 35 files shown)


sarah11918:
Hi @arty-name ! The team has decided that we'd rather show the redirects as currently is in documentation, so I'll be closing this PR. But of course, people are free to use whichever method in their own projects they like.

And, for future reference, we do ask that only changes to the English documentation, and not all the other languages, are made when changing docs content. So when making contributions in the future, please only change the English files!  🙌

arty-name:
Thank you Sara for confirming that this is a conscious decision!

PR #4974, opened 2023-10-05
Scripts and Event Handling: elaborate on module scripts

lilnasy:
#### Description
Clarify a few common points of confusion regarding processing of scripts.

#### Related issues & labels
None

### DIFF
--- a/src/content/docs/en/guides/client-side-scripts.mdx
+++ b/src/content/docs/en/guides/client-side-scripts.mdx
@@ -46,7 +46,7 @@ In this example, adding the `<Hello />` component to a page will log a message t
 </script>
 ```
 
-### Script bundling
+### How Astro processes scripts
 
 By default, `<script>` tags are processed by Astro.
 
@@ -62,7 +62,18 @@ By default, `<script>` tags are processed by Astro.
 </script>
 ```
 
-To avoid bundling the script, you can use the `is:inline` directive.
+#### All processed scripts become module scripts
+
+This has several benefits:
+- Rendering is not blocked. The browser continues to process the rest of the html while the module script and its dependencies load.
+- Browsers wait for html to processed before executing module scripts. You do not need to set "load" event listeners.
+- `async` and `defer` attributes become unnecessary. Module scripts are always deferred.
+
+The `async` attribute is valuable for normal scripts as it prevents them from blocking rendering. However, module scripts already have this behavior and adding this attribute will make module scripts execute before the page has fully loaded. This is probably not what you want.
+
+#### Opting out of processing
+
+To prevent Astro from processing a script, you can use the `is:inline` directive.
 
 ```astro title="src/components/InlineScript.astro" "is:inline"
 <script is:inline>
@@ -73,18 +84,18 @@ To avoid bundling the script, you can use the `is:inline` directive.
 ```
 
 :::note
-Astro's default bundling behavior will be disabled in some situations. In particular, adding `type="module"` or any attribute other than `src` to a `<script>` tag will cause Astro to treat the tag as if it had an `is:inline` directive. The same will be true when the script is written in a JSX expression.
+Astro will not process your script tags in some situations. In particular, adding `type="module"` or any attribute other than `src` to a `<script>` tag will cause Astro to treat the tag as if it had an `is:inline` directive. The same will be true when the script is written in a JSX expression.
 :::
 
 📚 See our [directives reference](/en/reference/directives-reference/#script--style-directives) page for more information about the directives available on `<script>` tags.
 
-### Script loading
+### Include javascript files on your page
 
 You may want to write your scripts as separate `.js`/`.ts` files or need to reference an external script on another server. You can do this by referencing these in a `<script>` tag’s `src` attribute.
 
 #### Import local scripts
 
-**When to use this:** If your script lives inside of `src/`.
+**When to use this:** when your script lives inside of `src/`.
 
 Astro will build, optimize, and add these scripts to the page for you, following its [script bundling rules](#script-bundling).
 
@@ -98,7 +109,7 @@ Astro will build, optimize, and add these scripts to the page for you, following
 
 #### Load external scripts
 
-**When to use this:** If your JavaScript file lives inside of `public/` or on a CDN.
+**When to use this:** when your JavaScript file lives inside of `public/` or on a CDN.
 
 To load scripts outside of your project's `src/` folder, include the `is:inline` directive. This approach skips the JavaScript processing, bundling, and optimizations that are provided by Astro when you import scripts as described above.
 


src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -73,18 +84,18 @@ To avoid bundling the script, you can use the `is:inline` directive.
...
+### Include javascript files on your page
 
 You may want to write your scripts as separate `.js`/`.ts` files or need to reference an external script on another server. You can do this by referencing these in a `<script>` tag’s `src` attribute.
 
 #### Import local scripts
 
-**When to use this:** If your script lives inside of `src/`.
+**When to use this:** when your script lives inside of `src/`.
 
 Astro will build, optimize, and add these scripts to the page for you, following its [script bundling rules](#script-bundling).
```
sarah11918, line 100:
```suggestion
Astro will build, optimize, and add these scripts to the page for you, following its [script bundling rules](#how-astro-processes-scripts).
```
This will fix the fact that the heading has been changed, but given that the wording has changed, please also check this sentence to see that it accurately describes what it's linking to. i.e. it claims to be taking you to "script bundling rules" but now takes you to a more generic section. Verify you think this still makes sense as a place to link and/or whether this sentence is still phrased correctly for the new context! 🙌 

(And, if you follow my suggestion to change that heading to something else, actually, that should be reflected here.)

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -62,7 +62,18 @@ By default, `<script>` tags are processed by Astro.
 </script>
 ```
 
-To avoid bundling the script, you can use the `is:inline` directive.
+#### All processed scripts become module scripts
+
+This has several benefits:
+- Rendering is not blocked. The browser continues to process the rest of the html while the module script and its dependencies load.
```
sarah11918, line 68:
```suggestion
- Rendering is not blocked. The browser continues to process the rest of the HTML while the module script and its dependencies load.
```

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -62,7 +62,18 @@ By default, `<script>` tags are processed by Astro.
...
+- The browser waits for html to be processed before executing module scripts. You do not need to listen for the "load" event.
```
sarah11918, line 69:
```suggestion
- The browser waits for HTML to be processed before executing module scripts. You do not need to listen for the "load" event.
```

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -62,7 +62,18 @@ By default, `<script>` tags are processed by Astro.
...
+- `async` and `defer` attributes become unnecessary. Module scripts are always deferred.
```
sarah11918, line 70:
```suggestion
- `async` and `defer` attributes are unnecessary. Module scripts are always deferred.
```
I think this fits the same verb tense as the others. We're describing what *does* happen.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -62,7 +62,18 @@ By default, `<script>` tags are processed by Astro.
...
+
+The `async` attribute is valuable for normal scripts as it prevents them from blocking rendering. However, module scripts already have this behavior and adding this attribute will make module scripts execute before the page has fully loaded. This is probably not what you want.
```
sarah11918, line 72:
```suggestion
:::note
The `async` attribute is valuable for normal scripts because it prevents them from blocking rendering. However, module scripts already have this behavior. Adding `async` to a module script will cause it to execute before the page has fully loaded. This is probably not what you want.
:::
```
Now, here's a case where I think this makes a great "note"! (Or "tip" even?)

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -62,7 +62,18 @@ By default, `<script>` tags are processed by Astro.
...
+
+#### Opting out of processing
+
+To prevent Astro from processing a script, you can use the `is:inline` directive.
```
sarah11918, line 76:
```suggestion
To prevent Astro from processing a script, add the `is:inline` directive.
```
(This is an excellent heading!)

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -73,18 +84,18 @@ To avoid bundling the script, you can use the `is:inline` directive.
...
 
 :::note
-Astro's default bundling behavior will be disabled in some situations. In particular, adding `type="module"` or any attribute other than `src` to a `<script>` tag will cause Astro to treat the tag as if it had an `is:inline` directive. The same will be true when the script is written in a JSX expression.
+Astro will not process your script tags in some situations. In particular, adding `type="module"` or any attribute other than `src` to a `<script>` tag will cause Astro to treat the tag as if it had an `is:inline` directive. The same will be true when the script is written in a JSX expression.
 :::
 
 📚 See our [directives reference](/en/reference/directives-reference/#script--style-directives) page for more information about the directives available on `<script>` tags.
 
-### Script loading
+### Include javascript files on your page
```
sarah11918, line 92:
```suggestion
### Running scripts from JavaScript files
```
I'm still not entirely sure about the wording, but something more like this seems to most precisely describe the action? Do you have something better for here?

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -46,7 +46,7 @@ In this example, adding the `<Hello />` component to a page will log a message t
 </script>
 ```
 
-### Script bundling
+### How Astro processes scripts
```
sarah11918, line 49:
```suggestion
### Script processing
```
Headings tend to be easier to scan when they are shorter, and "nouns" (things) people might care about. I think I'd have a preference for something like this?

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -62,7 +62,18 @@ By default, `<script>` tags are processed by Astro.
 </script>
 ```
 
-To avoid bundling the script, you can use the `is:inline` directive.
+#### All processed scripts become module scripts
```
sarah11918, line 65:
```suggestion
#### module scripts by default

All scripts processed and bundled by Astro become module scripts by default.
```
This feels more like a "key fact" than a "section heading". Can we have a more generic heading that is like a "thing", and then lead with the key fact?

Alternatively, this might mean that this really *isn't* its own section at all? (Or, the section is "advantages of ..."? Consider whether this really is A. Section. I kind of think maybe it's not?

Also, since this sentence then becomes kind of short, we can through "bundled" in here again because we do use that word a lot, and the reinforcement of terms people use/are familiar with helps.

sarah11918 (commented):
Some really nice additions here, @lilnasy ! See my comments below and let's see if we can really get this just right!

src/content/docs/en/guides/client-side-scripts.mdx
lilnasy: I agree, and I think note is still better - users do this hoping to delay things but they would be surprised to know it has the opposite effect.

src/content/docs/en/guides/client-side-scripts.mdx
lilnasy: I was thinking of the case where users have a javascript file written somewhere (`src/lib/smooth-scroll.js`). I think they would lookup how to "include".

sarah11918: OK, happy to defer to you on this one!

src/content/docs/en/guides/client-side-scripts.mdx
lilnasy: Yeah, I think this still applies.

src/content/docs/en/guides/client-side-scripts.mdx
lilnasy: The examples below show discrete files as well, so I feel good about "include".

src/content/docs/en/guides/client-side-scripts.mdx
lilnasy: What do you think about "Why `type=module`" as the heading? It is introduced as a bullet point a couple of lines above.

sarah11918: I don't *love* the "why" part but I like `type=module` better than anything so far.

Not everything needs to be a heading, and I'm trying to think of justification for this. Headings are useful for navigation, for separating content sections, and as visual breaks. Given that this is a list, we already have some natural visual break for the eyes.

I'm still not convinced this "section" is anything more than how Astro processes your scripts. This still seems content-wise to be in that category. 

And, it's not *awesome* practice for content/navigation to use questions as headers. When people are scanning, they want to know what this section is *about*. I think the most appropriate heading, if in fact we do need one (which I still am not convinced we do!) is "`type-module` benefits" -- that is what this section is *about*.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -62,7 +62,20 @@ By default, `<script>` tags are processed by Astro.
...
+This attribute makes the browser treat the script as a JavaScript module. This has several performance benefits:
+- Rendering is not blocked. The browser continues to process the rest of the HTML while the module script and its dependencies load.
+- The browser waits for HTML to be processed before executing module scripts. You do not need to listen for the "load" event.
+- `async` and `defer` attributes are unnecessary. Module scripts are always deferred.
+
+:::note
+The `async` attribute is valuable for normal scripts because it prevents them from blocking rendering. However, module scripts already have this behavior. Adding `async` to a module script will cause it to execute before the page has fully loaded. This is probably not what you want.
+:::
+
+#### Opting out of processing
```
sarah11918, line 76:
```suggestion
### Opting out of processing
```
Petition to bump this up, as something more important so it gets highlighted in the sidebar TOC!

lilnasy: Granted!

src/content/docs/en/guides/client-side-scripts.mdx
lilnasy: It does not have to be a section. I'd be happy with just a sentence that introduces the list.

lilnasy: Updated

sarah11918 (approved):
Another fine contribution!  🚀

PR #4978, opened 2023-10-05
i18n(fr): Adding `deploy/sst` French translation

thomasbnt:
#### Description (required)

Adding `deploy/sst` French translation. That it!

I am participating in the Hacktoberfest! 🚀☕ (*Yes, I alternate emojis because it's always the same message.*)
- [i18n(fr): Adding deploy/sst French translation](https://github.com/withastro/docs/commit/d40b9a6347d2e0d146eb3ca6e65b3b946ee5ea61)

PR #4979, opened 2023-10-05
i18n(en): Fix steps numbers in Quickstart part at `deploy/sst`

thomasbnt:
#### Description (required)

Seen this little misspell when I translated the page in French https://github.com/withastro/docs/pull/4978.


dreyfus92 (commented):
This is not necessary cause in MDX you can use this pattern and it will be rendered consecutively. This remained as it is for maintainability purposes 😁

PR #4983, opened 2023-10-06
Remove Netlify edge functions adapter

julienp:
#### Description (required)
Remove netlify edge functions adapter

Astro 3.0 removed this adapter https://github.com/withastro/adapters/blob/main/packages/netlify/CHANGELOG.md#major-changes

- Closes #
- Suggested label:


sarah11918:
Thank you for noticing this @julienp ! We actually merged a PR today that updates advice for this, since you can still use Netlify edge functions, just with the other adapter, and by adding middleware. There were a few things on this page that had to change, but we believe that it's all correct now! Thank you again for noticing, and I'll close this PR since it no longer matches the current page, but if you spot anything else, please open a new PR!  :raised_hands:

PR #4986, opened 2023-10-06
Update cloudflare.mdx wrangler publish -> deploy

sunapi386:
Wrangler deprecated publish, use deploy instead.

$ npx wrangler pages publish dist
▲ [WARNING] `wrangler pages publish` is deprecated and will be removed in the next major version.

  Please use `wrangler pages deploy` instead, which accepts exactly the same arguments.

- Closes #
- Suggested label:


ElianCodes (approved):
Thanks for keeping an eye on this @sunapi386!

PR #4989, opened 2023-10-07
✨ Improvement: rel="noreferrer" & png to webp

nermalcat69:
status

## ✨ Improvement
- added **rel="noreferrer"** to links which contain **target="_blank"**
- png to webp

Update as of 2021: All current versions of major browsers now automatically use the behavior of rel="noopener" for any target="_blank" link, nullifying this issue. See more at [chromestatus.com](https://chromestatus.com/feature/6140064063029248).


sarah11918:
This is not an appropriate PR as it combines two entirely different concepts and can't be easily reviewed. It also changes mutliple languages, which is not advised.

As this is the second PR I've had to close, I really suggest you file issues first, and seek confirmation that you're making helpful changes and confirm the way to proceed. I'm sorry that this PR cannot be accepted.

PR #5002, opened 2023-10-08
Reference:  `getEntries()` would return `Array<CollectionEntry>` - not a single entry

uceumice:
#### Description

Fixing a minor inconsistency in reference docs, so that `getEntries()` returns `Array<CollectionEntry>` - not a single entry.


sarah11918 (approved):
Thank you for this, @uceumice ! We appreciate you taking the time to contribute to the docs!  :partying_face:

PR #5031, opened 2023-10-11
use standard example.com domain for code sample

bennycode:
#### Description (required)

I found valuable information on how Astro generates anchor IDs in the German translation for [Heading IDs](https://docs.astro.build/de/guides/markdown-content/#%C3%BCberschriften-ids). However, this part was missing in the English version, which is why I am submitting this PR to address it.

- Suggested label: `hacktoberfest-accepted`

### DIFF
--- a/src/content/docs/en/guides/markdown-content.mdx
+++ b/src/content/docs/en/guides/markdown-content.mdx
@@ -119,6 +119,10 @@ I can link internally to [my conclusion](#conclusion) on the same page when writ
 I can use the URL `https://my-domain.com/page-1/#introduction` to navigate directly to my Introduction on the page. 
 ```
 
+Astro uses [github-slugger](https://github.com/Flet/github-slugger) to automatically assign generated IDs to all headings in Markdown files. If a heading already has a custom ID, it will be preserved and not overwritten.
+
+The automatic IDs are added after all other Markdown plugins have been executed. So, if you want to use a plugin like [rehype-toc](https://github.com/JS-DevTools/rehype-toc), which requires heading IDs beforehand, you should also include your own slugging plugin like [rehype-slug](https://github.com/rehypejs/rehype-slug).
+
 ### Escaping special characters
 
 Certain characters have a special meaning in Markdown and MDX. You may need to use a different syntax if you want to display them. To do this, you can use [HTML entities](https://developer.mozilla.org/en-US/docs/Glossary/Entity) for these characters instead.
--- a/src/content/docs/en/guides/markdown-content.mdx
+++ b/src/content/docs/en/guides/markdown-content.mdx
@@ -116,12 +116,8 @@ I can link internally to [my conclusion](#conclusion) on the same page when writ
 
 ## Conclusion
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Thank you for this! I'm going to have to do a bit of digging in the history, because usually if there's a discrepancy between languages, it means the English was updated and the translation did not get updated to match. It would be a little unusual in our system for the opposite to happen. So, it's *possible* that this information was removed from the English because it's no longer true!

We'll do some digging and verify to be sure! :raised_hands:

src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -119,6 +119,10 @@ I can link internally to [my conclusion](#conclusion) on the same page when writ
 I can use the URL `https://my-domain.com/page-1/#introduction` to navigate directly to my Introduction on the page.
```
sarah11918, line 119:
```suggestion
I can use the URL `https://example.com/page-1/#introduction` to navigate directly to my Introduction on the page. 
```

src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -119,6 +119,10 @@ I can link internally to [my conclusion](#conclusion) on the same page when writ
 I can use the URL `https://my-domain.com/page-1/#introduction` to navigate directly to my Introduction on the page. 
 ```
 
+Astro uses [github-slugger](https://github.com/Flet/github-slugger) to automatically assign generated IDs to all headings in Markdown files. If a heading already has a custom ID, it will be preserved and not overwritten.
+
+The automatic IDs are added after all other Markdown plugins have been executed. So, if you want to use a plugin like [rehype-toc](https://github.com/JS-DevTools/rehype-toc), which requires heading IDs beforehand, you should also include your own slugging plugin like [rehype-slug](https://github.com/rehypejs/rehype-slug).
+
```
sarah11918, line 125:
```suggestion
```

sarah11918 (commented):
Hi @bennycode , it turns out this information was intentionally removed from here and some information reorganized intentionally. Apparently we do still use `github-slugger`, but with modifications that would be a little much to explain here so it makes more sense to keep that out so it's not misleading info. The information about the plugin usage was moved down to another section, so we do think the current page contains all the info we wanted to include, in the places we wanted to include it.

Even though we've established that the current version of the English site is our preferred content, I appreciate you bringing this to our attention!  Since you're Hacktoberfesting, I've made an alternate addition to this page that we can give you to count for credit! We have been wanting to make sure our code snippet example domains follow standard advice and use `example.com`, and I spotted that right next to your original suggestion. 

So, I've made that change here and we can get you not only your first docs PR, but also a Hacktoberfest PR! It's still a valuable contribution, and we spotted it thanks to you!  :raised_hands:

bennycode:
Hi @sarah11918, thank you so much for your detailed response. I am truly amazed by the thorough documentation process at Astro. It's absolutely top-notch. 👍 

I just wanted to express my gratitude for giving me credits for Hacktoberfest. Your kindness is greatly appreciated. Wishing you a fantastic weekend. 💫

PR #5039, opened 2023-10-11
i18n(zh-CN): improve translation for `reading-time.mdx`

Genteure:
#### Description (required)

Mostly rewording and restructuring sentences.
Source: https://github.com/withastro/docs/blob/main/src/content/docs/en/recipes/reading-time.mdx

While doing the translation I found the npm package `reading-time` does not seems to output a `text` property, maybe this doc is outdated? I was checking if their `text` output is translated or not since it was translated in this page.


liruifengv (approved):
LGTM

Yan-Thomas (approved):
LGTM! Thanks peeps 🙌

PR #5043, opened 2023-10-11
Document `--outDir` CLI flag

delucis:
#### Description (required)

- This PR updates the CLI reference to mention the new `--outDir` CLI flag added in https://github.com/withastro/astro/pull/8808
- While there, I also removed the flags subsection under `astro build`, which was left in a slightly odd shape when the `--drafts` flag was removed. It now links to common flags with the same style as the docs for `astro preview` do.


sarah11918 (approved):
Looks great, @delucis !

PR #5045, opened 2023-10-11
Removing misleading 'including'

lschierer:
the use of 'including' here suggested other supported formats. This should make it clear that only these two formats are currently supported.

- Closes #
- Suggested label:


sarah11918 (approved):
Thank you for spotting an improvement, and following through to fix it @lschierer ! 💪 

Welcome to Team Docs!  🥳

PR #5052, opened 2023-10-12
Remove outdated shiki pnpm note

bluwy:
#### Description (required)

Remove note where `shiki` might need to be installed in your project when using pnpm. This is no longer the case since Astro 3.3 (releasing today). The PR (https://github.com/withastro/astro/pull/8502) switches to shikiji which doesn't have this caveat anymore.


sarah11918 (approved):
Thanks so much for this, @bluwy ! Always happy when code improvements allow us to remove a cautionary note! :raised_hands:

PR #5057, opened 2023-10-12
Update microsoft-azure.mdx to expand description and add community links

### LINKED ISSUE #4891: ⚠️ Issue with Azure Deployment Instructions
On https://docs.astro.build/en/guides/deploy/microsoft-azure/

You must also set the engine in your `packages.json` as the static website build defaults to node 16 which is not compatible with astro

```
  "engines": {
    "node": ">=18.0.0"
  },
```

https://edi.wang/post/2022/1/27/how-to-specify-nodejs-version-when-building-azure-static-web-app

blueboxes:
#### Description (required)
Added more descriptions and links along with a known issue/workaround

- Closes #4891


sarah11918:
Thank you for this contribution, @blueboxes ! I took an editing pass to conform to Astro docs style, and it looks great!

Before merging, I do want to confirm that the Node Version issue is still a problem. Node 16 reached end of life a couple of weeks ago, and I would be surprised if anyone were still using Node 16 as a default. This may have changed very recently, so it should be checked again to see whether this is still the case.

Can you confirm that and report back? Maybe we don't even need the known issues section anymore!

sarah11918:
Adding the `help-wanted` and `confirm-behaviour` labels here, because I'd love for anyone to confirm that the Node version is still a problem! (See above comment)

blueboxes:
I have been able to re-create and confirm myself however I have been looking to see if they have plan to upgrade but could not found anything.

sarah11918:
@blueboxes That's good enough for me! It is just surprising, since no one is supposed to be using those older versions anymore and the EOL *just* happened very recently. That's the only reason I wanted to double-check: maybe something happened between the time you wrote the guidance, and I reviewed it!  😄 

No reason not to get this live in docs then, although I do suspect at some point that issue will no longer be the case and we'll need to remove that. I'm sure someone will tell us when it happens.  😅

PR #5116, opened 2023-10-16
Fix typo in template directives

sasoria:
#### Description (required)

This pull request fixes a typo in the reference section for Template Directives.


sarah11918:
Thanks @sasoria , but these aren't actually typos! This is an acceptable way in English to say "the JS belonging to the component". It's possessive, it's not actually "the component with JavaScript."

It happens to be a nice form in English because it avoids the extra `'s` punctuation or wordier "belonging to" phrases. So, we'll keep it like this for now because it's currently nicer than the immediate alternatives and doesn't yet need a rewrite to be accurate. When I revisit this page for future refreshing (which will be soonish, because I will want to add our View Transitions directives here too), I'll keep in mind that maybe the wording could be improved.

PR #5155, opened 2023-10-19
Add insert on import

jacc:
#### Description (required)
In the documentation for SSR, the first example about adding `nodejs` to your Astro config is not highlighted as an import, whereas the second example (converting a static site) highlights `nodejs` as an import. This PR fixes that.


sarah11918 (approved):
Love it, and you're absolutely right.

Happy to welcome you and your eagle eyes to Team Docs! 🫡

PR #5183, opened 2023-10-23
Add documentation on making dev overlay plugins

Princesseuh:
#### Description (required)

Docs for https://github.com/withastro/astro/pull/8757.

Since the dev overlay API is *very* early still, the content is a bit light. But, this should provide a basic guide over the different hooks and features that should still be appreciated by the few people who will make dev overlay plugins already.

This is intentionally only in the references while the dev overlay is so early and very much lacking in features. In a very close future, I'll be sending a PR to add more references to it throughout the docs where relevant. (Right now, we don't mind too much people not using it, so no big marketing needed until it's very cool!)


src/content/docs/en/reference/dev-overlay-reference.mdx
```diff
@@ -0,0 +1,260 @@
...
+  }
+```
+
+`state` can be either `true` or `false`. When `true`, the plugin icon will be highlighted using a red dot. Conversely, when `false`, the highlight will be removed.
+
+## Client-Server Communication
+
+Using [Vite's methods for client-server communication](https://vitejs.dev/guide/api-plugin.html#client-server-communication), dev overlay plugins can communicate with the server.
+
+In addition to being able to send and receive arbritary messages, the dev overlay also sends the following messages, where `PLUGIN_ID` is the ID of a plugin:
```
sarah11918, line 138:
```suggestion
In addition to being able to send and receive custom messages, the dev overlay also sends the following messages, where `PLUGIN_ID` is the [plugin's ID](#id-string):
```
I'm not sure "custom" is the right word, but "arbitrary" sounds a lot like "random" so I'm wondering if we can choose something that more means "messages you've written for your plugin"...

> ... your plugin's messages ...

?

src/content/docs/en/reference/dev-overlay-reference.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+This message is sent when the plugin is initialized. The data for this message is empty.
+
+```ts title="integration.ts"
+
+```
+
+### `astro-dev-overlay:PLUGIN_ID:toggle`
+
+This message is sent when the user clicks on the plugin icon in the UI. The data for this message is a boolean indicating whether the plugin is enabled or not.
```
sarah11918, line 150:
```suggestion
This message is sent when the user clicks on the plugin icon in the UI. The data for this message is [the plugin's state](#state-boolean) indicating whether the plugin is enabled or not.
```
Would this be correct? This might tie it back nicely to the property!

Also, if we're showing an empty message above, is it worth showing the message being either true or false here, too?

src/content/docs/en/reference/dev-overlay-reference.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+:::note
```
sarah11918, line 152: Just a note that I want to come back to this note!

src/content/docs/en/reference/dev-overlay-reference.mdx
```diff
@@ -0,0 +1,260 @@
...
+The built-in `connection` event from Vite fire *before* dev overlay plugins are initialized. As such, trying to receive events from this event will not work inside plugins. Instead, use the `astro-dev-overlay:PLUGIN_ID:init` event.
+:::
+
+## UI Toolkit
+
+The dev overlay includes a set of webcomponents that can be used to build plugins with a consistent look and feel.
+
+### `astro-dev-overlay-window`
+
+Shows a window with optionally a title and an icon. `windowIcon` can either be a string of a SVG file or an icon from [the icon list](/en/reference/dev-overlay-plugin-api/#icons).
```
sarah11918, line 162:
```suggestion
Shows a window with optionally a title and an icon.

`windowTitle` is a string that will be shown at the top of the overlay window. `windowIcon` can either be a string of a SVG file or an icon from [the icon list](#icons).
```
I agree this maybe feels stupidly unnecessary, but it is a part of the API and it should be mentioned explicitly somewhere.

src/content/docs/en/reference/dev-overlay-reference.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+```html
+<astro-dev-overlay-window windowTitle="MyWindows" windowIcon="astro:logo">
+  <p>My content</p>
+</astro-dev-overlay-window>
+```
+
+### `astro-dev-overlay-card`
+
+Shows a card with a icon. Optionally, if `link` is passed, the card will be clickable and will open the link in a new tab.
```
sarah11918, line 174:
```suggestion
Shows a card with an [`icon`](#icons). Optionally, if a `link` is passed, the card will be clickable and will open the link in a new tab.
```
We have a section, so I like linking to it!

sarah11918 (commented):
This is great @Princesseuh ! I almost feel like I could write a plugin... :smile: 

Left some suggestions, and I do think changing the main filename to `dev-overlay-plugin-reference.mdx` makes a lot of sense, since this really is geared towards writing plugins specifically.

See what you think about some of my suggestions below, and of course, change whatever is necessary!

src/content/docs/en/reference/dev-overlay-reference.mdx
```diff
@@ -0,0 +1,260 @@
...
+highlight.style.width = `${rect.width + 15}px`;
+highlight.style.height = `${rect.height + 15}px`;
+highlight.icon = 'astro:logo';
+```
+
+### `astro-dev-overlay-tooltip`
+
+Shows a tooltip with different sections. This component is set to `display: none;` by default and can be made visible using a `data-show="true"` attribute.
+
+Sections are defined using the `sections` property. This property is an array of objects with the following shadpw:
```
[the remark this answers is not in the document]
Princesseuh: I have no idea what happened.

sarah11918:
@Princesseuh This all looks amazing! I'm going to just re-visit that one `note` today, and was also wondering whether it's possible/makes sense to actually show the two included icons visually at the bottom of the page. Otherwise, this is pretty much ready for tomorrow!

Princesseuh:
> @Princesseuh This all looks amazing! I'm going to just re-visit that one `note` today, and was also wondering whether it's possible/makes sense to actually show the two included icons visually at the bottom of the page. Otherwise, this is pretty much ready for tomorrow!

Hmm, that would definitely be nice, but I'm not sure *how*. Probably that the docs repo could take the icon's codes directly from the main repo and use them that way. Since it's still very possible for that file to move around and be changed fairly consequentially, I prefer doing it in a separate effort once the refactors are done!

src/content/docs/en/reference/dev-overlay-plugin-reference.mdx
```diff
@@ -0,0 +1,313 @@
...
+    server.ws.on("astro-dev-overlay:super-plugin:toggle", (data) => {
+      console.log(`My plugin is now ${data.state ? "enabled" : "disabled"}!`);
+    });
+  },
+  // ...
+}
+```
+
+:::note
+The built-in `connection` event from Vite fire *before* dev overlay plugins are initialized. As such, trying to receive events from this event will not work inside plugins. Instead, use the `astro-dev-overlay:PLUGIN_ID:init` event.
```
sarah11918, line 198:
```suggestion
The built-in `connection` event from Vite fires **before** dev overlay plugins are initialized and therefore cannot be used directly by plugins. Instead, use the `astro-dev-overlay:PLUGIN_ID:init` event.
```
I know this is a little more generic, but would this be sufficient here?

Princesseuh: That's perfectly fine with me!

sarah11918 (approved):
Ready for minor release!  :partying_face:

PR #5190, opened 2023-10-24
[api-reference] add more built-in code components

### LINKED ISSUE #4442: Built-in Components reference is outdated
### 📚 Subject area/topic

astro:components

https://docs.astro.build/en/reference/api-reference/#built-in-components

The reference for built-in components is outdated for v3.0. 

The section should probably include at least a reference for the `ViewTransitions` component, potentially linking back to https://docs.astro.build/en/guides/view-transitions/

```js
import { ViewTransitions } from 'astro:transitions'
```

n/a

sarah11918:
#### Description (required)

Documents our "newer" built-in components that were never added to the API reference page.

- Closes #4442

### DIFF
--- a/src/content/docs/en/guides/images.mdx
+++ b/src/content/docs/en/guides/images.mdx
@@ -5,6 +5,8 @@ i18nReady: true
 ---
 import Since from '~/components/Since.astro';
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
+import Badge from '~/components/Badge.astro';
+import RecipeLinks from "~/components/RecipeLinks.astro";
 
 
 Astro provides several ways for you to use images on your site, whether they are stored locally inside your project, linked to from an external URL, or managed in a CMS or CDN!
@@ -84,10 +86,6 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
 />
 ```
 
-Currently, the built-in assets feature does not include a `<Picture />` component.
-
-Instead, you can [generate images or custom components using `getImage()`](#generating-images-with-getimage) that use the HTML image attributes `srcset` and `sizes` or the `<picture>` tag [for art direction or to create responsive images](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#art_direction).
-
 #### Properties
 
 ##### src (required)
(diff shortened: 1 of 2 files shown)


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -779,7 +779,56 @@ Pagination will pass a `page` prop to every rendered page that represents a sing
...
-## Content Collections
+
+
+
+
+## `import.meta`
+
+All ESM modules include a `import.meta` property. Astro adds `import.meta.env` through [Vite](https://vitejs.dev/guide/env-and-mode.html).
+
+**`import.meta.env.SSR`** can be used to know when rendering on the server. Sometimes you might want different logic, for example a component that should only be rendered in the client:
```
Yan-Thomas, line 790: Here we'd need a comma after "for example", but since we also have the same sentence structure, I would rather change it to another word
```suggestion
**`import.meta.env.SSR`** can be used to know when rendering on the server. Sometimes you might want different logic, like a component that should only be rendered in the client:
```

src/content/docs/en/guides/images.mdx
Yan-Thomas, line 1: Are we changing anything here? If we're doing nothing here, I would remove this page from the PR to avoid marking this as outdated in the i18n tracker. No pressure though, if you can't remove this weird git behavior we can take over the tracking system when we merge this.

Yan-Thomas (commented):
This LGTM, just added some small nits to consider.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1154,6 +1191,133 @@ This component provides language-specific syntax highlighting for code blocks by
...
+Format to use as a fallback value for the `<img>` tag. Defaults to `.png` for static images, `.gif` for animated images, and `.svg` for SVG files.
+
+##### `pictureAttributes`
+
+An object of attributes to be added to the `<picture>` tag. Use this property to apply attributes to the outer `<picture>` element itself. Attributes applied to the `<Picture />` component directly will apply to the inner `<img>` element, except for those used for image transformation.
+
+
+### `<Content />`
+
+A component used to render the content of a [content collection entry](/en/guides/content-collections/#what-are-content-collections).
```
VoxelMC, line 1286:
```suggestion
A generic component used to render the content of a [content collection entry](/en/guides/content-collections/#what-are-content-collections).
```

Just a thought, as it is amorphous? Might be unnecessary to say, though.

src/content/docs/en/guides/images.mdx
sarah11918: No, we're not, and I have no idea why it's showing up as changed! :/

sarah11918: OK, managed to get rid of this one!

PR #5110, opened 2023-10-16
i18n(ko-KR): update `layouts.mdx`

jsparkdev:
#### Description (required)

- update `layouts.mdx`
- I am participating in Hacktoberfest, please label it (hacktoberfest-accepted)

#### Related issues & labels (optional)

N/A

PR #5202, opened 2023-10-26
i18n(ko) typescript.mdx link heading fix

sarah11918:
#### Description (required)

Corrects a link pointing to an English heading instead of a translated one.

If I'm right, all the link checks should pass.


jsparkdev:
I wasn't aware of the existence of this PR, so I added the same actions as this PR to #5110 .

Could this cause any problems?

ElianCodes (approved):
What a tough catch! Thanks @sarah11918! LGTM

jsparkdev:
> @jsparkdev It shouldn't be a problem! Checks should still pass over there, too. If not, I'll tag you in there to fix! 🙌 Thanks for being so responsive!

I'm so glad! Thank you :+1:

PR #5213, opened 2023-10-27
[nav.ts] New sidebar sections: built-ins and add-ons

sarah11918:
#### Description (required)
Adds two new sections to our main sidebar: Built-ins and Add-ons.

This intends to better highlight the natural progression of learning about Astro: Basics > not basic, but included stuff you probably want > things you can add-on that are part of our  story: framework islands and hyping our rendering options

Play around and see what you think!

Current Docs -> Proposal: 
Screenshot 2023-10-27 10 30 52Screenshot 2023-10-27 10 29 57


src/i18n/en/nav.ts
```diff
@@ -46,56 +46,52 @@ export default [
...
+	},
+
+	{ text: 'Add-ons', header: true, type: 'learn', key: 'examples' },
+	{ text: 'Add integrations', slug: 'guides/integrations-guide', key: 'guides/integrations-guide' },
+	{
+		text: 'UI Frameworks',
+		slug: 'core-concepts/framework-components',
+		key: 'core-concepts/framework-components',
 	},
+	{ text: 'Server-side Rendering', slug: 'guides/server-side-rendering', key: 'guides/server-side-rendering' },
```
natemoo-re, line 74: This is my only complaint! Seems like Server-side Rendering should be under Built-ins. I know this section is getting revamped, but I don't think it's an add-on.

src/i18n/en/nav.ts
```diff
@@ -46,56 +46,52 @@ export default [
...
 	},
-	{
-		text: 'UI Frameworks',
-		slug: 'core-concepts/framework-components',
-		key: 'core-concepts/framework-components',
-	},
-	{ text: 'Routing', slug: 'core-concepts/routing', key: 'core-concepts/routing' },
-	{ text: 'Markdown & MDX', slug: 'guides/markdown-content', key: 'guides/markdown-content' },
+
+	{ text: 'Built-ins', header: true, type: 'learn', key: 'examples' },
```
natemoo-re, line 55: Is `key: 'examples'` correct? I don't totally grok what `key` is for to be honest!

natemoo-re (approved):
I think this clarifies the structure a lot! Big fan of this change. 

Don't consider my comments blocking at all! Approved.

src/i18n/en/nav.ts
natemoo-re: Ah, I see there is discussion happening in Discord about this point.

ElianCodes:
Big fan of the change!

ElianCodes (approved):
The new sidebar feels and looks really nice!

src/i18n/en/nav.ts
sarah11918: Good old copy paste! :sweat_smile:

src/i18n/en/nav.ts
sarah11918: Mirroring discussion on Discord: because it requires adding an adapter, conceptually this falls closer to UI Frameworks because it doesn't work without something that needs to be `astro add`ed. So, literally, adding something for SSR to work **is** what makes it an add on. This category would be restricted to things very closely coupled to Astro's core offering, like our official packages where we also document their usage.

So noted, but as @delucis mentioned, from the outside looking in, "all you know is adding X makes SSR work" and I don't think this placement diminishes that it is something that is built into Astro core (but is helpful in understanding that it needs to be "unlocked" in order to actually use it.

And yes, I'd like this page to eventually be replaced with our Rendering Modes page, which maybe DOES go into basics, and then maybe we also keep an SSR page here that is devoted mostly to adding the adapter, marking individual pages in or out of pre-rendering etc.

wassfila:
- I like how this new clustering helps already place the user context when looking for docs, it just brings the wish for more, for example adding a cms is also adding integrations so it feels like it should be in add-ons but then it's also a recipe.

- same comment as above regarding the ssr as add-ons, impression jumped to me when I saw it. I understand that some adapters are addon, but node js is built in I think, but I might be wrong.

- one thing that triggers thoughts about add-ons vs built-ins, users might confuse them with officially supported and maintained vs provided by the community or third party, which come in mind when reading the heading "Official integrations", so it's an "official add on". If only there was a good spot where it's poosible to explain what is built in and what is add on, maybe in the intro of each first page in each category.
those are just thoughts, all in all get much better over time.

sarah11918:
@wassfila Even the Node adapter has to be added! (FYI)

sarah11918:
Final decision after consultation with the team:

SSR label in sidebar changed to "SSR Adapters" in keeping with the idea of the adapter being the "add on", and less to imply stuff about our actual rendering modes.

This is a non-content/breaking change that only moves items in the sidebar. But, the plan is to augment this and allow us to include a "Rendering Mode" explanatory page in basics, and eventually this SSR page will in fact be geared towards enabling and using the adapters

PR #5214, opened 2023-10-27
[i18nIgnore] remove specific heading from links that will change

sarah11918:
#### Description (required)

Future proofing some links! Instead of going to a specific spot on the page (which will be near the top of the new page anyway), just goes to the page itself! Will make other PRs link checks pass


ElianCodes (approved):
Looks perfect!

sarah11918:
I got all the translations that exist, I believe!

PR #5215, opened 2023-10-27
upgrade to v2 - future proof link

sarah11918:
#### Description (required)
The link for this very old page (upgrade from v1 to v2) is going to eventually cause problems. This removes the specific heading that will change and simply directs to the page itself.


TheOtterlord (approved):
lgtm!

PR #5222, opened 2023-10-28
fix import meta env reference in env docs

FredKSchott:
#### Description (required)

I ended up findings some Vite docs on how to fix this, and ended up trying a few things before realizing that this worked in dev (and assuming preview/prod as well).

Before: `Because Vite statically replaces import\u002Emeta.env, you cannot access it with dynamic keys like import\u002Emeta.env[key].`

After: `Because Vite statically replaces import.meta.env, you cannot access it with dynamic keys like import.meta.env[key].`


at-the-vr (approved):
sounds accurate. unless `VITE_` is used as a prefix, env variables are not accessible in client side [link to production replacement by Vite ](https://vitejs.dev/guide/env-and-mode.html#production-replacement)

sarah11918 (approved):
This works!

PR #5232, opened 2023-10-29
Add tab component to Keystatic CMS guide

jacobthesheep:
#### Description (required)

Add tab component to tip in Keystatic page

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/cms/keystatic.mdx
+++ b/src/content/docs/en/guides/cms/keystatic.mdx
@@ -14,9 +14,23 @@ import Aside from '~/components/Aside.astro'
 :::tip
 If you're starting a **new Astro + Keystatic project from scratch**, you can use the [Keystatic CLI](https://keystatic.com/docs/quick-start#keystatic-cli) to generate a new project in seconds:
 
-```bash
-npm create @keystatic@latest
-```
+<PackageManagerTabs>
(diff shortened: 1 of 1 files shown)


at-the-vr:
How come there are failed Checked Links ? This PR doesn't even include one 🤔. At least the fragment and scripts are displayed correctly on [Deployed Preview](https://deploy-preview-5232--astro-docs-2.netlify.app/en/guides/cms/keystatic/) 😊

sarah11918:
Thanks for adding this, @jacobthesheep!

And thanks for checking the preview and reporting back, @at-the-vr ! That's super helpful!

I've seen these failed link checks before, and these particular ones that don't have anything to do with the PR are usually  just a fluke and will pass if I rerun the tests. Doing that now... :raised_hands:

sarah11918:
Checks have passed! I will tag @simonswiss in here because I realized that even Keystatic's own docs only show `npm` and not the other package managers. There's always the possibility that they haven't tested and/or don't (want to) support the other package managers... but I'll merge this now and I guess the best way to find out is to have someone try and complain!  :joy: 

@jacobthesheep , did you try testing creating with `pnpm` and `yarn`? It *seems* safe to me, but Simon is now looped in, and if we get complains, we can revert to the simple `npm` install instructions!

PR #5233, opened 2023-10-30
chore: test view transitions

afonsojramos:
#### Description (required)

Test out ViewTransitions for the Astro Docs. Currently navigating between pages creates a flash that should be fixed before this is ready for review.

- Related to: https://github.com/withastro/astro/issues/8948

### DIFF
--- a/src/components/Header/ThemeToggleButton.tsx
+++ b/src/components/Header/ThemeToggleButton.tsx
@@ -43,17 +43,31 @@ const icons = [
 const ThemeToggle = ({ labels, isInsideHeader }: Props) => {
 	const [theme, setTheme] = useState<'light' | 'dark'>();
 
-	useEffect(() => {
-		setTheme(document.documentElement.classList.contains('theme-dark') ? 'dark' : 'light');
-	}, []);
+	const getTheme = () => {
+		if (typeof localStorage !== 'undefined' && localStorage.getItem('theme')) {
+			return localStorage.getItem('theme')!;
+		}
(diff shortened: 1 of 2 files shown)


sarah11918:
Hey @afonsojramos ! I can see that this is just a test/draft right now, but heads up that we will very soon be converting Astro Docs over to Starlight, so we won't be adding View Transitions to this repo as we'll soon be using a different model entirely.

You're welcome of course to keep experimenting with your own fork of docs site, but if you're thinking of contributing back to docs site infrastructure in the future, you could consider experimenting with a Starlight template! (I honestly don't know whether we are thinking of adding View Transitions there, so it might be worth discussing either in Issues or in the #starlight channel in our Discord where the Starlight development is happening.

So I'm closing this PR as it's not something we're going to add to Astro Docs, but I'll also leave a comment in your issue regarding the theme flash!

afonsojramos:
Hi @sarah11918! Just to clarify, what is the reason for not adding ViewTransitions in the meantime? Couldn't we maintain ViewTransitions in Starlight either way? I just think that ViewTransitions is, truly, very interesting and a big part of what makes Astro great, so not seeing it present in one of the biggest Astro-based sites seems a bit weird.

sarah11918:
Hi! We're just all-hands-on-deck with the migration to Starlight right now and we just don't have the resources to properly test, troubleshoot especially for something that won't be in production very long. That's all!

Our Docs site is pretty complicated, so we're just not changing any infra anymore so we can focus on what's going to integrate with our new Starlight version.

Glad you got your issue sorted!  🙌

PR #5237, opened 2023-10-31
fix: import images inside the custom image component recipe

Princesseuh:
#### Description (required)

Local images have to be imported, we don't support linking to images in `public` like that at the moment


sarah11918 (approved):
Perfect, thank you so much @Princesseuh

PR #5244, opened 2023-11-01
Adds me as an author for an article that was already listed

DBozhinovski:
Adds my name to resources, as it wasn't listed

#### Description (required)
Added my name to an article listed in WordPress resources that I wrote, but I wasn't mentioned by name (every other resource has its author mentioned by name).


sarah11918 (approved):
Looks great! Not sure how that oversight was made, but I'm so glad you came to correct it.  🙌  Let's get you your credit, and officially into Team Docs!  🥳

PR #5268, opened 2023-11-02
fix: `astro check` watches all files, not just Astro files

Princesseuh:
#### Description (required)

What the title says, the reference was outdated for Astro 3

### DIFF
--- a/src/content/docs/en/reference/cli-reference.mdx
+++ b/src/content/docs/en/reference/cli-reference.mdx
@@ -245,7 +245,7 @@ Use these flags to customize the behavior of the command.
 
 The command will watch for any changes in your project, and will report any errors.
 
-📚 Read more about [TypeScript support in Astro](/en/guides/typescript/).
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -243,11 +243,7 @@ Use these flags to customize the behavior of the command.
...
 #### `--watch`
 
-The command will watch for any changes to `.astro` files, and will report any errors.
-
-:::note
-This command only checks types within `.astro` files.
-:::
+The command will watch for any changes in your project, and will report any errors.
 
 📚 Read more about [TypeScript support in Astro](/en/guides/typescript/).
```
sarah11918, line 248:
```suggestion
📚 Read more about [Type checking in Astro](/en/guides/typescript/#type-checking).
```

sarah11918 (approved):
One thought, makes sense to go directly to the Type checking section of the TypeScript page here, I think?

Thanks for jumping on this!

PR #5284, opened 2023-11-05
[i18nIgnore] Fix pnpm command typos

mingjunlu:
#### Description

I found that many package installation snippets use `pnpm install <pkg>` instead of `pnpm add <pkg>`.

According to pnpm's documentation:
- [`pnpm install`](https://pnpm.io/cli/install) is used for installing **all** dependencies for a project
- [`pnpm add <pkg>`](https://pnpm.io/cli/add) is used for installing certain package(s)

This PR changes those to the latter ✍️

#### Related issues & labels

### DIFF
--- a/src/content/docs/en/guides/images.mdx
+++ b/src/content/docs/en/guides/images.mdx
@@ -656,7 +656,7 @@ If the image is merely decorative (i.e. doesn’t contribute to the understandin
 When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like `pnpm`, you may need to manually install Sharp into your project even though it is an Astro dependency:
 
 ```bash
-pnpm install sharp
+pnpm add sharp
 ```
 :::
 
--- a/src/content/docs/en/guides/images.mdx
+++ b/src/content/docs/en/guides/images.mdx
@@ -656,7 +656,7 @@ If the image is merely decorative (i.e. doesn’t contribute to the understandin
 When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like `pnpm`, you may need to manually install Sharp into your project even though it is an Astro dependency:
 
 ```bash
-pnpm add sharp
+pnpm install sharp
 ```
 :::
 
--- a/src/content/docs/en/install/manual.mdx
+++ b/src/content/docs/en/install/manual.mdx
@@ -71,7 +71,7 @@ Astro must be installed locally, not globally. Make sure you are *not* running `
   </Fragment>
   <Fragment slot="pnpm">
   ```shell
-  pnpm add astro
+  pnpm install astro
   ```
   </Fragment>
   <Fragment slot="yarn">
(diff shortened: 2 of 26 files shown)


dreyfus92 (approved):
You're right, there is a difference between `pnpm install` and `pnpm add` when using the pnpm package manager.

So in general:

- Use `pnpm install` when setting up a brand new project, or when you want to fully reinstall all dependencies.
- Use `pnpm add` when you just want to add a new dependency to an existing project.

The snippets likely use pnpm install because they are showing how to install a package in a generic, standalone way. But in the context of an existing project, pnpm add is preferrable for adding new packages.

Good catch on the distinction! Using the right command for the situation can save time and avoid unnecessary installs.

src/content/docs/en/install/manual.mdx
```diff
@@ -71,7 +71,7 @@ Astro must be installed locally, not globally. Make sure you are *not* running `
   </Fragment>
   <Fragment slot="pnpm">
   ```shell
-  pnpm install astro 
+  pnpm add astro
```
dreyfus92, line 74: I wouldn't replace install here, since this guide is install a brand new project.

src/content/docs/en/guides/images.mdx
```diff
@@ -656,7 +656,7 @@ If the image is merely decorative (i.e. doesn’t contribute to the understandin
 When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like `pnpm`, you may need to manually install Sharp into your project even though it is an Astro dependency:
 
 ```bash
-pnpm install sharp
+pnpm add sharp
```
dreyfus92, line 659: I'm not completely sure about this one, i'd stick with a fresh re install of sharp tho.

dreyfus92 (commented):
Thank you for handling this @mingjunlu, I've left you a couple of suggestions let me know your thoughts.

sarah11918:
Thanks so much @mingjunlu for helping us get the commands just right. And thank you for this excellent review, @dreyfus92 with extra context and for thoughtfully considering each individual change. Go Team!  🙌 

I'm going to add the title keyword so that this PR touching multiple files doesn't affect our translation tracking status, and I'm going to apply the suggestions so this can be merged and helping people in docs right away!

@mingjunlu - If you find any error in the published version of this PR, please do make another one and let us know! I do agree with @dreyfus92 's assessment that in some cases, install is preferable, and will make those changes.

Thanks again, and welcome to Team Docs!  🥳

src/content/docs/en/guides/images.mdx
sarah11918, line 659:
```suggestion
pnpm install sharp
```

src/content/docs/en/install/manual.mdx
sarah11918, line 74:
```suggestion
  pnpm install astro
```

mingjunlu:
Woah! To be honest, I didn't expect to receive response that fast! 🚀
@dreyfus92 @sarah11918 Thank you both for taking care of this PR and providing such valuable feedback. I'm glad that I could help 😊

src/content/docs/en/install/manual.mdx
mingjunlu: Got it! Thanks for pointing it out 💡

src/content/docs/en/guides/images.mdx
mingjunlu: No problem! Let's keep the original one

PR #5294, opened 2023-11-06
Guides Gitlab update .gitlab-ci.yml example

Ethirallan:
#### Description (required)
This commit addresses a syntax error in the .gitlab-ci.yml file that was causing the pipeline to fail. The error was due to incorrect indentation in the YAML file.

Additionally, the `before_script` block, which was previously defined globally, has been moved inside the `pages` job. This change provides better flexibility for users who have multiple stages in their pipelines, as the `before_script` will now only run before the `pages` job, rather than before every job in the pipeline.

These changes should resolve the pipeline failure and prevent potential issues in multi-stage pipelines.


Ethirallan:
Hello @dreyfus92, thank you for a fast reply. We just started migrating our docs to Astro (starlight) and successfully deployed the first one on Friday. Today we had problems with the second one (copied again pipeline from the docs) and had a failed pipeline because of the invalid yaml syntax: `image`, `before_script` and `pages` stage were not indented on the same level. You can double check this with yaml validator. It was an easy fix and after testing with `gitlab-runner` I prepared this PR.

On the other hand, I'm now thinking about if it would be a good idea to move `image` inside `pages` stage as well. This way the whole snippet can me easily copied to any exiting pipeline as well as used as a standalone one. What is you opinion on this?

dreyfus92:
Thanks for the additional context about starlight @Ethirallan . I appreciate you taking the time to explain the full background and proposal.

You're absolutely right that the YAML indentation was invalid in the original - the image, `before_script`, and pages all needed to be aligned under stages for it to be valid syntax. I'm glad the fix resolved the pipeline failure.

Regarding moving image under the pages stage, I think that's a great idea for improving reusability as you mentioned. Defining the runner image only once where it's needed makes the config more modular. People can easily copy just the pages stage to add Astro building to an existing pipeline, without having to define the image globally.

The only consideration is if you have any other stages in the pipeline that may also depend on that image. But for a simple Astro docs pipeline focused just on building pages, consolidating everything into the one pages stage makes total sense.

Overall I think that change would provide a cleaner, more portable Astro config snippet that can drop easily into other pipelines. Feel free to update the PR if you'd like 😁.

I'd also like some input from @delucis and @sarah11918 to see what are their thoughts regarding this matter 😊

Ethirallan:
You are welcome and thank you for this amazing feedback @dreyfus92 . Your detailed explanation of all points is greatly appreciated 😁.

Regarding the consideration you have a good point, but I think this is beyond the scope of Astro docs and depends from use case to use case. In my opinion the provided snippet should be designed to function as a standalone pipeline initially (as it already is in the published docs). If it can be consolidated (like in this case) it is just a nice bonus and it would serve as a good starting point for integration into existing and more complex pipelines. By this stage, users are likely equipped with the knowledge to optimize their solutions accordingly anyway.

I'll wait for further feedback and update this PR according afterwards 😄.

sarah11918:
Thanks everyone for this lovely discussion! 

@Ethirallan We really appreciate you taking the time to correct the docs with what worked for you, and helping us to have a working, standalone code sample.

Since @dreyfus92 has the most experience with this workflow, I am happy to let him make the final call. When I see that he has approved this PR, I will consider it ready to merge!

delucis (commented):
Thanks for the PR and discussion! This also looks good to me, although quick note that currently as well as adjusting the YAML indentation, this PR reduces the indentation of the entire codeblock from 3 to 2 spaces. This means Markdown no longer treats it as part of the numbered item above:

| Current Docs | This PR |
|---|---|
| image | image |

It’s a subtle difference (look at how the code block’s left edge aligns with the text above it) but we should make sure to restore the three-space indentation before merging (while also making sure the resulting YAML indentation is correct).

Ethirallan:
Thank you for your feedback, @sarah11918 and @delucis, really nice catch with wrong codeblock indentation. I have made the following changes based on all comments:

- Fixed the codeblock indentation issue
- Moved image inside pages for better reusability
- Tested the changes with gitlab-runner and in one of our projects

If you have any further comments or suggestions, please let me know and I will be happy to make the necessary changes 😄.

delucis (approved):
Fantastic — thanks for the thorough testing @Ethirallan!

PR #5298, opened 2023-11-06
fix: add note about dynamic slot support

### LINKED ISSUE #5297: No mention of lack of support for dynamically named slots
Although the docs mention astro has support for named slots, and that these slots can be used in UI framework components, it doesn't mention anywhere that there are severe limitations when it comes to dynamically named slots.

This is a known limitation of the astro compiler (https://github.com/withastro/astro/issues/5066) and looks like it is unlikely to be supported any time soon.

Not documenting known limitations can be frustrating as it often leads to many wasted hours.

I will submit a PR shortly to amend this.

Alex-S-Davies:
Add a note in the docs about limited support regarding dynamically named slots

Closes #5297

Happy to change the wording or the location of the note if necessary


sarah11918 (approved):
Thanks for this note! 🙌

PR #5302, opened 2023-11-07
feat(overlay): Add information on new exit hook

Princesseuh:
#### Description (required)

https://github.com/withastro/astro/pull/9014 adds a new hook that happens before a plugin is toggled off. It can be used to display a confirmation dialog, or do an animation etc.


sarah11918 (approved):
This is flawlessly written, @Princesseuh.

If `canvas` is something that will be used, it would be nice to show an example usage here. (Admittedly, we're well beyond things I grok here.) But this is non-blocking.

Princesseuh:
> This is flawlessly written, @Princesseuh.
> 
> If `canvas` is something that will be used, it would be nice to show an example usage here. (Admittedly, we're well beyond things I grok here.) But this is non-blocking.

I kinda purposely omitted an example here because it's the exact same idea as for the main hook, that you're required to be using and has an example right before, I'm happy to add one though if you feel like it'd be better

sarah11918:
I would lean towards adding one, because I always think of people who just land on one particular section/heading and even what's immediately above them might not be visible on screen. So if it's not a problem, even if repetitive, I would!

sarah11918:
@Princesseuh It's up to you whether or not to add an example here! I'll merge this tomorrow with all the minor release docs, in whatever state it's in!  🙌

PR #5326, opened 2023-11-09
[i18nIgnore] feat(image-service-reference): Change types to Uint8Array

Princesseuh:
#### Description (required)

Documentation for https://github.com/withastro/astro/pull/9029

99.99999% of people won't be affected by this change, nonetheless we'll document what you can do if you are affected in the changeset.


sarah11918 (approved):
Thank you!

PR #5341, opened 2023-11-10
Refactor Configuration Reference docgen script

delucis:
#### Description (required)

- Updates our configuration reference generation script to add support for `@kind h4` (and other heading levels) in addition to our existing `@kind heading` support.
- Accompanies https://github.com/withastro/astro/pull/9052 and already includes a preview of what those updates would look like with this new version of the script.
- Also refactored this script a bunch because it was kind of gnarly how some of that was written.

### DIFF
--- a/scripts/docgen.mjs
+++ b/scripts/docgen.mjs
@@ -1,3 +1,5 @@
+// @ts-check
+
 import fs from 'fs';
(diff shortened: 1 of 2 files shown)


sarah11918 (approved):
I have been told this is how I get my hands on it.

PR #5392, opened 2023-11-13
initial attempt at way less SSR/SSG

sarah11918:
#### Description (required)
This takes a sweep through docs and updates *most* references to "SSR", "server-side rendering", and "SSG" to more closely match the new rendering modes page that emphasizes "pre-rendering" and "on-demand rendering".

Does not make any changes to the files that are currently generated outside of the docs repo: error messages, integration READMES and config reference

This gives us an idea of what more consistent wording would look like!

We don't have to entirely remove SSR etc. from docs, but we have found in general that it is more helpful in support to refer to our rendering modes (`static`, `server`, and `hybrid`) directly and to distinguish them by what is pre-rendered vs rendered on-demand. This also provides us a nice way of distinguishing several things that happen "on the server" as opposed to "client side." This would also help align our support vocabulary, and help reinforce the `const prerender = true/false` pattern.


src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -38,9 +38,9 @@ There is no separate "routing config" to maintain in an Astro project! When you
 
 An Astro page file can specify dynamic route parameters in its filename to generate multiple, matching pages. For example, `src/pages/authors/[author].astro` generates a bio page for every author on your blog. `author` becomes a _parameter_ that you can access from inside the page.
 
-In Astro's default static output mode, these pages are generated at build time, and so you must predetermine the list of `author`s that get a corresponding file. In SSR mode, a page will be generated on request for any route that matches.
+In Astro's default `static` output mode, these pages are generated at build time, and so you must predetermine the list of `author`s that get a corresponding file. In `server` or `hybrid` mode, a page will be generated on request for any route that matches.
```
natemoo-re, line 41: This section is specifically calling out on-demand rendering so conflating “server” and “hybrid” is a little confusing IMO.

src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -180,9 +180,9 @@ const { resource, id } = Astro.params;
 ```
 This page will be served for any value of `resource` and `id`: `resources/users/1`, `resources/colors/blue`, etc.
 
-#### Modifying the `[...slug]` example for SSR
+#### Modifying the `[...slug]` example for on-demand rendering
 
-Because SSR pages can't use `getStaticPaths()`, they can't receive props. The [previous example](#example-dynamic-pages-at-multiple-levels) can be adapted for SSR mode by looking up the value of the `slug` param in an object. If the route is at the root ("/"), the slug param will be `undefined`. If the value doesn't exist in the object, we redirect to a 404 page.
+Because pages rendered on demand can't use `getStaticPaths()`, they can't receive props. The [previous example](#example-dynamic-pages-at-multiple-levels) can be adapted for `server` or `hybrid` mode by looking up the value of the `slug` param in an object. If the route is at the root ("/"), the slug param will be `undefined`. If the value doesn't exist in the object, we redirect to a 404 page.
```
natemoo-re, line 185: Same comment! Not quite sure how to talk about “server” or “hybrid + prerender = false” being equivalent

natemoo-re (commented):
I’m a huge fan of this change! SSR + SSG are generally confusing terms and I think we have an opportunity to use language more tailored to audience. Since this has already been validated by ongoing support conversations, I’m pretty comfortable with this change!

Two additional things to consider:
- Search! Do users search specifically for “ssg” and “ssr”? Would removing these references make it harder to find the relevant pages? We could probably set up an alias/keyword if needed.
- Somewhat tangential, but I see a few references to “on the client”. How do we feel about that terminology? Would a preference for the phrase “in the browser” also provide some additional clarity?

It would be great to align Astro’s APIs with the terminology as well, when we have to opportunity to do so.

src/content/docs/en/core-concepts/routing.mdx
sarah11918: Yes, agreed! It's not a perfect science, so we'll flag all the awkward cases. I do like "pages (or routes??) rendered on demand" because that covers our bases.

I've been alternating with talking about the "mode enabled", but yes, we should make extra sure that in a case like this, we're referencing what is actually happening for this exact page. I'm sure there will be several of these we'll refine as we figure out when we mean/need to say:

- you have configured this rendering mode
- the actual page/route in question is rendered on-demand

sarah11918:
As for search, I am actually less concerned with that because our rendering modes page clearly makes the analogy between SSG/SSR to our modes. I think it's actually *better* that someone searching SSR gets taken e.g. straight to our page on rendering modes vs.... like, hits for 70 different pages in docs?

sarah11918:
FWIW: The link checks are related to two links in config-reference (not updated here) and would not be a problem once config reference is updated at the source. I'm kind of impressed this caused as few issues as it did. :smile:

sarah11918:
Closing this PR for now as too disruptive for upcoming 4.0/Starlight migration and causing a bunch of translation needed updates. Will work on this one later!

PR #5396, opened 2023-11-14
i18n(es): Update `add-view-transitions`

dreyfus92:
#### Description (required)

Updated `add-view-transitions`, there wasn't much to update but, this was required so the tracker won't keep saying it requires an update.

- Suggested label: i18n


Waxer59 (approved):
LGTM ✅

Yan-Thomas (approved):
LGTM! Thanks peeps 🙌

PR #5407, opened 2023-11-14
Feat:(guides):  #5406 Update client-side-scripts.mdx

### LINKED ISSUE #5406: 💡 Idea: add alternative way to include Astro.props on the client
https://docs.astro.build/en/guides/client-side-scripts/#pass-frontmatter-variables-to-scripts

This section can also be accomplished using the method on this page 

https://docs.astro.build/en/reference/directives-reference/#definevars 

However there is no reference to it here. It would be helpful to give people an easy reference to this alternative.

endymion1818:
#### Description (required)

- Add alternative to recommended approach to the guide on how to pass `Astro.props` to the client

Closes #5406

### DIFF
--- a/src/content/docs/en/guides/client-side-scripts.mdx
+++ b/src/content/docs/en/guides/client-side-scripts.mdx
@@ -196,7 +196,9 @@ There are two advantages to using a custom element here:
 
 In Astro components, the code in [the frontmatter](/en/core-concepts/astro-components/#the-component-script) between the `---` fences runs on the server and is not available in the browser. To send variables from the server to the client, we need a way to store our variables and then read them when JavaScript runs in the browser.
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @endymion1818 ! Just following up on the conversation from the issue I see you've been having with Chris! Yes, as he said, we have some reservations about super strongly recommending it since it is a feature from an earlier time that now brings some extra baggage along with it to modern Astro. We may even decide to change how it works in future, so at this time we're happy with the recommendation on that page, and just leaving this in the reference section.

That being said, we do appreciate you bringing it to our attention and taking the time to improve the docs! This just happens to be one of those things we think is better left available, but not too strongly positioned, and would prefer to instead force ourselves to make a solution that works better! I hope you'll come back and let us know the next time you see something that could be improved!

PR #5410, opened 2023-11-14
feat(i18n): current locale

ematipico:
#### Description (required)

This PR adds documentation for `Astro.currentLocale`

RFC change: https://github.com/withastro/roadmap/pull/734/commits/f934880ab5835492e56ddbb69e3f32662310c8bf

 #### For Astro version: `3.5.*`. See astro PR: https://github.com/withastro/astro/pull/9101

We are in an experimental phase, so we can add new features in any release.


src/content/docs/en/guides/internationalization.mdx
```diff
@@ -169,9 +169,8 @@ Astro's i18n routing combined with one of Astro's  [on-demand server rendering m
 These combine the browser's `Accept-Langauge` header, and your `locales` to automatically respect your visitor's preferred languages.
 
 - `Astro.preferredLocale`:  Astro can compute a **preferred locale** for your visitor if their browser's preferred locale is included in your `locales` array. This value is undefined if no such match exists.
-
-
 - `Astro.preferredLocaleList`: An array of all locales that are both requested by the browser and supported by your website. This produces a list of all compatible languages between your site and your visitor. The value is `[]` if none of the browser's requested languages are found in your `locales` array. If the browser does not specify any preferred languages, then this value will be [`i18n.locales`].
+- `Astro.currentLocale`: The locale computed from the current URL, that matches what's configured in [`i18n.locales`]. When the `routingStrategy` is set to `"prefix-other-locales"`, the value will be `i18n.defaultLocale` for any URL that doesn't contain any locale.
```
sarah11918, line 173:
```suggestion
- `Astro.currentLocale`: The locale computed from the current URL. If the URL does not contain a `/[locale]/` prefix, then the value will default to `i18n.defaultLocale`.
```
1. Is it important to say "matching what is configured in `i18.locales`? That is an array, and this is a single locale, right? So I'm not sure how they can be matching. And if it's in your current URL then doesn't it already *have* to be in the locales array? If that is the case, then I think it's not necessary, and more confusing, to mention `locales`.  Let me know if I have this wrong!

3. While the extra context about the `routingStrategy` is correct, I think it's more the cause of what would lead there to be no URL prefix shown, and we don't have to spell that out. I think, if this is true, then it's fine to just say that "if there's no language prefix in the URL, it defaults to your default language". I think again the extra information adds more complication.

sarah11918 (commented):
Hi @ematipico ! I took a shot at simplifying the explanation, because I think we can remove some complication here (if I understand this correctly!)

Please let me know if the following is correct, and if so, I think it's OK just to say that!

src/content/docs/en/guides/internationalization.mdx
ematipico: 1. Yeah, sorry, that was poorly worded. The locale returned will **one** of the values contained in `i18n.locales`. The reason why I mention it is because a user could have configured `pt-BR`, but have `/pt-br/blog/` in the URL. We want to return the value **configured**.

2. I believe it's fine to use your suggestion **for now**. This will change once we will add more features

sarah11918: Ah, that's a great point. We will return the CONFIGURED value! Got it. I can add that!

sarah11918: @ematipico Updated to briefly include that (as you say, for now!) See what you think!

sarah11918:
Note that this is set for 3.5.6... next release!

sarah11918 (approved):
Approving this for the next patch release.

PR #5425, opened 2023-11-15
VT event lifecycle docs

matthewp:
#### Description (required)

Documents the new lifecycle events from: https://github.com/withastro/astro/pull/9090


matthewp:
I could see the following changes made:

- Make the sections which explain the events purely about *what the event does*, when it occurs, etc. and not about use-cases.
- Move use-cases to a separate headings. I like this idea because it's easier to find how to achieve something vs. finding which event lists the use-case.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -488,20 +547,24 @@ An event that fires immediately after the new page replaces the old page. You ca
 
 This event, when listened to on the **outgoing page**, is useful to pass along and restore any state on the DOM that needs to transfer over to the new page.
 
-For example, if you are implementing dark mode support, this event can be used to restore state across page loads:
+Previously this event was used to implement dark mode. You could listen to the event and then add the dark mode class name to, for example: `<html class="dark-mode">`. This is still a valid way to do it, but it might be better to do so in an earlier event, like `astro:after-preparation` which occurs before the swap and before the view transition occurs.
 
-```astro
+As the `astro:after-swap` event occurs immediately before the view transition ends, it's a good time to stop any loading indicators that might be running.
```
martrapp, line 563: "occurs immediately before the view transition ends" could be misleading: Surely it is near the end of the navigation, but "astro:after-swap" takes place immediately before the start of the animation (for the simulation immediately before the start of the "new" part). In any case, the loading indicator is already recorded on the screenshot and will not disappear so quickly, depending on the animation. A better place to remove it would be `astro:after-preparation` because the photo has not yet been taken :)

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -465,20 +465,79 @@ If you have code that sets up global state, this state will need to take into ac
...
+    ev.loader = async function() {
+      const { startSpinner } = await import('./spinner.js');
+      const stop = startSpinner();
+      await originalLoader();
+      stop();
+    };
+  });
+</script>
+```
+
```
martrapp, line 510: I like this example where the spinner is set and removed with the loader callback.    But if there is another wrapper around it that needs some time to do e.g. further prefetching based on the loaded data, then the spinner is gone too soon. It would be much better to remove the spinner in astro:after-preparation, because all loaders/prefetchers/... are through and the screenshot has not yet been taken.

sarah11918:
@matthewp @martrapp I took a pass at this, and added a bit more intro content to the section. It was really well done! I would be happy with this releasing with the new features, and as we have time adding more of Martin's use-cases listed in PR 9090... either in time for launch or otherwise.

I will call out:

- `astro:after-swap` was already written for docs, and was previously **the only** point at which state was passed, so I tried to update more like to indicate that this is the "last responsible moment" for passing along state, and instead focus on the end of the view transition. I would ask you in particular to look at this and see whether you'd describe this differently now.
- I'm not entirely sure *I've* nailed the difference between `after-preparation` and `before-swap`. Only one of them talked about being inside/outside of the view transition, so I made it clearer that one is before and one is inside. And I'm inferring that the difference is that before you have access to the old page only, not the new one, and inside the view transition you have access to BOTH DOMs so you can swap. I just want to make sure that that is correct and that you think that comes across. We haven't really said what "being inside the view transition" means explicitly, but I think that's the connection.

Otherwise, please read for both accuracy but also nuance! We have lots of time to get this *exactly* right, not just *sort of right*.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -465,20 +465,90 @@ If you have code that sets up global state, this state will need to take into ac
...
 Module scripts are only ever executed once because the browser keeps track of which modules are already loaded. For these scripts, you do not need to worry about re-execution.
 
-### `astro:page-load`
+## Lifecycle events
 
-An event that fires at the end of page navigation, after the new page is visible to the user and blocking styles and scripts are loaded. You can listen to this event on the `document`.
+The `<ViewTransition />` router fires a number events on the `document` during navigation. These events provide hooks into the lifecycle of navigation, allowing to you do things like showing indicators that a new page is loading, override default behavior, and restore state as navigation is completing.
 
-The `<ViewTransitions />` component fires this event both on initial page navigation (MPA) and any subsequent navigation, either forwards or backwards.
+The navigation phase involves a **preparation** phase, when new content is loaded; a **DOM swap** phase, where the old page's content is replaced by the new page's content; and a **completion** phase where the new page is loaded, scripts are executed and clean-up work is carried out.
```
sarah11918, line 472:
```suggestion
The navigation proccess involves a **preparation** phase, when new content is loaded; a **DOM swap** phase, where the old page's content is replaced by the new page's content; and a **completion** phase where the new page is loaded, scripts are executed and clean-up work is carried out. 
```

src/content/docs/en/guides/view-transitions.mdx
sarah11918: Is this easy to show in a code sample? Do we just show **starting** the spinner here? I will ask you or Matthew to edit the sample to your liking!

src/content/docs/en/guides/view-transitions.mdx
sarah11918: I'm hearing two different comments here:

1. "immediately before the view transition ends" isn't *exactly* right. What's a better, concise, way to describe it?
2. Spinner has come back to haunt us. This now is the only real example (that's not, it's still safe to do this) example for this section. Can you suggest (code sample even better, but just in words like above is fine) what a good thing to happen here is? So that this section can have a proper example of its own that makes sense and is a realistic use case? Note: it's also OK to say that "there's very little that might target this event specifically, but it does provide you..."

matthewp:
@sarah11918 Yeah, that different makes sense. Another thing that @martrapp and I were talking about, is that the `before-` events are points in which you can modify what is about to happen, whereas the `after-`  events are notifications of what just happened. I'm not sure if that's something we should mention or not.

Doing another pass myself now.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -465,20 +465,105 @@ If you have code that sets up global state, this state will need to take into ac
...
+- [`astro:after-swap`](#astroafter-swap)
+- [`astro:page-load`](#astropage-load)
 
-```astro "{ once: true }"
-<script>
-  document.addEventListener('astro:page-load', () => {
-    // This only runs once.
-    setupStuff();
-  }, { once: true });
+While some actions can be trigged during any event, some tasks can only be performed during a specific event for best results, such as displaying a loading spinner before preparation or overriding animation pairs before swapping content.
```
sarah11918, line 481:
```suggestion
:::tip
`before-` events allow you to influence and modify actions that are about to take place, and `after-` events are notifications that a phase is complete.
:::

While some actions can be triggered during any event, some tasks can only be performed during a specific event for best results, such as displaying a loading spinner before preparation or overriding animation pairs before swapping content.
```
Wondering what we think about something like this? I'll commit so we can see it on the page.

src/content/docs/en/guides/view-transitions.mdx
sarah11918: @martrapp Given @matthewp 's new example below using both `before-` and `after-` preparation, should we remove this example altogether, and let the combined example below serve for both?

src/content/docs/en/guides/view-transitions.mdx
matthewp: Yes, I like this!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -488,20 +577,24 @@ An event that fires immediately after the new page replaces the old page. You ca
 
 This event, when listened to on the **outgoing page**, is useful to pass along and restore any state on the DOM that needs to transfer over to the new page.
 
-For example, if you are implementing dark mode support, this event can be used to restore state across page loads:
+This is the latest point in the lifecyle where it is still safe to, for example, add a dark mode class name (`<html class="dark-mode">`), though you may wish to do so in an earlier event.
 
-```astro
+As the `astro:after-swap` event occurs immediately before the view transition ends, it's a good time to stop any loading indicators that might be running.
```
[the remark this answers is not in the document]
martrapp: Meaningful example pending

martrapp:
As for the question of relevant examples for each event, you might want to check out the new examples at [https://events-3bg.pages.dev/](https://events-3bg.pages.dev/).

For the document we could do this:

* before-preparation + loader() replacement: (simplified) image prefetch example from the demo page
* after-preparation: existing example (loading indicator)
* before-swap: existing example (darkMode) 
* before-swap + swap() replacement: existing example (diff)
* after-swap: scroll to top example from the demo page
* page-load: existing example

sarah11918:
@martrapp Did you make that page?? That's lovely! I would suggest we link to it, rather than reproduce it. Many of our docs pages link to "community resources" and then it's less for our translators to translate, less of a maintenance burden on docs if things update. it also encourages community members to produce content which scales more than us doing everything ourselves.  :smile: 

What would you think about that?

martrapp:
> What would you think about that?

That would be great! (Once the events are published and people can download and run the examples with a current Astro version)

But we can definitely borrow for the missing after-swap example :-)

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -465,20 +465,109 @@ If you have code that sets up global state, this state will need to take into ac
 
 Module scripts are only ever executed once because the browser keeps track of which modules are already loaded. For these scripts, you do not need to worry about re-execution.
 
-### `astro:page-load`
+## Lifecycle events
 
-An event that fires at the end of page navigation, after the new page is visible to the user and blocking styles and scripts are loaded. You can listen to this event on the `document`.
+The `<ViewTransition />` router fires a number events on the `document` during navigation. These events provide hooks into the lifecycle of navigation, allowing you to do things like showing indicators that a new page is loading, override default behavior, and restore state as navigation is completing.
```
sarah11918, line 470:
```suggestion
The `<ViewTransition />` router fires a number events on the `document` during navigation. These events provide hooks into the lifecycle of navigation, allowing you to do things like show indicators that a new page is loading, override default behavior, and restore state as navigation is completing.
```
nit: consistent verb tense

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -465,20 +465,109 @@ If you have code that sets up global state, this state will need to take into ac
...
+    document.querySelector('#loading').classList.remove('show');
+  });
+</script>
+```
+Compared to the spinner example above, this example looks simpler: If all of the listener's code can be executed synchronously, there is no need to hook into the loader's callback.  
+### `astro:before-swap`
+
+<Since v="3.6.0" />
+
+An event that fires befores the new document, populated during the preparation phase, replaces the current document. This event occurs inside of the view transition, where the user is still seeing a snapshot of the old page.
```
sarah11918, line 538:
```suggestion
An event that fires before the new document (which is populated during the preparation phase) replaces the current document. This event occurs inside of the view transition, where the user is still seeing a snapshot of the old page.
```
I feel like the parenthesis can help with the more complicated sentence structure here.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -465,20 +465,109 @@ If you have code that sets up global state, this state will need to take into ac
...
+    setDarkMode(ev.newDocument);
+  });
+</script>
+```
+
+
+
+The `astro:before-swap` event can also be used to change the *implementation* of the swap. The default swap implementation diffs head content, moves __persistent__ elements from the old document to the `newDocument`, and then replaces the entire `body` with the body of the new document.
+
+At this point of the lifecycle, you could choose to define your own swap implementation, for example to diff the entire contents of the existing document (which some other routers do):
```
sarah11918, line 562:
```suggestion
The `astro:before-swap` event can also be used to change the *implementation* of the swap. The default swap implementation diffs head content, moves __persistent__ elements from the old document to the `newDocument`, and then replaces the entire `body` with the body of the new document.

At this point of the lifecycle, you could choose to define your own swap implementation, for example to diff the entire contents of the existing document (which some other routers do):
```
just removed some blank lines!

sarah11918:
@martrapp I would love it if you could add the `after-swap` event example here, and then I think this would be good to go!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -488,20 +576,24 @@ An event that fires immediately after the new page replaces the old page. You ca
 
 This event, when listened to on the **outgoing page**, is useful to pass along and restore any state on the DOM that needs to transfer over to the new page.
 
-For example, if you are implementing dark mode support, this event can be used to restore state across page loads:
+This is the latest point in the lifecyle where it is still safe to, for example, add a dark mode class name (`<html class="dark-mode">`), though you may wish to do so in an earlier event.
 
-```astro
+The "astro:after-swap" event occurs immediately after the browser history has been updated and the scroll position has been set. 
+
```
martrapp, line 582:
```suggestion
Therefore, it is save now to override the standard scroll restoration if desired. The following example resets the horizontal and vertical scroll position to the top left corner of the page. 
```

martrapp:
```astro
document.addEventListener('astro:after-swap', 
  () => window.scrollTo({ left: 0, top: 0, behavior: 'instant' }))
```

sarah11918: Perfect, thanks, @martrapp  I'll add these in!

Just because GitHub is unintuitive and you might not realize you can add the code directly in your suggestion, here's my trick:

When you click to make a "suggestion", it will automatically format with three backticks. If you **increase the number of backticks to open and close your suggestions** (I add three, to make it six, out of habit), then you can have a regular three-backtick code sample directly in your suggestion! Like this!

image

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -488,20 +576,30 @@ An event that fires immediately after the new page replaces the old page. You ca
 
 This event, when listened to on the **outgoing page**, is useful to pass along and restore any state on the DOM that needs to transfer over to the new page.
 
-For example, if you are implementing dark mode support, this event can be used to restore state across page loads:
+This is the latest point in the lifecyle where it is still safe to, for example, add a dark mode class name (`<html class="dark-mode">`), though you may wish to do so in an earlier event.
 
-```astro
+The "astro:after-swap" event occurs immediately after the browser history has been updated and the scroll position has been set. 
+Therefore, it is now possible to override the default scroll restore for history navigation if desired. The following example resets the horizontal and vertical scroll position to the top left corner of the page for each navigation.
```
sarah11918, line 582:
```suggestion
Therefore, one use of targeting this event is to override the default scroll restore for history navigation. The following example resets the horizontal and vertical scroll position to the top left corner of the page for each navigation. 
```

sarah11918:
OK! I'm happy with the content/structure of this page!

@matthewp @martrapp - happy for you to do any final review here. Otherwise, Team Docs will just handle any last minute polishing as we notice in the morning! Thank you both for your contribution for this one! An exciting feature we're happy to get into docs! 🙌

martrapp (approved):
Layout error due to wrong quotation marks, but otherwise I'm very happy with this section!

PR #5434, opened 2023-11-17
i18n(ko-KR): update `docsearch.ts`

jsparkdev:
#### Description (required)

update `docsearch.ts`


sarah11918 (approved):
Did some checks so that we could merge this, even without a native language review. Looks great, thank you! 🥳

PR #5450, opened 2023-11-17
Adds Sentry to the list of sponsors

tony-sull:
#### Description (required)

This adds Sentry to the list of sponsors in the main navigation sidebar

- Related astro.build [PR 879](https://github.com/withastro/astro.build/pull/879)


sarah11918:
@tony-sull Is there a specific date for merging this? Does it wait for the launch day announcement?

delucis:
+1 to Sarah’s question. And if we are waiting, this PR may not be so useful as we’ll be migrating to Starlight by then.

clearlyTHUYDOAN (approved):

tony-sull:
@sarah11918 @delucis great question, I hadn't realized the Starlight migration was so close!

We have the green light to ship this update before the announcement, I'll be doing the astro.build merge shortly

sarah11918:
@tony-sull @delucis It was just confirmed on Launch week call that yes, these logos can merge any time. So, we *can* merge here, and then we can carry it ourselves over to the migration, would you say, Chris?

delucis:
Yup, that works!

src/components/LeftSidebar/Sponsors.astro
```diff
@@ -26,6 +26,13 @@ import UIString from '../UIString.astro';
 			</g>
 		</svg>
 	</a>
+	<a href="https://sentry.io/" aria-label="Sentry">
+		<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 44" width="110" height="24">
```
delucis, line 30: LGTM! Just needs an `aria-hidden` on the SVG itself:

```suggestion
		<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 44" width="110" height="24" aria-hidden="true">
```

tony-sull: Great catch, I even saw that when copying it in and forgot about it when sitting back down to setup the PR 🤦

PR #5457, opened 2023-11-18
i18n(ru): Add new core-concepts/astro-components.mdx translation

tem235:
#### Description (required)
- Translated `core-concepts/astro-components.mdx` page into Russian


sarah11918 (approved):
Thank you, @tem235 ! There is great attention to detail here including all the links from other pages! Normally, we would prefer to wait for a native language speaker to review. But, since we are close to migrating to Starlight and want to wrap up our existing PRs, I did some checks and this looks good to me! If we need any corrections for typos etc., then future PRs can be made. But right now, this can be merged and helping people right away!  🙌

PR #5478, opened 2023-11-23
Simplify robots.txt

bluwy:
#### Description (required)

`robots.txt` can be a simple file in `/public` rather than an endpoint. This also unblocks https://github.com/withastro/astro/pull/9181 which we'll be removing support for returning simple objects from endpoints. 

- Closes n/a
- Suggested label: n/a

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Allow: /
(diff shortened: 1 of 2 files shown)


sarah11918 (approved):
Go ahead, @bluwy !

PR #5500, opened 2023-11-26
DOCS: Add to VS Code settings, Bun recipe and Astro package for Subli…

Afrowave:
#### Description (required)

The purpose of this PR is to add missing instructions for getting the Sublime Text Astro package and adding Cypress and Playwright testing instructions into the Bun recipe page.

This PR:

- Add integrations with VS Code and ESlint
- Add testing instructions for Cypress and Playwright for the Bun runtime


sarah11918:
Hey @Afrowave ! Thanks for these, but just a heads up that it's much easier on us if we could keep the different concerns separate in multiple PRs here. 

We have different people reviewing different areas of docs, so this allows us to tag in the appropriate people based on the content. And, it keeps us, and the translators who update pages in response to PRs, better organized when PRs aren't just "random collection of things someone decided to contribute", if you know what I mean.  😄 

The two ones related to editor setup are fine together, as I'd be asking Erika to take a look at both of those. But would you mind separating out the Bun one into its own PR?  That way it can be reviewed and addressed separately by other maintainers who would normally review that content. It also lets our translators separate the work. And, you get TWO PRs for the price of one on your Astro Badge... 🚀 

Thank you!

Afrowave:
Thanks @sarah11918,

Let me do just that.

PR #5501, opened 2023-11-26
ADD settings for Code Editors

Afrowave:
#### Description (required)

Add to VS Code settings, Bun recipe and Astro package for Sublime Text

- Add integrations with VS Code and ESlint

- Closes #5500


Afrowave:
Hi @sarah11918,  
  
I now know a contributor cannot have two pending PRs. I tried to separate them but I see I need to create two different branches to make that work. Learning is coming quick and fast.

PR #5502, opened 2023-11-26
Add to VS Code settings and the Astro package for Sublime Text

Afrowave:
#### Description (required)
This PR is to add details to enhancing code editors for Astro
- Add ESlint integrations with VS Code


src/content/docs/en/editor-setup.mdx
```diff
@@ -37,6 +37,7 @@ Our amazing community maintains several extensions for other popular editors, in
 - [Vim Plugin](https://github.com/wuelnerdotexe/vim-astro) <span style="margin: 0.25em;"><Badge variant="neutral">Community</Badge></span> - Provides syntax highlighting, indentation, and code folding support for Astro inside of Vim or Neovim
 - Neovim [LSP](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#astro) and [TreeSitter](https://github.com/virchau13/tree-sitter-astro) Plugins <span style="margin: 0.25em;"><Badge variant="neutral">Community</Badge></span> - Provides syntax highlighting, treesitter parsing, and code completion for Astro inside of Neovim
 - Emacs - See instructions for [Configuring Emacs and Eglot](https://medium.com/@jrmjrm/configuring-emacs-and-eglot-to-work-with-astro-language-server-9408eb709ab0) <span style="margin: 0.25em;"><Badge variant="neutral">Community</Badge></span> to work with Astro
+- [Astro syntax highlighting for Sublime Text](https://packagecontrol.io/packages/Astro) <span style="margin: 0.25em;"><Badge variant="accent">Official</Badge></span> - The official package for Sublime Text, available on the Sublime Text package manager.
```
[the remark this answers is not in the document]
Afrowave: Thank you @sarah11918. I had no idea "official" meant Astro-official. 😃

.vscode/settings.json
```diff
@@ -1,3 +1,18 @@
 {
```
Princesseuh, line 1: All the settings here looks great to me, however they're opinionated and since I don't work on the docs, I don't really know. If they're what people want, maybe @delucis would have a better idea!

sarah11918: Thanks! Will ask Chris to weigh in here!

sarah11918:
Also, @Afrowave , reminder that we might have the same issue as the other PR in that any suggested changes should be "accepted" and ideally you can mark this PR as editable by maintainers so we can swoop in and make any changes we need without bothering you again!  :raised_hands:

.vscode/settings.json
sarah11918: Hi @Afrowave - Discussed this with Chris and we think these settings are too opinionated for putting as official documentation. We are happy to add the Sublime link, though, so if you can amend this PR so that the only change is the link that would be wonderful!

Reminder, any future PRs should be marked as "maintainers may edit" so that we can just fix little things like this ourselves and not bother you.  🙌

PR #5518, opened 2023-11-28
docs(overlay): Add mention of new event

Princesseuh:
#### Description (required)

Docs for https://github.com/withastro/astro/pull/9215


Princesseuh:
This doesn't make as much sense with https://github.com/withastro/astro/pull/9230 now, because the overlay gets minimized a lot more easily and is often not a conscious choice.

PR #5522, opened 2023-11-29
Improve description of `locals` on guides/middleware

### LINKED ISSUE #4207: Improve `locals` documentation on Middleware
### 📚 Subject area/topic

Middleware

https://docs.astro.build/en/guides/middleware/

Some code snippets, like the one below, include a comment about intercepting the response data from a request. No data is being intercepted in such snippets and the comments are placed before populating the `locals` object, where one might intercept _the request_. Intercepting _the response_ would be done by modifying the return value of `next()`.

```js
export function onRequest ({ locals, request }, next) {
    // intercept response data from a request
    // optionally, transform the response by modifying `locals`
    locals.title = "New title";

    // return a Response or the result of calling `next()`
    return next();
};
```

I think this snippet would be clearer with the following comments:

```js
export function onRequest ({ locals, request }, next) {
    // intercept data from a request
    // optionally, share data with the rendering process by modifying `locals`
    locals.title = "New title";

    // return a Response or the result of calling `next()`
    // optionally, transform the response by modifying the result
    return next();
};
```

The reference for `locals` also causes the same confusion, it says:

> An object containing data from a `Response` that can be manipulated inside middleware.

From what I understand, the `locals` is used by the routes and pages to create the response, so it contains data _for_ the response, possibly _from the request_.

Is my interpretation correct?

Fryuni:
#### Description (required)

Change the comments and description of the `locals` object on `guides/middleware`.

- Closes #4207 

### DIFF
--- a/src/content/docs/en/guides/middleware.mdx
+++ b/src/content/docs/en/guides/middleware.mdx
@@ -19,11 +19,12 @@ Middleware is available in both SSG and SSR Astro projects.
 
     ```js title="src/middleware.js"
     export function onRequest ({ locals, request }, next) {
-        // intercept response data from a request
-        // optionally, transform the response by modifying `locals`
+        // intercept data from a request
+        // optionally, share data with the rendering process by modifying `locals`
         locals.title = "New title";
 
         // return a Response or the result of calling `next()`
+        // optionally, transform the response by modifying the result
         return next();
     };
     ```
@@ -46,7 +47,7 @@ This is an optional argument passed to `onRequest()` that may contain the `local
 
 ### Storing data in `context.locals`
 
-`context.locals` is an object containing data from a `Response` that can be manipulated inside middleware. 
+`context.locals` is an object that can be manipulated inside the middleware. Adapters _might_ add data to it.
 
 This `locals` object is forwarded across the request handling process and is available as a property to [`APIContext`](/en/reference/api-reference/#contextlocals) and [`AstroGlobal`](/en/reference/api-reference/#astrolocals) . This allows data to be shared between middlewares, API routes, and `.astro` pages. This is useful for storing request-specific data, such as user data, across the rendering step.
 
@@ -54,14 +55,15 @@ You can store any type of data inside `locals`: strings, numbers, and even compl
 
 ```js title="src/middleware.js"
 export function onRequest ({ locals, request }, next) {
-    // intercept response data from a request
-    // optionally, transform the response by modifying `locals`
+    // intercept data from a request
+    // optionally, share data with the rendering process by modifying `locals`
     locals.user.name = "John Wick";
     locals.welcomeTitle = () => {
         return "Welcome back " + locals.user.name;
     };
 
     // return a Response or the result of calling `next()`
+    // optionally, transform the response by modifying the result
     return next();
 };
 ```


src/content/docs/en/guides/middleware.mdx
```diff
@@ -46,22 +47,23 @@ This is an optional argument passed to `onRequest()` that may contain the `local
 
 ### Storing data in `context.locals`
 
-`context.locals` is an object containing data from a `Response` that can be manipulated inside middleware. 
+`context.locals` is an object that can be manipulated inside the middleware. Adapters _might_ add data to it.
```
sarah11918, line 50: "Adapters might add data to it" is a little too vague to be helpful.

It's not clear whether the reader should take this as some kind of warning (i.e. if you're writing your own middleware , beware that there might be something else that's also manipulating the data in `context.locals`), or it's being used as an example of something that manipulates `context.locals` to provide real-usage context, or something else entirely.... fun fact!  Did you know...! :smile: 

Can you reword that statement in a way that makes it clear *why* they are being told that?

src/content/docs/en/guides/middleware.mdx
```diff
@@ -19,11 +19,12 @@ Middleware is available in both SSG and SSR Astro projects.
 
     ```js title="src/middleware.js"
     export function onRequest ({ locals, request }, next) {
-        // intercept response data from a request
-        // optionally, transform the response by modifying `locals`
+        // intercept data from a request
+        // optionally, share data with the rendering process by modifying `locals`
```
sarah11918, line 23:
```suggestion
        // optionally, modify the data stored in `locals`
```
While yours may be the more technically accurate description, I don't think this exactly hits the mark on the helpfulness scale.

"Transform some data" is really clear to a reader that you're going to change the value of something. "Share data with a process" really doesn't convey with the same clarity the idea that middleware is used to change an object that is then available somewhere else.

Can we find a middle ground here? Is it "good enough" just to do the above?

src/content/docs/en/guides/middleware.mdx
```diff
@@ -19,11 +19,12 @@ Middleware is available in both SSG and SSR Astro projects.
...
         locals.title = "New title";
 
         // return a Response or the result of calling `next()`
+        // optionally, transform the response by modifying the result
```
sarah11918, line 27: This doesn't feel well-placed to me, and in fact I don't even think it's necessary at all, if the we get the above code comment just right. The existing comment here is meant to highlight that you have to RETURN something, either `next()`  or a `Response` directly. I think this statement here only adds extra information gets (literally!) in the way of the point it's trying to make.

Is this also in fact correct? is the modifying done after what is returned in this function? Isn't returning *something* the final step here? So, if there are in fact two code comments, it seems to me like the one about what is returned by the middleware should go last.

I can think of some options off the top of my head, depending on what's most accurate, that I think feel more *helpful* to the reader:

1. change the existing statement to include the idea that the data, potentially modified, is passed along in the thing that's returned. e.g. "return a Response or the result of calling `next`, passing along `context.locals`"
2. swap the order of these, and reword the first one include information you believe is truly useful to include.
3. delete this, because it probably doesn't need to be spelled out HERE that the new info is available if the code comment above clearly explains the purpose behind transforming some data to mutate and pass along `context.locals`.

sarah11918 (commented):
Thank you for jumping in and helping us get this example just right, @Fryuni ! I'm not sure we're quite there yet, so I've left my thoughts so we can keep workshopping it. See what you think!

src/content/docs/en/guides/middleware.mdx
Fryuni: The distinction is that:
- You can modify the `locals` to affect how the page/endpoint _will_ behave
- You can modify the return of `next` _after_ the page/endpoint has already completed

There are examples of this in the same page. For the first case, there is an example of changing the title inside `locals` allowing the page to render it differently. For the second, there is an example of redacting sensitive information by replacing segments of the resulting HTML.

src/content/docs/en/guides/middleware.mdx
Fryuni: I agree that my wording was _too_ technical, but it is in line with the language just above it:

> This `locals` object is forwarded across the request handling process and is available as a property to `APIContext` and `AstroGlobal`.

I think a middle ground would be best. The original "transform the response" is what I thought was misleading since it may not affect the response at all, and there is another place to transform the actual `Response` object (after the `next` call), so it could be confusing.

I like your suggestion, we can go with it. I'd prefer "modify the _properties_" instead of "modify the _data_" just because you can set functions in there, but this is quite minor, WDYT?

src/content/docs/en/guides/middleware.mdx
Fryuni: Taking another look at it, I agree that it gets in the way of the most essential information. Since there is already another section explaining about modifying the result of `next` I'll remove this comment.

src/content/docs/en/guides/middleware.mdx
Fryuni: I think this could be added on a tip. What do you think of removing that last sentence and adding this:

```md
:::tip[Adapter-provided properties]
Adapters may provide extra runtime information and functionality through the `locals` object. For example, SSR adapters may include information about the request or runtime.
:::
```

We can link to some examples. Of the official integrations, 4 modify the `locals` object.

sarah11918: I still don't know **why** you're telling me this, though? Again, it's not clear why I should care about this, whether it's something for me to watch out for...

src/content/docs/en/guides/middleware.mdx
sarah11918: Totally cool with "modify the properties" !

src/content/docs/en/guides/middleware.mdx
sarah11918: That works for me!

src/content/docs/en/guides/middleware.mdx
```diff
@@ -46,22 +47,23 @@ This is an optional argument passed to `onRequest()` that may contain the `local
...
+    // intercept data from a request
+    // optionally, share data with the rendering process by modifying `locals`
     locals.user.name = "John Wick";
     locals.welcomeTitle = () => {
         return "Welcome back " + locals.user.name;
     };
 
     // return a Response or the result of calling `next()`
+    // optionally, transform the response by modifying the result
     return next();
```
sarah11918, line 67:
```suggestion
    // optionally, modify the properties stored in `locals`
    locals.user.name = "John Wick";
    locals.welcomeTitle = () => {
        return "Welcome back " + locals.user.name;
    };

    // return a Response or the result of calling `next()`
    return next();
```
Just a reminder that whatever we agree on above should also be done here, e.g.

src/content/docs/en/guides/middleware.mdx
Fryuni: I think it is a mix of everything. It is something to be aware:
- An integration may provide something that you are checking manually
- Something you write might override a value used by an integration
- Information that would be hard to access or infer otherwise might already be included there (like CloudFlare's runtime)

sarah11918: OK, gotcha! No, I don't want examples here to clutter this up. The reasoning/why I should care is way more important. I think what you mentioned above is the kind of content that would be good to go inside a note!

sarah11918:
So @Fryuni , happy if you add a new note based on the "why I need to know/care" ideas and then adjust the two code samples accordingly!

Let me know when you'd like me to take another look!

src/content/docs/en/guides/middleware.mdx
Fryuni: I sent a new version with a different tip. WDYT?

sarah11918 (approved):
Looks great, @Fryuni ! Thanks for your patience and cooperation during a particularly stressful docs time!  :sweat_smile: 

Happy to have this improvement in the docs! :partying_face:

PR #5523, opened 2023-11-29
Prevent scrolling when feedback modal is open

mingjunlu:
#### Description

- This PR prevents scrolling when the feedback modal is open. It should lead to:
  - Better user experience
  - Consistent behavior with the search modal, which also prevents scrolling
- Demonstration:

  | Before | After |
  | ------ | ----- |
  | <video src="https://github.com/withastro/docs/assets/40516784/dcb95588-bd2a-4ecc-9ee6-b639438ed89a" /> | <video src="https://github.com/withastro/docs/assets/40516784/3cde158b-3977-4219-92c7-ebc1e14335a1" />  |

#### Related issues & labels

- Closes #5334

### DIFF
--- a/src/components/tutorial/FeedbackButton.astro
+++ b/src/components/tutorial/FeedbackButton.astro
@@ -150,6 +150,7 @@ const instanceId = Math.round(Math.random() * 10e9).toString(36);
 			const openModal = (event: MouseEvent) => {
 				showSuccess(false);
 				dialog.showModal();
+				document.body.classList.add('overflow-hidden');
 				form.querySelector<HTMLInputElement>('input:checked')?.focus();
 				// Prevent click bubbling up and immediately closing the modal again.
 				event.stopPropagation();
@@ -158,6 +159,7 @@ const instanceId = Math.round(Math.random() * 10e9).toString(36);
 
 			const closeModal = () => {
 				dialog.close();
+				document.body.classList.remove('overflow-hidden');
 				window.removeEventListener('click', onWindowClick);
 			};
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @mingjunlu ! I can see you created an issue for this, but didn't wait for any feedback to see whether this was actually a fix we wanted to implement. In future, please wait for some feedback there first before submitting a PR. I actually kind of like the scroll underneath, and I don't consider this a problem to fix, myself! Given that, it makes sense that this should have been discussed in the issue first, because we don't know whether this is something that a lot of people want.

But perhaps more importantly, we are just about to change this site over to Starlight in the next few days, so we're not really accepting any changes to the site infrastructure here right now. This version will be going away very soon! So, I'm closing this PR, but will leave your issue open for discussion about the idea.

If you're interested in implementing this in the new version, you can look at the open PR to Migrate to Starlight to see what the new site will be. But again, I prefer the existing scrolling behaviour myself and if you asked me, I would not want to change it. So, it probably makes more sense to wait and see whether people contribute to your issue and also think it's a problem.

mingjunlu:
Hi @sarah11918,

Thank you for your swift response and detailed explanation. I value your perspective on the current scrolling behavior, and I want to apologize for not waiting for feedback before submitting the PR.

Not receiving any responses from the initiation of the discussion, I mistakenly believed that the discussion might have been overlooked. In an attempt to move the conversation forward and contribute positively, I took the initiative to submit the PR and propose an adjustment.

I now understand that this approach was premature, and I acknowledge the upcoming transition to Starlight, which makes changes to the current version's infrastructure impractical. Thank you for closing the PR and providing clarification on the roadmap.

I'm sorry if this caused any inconvenience. I appreciate your guidance, and moving forward, I will ensure to be more patient and considerate in the future.

Thank you once again for your guidance and understanding.

PR #5527, opened 2023-11-29
refactor(i18n): remove `routingStrategy` in favour of `routing` object

ematipico:
#### Description (required)

RFC change: https://github.com/withastro/roadmap/pull/734/commits/089253d317e4327ee4e3b93eb303ddc28859952b

This PR updates the page to use the new `routing` object. The old `routingStrategy` has been removed.

Here's how the old configuration is mapped to the new one

```diff
export default defineConfig({
  experimental: {
      i18n: {
-          routingStrategy: "prefix-always",
+          routing: {
+              prefixDefaultLocale: true,
+          }  
      }
  }
})
```

```diff
export default defineConfig({
  experimental: {
      i18n: {
-          routingStrategy: "prefix-other-locales",
+          routing: {
+              prefixDefaultLocale: false,
+          }  
      }
  }
})
```

Also, it's worth noting that `routing` contains a key called [`strategy`](https://github.com/withastro/astro/blob/8deeb313ca98f74a90ffd5a1f6d109298440b540/packages/astro/src/core/config/schema.ts#L354), but since its value is only one, ** for now**, I decided not to mention it at all.

### DIFF
--- a/src/content/docs/en/guides/internationalization.mdx
+++ b/src/content/docs/en/guides/internationalization.mdx
@@ -35,7 +35,7 @@ This routing API helps you generate, use, and verify the URLs that your multi-la
    })
    ```
   
-2. Choose and configure a [`routingStrategy`](#routingstrategy) based on the desired URL path for your `defaultLocale`:
+2. Choose and configure a [`routing`](#routingstrategy) based on the desired URL path for your `defaultLocale`:
 
   - `"prefix-other-locales"` (default): URLs in your default language will **not** have a `/[locale]/` prefix. All other locales will.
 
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
This looks great to me! I'll just mention that I think the other open PR has `routingStrategy` in one of its examples, so that (and the implementation PR and changeset) should also be updated, if this is the new config option!

ematipico:
> This looks great to me! I'll just mention that I think the other open PR has `routingStrategy` in one of its examples, so that (and the implementation PR and changeset) should also be updated, if this is the new config option!

Great call out! I update the PR once this one is merged

ElianCodes (approved):

PR #5534, opened 2023-11-30
chore: add MDX to recommended plugins

ElianCodes:
#### Description (required)

Add MDX extension to recommended plugins. So it's easier to install when we're sharing codespaces!


delucis (approved):
Great idea!

PR #5542, opened 2023-12-01
fix: fix dark mode code example

azat-io:
#### Description (required)

When you have the light theme on, the theme toggle button should display the moon, and when you have the dark theme, it should display the sun.

This is done on all sites, including the Astro documentation site.

N/A

### DIFF
--- a/src/content/docs/ru/tutorial/6-islands/2.mdx
+++ b/src/content/docs/ru/tutorial/6-islands/2.mdx
@@ -42,12 +42,12 @@ import PreCheck from '~/components/tutorial/PreCheck.astro';
         border: 0;
         background: none;
       }
-      .sun { fill: black; }
-      .moon { fill: transparent; }
+      .sun { fill: transparent; }
+      .moon { fill: black; }
+      
 
-
-      :global(.dark) .sun { fill: transparent; }
-      :global(.dark) .moon { fill: white; }
+      :global(.dark) .sun { fill: white; }
+      :global(.dark) .moon { fill: transparent; }
     </style>
     ```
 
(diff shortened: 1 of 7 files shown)


at-the-vr (commented):
I guess its a matter of choice which icon to display, that being said having this change is good. 
One thing you did not covered the same changes for pt-br locale

src/content/docs/ru/tutorial/6-islands/2.mdx
```diff
@@ -42,12 +42,12 @@ import PreCheck from '~/components/tutorial/PreCheck.astro';
         border: 0;
         background: none;
       }
-      .sun { fill: black; }
-      .moon { fill: transparent; }
-
+      .sun { fill: transparent; }
+      .moon { fill: black; }
+
```
at-the-vr, line 47: minor nit: additional newline + tabs (could be removed)

azat-io:
> One thing you did not covered the same changes for pt-br language

✅️️️️

at-the-vr (approved):
definitely needs additional review from Maintainers cause of subjective changes
from my perspective its LGTM

at-the-vr:
Bumping the PR up a bit + stackblitz instance for faster reviews ⤵
| Blog  | Link |
| Original Blog  | [Former Theme](https://stackblitz.com/github/withastro/blog-tutorial-demo/tree/complete?file=src%2Fcomponents%2FThemeIcon.astro)  |
| Azat's suggestion applied into same Blog  | [Suggested Theme](https://stackblitz.com/edit/github-mtyekr?file=src%2Fcomponents%2FThemeIcon.astro)  |

delucis:
Thanks again for the PR @azat-io — and thanks for the super helpful review comments @at-the-vr, which made it very easy to understand this PR.

We really appreciate you taking the time to contribute, but I think we want to stick with the current behaviour where the displayed icon matches the current colour theme. I’ll admit this is slightly subjective, and partly an issue with this kind of toggle button which relies on a single icon where it’s not entirely clear if the icon should represent current state or the next state triggered by a click. In this case, we’ll stick with the existing behaviour, but thanks again for your time and I hope you’ll consider making other contributions in the future. It can be a good idea to make a new issue for ideas so we can discuss them before making any code changes. That way we can be sure we’re all on the same page!

PR #5554, opened 2023-12-03
Fix stylying and formatting on API reference

Fryuni:
Update the title for one of the methods that was missing the `` ` ` ``.

Since I was already fixing a markdown formatting I also:
- Removed a chunk of newlines that were unnecessary
- Aligned a table
- Added `;` to the imports that were missing
- Removed unnecessary indentation from one of the examples


sarah11918 (approved):
Looks great, thank you @Fryuni !  🙌

PR #5585, opened 2023-12-04
Fix incorrect endpoint return values in API Reference

mingjunlu:
#### Description

- Some endpoint snippets in [API Reference](https://docs.astro.build/en/reference/api-reference) return plain objects
- This PR changes those return values to `Response` objects

#### Related issues & labels


sarah11918 (approved):
Thank you @mingjunlu ! We will merge these tomorrow with all the other 4.0 docs! 🙌

PR #5593, opened 2023-12-04
[v4 upgrade] show i18n config added back

sarah11918:
#### Description (required)

Since `i18n` still needs configuring, just not within `experimental`, this adds the config at the root level to show where it goes.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v4.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v4.mdx
@@ -57,7 +57,7 @@ See [the changelog](https://github.com/withastro/astro/blob/main/packages/astro/
 
 Remove the following experimental flags from `astro.config.mjs`:
 
-```js del={5-9} title="astro.config.mjs"
+```js del={5-9} ins={11-14} title="astro.config.mjs"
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/upgrade-to/v4.mdx
```diff
@@ -57,7 +57,7 @@ See [the changelog](https://github.com/withastro/astro/blob/main/packages/astro/
 
 Remove the following experimental flags from `astro.config.mjs`:
```
delucis, line 58: How do you feel about something in this direction?

```suggestion
Remove the `devOverlay` experimental flag and move any `i18n` config to the top level in `astro.config.mjs`:
```

delucis (commented):
Thanks for tackling @sarah11918! One thought regarding the accompanying copy but code example looks good to me now. 🙌

src/content/docs/en/guides/upgrade-to/v4.mdx
sarah11918: Looks good!

PR #5595, opened 2023-12-04
[install-auto] add note about dev toolbar

sarah11918:
#### Description (required)

This adds a line after "start the dev server" to the getting started instructions mentioning that you have access to the dev toolbar.

> When viewing your site in the browser, you'll have access to the [Astro dev toolbar](/en/reference/configuration-reference/#dev-toolbar-options). As you build, it will help you inspect your [islands](/en/concepts/islands/), spot accessibility issues, and more.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/install/auto.mdx
+++ b/src/content/docs/en/install/auto.mdx
@@ -85,6 +85,8 @@ If all goes well, Astro should now be serving your project on [http://localhost:
 
 Astro will listen for live file changes in your `src/` directory, so you will not need to restart the server as you make changes during development.
 
+When viewing your site in the browser, you'll have access to the [Astro dev toolbar](/en/reference/configuration-reference/#dev-toolbar-options). As you build, it will help you inspect your [islands](/en/concepts/islands/), spot accessiblity issues, and more.
+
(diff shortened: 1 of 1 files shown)


src/content/docs/en/install/auto.mdx
```diff
@@ -85,6 +85,8 @@ If all goes well, Astro should now be serving your project on [http://localhost:
 
 Astro will listen for live file changes in your `src/` directory, so you will not need to restart the server as you make changes during development.
 
+When viewing your site in the browser, you'll have access to the [Astro dev toolbar](/en/reference/configuration-reference/#dev-toolbar-options). As you build, it will help you inspect your [islands](/en/concepts/islands/), spot accessibility issues, and more.
```
martrapp, line 88: Nitpicking: It already says "Astro _dev_ toolbar", but wouldn't it be helpful to say that this only happens in DEV mode?

sarah11918: I actually started with that included, but there is a whole lot of "dev" in this section already, so it felt a bit much in context!  😄

PR #5606, opened 2023-12-05
fix: Fix a link in getting-started.mdx

san-ghun:
#### Description (required)

Update `getting-started.mdx` file in `src/content/docs/en/` to correct the link to be consistent with the content.
- The original content is linked to the Astro theme page, but suggest it would be better to be linked to the actual tutorial page in Astro docs.
```md
[Build a Blog Tutorial]
(https://astro.build/themes/) --> (https://docs.astro.build/en/tutorial/0-introduction/)
```


ElianCodes (commented):
Thanks for catching this! Just quickly updating the link here

PR #5621, opened 2023-12-05
Fixes site title UX issues

natemoo-re:
#### Description (required)

Tweaks to the site title link UX to address a few usability issues

#### Hover Behavior

Previously, the `:hover` behavior could lead to stuttering because the bounding boxes of the `a` elements were not normalized. There was also not enough padding on the inline axis of the elements.

https://github.com/withastro/docs/assets/7118177/c2e3675c-8bf3-4f5e-bc6e-c068b9a2a07a

This PR adjusts the `a` elements so that the `:hover` behavior is consistent. The inline padding has been adjusted to provide a larger interaction target.

https://github.com/withastro/docs/assets/7118177/242fc0f7-5c33-492f-b9d6-8305cac1d62e

#### Focus (Visible) Behavior

Previously, the `:focus-visible` behavior was unclear because of the size of the `a` elements.

https://github.com/withastro/docs/assets/7118177/294874cc-ccef-455e-bcaf-ce53fdc1f153

This PR adds a pseudo-element to both `a` elements when `:focus-visible` matches to ensure that the focus ring remains visible.

https://github.com/withastro/docs/assets/7118177/b60888a3-c247-4772-ba3f-4c4c0a22ecfa


src/components/starlight/SiteTitle.astro
```diff
@@ -48,14 +49,36 @@ const href = pathWithBase((Astro.props.locale || '') + '/getting-started/');
...
+		top: var(--sl-nav-pad-y);
+		right: 0;
+		bottom: var(--sl-nav-pad-y);
+		left: 0;
+		outline: var(--sl-color-text-accent) auto 2px;
+	}
+	.site-title a:first-child:focus-visible::after {
+		left: -0.5em;
+		right: 0.25em;
 	}
```
delucis, line 82: Might be a cheaper fix if we just `outline-offset: -3px` or something on the `a` directly? That will move the outline inside so it’s visible.

delucis (commented):
Thanks @natemoo-re! This was on my list to revisit — thanks for checking it off :-D

One small suggestion to avoid the complexity of the pseudo element.

src/components/starlight/SiteTitle.astro
natemoo-re: That would be cheaper, yes! I started there but quickly found it much trickier than I anticipated. The height adjustments required for the `:hover` behavior made changing the `outline-offset` impossible. `outline-offset` is applied consistently in all four directions, but we have to bring in the block axis while maintaining the inline axis.

Since we needed more control than the `offset` properties provided, the pseudo-element made the most sense.

natemoo-re: Hmm now that I think of it, it might be easier to apply the outline to the `svg` element rather than the `a` element. Let me try that quickly.

delucis (approved):
Thanks @natemoo-re! I feel like we’re maybe patching over some decisions we could improve — like relying on the space inside the Astro wordmark SVG for spacing — but happy to get this in for now.

PR #5622, opened 2023-12-05
Replace non-break space with regular space [en-only]

Genteure:
#### Description (required)

Replaced the non-break space (`U+00A0`) between "file" and "inside" with regular space.


sarah11918 (approved):
LGTM, thank you!

Yan-Thomas (approved):
LGTM! Thanks @Genteure 🙌

PR #5630, opened 2023-12-05
[en-only] Clarify SSR for bun recipe

jacobthesheep:
#### Description (required)

Update SSR section to reflect that now all Astro SSR adapters support Bun and no longer node specific.

- Closes #
- Suggested label:


sarah11918 (approved):
Always love it when we delete more lines than we add! 😂 

Thanks for this, @jacobthesheep ! Sorry I'm just catching up with even some of these smaller PRs after a pretty hectic launch prep/week.  :raised_hands:

PR #5633, opened 2023-12-05
Update view-transitions.mdx

OliverSpeir:
Not 100% sure about these changes, but I figured might as well see what y'all think.

#### Description (required)

- Remove code block in `### Transitions with forms` because it does not add any value ( in my opinion )

- Add tip to `### Transitions with forms` about `event.preventDefault()` because this was not obvious to me personally 

- Add tip to `## Script behavior during page navigation` about `DOMContentLoaded` ( struggled with the wording here, not sure if it is the best, potentially could remove this tip altogether ) because this is a very common question in the discord

- Add link to `Update Scripts Tutorial` in `## Script behavior during page navigation` because this is a very valuable resource that most don't seem to find and is relevant to this section


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -374,20 +374,6 @@ This has the effect that if you go back from the `/main` page, the browser will
 
 The `<ViewTransitions />` router also supports navigation triggered by form submissions.
```
lilnasy, line 375:
```suggestion
The `<ViewTransitions />` router also animates navigations triggered by form submissions. 
```
You are right, the block below seems redundant to me as well. Its presence may also give the impression that there is something users will have to do make it work. I tried rewording this line hoping to make it clear that not only is it supported, it will work by default.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -396,6 +382,10 @@ The router will trigger in-page transitions from `<form>` elements, supporting b
 </form>
 ```
 
+:::tip
+By default your page will reload after the `POST` request is sent, if you are using Javascript to prevent this with `event.preventDefault()` you will also want to opt out of router transitions for this form. 
+:::
```
lilnasy, line 387: What's the alternative? What happens when `event.preventDefault()` is used without opting out?

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -465,6 +455,13 @@ If you have code that sets up global state, this state will need to take into ac
 
 Module scripts are only ever executed once because the browser keeps track of which modules are already loaded. For these scripts, you do not need to worry about re-execution.
 
+:::tip
+Because module scripts that add event listeners for `DOMContentLoaded` only execute once, you will also need to add listeners for a [Lifecycle event](#lifecycle-events).
+:::
+
+See the [Update Scripts Tutorial](/en/tutorials/add-view-transitions/#update-scripts) for more information.
+
+
```
lilnasy, line 464: This is an important caveat to communicate and deserves more elaboration. I think `Full site view transitions (SPA mode)` would be the appropriate place to explain the architecture that ultimately makes this necessary.

sarah11918:
Hey @OliverSpeir , thanks so much for hopping in here, and just a quick note to let you know this is now finally on my plate for this week!

One reason I haven't yet looked over this is because I know it's going to take a bit of time/work, and I'll let you in on a little secret how I know: suggestions that add "tips" are almost never the right way to go about adding new information... and you added two of them!  :sweat_smile: 

Asides should almost always be left to truly tangential information. If I let everyone add their idea in a note or a tip or a caution, then that's all docs would be.. people tacking on notes all over the place! Without even reading your submission, what this tells me is we need some content rework here so that the necessary information is actually *part of* our text: if we say things right the first time, then we don't NEED flashy tips pointing things out!

So, that's why this one has taken me a while to return to (not a "quick win" for me during a busy launch week), and I'll be spending time on this before the weekend for sure!

src/content/docs/en/guides/view-transitions.mdx
OliverSpeir: ViewTransitions *transitions* to the default (either sends post or get) which if user is not set up for that seems like it just refreshes the page

OliverSpeir:
Makes sense! I figured as much 😅

sarah11918:
Thanks for your patience @OliverSpeir ! I tried to rework the content into the text here. Please check this for accuracy and make any corrections, and let's see whether we can get your content incorporated in more like this kind of style/structure!

OliverSpeir:
Love it,

I reworded to clarify both post and get methods will transition the page

sarah11918:
Fantastic! Happy with this one! 🥳

PR #5635, opened 2023-12-06
Add tip for GitHub Private Pages

alvgaona:
#### Description

The current documentation does not completely take into account private GitHub Pages within organizations. GitHub will create a random URL that needs a different treatment.


src/content/docs/en/guides/deploy/github.mdx
```diff
@@ -42,6 +42,17 @@ Astro maintains the official `withastro/action` to deploy your project with very
...
 
+    :::tip[GitHub Organizations]
+    If you're not using custom domains for your [private pages](https://docs.github.com/en/enterprise-cloud@latest/pages/getting-started-with-github-pages/changing-the-visibility-of-your-github-pages-site),
+    GitHub will create a random URL for you. 
+
+    ```text
+    https://<random-string>.pages.github.io/
+    ```
+
+    Make sure you configure the `site` key with the given URL - no `base` needed.
```
dreyfus92, line 53:
```suggestion
    Make sure to set the `site` key to match the auto-generated URL. There's no need to specify a `base` key.
```

dreyfus92 (changes requested):
Hey @alvgaona, thanks for tackling this 😁. I left a small suggestion.

src/content/docs/en/guides/deploy/github.mdx
alvgaona: Awesome. Thank you!

src/content/docs/en/guides/deploy/github.mdx
```diff
@@ -42,6 +42,17 @@ Astro maintains the official `withastro/action` to deploy your project with very
...
 
+    :::tip[GitHub Organizations]
+    If you're not using custom domains for your [private pages](https://docs.github.com/en/enterprise-cloud@latest/pages/getting-started-with-github-pages/changing-the-visibility-of-your-github-pages-site),
+    GitHub will create a random URL for you. 
+
+    ```text
+    https://<random-string>.pages.github.io/
+    ```
+
+    Make sure to set the `site` key to match the auto-generated URL. There's no need to specify a `base` key.
```
lilnasy, line 53: What happens when the `site` config is not set or set to an incorrect origin?

sarah11918:
Hi @alvgaona ! This is just a quick note to let you know I'll be looking at this submission this week!

I know it might seem like a small addition, and maybe shouldn't need much review, but even though you've only added a "tip"... when you look at your submission in the surrounding context of the page, it looks like this:

notecautiontip

This... does not make for good docs.  :sweat_smile: 

What this tells me is that we need to rework our actual documentation with better guidance on using `base`... just adding on more notes, cautions and tips is not the solution here. This is why I haven't yet had the time to review your PR (it's not a "quick win" for me during a busy launch week) and clearly is an indication of much bigger docs work that needs to happen. 

But, it is on my plate to address this! Thank you for your patience!

sarah11918:
OK! Thank you for your patience @alvgaona !  :raised_hands: 

There were just too many notes and cautions etc. so that told me that this content needed to be restructured. In fact, I made sections for using a custom domain vs not, since that often determines what you do or don't have to set.

I did of course include the information you provided, just in a different place/way. Thank you for contributing to this page!

alvgaona:
> OK! Thank you for your patience @alvgaona ! 🙌
> 
> There were just too many notes and cautions etc. so that told me that this content needed to be restructured. In fact, I made sections for using a custom domain vs not, since that often determines what you do or don't have to set.
> 
> I did of course include the information you provided, just in a different place/way. Thank you for contributing to this page!

Thank you, @sarah11918! I

PR #5653, opened 2023-12-06
Update astro version to v4.0 in the docs

jdbruxelles:
This PR addresses the recent update of Astro from v3 to v4 The changes include updating the version number in the sidebar, the version announcement panel, add-content-collections.mdx and add-view-transitions.mdx to reflect the latest release.


Yan-Thomas:
Thanks for this @jdbruxelles, we recently migrated to Starlight so this text is completely different, so I'll have to close this one. The UI text we let translators update at their time, so the label is not misleading if the page itself isn't updated!

jdbruxelles:
Thanks @Yan-Thomas for the clarification. YOU're making good job :tada:

PR #5676, opened 2023-12-07
Add integration page headers

delucis:
#### Description (required)

- Adds integration metadata to integration page headers with a page title override similar to the header we had in the previous site iteration.

   Example Alpine.js integration page header showing version number, GitHub, npm, and changelog links


sarah11918:
Oooh, tried making my screen smaller and saw this though: 
image

delucis:
Great catch @sarah11918! We forgot to test narrower viewports on stream. Just pushed a fix — should be good now 🤞

sarah11918:
Confirmed @delucis ! I cannot break this version, so approved!

PR #5680, opened 2023-12-07
Update astro-components.mdx to add boolean attribute prop feature.

jinsupark:
Added notes on how Astro props supports boolean attributes without explicitly writing prop={true}

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/core-concepts/astro-components.mdx
+++ b/src/content/docs/en/core-concepts/astro-components.mdx
@@ -163,6 +163,29 @@ const { greeting = "Hello", name = "Astronaut" } = Astro.props;
 <h2>{greeting}, {name}!</h2>
 ```
 
+### Boolean Attribute Props
+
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @jinsupark , thank you for this contribution!

I think, however, this might be more suited to a short recipe. It doesn't quite fit here. 

Firstly, it's more about building a custom component than explaining literally how props are passed, which is what this "basics" section is for: a super basic introduction that doesn't bog anyone down in too many options or details. Secondly, it *uses* a `<slot />` before they've been introduced and explained.

If you'd like to take a look at the recipes on our More Recipes page, I could see this fitting as a dedicated recipe to build a custom component that handles Boolean props. But it's not really a fit for this page here. If you do want to expand it out and make this a recipe with some steps to follow to guide people through scenario, then you can take a look at our guide for submitting a recipe: https://contribute.docs.astro.build/guides/recipe-writing/ This will also give you a little more flexibility to develop the example, or get creative, rather than try to squish it in to a "basics" example, if you want!

So, I'm going to close *this* PR, but I think there's a place for this content if you'd like to create a recipe!

PR #5707, opened 2023-12-09
Make hero layout more robust

delucis:
#### Description (required)

- Use a grid layout instead of absolute positioning to overlay hero elements (avoids relying on padding for positioning which breaks if text wraps in some languages).

- Tweak maximum text width so it doesn’t overlap with facepile items on large viewports.

- Tweak style on “Join us” tagline in facepile so it doesn’t push the grid row larger if the text wraps onto three lines.

| before | after |
|---|---|
| Portuguese getting started page with headline overlapping avatar images in the face pile | Portuguese getting started page with all elements laid out correctly |


TheOtterlord (approved):
This looks great in every viewport size & language combination I tried. Thanks!

Yan-Thomas (approved):
LGTM!

PR #5709, opened 2023-12-09
Update middleware.mdx

azakiio:
#### Description (required)

The import path for `defineMiddleware` and `sequence` seem to be outdated.
Using `astro/middleware` instead of `astro:middleware` solves it

- Suggested label: typo

### DIFF
--- a/src/content/docs/en/guides/middleware.mdx
+++ b/src/content/docs/en/guides/middleware.mdx
@@ -115,7 +115,7 @@ You can import and use the utility function `defineMiddleware()` to take advanta
 
 ```ts
 // src/middleware.ts
-import { defineMiddleware } from "astro:middleware";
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi! I know that `astro/middleware` should also work, but this was an intentional documentation change once middleware was moved out of an experimental flag. The correct (and preferred) import should be from `astro:middleware`, so if that's not working, that maybe points to an astro error!

Would you mind filing an issue in the main astro repo https://github.com/withastro/astro/issues with a reproduction to let them know? 

I'm going to close this documentation PR because I know we intend this documented import to be the correct one. And if it's not working, it's on our devs to fix it. But thank you for taking the time to contribute the fix that worked for you! If it turns out that this change is needed in docs, you are welcome to reopen it!

PR #5711, opened 2023-12-09
fix(tabs): disable the smooth scrolling

### LINKED ISSUE #5678: Inconsistent scroll behavior between tabs and sidebar
### 📝 Issue Description

The links on the sidebar jump directly to the header, without any scroll animation.

The links for the tabs are scrolling smoothly to the tabs.

Every page with tabs

### 🤔 Expected Behavior

Either neither should have a smooth scroll or both should.

But I don't think the tabs should have any scroll at all, Starlight's [own docs](https://jam.dev/c/60a16be9-b177-4d66-adbc-1a51e3e6dfb1) don't have that.

### 👀 Current Behavior

https://jam.dev/c/3c56e87d-407c-42c4-a216-ec46ebe063ae

### 🖥️ Browser

Chrome and Firefox on Linux

### 📄 Additional Information

_No response_

yasnbouz:
#### Description (required)
Disable the smooth scrolling on tabs

- Closes #5678 
- Suggested label: bug


sarah11918:
Hi @yasnbouz , thank you for making a contribution to Astro Docs!  🙌 

I can see that you tackled an open issue, but I should warn you that there was no discussion around that issue, so we haven't had a chance to decide whether this is in fact an issue that we want or intend to fix.  😄  That just means this contribution will have to wait as we figure out what we in fact want to do about that opened issues, so you'll have to hang on a little longer!

It's always safe to stick to the "help wanted" labels when working on issues. Sometimes people open what they feel are issues, but for various reasons are not something we decide or intend to change. We have also just recently migrated our docs to Starlight, so we have a lot of review planned as we work out some of the new internals, and it's possible that this contribution may not end up fitting into our larger planned work. That doesn't mean it was a bad solution to the problem, just that this is not something we have yet requested assistance with, and it may not end up being compatible with other site maintenance we are implementing.

So, please hold on for a while longer as this is not something we had planned to look at, and are trying to incorporate reviewing this along with our other maintenance!

src/components/tabs/Tabs.astro
```diff
@@ -47,8 +47,6 @@ export interface Props {
 						index
 					);
 				});
-
-				starlightTab.scrollIntoView({ behavior: 'smooth' });
```
delucis, line 51: Thanks for the PR @yasnbouz — I don’t think this is quite what was meant in #5711. This change prevents all scrolling when changing tabs, but this results in a bad user experience with tab content jumping up and down on the page when tabs are toggled.

For example, try switching between the tabs in the ‘Using “atoms”’ section of the sharing state guide: https://docs-git-fork-yasnbouz-fix-tabs-astrodotbuild.vercel.app/en/core-concepts/sharing-state/#using-atoms

When I switch from Preact to Svelte for example, the tabbed content jumps down the page.

yasnbouz: Thank you, @sarah11918 and @delucis, for your feedback. I apologize for misunderstanding the issue

TheOtterlord (approved):
Thanks @yasnbouz! :raised_hands: 

This feels like a good compromise for the time being. We're working on improving the behaviour of tabs in Starlight, so hopefully soon, we'll no longer need to anchor tabs to the top!

PR #5727, opened 2023-12-10
[i18nIgnore] remove unused listener in sharing state recipe

Tambouil:
#### Description (required)

There is no need to add a listener on the Dialog.astro component is this recipe. The listener is added on the Button.astro example above. The element with the id #openDialog is undefined here

- Closes #
- Suggested label:


delucis (approved):
Thanks for the fix @Tambouil! Added a commit removing this line in the `pt-br` translation too and this should be good to go — thanks for contributing 🚀

PR #5733, opened 2023-12-12
add: favicon to head in template

dreyfus92:
#### Description (required)

Added favicon to the i18n tracker


scripts/lib/translation-status/template.html
```diff
@@ -12,6 +12,8 @@
 			content="Translation progress tracker for the Astro Docs site. See how much has been translated in your language and get involved!"
 		/>
 		<link rel="canonical" href="https://i18n.docs.astro.build/" />
+		<link rel="icon" href="/favicon.ico" sizes="32x32" />
+		<link rel="shortcut icon" href="/favicon.svg" type="image/svg+xml">
```
HiDeoo, line 16: Thanks for the contribution 🙌 

Small question: I tested locally and I did not get any favicon. Where are the assets supposed to come from? Should they also be part of the PR?

dreyfus92: They are supposed to be coming from public, but apparently they are not being loaded. I don't know if the `href` should be replaced with something like `https://docs.astro.build/favicon.ico`.

delucis: The i18n dashboard is not built with Astro so doesn’t use `public/`. Given we’ll soon switch to [Lunaria](https://lunaria.dev/), maybe this is something to fix there if @Yan-Thomas hasn’t already considered it?

delucis: > I've considered and was about to implement inline SVG favicon support, you can see more about this method here: [austingil.com/svg-favicons](https://austingil.com/svg-favicons/)

TIL!

> I can just replace with what I just previously mentioned, it works tho. And maybe later patch it when Lunaria has figured out the best way to tackle this.

@dreyfus92 that makes sense! Yeah we can just point at the docs.astro.build version for now 👍

dreyfus92: @delucis cool, cool. I will make the fix rn.

ElianCodes (approved):
Looks like it does indeed work now!

PR #5634, opened 2023-12-06
Move and Adapt Glossary Page

VoxelMC:
#### Description (required)

- Moved the glossary file to `src/content/docs/en/glossary.mdx`.
- Added frontmatter to ensure that it is compatible with the content collection.
- Added a small overview.

#### Related issues & labels (optional)

- Suggested label: Glossary

PR #5739, opened 2023-12-12
glossary: add `glossary.md`

at-the-vr:
#### Description (required)
- Introduce glossary.md to `main`

- Suggested label: glossary 

Former `glossary.md` from #5634 (`glossary` branch)

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -21,4 +21,6 @@ Some future guidelines for this:
 
 ### Islands
 
-**Islands** is a frontend architecture that results in better frontend performance by helping you avoid monolithic JavaScript patterns and stripping all non-essential JavaScript from the page automatically. The technique that this architectural pattern builds on is also known as partial or selective hydration. In Astro, an `island` refers to any interactive UI component on the page. Read more about islands in the [Astro Islands](https://docs.astro.build/en/concepts/islands/) page.
+An [Astro Island](https://docs.astro.build/en/concepts/islands/) is an [interactive UI framework component](https://docs.astro.build/en/core-concepts/framework-components/) on the page.
+
+**Islands** is a frontend architecture that results in better frontend performance by helping you avoid monolithic JavaScript patterns and stripping all non-essential JavaScript from the page automatically. The technique that this architectural pattern builds on is also known as partial or selective hydration. In Astro,  Read more about islands in the  page.
--- a/glossary.md
+++ b/glossary.md
@@ -1,0 +1,24 @@
+---
+title: Glossary 
+description: A collection of technical terminology commonly used in the Astro documentation and community.
+i18nReady: false
+---
+
+This glossary describes common terminology used in the Astro documentation and the Astro community.
+
+{ /* // Uncomment this section when we have more terms to add to the glossary.
+It is separated into Astro-specific terms and general web development terms.
+## Astro-Specific Terms 
+
+Some future guidelines for this:
+- Astro-Specific: Content Collections, ViewTransitions, Astro Component
+- General: Editor, LSP, Front-End Library, Framework, Metaframework, Framework Component
+*/}
+
+### Component
+
+Astro components are the basic building blocks of any Astro project. They are HTML-only templating components with no client-side runtime. You can spot an Astro component by its file extension: `.astro`. Read more about components in the [Astro Components](https://docs.astro.build/en/core-concepts/astro-components/) page.
+
+### Islands
+
+**Islands** is a frontend architecture that results in better frontend performance by helping you avoid monolithic JavaScript patterns and stripping all non-essential JavaScript from the page automatically. The technique that this architectural pattern builds on is also known as partial or selective hydration. In Astro, an `island` refers to any interactive UI component on the page. Read more about islands in the [Astro Islands](https://docs.astro.build/en/concepts/islands/) page.
--- a/glossary.md
+++ b/glossary.md
@@ -17,7 +17,7 @@ Some future guidelines for this:
 
 ### Component
 
-Astro components are the basic building blocks of any Astro project. They are HTML-only templating components with no client-side runtime. You can spot an Astro component by its file extension: `.astro`. Read more about components in the [Astro Components](https://docs.astro.build/en/core-concepts/astro-components/) page.
+[Astro Components](https://docs.astro.build/en/core-concepts/astro-components/) are the basic building blocks of any Astro project. They are HTML-only templating components with no client-side runtime. You can spot an Astro component by its file extension: `.astro`. 
 
 ### Islands
 


glossary.md
```diff
@@ -0,0 +1,24 @@
...
+## Astro-Specific Terms 
+
+Some future guidelines for this:
+- Astro-Specific: Content Collections, ViewTransitions, Astro Component
+- General: Editor, LSP, Front-End Library, Framework, Metaframework, Framework Component
+*/}
+
+### Component
+
+Astro components are the basic building blocks of any Astro project. They are HTML-only templating components with no client-side runtime. You can spot an Astro component by its file extension: `.astro`. Read more about components in the [Astro Components](https://docs.astro.build/en/core-concepts/astro-components/) page.
```
sarah11918, line 20:
```suggestion
Astro components are the basic building blocks of any Astro project. They are HTML-only templating components with no client-side runtime. You can spot an Astro component by its file extension: `.astro`. 

Source: [Astro Components](https://docs.astro.build/en/core-concepts/astro-components/)
```
What would you think about a format more like this, that avoids needing an extra sentence? Simply referencing the most appropriate page/section? Play around with the format (maybe even linking inside the main text like:

```suggestion
[Astro Components](https://docs.astro.build/en/core-concepts/astro-components/) are the basic building blocks of any Astro project. They are HTML-only templating components with no client-side runtime. You can spot an Astro component by its file extension: `.astro`. 
```
... but I think the fewer non-essential sentences here, the better?

Genteure: What about a mix of both styles? For "Components" I like the second one (link within the main text) a lot more, but for some terms the page title might not appear in the text.

glossary.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Islands
+
+**Islands** is a frontend architecture that results in better frontend performance by helping you avoid monolithic JavaScript patterns and stripping all non-essential JavaScript from the page automatically. The technique that this architectural pattern builds on is also known as partial or selective hydration. In Astro, an `island` refers to any interactive UI component on the page. Read more about islands in the [Astro Islands](https://docs.astro.build/en/concepts/islands/) page.
```
sarah11918, line 24:
```suggestion
An [Astro Island](https://docs.astro.build/en/concepts/islands/) is an [interactive UI framework component](https://docs.astro.build/en/core-concepts/framework-components/) on the page.

**Islands** is a frontend architecture that results in better frontend performance by helping you avoid monolithic JavaScript patterns and stripping all non-essential JavaScript from the page automatically. The technique that this architectural pattern builds on is also known as partial or selective hydration. In Astro,  Read more about islands in the  page.
```
So, I didn't edit here, but I will point out that I think it's way more helpful for a glossary entry to not bury the lede. 😅  If people wanted to read docs, they'd.... read docs?

FOR ALL OF YOU:

I think part of making these entries helpful is deciding what your *purpose* is. I would argue that (mostly) repeating docs is not a terribly useful thing to have, nor for us to spend time on, when there are so many other things to do.

So, I think before all of you start adding more and more entries, really think about what this glossary is going to ACHIEVE. How should it be different from what I've already written, and edited, in docs itself? How are you going to make it *useful* and worthwhile? WHO is going to read this over docs, and for what goal, and how can you cater it to THAT audience to serve that purpose?

lilnasy: While the edit was included, the overarching question remains.

I would think search serves the purpose of "I am not familiar with this term". Is there a case where search is inadequate and a glossary page is necessary?

glossary.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Component
+
+[Astro Components](https://docs.astro.build/en/core-concepts/astro-components/) are the basic building blocks of any Astro project. They are HTML-only templating components with no client-side runtime. You can spot an Astro component by its file extension: `.astro`. 
+
+### Islands
+
+An [Astro Island](https://docs.astro.build/en/concepts/islands/) is an [interactive UI framework component](https://docs.astro.build/en/core-concepts/framework-components/) on the page.
+
+**Islands** is a frontend architecture that results in better frontend performance by helping you avoid monolithic JavaScript patterns and stripping all non-essential JavaScript from the page automatically. The technique that this architectural pattern builds on is also known as partial or selective hydration. In Astro,  Read more about islands in the  page.
```
at-the-vr, line 26: Sarah's questions this suggestion tries to satisfy -
1. How should it be different from what I've already written, and edited, in docs itself?
Provide a summary that gives **"Ahh haa 💡 that's what this term means"**  to the person searching.

2. How are you going to make it useful and worthwhile?
This is intended to give readers a baseline to what `Astro Docs describes in verbose`. Get a quick summary and take decisive steps accordingly

3. Who is going to read this over docs, and for what goal, and how can you cater it to THAT audience to serve that purpose?
A person fed up from searching stuff in Docs and Internet, and wants a simpler & quicker explanation to what they are after. Their motive is to know if the intended term is useful in their project or not. That's what this Glossary should provide with, help the person with minimum research for maximum coding perf

Here's a new take on terminologies mentioned:

```suggestion
### Component

[Astro Components](https://docs.astro.build/en/core-concepts/astro-components/) are a flexible group of HTML snippets providing Developers with code fence (---) that allows them to import different components, run async functions and standard Javascript variables to use inside their snippets.
Astro Components are server-rendered that is they are turned into their HTML equivalents during build-time with 0 Javascript which improves performance of your website tremendously.

### Islands
Island is a component in a Project that allows developers to choose when and where they want client side JavaScript to be injected,  providing the same component based experience. Island allows developers to control JavaScript flow by server rendering your components into HTML equivalents and when browsed by clients, inject JavaScript in order to make specific, isolated components interactive. This shaves a lot of unused JavaScript, improves SEO alongside User Experience and upholds major accessibility guidelines, since the server practically renders only standard HTML.

[Astro Island](https://docs.astro.build/en/concepts/islands/) is a interactive UI Framework component running on a Astro Site(which is essentially server-rendered static page of HTML). Developers get to control its hydration and interaction prior build time and in return they get to:
1. Limit JavaScript in their Astro Project
2. Parallel load all your interactive components
3. Improve Lighthouse metrics by fastening page load speed
```

VoxelMC: > Sarah's questions this suggestion tries to satisfy -
> 
> 1. How should it be different from what I've already written, and edited, in docs itself?
>    Provide a summary that gives **"Ahh haa 💡 that's what this term means"**  to the person searching.
> 2. How are you going to make it useful and worthwhile?
>    This is intended to give readers a baseline to what `Astro Docs describes` but concise. Get a quick summary and take decisive steps accordingly
> 3. Who is going to read this over docs, and for what goal, and how can you cater it to THAT audience to serve that purpose?
>    A person fed up from searching stuff in Docs and Internet, and wants a simpler & quicker explanation to what they are after. Their motive is to know if the intended term is useful in their project or not. That's what this Glossary should provide with, help the person with minimum research for maximum coding perf
> 
> Here's a new take on terminologies mentioned:

What would you think about having the first line of the definition be one very clear and concise sentence, such as my example below
> Islands is an architectural pattern that allows HTML content to be rendered at build-time without immediate javascript interactivity, then provided with that functionality on the client side when the page is loaded (hydration).

I am also interested in having a longer, slightly more thorough explanation of the term. However, as you noted, it must be abstracted from what is already within the Docs and made more straightforward. 

Here is my proposition: we can write a REALLY good one- or two-sentence description for these two terms. Then, we can see how much more information we need. I think we should aim to provide enough information that the reader doesn't *need* to read more Docs if all they were looking for was a short explanation anyway. Even if the glossary were present, anyone who *really* wants to read the docs on it would just read the docs, as Sarah mentioned.  

I'm not sure if I'm making any sense here, but I hope you get what I mean! Haha

I think this approach answers lilnasy's question:
> Is there a case where search is inadequate and a glossary entry is necessary?

I think we can treat the glossary entry as an entry point of understanding. `Search` enables the reader to undertake a lot of information gathering at once. If the glossary entry can be the first search suggestion, perhaps we can save the time of those who want an easy but **sufficient** description by severely reducing the size of the information undertaking. **This is the sole reason I believe it has a place. To give the reader a choice between reading many and a few pages.**

sarah11918:
I think we said it was OK to close this one and move to the HackMD!

PR #5741, opened 2023-12-12
Simplify site title styling and logic

natemoo-re:
#### Description (required)

Simplifies the implementation and CSS for the `SiteTitle` component.
- Updates the SVG files used as the logos. Both have been normalized and optimized. 
- Spacing is no longer determined by the spacing in the SVG files, it is all controlled by CSS.
- Removes the light/dark logo from the Starlight config, preferring to import a responsive `.svg` that updates its styles to match the `[data-theme]` attribute.
- Removes the hacky `:focus-visible` behavior that required a pseudo-element. Since the SVGs have been normalized, we can now rely on the browser's `outline` behavior for these elements.
- Removed all CSS in this file that was no longer needed.

- Follow-up to #5621
- Closes DX-870


src/styles/custom.css
```diff
@@ -1,5 +1,9 @@
 /** Custom Styles **/
 
+/* Ensure that <header> nav element doesn't allow overflow-y */
+.page > .header {
+	overflow-y: hidden;
+}
```
natemoo-re, line 6: Should this change be upstreamed to Starlight? Feels gross to have it here, but I didn't want to add a component override just to add an overflow rule.

src/components/starlight/SiteTitle.astro
```diff
@@ -1,121 +1,59 @@
...
-						src={logos.light?.src}
-						width={logos.light?.width}
-						height={logos.light?.height}
-					/>
-				)}
-			</a>
-		)
-	}
-	<a {href} class="docs" set:html={docs} aria-label="Astro Docs" />
+	<a class="astro-logo" href={main} set:html={logo} aria-label="Astro" />
```
natemoo-re, line 12: Vastly simplified this by using one SVG that has a `style` block

src/styles/custom.css
natemoo-re: Ah you're totally correct. Putting the `transform` on the `<svg>` should solve the problem. Good catch!

PR #5765, opened 2023-12-13
fix: add spacing to 'added in' texts

kevinzunigacuellar:
#### Description

Add spacing to `<Since />` elements

| Before                                                                                                                     | After                                                                                                                      |
|----------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|
| image | image |


ElianCodes:
Wouldn't it make more sense to add padding to the `<Since />` component instead of wrapping all of them in a `<p />?

kevinzunigacuellar:
> Wouldn't it make more sense to add padding to the `<Since />` component instead of wrapping all of them in a `<p />`?

It would but there are some instances of Since that are inlined 😅

delucis (approved):
LGTM! Thanks for tackling this @kevinzunigacuellar 💜

sarah11918:
Thanks, Kevin! I updated Astro Docs Docs to show this usage of the component in a standalone line!

PR #5740, opened 2023-12-12
Update Astro to v4 and Starlight to v0.15

delucis:
#### Description (required)

- Updates docs dependencies on Astro and Starlight to latest and also the full unified JS ecosystem to match

PR #5772, opened 2023-12-13
[i18nIgnore] Fix type errors

delucis:
#### Description (required)

While working on #5740 I noticed a bunch of type errors in non-Astro files that are not getting checked currently by `astro check` (caused by bug fixed in https://github.com/withastro/language-tools/pull/722).

Some are fixed in #5740 where they relate to rehype/remark types updated in that PR, but this PR follows up fixing some more. Once this PR and #5740 are both merged we should have no more type errors (except for two upstream in Starlight which I’ll tackle separately).


dreyfus92 (commented):
LGTM ✅, Just one question, shouldn't you add `i18nIgnore` in the title of the PR so it wouldn't be tracked by i18n tracker or just apply the changes to en version?

delucis:
> shouldn't you add `i18nIgnore` in the title of the PR

Yes! I meant to do that — thanks for catching @dreyfus92 🙌

sarah11918 (approved):
Quick preview check LGTM

PR #5773, opened 2023-12-13
[issue template] - remove Stackblitz requirement for content issues

sarah11918:
#### Description (required)

This sets the StackBlitz field as not required, plus tweaks some wording to make this issue for reporting incorrect content. I'll add another template itself specifically for new content requests, so this one does not have to pull double duty.

- Closes #
- Suggested label:


delucis (approved):
LGTM! ✅

kevinzunigacuellar:
LGTM! I like that the template is more focused on content rather than individual user issues.

PR #5779, opened 2023-12-13
Refactor `cachedFetch()` to always retry failed fetches

delucis:
#### Description (required)

This PR updates our `cachedFetch()` utility to use `p-retry` to automatically retry up to 10 times if a fetch fails. We were already using this strategy in the `<Version>` component — wrapping `cachedFetch()` in a retry — and this is now our default behaviour, so I refactored that component to remove the now redundant retry logic.

Should help with some flaky GitHub API calls we’ve been encountering (with so many contributors, we now need 8 calls just to get them all, so the chance of one flaking out has increased).

I’ve also switched the face pile component used in the new landing page hero to use `cachedFetch()` instead of a plain `fetch()`.

### DIFF
--- a/src/util-server.ts
+++ b/src/util-server.ts
@@ -5,6 +5,7 @@
 
 // @ts-expect-error Package without types we can’t do anything about.
 import EleventyFetch from '@11ty/eleventy-fetch';
+import retry from 'p-retry';
 
 export type CachedFetchOptions = {
 	duration?: string;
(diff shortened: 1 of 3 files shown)


src/util-server.ts
```diff
@@ -5,6 +5,7 @@
 
 // @ts-expect-error Package without types we can’t do anything about.
 import EleventyFetch from '@11ty/eleventy-fetch';
+import retry from 'p-retry';
 
 export type CachedFetchOptions = {
 	duration?: string;
```
hippotastic, line 11: It would be great to have an optional `retryOptions?: Options` property here. `Options` is exported by `p-retry`. This should be passed to the `retry` call to allow configuring its behavior, including the number of retries and the function that gets called on error.

delucis: I did consider this, but seeing as we don’t actually need it right now, I thought it probably falls under [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)?

hippotastic: True, we can always add it later if we actually need it.

hippotastic (approved):
LGTM!

In case this still isn't enough to prevent the flakiness in CI, we can try moving the 11ty cache into a folder that gets persisted across builds.

PR #5782, opened 2023-12-13
Update 404 regex

### LINKED ISSUE #5730: 404 link fails for one level deep paths

TheOtterlord:
#### Description (required)

Improves the 404 regex to catch more URLs (e.g. `/test/`) & removes some excess padding along the bottom

- Closes #5730


delucis:
Could you explain a bit more what this PR addresses? I can’t see any change in the previews 😅

TheOtterlord:
It just removes a bit of padding at the bottom of the page, preventing a slight scroll on some viewport sizes where that padding isn't doing anything. Also updates the regex to fix one specific issue. Thinking back though, preferably the regex would instead be made up of every active language. It only caught my eye because I use `/test` sometimes to work on 404 pages :sweat_smile:

delucis:
> Also updates the regex to fix one specific issue.

What was the issue? I tested `/test` and `/x` and it seemed ok but maybe I missed something!

TheOtterlord:
An example of the link break is here: https://docs.astro.build/test/

src/components/NotFound.astro
```diff
@@ -132,7 +140,7 @@ const dictionary = Object.fromEntries(
 				const dictionary = JSON.parse(this.dataset.dictionary);
 
 				// Get current language from the URL path.
-				const langCodeMatch = location.pathname.match(/\/([a-z]{2}-?[a-z]{0,2})\//);
+				const langCodeMatch = location.pathname.match(/\/([a-z]{2}(-[a-z]{0-2})?)\//);
```
delucis, line 143: Thanks for explaining! I’d missed the original issue about this, so didn’t have that context — sorry if I seemed a bit dense 😅 

So thinking about this a little more, this issue still persists if the route is `/aa` or `/zz` or something similar that matches our regex but is not actually a language. We may want to move all the mutations behind a check that `translations` exists. This would assure us that the match was a valid language and only update the link, `lang` attribute, and text if so.

TheOtterlord: Yep! Sounds great. Pushed a new version that builds a regex from our available languages, so I believe `lang` should already default to `en` if a known language isn't found, therefore `translations` should always exist.

src/components/NotFound.astro
```diff
@@ -130,9 +138,10 @@ const dictionary = Object.fromEntries(
 			try {
 				if (!this.dataset.dictionary) return;
 				const dictionary = JSON.parse(this.dataset.dictionary);
+        const languages = Object.keys(dictionary);
 
 				// Get current language from the URL path.
-				const langCodeMatch = location.pathname.match(/\/([a-z]{2}-?[a-z]{0,2})\//);
+				const langCodeMatch = location.pathname.match(`\\/(${languages.join('|')})\\/`);
```
delucis, line 144: I don’t think this works? The argument to `match()` here is just treated as a string and tested for basic equality. And when constructing a regular expression from a string, you don’t need to escape the `/` either:

```suggestion
				const langCodeMatch = location.pathname.match(new RegExp('/(' + languages.join('|') + ')/'));
```

TheOtterlord: The string gets converted into a regex by [`.match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#parameters). Ahh, yep. I think the `\\` was left over from when I was working on the inline regex. Made a commit to remove

delucis (approved):
LGTM! Thanks @TheOtterlord — I had no idea `String.prototype.match()` did that conversion for you. Tested out a few different 404 routes and it all worked as expected.

Re: the CSS changes, FWIW, I’d likely not have bothered with them as I don’t think the extra space mattered too much and my bias is towards as little code as possible. Was there a specific visual issue that pushed you to add those?

TheOtterlord:
The changes were just to deal with some scroll people in discord and in the issue were talking about. Although, I may revisit this by moving to an absolute positioning or something in the future :thinking: I have some other plans to discuss for the 404 when we get to design retro anyway, so for now the layout can probably remain as-is.

delucis (approved):
Gotcha — sounds good to me and helps keep this PR scoped to a single issue too which is a nice bonus. Thanks for putting up with me nitpicking here 😅 

LGTM! Fix ready to launch 🚀

PR #5793, opened 2023-12-15
Add import to i18n recipe - translateRoutes

Gelbotron:
#### Description (required)

There seems to be a missing import for a function that was added previously to `i18n/utils.ts`

- Closes: N/A
- Suggested label: N/A


sarah11918 (approved):
Thank you so much for fixing this for us! Looks great, and welcome to Team Docs!  :partying_face:

PR #5811, opened 2023-12-17
i18n(ko-KR): add `server-side-rendering.mdx`

jsparkdev:
#### Description (required)

add `server-side-rendering.mdx`

- Closes #
- Suggested label:


sarah11918 (approved):
Reviewed using our Docs checklist, because we want as many translated pages before the new year as we can get!  :muscle:

PR #5817, opened 2023-12-17
[i18nIgnore] Fix highlighted code errors

itsmatteomanf:
Fixes a small issue in rows selected in a code block, off-by-one errors.

### DIFF
--- a/src/content/docs/es/core-concepts/astro-components.mdx
+++ b/src/content/docs/es/core-concepts/astro-components.mdx
@@ -287,7 +287,6 @@ const { titulo } = Astro.props;
 Los slots pueden ser transferidos a otros componentes. Por ejemplo, cuando se crean plantillas anidadas:
 
 ```astro title="src/layouts/BaseLayout.astro" {9,12}
-// src/layouts/BaseLayout.astro
 ---
 ---
 <html lang="en">
--- a/src/content/docs/es/core-concepts/astro-components.mdx
+++ b/src/content/docs/es/core-concepts/astro-components.mdx
@@ -286,11 +286,10 @@ const { titulo } = Astro.props;
 
 Los slots pueden ser transferidos a otros componentes. Por ejemplo, cuando se crean plantillas anidadas:
 
-```astro title="src/layouts/BaseLayout.astro" {10,13}
+```astro title="src/layouts/BaseLayout.astro" {9,12}
 // src/layouts/BaseLayout.astro
 ---
 ---
-
 <html lang="en">
 	<head>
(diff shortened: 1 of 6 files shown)


itsmatteomanf:
Not sure if I should probably update translations, if they exist...

delucis (approved):
Looks good to me! Usually we suggest avoiding updating multiple languages at once, but for a small fix like this, if you can fix all other languages where this is present, that would also be acceptable (and we’d stick the `i18nIgnore` keyword into the commit message to avoid triggering translation status changes).

itsmatteomanf:
@delucis done all other languages, the issue was that some space was added (or removed) and the numbers didn't align. Found a language which had numbers wrong even more ahah

Do check that the commit message is correct with the correct usage of the keyword.

itsmatteomanf:
> Usually we suggest avoiding updating multiple languages at once, but for a small fix like this, if you can fix all other languages where this is present, that would also be acceptable (and we’d stick the `i18nIgnore` keyword into the commit message to avoid triggering translation status changes).

I asked because of this, I know the policy, which makes sense, but it was such a small fix...

src/content/docs/es/core-concepts/astro-components.mdx
```diff
@@ -286,11 +286,10 @@ const { titulo } = Astro.props;
 
 Los slots pueden ser transferidos a otros componentes. Por ejemplo, cuando se crean plantillas anidadas:
 
-```astro title="src/layouts/BaseLayout.astro" {10,13}
+```astro title="src/layouts/BaseLayout.astro" {9,12}
 // src/layouts/BaseLayout.astro
```
delucis, line 290: I think in this case, this comment also needs removing:

```suggestion
```

Currently it looks like this: 
code snippet with two titles and the wrong lines highlighted

delucis (changes requested):
Thanks @itsmatteomanf! Noticed one of these that looks off.

itsmatteomanf:
> Thanks @itsmatteomanf! Noticed one of these that looks off.

I accepted it... but locally looks like this.

VSCode doesn't handle it, as it doesn't align with what it has locally, as you are asking to change it how it already is ahah

image

delucis:
I see the comment still their in your screenshot (second line here):

Two lines both of which include the code sample title, first as a title attribute, then as a code comment

So I think all’s well — let’s see how the deploy preview looks after that commit.

delucis (approved):
Yup, looks correct now! 🙌 

code block with the expected lines highlighted

PR #5825, opened 2023-12-18
replace prefetch integration's docs with removal notice.

sarah11918:
#### Description (required)

Removes the `@astrojs/prefetch` integration docs. Replaces with a deprecation warning and front matter changes that should make this a regular, static page.

Page Preview: https://docs-git-sarah11918-patch-1-astrodotbuild.vercel.app/en/guides/integrations-guide/prefetch/

Also, removed from integrations nav

image


dreyfus92 (approved):
LGTM 🫡

sarah11918:
@dreyfus92 I will await your incoming Spanish translation in 4... 3... 2...

src/content/docs/en/guides/integrations-guide/prefetch.mdx
```diff
@@ -1,161 +1,11 @@
...
-You can also check our [Astro Integration Documentation][astro-integration] for more on integrations.
-
-## Contributing
-
-This package is maintained by Astro's Core team. You're welcome to submit an issue or PR!
-
-## Changelog
-
-See [CHANGELOG.md](https://github.com/withastro/astro/tree/main/packages/integrations/prefetch/CHANGELOG.md) for a history of changes to this integration.
+:::note[removed]
```
delucis, line 7: I guess we usually sentence case headings, so this should be too?

```suggestion
:::note[Removed]
```

delucis: More optionally, should this be a caution rather than a note?

```suggestion
:::caution[Removed]
```

src/content/docs/en/guides/integrations-guide/prefetch.mdx
```diff
@@ -1,161 +1,11 @@
...
+`@astrojs/prefetch` has been replaced by the [built-in `prefetch` feature](/en/guides/prefetch/) introduced in Astro 3.5. See the [migration guide](/en/guides/prefetch/#migrating-from-astrojsprefetch) for instructions on updating an older project.
 
-[astro-integration]: /en/guides/integrations-guide/
+If you still are using this integration in pre-v3.5 Astro project, you can find [its previous documentation](https://github.com/withastro/astro/blob/c47478bbf6b21973419f25234c68efb59466b368/packages%2Fintegrations%2Fprefetch%2FREADME.md) from an older version of Astro.
```
delucis, line 10: “its previous documentation” struck me as not great for a link label, so I think updating that to “the `@astrojs/prefetch` README” would be a nice a11y win. Also suggesting some tweaks because I found “previous docs”/“older version” a bit brain bending, but it is late, so my brain may be bend-prone.

```suggestion
If you are still using this integration in a pre-v3.5 Astro project, you can read an archived copy of [the `@astrojs/prefetch` README](https://github.com/withastro/astro/blob/c47478bbf6b21973419f25234c68efb59466b368/packages%2Fintegrations%2Fprefetch%2FREADME.md) on GitHub.
```

delucis (commented):
I wasn’t 100% sure of the approach removing the `type: integration` and related frontmatter at first, but thinking through the various options, I think you’re right and this is a good approach. Thanks for updating this @sarah11918! Made a couple of suggestions on the new copy.

PR #5857, opened 2023-12-20
Update example to use webp instead of avif

florian-lefebvre:
#### Description (required)

Following a discussion from Discord, it updates the background image format to use webp instead avif because:
- Avif is not the most standard format, with known issues on MS Edge
- A lot of beginners use this code snippet as is so it would be better to give them a "fully working" code
- This snippet must show we can pass a custom format. Even if webp is the default, it shows that it can be set

- Closes N/A
- Suggested label: code snippet update, low effort


sarah11918 (approved):
I followed the discussion on Discord, so I believe Erika is already fine with the change!

PR #5880, opened 2023-12-22
public files copied before build

### LINKED ISSUE #5464: Document how to dynamically create/add/download files to public folder during the build process
### 📚 Subject area/topic

Bundling

https://docs.astro.build/en/guides/imports/#files-in-public

In multiple locations, the Astro docs state that any files in the `public` folder will be copied to the output bundle, without any modification.

However the docs are missing one big caveat: This is only true, if the files are already existing before the build process starts. It's not possible to add files to `public` during the build process (e.g. downloading images via API) and have them copied/bundled to the output. It seems that Astro is indexing the public folder BEFORE the build process, and will only copy/bundle all the files that have been there before the build process started.

This should be mentioned in the docs, and maybe there could maybe even be a mention of possible workarounds (e.g. use a shell script to copy the files after the `npm run build` call).

https://stackblitz.com/edit/github-aqszdy?file=src%2Fpages%2Findex.astro

sarah11918:
#### Description (required)

Adds a short clarification about the timing of when assets from public are copied into the build folder.

- Closes #5464

### DIFF
--- a/src/content/docs/en/core-concepts/project-structure.mdx
+++ b/src/content/docs/en/core-concepts/project-structure.mdx
@@ -96,7 +96,7 @@ It is a common convention to store your CSS or Sass files in a `src/styles` dire
 
 ### `public/`
 
-The `public/` directory is for files and assets that do not need to be processed during Astro's build process. These files will be copied into the build folder untouched.
+The `public/` directory is for files and assets in your project that do not need to be processed during Astro's build process. These existing files will be copied into the build folder untouched, and then your site will be built.
 
 This behavior makes `public/` ideal for common assets like images and fonts, or special files such as `robots.txt` and `manifest.webmanifest`.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/core-concepts/project-structure.mdx
```diff
@@ -96,7 +96,7 @@ It is a common convention to store your CSS or Sass files in a `src/styles` dire
 
 ### `public/`
 
-The `public/` directory is for files and assets that do not need to be processed during Astro's build process. These files will be copied into the build folder untouched.
+The `public/` directory is for files and assets in your project that do not need to be processed during Astro's build process. These existing files will be copied into the build folder untouched, and then your site will be built.
```
sarah11918, line 99:
```suggestion
The `public/` directory is for files and assets in your project that do not need to be processed during Astro's build process. The files in this folder will be copied into the build folder untouched, and then your site will be built.
```

TheOtterlord (approved):
Sounds good :+1:

PR #5884, opened 2023-12-22
Fix issue with rendering VT events

Yan-Thomas:
#### Description (required)

It seems there's currently an issue with displaying View Transition events inside Markdown, causing this reported issue:
image

This PR fixes this in docs by wrapping it around with backticks. I'll make an repro and report as bug later.

- Closes #
- Suggested label:


TheOtterlord (approved):
Thanks! This looks good to me. Looks like both translated versions already use backticks, so this should be the only case.

PR #5885, opened 2023-12-22
Fix typos in Internationalization guide

matiboux:
#### Description (required)

This PR fixes two typos in the **Internationalization** guide:

- Fix wrong import for `getLocaleByPath` in example code:
  Imported from `astro:18n`, should be from `astro:i18n`.

- Fix wrong prototype for `getRelativeLocaleUrlList`:
  First argument is `locale: string`, should be `path?: string` (from [Astro's source code](https://github.com/withastro/astro/blob/4660aaa4051bd23fb77811968bbe56e2021161a2/packages/astro/client.d.ts#L221)).

- Fix wrong prototype for `getAbsoluteLocaleUrlList`:
  First argument is `locale: string`, should be `path?: string` (from [Astro's source code](https://github.com/withastro/astro/blob/4660aaa4051bd23fb77811968bbe56e2021161a2/packages/astro/client.d.ts#L229)).

### DIFF
--- a/src/content/docs/en/guides/internationalization.mdx
+++ b/src/content/docs/en/guides/internationalization.mdx
@@ -367,7 +367,7 @@ export default defineConfig({
 ```astro
 ---
 // src/pages/index.astro
-import { getLocaleByPath } from "astro:18n";
+import { getLocaleByPath } from "astro:i18n";
 console.log(getLocaleByPath("en")) // will log "en"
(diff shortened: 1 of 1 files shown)


TheOtterlord (approved):
Good catch! Thanks a lot! :raised_hands:

Confirmed the type arguments match up with the virtual module & reference. Let's get this merged!

PR #5886, opened 2023-12-22
feat: clarification on dynamic file creation

### LINKED ISSUE #5464: Document how to dynamically create/add/download files to public folder during the build process

pReya:
#### Description (required)

Adding another paragraph on top of #5880 to clarify that dynamically created files during build process will not be part of the output folder, but'll have to be copied manually.

- Closes #5464

### DIFF
--- a/src/content/docs/en/core-concepts/project-structure.mdx
+++ b/src/content/docs/en/core-concepts/project-structure.mdx
@@ -98,6 +98,10 @@ It is a common convention to store your CSS or Sass files in a `src/styles` dire
 
 The `public/` directory is for files and assets in your project that do not need to be processed during Astro's build process. The files in this folder will be copied into the build folder untouched, and then your site will be built.
 
+:::caution
+Since `public/` will be copied to the build folder **before** the build process runs, this means that any files that you add during your build process (e.g. images that you fetch from external sources) will not end up in your build folder. You'll have to copy them manually or via a script.
+:::
+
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi! I appreciate the submission, but as I wrote in the issue, I really think this is too much for this page. Certainly, I'm really hesitant to have a flaming caution on our most basic project structure page. As I wrote in the issue, I would consider a recipe with a workaround, and we could use our "related recipes" link component here (e.g. "Related Recipe: Adding files to the public folder during the build process". But I really do think this goes a little too far here. So, I'm closing this, but would be open to having a recipe submitted!

PR #5888, opened 2023-12-23
fix: add `--no-pager` option to git command

CaptainOfPhB:
#### Description (required)

The previous git command can not work as expected. With no `--no-pager` option, git command will output the paging result, so `execSync` can't get the returned result.

image

After adding the `--no-pager` option(meanwhile set `pretty` to `%cI`and wrap `filepath` with double quote), we get what we want.

image

- Suggested label: update doc


TheOtterlord (commented):
Thanks for the PR @CaptainOfPhB!

I see what you mean about the pager when using the command in the terminal. However, I found that calling `execSync` with the git command works without `--no-pager`, and only in the terminal did I have that issue.

Looking at the new formatting, running `execSync` with `--pretty=%cI`, it threw an invalid date error, displaying the last character as a `\n`, something that doesn't happen with the old argument. Did you have a specific issue with the old formatting?

CaptainOfPhB:
> Thanks for the PR @CaptainOfPhB!
> 
> I see what you mean about the pager when using the command in the terminal. However, I found that calling `execSync` with the git command works without `--no-pager`, and only in the terminal did I have that issue.
> 
> Looking at the new formatting, running `execSync` with `--pretty=%cI`, it threw an invalid date error, displaying the last character as a `\n`, something that doesn't happen with the old argument. Did you have a specific issue with the old formatting?

Thank you for your reply!

The previous command run failed yesterday but worked just now when I tried again. Looks so wired :(

But the `filepath` should wrap with double quotes or the command will fail if there is blank space in the filename.

sarah11918:
Hi @CaptainOfPhB ! Would you please update your suggestion if you think all that's now needed are the double quotes? Then, we can get this merged!  :partying_face:

src/content/docs/en/recipes/modified-time.mdx
```diff
@@ -43,7 +43,7 @@ Learn how to build a [remark plugin](https://github.com/remarkjs/remark) that ad
      export function remarkModifiedTime() {
        return function (tree, file) {
          const filepath = file.history[0];
-         const result = execSync(`git log -1 --pretty="format:%cI" ${filepath}`);
+         const result = execSync(`git --no-pager log -1 --pretty=%cI "${filepath}"`);
```
TheOtterlord, line 46:
```suggestion
         const result = execSync(`git log -1 --pretty="format:%cI" "${filepath}"`);
```

TheOtterlord (approved):

TheOtterlord:
Thanks for the update! Let's get this merged! :partying_face:

PR #5932, opened 2023-12-24
i18n(it): translated some files of `recipes/`

BlackdestinyXX:
**Description (required)**

translated some files of `recipes/`

**Related issues & labels (optional)**

- Suggested label: i18n

### DIFF
--- a/src/content/docs/it/recipes/add-yaml-support.mdx
+++ b/src/content/docs/it/recipes/add-yaml-support.mdx
@@ -1,0 +1,64 @@
+---
+title: Installare un plugin Vite o Rollup
+description: Impara come importare dati YAML aggiungendo un plugin Rollup al tuo progetto.
+i18nReady: true
+type: recipe
+---
+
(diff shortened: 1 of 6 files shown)


sarah11918:
Thanks for these contributions, @BlackdestinyXX ! There are some problems with the structure of the links. Can you go back and fix these? There are some key rules our links have to follow:

- Use the language prefix for the language you're translating in. (e.g. `href="/it/guides/..."`). This makes sure that a reader is always staying in the Italian version of the site. Even if the page content is not translated, they will still see the page structure in Italian.
- If the page you are linking to is already translated, then use the **translated** heading (after the `#` symbol), even though the rest of the URL structure is in English. (e.g. `/it/core-concepts/astro-components/#componenti-html`). If you are ever not sure, visit the page in Italian, go to the section and copy the link directly from the page.

These two rules should fix almost all of the link errors you might see on your PRs. Thank you! :purple_heart:

sarah11918:
I am closing this PR because all of the changes were reviewed and fixed in the other PR. Next time, please be sure that your PRs don't also include content from other PRs!

PR #5955, opened 2023-12-26
[adapter-reference.mdx] update `AstroAdapter` interface

### LINKED ISSUE #5181: Adapter implementation documentation
### 📚 Subject area/topic

Integration API, Adapter API

https://docs.astro.build/en/reference/integrations-reference/
https://docs.astro.build/en/reference/adapter-reference/

Hello, I've noticed some small inconsistencies in the documentation about implementing adapters.

1. The [Quick API Reference](https://docs.astro.build/en/reference/integrations-reference/#quick-api-reference) doesn't include the `setAdapter` option for the `astro:config:done` hook that is mentioned in the [Adapter API](https://docs.astro.build/en/reference/adapter-reference/#building-an-adapter) page.
2. The definition for the AstroAdapter interface that is passed into the `setAdapter` call [here](https://docs.astro.build/en/reference/adapter-reference/#building-an-adapter) is slightly different than its typing in the code [here](https://github.com/withastro/astro/blob/d5dc5ce04acd19d9c9b7a29c2519408f6bdc3b83/packages/astro/src/%40types/astro.ts#L1828).
    1.  `previewEntrypoint` and `args` are missing
    2. There are a few differences in which properties are optional, but that might be intentional
    3. In the code `args` is typed as `any`, but in practice it seems like trying to pass a function or an object with a function value won't work. It might be helpful to note any restrictions (if args has to be serializable for example). 

One other small question I ran into was about the `serverEntrypoint` option. The documentation describes it as "The entrypoint for server-side rendering." The example value given is `"@matthewp/my-adapter/server.js"`. It wasn't obvious to me how the string should be structured in general. The example appears to be pointing to a node module, so it might be nice to mention you can also use a relative path from the root of your Astro project. This question came up in the discord [here](https://discord.com/channels/830184174198718474/830184175176122389/1150936602118459482) and the asker concluded it wasn't possible, but it has been working for me. 

n/a

sarah11918:
#### Description (required)

The docs for the `AstroAdapter` interface don't match the code in `@types/astro.ts`. Assuming the code is correct, then the docs should be updated! [SOURCE](https://github.com/withastro/astro/blob/d5dc5ce04acd19d9c9b7a29c2519408f6bdc3b83/packages/astro/src/%40types/astro.ts#L1828)

Also adds  the `setAdapter` option for the `astro:config:done` hook in on the Integrations API page in another commit, to fully close the issue below. [SOURCE](https://github.com/withastro/astro/blob/d5dc5ce04acd19d9c9b7a29c2519408f6bdc3b83/packages/astro/src/%40types/astro.ts#L2060)

- Closes #5181


bhainesva (approved):
lgtm thanks for the update!

PR #5977, opened 2023-12-28
docs: update Netlify import

orimdominic:
#### Description (required)
A warning message about @astrojs/netlify/functions folder is reported on the terminal when I run the build script.

See screenshot attached in the PR
Screenshot from 2023-12-28 13-57-30

- Closes #
- Suggested label:


TheOtterlord (approved):
Thanks @orimdominic! This looks good to me. Looks like this is also still recommended in the netlify adapter's README. So we'll need to update it there as well

PR #5980, opened 2023-12-28
[i18nIgnore] New bun types package @types/bun

gtrabanco:
#### Description (required)

Update to use new @types/bun package instead of `bun-types` which will be deprecated in the future.

https://bun.sh/blog/bun-v1.0.19#re-introducing-types-bun-formerly-bun-types

- Closes #
- Suggested label: documentation

#### Other information

I have updated the link and installation on all languages that bun recipe is available because only requires to update the link, delete `tsconfig` configuration and update shell command.


sarah11918 (approved):
Thanks so much @gtrabanco ! These all look great. Welcome to Team Docs!  :partying_face:

PR #5982, opened 2023-12-29
update: added Response.json tip

jprando:
#### Description (required)

Added tip showing how to use `Response.json`  
so you don't need to enter `application/json` in the `content-type` `header`, every time you need to return JSON

- Closes #
- Suggested label: api-routes

#### For Astro version: `4.x`.

[doc preview link - Guides > Enpoints](https://docs-2o3iug2h0-astrodotbuild.vercel.app/en/core-concepts/endpoints/#server-endpoints-api-routes)


jprando:
ChatGPT suggested using the following text, with the intention of making it easier to read, if you agree and it makes sense, please change the text:

You can use the `Response.json` method to generate a response without the need to specify the `content-type` in the `header` as `application/json`.
```
// code here
```
Note that we refrain from employing the `new` operator preceding `Response.json`.

TheOtterlord:
Thanks for the PR @jprando!

Astro docs exists to document Astro, and how to best use it's APIs, and not really for documenting different ways of achieving the same thing in JavaScript itself. I love this new API, and I'm sure future examples could use this. However, if we do begin to use this, we'd want to look at using it more widely. Rather than looking at it as "a second way of doing the same thing", instead at how we can use it while still giving examples that can be adapted easily to fit a reader's needs.

We might look at updating a bunch of examples to use this API in the new year, to better present this as how you write JSON endpoints, rather than one more way of writing JSON endpoints.

PR #5978, opened 2023-12-28
Update integrations to MDX syntax

ElianCodes:
#### Description (required)

Updates all integration guides to use MDX & Expressive Code syntax, instead of regular MD.

More info on [the HackMD](https://hackmd.io/ye3ciu4QRd-2UdyE_BxQPQ)

PR #5999, opened 2023-12-30
Standardize the Integration Installation Steps 

VoxelMC:
#### Description (required)

I have updated every integration guide with a standardized version of the installation steps outlined below.
I also updated the formatting to be consistent with the rest of the `docs` (at least in the installation sections, I did not go past that)

This is meant for #5978

---

### Manual Install

Instead of lists: 

---

Please let me know if I should add or remove anything to this! Sorry that it's such a big PR... I figured that if I am going to do this, I should do it right the first time!!!

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
+++ b/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
@@ -12,14 +12,7 @@ This **[Astro integration][astro-integration]** adds [Alpine.js](https://alpinej
 
 ## Installation
 
-There are two ways to add integrations to your project. Let's try the most convenient option first!
-
-### Quick Install
-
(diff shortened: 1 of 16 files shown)


sarah11918 (approved):
Thank you so much for this! This was incredibly helpful and thorough. Even though I edited some of the text, all I had to do was find and replace and update tons at once. Really appreciate this one!  💜

VoxelMC:
> Thank you so much for this! This was incredibly helpful and thorough. Even though I edited some of the text, all I had to do was find and replace and update tons at once. Really appreciate this one! 💜

I am glad to have been helpful! I think the next skill for me and these sorts of things is to think a little more outside the box and rewrite what has already been written a bit more :)

PR #6040, opened 2024-01-02
Document --host and --port flags for preview

bluwy:
#### Description (required)

`--host` and `--port` works for `astro preview` for a very long time (since the command existed?). I moved the two flags under `dev` flags to common flags. And roughly updated the "dev server" phrase to "dev server and preview server".

- Closes #
- Suggested label: 

The Astro core PR that updated the flags is https://github.com/withastro/astro/pull/9572. We don't have to align the merge release with it though. I think it's more of missing docs.


VoxelMC:
The change makes sense to me. Thank you bluwy :)

PR #6073, opened 2024-01-03
Update default image service section

bluwy:
#### Description (required)

From https://github.com/withastro/astro/pull/9546#discussion_r1439775572, Sarah suggested that the "Default image service" section could have a link to the `image.service` option explaning how to configure options for an image service.

- Closes #
- Suggested label:  

#### For Astro version: `4.1`. See astro PR https://github.com/withastro/astro/pull/9546

Don't think there's a big harm if we merge this without waiting for  https://github.com/withastro/astro/pull/9546? I guess I'll let y'all decide so the PR is properly labelled 😅


VoxelMC (approved):
LGTM!

Yan-Thomas (approved):
LGTM! I'll let Sarah make the call of when to merge.

sarah11918 (approved):
Looks great! Since the specific Sharp config option isn't actually live until the minor release, I'll save this to merge with the other release docs. But, we should have had a link to config reference regardless, so this is perfect!

PR #6075, opened 2024-01-03
Update framework-components.mdx

leochenliu:
add some explainations to named-slot

- Closes #
- Suggested label:


liruifengv:
Hello. Can you describe in more detail why you made this change?

If it is necessary to add additional instructions, I think we should add them first in the original English docs.

I thought maybe we could add a link to https://docs.astro.build/en/core-concepts/astro-components/#named-slots to achieve the same purpose?

Yan-Thomas:
Hey @leochenliu, welcome to Team Docs, thanks for your contribution!

To reiterate liruifengv's comment,  as this seems to be a **content** change, we need it to be proposed to the English version of the docs so that all translations can benefit from it afterwards.

Could you update this PR to change the English version instead with your proposed change?

sarah11918:
Hi @leochenliu ! I translated the content addition, and if the translation is accurate, this is not something we would add to the English text. The note would be too large and seems to repeat information.

So, I am going to close this PR. But, it is possible that this is just how my browser translated and if you contribute content in English it will read differently! You can make a new English PR if you have content that is **significantly different** from what is already inside the note and we can look at that.

PR #6076, opened 2024-01-03
Fix: Type "string" is not assignable to type "Provider".

lucialv:
If you copy and paste the actual code, you will have the error: `Type "string" is not assignable to type "Provider".`
So I just changed the line 438 and 426 to fix the error :)

Before:
image

After:
image


sarah11918 (approved):
Looks great, @lucialv ! Thank you so much for taking the time to correct this. Welcome to Team Docs! 🥳

PR #6150, opened 2024-01-05
Update cloudflare.mdx

alexanderniebuhr:
#### Description (required)

- relates to withastro/adapters#58
- merge with Cloudflare adapter version: **>8.1.0**


alexanderniebuhr:
Even if this does not depend on core PR, it does depend on the linked adapters PR. So I still added `merge-on-release`, let me know if that was wrong ❤️

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -201,12 +201,30 @@ export default defineConfig({
...
 ### `imageService`
 
-`imageService: "passthrough" | "cloudflare"`
+`imageService: "passthrough" | "cloudflare" | "compile"`
 
 Determines which image service is used by the adapter. The adapter will default to `passthrough` mode when an incompatible image service is configured. Otherwise, it will use the globally configured image service:
 
 * **`cloudflare`:** Uses the [Cloudflare Image Resizing](https://developers.cloudflare.com/images/image-resizing/) service.
 * **`passthrough`:** Uses the existing [`noop`](/en/guides/images/#configure-no-op-passthrough-service) service.
+* **`compile`:** Uses Astro's default service (sharp) to optimize images during build.
```
sarah11918, line 210:
```suggestion
* **`compile`:** Uses Astro's default service (sharp) to optimize images on pre-rendered routes at build time. This disables image optimization during SSR for pages rendered on-demand.
```
I'm wondering whether we can just say something like this instead of the caution. Would something of this style this cover it?

sarah11918 (commented):
Thanks @alexanderniebuhr - Quick thought re: the description below, and how we can maybe avoid having a caution by getting the idea across in the description itself!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
sarah11918: @alexanderniebuhr Changed my suggestion above. What do you think? This also matches the others that simply say "Uses service X" (and don't say "to optimize images") so that the focus can be on when it does/doesn't work.

sarah11918 (approved):
LGTM!

sarah11918:
@alexanderniebuhr Can you please update the PR description to include the version number of Cloudflare this is to be released with? It just helps us when scanning. Any maintainer can merge when they notice the announcement post of the correct version, if by chance this slips your attention when the next minor is released.

sarah11918:
@alexanderniebuhr Heads up there's now a conflict to resolve here!

PR #6160, opened 2024-01-05
Update external-links.mdx

jeromeabel:
Add common optional best practices for external links, like `rel="nofollow, noopener, noreferrer" target="_blank"` attributes

#### Description (required)
The generated output of external links should follow some common security practices.

Instead of having: `<a href="https://website.com">Link</a>`

We could prefer: `<a href="https://website.com" rel="nofollow, noopener, noreferrer" target="_blank">Link</a>`

Some resources:
- https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener
- https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable

- Closes #

### DIFF
--- a/src/content/docs/en/recipes/external-links.mdx
+++ b/src/content/docs/en/recipes/external-links.mdx
@@ -1,13 +1,16 @@
 ---
-title: Add icons to external links
-description: Learn how to install a rehype plugin to add icons to external links in your Markdown files.
+title: Customize external links
+description: Learn how to install a rehype plugin to add attributes and icons to external links in your Markdown files.
 i18nReady: true
 type: recipe
 ---
 
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
-Using a rehype plugin, you can identify and modify links in your Markdown files that point to external sites. This example adds icons to the end of each external link, so that visitors will know they are leaving your site.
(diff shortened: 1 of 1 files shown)


jeromeabel:
@florian-lefebvre Do you want to check this out ?

delucis:
Hi @jeromeabel! Thanks for taking the time to open this PR.

I would not recommend this pattern in our docs I don’t think. In general, opening links in a new tab just because they are from a different site is not considered a best practice. See https://www.nngroup.com/articles/new-browser-windows-and-tabs/ for example for some of the UX research on this. Doing so overrides a user's ability to choose how they want a link to open. While there are exceptions to this rule for contexts like form flows etc. setting this for all external links like this recipe shows would generally be a bad idea.

I appreciate you looking to improve the Astro docs and sorry this PR didn’t work out. For future reference, sometimes it can be better to first open an issue suggesting your idea before making the PR — that way we can get feedback to you before the work on the PR starts.

jeromeabel:
Hi @delucis. 

Sorry for the noise. You point me the right way to contribute, thanks. Or even discuss on Discord first ?

I was quite optimistic about this one... I followed the advice from this FrontEnd Masters course "Advanced Web Development Quiz - Q16: Refer Policies".I will consider changing my external links strategy thanks to your feedback.

PR #6181, opened 2024-01-06
Add Pwa Support (no offline support )

abdullhakim-sami:
just just the ability to "add to home screen " on any device

#### Description (required)

This feature might be small but i see it everywhere in docs for example

https://developer.mozilla.org/en-US/

you can add this site to your home screen and have it take 100% of screen height and width

also you can add shortcuts for example "/reference/configuration-reference/" etc maybe usefull maybe not

also you can now have Astro Docs on Google Play or Ios or chromeOS.


abdullhakim-sami:
i used npm ig that's why it errored?

abdullhakim-sami:
you can fix it by looking at this guide

abdullhakim-sami:
you can fix it by looking at this guide


https://astro-pwa-recipe.vercel.app/recipe/generate-sw

TheOtterlord (commented):
Thanks for the PR @abdullhakim-sami!

Taking a first look at this, I spotted a couple issues with this implementation at this point:

As-is, this returns an English web manifest. I think if possible, we'd like to make this more accessible to people using translated content. The web manifest schema doesn't seem to support multi-lingual sites itself, and all implementations I've seen so far to handle multi-lingual sites seem to rely on server-rendering the manifest based on a cookie preference.

Also, the generated `sw.js` file for the service worker seems to reference every existing page twice, causing it to swell to ~740kb in size. But more importantly, it appears to attempt to precache all pages on the site.

abdullhakim-sami:
> Thanks for the PR @abdullhakim-sami!

you are welcome this isn't a "real " PR just an Experiment 

I tried running the site and hosted it using zrok it litterally was sending 2Gb of files to any user that visits the site lol 


this is because im using the default option of caching everything on first load

there are better approches of caching only viewed pages but honstly docs shouldn't be cached most of the time cuz they update alot

you can remove the glob patterns from the astro config in workbox and the site won't cache anything 


> 
> The web manifest schema doesn't seem to support multi-lingual sites itself
> 
Having the Manifest be in English or the main language of the site isn't a big problem honstly cuz the Astro logo and title is always in English 

>  But more importantly, it appears to attempt to precache all pages on the site.


you can have an injected sw which will allow you to customize the caching in the way you like by writting the sw its more advance

TheOtterlord:
After some further discussion, we've decided it's not worth taking on the maintenance of a PWA (at least right now). There are definitely ways of making this world, and it would be cool to find a nice caching solution for a better offline experience, but with docs updating as frequently as it does, this could be a bit of a pain to perfect. There are other issues with manifests too, like how the base URL will always remain the same as when you installed the app (e.g. whatever locale you chose at the time, or just always `en`). Thanks for the experiment though! It's great to be able to explore options like this.

PR #6186, opened 2024-01-07
add package.json hint

nadar:
For non-super-duper experts, it can be helpful to have a hint that those files must be exported, see my struggle and the helpful answer of @Fryuni 

https://github.com/withastro/astro/issues/7721#issuecomment-1878267678


sarah11918:
Hi @nadar! I'm glad that you found a solution that worked for you, and that you want to help others who have the same issue! Thank you for coming to add extra information to the docs.

We really try not to add "notes" and "tips" unless they fit certain criteria. These should be things that are almost *not* a part of docs at all, but extra, external or tangential things. If the information is truly important, then it should be properly documented as a part of the instructions or explanation. 

For example, if it is necessary to also update another file, then we should properly show that file being updated as part of the complete instructions. This does not go in a "note" if it's key information. Would you be willing to revise this to include it at the appropriate step, not using a "note" for things that should have been done in the first place?

nadar:
Thank you @sarah11918 for guiding me in the right direction. I've revised the sentence and eliminated the note syntax. Hopefully, it's better now. Since English is not my first language, I don't want to mess up the quality of the Astro documentation. 🙂

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -339,6 +339,8 @@ injectRoute({
 });
 ```
 
+When publishing the above mentioned example package `@fancy/dashboard` to npm, it's required to export the `dashboard.astro` file in the `package.json` export section as follows: `"exports": {"./dashboard.astro": "./dashboard.astro"}`. Otherwise, the file will be looked up in the project folder, which will cause a `no such file or directory` exception.
```
VoxelMC, line 342:
``````suggestion
When publishing your package (`@fancy/dashboard`, in this case) to npm, you must export `dashboard.astro` in your `package.json`. Otherwise, the file will be looked up in the project folder, which will cause a `no such file or directory` exception.

```json {"exports"}
{
  "name": "@fancy/dashboard",
  // ...
  "exports": { "./dashboard.astro": "./dashboard.astro" }`
}
```
``````

**Review by VoxelMC — COMMENTED** [review]
> Thank you @sarah11918 for guiding me in the right direction. I've revised the sentence and eliminated the note syntax. Hopefully, it's better now. Since English is not my first language, I don't want to mess up the quality of the Astro documentation. 🙂

Hello! Thank you for taking on the task of ensuring no one makes this mistake in the future!

I have taken a stab at revising this in a way that is more consistent with the Astro docs at the moment.

I would also like to request someone in `core` to confirm that this is the correct solution, before moving forward!

I have a question about this, as I have never written an integration. Would
```json
{
  "exports": "./dashboard.astro"
}
```
be the same as
```json
{
  "exports": { "./dashboard.astro": "./dashboard.astro" }
}
```
?

Fryuni:
Those are not the same @VoxelMC

`"export": "./dashboard.astro"` allows importing that page from the root of the component:
```ts
import Dashboard from '@some/integration';
```

It is equivalent to this:
```json
"exports": {
  ".": "./dashboard.astro"
}
```

To import from a subpath like this
```ts
import Dashboard from '@some/integration/dashboard.astro';
```

Is what requires this export:
```json
"exports": {
  "./dashboard.astro": "./dashboard.astro"
}
```

The same applied for entrypoints passed to `injectRoute`

More information here: https://nodejs.org/api/packages.html#subpath-exports

VoxelMC:
> Those are not the same @VoxelMC
> 
> `"export": "./dashboard.astro"` allows importing that page from the root of the component:
> ```ts
> import Dashboard from '@some/integration';
> ```
> 
> It is equivalent to this:
> ```json
> "exports": {
>   ".": "./dashboard.astro"
> }
> ```
> 
> To import from a subpath like this
> ```ts
> import Dashboard from '@some/integration/dashboard.astro';
> ```
> 
> Is what requires this export:
> ```json
> "exports": {
>   "./dashboard.astro": "./dashboard.astro"
> }
> ```
> 
> The same applied for entrypoints passed to `injectRoute`
> 
> More information here: https://nodejs.org/api/packages.html#subpath-exports
> 
> It's not an Astro requirement, Node won't even try to resolve the import if it is not declared in the package, so Vite's module loader won't be called, and neither is Astro's Vite plugin.

Ohhhh, that makes a lot more sense. You explained it much better than the npm and nodejs docs, imo! Thanks Fryuni :)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -339,6 +339,16 @@ injectRoute({
 });
 ```
 
+When publishing your package (`@fancy/dashboard`, in this case) to npm, you must export `dashboard.astro` in your `package.json`. Otherwise, the file will be looked up in the project folder, which will cause a `no such file or directory` exception.
```
sarah11918, line 342:
```suggestion
When publishing your package (`@fancy/dashboard`, in this case) to npm, you must export `dashboard.astro` in your `package.json`:
```
This is a case where I'm not sure all the negative information is helpful. If they follow the instructions, they shouldn't need to worry about how it could go wrong.

sarah11918:
Would love a final review over this content from @nadar and @Fryuni ! (Especially since I proposed shortening the suggestion.)

What is the group consensus here? Do we think this is enough to be helpful in the docs??

VoxelMC:
> Would love a final review over this content from @nadar and @Fryuni ! (Especially since I proposed shortening the suggestion.)
> 
> What is the group consensus here? Do we think this is enough to be helpful in the docs??

I think it is a pretty niche case. However, I can see it being a very annoying bug to research if it happens to someone. I agree with shortening the suggestion. I say keep it!

What do you say to it being a `:::note` aside? Probably no, but a thought!

Fryuni:
I think the shortened version is enough, no need to get in the details of how it fails otherwise.

Users will see that on their terminal if they don't follow.

LGTM

sarah11918:
Alright, I'm going to take this all as NWTWWHB! 

(PS: Not a "note" because it's actually a required step. All the asides are for things outside of "normal procedure")

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -339,6 +339,16 @@ injectRoute({
 });
 ```
 
+When publishing your package (`@fancy/dashboard`, in this case) to npm, you must export `dashboard.astro` in your `package.json`:
+
+```json {"exports"}
```
sarah11918, line 344:
``````suggestion
```json title="package.json" "exports"
``````

PR #6206, opened 2024-01-09
warn about extra sanitization on RSS page

mayank99:
#### Description (required)

Discussed beforehand with Sarah, this PR expands the tip about `sanitize-html` to warn about potentially too much sanitization.

When I first added RSS to my blog, I simply followed the tutorial and it took me quite a bit of time to understand that many parts of the post content were broken because `sanitize-html` was stripping out elements and attributes beyond just `<script>` and `<style>`.

- Closes #

### DIFF
--- a/src/content/docs/en/guides/rss.mdx
+++ b/src/content/docs/en/guides/rss.mdx
@@ -154,7 +154,7 @@ items: import.meta.glob('./blog/*.{md,mdx}'),
 The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
 
 :::tip
-A package like [`sanitize-html`](https://www.npmjs.com/package/sanitize-html) will make sure that your content is properly sanitized, escaped, and encoded.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/rss.mdx
```diff
@@ -154,7 +154,7 @@ items: import.meta.glob('./blog/*.{md,mdx}'),
 The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
 
 :::tip
-A package like [`sanitize-html`](https://www.npmjs.com/package/sanitize-html) will make sure that your content is properly sanitized, escaped, and encoded.
+A package like [`sanitize-html`](https://www.npmjs.com/package/sanitize-html) will make sure that your content is properly sanitized, escaped, and encoded. In the process, such a package might also remove some harmless elements and attributes, so make sure to verify the output and configure the package according to your needs.
```
mayank99, line 157: Could be useful to link to something like https://github.com/apostrophecms/sanitize-html#common-use-cases, but I didn't want to make this tip too verbose

sarah11918: @mayank99 Although not a direct link to the anchor, the npm readme does include the "common use cases" so someone can certainly get there just by reading further down the page.

Let's merge the tip as is for now, which I think sounds great!

sarah11918 (approved):
Thank you so much for picking this up, @mayank99 ! I think it's a great addition!

PR #6208, opened 2024-01-09
Transitions Tables to a Heading Layout for API Reference

### LINKED ISSUE #6201: Replace terrible tables with headings, bullets etc.
### 📚 Subject area/topic

API reference

https://docs.astro.build/en/reference/api-reference/
(check for more)

We really can't put up with these tables any longer!  :sweat_smile: 

image

Use other entries on this, and on the configuration reference page, for inspiration to display this content sensibly. (Headings are never bad. Listed as bullet points *may* work, but we'd need to see them in action.)

_No response_

SatanshuMishra:
#### Description (required)

This PR replaces the Tables on the API Reference page to use Headings instead. The changes maintain the same heading structure in the side-bar and bring the page to a similar layout to the rest of the page and the configuration reference page. Preview of changes attached below:

Side-by-side preview of changes.

- Closes #6201 

Special thanks to @VoxelMC for his help!


SatanshuMishra:
There is also another table present further down the page under the `The pagination page prop` header. Should I also transition that to a similar format? (Screenshot attached)

image

VoxelMC:
Thanks for starting to work on this, @SatanshuMishra! I would suggest making this a Draft PR for now, as it's not quite done yet. 

@sarah11918 for the sake of consistency, how do you feel about the way that these properties are documented?
image

I don't think they would work for the properties in `Astro.cookies`, because of how long the descriptions are. Whatever we come up with for `Astro.cookies`, should we apply it wherever else properties are documented, such as in `Astro.response`?

sarah11918:
So to answer the questions so far:

YES DESTROY ALL TABLES. I want no tables remaining.  :smile: 

As to the `Astro.response`, I'm tempted to leave it as is because it's so visually compact, and I think maybe more helpful when you can see it that way? Looking at the `cookies.set` option, for example, it feels sooooooo long and drawn out, for mostly short entries, I might prefer to see what this looks like in the bulleted form too.

I'm not sure stretching everything out into headings is visually helpful, even if structurally it makes sense. I'll ask some others to take a look here!

SatanshuMishra:
Sounds good! I have a meeting in an hour or so. I will make the changes after :)!

sarah11918:
Alright, so after some chatting, I say leave `Astro.response` as is, since in fact we used to just have a link out to a standard response object, but that content was removed so we added it in here to be helpful. (But, it's really not documenting an Astro thing; just saying that we conform to a standard).

As for the tables, let's keep going with the headings. It's the best structure, and so it does make sense to use it. If we need to do visual adjusting from that, we always can later.

SatanshuMishra:
The last commit changes the final Table into a heading based layout. The style is consistent with the previous changes made above. Preview of change is attached below:

image

VoxelMC:
> So to answer the questions so far:
> 
> YES DESTROY ALL TABLES. I want no tables remaining. 😄
> 
> As to the `Astro.response`, I'm tempted to leave it as is because it's so visually compact, and I think maybe more helpful when you can see it that way? Looking at the `cookies.set` option, for example, it feels sooooooo long and drawn out, for mostly short entries, I might prefer to see what this looks like in the bulleted form too.
> 
> I'm not sure stretching everything out into headings is visually helpful, even if structurally it makes sense. I'll ask some others to take a look here!

I definitely agree with this. Just wanted to make sure it aligned well with you and DX's standpoint.
I will now log this in my memory as maintainable formatting - and be on the lookout for inconsistencies as the self-proclaimed `Docs Inconsistency Sheriff` :)

SatanshuMishra:
@sarah11918, I've made all the changes you were looking for, and they're now available for you to review together :)

sarah11918:
Thank you @SatanshuMishra ! We were even talking about asking our designer if he had an opinion to share. I think structurally, this is the best format. Some other docs maintainers and I will review the specific changes to see what we think, but we VERY MUCH appreciate you starting this off and giving us something to look at!

doodlemarks:
@SatanshuMishra chiming in real quick.

I like this format better than the table! My only comment would be that the vertical rhythm needs some help.

image

As of now in the example above taken from the preview link all the spacing between lines is very similar if not identical. Where there are arrows I would shrink the space so each line of each section can show that they are connected. Also, helps with readability.

Let me know if I can provide more guidance!

VoxelMC:
> @SatanshuMishra chiming in real quick.
> 
> I like this format better than the table! My only comment would be that the vertical rhythm needs some help.
> 
> image
> As of now in the example above taken from the preview link all the spacing between lines is very similar if not identical. Where there are arrows I would shrink the space so each line of each section can show that they are connected. Also, helps with readability.
> 
> Let me know if I can provide more guidance!

The space between the `type` and the body text can probably be removed

sarah11918:
Thanks to Mark, our designer, for weighing in! We've discussed and decided that the scope of *this* PR should just be moving to headings... freeing the data from the tables and putting it reasonably out into headings. 

Yeah, spacing is a *little* wide (Chris thinks he found a specific docs issue) and so things aren't visually grouped as well as they could be, but Chris will work separately on that! All we need to do here is liberate the data!

I'll do a closer review of this tomorrow just to make sure, but right now, all we need to worry about is the info structure. We'll tackle styling separately!  🙌

SatanshuMishra:
Sounds good! Even though the focus for this PR is content, I am happy to maybe start a thread on discord or some form of discussion (maybe another issue) to start exploring different spacing options and styles with @doodlemarks and @VoxelMC.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -808,6 +921,88 @@ const { page } = Astro.props;
...
+
+##### `page.url.next`
+
+<p>
+**Type:** `string | undefined`
+</p>
+
+Get the URL of the next page (will be `undefined` if no more pages).
+
+{/*
```
VoxelMC, line 1004: You should probably remove this now!

VoxelMC (commented):
LGTM, but pending my comment :)

src/content/docs/en/reference/api-reference.mdx
SatanshuMishra: Thanks for catching this. It's been resolved in my new commit :)!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -808,18 +921,85 @@ const { page } = Astro.props;
...
+
+The total number of pages.
+
+##### `page.url.current`
+
+<p>
+**Type:** `string`
+</p>
+
+Get the URL of the current page (useful for canonical URLs)
```
sarah11918, line 986:
```suggestion
Get the URL of the current page (useful for canonical URLs).
```
Was missing in the original!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -808,18 +921,85 @@ const { page } = Astro.props;
...
+
+Array of data returned from `data()` for the current page.
+
+##### `page.start`
+
+<p>
+**Type:** `number`
+</p>
+
+Index of first item on current page, starting at `0` (e.g. if `pageSize: 25`, this would be `0` on page 1, `25` on page 2, etc.).
```
sarah11918, line 938:
```suggestion
Index of first item on current page, starting at `0`. (e.g. if `pageSize: 25`, this would be `0` on page 1, `25` on page 2, etc.)
```
Better punctuation than original!

sarah11918 (approved):
Looks great! Thank you so much for grabbing this issue, responding to feedback, and working collaboratively with us! This is an amazing first-time contribution, and we're so happy to welcome you to Team Docs!  :partying_face:

PR #6228, opened 2024-01-09
Add Paraglide to i18n community libraries

LorisSigrist:
#### Description (required)

This PR adds [Paraglide](https://inlang.com/c/astro) to the community libraries in the i18n recipie.

Paraglide plays really well with islands due to it's treeshakeable output. It only ships messages for an island if they are actually used, making it a great fit for translating interactive components and site shells. The resulting bundle can be as small as 120bytes, depending on the message and number of languages.

The Paraglide team is also committed to maintaining good Astro [guides](https://inlang.com/g/utqgkmzp/guide-lorissigrist-buildAGlobalAstroApp)  and [examples](https://github.com/opral/monorepo/tree/main/inlang/source-code/paraglide/paraglide-js-adapter-astro/example).

### DIFF
--- a/src/content/docs/en/recipes/i18n.mdx
+++ b/src/content/docs/en/recipes/i18n.mdx
@@ -505,3 +505,4 @@ Translate the routes of your pages for each language.
 - [astro-i18next](https://github.com/yassinedoghri/astro-i18next) — An Astro integration for i18next including some utility components.
 - [astro-i18n](https://github.com/alexandre-fernandez/astro-i18n) — A TypeScript-first internationalization library for Astro.
 - [astro-i18n-aut](https://github.com/jlarmstrongiv/astro-i18n-aut) — An Astro integration for i18n that supports the `defaultLocale` without page generation. The integration is adapter agnostic and UI framework agnostic.
+- [paraglide](https://inlang.com/c/astro) - A fully typesafe i18n library that plays really well with islands.
--- a/src/content/docs/en/recipes/i18n.mdx
+++ b/src/content/docs/en/recipes/i18n.mdx
@@ -505,4 +505,4 @@ Translate the routes of your pages for each language.
 - [astro-i18next](https://github.com/yassinedoghri/astro-i18next) — An Astro integration for i18next including some utility components.
 - [astro-i18n](https://github.com/alexandre-fernandez/astro-i18n) — A TypeScript-first internationalization library for Astro.
 - [astro-i18n-aut](https://github.com/jlarmstrongiv/astro-i18n-aut) — An Astro integration for i18n that supports the `defaultLocale` without page generation. The integration is adapter agnostic and UI framework agnostic.
-- [paraglide](https://inlang.com/c/astro) - A fully type-safe i18n library specifically designed for partial hydration patterns, like Astro islands.
+- [paraglide](https://inlang.com/c/astro) - A fully type-safe i18n library specifically designed for partial hydration patterns like Astro islands.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/recipes/i18n.mdx
```diff
@@ -505,3 +505,4 @@ Translate the routes of your pages for each language.
 - [astro-i18next](https://github.com/yassinedoghri/astro-i18next) — An Astro integration for i18next including some utility components.
 - [astro-i18n](https://github.com/alexandre-fernandez/astro-i18n) — A TypeScript-first internationalization library for Astro.
 - [astro-i18n-aut](https://github.com/jlarmstrongiv/astro-i18n-aut) — An Astro integration for i18n that supports the `defaultLocale` without page generation. The integration is adapter agnostic and UI framework agnostic.
+- [paraglide](https://inlang.com/c/astro) - A fully typesafe i18n library that plays really well with islands.
```
sarah11918, line 508:
```suggestion
- [paraglide](https://inlang.com/c/astro) - A fully type-safe i18n library that works with islands.
```
I checked, and it seems that "type-safe" is the correct adjective form (I might need to check the rest of our site, too!) and we really need a more literal, neutral phrase that will be easier on our translators than "plays well with." Maybe "works with"? Do you have a better idea here?

sarah11918 (commented):
Thank you for adding this helpful link @LorisSigrist ! See my comment below, and maybe you have a better idea for a more straight-ahead phrasing there?

src/content/docs/en/recipes/i18n.mdx
LorisSigrist: Good note on the "type-safe", thanks!

"plays well with" and "works with" don't convey the same meaning. "plays well with" implies a level of synergy, whereas "works with" only states that it does function. 

What do you think of the phrasing "A fully type-safe i18n library that's great for islands"?

sarah11918: if it were "designed for islands", that's a feature worth stating.

Is there a particular way in which it's great for islands? A particular noteworthy feature to call out? (Honestly, my initial reaction was to delete that entirely. Unless you can tell me something significant about it that means we *should* mention islands, it's not really worth calling out.)

LorisSigrist: Paraglide wasn't designed for Astros Islands in particular, but for any partial hydration pattern. 

With paraglide messages are treeshakeable. That way any island only adds exactly the messages that it needs to the JS bundle no more. Usually this would require developers to sort their messages into namespaces, but with Paraglide it's automatic.

LorisSigrist: That's fits really well, great suggestion!
I'll change it to that. Thank you

sarah11918 (approved):
Happy to have this added, and happy with where we landed on the description!  :raised_hands:

Now, I just need @Yan-Thomas to weigh in on that comma, and we'll be good to get this merged. Welcome to Team Docs! :partying_face:

src/content/docs/en/recipes/i18n.mdx
```diff
@@ -505,3 +505,4 @@ Translate the routes of your pages for each language.
 - [astro-i18next](https://github.com/yassinedoghri/astro-i18next) — An Astro integration for i18next including some utility components.
 - [astro-i18n](https://github.com/alexandre-fernandez/astro-i18n) — A TypeScript-first internationalization library for Astro.
 - [astro-i18n-aut](https://github.com/jlarmstrongiv/astro-i18n-aut) — An Astro integration for i18n that supports the `defaultLocale` without page generation. The integration is adapter agnostic and UI framework agnostic.
+- [paraglide](https://inlang.com/c/astro) - A fully type-safe i18n library specifically designed for partial hydration patterns, like Astro islands.
```
VoxelMC, line 508:
```suggestion
- [paraglide](https://inlang.com/c/astro) - A fully type-safe i18n library specifically designed for partial hydration patterns like Astro islands.
```
If I may weigh in here, I believe the comma here is a comma-splice!

VoxelMC (commented):
Here is my take on the comma, if I may!

dreyfus92 (approved):
Welcome to `astro-docs` @LorisSigrist, LGTM 😁

LorisSigrist:
Thank you all!

PR #6248, opened 2024-01-10
[api-reference]: Refactoring Markdown Table to `ApiTable` component

liruifengv:
#### Description (required)

Refactoring Markdown Table to `ApiTable` component

Since the original markdown table causes the text in the cell to collapse on newlines and is difficult to maintain.

And it is not possible to set the width, so I refactoring the Table component to use HTML table and specify the column width.

#### Before
image

#### After
image

### DIFF
--- a/src/components/ApiTable.astro
+++ b/src/components/ApiTable.astro
@@ -1,0 +1,23 @@
+---
+import { useTranslations } from '~/i18n/util';
+
+const t = useTranslations(Astro);
+---
+
+<table>
+	<colgroup>
+		<col width="15%">
+		<col width="35%">
+		<col width="50%">
+	</colgroup>
+	<thead>
+		<tr>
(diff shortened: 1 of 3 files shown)


liruifengv:
Maybe We need a better built-in table component for Starlight.

sarah11918:
Hi @liruifengv , thank you for recognizing that our tables are terrible! :sweat_smile: We actually have an open issue, and another open PR for this right now. We've decided to not use a table at all because we really can't guarantee a great experience on all screen sizes. In fact, when I view this PR's preview at a smaller screen size, I get this:

image

We think we're pretty close to a solution in #6208 , and Chris is already working on some tighter spacing to better visually group the entries. So, I'll close this one because we're not going to use a table, but you're welcome to join us in the other PR and include any feedback there! It will be so nice to be rid of these existing tables.  :smile:

PR #6265, opened 2024-01-11
feat(i18n): document new option

ematipico:
#### Description (required)

Closes PLT-1431

This PR adds documentation for the upcoming new option. I added a small chapter, I don't know if I need to repeat the examples.  Let me know what you think.

- Closes #
- Suggested label:


src/content/docs/en/guides/internationalization.mdx
```diff
@@ -152,6 +152,10 @@ Set this option when all routes will have their `/locale/` prefix in their URL a
 - URLs without a locale prefix, (e.g. `example.com/blog/`) will return a 404 (not found) status code.
 - The home URL (`/`) redirects to `/<defaultLocale>`.
 
+#### `redirectToDefaultLocale: true`
+
+When `prefixDefaultLocale` is `true`, the redirect form the home URL to `/<defaultLocale>` is disabled.
```
VoxelMC, line 157:
```suggestion
When `prefixDefaultLocale` is `true`, `redirectToDefaultLocale: false` will disable redirecting from the home URL to `/<defaultLocale>`. 
```

Some alternatives:
- When `prefixDefaultLocale` is `true`, `redirectToDefaultLocale` controls whether to redirect from the home URL to `/<defaultLocale>`. 
- When `prefixDefaultLocale` is `true` and `redirectToDefaultLocale` is `false`, redirecting from the home URL to `/<defaultLocale>` is disabled.

VoxelMC (commented):
Hello! Thank you for making sure this is documented!

I have added some potential alternative phrasing, should you desire to choose one. 
I hope this helps!

src/content/docs/en/guides/internationalization.mdx
ematipico: I like the first alternative!

sarah11918:
Editing now, but also adding a tiny bit to the section directly above to help with the issue where people are forgetting to make an index page when `prefixDefaultLocale` is set to true:

- highlights, and adds an inline comment for `index.astro`
- adds the file name in the second bullet point

image

sarah11918 (commented):
@ematipico - See what you think about these changes. Since I knew you were also adding an error message about a missing `index.astro` file, I figured we could help the existing documentation here too, at the same time.

Please check the additions below and edit for correctness as necessary. e.g. I invented a reason why someone might configure this as `false`, to give more context.   I think it's ok to expand this, but I'd like to make sure this is accurate/gives the right nuance!

ematipico:
Thank you, Sarah; I like very much the new phrasing!

sarah11918:
Note to self: ADD A LINK TO THE CONFIG REFERENCE ITEM

But otherwise, happy with the wording!

src/content/docs/en/guides/internationalization.mdx
```diff
@@ -150,7 +150,14 @@ Set this option when all routes will have their `/locale/` prefix in their URL a
...
 - URLs without a locale prefix, (e.g. `example.com/blog/`) will return a 404 (not found) status code.
-- The home URL (`/`) redirects to `/<defaultLocale>`.
+
+#### `redirectToDefaultLocale`
+
+Configures whether or not the home URL (`/`) generated by `src/pages/astro.index` will redirect to `/<defaultLocale>`.
+
+Setting `prefixDefaultLocale: true` will also automatically set `redirectToDefaultLocale: true` in your `routing` config object. By default, the required `src/pages/index.astro` file will automatically redirect to the index page of your default locale.
+
+You can opt out of this behavior by setting `redirectToDefaultLocale: false`. This allows you to have a site home page that exists outside of your configured locale folder structure.
```
sarah11918, line 160:
```suggestion
You can opt out of this behavior by [setting `redirectToDefaultLocale: false`](/en/reference/configuration-reference/#i18nroutingredirecttodefaultlocale). This allows you to have a site home page that exists outside of your configured locale folder structure.
```

sarah11918 (approved):
Link will fail until #6437 is merged.

PR #6316, opened 2024-01-12
Try removing imports from `'virtual:starlight/components'`

delucis:
#### Description (required)

Test PR that removes imports from Starlight’s virtual components module. Trying this to see if it is related to some errors @sarah11918 in particular has been experiencing.

Doing this has shifted CSS order, so I made a couple of hotfixes to CSS. We can revisit those to do them more correctly if this seems to work.

### DIFF
--- a/src/components/starlight/PageSidebar.astro
+++ b/src/components/starlight/PageSidebar.astro
@@ -1,7 +1,8 @@
 ---
 import type { Props } from '@astrojs/starlight/props';
+import MobileTableOfContents from './MobileTableOfContents.astro';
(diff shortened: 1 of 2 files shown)


sarah11918:
Less than a minute into typing text into an `.mdx` page:

image

This is the same error as before. But, this time it did not actually crash the dev server. So, progress!

sarah11918:
and... this time it actually did crash.
[devservercrash.txt](https://github.com/withastro/docs/files/13922889/devservercrash.txt)

delucis:
THIS IS NOT THE FIX YOU WERE LOOKING FOR

PR #6319, opened 2024-01-13
Update turso.mdx

alexnguyennz:
#### Description (required)

- fix imports (one level too deep referencing `src/turso.ts`)


sarah11918 (approved):
Thank you, Alex! :rocket:

PR #6386, opened 2024-01-14
Document new routing priority behavior

Fryuni:
#### Description

- Document the changes being implemented on withastro/astro#9439


src/content/docs/en/core-concepts/routing.mdx
```diff
@@ -293,32 +293,45 @@ if (!isLoggedIn(cookie)) {
...
-- Pre-rendered dynamic routes take precedence over server dynamic routes
-- Rest parameters have the lowest priority
-- Endpoints always take precedence over pages
-- Ties are resolved alphabetically
+- More specific rotues (with more path segments) will take precedence over less specific routes. In the example above all routes under `/posts/` take precedence over `/[...slug].astro` at the root.
+- Static routes without path parameters will take precedence over dynamic routes. `/posts/create.astro` takes precedence over all the other routes in the example.
+- Dynamic routes using named parameters take precedence over rest parameters. `/posts/[page].astro` takes precedence over `/posts/[...slug].astro`.
+- Pre-rendered dynamic routes take precedence over server dynamic routes.
+- Endpoints take precedence over pages.
+- If all the above doesn't decide the order, routes are sorted alphabetically based on the default locale of your Node instalation.
```
VoxelMC, line 315:
```suggestion
- Static routes without path parameters will take precedence over dynamic routes. E.g. `/posts/create.astro` takes precedence over all the other routes in the example.
- Dynamic routes using named parameters take precedence over rest parameters. E.g. `/posts/[page].astro` takes precedence over `/posts/[...slug].astro`.
- Pre-rendered dynamic routes take precedence over server dynamic routes.
- Endpoints take precedence over pages.
- If none of the rules above decide the order, routes are sorted alphabetically based on the default locale of your Node installation.
```

Some typo fixes here, and perhaps add "E.g." to the examples, or something along the lines of that?

VoxelMC (commented):
Nice work Fryuni! Thank you for documenting all of these rules!

I have left some typo fixes and suggestions for the `routing.mdx` file, but not the reference files because I believe those are auto-generated from the JSDocs. 

Hope they help!

sarah11918:
I am noting that this feature will be released as an experimental flag for 4.2.0, which means that the ONLY documentation should be in the changeset and `types.ts` file under "experimental flags" for configuration reference.

This PR should not be merged as there will be no wider docs changes (as there never are for experimental flag features changing behaviour. All documentation is contained to the experimental config section). BUT, we can use this to help craft that!

We will use https://github.com/withastro/astro/pull/9644 as a model/reference for how to document for an experimental flag that builds on/changes existing features.

sarah11918 (dismissed):
This is not to be merged, used in helping us prepare documentation in https://github.com/withastro/astro/pull/9439 only.

Fryuni:
Lovely, thanks for sending the changes @sarah11918. You're the best!

PR #6425, opened 2024-01-16
Fix: missing interface assign in code example

poooow:
#### Description (required)

Code doesn't work without assigning props variables to interface

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -147,7 +147,7 @@ interface Props {
   name: string;
   greeting?: string;
 }
-const { greeting = 'Hello', name } = Astro.props;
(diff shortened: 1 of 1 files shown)


at-the-vr (commented):
Props do actually work without further assigning
Check this [Stackblitz Instance](https://stackblitz.com/edit/github-scpmmf?file=src%2Fpages%2Findex.astro) for verifying. I don't think there's a need to be verbose here

sarah11918:
Thanks for the StackBlitz Atharva! 

Yes, this code should work as written in the documentation, so I'll close this now. But if you're having trouble getting this to work without assigning, you can ask for assistance figuring out what might be happening in your project in our Discord! Maybe there's a bug you've encountered, or a configuration setting affecting your project. People would be happy to figure out what's going on there!  https://astro.build/chat

PR #6431, opened 2024-01-16
typo: correct .mdoc extension to .astro

schalkneethling:
It seems there is a typo suggesting that `.mdoc` files can be used inside the `pages` folder. After speaking with nate on Discord, it seems it should be `.astro`.

The makes the relevant change.


sarah11918 (approved):
Thank you so much for taking the time to fix this! We appreciate your help, and are delighted to welcome you to Team Docs!  🥳

PR #6432, opened 2024-01-16
typo: fix a small typo in the welcome bot message

schalkneethling:
The welcome bot states that a preview will be available on Netlify, but it seems deploys are now done to Vercel.


sarah11918 (approved):
Indeed! A recent change we neglected to update. Thank you! 🙌

PR #6440, opened 2024-01-17
Update mention about dynamically generated slot names

MoustaphaDev:
#### Description

In https://github.com/withastro/compiler/pull/933, we added support for dynamically generated slots. This PR updates a reference about that feature not being supported and underlines the minimum astro version required to use it

#### For Astro version: `4.2.0`.

See astro PR https://github.com/withastro/astro/pull/9605

### DIFF
--- a/src/content/docs/en/core-concepts/astro-components.mdx
+++ b/src/content/docs/en/core-concepts/astro-components.mdx
@@ -291,7 +291,7 @@ Named slots can also be passed to [UI framework components](/en/core-concepts/fr
 
 
 :::note
-An astro slot name can not be dynamically generated, such as within a map function. If this feature is needed within UI framework components, it might be best to generate these dynamic slots within the framework itself.
+To dynamically generate an astro slot name, such as within a map function, you need use Astro versions later than 4.2.0. If this feature is needed within UI framework components on older Astro versions, it might be best to generate these dynamic slots within the framework itself.
 :::
 
 


dreyfus92 (approved):
LGTM ✅

src/content/docs/en/core-concepts/astro-components.mdx
```diff
@@ -291,7 +291,7 @@ Named slots can also be passed to [UI framework components](/en/core-concepts/fr
 
 
 :::note
-An astro slot name can not be dynamically generated, such as within a map function. If this feature is needed within UI framework components, it might be best to generate these dynamic slots within the framework itself.
+To dynamically generate an astro slot name, such as within a map function, you need use Astro versions later than 4.2.0. If this feature is needed within UI framework components on older Astro versions, it might be best to generate these dynamic slots within the framework itself.
```
sarah11918, line 294:
```suggestion
Dynamically generating an astro slot name, such as within a map function, was added in Astro v4.2.0. For earlier versions of Astro, it might be best to generate these dynamic slots within the framework itself.
```

sarah11918: Typically, we write docs for the current version, so ideally I wouldn't want this as a note at all. If this is possible now, I think an example of usage would be fabulous, and the better way to go in addition to a note like the above.

BUT, for a super quick fix, I would just amend the note this way, and am happy to approve this.

sarah11918 (commented):
@MoustaphaDev Let me know if you're OK with the proposed changes below, or slighty fix if necessary.

If it's for tomorrow's minor release, I'll take responsibility for merging it, along with the other docs!

src/content/docs/en/core-concepts/astro-components.mdx
MoustaphaDev: Sounds good, I'll add an example usage of the feature on top of this note.

MoustaphaDev: Yeah it should I'll edit that @VoxelMC, thanks!

src/content/docs/en/core-concepts/astro-components.mdx
```diff
@@ -291,7 +291,7 @@ Named slots can also be passed to [UI framework components](/en/core-concepts/fr
 
 
 :::note
-An astro slot name can not be dynamically generated, such as within a map function. If this feature is needed within UI framework components, it might be best to generate these dynamic slots within the framework itself.
+Dynamically generating an astro slot name, such as within a map function, was added in Astro v4.2.0. For earlier versions of Astro, it might be best to generate these dynamic slots within the framework itself.
```
MoustaphaDev, line 294:
```suggestion
Dynamically generating an Astro slot name, such as within a map function, was added in Astro v4.2.0. For earlier versions of Astro, it might be best to generate these dynamic slots within the framework itself.
```

MoustaphaDev:
I said I would add an example, but I think it's okay to keep it as-is, for now at least.
I need to get another Astro PR merged in order to make a good example (using the feature introduced in that PR). I'll add an example in a follow-up docs PR if that's okay for you!

sarah11918:
Sounds perfect, thank you @MoustaphaDev ! NWTWWHB!

PR #6505, opened 2024-01-20
Fix a confusing React component example

Kirshach:
#### Description (required)

I believe the navigation triggering React component has a just very slightly confusing name and an unused prop which somewhat misguide the reader's attention

- Suggested label: Fix a confusing React component example

### DIFF
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -309,7 +309,7 @@ The following example implements the same using `navigate()` in a React `<Form /
 import React from 'react';
 import { navigate } from 'astro:transitions/client';
 
-export default function ClickToNavigate({ to }) {
+export default function Form() {
   return <select onChange={(ev) => navigate(e.target.value)}>
     <option value="/play">Play</option>
     <option value="/blog">Blog</option>


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -309,7 +309,7 @@ The following example implements the same using `navigate()` in a React `<Form /
 import React from 'react';
 import { navigate } from 'astro:transitions/client';
 
-export default function ClickToNavigate({ to }) {
+export default function Form() {
```
at-the-vr, line 312: Missing `to` prop ⬇️ 
```suggestion
export default function Form({ to }) {
```

at-the-vr (commented):
Hi there @Kirshach `ClickToNavigate` suggests readers that this function is meant to well "navigate to different page on click". Changing its name to plain `Form` semantically raises questions eg: what's Form supposed to do ? any action related with it (submit, fetch) ? 

But importantly at [Line 325](https://github.com/withastro/docs/blob/main/src/content/docs/en/guides/view-transitions.mdx#L325) there's the same thing your thinking off

That being said I only have 1 suggestion

src/content/docs/en/guides/view-transitions.mdx
Kirshach: Hi! Thank you for your reply!
Why would we include a "to" prop and not use it? 🤔

at-the-vr: your right, `to` can be removed 😄 , suggestion cancelled

sarah11918 (approved):
Agreed! This is a great change. Thank you for taking the time to submit! We really appreciate it, and are happy to welcome you to Team Docs!

PR #6519, opened 2024-01-21
Fix the logic error in docs [Build a blog 5:3]

md-abid-hussain:
#### Description (required)
There was logic error in **Build a blog** tutorial at **Unit 5: Astro api** 3rd section
Screenshot 2024-01-21 114828
instead of `{tags.map ...}` it should be `{frontmatter.tags.map....}`


sarah11918:
Nicely spotted, you saw this before it was even pointed out! This is not a mistake. Immediately under that code snippet, you will find:

image

PR #6523, opened 2024-01-21
Fix port typo compared to the source.

santiemanuel:
#### Description (required)

Fixed the default port Astro uses, not the port that's used if `4321` is not available.

- Closes #
- Suggested label:


Yan-Thomas (approved):
Nice catch, thanks @santiemanuel!

PR #6362, opened 2024-01-13
i18n(de): update `editor-setup.mdx`

DerTimonius:
#### Description (required)

This PR will update the german translation for the `editor-setup` documentation to match the english version. I also made a small change to the english version, because Fleet has been released some time ago.

PR #6532, opened 2024-01-22
i18n(en): update editor-setup

DerTimonius:
#### Description (required)

Small change to the english version of the `editor-setup.mdx` file, related to the update of the german translation in  #6362


src/content/docs/en/editor-setup.mdx
```diff
@@ -28,7 +28,7 @@ import ReadMore from '~/components/ReadMore.astro';
 
 Initial support for Astro landed in WebStorm 2023.1. You can install the official plugin through [JetBrains Marketplace](https://plugins.jetbrains.com/plugin/20959-astro) or by searching for "Astro" in the IDE's Plugins tab. This plugin includes features like syntax highlighting, code completion, and formatting, and plans to add even more advanced features in the future. It is also available to all other [JetBrains IDEs with JavaScript support](https://www.jetbrains.com/products/#lang=js&type=ide).
 
-JetBrains' upcoming [Fleet IDE](https://www.jetbrains.com/fleet/) includes language server support and will be able to use the official Astro tooling.
+JetBrains' [Fleet IDE](https://www.jetbrains.com/fleet/) includes language server support and uses the official Astro tooling.
```
Princesseuh, line 31: This is not really true, Fleet still requires a plugin (that doesn't exist). It's just intended to be compatible with the underlying technology we use and so developing a plugin is *possible*

Though, I'm not even sure if it has its LSP support released yet.

sarah11918: OK this is causing me way too much grief. FLEET IS DEAD TO ME.

sarah11918, line 31:
```suggestion
```
Suggest deleting this line entirely, since it's not clear exactly how Fleet does or will work. If they want to be in our docs, they can PR it in.

sarah11918 (commented):
Thanks @DerTimonius ! Given Erika's comment, let's remove this entirely, and you can update the German to also delete.

src/content/docs/en/editor-setup.mdx
DerTimonius: I'm all for getting out the pitchforks 😄

Yan-Thomas (approved):
LGTM!

PR #6534, opened 2024-01-22
Update alpinejs.mdx

florian-lefebvre:
#### Description (required)

Docs for https://github.com/withastro/astro/pull/9751, mostly copied from the vue docs

- Suggested label: `minor-release`

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
+++ b/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
@@ -98,7 +98,7 @@ Check our [Astro Integration Documentation][astro-integration] for more on integ
 
 Note that you can use the CDN builds of plugins without any issues, just add them to the `head`:
 
-```diff lang="astro" title="src/layouts/Layout.astro"
+```diff lang="astro" title="src/pages/index.astro"
 <html>
 	<head>
 		<!-- ... -->
--- a/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
+++ b/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
@@ -96,31 +96,35 @@ Once the integration is installed, you can use [Alpine.js](https://alpinejs.dev/
 
 Check our [Astro Integration Documentation][astro-integration] for more on integrations.
 
-## Limitations
+## Options
 
-The Alpine.js integration does not give you control over how the script is loaded or initialized. If you require this control, consider [installing and using Alpine.js manually](https://alpinejs.dev/essentials/installation). Astro supports all officially documented Alpine.js manual setup instructions, using `<script>` tags inside of an Astro component.
+### entrypoint
 
-**It is not currently possible to [extend Alpine.js](https://alpinejs.dev/advanced/extending) when using this component.** If you need this feature, consider following [the manual Alpine.js setup](https://alpinejs.dev/essentials/installation) instead using an Astro script tag:
+You can extend Alpine by setting the `entrypoint` option to a root-relative import specifier (for example, `entrypoint: "/src/entrypoint"`).
 
-```astro title="src/pages/index.astro"
----
----
+The default export of this file should be a function that accepts an Alpine instance prior to starting, allowing the use of custom directives, plugins and other customizations for advanced use cases.
 
-<!-- Example: Load AlpineJS on a single page. -->
-<script>
-  import Alpine from 'alpinejs';
+```js
+// astro.config.mjs
+import { defineConfig } from 'astro/config';
+import alpine from '@astrojs/alpinejs';
 
-  // Optional: Extend Alpine.js
-  // Alpine.directive('foo', ...)
-
-  window.Alpine = Alpine;
-  Alpine.start();
-</script>
+export default defineConfig({
+  // ...
+  integrations: [alpine({ entrypoint: '/src/entrypoint' })],
+});
 ```
 
-## Configuration
+```js
+// src/entrypoint.ts
+import type { Alpine } from 'alpinejs'
 
-The Alpine.js integration does not support any custom configuration at this time.
+export default (Alpine: Alpine) => {
+    Alpine.directive('foo', el => {
+        el.textContent = 'bar';
+    })
+}
+```
 
 ## Examples
 
--- a/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
+++ b/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
@@ -118,11 +118,10 @@ export default defineConfig({
 ```js
 // src/entrypoint.ts
 import type { Alpine } from 'alpinejs'
+import intersect from '@alpinejs/intersect'
 
 export default (Alpine: Alpine) => {
-    Alpine.directive('foo', el => {
-        el.textContent = 'bar';
-    })
+    Alpine.plugin(intersect)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/alpinejs.mdx
```diff
@@ -96,31 +96,45 @@ Once the integration is installed, you can use [Alpine.js](https://alpinejs.dev/
 
 Check our [Astro Integration Documentation][astro-integration] for more on integrations.
 
-## Limitations
+## Options
 
-The Alpine.js integration does not give you control over how the script is loaded or initialized. If you require this control, consider [installing and using Alpine.js manually](https://alpinejs.dev/essentials/installation). Astro supports all officially documented Alpine.js manual setup instructions, using `<script>` tags inside of an Astro component.
+### entrypoint
```
dreyfus92, line 101:
```suggestion
### `entrypoint`
```

dreyfus92 (commented):
Just a tiny suggestion @florian-lefebvre, awesome job 😁

src/content/docs/en/guides/integrations-guide/alpinejs.mdx
florian-lefebvre: I checked and it looks like no integration docs is using backticks in headlines. Should we stay consistent?

dreyfus92: There are a couple of examples, here you have one. https://github.com/withastro/docs/blob/a5bc295a807b1493b34d68d19f26eda18b6b165a/src/content/docs/en/guides/integrations-guide/mdx.mdx#L136
I was thinking as well about adding the the type and the def type like in that example, let me know what you think.

florian-lefebvre: I don't mind tbh 👍

hippotastic (commented):
Thank you for this docs addition! As you added this new feature to address common Alpine developer issues with extending Alpine as we discussed on Discord, it might be useful to include an example that shows how to add a popular Alpine plugin.

In the Discord thread, developers using Alpine mentioned [`@alpinejs/intersect`](https://alpinejs.dev/plugins/intersect) and [`@alpinejs/persist`](https://alpinejs.dev/plugins/persist), which might serve as an example.

Also, developers mentioned that they tried installing a plugin by adding `<script defer src="https://unpkg.com/@alpinejs/intersect@3.x.x/dist/cdn.min.js"></script>`, which seems to be a common method of installing Alpine plugins on other sites. It would be great to validate if this method works reliably with your new feature, and if not, document this and recommend the alternative instead (installing the package as a dependency).

florian-lefebvre:
I have updated the entrypoint example to use the the intersect plugin as a npm package. As for the the script import, I have no idea if it works. I'm first going to test it

src/content/docs/en/guides/integrations-guide/alpinejs.mdx
```diff
@@ -96,31 +96,52 @@ Once the integration is installed, you can use [Alpine.js](https://alpinejs.dev/
 
 Check our [Astro Integration Documentation][astro-integration] for more on integrations.
 
-## Limitations
+Note that you can use the CDN builds of plugins without any issues, just add them to the `head`:
 
-The Alpine.js integration does not give you control over how the script is loaded or initialized. If you require this control, consider [installing and using Alpine.js manually](https://alpinejs.dev/essentials/installation). Astro supports all officially documented Alpine.js manual setup instructions, using `<script>` tags inside of an Astro component.
+```diff lang="html"
+<head>
```
hippotastic, line 102: Sorry for being so insistent on this, but I have to ask: Were you really able to successfully test this? :) From what I've read, this was one of the issues that users were struggling with in the Discord support thread: the Alpine plugins loaded via deferred script tags were not guaranteed to have finished loading before the Alpine integration would initialize Alpine, and were therefore not reliably picked up by Alpine.

Did you change anything in the code that makes this possible now?

If it's possible now, awesome! If so, I'd love to see where this `<head>` tag should be placed by adding more context, e.g. a file name tab and the surrounding code - users should be able to know where they should add this so that the plugin is available on all pages of their site.

florian-lefebvre: Sure no problem! Have a look at this e2e test https://github.com/withastro/astro/pull/9812, I think that's exactly the scenario you're mentioning. It seems to work without my recent changes

hippotastic: It's awesome that you have actually added an E2E test! However, is this test actually effective in checking if the collapse plugin is working? From what I see, it's checking whether `<p id="foo" x-show="expanded" x-collapse>` has a class of `hidden`, and if not, it assumes the test has passed. But is this not the case even when the plugin is not loaded at all?

florian-lefebvre: I manually checked and yes, the collapse plugin adds a `hidden` attribute so that's a safe check

hippotastic: Ok, thank you for confirming. I also checked it manually now, even while simulating a slow connection using devtools, and the plugin always worked. I wonder why it didn't for the users reporting issues on Discord.

So, the only thing that's left is providing a code block file name and a little bit of context in your example so that users reading it will know immediately where to put this.

florian-lefebvre: Tell me if that looks good enough to you!

src/content/docs/en/guides/integrations-guide/alpinejs.mdx
```diff
@@ -96,31 +96,65 @@ Once the integration is installed, you can use [Alpine.js](https://alpinejs.dev/
 
 Check our [Astro Integration Documentation][astro-integration] for more on integrations.
 
-## Limitations
+Note that you can use the CDN builds of plugins without any issues, just add them to the `head`:
+
+```diff lang="astro" title="src/layouts/Layout.astro"
```
hippotastic, line 101:
```suggestion
```diff lang="astro" title="src/pages/index.astro"
```

As the code example (which is great) doesn't seem to show a real layout, but a page without any slots, I'd suggest changing the tab title accordingly. Alternatively, you could also leave this as a layout, include a `<slot />`, and add a second code block below that shows a page importing this layout and using the collapse plugin. Your call!

src/content/docs/en/guides/integrations-guide/alpinejs.mdx
hippotastic: A lot better, thank you! I just left a comment regarding file naming (layout vs. page). It was actually already helpful that we talked about this because I didn't even have a layout in mind, but that's actually a great way to add the script to each page head!

hippotastic (approved):
LGTM!

dreyfus92 (approved):
Such an amazing job @florian-lefebvre, thank you as well @hippotastic for deep diving into this, LGTM 😁

sarah11918 (approved):
Thank you Florian! I took an editing pass and mostly: moved config up before usage, since you will likely configure before using; Expressive Code'd the code sample, and just tightened up some wording!  Since the feature is released, I'll merge this now, and you are welcome to PR changes/corrections if I messed it up too badly.  ;)

PR #6545, opened 2024-01-23
Clarify route precedence in routing.mdx

mb21:
#### Description (required)

- Clarify what "them" and "in order" exactly refer to.
- "Routes with ... will take precedence over less specific routes" -> seems like a circular definition, since "precedence" and "specificity" seem to refer to the same thing here...


sarah11918:
Docs is happy with the existing wording here!

mb21:
Any rationale? Maybe you wanna take a look at how many people are confused about this, e.g. in https://github.com/withastro/astro/issues/9757

PR #6546, opened 2024-01-23
Netlify Adapter: Add callout on 404 behaviour

Skn0tt:
#### Description (required)

As discussed in https://github.com/withastro/adapters/issues/133.

- Closes #
- Suggested label:


sarah11918:
Hi @Skn0tt ! I chatted with the team further about this, and we don't think anything added to docs here is needed at this time! We appreciate you adding this to cover the bases, but let's wait until there's evidence that this is causing a widespread problem with a substantial number of users before calling attention to it, especially since there's active feature development being done around 404 pages.

So, you're off the hook here! :raised_hands:

PR #6571, opened 2024-01-24
Update 1.mdx

alexnguyennz:
#### Description (required)

- add package manager tabs for installing Preact
- remove .jsx extension like the following two examples without .jsx with `Greeting` import
- add space in component after client directive


sarah11918 (approved):
Looks great!

PR #6653, opened 2024-01-26
Update install-manual.mdx

liruifengv:
#### Description (required)

Update install-manual.mdx

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/install/manual.mdx
+++ b/src/content/docs/en/install/manual.mdx
@@ -12,7 +12,9 @@ This guide will walk you through the steps to manually install and configure a n
 
 #### Prefer a quicker way to get started?
 
+<p>
 <Button href="/en/install/auto/">Try the create astro CLI wizard →</Button>
+</p>
 
 #### Prerequisites
 
--- a/src/content/docs/en/install/manual.mdx
+++ b/src/content/docs/en/install/manual.mdx
@@ -10,11 +10,9 @@ import ReadMore from '~/components/ReadMore.astro'
 
 This guide will walk you through the steps to manually install and configure a new Astro project if you prefer not to use [the automatic CLI tool](/en/install/auto/).
 
-#### Prefer a quicker way to get started?
-
-<p>
-<Button href="/en/install/auto/">Try the create astro CLI wizard →</Button>
-</p>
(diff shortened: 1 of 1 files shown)


at-the-vr (commented):
will it be a good thing to specify `max-width` so that the button is not stretched out
| Instance | Button Look |
| -------- | ------- |
| Old Docs | image |
| New Docs | image |

src/content/docs/en/install/manual.mdx
```diff
@@ -12,7 +12,9 @@ This guide will walk you through the steps to manually install and configure a n
 
 #### Prefer a quicker way to get started?
 
+<p>
 <Button href="/en/install/auto/">Try the create astro CLI wizard →</Button>
+</p>
```
delucis, line 17: Thanks for spotting this @liruifengv! I wonder if actually the button treatment can be dropped now in favour of something more like what we have on https://docs.astro.build/en/install/auto/ ?

```suggestion
:::tip[Prefer a quicker way to get started?]
Follow the [create astro CLI wizard](/en/install/auto/) instead.
:::
```

liruifengv: It's great!

sarah11918:
Just chiming in that we probably want to rework some text here:

Do this if you prefer not to use the CLI tool.
Prefer to use the CLI tool???

😆 

image

src/content/docs/en/install/manual.mdx
```diff
@@ -10,9 +10,9 @@ import ReadMore from '~/components/ReadMore.astro'
 
 This guide will walk you through the steps to manually install and configure a new Astro project if you prefer not to use [the automatic CLI tool](/en/install/auto/).
```
sarah11918, line 11:
```suggestion
This guide will walk you through the steps to manually install and configure a new Astro project.
```
Let the tip do the talking?

sarah11918 (approved):
Looks great! Thank you for making this way better. 💜

sarah11918:
This one needs:

Co-authored-by: Chris Swithinbank <357379+delucis@users.noreply.github.com>
Co-authored-by: Atharva <88548999+at-the-vr@users.noreply.github.com>

PR #6654, opened 2024-01-26
Update the MultipleChoice submit button style

liruifengv:
#### Description (required)

Update the MultipleChoice submit button style when hover

before

image

after

image

- Closes #
- Suggested label:

### DIFF
--- a/src/components/tutorial/MultipleChoice.astro
+++ b/src/components/tutorial/MultipleChoice.astro
@@ -65,7 +65,9 @@ const t = useTranslations(Astro);
 	}
 
 	.submit:hover {
-		background-color: var(--sl-color-gray-1);
+		background: var(--sl-color-text-accent);
+		color: var(--sl-color-black);
+		border-color: transparent;
 	}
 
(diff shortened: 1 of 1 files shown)


at-the-vr (commented):
Nice catch 💯 for Light theme
Where can I find these color shades in Docs Repo by the way ? I could not find their definition in any css file 😕

liruifengv:
> Nice catch 💯 for Light theme
> 
> Where can I find these color shades in Docs Repo by the way ? I could not find their definition in any css file 😕 

It's in Starlight https://github.com/withastro/starlight/blob/main/packages/starlight/style/props.css

delucis (approved):
Thanks for the fix @liruifengv! 🙌 

If anyone’s interested, it looks like there are quite a few styling improvements needed to this component. Here is a design reference in case someone would like to have a go at making this match more closely: https://www.figma.com/file/GrDkcguAR7tSWPFuLVDixq/Theme---Starlight-Docs?node-id=454%3A34245&mode=dev

PR #6664, opened 2024-01-27
Upgrade dependencies (and downgrade Sharp)

delucis:
#### Description (required)

- This PR updates Astro and Starlight to the latest releases
- It also downgrades `sharp` back to 0.32.6 as the newest versions have been causing “Could not find Sharp.” errors


sarah11918 (approved):
No objections!

PR #6666, opened 2024-01-27
Fix headings spacing

delucis:
#### Description (required)

Because we wrap our heading elements in a `<div>` to support the layout of the anchor link icons, Starlight’s built-in `:not(heading) + heading` styles weren’t applying.

This PR adds back those spacing styles for less ugly pages 💅 

Click around several pages to see the difference, but here is one before/after comparison:

| Before | After |
|--------|--------|
| Page with minimal spacing before headings | Page with more generous spacing before headings |


dreyfus92 (approved):
LGTM 🚢

sarah11918 (approved):
Looks much better! Thank you Chris!

PR #6695, opened 2024-01-29
i18n(zh-cn): Update translate words “you”

huyikai:
#### Description (required)

Update translate words “you”
Because in the Chinese word convention, “you” suggest using "你" instead of "您"

- Closes #
- Suggested label:


liruifengv (approved):
LGTM. nice catch.

dreyfus92 (approved):
Thank you both, LGTM 😁

PR #6699, opened 2024-01-29
Add first-timer message to PULL_REQUEST_TEMPLATE.md

sarah11918:
#### Description (required)

This might help when trying to find first-time contributors on Discord for their first-time contributor alert!


dreyfus92 (approved):
Amazing idea, loved it 😁

ElianCodes (approved):
This is actually amazing!

TheOtterlord (approved):
Great idea!

delucis (commented):
Love this idea!

Only tiny nit pick is that this format requires people to remember to delete a new heading in the template each time they open a PR.

Perhaps could be *only* a comment in the description section?

VoxelMC (approved):
Awesome, no more missing people!
I agree with Chris about the deleting, perhaps it is worth commenting out the heading as well, and add an "uncomment this" like the above.

Either way, it's great!

sarah11918:
Thanks all! We'll try out this new version and see how we like it!

PR #6702, opened 2024-01-29
feat(types): add `ComponentProps` reference to TypeScript page

natemoo-re:
#### Description (required)

Adds reference for the new `ComponentProps` type added in Astro 4.3. This type allows users to reference the `Props` accepted by another component, even if that component doesn't export that `Props` type directly.

#### For Astro version: `4.3`. See astro PR [#9839](https://github.com/withastro/astro/pull/9839).


src/content/docs/en/guides/typescript.mdx
```diff
@@ -214,6 +214,22 @@ type MyAttributes = astroHTML.JSX.ImgHTMLAttributes;
 ```
 :::
 
+### Component Props type
+
+<p><Since v="4.3.0" /></p>
+
+In order to reference the type of `Props` accepted by another component, Astro exposes the `ComponentProps` utility from `astro/types`.
+
+```astro
```
sarah11918, line 223:
``````suggestion
### `ComponentProps` type

<p><Since v="4.3.0" /></p>

This type export allows users to reference the `Props` accepted by another component, even if that component doesn't export that `Props` type directly.

The following example shows using the `ComponentProps` utility from `astro/types` to reference a `<Button />` component's `Props` types:

```astro title="src/pages/index.astro"
``````
Would something like this be correct, @natemoo-re ? Please fine-tune as necessary. I really liked what you wrote in the PR description as an intro!

Comment (natemoo-re) [reply src/content/docs/en/guides/typescript.mdx]: Love this suggestion as-is, thanks!

File: src/content/docs/en/guides/typescript.mdx [hunk 2]
```diff
@@ -214,6 +214,24 @@ type MyAttributes = astroHTML.JSX.ImgHTMLAttributes;
 ```
 :::
 
+### `ComponentProps` type
+
+<p><Since v="4.3.0" /></p>
+
+This type export allows users to reference the `Props` accepted by another component, even if that component doesn't export that `Props` type directly.
```
sarah11918, line 221:
```suggestion
This type export allows you to reference the `Props` accepted by another component, even if that component doesn't export that `Props` type directly.
```

sarah11918 (approved):
Ready for 4.3! 🫡

PR #6711, opened 2024-01-30
i18n(zh-cn): Update translate words “other”

huyikai:
#### Description (required)

Update translate words “other”
Because in the Chinese word convention, “other” suggest using "其他" instead of "其它"

- Closes #
- Suggested label:


liruifengv (approved):
LGTM. Nice catch!

Yan-Thomas (approved):
LGTM!

PR #6718, opened 2024-01-30

Nin3lee:
#### Description (required)

the type of { frontmatter.pubDate } is a "date", but not a "string"! It needs to be transform to the "string"


sarah11918 (approved):
Thank you for this helpful fix! We apprecate you taking the time to contribute! :rocket:

PR #6734, opened 2024-02-01
i18n(zh-cn): Update strapi.mdx

webvs2:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/zh-cn/guides/cms/strapi.mdx
+++ b/src/content/docs/zh-cn/guides/cms/strapi.mdx
@@ -26,9 +26,14 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 ### 在 `.env` 文件中添加 Strapi URL
 
 为了将 Strapi URL 添加到 Astro 中，你可以在项目的根目录中创建一个 `.env` 文件（如果还没有的话），并添加以下变量：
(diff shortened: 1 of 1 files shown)


webvs2:
This is a problem I encountered when I was learning how to use strapi cms, which caused me to have doubts for a long time, and then I found out the solution and updated it

Yan-Thomas:
> This is a problem I encountered when I was learning how to use strapi cms, which caused me to have doubts for a long time, and then I found out the solution and updated it

Hey @webvs2, thanks for this! I'll note that we don't accept content changes in other languages, but we do appreciate the intention to improve our docs: would you be able to make a PR changing the English version, so that we can review it and then allow this change to roll out for all languages, not only Simplified Chinese?

webvs2:
Sorry, I will change and close that PR

webvs2:
> > This is a problem I encountered when I was learning how to use strapi cms, which caused me to have doubts for a long time, and then I found out the solution and updated it
> 
> Hey @webvs2, thanks for this! I'll note that we don't accept content changes in other languages, but we do appreciate the intention to improve our docs: would you be able to make a PR changing the English version, so that we can review it and then allow this change to roll out for all languages, not only Simplified Chinese?

@Yan-Thomas examine  #6762

PR #6740, opened 2024-02-01
chore(i18n): add one more example for APIs

ematipico:
#### Description (required)

The APIs `getRelativeLocaleUrl` and `getAbsoluteLocaleUrl` accept three parameters, but only the first parameter is mandatory. 

I added a new example that shows that. 

I received an issue where a user tried to use `getRelativeLocaleUrl("es", "/")` because they wanted to have the home URL of the locale.

Core repository https://github.com/withastro/astro/issues/9919


VoxelMC (approved):
It makes sense to me!
LGTM

sarah11918 (approved):
Thanks! Just getting around to this now, but looks great!

PR #6762, opened 2024-02-02
Update strapi.mdx

webvs2:

- Closes #
- Suggested label:


src/content/docs/en/guides/cms/strapi.mdx
```diff
@@ -34,7 +34,9 @@ STRAPI_URL="http://127.0.0.1:1337" // or use your IP address
 Now, you can use this environment variable in your Astro project.
```
sarah11918, line 34:
```suggestion
Restart the dev server to use this environment variable in your Astro project.
```

src/content/docs/en/guides/cms/strapi.mdx
```diff
@@ -34,7 +34,9 @@ STRAPI_URL="http://127.0.0.1:1337" // or use your IP address
...
 
 If you would like to have IntelliSense for your environment variable, you can create a `env.d.ts` file in the `src/` directory and configure `ImportMetaEnv` like this:
-
+:::note
+.env files are loaded at the beginning of Vite's startup and changes take effect after restarting the server.
+:::
```
sarah11918, line 39:
```suggestion
```

sarah11918 (commented):
Hi @webvs2 ! Thank you very much for sharing this useful information in our docs! I'm sure it will help many people.

We try to not use notes very often, though, and only for important or unusual pieces of information that require the visual interruption when someone is reading.

For a normal part of the process, we would instead just write the instruction as normal text.

So, I've made a change here to the text in the line above that I think will serve the same purpose, and actually tell people to do this exactly when they need to.

Thank you again for this contribution, and welcome to Team Docs!  :partying_face:

PR #6763, opened 2024-02-02
Fix an issue where 404 page couldn't switch to English

### LINKED ISSUE #6182: Language selector can't go back to English 404 page
### 📝 Issue Description

On any unknown URL, like https://docs.astro.build/foo/, the language selector doesn't work when trying to go back to English.

Any 404 page

### 🤔 Expected Behavior

Selecting English on a translated 404 page should go to the English 404 page

### 👀 Current Behavior

Going from English to any other language and switching between them works, but not going back to the default language.

### 🖥️ Browser

Chrome 120.0.6099.129

### 📄 Additional Information

This bug doesn't happen on Starlight. If you go to https://starlight.astro.build/foo/, you can switch back and forth between any language including English.

mingjunlu:
#### Description

- This PR adds a `<script>` to `NotFound.astro`, doing the following things:
  1. Check if current language in URL matches any option in the language selector
  2. If there's a match, apply the option's value to the language selector
- It should be able to:
  - Allow the user to switch to English 404 page
  - Resolve inconsistent displayed languages (main content of 404 vs language selector on the top right)
- Demonstration:
  | Before | After |
  | ------ | ----- |
  | <video src="https://github.com/withastro/docs/assets/40516784/08560478-c9d4-47f8-ab6d-367c406f7933" /> | <video src="https://github.com/withastro/docs/assets/40516784/f7396786-760f-4237-8e45-ff1d54795342" /> |
  | before-2 | after-2 |

#### Related issues & labels

- Closes #6182
- Suggested label: bug


delucis (commented):
Thanks for looking into this @mingjunlu! Really appreciate it.

We already have some scripts for the 404 inside the `NotFound` component itself, so I think it would be best to do any logic like this there:

https://github.com/withastro/docs/blob/3a5e1c22e0f9a0238a393399083af20fba91e426/src/components/NotFound.astro#L122-L158

dreyfus92 (approved):
Amazing job @mingjunlu, I've tested this on the preview and it is working as expected. LGTM 😁

src/components/NotFound.astro
```diff
@@ -156,3 +156,15 @@ const dictionary = Object.fromEntries(
 	}
 	customElements.define('four-oh-four', FourOhFour);
 </script>
+
+<script is:inline>
+	// Script is inlined to avoid visible flash
+	(() => {
+		// Get current language from URL, then apply it to language selector if matched
+    const cssSelector = `starlight-lang-select option[value="${location.pathname}"]`;
```
delucis, line 164: I don’t think this actually works in production. See https://docs-git-fork-mingjunlu-fix-language-selec-3866c6-astrodotbuild.vercel.app/de/x/ for example where the language is German, but the language select remains on English.

This is because the page is served statically and the language select always moves between routes like `/en/404/` and `/de/404/`, not the current page’s location. Instead we should probably parse the language segment and find the option that matches `/${lang}/404/`.

(In dev mode your approach works because the page is rendered on-demand and uses the actual URL instead of the `/404/` one.)

src/components/NotFound.astro
```diff
@@ -156,3 +156,15 @@ const dictionary = Object.fromEntries(
...
+    const matchedLanguageOption = document.querySelector(cssSelector);
+    if (!!matchedLanguageOption && !!matchedLanguageOption.parentElement) {
```
delucis, line 166: Small detail, but this can be simplified like this:

```suggestion
    if (matchedLanguageOption?.parentElement) {
```

delucis (changes requested):
Thanks again for the refactor @mingjunlu! Left a couple of comments — looks like we’ll need a little bit more work to make this work as desired. Let me know if you need a hand with any of this!

src/components/NotFound.astro
mingjunlu: Sure thing 👍

https://github.com/withastro/docs/pull/6763/commits/d03a4b97431b66e83ed09be8d5f39e25f4f1146d#diff-3f71d04203961deb2eef1334ac685c4596b67a98a96f0551b33fcf2a454cbd91R167

src/components/NotFound.astro
mingjunlu: Thanks for the explanation! I've made some [adjustments](https://github.com/withastro/docs/pull/6763/commits/d03a4b97431b66e83ed09be8d5f39e25f4f1146d#diff-3f71d04203961deb2eef1334ac685c4596b67a98a96f0551b33fcf2a454cbd91R164-R165).

Instead of looking for an exact match (`[value="/${lang}/404/"]`), I changed the CSS selector to match the beginning part only (`[value^="/${lang}/"]`), so that things would work in both production and development.

delucis (approved):
Looks great! Nice work @mingjunlu 🚀

PR #6790, opened 2024-02-06
feat: better sidebar UX

jw-12138:
#### Description (required)

Previously, the sidebar did not retain its scrolling position, which required readers to scroll down within the sidebar in order to continue reading. 

This PR will change the behavior of the sidebar, the highlighted item will instantly scroll to the view, this is somehow a better UX in my opnion.

Before: 

https://github.com/withastro/docs/assets/29943110/e00d30bb-52fb-4966-8158-81d2b4de5224

After: 

https://github.com/withastro/docs/assets/29943110/42132002-d75b-4307-b8d8-01e613c69954

### DIFF
--- a/src/components/starlight/Sidebar.astro
+++ b/src/components/starlight/Sidebar.astro
@@ -53,3 +53,13 @@ sidebar = await Promise.all(sidebar.map(remap));
 		font-weight: 700;
 	}
 </style>
+
+<script>
+	const currentActiveMenu : Element = document.querySelector('a[aria-current="page"]');
+
+	currentActiveMenu.scrollIntoView(<ScrollIntoViewOptions>{
+		behavior: 'instant',
+		block: 'center',
+		inline: 'center',
+	});
+</script>


liruifengv:
Cool! Call @delucis review.

delucis:
Thanks very much for the PR @jw-12138 — I agree we should improve this UX.

I think the best place to do this is upstream in Starlight, which is the tool Astro’s Docs uses to build the site.

We have an open discussion there — https://github.com/withastro/starlight/discussions/971 — and a PR which started to do some work in this direction — https://github.com/withastro/starlight/pull/1121.

The solution can ideally preserve exact scroll position across page navigations, so we’ll probably end up using something like `sessionStorage` to persist the position rather than a solution like this, which causes a visible scrolling motion on page load which is a bit distracting.

Because of this, I hope you don’t mind if we close this PR in favour of the upstream work. But I would still like to thank you for your time and care in trying to approve Astro’s docs 💜

jw-12138:
no problem, thanks for letting me know that this improvement is on the way.🥰

PR #6791, opened 2024-02-06
Bring community back

liruifengv:
#### Description (required)

Bring community back.

I noticed that this part was removed when the new docs were migrated to starlight. 

I don't know if this was done on purpose. I brought it back.

- Closes #
- Suggested label:


delucis:
Thanks for the PR @liruifengv and for noticing these details!

This was actually (at least semi 😁) intentional for a couple of reasons.

1. We want to move away completely from using the right sidebar for these links as they are not accessible at smaller viewports.

2. Now that we’ve started work on https://community.astro.build/ it is now the focus for links related to the community (although we should probably think about linking to THAT from docs somewhere).

That said, we didn’t really follow up on completing that work yet, so this is a good opportunity to discuss it — thanks for creating that opportunity!

I’ve posted a new issue with some details of what we’d want: #6845 
Would love to hear your thoughts there and will close this PR for now.

PR #6792, opened 2024-02-06
Fix typo in configuration-reference.mdx

mkrogemann:
Fix typo in output section

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/configuration-reference.mdx
+++ b/src/content/docs/en/reference/configuration-reference.mdx
@@ -166,7 +166,7 @@ You can customize the [redirection status code](https://developer.mozilla.org/en
 
 Specifies the output target for builds.
 
-- `'static'` - Building a static site to be deploy to any static host.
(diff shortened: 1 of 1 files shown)


sarah11918:
Oh my goodness @mkrogemann -- how long have we had this typo??  :sweat_smile: 

I really appreciate you taking the time to fix this for us! This is one of our few "special pages" though where the source code actually exists in the main `astro` repo and then gets pulled in here automatically.
 
So, the place to fix this is right here:

https://github.com/withastro/astro/blob/6e30bef652b23c5ae2e097f003bcc9f7a4f1e5c3/packages/astro/src/%40types/astro.ts#L556

Would you be willing to make a PR to the astro repo instead and correct this for us? If so, you can close this PR and make one there instead!

mkrogemann:
A new PR in astro repository has been created to affect the desired change in config documentation
New PR in astro repo:  #10008

PR #6793, opened 2024-02-06
Revert "Update mention about dynamically generated slot names (#6440)"

MoustaphaDev:
#### Description (required)

We're [temporarily rolling back a feature](https://github.com/withastro/compiler/pull/963#pullrequestreview-1865102712) (dynamically generated slots) which we added docs for in #6440, so I removed its mention in the docs

### DIFF
--- a/src/content/docs/en/basics/astro-components.mdx
+++ b/src/content/docs/en/basics/astro-components.mdx
@@ -291,7 +291,7 @@ Named slots can also be passed to [UI framework components](/en/guides/framework
 
 
 :::note
-Dynamically generating an Astro slot name, such as within a map function, was added in Astro v4.2.0. For earlier versions of Astro, it might be best to generate these dynamic slots within the framework itself.
+An astro slot name can not be dynamically generated, such as within a map function. If this feature is needed within UI framework components, it might be best to generate these dynamic slots within the framework itself.
 :::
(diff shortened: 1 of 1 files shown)


src/content/docs/en/basics/astro-components.mdx
```diff
@@ -291,7 +291,7 @@ Named slots can also be passed to [UI framework components](/en/guides/framework
 
 
 :::note
-Dynamically generating an Astro slot name, such as within a map function, was added in Astro v4.2.0. For earlier versions of Astro, it might be best to generate these dynamic slots within the framework itself.
+An astro slot name can not be dynamically generated, such as within a map function. If this feature is needed within UI framework components, it might be best to generate these dynamic slots within the framework itself.
```
sarah11918, line 294:
```suggestion
It is not possible to dynamically generate an Astro slot name, such as within a map function. If this feature is needed within UI framework components, it might be best to generate these dynamic slots within the framework itself.
```
I know you just faithfully reproduced the old message, but I'd improve a little now that I have the chance! (since search hits tend to be for e.g. "How do I dynamically generate an Astro slot name"

sarah11918 (commented):
Thank you for not blowing up people's sites!  🫡

PR #6797, opened 2024-02-06
Get remote dimensions

OliverSpeir:
#### Description (required)

Add inferSize property and update the wording of widths and heights property to Images component (which is source of truth for getImage and Picture as well)

#### Related PR

https://github.com/withastro/astro/pull/9976


src/content/docs/en/guides/images.mdx
```diff
@@ -240,6 +240,22 @@ By default, the `<Image />` component will produce a `.webp` file.
...
+By default, this value is set to `false` and you must manually specify both dimensions for your remote image.
+
+Set `inferSize: true` to infer these values from the image content when fetched. This is helpful if you don't know the dimensions of the remote image, or if they might change:
+
+```astro title="src/component/Avatar.astro" ins="inferSize={true}" del='"height="64" width="64"'
+---
+import { Image } from 'astro:assets';
+---
+<Image src="https://avatars.githubusercontent.com/u/622227?s=64," height="64" width="64" inferSize={true} alt="My GitHub Avatar" />
+```
```
sarah11918, line 258:
``````suggestion
```astro
---
import { Image, Picture, getImage } from 'astro:assets';
const myPic = await getImage({src: "https://example.com/example.png", inferSize: true, alt: ""})
---
<Image src="https://example.com/example.png" inferSize={true} alt="">
<Picture src="https://example.com/example.png" inferSize={true} alt="">
``````

**Comment by OliverSpeir** [comment]
> I'm curious why this example shows BOTH the dimensions AND inferSize={true} 

I think the idea was going to be we'd do ins and del with expressive code, but I don't think it actually looks as good as just marking the inferSize property


I added a note about unauthorized remote images, it made things feel a bit long but I'll see what y'all think

File: src/content/docs/en/guides/images.mdx [hunk 2]
```diff
@@ -240,6 +240,22 @@ By default, the `<Image />` component will produce a `.webp` file.
 - a preset (`low`, `mid`, `high`, `max`) that is automatically normalized between formats.
 - a number from `0` to `100` (interpreted differently between formats).
 
+##### inferSize
+
+<p><Since v="4.4.0" /></p>
+
+Allows you to set the original `width` and `height` of a remote image automatically.
+
+By default, this value is set to `false` and you must manually specify both dimensions for your remote image. If used with an unauthorized remote image, a request will still be made to the image domain, so the dimensions can be fetched.
```
sarah11918, line 249: How is using with an unauthorized remote image different from an authorized one? Is there a situation where you would want/need to fetch the dimensions, even if you're not allowed to fetch the image?

I could see this being useful if there's a common thing you'd want to do that requires that, and we explain that you can do what that thing is. Here, I'm a left a little confused thinking, "OK, so what? So, it's the same thing."

src/content/docs/en/guides/images.mdx
```diff
@@ -240,6 +240,22 @@ By default, the `<Image />` component will produce a `.webp` file.
...
+
+Set `inferSize: true` to infer these values from the image content when fetched. This is helpful if you don't know the dimensions of the remote image, or if they might change:
+
+```astro mark="inferSize"
+---
+import { Image } from 'astro:assets';
+---
+<Image src="https://example.com/cat.png" inferSize alt="A cat sleeping in the sun.">
```
sarah11918, line 257: I was going to ask whether explicitly setting true is necessary, or whether just adding the attribute is enough! (assuming the `={true}` wasn't deleted by mistake)

It seems like there are a few ways to set this attribute. Ideally, we'd like to demonstrate all of them, and make sure it's clear when each is properly used:

- `inferSize = {true}` or simply `inferSize` as a component attribute
- `inferSize: true` when used with `getImage()`

I don't think we should show `inferSize: true` in the text explanation if that's not how it's used in the `<Image />` component. We could say:

> Add the `inferSize` attribute to infer...
or
> Set `inferSize` to `true` to infer... (and let the example show the `={true}` syntax)
or
> Add `inferSize={true}` to the `<Image />` component (or just `inferSize`) to infer...

Do you think any of those work?

src/content/docs/en/guides/images.mdx
OliverSpeir: Good question, the basic usecase is if you want the image to be processed by sharp. Currently if a remote <Image /> is authorized it will be served by the astro site, if it is unauthorized it will become an . There might be some use cases for doing this, as the Image component still adds some optimzations even to unauthrized images ( `loading=lazy`, `decoding=async`, `width`, `height` ) 

The point of authorization is SSR security, it is hypothetically possible that some image hosting domain goes rogue and starts serving malware instead of images, so authorization implies that your astro web server will make a request to the image domain and you are okay with that

inferSize makes a request to the image domain, and when it is added to an unauthorized image it will result in a  

I think it's worth documenting but I agree its difficult to convey concisely why its important 

Basically:

Authorized without inferSize = requires width / height results in your own static version of the image
Unauthorized without inferSize = requires width / height but keeps the img "dynamic" so if the domain changes the image served at that url website will change 

Authorized with inferSize = your own static version of image
Unauthorized with inferSize = "dynamic" version of img

src/content/docs/en/guides/images.mdx
OliverSpeir: Yeah I see what you mean, I went with:
> Add inferSize to the <Image /> component (or inferSize: true to getImage()) 

With the width property for <Image /> there are two different documented syntaxes ( using a `number` and using a `string` ) and it basically is just left to the user to figure it out, so I think just showing them adding the  `inferSize` attribute to <Image /> is good enough. It could be a matter of persona preference though, I personally think `inferSize` looks much better than `inferSize={true}`

src/content/docs/en/guides/images.mdx
sarah11918: OK, gotcha. I propose then that we mention it, but *after* the code snippet. This isn't necessarily "up front" info to distract from just how to generally use it. So, I'd scrap that sentence from this paragraph, then after the code snippet, add something in like...

> `inferSize` can fetch the dimensions of a [remote image from a domain that has not been authorized](#authorizing-remote-images), however the image itself ...

And fill in the blank with what will/won't happen. Would something like that work?

src/content/docs/en/guides/images.mdx
sarah11918: Looks good to me!

If we're going with the simpler attribute here, then I'll review our blog post for 4.4 and make sure that's what is shown!

src/content/docs/en/guides/images.mdx
OliverSpeir: Sounds good I'll do that

PR #6798, opened 2024-02-06
update script behavior section

OliverSpeir:
#### Description (required)

Reworks the Scripts section of `view-transitions.mdx`, relativey significantly. Adds documentation for new attribute of `<script>` tags `data-astro-rerun` 

#### Related PR

https://github.com/withastro/astro/pull/9977

 #### For Astro version: `4.5`
 
 
 Would love your review on this @martrapp


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -443,11 +443,15 @@ When using the `<ViewTransitions />` router, the following steps occur to produc
 7. The router executes any new scripts added to the page.
 8. The `astro:page-load` event fires. This is the end of the navigation process.
 
-## Script behavior during page navigation
+## Script behavior
```
martrapp, line 446:
```suggestion
<a id="#script-behavior-during-page-navigation"/>
## Script behavior
```

martrapp, line 446:
```suggestion
<a id="#script-behavior-during-page-navigation"/>
## Script behavior
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -443,11 +443,15 @@ When using the `<ViewTransitions />` router, the following steps occur to produc
...
```
martrapp, line 447:
```suggestion
<a id="script-behavior-during-page-navigation"/>
## Script behavior

```

martrapp:
As for the broken link: It is mdx. You could replace the `##` with an `<h2>` and the old id or insert an anchor with the old id right in front of the `##`, see suggestion.

OliverSpeir:
> As for the broken link: It is mdx. You could replace the `##` with an `<h2>` and the old id or insert an anchor with the old id right in front of the `##`, see suggestion.

Thank you that's brilliant

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -443,11 +443,16 @@ When using the `<ViewTransitions />` router, the following steps occur to produc
...
 8. The `astro:page-load` event fires. This is the end of the navigation process.
 
-## Script behavior during page navigation
+<a id="script-behavior-during-page-navigation"/>
+## Script behavior
 
 When navigating between pages with the `<ViewTransitions />` component, scripts are run in sequential order to match browser behavior. 
 
-If you have code that sets up global state, this state will need to take into account that the script might execute more than once. Check for the global state in your `<script>` tag, and conditionally execute your code where possible:
+Module scripts, which the default scripts in Astro are, only ever execute once. If a module script exists on pages that are transitioned to after its initial execution they will be ignored. Inline scripts, ones with `is:inline` attribute, will not reexecute if they exist on both the page being transitioned to and the one the user is currently on.
```
martrapp, line 451:
```suggestion
Bundled module scripts, which are the default scripts in Astro, only ever execute once. If a bundled module script exists on pages that are transitioned to after its initial execution they will be ignored. Inline scripts, ones with `is:inline` attribute, will not reexecute if they exist on both the page being transitioned to and the one the user is currently on. 
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -443,11 +443,16 @@ When using the `<ViewTransitions />` router, the following steps occur to produc
...
-## Script behavior during page navigation
+<a id="script-behavior-during-page-navigation"/>
+## Script behavior
 
 When navigating between pages with the `<ViewTransitions />` component, scripts are run in sequential order to match browser behavior. 
 
-If you have code that sets up global state, this state will need to take into account that the script might execute more than once. Check for the global state in your `<script>` tag, and conditionally execute your code where possible:
+Module scripts, which the default scripts in Astro are, only ever execute once. If a module script exists on pages that are transitioned to after its initial execution they will be ignored. Inline scripts, ones with `is:inline` attribute, will not reexecute if they exist on both the page being transitioned to and the one the user is currently on. 
+
+Unlike module scripts, inline scripts have the potential to be reexcuted during a users visit to a site, if they exist on a page that is visited multiple times. Inline scripts might also reexecute if they exist on some pages but not all, if a user goes to a page without the script then back to one with the script it will be reexecuted.
```
martrapp, line 453:
```suggestion
Unlike bundled module scripts, inline scripts have the potential to be reexcuted during a users visit to a site, if they exist on a page that is visited multiple times. Inline scripts might also reexecute if they exist on some pages but not all, if a user goes to a page without the script then back to one with the script it will be reexecuted.
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -457,7 +462,13 @@ If you have code that sets up global state, this state will need to take into ac
 </script>
 ```
 
-Module scripts, including those that add event listeners for `DOMContentLoaded`, are only ever executed once because the browser keeps track of which modules are already loaded. You must add listeners for a [Lifecycle event](#lifecycle-events) to re-execute scripts during client-side page navigation.
+Inline scripts can be reexecuted after every transition by adding the `data-astro-reload` property. This will make the script `is:inline` implicitly, so it will not work for module scripts.
```
martrapp, line 465:
```suggestion
Inline scripts can be reexecuted after every transition by adding the `data-astro-reload` property. This will make the script `is:inline` implicitly, so it will not work for bundled scripts.
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -457,7 +462,13 @@ If you have code that sets up global state, this state will need to take into ac
...
+
+```astro
+<script data-astro-reload></script>
```
martrapp, line 468:
```suggestion
<script data-astro-reload>...</script>
```

martrapp:
Sorry for the nitpicking. The main reason for the suggested changes are scripts like this: 
```
<script type="module" data-astro-reload>...</script>
```
which are module scripts (= deferred loading) but not bundled (instead inlined) and can be re-executed without problems.

OliverSpeir:
I appreciate the nitpick especially when its as easy as hitting commit suggestions haha and I really do want this section of the docs to be as clear as possible because this is a huge part of the view tranistions experience 

haha at least running a module will run when its meant to (deffered) but yeah the whole any attribute creates is:inline is weird

OliverSpeir:
@sarah11918  Could you take another look at this when you get a chance?

sarah11918:
Yup! These 4.5 docs are on deck for this week! 🙌

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -443,11 +443,15 @@ When using the `<ViewTransitions />` router, the following steps occur to produc
...
+Bundled module scripts, which are the default scripts in Astro, only ever execute once. If a bundled module script exists on pages that are transitioned to after its initial execution they will be ignored. Inline scripts, ones with `is:inline` attribute, will not reexecute if they exist on both the page being transitioned to and the one the user is currently on. 
+
+Unlike bundled module scripts, inline scripts have the potential to be reexcuted during a user's visit to a site, if they exist on a page that is visited multiple times. Inline scripts might also reexecute if they exist on some pages but not all, if a user goes to a page without the script then back to one with the script it will be reexecuted.
+
+If you have code that sets up global state in an inline script, this state will need to take into account that the script might execute more than once. Check for the global state in your `<script>` tag, and conditionally execute your code where possible. This works because `window` is preserved.
```
yanthomasdev, line 454: A few grammar/typo nits. From what I saw the correct spelling for re-execute is using a hyphen.
```suggestion
Bundled module scripts, which are the default scripts in Astro, are only ever executed once. If a bundled module script exists on pages that are transitioned to after its initial execution they will be ignored. Inline scripts, ones with `is:inline` attribute, will not re-execute if they exist on both the page being transitioned to and the one the user is currently on. 

Unlike bundled module scripts, inline scripts have the potential to be re-executed during a user's visit to a site, if they exist on a page that is visited multiple times. Inline scripts might also re-execute if they exist on some pages but not all, if a user goes to a page without the script then back to one with the script it will be re-executed.

If you have code that sets up a global state in an inline script, this state will need to take into account that the script might execute more than once. Check for the global state in your `<script>` tag, and conditionally execute your code where possible. This works because `window` is preserved.
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -457,7 +461,13 @@ If you have code that sets up global state, this state will need to take into ac
 </script>
 ```
 
-Module scripts, including those that add event listeners for `DOMContentLoaded`, are only ever executed once because the browser keeps track of which modules are already loaded. You must add listeners for a [Lifecycle event](#lifecycle-events) to re-execute scripts during client-side page navigation.
+Inline scripts can be reexecuted after every transition by adding the `data-astro-rerun` property. This will make the script `is:inline` implicitly, so it will not work for bundled scripts.
```
yanthomasdev, line 464:
```suggestion
Inline scripts can be re-executed after every transition by adding the `data-astro-rerun` property. This will make the script `is:inline` implicitly, so it will not work for bundled scripts.
```

yanthomasdev (commented):
Thanks @OliverSpeir, just a few nits suggestions, after this it should be good to go 🙌

OliverSpeir:
Thank you! Great catch

src/content/docs/en/guides/view-transitions.mdx
sarah11918, line 446:
```suggestion
## Script behavior with view transitions
```
Do you have a reason for removing this? This original title attempted to disambiguate for people not using view transitions, so they wouldn't necessarily end up on this heading when doing a search. Perhaps this is better?

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -443,11 +443,15 @@ When using the `<ViewTransitions />` router, the following steps occur to produc
 7. The router executes any new scripts added to the page.
 8. The `astro:page-load` event fires. This is the end of the navigation process.
 
-## Script behavior during page navigation
+## Script behavior
 
 When navigating between pages with the `<ViewTransitions />` component, scripts are run in sequential order to match browser behavior. 
 
-If you have code that sets up global state, this state will need to take into account that the script might execute more than once. Check for the global state in your `<script>` tag, and conditionally execute your code where possible:
+Bundled module scripts, which are the default scripts in Astro, are only ever executed once. If a bundled module script exists on pages that are transitioned to after its initial execution they will be ignored. Inline scripts, ones with `is:inline` attribute, will not re-execute if they exist on both the page being transitioned to and the one the user is currently on.
```
sarah11918, line 450:
```suggestion
### Script re-execution

[Bundled module scripts](/en/guides/client-side-scripts/#script-processing), which are the default scripts in Astro, are only ever executed once. After initial execution they will be ignored, even if the script exists on the new page after a transition.
```
It seems very confusing to have the sentence about inline scripts running once, then immediately saying that they *can* run more than once in a following paragraph. I think keeping inline stuff to its own paragraph makes sense.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -443,11 +443,15 @@ When using the `<ViewTransitions />` router, the following steps occur to produc
...
 
-## Script behavior during page navigation
+## Script behavior
 
 When navigating between pages with the `<ViewTransitions />` component, scripts are run in sequential order to match browser behavior. 
 
-If you have code that sets up global state, this state will need to take into account that the script might execute more than once. Check for the global state in your `<script>` tag, and conditionally execute your code where possible:
+Bundled module scripts, which are the default scripts in Astro, are only ever executed once. If a bundled module script exists on pages that are transitioned to after its initial execution they will be ignored. Inline scripts, ones with `is:inline` attribute, will not re-execute if they exist on both the page being transitioned to and the one the user is currently on. 
+
+Unlike bundled module scripts, inline scripts have the potential to be re-executed during a user's visit to a site, if they exist on a page that is visited multiple times. Inline scripts might also re-execute if they exist on some pages but not all, if a user goes to a page without the script then back to one with the script it will be re-executed.
```
sarah11918, line 452:
``````suggestion
Unlike bundled module scripts, [inline scripts](/en/guides/client-side-scripts/#opting-out-of-processing) have the potential to be re-executed during a user's visit to a site if they exist on a page that is visited multiple times. Inline scripts might also re-execute when a visitor navigates to a page without the script, and then back to one with the script.

#### `data-astro-rerun`

<p><Since v="4.5.0" /></p>

To force inline scripts to re-execute after every transition, add the `data-astro-rerun` property. Adding any attribute to a script also implicitly adds `is:inline`, so this is only available for scripts that are not bundled and processed by Astro.

```astro
<script is:inline data-astro-rerun>...</script>
```

To ensure that a script runs every time a page is loaded during client-side navigation, it should be executed by a [lifecycle event](#lifecycle-events). For example, event listeners for `DOMContentLoaded` can be replaced by the [`astro:page-load`](/en/guides/view-transitions/#astropage-load) lifecycle event.
``````

File: src/content/docs/en/guides/view-transitions.mdx [hunk 11]
```diff
@@ -457,7 +461,13 @@ If you have code that sets up global state, this state will need to take into ac
...
 ```
 
-Module scripts, including those that add event listeners for `DOMContentLoaded`, are only ever executed once because the browser keeps track of which modules are already loaded. You must add listeners for a [Lifecycle event](#lifecycle-events) to re-execute scripts during client-side page navigation.
+Inline scripts can be re-executed after every transition by adding the `data-astro-rerun` property. This will make the script `is:inline` implicitly, so it will not work for bundled scripts.
+
+```astro
+<script is:inline data-astro-rerun>...</script>
+```
+
+If a script should execute every time a page is loaded, it should be executed by a [lifecycle event](#lifecycle-events). Event listeners for `DOMContentLoaded` can be replaced by the [`astro:page-load`](http://localhost:4321/en/guides/view-transitions/#astropage-load) lifecycle event.
```
sarah11918, line 470:
```suggestion
```
I suggest moving this earlier to where you're talking about whether or not scripts re-execute!

sarah11918 (commented):
OK! Thanks for waiting patiently @OliverSpeir ! I went through this and think some restructuring of the content (both the original and yours) makes it even more helpful, so see whether my attempt makes sense here, and let me know what you think!

OliverSpeir:
@sarah11918 brilliant suggestions thank you

sarah11918 (approved):
Thanks, @OliverSpeir ! I really appreciate you adding the extra context over and above documenting `astro-data-rerun`!

PR #6815, opened 2024-02-07
Update StoryblokComponent import source path on tutorial

schabibi1:
#### Description (required)

Update from below:

```js
import StoryblokComponent from '@storyblok/astro/components/StoryblokComponent.astro';
```

to this:

```js
import StoryblokComponent from "@storyblok/astro/StoryblokComponent.astro";
```

- Suggested label: docs fix


sarah11918 (approved):
Thank you for submitting this fix! This will help our Storyblok users, and we all appreciate it!  :rocket:

PR #6843, opened 2024-02-09
[i18nIgnore] Change event name

ktym4a:
#### Description (required)

The event name `app-notification` does not exist and can be misleading to users, so change it to an event name that does exist.


Yan-Thomas (approved):
LGTM! Thanks for catching this @ktym4a 🙌

PR #6851, opened 2024-02-10
[i18nIgnore] Fix date formatting in content-collections.mdx

AitorMT:
#### Description (required)

The content of pull request #6718 has been added in the available languages.


Yan-Thomas:
Since the original change was done without `i18nIgnore`, if we merge this one using it as well, the pages won't be marked as updated if that's the only change they're missing. If we merge without the keyword, we might lose the tracking status of other changes in the file if the translation is not totally up to date, so I'll close this one and let translators do individual PRs to this file. Thanks for the contribution!

PR #6856, opened 2024-02-10
Introduces transition:reload instead of data-astro-reload

martrapp:
## Background:
Since version 3.0.0 you can use `data-astro-reload` to disable view transitions for certain links.
At first this was for `<a>` tags, but then it was extended to `<form>`, `<area>` and `<svg>` anchors.
https://github.com/withastro/astro/pull/9977 now also introduces `data-astro-reload` for `<script>` tags to allow users to mark scripts for re-execution during view transitions.

Astro only has a few places where it relies on `data-astro-*` attributes. These are typically for low-level techniques or implementation details. Most of the time Astro uses `xyz:*` directives to control things. These look more official and, more importantly, they have better compiler and language tool support.

## This PR ...
documents the  `transition:reload` directive introduced by  https://github.com/withastro/compiler/pulls that will replace `data-astro-reload` for the sake of better warning messages when misused and better language tool support. 
`data-astro-reload` will still continue to work.

Documents [compiler PR #965](https://github.com/withastro/compiler/pull/965)
Not sure yet what implication a compiler minor(?) has on Astro releases and docs.
This PR should be merged only after [#6798](https://github.com/withastro/docs/pull/6798) and  [compiler PR #965](https://github.com/withastro/compiler/pull/965).

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v4.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v4.mdx
@@ -44,7 +44,7 @@ After upgrading Astro to the latest version, you may not need to make any change
 But, if you notice errors or unexpected behavior, please check below for what has changed that might need updating in your project.
 :::
 
-Astro v4.0 includes [potentially breaking changes](#breaking-changes), as well as the [removal of some previously deprecated features](#previously-deprecated-features-now-removed). 
+Astro v4.0 includes [potentially breaking changes](#breaking-changes), as well as the [removal of some previously deprecated features](#previously-deprecated-features-now-removed).
 
 If your project doesn't work as expected after upgrading to v4.0, check this guide for an overview of all breaking changes and instructions on how to update your codebase.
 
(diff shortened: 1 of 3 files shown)


sarah11918:
Hi @martrapp !

If this is simply an alternative way of achieving the same thing, then we should probably document it as that: an alternative, and *not* replace the old documentation. Even if this way is "better", people will still have the old way in their projects, and people are notoriously bad at updating to the latest version. So, we can't have this as the *only* documented way when it simply will not work for a lot of people.

So, this becomes tricky from my perspective because we can NOT simply document what it is as "the way to do something", but in fact have to provide guidance on this as a replacement.

My suggestion would be to document this as if it were in fact a new thing, immediately after `astro-data-reload` like:

> - `data-astro-reload`: an `<a>` tag attribute to [force a full-page navigation](#preventing-client-side-navigation). Deprecated in Astro v4.x.0. Use `transition: reload` instead.
> - <Since 4.x.0 /> `transition: reload`: a directive for `<a>`, `<area>`,  `<form>` and `<script>` tags to [force a full-page navigation](#preventing-client-side-navigation). Replaces `astro-data-reload`.

Does that give you enough to go on to reshape this a bit?

martrapp:
We stay with data-astro-reload and introduce data-astro-rerun, see https://github.com/withastro/astro/pull/9977

PR #6861, opened 2024-02-12
corrects misspelled data-astro- attribute in ko and pt-br 

martrapp:
#### Description (required)

Corrects misspelled data-astro- attribute in ko and pt-br 

It seems like a no-brainer to me, but I don't understand the text around my change :)
So I would be grateful for a second pair of eyes.


Yan-Thomas (approved):
Nice catch, LGTM

PR #6899, opened 2024-02-15
Reminder to add the app to PNPM workspace

eric-burel:
If the app is not added to the workspace, user will hit an error claiming "astro" is not installed

- Closes #
- Suggested label:


yanthomasdev:
Thanks @eric-burel, appreciate the contribution! I'll note we prefer not to document these more specific, monorepo-related setup instructions since these can change and vary widely depending on the tool used, meaning it isn't really in our control to give proper instructions on this.

PR #6911, opened 2024-02-16
i18n(zh-CN): Fix incorrect translation text in `guides/integrations-guide/node.mdx` 

mitian233:
#### Description (required)

Unify the style used in translation: 'Why Astro Node.js' should be translate to "为什么选择 Astro Node.js". 

This is a small correction.


liruifengv (approved):
LGTM! Thank you!

PR #6914, opened 2024-02-16
Clearer instructions for using the Vercel adapter with additional services

TudorZgimbau:
#### Description (required)

While updating the Vercel adapter, I read the Astro docs to make sure I'm using the latest APIs and mistakenly concluded that it is not needed in static builds, even if using other Vercel services which require a set up, such as Web Analytics.

This PR mentions the need for installing and configuring the adapter even when using static builds, if other services need to be set up or customly configured.

- Closes #

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/vercel.mdx
+++ b/src/content/docs/en/guides/integrations-guide/vercel.mdx
@@ -10,9 +10,9 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 import Since from '~/components/Since.astro';
 import ReadMore from '~/components/ReadMore.astro'
 
-This adapter allows Astro to deploy your [`hybrid` or `server` rendered site](/en/basics/rendering-modes/#on-demand-rendered) to [Vercel](https://www.vercel.com/).
+This adapter allows Astro to deploy your [`static`, `hybrid` or `server` rendered site](/en/basics/rendering-modes/#on-demand-rendered) to [Vercel](https://www.vercel.com/) and configure additional services such as [Vercel Web Analytics](https://vercel.com/docs/analytics).
 
-If you're using Astro as a [static site builder](/en/basics/rendering-modes/#pre-rendered), you don't need an adapter.
+If you're using Astro as a [static site builder](/en/basics/rendering-modes/#pre-rendered) and do not need to set up services such as [Vercel Web Analytics](https://vercel.com/docs/analytics) and [Vercel Image Optimization](https://vercel.com/docs/image-optimization) or use a custom configuration, you don't need an adapter.
 
 Learn how to deploy your Astro site in our [Vercel deployment guide](/en/guides/deploy/vercel/).
 


src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -10,9 +10,9 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 import Since from '~/components/Since.astro';
 import ReadMore from '~/components/ReadMore.astro'
 
-This adapter allows Astro to deploy your [`hybrid` or `server` rendered site](/en/basics/rendering-modes/#on-demand-rendered) to [Vercel](https://www.vercel.com/).
+This adapter allows Astro to deploy your [`static`, `hybrid` or `server` rendered site](/en/basics/rendering-modes/#on-demand-rendered) to [Vercel](https://www.vercel.com/) and configure additional services such as [Vercel Web Analytics](https://vercel.com/docs/analytics).
```
sarah11918, line 13:
```suggestion
This adapter allows Astro to deploy your [`hybrid` or `server` rendered site](/en/basics/rendering-modes/#on-demand-rendered) to [Vercel](https://www.vercel.com/).
```
Since adapters are not needed for `static` sites *in general*, I think it's more helpful if this statement stays as is, and we instead explain further below where we do talk about static sites!

src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -10,9 +10,9 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
...
 
-If you're using Astro as a [static site builder](/en/basics/rendering-modes/#pre-rendered), you don't need an adapter.
+If you're using Astro as a [static site builder](/en/basics/rendering-modes/#pre-rendered) and do not need to set up services such as [Vercel Web Analytics](https://vercel.com/docs/analytics) and [Vercel Image Optimization](https://vercel.com/docs/image-optimization) or use a custom configuration, you don't need an adapter.
```
sarah11918, line 15:
```suggestion
If you're using Astro as a [`static` site builder](/en/basics/rendering-modes/#pre-rendered), you only this adapter if you are using additional Vercel services (e.g. [Vercel Web Analytics](https://vercel.com/docs/analytics), [Vercel Image Optimization](https://vercel.com/docs/image-optimization)). Otherwise, you do not need an adapter to deploy your `static` site.
```
I think keeping all the static stuff together makes sense!

sarah11918 (commented):
Thank you for this helpful addition, @TudorZgimbau ! You're absolutely right that there are cases where `static` sites will use this Vercel adapter, and we appreciate the call out.

I did restructure the text to keep separate the on-demand rendering from the static rendering, because in general we do want people to understand that adapters are primarily an SSR feature. So, I moved your extra note to the second paragraph where static sites were already mentioned.

We also do try to phrase things positively, "do this if you need" instead of negatively "only do this if you don't need", so I made some tiny adjustments like that to match our docs style.

See what you think of the changes!

TudorZgimbau:
Hi Sarah, thanks for reviewing and suggesting the changes! I agree clarity is important and I think that phrasing instructions positively might also contribute to that. I applied your suggestions.

sarah11918 (approved):
Alright, this looks great! Thank you so much for this helpful addition @TudorZgimbau , and welcome to Team Docs!  :partying_face:

TudorZgimbau:
Thank you too, @sarah11918! Have a great day!

PR #6915, opened 2024-02-16
Explain that props are updated on navigation

matthewp:
#### Description (required)

With https://github.com/withastro/astro/pull/10136 `transition:persist` is changing to update props upon navigation when used on an island. This explains that props will trigger a re-render.

### DIFF
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -134,6 +134,8 @@ As a convenient shorthand, `transition:persist` can alternatively take a transit
 <video controls="" autoplay="" transition:persist="media-player">
 ```
 
+When you add `transition:persist` to an island, the state is retained upon navigation, but any props that change will be carried over to the new page, causing your component to re-render with new props, but the existing state.
+
 ### Built-in Animation Directives
 
 Astro comes with a few built-in animations to override the default `fade` transition. Add the `transition:animate` directive to individual elements to customize the behavior of specific transitions.


yanthomasdev (approved):

sarah11918:
@matthewp - Is this example (in the same section, but just above what you added) still accurate? How are we making clear the distinction between a component's props vs its state?

This change now updates island props upon navigation, so would this counter get set back to 5 by default?

image

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -134,6 +134,8 @@ As a convenient shorthand, `transition:persist` can alternatively take a transit
 <video controls="" autoplay="" transition:persist="media-player">
 ```
 
+When you add `transition:persist` to an island, the state is retained upon navigation, but any props that change will be carried over to the new page, causing your component to re-render with new props, but the existing state. You can control this behavior with the `transition:persist-props` directive. Adding this directive will keep the existing props and not copy over new values.
```
sarah11918, line 137:
```suggestion
#### `transition:persist-props`
<p><Since v="4.5.0" /></p>
**Default**: `false`

This allows you to control whether or not an island's props should be persisted upon navigation.

By default, when you add `transition:persist` to an island, the state is retained upon navigation, but your component will re-render with new props. This is useful, for example, when a component receives page-specific props such as the current page's `title`.

You can override this behavior with `transition:persist-props="true"`. Adding this directive will keep an island's existing props (not re-render with new values) in addition to maintaining its existing state.
```
Here's my suggestion, so we can mention when this was introduced too.

sarah11918 (commented):
Thanks @matthewp !

I took some of the description from the language of the changeset in the implementation PR to provide a use-case, and to include the version it was introduced.

Please also read the entire section now that the default props behaviour has changed, and make sure the existing description here is still accurate, or whether mention of the new default behaviour might need to be corrected earlier, too!

matthewp:
@sarah11918 It might be a good idea to rename `count` to `initialCount` or something, would that make it more clear? In a case like this you are really providing the initial value to the component. The component then keeps the "real" count in its state. That state will not be affected by navigation.

src/content/docs/en/guides/view-transitions.mdx
matthewp: You don't need the `="true"` part but otherwise very good! I can't edit the suggestion so I think i'll commit and then change.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -134,6 +134,16 @@ As a convenient shorthand, `transition:persist` can alternatively take a transit
...
 
+#### `transition:persist-props`
+<p><Since v="4.5.0" /></p>
+**Default**: `false`
+
+This allows you to control whether or not an island's props should be persisted upon navigation.
+
+By default, when you add `transition:persist` to an island, the state is retained upon navigation, but your component will re-render with new props. This is useful, for example, when a component receives page-specific props such as the current page's `title`.
+
+You can override this behavior with `transition:persist-props`. Adding this directive will keep an island's existing props (not re-render with new values) in addition to maintaining its existing state.
```
sarah11918, line 145:
``````suggestion
You can override this behavior with `transition:persist-props`. Adding this directive will keep an island's existing props (not re-render with new values) in addition to maintaining its existing state.

```astro
<ProductListing transition:persist-props=true products={products} />
```
``````
Thanks @matthewp ! Wording looks fine to me, but if they do have to set this to `true`, then we should probably show an example of that. (If they *don't* have to set it to true, and just merely adding it adds it as true, then I would choose to remove the "Default: false" from above.)

Comment (matthewp) [reply src/content/docs/en/guides/view-transitions.mdx:145]: They don't have to set it to true. It works like a boolean attribute where the presence of the attribute makes it be true. I can see how the "defaults to false" part might be confusing, even if its technically accurate, so we can remove if you think that's best.

Comment (sarah11918) [reply src/content/docs/en/guides/view-transitions.mdx:145]: Great, then I'll remove the default line and we're good to go!

File: src/content/docs/en/guides/view-transitions.mdx [hunk 3]
```diff
@@ -134,6 +134,16 @@ As a convenient shorthand, `transition:persist` can alternatively take a transit
 <video controls="" autoplay="" transition:persist="media-player">
 ```
 
+#### `transition:persist-props`
+<p><Since v="4.5.0" /></p>
+**Default**: `false`
```
sarah11918, line 139:
```suggestion
```

sarah11918 (approved):
Approved!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -112,10 +112,10 @@ For example, the following `<video>` will continue to play as you navigate to an
 
 You can also place the directive on an [Astro island](/en/concepts/islands/) (a UI framework component with a [`client:` directive](/en/reference/directives-reference/#client-directives)). If that component exists on the next page, the island from the old page **with its current state** will continue to be displayed, instead of replacing it with the island from the new page.
 
-In the example below, the `count` will not be reset when navigating back and forth across pages that contain the `<Counter />` component with the `transition:persist` attribute.
+In the example below, the component's `count` internal state will not be reset when navigating back and forth across pages that contain the `<Counter />` component with the `transition:persist` attribute.
```
sarah11918, line 115:
```suggestion
In the example below, the component's `initialCount` internal state will not be reset when navigating back and forth across pages that contain the `<Counter />` component with the `transition:persist` attribute.
```

PR #6926, opened 2024-02-17
i18n(zh-cn): Update `guides/migrate-to-astro/from-gatsby.mdx`

100gle:
#### Description (required)

Update `guides/migrate-to-astro/from-gatsby.mdx`

Upstream PR:

- #6833

#### Related issues & labels (optional)

- Closes #
- Suggested label:

PR #6931, opened 2024-02-17
i18n(zh-cn): fix link from-gatsby.mdx

AitorMT:
#### Description (required)

The link has been fixed, it was pointing to the English version edition. #6926

### DIFF
--- a/src/content/docs/zh-cn/guides/migrate-to-astro/from-gatsby.mdx
+++ b/src/content/docs/zh-cn/guides/migrate-to-astro/from-gatsby.mdx
@@ -573,5 +573,5 @@ export const pageQuery = graphql`
 - 博客帖子：[再次迁移：从 Gatsby 到 Astro](https://logarithmicspirals.com/blog/migrating-from-gatsby-to-astro/)。
 
 :::note[你有资源分享吗？]
-如果你找到（或制作）了一个关于将 Gatsby 网站转换为 Astro 的有用的视频或博客文章，[编辑此页面](https://github.com/withastro/docs/edit/main/src/content/docs/en/guides/migrate-to-astro/from-gatsby.mdx)并将其添加到这里！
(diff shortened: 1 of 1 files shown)


liruifengv:
Thank you! @AitorMT 

I've noticed this, and I think that if new content is to be added, it should be added to the English page first, and then updated for the other i18n pages.

So I think this makes sense now.

AitorMT:
> Thank you! @AitorMT
> 
> I've noticed this, and I think that if new content is to be added, it should be added to the English page first, and then updated for the other i18n pages.
> 
> So I think that makes sense for now.

So, do I close the pull request ?

liruifengv:
This may need to be discussed. cc @Yan-Thomas @sarah11918

sarah11918:
Ah, great question! Thank you for asking!

I will also wait for @Yan-Thomas to answer, but my opinion is yes, the link to add a new piece of content *should* go to the English site. That is the only way that the other languages would know to update. 

Links to update the page itself (like in the sidebar), should point to the proper language. But, all new content must first be added to the English page.

So, one option is to just close this PR. Another option is to update the text, if you feel that would be helpful, in other languages to make it clear that these links should be added to the English site.

e.g., the translation of:

> If you find (or make) a useful video or blog post about converting a Gatsby site to Astro , [ Edit the source page] ( https://github.com/withastro/docs/edit/main/src/content/ docs/ en-us /guides/migrate-to-astro/from-gatsby.mdx ) to add it here!

Let's see if Yan thinks anything different!

AitorMT:
If you tell me how to do this, I will do it, thank you.

Yan-Thomas:
Oh right, sorry I wasn't clear @AitorMT, I believe something along what Sarah suggested works:

> If you found (or made!) a helpful video or blog post about converting a Gatsby site to Astro, [edit the source page](https://github.com/withastro/docs/edit/main/src/content/docs/en/guides/migrate-to-astro/from-gatsby.mdx) to add it here!

liruifengv:
> If you tell me how to do this, I will do it, thank you.

Yeah. You can update the en page use Sarah's suggestion. And other language page is in the hands of our translators

yanthomasdev:
Looks like this PR got a bit stale, I'll close this one but I would appreciate a new one to be open with the change to English as suggested above when you have time @AitorMT!

PR #6938, opened 2024-02-18
Update Astro to 4.4

sasoria:
#### Description

This pull request updates Astro to 4.4.0. It looks like there are some performance improvements in Astro [4.4](https://astro.build/blog/astro-440/) that might be relevant: 

> Notably, this change reduced the build time of Starlight websites with large sidebars by up to 47% in extreme cases


sarah11918:
Thanks, @sasoria ! I want to wait for @delucis's approval on this, since Astro Docs is a highly customized Starlight site, and sometimes we do have to be more careful about our updates.

But the preview is looking good to me, and we'll be happy to update as soon as we can!  💪

delucis (approved):
Looks good to me! Thanks for the PR @sasoria 💜

PR #6942, opened 2024-02-18
ci: Bump GitHub workflow to latest versions

deining:
This PR bumps GitHub workflow to latest released versions.


delucis:
Thanks for the PR @deining! Could you summarise if there are any breaking changes in these version updates that could impact us, which we should be aware of?

deining:
> Thanks for the PR @deining! Could you summarise if there are any breaking changes in these version updates that could impact us, which we should be aware of?

No, I don't think there is any breaking chance involved.

delucis:
Thanks for the response — the CI tests are running fine in this PR, but there also the other 5 workflows updated here. These all run in other circumstances (e.g. when a PR merges, once per day at a set time, etc.). Given that most of the version updates are major (including https://github.com/peter-evans/create-pull-request bumping v3 to v6), there are likely some breaking changes listed in the CHANGELOG for those actions. It would be good to check those CHANGELOGs to understand what those are and if they in fact cause any issues for us or not.

yanthomasdev:
Thanks for this @deining, I'll be closing this PR since we have updated the Action versions used in other PRs, we would appreciate a new PR if you find we missed updating it in any place, though! 🙌

PR #6959, opened 2024-02-20
Update images.mdx

alexnguyennz:
#### Description (required)

- add Picture to note about required alt attribute


OliverSpeir:
https://docs.astro.build/en/guides/images/#properties-1

> <Picture /> accepts all the properties of the <Image /> component 

I think this sentence is the reason why the <Image> properties don’t explicitly mention they apply to <Picture/> but I see that accepts doesn’t specify exactly requires

src/content/docs/en/guides/images.mdx
```diff
@@ -679,7 +679,7 @@ It returns an object with the following properties:
 
 Not all users can see images in the same way, so accessibility is an especially important concern when using images. Use the `alt` attribute to provide [descriptive alt text](https://www.w3.org/WAI/tutorials/images/) for images.
 
-This attribute is required for the `<Image />` component. `<Image />` will throw an error if no alt text is provided.
+This attribute is required for the `<Image />` and `<Picture />` components. `<Image />` and `<Picture />` will throw an error if no alt text is provided.
```
sarah11918, line 682:
```suggestion
This attribute is required for both the `<Image />` and `<Picture />` components. If no alt text is provided, a helpful error message will be provided reminding you to include the `alt` attribute.
```
How about something like this, to avoid some text duplication?

sarah11918 (commented):
Thank you for updating here, @alexnguyennz -- great catch! I left a suggestion because now it feels like a lot of duplication. See what you think!

sarah11918 (commented):
Thank you for updating here, @alexnguyennz -- great catch! I left a suggestion because now it feels like a lot of duplication. See what you think!

src/content/docs/en/guides/images.mdx
alexnguyennz: Makes sense to me!

Yan-Thomas (approved):
Thanks!

PR #6968, opened 2024-02-21
[i18nIgnore]: Fix pnpm command

Fryuni:
#### Description

Continues the work of #5284 in all other languages.

There is one remaining case of `pnpm install` on the outdated Portuguese translation of the Netlify guide. In that case, the correct command is indeed `pnpm install`, but it will be removed once that page is updated.

Also fixes the code blocks for one Russian case that was missing newlines

- English fix made in #5284


Yan-Thomas (approved):
LGTM! Thanks 🚀

PR #6986, opened 2024-02-22
Update aliases.mdx

eric-burel:
#### Description (required)

Currently I think the change in tsconfig.json is not picked up by the dev server, which can be confusing.
Not sure if it's a limitation of Astro or Vite but anyway the doc could mention it

- Closes #
- Suggested label:


AitorMT (approved):

Mayank-Malviya:
I did not faced alias changes affecting the development server, it can also be because I am using existing themes from astro.build with nothing much todo in tsconfig file

dreyfus92 (commented):
Hey everybody, I've bootstrapped a brand new project and here's what I found:
- If you're running your project locally and if apply changes to your `tsconfig.json` file it will restart the `ts-server` for you
image
- If you're not running your project, the `ts-server` will instantly notice about the changes made to the file.
- I tested this on macOS.

src/content/docs/en/guides/aliases.mdx
```diff
@@ -33,6 +33,7 @@ You can add import aliases from either `tsconfig.json` or `jsconfig.json`.
 
 :::note
 Make sure `compilerOptions.baseUrl` is set so the aliased paths can be resolved.
+You may need to restart your development server if it's running.
 :::
 
 With this change, you can now import using the aliases anywhere in your project:
```
sarah11918, line 39:
```suggestion
:::note
Make sure `compilerOptions.baseUrl` is set so the aliased paths can be resolved.
:::

The development server will automatically restart after this configuration change. You can now import using the aliases anywhere in your project:
```

How is something like this for getting across the idea that a restart is required, but is expected to happen automatically?

We try to avoid saying "should" in docs, so I'd prefer to say it "will" restart if that's actually the expected behavior. (Sure, it occasionally might not. But there are a lot of things that occasionally go wrong. 😄 )

But, this is a clue that if something isn't working, they can check to see whether the dev server actually *did* restart or do so manually.

sarah11918 (commented):
Thanks for bringing your experience into the docs to help others, @eric-burel, and thanks to everyone here for jumping in, confirming, and discussing!

I went through the conversation and made a proposal that I think captures where everything left off here? See what you think!

BryceRussell:
Hey @eric-burel thanks for opening a PR on this! I tried reproducing this locally, but I couldn't get it to break. The dev server is restarting, and types are updating when I make changes to `tsconfig.ts`. Even though this doesn't seem to be an issue, the docs could be more explicit about the dev server restarting when making these changes. I am going to approve Sarah's suggestion and merge this, thanks again! 🚀

PR #7045, opened 2024-02-26
chore(middleware): clarify usage

ematipico:
#### Description (required)

In core, we have had some feedback where users expected to have the middleware run also in SSG when the website is deployed.

From one point of view, I can understand this. A middleware should run every time, at runtime, because that's what a middleware is meant for. Although, this isn't the case for our static and prerendered pages.

This PR is meant to add a small note to clarify how the middleware works in Astro.

### DIFF
--- a/src/content/docs/en/guides/middleware.mdx
+++ b/src/content/docs/en/guides/middleware.mdx
@@ -11,6 +11,11 @@ This also allows you to set and share request-specific information across endpoi
 
 Middleware is available in both SSG and SSR Astro projects.
 
+:::note
+In SSG, the middleware runs during the **build process**. Once the app is built, the middleware won't run in the deployed website. This is valid also for prerendered pages.
+In SSR, the middleware runs once the website is deployed.
+:::
+
 ## Basic Usage
 
 1. Create `src/middleware.js|ts` (Alternatively, you can create `src/middleware/index.js|ts`.)


src/content/docs/en/guides/middleware.mdx
```diff
@@ -11,6 +11,11 @@ This also allows you to set and share request-specific information across endpoi
 
 Middleware is available in both SSG and SSR Astro projects.
 
+:::note
+In SSG, the middleware runs during the **build process**. Once the app is built, the middleware won't run in the deployed website. This is valid also for prerendered pages.
+In SSR, the middleware runs once the website is deployed.
```
sarah11918, line 16:
```suggestion
On pre-rendered pages, the middleware runs only during the **build process** to create static pages and does not run in the deployed website. This includes all pages in a `static` (SSG) project.

For on-demand rendered pages in `server` or `hybrid` mode, the middleware runs in the deployed website when the route is requested.
```
What do you think about this? I tried to split up prerendered vs. on-demand. rendered because SSR can have 2 kinds of pages.

sarah11918 (commented):
Thanks for adding this clarification @ematipico ! See what you think about my suggestion!

src/content/docs/en/guides/middleware.mdx
ematipico: I think the `hybrid` requires an additional clarification. Astro pages are rendered statically by default unless a user sets `prerender = false`.

sarah11918: Thanks @ematipico, I updated to be even more explicit. What do you think now? Too much?  :smile:

ematipico: That's great! Thank you

sarah11918 (approved):
Looks good to me @ematipico ! Happy for you to merge when you're happy!

PR #7046, opened 2024-02-26
chore(i18n): clarify routing logic

ematipico:
#### Description (required)

We've had some reports where users were surprised by **where** the logic of our internal i18n middleware comes into play.

This PR adds a new paragraph that attempts to shed some lights in this technical aspects of the feature.

https://github.com/withastro/astro/issues/10214


src/content/docs/en/guides/internationalization.mdx
```diff
@@ -220,6 +218,15 @@ Astro relies on the following headers in order to support the feature:
...
 Make sure that your server proxy/hosting platform is able to provide this information. Failing to retrieve these headers will result in a 404 (status code) page.
 
+#### Routing in details 
+
+Astro uses a [middleware](/en/guides/middleware/) to implement its routing logic. This middleware function is placed in the [first position](/en/guides/middleware/#chaining-middleware), even before the user's middleware, and it checks every `Response` coming from each route.
+
+This means two things:
+- the i18n middleware business logic comes into play **after** a route is rendered;
+- any operation (redirects, etc.) done in the users middleware are checked by the i18n middleware;
+
```
sarah11918, line 228:
```suggestion
#### Routing logic

Astro uses a [middleware](/en/guides/middleware/) to implement its routing logic. This middleware function is placed in the [first position](/en/guides/middleware/#chaining-middleware) where it awaits every `Response` coming from any additional middleware and each page route before finally executing its own logic.

This means that operations (e.g. redirects) from your own middleware and your page logic are run first, your routes are rendered, and then the i18n middleware performs its own actions such as verifying that a localized URL corresponds to a valid route.
```

OK, as usual, I'm not sure whether this is *entirely* correct or the right nuance, so you'll need to fact check and rewrite as necessary to be accurate!  :smile: 

But, what about something like this? Notably:

- since we are talking about order of things happening, I thought highlighting the "await" (vs "checks") from the first position example might make that clearer. (The second paragraph could say "... and your page logic are run and checked first" if you think that's more accurate to say. But I honestly just don't know enough to know what the i18n middleware is actually "checking on" in the other logics -- I just know it's letting the other stuff run first! This is why I'm not as clear on the use of "checks/checking" vs "await")

- the linked discussion refers to "page logic" executing, and this only mentioned additional user middlewares. So, maybe it's also correct to include logic from the page iteself, even with no additional middleware? (That's the idea I tried to capture, if it's correct/helpful.)

- instead of a list of "things that are true", I thought attempting to illustrate the actual sequence of events could be helpful here! So, I took a shot at making this sequence up, but again, make it true!

sarah11918 (commented):
Thanks for this, @ematipico ! This was a helpful explanation, and especially the linking to the example of chaining middleware was an excellent call!

I tried an alternative version, and gave some comments as to what my thinking was when I wrote it. See if you think any of that is helpful!

src/content/docs/en/guides/internationalization.mdx
ematipico: Thank you Sarah. Yes the suggestion is perfect

sarah11918 (approved):
Great! This is a super helpful addition @ematipico and thanks for providing this extra context. Ready for you to merge when you're happy!

PR #7051, opened 2024-02-27
Add LightningCSS to preprocessors

jasikpark:
#### Description (required)

"How to set up LightningCSS w/ Astro"

LightningCSS is supported by Astro, and allows using several features that something like Sass might support but via current CSS without leaving old browsers behind. We should document how to set it up

- Closes #
- Suggested label: 

Feel free to workshop / move this / update my PR, I just wanted to make sure it's documented *somewhere*

### DIFF
--- a/src/content/docs/en/guides/styling.mdx
+++ b/src/content/docs/en/guides/styling.mdx
@@ -491,17 +491,20 @@ Use `<style lang="less">` in `.astro` files.
 npm install lightningcss
 ```
 
-Update your `astro.config.ts`
+Update your `vite` configuration in `astro.config.mjs`:
 
-```diff
-+  vite: {
-+    css: {
-+      transformer: "lightningcss",
-+    },
-+  },
+```js title="astro.config.mjs" ins={4-8}
+import { defineConfig } from 'astro/config'
+
+export default defineConfig({
+  vite: {
+    css: {
+      transformer: "lightningcss",
+    },
+  },
+})
+  
 ```
-
-Then LightningCSS will optimize and down-level your vanilla CSS.
 
 ### In framework components
 
--- a/src/content/docs/en/guides/styling.mdx
+++ b/src/content/docs/en/guides/styling.mdx
@@ -485,6 +485,24 @@ npm install less
 
 Use `<style lang="less">` in `.astro` files.
 
+### LightningCSS
+
+```shell
+npm install lightningcss
+```
+
+Update your `astro.config.ts`
+
+```diff
++  vite: {
++    css: {
++      transformer: "lightningcss",
++    },
++  },
+```
+
+Then LightningCSS will optimize and down-level your vanilla CSS.
+
 ### In framework components
 
 You can also use all of the above CSS preprocessors within JS frameworks as well! Be sure to follow the patterns each framework recommends:


src/content/docs/en/guides/styling.mdx
```diff
@@ -485,6 +485,24 @@ npm install less
...
+
+```diff
++  vite: {
++    css: {
++      transformer: "lightningcss",
++    },
++  },
+```
+
+Then LightningCSS will optimize and down-level your vanilla CSS.
```
sarah11918, line 504:
``````suggestion
### LightningCSS

```shell
npm install lightningcss
```

Update your `vite` configuration in `astro.config.mjs`:

```js title="astro.config.mjs" ins={4-8}
import { defineConfig } from 'astro/config'

export default defineConfig({
  vite: {
    css: {
      transformer: "lightningcss",
    },
  },
})
  
```
``````
Just updated this for Expressive code!

Also, in [the Lightning CSS docs](https://lightningcss.dev/docs.html#with-vite) they show also setting a `build` option for Vite. Do we not have to do that because Astro does the building part?

Lastly, we don't really say what any of the other packages do, so I'd suggest not having any description for now. I can see that we also don't use our package manager tabs here either, so that means this page is probably do for a bit of an overhaul anyway, and maybe we can in fact add at least a link to each package and/or a one-liner to make this a little nicer. But for now, how is this?

sarah11918 (commented):
Thanks @jasikpark ! Lightning CSS looks pretty nice, actually, so happy to have in docs! Comments below!

src/content/docs/en/guides/styling.mdx
jasikpark: I agree with your assertions, I'm entirely uncertain about how much config is required. I think in the form I suggested it just does down-leveling and not minifying? And we rely on Astro for that? Adding the entire Vite guide might be helpful.

sarah11918 (approved):
Ooh, I am noticing that the other sections show *usage*: you need to add a specific `lang` attribute to `<style>` tags.

Is anything like that required here? If so, we could show that like the other sections. If no, it might be worth a quickie explicit mention like e.g. 

> Lightning CSS will automatically optimize all your `<style>` tags.

Or whatever is accurate. (Admittedly, I'm not familiar with the term "down-leveling", and googling didn't give me any hits for this phrase with CSS. So maybe it's not standard? Is it OK just to say optimize, or is there a more standard term?)

In any case, happy to approve this as is! Also happy if you want to add a usage statement if you think that's helpful.

jasikpark:
Great, I'm trying to figure out how to say 

> Lightning CSS lets you use modern CSS features and future syntax today. Features such as CSS nesting, custom media queries, high gamut color spaces, logical properties, and new selector features are automatically converted to more compatible syntax based on your browser targets.
>
> Lightning CSS also automatically adds vendor prefixes for your browser targets, so you can keep your source code clean and repetition free.

succinctly...

sarah11918:
Yeah, as I said, cool with merging as-is!

 I don't need a description of what it DOES (none of the others have that), only if you need to do something else to USE/ENABLE it (e.g. an attribute in your `<style>` tag). It's really just, does it *work* with all that you've shown, or not? Is there literally anything else to DO?

jasikpark:
Nope! It should process all vanilla CSS 👍 

I'm going to merge in that case 👍

PR #7077, opened 2024-02-28
i18n(it): Updated every file to parity with the english version

casungo:
#### Description (required)

Updated basically every file to be true to the english version. Made it a draft since this is a pretty big change and i'm still working on some changes :)


casungo:
ok, finally fixed the deployment :P didnt realize that I translated Since

casungo:
The PR is being divided in much more small pieces to allow for easier managing and is being closed.

For future reference here is the list of such PRs:

PR #7081, opened 2024-02-28
i18n(es): update `nav.ts`

dreyfus92:
#### Description (required)

There wasn't anything to update on this file, but spotted that installation was set to be header when it isn't and fixed it. 

- Closes # N/A
- Suggested label: i18n


Waxer59 (approved):
LGTM ✅

PR #7098, opened 2024-02-28
[i18nIgnore] Updates deprecated Node.js 16 github actions

martrapp:
#### Description (required)

Get rid of those warnings:
image

In addition, I updated the action versions in the examples so that no warnings are issued when copying and pasting.


yanthomasdev (approved):
LGTM! Nice work 🙌

martrapp:
I kept actions/labeler@v4 as updating to v5 requires a massive change of the configuration. This will be a PR on its own.

PR #7108, opened 2024-02-29
Add section about RSS autodiscovery to RSS guide

### LINKED ISSUE #6802: Mention RSS Autodiscovery in RSS Guide
### 📚 Subject area/topic

RSS guide

### 📋 Suggested page

https://docs.astro.build/en/guides/rss/

Something like this, as a new section before the `Next Steps` section in the RSS guide.

> ## Enabling auto-discovery of your RSS feed
>
> RSS autodiscovery, [as explained by the RSS Advisory Board](https://www.rssboard.org/rss-autodiscovery), is "a technique that makes it possible for browsers and other software to automatically find a site's RSS feed."
> 
> All you need to do is add a `link` tag to your site's `head` element, like so:
>
> ```html
> <link
>     rel="alternate"
>     type="application/rss+xml"
>     title="Your Site's Title"
>     href={`${Astro.site}rss.xml`}
> />
> ```
> 
> After you've done this, your readers can put your site's base URL into their RSS reader to subscribe to your posts, without having to find the specific URL of your RSS feed!

_No response_

tylermercer:
Inspired by [this article](https://rknight.me/blog/please-expose-your-rss/). I figured it would be good to steer Astro users in that direction via the guide.

#### Description (required)

Add section about RSS autodiscovery to RSS guide, so that people can add auto-discovery when setting up their RSS feeds.

- Closes #6802 
- Suggested label: Add new content

Discord: `captainnemo1`


src/content/docs/en/guides/rss.mdx
```diff
@@ -230,6 +230,23 @@ rss({
...
+```html
+<link
+    rel="alternate"
+    type="application/rss+xml"
+    title="Your Site's Title"
+    href={`${Astro.site}rss.xml`}
+/>
+```
+
+After you've done this, your readers can put your site's base URL into their RSS reader to subscribe to your posts, without having to find the specific URL of your RSS feed!
```
sarah11918, line 248:
``````suggestion
## Enabling RSS feed auto-discovery

[RSS autodiscovery](https://www.rssboard.org/rss-autodiscovery) allows browsers and other software to automatically find a site's RSS feed from the main URL.

To enable, add a `<link>` tag with the following attributes to your site's `head` element:

```html
<link
    rel="alternate"
    type="application/rss+xml"
    title="Your Site's Title"
    href={`${Astro.site}rss.xml`}
/>
```

With this tag, readers of your blog can enter your site's base URL into their RSS reader to subscribe to your posts without needing the specific URL of your RSS feed.
``````
Just tightened up some of the wording to match our docs language style!

sarah11918 (commented):
Hi @tylermercer , this is an amazing addition to our docs! I am a huge RSS fan, and I think this is a very helpful inclusion to this page.

I've just edited the text to match our style a bit more, so see what you think!

src/content/docs/en/guides/rss.mdx
tylermercer: This was not at all clear from my original wording, but RSS autodiscovery works from any URL that has the meta tag—even a specific post, for example. That could be good to capture in the wording. Do you have ideas about how best to word that?

sarah11918: @tylermercer So, if I understand, even if one single page has this tag, then your ENTIRE RSS feed is discoverable?

I will also say that "not everything that is true" belongs in docs!  :smile: It is fine to keep the readers on a "happy path" with something that works, and not give them too many options that might only confuse them. So, I would lean towards not even saying anything else at all, if the above is enough guidance for them to be successful!

Is there a very specific, common use case where putting this in the head of your main site layout would not work, and you would absolutely want it somewhere else? If no, then this is fine as is, I think!

tylermercer: I meant that if this tag is in your base layout's `head`, it doesn't matter which URL from your site the person puts in their reader—it doesn't have to be the base URL like I said in my original wording. E.g. a person could drop `https://tylermercer.net/posts/software/partial-clone-git-repo/` into their reader and it would subscribe to my sitewide RSS feed.

But, point taken! We can keep it as-is; it's clear enough.

sarah11918: Ah, I get you now! Yes, I think that's extra info that takes away from the simplicity of enabling the tag. So let's get this puppy merged and helping people for the weekend!

sarah11918 (approved):
I am super happy to have this in docs, @tylermercer and super happy to welcome you to Team Docs!  :partying_face:

yanthomasdev (approved):
LGTM!

PR #7115, opened 2024-03-01
Fix incorrect list numbering in `ecommerce.mdx`

mearashadowfax:
#### Description

I’ve noticed a small inconsistency in the Snipcart configuration section of the `ecommerce.mdx` page that could be slightly confusing to translators. The step numbering within the "Basic usage" subsection does not follow a sequential order. The current numbering sequence in the markdown is 1, 2, 4, 6. For clearer readability and future maintenance of the document, it may be beneficial to update the numbering to a sequential order: 1, 2, 3, 4.

I understand that when markdown is rendered, the list numbering is corrected automatically. However, for the ease of future documentation maintenance and clarity in the markdown source, adjusting the raw markdown numbers to be sequential could help reduce any potential confusion.

### DIFF
--- a/src/content/docs/en/guides/ecommerce.mdx
+++ b/src/content/docs/en/guides/ecommerce.mdx
@@ -146,7 +146,7 @@ The following is an example of configuring a Snipcart checkout and adding button
     </script>
     ```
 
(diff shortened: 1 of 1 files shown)


Fryuni (approved):
Nice catch, thank you!

AitorMT (approved):
LGTM!

sarah11918 (approved):
You're absolutely right, this is exactly what we should do! Sorry, this was likely a leftover from shortening the example.  Thank you for fixing!

sarah11918:
!coauthor

PR #7163, opened 2024-03-03
Fix incorrect list numbering in `fonts.mdx`

mearashadowfax:
#### Description 

Corrected the numbering to be sequential (1, 2, etc. ) for improved readability and ease of future maintenance.


sarah11918 (approved):
Thank you! I do prefer this!  🚀

PR #7190, opened 2024-03-04
Update images.mdx

alexnguyennz:
#### Description (required)

- code block titles - MyImages implies a component but is in pages folder, add src/components/
- formatting - same quotes and semicolons used in each code block, closing img and Image tags, spaces
- add Output comment like other img and picture examples
- making Image or Picture components multiline formatted if they're probably too long for one line (like [other examples](https://docs.astro.build/en/guides/images/#src-required))


sarah11918:
I will have to check, but I don't think we typically show a closing tag for the regular HTML `` tag, especially to further distinguish it from the components! Hang tight!

alexnguyennz:
Hey I just had a check, there seems to be usages for both e.g.

`` on other pages:
- https://docs.astro.build/en/guides/migrate-to-astro/from-gatsby/#gatsby-image-plugin-to-astro
- https://docs.astro.build/en/guides/migrate-to-astro/from-nextjs/#next-image-plugin-to-astro
- https://docs.astro.build/en/guides/migrate-to-astro/from-nuxtjs/#nuxt-image-plugin-to-astro

`` on other pages:
- https://docs.astro.build/en/reference/api-reference/#image-
- https://docs.astro.build/en/reference/api-reference/#picture-

`` seems to always be used in a sentence or outside of code otherwise `/>` is usually used e.g. https://github.com/search?q=repo%3Awithastro%2Fastro+%3Cimg&type=code

sarah11918 (commented):
OK, thanks for your patience @alexnguyennz ! I chatted with Chris and we decided that for HTML `` elements in `.astro` files, we're NOT going to show the slash at the end. This helps emphasize and differentiate from e.g. the Image component and JSX etc. (We do like emphasizing that you use plain HTML in Astro files!)

So I've just undone those changes, but the rest is superhelpful! Thank you!

alexnguyennz:
Makes sense! Have put those through.

TheOtterlord (approved):

PR #7192, opened 2024-03-04
Update getting-started.mdx

alexnguyennz:
#### Description (required)

- add package manager tabs like on other setup pages so same access to npm/pnpm/Yarn commands
  - https://docs.astro.build/en/install/auto/#1-run-the-setup-wizard
  - https://docs.astro.build/en/tutorial/1-setup/2/#launch-the-astro-setup-wizard
- grammar/nitpicky - remove comma

Untitled

### DIFF
--- a/src/content/docs/en/getting-started.mdx
+++ b/src/content/docs/en/getting-started.mdx
@@ -31,6 +31,7 @@ import Card from '~/components/Landing/Card.astro'
 import ListCard from '~/components/Landing/ListCard.astro'
 import SplitCard from '~/components/Landing/SplitCard.astro'
 import Discord from '~/components/Landing/Discord.astro'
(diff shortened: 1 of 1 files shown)


sarah11918:
I will have to check with @delucis whether we want the package manager tabs on the front page! This page is meant to be more visual in nature, and we still have the issue that when you click a tab, it changes your scroll position, which could be a distracting user experience.

sarah11918:
OK! I finally did get a chance to get Chris's opinion, and in addition to our scroll problem, now the card is less balanced because the package manager headings make the entire left side bigger.

image

So we've decided just to keep this simple on the landing page! It was a good idea, but it messes a bit with the design so we'll just go for "looking good" here on this page.  :smile:

PR #7254, opened 2024-03-08
feat: middleware and virtual routes

lilnasy:
#### Description (required)

- Before the linked PR, a 404.astro had to exist for the middleware to run for the "not found" case.
- After it merges, astro will run the middleware for a blank page, but only if the adapter doesn't override this behavior (Cloudflare, SST, and Vercel adapters do.)
- If the user project had a `404.astro`, there is no difference.

- https://github.com/withastro/astro/pull/10206


src/content/docs/en/guides/middleware.mdx
```diff
@@ -12,9 +12,9 @@ This also allows you to set and share request-specific information across endpoi
...
 
 :::note
-On all prerendered pages, the middleware runs only during the **build process** to create static pages and does not run in the deployed website. This applies to all pages in a `static` (SSG) project. This also includes prerendered pages in `hybrid` mode (default) and any pages in `server` mode with `export const prerender = true`.
+On all prerendered pages, the middleware runs only during the **build process** to create static pages and does not run in the deployed website. This applies to all pages in a `static` (SSG) project. This also includes prerendered pages in `hybrid` mode (default) and any pages in `server` mode with `export const prerender = true`, including 404 and 500 pages.
 
-For on-demand rendered pages in `server` (default) or `hybrid` (with `export const prerender = false`) mode, the middleware runs in the deployed website when the route is requested.
+For on-demand rendered pages in `server` (default) or `hybrid` (with `export const prerender = false`) mode, the middleware runs in the deployed website when the route is requested. If a route is not found, the middleware will run for the 404 page, unless it is prerendered.
```
lilnasy, line 17:
```suggestion
For on-demand rendered pages in `server` (default) or `hybrid` (with `export const prerender = false`) mode, the middleware runs in the deployed website when the route is requested. If a route is not found, the middleware will run for the 404 page, unless it is prerendered. Some adapters may show a platform-specific 404 page instead, in which case the middleware will not run.
```

src/content/docs/en/guides/middleware.mdx
```diff
@@ -11,10 +11,15 @@ This also allows you to set and share request-specific information across endpoi
 
 Middleware is available in both SSG and SSR Astro projects.
 
-:::note
-On all prerendered pages, the middleware runs only during the **build process** to create static pages and does not run in the deployed website. This applies to all pages in a `static` (SSG) project. This also includes prerendered pages in `hybrid` mode (default) and any pages in `server` mode with `export const prerender = true`.
+### Timing of middleware execution
+By default, astro builds your project into a static site. In this [mode](https://docs.astro.build/en/basics/rendering-modes/), the middleware runs during the build as well. This allows the pages access to [`locals` object](#storing-data-in-contextlocals). Additionally, the static files can be served immediately since no server-side code is run on-request, ensuring fast response times.
```
lilnasy, line 15: Users ask "why not run middleware only on-request for static pages". So I tried to give the 2 reasons here:
- you'll need `Astro.locals`, and
- you probably don't want server-side code to run on-request if it can run ahead of time for performance.

lilnasy, line 15:
```suggestion
By default, astro builds your project into a static site. In this [mode](/en/basics/rendering-modes/), the middleware runs during the build as well. This allows the pages access to [`locals` object](#storing-data-in-contextlocals). Additionally, the static files can be served immediately since no server-side code is run on-request, ensuring fast response times.
```

sarah11918 (commented):
Hey @lilnasy ! It felt like there was more here than needed to be, and with the deleted lines and everything, it was awkward to use GitHub here to suggest changes.

So sorry, I just edited the file so I could do something different here. So, we're going to start small, and build back up with info that is truly needed. How is this for a start?

lilnasy:
It is great for a start!

Some bullet points we should get across:
- middleware will run for not-found case **even if** the user didn't author a custom 404 page (new in 4.5)
- it will NOT run for not-found case if the user's custom 404 page is prerendered (unchanged from before, still true)
- it may run twice on errors - 
first for the original on-demand rendered page that error'd, then for the on-demand rendered 500.astro page
- locals will not be available to the on-demand rendered custom 500 page if the middleware itself errors when run the second time

It's fine to leave some out if they feel too specific of details.

I'll add suggestions for these points.

src/content/docs/en/guides/middleware.mdx
```diff
@@ -200,3 +192,7 @@ greeting response
 auth response
 validation response
 ```
+
+### Handling 404 pages
+
+Middleware will run for an on-demand rendered route while Astro attempts to render a 404 page. However, it is up to the [adapter](/en/guides/server-side-rendering/) to decide whether that code runs. Some adapters may serve a platform-specific error page instead.
```
lilnasy, line 198:
```suggestion
If your custom 404 page is on-demand rendered, or if astro's default 404 page is being used, middleware will run for requests that did not match a route. However, it is up to the [adapter](/en/guides/server-side-rendering/) to decide whether that code runs. Some adapters may serve a platform-specific error page instead.
```

lilnasy: I feel a little hesitant using the term "default 404 page" because it is just a blank white page. There may not be any avoiding it, and that's fine too.

src/content/docs/en/guides/middleware.mdx
```diff
@@ -200,3 +192,7 @@ greeting response
 auth response
 validation response
 ```
+
+### Handling 404 pages
```
lilnasy, line 196:
```suggestion
## Error pages
```
It is not as recognizable as having "404" in the heading, but we don't have enough to say about 500 pages to give it its own section.

`###` to `##` because it isn't a subsection of "Chaining middleware".

src/content/docs/en/guides/middleware.mdx
lilnasy, line 198:
```suggestion
Middleware will run for an on-demand rendered route while Astro attempts to render a 404 page. However, it is up to the [adapter](/en/guides/server-side-rendering/) to decide whether that code runs. Some adapters may serve a platform-specific error page instead.

If a page fails to render due to an error, astro will attempt to render your custom 500 page. If the error page is set to render on-demand, the middleware will run a second time for that request.
```

src/content/docs/en/guides/middleware.mdx
sarah11918: Agreed!

src/content/docs/en/guides/middleware.mdx
sarah11918: I feel like my sentence is a cleaner user-facing statement -- middleware will run when...

What about something closer to:

> Middleware will attempt to run for all on-demand rendered pages, even when a matching route cannot be found. This includes Astro's default (blank) 404 page and any custom 404 pages. However, ...

sarah11918: I'm trying to figure out how *important* it is to add these two paragraphs. 500 is generic error, right? Does it *matter* that middleware runs if the page 500s? What practical value is this to the author, and would they change anything about how they write their code knowing this? In which situations would it make a difference?

Similarly, if a 500 page is being served,how much does it *matter* that they don't have `Astro.locals`? 

I'm going by the "not everything is true belongs in docs" principle, and want to include things that will affect people's decisions, and that they need to know to build properly. I'm going to need your help to figure out if these fit that criteria.

sarah11918: Updated your suggestion with min to commit. See what you think of this

src/content/docs/en/guides/middleware.mdx
sarah11918: OK, so I see this less about timing and more about "why would I run middleware on static pages?"

In that framing, I can see adding something here. Let me try something!

src/content/docs/en/guides/middleware.mdx
```diff
@@ -5,17 +5,9 @@ i18nReady: true
...
-This also allows you to set and share request-specific information across endpoints and pages by mutating a `locals` object that is available in all Astro components and API endpoints
-
-Middleware is available in both SSG and SSR Astro projects.
-
-:::note
-On all prerendered pages, the middleware runs only during the **build process** to create static pages and does not run in the deployed website. This applies to all pages in a `static` (SSG) project. This also includes prerendered pages in `hybrid` mode (default) and any pages in `server` mode with `export const prerender = true`.
-
-For on-demand rendered pages in `server` (default) or `hybrid` (with `export const prerender = false`) mode, the middleware runs in the deployed website when the route is requested.
-:::
+Middleware also allows you to set and share request-specific information across endpoints and pages by mutating a `locals` object that is available in all Astro components and API endpoints.
```
sarah11918, line 10:
```suggestion
Middleware also allows you to set and share request-specific information across endpoints and pages by mutating a `locals` object that is available in all Astro components and API endpoints. This object is available even when this middleware runs at build time, giving your static routes access to information during the lifecycle of a request.
```
What do you think about incorporating this idea here?

src/content/docs/en/guides/middleware.mdx
```diff
@@ -200,3 +192,7 @@ greeting response
 auth response
 validation response
 ```
+
+## Error pages
+
+Middleware will attempt to run for all on-demand rendered pages, even when a matching route cannot be found. This includes Astro's default (blank) 404 page and any custom 404 pages. However, it is up to the [adapter](/en/guides/server-side-rendering/) to decide whether that code runs. Some adapters may serve a platform-specific error page instead.
```
sarah11918, line 198:
```suggestion
Middleware will attempt to run for all on-demand rendered pages, even when a matching route cannot be found. This includes Astro's default (blank) 404 page and any custom 404 pages. However, it is up to the [adapter](/en/guides/server-side-rendering/) to decide whether that code runs. Some adapters may serve a platform-specific error page instead.

Middleware will also attempt to run before serving a 500 error page, including a custom 500 page, unless the server error occured in the execution of the middleware itself. If your middleware does not run successfully, then you will not have access to `Astro.locals` to render your 500 page.
```
Does something like this capture it?

sarah11918:
OK @lilnasy - I made one adjustment and then a couple of suggestions that I think might address your outstanding points? See what you think of this, whether you feel those both a) are correct 😅 and b) provide helpful information to be worth including!

Then, we'll be finished here!

sarah11918:
@arsh For the sake of having something ready for minor release tomorrow, I'm committing my suggestions, but I'm totally open to your feedback! Still at least 12 hours before I'm merging.  :joy:

If you don't have a chance to look at this before it merges with the minor release and you want to change something, then happy to PR an update to it!

src/content/docs/en/guides/middleware.mdx
lilnasy: Not that important.

The most important point would be that 500 page could miss `locals` _sometimes_. However, even that I don't feel too strongly about - it came up only once.

src/content/docs/en/guides/middleware.mdx
```diff
@@ -5,17 +5,9 @@ i18nReady: true
...
-This also allows you to set and share request-specific information across endpoints and pages by mutating a `locals` object that is available in all Astro components and API endpoints
-
-Middleware is available in both SSG and SSR Astro projects.
-
-:::note
-On all prerendered pages, the middleware runs only during the **build process** to create static pages and does not run in the deployed website. This applies to all pages in a `static` (SSG) project. This also includes prerendered pages in `hybrid` mode (default) and any pages in `server` mode with `export const prerender = true`.
-
-For on-demand rendered pages in `server` (default) or `hybrid` (with `export const prerender = false`) mode, the middleware runs in the deployed website when the route is requested.
-:::
+Middleware also allows you to set and share request-specific information across endpoints and pages by mutating a `locals` object that is available in all Astro components and API endpoints. This object is available even when this middleware runs at build time, giving your static routes access to stored information during the lifecycle of a request.
```
lilnasy, line 10:
```suggestion
Middleware also allows you to set and share request-specific information across endpoints and pages by mutating a `locals` object that is available in all Astro components and API endpoints. This object is available even when this middleware runs at build time.
```
Not blocking, but this could be confusing - there is no "real" request at build time to derive information from.

lilnasy:
Thanks for getting this all the info in a much better shape!

Just one suggestion. Even without it, looks ready to publish.

src/content/docs/en/guides/middleware.mdx
sarah11918: Love it, thank you!

PR #7284, opened 2024-03-11
[i18nIgnore] removes the video component and deletes it from the pt-br Tailwind page

sarah11918:
#### Description (required)

The Video was already removed from the English `tailwind.mxd` page. This updates the `pt-br` page to remove it, and deletes the unused component entirely.

(Was also causing a warning in the build due to deprecated usage, so this removes that for something we're not using anyway.)

Will let @yanthomasdev determine whether there are other files (e.g. captions, subtitles) that should also be deleted if we are not using this component anymore.


yanthomasdev (approved):
LGTM!

PR #7292, opened 2024-03-11
Adds a note about what internal state of elements can't be preserved by transition:persist

martrapp:
#### Description (required)

A user on discord complained that the current text gives the impression that `transition:persist` would keep all states of an element, while he suffered from restarting animations.

He referred to this sentence, which he had taken out of context
> the [...] internal state [...] will not be reset when navigating back and forth across pages 

It might help to mention the known limitations.

### DIFF
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -118,6 +118,10 @@ In the example below, the component's internal state of the count will not be re
 <Counter client:load transition:persist initialCount={5} />
 ```
 
+:::note[Known limitations]
+Not all element state can be preserved in this way. The restart of CSS animations and reloads of iframes during view transitions cannot be avoided even with transition:persist.
+:::
+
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -118,6 +118,10 @@ In the example below, the component's internal state of the count will not be re
 <Counter client:load transition:persist initialCount={5} />
 ```
 
+:::note[Known limitations]
+Not all state can be preserved in this way. The restart of CSS animations and the reload of iframes cannot be avoided during view transitions even when using transition:persist.
```
sarah11918, line 122:
```suggestion
Not all state can be preserved in this way. The restart of CSS animations and the reload of iframes cannot be avoided during view transitions even when using `transition:persist`.
```

sarah11918 (approved):
Just some code formatting, otherwise, good to merge @martrapp ! Thank you for adding this clarification!

martrapp:
Many thanks for the quick review and improvement!

PR #7307, opened 2024-03-12
Fix sidebar highlighting for studio recipe

delucis:
#### Description (required)

Fixes a highlighting bug in the docs sidebar

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/src/util/isSubPage.ts
+++ b/src/util/isSubPage.ts
@@ -26,22 +26,32 @@ const categoryIndex: Partial<Record<ReturnType<typeof getPageCategory>, string>>
 	'Error Reference': 'reference/error-reference',
 };
 
+/** Slugs of pages that appear at the top level even though they are in a sub-category. */
+const topLevelExceptions = ['recipes/studio'];
+
 /**
  * Test if `currentPage` is considered a sub-page of `parentSlug`.
  * @param currentPage The full slug for the current page, e.g. `'en/guides/rss'`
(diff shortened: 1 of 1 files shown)


TheOtterlord (commented):
State is still incorrect when visiting http://localhost:4321/en/recipes/
image

src/util/isSubPage.ts
```diff
@@ -26,22 +26,32 @@ const categoryIndex: Partial<Record<ReturnType<typeof getPageCategory>, string>>
...
  * @param parentSlug The language-less slug for the parent to test against e.g. `'guides/content-collections'`
  */
 export function isSubPage(currentPage: string, parentSlug: string): boolean {
-	// Test 1: do the two pages share a base URL segment?
+	// Test 1: is this page a known, top-level page? Don’t match its parent.
+	for (const slug of topLevelExceptions) {
+		if (currentPage.endsWith('/' + slug) && parentSlug !== slug) {
+			return false;
+		}
+	}
```
TheOtterlord, line 43: This seemed to work for me
```suggestion
		if (parentSlug === slug && !currentPage.endsWith('/' + slug)) {
			return false;
		}
	}
```

delucis: Yeah just pushed up something equivalent to this! Thanks for catching this @TheOtterlord.

TheOtterlord (approved):
LGTM locally!

PR #7334, opened 2024-03-12
i18n(es): Update dev-toolbar-app-reference.mdx

Eric-llos:
Update `dev-toolbar-app-reference.mdx` for ES translation


Eric-llos:
I've seen @AitorMT already included it :)

PR #7336, opened 2024-03-12
Add missing li element in ul

bart-nnx:
#### Description (required)

The example is missing the `<li>` element in the `<ul>`.

- Suggested label: i18n


delucis (approved):
Thanks for this fix @bart-nnx!

PR #7363, opened 2024-03-13
fix: fix broken link

aqyuki:
#### Description (required)

Fixed a broken link to the Fontsource font catalogue in the Japanese document.

No Issue has been created as it is only a fix for broken links.
- Suggested label: bug

# Additional information

In addition to the Japanese document, I found some broken links to other documents, but I didn't know whether I should put them in one PR or not, so I fixed only the links to the Japanese document. Additional corrections may be made in this PR if it is acceptable to publish them together.

### DIFF
--- a/src/content/docs/ja/guides/fonts.mdx
+++ b/src/content/docs/ja/guides/fonts.mdx
@@ -49,7 +49,7 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 
 [Fontsource](https://fontsource.org/)プロジェクトにより、Google Fontsやその他のオープンソースのフォントを簡単に使用できます。使用したいフォントをnpmモジュールとしてインストールできます。
 
-1. [Fontsourceのカタログ](https://fontsource.org/fonts)で使用したいフォントを探します。例として、ここでは[Twinkle Star](https://fontsource.org/fonts/twinkle-star)を使用します。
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Excellent! Thank you so much for taking the time to contribute this fix @aqyuki . We really appreciate it, and are happy to welcome you to Team Docs!  :partying_face:

sarah11918:
About your comment about broken links:

Are they broken in the English version? If so, we always prefer to make changes THERE. Then, our systems are in place to alert all the translators that there has been a chance, and each language typically takes care of updating their own pages. If, like this case, the English version is fine, then making a PR fix directly like this to the non-English page is perfect.

The most helpful thing to do if you find a broken link in the English version of the site is to PR the English page only. Then, when that has been merged and fixed, our tracker at https://i18n.docs.astro.build will show that other pages need updating. Then, each page in other languages can be fixed by anyone monitoring there.

We prefer smaller, individual PRs because they are very easy to review! You can fix multiple things in the same page, and sometimes 2 or 3 pages together is fine if the changes are very small, but we prefer small, self-contained PRs. This also makes it easier for someone who wants to just make one quick link fix themselves, and not update *everything* that the original PR updated, so it's easier for everyone to keep track.

Thanks again for your contribution, and for asking about our process!

aqyuki:
No, the English version of the document was fine, but there were broken links in other languages.

Thanks for the merge.

PR #7378, opened 2024-03-13
edit: move DB "Columns" section after "Tables"

bholmesdev:
#### Description (required)

There are a couple reasons I'd like to move the "columns" section after "tables:"

1. **Order of explanation.** It's impossible to describe a column without first describing a table, and this was clear from the preamble. It's also odd to show users all available column types before explaining how a table can be configured in the first place. By moving the column preamble into the "tables" section, and pushing the "columns" snippet below, the user is presented with the most minimal example for configuration before reading the details.

2. **Copy / paste warriors.** While watching a user walk through the docs ([t3dotgg](https://www.twitch.tv/theo)), they reached the "columns" section of the docs and naturally copied the snippet to build their config. Trouble is, this snippet includes a number of complex column types, while the simpler Tables snippet would make for an easier demo. By presenting the sample Tables config first, users can copy the simplest demo.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/astro-db.mdx
+++ b/src/content/docs/en/guides/astro-db.mdx
@@ -49,9 +49,36 @@ export default defineDb({
 })
 ```
 
+### Tables
+
+Data in Astro DB is stored using SQL tables. Tables structure your data into rows and columns, where columns enforce the type of each row value.
+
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
I did have my thought on first reading this, too!

PR #7396, opened 2024-03-14
new Upgrade Astro page incl semantic versioning

ematipico:
#### Description (required)

This PR adds a new page called `Upgrade Astro`. I mainly created this page because we—the platform team—needed a document that frames Astro's semantic versioning contract.

Considering the entity of the document, we felt that the docs website was the best place to have it. This document is meant for end users, integration developers, adapters developers and us - the core team. 

We recently discussed some changes we wanted to ship, and we couldn't agree. This PR will serve as:
- discussion hub to land on a unified semantic versioning contract;
- to have a broader page that involves everything about "updating" Astro.


src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,49 @@
+---
+title: Upgrade Astro
```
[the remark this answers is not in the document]
ematipico: That's on purpose. Sarah envisions this page to be a page that will contain upgrading guides.

The semantic versioning bit will be just a chapter

sarah11918:
Thanks @ematipico ! I will be looking into this over the next few days. We will probably wait to merge it until we can add more content to make this a great, general "Update" page... but we'll get on it!

Some of the things I have in mind include:

- very clearly displaying the current version of Astro and instructions to update version (and/or to downgrade/install any particular version)
- link to changelog
- links to all past migration guides
- ... whatever else we think of! ideas welcome!

sarah11918:
@ematipico I took a first pass at Semantic versioning editing, and added my own TODOS for the rest of the page. You are welcome to review/edit/fix any of my changes so far!

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+### Major changes
+
+Major releases will include breaking changes to at least some existing code. These breaking changes are always documented in an "Upgrade to vX" guide in Astro.
+
+Major releases allow Astro to make significant changes not only to internal logic, but also in intended behavior and usage. Documentation will be updated to reflect the latest version only, and a **static, unmaintained snapshot of older docs** are available as a historical record for older projects that are not yet upgraded.
+
+Major releases include changes such as:
+
+- Upgrade of minimal required Node.js version (minor or major), e.g. from `v18.4.0` to `v20.0.0`.
```
Fryuni, line 65: A minor bump in Node can be both a minor and a major for us?

ematipico: Yes. That's the controversial part that started the discussion, which ended up with the creation of this document. Usually, we want to try to keep Node.js **minor** bumps to a major release, but some situations came up that we felt needed clarification.

For example, the latest version of `sharp` requires a Node.js version that is more recent than our minimal version. Updating to that `sharp` version would allow us to fix some nasty workarounds and bugs.

Fryuni: I don't see any problem in bumping a minor in a minor in general. A minor version guarantees no breaking changes in the contract between the consumer and the thing being versioned. And that is upheld when bumping a dependency, one upgrade requiring another and when both are applied and both are minor the consumer project should not break.

We can't make guarantees about anything else in the project, just between the project and Astro itself. A bump in a patch version of a dependency can be a breaking change because some other dependency outside of Astro may no longer be de-duplicated and it's types won't match up. That doesn't affect the semver guarantees of Astro, it is the same as relying on an implementation detail

VoxelMC (commented):
Great work! I think this will clear up a lot of confusion that I see around. Here are a couple of tiny notes (that I took a few days ago and never submitted...).

src/i18n/en/nav.ts
```diff
@@ -20,6 +20,7 @@ export default [
 	{ text: 'Installation', slug: 'install/auto', key: 'install' },
 	{ text: 'Editor Setup', slug: 'editor-setup', key: 'editor-setup' },
 	{ text: 'Upgrade to v4', slug: 'guides/upgrade-to/v4', key: 'guides/upgrade-to/v4' },
+	{ text: 'Upgrade Astro', slug: 'upgrade-astro', key: 'upgrade-astro' },
```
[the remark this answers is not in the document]
dreyfus92: The semantic versioning bit will be just a chapter of what Sarah envisions for this page that will contain upgrading guides.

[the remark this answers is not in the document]
sarah11918: Yeah, if you look at all my TODOs, I think the *most* use people will get out of this page is actually *how* to upgrade. (Sorry Ema 😅 ) The versioning stuff is great to give more context *about* upgrading, but for example, right now if we want to show someone how to upgrade their version of Astro, we have to send them to the current migration guide... which is not good! That should be RIGHT ON THE BOX.  :smile:

sarah11918:
Latest version open for reviews!

NOTE: Still to-do will be to remove the specific link "Upgrade to v4" from the sidebar

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+Patches may be released for reasons such as:
+
+- Re-releases after a failed release.
+- Internal changes that do not change Astro’s functionality:
+  - refactors
+  - performance improvements
+  - increase or change in test coverage
+  - aligning with stated documentation and expected behavior
+- Improvements to logging and error messages.
```
Princesseuh, line 141:
```suggestion
- Internal changes that do not change Astro’s functionality:
  - refactors
  - performance improvements
  - increase or change in test coverage
  - aligning with stated documentation and expected behavior
- Improvements to logging and error messages.
- Re-releases after a failed release.
```

Princesseuh: This order made more sense to me, since the vast majority of changes are the first point.

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+### Minor changes
+
+
+Minor releases primarily introduce new features and improvements that you may wish to try, but require no changes to your code. Some existing features may also be **deprecated** (marked for deletion in a future version while continuing to function) in a minor release, giving you the opportunity to prepare for their eventual removal.
+
+Minor releases include changes such as:
+
+- **Deprecations** of existing features/options.
+- Introduction of new functionalities.
+- Upgrade of minimal required Node.js version (minor or patch), e.g. from `v18.4.0` to `v18.7.0`.
```
bluwy, line 155:
```suggestion
- Upgrade of minimum required Node.js version (minor or patch), e.g. from `v18.4.0` to `v18.7.0`.
```

I _think_ minimum fits better than minimal here

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+### Major changes
+
+Major releases will include breaking changes to at least some existing code. These breaking changes are always documented in an "Upgrade to vX" guide in Astro.
+
+Major releases allow Astro to make significant changes not only to internal logic, but also in intended behavior and usage. Documentation will be updated to reflect the latest version only, and a **static, unmaintained snapshot of older docs** are available as a historical record for older projects that are not yet upgraded.
+
+Major releases include changes such as:
+
+- Upgrade of minimal required Node.js version (minor or major), e.g. from `v18.4.0` to `v20.0.0`.
```
bluwy, line 172:
```suggestion
- Upgrade of minimum required Node.js version (minor or major), e.g. from `v18.4.0` to `v20.0.0`.
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+- Removal of previously deprecated functionalities.
+- Changes of existing functionalities.
+- Changes of existing options in the integration hooks.
+- Changes of existing options and functionalities in `astro/app`, notably used for creating new adapters.
+
+A major release may also include some non-breaking changes and improvements that would normally be released separately in a minor or patch release.
+
+### Exceptions
+
+- **Experimental features** are exempt from semantic versioning. This allows Astro developers the greatest flexibility to explore, and even radically change course, during development. Therefore, the behaviour  of these features can break in minor and patch changes.
```
Princesseuh, line 182:
```suggestion
- **Experimental features** are exempt from semantic versioning. This allows Astro developers the greatest flexibility to explore, and even radically change course, during development. Therefore, the behaviour of these features can break in minor and patch changes.
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+  These features are usually accompanied by ongoing, public [Request for Consideration (RFC) stage 3](https://github.com/withastro/roadmap#stage-3-rfc--development). It is expected that beta users will follow for updates, and leave early feedback to help guide development of these features.
+
+  Once these features are out of their experimentation period, they will follow the normal semantic versioning contract.
+
+- **Improvements to the documentation** (e.g. reference and error messages) are also exempt from semantic versioning. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repo.
```
Princesseuh, line 188: I would say that mentioning this is unnecessary, as the previous paragraphs already either imply or purposely say this.

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+Minor releases primarily introduce new features and improvements that you may wish to try, but require no changes to your code. Some existing features may also be **deprecated** (marked for deletion in a future version while continuing to function) in a minor release, giving you the opportunity to prepare for their eventual removal.
+
+Minor releases include changes such as:
+
+- **Deprecations** of existing features/options.
+- Introduction of new functionalities.
+- Upgrade of minimal required Node.js version (minor or patch), e.g. from `v18.4.0` to `v18.7.0`.
+  :::note
+  We take these kind of changes very seriously, and approve them only if we think that a new version of Node.js can improve the DX of our users, and impact core functionalities.
```
bluwy, line 157: Discussion: Something I proposed before bumping the node version is perhaps to issue a warning from the CLI if they are using the soon-to-be-dropped nodejs versions. This way users will see warnings a couple minors before before we bump it. I'm not sure if we want to do this, but if you think this is a good idea, maybe we can note about this here. But if not (for now), I'm also fine 😄

bluwy (approved):
The general flow of it looks great to me 👍

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+  :::
```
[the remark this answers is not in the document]
ematipico: > Should a minor be able to deprecate the usage of an unsupported version of Node? For instance, should 4.6 be able to formally remove support for Node 19, which is completely unsupported by Node. (and overall is IIRC, less than 1% of our users, but part of the reason we can't update to latest Sharp)

I think so, as long as this is justified and in line with this document. If relying on Node.js 19 is a security hazard, then `4.6` is entitled to drop that Node.js version

src/content/docs/en/upgrade-astro.mdx
ematipico: I don't mind the warning, although we have to consider that this is another "chore" to take into consideration between releases. If there's an automated way to do so, that would be great

src/content/docs/en/upgrade-astro.mdx
ematipico: Done!

src/content/docs/en/upgrade-astro.mdx
bluwy: Yeah we'd have to remind ourselves if we do this, but given that we'd rarely do this unless we need to, I think it shouldn't take too much work. e.g. add the semver warnings, add a linear ticket to remind at a certain time, remove the semver warnings and update versions.

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+Documentation for older versions of Astro is not maintained, but is available as a static snapshot. Use these versions of docs if you are unable to upgrade your project, but still wish to consult guides and reference:
+
+- [unmaintained v3.6.3 snapshot](https://docs-git-v3-docs-unmaintained-astrodotbuild.vercel.app/)
+- [unmaintained v2.10.15 snapshot](https://deploy-preview-4405--astro-docs-2.netlify.app/en/getting-started/)
+
+
+## Semantic versioning
+
+Astro attempts to adhere as much as possible to [semantic versioning](https://semver.org/), which is a set of rules developers use to determine how to assign a version number to a release. Semantic version follows predictable to inform users of the kind of changes they can expect from one version to the next.
```
sarah11918, line 123:
```suggestion
Astro attempts to adhere as much as possible to [semantic versioning](https://semver.org/), which is a set of rules developers use to determine how to assign a version number to a release. Semantic version follows a predictable pattern to inform users of the kind of changes they can expect from one version to the next.
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+- Internal changes that do not change Astro’s functionality:
+  - refactors
+  - performance improvements
+  - increase or change in test coverage
+  - aligning with stated documentation and expected behavior
+- Improvements to logging and error messages.
+- Re-releases after a failed release.
+
+Patch changes also include bug fixes, even in cases where users were taking advantage of existing unintended or undesirable behavior.
```
sarah11918, line 143:
```suggestion
Patch changes also include **most bug fixes**, even in cases where users were taking advantage of existing unintended or undesirable behavior.
```
Drawing attention to bug fixes here!

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+
+### Minor changes
+
+
```
sarah11918, line 148:
```suggestion
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+  We take these kind of changes very seriously, and approve them only if we think that a new version of Node.js can improve the security or developer experience of our users, and impact core functionalities.
+  :::
+- Introduction of new options in the integration hooks.
+- Introduction of new functionalities in `astro/app`, notably used for creating new adapters.
+
+A minor release may also include smaller, patch changes at the same time.
+
+### Major changes
+
+Major releases will include breaking changes to at least some existing code. These breaking changes are always documented in an "Upgrade to vX" guide in Astro.
```
sarah11918, line 166:
```suggestion
Major releases will include breaking changes to at least some existing code. These breaking changes are always documented in an ["Upgrade to vX" guide](#upgrade-guides) in Astro.
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+- Removal of previously deprecated functionalities.
+- Changes of existing functionalities.
+- Changes of existing options in the integration hooks.
+- Changes of existing options and functionalities in `astro/app`, notably used for creating new adapters.
+
+A major release may also include some non-breaking changes and improvements that would normally be released separately in a minor or patch release.
+
+### Exceptions
+
+- **Experimental features**. This allows Astro developers the greatest flexibility to explore, and even radically change course, during development. Therefore, the behaviour of these features can break in minor and patch changes.
```
sarah11918, line 182:
```suggestion
- **Experimental features**. Releasing versions of Astro without adhering to semantic versioning allows Astro developers the greatest flexibility to explore, and even radically change course, during the development of experimental features. Therefore, the behavior of these features can break in minor and patch changes.
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+### Minor changes
+
+
+Minor releases primarily introduce new features and improvements that you may wish to try, but require no changes to your code. Some existing features may also be **deprecated** (marked for deletion in a future version while continuing to function) in a minor release, giving you the opportunity to prepare for their eventual removal.
+
+Minor releases include changes such as:
+
+- **Deprecations** of existing features/options.
+- Introduction of new functionalities.
+- Upgrade of minimum required Node.js version (minor or patch), e.g. from `v18.4.0` to `v18.7.0`.
```
matthewp, line 155: Want to flag this one, I don't think we've done this in the past as a normal thing in a minor release. It's possible that we have. Before we document this as a fact I would like to know:

* Have we ever bumped the minimum Node.js version in the past and why did we do it?

My understanding of Semver is that this would be a breaking change and not allowed in a minor release; you are using `18.4.0` and Astro `4.1.0` and Astro releases `4.2.0` and now your builds break; that shouldn't happen.

If there's a very good reason to do so then I think we can make this change but would like to discuss them first.

Princesseuh: We have never bumped a Node version outside of a major before. The motivation for this change is that right now we're blocked by a very small amount (10-15% depending how you filter the telemetry) of users being on versions that are unsupported by Node themselves (Node only supports the latest minor of extended maintenances releases and even releases)

No production system actually runs those outdated versions, they all run the latest of every major. It's only local dev machines that are outdated (and vulnerable to security issues, since Node doesn't backport security fixes, they expect you to be on the latest of the extended maintenance). It's unlikely that those users are unable to upgrade, as even old LTS of Linux systems will have the latest 

What I suggested (somewhere... I don't remember where) is to allow ourselves to change the minimum minor of *only* the extended maintenance (For Astro 4.x, that'd be requiring to latest minor of 18) and perhaps the LTS.

matthewp: Ok, I think that's a reasonable direction to go. Would still love @bluwy's input here as well.

If we do decide on this policy, I think we should state that explicitly and give the reason why. The current text (below this) says:

> approve them only if we think that a new version of Node.js can improve the security or developer experience of our users, and impact core functionalities.

this gives us too much wiggle room in my opinion. I think this is one case where we should be more explicit to give users more confidence that we're not going to change the version for arbitrary reasons.

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
+import Version from '~/components/Version.astro';
+
+
+This guide covers how to update your version of Astro and related dependencies, how to learn what has changed from one version to the next, and how to understand Astro's versioning system and corresponding documentation updates.
+
+## What has changed?
+The latest release of Astro is <Version pkgName="astro" />.
+
+You can find an exhaustive list of all changes in [Astro's changelog](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md), and important instructions for upgrading to each new [major version](#major-changes) in our [upgrade guides](#upgrade-guides).
```
SatanshuMishra, line 15: Perhaps this is suggestion needs it's own separate discussion (to ensure consistency throughout the docs). Thoughts on making the external link to Astro's Changelog open in a new tab? 

Since markdown doesn't have a built in way to do this, a simple component could be created to implement this. While this is a subjective suggestion, personally, I do find it a bit annoying to have the link open in the current page.

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+  ```
+  </Fragment>
+</PackageManagerTabs>
+
+
+## Documentation updates
+
+This documentation is updated for each [minor release](#minor-changes) and [major version release](#major-changes). When new features are added, or existing usage changes, the docs will update to reflect the **current behavior of Astro**. If your project is not updated, then you may notice some behaviors do not match the up-to-date documentation. 
+
+New features are added to docs with the specific version number in which they were added. This means that if you have not updated to the latest release of Astro, some documented features may be unavailable. Always check the `Added in:` version number and make sure your project is updated before attempting to use new features!
```
SatanshuMishra, line 96: Possible improvement to the wording of this sentence:
```suggestion
New features are added to docs with the specific version number in which they were added. This means that if you have not updated to the latest release of Astro, some documented features may be unavailable. Always check the `Added in:` version number to ensure your project version is compatible with the features you want to use.
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+### Exceptions
+
+- **Experimental features**. This allows Astro developers the greatest flexibility to explore, and even radically change course, during development. Therefore, the behaviour of these features can break in minor and patch changes.
+
+  These features are usually accompanied by ongoing, public [Request for Consideration (RFC) stage 3](https://github.com/withastro/roadmap#stage-3-rfc--development). It is expected that beta users will follow for updates, and leave early feedback to help guide development of these features.
+
+  Once these features are out of their experimentation period, they will follow the normal semantic versioning contract.
+
+- **Improvements to the documentation** (e.g. reference and error messages). They are built from source from the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repo.
```
SatanshuMishra, line 188: Ensuring consistency.
```suggestion
- **Improvements to the documentation** (e.g. reference and error messages). They are built from source from the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
```

SatanshuMishra (commented):
The page looks great and will definitely be a helpful addition to the docs! Added a few suggestions :)!

src/content/docs/en/upgrade-astro.mdx
natemoo-re: I agree with that! These cases should be super explicit and hopefully rare. Not really sure how best to communicate the policy on that, but I'd hate for us to develop a reputation like TypeScript where "oh they don't actually follow SemVer"

ematipico: @matthewp and @natemoo-re, could you please provide some suggestions on how to phrase this?

bluwy: I'm also fine if we really need to bump the node version if there's a big benefit for most users. I suggested emitting warnings beforehand to help soften the impact: https://github.com/withastro/docs/pull/7396#discussion_r1530685955, but otherwise I think projects can bump the node version as long as they document this possibility beforehand.

src/content/docs/en/upgrade-astro.mdx
sarah11918: Our practice is to never open links in new tabs for consistency and accessibility. It's generally considered better practice to allow users to control that behaviour (everyone can choose to open in a new tab, but if you force a new tab, you take that control away from them).

src/content/docs/en/upgrade-astro.mdx
sarah11918: I think both versions are pretty comparable! In this specific context, on a page about upgrading/updating and how we document *new* features, I *do* slightly prefer the extra nudge to update their project to use new features, but I don't have a *strong* preference. Yours uses *slightly* move elevated language, which can be *slightly* more work to translate properly, or a bit of friction for people reading in their second language, but it's certainly a fine edit!

Is there a problem you're attempting to fix here, so I know the motivation and can better decide between the two wordings?

src/content/docs/en/upgrade-astro.mdx
sarah11918: I will note that often we *intentionally* switch between equivalent words for sentence variety which helps for readability! Also, sometimes a shorter version of a word works better for rhythm in longer sentences and vice versa. 

No problems changing here, but just pointing out that 100% consistency isn't *always* the top priority!

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,187 @@
...
+- **Deprecations** of existing features/options.
+- Introduction of new functionalities.
+- Upgrade of minimum required Node.js version (minor or patch), e.g. from `v18.4.0` to `v18.7.0`.
+  :::note
+  We take these kind of changes very seriously, and approve them only if we think that a new version of Node.js can improve the security or developer experience of our users, and impact core functionalities.
+  :::
+- Introduction of new options in the integration hooks.
+- Introduction of new functionalities in `astro/app`, notably used for creating new adapters.
+
+A minor release may also include smaller, patch changes at the same time.
```
sarah11918, line 161:
```suggestion
Minor releases include changes such as:

- **Deprecations** of existing features/options with a warning that they will be removed in an upcoming major release.
- Introduction of new functionalities.
- Introduction of new options in the integration hooks.
- Introduction of new functionalities in `astro/app`, notably used for creating new adapters.

:::note
Minor releases are meant to include only non-breaking changes.  However, *in rare cases*, Astro will include an upgrade of the minimum required Node.js version in a minor release (e.g. from `v18.4.0` to `v18.7.0`).

Astro is committed to adhering to semantic versioning, and will include an upgrade to a new version of Node.js in a minor release after careful analysis of a variety of factors, including the number of Astro projects running on incompatible versions, the security risk to developers of supporting an older version of Node.js, and the impact on Astro's core functionalities.
:::

A minor release may also include smaller, patch changes at the same time.
```

So, here's a suggested edit to start from! I'm not sure whether this is still too handwavy, BUT:
- I moved this rare edge case to the end, since we don't really need to stick something that almost never happens in the middle of a list
- I put the whole thing in the note, since the list doesn't have to be exhaustive (include changes such as...) and this is *all* kind of an aside. This gives you more flexibility in how you discuss it.

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,189 @@
...
+
+Minor releases include changes such as:
+
+- **Deprecations** of existing features/options with a warning that they will be removed in an upcoming major release.
+- Introduction of new functionalities.
+- Introduction of new options in the integration hooks.
+- Introduction of new functionalities in `astro/app`, notably used for creating new adapters.
+
+:::note
+Minor releases are meant to include only non-breaking changes.  However, *in rare cases*, Astro will include an upgrade of the minimum required Node.js version in a minor release (e.g. from `v18.4.0` to `v18.7.0`).
```
sarah11918, line 158:
```suggestion
Minor releases are meant to include only non-breaking changes. However, *in rare cases* Astro will include an upgrade of the minimum required Node.js version in a minor release (e.g. from `v18.4.0` to `v18.7.0`).
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,189 @@
...
+- **Deprecations** of existing features/options with a warning that they will be removed in an upcoming major release.
+- Introduction of new functionalities.
+- Introduction of new options in the integration hooks.
+- Introduction of new functionalities in `astro/app`, notably used for creating new adapters.
+
+:::note
+Minor releases are meant to include only non-breaking changes. However, *in rare cases* Astro will include an upgrade of the minimum required Node.js version in a minor release (e.g. from `v18.4.0` to `v18.7.0`).
+
+Astro is committed to adhering to semantic versioning, and will include an upgrade to a new version of Node.js in a minor release after careful analysis of a variety of factors, including the number of Astro projects running on incompatible versions, the security risk to developers of supporting an older version of Node.js, and the impact on Astro's core functionalities.
+:::
```
sarah11918, line 161:
```suggestion
:::note
Minor releases are meant to include only non-breaking changes. However, *in rare cases* Astro will include an upgrade of the minimum required Node.js version in a minor release (e.g. from `v18.4.0` to `v18.7.0`).

Astro is committed to adhering to semantic versioning and will include an upgrade to a new version of Node.js in a minor release **only after careful analysis of a variety of factors** including, but not limited to, the number of Astro projects running on incompatible versions, the security risk to developers of supporting an older version of Node.js, and the impact on Astro's core functionalities.
:::
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,189 @@
...
+
+### Exceptions
+
+- **Experimental features**. Releasing versions of Astro without adhering to semantic versioning allows Astro developers the greatest flexibility to explore, and even radically change course, during the development of experimental features. Therefore, the behavior of these features can break in minor and patch changes.
+
+  These features are usually accompanied by ongoing, public [Request for Consideration (RFC) stage 3](https://github.com/withastro/roadmap#stage-3-rfc--development). It is expected that beta users will follow for updates, and leave early feedback to help guide development of these features.
+
+  Once these features are out of their experimentation period, they will follow the normal semantic versioning contract.
+
+- **Improvements to the documentation** (e.g. reference and error messages). They are built from source from the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
```
sarah11918, line 189:
```suggestion
- **Improvements to the documentation** (e.g. reference and error messages). They are built from source for the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
```
I think this should be *for* the docs repository, because they're not built *from* the docs repository. They're pulled in.

liruifengv:
Great work!

About deprecated APIs, our strategy seems to be to keep one major version each time? 
Until the second major version completely removes it?

Do you think it needs to be stated on this page?

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -0,0 +1,182 @@
...
+
+A major release may also include some non-breaking changes and improvements that would normally be released separately in a minor or patch release.
+
+### Exceptions
+
+- **Experimental features**. Releasing versions of Astro without adhering to semantic versioning allows Astro developers the greatest flexibility to explore, and even radically change course, during the development of experimental features. Therefore, the behavior of these features can break in minor and patch changes.
+
+  These features are usually accompanied by ongoing, public [Request for Consideration (RFC) stage 3](https://github.com/withastro/roadmap#stage-3-rfc--development). It is expected that beta users will follow for updates, and leave early feedback to help guide development of these features.
+
+  Once these features are out of their experimentation period, they will follow the normal semantic versioning contract.
```
sarah11918, line 180:
```suggestion
  These features are usually accompanied by an ongoing, public [Request for Consideration (RFC) stage 3](https://github.com/withastro/roadmap#stage-3-rfc--development). It is expected that beta users will follow for updates, and leave early feedback on the discussion to help guide development of these features.

  Once these features are out of their experimental period, they will follow the normal semantic versioning contract.
```

src/content/docs/en/upgrade-astro.mdx
SatanshuMishra: Sounds good! Will keep this in mind moving forward :)!

src/content/docs/en/upgrade-astro.mdx
SatanshuMishra: My main goal with this change was to be a bit more clear about that new features meant. Someone could be a version behind from where a feature was introduced but that feature doesn't necessarily have to be in the latest version of Astro. Rewording it to `compatible with...` makes it a bit more clear in my mind.

This is a very minor change and as you mentioned, the two versions are very comparable. I think we could go either way.

src/content/docs/en/upgrade-astro.mdx
SatanshuMishra: Good to know! Thank you for clarifying this :)!

src/content/docs/en/upgrade-astro.mdx
sarah11918: s nAgreed, they are comparable and my point here was to keep with the theme of upgrading/new features. So, there's nothing wrong with your suggestion, and on a more generic page I probably would have used it! That's why I asked exactly what problem you were trying to solve, to see what I might be missing.

I think I'll keep this nuance as is for now, but I do appreciate these checks for clarity! :purple_heart:

casungo (approved):
LGTM!

yanthomasdev:
🚨 **ATTENTION** 🚨 

This one shouldn't be `i18nIgnore`'d, paste this into the merge commit's description and Lunaria will do its thing:

```
@lunaria-ignore:src/i18n/**/nav.ts

@tracker-minor:src/i18n/**/nav.ts;./src/i18n/**/nav.ts;
```

PR #7405, opened 2024-03-14
docs: fix typo in `configuration-reference.mdx`

jsparkdev:
#### Description (required)

- fix typo in `configuration-reference.mdx`

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/configuration-reference.mdx
+++ b/src/content/docs/en/reference/configuration-reference.mdx
@@ -512,7 +512,9 @@ To fetch all assets uploaded to the same domain (e.g. `https://cdn.example.com/_
 }
 ```
 
-**Added in:** `astro@4.5.0`
+<p>
+  <Since v="4.5.0" />
+</p>
 
 You can also pass an object to `assetsPrefix` to specify a different domain for each file type.
 In this case, a `fallback` property is required and will be used by default for any other files.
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @jsparkdev , unfortunately, this is the best we can do here!  :sweat_smile: 

This file is generated by JSDoc in the main Astro repo, and so we use a JSDoc heading which gets converted to the Since component when it's pulled in here. However, we don't have access to that heading partway through a description, so I did this manually to best approximate that a certain aspect of this feature (passing an object instead of just a string) was added in 4.5. There is no other good way to do this, so once again, Sarah :hotdog: 'd it in!

jsparkdev:
Thanks for the great explanation! It was helpful.

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -1161,7 +1163,7 @@ and deduplicating scripts). You can also now conditionally render scripts in you
 However, this means scripts are no longer hoisted to the `<head>` and multiple scripts on a page are no longer bundled together.
 If you enable this option, you should check that all your `<script>` tags behave as expected.
 
-This option will be enabled by default in Astro 5.0.
+This option will be enabled by default in Astro 4.5.0.
```
jsparkdev, line 1166: How about this part?

If the PR for this part is correct, I would like to open a new PR.

PR #7423, opened 2024-03-16
Added `prerender = false` option to rendering-modes.mdx documentation

marceloverdijk:
#### Description (required)

Added `prerender = false` to documentation to opt-out of pre-rendering in Hybrid mode.

marceloverdijk

### DIFF
--- a/src/content/docs/en/basics/rendering-modes.mdx
+++ b/src/content/docs/en/basics/rendering-modes.mdx
@@ -47,6 +47,38 @@ Since they are generated per visit, these routes can be customized for each view
 
 - **Frequently changing content**: Generate individual pages without requiring a static rebuild of your site. This is useful when the content of a page updates frequently, for example displaying data from an API called dynamically with `fetch()`.
 
+When using `hybrid` output mode, pages will still be pre-rendered by default. Top opt-out of pre-rendering add `export const prerender =  false` to any page or endpoint.
+
+```astro
+---
+export const prerender = false;
+
+if (Astro.request.method === 'POST') {
+  // handle form submission
+}
+---
+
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @marceloverdijk ! We do cover this content in detail on this page: https://docs.astro.build/en/guides/server-side-rendering/

For the Basics section, we really are just making the reader aware of the various output modes so they understand what they are and have a vocabulary going forward. We state that `output: 'hybrid'` is for mostly static sites with some on-demand routes without getting into more detail at this point. We link to the other page with code samples in the more complete guide for people who are actually ready to install an adapter and configure their pages.

So, this information is actually a little too specific for this introductory page, and we don't like to include things like exact usage/code snippets on multiple pages for maintenance: if something changes in actual usage, now we have multiple pages we have to update.

If you'd like to create an issue to discuss an improvement to this page, you are welcome to! That's often the best way to make sure that your ideas are "on the same page" ( :smile: ) as the docs team before you make a PR. But, this is not content to add to this page according to our overall docs structure/intentions, so I'll be closing this PR, but we hope to see you in docs again!

marceloverdijk:
thx @sarah11918 I probably missed the other page.

Maybe good to only add the following then:

```
When using `hybrid` output mode, pages will still be pre-rendered by default. Top opt-out of pre-rendering add `export const prerender =  false` to any page or endpoint.
```

If so I can update the PR...

sarah11918:
Hi! Like I said, you are welcome to open an issue to discuss problems with this page but it's not my intention to add this statement about specific code usage. If you'd like to describe problems with the page, an issue is the place for that!

Issue #7376, opened 2024-03-13
Add Bun to the Package Manager Tabs across the docs


Dax911:
### 📚 Subject area/topic

Package Managers

### 📋 Suggested page

https://docs.astro.build/en/install/auto/
https://docs.astro.build/en/editor-setup
... most pages tbh

### 📋 General description or bullet points (if proposing new content)

Just adding the bun commands to the documentation.

Dax911:
Working on the changes [here](https://github.com/Dax911/docs.git) if this is something the team is interested in.

TheOtterlord:
Hi @dax911! We know Bun is working on supporting Astro, but Bun is not officially supported by Astro, so we are not including it in our documentation at this time. We do have a user-submitted recipe for [using Bun with Astro](https://docs.astro.build/en/recipes/bun/), and if there is an error on that page you are welcome to PR a correction! But we've decided that for the time being, we will not be adding additional package managers or runtimes to the docs.

PR #7425, opened 2024-03-16
Add Bun to installation options in DB docs

Dillonpw:
Added bun install fragment to new DB docs

#### Description (required)

Add a fragment for bun install step in the DB documentation

- Closes #
- Suggested label:


TheOtterlord:
Hi @Dillonpw! Thanks for the PR!

We've decided that for the time being, we will not be adding additional package managers to the docs. We talk a little more about this in #7376

PR #7460, opened 2024-03-18
 Add a note for using `id` instead of `slug` when using type: 'data'

KoriIku:
#### Description

The `slug` property only exists when type is set to `content`. If type is set to `data`, using the example code will result in an error indicating that slug must be a string. 

While the "Custom Slugs" and "Filtering Collection Queries" sections of this document mention that `slug` exists only in `content`, a reminder here may help beginners who overlook this point avoid trouble.


sarah11918:
Hi @KoriIku ! Thanks for providing some helpful extra clarification for the docs on data collections!
 
Can you please remove the translation file in this PR? When content changes are proposed to the English site, the PR should contain *only* a change to the English file, which is treated as the source file for our i18n tracking system.

Only after the English is reviewed, edited, and approved do we consider changes ready for translation. Even small changes to the English site are often revised before publishing, so we ask that you not worry about submitting a translation at the same time, to save you time, and because we have a system for translations already!  :raised_hands:

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -482,7 +482,7 @@ The exact method for generating routes will depend on your build [`output`](/en/
 
 If you are building a static website (Astro's default behavior), you would use the [`getStaticPaths()`](/en/reference/api-reference/#getstaticpaths) function to create multiple pages from a single `src/pages/` component during your build.
 
-Call [`getCollection()`](/en/reference/api-reference/#getcollection) inside of `getStaticPaths()` to query your content. Then, create your new URL paths using the `slug` property of each content entry.
+Call `getCollection()` inside of `getStaticPaths()` to [query your content or data collection](https://docs.astro.build/en/guides/content-collections/#querying-collections). Then, create your new URL paths using the `slug` property (content collections) or `id` property (data collections) of each content entry.
```
sarah11918, line 485:
```suggestion
Call `getCollection()` inside of `getStaticPaths()` to [query your content or data collection](/en/guides/content-collections/#querying-collections). Then, create your new URL paths using the `slug` property (content collections) or `id` property (data collections) of each content entry.
```

sarah11918:
Hi @KoriIku ! I'm so sorry that it took me so long to get back to this one! Thank you for your patience!

We always prefer to not add extra notes/tips/cautions etc. for regular information that should be part of the text or instruction, so I was waiting until I could look at the page more closely and figure out how to best add this content inside the regular text.

See what you think of this instead! I think this makes the properties clear, and also links back to where on the page the properties are listed out, so that someone can find for themselves where it's mentioned that data collections use `id`. Are you happy with this change instead?

KoriIku:
Thank you for your revisions, it does indeed look better now.

sarah11918 (approved):
Thank you again for this helpful extra content @KoriIku ! I think this section is indeed more helpful now, and I'm so glad you contributed it! :raised_hands:

PR #7461, opened 2024-03-18
i18n(zh-cn): Update framework-components.mdx

KoriIku:
#### Description (required)

The block for the tip is missing a colon.


liruifengv (approved):
LGTM. Nice catch!

PR #7462, opened 2024-03-18
added a tip for markdown with `layout` frontmatter

liruifengv:
#### Description (required)

added a warning tip for use `image()` helper

- https://github.com/withastro/astro/issues/10468


sarah11918:
Hi @liruifengv!

Let's discuss this, because I think the fix for docs is actually to emphasize in *content collections* that you are no longer using the layout property anymore, because your Markdown/MDX files are no longer actual pages (that can control their own layout). In content collections, these files now only provide some content and metadata.

See the Step 10 here: https://docs.astro.build/en/tutorials/add-content-collections/#generate-pages-from-a-collection I think this is the real issue, that your frontmatter no longer handles your layout. Instead, your `[slug].astro` file does.

So, I don't see the solution as something for the image page, but instead something that makes it clear that your Markdown files should NOT attempt to use a layout when they are entries in a collection, because they are no longer a page that can/does control that.

What do you think? Is there a spot this could be clearer in one of these pages: content collections, Markdown/MDX guide, or even Pages (where we discuss markdown pages) because not every markdown file *is* a "Markdown page" -- only the ones in `src/pages/` are. Those last two pages were written before content collections existed, so it makes sense that they don't fully explain the difference when *collections* are involved.

sarah11918 (commented):
Thank you for adding this helpful content @liruifengv ! It took me a while, but what do you think about my suggestion instead? This emphasizes right upfront that the property is only available for Markdown/MDX *pages* (with a link and an extra definition).

Although it doesn't say you can't use this with content collections, I think it makes it clear about when you CAN use it. What do you think?

liruifengv:
> Thank you for adding this helpful content @liruifengv ! It took me a while, but what do you think about my suggestion instead? This emphasizes right upfront that the property is only available for Markdown/MDX *pages* (with a link and an extra definition).
> 
> 
> 
> Although it doesn't say you can't use this with content collections, I think it makes it clear about when you CAN use it. What do you think?

Looks great!

sarah11918 (approved):
LGTM! Thank you, this is really helpful content to have !

PR #7471, opened 2024-03-18
i18n(fr) Wrong code example on rss feed generation

borelkoumo:
#### Description (required)
In the tutorial, we are using this code to generate the rss feed
[Create an .xml feed document](https://docs.astro.build/en/tutorial/5-astro-api/4/#create-an-xml-feed-document)

image

Running `npm run build` with above code from the tutorial outputs a WARN:

image

Changes made in this PR fix that, and allows the rss.xml file to be correctly generated


src/content/docs/fr/tutorial/5-astro-api/4.mdx
```diff
@@ -95,7 +95,7 @@ Les personnes peuvent s'abonner à votre flux dans un lecteur de flux et recevoi
     import { defineConfig } from "astro/config";
 
     export default defineConfig({
-      site: "https://exemple.com"
+      site: "https://mon-site-de-blog.netlify.app"
```
sarah11918, line 98:
```suggestion
      site: "https://example.com"
```
I'm going to ask our French translators for advice here! Normally "example.com" is a reserved URL for documentation etc. and it is recommended to use instead of other specific domain names.

But, I'm not sure whether the translated version is *also* a protected domain name specifically for testing and documentation! What do we think??

sarah11918 (commented):
Merci, @borelkoumo ! Yes, the `GET` must now be capitalized and you have matched the English version well. Thank you for catching this, and taking the time to contribute!

I have one question for our group, in a comment below, so we'll get that figured out and then can add this helpful contribution to our docs! :muscle:

src/content/docs/fr/tutorial/5-astro-api/4.mdx
thomasbnt: Owwh good catch @borelkoumo ! Yes @sarah11918, it's a translated URL, but it links to a fraudulent site or whatever. So it's better to keep `example.com`!

borelkoumo: Hi @sarah11918 and @thomasbnt 
I'm excited to use Astro, and to make my very first contribution to this framework.
I also think it's better to keep `example.com`, which is a well known placeholder

thomasbnt (approved):
LGTM! 🚀

yanthomasdev (approved):
LGTM! Thanks 🙌

PR #7481, opened 2024-03-19
Update github.mdx

ben-milanko:
This example does not work with the leading slash.

#### Description (required)

When trying this example out, the leading slash would deploy the site with an empty base, breaking the references.


sarah11918 (approved):
Thank you for providing this helpful update @ben-milanko , and welcome to Team Docs! :partying_face:

PR #7482, opened 2024-03-19
[i18nIgnore] fix(db): use `where` correctly

ematipico:
#### Description (required)

This PR updates a section of the Astro DB documentation. The section that shows how to delete a comment is incorrect. The `where` function doesn't accept a "raw" object, but we should always use the utility functions provided.

Reference: https://orm.drizzle.team/docs/sql#sql-in-where

Filed here https://github.com/withastro/astro/issues/10489

### DIFF
--- a/src/content/docs/ru/guides/astro-db.mdx
+++ b/src/content/docs/ru/guides/astro-db.mdx
@@ -233,7 +233,7 @@ import type { APIRoute } from "astro";
 import { db, Comment, eq } from 'astro:db';
 
 export const DELETE: APIRoute = async (ctx) => {
-  await db.delete(Comment).where(eq(Comment.id, ctx.params.id );
+  await db.delete(Comment).where(eq(Comment.id, ctx.params.id ));
   return new Response(null, { status: 204 });
 }
 ```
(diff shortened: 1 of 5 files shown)


delucis (approved):
Thanks for fixing this @ematipico! Noting that I’ll update the title to use `i18nIgnore` just in case any of the updated languages have other pending updates from the English so they aren’t marked as up-to-date.

src/content/docs/ru/guides/astro-db.mdx
```diff
@@ -230,10 +230,10 @@ const comments = await db.select().from(Comment);
 ```ts
 // src/pages/api/comments/[id].ts
-import { db, Comment } from 'astro:db';
+import { db, Comment, eq } from 'astro:db';
 
 export const DELETE: APIRoute = async (ctx) => {
-  await db.delete(Comment).where({ id: ctx.params.id });
+  await db.delete(Comment).where(eq(Comment.id, ctx.params.id );
```
sarah11918, line 236:
```suggestion
  await db.delete(Comment).where(eq(Comment.id, ctx.params.id ));
```

sarah11918 (approved):
Thank you, @ematipico , looks great! Just spotted a missing parenthesis.

PR #7498, opened 2024-03-20
Update server-side-rendering.mdx and integrations-reference.mdx

morinokami:
#### Description (required)

- Corrected the spelling of "NodeJS" and "NodeJs" to the official "Node.js".
- Removed redundant sections in server-side-rendering.mdx.

- Closes #
- Suggested label:


src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -129,18 +129,6 @@ export default defineConfig({
...
-
-```astro title="src/pages/my-static-page.astro" ins={2}
----
-export const prerender = true;
-// ...
----
-<html>
-  <!-- Static, pre-rendered page here... -->
-</html>
-```
```
morinokami, line 142: The method for opting into pre-rendering in `server` mode is described immediately below, and it seems to be redundant. So, I think this part can be safely removed.

sarah11918 (approved):
Good call @morinokami ! Thank you!

PR #7507, opened 2024-03-20
[i18nIgnore] Remove prettier pnpm note

bluwy:
#### Description (required)

Since Prettier 3, I don't think `prettier-plugin-astro` should have issues with pnpm now. Original pnpm issue: https://github.com/prettier/prettier/issues/8056

The current [prettier setup recommendation](https://github.com/withastro/prettier-plugin-astro?tab=readme-ov-file#recommended-configuration) works well for me.

I removed the caution block for all translations.

cc @Princesseuh 

- Closes #
- Suggested label:


Princesseuh (approved):
Yep, that's correct! Prettier removed the auto plugin discovery completely, so it's no longer relevant. Good catch!

yanthomasdev (approved):
LGTM!

PR #7529, opened 2024-03-21
Enable `experimental.directRenderScript` option

bluwy:
#### Description (required)

Enable `experimental.directRenderScript` option so that scripts are rendered at the correct pages. But mainly also, I found this to improve performance during builds. The hoisted script analyzer before took some work to complete:

image

I navigated around the pages locally and everything seem to work right. One report I got (from Erika) is that there may be whitespace issues (as the script is now rendered where it's declared), but I also don't see it happening locally.

---

I also updated astro and rollup to latest. The current rollup build time locally for me went from 3m25s to 1m55s.

- Closes #
- Suggested label:


sarah11918:
Just a heads up that @delucis is away until Friday, and he should really be the one to take a look at this!

delucis:
@sarah11918 This might be a good one to tackle with a mass of Team Docs eyes! The change should be good, but the main thing will be to go through the site carefully checking that nothing has changed visually, so could be a good T&D candidate?

If all looks good, we might consider setting this as the default in Starlight too.

sarah11918:
Excellent! We'll review this one Thursday on Talking and Doc'ing and assuming no issues, we'll merge it then! Or, we'll provide feedback as to what might seem off.

alexanderniebuhr:
Very interested if this works.. I enabled that for the Astrolicious Starlight site which uses Tailwind and it got broken.. However that might have be caused by the combination of Starlight + Tailwind 🤔

bluwy:
Is that related to this?

> One report I got (from Erika) is that there may be whitespace issues (as the script is now rendered where it's declared), but I also don't see it happening locally.

I don't think we can fix that because that's how the new rendering technique works. You need re-structure/style your elements to ignore the whitespace.

alexanderniebuhr:
@bluwy It could be due to that, not sure. However since it's Starlight + Tailwind, there is no way I can restructure them. So that would need to be something Starlight or the Starlight Tailwind integration, would need to do 🤔

BryceRussell:
I am running into issues with the mobile table of contents, clicking any of the links does not collapse the view. In my console I am getting these errors: 
image

sarah11918:
@bluwy One issue reported by Bryce!

bluwy:
Thanks @BryceRussell! That seems like a real bug, it looks like the `/_astro/TableOfContents...` chunk was accidentally deleted because it was inlined in another component. I'll fix that upstream.

@alexanderniebuhr Happy to take a look into it if you can share an example. I think a fix would be to mark certain styles within Starlight to be non-whitespace sensitive so that Tailwind doesn't change it, but I'm not sure where.

alexanderniebuhr:
@bluwy forget what I said. It works now, so maybe a recent Astro release fixed it. I tested it two weeks back 🤔 :)

bluwy:
This should be ready now. The bug above was fixed in https://github.com/withastro/astro/pull/10686 and I've updated Astro to latest.

BryceRussell (approved):
Looks great! I tested the preview some more and everything is working with no errors in console

yanthomasdev (approved):

PR #7531, opened 2024-03-21
Improve tab scroll on click

bluwy:
#### Description (required)

At https://docs.astro.build/en/install/auto/#1-run-the-setup-wizard, when clicking a package manager tab, it auto scrolls to the tab and I lost focus of where I am 😅 This PR replaces the auto-scroll with a logic that maintains the scroll position when clicking on a tab.

A more complex example where the scroll position can shift is at https://docs.astro.build/en/recipes/sharing-state-islands/#using-atoms. With this PR, the tab I clicked will also maintain its position.

(I think it's best to showcase a video of how it works, but I haven't figured how to do that on my laptop without large file sizes)

- Closes #
- Suggested label:


sarah11918:
Thanks, @bluwy ! I *think* there's current work on Starlight to fix this behaviour there, which we would then get when we update our Starlight version. So, I'll let you, @delucis and @HiDeoo negotiate this one!  :smile:

liruifengv:
I've encountered this problem and this behavior is very annoying to me. I wanted to file an issue but I forgot.

HiDeoo:
To comment on the Starlight part of things:

- There is a [PR](https://github.com/withastro/starlight/pull/640) adding the syncing behavior to Starlight tabs (old one but I want to revive it soon)
- The tab scroll lock mechanism is something @TheOtterlord and I tried to implement for the PR I linked but there was a few [edge cases](https://github.com/HiDeoo/starlight/pull/3#issuecomment-1855497391) that we didn't solve at that time (specially when the tab heights change) but my plan is to revisit it and add it to the Starlight PR.

bluwy:
I think since upstream support may take a while to iron out, it might be worth to push this along first? The auto-scroll prevention would improve the experience at the meantime.

TheOtterlord (approved):
Yep, I'm happy with adding this for now. Once upstream support is released, we can update this to use the new implementation.

PR #7535, opened 2024-03-21
feat: enable Content Collection Caching

Princesseuh:
#### Description (required)

Maybe I could've tried this before, sorry for the time loss.

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -69,6 +69,9 @@ export default defineConfig({
 	trailingSlash: 'always',
 	scopedStyleStrategy: 'where',
 	compressHTML: false,
+	experimental: {
+		contentCollectionCache: true,
+	},
 	markdown: {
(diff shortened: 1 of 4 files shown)


delucis:
Adding a quick note that I’m not 100% confident we can rely on Vercel to test the cache hit in this case — tried rerunning the deploy there a few times with build cache enabled, but it always logs:
```
Cache invalid, rebuilding from source. Reason: The cache manifest version has changed.
```
I think this is because Vercel only caches `node_modules` from `main` not from deploy branches so the lock file has always changed on branch deploy? Not sure, but just a heads up.

matthewp:
@delucis that's what I was going to test today, thanks!

matthewp:
I think in this case a code issue is causing it to continuously be invalidated. I'm working on a fix while writing tests.

PR #7538, opened 2024-03-22
Update routing.mdx, typo fix

Nepsha1986:


ElianCodes (commented):
This actually is correct as a .ts iirc. Further down it also writes .ts

sarah11918:
Thanks @Nepsha1986 ! As Elian said, this is correct as written, and is meant to illustrate using this kind of file here. So, I'll close this PR but we appreciate you taking the time to contribute and look forward to you spotting more corrections to make in the future! :rocket:

PR #7555, opened 2024-03-23
i18n(pt-br): Update `guides/images.mdx`

Fryuni:
#### Description

Update the Portuguese translation of the images guide.

#### Notes

I asked on Discord about how to translate "no-op passthrough service", which is used to refer to the image service configuration that does no optimization. I went with Yan's suggestion of "serviço de passagem não-operacional" but others might have other ideas for that.


yanthomasdev (approved):
LGTM, good job!

PR #7556, opened 2024-03-23
i18n(es): update vercel.mdx

AitorMT:
#### Description (required)

Updated to comply with English documentation


yanthomasdev:
Looks like this was already updated in another PR, thanks for it in any case, but since there's no changes I'll close it! 🙌

PR #7565, opened 2024-03-25
i18n (ar): Updating README.md file

mahmood601:
Some of these words are not widely used and are strange to developers


yanthomasdev (approved):
Thanks @mahmood601, welcome to Team Docs 🙌

PR #7576, opened 2024-03-25
Adds bun as second tab for npm commands

bitbonsai:
#### Description (required)

Proposing to add bun to the shell commands.

- Closes #
- Suggested label:


TheOtterlord:
Hi @bitbonsai! Thanks for the PR! We know Bun is working on supporting Astro, but Bun is not officially supported by Astro, so we are not including it in our documentation at this time. We do have a user-submitted recipe for [using Bun with Astro](https://docs.astro.build/en/recipes/bun/), and if there is an error on that page you are welcome to PR a correction! But we've decided that for the time being, we will not be adding additional package managers or runtimes to the docs.

PR #7594, opened 2024-03-26
i18n(ja): Update tutorial/1-setup/index.mdx

morinokami:
#### Description (required)

- Update tutorial/1-setup/index.mdx.
  - [diff](https://github.com/withastro/docs/commits/main/src/content/docs/en/tutorial/1-setup/index.mdx?since=2023-10-08T16:01:17.000Z)
    - https://github.com/withastro/docs/pull/7496

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/ja/tutorial/1-setup/index.mdx
+++ b/src/content/docs/ja/tutorial/1-setup/index.mdx
@@ -10,6 +10,7 @@ i18nReady: true
 import Badge from '~/components/Badge.astro';
 import Checklist from '~/components/Checklist.astro';
 import Box from '~/components/tutorial/Box.astro';
+import { Steps } from '@astrojs/starlight/components';
 
 これから何を作成するのかわかったところで、必要なツールを準備していきましょう！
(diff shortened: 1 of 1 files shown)


liruifengv:
Thank you @morinokami A suggestion. Would you be willing to merge all the tutorial sections into one PR, since we have 24 files and our docs repository is very large and the build time is very long, which will cause us to spend a long time waiting for the CI and then merge it!

morinokami:
@liruifengv Thanks for your suggestion! I understand the concern about the CI build times given the repository's size and agree that merging PRs could take considerable time. However, some files may require not a few changes, and I was worried that combining them all into a single PR might overwhelm reviewers. To strike a balance between reducing CI time and easing the review process, what if we group the updates by chapters rather than consolidating everything into one PR? This approach might offer a middle ground that addresses both concerns. What do you think?

liruifengv:
Yeah. I think updating by chapter would be a good way to do it

PR #7628, opened 2024-03-27
Adds Hookdeck website and docs as an example production site

leggetter:
#### Description (required)

I believe [hookdeck.com](https://hookdeck.com) and [hookdeck.com/docs](https://hookdeck.com/docs) are good production examples of what's possible with Astro.

I'm `leggetter` in the Astro Discord.

Website

CleanShot 2024-03-27 at 11 27 31@2x

Docs

CleanShot 2024-03-27 at 11 27 48@2x


sarah11918 (approved):
Thank you for adding a new resource, @leggetter ! We admittedly don't have a large sample of Markdoc examples, so it's always interesting to see one!

PR #7630, opened 2024-03-27
fix(markdown-content.mdx): fix incorrect remote markdown example

### LINKED ISSUE #7629: Incorrect remote markdown example
### 📚 Subject area/topic

Remote markdown

https://docs.astro.build/en/guides/markdown-content/#fetching-remote-markdown

The original code may initially appear to work as `npm run dev` will re-fetch the content every time the page is refreshed, but running `npm run build` will reveal that in the actual behaviour, the content is only fetched at build time and is not in fact dynamically fetched during runtime.

_No response_

tinnamchoi:
#### Description (required)

The original code may initially appear to work as `npm run dev` will re-fetch the content every time the page is refreshed, but running `npm run build` will reveal that in the actual behaviour, the content is only fetched at build time and is not in fact dynamically fetched during runtime.

The new code fixes this by moving the JavaScript outside of the component script and instead placing it in a `<script>` tag.

- Closes #7629 
- Suggested label: [code snippet update](https://github.com/withastro/docs/labels/code%20snippet%20update), [improve documentation](https://github.com/withastro/docs/labels/improve%20documentation)

### DIFF
--- a/src/content/docs/en/guides/markdown-content.mdx
+++ b/src/content/docs/en/guides/markdown-content.mdx
@@ -670,10 +670,13 @@ Astro was primarily designed for local Markdown files that could be saved inside
 // Example: Fetch Markdown from a remote API
 // and render it to HTML, at runtime.
 // Using "marked" (https://github.com/markedjs/marked)
-import { marked } from 'marked';
-const response = await fetch('https://raw.githubusercontent.com/wiki/adam-p/markdown-here/Markdown-Cheatsheet.md');
-const markdown = await response.text();
-const content = marked.parse(markdown);
 ---
-<article set:html={content} />
+<script>
+  import { marked } from 'marked';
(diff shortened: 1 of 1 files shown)


TheOtterlord (commented):
Thanks for the PR @tinnamchoi!

Looking at the documentation presented above the code snippet, it looks like this example is intended to be run on the server. 

> For example, you may need to fetch and render Markdown from a remote API **when you build your website** (or **when a user makes a request to your website, when using [SSR](https://docs-git-fork-tinnamchoi-remote-markdown-astrodotbuild.vercel.app/en/guides/server-side-rendering/)**).

If you wanted to use this example to fetch on each request, you'd add an SSR adapter. I don't think we'd want to document client-side markdown parsing here, so I'll close the PR and issue.

Let me know if you have any further questions, and if you need any assistance with markdown rendering, let us know in the [Discord](https://astro.build/chat)!

PR #7652, opened 2024-03-28
feat: better explanation of the Node.js policy

ematipico:
#### Description (required)

This PR adds a new paragraph that explains how Astro will manage the support and upgrade of the Node.js versions.

I took the opportunity to add also a paragraph around our extended maintenance

### DIFF
--- a/src/content/docs/en/upgrade-astro.mdx
+++ b/src/content/docs/en/upgrade-astro.mdx
@@ -186,6 +186,7 @@ A major release may also include some non-breaking changes and improvements that
 #### Support
 
 - Astro supports the [**latest _Maintenance_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule).
+- Astro supports the [**_Active_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule)
 - Astro can support odd versions of Node.js.
 
 #### Upgrade
--- a/src/content/docs/en/upgrade-astro.mdx
+++ b/src/content/docs/en/upgrade-astro.mdx
@@ -192,7 +192,7 @@ A major release may also include some non-breaking changes and improvements that
 
 The following rules define when Astro may deprecate, drop, or add support for versions of Node.js:
 
-- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
+- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after a reasonable period of extended support as decided by the Astro Core team.
 - Upgrading the minimum **_Maintenance_ LTS** (withing the same major range, e.g. from `v18.14.*` to `v18.20.*`) version of Node.js can occur in a **minor** release of Astro.
   - Security exception: If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_Maintenance_ LTS** in a **patch** release.
 - Upgrading minor or major versions of Node.js (**not** Maintenance LTS) occurs only in major versions of Astro.
--- a/src/content/docs/en/upgrade-astro.mdx
+++ b/src/content/docs/en/upgrade-astro.mdx
@@ -180,3 +180,22 @@ A major release may also include some non-breaking changes and improvements that
   Once these features are out of their experimental period, they will follow the normal semantic versioning contract.
 
 - **Improvements to the documentation** (e.g. reference and error messages). They are built from source for the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
+
+### Node.js support and upgrade policies
+
+#### Support
+
+- Astro supports the **latest _extended_ LTS** versions of Node.js.
+- Astro can support odd versions of Node.js.
+
+#### Upgrade
+
+- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
+- Upgrade of the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
+  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_extended_ LTS** in a **patch** release.
+- Upgrade of minor or major versions of Node.js - **not** extended LTS - occurs only in major versions of Astro.
+  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version in a **minor** release.
+
+### Extended maintenance
+ 
+The Core team will provide extended maintenance e.g. security fixes, only to the previous major of the current release. This means that of the **current major** is `v4.*`, the Core team will backport security fixes on for `v3.*` versions. 
--- a/src/content/docs/en/upgrade-astro.mdx
+++ b/src/content/docs/en/upgrade-astro.mdx
@@ -198,4 +198,4 @@ A major release may also include some non-breaking changes and improvements that
 
 ### Extended maintenance
  
-The Core team will provide extended maintenance e.g. security fixes, only to the previous major of the current release. This means that of the **current major** is `v4.*`, the Core team will backport security fixes on for `v3.*` versions. 
+The Core team will provide extended maintenance for security fixes, only to the previous major of the current release. This means that of the **current major** is `v4.*`, the Core team will backport security fixes on for `v3.*` versions. 
--- a/src/content/docs/en/upgrade-astro.mdx
+++ b/src/content/docs/en/upgrade-astro.mdx
@@ -186,7 +186,7 @@ A major release may also include some non-breaking changes and improvements that
 #### Support
 
 - Astro supports the [**latest _Maintenance_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule).
-- Astro supports the [**_Active_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule)
+- Astro supports the [**current _Active_ LTS** version of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule)
 - Astro can support odd versions of Node.js.
 
 #### Upgrade
--- a/src/content/docs/en/upgrade-astro.mdx
+++ b/src/content/docs/en/upgrade-astro.mdx
@@ -190,11 +190,13 @@ A major release may also include some non-breaking changes and improvements that
 
 #### Upgrade
 
+The following rules define when Astro may deprecate, drop, or add support for versions of Node.js:
+
 - Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
-- Upgrade of the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
-  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_extended_ LTS** in a **patch** release.
-- Upgrade of minor or major versions of Node.js - **not** extended LTS - occurs only in major versions of Astro.
-  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version in a **minor** release.
+- Upgrading the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
+  - Security exception: If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_extended_ LTS** in a **patch** release.
+- Upgrading minor or major versions of Node.js (**not** extended LTS) occurs only in major versions of Astro.
+  - Security exception: If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version in a **minor** release.
 
 ### Extended maintenance
  
--- a/src/content/docs/en/upgrade-astro.mdx
+++ b/src/content/docs/en/upgrade-astro.mdx
@@ -185,7 +185,7 @@ A major release may also include some non-breaking changes and improvements that
 
 #### Support
 
-- Astro supports the **latest _extended_ LTS** versions of Node.js.
+- Astro supports the **latest _Maintenance_ LTS** versions of Node.js.
 - Astro can support odd versions of Node.js.
 
 #### Upgrade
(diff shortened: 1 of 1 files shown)


florian-lefebvre (commented):
May be worth linking to the official node docs where there's the versions diagram, so that people know what versions are currently supported without needing us to update the data

ematipico:
> without needing us to update the data

Which data?

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,22 @@ A major release may also include some non-breaking changes and improvements that
...
+
+- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
+- Upgrade of the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
+  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_extended_ LTS** in a **patch** release.
+- Upgrade of minor or major versions of Node.js - **not** extended LTS - occurs only in major versions of Astro.
+  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version in a **minor** release.
+
+### Extended maintenance
+ 
+The Core team will provide extended maintenance e.g. security fixes, only to the previous major of the current release. This means that of the **current major** is `v4.*`, the Core team will backport security fixes on for `v3.*` versions.
```
sarah11918, line 201: Just pointing out that this implies that you will provide other maintenance too, and not JUST security fixes. (Not sure I've seen that actually happen? I *think* I've only seen older versions released for security reasons?)

If you really mean "just security fixes" then I would say that! Otherwise, this is pretty vague and makes it sound like v3 is *regularly maintained* too.

florian-lefebvre:
> > without needing us to update the data
> 
> Which data?

I mean like if Node 20 becomes the LTS we don't have to update this page to state "Node 20". I was referring to https://nodejs.org/en/about/previous-releases

matthewp (approved):

ematipico:
> > > without needing us to update the data
> > 
> > 
> > Which data?
> 
> I mean like if Node 20 becomes the LTS we don't have to update this page to state "Node 20". I was referring to [nodejs.org/en/about/previous-releases](https://nodejs.org/en/about/previous-releases)

I didn't use versions on purpose in this PR, and used only specific terms, so we don't need to update it every time. If `v18` is discontinued, this document still applies. If it doesn't, we should change the wording. 

The only change we should apply when updating the minimum version is here:  https://docs.astro.build/en/install/auto/#prerequisites

src/content/docs/en/upgrade-astro.mdx
ematipico: That's fair, I'll update it. Only security fixes

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,22 @@ A major release may also include some non-breaking changes and improvements that
...
+
+- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
+- Upgrade of the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
+  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_extended_ LTS** in a **patch** release.
+- Upgrade of minor or major versions of Node.js - **not** extended LTS - occurs only in major versions of Astro.
+  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version in a **minor** release.
+
+### Extended maintenance
+ 
+The Core team will provide extended maintenance for security fixes, only to the previous major of the current release. This means that of the **current major** is `v4.*`, the Core team will backport security fixes on for `v3.*` versions.
```
sarah11918, line 201:
```suggestion
The Core team will provide extended maintenance **for security fixes only** for one previous major version. This means that if the **current major** is `v4.*`, the Core team will back port security fixes and issue a new `v3.*` release. 
```
Just checking, but I think last time we issued a new v3.x release, not backported compatibility for everyone on every single previous v3.* version? I would frame it more like above, where "we'll issue an updated v3 version you can update to" and not "you can stay on whichever version of 3 you want"

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,22 @@ A major release may also include some non-breaking changes and improvements that
...
+- Astro supports the **latest _extended_ LTS** versions of Node.js.
+- Astro can support odd versions of Node.js.
+
+#### Upgrade
+
+- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
+- Upgrade of the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
+  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_extended_ LTS** in a **patch** release.
+- Upgrade of minor or major versions of Node.js - **not** extended LTS - occurs only in major versions of Astro.
+  - If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version in a **minor** release.
```
sarah11918, line 197:
```suggestion
The following rules define when Astro may deprecate, drop, or add support for versions of Node.js:

- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
- Upgrading the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
  - Security exception: If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_extended_ LTS** in a **patch** release.
- Upgrading minor or major versions of Node.js (**not** extended LTS) occurs only in major versions of Astro.
  - Security exception: If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version in a **minor** release.
```
I like having an intro sentence to explain what these points are!

src/content/docs/en/upgrade-astro.mdx
ematipico: That's a good point, thank you for the clarification

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,24 @@ A major release may also include some non-breaking changes and improvements that
...
+
+- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
+- Upgrading the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
+  - Security exception: If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version of the **_extended_ LTS** in a **patch** release.
+- Upgrading minor or major versions of Node.js (**not** extended LTS) occurs only in major versions of Astro.
+  - Security exception: If a security flaw in Node.js that **affects Astro** is disclosed and fixed, the Core team can bump the minimum version in a **minor** release.
+
+### Extended maintenance
+ 
+The Core team will provide extended maintenance **for security fixes only** for one previous major version. This means that if the current major is `v4.*`, the Core team will back port security fixes and issue a new `v3.*` release.
```
sarah11918, line 203:
```suggestion
The Core team will provide extended maintenance **for security fixes only** for one previous major version. This means that if the current major is `v4.*`, the Core team will back port security fixes and issue new `v3.*` releases when necessary. 
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,24 @@ A major release may also include some non-breaking changes and improvements that
   Once these features are out of their experimental period, they will follow the normal semantic versioning contract.
 
 - **Improvements to the documentation** (e.g. reference and error messages). They are built from source for the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
+
+### Node.js support and upgrade policies
+
+#### Support
+
+- Astro supports the **latest _extended_ LTS** versions of Node.js.
```
Fryuni, line 188: What does "extended LTS" stand for here? On Node's site, they have "Active LTS" and "Maintenance LTS". I suppose this refers to "Maintenance LTS", the longest of the two.
Or is this referring to the extended support that some, like Meteor.js, offer one year beyond the end of the official support from the Node team?

Also, should we link to the Node page so the user can see which version this refers to? I'm guessing that not every user reading this will know from memory which Node version is currently in each support stage.

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,24 @@ A major release may also include some non-breaking changes and improvements that
...
+- Astro can support odd versions of Node.js.
+
+#### Upgrade
+
+The following rules define when Astro may deprecate, drop, or add support for versions of Node.js:
+
+- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
+- Upgrading the minimum **_extended_ LTS** version of Node.js can occur in a **minor** release of Astro.
```
Fryuni, line 196: By "Upgrading the minimum extended LTS version", does it mean an Astro minor can bump the minimum version only _within the same LTS_ (from the current `18.14.1` to `18.20.0` for example) or can it also bump to the next LTS once the current loses support (from `18` to `20` once support for Node 18 ends in June 2025)?

ematipico: Within the same LTS, as per your example

src/content/docs/en/upgrade-astro.mdx
ematipico: I will update the terminology, using _Maintainance LTS_

florian-lefebvre: I already suggested linking to the node versions diagram page and still think it's a good idea

sarah11918: Can someone drop the link here to the thing you want to link to?

Fryuni: https://nodejs.org/en/about/previous-releases#release-schedule

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,24 @@ A major release may also include some non-breaking changes and improvements that
   Once these features are out of their experimental period, they will follow the normal semantic versioning contract.
 
 - **Improvements to the documentation** (e.g. reference and error messages). They are built from source for the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
+
+### Node.js support and upgrade policies
+
+#### Support
+
+- Astro supports the **latest _Maintenance_ LTS** versions of Node.js.
```
sarah11918, line 188:
```suggestion
- Astro supports the [**latest _Maintenance_ LTS** versions of Node.js]([nodejs.org/en/about/previous-releases#release-schedule](https://nodejs.org/en/about/previous-releases#release-schedule).
```
I think the link here makes sense!

sarah11918 (approved):
Added the link as suggested, and otherwise, when everyone here is happy, you're welcome to merge this Ema!

src/content/docs/en/upgrade-astro.mdx
Fryuni: There is an extra portion in the middle.

```suggestion
- Astro supports the [**latest _Maintenance_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule).
```

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,24 @@ A major release may also include some non-breaking changes and improvements that
   Once these features are out of their experimental period, they will follow the normal semantic versioning contract.
 
 - **Improvements to the documentation** (e.g. reference and error messages). They are built from source for the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
+
+### Node.js support and upgrade policies
+
+#### Support
+
+- Astro supports the [**latest _Maintenance_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule).
```
Princesseuh, line 188: Presumably this should also say that the current LTS is supported. Right now, this (and the line below) implies that Astro only supports Node 18, 19, and 21, but not 20.

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,24 @@ A major release may also include some non-breaking changes and improvements that
...
+- Astro can support odd versions of Node.js.
+
+#### Upgrade
+
+The following rules define when Astro may deprecate, drop, or add support for versions of Node.js:
+
+- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after an arbitrary period of time decided by the Astro Core team.
```
sarah11918, line 195:
```suggestion
- Odd versions of Node.js can be deprecated and/or dropped when the next even version of Node.js published. This change can occur in a **minor** release of Astro, after a reasonable period of extended support as decided by the Astro Core team.
``` 
Maybe a bit friendlier?  :smile:

src/content/docs/en/upgrade-astro.mdx
sarah11918: Good call! I wasn't sure whether this line was meant to "assure about maintenance", but if it starts as "what we support", it should be exhaustive!

florian-lefebvre (approved):
LGTM

src/content/docs/en/upgrade-astro.mdx
ematipico: Added in [`12dc2df` (#7652)](https://github.com/withastro/docs/pull/7652/commits/12dc2dfaf77278cfb837d0234ea338f7bb891cc9)

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,25 @@ A major release may also include some non-breaking changes and improvements that
   Once these features are out of their experimental period, they will follow the normal semantic versioning contract.
 
 - **Improvements to the documentation** (e.g. reference and error messages). They are built from source for the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
+
+### Node.js support and upgrade policies
+
+#### Support
+
+- Astro supports the [**latest _Maintenance_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule).
+- Astro supports the [**_Active_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule)
```
Fryuni, line 189:
```suggestion
- Astro supports the [**_Active_ LTS** version of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule)
```

There is only one Active LTS version at any given time.

Fryuni (approved):
LGTM

src/content/docs/en/upgrade-astro.mdx
ematipico: Can you change the suggestion and add `current _Active_ LTS` instead?

Princesseuh: > Can you change the suggestion and add `current _Active_ LTS` too?

You can edit people's messages to add it

src/content/docs/en/upgrade-astro.mdx
```diff
@@ -180,3 +180,25 @@ A major release may also include some non-breaking changes and improvements that
   Once these features are out of their experimental period, they will follow the normal semantic versioning contract.
 
 - **Improvements to the documentation** (e.g. reference and error messages). They are built from source for the `docs` repository. This allows Astro to quickly update docs fixes and improvements in the cases where documentation source content is stored in the main `astro` repository.
+
+### Node.js support and upgrade policies
+
+#### Support
+
+- Astro supports the [**latest _Maintenance_ LTS** versions of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule).
```
ematipico, line 188:
```suggestion
- Astro supports the [**latest _Maintenance_ LTS** version of Node.js](https://nodejs.org/en/about/previous-releases#release-schedule).
```

Princesseuh (approved):

PR #7670, opened 2024-03-29
docs: fix typo in `integrations-guide/cloudflare.mdx`

jsparkdev:
#### Description (required)

- fix typo in `integrations-guide/cloudflare.mdx`

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -256,7 +256,7 @@ const { env } = Astro.locals.runtime;
 
 You can access the runtime from API endpoints through `context.locals`:
 
-```ts title="src/pages/api/someFile.ts"
+```js title="src/pages/api/someFile.js"
 export function GET(context) {
   const runtime = context.locals.runtime;
 
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -256,7 +256,7 @@ const { env } = Astro.locals.runtime;
 
 You can access the runtime from API endpoints through `context.locals`:
 
-```ts title="src/pages/api/someFile.js"
+```ts title="src/pages/api/someFile.ts"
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -256,7 +256,7 @@ const { env } = Astro.locals.runtime;
 
 You can access the runtime from API endpoints through `context.locals`:
 
-```ts title="src/pages/api/someFile.js"
+```ts title="src/pages/api/someFile.ts"
```
alexanderniebuhr, line 259: > To create a custom endpoint, add a .js or .ts file to the /pages directory. The .js or .ts extension will be removed during the build process [...]

Astro docs state that endpoints can be `.js` or `.ts` files. So why do we need this change?

jsparkdev: The code block is specified as `ts`, but the file name is `.js`, so I thought it would be inconsistent and should be fixed.

As you said, if this is what you intended, I'll close this PR :)

alexanderniebuhr:
```suggestion
```js title="src/pages/api/someFile.js"
```

Oh sorry I missed that, are you happy with this suggestion?

sarah11918: My thinking was that this file didn't *look* explicitly like a ts file. I would instead have changed the snippet language to js!

I'm happy for it to be either one, but this looks like just js to me, so that would be my preference. We usually default to "people who want to chose ts instead will do that on their own". BUT, if you have reason to believe that e.g. 90% of these adapter users WOULD choose ts instead, I don't mind showing one.

I will leave the final call to @alexanderniebuhr !

jsparkdev: > Oh sorry I missed that, are you happy with this suggestion?

Sounds good!

alexanderniebuhr (approved):

PR #7723, opened 2024-04-01
Document ASTRO_DATABASE_FILE

matthewp:
#### Description (required)

- Documents the `ASTRO_DATABASE_FILE` environment variable.

None


src/content/docs/en/guides/astro-db.mdx
```diff
@@ -499,3 +499,21 @@ export default async function() {
...
+
+This path should be to your `.db` file on the host. It's important that you *do not* override this file on deployment, ever, or you will lose your production data. Additionally, any schema migrations must be managed manually using this method.
+
+If you're comfortable with the risks and are able to manage deployment yourself, you can set the `ASTRO_DATABASE_FILE` variable during your build and omit:
+
+```shell
+ASTRO_DATABASE_FILE=/srv/files/database.db astro build
+```
+
+The build will statically compile with this path as your production database. When you deploy and launch your server it will connect to the file at this path on the production host.
```
TheOtterlord, line 519: Moved some things about and tried to clarify other things. I think it's probably OK to say explicitly that this method won't work with serverless environments.

> Additionally, any schema migrations must be managed manually using this method.

I wasn't 100% sure what this meant in the context, so I've left it as-is, along with the surrounding content. Does this essentially mean you need to run your migration in your project directory on the server, with the `ASTRO_DATABASE_FILE` configured in `.env`.

``````suggestion
## Self-hosted production deployment

If you deploy your site to a self-managed host such as a [virtual private server](https://en.wikipedia.org/wiki/Virtual_private_server), you can choose to use a database file instead of connecting to Astro Studio.

:::caution
Using a database file is an advanced feature, and care should be taken when deploying to prevent overriding your database. This method will also not work in serverless deployments, as the file system is not persisted in those environments. For a fully managed solution, try [connect to the Astro Studio platform](/en/recipes/studio/) instead.
:::

To use a database file, set the `ASTRO_DATABASE_FILE` environment variable. This path should point to your `.db` file within the host environment. It's important that you *do not* override this file on deployment, ever, or you will lose your production data. Additionally, any schema migrations must be managed manually using this method.

If you're comfortable with the risks and are able to manage deployment yourself, you can set the `ASTRO_DATABASE_FILE` variable during your build and omit:

```shell
ASTRO_DATABASE_FILE=/srv/files/database.db astro build
```

The build will statically compile with this path as your production database. When you deploy and launch your server it will connect to the file at this path on the production host.
``````

**Review by TheOtterlord — COMMENTED** [review]
Left an initial comment proposing some updates to the structure, and to clarify a few points. I wrote it in one piece, but feel free to pick apart anything you find useful.

File: src/content/docs/en/guides/astro-db.mdx (hunk 1 above)
Comment (matthewp) [reply src/content/docs/en/guides/astro-db.mdx:519]: I like it!

File: src/content/docs/en/guides/astro-db.mdx [hunk 2]
```diff
@@ -499,3 +499,21 @@ export default async function() {
 
 The value returned by `asDrizzleTable('Pets', Pets)` is equivalent to `import { Pets } from 'astro:db'`, but is available even when Astro’s type generation can’t run.
 You can use it in any integration code that needs to query or insert into the database.
+
+## Self-hosted production deployment
+
+If you deploy your site to a self-managed host such as a [virtual private server](https://en.wikipedia.org/wiki/Virtual_private_server), you can choose to use a database file instead of connecting to Astro Studio.
```
sarah11918, line 505:
```suggestion
If you deploy your site to a self-managed host such as a [virtual private server](https://en.wikipedia.org/wiki/Virtual_private_server), you can choose to use a database file instead of connecting to a database hosted at Astro Studio.
```

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -499,3 +499,21 @@ export default async function() {
...
+
+:::caution
+Using a database file is an advanced feature, and care should be taken when deploying to prevent overriding your database. This method will also not work in serverless deployments, as the file system is not persisted in those environments. For a fully managed solution, try [connect to the Astro Studio platform](/en/recipes/studio/) instead.
```
sarah11918, line 508:
```suggestion
Using a database file is an advanced feature, and care should be taken when deploying to prevent overriding your database and losing your production data.

Additionally, this method will not work in serverless deployments, as the file system is not persisted in those environments.

For a fully managed solution, [connect to databases hosted on the Astro Studio platform](/en/recipes/studio/) instead.
```

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -499,3 +499,21 @@ export default async function() {
...
+
+To use a database file, set the `ASTRO_DATABASE_FILE` environment variable. This path should point to your `.db` file within the host environment. It's important that you *do not* override this file on deployment, ever, or you will lose your production data. Additionally, any schema migrations must be managed manually using this method.
+
+If you're comfortable with the risks and are able to manage deployment yourself, you can set the `ASTRO_DATABASE_FILE` variable during your build and omit:
+
+```shell
+ASTRO_DATABASE_FILE=/srv/files/database.db astro build
+```
+
+The build will statically compile with this path as your production database. When you deploy and launch your server it will connect to the file at this path on the production host.
```
sarah11918, line 519:
```````suggestion
If you are comfortable with the risks, and can manage deployment yourself, you can use a database file instead of connecting to Studio. 

Set the `ASTRO_DATABASE_FILE` environment variable to a path pointing to your `.db` file within the host environment during your build:

```shell
ASTRO_DATABASE_FILE=/srv/files/database.db astro build
```

The build will statically compile with this path as your production database. When you deploy and launch your server it will connect to the file at this path on the production host.

:::danger
If you override your `.db` file on deployment, you will lose your production data.** MATTHEW, CAN YOU SUGGEST ONE GENERAL OR SPECIFIC EXAMPLE HERE OF HOW NOT TO DO THAT, OR ONE THING THAT DEFINITELY WILL CAUSE THAT? I think that would be a helpful nudge here. Even if it's just, "Check the settings of your deploy host/deployment environment for... " or something like that? ANYTHING we can add here?
:::

Additionally, [pushing any table schema changes](#pushing-table-schemas) (also known as "schema migrations") must be managed manually using this method:

```shell
ASTRO_DATABASE_FILE=/srv/files/database.db astro db push
```
``````
Is this the correct example for pushing schema migrations? Note, we don't use that phrase anywhere else on this page (we do say "pushing table schemas") so we can't just use this phrase as if it means something already explained, so we either have to commit to "pushing table schemas" or acknowledge that we haven't yet used the word "migration"
```

sarah11918 (commented):
This is looking pretty good, and thank you @TheOtterlord Otterlord for a lovely flow here! Two quick additions for explicitness but the opening is really nice!

@matthewp See what you think of my slight reordering of the bottom section, and I wanted to show a second example for the schema migration (which brought up its own naming issue). Can you work with that kind of structure and make it accurate?

PS - I removed the word "omit" (remove?) because I wasn't sure what that was telling me to remove, nor from where. If that's a key word then we'll need to find a way to put it back in with a little more context.

src/content/docs/en/guides/astro-db.mdx
matthewp: I don't think we should instruct people on how to do manual deployments here. There's many ways to do it (like literally dozens) because it all depends on your host environment. That's why we wanted to add a warning here as this is an advanced feature.

sarah11918: Sure! let me restructure this a bit.

If we stick with the original wording, can you let me know what "omit" refers to? (I'm not familiar with it other than to remove, and I'm not clear on what is being removed from what)

sarah11918: Updated the comment @matthewp - now we have a caution AND a danger!  :joy: how does that sound? What specifically would you change from this?

matthewp: Looks good aside from the fact that there are some out of control backticks at the end which pushes a question into the file change...

sarah11918: I *think* I fixed it. I'll commit so we can see the preview, and if anything needs changing ,we totally can.

sarah11918 (approved):
OK, I think all the ticks are worked out here, @matthewp ! I checked the preview and it looks in order.

When you're happy with this, it's good to merge on my end. Happy for you to update branch and merge, or ping me to do it! (I don't think this is waiting for any feature PR first).

PR #7744, opened 2024-04-02
Rework Markdoc info hierarchy

bholmesdev:
#### Description (required)

- Restructure table-of-contents with clearer h2 sections. Before, everything was nested under Configuration. This breaks out related sections, and moves higher-traffic content (ex. how to use UI components) further up).

**Before | After**
image image

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/markdoc.mdx
+++ b/src/content/docs/en/guides/integrations-guide/markdoc.mdx
@@ -117,6 +117,63 @@ const { Content } = await entry.render();
 
 <ReadMore>See the [Astro Content Collection docs][astro-content-collections] for more information.</ReadMore>
 
+## Pass Markdoc variables
+
+You may need to pass [variables][markdoc-variables] to your content. This is useful when passing SSR parameters like A/B tests.
+
+Variables can be passed as props via the `Content` component:
+
+```astro title="src/pages/why-markdoc.astro"
+---
+import { getEntryBySlug } from 'astro:content';
+
+const entry = await getEntryBySlug('docs', 'why-markdoc');
+const { Content } = await entry.render();
+---
+
+<!--Pass the `abTest` param as a variable-->
+<Content abTestGroup={Astro.params.abTestGroup} />
+```
+
+Now, `abTestGroup` is available as a variable in `docs/why-markdoc.mdoc`:
+
+```md title="src/content/docs/why-markdoc.mdoc"
+{% if $abTestGroup === 'image-optimization-lover' %}
+
+Let me tell you about image optimization...
+
+{% /if %}
+```
+
+To make a variable global to all Markdoc files, you can use the `variables` attribute from your `markdoc.config.mjs|ts`:
+
+```js title="markdoc.config.mjs"
+import { defineMarkdocConfig } from '@astrojs/markdoc/config';
+
+export default defineMarkdocConfig({
+  variables: {
+    environment: process.env.IS_PROD ? 'prod' : 'dev',
+  },
+});
+```
+
+### Access frontmatter from your Markdoc content
+
+To access frontmatter, you can pass the entry `data` property as a variable where you render your content:
+
+```astro title="src/pages/why-markdoc.astro"
+---
+import { getEntry } from 'astro:content';
+
+const entry = await getEntry('docs', 'why-markdoc');
+const { Content } = await entry.render();
+---
+
+<Content frontmatter={entry.data} />
+```
+
+This can now be accessed as `$frontmatter` in your Markdoc.
+
 ## Render components
 
 `@astrojs/markdoc` offers configuration options to use all of Markdoc's features and connect UI components to your content.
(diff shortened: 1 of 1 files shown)


sarah11918:
@bholmesdev the table of contents does indeed look much nicer! 

Just because the diff is going to look terrible here, and not really reflect what you actually did, is the section on Partials the only new/changed *content* (other than reordering?) This will save me a bunch of close reading trying to figure out what content actually did change, and will also help the translators who will have to make sense of this PR when they update in all the other languages. :smile:

bholmesdev:
Yes, apologies! Let's get [the Partials PR](https://github.com/withastro/docs/pull/7743) reviewed first, then rebase this PR once it is merged. That way we don't have to untangle new content from reorganization.

bholmesdev:
@sarah11918 Okay, rebased and ready for review!

sarah11918:
Great! Assuming this is *just* reorganization of existing content for flow, this now should be an easier read!

src/content/docs/en/guides/integrations-guide/markdoc.mdx
```diff
@@ -283,45 +288,54 @@ export default defineMarkdocConfig({
 
 <ReadMore>To learn about configuring Prism stylesheets, [see our syntax highlighting guide](/en/guides/markdown-content/#prism-configuration).</ReadMore>
 
-### Set the root HTML element
+<ReadMore>See the [Markdoc nodes documentation](https://markdoc.dev/docs/nodes#built-in-nodes) to learn about all the built-in nodes and attributes.</ReadMore>
```
sarah11918, line 291: Does this belong here? It looks like it's part of the current Syntax Highlighting section when it's here. What do you think about moving this to line 312, right after the code sample about standard nodes/elements?

src/content/docs/en/guides/integrations-guide/markdoc.mdx
```diff
@@ -283,45 +288,54 @@ export default defineMarkdocConfig({
...
-<ReadMore>See the [Markdoc nodes documentation](https://markdoc.dev/docs/nodes#built-in-nodes) to learn about all the built-in nodes and attributes.</ReadMore>
+All Markdown headings will render the `Heading.astro` component and pass the following `attributes` as component props:
 
-### Custom image components in Markdoc
+* `level: number` The heading level 1 - 6
+* `id: string` An `id` generated from the heading's text contents. This corresponds to the `slug` generated by the [content `render()` function](/en/guides/content-collections/#rendering-content-to-html).
+
+For example, the heading `### Level 3 heading!` will pass `level: 3` and `id: 'level-3-heading'` as component props.
+
+### Custom image components
```
sarah11918, line 338: Just a note for translators: this heading has changed from "Custom image components in Markdoc" to just "Custom image components" It's more than just moved!!  :purple_heart:

src/content/docs/en/guides/integrations-guide/markdoc.mdx
```diff
@@ -436,38 +450,8 @@ The following steps will create a custom Markdoc image tag to display a `<figure
...
-      attributes: {
-        type: { type: String },
-      },
-    },
-  },
-});
-```
-
-### Markdoc config
+## Advanced Markdoc configuration
```
sarah11918, line 454: Note for translators: this heading has changed! Not just moved! 

It was previously (lol) "### Markdoc config" (line 470 in English) nested under "## Markdoc config"

We heard you liked Markdoc config, so we put some Markdog config in your Markdoc config...

sarah11918 (approved):
The reorganization looks fantastic @bholmesdev !

I left a couple of notes for translators to indicate when a heading actually changed content (vs just moving around). You can ignore those, that's just for their convenience.

I also left a comment about a "read more" line that I think got missed or errantly moved in the shuffle. Pretty sure that was an unintentional change and I identified where it should end up instead.

**The only thing about the reorganization itself I could see shuffling:**

Pass Markdoc variables and Access frontmatter from your Markdoc *could* go much higher, I'd argue maybe even under Usage?

This seems like a more common use case, and related to just "writing Markdoc", enough so that it feels like it could go right above using components. Also noting that the Access frontmatter is *very* related to the usage example showing accessing frontmatter variables. So, if you think those two sections show reasonably common use cases, I'd throw them up at the top there. If you think they are low-usage things that aren't all that common, then no issue keeping them lower. They just feel much more related to writing your Markdoc to me.

In any event, whatever you decide, this is a fabulous change! 🫡

src/content/docs/en/guides/integrations-guide/markdoc.mdx
```diff
@@ -75,18 +75,29 @@ export default defineConfig({
 });
 ```
 
-### Editor Integration
+### VS Code Editor Integration
```
sarah11918, line 78: Also note to translators: This new section is a combination of the older Editor Integration and #Markdoc langauge server. They have both been combined into this one!

sarah11918 (approved):
This looks good to me now, Ben! I assume we're not waiting on any other PR for this, so as soon as you're happy, feel free to merge this.

PR #7771, opened 2024-04-05
fix: remove unused import and add missing syntax

Egpereira:
# Remove unused import and add missing syntax

## Description (required)

This PR removes the unused import of the Badge component from the `cli-reference.mdx` file as well as some extra line breaks. It also adds the code syntax to the `astro link` command at the end of the file, to match the other occurrences of commands inside the main text.


sarah11918 (approved):
Thank you, @Egpereira ! We really appreciate you spotting these, and taking the time to fix them. Welcome to Team Docs!  🥳

Issue #7792, opened 2024-04-07
Fixing Data Insertion Type Mismatch in Astro DB Table Definitions


suzulang:
### 📚 Subject area/topic

astro-db

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/astro-db/

### 📋 Description of content that is out-of-date or incorrect

I will try to follow the documentation and configure astro-db.

The field names of the table provided in the documentation do not match the field names of the mock data in seed.ts

### 🖥️ Reproduction in StackBlitz (if reporting incorrect content or code samples)

**Title:**
Issues with Inserting Data into Astro DB using `astro:db`

**Body:**

Hello,

I'm working with Astro DB and encountered an issue while trying to insert data into a table. Below are the details of my setup and the problem I've encountered.

**Setup**
**Database Configuration (db/config.ts):**

```ts
// db/config.ts
import { defineDb, defineTable, column } from 'astro:db';

const Comment = defineTable({
  columns: {
    author: column.text(),
    body: column.text(),
  }
})

export default defineDb({
  tables: { Comment },
})
Data Seeding Script (db/seed.ts):
```
```ts
// db/seed.ts
import { db, Comment } from 'astro:db';

export default async function() {
  await db.insert(Comment).values([
    { authorId: 1, body: 'Hope you like Astro DB!' },
    { authorId: 2, body: 'Enjoy!'},
  ])
}
```
When I run this setup, it produces the following output (as seen in the attached image):

<img width="862" alt="image" src="https://github.com/withastro/docs/assets/155881657/bc9248ba-7fee-440a-8b21-7715af529835">

**Issue**
The compiler throws an error regarding the **authorId** field, which does not match the defined structure in **Comment** table where the field is named **author**.

**Solution Attempt**
I modified the database configuration and the seeding script as follows:

**Modified Database Configuration in db/config.ts:**

Changed `author: column.text(),` to `author: column.number(),`

**Modified Data Seeding Script (db/seed.ts):**

```ts
// db/seed.ts
import { db, Comment } from 'astro:db';

export default async function() {
  await db.insert(Comment).values([
    { author: 1, body: 'Hope you like Astro DB!' },
    { author: 2, body: 'Enjoy!'},
  ])
}
```
After making these changes, the script runs successfully, as shown in the attached image:
<img width="303" alt="image" src="https://github.com/withastro/docs/assets/155881657/bfdcb18f-e586-4542-a063-ba6263fc4b4d">

Questions
[...]

PR #7793, opened 2024-04-07
Fixing Data Insertion Type Mismatch in Astro DB Table Definitions #7792

suzulang:
i fixed this issue
https://github.com/withastro/docs/issues/7792

### DIFF
--- a/src/content/docs/en/guides/astro-db.mdx
+++ b/src/content/docs/en/guides/astro-db.mdx
@@ -65,7 +65,7 @@ import { defineDb, defineTable, column } from 'astro:db';
 
 const Comment = defineTable({
   columns: {
-    author: column.text(),
+    authorId: column.number(),
     body: column.text(),
   }
 })
@@ -87,7 +87,7 @@ import { defineTable, column } from 'astro:db';
 const Comment = defineTable({
   columns: {
     // A string of text.
-    author: column.text(),
+    authorId: column.number(),
     // A whole integer value.
     likes: column.number(),
     // A true or false value.
@@ -422,7 +422,7 @@ The following example renames a table from `Comment` to `Feedback`:
     const Comment = defineTable({
       deprecated: true,
     	columns: {
-    		author: column.text(),
+    		authorId: column.number(),
     		body: column.text(),
   		}
     });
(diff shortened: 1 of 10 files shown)


suzulang:
this is minimal-reproductions https://github.com/suzulang/minimal-reproduction

liruifengv:
Thanks for you fix @suzulang! We only need to update the original content in English and leave it to our translators for other languages.

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -87,7 +87,7 @@ import { defineTable, column } from 'astro:db';
 const Comment = defineTable({
   columns: {
     // A string of text.
```
liruifengv, line 89: I think the comment need to be updated too

liruifengv:
image
image

I found some similar questions with some code examples. Not sure if we want to update them. @sarah11918

sarah11918:
Hi @suzulang !

Thank you for taking the time to contribute! As @liruifengv said, we only accept changes to the English documentation, but also, these examples are all individual and not meant to be used together. Each one describes individually what it does, and we did not write them to match with each other. Instead, some show different shapes/content so that people can see different examples of what a database might look like. These are not steps to follow, but individual sections to consult for reference.

So, I will close this PR as it is not an issue for us that these samples do not match. Instead, I think we should make sure we produce some content that does intend to show content linked together! I will ask our team to make a recipe to follow based on your issue linked here.

Thank you very much for bringing this up!

suzulang:
> Hi @suzulang !
> 
> Thank you for taking the time to contribute! As @liruifengv said, we only accept changes to the English documentation, but also, these examples are all individual and not meant to be used together. Each one describes individually what it does, and we did not write them to match with each other. Instead, some show different shapes/content so that people can see different examples of what a database might look like. These are not steps to follow, but individual sections to consult for reference.
> 
> So, I will close this PR as it is not an issue for us that these samples do not match. Instead, I think we should make sure we produce some content that does intend to show content linked together! I will ask our team to make a recipe to follow based on your issue linked here.
> 
> Thank you very much for bringing this up!

Hello    @liruifengv  and team,

Thank you very much for your prompt response and clarification. I understand the guidelines presented and the explanation regarding the individual examples. I will keep an eye on the updates and new content, especially looking forward to those recipes that intend to show linked content. I am glad to contribute to the improvement of the documentation, and I will be keen to participate more actively if there are more suitable opportunities for contributions in the future.

Thank you again for your hard work and feedback, looking forward to seeing more excellent content.

Best regards

PR #7810, opened 2024-04-08
fix(docs): internationalisation domains config demo

afonsojramos:
#### Description (required)

Fixing a small docs bug in the internationalisation domains config demo.


sarah11918 (approved):
Well spotted! Thanks so much for fixing this! :raised_hands:

PR #7821, opened 2024-04-09
feat: rewrite toolbar reference page for new APIs

Princesseuh:
#### Description (required)

Rewrite the toolbar api reference to match new APIs

https://github.com/withastro/astro/pull/10665 & https://github.com/withastro/astro/pull/10667 rework how to make dev toolbar APIs to make it easier to make apps


src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -9,271 +9,327 @@ The Astro Dev Toolbar App API allows you to create apps that can be used to exte
...
+'astro:server:setup': ({ toolbar }) => {
+  toolbar.onInitialized('my-app', () => {
+    console.log('The app is now initialized!');
+  });
+},
 ```
 
 :::note
-The built-in `connection` event from Vite fires **before** Dev Toolbar apps are initialized and therefore cannot be used directly by apps. Instead, use the `astro-dev-toolbar:APP_ID:initialized` event.
+The built-in `connection` event from Vite fires **before** Dev Toolbar apps are initialized and therefore cannot be used directly by apps, as such you should always use the `onInitialized` method to ensure that the app is fully initialized before sending messages to it.
```
sarah11918, line 308:
```suggestion
The built-in `connection` event from Vite fires **before** Dev Toolbar apps are initialized and therefore cannot be used directly by apps. Use the `onInitialized` method to ensure that the app is fully initialized before sending messages to it.
```

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -9,271 +9,327 @@ The Astro Dev Toolbar App API allows you to create apps that can be used to exte
...
+    });
+  },
+});
+```
 
-An [`EventTarget`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) that can be used to send and receive events from the Dev Toolbar.
+#### `server`
 
-### `beforeTogglingOff`
+An object that can be used to communicate with the server, see the [Client-Server Communication](#client-server-communication) section for more information.
```
sarah11918, line 133:
```suggestion
An object that can be used to communicate with the server. See the [Client-Server Communication](#client-server-communication) section for more information.
```

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -9,271 +9,327 @@ The Astro Dev Toolbar App API allows you to create apps that can be used to exte
...
 
 #### canvas
 
-The ShadowRoot of the app, can be used to render any UI needed.
+The ShadowRoot of the app, can be used to render any UI needed before closing. Same as the `canvas` argument of the `init` function.
 
 ## Client-side Events
 
-Using the [`eventTarget`](#eventtarget) argument on the [`init`](#init-canvas-shadowroot-eventtarget-eventtarget--void) hook, apps can send and receive events from the Dev Toolbar. The following events are available:
+In addition to the standard methods of an `EventTarget` (addEventListener, dispatchEvent, etc), the `app` object also has the following methods:
```
sarah11918, line 171: Note here that I think we've lost the previous example showing `eventTarget.addEventListener` which is probably a helpful one to show! I don't mind if it's also here!

also, just double checking capital E on `EventTarget` here in this context!

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -9,271 +9,327 @@ The Astro Dev Toolbar App API allows you to create apps that can be used to exte
...
+In your integration's `astro:server:setup` hook, you can use the `toolbar` object to send and receive messages from your apps.
 
-## Client-Server Communication
+#### `send()`
 
-Using [Vite's methods for client-server communication](https://vitejs.dev/guide/api-plugin.html#client-server-communication), Dev Toolbar Apps can communicate with the server.
+**Signature:** `send<T>(event: string, data: T) => void`
 
-In addition to being able to send and receive custom messages, the Dev Toolbar also sends the following messages, where `APP_ID` is the [app's ID](#id-string):
+Sends data to the client
```
sarah11918, line 271:
```suggestion
Sends data to the client.
```

sarah11918 (commented):
Just a few initial comments reading through the file diff here (because some of what's showing up as "new" isn't really new).

I will also do a review comparing both live docs and the deploy preview here, but just a few tiny things that stood out to me here!

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -9,271 +9,327 @@ The Astro Dev Toolbar App API allows you to create apps that can be used to exte
...
+
+```ts title="my-integration.js"
+'astro:server:setup': ({ toolbar }) => {
+  toolbar.onInitialized('my-app', () => {
+    console.log('The app is now initialized!');
+  });
+},
 ```
 
 :::note
```
sarah11918, line 307: Quick note that in the current version of docs, this "note" came after both `astro-dev-toolbar:APP_ID:toggled` and  `astro-dev-toolbar:APP_ID:initialized` as if it applied to both.

In this version, this note is *between* the two. So just double check whether this applies only to `onInitialized` or also to `onToggled`!

sarah11918:
OK! Looking great @Princesseuh ! See my commit above with some light touches (some of them outside of GitHub's editing area, so easier just to make directly going through the page) and confirm everything I did is correct...and desirable! Also a couple of questions/notes/confirmation left unresolved here! But, very close to good to go, I think!

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
Princesseuh: This is okay! The note is only about how you should use the app's event vs Vite, it doesn't matter for other events.

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
Princesseuh: It's okay to omit it, there's no point (at this time) to use the normal `addEventListener` method, the helpers methods cover the entire API.

`EventTarget` is also correct, as it refer to the class here.

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -3,277 +3,369 @@ title: Dev Toolbar App API
...
+
+This function takes an object with an [`init()`](#init) function that will be called when the dev toolbar app is loaded. It can also optionally take a [`beforeTogglingOff()`](#beforetogglingoff) function that will run when the toolbar app is clicked to toggle off the its active status.
+
+### `init()`
 
-This is the core part of the app. This function will be called when the app is loaded, which will either be when the browser is idle or when the user clicks on the app in the UI.
+**Signature:** `init(canvas: ShadowRoot, app: ToolbarAppEventTarget, server: ToolbarServerHelpers) => void`
 
-The function receives two arguments:
+This required function defines the core behavior of the app. This function will be called only once when the app is loaded, which will either be when the browser is idle or when the user clicks on the app in the UI, depending on which one comes first.
```
Princesseuh, line 109: This function is actually not required, it's unlikely you wouldn't use it, but we've actually already seen an app that does not use it. (It does its logic exclusively in its icon's CSS)

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -3,277 +3,369 @@ title: Dev Toolbar App API
...
+{
   // ...
+  entrypoint: './my-app.js',
 }
 ```
 
-### `init: (canvas: ShadowRoot, eventTarget: EventTarget) => void`
+## Structure of a Dev Toolbar App
+
+A Dev Toolbar App is a `.js` or `.ts` file that default exports a [`defineToolbarApp()` function](#definetoolbarapp) which is available from the `astro/toolbar module`.
```
Princesseuh, line 83: You do not export the function, you export its result

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -3,277 +3,369 @@ title: Dev Toolbar App API
...
+    console.log(data.message);
+  });
 }
 ```
 
-### `astro-dev-toolbar:APP_ID:toggled`
+### On the server
 
-This message is sent when the user clicks on the app icon in the UI. The data for this message is a boolean indicating whether the app is enabled or not.
+In [the integration that adds your toolbar app](#toolbar-app-integration-setup), use the [`astro:server:setup` hook](/en/reference/integrations-reference/#astroserversetup) to access the `toolbar` object to send and receive messages to and from your apps.
```
Princesseuh, line 291: It can be in any integration, not just the one where the toolbar app is added. This is not super important, but it's a neat thing to be able to do cross integration/app communication

sarah11918:
OK! I made a commit to address those comments, and otherwise this is ready for your approval!

I'll be making a 4.7.0 branch and pointing these two PRs to them so that they are aware of each other. That will fix the failing link in the other one, and allow me to add a link here to the recipe that won't fail.

sarah11918:
Merging into 4.7.0 branch for final checks!

PR #7824, opened 2024-04-09
Document type checking for the `astro.config` file

bholmesdev:
#### Description (required)

We [got a user bug](https://github.com/withastro/astro/issues/10507) when using Astro DB with the `strictest` type settings and an `.mjs` extension for the Astro config. This seemed like a symptom of a larger problem: if you use`.mjs`, your Astro config is not type checked. Made two tweaks here:
- Add a TypeScript setup section to Astro DB to tell `strictest` users to use a `.ts` extension
- Add a "types in your Astro config file" section to TypeScript to call out options for inferred types

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -19,6 +19,10 @@ Astro starter projects include a `tsconfig.json` file in your project. Even if y
 
 Three extensible `tsconfig.json` templates are included in Astro: `base`, `strict`, and `strictest`. The `base` template enables support for modern JavaScript features and is also used as a basis for the other templates. We recommend using `strict` or `strictest` if you plan to write TypeScript in your project. You can view and compare the three template configurations at [astro/tsconfigs/](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/).
 
+:::tip
+We also recommend using the `.ts` extension for your `astro.config.ts` file when using `strict` or `strictest` TypeScript configurations. This surfaces type hints in your editor and ensures that generated types are correctly injected for certain features, including Astro DB.
+:::
+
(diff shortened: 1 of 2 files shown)


src/content/docs/en/guides/typescript.mdx
```diff
@@ -67,6 +67,20 @@ Then, add the following to your `tsconfig.json`:
 
 To check that the plugin is working, create a `.ts` file and import an Astro component into it. You should have no warning messages from your editor.
 
+### Types in your Astro config file
+
+You may want type hints and errors in your `astro.config.*` file as well. When using the `strict` or `strictest` TypeScript configurations, you can use the `.ts` extension instead of the `.mjs` extension to enable TypeScript parsing. The `.ts` extension is also necessary to support generated types from Astro integrations, [including Astro DB](/en/guides/astro-db/).
```
bholmesdev, line 72: Debating a change from "you can use" to "we _recommend_ using." That would be a strong stance since we DON'T recommend `.ts` today. But perhaps we should!

sarah11918: So in general, we would remove all tone of that sort and just say something like:

"For type hints and errors in your `astro.config` file..." and just say what it requires to get that. We don't have to make any judgments, recommendations etc. Just "To get X, do Y."

Note that we also avoid "you CAN" when we mean "you MUST do Y to get X". That wording is almost exclusively reserved for a choice: "You CAN set a default, but you don't have to."

bholmesdev:
Update: we've [now added JS checking](https://github.com/withastro/astro/issues/10507) to the `strictest` setting as well. We no longer need to document the nuance! Closing this issue

PR #7838, opened 2024-04-10
feat: update minimum Node.js version

ematipico:
#### Description (required)

This PR updates the minimum Node.js version. 

 #### For Astro version: `4.6`. See astro PR [#10689](https://github.com/withastro/astro/pull/10689). -->

### DIFF
--- a/src/content/docs/en/install/auto.mdx
+++ b/src/content/docs/en/install/auto.mdx
@@ -13,7 +13,7 @@ Read our [step-by-step manual installation guide](/en/install/manual/) instead.
 :::
 #### Prerequisites
 
-- **Node.js** - `v18.14.1` or higher.
+- **Node.js** - `v18.17.1` or higher.
 - **Text editor** - We recommend [VS Code](https://code.visualstudio.com/) with our [Official Astro extension](https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode).
 - **Terminal** - Astro is accessed through its command-line interface (CLI).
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/install/auto.mdx
```diff
@@ -13,7 +13,7 @@ Read our [step-by-step manual installation guide](/en/install/manual/) instead.
 :::
 #### Prerequisites
 
-- **Node.js** - `v18.14.1` or higher.
+- **Node.js** - `v18.17.1` or higher.
```
Princesseuh, line 16: Wonder if this should precise the full requirements, which is 18.17.1 or higher for 18, not 19, and 20.3.0 or higher for 20.

ematipico: Good point. If users are on Node.js `v19` now try to install Astro, they will get the engine warning

ematipico: I applied the change in [`3a32833` (#7838)](https://github.com/withastro/docs/pull/7838/commits/3a328331e8fc6791db495c94d9990496fc39de1a), I will let Sarah decide on the format of the sentence.

src/content/docs/en/install/auto.mdx
```diff
@@ -13,7 +13,7 @@ Read our [step-by-step manual installation guide](/en/install/manual/) instead.
 :::
 #### Prerequisites
 
-- **Node.js** - `v18.14.1` or higher.
+- **Node.js** - `v18.17.1` or higher, not `v19`, and `v20.3.0` or higher.
```
sarah11918, line 16:
```suggestion
- **Node.js** - `v18.17.1` or higher, and `v20.3.0` or higher. ( `v19` is not supported.)
```

PR #7843, opened 2024-04-10
Update Lunaria to latest and add Action

yanthomasdev:
#### Description (required)

This PR does two main things:
- Updates our Lunaria version to latest (I checked and no status differences were found between deploy vs local)
- Adds the Lunaria Action (it will now give internationalization insights on every PR to main)

I've also made our Setup Node use Node 20 instead of 18 since that's what I used for the Lunaria Action in my tests, I suppose that won't break anything (hopefully)

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -10,7 +10,7 @@ runs:
     - name: Setup Node
       uses: actions/setup-node@v4
       with:
-        node-version: 18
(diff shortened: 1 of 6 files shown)


sarah11918:
My approval means nothing, so just moral support! :purple_heart:

TheOtterlord (approved):

PR #7856, opened 2024-04-11
Update images.mdx

rishi-raj-jain:
#### Description (required)

Showcases the use of the relative image path without leading `./` in MDX file(s).

I am proposing this change in Astro docs to show how Astro users can use the particular functionality.

Yup, I've made change in only the english docs.

The related PR is https://github.com/withastro/astro/pull/10754


src/content/docs/en/guides/images.mdx
```diff
@@ -530,6 +530,12 @@ import rocket from '../assets/rocket.png';
 
 # My MDX Page
 
+// Local image stored in the src/pages/
+![Houston in the wild](houston.png)
+
+// Local image stored in the src/pages/foo/
+![Houston in the dense forests](foo/houston.png)
```
sarah11918, line 537:
```suggestion
// Local image stored in the the same folder
![Houston in the wild](houston.png)
```
I don't think we should show the second example because it's not really common that someone would keep images in `src/pages/otherfolder/` anyway. I think just showing the same folder is fine, since if you were going to keep images in a different folder, you'd probably NOT put that folder in pages. It just looks weird. But at least this way, collocated images are shown.

sarah11918 (commented):
Suggestion below!

src/content/docs/en/guides/images.mdx
rishi-raj-jain: Alright 👍🏻

sarah11918 (approved):
Thanks, this looks good and will be ready to merge when the next release goes out!

sarah11918:
Looks like this was released, so good to merge now! :partying_face:

PR #7922, opened 2024-04-17
Recipe for building a toolbar app

Princesseuh:
#### Description (required)

Build upon https://github.com/withastro/docs/pull/7821, adding a recipe showing how to build a toolbar app. This is intended as the "Non-Reference Guide to Making A Toolbar App:tm:" as the reference can sometimes be a bit too technical when you're getting started.

I tried my best to toe the line regarding "How to build a package" vs "How to build an app". In a fair amount of cases, apps are inside packages, but explaining how to build a package is kinda not the responsability of this page, so I instead choose to show moreso the use case of creating an app for your own website in the code examples, while giving more general advice in the text. I don't know if I succeeded, but I tried, which is what matters.


src/content/docs/en/recipes/making-toolbar-apps.mdx
```diff
@@ -0,0 +1,229 @@
+---
+title: Creating a dev toolbar app
+description: Learn how to create a dev toolbar app for your site.
+type: recipe
+---
+
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
+import { FileTree } from '@astrojs/starlight/components';
+
+Astro includes a [development toolbar](https://docs.astro.build/en/guides/dev-toolbar/) that you can use to inspect your site, check for accessibility and performance issues, and more.
```
sarah11918, line 10:
```suggestion
Astro includes a [development toolbar](/en/guides/dev-toolbar/) that you can use to inspect your site, check for accessibility and performance issues, and more. 
```

src/content/docs/en/recipes/making-toolbar-apps.mdx
```diff
@@ -0,0 +1,229 @@
...
+      
+      // Send a message to the server
+      server.send("my-app:hello-server", {message: "Hello, server!"});
+    },
+});
+```
+
+These arguments provide multiple utilities and helpers to interact with the dev toolbar and the server, such as executing code when the app is toggled on or off, showing a notification dot on the app, sending messages to the server, and more.
+
+For more information on the hooks and helpers available in the `init()` function, see the [Astro Toolbar API documentation](https://docs.astro.build/api/astro/toolbar).
```
sarah11918, line 135:
```suggestion
For more information on the hooks and helpers available in the `init()` function, see the [Astro Toolbar API documentation](/en/reference/dev-toolbar-app-reference/).
```
Now you're just making things up!  :smile:

Princesseuh: I put it as a placeholder while writing and forgot to change it, ha

src/content/docs/en/recipes/making-toolbar-apps.mdx
```diff
@@ -0,0 +1,229 @@
...
+  hooks: {
+    'astro:config:setup': ({ addDevToolbarApp }) => {
+      const __filename = fileURLToPath(import.meta.url);
+      const __dirname = dirname(__filename);
+  
+      addDevToolbarApp({
+        id: "my-toolbar-app",
+        name: "My Toolbar App",
+        icon: "🚀",
+        entrypoint: join(__dirname, "./app.ts"),
```
matthewp, line 60: All the path code doesn't look great. I don't know why we left all of these entrypoints as always being string. Can we at least clear this up a little by reducing to `fileURLToPath(new URL('./app.ts', import.meta.url))` ?

src/content/docs/en/recipes/making-toolbar-apps.mdx
```diff
@@ -0,0 +1,229 @@
...
+    ```
+  </Fragment>
+  <Fragment slot="yarn">
+    ```shell
+    yarn add --dev typescript
+    ```
+  </Fragment>
+</PackageManagerTabs>
+
+Next, create a `tsconfig.json` file in the root of your project with the appropriate settings to build and for the framework you're using ([React](https://react-typescript-cheatsheet.netlify.app/docs/basic/setup), [Preact](https://preactjs.com/guide/v10/typescript), [Solid](https://www.solidjs.com/guides/typescript)). For example, for Preact:
```
matthewp, line 192: I read "root of your project" as top-level. Would "root of your toolbar app" make it more clear?

src/content/docs/en/recipes/making-toolbar-apps.mdx
```diff
@@ -0,0 +1,295 @@
...
+  "You're awesome!",
+  "You're a star!",
+];
+
+export default defineToolbarApp({
+    init(canvas) {
+      const h1 = document.createElement('h1');
+      h1.textContent = motivationalMessages[Math.floor(Math.random() * motivationalMessages.length)];
+
+      canvas.appendChild(h1);
```
matthewp, line 119:
```suggestion
      canvas.append(h1);
```

src/content/docs/en/recipes/making-toolbar-apps.mdx
```diff
@@ -0,0 +1,319 @@
...
+    ```js title="astro.config.mjs" ins={2,5}
+    import { defineConfig } from 'astro/config';
+    import myIntegration from './my-toolbar-app/my-integration.ts';
+
+    export default defineConfig({
+      integrations: [myIntegration],
+    })
+    ```
+
+4. If not already running, start the dev server. If your integration has been successfully added to your project, you should see a new "undefined" app in the dev toolbar.
```
Princesseuh, line 101: Hmm, does it really show as undefined? That's not intended

sarah11918: It did for me, yes! Just a question mark for the icon, and title "undefined" when I hovered over it. I built a couple of different times from scratch and saw this behaviour both times. If it's not intended, and it will work to show the icon (just nothing happens when you click) then of course we can go back to the original wording!

I wondered whether it made sense to build the app first, before the integration to also avoid the console error, but this also matches the order of things on the API page, and I like how it emphasizes that you're building an integration.

Princesseuh: Aah I see yes, it does that when the file doesn't exist at all. An empty file would probably work so it at least shows the icon and title correctly.

sarah11918: I tried with an empty file too, and got the same result!

sarah11918:
@Princesseuh I added a bit more polish, linked to/from other pages for discoverability etc.

The only comment I'm not sure about is Matthew's -- I tried both `append` and `appendChild` and they seemed to work the same for me in this situation. Can you make the call which one we should use here?

Then, I think this is just ready for your review! The failing test won't happen once the new dev toolbar API page is also available. I'll start pointing both of these PRs to a new 4.7.0 branch and then that link should be fine.

sarah11918:
Merging into 4.7.0 branch! Will perform final reviews/checks/polish in there if needed!

PR #7932, opened 2024-04-18
i18n(es): Fix greeting banner code snippet in blog tutorial [i18nIgnore]

snake:
#### Description (required)

The text at the end of step 3 says to expect a 'random' greeting (I've checked this is the case in en and fr translations), but this won't happen on page load unless randomMessage is used instead of messages[0]. It appears this was already fixed in 'es' translation as a part of c81a980b6a, where the snippet in es already refers to randomMessage().

I'm not sure if patching this all in one PR is acceptable now that I've just seen the "PRs only in one language please" note, but this seems like the obvious fix to me. Let me know if I need to change it.


snake:
I've just realised that this change probably also means the sample code in https://github.com/withastro/blog-tutorial-demo/blob/complete/src/components/Greeting.tsx would need an update too.

As an aside, using randomMessage() instead of messages[0] also serves to highlight a potentially important learning goal which is hinted at in step 5: the fact that when using client:load directive the js there is run twice. Once when building the initial HTML, and again when the js is passed to the client (evident on page load, as you can sometimes see the message briefly load and then change as the front end code is run, getting back a different value from the random call).

at-the-vr (commented):
LGTM 👍

sarah11918:
Hi @snake ! This code is in fact intentional, from #2908.

You're right that the Spanish translation is currently incorrect showing the old version. An update to *that* language file to match the English would be a helpful contribution! 

I'll let you decide whether to update this PR or close this and make a new one with just the update to the `es` file. If you do make a new PR, mentioning this PR #7932, and/or #2908 in that PR description would be great and would help the maintainers know what's going on!  :raised_hands: 

(Also, in future, please **do** follow the request to only PR one language. The reason is not related to the size/nature of the fix, but how our translation tracking works and how our GitHub labeling works. In fact, the reason this PR was not noticed sooner was because it was mislabeled! :sweat_smile: )

snake:
Thanks, @sarah11918. Interesting. So the situation I explained above IS the hydration mismatch in action (the message flicker), and that's been fixed by hardcoding (well, popping the first item) in #2908. Got it. I think there's perhaps merit in making this clear in the tutorial docs then (which currently mention 'random message' in a number of places last I checked) - which I guess is a job for a separate PR.

I'll try to update this PR to fix the 'es' lang pack instead.

Thanks!

PR #7940, opened 2024-04-18
Generate OG images as WebP

delucis:
#### Description (required)

- Updates `astro-og-canvas` to the latest release (required for WebP output)
- Updates our Open Graph endpoint to output WebP images
- Updates our helper for getting the path to a generated image to also use `.webp` instead of `.png`
- Should reduce our cache size significantly. OG images were roughly 1 MB each on average and are now less than 10% of that.

**N.B.** When first building on Vercel the build was still over the build cache limit because I guess it includes all those old OG images which aren’t ever cleared out by `astro-og-canvas`. I reran without the cache and then it successfully cached the result with a size of only c. 250 MB. So we just need to vacate the cache once we merge this to `main` I suspect and we should be good to go.

## But the big news is…

With a warm cache and content collection caching working as added in #7535, I got a build completing in **2m46s** — <https://vercel.com/astrodotbuild/docs/CTPGZ2Z2juaEqgj4xbfqqBMcxi3v>


delucis (commented):
Seeing some errors in the build now it redeployed with content collection caching enabled, for example on the homepage:

image

From a quick look through, it seems to all be missing CSS for components in MDX. Their JS _is_ loading though — only styles are missing.

Something for @matthewp to maybe look at now we have this test case.

delucis:
Update branch to bring in #7952 and now the deployment preview looks OK again!

sarah11918:
Did some checks in online preview generators and all looks good I think!

image
image
image

delucis:
Oooh, nice work — thank you @sarah11918!

PR #7972, opened 2024-04-22
Manually configure MDX integration

bluwy:
#### Description (required)

With the upcoming upgrade to `astro-expressive-code` from Starlight (https://github.com/withastro/starlight/pull/1720), I decided to test it out and found issues with the syntax highlighting. Because our `rehype-optimize-static` plugin runs before `rehype-expressive-code`, it wasn't able to highlight any code blocks because the elements has been optimized as plain strings.

`rehype-optimize-static` needs to be last for it to work correctly. This is already guaranteed by the `@astrojs/mdx` `optimize` option. So in this PR, I decided to use it instead.

However, manually configuring `@astrojs/mdx` is tricky because it needs to be after `astro-expressive-code`, however Starlight will always append `astro-expressive-code`, causing an error with the ordering. I fixed this by patching with `integrations/mdx.ts`, but this could perhaps be better supported in Starlight in the future.

---

Performance (Rollup build):
- Before: 114s
- After: 109s
- After + `astro-expressive-code` latest: 99s

cc @hippotastic (Awesome work on the optimizations!)

- Closes #
- Suggested label:


sarah11918:
@delucis @hippotastic - Can I leave this to you two?

bluwy:
Closing this for now. The changes can be done in another PR that upgrades Starlight instead.

Issue #7981, opened 2024-04-23
Abount Fontsource package name


magurotabetai:
### 📚 Subject area/topic

Usinc custom fonts

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/fonts/

### 📋 Description of content that is out-of-date or incorrect

Since Variable Fonts in fontsource start with @fontsource-variable/,  I thought it might be better to display "It will start with @fontsource/ or @fontsource-variable/ followed by the name of the font." instead of the tip "It will start with @fontsource/ followed by the name of the font."

delucis:
Thank you for the issue @magurotabetai — would you like to make a PR with this change?

I checked the Fontsource docs and confirmed they are now also using that scope, see for example https://fontsource.org/fonts/inter/install

magurotabetai:
I would like to gain experience contributing to OSS, so I would like to create a PR for sure!

PR #7982, opened 2024-04-23
Update Fontsource tip to add @fontsource-variable scope

magurotabetai:
#### Description (required)

Since Variable Fonts in fontsource start with @fontsource-variable/, I thought it might be better to display `It will start with @fontsource/ or @fontsource-variable/ followed by the name of the font.` instead of the tip `It will start with @fontsource/ followed by the name of the font.`

- Closes # #7981 
- Suggested label:


sarah11918 (approved):
Thank you so much for taking the time to make the docs more accurate and helpful @magurotabetai ! We really appreciate the contribution, and are happy to welcome you to Team Docs!  :partying_face:

PR #7983, opened 2024-04-23
feat: add small mention of available preferences

Princesseuh:
#### Description (required)

Astro 4.7 adds a new preference: `checkUpdates`. We currently don't have a place where the list of preferences are listed, and in core we honestly don't add preferences all that often, so I figured just a small list in the CLI reference was good enough, happy to do something else, though.


sarah11918:
Merging into the 4.7.0 branch for final checks/polish there!

PR #7986, opened 2024-04-23
Fix typing mismatch in Integration API docs

BlenderDude:
While reading through the integration docs, I noticed that `astro:config:setup` appeared synchronous in the documentation, while the actual type allows a `Promise`.


TheOtterlord (approved):
Thanks @BlenderDude!

PR #8011, opened 2024-04-25
Add info to SSR page on using Astro.response

ascorbic:
#### Description (required)

The SSR docs currently have info on using `Astro.request`, but nothing about `Astro.response`. There is detail in the API guide, but I think it's helpful to have it on the SSR page too, so this PR adds it.

- Closes #
- Suggested label:


src/content/docs/en/guides/server-side-rendering.mdx
```diff
@@ -222,7 +222,43 @@ See more details about [`Astro.cookies` and the `AstroCookie` type](/en/referenc
...
+  Astro.response.status = 404;
+  Astro.response.statusText = 'Not found';
+}
+---
+<html>
+  <!-- Page here... -->
+</html>
+```
+
+Setting these values will change the response status and status text for the page.
```
sarah11918, line 244:
``````suggestion
[`Astro.response`](/en/reference/api-reference/#astroresponse) is a standard [`ResponseInit`](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response#options) object. It can be used to set the response status and headers.

The example below sets a response status and status text for a product listing page when the product does not exist:

```astro title="src/pages/my-product.astro" {8-9}
---
import { getProduct } from '../api';

const product = await getProduct(Astro.params.id);

// No product found
if (!product) {
  Astro.response.status = 404;
  Astro.response.statusText = 'Not found';
}
---
<html>
  <!-- Page here... -->
</html>
```
``````
Great example here! Fix this for accuracy, but we typically like to introduce code snippets with the explanation of what's happening first to prime them for reading the code. This keeps the reader flowing down the page, instead of seeing code, then reading an explanation, then scrolling back up to check what the explanation referred to.

I also presented a suggestion for making the text context-specific for the example, but feel free to use something that similarly explains the code in context! (Including the file name of the snippet!)

sarah11918 (commented):
This is *really* well done, @ascorbic ! So easy to read and great examples. This is a real improvement to the page! :raised_hands: 

Just made one small suggestion to align with our pattern of explanations moving up ahead of the code block, instead of after.

The second code example is pretty small and I don't think it *needs* a similar sentence (as also supported by you not feeling the need to write one before.) But if we're going to have one for the first code block, let's do it *before* the sample, and give it some context! :partying_face:

src/content/docs/en/guides/server-side-rendering.mdx
ascorbic: That is better, thanks!

sarah11918 (approved):
Looks great, thanks for the helpful contribution!

PR #8014, opened 2024-04-25
Enforce __session name for cookie

leoxs22:
#### Description (required)

Docs snippets are saving a cookie called `session`

Firebase will remove all cookies except for `__session` https://firebase.google.com/docs/hosting/manage-cache#using_cookies

Using the current code in the docs will result in a 500 error. Adding a log in a middleware will show that no cookies are being received in the firebase function.


src/content/docs/en/guides/backend/google-firebase.mdx
```diff
@@ -219,14 +219,18 @@ export const GET: APIRoute = async ({ request, cookies, redirect }) => {
...
+  cookies.set("__session", sessionCookie, {
     path: "/",
   });
 
   return redirect("/dashboard");
 };
 ```
 
+:::caution
+Cookie must be called '__session'. Firebase functions [will remove all cookies except for that one](https://firebase.google.com/docs/hosting/manage-cache#using_cookies).
```
Fryuni, line 231: To me, this reads like "a session cookie must have this name" when, in fact, it is the only cookie allowed _at all_. 

I think the warning should be a little stronger.

```suggestion
Firebase only allows the use of [one cookie, and it must be named `__session`](https://firebase.google.com/docs/hosting/manage-cache#using_cookies). Any other cookies the client sends will not be visible to your application.
```

Fryuni (changes requested):
Hey @leoxs22, thanks for the PR, and welcome to the team 🎉 🎉 🎉 

The code seems to be working fine with the fix. I sent a suggestion to the text because it implies more than it should (at least to me, maybe I'm being a bit paranoid). Besides that LGTM

leoxs22:
Thanks! I agree, it looks better that way :smile:

Fryuni (approved):
LGTM!

PR #8015, opened 2024-04-25
Fix double/triple click selection issues on headings

### LINKED ISSUE #8001: Double/triple click headings copy issues
### 📝 Issue Description

It looks like there may be some issues when copying headings when using either double click (select a word) or triple click (select a paragraph/logical unit). The behavior seems to be different depending on the browser and be related to heading anchor links.

### Chrome

When the viewport width is less than 95em (the heading anchor link is at the end of the heading) and double clicking and copying a heading with a single word or the last word of a multiple words heading will inject `Section` in the clipboard text (I am assuming this is the first word of the heading anchor link screen reader text).

For example, going to this [page](https://docs.astro.build/en/concepts/why-astro/#features) and double clicking and copying the heading `Features` will copy `FeaturesSection` instead of `Features`. Double clicking and copying the word `Principles` in the heading `Design Principles` will copy `PrinciplesSection` instead of `Principles`.

Triple clicking is working as expected.

### Firefox

When the viewport width is less than 95em (the heading anchor link is at the end of the heading) and triple clicking a heading and copying will inject the entire heading anchor link screen reader text in the clipboard text.

For example, going to this [page](https://docs.astro.build/en/concepts/why-astro/#features) and triple clicking and copying the heading `Features` will copy `FeaturesSection titled Features` instead of `Features`.

### Safari

When the viewport width is less than 95em, double clicking or triple clicking a heading with a single word or the last word of a multiple words heading will visually select the entire page and copying will inject `Section` in the clipboard text too.

SCR-20240424-lddg

When the viewport width is greater than 95em, triple clicking a heading with a single word or the last word of a multiple words heading will visually select the entire page (the clipboard content is fine in this case).

I used https://docs.astro.build/en/concepts/why-astro/#features as an example but this can repro on every pages with headings.

### 🤔 Expected Behavior

The selection should visually match what is expected (either a word or an entire heading but never the entire page) and the copied text should match such selection and never contains invisible text.

### 👀 Current Behavior

The selection can be visually wrong and the copied text can include invisible text.

### 🖥️ Browser

Chrome 125.0.6422.4 -  Firefox 125.0.2 - Safari 17.4.1

### 📄 Additional Information

_No response_

HiDeoo:
#### Description (required)

This PR fixes some text selection issues when double or triple clicking headings.

A short summary of the issue would be that double/triple click selection on headings could include some extra content when copying the selection. Refer to #8001 for a detailed list of all the issues by browser as the behavior is different in each one and I don't think it is worth duplicating the information here.

The changes fixes all the issues at the same time. I tried various approaches before settling on this one:

- A basic space does not fix the issue.
- A [zero-width space](https://en.wikipedia.org/wiki/Zero-width_space) would fix the issue in Chrome and Firefox, but not in Safari. It also has the downside of adding a character, non-visible (and non-conventional) to the text which browsers may or may not include when copying the selected heading.
  On top of that, some editors or note-taking apps may highlight such a character if the browser used includes it in the copied text.
  For example, here is how it looks in my VS Code when copying the heading with a zero-width space (notice the yellow rectangle at the end of the heading):

  SCR-20240425-qbzi

- The CSS approach used in this PR.

Accessiblity-wise, I couldn't think of any downside of using this approach in this specific case on this specific class (nor could I find any mention of this in my researches) and testing with VoiceOver and NVDA didn't show any change in behavior whatsoever.

For reference, here is a preview of the VoiceOver behavior before and after the change:

headings-voiceover

- Closes #8001
- Suggested label: bug a11y


TheOtterlord (approved):
Confirmed on Firefox & Chrome. Thanks for tackling this!

delucis (approved):
Smart fix! Also wonder if `.sr-only` should always include `user-select: none`, but I’m not 100% confident there as those are quite different semantically — maybe there’s some cases where text selection could make sense with visually hidden text, not sure.

HiDeoo:
> Also wonder if `.sr-only` should always include `user-select: none`, but I’m not 100% confident

Same, and even after talking with people having a better experience with this than me, the specificity of the selector seems like a wise choice for the time being.

PR #8040, opened 2024-04-27
i18n(pt-BR): Update `aliases.mdx` translation

Egpereira:
# Update `aliases.mdx` translation

Updates the pt-br translation of the`aliases.mdx` file.

I switched to not translating the word "alias" on the file, for consistency, since it was sometimes using it as "alias" and sometimes as "atalhos", which doesn't seems like the best word for it. Furthermore, I believe "alias" is a term commonly used in english even for Portuguese speakers, but I can change it back if needed.


yanthomasdev (approved):
LGTM! Thank you 💜 Pointing out that the sidebar entry for this page in Portuguese is still `Atalhos de Importação`, feel free to make a PR updating that as well!

PR #8056, opened 2024-04-30
feat: defineIntegration

florian-lefebvre:
#### Description (required)

- [x] Replaces occurences of `AstroIntegration` by `defineIntegration`
- [x] Add a section documenting usage of `defineIntegration`
- [x] Updates Astro DB integration docs
- [ ] TODO: show how to use a custom type using `z.custom<MyType>()`

- Suggested label: `add new content`, `minor release`


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -611,6 +622,129 @@ A list of all generated pages. It is an object with one property.
 
 - `pathname` - the finalized path of the page.
 
+## Helpers
```
sarah11918, line 625: Just getting ahead of this a bit, not full review:

This says "helpers" (plural), but then only seems to define `defineIntegration()`. Also, since this is an API reference page, thinking about what's scannable in the TOC, I think it might make more sense to have the actual code options visible vs descriptive headings. (People may often be using these pages to *look things up*, and seeing the various helpers/arguments etc. in the TOC makes that very easy to do.)

Would a structure like this make sense?

```
## `defineIntegration()`
<p><Since v="4.8.0" /></p>

### `optionsSchema`
<p><Since v="4.8.0" /></p>

### `setup`
<p><Since v="4.8.0" /></p>
```

florian-lefebvre: I have been thinking about this and my thinking was that, since we have a new entrypoint (`astro/integration`), more things could be exported in the future. So I decided to create a section for all the things that may end up in there. Let me know what you think! But I don't think it's worth reviewing just yet because we need to go through all the rfc stages

florian-lefebvre:
There's still no consensus on this for now, so I'm closing it until we want to revisit. RFC at https://github.com/withastro/roadmap/discussions/886

PR #8062, opened 2024-04-30
Remove unnecessary character from integrations-reference.mdx

Egpereira:
# Remove unnecessary character from `integrations-reference.mdx`

Removes an extra backtick ("\`")  character from a code snipper in `integrations-reference.mdx`


sarah11918 (approved):
Thank you!

PR #8086, opened 2024-05-01
Add Steps component to `publish-to-npm.mdx`

Egpereira:
# Add Steps component to `publish-to-npm.mdx`

Adds the `Step` component where needed.

### DIFF
--- a/src/content/docs/en/reference/publish-to-npm.mdx
+++ b/src/content/docs/en/reference/publish-to-npm.mdx
@@ -242,14 +242,14 @@ In the meantime, our current recommendation for testing is:
 3. Each page should include some different component usage that you'd like to test.
 
 4. Run `astro build` to build your fixtures, then compare the output of the `dist/__fixtures__/` directory to what you expected.
+   <FileTree>
+   - my-project/demo/src/pages/\_\_fixtures\_\_/
+     - test-name-01.astro
+     - test-name-02.astro
+     - test-name-03.astro
+   </FileTree>
 </Steps>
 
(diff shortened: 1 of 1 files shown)


at-the-vr (commented):
LGTM 👍

src/content/docs/en/reference/publish-to-npm.mdx
```diff
@@ -233,10 +234,15 @@ Astro does not currently ship a test runner. _(If you are interested in helping
...
 1. Add a test `fixtures` directory to your `demo/src/pages` directory.
+
 2. Add a new page for every test that you'd like to run.
+
 3. Each page should include some different component usage that you'd like to test.
+
 4. Run `astro build` to build your fixtures, then compare the output of the `dist/__fixtures__/` directory to what you expected.
+</Steps>
 
 <FileTree>
```
HiDeoo, line 247: Feels like the file tree should be included in the steps 4 as it's the output that you should compare to, and this comparison is part of the step 4. Any thoughts?

sarah11918: Yes! It wasn't in the original but I would say this does belong with Step 4, so let's include the file tree inside the steps.

TheOtterlord: Moved inside steps

Egpereira: Agreed, makes total sense

sarah11918:
!coauthor

PR #8130, opened 2024-05-01
Test build with NODE_OPTIONS

bluwy:
#### Description (required)

Checking if this one weird trick can speed up the Vercel CI


bluwy:
Unfortunately there's no improvement

PR #8131, opened 2024-05-01
Set docgen files to read-only

natemoo-re:
#### Description (required)

Not a perfect solution, but this change should help _some_ users avoid editing auto-generated files. These files will now be shown as read-only when attempting to edit the file in VS Code.


delucis (approved):
Checked out this branch locally and it worked great!

File tab shows a 🔒 and trying to type doesn’t work, displaying an explanatory tooltip:

VS Code showing “editor is read-only” tooltip and file tab with a padlock icon

sarah11918:
Just realized that it's *also* going to stop *me* from hotfixing things... but we'll see how it goes for now.  😅

sarah11918:
!coauthor

PR #8139, opened 2024-05-01
Remove obsolete warning about using `import.meta` in config files

kripod:
#### Description (required)

The Astro config file is now loaded with Vite if Node fails to resolve it:

- https://github.com/withastro/astro/blob/5248ed1fb35bc1407660a189f23155812787f9e4/packages/astro/src/core/config/config.ts#L146
- https://github.com/withastro/astro/blob/5248ed1fb35bc1407660a189f23155812787f9e4/packages/astro/src/core/config/vite-load.ts#L40-L66

When referencing Vite-specific properties like `import.meta.env.anything` or calling `import.meta.glob(…)`, Node throws an error unless the presence of the former variables is checked e.g. via optional chaining: `import.meta.env?.anything` or `import.meta.glob?.(…)`. In the latter cases, nullishness of those values is anticipated explicitly, while the former, more common usages should be handled by Vite as the fallback mechanism triggers.

_(Previously, Vite was [only used for TS-based configs](https://github.com/withastro/astro/pull/5377/files#diff-fd8d6e4dae76cfab252a32fdf56bf4b35cbbfcefbdc9867498cd1ed956db8eb2R101-R119), but the code has changed since then.)_

- Suggested label: 'improve documentation'

Yes.

### DIFF
--- a/src/content/docs/en/guides/configuring-astro.mdx
+++ b/src/content/docs/en/guides/configuring-astro.mdx
@@ -116,10 +116,6 @@ export default defineConfig({
 })
 ```
 
-:::note 
-Vite-specific `import.meta` properties, like `import.meta.env` or `import.meta.glob`, are _not_ accessible from your configuration file. We recommend alternatives like [dotenv](https://github.com/motdotla/dotenv) or [fast-glob](https://github.com/mrmlnc/fast-glob) for these respective use cases. In addition, [tsconfig path aliases](https://www.typescriptlang.org/tsconfig#paths) will not be resolved. Use relative paths for module imports in this file.
-:::
-
 ## Customising Output Filenames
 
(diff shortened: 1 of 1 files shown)


matthewp:
Depending on fallback is subject to your code breaking in the future. `import.meta.env` is not supported inside of Astro configs officially. The only reason we load with Vite is for `.ts` IIRC. Maybe @bluwy knows of another reason. But we definitely don't do it so you can take advantage of Vite features.

bluwy:
Yep exactly 👍 You shouldn't rely on Vite-specific features in the config as Vite is an implementation detail to load TS files.

sarah11918:
Thanks for participating everyone! Since this warning should stay in place, I'll close *this* PR, but I hope to see more contributions from you in the future!

PR #8096, opened 2024-05-01
Add Steps component to `testing.mdx`

Egpereira:
# Add Steps component to `testing.mdx`

Adds the `Step` component where needed.

PR #8187, opened 2024-05-06
Update lists in `testing.mdx`

Egpereira:
# Update lists in `testing.mdx`

As commented here: https://github.com/withastro/docs/pull/8096#issuecomment-2091086809, this PR proposes a change to two lists in the `testing.mdx` file:

1. Changes the ordered list in the NightwatchJS section to be an unordered list and and changes the wording from "CLI steps" to "CLI commands", since they are not needed in a particular order.
2. Updates the ordered lust in the Playwright section to use the `Steps` component, since they are actually instructions than only can run one after the other.

### DIFF
--- a/src/content/docs/en/guides/testing.mdx
+++ b/src/content/docs/en/guides/testing.mdx
@@ -223,21 +223,19 @@ You can install NightwatchJS within your Astro project using the package manager
 
 You can run a single test or several tests at once, testing one or multiple browsers. By default, your test results will be shown in the terminal. Optionally, you can open the HTML Test Reporter to show a full report and filter test results.
 
-You can run the tests with the [NightwatchJS VSCode Extension](https://marketplace.visualstudio.com/items?itemName=browserstackcom.nightwatch) or using the CLI steps below:
+You can run the tests with the [NightwatchJS VSCode Extension](https://marketplace.visualstudio.com/items?itemName=browserstackcom.nightwatch) or using the CLI commands below:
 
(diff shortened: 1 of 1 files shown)


Egpereira:
@HiDeoo only managed to open the proposal now!

Egpereira:
> This change makes sense! Well spotted!
> 
> I think the code blocks will need to be UN-indented, though, because the indentation was because of being inside of an ordered list.

you mean this one?
unordered list after change

should I un-indent one tab on the code blocks (they are with currently two, doesn't affect the visual outcome)?
code of the unordered list change

sarah11918:
Oh, you're right! I was thinking that, but seeing it, it does fit.  Thank you!

sarah11918 (approved):
Thank you, this one looks great!

sarah11918:
Hmm, just noticing that the middle one is viewing the test report. As you indicated in the section below, does this require first running one of the above tests?

I wonder whether we should at least reorder and put viewing the report at the bottom?

sarah11918:
I think it's not too weird! At some point we may revisit the page as a whole, but I say for now, your judgement makes sense! 2 steps for running/reporting, then present the other as an alternate.

Egpereira:
> I think it's not too weird! At some point we may revisit the page as a whole, but I say for now, your judgement makes sense! 2 steps for running/reporting, then present the other as an alternate.

Commited the changes as discussed!

sarah11918:
Looks great, let's do it!  💪

PR #8190, opened 2024-05-06
Update banner to mention Astro Together

TheOtterlord:


sarah11918 (approved):
Deploy preview isn't live yet, but looks perfect to me!

I'll let others weigh in on date format. Some prefer to show a year, no matter what, e.g. May 28, 2024.

PR #8191, opened 2024-05-06
Shorter Relative Path

felipeaprotazio:
#### Description (required)
The relative path `const allPosts = await Astro.glob('./posts/*.md');` works as well as `const myPosts = await Astro.glob('../pages/posts/*.md');`, which is the original. The latter unnecessarily returns to the pages folder.

- Suggested label: code snippet update, good first issue, help leave feedback.

felipeaprotazio


sarah11918:
Hi @felipeaprotazio , thank you for submitting this fix! Yes, you're right, we can make the path shorter here. 

In fact, I reviewed everywhere to make sure that if we change this line, we change it in every place, and there is one more place it needs updating: line 51.

Would you please make the change there too, and then I think we have every occurance!

sarah11918 (approved):
Thank you for this contribution @felipeaprotazio , and welcome to Team Docs!

PR #8192, opened 2024-05-06
Document new `getViteConfig()` option for Vitest

delucis:
#### Description (required)

Updates the Vitest testing docs to mention a new config option.

#### For Astro version: `4.8`. See astro PR https://github.com/withastro/astro/pull/10963.

I did not use `<Since />` in this case because it doesn’t really work well inline in the way I integrated these new docs. Any thoughts on whether this should be added somehow? There _could_ be a separate heading to fit it under, but that felt a little artificial.


src/content/docs/en/guides/testing.mdx
```diff
@@ -28,6 +28,19 @@ export default getViteConfig({
 });
 ```
 
+By default, `getViteConfig()` will try to load an Astro config file in your project and apply it to the test environment.
```
sarah11918, line 31:
```suggestion
<p><Since v="4.8.0" /><p>

By default, `getViteConfig()` will try to load an Astro config file in your project and apply it to the test environment.
```

It's not amazing, but we *could* do it this way

delucis: Ah yeah, that’s why that would be confusing. The “By default…” clause has always been the case — that hasn’t changed. The new part is the second argument allowing you to override that default with inline options. So I guess if anything it would happen mid paragraph?

> By default, `getViteConfig()` will try to load an Astro config file in your project and apply it to the test environment.
> As of Astro 4.8, if you need to customize the Astro configuration applied in your tests, pass a second argument to `getViteConfig()`:

sarah11918: Yes, your suggestion works perfectly well!! If you cared, one other option could be to split up into two paragraphs like:


> By default, getViteConfig() will try to load an Astro config file in your project and apply it to the test environment. 

 `<Since v="4.8.0"/>`

> `getViteConfig()` can also receive a second argument to customize the Astro configuration applied in your tests. Pass an object with any additional [configuration options](/en/reference/configuration-reference/):

src/content/docs/en/guides/testing.mdx
```diff
@@ -28,6 +28,19 @@ export default getViteConfig({
...
+If you need to customize the Astro configuration applied in your tests, pass a second argument to `getViteConfig()`:
```
delucis, line 32:
```suggestion
As of Astro 4.8, if you need to customize the Astro configuration applied in your tests, pass a second argument to `getViteConfig()`:
```

src/content/docs/en/guides/testing.mdx
delucis: Thanks! Added it to the paragraph — I think that makes more sense than breaking it up.

sarah11918 (approved):
Looks great, thanks @delucis  !

sarah11918:
updating branch and pointing to 4.8.0 branch!

src/content/docs/en/guides/testing.mdx
```diff
@@ -28,6 +28,19 @@ export default getViteConfig({
 });
 ```
 
+By default, `getViteConfig()` will try to load an Astro config file in your project and apply it to the test environment.
+As of Astro 4.8, if you need to customize the Astro configuration applied in your tests, pass a second argument to `getViteConfig()`:
+
+```js
```
sarah11918, line 34:
``````suggestion
```js title="vitest.config.ts"
``````
Should this be the title, @delucis ?

PR #8207, opened 2024-05-07
i18n(pl): update getting-started.mdx

gacek1123:
#### Description (required)

Update the Polish "Getting Started" page to the new layout.

- Suggested label: i18n

### DIFF
--- a/src/content/docs/pl/getting-started.mdx
+++ b/src/content/docs/pl/getting-started.mdx
@@ -56,7 +56,7 @@ import Discord from '~/components/Landing/Discord.astro'
   <ListCard title="Rozszerzanie" icon="puzzle">
     - [Dodaj integracje jak React czy Tailwind](/pl/guides/integrations-guide/)
     - [Twórz otypowane kolekcje zawartości](/pl/guides/content-collections/)
-    - [Ulepsz nawigację dzięki przejściom widoku](/pl/guides/view-transitions/)
+    - [Ulepsz nawigację dzięki View Teansitions](/pl/guides/view-transitions/)
     - [Połącz bezgłowy CMS ze swoim projektem](/pl/guides/cms/)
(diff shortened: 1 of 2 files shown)


PaTRlX (approved):

src/content/docs/pl/getting-started.mdx
```diff
@@ -2,108 +2,63 @@
...
+    - [Główne funkcje Astro](/pl/concepts/why-astro/)
+    - [Architektura wysp](/pl/concepts/islands/)
+    - [Komponenty Astro](/pl/basics/astro-components/)
+    - [Składnia plików Astro](/pl/basics/astro-syntax/)
+  </ListCard>
+
+  <ListCard title="Rozszerzanie" icon="puzzle">
+    - [Dodaj integracje jak React czy Tailwind](/pl/guides/integrations-guide/)
+    - [Twórz otypowane kolekcje zawartości](/pl/guides/content-collections/)
+    - [Ulepsz nawigację dzięki przejściom widoku](/pl/guides/view-transitions/)
```
V3RON, line 59: I suggest using 'Ulepsz nawigację dzięki View Transitions' instead. Usually proper names stay in the original form. The translated one sounds weird to me.

What do you think?

gacek1123: Yeah, that'd make sense. I was actually considering doing that at first but then saw that "view transitions" was written in lowercase, which made me not treat is as a name and translate it too.

Let me take a look at how people solved it for the other languages.

gacek1123: Opinions seem to be mixed, although in most languages the name is left as-is. In that case, I think we shouldn't stick out.
And I have to agree, it does sound kind of weird

V3RON (approved):

PR #8211, opened 2024-05-07
Update cloudflare adapter docs for new import types

adrianlyjak:
#### Description (required)

Updates cloudflare adapter docs for features to-be-added in https://github.com/withastro/adapters/pull/251. This change enhances `wasmModuleImports` to support more extensions. The change is non-breaking, since the existing `wasmModuleImports` option will be respected if defined. The change also changes the default to enabled. I chose to exclude mention of the deprecated `wasmModuleImports` to keep it simple

#### First-time contributor to Astro Docs

I contributed here once before to add the `wasmModuleImports` docs, I think this was before this repo was separate. Discord user is "lyjackal"


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -338,14 +326,23 @@ You can [specify additional routing patterns to follow](#routesextend) in your a
...
-## Use Wasm modules
+## Cloudflare Module Imports
+
+The cloudflare worker runtime supports imports of some [non-standard module types](https://developers.cloudflare.com/pages/functions/module-support/). Most additional file types are also available in astro:
+
+- `.wasm` or `.wasm?module`: exports a [`WebAssembly.Module`](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Module) that can then be instantiated
+- `.bin`: exports an [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) of the raw binary contents of the file
+- `.txt`: Exports a string of the file contents
+
+All module types export a single default value. Modules can be imported both from server side rendered pages, or from pre-rendered pages for static site generation.
```
sarah11918, line 337:
```suggestion
All module types export a single default value. Modules can be imported both from server-side rendered pages, or from prerendered pages for static site generation.
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -204,28 +204,16 @@ export default defineConfig({
...
+### `cloudflareModules`
 
 <p>
 **Type:** `true | false`<br />
-**Default:** `false`
+**Default:** `true`
 </p>
 
-Whether or not to import `.wasm` files [directly as ES modules](https://github.com/WebAssembly/esm-integration/tree/main/proposals/esm-integration) using the `.wasm?module` import syntax.
+Enables [imports of `.wasm`, `.bin`, and `.txt` modules](#cloudflare-module-imports).
```
sarah11918, line 214: I'll ask @alexanderniebuhr to also review this line because admittedly I'm not familiar with this, so I can only go on language!

Just checking: is this meant to be essentially the same as the option it replaces, only adding extra kinds of files that can imported? If so, is it intentional that this no longer says "directly as ES modules"? Is that an important nuance/detail that should still be included here?

sarah11918 (commented):
Thank you @adrianlyjak ! This looks pretty good and will be a welcome addition to the docs! Just one line I think is worth clarifying before I sign off on it, and I fixed some tiny nits in another sentence. But, soon you'll get your *official* welcome to Team Docs! 🥳

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
adrianlyjak: You are correct that this effectively just adds a couple of new export types. I removed this line because it’s only relevant to Wasm. If you are `import`ing, that’s inherently an ES module. I think the reason that this is mentioning ES modules in the context

adrianlyjak: You are correct that this effectively just adds a couple of new exported file types. I removed the sentence and link because it's a proposal that's only for Wasm ES modules. (It contains info about the the specific nuances of how Wasm is exported). Yes, all three types are ES module (that's somewhat inherent to `import`ing them).

I think the breakdown of each exported type in the subsequent section is sufficient and more to the point, but it might make sense to link to this proposal within the Wasm bullet? I think we should leave it out for now, reading through the Wasm ES Module proposal, I think the semantics are starting to diverge from cloudflare's implementation. Effectively in the proposal, a vanilla import looks like it will pre-instantiate the wasm module (which is kind of nice, but not what cloudflare does)

proposal
```js
import wasmInstance from "./myModule.wasm";
wasmInstance.add(1, 2)
// or, if you want to get the "source" module in order to instantiate it, there's some sort of special "source" keyword
import source addModule from "./myModule.wasm");
const instantiated = new WebAssembly.Instance(myModule);
instantiated.add(1, 2)
```

As opposed to cloudflare, which always exports the source module. Note that there's no `source` keyword

```js
import addModule from "./myModule.wasm");
const instantiated = new WebAssembly.Instance(myModule);
instantiated.add(1, 2)
```

sarah11918: Gotcha! Thanks for the explanation. I'm not sure an external link is necessary here, I just wanted to make sure that the deletion of the extra detail was intentional. I'll wait for Alex to confirm that this expresses what he wants conveyed, and if he's happy, I'm happy! :raised_hands:

alexanderniebuhr (approved):
Looks to me too, @sarah11918 anything missing from your side?

PR #8217, opened 2024-05-08
Update deploy/vercel.mdx

liruifengv:
#### Description (required)

- Update deploy/vercel.mdx

This section is outdated, the current Astro version is 4.7.1 and I think we should update it

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/deploy/vercel.mdx
+++ b/src/content/docs/en/guides/deploy/vercel.mdx
@@ -94,8 +94,3 @@ After your project has been imported and deployed, all subsequent pushes to bran
 You can use `vercel.json` to override the default behavior of Vercel and to configure additional settings. For example, you may wish to attach headers to HTTP responses from your Deployments. 
 
 <ReadMore>Learn more about [Vercel’s project configuration](https://vercel.com/docs/project-configuration).</ReadMore>
-
-### Upgrade Astro version
-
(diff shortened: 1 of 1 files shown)


liruifengv:
netlify has the same issue

src/content/docs/en/guides/deploy/vercel.mdx
```diff
@@ -6,6 +6,7 @@ i18nReady: true
 ---
 import ReadMore from '~/components/ReadMore.astro';
 import { Steps } from '@astrojs/starlight/components';
+import Version from '~/components/Version.astro';
```
sarah11918, line 9:
```suggestion
```

sarah11918 (approved):
Thanks @liruifengv ! As discussed on Discord, this doesn't even belong here any more! It was from before we had good upgrade guides.

To save us "upgrading" this section every time, we should just delete it and let the new Updating Astro page do this work!  :muscle: 

So let's just delete it entirely, and go ahead and merge this when you're ready!

sarah11918 (approved):
Thank you @liruifengv ! As discussed, we now have an Upgrade Astro page for all this information where it can be more easily kept up to date in one place. So lets just deleted it, then go ahead and merge when you're ready!

sarah11918 (approved):
Thank you @liruifengv ! As discussed, we now have an Upgrade Astro page for all this information where it can be more easily kept up to date in one place. So lets just deleted it, then go ahead and merge when you're ready!

dreyfus92 (approved):
Thanks everyone! LGTM ✅

PR #8218, opened 2024-05-08
Add support for nesting and collapsing sidebar groups

delucis:
### Description (required)

Updates our utility function mapping our `nav.ts` files to Starlight’s sidebar config shape to support nesting and collapsing groups.

#### Syntax

- Nesting/collapsing is controlled from `src/i18n/en/nav.ts`. Other languages continue to work the same way, providing just the key–value mapping of translations.

- Each heading (entries with `header: true`) can be marked as a nested group by adding `nested: true`.

- Each heading (entries with `header: true`) can be marked as collapsed by default by adding `collapsed: true`.

- I took a few shortcuts to keep this simple, which means nested groups have to be at the end of their containing group, i.e. the following is possible:

   ```
   Parent group     ▼
   |   Link 1
   |   Link 2
   |
   |   Sub group    ▼
   |   |   Link 3
   |   |   Link 4
   |
   |   Other group  ▼
   |   |   Link 5
   |
   Other group      ▼
   |   Link 6
   ```
   
   But this is not:
   
   ```
   Parent group     ▼
   |   Link 1
   |   Link 2
   |
   |   Sub group    ▼
   |   |   Link 3
   |   |   Link 4
   |
   |   Link 5
   |   Link 6
   |
   Other group      ▼
   |   Link 7
   ```
   
   Based on what I’ve seen so far, that seemed to align with intended usage, but let me know if it’s not!

### DIFF
--- a/config/sidebar.ts
+++ b/config/sidebar.ts
@@ -22,22 +22,30 @@ type StarlightSidebarConfig = NonNullable<Parameters<typeof starlight>[0]['sideb
 
 /** Generate a Starlight sidebar config object from our existing `nav.ts` files. */
 export function makeSidebar(): StarlightSidebarConfig {
+	let currentSubGroup: Extract<StarlightSidebarConfig[number], { items: any }>;
 	return navTranslations.en.reduce((sidebar, item) => {
 		if ('header' in item) {
-			sidebar.push({
+			const newGroup = {
 				label: item.text,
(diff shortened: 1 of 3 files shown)


TheOtterlord (approved):
LGTM! Managed to test implement the sidebar redesign with these changes

PR #8226, opened 2024-05-08
docker.mdx: use `--omit=dev` for npm over deprecated `--production`

corneliusroemer:
#### Description (required)

This fixes npm warnings that I get using the current suggested Dockerfile, see warnings:

```
[linux/arm64 build-deps 1/1] 
RUN npm install --production=false
npm WARN config production Use `--omit=dev` instead.
RUN npm install --production
npm WARN config production Use `--omit=dev` instead.
```

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/recipes/docker.mdx
+++ b/src/content/docs/en/recipes/docker.mdx
@@ -148,10 +148,10 @@ WORKDIR /app
 COPY package.json package-lock.json ./
 
 FROM base AS prod-deps
-RUN npm install --production
+RUN npm install --omit=dev
 
 FROM base AS build-deps
-RUN npm install --production=false
+RUN npm install
 
 FROM build-deps AS build
 COPY . .


src/content/docs/en/recipes/docker.mdx
```diff
@@ -148,10 +148,10 @@ WORKDIR /app
...
 FROM base AS prod-deps
-RUN npm install --production
+RUN npm install --omit=dev
 
 FROM base AS build-deps
-RUN npm install --production=false
+RUN npm install
 
 FROM build-deps AS build
 COPY . .
```
adaliszk, line 157: Niptick: While it works, I would instead specify the intended folders to copy. This solution heavily relies on `.dockerignore` being correct and not allowing anything unwanted to be copied.

src/content/docs/en/recipes/docker.mdx
```diff
@@ -148,10 +148,10 @@ WORKDIR /app
 COPY package.json package-lock.json ./
 
 FROM base AS prod-deps
-RUN npm install --production
+RUN npm install --omit=dev
```
adaliszk, line 151: Suggestion: Instead of `install`, it would be safer to use `ci`. At this stage, installing is essentially a clean install, but if someone copies this into a different context, they might face some unintended behaviors.

``` suggestion
RUN npm ci--omit=dev
```

src/content/docs/en/recipes/docker.mdx
```diff
@@ -148,10 +148,10 @@ WORKDIR /app
...
 
 FROM base AS build-deps
-RUN npm install --production=false
+RUN npm install
```
adaliszk, line 154: Suggestion: Instead of `install`, it would be safer to use `ci`. At this stage, installing is essentially a clean install, but if someone copies this into a different context, they might face some unintended behaviors.

``` suggestion
RUN npm ci
```

src/content/docs/en/recipes/docker.mdx
adaliszk: Question: What should happen with peer and optional dependencies? Should we also omit those, and if so, then perhaps `--include prod` may be better? However, I am not sure that `include` would automatically understand that everything else should be excluded.

src/content/docs/en/recipes/docker.mdx
```diff
@@ -148,10 +148,10 @@ WORKDIR /app
 COPY package.json package-lock.json ./
```
adaliszk, line 148: Suggestion/Nitpick: While `./` works fine, perhaps `/app/` would convey more information without relying on the previous state of the `WORKDIR`.

src/content/docs/en/recipes/docker.mdx
adaliszk: Suggestion/Nitpick: While `.` works fine, perhaps `/app/` would convey more information without relying on the previous state of the `WORKDIR`.

wassfila: it's different working styles, I guess the doc purpose is only to convey examples, it's hard to cover all use cases, and sometimes people use them in different context, some even develop with live reload under docker. relying on workdir also prevents changing it everywhere in case of rename.

src/content/docs/en/recipes/docker.mdx
wassfila: I agree that ci is better for docker in production for repeatability, but well, for whom that matters this is also easy to modify.

wassfila (approved):
impact of changes is low, changes are strictly equivalent but newer notation so fancy future proof.

src/content/docs/en/recipes/docker.mdx
adaliszk: I would still advise against using `COPY . .` to avoid beginner mistakes. With HRM you would mount the files, and that is fine, but when you `COPY`, you do copy files into your image! You can very easily make a mistake and blow up your image size and even leak sensitive information with things like `.env`.

sarah11918:
Thank you everyone for discussing this so thoughtfully and thoroughly!

(And reminder, at this point it's only the submitted change that we're considering. If there's other text on the page someone feels should be changed, that's out of the scope of this proposed change and a separate PR can be made.)

src/content/docs/en/recipes/docker.mdx
sarah11918: I'm going to go with what the error message suggested, and what worked for the OP, and in the spirit of NWTWWHB, `--omit dev` seems appropriate here.

src/content/docs/en/recipes/docker.mdx
sarah11918: I will keep this consistent with the original instruction, just removing the flag. If there are changes to be made to the recipe otherwise, we can always revisit!

sarah11918 (approved):
Thank you for taking the time to contribute back the fix that worked for you @corneliusroemer ! We appreciate your efforts, and are happy to welcome you to Team Docs!  :partying_face:

src/content/docs/en/recipes/docker.mdx
adaliszk: FYI: https://discord.com/channels/830184174198718474/1245118392793436201

PR #8246, opened 2024-05-10
Update prefetch docs with browser support information

bluwy:
#### Description (required)

Update the prefetch docs based on the core PR linked below, which includes some internal behaviour changes.

I also added a new browser support section to explain the behaviour difference between browsers, and how to ensure prefetching works reliably.


src/content/docs/en/guides/prefetch.mdx
```diff
@@ -51,7 +51,6 @@ Each strategy is fine-tuned to only prefetch when needed and save your users' ba
 
 - If a visitor is using [data saver mode](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/saveData) or has a [slow connection](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType), prefetch will fallback to the `tap` strategy.
 - Quickly hovering or scrolling over links will not prefetch them.
-- Links that use the `viewport` or `load` strategy are prefetched with a lower priority to avoid clogging up the network.
```
bluwy, line 54: This will no longer be the case as all strategies will internally use `<link rel="prefetch">` which has the same (low) priority.

src/content/docs/en/guides/prefetch.mdx
```diff
@@ -106,17 +105,6 @@ As some navigation might not always appear as `<a />` links, you can also prefet
...
-You can additionally configure the priority of prefetching by passing the `with` option:
-
-```js
-// Prefetch with `fetch()`, which has a higher priority.
-prefetch('/about', { with: 'fetch' });
-
-// Prefetch with `<link rel="prefetch">`, which has a lower priority
-// and manually scheduled by the browser. (default)
-prefetch('/about', { with: 'link' });
-```
```
bluwy, line 118: The `with` option is deprecated so I removed this to discourage people from using it.

src/content/docs/en/guides/prefetch.mdx
```diff
@@ -154,6 +142,32 @@ export default defineConfig({
...
+
+1. It supports `<link rel="prefetch">` but errors with `NS_BINDING_ABORTED` if an explicit cache header like [`Cache-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) or [`Expires`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires) doesn't exist.
+2. If it errors but the response have a proper [`ETag`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) header, it will be re-used on navigation.
+3. Otherwise, if it errors with no other cache headers, the prefetch will not work.
+
+### Safari
+
+1. It doesn't support `<link rel="prefetch">` and will fall back to the `fetch()` API.
+2. The `fetch()` API requires cache headers, like [`Cache-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control), [`Expires`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires), and [`ETag`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag), to be set in order to work. Otherwise, the prefetch will not work.
+3. Edge case: `ETag` headers do not work in private windows.
```
sarah11918, line 163:
```suggestion
**Support:** Safari does not support `<link rel="prefetch">` and will fall back to the `fetch()` API, which requires cache headers (e.g. [`Cache-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control), [`Expires`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires), and [`ETag`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)) to be set. Otherwise, the prefetch will not work.

**Edge case:** `ETag` headers do not work in private windows.
```

src/content/docs/en/guides/prefetch.mdx
```diff
@@ -154,6 +142,32 @@ export default defineConfig({
...
+
+### Recommendations
+
+To best support all browsers, make sure your pages have the proper cache headers. For SSG or pre-rendered pages, the `ETag` header is often automatically set by the deployment platform, so they should work out of the box.
```
sarah11918, line 167:
```suggestion
To best support all browsers, make sure your pages have the proper cache headers. 

For static or prerendered pages, the `ETag` header is often automatically set by the deployment platform, so they should work out of the box.
```

src/content/docs/en/guides/prefetch.mdx
```diff
@@ -154,6 +142,32 @@ export default defineConfig({
...
+
+For dynamic and server-side rendered pages, you should set the appropriate cache headers based on the page content. Visit the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching) for more information on HTTP caching.
```
sarah11918, line 169:
```suggestion
For dynamic and server-side rendered pages, you should set the appropriate cache headers based on the page content. Visit the [MDN documentation on HTTP caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching) for more information.
```

sarah11918 (commented):
Looks great, @bluwy !

I tried something different with the browser support section (haven't seen it in preview yet, so it might be awful). Let's see what we think!

bluwy:
Thanks! I like your suggested formatting too. I committed them and view locally and it looks like this:

<details>
<summary>Screenshot (Vercel link should also be possible)</summary>

image

</details>

Seeing this, maybe we could remove the `Support:` prefix on each paragraphs? Since the heading "Browser support" already implies so.

sarah11918:
Yup, agreed! I had it in my head that maybe the different original list items were going to be different, repeatable "things"... but nope, I think without will be just fine!

bluwy:
Thanks! I've updated and removed them now.

bluwy:
@sarah11918 the core PR has been released. Waiting on your approval just in case and I think we can then merge this

sarah11918:
@bluwy Thanks, taking one more final pass now, then I'll merge!

sarah11918 (approved):
Thank you, @bluwy , LGTM!

PR #8255, opened 2024-05-11
A typo in the code sample 4.mdx

stixijin:
A typo in the get code example - because of this I get the error "No API Route handler exists for the method "GET" for the route "/rss.xml". Found handlers: "Get" When building xml is not created. When editing from get to GET everything works.

- Closes #
- Suggested label:


at-the-vr (commented):
Can confirm this is correct after checking [en/tutorial/5-astro-api/4](https://docs.astro.build/en/tutorial/5-astro-api/4/#create-an-xml-feed-document) LGTM 👍

yanthomasdev (approved):
LGTM! Thanks

PR #8281, opened 2024-05-15
feat: container API docs

ematipico:
#### Description (required)

This PR adds the first batch of documentation of the container APIs.

I added a new page in the reference called "Container API" and a new paragraph in the "Testing" section.

#### For Astro version: `4.9`. See astro PR [#11051](https://github.com/withastro/astro/pull/11051).


src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+
+```js name="Card.test.js"
+import Card from "../src/components/Card.astro";
+
+const result = await container.renderToString(Card, { 
+  slots: { "header": "Header content", "footer": "Footer" }
+});
+```
+
+You can also render components in cascade:
```
sarah11918, line 145: I'm not sure we'd say "in cascade" but I get what you mean. Bookmarking this!

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+});
+```
+
+#### `request` option
+
+**Type**: `Request`
+
+The request is used to understand which path/URL the component is about to render.
+
+Use this option in case your component needs to read information like `Astro.url` or `Astro.request`.
```
sarah11918, line 175:
```suggestion
An option to pass a `Request` with information about the path/URL the component will render.

Use this option when your component needs to read information like `Astro.url` or `Astro.request`.
```

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+    }
+  })
+});
+```
+
+#### `params` option
+
+**Type**: `Record<string, string | undefined>`;
+
+Use this option in case your component needs to read `Astro.params`. As opposed to `getStaticPaths`, you must provide only **one** item of the params returned by that function.
```
sarah11918, line 195:
```suggestion
An object to pass information about the path parameter to an Astro component responsible for [generating dynamic routes](/en/guides/routing/#dynamic-routes).

Use this option when your component needs a value for `Astro.params` in order to generate a single route dynamically.
```
Check this for accuracy, but I think this seems like what you're saying. 

And, I thought mentioning `getStaticPaths()` *might* be more confusing here than helpful? I think we can say this like above to emphasize that, like in other examples shown, you're providing *ONE* value of something so that you can test a particular case by rendering a single version of the component.

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+    slug: "getting-started"
+  }
+});
+```
+
+#### `locals` options 
+
+**Type**: `App.Locals`
+
+Use this option to stub the `Astro.locals` object:
```
sarah11918, line 219:
```suggestion
An option to pass information from [`Astro.locals`](/en/reference/api-reference/#astrolocals) for rendering your component.

Use this option to when your component needs information stored during the lifecycle of a request in order to render, such as logged in status.
```
These words might be kind of wrong, but just trying to fit this into the pattern!

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+  
+  // assert result contains "You're out"
+})
+```
+
+#### `routeType` option
+
+**Type**: `"page" | "endpoint"`
+
+Useful in case you're rendering an endpoint. In this case, you want to use `renderToResponse`:
```
sarah11918, line 258:
```suggestion
An option available when using `renderToResponse` to specify that you are rendering an [endpoint](/en/guides/endpoints/):
```

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+```js file="endpoint.test.js"
+import * as Endpoint from "../src/pages/api/endpoint.js";
+
+const response = await container.renderToResponse(Endpoint, { 
+  routeType: "endpoint"
+});
+const json = await response.json();
+```
+
+If your endpoint needs to be tested on methods such as `POST`, `PATCH`, etc., you'll have to use the `request` option to signal Astro to call the correct function:
```
sarah11918, line 273:
```suggestion
To test your endpoint on methods such as `POST`, `PATCH`, etc., use the `request` option to call the correct function:
```

src/content/docs/en/guides/testing.mdx
```diff
@@ -43,6 +44,30 @@ export default getViteConfig(
 
 See the [Astro + Vitest starter template](https://github.com/withastro/astro/tree/latest/examples/with-vitest) on GitHub.
 
+## Vitest and Container API
+
+<Since v="4.9.0" />
+
+You can natively test Astro components using the new [container API](/en/reference/container-reference/). First, setup [`vitest` as explained above](#vitest), then create a `.test.js` file to test your component:
```
sarah11918, line 51:
```suggestion
You can natively test Astro components using the [container API](/en/reference/container-reference/). First, setup [`vitest` as explained above](#vitest), then create a `.test.js` file to test your component:
```
If we don't say "new", then we don't have to remember to update whenever we decide it's not "new" anymore! :smile:

sarah11918 (commented):
Looks great @ematipico !

I mostly added some links throughout, and tried to add a bit of "glue" text on the reference page to explain things a little. But of course, I'm going to need you to make my "explanations" make perfect sense!  :sweat_smile: 

See what you think about some of my suggestions below! (I think you'll find some are tiny fixes, some are repeated "let's use this format" so it's not really as much as it looks)

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+	renderers?: AstroRenderer[];
+	astroConfig?: AstroContainerUserConfig;
+};
+```
+
+#### `streaming` option
+
+Enables rendering components using the streaming.
+
+**Type:** `boolean`
```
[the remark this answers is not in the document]
ematipico: Yeah that's correct

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+
+#### `renderers` option
+
+A list of client renderers. Add a renderer if your Astro component renders some client components such as React, Preact, etc.
+
+**Type:**: `AstroRenderer[]`;
+
+For example, if you're rendering a Astro component that renders a React component, you'll have to provide the following renderer:
```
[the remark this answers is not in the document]
ematipico: Since we typed the options with `AstroRenderer`, users will be required to provide `name`, a `clientEntrypoint` renderer and `serverEntrypoint` renderer. 

Unfortunately, this is a very low-level API, and no end-user is aware of that other than developers who create Astro renderers. Renders that we support don't expose or document these things. 

Ideally, the example should provide *enough* knowledge as a starting point, but we plan to expose these renderers to the users, so they won't need to manually write them like in the example.

Overall, I think the suggestion looks

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+    ]
+})
+const result = await container.renderToString(ReactWrapper);
+```
+
+#### `astroConfig` option
+
+A subset of the Astro user configuration.
+
+**Type**: `AstroContainerUserConfig`
```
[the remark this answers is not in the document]
ematipico: We decided to remove this option from the container for the time being. We will wait for users to request specific options if it makes sense. We've had hard times to find a use case that made sense.

src/content/docs/en/reference/container-reference.mdx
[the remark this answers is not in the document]
sarah11918: (NOTE: Part of my thinking by exposing the vocab in a sentence is that even if they can't find these things themselves, it's easy enough for them to come into support and ask, "I can't find the `clientEntrypoint` and `serverEntrypoint` for Svelte..." and then that's an easy question for them to ask and get a direct answer to by someone who knows.)

sarah11918 (commented):
So close with this one! I'm really loving how it turned out! Just one little thing I think was missed in applying suggestions (unless I'm wrong, so speak up!) and one example I'd like you to double and triple check, just to be sure!

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+#### `params` option
+
+**Type**: `Record<string, string | undefined>`;
+
+An object to pass information about the path parameter to an Astro component responsible for [generating dynamic routes](/en/guides/routing/#dynamic-routes).
+
+Use this option when your component needs a value for `Astro.params` in order to generate a single route dynamically.
+
+```astro name="Card.astro"
```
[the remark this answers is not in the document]
ematipico: Yeah, I think it makes more sense to use a file with a dynamic route

PR #8300, opened 2024-05-16
fix minor detail in tutorial 3

garysassano:
#### Description (required)

The code wasn't referencing the previous state of `index.astro` from tutorial 2.


sarah11918 (approved):
Thank you @garysassano !

We don't always show the full existing content in these samples or else the tutorial would get very long and complicated! This is why we have code check-ins periodically where you can compare to a full version.

But, in this case, since there is very little existing code, I think it's fine to add here! We can be complete and still not overwhelming with too much code, so let's do this!

Thank you for your submission, and welcome to Team Docs!  :partying_face:

PR #8303, opened 2024-05-17
change wording in tutorial

garysassano:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/tutorial/4-layouts/1.mdx
+++ b/src/content/docs/en/tutorial/4-layouts/1.mdx
@@ -147,7 +147,7 @@ Don't forget to:
 
 - Let the layout be responsible for the HTML rendering of any common elements.
 
-- Delete anything from each page that that page is no longer responsible for rendering, because it is being handled by the layout, including: 
+- Delete anything from each page that is no longer responsible for rendering, because it is being handled by the layout, including: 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/tutorial/4-layouts/1.mdx
```diff
@@ -147,7 +147,7 @@ Don't forget to:
 
 - Let the layout be responsible for the HTML rendering of any common elements.
 
-- Delete anything from each page that that page is no longer responsible for rendering, because it is being handled by the layout, including: 
+- Delete anything from each page that is no longer responsible for rendering, because it is being handled by the layout, including:
```
sarah11918, line 150:
```suggestion
- Delete anything from each individual page that is now being handled by the layout, including: 
```
In fact, the original is grammatically correct, but, I agree that it's an awkward sentence to read (feels more natural when spoken) so let's change it!

I think maybe we can make this even better. What do you think about something like this?? Does this feel *too* short, or does this still get the idea across?

sarah11918 (commented):
Thanks for contributing @garysassano ! See my comment and suggestion below!

sarah11918 (approved):
Looks great, thank you for helping to improve this line!

PR #8304, opened 2024-05-17
fix css in tutorial 3

garysassano:
The stylesheet in part 4 is missing a piece that was defined in part 3.


sarah11918:
Hi! As commented in the other PR, we don't *always* show the full files. This says "add the following styles" so it can be assumed that there are some existing styles already here. This would make this code sample very long and not offer much benefit, so that's why we explicitly chose to leave out the existing part of the file here.

PR #8318, opened 2024-05-17
replace astro light logo with dark logo

garysassano:
When you reach this part of the tutorial:
image

You will end up with this blog post page being rendered:

before

You may think something isn't working, but you are actually just displaying a white logo on a white page, which isn't ideal.

So we can just switch to the dark logo instead:

after


yanthomasdev:
Nice catch @garysassano, can I just ask you to remove the translations from this PR and leave only the English changes? We have a system in-place to let translators know of this change and do it on their own 🙌

at-the-vr (commented):
Hey Gary 👋 I find a similar issue with the Dark theme now: [Stackblitz Instance](https://stackblitz.com/edit/github-wnthds?file=src%2Fpages%2Fposts%2Fpost-1.md) shows a dark logo on a dark page. Its still more visible as compared to former page. letting it know before someone asks to change it back to `logo-light.png` 😄 
image

at-the-vr:
yeah i was worried that this is not available from Astro Press and having an unofficial logo would be uncalled for

src/content/docs/en/tutorial/2-pages/2.mdx
```diff
@@ -49,8 +49,8 @@ Now that you have built pages using `.astro` files, it's time to make some blog
     description: 'This is the first post of my new Astro blog.'
     author: 'Astro Learner'
     image:
-        url: 'https://docs.astro.build/assets/full-logo-light.png' 
-        alt: 'The full Astro logo.'
+        url: 'https://docs.astro.build/assets/rose.webp' 
+        alt: 'Thumbnail of astro rose.'
```
sarah11918, line 53:
```suggestion
        alt: 'The Astro logo on a dark background with a pink glow.'
```

sarah11918 (approved):
Thanks for updating the graphic! LGTM! Let's make this work for all view modes! :rocket:

PR #8323, opened 2024-05-17
update:  vite resolve guide for path aliases

samuelhulla:
#### Description (required)
Adds information on how to configure astro with vite in order to prevent failing builds when using path aliases.

This was caused by `vite` requiring a speicifc `resolve.alias` configuration, which was not mentioned anywhere in the docs. Furthermore it's super confusing for everyday users, since the build does work locally and only fails in deployment environments (i.e. vecel or render)

#### Implementation details

I've noticed that `__dirname` is not defined in `astro.config.mjs` so I've resorted to using `path.dirname('')` since it's bound to work at all config types.

I'd also like to note that defining own `vite.config.ts` does not work _(burnt me when trying to resolve the issue)_ as you need to include the vite value in the astro config itself. But I've decided not to mention this in the docs.

#### Final thoughts

In an ideal scenario, when using `npx/bunx astro add vercel/or-any-other-deployment-service` there could be a codemod which modifies the astro config for values set in tsconfig.json, as we can infer from the values in `tsconfig.json`. But  then it kind of obfuscates that you need to update these values in both places, so I'm kind of torn on this _(ideal scenario is imho to have both docs warning people like this and codemode do the rest for them on install / update)_

___

pr meta:

- Closes  https://github.com/withastro/astro/issues/11090
- Suggested label: `docs pr`

yes

#### If you are a member of the Astro Discord, please add your username in the description so we can welcome you there!
`@hulla`

### DIFF
--- a/src/content/docs/en/guides/aliases.mdx
+++ b/src/content/docs/en/guides/aliases.mdx
@@ -56,7 +56,7 @@ This is not an `astro` related issue. However, since we rely on `vite` to produc
 
 To resolve this issue, add the following fields to your Astro configuration:
 
-```mjs title="astro.config.mjs" ins={1} ins={4-13}
+```ts title="astro.config.mjs" ins={1} ins={4-13}
 import { path } from 'path'
 
 export default defineConfig({
@@ -71,5 +71,3 @@ export default defineConfig({
     }
   }
 })
-
-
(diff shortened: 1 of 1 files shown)


at-the-vr (commented):
Hey @samuelhulla your PR's accurate 💯 but will this require [eslint-import-resolver-alias](https://www.npmjs.com/package/eslint-import-resolver-alias) in a project using `eslint` for future type checks? Can you verify this part?

samuelhulla:
@at-the-vr I'm not sure if it necessitates it, I got the resolution working out of the box without any issues with just the lines in the PR change. 

That being said, I'll check it out, but I have a busy weekend so quickest I can get to it is late Sunday/Monday

at-the-vr:
Its totally fine @samuelhulla. In any case you find a cloud service that can benefit from your PR, you can proceed to reopen this one 😄

PR #8335, opened 2024-05-18
Move Houston to avoid colliding with text

TheOtterlord:
- Move Houston to avoid colliding with heading text
image
- Add spacing above card to prevent collision with asides.
image


sarah11918 (approved):
Yup, confirming I see Shrunkston on various views of "Analyze the Pattern". Great job, thanks for jumping on this @TheOtterlord !

delucis:
Noting that one other way to avoid the overlap that Erika observed would be to set `margin-inline-end` for the title of the box, i.e. something like:

```css
.highlight-box .heading-wrapper:first-child {
    margin-inline-end: 6rem;
}
```

That said, moving up and adding some spacing is also still a good idea to avoid stuff like this:

Houston overlapping with body text and sitting too snugly close to text below it

Might be worth doing that as well to be safe and to avoid Houston sitting too snuggly immediately above the header like in your first screenshot @TheOtterlord?

TheOtterlord:
Yep! Much better.
image

TheOtterlord:
Fixed! Scoped it to the questions & shrunk houston again

delucis (approved):
LGTM! Thanks for figuring all that out @TheOtterlord 💜 

One day I guess we’ll rename those classes to no longer be `.question-mark` 😁

PR #8364, opened 2024-05-21
content-collections: Add example to render list of contents

tordans:
#### Description (required)

I was looking all over for an example on how to use the `render()` function to render a list of blog posts like on any old WordPress index page.

It turns out, one can just pass the `Content` component in a custom object and render this as `<Content/>` just like always.

I think it would be great to mention this somewhere. I cannot be the only one searching for this? :-D.

Please feel free to change this PR as you see fit.

**=> [This is the section in the preview](https://docs-git-fork-tordans-patch-4-astrodotbuild.vercel.app/en/guides/content-collections/#render-a-list-of-contents)**

./.

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -476,7 +476,7 @@ const { Content, headings } = await entry.render();
 
 Sometimes, you don't just want to [render one Content to HTML](#rendering-content-to-html) but a list of contents, like on the index page of a blog.
 
-```astro {5}
+```astro
 ---
 // src/pages/render-list-example.astro
 import { getCollection } from 'astro:content';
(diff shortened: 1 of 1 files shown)


sarah11918:
We've discussed this PR and don't think this fits our needs in that section, which is only meant to show the functions used, and not the various ways in which they can be used. So I'll be closing this PR, but thank you as always for contributing! 🙌

tordans:
@sarah11918 thank you for your feedback. Can you help me understand where in the docs I can find information on how to render a list of full posts like I would do with any WordPress side? In case I am right and there is none, do you consider it a trivial thing to find out by oneself … or do you consider it a non-relevant use case for Astro to render full lists of content?

tordans:
Reverse ping https://github.com/withastro/docs/issues/10317 which will show how to create a list of content for RSS.

PR #8373, opened 2024-05-22
Update api-reference.mdx

hfournier:
The description for Astro.redirect() makes no mention of the optional second parameter, so I added a "note" about it.

#### Description (required)

The description for Astro.redirect() makes no mention of the optional second parameter, so I propose adding a "note" that it accepts an optional HTTP Status Code (linked it to MDN), and that the default is 302.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -385,6 +385,11 @@ if (!isLoggedIn(cookie)) {
 ---
 ```
 
+:::note
+`Astro.redirect()` accepts an [HTTP response status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) as an optional second parameter. The default is 302.
+:::
+
+
 ### `Astro.canonicalURL`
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks for this @hfournier ! We should absolutely be showing this! I took a quick editing pass to work the content into the actual description, since we don't like to add essential content by tacking on notes. (Docs can start to look pretty wild if we overuse that pattern!)

Take a look at the version I have and see whether you'd suggest any changes!

hfournier:
> Thanks for this @hfournier ! We should absolutely be showing this! I took a quick editing pass to work the content into the actual description, since we don't like to add essential content by tacking on notes. (Docs can start to look pretty wild if we overuse that pattern!)
> 
> Take a look at the version I have and see whether you'd suggest any changes!

That looks good to me... and thanks for the heads up on not overusing notes on future updates.

sarah11918 (approved):
Great, thank you so much, let's do this!

PR #8393, opened 2024-05-24
feat: add empty slot fallback content info (#8302)

### LINKED ISSUE #8302: Improve documentation on fallback content for slots
### 📚 Subject area/topic

Fallback Content for Slots

### 📋 Suggested page

https://docs.astro.build/en/basics/astro-components/#fallback-content-for-slots

There is a case where the default fallback content for slots is not displayed. This has been discussed on [this](https://github.com/withastro/astro/issues/11047#issuecomment-2112538371) issue in the withastro/astro repo. 

So given component A and B below, the default fallback content is not shown when `<slot name="s" slot="s" />` is present and empty. That is, when the slot is empty in component A, the fallback content is not shown.

### Component B
```astro
<slot name="s">
  <div>Default content</div>
</slot>
```

### Component A
```astro
<B>
  <slot name="s" slot="s" />
</B>
```

As Mathew wrote in the issue mentioned earlier, "Default content is rendered when a slot is not provided. Not when it is empty." This should be documented in the section about [Fallback Content for Slots](https://docs.astro.build/en/basics/astro-components/#fallback-content-for-slots).

[Link](https://stackblitz.com/edit/github-mvb5lj?file=src%2Fpages%2Findex.astro,src%2Fincludes%2FB.astro,src%2Fincludes%2FA.astro) to the fallback behaviour on Stackblitz

jdtjenkins:
#### Description (required)

This PR adds an example of how passing empty slots causes that slot's default content to not render.

- Closes #8302 

**For a new/changed feature in an upcoming Astro release?**

No

Yes :3

Discord: Jumper#8174


at-the-vr (approved):
PR looking great already Jumper 🔥 although 
⚠️ nit warning ⚠️ 
I believe the `title` for components could match docs theme that is name for components themselves. Besides that, this is LGTM 👍

src/content/docs/en/basics/astro-components.mdx
```diff
@@ -318,6 +318,40 @@ const { title } = Astro.props;
...
+---
+import ComponentB from ""./ComponentB.astro";
+---
+
+<ComponentB />
+```
+
+Here we are not passing a slot into `ComponentB` from `index.astro`. But `ComponentB` is still passing an empty slot into `ComponentA`. Therefore `ComponentA` will not render it's default content.
+
+This is because a component checks that a slot exists - not if it's empty.
```
sarah11918, line 353:
```suggestion
Fallback content will only be displayed when no matching elements with the `slot="name"` attribute are found to pass into a named slot. 

If an empty slot (a slot that exists but contains no content) is passed into a component, then the fallback content will not be displayed.
```
I'm wondering whether we can get away with this, and not showing an example here. Would you say this explanation is correct? And if so, do you think it's *sufficient* without showing three different code blocks?

I would argue that if it's correct, it's NWTWWHB! And, if we get questions/complaints, we can always add code snippets in later. What do you think?

sarah11918 (commented):
LOOK WHO MADE A DOCS PR!!!  😄 

Thanks for jumping on this issue, @jdtjenkins ! So, I kind of started by trying to go extreme minimum to figure out what is *enough* to be helpful here.

See what you think, and we can negotiate how much more is needed!

src/content/docs/en/basics/astro-components.mdx
jdtjenkins: @sarah11918 Thank you! 🙌 

I think that explanation is pretty perfect and concise! The only thing I'm thinking is: it's a fairly uncommon situation to be in. I think the code snippet sort of highlights that really you should never really have this scenario in your code - but ofc it _will_ happen to someone eventually.

Also I feel people, myself included, would assume that an empty slot _wouldn't_ pass an empty slot. So I feel like this section is less about telling people that fallback content will only displayed when blah blah blah. It's more about informing people that it _is_ in fact possible to pass an empty slot to a child. I would've assumed the slot would _only_ get passed if it contains content, which clearly is not the case.

I'm more than happy to have just this explanation for now and get feedback before adding more to it. I feel it's a very uncommon use case, and at least those unlucky enough to run into this can see this bit here immediately with the rest of the slots documentation.

sarah11918: OK, if it's less about the fallback then we could shift the second sentence to something like...??

> Astro will pass an **empty slot** when a slot element exists but has no content to pass, and fallback content will not be displayed. Therefore, fallback content cannot be used as a default or check for a case where no content is passed. Fallback content is only displayed when no slot element can be found.

jdtjenkins: Yeah I think that makes more sense tbh from that angle! But it is still quite complex to explain! What about:

> Astro will pass an **empty** slot when a slot element exists but has no content to pass. Fallback content cannot be used as a default when an empty slot is passed. Fallback content is _only_ displayed when no slot element can be found.

sarah11918: Ooh, even simpler! I like it! Ship it!

src/content/docs/en/basics/astro-components.mdx
```diff
@@ -318,6 +318,8 @@ const { title } = Astro.props;
 </div>
 ```
 
+Astro will pass an empty slot when a slot element exists but has no content to pass. Fallback content cannot be used as a default when an empty slot is passed. Fallback content is only displayed when no slot element can be found.
```
sarah11918, line 321:
```suggestion
Fallback content will only be displayed when no matching elements with the `slot="name"` attribute are found to pass into a named slot. 

Astro will pass an empty slot when a slot element exists but has no content to pass. Fallback content cannot be used as a default when an empty slot is passed. Fallback content is only displayed when no slot element can be found.
```
Do we think leaving in that first sentence is too much?

sarah11918: I don't think it's "passed IN AS a named slot?" (It is not the named slot itself, that's the destination?) 

I would still say passed (in) TO a named slot? Are you 100% on that?

But yes, we are very close!  :joy:

jdtjenkins: Y'know what, you're right! You pass things _into_ a slot!

So then the final™️ wording:

---

Fallback content will only be displayed when there are no matching elements with the slot="name" attribute being passed in to a named slot.

Astro will pass an empty slot when a slot element exists but has no content to pass. Fallback content cannot be used as a default when an empty slot is passed. Fallback content is only displayed when no slot element can be found.

sarah11918: final-1(1)-a-2 final!

sarah11918 (approved):
Finally getting you into Team Docs @jdtjenkins ! :partying_face:

PR #8394, opened 2024-05-25
feat: upgrade pnpm

florian-lefebvre:
Upgrade pnpm, required for https://github.com/withastro/astro/pull/11146


.npmrc
```diff
@@ -1 +1,3 @@
 shell-emulator=true
+# Required not to fail on Vercel
+package-manager-strict=false
```
delucis, line 3: Don’t love that we need to add this, but I guess given it’s needed for smoke tests in the main repo, it’s ok 👍

delucis (approved):

.npmrc
florian-lefebvre: There was an update to this in 9.2.0 I think, let's try to comment it

PR #8396, opened 2024-05-26
Update `talks.mdx`

at-the-vr:
#### Description (required)

- More talks 🔥 

- Suggested label: 1i8n


sarah11918:
Hey @at-the-vr , thanks for always keeping up with this!

When I created this section, it wasn't meant to be comprehensive, like some of the other sections, and it was mostly intended for *our* talks and appearances from core team members explaining what Astro is. And, Elian gives similar talks, so we don't need to keep adding them! (Sorry Elian 😄 )

So none of these are really what I had envisioned for this section, because we do have a YouTube playlist, the community calls/monthly blog posts for this etc. This should only be a small sample of what WE produce, so that people are finding out about Astro from us. So, let's not add to this section for now, but make sure these are posted in the #showcase channel so that they will end up in the monthly blog posts! Sorry that I hadn't made my intention for this section clear!

at-the-vr:
makes sense, thanks for the clarification 🙌. Thinking out loud, none of the talks were held in May 2024 and won't clarify logic for Astro community calls or the Monthly blogpost, I guess i will abstain from posting them in showcase channel if that is fair. As long as they are saved in the playlist I won't be bothered

PR #8397, opened 2024-05-26
Fix content collection Date string example

ascheucher-shopify-partner:
the given method of "slicing" out the date part of a Date.toString() did give me some "Mon May 25" and not the expected 2024-05-25 I wanted.

Using the Date.toISOString() solved this for me.

#### Description (required)

See introductional description. It's simple as that.

- Closes #
- Suggested label:


Fryuni (approved):
Hey @ascheucher-shopify-partner!! Thanks for the PR and welcome to the Docs Team 🎉 🎉 🎉 

The result of `Date.toString()` depends on your system's configuration; it is indeed not the appropriate method for that use case.

PR #8400, opened 2024-05-27
feat(container-api): add props reference

adaliszk:
#### Description (required)

Adding the reference to the Container API props.


adaliszk:
ehm, github...

PR #8404, opened 2024-05-27
feat(container-api): add props reference

adaliszk:
#### Description (required)

Adding the reference to the Container API props.


src/content/docs/en/reference/container-reference.mdx
```diff
@@ -96,11 +96,11 @@ Both [`renderToResponse`](#rendertoresponse) and [`renderToString`](#rendertostr
 ```ts
 export type ContainerRenderOptions = {
 	slots?: Record<string, any>;
+  props?: Record<string, unknown>;
```
ematipico, line 99: It seems it's not formatted properly

ematipico (approved):

src/content/docs/en/reference/container-reference.mdx
adaliszk: Nice catch, fixed!

dreyfus92 (approved):
Thank you @adaliszk, LGTM ✅

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -165,6 +165,66 @@ const result = await container.renderToString(Card, {
...
+import { CardProps } from "./types.ts";
+
+interface Props extends CardProps {};
+
+const { name } = Astro.props;
+---
+<div>
+  {name}
+</div>
+```
```
sarah11918, line 226:
``````suggestion
#### Type-hints
 
When using the Container API outside of `.astro` files (e.g. `test.js`), define and import any interfaces to make them available for type-hints:

```js name="Card.test.js"
import Card from "../src/components/Card.astro";
import { CardProps } from "../src/components/types.ts";

const props: CardProps = { name: "Hello, world!" };
const result = await container.renderToString(Card, { 
  props,
});
```

```astro name="types.ts"
export interface CardProps {
  name: string;
};
```

```astro name="Card.astro"
---
import { CardProps } from "./types.ts";

interface Props extends CardProps {};

const { name } = Astro.props;
---
<div>
  {name}
</div>
```
``````
I'll need your help to make sure the wording is accurate, but we try not to use "notes" etc. for things that are just regular parts of the documentation. Notes, cautions, tips are really for "out of the ordinary" content.

So, I'm assuming it makes sense that instead of a note, we can have a section heading. The advantage here is that these get indexed, are easy to link to for support etc. vs being in a note.

With that being said, how do you feel about the suggestion above? What should be changed to make it completely accurate/helpful?

**Review by sarah11918 — COMMENTED** [review]
Thank you for picking this up, @adaliszk ! 🙌  See my comment / question below!

File: src/content/docs/en/reference/container-reference.mdx (hunk 2 above)
Comment (adaliszk) [reply src/content/docs/en/reference/container-reference.mdx:226]: Hmm, the only thing I see now is not doing
```typescript
import { type CardProps } from "./types.ts";
```

adaliszk: Incorporated the suggestion and added a little more at the end to encourage astro component maintainers to expose their properties through interfaces.

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -165,6 +165,75 @@ const result = await container.renderToString(Card, {
...
+
+With the dedicated interface for the component properties, astro components 
+can be integrated with any UI frameworks and gain full type-hints support.
+
+Therefore, UI components authors for Astro components can provide a way to
+integrate their components with any UI frameworks with fairly minimal effort!
+
```
delucis, line 236: Sorry if this suggestion seems a bit drastic — DELETE ALL THE THINGS 😅 

I think while it can seem helpful, this is mostly **re**documenting common practices from elsewhere, i.e. how Astro props typing works is not container API specific, and nor is using an imported type for the `props` object. It might be different if we’d made the API take generic parameters (like `renderToString<Props>()` or something), so that we needed to document the specific pattern, but given this is all regular usage, I’d suggest keeping the docs slim and focused is actually more helpful:

```suggestion
```

delucis (changes requested):
Thanks for adding this @adaliszk! I think we can actually slim this down to help keep the reference here focused.

src/content/docs/en/reference/container-reference.mdx
adaliszk: Sure thing. If you want to omit available information on integrating with typescript, that is fine; who needs types or a healthy ecosystem anyway?
The generic type solution is a workaround solution, just like casting the props; I would not spend effort on it when your own JSX types should infer the props correctly - without a dedicated LSP - and thus allow any usage of your Astro components to be correct.

However, if removing the context is a problem (181-195), then all other occurrences on this page should change to keep it consistent. I would suggest not doing that in PR but raising a new one if you desire to change the pattern of the documentation.
Admittedly I have overstepped on the same.

adaliszk: I removed the type-hint portion for now and am waiting for your response on changing the reference pattern.

delucis (approved):
Thanks again @adaliszk!

PR #8419, opened 2024-06-02
Remove step 4 and 6 from the tutorial

### LINKED ISSUE #8418: Tutorial says that the rss.xml can't be seen in development but it actually can be
### 📚 Subject area/topic

Tutorials -> Build a blog

https://docs.astro.build/en/tutorial/5-astro-api/4/#create-an-xml-feed-document

On step 4 (This rss.xml document is only created when your site is built, so you won’t be able to see this page in your browser during development...) of Create an .xml feed document

It says that rss.xml is only created when the site is built, but that isn't true since I can see it in the development server.

I guess that this could be fixed by just removing step 4 or replacing it with a run a development server?

Bildschirmfoto_20240601_211612

Ash882:
#### Description (required)

I removed step 4 and 6 of Create an .xml feed document in the tutorial because it says that rss.xml is only generated during build, even though it is also generated during development.

- Closes #8418
- Suggested label: tutorial


sarah11918 (approved):
Thank you, @Ash882 ! I don't know when we updated to show this page in dev, but apparently we did! 😄 

Thanks for the help keeping this updated, and welcome to Team Docs! 🥳

yanthomasdev (approved):
LGTM! Thanks @Ash882!

PR #8422, opened 2024-06-03
Fix service account for firebase prod.

leoxs22:
#### Description (required)

The current recommended configuration produces an error when deploying to Firebase. Randomly, the server will return Service unavailable, saying that project_id is undefined when loading the site. After reloading the site a few times it will load, but it will break again after some inactivity _(I guess is when firebase shut downs the function)_ or after a new deploy.

It is pretty random, so it was extremely hard to find what was causing the error. I'm still not sure, but I'm sure this fixes it. My best guess is:

meta.env vars are solved in build time, when running this on a CI, it won't find anything so it will end up as a proccess.env, waiting for the var to be solved in runtime. Firebase already has a default account loaded in every enviroment, so just calling `initializeApp()` without anything works perfectly, but the env vars are not available. I'm not sure why after 4 reloads it works, it might be because Firebase loads the env vars but with a delay, or maybe because the app is created with a delay so it ends up in `getApps()`

In any case, calling the initialize without parameters is the recommended way to initialize it in production, so I would say this is the best approach

(btw, firebase recently released app hosting which seems to be for SSR, I haven't tried it yet, but it might change a lot of things)


sarah11918:
Thank you for looking into this, @leoxs22 ! I'm going to see who I can get to look further into this on our end in case someone has any extra insight before we go ahead and publish this. Hang tight, and thanks for helping!  🙌

sarah11918 (approved):
OK @leoxs22 , since no one seems to have anything better, and you're confident that this at least doesn't produce an error, I'm going to go with you here and let's get this published and helping people!  🙌

PR #8428, opened 2024-06-04
feat(container): document new way to load renderers

ematipico:
#### Description (required)

This PR updates the container API, and documents the new way to load renderers.

### DIFF
--- a/src/content/docs/en/reference/container-reference.mdx
+++ b/src/content/docs/en/reference/container-reference.mdx
@@ -42,9 +42,13 @@ Enables rendering components using [HTML streaming](/en/guides/server-side-rende
 
 **Type:** `SSRLoadedRenderer[]`
 
-A list of client renderers required by the component. Use this if your `.astro` component renders any [UI framework components](/en/guides/framework-components/) using an official Astro integration (e.g. React, Vue, etc.). For each framework rendered, you must use function `getContainerRenderer` that can be imported from the renderer library.
+A list of loaded client renderers required by the component. Use this if your `.astro` component renders any [UI framework components](/en/guides/framework-components/) or MDX using an official Astro integration (e.g. React, Vue, etc.).
 
-When using `vite` (`vitest`, Astro integrations, etc.), the renderers should be loaded by using the function `loadRenderers`, from the virtual module `astro:container`. Outside `vite`, you'll have to load the renderers manually.
+For each official Astro integration, import and use the `getContainerRenderer()` helper function to expose its client and server rendering scripts. These are available for `@astrojs/react`, `@astrojs/preact`, `@astrojs/solid-js`, `@astrojs/svelte`, `@astrojs/vue`, `@astrojs/lit`, and `@astrojs/mdx`.
+
+For renderer packages outside the `@astrojs` npm org, look in their documentation for `getContainerRenderer()` or a similar function provided.
+
+When using `vite` (`vitest`, Astro integrations, etc.), the renderers are loaded with the function `loadRenderers()` from the virtual module `astro:container`. Outside `vite`, you'll have to load the renderers manually.
 
 The following example provides the necessary object to render an Astro component that renders a React component and a Svelte component: 
 
@@ -59,10 +63,6 @@ const container = await experimental_AstroContainer.create({
 })
 const result = await container.renderToString(ReactWrapper);
 ```
-
-:::note
-The function `getContainerRenderer` is only available for renderers inside the `@astrojs/` npm org. Renderers outside this org will provide a similar/same function.
-:::
 
 ## `renderToString()`
 
--- a/src/content/docs/en/reference/container-reference.mdx
+++ b/src/content/docs/en/reference/container-reference.mdx
@@ -40,7 +40,7 @@ Enables rendering components using [HTML streaming](/en/guides/server-side-rende
 
 #### `renderers` option
 
-**Type:** `SSRLoadedRendere[]`
+**Type:** `SSRLoadedRenderer[]`
 
 A list of client renderers required by the component. Use this if your `.astro` component renders any [UI framework components](/en/guides/framework-components/) using an official Astro integration (e.g. React, Vue, etc.). For each framework rendered, you must provide an object stating the integration package `name`, as well as both its client and server rendering script.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/container-reference.mdx
```diff
@@ -40,21 +40,21 @@ Enables rendering components using [HTML streaming](/en/guides/server-side-rende
 
 #### `renderers` option
 
-**Type:** `AstroRenderer[]`
+**Type:** `SSRLoadedRendere[]`
 
-A list of client renderers required by the component. Use this if your `.astro` component renders any [UI framework components](/en/guides/framework-components/) using an official Astro integration (e.g. React, Vue, etc.). For each framework rendered, you must provide an object stating the integration package `name`, as well as both its client and server rendering script
+A list of client renderers required by the component. Use this if your `.astro` component renders any [UI framework components](/en/guides/framework-components/) using an official Astro integration (e.g. React, Vue, etc.). For each framework rendered, you must provide an object stating the integration package `name`, as well as both its client and server rendering script.
```
sarah11918, line 45: We've now lost the example that the second sentence describes, so people will have no reference how to do it the other way in "non-viteful environments". Did you mean to leave that earlier paragraph as is with no changes? Is that previous way still OK? If so, when/for what things?

Do we want to instead "switch" to only talking about the new way to load renderers, rather than make this an alternative?

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -40,21 +40,21 @@ Enables rendering components using [HTML streaming](/en/guides/server-side-rende
...
+
+In a vite-full environment, renderers should be loaded by using the function `loadRenderers`, from the virtual module `astro:container`.
```
sarah11918, line 47: I'm not sure "a vite-full environment" is clear enough to be helpful. I'm not sure what specific conditions this does or doesn't include. Can you be more descriptive here? (Even to me, then I can put it into words if you like!)

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -40,21 +40,21 @@ Enables rendering components using [HTML streaming](/en/guides/server-side-rende
...
 
 The following example provides the necessary object to render an Astro component that renders a React component: 
 
 ```js
+import { getContainerRenderer } from "@astrojs/react";
```
sarah11918, line 52:
```suggestion
import { getContainerRenderer } from "@astrojs/react";
```
I'm assuming this is what you import from each different renderer integration. What happens when you have multiple to render (e.g. from `@astrojs/vue` or `@astrojs/svelte` at the same time? Are they all called exactly the same thing, and is that OK? (No naming collision?)

We do say above you need to do this "for each framework rendered" and I'm not immediately sure what that would look like if you tried to render two frameworks. Maybe the example below could be updated to show multiple? (I think then, it will be pretty easy for the reader to visualize what only one would look like.)

sarah11918 (commented):
Thanks @ematipico ! Exposing these is going to make this much easier, for sure!  🙌 

Most of my comments are questions, not edits, to make sure we are guiding people in the most helpful way!

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -40,25 +40,30 @@ Enables rendering components using [HTML streaming](/en/guides/server-side-rende
 
 #### `renderers` option
 
-**Type:** `AstroRenderer[]`
+**Type:** `SSRLoadedRenderer[]`
 
-A list of client renderers required by the component. Use this if your `.astro` component renders any [UI framework components](/en/guides/framework-components/) using an official Astro integration (e.g. React, Vue, etc.). For each framework rendered, you must provide an object stating the integration package `name`, as well as both its client and server rendering script
+A list of client renderers required by the component. Use this if your `.astro` component renders any [UI framework components](/en/guides/framework-components/) using an official Astro integration (e.g. React, Vue, etc.). For each framework rendered, you must use function `getContainerRenderer` that can be imported from the renderer library.
```
sarah11918, line 45:
```suggestion
A list of loaded client renderers required by the component. Use this if your `.astro` component renders any [UI framework components](/en/guides/framework-components/) or MDX using an official Astro integration (e.g. React, Vue, etc.).

For each official Astro integration, use its `getContainerRenderer()` helper function that can be imported from package to expose its client and server rendering scripts. These are available for `@astrojs/react`, `@astrojs/preact`, `@astrojs/solid-js`, `@astrojs/svelte`, `@astrojs/vue`, `@astrojs/lit`, and `@astrojs/mdx`.

For renderer packages outside the `@astrojs` npm org, look in their documentation for `getContainerRenderer()` or a similar function provided.
```
Tried to incorporate the note here!

Also, I added "a list of LOADED client renderers" to further emphasize the `loadRenderers()` function use, and because the type has changed. I thought maybe this is nice for consistency? Up to you if you think that is helpful!

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -40,25 +40,30 @@ Enables rendering components using [HTML streaming](/en/guides/server-side-rende
...
 
-The following example provides the necessary object to render an Astro component that renders a React component: 
+When using `vite` (`vitest`, Astro integrations, etc.), the renderers should be loaded by using the function `loadRenderers`, from the virtual module `astro:container`. Outside `vite`, you'll have to load the renderers manually.
```
sarah11918, line 47:
```suggestion
When using `vite` (`vitest`, Astro integrations, etc.), the renderers are loaded with the function `loadRenderers()` from the virtual module `astro:container`. Outside `vite`, you'll have to load the renderers manually.
```

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -40,25 +40,30 @@ Enables rendering components using [HTML streaming](/en/guides/server-side-rende
...
-    ]
+    renderers
 })
 const result = await container.renderToString(ReactWrapper);
 ```
 
+:::note
+The function `getContainerRenderer` is only available for renderers inside the `@astrojs/` npm org. Renderers outside this org will provide a similar/same function.
+:::
+
```
sarah11918, line 66:
```suggestion
```
We try to avoid notes when we can, especially if we can work the text into a normal paragraph. I tried to do that above. See what you think!

sarah11918 (commented):
Some polishing thoughts here. See what you think!

PR #8447, opened 2024-06-06
[i18nIgnore] Use Starlight new `Badge` component & remove old badge

liruifengv:
#### Description (required)

- Use Starlight new `Badge` component
- Wrap StarlightBadge component twice `src/components/Badge.astro`
- Updated component: `src/components/CardsNav.astro` and `src/components/Since.astro`
- Removed unused import
- Added custom css `.neutral-badge` for customize Badge style
- Update pages (all lanuage):
  - editor-setup.mdx
  - guides/cms.mdx
  - guides/content-collections.mdx
  - from-create-react-app.mdx
  - tutorial/1-setup/1.mdx
  - tutorial/1-setup/2.mdx
  - tutorial/1-setup/4.mdx
  - tutorial/1-setup/5.mdx
  - tutorial/2-pages/2.mdx
  - tutorial/3-components/1.mdx
  - tutorial/3-components/3.mdx
  - tutorial/4-layouts/2.mdx

- Closes #
- Suggested label:


sarah11918:
I think we needed to have this one in time for Talking and Doc'ing so we could set everyone on checking the preview in batches!  :sweat_smile: 

Let me figure out when and how we can get around to reviewing this one! :smile:

liruifengv:
> I think we needed to have this one in time for Talking and Doc'ing so we could set everyone on checking the preview in batches! 😅
> 
> Let me figure out when and how we can get around to reviewing this one! 😄

Lol. Too many files 😄

sarah11918:
So just seeing what kind of review I can get in here, and I'm noticing that there isn't a difference between "official" and "community" like there is on the original page. Can you play around with that so that the "community" badges are less noticable? The "official" one should stand out, I think!

image

sarah11918:
@liruifengv - can you resolve the conflicts from #8451 ? It looks like there were some typos in the Russian, so assume the new text/spelling is correct, but you'll need to add your badges to that. Thank you!

sarah11918 (approved):
I have only verified the English pages myself, but I trust that if something is wrong in the other languages, someone will notice! See my two comments above, and when you've resolved those, it's OK to merge! Thank you for this huge effort!!

liruifengv:
> So just seeing what kind of review I can get in here, and I'm noticing that there isn't a difference between "official" and "community" like there is on the original page. Can you play around with that so that the "community" badges are less noticable? The "official" one should stand out, I think!
> 
> image

I think I've run into some tricky issues, my custom css doesn't work after build, it doesn't have the same behavior as dev

Build: 
image

Dev:
image

src/content/docs/en/editor-setup.mdx
```diff
@@ -4,8 +4,9 @@ description: Set up your editor to build with Astro.
 i18nReady: true
 ---
 
-import Badge from '~/components/Badge.astro';
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
+import { Badge } from "@astrojs/starlight/components"
+import "~/styles/custom.css" // import custom.css for customize starlight components
```
liruifengv, line 9: Import `custom.css` instead of Starlight `customCss` option.

Because `customCss` option will bundle to a external style link tag. It can't override Starlight component's scoped style.

But, I only import it in `/en/editor-setup.mdx`, it works for all pages.

liruifengv:
Ship it! 🚀

PR #8453, opened 2024-06-06
Bump `.nvmrc` Node.js version to `18.17.1`

HiDeoo:
#### Description (required)

This pull request updates the Node.js version in the `.nvmrc` file to the minimum required version for the Astro version used.

- Astro `4.9.3` is used.
- Astro [`4.6.0`](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#460) deprecated support for versions of Node.js older than v18.17.1 for Node.js 18.


dreyfus92 (approved):
Thanks @HiDeoo, LGTM ✅

delucis (commented):
Thanks!

Noting we now [use Node 20 in our GitHub Actions](https://github.com/withastro/docs/blob/main/.github/actions/install/action.yml#L13), not sure if we want to try using that here too?

HiDeoo:
> Noting we now [use Node 20 in our GitHub Actions](https://github.com/withastro/docs/blob/main/.github/actions/install/action.yml?rgh-link-date=2024-06-07T09%3A57%3A08Z#L13), not sure if we want to try using that here too?

This is a good point and I have to say, in the case of `.nvmrc` files, I never know what the best approach is, e.g. the minimum supported version, the one from the integration workflows, etc.

More than happy to changing this if someone has a strong opinion on it or if we have a clear guideline on this.

delucis (commented):
It’s not clear to me if Vercel respects `.nvmrc` currently (unlike Netlify, which I know uses `.nvmrc` if found), but just tried setting that to Node 20 in the dashboard — might as well try to get onto the latest version if possible as I remember seeing some performance improvements in Node 20 too.

Build running here: https://vercel.com/astrodotbuild/docs/4QYXxhgPFAZA7n8UcrxxkAj17o32
(Will take a little while as new Node version == new build cache)

.nvmrc
```diff
@@ -1 +1 @@
-18.14.1
+18.17.1
```
delucis, line 1: Vercel is also building fine with v20, so I’d suggest we bump this up to the latest too:

```suggestion
20.14.0
```

> I never know what the best approach is, e.g. the minimum supported version, the one from the integration workflows, etc.

I don’t know if there’s a “best practice”, but given that tools that respect `.nvmrc` tend to switch to exactly the version specified, I think it can be OK to upgrade to the latest rather than minimum (it’s not declaring “x or higher” unless you specify a loose version like just `20`, but we’ve had issues in the past with dodgy versions of Node, so I think a fixed version might be safest).

HiDeoo: This makes total sense, thanks for the Vercel test :thumbsup:

delucis (approved):
Thanks again @HiDeoo 🎉

PR #8466, opened 2024-06-07
Added caution element to the anchor section.

### LINKED ISSUE #8465: Add a tip telling the user that anchor elements are not affected by base config
### 📚 Subject area/topic

Routing

https://docs.astro.build/**/guides/routing/

> ## Navigating between pages
> 
> Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) to navigate between routes. There is no framework-specific `<Link>` component provided.
> 
> ```astro title="src/pages/index.astro"
> <p>Read more <a href="/about/">about</a> Astro!</p>
> ```

This should explicitly say that the href will not change with different "base" values in the config.

See issue https://github.com/withastro/astro/issues/11159

nicogutz:
This will help users understand that Astro does not manage anchor elements

#### Description (required)

When building with a base config other than `/` , none of the hrefs on anchor elements are changed. This is never explicitly mentioned in the documentation. I added a caution element to inform users that they have to change it manually and gave them two main options to do so.

 
- Closes #8465, closes https://github.com/withastro/astro/issues/11159
- Suggested label: add+new+content


at-the-vr (approved):
Great PR @nicogutz 🔥🔥. LGTM 👍

src/content/docs/en/guides/routing.mdx
```diff
@@ -18,6 +18,10 @@ Astro uses standard HTML [`<a>` elements](https://developer.mozilla.org/en-US/do
 <p>Read more <a href="/about/">about</a> Astro!</p>
 ```
 
+:::caution
+Changing the [base](/en/reference/configuration-reference/#base) configuration option does not change the `href` of `<a>` elements. This has to be done manually using `Astro.url` or `import.meta.env.BASE_URL` on each element.
+:::
```
sarah11918, line 23:
``````suggestion
<p>Read more <a href="/about/">about</a> Astro!</p>

<!-- With `base: "/docs"` configured -->
<p>Learn more in our <a href="/docs/reference/">reference</a> section!</p> 
```
``````
We only use caution very rarely, for things like security issues. In general, we prefer to document "how something works" properly in the first place, not add notes on like we forgot to say something. :sweat_smile: 

When multiple options are possible, we usually extend the code block to show more examples. I think this is helpful!

sarah11918 (commented):
Thank you for this helpful addition, @nicogutz ! The information is absolutely great to have here! I've just changed the formatting to be more consistent with Astro Docs style!

I think the example below serves the purpose of demonstrating that `base` needs to be taken into account when creating URLs. Given that this is a very introductory section, and there are a variety of ways that handling the base can be accomplished, we prefer to keep the documentation simple and illustrative here.

Thank you again for this very helpful contribution, and welcome to Team Docs! :partying_face:

src/content/docs/en/guides/routing.mdx
nicogutz: > We only use caution very rarely, for things like security issues. In general, we prefer to document "how something works" properly in the first place, not add notes on like we forgot to say something. 😅

I will keep that in mind for next time :smiley: , nevertheless, the example shows the opposite of what happens, see https://github.com/withastro/astro/issues/11159.

This is why I thought it was important to explicitly indicate that the anchors are not changed with the base config parameter. For a user coming from other frameworks, they might assume that since the assets are changed by the config, then the anchors must be automatically changed as well.

PR #8476, opened 2024-06-10
Be more explicit about cookie deletion

whitep4nth3r:
#### Description (required)

Whilst the word "delete" feels intuitive in the Astro.cookies API, it hides the fact when "deleting" a cookie, you're invalidating it by setting an expiry date in the past. It might be helpful to developers to call this out (it would have been _really_ helpful to me when using this API 😅). I'm not sure of the correct wording so feel free to update or deem it unnecessary.

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -238,7 +238,7 @@ Sets the cookie `key` to the given value. This will attempt to convert the cooki
 **Type:** `(key: string, options?: CookieDeleteOptions) => void`
 </p>
 
-Marks the cookie as deleted. Once a cookie is deleted `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options available when deleting a cookie are: `domain`, `path`, `httpOnly`, `sameSite`, and `secure`.
+Marks the cookie as deleted by setting the expiration date to 0 in Unix time. Once a cookie is deleted `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options available when deleting a cookie are: `domain`, `path`, `httpOnly`, `sameSite`, and `secure`.
 
 ##### `headers`
 


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -238,7 +238,7 @@ Sets the cookie `key` to the given value. This will attempt to convert the cooki
 **Type:** `(key: string, options?: CookieDeleteOptions) => void`
 </p>
 
-Marks the cookie as deleted. Once a cookie is deleted `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options available when deleting a cookie are: `domain`, `path`, `httpOnly`, `sameSite`, and `secure`.
+Marks the cookie as deleted by setting the expiration date to 0 in Unix time. Once a cookie is deleted `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options available when deleting a cookie are: `domain`, `path`, `httpOnly`, `sameSite`, and `secure`.
```
sarah11918, line 241:
```suggestion
Invalidates a cookie by setting the expiration date in the past (0 in Unix time).

Once a cookie is "deleted" (expired), `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options available when deleting a cookie are: `domain`, `path`, `httpOnly`, `sameSite`, and `secure`.
```
Great call out, @whitep4nth3r ! In fact, I like your plain language explanation in your PR description and blog post, so I think it's great to include it here so people have another way to parse what "0 in Unix time" means.

What do you think about my suggestion above?

whitep4nth3r: This is great, your suggestion gives more context than mine! 👍🏼

ematipico (approved):

sarah11918 (approved):
Thank you @whitep4nth3r for adding some very helpful context to our docs! We are so pleased (and can't believe it took this long!!) to welcome you to Team Docs! :partying_face:

florian-lefebvre (approved):

whitep4nth3r:
Happy to help! ✨

PR #8478, opened 2024-06-10
Update astro-db.mdx

wlai-lp:
#### Description (required)
add some missing details to make the tutorial more fluid.   if you follow the tutorial, at this point, your comment table has author and content columns.   The seed script is going to fail because it's missing data in the author table and the Comment table does not have body column

- Closes #
- Suggested label:


sarah11918 (commented):
Hi @wlai-lp ! Thank you for pointing this out! However, these two examples are not meant to follow each other. This is not a tutorial where you complete every step in order, but just a very minimal example of each process. In fact, none of the examples necessarily work with each other, so that's entirely expected!

I do think your changes are helpful though, for people who are treating it like a tutorial! I will just make a minor change as we like to try to be representative of our global audience when adding new names in our code examples. Thank you again for this contribution, and I am happy to welcome you to Team Docs! :partying_face:

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -141,6 +141,11 @@ The following example defines two rows of development data for a `Comment` table
 import { db, Comment } from 'astro:db';
 
 export default async function() {
+  await db.insert(Author).values([
+    { id: 1, name: "Joe" },
+    { id: 2, name: "Jane" },
```
sarah11918, line 146:
```suggestion
    { id: 1, name: "Kasim" },
    { id: 2, name: "Mina" },
```

sarah11918 (approved):

sarah11918:
NOTE: Not merging until we figure out why the deploy build is failing, just to be safe!  :smile: (Probably not related to this PR, but want to make sure our builds are working first before merging new PRs)

PR #8491, opened 2024-06-11
Add timeout to CI

yanthomasdev:
#### Description (required)

This PR adds `timeout-minutes: 30` to our CI actions that tend to hang in some cases, so we avoid wasting resources when that happens.


delucis (approved):
Thanks @yanthomasdev! I think this looks good.

I did take a quick look through past run times and I think we might even be safe to bring this down lower? Maybe to 15 minutes? I didn’t see CI running longer than 10 minutes while looking through, even when going back several months to before we did all the performance work.

delucis (approved):

PR #8533, opened 2024-06-15
Add explanation about Vite environment variables being replaced at build time

### LINKED ISSUE #8509: Docs doesn't specify that `import.meta.env` is evaluated at build time with SSR adapter.
### 📋 Explain your issue

In the [docs](https://docs.astro.build/en/guides/environment-variables/) it's not mentioned that using `import.meta.env` will be evaluated at build time. If you are using static generation, that goes without saying. But with a SSR adapter, I think many will expect this to be evaluated per request at runtime. I at least did until i looked in the transpiled server files.

I believe some clarification would be good.

feelixe:
#### Description (required)

Add explanation about Vite environment variables being replaced at build time

closes #8509

### DIFF
--- a/src/content/docs/en/guides/environment-variables.mdx
+++ b/src/content/docs/en/guides/environment-variables.mdx
@@ -5,7 +5,7 @@ i18nReady: true
 ---
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Looks great! Thank you so much for taking the time to contribute @feelixe , and welcome to Team Docs!  🥳

feelixe:
> Looks great! Thank you so much for taking the time to contribute @feelixe , and welcome to Team Docs! 🥳

Thanks! 😊

PR #8583, opened 2024-06-19
[recipe] emphasize RTL for search result in docs

### LINKED ISSUE #8488: Missing results for query "rtl"
Hi, it would be great if you could provide a reference on how to implement the right-to-left feature in the themes.

sarah11918:
#### Description (required)

This adds a `<ReadMore>` component linking to the Resources section and calling them out, specifically including the acronym `RTL` so it will show up in search results.

- Closes #8488


src/content/docs/en/recipes/i18n.mdx
```diff
@@ -36,7 +36,11 @@ If you prefer the default language to not be visible in the URL unlike other lan
           - index.astro
         - index.astro
     </FileTree>
-    
+
+    :::tip[RTL languages]
+    If you are including right-to-left (RTL) languages such as Arabic, then you will also need to [add supporting styles](https://rtlstyling.com/posts/rtl-styling)
```
sarah11918, line 41:
```suggestion
    If you are including right-to-left (RTL) languages such as Arabic, then you will also need to [add supporting styles](https://rtlstyling.com/posts/rtl-styling).
```

dreyfus92 (approved):

PR #8585, opened 2024-06-19
Add additional explanation about define:vars directive

### LINKED ISSUE #6477: Add a warning about dynamic tags and reword another in Template Directives
### 📚 Subject area/topic

Templates directives

https://docs.astro.build/en/reference/directives-reference/

I noted two improvements more or less related that can be made on the "Template Directives" page. I could open a pull request but:
1. I'm not sure what is the best approach to improve the documentation in relation to your style guide,
2. I'm not enough familiar with the internal mechanism of Astro to be accurate.

### The issue

Using `define:vars` on a `<style />` tag in a component which depends on [dynamic tags](https://docs.astro.build/en/core-concepts/astro-syntax/#dynamic-tags) does not work. An [issue was reported on the Astro repository](https://github.com/withastro/astro/issues/7534) but has been closed. It seems a fix is not planned. So I think it would be worth mentioning it in the documentation.

### My concerns

There is already [a warning about `<script />` tags](https://docs.astro.build/en/reference/directives-reference/#definevars):

define-vars-warning

So I'm not sure what is the best approach to differentiate the two warnings:
* two warning blocks with an explicit title for each (I never saw this pattern on your documentation),
* a warning block for `<script />` and a note block for `<style />` (I saw an example of two different blocks next to each other in the documentation),
* a single generic "*Caution*" block with the two warnings next to each other (it can be confusing if the two warnings are not related since the current warning is already multi-lines... I don't know if there is any relation between the the two behaviors),
* to consider the `<style />` tag case as less important and to put it in a new paragraph between the code block and the warning block (there is a similar mention on the "[Client directives](https://docs.astro.build/en/reference/directives-reference/#client-directives)" section: "*Hydration directives are not supported when using [dynamic tags](https://docs.astro.build/en/core-concepts/astro-syntax/#dynamic-tags)*".

### Proposal

Regardless of the chosen format, we could add the following sentences:

```
The `define:vars` directive is not supported when using [dynamic tags](https://docs.astro.build/en/core-concepts/astro-syntax/#dynamic-tags). If you cannot wrap the children with an extra element like a `div`, the best approach is to manually add a ``style={`--myVar:${value}`}`` to your `Element`.
```

### Issue

There is a [warning block](https://docs.astro.build/en/reference/directives-reference/#isinline) about the directive being applied automatically for any attribute other than `src` for both `<script />` and `<style />` tags:

is-inline-warning

But in the next section (`define:vars`), the warning only mentions the `<script />` tag:

```
Using `define:vars` on a `<script>` tag implies the [`is:inline` directive](https://docs.astro.build/en/reference/directives-reference/#isinline)
```

There was a previous issue (#3165) about `is:inline` behavior with `define:vars`. It seems the statement in `is:inline` section is wrong: using `define:vars` on `<style />` tag does not automatically apply the `is:inline` directive.

So I think the warning block in `is:inline` section should be reworded to reflect the current behavior.

### Proposal

```
Most of the attributes used on a `<script>` or `<style>` tag will imply the `is:inline` directive. The few exceptions are the `src` attribute and, only on the `<style>` tag, the [`define:vars` directive](https://docs.astro.build/en/reference/directives-reference/#definevars).
```

Perhaps I'm wrong but the `lang` attribute (ie. `<style lang="scss">`) does not apply the `is:inline` directive. Maybe it should be added to the list to prevent confusion.

About the `define:vars` bug, here is the StackBlitz from the original issue (not mine): https://stackblitz.com/edit/astro-minimal-example-custom-tags?file=README.md

ArmandPhilippot:
#### Description (required)

I added additional information about `define:vars` directive on two pages:

* https://docs.astro.build/en/basics/astro-syntax/#dynamic-tags

The `define:vars` directive does not work with dynamic tags and I have seen some issues related to that on the main repository so I think it should be clarified.

* https://docs.astro.build/en/reference/directives-reference/#isinline

The statement was not exact since, as stated in #3165: using `define:vars` on `<style />` tag does not automatically apply the `is:inline` directive.

- Closes #6477.
- Suggested label: improve-documentation

ar_phi

(I'm not often connected... but I should get an email when mentioned if needed.)

### DIFF
--- a/src/content/docs/en/reference/directives-reference.mdx
+++ b/src/content/docs/en/reference/directives-reference.mdx
@@ -225,7 +225,6 @@ The `is:inline` directive means that `<style>` and `<script>` tags:
 
 :::caution
 The `is:inline` directive is implied whenever any attribute other than `src` is used on a `<script>` or `<style>` tag.
-
 The one exception is using the [`define:vars` directive](/en/reference/directives-reference/#definevars) on the `<style>` tag, which does not automatically imply `is:inline`.
 :::
 
--- a/src/content/docs/en/reference/directives-reference.mdx
+++ b/src/content/docs/en/reference/directives-reference.mdx
@@ -225,6 +225,8 @@ The `is:inline` directive means that `<style>` and `<script>` tags:
 
 :::caution
 The `is:inline` directive is implied whenever any attribute other than `src` is used on a `<script>` or `<style>` tag.
+
+The one exception is using the [`define:vars` directive](/en/reference/directives-reference/#definevars) on the `<style>` tag, which does not automatically imply `is:inline`.
 :::
 
 ```astro
--- a/src/content/docs/en/basics/astro-syntax.mdx
+++ b/src/content/docs/en/basics/astro-syntax.mdx
@@ -120,6 +120,8 @@ When using dynamic tags:
 
 - **Hydration directives are not supported.** When using [`client:*` hydration directives](/en/guides/framework-components/#hydrating-interactive-components), Astro needs to know which components to bundle for production, and the dynamic tag pattern prevents this from working.
 
+- **The [define:vars directive](/en/reference/directives-reference/#definevars) is not supported.** If you cannot wrap the children with an extra element (e.g `<div>`), then you can manually add a `style={``--myVar:${value}``}` to your Element.
+
 ### Fragments
 
 Astro supports using either `<Fragment> </Fragment>` or the shorthand `<> </>`.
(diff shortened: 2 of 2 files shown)


src/content/docs/en/reference/directives-reference.mdx
```diff
@@ -225,6 +225,8 @@ The `is:inline` directive means that `<style>` and `<script>` tags:
 
 :::caution
 The `is:inline` directive is implied whenever any attribute other than `src` is used on a `<script>` or `<style>` tag.
+
```
sarah11918, line 228:
```suggestion
```
Just making this all one paragraph so the exception doesn't seem so far removed from the earlier statement!

sarah11918 (approved):
Looks great @ArmandPhilippot and I really appreciate you taking the time both to point out the issues and suggested fixes and to come and make a PR to improve the documentation. Welcome to Team Docs! 🥳

PR #8618, opened 2024-06-21
Update the code example of `<Code />` with `transformers` in`api-reference.mdx`

liruifengv:
#### Description (required)

- Update the code example of `<Code />` with `transformers`
- `<Code />` alone in the .astro file does not render correctly. 
- I don't know if this is a bug, only update the docs to make sense for people that want quickrun.

   | Before | After |
   |---|---|
   | Steps list numbered 1, 4, 5 | Steps list numbered 1, 2, 3 |

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -1720,16 +1720,18 @@ const bar = ' world'
 console.log(foo + bar) // [!code focus]
 `
 ---
-<Code
-  code={code}
-  lang="js"
-  transformers={[transformerNotationFocus()]} />
-  
-  <style is:global>
-    pre.has-focused .line:not(.focused) {
-      filter: blur(1px);
-    }
-  </style>
+<div>
+  <Code
+    code={code}
+    lang="js"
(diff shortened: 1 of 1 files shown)


sarah11918:
I'm not sure it's expected that the `<Code />` component should need a wrapping `<div>` to style it!

Maybe we should first check to see whether this is a bug, or something about the implementation in https://github.com/withastro/astro/pull/11197

Can @ematipico @bluwy or @braebo help us out here with the code sample we should document, and whether this behavior is expected?

braebo:
That particular div in the example is indeed extraneous and unrelated to styling.

sarah11918:
@liruifengv - You might have found a bug then! If you can reproduce for the team, can you please file an Astro issue?

I'll close this as a Docs PR since the intended behaviour is already documented.

liruifengv:
> @liruifengv - You might have found a bug then! If you can reproduce for the team, can you please file an Astro issue?
> 
> 
> 
> I'll close this as a Docs PR since the intended behaviour is already documented.

Ok. I will open a issue for core repo

PR #8620, opened 2024-06-21
i18n(es): Correction of missing imports for the function 'translatePath'

fabriciodev26:
Imports should have been added so that the "translatePath" function can be used.

#### Description

I have corrected the lack of required imports for the 'translatePath' function. This ensures that the code can correctly access and use the 'translatePath' function, thus improving the integrity and functionality of the project.


TheOtterlord (commented):
Thanks for the PR!

I like being more verbose here, although if we're adding the imports, we should also add the statement defining `translatePath` to complete the snippet, as it's not exported directly. I guess we could import the `defaultLang` as the value to pass to `useTranslatedPath` (it's only needed as it's a required typed param).

```ts
const translatePath = useTranslatedPath(defaultLang);
```

I've not made any code suggestions in GitHub because our translated pages source their content from the english versions, so any changes need to be made to the `en` version of this file. We accept the english version first, then translated versions as separate PRs.

sarah11918:
Hi @fabriciodev26 - Just a reminder that if you'd like to contribute this change, it must be made to the `/en/` version of the file! (not `/es/`)

Please let us know if you're interested in making that change! :rocket:

yanthomasdev (changes requested):
Thanks @fabriciodev26, I'll block this PR from being merged till you make these changes match whatever is merged on #8645.

src/content/docs/es/recipes/i18n.mdx
```diff
@@ -398,6 +398,7 @@ Crea enlaces a los diferentes idiomas que admites para que los usuarios puedan e
     ---
     // src/components/LanguagePicker.astro
     import { languages } from '../i18n/ui';
+    import { getLangFromUrl, useTranslatedPath, useTranslations } from '../i18n/utils';
```
delucis, line 401: Can you update this to match the change in #8645?

```suggestion
    import { getLangFromUrl, useTranslatedPath } from '../i18n/utils';
    
    const lang = getLangFromUrl(Astro.url);
    const translatePath = useTranslatedPath(lang);
```

delucis (changes requested):
Thanks @fabriciodev26! I think this needs a small change.

sarah11918 (approved):
Thank you everyone! I've just gone ahead and committed the changes to match so this PR can be published and helping people!

PR #8623, opened 2024-06-21
Update Lunaria to latest

yanthomasdev:
#### Description (required)

This PR updates Lunaria to `v0.1.1` which is important since it fixes a nasty regression in how we link to GitHub's source history.


liruifengv (approved):
LGTM

PR #8632, opened 2024-06-22
RSS guide: Use `new URL()` instead of template literal

evadecker:
#### Description (required)

The RSS guide currently uses a template literal to concatenate `rss.xml` and `Astro.site`. This works without issue when the site is built, but it raises an ESLint warning:

image

This PR updates the guide to recommend the safer `URL()` constructor to create the link to the `rss.xml` file.

### DIFF
--- a/src/content/docs/en/community-resources/content.mdx
+++ b/src/content/docs/en/community-resources/content.mdx
@@ -53,6 +53,7 @@ Have you published a recipe or guide for working with Astro? [Edit this page](ht
 ### CSS
 - [Use UnoCSS in Astro](https://www.elian.codes/blog/23-02-11-implementing-unocss-in-astro/)
 - [Add dark mode to Astro with Tailwind CSS](https://www.kevinzunigacuellar.com/blog/dark-mode-in-astro/)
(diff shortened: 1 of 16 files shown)


at-the-vr (approved):
Great PR Eva 🙌 and thank you for covering all locales as well

sarah11918:
Hi @evadecker , thanks so much for providing this!

Because of our translation status tracking, however, we do ask that PRs only be submitted in one language, then we let the translators catch up. Some of these pages might be out of date translations, some not. Someone might be working on updating a page at the moment!  So, it's always easiest for us to only accept English versions of fixes, even when it *looks* like it should just be a simple fix across the board.

When you remove the other language files, we can get this merged!  🙌

evadecker:
> Hi @evadecker , thanks so much for providing this!
> 
> Because of our translation status tracking, however, we do ask that PRs only be submitted in one language, then we let the translators catch up. Some of these pages might be out of date translations, some not. Someone might be working on updating a page at the moment! So, it's always easiest for us to only accept English versions of fixes, even when it _looks_ like it should just be a simple fix across the board.
> 
> When you remove the other language files, we can get this merged! 🙌

No worries, updated!

yanthomasdev (approved):
LGTM! Thank you!

PR #8639, opened 2024-06-23
Update integrations-reference.mdx

nadar:
I am far away from an astro expert, but it confused me that there is no next() function call, like in all the other middleware examples.

+ the example code does not work and will throw an error if the route is not some-test-path
+ also everywhere in the docs it says next instead of request(?)


florian-lefebvre (approved):

sarah11918 (approved):
Thank you for the fix! 💪

ematipico (approved):
Looks good to me

yanthomasdev (approved):
Thanks @nadar, good work!

PR #8640, opened 2024-06-24
Add recipe to use Boostrap with Astro

druettiger:
#### Description

[Bootstrap](https://getbootstrap.com/) is still one of the most popular CSS frameworks and I was surprised to not find a recipe on the Astro Docs. So after figuring out how to integrate it properly in Astro, I decided to write a recipe myself.

There is an [astro-bootstrap](https://astro-bootstrap.github.io/) integration, but this is not how I would recommend developers to use Bootstrap in Astro.

Astro Discord username: druettiger


sarah11918 (approved):
Thanks so much for sharing your guide, @druettiger ! It's great to finally be able to have a reference for some Bootstrap content. Welcome to Team Docs! 🥳

PR #8645, opened 2024-06-24
added imports already created

fabriciodev26:
the 'getLangFromUrl', 'useTranslatedPath' have been added and have been called in the variable 'lang' and 'translatePath'.

#### Description (required)

I have corrected the lack of required imports for the 'translatePath' function. This ensures that the code can correctly access and use the 'translatePath' function, thus improving the integrity and functionality of the project.


src/content/docs/en/recipes/i18n.mdx
```diff
@@ -301,6 +301,10 @@ Create links to the different languages you support so users can choose the lang
     ---
     // src/components/LanguagePicker.astro
     import { languages } from '../i18n/ui';
+    import { getLangFromUrl, useTranslatedPath } from '../i18n/utils';
+    
+    const lang = getLangFromUrl(Astro.url);
+    const translatePath = useTranslatedPath(lang);
```
at-the-vr, line 307: Hey @fabriciodev26 I followed your changes and Typescript likes to complain with your current changes, primarily because `useTranslatedPath` does not exists prior to this step

at-the-vr: same changes on https://github.com/withastro/docs/blob/fbc2322ac47ae1754b5923fbb758f8c919191d67/src/content/docs/en/recipes/i18n.mdx?plain=1#L411 are resolving the mentioned issue

at-the-vr (commented):
Tried and tested on minimal template, leaving 1 suggestion, thanks a lot for your suggestion @fabriciodev26 🔥

src/content/docs/en/recipes/i18n.mdx
delucis: Ah I see that this change here was made in the wrong place. The missing code is in step 5 of the [“Hide default language in the URL”](https://docs.astro.build/en/recipes/i18n/#hide-default-language-in-the-url) section.

This was changed correctly in #8620 but not in this PR — I’ll update it and it should be good to go.

delucis (approved):
Thanks for the flagging this @fabriciodev26!

PR #8649, opened 2024-06-24
[content collections] linking to examples for generating routes

sarah11918:
#### Description (required)

Although we have examples for generating dynamic routes, and we have examples of querying with `getCollection()`, we don't (yet?) show some basic examples of e.g. blog index, tags pages etc. outside of the tutorial code.

This adds a tip to check out that code on GitHub or StackBlitz so that people can use the basic example code as a model.

Note: Must first merge https://github.com/withastro/blog-tutorial-demo/pull/31 which updates the version of Astro so that it plays nicely when opening on StackBlitz


dreyfus92 (approved):
Great tip! LGTM ✅

yanthomasdev (approved):

PR #8659, opened 2024-06-25
Fix code diff syntax in v3.mdx

liruifengv:
#### Description (required)

- Update code diff syntax in v3.mdx
- use diff syntax instead of `mark`
- Fix code highlight in VSCode and GitHub

#### Before

image
image

#### After

image

- Closes #
- Suggested label:


src/content/docs/en/guides/upgrade-to/v3.mdx
```diff
@@ -554,9 +554,9 @@ Rename all functions to their uppercase equivalent:
 - `all` to `ALL`
 - `del` to `DELETE`
 
-```js title="endpoint.ts" del={1} ins={2}
-export function get() {
-export function GET() {
+```diff lang="js" title="endpoint.ts"
+- export function get() {
++ export function GET() {
```
yanthomasdev, line 559: Thanks @liruifengv, just a nit: I would remove the identation between the code and the diff mark (e.g. `-`) so the indentation is kept the same
```suggestion
-export function get() {
+export function GET() {
```

yanthomasdev (approved):
LGTM!

PR #8674, opened 2024-06-25
i18n(fr): update i18n recipe in French

ArmandPhilippot:
#### Description (required)

* add ReadMore section added in #8583 
* add RTL abbreviation to mimic #8583 (even if this is the French translation I think French developers are more used to see RTL abbreviation than `style de droite à gauche` so it could make sense... not sure)
* make instructions consistent (each bullet point should start with imperative or infinitive form and not a mix of both)
* slightly reword some parts (eg. `basée sur Typescript` seems less literal translation than `en premier lieu Typescript`).
* fix a verb agreement

- Suggested label: i18n


thomasbnt (approved):
LGTM! Thanks @ArmandPhilippot

yanthomasdev (approved):
LGTM! Thanks!

PR #8701, opened 2024-06-28
Document how to extend integration hooks

Fryuni:
#### Description

- Document new feature allowing integrations to expose hooks for other integrations.
- Implementation PR: https://github.com/withastro/astro/pull/11304


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +615,90 @@ A list of all generated pages. It is an object with one property.
...
+export default (): AstroIntegration => {
+  return {
+    name: 'your-integration',
+    hooks: {
+      'otherLib:eventHappened': async (your, values) => {
+        // ...
+      },
+    }
+  }
+}
```
delucis, line 699: I think it’s a tiny bit confusing to keep the `your-integration` naming here when in the previous examples `your-integration` was the one defining the types etc. So maybe it makes sense to flip the naming here?

```suggestion
// other-integration/index.ts
import 'your-integration/types.ts';

export default (): AstroIntegration => {
  return {
    name: 'other-integration',
    hooks: {
      'myLib:eventHappened': async (your, values) => {
        // ...
      },
    },
  };
};
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +615,90 @@ A list of all generated pages. It is an object with one property.
...
+        }
+      },
+    }
+  }
+}
+```
+
+#### Using a custom hook from another integration
+
+Integrations can declare hooks defined by other integrations by importing their types:
```
delucis, line 684: This is a bit more verbose, but hopefully also slightly more precise (i.e. hooks aren’t declared _by_ importing the types, instead importing the types enables type checking etc. when the hooks are declared)

```suggestion
An integration that uses your custom hook will need to import your types file. It can then provide a value for your hook in `hooks` with type checking and editor hints.
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +615,90 @@ A list of all generated pages. It is an object with one property.
...
+    }
+  }
+}
+```
+
+#### Calling custom hooks on other integrations
+
+Custom hooks implemented by all the integrations in use can be accessed from the `astro:config:setup` and `astro:config:done` hooks.
+
+```ts
```
delucis, line 637: Suggesting highlighting the core for-loop in this example:

````suggestion
```ts {9-11}
````

File: src/content/docs/en/reference/integrations-reference.mdx [hunk 4]
```diff
@@ -613,6 +615,90 @@ A list of all generated pages. It is an object with one property.
...
+        }
+      },
+    }
+  }
+}
+```
+
+Custom hooks don't need to be triggered inside of those hooks, they can be stored and called later:
+
+```ts
```
delucis, line 657: And the core lines in this example too:

````suggestion
```ts {5,11-12,15-18}
````

File: src/content/docs/en/reference/integrations-reference.mdx [hunk 5]
```diff
@@ -613,6 +615,90 @@ A list of all generated pages. It is an object with one property.
...
+    }
+  }
+}
+```
+
+#### Using a custom hook from another integration
+
+Integrations can declare hooks defined by other integrations by importing their types:
+
+```ts
```
delucis, line 686: And in this one too
````suggestion
```ts {2,8-10}
````

File: src/content/docs/en/reference/integrations-reference.mdx [hunk 6]
```diff
@@ -613,6 +615,90 @@ A list of all generated pages. It is an object with one property.
...
+    }
+  }
+}
+```
+
+Custom hooks don't need to be triggered inside of those hooks, they can be stored and called later:
+
+```ts
+// your-integration/index.ts
+import './types.ts';
```
delucis, line 659: Realised we kind of skim over this import. Maybe should be mentioned in the introduction that this is needed?

delucis (commented):
Thanks Luiz! Left some notes.

sarah11918:
Hey @Fryuni , thanks for this addition! In general, I agree with all of Chris's edits here, but I'd like your confirmation before committing them to have as a base for me to start editing.

Can you take a look over Chris's comments here and resolve them all before I do my (more text-based/structural) reviewing and possibly editing for language? Thanks! :raised_hands:

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: It's a bit complicated. It is not required at all; you just have to have some way for the consumer integration to load the type augmentation from the provider integration, and this makes it the most intuitive (by just importing the integration).

But there are multiple ways of doing it and IMHO none of them are intuitive for non-TS-power-users

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: I'm a bit on the fence regarding whether this section should be directed to the author making the providing or consuming integration.

- Authors of the providing integration might refer to this when writing their own docs.
- Authors of the consuming integration might look for this information directly.

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: I think those two snippets are intended for different audiences and in each case what is "your" integration is inverse

src/content/docs/en/reference/integrations-reference.mdx
delucis: This is in the integration code itself though. If it’s not needed, maybe we remove it?

src/content/docs/en/reference/integrations-reference.mdx
delucis: I think in this case it’s on the integration that provides the custom hook to document using it, i.e. `fryunis-event-bus` calls a custom `event:happened` hook on any integrations defining it. It’s then on `fryunis-event-bus` to document a) that it expects `event:happened` and b) what the correct type import is. So I think these docs are educating integration authors about this pattern rather than how to use an integration that implements this pattern.

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: Maybe I wasn't clear. Loading the type augmentation is required, but it isn't required to be done this way.

I'll explain this in the text

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
...
+    interface IntegrationHooks {
+      'myLib:eventHappened': (your: string, parameters: number) => Promise<void>;
+    }
+  }
+}
+```
+
+#### Calling custom hooks
+
+Installed integrations can be accessed from the `astro:config:setup` and `astro:config:done` hooks. Loop over these to call your custom hook on each if available:
```
sarah11918, line 637: I think this first sentence is too general here, and is a little unclear.  You haven't shown installing anything. Installed by who? Installed in a users project? Installed by you (somewhere) as an integration author? Is it possible that you don't need the word installed here? Would it be any less true to say "Integrations can be accessed from these hooks.."?

Also, because the example is so abstract, I'm not sure *what* is actually happening below. When this becomes a concrete example (at setup, your integration will do this thing in a user's project...) This will become more clear.

"Loop over these" grammatically is also vague in that positionally, it kind of reads like "Loop over the hooks" vs "Loop over the integrations". I'm also unclear without an example why my custom hook *needs* to access every integration (again, where? installed by the user?). Will every custom hook NEED to talk to EVERY integration installed? (Maybe it does! Would be worth an explanation.) If I've written something very custom that affects my MDX pages, it's not intuitively obvious why some other unrelated random integration I have needs this hook called. 

This is about calling your custom hook, but I can't picture what it actually might be doing here.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
...
+        for (const integration of config.integrations) {
+          await integration.hooks['myLib:eventHappened'].?('your values', 123);
+        }
+      },
+    }
+  }
+}
+```
+
+To call custom hooks outside of `astro:config:setup` and `astro:config:done`, store the integrations array and use it later. `astro:config:done` should be preferred to work with integrations that get added by other integrations during `astro:config:setup`.
```
sarah11918, line 655: This actually feels like it contradicts the previous part, which honestly made it sound like your hook happens "tied to" or "at" setup or done (only).

Also, I think instead of "outside of" setup/done, maybe instead referring to "another hook", because from the example it looks like you are using this by tapping in to a different hook.

So, I'm left wondering why the section below told me things could be accessed by two hooks (which I will take as an exhaustive list), then this part shows me using a different Astro hook.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
...
+        for (const integration of config.integrations) {
+          await integration.hooks['myLib:eventHappened'].?('your values', 123);
+        }
+      },
+    }
+  }
+}
+```
+
+#### Using a custom hook from another integration
```
sarah11918, line 680: This doesn't immediately feel like the right description of what's described below, and I can't quite put my finger on it yet. This sounds like a thing *you* want to do (use someone else's integration), but the text below is about *someone else* using *your* hook. It's not wrong, but it doesn't *feel* exactly right.

I would be tempted to either change this heading to more imply "making your hook available to other integrations" to match the example below, or to change the example below to show YOU tapping into someone ELSE's custom hook. These feel like two different things.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
...
+    hooks: {
+      'myLib:eventHappened': async (your, values) => {
+        // ...
+      },
+    }
+  }
+}
+```
+
+Make sure that your type augmentation can be loaded by other integrations; otherwise, their hook implementations won't be properly typed. The easiest way to ensure this is to either have that declaration in the same file that exports your integration or import it from that file:
```
sarah11918, line 698: I would consider including this where you've defined the type in the first place. It could be a sub-heading about making your custom hook available to other integration authors.

It seems a little off to go backwards to something already discussed and duplicate a lot of existing content when this could have been an addendum (Oh, and hey, if you want to make this hook available to other integration authors...)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
...
+
+```ts {2-3} {5-12}
+// your-integration/index.ts
+// Import the types
+import './types.ts';
+
+// Or declare it in the same file
```
sarah11918, line 705: As Chris mentioned on Discord, we would prefer not to have two conflicting bits of code in the same code sample. You have the power to just choose one method to show! Alternatively, you can present the alternative separately. (But it's honestly fine to just show one thing if it will always work)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
...
+
+export default (): AstroIntegration => {
+  return {
+    name: 'your-integration',
+    hooks: { /* ... */ }
+  }
+}
+```
+
+In either case, other integrations can load your types by importing your integration or declaring it as a reference:
```
sarah11918, line 722: Then, we won't need an "in either case"

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
...
+
+```ts
+// other-integration/index.ts
+import yourIntegration from 'your-integration';
+// or
```
sarah11918, line 727: Pick one!

sarah11918 (commented):
Hey @Fryuni , so excited to have this feature in docs!

I've left some comments/questions, some of which are absolutely coming from a lack of familiarity with authoring integrations, so I apologize for the hand-holding I'll need.

The *main* takeaways here are: I'd prefer a concrete example showing an actual use case, and I'm thinking that the page hierarchy might best change here.

I think there's a case to be made for this entire section being an `<h2>` and the whole thing is not about "defining custom hooks" but rather "creating and using custom hooks" (As I've mentioned in one comment, maybe that means that the existing "Hooks" would be improved as "Provided Hooks" or something like that)

Anyway, process this and see what of my comments make any sense!  :smile:

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: Contrary to the one above, these two options are not equal.

If you want to integrate with another integration _if_ it is there, but don't want to require your users to install them if they don't want to, you need to use the `/// <directive .../>` option.
If you are going to add the integration yourself if it is not already present in the project then you can just import it.

JS ecosystem is quite bloated with dependencies that are not used at all at runtime because people just go for the import 😞

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
 
 - `pathname` - the finalized path of the page.
 
+### Defining custom hooks
+
+As an integration author, you can define custom hooks that your integration will call. This allows for runtime interactions between your integration and other integrations. Other integrations can provide implementations for your custom hooks and your integration can call those at the appropriate moment in your integration's lifecycle.
```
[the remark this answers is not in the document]
Fryuni: > Here, I think it makes sense to give a very concrete example and wherever possible, use that in all the examples below.

Hooks are very contextual; an example showcasing _just_ the hook will seem somewhat weird. The most compact and to-the-point I can think of would be for some post-processing step like a search indexer or a sitemap generator.

> Also, would like to clarify that in order to create a custom hook, you need to have an integration. You can't just make a hook without having an integration?

Yes, they are for one integration to hook into another. Both sides are integrations.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +613,121 @@ A list of all generated pages. It is an object with one property.
...
+
+#### Defining hook types
+
+Provide the type for your custom hook by augmenting Astro’s `IntegrationHooks` interface within your integration:
+
+```ts title="types.ts"
+// on an implementation file
```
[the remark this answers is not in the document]
Fryuni: The file is a `something.ts` not a `something.d.ts`. Meaning this is part of actual code that you can add wherever you want not just an ambient declaration (TS terminology) that has to go on a special file.

But yeah, it is already in the title

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: > I think this first sentence is too general here, and is a little unclear. You haven't shown installing anything. Installed by who? Installed in a users project? Installed by you (somewhere) as an integration author? Is it possible that you don't need the word installed here? Would it be any less true to say "Integrations can be accessed from these hooks.."?

Installed integrations here are the integrations used on the project your integration is running.

```ts
defineConfig({
  integrations: [
    yourIntegration(),
    mdx(),
    process.env.VERCEL === '1' ? astroIcon() : null,
  ],
});
```

`yourIntegration` and `mdx` are "installed." `astroIcon` is only "installed" when running on Vercel, even though it is installed as a project dependency.

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: Using the same example as the other comment:

```ts
defineConfig({
  integrations: [
    yourIntegration(),
    mdx(),
  ],
});
```

If `mdx` calls `updateConfig` during `astro:config:setup` to add a `tsx` integration, that new integration will only be visible to `yourIntegration` on `astro:config:done`. So, if your custom hook happens during your own `astro:config:setup` there is nothing to do, it happens before the other integration is even present to react to it. But if that is not the case, it is better to use `astro:config:done` so all integrations can react

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: Ah, cool. (I mean, not cool, but good to know)...

In this case, I'd like to add that extra knowledge here then! I know I'm asking stupid questions that maybe integration authors already know, but to let you know where I'm coming from.. historically I haven't pushed back too much on any of these API reference docs since they are a more advanced use case and many people reading them will probably already have this context or understand. But, I'm trying to bring more explicitness to them as we add new content, and the best way for me to know that they really do make sense to any audience is if they make at least a bit of sense to me! (A tough ask, I know.)

So in a case like this, where someone has two choices (and they actually do different things), I am leaning towards explaining what these different options actually do. One option could be two tiny (whatever appropriate level) sub headings, e.g.

> ####### Without installing the other integration as a dependency
> ####### Installing the other integration as a user dependency

src/content/docs/en/reference/integrations-reference.mdx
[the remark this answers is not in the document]
sarah11918: Gotcha. If you think it's helpful information to include, to differentiate from a `*.d.ts` file, then maybe we could add it to the text above? Something like

> Provide the type for your custom hook by augmenting Astro’s `IntegrationHooks` interface in any implementation file within your integration.

or ideally, if there's maybe a general way to describe the file that someone would use this in, then this could be a nice way to introduce it

> Provide the type for your custom hook by augmenting Astro’s `IntegrationHooks` interface in YOUR integration's [ADJECTIVE] file:

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: Right, this is more a note of the actual text wording here! The paragraph before sounded like "this is how you do it" and this sentence sounded like, "To do it another way" which felt a little jarring.

So this is all great stuff, I just am at the same time doing any combination of confirming some things that I honestly don't have experience with, and also pointing out where generic text/language use feels a little off or out of flow. This note was more about massaging the text a bit more to remove that feeling, and the extra context helps!

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: @Fryuni Did you have any particular thoughts here re: what an integration author might more often be thinking/their need or perspective when they read these docs?

Would you say it's more, "I need to use someone else's custom hook in their integration" or "I need to make my hook available for other people to use in their integration" ? Although practically speaking they may result in similar examples, I feel like these are two different use cases and would like the text to match the thing someone is trying to do. 

As I think I put in a comment below, if the *making your hook available* part is really like an extra step in your process when you create your hook, that could go earlier, then this section could really be about how to use someone else's hook.

src/content/docs/en/reference/integrations-reference.mdx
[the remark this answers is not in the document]
sarah11918: OK, sounds good!

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: > Is it true that you will ALWAYS want your integration to "interact" (for lack of a better word) with all of these installed integrations?

In the same sense that Astro hooks are executed for all integrations that define them, custom hooks should go through all the integrations. It is up to the integrations to implement the hook; filtering on the integration that is calling the hooks is a bit against the idea of loosely coupling things using hooks.

> You will want your custom hook called multiple times?

Using my own integration hooks as an example:
- There is [one hook](https://inox-tools.vercel.app/content-utils/git-time#itcontentgitlisted) that is called only once and only during `astro build`
- And [another hook](https://inox-tools.vercel.app/content-utils/git-time#itcontentgitresolved) that is called multiple times, twice per file, and triggers on both `astro build` and `astro dev`

The semantics of what the hook means, when it is called, and how many times it is called is up to the integration that defined it (and is calling it).

> Would an example be like, your custom hook needs to set a configuration, so you need to systematically check what each of these integrations might have set so that yours doesn't conflict, or knows what might have already been set?

I think this is more of a use case for an inter-integration API for proactive actions, not reactive (which is not a thing in Astro itself, at least not yet, but I also have a [lib](https://inox-tools.vercel.app/modular-station/api) enabling that 👉🏼👈🏼).

The use cases are much more similar to what Astro's own hooks are for. A hook just before something happens internally so integrations can modify it, another just after it happens so integrations can use the result.

Some examples from Astro's integrations:
- Sitemap could have an `astro:sitemap:discovered` hook triggered after it discovers the pages but before it generates the sitemaps to allow integrations to add or remove items programmatically (which currently can't be done)
- Sitemap could have an `astro:sitemap:done` hook triggered after it finishes generating the sitemap, which is currently unreliable since even if you use `astro:build:done`, your integration might run before the sitemap integration
- Astro DB could have a hook after the dev DB is created and seeded, giving a handle for it so integration can interact with it
- An adapter could define a hook like a `cloudflare:before:build` such that integrations can switch their internals into a Cloudflare-compatible mode with some tradeoff (like a pure JS implementation instead of a native module that would be faster)

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: Originally, I wrote each section from the perspective of who I'd expect to use them. Chris suggested keeping the same perspective throughout all sections, so I changed the text and examples to be from the perspective of the integration author defining the hooks.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +619,22 @@ A list of all generated pages. It is an object with one property.
 
 - `pathname` - the finalized path of the page.
 
+### Custom hooks
+
+Since the `IntegrationHooks` is a global interface, custom hooks can be added to integrations by extending the `IntegrationHooks` interface through [global augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#global-augmentation).
```
sarah11918, line 624:
```suggestion
Custom hooks can be added to integrations by extending the `IntegrationHooks` interface through [global augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#global-augmentation).
```
I think we can reduce repetition here!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,6 +619,22 @@ A list of all generated pages. It is an object with one property.
...
+declare global {
+  namespace Astro {
+    export interface IntegrationHook {
+      'your:hook': (params: YourHookParameters) => Promise<void>
+    }
+  }
+}
+```
+
+Astro reserves the `astro:` prefix for future built-in hooks, defining a hook with that prefix may break in future versions.
```
sarah11918, line 636:
```suggestion
Astro reserves the `astro:` prefix for future built-in hooks. Please choose a different prefix when naming your custom hook.
```
It probably makes sense to just advise people not to do it?

sarah11918 (commented):
I think this is looking pretty good?? In fact, when reviewing your PR initially, it kind of bothered me that we didn't say *anything* for Hooks, so this seems like a real improvement across the board!

Are you happy if this PR becomes *just this* reference documentation, or would you like to also try to fit in a recipe/guide in the same way we did for the Dev Toolbar API page?

Fryuni:
> Are you happy if this PR becomes just this reference documentation, or would you like to also try to fit in a recipe/guide in the same way we did for the Dev Toolbar API page?

I think the recipe/guide should go on a separate PR. It is not a requirement to get the feature released.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -54,12 +54,18 @@ interface AstroIntegration {
         logger: AstroIntegrationLogger;
     }) => void | Promise<void>;
     'astro:build:done'?: (options: { dir: URL; routes: RouteData[]; logger: AstroIntegrationLogger; }) => void | Promise<void>;
+
+    // ... integration-defined hooks gets added here
```
sarah11918, line 58:
```suggestion
    // ... any custom hooks from integrations
```
I like this matching the language that's already on the page. Since most people won't have to worry about this, I like this looking like, "oh, and maybe some stuff will get added here" with a direct heading ("Custom hooks") they can go see what it means, if they're curious.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -54,12 +54,18 @@ interface AstroIntegration {
...
   };
 }
 ```
 
 ## Hooks
 
+Astro defines hooks that integrations can implement to execute during certain parts of Astro's lifecycle and to extend its functionality. The Astro hooks are defined in the `IntegrationHooks` interface, which is part of the global `Astro` namespace.
```
sarah11918, line 65:
```suggestion
Astro provides hooks that integrations can implement to execute during certain parts of Astro's lifecycle to extend their functionality. Astro hooks are defined in the `IntegrationHooks` interface, which is part of the global `Astro` namespace.
```

sarah11918 (approved):
Approving for 4.12 release!

sarah11918:
!coauthor

PR #8706, opened 2024-06-30
example does not work, missing comma

nadar:
The example has an error because of the missing comma; so if you copy & paste the example code to test, it won't work.


kevinzunigacuellar (approved):
LGTM! ✅ Thank you for catching that error. The missing comma in the example would indeed cause issues if someone tried to copy and paste the code.

sarah11918 (approved):
Thank you, @nadar !

PR #8707, opened 2024-06-30
Fix view transitions tutorial

### LINKED ISSUE #8497: View transitions flicker
I didn't run into any flicker while navigating between routes. But the theme does flicker if the browser is refreshed. To fix this you can just add this inline script, instead of the suggested after-swap event
`<script is:inline>
  localStorage.theme === 'dark'
  ? document.documentElement.classList.add("dark")
  : document.documentElement.classList.remove("dark");
</script>`

TheOtterlord:
- Closes #8497


sarah11918 (approved):
Thank you @TheOtterlord!

PR #8728, opened 2024-07-04
markdown plugin changes

gllrmzndm:
#### Description (required)

Added an example of how the plugin works.

- Confusing examples @ https://docs.astro.build/en/guides/markdown-content/#markdown-plugins
- Sarah asked me to do a PR.

English

- Closes #
- Suggested label:


sarah11918 (approved):
Thank you for contributing this addition to docs, @gllrmzndm ! All I was hoping for was to show a config setting in the example in place, so that if readers did try to copy it, they'd get something that worked out of the box.

I appreciate the full example, but I think we can achieve the same thing just by not showing a plugin that **requires** configuration without any config. I also think this makes it obvious that if they don't like the default behaviour this produces, they can go look up remarkToc in its own docs and adjust settings.

I'm happy to have a more complete example here, and also happy to welcome you to Team Docs!  :partying_face:

PR #8740, opened 2024-07-05
Make clear that transition:persist-props does not work without transition:persist

martrapp:
#### Description (required)

Docs are not super clear on how to use `transition:persist-props`. Hopefully saying that it is an additional directive meant to be used together with `transition:persist` might help.

See https://github.com/withastro/astro/pull/10136.


martrapp:
Newline changed at end of file. No idea where that came from nor how to get rid of it.

sarah11918 (approved):
Thank you for jumping on this so quickly, @martrapp ! 🫡

PR #8749, opened 2024-07-08
Fix broken reference links in Markdown content guide

alfawal:
#### Description (required)

Fixed broken reference links in the English docs. I also did it for the other languages but reverted since its mentioned that I should edit one language at a time. Is it a PR per language or a commit per language?


delucis (approved):
Thanks for the fix! Yes, we prefer to update the English on its own in general, and then leave updating other languages to our translators — appreciate the care you took reading those instructions 🙌

PR #8750, opened 2024-07-08
i18n(fr): Fix escaping typo in Rss guide

boherm:
#### Description (required)

We have some escape issues in `description` key with French in RSS Guide, that cause some highlighting problems:

Capture d’écran 2024-07-08 à 10 55 48


delucis (approved):
Thanks for fixing these @boherm! 🎉

PR #8763, opened 2024-07-10
add information about using TypeScript with layouts

### LINKED ISSUE #8692: How do I use one layout for .md, .mdx, and .astro files with TypeScript?
### 📚 Subject area/topic

Using one Layout for .md, .mdx, and .astro

https://docs.astro.build/en/basics/layouts/#using-one-layout-for-md-mdx-and-astro

The article linked above gives the following example of an Astro layout file that accepts both JSX props and frontmatter props:

```astro
---
const { title } = Astro.props.frontmatter || Astro.props;
---
<html>
  <head></head>
  <body>
    <h1>{title}</h1>
    <slot />
  </body>
</html>
```

But it's unclear how this can be used together with TypeScript to get typesafety.

If I just do this, then TypeScript complains that `frontmatter` does not exist on `Astro.props`:

```astro
---
type Props = { title: string };
const { title } = Astro.props.frontmatter || Astro.props;
---
<html>
  <head></head>
  <body>
    <h1>{title}</h1>
    <slot />
  </body>
</html>
```

```
Property 'frontmatter' does not exist on type 'Props'.
```

So then I tried this, but that doesn't work either:

```astro
---
import type { MarkdownLayoutProps } from 'astro';

type LayoutProps = {
  title: string;
}

type Props = MarkdownLayoutProps<LayoutProps> | LayoutProps;

const { title } = Astro.props.frontmatter || Astro.props;
---
<html>
  <head></head>
  <body>
    <h1>{title}</h1>
    <slot />
  </body>
</html>
```

TypeScript still complains that `Astro.props.frontmatter` does not exist:

```
Property 'frontmatter' does not exist on type 'Props'.
  Property 'frontmatter' does not exist on type 'LayoutProps'.ts(2339)
```

Finally, I also tried:

```astro
---
import type { MarkdownLayoutProps } from 'astro';
type Props = MarkdownLayoutProps<{ title: string }>;
const { title } = Astro.props.frontmatter || Astro.props;
---
<html>
  <head></head>
  <body>
    <h1>{title}</h1>
    <slot />
  </body>
</html>
```

But then the JSX usage gives a different error:

**`src/pages/index.astro`**

```astro
---
import BaseLayout from '../layouts/BaseLayout.astro'
---

<BaseLayout title="Home page!">
    <p>test</p>
</BaseLayout>

```

```
Type '{ children: any; title: string; }' is not assignable to type 'IntrinsicAttributes & Props'.
  Property 'title' does not exist on type 'IntrinsicAttributes & Props'.
```

_No response_

### LINKED ISSUE #8934: Using a common Layout component for Markdown pages and content collections
### 📚 Subject area/topic

Layouts

https://docs.astro.build/en/basics/layouts/#using-one-layout-for-md-mdx-and-astro

I can't decide if this is a proposal, a bug report against Astro, or a documentation issue. It boils down to an ugliness with trying to re-use a Layout component between a Markdown page and a Markdown content collection entry, and how frontmatter keys are passed to the Layout component.

The Layout component of a Markdown *page* accesses frontmatter like `Astro.props.frontmatter.author` ([docs](https://docs.astro.build/en/basics/layouts/#markdown-layout-props)).

The `.astro` file that renders a Markdown *content entry* accesses the frontmatter via `Astro.props.data` ([docs](https://docs.astro.build/en/guides/content-collections/#using-content-in-astro-templates)). Following the documented conventions, these are passed through to a Layout component, which can then access the frontmatter like `Astro.props.author`.

This leads to confusion using the same Layout for both content entries and Markdown pages, how should the frontmatter be accessed? This is [partially covered in the documentation](), giving a workaround like this:

```js
const { author } = Astro.props.frontmatter || Astro.props;
```

However, this is an incomplete solution, because content collections have a Zod schema that can include field transformations. For example, "dates written without quotes around them are interpreted as Date objects", and arbitrary transformations can be applied with [`.transform()`](https://github.com/colinhacks/zod?tab=readme-ov-file#transform). These transformations are skipped entirely for Markdown pages.

I think a more complete version would be:

```js
import { collections } from '../content/config';
const { author } = Astro.props.frontmatter ? collections.blog.schema.parse(Astro.props.frontmatter) : Astro.props;
```

However, I think it would be conceptually clearer to address this another way. It's odd for a Layout component to have to invoke a parser / validator on its props.

_No response_

at-the-vr:
#### Description (required)

- Quickly guide the user how to get typesafety and autocompletion in any astro layout

- Closes #8692 
- Closes #8934


src/content/docs/en/basics/layouts.mdx
```diff
@@ -258,6 +258,28 @@ const { title } = Astro.props.frontmatter || Astro.props;
...
+
+or by overriding default Props via
+
+```astro ins={3} 
+---
+// src/components/MyLayout.astro
+interface Props { title: string }
+const { title } = Astro.props;
+---
+```
```
sarah11918, line 281:
``````suggestion
Any astro layout can be modified to introduce typesafety & autocompletion by providing the types for your props:

```astro ins={2} title="src/components/MyLayout.astro"
---
interface Props { 
	title: string
}
---
```

You can override Astro's default `Props` to handle situations where your layout recieves props of different types:

```astro ": CustomProps" title="src/components/MyLayout.astro"
---
type CustomProps = { title: string };
const { title } : CustomProps = Astro.props.frontmatter || Astro.props;
---
```
``````
So here's what I was thinking...

- Showing the default using TypeScript first feels more natural, like "this is how you type your props" before a more niche/advanced case like overriding the props.
- I do like how compact the code sample is, but I think whenever I've seen this actually used in someone's project, the `interfaceProps` is all expanded. So I'm thinking maybe we show it like that? Someone might often have more than one prop, and this does help keep their code readable, too.
- each of these need some kind of actual explanation of what's happening (not just "via", which is like "like so" or "like this" and makes the reader guess which part of the next code is the part you mean). I've tried to explain with words I think make sense, but I will need someone more familiar with TypeScript terminology to verify!

**Review by sarah11918 — COMMENTED** [review]
Thank you so much for running with this @at-the-vr ! I love clearing issues, and I think this is really helpful! Take a look at my comments and see what you think.

P.S. idea for another PR: the basic example, maybe we could also find a spot for something just like it on the astro components page so that where we talk about props, we show a TypeScript example there, too!

File: src/content/docs/en/basics/layouts.mdx (hunk 1 above)
Comment (at-the-vr) [reply src/content/docs/en/basics/layouts.mdx]: - that makes sense, I went autopilot with this one just being worried for the whole `Astro.props.frontmatter` issue getting addressed with high priority
- Sure, I will look into this. There are exemplar props I can/will refer from https://docs.astro.build/en/basics/layouts/#markdown-layout-props
- that helps the situation significantly 🙌 I will be `ptal`ing this to get accurate info

Comment (sarah11918) [reply src/content/docs/en/basics/layouts.mdx]: I was thinking like our example packages show the props like this:

image

File: src/content/docs/en/basics/layouts.mdx [hunk 2]
```diff
@@ -258,6 +258,52 @@ const { title } = Astro.props.frontmatter || Astro.props;
...
+      <p>Viewed by {viewCount} folks</p>
+    </header>
+    <main>
+      <slot />
+    </main>
+  </body>
+</html>
+```
+
+You can override Astro's default `Props` to handle situations where your layout recieves props of different types:
```
at-the-vr, line 293: ⚠️ friction alert ⚠️ `You can override ... different types` is incorrect afaik because `Props` can be defined/override to include different types as described in snippet above and I can't think of what it should be 😓
Can any experienced TypeScript developers confirm suggest a better approach? Thanks in advance ❤️

sarah11918, line 293:
```suggestion
You can also create custom `Props` to handle situations where your layout receives props of different types:
```
@at-the-vr , what do you think if we just go a little more general like this? :sweat_smile:

at-the-vr: Apologies for being blunt but the whole sentence `You can override Astro's ... props of different types:` happens to be incorrect with what I know of Types and Astro handling them. The suggestion is not doing any good here honestly.

Props can be custom in Astro and that's correct, but `Astro.props.frontmatter` is very specific. I believe its mentioned in [Markdown Layout Props](<https://docs.astro.build/en/basics/layouts/#markdown-layout-props>) - Please correct me if I happen to be faulty here 😓 

So when someone attempts the following (similar to OP's Issue linked)
```js
---
// snippet from Blog Template
type Props = CollectionEntry<"blog">["data"];

const { title, description, pubDate, updatedDate, heroImage } = Astro.props.frontmatter || Astro.props;
---
```
They receive the Property `frontmatter` does not exists Error.

With respect to the Blog template this issue can be resolved by CustomProps like this:
```diff
+ type CustomProps = CollectionEntry<"blog">["data"] & {
  title: string,
  description: string,
  pubDate: Date,
  updatedDate: Date,
  heroImage: string,
}

+ const { title, description, pubDate, updatedDate, heroImage } :CustomProps = 
 Astro.props.frontmatter || Astro.props;
---
```
and that's what I want to address in that sentence

My direction: Here's a way you can avoid Property `frontmatter` does not exists error
Docs direction: Have to rely on Sarah and Team Astro for this one 🙏

rgov:
As I commented in #8934, an overlooked aspect of this is that content entry data is processed through Zod, which can transform frontmatter entries. For example, in the blog example, the `pubDate` key is coerced to a Date object.

https://github.com/withastro/astro/blob/0dcef3ab171bd7f81c2f99e9366db3724aa7091b/examples/blog/src/content/config.ts#L10

I believe all frontmatter entries for a page would remain as they are parsed, as Strings or whatever basic types. (My experience recently trying to use the BlogEntry.astro layout for a Markdown page was that `pubDate` got parsed as a String; the documentation says they're interpreted as dates when not quoted, but I didn't find that to be true.)

Arbitrary transformations are possible through the Zod API, so it's not enough to rely on the frontmatter parser assigning types automatically, in the general case.

I'm not well-versed in TypeScript but this is the hack way I found to do it:

```typescript
import { collections } from '../content/config';
const { author } = Astro.props.frontmatter ? collections.blog.schema.parse(Astro.props.frontmatter) : Astro.props;
```

It might be worth rethinking and unifying frontmatter handling between pages and content entries, which would be more conceptually ideal, rather than requiring the user either use (and know to use) hacks like applying the schema parser in their layout, or duplicating type definitions.

src/content/docs/en/basics/layouts.mdx
```diff
@@ -258,6 +258,52 @@ const { title } = Astro.props.frontmatter || Astro.props;
...
+type CustomProps = { 
+	title: string;
+  description: string;
+  publishDate: Date;
+  viewCount: number;
+}
+const { title, description, publishDate, viewCount } : CustomProps = Astro.props.frontmatter || Astro.props;
+---
+```
+
```
sarah11918, line 306:
```suggestion
```
As I mentioned in the recent (second!) issue concerning this section, I think this section that was originally provided to serve one purpose for **v1** of Astro, long before content collections, has outlived its use. 

My suggestion here is to only add the new "Using TypeScript with layouts" section, and in fact remove lines 241-260 entirely! I don't think this section is helping people (vs causing problems) and I don't think with content collections now we really do recommend this anyway.

at-the-vr:
> quoting @rgov comment

Is the "`pubDate` getting passed as date" something you found in Astro Docs/Blog Repo because that should be rectified(coercion helps with Formatting for some context).

I am afraid to tell ya, a hack is not what Docs is supposed to tell its users. Additionally I could not make it work on my end 😞 without TypeScript complaining. Trying hard not to put down your suggestion but because its working I will ask if you can document this and share for future usecases.

As for unifying frontmatter handling, perhaps someone with more experience in this area could provide insights or suggestions on how to approach this 🙌

src/content/docs/en/basics/layouts.mdx
```diff
@@ -258,6 +258,38 @@ const { title } = Astro.props.frontmatter || Astro.props;
...
+### Working with TypeScript
+
+Any astro layout can be modified to introduce typesafety & autocompletion by providing the types for your props:
+
+```astro ins={2-7} title="src/components/MyLayout.astro"
+---
+interface Props { 
+	title: string;
+  description: string;
+  publishDate: Date;
```
at-the-vr, line 270:
```suggestion
  publishDate: string;
```

sarah11918:
Thanks for updating this, @at-the-vr ! I do think we should maybe also remove the "two layouts in one" section here. I'm not sure it's really helpful anymore. Would you be willing to remove that entire section, too? Then, when you're happy with this, I think it's good to merge!

sarah11918 (approved):
Approving the new section as-is, and would also approve if you decided to delete the problematic existing section, too!  :smile:

rgov:
> I am afraid to tell ya, a hack is not what Docs is supposed to tell its users.

I agree, but I think that the existing suggestion of `Astro.props.frontmatter || Astro.props` is a hack too. `Astro.props` (from a content entry) would be properly transformed, `Astro.props.frontmatter` (from a page) would not be. So I think it's incorrect for the documentation to suggest these are interchangeable/equivalent with the `||` operator. It may work in a few cases, but users will encounter issues even in fairly basic use cases, like the Blog example.

at-the-vr:
> existing suggestion of `Astro.props.frontmatter` || `Astro.props` is a hack too

This is a can of worms, and clearly I lack solution for this. There is one way how [Official Portfolio](https://github.com/withastro/astro/blob/main/examples/portfolio/src/pages/work/%5B...slug%5D.astro) does it which I find fitting your usecase. If it doesn't I gotta ask you to share your thoughts in [Astro Lounge](https://astro.build/chat) to receive a much proper response. Thanks for chiming in the PR ❤️

PR #8764, opened 2024-07-10
i18n(fr): add migrate-to-astro/from-nuxtjs.mdx

ArmandPhilippot:
#### Description (required)

* Translate in French [Migrating from NuxtJS](https://docs.astro.build/en/guides/migrate-to-astro/from-nuxtjs/) (`migrate-to-astro/from-nuxtjs.mdx`) added in #2232

#### Related issues & labels (optional)

- Suggested label: i18n

Since this page does not exist yet, I think we should wait for #8765 to be merged before (I'll update the translation once it's done) so I converted this PR as draft for now.

PR #8765, opened 2024-07-10
fix(content): update code diff in migrate-to-astro/from-nuxtjs.mdx

ArmandPhilippot:
#### Description (required)

While translating this page in French (#8764), I noticed an error in a code block: the `del`/`add` attributes was not matching the sentence above ("*change the set of two curly braces to one set of curly braces*"): the line numbers should be incremented by one.

See the code diff in [Nuxt Local Variables to Astro](https://docs.astro.build/en/guides/migrate-to-astro/from-nuxtjs/#nuxt-local-variables-to-astro), it should be:

The code diff after changes


sarah11918 (approved):
Merci, @ArmandPhilippot ! We are very fortunate to have our translators also proofreading our English pages for us! 🫡

PR #8803, opened 2024-07-15
Add dev flag to css preprocessor install

ajitsinghkamal:
#### Description (required)
A better way to add CSS pre-processors in projects is as dev dependency as they are needed only during the development phase of the project. Same for CSS parsers like `LightningCSS`. This PR just adds the `devDependency` flag `-D` in the install script to reflect that these packages are not required in the final production build.

### DIFF
--- a/src/content/docs/en/guides/styling.mdx
+++ b/src/content/docs/en/guides/styling.mdx
@@ -464,7 +464,7 @@ Astro supports CSS preprocessors such as [Sass][sass], [Stylus][stylus], and [Le
 ### Sass and SCSS
 
  ```shell
- npm install sass
+ npm install -D sass
  ```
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @ajitsinghkamal ! Thank you for opening up a PR!

Because of how Astro works, unlike some other frameworks, installing as a dev dependency is no different than installing as a regular dependency. And we even recommend putting all dependencies as regular dependencies unless you find a specific need not to: https://docs.astro.build/en/basics/project-structure/#packagejson

So even though in other projects this might be the recommended advice, it's an optimization that doesn't usually benefit an Astro project, so we prefer not to complicate our general guidance. We won't be making this change to docs as this isn't our recommendation for an Astro project, but you are of course welcome to put your dependencies where you like in your own project! :raised_hands:

PR #8808, opened 2024-07-15
feat(images): Add small snippet about the inferRemoteSize function

Princesseuh:
#### Description (required)

Documentation for https://github.com/withastro/astro/pull/11098


sarah11918:
Looking good @Princesseuh ! I didn't want to edit directly over top of you, but I did want to access some other parts of this file. We can still figure out the best wording, but I wanted to make sure that we thought about other areas where we might want to reference this!

sarah11918:
NOTE TO SELF: you can also use `inferSize` on `getImage`, so take a sweep and make sure nothing implies that you can't, and that this (more advanced) use case is at least mentioned.

PR #8811, opened 2024-07-15
Update GitLab Pages deployment guide

### LINKED ISSUE #8780: Error in Gitlab deploy code?
### 📚 Subject area/topic

deploy guide

https://docs.astro.build/en/guides/deploy/gitlab/

From Feedback Fish:

the .gitlab-ci.yml on your docs are not valid

_No response_

menghif:
#### Description (required)

This PR improves the [Deploy your Astro Site to GitLab Pages](https://docs.astro.build/en/guides/deploy/gitlab/) guide in the following ways:
- Use Expressive Code `title` and `diff` features for code blocks.
- Separate `site` and `base` from the main deployment steps and add more information about the two options (similar to the "Deploy to GitHub Pages" Guide).
- Add `.gitignore` changes for the build output folder.

I tested deployments on Gitlab.com with a [repository](https://gitlab.com/menghif/astro-site) under my profile and [another one](https://gitlab.com/francescomenghi/astro-site) under a group.

- Closes #8780
- Suggested label:  "improve documentation"


sarah11918:
Thank you @menghif for testing deployment and adding helpful information to the docs!

So that I can better review the deploy preview, could you please make the following changes first:

1. Please do not use all `1.` for our ordered lists. Please revert these back to their original numbers. I know it renders the same in Markdown, but it is harder to maintain if we know there is a typo in "Step 3" -- it's easier to compare the code to the output for community members making small edits.

2. It's very helpful to have the extra context from the GitHub deploy guide! Thank you for adding! However, like that page, this information is helpful to have *first*, not *last* on the page. This is what messes people up the most, and it's the most complicated thing to understand, so they should tackle it first while they're fresh. And, after finishing a set of steps, people will expect things to work... and it will not! So like the other page, can you bring that information up as the *first* thing people should set! 

3. With the above changes: you should *keep* `base` and `site` from the code example in lines 32-33 because they will have already set those, which is a good thing! Now we do not show an "incomplete" example of what their config should look like.

4. Double check that everything that did say gitHUB now says gitLAB (I notice at least lines 92 and 93)

Thank you again, this contribution is very helpful! After you make these changes, I'll read through it again and look closely for any typos or small fixes etc.!

menghif:
Hi @sarah11918, thank you for the feedback! I can definitely make all those changes.

I thought I would separate `base` and `site` since they are not required for deploying an Astro site to GitLab (as opposed to GitHub Pages). The official [example](https://gitlab.com/pages/astro) from GitLab that we link to does not include them. I tried to do the same on one of my test [repos](https://gitlab.com/francescomenghi/astro-site) and the website was deployed successfully.

Let me know if this information changes anything on how I should rewrite the page.

sarah11918:
Just to clarify what you mean: they are *not needed*, but *if you have set them, this is how they work*?

If they're not needed then I'm not sure why they are included here at all. Can you explain more about why you added them here, and what you mean when you say you tested them?

menghif:
The `site` option is not needed but strongly recommended. The Astro [docs](https://docs.astro.build/en/reference/configuration-reference/#site) explain that it is used to generate the sitemap and canonical URLs.

The `base` option is technically not needed but, not having one can lead to problems like failing to load css files or images. My example [repository](https://gitlab.com/menghif/astro-site) was [deployed](https://menghif.gitlab.io/astro-site/) successfully without a `base` option only because it is a simple Astro "Get Started" single page. A more complex site would not work correctly without `base`.

I understand the logic of putting `site` and `base` as the first step of the instructions to avoid causing potential problems. I modified the page accordingly!

src/content/docs/en/guides/deploy/gitlab.mdx
```diff
@@ -6,39 +6,90 @@ i18nReady: true
...
+      site: 'https://<username>.gitlab.io',
+      base: '/<my-repo>',
+      outDir: 'public',
+      publicDir: 'static',
+    });
+    ```
+
+    `site`
+
+    The value for `site` must be one of the following:
```
sarah11918, line 36: These requirements were true specifically for GitHub (these are not requirements for Astro at all, but very specific to GitHub.)

Have you confirmed that this is necessary for GitLab? Is there a reason to assume it works the same way?

src/content/docs/en/guides/deploy/gitlab.mdx
```diff
@@ -6,39 +6,90 @@ i18nReady: true
...
+
+    `base`
+
+    A value for `base` may be required so that Astro will treat your repository name (e.g. `/my-repo`) as the root of your website.
+
+    :::note
+      Don't set a `base` parameter if your page is served from the root folder.
+    :::
+
+    The value for `base` should be your repository’s name starting with a forward slash, for example `/my-blog`. This is so that Astro understands your website's root is `/my-repo`, rather than the default `/`.
```
sarah11918, line 52: Again, this is true for GitHub specifically, but is not something needed for every deploy host. Have you confirmed that GitLab works the same way?

sarah11918:
Thanks! I added a couple of more questions calling out specific content because I'm worried this may have been just taken from GitHub, which works differently from most other hosts. We went through a LOT of careful pain figuring out exactly what was needed for GitHub pages, and we haven't had similar comments/complaints about GitLab. So I just want to be EXTRA sure that we are not overcopying and using GitHub as a kind of general model... because from our experience, GitHub pages has been anything but typical!  :joy:

src/content/docs/en/guides/deploy/gitlab.mdx
menghif: I have confirmed these requirements by reading the official [GitLab Pages docs](https://docs.gitlab.com/ee/user/project/pages/getting_started_part_one.html):

> You created a project called blog under your username john, therefore your project URL is https://gitlab.com/john/blog/. After you enabled GitLab Pages for this project, and build your site, you can access it at https://john.gitlab.io/blog/.

> You created a group for all your websites called websites, and a project in this group is called blog. Your project URL is https://gitlab.com/websites/blog/. After you enabled GitLab Pages for this project, the site is available at https://websites.gitlab.io/blog/.

src/content/docs/en/guides/deploy/gitlab.mdx
menghif: I also confirmed this works the same way.

Other than the GitLab Pages docs, there is also this blog post from GitLab:
[How to publish your Astro Site with GitLab Pages](https://about.gitlab.com/blog/2022/10/24/publishing-an-astro-site-with-pages/)

> Depending on your project configuration, GitLab Pages will deploy your site at a URL that follows the format simlar to https://<user-or-group>.gitlab. io/<project-name>. If you want to use the default URL, you need to adjust Astro to the fact that the site is not mounted at the root path, otherwise it may not load static assets (such as the CSS files) correctly.

PR #8812, opened 2024-07-15
Migrating from Pelican: comparing integrations/plugins

Vanuan:
#### Description (required)

Some comparison of extending functionality in Pelican and Astro.


src/content/docs/en/guides/migrate-to-astro/from-pelican.mdx
```diff
@@ -25,6 +25,8 @@ When you rebuild your Pelican site in Astro, you will notice some important diff
 
 - Pelican supports writing content in Markdown and reStructured Text (`.rst`). Astro supports [creating pages from Markdown and MDX](/en/guides/markdown-content/) files, but does not support reStructured Text.
 
+- Pelican is written in Python and supports plugins to extend Markdown with custom Python code (Markdown extensions, or MDX), and custom plugins to generate any html from any file format. Astro also supports MDX, that uses [Makrdown remark plugin system](https://github.com/remarkjs/remark) written in JavaScript, but is more opinionated and provides its own Astro file format that allows to write content, html templates, client-side and server-side code and styling all in the same file.
```
sarah11918, line 28: There are a few things mentioned here, and I'm not sure what the key difference is that you want to explain. It seems like a *similarity* that both support Markdown and MDX, but in different ways. Then, I'm not sure how the Astro file format is relevant to describing Markdown and MDX, which are different file formats.

Can you tell me briefly what difference you mean to say here?

src/content/docs/en/guides/migrate-to-astro/from-pelican.mdx
```diff
@@ -57,6 +59,8 @@ Bring your existing Markdown content files to [create Markdown pages](/en/guides
 
 Pelican may have handled much of your site layout and metadata for you. You may wish to read about [building Astro Layouts as Markdown page wrappers](/en/basics/layouts/#markdownmdx-layouts) to see how to manage templating yourself in Astro, including your page `<head>`.
 
+Pelican has many plugins that extend its functionality. Explore the [official list of integrations](https://docs.astro.build/en/guides/integrations-guide/), find more in [Astro Integrations Directory](https://astro.build/integrations/), and learn [Astro Integration API](https://docs.astro.build/en/reference/integrations-reference/) to build your own integration.
```
sarah11918, line 62:
```suggestion
Like Pelican, Astro has many plugins that extend its functionality. Explore the [official list of integrations](https://docs.astro.build/en/guides/integrations-guide/) for adding features such as MDX support, and find hundreds more of community-maintained integrations in the [Astro Integrations Directory](https://astro.build/integrations/). You can even use the [Astro Integration API](https://docs.astro.build/en/reference/integrations-reference/) to build your own custom integration to extend your project's features.
```

sarah11918 (commented):
Thank you for this helpful contribution, @Vanuan ! We really appreciate you helping people compare Astro and Pelican!  🙌 

I left a small edit below, and a question for your proposed paragraph so that I can better understand to help you write something that will fit in with the rest of the page!

src/content/docs/en/guides/migrate-to-astro/from-pelican.mdx
Vanuan: You're right, as I read it it doesn't look clear. I think we can standardize the migration guide (from any content-driven framework) by analyzing which pieces of *insert framework* content could be preserved and which should be edited or written from scratch. And compare how the customization needs are implemented in each framework. I'm not sure I can be brief. Maybe it deserves a blog post, but ironically, I don't have a blog.

---
There are two concerns when doing static site generation (or handcoded website).
One is content, another is templating and styling, i.e. how this content is turned into a static html page.

I think the portability of content using human-readable text files is a key selling point of Pelican, Astro, or any content-driven framework. Without it, it's no more than just another web framework, such as NextJS, React, Vue, etc.

So, for portability, we have human readable text files (html was supposed to be human readable but failed). So, there should be some markup to specify how each piece of the content would look like.

Markdown is used in both Pelican and Astro. So far, this is standardized using CommonMark specification. And you can also inject some limited set of html tags here and there.

As a side note, Pelican, in addition, supports RST which is more friendly towards printable matters and technical documentation rather than to web content. Pelican can also be extended to consume any content and generate any output. I haven't figured out yet what it takes for Astro to consume, say, csv, json or yaml files.

The markdown syntax alone might not be enough to specify all the different types of content and text blocks. In addition, there are some elements that are common to all pages, and there are some parameters to content (front matter, metadata).

Here's where all the differences are.

First, almost de-facto standardized portion is Markdown front matter. I think it was introduced by Jekyll, and uses YAML format. It allows to specify metadata of the piece of content saved in the markdown file. Sadly, it can only be one per file, you can't have a single markdown to generate multiple sections each having its own metadata. At least not using the standardised convention. And the fields are completely custom. So, this is the first thing  you should edit when migrating from Pelican markdown files. Some comparison table of front matter fields or a link to the guide on how to use them in templates might be useful.

Then, there are Markdown extensions. They allow to push Markdown syntax beyond the CommonMark specification. GitHub has its own set of extensions, abbreviated GFM. And any decent Markdown parser allows you to inject your own implementation on how to process your custom syntax blocks. Pelican used the Python Markdown package, Astro uses [Remark](https://github.com/remarkjs/remark) which is a part of [unified](https://github.com/unifiedjs/unified) ecosystem. I think in this section we can introduce a link to the guide on how to write your own Remark plugin or integrate an existing one.

After you have your little pieces of content transformable into html, the question arises on how to piece them together. This is what templating and content organization is for. Pelican uses Jinja templates (*.html) for layout and a set of conventions for the directory structure. Astro uses its own *.astro file extension to specify different types of templates (components, pages, layouts) using Astro Template syntax. It has similar capabilities to Jinja templates, but allows seamless integration into JS ecosystem, client-side and server-side code, styling and content in the same file, etc.

And then, you have a question of styling, plugins, themes, etc. There are inherently non-migratable. While you can migrate some CSS, if you're using a custom plugin, you have to reimplement it or find similar functionality in Astro on your own.

Lastly, if you're not using Markdown, you have to create completely custom implementation to migrate your content. This can link to extensive Astro API or a guide on how to write your own extension.
---

That's the basic idea I had in my mind.

sarah11918: Thanks for the clarification!

Just to point out, the scope of these guides is rather small: it's only to give a basic idea to someone so they "know what they'd be getting into" if they migrated. It is not intended to walk someone through the process, nor to explain what Astro *is*, since we have other docs pages for that. It's important not to confuse these as "intros to Astro" which is what a lot of your addition is doing here. So, what you're describing falls a little outside of that.

Three of our guides are "larger": Next, Nuxt, and Gatsby. These include actual tips for walking through some code conversion, to show the difference. It's not really possible to get much detailed than the other guides without showing people what you mean. Even still, these are not "here is what Astro is and how it works", they are expanding on the actual process of changing some project files to their Astro counterparts.

Yes, what you're describing sounds very much like you'd like to write a blog post. :smile: (We would absolutely link to one in the community resources section if you did!) It feels very much like the content you want to write also introduces/describes Astro. For our docs, that's not what goes here since we do that on other pages.

So your options here are: distill your proposal into a key idea that's not already in a bullet on this page about a **difference in your project as an Astro project** (not "what Astro is/does"); expand the entire page to be more like the Gatsby/Next one and actually take a couple of key project files from Pelican and show converting them to Astro; or... write a "Why you should consider moving your Pelican project to Astro" blog post!  :smile:

src/content/docs/en/guides/migrate-to-astro/from-pelican.mdx
```diff
@@ -25,6 +25,8 @@ When you rebuild your Pelican site in Astro, you will notice some important diff
 
 - Pelican supports writing content in Markdown and reStructured Text (`.rst`). Astro supports [creating pages from Markdown and MDX](/en/guides/markdown-content/) files, but does not support reStructured Text.
 
+- Pelican is written in Python and supports plugins to extend Markdown with custom Python code (Markdown extensions, or MDX), and custom plugins to generate any html from any file format. Astro also supports MDX, that uses [Makrdown remark plugin system](https://github.com/remarkjs/remark) written in JavaScript, but is more opinionated and provides its own Astro file format that allows to write content, html templates, client-side and server-side code and styling all in the same file.
+
```
sarah11918, line 29:
```suggestion
```
I think it makes sense to merge the section below independently, since it's not yet clear what could or should go here at the bullet point level!

sarah11918 (commented):
Hi @Vanuan ! I would like to get the additional paragraph near the bottom of the page published and helping people as soon as possible! 

I think the other proposed change is not at the stage where it is as simple as adding a bullet point, so I'm going to remove that one in this PR and not put any pressure on either of us to figure out exactly how to handle your ideas at the same time.

You are absolutely welcome to think about what path you'd like to take going forward (I listed a few ideas!) and if appropriate, make a new PR to docs to update this page again! We are all about continually iterating and small improvements, so since we can make the page better right now, let's do it!  :muscle: 

Welcome to Team Docs! :partying_face:

sarah11918 (approved):
Approving the more scoped change to the page at this time!

Vanuan:
I'm leaning towards changing this page from an explanation article to a more practical how-to or even a small tutorial.

I’m considering two main approaches:

1. **Tutorial for Beginners**: Create a tutorial aimed at those completely new to Astro. This could help users learn Astro while converting their Pelican project, providing both an introduction and hands-on experience.

2. **Quick How-To for Experienced Users**: Develop a guide for those who are already familiar with Astro and just need a quick reference for the migration process. This would focus on practical steps and tips to achieve a successful migration.

Currently, none of the guides offer a practical example that combines various frameworks functionalities being converted to Astro while preserving the original website behavior. A guide like this could be more useful than the existing ones, which are either too abstract or sales-oriented.

Additionally, I noticed there's no migration guide for Vite/Vitepress, which is surprising given Astro’s foundation on it. Is this intentional?

I haven’t decided which direction to take yet. What do you think would be most beneficial for Astro users?

Thank you for accepting my suggestion!

sarah11918:
Hi! At this point, neither of those are exactly what we're looking for in docs at the moment, but I think those would be excellent blog posts/videos/articles to post that we can link to from here! These pages all follow a similar format and we're keeping that at this time. You may be interested in contributing to https://astro-tips.dev which is accepting a wide range of community contributions, since you mentioned that you don't currently have a blog yourself.

The inclusion/exclusion of guides is simply what we had resources available for that we felt might be illustrative for our readers. We included short guides, like Pelican, as an example of coming from a very different framework. We included guides like Gatsby and Next because they are popular similar frameworks that many are curious about seeing how they compare.

src/content/docs/en/guides/migrate-to-astro/from-pelican.mdx
```diff
@@ -57,6 +57,8 @@ Bring your existing Markdown content files to [create Markdown pages](/en/guides
 
 Pelican may have handled much of your site layout and metadata for you. You may wish to read about [building Astro Layouts as Markdown page wrappers](/en/basics/layouts/#markdownmdx-layouts) to see how to manage templating yourself in Astro, including your page `<head>`.
 
+Like Pelican, Astro has many plugins that extend its functionality. Explore the [official list of integrations](https://docs.astro.build/en/guides/integrations-guide/) for adding features such as MDX support, and find hundreds more of community-maintained integrations in the [Astro Integrations Directory](https://astro.build/integrations/). You can even use the [Astro Integration API](https://docs.astro.build/en/reference/integrations-reference/) to build your own custom integration to extend your project's features.
```
sarah11918, line 60:
```suggestion
Like Pelican, Astro has many plugins that extend its functionality. Explore the [official list of integrations](/en/guides/integrations-guide/) for adding features such as MDX support, and find hundreds more of community-maintained integrations in the [Astro Integrations Directory](https://astro.build/integrations/). You can even use the [Astro Integration API](https://docs.astro.build/en/reference/integrations-reference/) to build your own custom integration to extend your project's features.
```

src/content/docs/en/guides/migrate-to-astro/from-pelican.mdx
```diff
@@ -57,6 +57,8 @@ Bring your existing Markdown content files to [create Markdown pages](/en/guides
 
 Pelican may have handled much of your site layout and metadata for you. You may wish to read about [building Astro Layouts as Markdown page wrappers](/en/basics/layouts/#markdownmdx-layouts) to see how to manage templating yourself in Astro, including your page `<head>`.
 
+Like Pelican, Astro has many plugins that extend its functionality. Explore the [official list of integrations](/en/guides/integrations-guide/) for adding features such as MDX support, and find hundreds more of community-maintained integrations in the [Astro Integrations Directory](https://astro.build/integrations/). You can even use the [Astro Integration API](https://docs.astro.build/en/reference/integrations-reference/) to build your own custom integration to extend your project's features.
```
sarah11918, line 60:
```suggestion
Like Pelican, Astro has many plugins that extend its functionality. Explore the [official list of integrations](/en/guides/integrations-guide/) for adding features such as MDX support, and find hundreds more of community-maintained integrations in the [Astro Integrations Directory](https://astro.build/integrations/). You can even use the [Astro Integration API](/en/reference/integrations-reference/) to build your own custom integration to extend your project's features.
```

Issue #8781, opened 2024-07-11
Suggestion to improve web components example code


sarah11918:
### 📚 Subject area/topic

web components

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/client-side-scripts/#web-components-with-custom-elements

### 📋 Description of content that is out-of-date or incorrect

From Feedback Fish:

Hello! Great docs as usual, but
From the custom elements best praactices in https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements

I think that the best way to make the logic of the component inside
connectedCallback() to use best practices.

PR #8817, opened 2024-07-16
update Web Components example code

liruifengv:
#### Description (required)

update Web Components example code.

- Closes https://github.com/withastro/docs/issues/8781
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/client-side-scripts.mdx
+++ b/src/content/docs/en/guides/client-side-scripts.mdx
@@ -167,18 +167,21 @@ In this example, we define a new `<astro-heart>` HTML element that tracks how ma
   // Define the behaviour for our new type of HTML element.
   class AstroHeart extends HTMLElement {
     constructor() {
-			super();
+      super();
+    }
+
+    connectedCallback() {
       let count = 0;
 
-      const heartButton = this.querySelector('button');
-      const countSpan = this.querySelector('span');
+      const heartButton = this.querySelector('button')!;
+      const countSpan = this.querySelector('span')!;
 
       // Each time the button is clicked, update the count.
-			heartButton.addEventListener('click', () => {
+      heartButton.addEventListener('click', () => {
         count++;
         countSpan.textContent = count.toString();
       });
-		}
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks for tackling this issue, @liruifengv ! When I made the issue, I wasn't sure whether the suggestion from the feedback was good/appropriate but wanted people to comment or discuss whether it was a change we should make.

So I will ask @delucis to review this and decide if this is a change he agrees with! (I'm sure your code is great, and thank you for doing it in advance! 🙌 )

delucis:
Yes, I think we can leave it as is — but thank you @liruifengv!

The feedback is technically correct that using `connectedCallback()` in this case is a best practice — especially when authoring components that can be instantiated in JS itself (e.g. using `document.createElement('custom-element')`) — but in the context demonstrated for Astro I think this is not actually all that helpful.

- What to do in `constructor()` and what to do in `connectedCallback()` requires quite a bit of explaining
- In theory, if adding event listeners in `connectedCallback()`, we should also show removing them in `disconnectedCallback()`

For example, a fully accurate version of the first example would probably be:

```js
  class AstroHeart extends HTMLElement {
    count = 0;
    listener = () => {
      this.count++;
      const countSpan = this.querySelector('span')!;
      countSpan.textContent = this.count.toString();
    };

    connectedCallback() {
      const heartButton = this.querySelector('button')!;
      heartButton.addEventListener('click', this.listener);
    }
    
    disconnectedCallback() {
      const heartButton = this.querySelector('button')!;
      heartButton.removeEventListener('click', this.listener);
    }
  }
```

i.e. `count` and `listener` defined as a class fields, `constructor` omitted because the default constructor is all that is required, removing the listener on disconnect.

In an Astro context going into this detail is complex and probably unnecessary given the point of these examples is demonstrating using custom elements for encapsulation of logic, not the full lifecycle (which doesn’t generally apply when using custom elements in this pattern).

Given that, I think keeping things simple is acceptable and other documentation can do the work of covering the full complexity of custom elements.

Issue #8782, opened 2024-07-11
Is it true that you can use a `jsconfig.json` ?


sarah11918:
### 📚 Subject area/topic

import aliases

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/aliases/

### 📋 Description of content that is out-of-date or incorrect

From feedback fish:

I tried setting up import aliases with a jsconfig.json file, per the docs. And this seemed to be ignored. I switched over to a tsconfig.json file, and everything worked as expected. Does this page need to be updated or is this a bug?

Fryuni:
`jsconfig.json` is the same as `tsconfig.json` but with `allowJs` implied to be `true`.

This feature comes from Vite, it is even used on [some of their examples](https://github.com/vitejs/vite/blob/167006e74751a66776f4f48316262449b19bf186/packages/create-vite/template-svelte/jsconfig.json), so it should just work. Astro doesn't do anything special regarding that. Might be worth looking into whether something we set implicitly disables this.

sarah11918:
Thanks Fyuni!

@Princesseuh - would you be able to confirm here whether there might be anything we do that makes `jsconfig.json` wonky, or not "just work"?

As with all Feedback Fish reports, it's a single user so a very specific problem with an individual project is always a possibility! But I'm also not sure how many people are actually using `jsconfig` vs `tsconfig`, so it might be worth taking a peek at even one report.

Princesseuh:
It is true, however I would never recommend it. `jsconfig.json` is not exactly a standard, it's kinda a thing the VS Code team made up.

Since it's so rare, we don't test it often (I think we do have a test for it in the core repo, not sure), nor does any of our dependencies (we do a fair amount of tsconfig things ourselves, using the same library that Vite uses)

sarah11918:
OK, sounds like we should maybe just remove the suggestion and stick to the standard?

Our `base` configuration includes `allowJS; true`, so it should be good enough to just standardize and only document `tsconfig.json`.

Unless anyone screams loudly in protest, I'll remove mention of also being able to use `jsconfig.json`.  Thanks everyone!

sarah11918:
GOOD FIRST ISSUE TASK:

Find everywhere in the docs that `jsconfig.json` is mentioned as an alternative to `tsconfig.json` and remove it. (Updating all sentences  appropriately now that there is only one option instead of two).

[...]

PR #8818, opened 2024-07-16
Remove all mentioned  `jsconfig`

liruifengv:
#### Description (required)

Remove all mentioned  `jsconfig`

- Closes https://github.com/withastro/docs/issues/8782
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/imports.mdx
+++ b/src/content/docs/en/guides/imports.mdx
@@ -161,7 +161,7 @@ import logoUrl from '../../assets/logo.png?url';
 
 In this example, a developer would need to understand the tree relationship between `src/pages/about/company.astro`, `src/components/controls/Button.astro`, and `src/assets/logo.png`. And then, if the `company.astro` file were to be moved, these imports would also need to be updated.
 
-You can add import aliases from either `tsconfig.json` or `jsconfig.json`.
+You can add import aliases in `tsconfig.json`.
 
 ```json title="tsconfig.json" ins={5-6}
 {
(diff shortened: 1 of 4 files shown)


src/content/docs/en/guides/imports.mdx
```diff
@@ -188,9 +188,6 @@ import logoUrl from '@assets/logo.png?url';
 ---
 ```
 
-These aliases are also integrated automatically into [VS Code](https://code.visualstudio.com/docs/languages/jsconfig) and other editors.
```
sarah11918, line 191: I think this line is OK to keep! It doesn't mention `jsconfig` by name, even though it goes to page about it. And that page says that `jsconfig` is a descendent of `tsconfig`, so it shouldn't be confusing if people go read it.

Let's keep it in, because this line tells people that the aliases they set will be used in their editor!

liruifengv: Ok

src/content/docs/en/guides/upgrade-to/v1.mdx
```diff
@@ -592,8 +592,6 @@ In Astro v0.21+, import aliases can be added from `tsconfig.json` or `jsconfig.j
 }
 ```
 
-_These aliases are integrated automatically into [VSCode](https://code.visualstudio.com/docs/languages/jsconfig) and other editors._
```
liruifengv, line 595: @sarah11918 Should we keep this line also?

sarah11918: Yeah, I was thinking less necessary there!  🙌

sarah11918 (approved):
Looks amazing, thank you @liruifengv ! 🚀

dreyfus92 (approved):
LGTM ✅

liruifengv:
!coauthor

PR #8836, opened 2024-07-17
docs: use `var` instead of `const` for declarations

### LINKED ISSUE #8834: Types for `globalThis` aren't recognized
### Astro Info

```block
Astro                    v4.11.5
Node                     v18.20.3
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none

(also happens on Node 20 and 22)
```

_No response_

### Describe the Bug

Declaring types for `globalThis` doesn't work as described in the [docs](https://docs.astro.build/en/guides/typescript/#extending-window-and-globalthis). Steps to reproduce:

1. Declare type such as adding `declare const pagefind: string | undefined;` to `env.d.ts
1. Use `globalThis` inside an Astro component like:
  ```html
<script>
if (!globalThis.pagefind) {
	console.log("Works")
}
</script>
```
3. Run `npm run astro check`

The following error appears:

```
22:54:08 [types] Added src/env.d.ts type declarations.
22:54:08 Types generated 54ms
22:54:08 [check] Getting diagnostics for Astro files in /home/projects/qgnirbojgz.github...
src/components/Card.astro:63:17 - error ts(2339): Property 'pagefind' does not exist on type 'typeof globalThis'.

63 if (!globalThis.pagefind) {
                   ~~~~~~~~

Result (5 files): 
- 1 error
- 0 warnings
- 0 hints
```

`globalThis.pagefind` should be typed with `string | undefined` on client and server side code.

https://stackblitz.com/edit/github-rovuba?file=src%2Fcomponents%2FCard.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

### PR DESCRIPTION
**Pull request by openscript**
#### Description (required)

by @Princesseuh:
> You can't use const when declaring on the global scope, you have to use var. See https://stackoverflow.com/questions/55030498/why-dont-const-and-let-statements-get-defined-on-the-window-object/55031001#55031001

- Closes #8834
- Suggested label:

### CODE REVIEW

File: src/content/docs/en/guides/typescript.mdx [hunk 1]
```diff
@@ -122,11 +122,11 @@ import Layout from '@layouts/Layout.astro';
 You may want to add a property to the global object. You can do this by adding top-level declarations using the `declare` keyword to your `env.d.ts` file:
```
sarah11918, line 122:
```suggestion
You may want to add a property to the global object. You can do this by adding top-level declarations using `declare var` in your `env.d.ts` file:
```
So, what do you think about just mentioning here so that we don't need a whole other sentence?

src/content/docs/en/guides/typescript.mdx
```diff
@@ -122,11 +122,11 @@ import Layout from '@layouts/Layout.astro';
...
 
 ```ts title="env.d.ts"
-declare const myString: string;
+declare var myString: string;
 declare function myFunction(): boolean;
 ```
 
-This will provide typing to `globalThis.myString` and `globalThis.myFunction`, as well as `window.myString` and `window.myFunction`.
+This will provide typing to `globalThis.myString` and `globalThis.myFunction`, as well as `window.myString` and `window.myFunction`. Make sure to use `var` instead of `const` for type declarations.
```
sarah11918, line 129:
```suggestion
This will provide typing to `globalThis.myString` and `globalThis.myFunction`, as well as `window.myString` and `window.myFunction`.
```
If we incorporate the information earlier, then we don't need a "make sure" comment!

sarah11918 (commented):
Thank you for taking that discussion and coming right here to update the docs! 

Protip: it's very common for people to want to add sentences like "make sure to" to docs... but if we've told you to do it in the first place, we shouldn't need that sentence! (And, then docs doesn't have to become a whole mess of "Make sure! Make sure!) sentences.

See whether you think my suggestion is on the right track and would achieve your goal!

src/content/docs/en/guides/typescript.mdx
openscript: In my opinion this change doesn't fit, because it is also possible to declare a function with `declare function`. With this change it becomes to specific.

We should leave this sentence as it is.

src/content/docs/en/guides/typescript.mdx
openscript: It's okay to remove it, even though we didn't incorporate the change from above.

src/content/docs/en/guides/typescript.mdx
sarah11918: Ah, great call out! OK, then yeah, change doesn't work.

So, we'll let the example do the speaking. At least people won't be led astray by a faulty example, and they will tend to copy the example.

sarah11918 (approved):
Thanks very much for this correction to the docs @openscript ! I appreciate your patience, and attention to detail! Welcome to Team Docs! :partying_face:

PR #8848, opened 2024-07-18
Fix environment variable check for Netlify

delucis:
#### Description (required)

I forgot to add an environment variable on Netlify used for downloading our brand font for open graph images. This resulted in images building using Inter instead of Obviously as the headline font.

Example:

Open Graph image for the getting started page with the headline set in Inter instead of Obviously

I have now fixed that in the Netlify dashboard, but the reason this error was possible was that we were still checking for a `VERCEL` environment variable before throwing errors. This PR updates this to check the equivalent `NETLIFY` variable instead.

I also considered using the more generic `CI` variable, but docs gets built a bunch of CI places that don’t need the custom font (our GitHub Actions, `withastro/astro` smoke tests, etc.), so I think we probably can’t.

### DIFF
--- a/src/pages/open-graph/_fetchFont.ts
+++ b/src/pages/open-graph/_fetchFont.ts
@@ -32,7 +32,7 @@ export async function fetchBrandFont() {
 	} catch (error) {
 		// When running locally, if anything goes wrong, we can safely return
 		// nothing and continue with the default local fonts.
-		if (!import.meta.env.CI) return undefined;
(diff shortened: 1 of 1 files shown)


at-the-vr (approved):
The attention to details is impeccable.`Obviously` is back again 👍

TheOtterlord:
Perfect!

TheOtterlord (approved):

PR #8850, opened 2024-07-18
feat: document `addClientRenderer`

ematipico:
#### Description (required)

This PR documents a API that was added to the container API

https://github.com/withastro/astro/pull/11486

### DIFF
--- a/src/content/docs/en/reference/container-reference.mdx
+++ b/src/content/docs/en/reference/container-reference.mdx
@@ -90,6 +90,7 @@ When the container is called at runtime, or inside other "shells", the `astro:co
 
 Server renderers are required to build your project, and must be stored in the container for every framework used. Client renderers are additionally needed to any hydrate client-side components using [`client:*` directives](/en/reference/directives-reference/#client-directives).
 
+Only one import statement is needed per framework. Importing a renderer makes both the server and client renderers available to your container. However, **server renderers must be added to your container before client renderers**. This allows your entire container to render first, and then hydrate any interactive components.
 
 The following example manually imports the necessary server renderers to be able to display static Vue components and `.mdx` pages. It additionally adds both server and client renderers for interactive React components.
 
--- a/src/content/docs/en/reference/container-reference.mdx
+++ b/src/content/docs/en/reference/container-reference.mdx
@@ -84,7 +84,7 @@ const container = await experimental_AstroContainer.create({
 const result = await container.renderToString(ReactWrapper);
 ```
 
-#### Adding a renderer manually
+#### Adding a server renderer manually
 
 When the container is called at runtime, or inside other "shells", the `astro:container` virtual module's helper functions are not available. Import the necessary server renderer manually and store it inside the container using `experimental_Astro.Container.addServerRenderer`.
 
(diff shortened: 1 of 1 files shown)


sarah11918 (commented):
Left an initial review edit to see what you think, @ematipico !

I think there's so much in common here that it really can be the *same* section about adding renderers manually. See how you think this feels, and adjust if it's not accurate in some way!

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -10,7 +10,7 @@ The Container API allows you to render Astro components in isolation.
 
 This experimental server-side API unlocks a variety of potential future uses, but is currently scoped to allow [testing of `.astro` component output](/en/guides/testing/#vitest-and-container-api) in `vite` environments such as `vitest`.
 
-It also allows you to [manually load rendering scripts](#adding-a-renderer-manually) for creating containers in pages rendered on demand or other "shell" environments outside of `vite` (e.g. inside a PHP or Elixir application).
+It also allows you to [manually load rendering scripts](#adding-a-server-renderer-manually) for creating containers in pages rendered on demand or other "shell" environments outside of `vite` (e.g. inside a PHP or Elixir application).
```
sarah11918, line 13:
```suggestion
It also allows you to [manually load rendering scripts](#adding-a-renderer-manually) for creating containers in pages rendered on demand or other "shell" environments outside of `vite` (e.g. inside a PHP or Elixir application).
```

ematipico:
Thank you @sarah11918, I like the new edits! I still think we should add the `:::caution` block because the order matters.

It isn't possible to use `addClientRenderer` before `addServerRenderer`. The function `addServerRenderer` must be called first.

sarah11918:
Ah, cool. I didn't get that the *order* mattered. Let me get back to that then! Thank you!

src/content/docs/en/reference/container-reference.mdx
```diff
@@ -86,19 +86,24 @@ const result = await container.renderToString(ReactWrapper);
 
 #### Adding a renderer manually
 
-When the container is called at runtime, or inside other "shells", the `astro:container` virtual module's helper functions are not available. Import the necessary server renderer manually and store it inside the container using `experimental_Astro.Container.addServerRenderer`.
+When the container is called at runtime, or inside other "shells", the `astro:container` virtual module's helper functions are not available.  You must import the necessary server and client renderers manually and store them inside the container using `addServerRenderer` and `addClientRenderer`.
 
-The following example manually imports the server renderers for React and Vue components, and MDX for `.mdx` pages. 
+Server renderers are required to build your project, and must be stored in the container for every framework used. Client renderers are additionally needed to any hydrate client-side components using [`client:*` directives](/en/reference/directives-reference/#client-directives).
+
```
sarah11918, line 92:
```suggestion
Server renderers are required to build your project, and must be stored in the container for every framework used. Client renderers are additionally needed to any hydrate client-side components using [`client:*` directives](/en/reference/directives-reference/#client-directives).

Only one import statement is needed per framework. Importing a renderer makes both the server and client renderers available to your container. However, **server renderers must be added to your container before client renderers**. This allows your entire container to build first, and then hydrate any interactive components.
```
Instead of a note, does this accurately describe what's happening? A paragraph here that both mentions the order requirement and explains why might make the information more "sticky" for people, since this parallels how they think about Astro sites in general.

If this is the case, I might adjust the blank line in the example below to group all the server renderers together, and leave the client renderer on its own after the blank line. This would be another way to subtly enforce this pattern. 

(I don't actually know whether ALL the server renders need to be imported first, before ANY client renderers, or whether it's just to make sure for each framework, the client comes second. But if we show importing ALL the server renderers first, people will get the idea.)

sarah11918: @ematipico If you're OK with the above change, (or just edit it slightly if it needs correcting) then I think this can merge any time!

ematipico: Thank you! I applied the edits, and I'm going to merge

PR #8866, opened 2024-07-19
fix: typo in `configuration-reference.mdx`

jsparkdev:
#### Description (required)

- fix typo in `configuration-reference.mdx`

- Closes #
- Suggested label:


liruifengv (approved):
LGTM!

sarah11918:
Thank you @jsparkdev , but since this is the configuration reference page, this will need to be fixed in the `withastro/astro` repo itself! The specific file is mentioned at the top of this file in the frontmatter. I need to fix another typo, so I can make this there at the same time too, thanks!

sarah11918:
I've made this PR to correct this, and another typo I spotted in the source (and had fixed only locally before publishing yesterday) Thanks again, and this will be fixed when I run the action to bring these changes here!

PR #8871, opened 2024-07-19
Visual separation of styles

JavGuerra:
The separation makes code reading easier.

### DIFF
--- a/src/content/docs/es/tutorial/5-astro-api/3.mdx
+++ b/src/content/docs/es/tutorial/5-astro-api/3.mdx
@@ -221,6 +221,7 @@ const pageTitle = "Índice de etiquetas";
     ))}
   </div>
 </BaseLayout>
+
 <style>
   a {
     color: #00539F;
@@ -295,6 +296,7 @@ Sigue los pasos que se indican a continuación y comprueba tu trabajo comparánd
 
      <slot />
    </BaseLayout>
+
    <style>
      a {
        color: #00539F;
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi @JavGuerra ! I appreciate you've been helping with a lot of fixes!

We do have some processes though, and so that we can all function as a team we ask you to follow them. I would like to give you some feedback so that you can contribute in the best possible way that helps us!

- Please do not make changes that are not **corrections to spelling/grammar/language** in translated Spanish language files. The **only** changes to docs code, including code samples, must happen in English files first. All other languages copy what is in the English files, and that is how changes are made.

- Please remember that every change you make means 14 languages receive notification that their translations are "out of date and need fixing" and several people must work to update the docs. When you make a change, it is important to consider how important the change is. This is not a very significant change to make, and in this case, we often wait until something needs fixing on a page and then if we decide improvements like this are necessary, make them at that time. If you are not sure whether it is a good idea to make a PR to fix something, we ask that you file an Issue on GitHub first, instead of making a PR.

- You must use a proper title format to help organize our PRs in other languages. I have mentioned this a few times now and changed a few of your titles for you, but it is not helpful if you keep submitting PRs and do not follow our format. If you want to particpate, we must ask that you follow our processes. Not doing so creates extra work for us, and we have several PRs to manage!

So, I must close this PR because it does not change the `/en/` file. I also do not think this is a necessary change to documentation to make right now, so I think we can not do this one yet!

JavGuerra:
Thank you very much Sarah for all the instructions. You editors are doing an impressive job. I will be attentive when presenting new changes. Thank you for the modifications you made to adapt the contents appropriately.

PR #8914, opened 2024-07-24
feat(rewrite): document rewrite APIs

ematipico:
#### Description (required)

This PR documents the rewrite APIs, which we plan to release in the next minor


sarah11918 (commented):
Just publishing a partial review (API reference and first paragraph of Routing page) just so nothing happens to it in the meantime.

Still to review: Routing page i18n example and middleware page!

ematipico:
@sarah11918 FYI this commit [`e4fd1cc` (#8914)](https://github.com/withastro/docs/pull/8914/commits/e4fd1cc470c1fcef041e6125c96790d0fa1d7106) added more information that was missing.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -399,6 +399,47 @@ Use [`Astro.url`](#astrourl) to construct your own canonical URL.
 
 The [canonical URL][canonical] of the current page.
 
+### `Astro.rewrite()`
+
+<p><Since v="4.13.0" /></p>
+
```
ArmandPhilippot, line 405: As suggested by @sarah11918 in #8917, we should add the type of the function:

```suggestion

**Type:** `(rewritePayload: string | URL | Request) => Promise<Response>`

```

I replaced the [`RewritePayload` type](https://github.com/withastro/astro/blob/208f7f6470e873c715806e006747d0bd6351c3f9/packages/astro/src/%40types/astro.ts#L2818) (the parameter) by its value to make it more clear by reading the documentation (even if the possible values ​​are detailed below).

sarah11918: Thank you! Will ask @ematipico to confirm this!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -759,6 +800,51 @@ export function GET({ redirect }: APIContext) {
 
 See also: [`Astro.redirect()`](#astroredirect)
 
+
```
sarah11918, line 803:
```suggestion
### `context.rewrite()`

```
Are we missing a heading here?

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -759,6 +800,51 @@ export function GET({ redirect }: APIContext) {
...
+     "x-custom-header": JSON.stringify(Astro.locals.someValue)
+   }
+ }));
+}
+```
+
+The status code returned by the request will change based on the result of the rewrite. For example, if you call `context.rewrite("/foo/bar")`, but `/foo/bar` doesn't exist even though you have a dynamic route, Astro will return a `404` status code. 
+
+You can take advantage of this logic by conditionally return different status code. For example, for a certain user can't access certain routes, you can call `context.rewrite("/404")` and Astro will render the content of the 404 page (built-in or custom). 
+
```
sarah11918, line 847:
```suggestion

See also: [`Astro.rewrite()`](#astrorewrite)

```
Are we also missing the "see the main entry" line that redirects had?

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -399,6 +399,47 @@ Use [`Astro.url`](#astrourl) to construct your own canonical URL.
...
+  headers: {
+    "x-custom-header": JSON.stringify(Astro.locals.someValue)
+  }
+}))
+---
+```
+
+The status code returned by the request will change based on the result of the rewrite. For example, if you call `Astro.rewrite("/foo/bar")`, but `/foo/bar` doesn't exist even though you have a dynamic route, Astro will return a `404` status code. 
+
+You can take advantage of this logic by conditionally return different status code. For example, for a certain user can't access certain routes, you can call `Astro.rewrite("/404")` and Astro will render the content of the 404 page (built-in or custom).
```
sarah11918, line 440: This doesn't feel very clear right now. 

Is the first paragraph anything more than routes that don't exist will 404 (like any other routes? not sure why this is special?). I don't feel like this paragraph is helpful, or provides any specific information about the rewrite. So you'll have to connect the dots a little stronger for me if this is really important!

Is the second paragraph really about conditionally *rewriting* (not about returning status codes?) If there's a 404 page, then the 404 page *itself* isn't 404'ing, right? So from the user perspective using this feature, you are rewriting with a path that does exist, just like any other path? (It just happens to be a 404 page) As written, this doesn't feel like it's about status codes since it's describing rewriting to a page that exists.

Since the other examples of usage have been very simple, if you want to talk about conditional rewrites it might be nice to actually show the code for a conditional rewrite here. Using the example of 404'ing for a restricted page is a good example, but I think it's more helpful to show what that code looks like!

(And of course, whatever we decide here, we can use for the context entry below, too, or decide just to link up here!)

ematipico: > Is the first paragraph anything more than routes that don't exist will 404 (like any other routes? not sure why this is special?). I don't feel like this paragraph is helpful, or provides any specific information about the rewrite. So you'll have to connect the dots a little stronger for me if this is really important!

Ok, I'll try to use some bridge phrase.

> Is the second paragraph really about conditionally rewriting (not about returning status codes?)

It is **not** about conditional rewriting, it's about status code and expectations, but it's a bit silly to say if you do this, this other thing happen. So, I was trying to put some real usage context. Let me see if I can do something better with an example.

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,32 @@ if (!isLoggedIn(cookie)) {
...
+
+A rewrite allows you serve a different route without redirecting the browser to a different page. The browser will show the same address in the URL bar, but will display the content of the URL provided to `Astro.rewrite()`.
+
+This can be useful for showing the same content at multiple paths (e.g. `/products/shoes/men/` and `/products/men/shoes`/).
+
+For content that has permanently moved, or to direct your user to a different page (e.g. a user dashboard after logging in), use a [redirect](#redirects) instead.
+
+A rewrite is really useful in an i18n context, for example. Let's suppose your website serves localised version of Spanish e.g. `/es-es/articles/introduction` (Spain) and `/es-mx/articles/introduction` (Mexico). However, your website also serves pages for `es-CU` (Cuba), but you're missing the localised content for `/es-cu/articles/introduction`. 
+
+However, you don't to serve that page as a 404, and you don't want to serve a redirect for SEO purpose. That's where the rewrites comes into play. You can decide to **fallback** to the Mexican localised page using the new `Astro.redirect` method:
```
sarah11918, line 311:
```suggestion
Rewrites are also useful to provide fallback content, such as for a localized page route whose content does not exist. If you don't want to serve a redirect for SEO purposes, and do not want to show a 404, then you can use a rewrite to display a different page at the requested route. The following example uses a rewrite to display the `/es-mx/` version of a page (Mexican Spanish) without from the `/es-CU/` (Cuban Spanish) URL path. 
```

Just rough draft for now, but I'm thinking all of this could be condensed to something like this...

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,32 @@ if (!isLoggedIn(cookie)) {
...
+
+```astro title="src/pages/es-cu/articles/introduction.astro"
+---
+return Astro.rewrite("/es-mx/articles/introduction")
```
sarah11918, line 315:
```suggestion
return Astro.rewrite("/es/articles/introduction")
```
See the comment I made above for the corresponding slight change in the example -- I think it's a simpler example that someone wants to just write one general Spanish page and use that for variants? (At least, is visually simpler, so the idea gets across better, I think!)

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,32 @@ if (!isLoggedIn(cookie)) {
...
+---
+```
+
+When you'll navigate the URL `/es-cu/articles/introduction`, Astro will render the route `src/pages/es-mx/articles/introduction.astro`.
```
sarah11918, line 319:
```suggestion
In the previous example, when a visitor navigates to the URL `/es-cu/articles/introduction`, Astro will render the content generated by the file `src/pages/es/articles/introduction.astro`
```

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,32 @@ if (!isLoggedIn(cookie)) {
...
+
+The function `Astro.redirect()` will trigger a new complete rendering phase, which means that, if you have middleware, it will be executed again for the new route/request.
```
sarah11918, line 321: This feels like maybe it should go somewhere before the example, as it's more general about how it works. It also feels like it might get lost or grouped in with the example, when this is a standalone fact. I'll try moving it up and see what we think! 

```suggestion
```

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,32 @@ if (!isLoggedIn(cookie)) {
 </html>
 ```
 
+## Rewrites
+
+<Since v="4.13.0" />
+
+A rewrite allows you serve a different route without redirecting the browser to a different page. The browser will show the same address in the URL bar, but will display the content of the URL provided to `Astro.rewrite()`.
```
sarah11918, line 303:
```suggestion
A rewrite allows you serve a different route without redirecting the browser to a different page. The browser will show the same address in the URL bar, but will display the content of the URL provided to [`Astro.rewrite()`](/en/reference/api-reference/astrorewrite).
```
Realized we should have a link to the reference here!

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,32 @@ if (!isLoggedIn(cookie)) {
...
+
+This can be useful for showing the same content at multiple paths (e.g. `/products/shoes/men/` and `/products/men/shoes`/).
```
sarah11918, line 305:
```suggestion
Rewrites can be useful for showing the same content at multiple paths (e.g. `/products/shoes/men/` and `/products/men/shoes`/).
```

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,58 @@ auth response
...
+  // We rewrite to a new route, the homepage 
+  return next("/")
+}
+
+// This is the second middleware function that will be triggered by Astro
+async function second(ctx, next) {
+  // `ctx` will contain new information, computed by the previous rewrite
+  console.log(ctx.url.pathname) // this will log  "/"    
+  return next()
+}
```
sarah11918, line 243:
```suggestion
// Current URL is https://example.com/blog

// First middleware function
async function first(_, next) {
  console.log(ctx.url.pathname) // this will log "/blog"
  // Rewrite to a new route, the homepage 
  // Return updated `ctx` which is passed to next function
  return next("/")
}

// Current URL is still https://example.com/blog

// Second middleware function
async function second(ctx, next) {
  // Receives updated `ctx`
  console.log(ctx.url.pathname) // this will log  "/"    
  return next()
}
```
Just tried to make this a bit more compact, and also emphasizing that the URL hasn't changed. Maybe something like this?

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,58 @@ auth response
...
+
+  return next();
+};
+```
+
+The `next` function accepts the same payload of the [`Astro.redirect() function`](/en/reference/api-reference/#astrorewrite).
+
+In case you have multiple middleware functions chained via [sequence](#chaining-middleware), using `next("/")` with  a payload will rewrite the request **in place** which means that the next middleware function will receive the **new `Request`**:
+
+```js
```
sarah11918, line 229:
``````suggestion
```js title="src/middleware.js"
``````

**Review by sarah11918 — APPROVED** [review]
Whew! OK, here are some thoughts for the rest of the document!

See what you like, (and what is correct) and change/commit whatever makes you happy. When there are lots of comments like this, it can be distracting to read and review, so please do go ahead and commit whatever changes you want. It's easier for me to read from a fresh file again.

File: src/content/docs/en/guides/middleware.mdx [hunk 13]
```diff
@@ -196,6 +197,58 @@ auth response
 validation response
 ```
 
+## Rewriting
+
+<Since v="4.13.0" />
+
+The `APIContext` exposes a method called `rewrite`, which works the same way as [Astro.rewrite](/en/guides/routing/#rewrites).
+
+The `next` function optionally accept a parameter that allows to **rewrite** the current `Request`. For example, if a user isn't logged in, you could **rewrite** the current request without the need of a redirect:
```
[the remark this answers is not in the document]
ematipico: I removed this part:
> This executes [`context.rewrite()`](/en/reference/api-reference/#contextrewrite) using the provided value (`rewritePayload`)

Because it's incorrect. They are two separate functions with two separate business logic.

src/content/docs/en/guides/routing.mdx
ematipico: I leave the example to you, since you have more hands-on experience :)

src/content/docs/en/reference/api-reference.mdx
ematipico: @sarah11918 please have a look at this new version [`cc081b4` (#8914)](https://github.com/withastro/docs/pull/8914/commits/cc081b450ce5260672b9b3f215dba4df9727c460)

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,76 @@ auth response
...
+        "x-redirect-to": context.url.pathname
+      }
+    }));
+  }
+
+  return next();
+};
+```
+
+The `next` function optionally accept a parameter that allows to **rewrite** the current `Request`. For example, if a user isn't logged in, you could **rewrite** the current request without the need of a redirect:
```
sarah11918, line 223: I'm not sure what the difference between these two slightly different methods is, practically speaking. Do they both do the exact same thing in the end?

If we are going to show both, is the point to say that you can "do the same thing either way"? Is the second one a "convenient shortcut" or preferable in any way? These examples look very similar on first glance, and someone might have trouble seeing the difference without looking closely. So it's helpful if we can draw the reader's attention to why we are showing them two slightly different examples, and what we want them to learn from them.

And, if you expect people to use only ONE of them 80% of the time, it's often OK to only show the common/preferred one. If they both have their place/advantages, then it's helpful if we explain why when we show both.

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,76 @@ auth response
...
+    }));
+  }
+
+  return next();
+};
+```
+
+The `next` function accepts the same payload of the [`Astro.rewrite() function`](/en/reference/api-reference/#astrorewrite). The location of the rewrite path can be provided as a string, URL, or `Request`.
+
+When you have multiple middleware functions chained via [sequence()](#chaining-middleware), submitting a path to `next()` will rewrite the `Request` in place and **it _won't_ execute again the middleware**. The next middleware function in the chain will receive the new `Request`:
```
sarah11918, line 244:
```suggestion
When you have multiple middleware functions chained via [sequence()](#chaining-middleware), submitting a path to `next()` will rewrite the `Request` in place and the middleware will not execute again. The next middleware function in the chain will receive the new `Request` with its updated `context`:
```
How does this sound?

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,76 @@ auth response
 validation response
 ```
 
+## Rewriting
+
+<Since v="4.13.0" />
+
+The `APIContext` exposes a method called `rewrite`, which works the same way as [Astro.rewrite](/en/guides/routing/#rewrites).
```
sarah11918, line 204:
```suggestion
The `APIContext` exposes a method called `rewrite`, which works the same way as [Astro.rewrite](/en/guides/routing/#rewrite).
```

src/content/docs/en/guides/middleware.mdx
ematipico: Ok, I think I see the confusion, probably because I put the important bits in the incorrect context. Here's the breakdown:

`ctx.rewrite`:
- it retriggers a new rendering phase (the middleware is called again)
- it doesn't call subsequent middleware functions, in case `sequence` is used

`next()` with payload
- it doesn't retrigger a new rendering phase
- it calls subsequent middleware functions, in case `sequence` is used. The next middleware functions will receive the new `Request`

sarah11918: Thanks! This is helpful! I will look at this more closely after our calls!

sarah11918:
@ematipico heads up that I moved some of the newer API page content into the routing page because I felt it matched more with the Routing/guide style of the page, and kept the API page short (matching what we have for redirects)

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,58 @@ if (!isLoggedIn(cookie)) {
 </html>
 ```
 
+## Rewrites
+
+<p><Since v="4.13.0" /></p>
+
+A rewrite allows you serve a different route without redirecting the browser to a different page. The browser will show the original address in the URL bar, but will instead display the content of the URL provided to [`Astro.rewrite()`](/en/reference/api-reference/astrorewrite).
```
sarah11918, line 303:
```suggestion
A rewrite allows you serve a different route without redirecting the browser to a different page. The browser will show the original address in the URL bar, but will instead display the content of the URL provided to [`Astro.rewrite()`](/en/reference/api-reference/#astrorewrite).
```

src/content/docs/en/guides/middleware.mdx
sarah11918: Ok, @ematipico - Please check [this commit](https://github.com/withastro/docs/pull/8914/commits/7dc8649f13631385971dc0d15842ce48798dd243) to see if this might be an accurate way to distinguish the two samples.

Since they do in fact act differently but the code comments are identical, I added a third `//` code comment line which should be different to help illustrate how the two examples work differently, or when you might use one instead of the other, or what the different effects are.

I also edited the intro paragraph to each one a tiny bit, but I think if we get that third code comment line just right, then this will be really helpful! So please check those in particular and correct them as necessary!

ematipico:
I think we achieved what I wanted and what you wanted :D I am happy with the final changes

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1392,6 +1476,11 @@ A function that intercepts (reads and modifies) the `Response` of a `Request` or
 
 This is an optional argument of `onRequest()`, and may provide the required `Response` returned by the middleware.
 
+<Since v="4.13.0" />
+
+**Type:** `(rewritePayload?: string | URL | Request) => Promise<Response>`
+
+It accepts, optionally, a string, `URL`, or `Request` to trigger a [rewrite](/en/guides/routing/#rewrites) of the current request.
```
sarah11918, line 1483:
```suggestion
It accepts an optional parameter in the form of a a string, `URL`, or `Request` to trigger a [rewrite](/en/guides/routing/#rewrites) of the current request.  
```

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,79 @@ auth response
...
+
+Use `context.rewrite()` inside middleware to display a different page's content without [redirecting](/en/guides/routing/#dynamic-redirects) your visitor to a new page. This will trigger a new rendering phase, causing any middleware to be re-executed.
+
+```js title="src/middleware.js"
+import { isLoggedIn } from "~/auth.js"
+export function onRequest (context, next) {
+  if (!isLoggedIn(context) {
+    // If the user is not logged in, update the Request to render the `/login` route and
+    // add header to indicate where the user should be sent after a successful login.
+    // Re-execute middleware. CHECK THIS IS TRUE/THE POINT!!
```
sarah11918, line 214:
```suggestion
    // Re-execute middleware.
```

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,79 @@ auth response
...
+
+You can also pass the `next()` function an optional URL path parameter to rewrite the current `Request` without retriggering a new rendering phase. The location of the rewrite path can be provided as a string, URL, or `Request`:
+
+```js title="src/middleware.js"
+import { isLoggedIn } from "~/auth.js"
+export function onRequest (context, next) {
+  if (!isLoggedIn(context) {
+    // If the user is not logged in, update the Request to render the `/login` route and
+    // add header to indicate where the user should be sent after a successful login.
+    // Return a new `context` to any following middlewares. CHECK THIS IS TRUE/THE POINT!!
```
sarah11918, line 234:
```suggestion
    // Return a new `context` to any following middlewares.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1392,6 +1476,11 @@ A function that intercepts (reads and modifies) the `Response` of a `Request` or
 
 This is an optional argument of `onRequest()`, and may provide the required `Response` returned by the middleware.
 
+<Since v="4.13.0" />
+
+**Type:** `(rewritePayload?: string | URL | Request) => Promise<Response>`
+
+It accepts an optional URL path parameter in the form of a string, `URL`, or `Request` to trigger a [rewrite](/en/guides/routing/#rewrites) of the current request.
```
sarah11918, line 1483:
```suggestion
<p><Since v="4.13.0" /></p>

**Type:** `(rewritePayload?: string | URL | Request) => Promise<Response>`

It accepts an optional URL path parameter in the form of a string, `URL`, or `Request` to [rewrite](/en/guides/routing/#rewrites) the current request without retriggering a new rendering phase.
```

sarah11918, line 1483:
```suggestion
<p><Since v="4.13.0" /></p>

**Type:** `(rewritePayload?: string | URL | Request) => Promise<Response>`

It accepts an optional URL path parameter in the form of a string, `URL`, or `Request` to [rewrite](/en/guides/routing/#rewrites) the current request without retriggering a new rendering phase.
```

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,58 @@ if (!isLoggedIn(cookie)) {
 </html>
 ```
 
+## Rewrites
+
+<p><Since v="4.13.0" /></p>
+
+A rewrite allows you serve a different route without redirecting the browser to a different page. The browser will show the original address in the URL bar, but will instead display the content of the URL provided to [`Astro.rewrite()`](/en/reference/api-reference/#astrorewrite).
+
+For content that has permanently moved, or to direct your user to a different page (e.g. a user dashboard after logging in), use a [redirect](#redirects) instead.
```
sarah11918, line 305:
```suggestion
:::tip
For content that has permanently moved, or to direct your user to a different page with a new URL (e.g. a user dashboard after logging in), use a [redirect](#redirects) instead.
:::
```
Feels tippy to me!

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,60 @@ if (!isLoggedIn(cookie)) {
...
+
+A rewrite allows you serve a different route without redirecting the browser to a different page. The browser will show the original address in the URL bar, but will instead display the content of the URL provided to [`Astro.rewrite()`](/en/reference/api-reference/#astrorewrite).
+
+:::tip
+For content that has permanently moved, or to direct your user to a different page with a new URL (e.g. a user dashboard after logging in), use a [redirect](#redirects) instead.
+:::
+
+Rewrites can be useful for showing the same content at multiple paths (e.g. `/products/shoes/men/` and `/products/men/shoes`/) without needing to maintain two different source files.
+
+Rewrites are also useful for SEO purposes and user experience. They allow you to display content that otherwise would require redirecting your visitor to a different page or would take them to a 404 page. One common use of rewrites is to show the same localized content for different variants of a language.
```
sarah11918, line 311:
```suggestion
Rewrites are also useful for SEO purposes and user experience. They allow you to display content that otherwise would require redirecting your visitor to a different page or would return a 404 status. One common use of rewrites is to show the same localized content for different variants of a language.
```

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,79 @@ auth response
...
+}
+
+// Current URL is still https://example.com/blog
+
+// Second middleware function
+async function second(ctx, next) {
+  // Receives updated `ctx`
+  console.log(ctx.url.pathname) // this will log  "/"    
+  return next()
+}
```
sarah11918, line 268:
```suggestion
// Current URL is https://example.com/blog

// First middleware function
async function first(_, next) {
  console.log(context.url.pathname) // this will log "/blog"
  // Rewrite to a new route, the homepage 
  // Return updated `context` which is passed to next function
  return next("/")
}

// Current URL is still https://example.com/blog

// Second middleware function
async function second(context, next) {
  // Receives updated `context`
  console.log(context.url.pathname) // this will log  "/"    
  return next()
}
```

sarah11918 (approved):
Marking ready for 4.13! :tada:

src/content/docs/en/guides/middleware.mdx
```diff
@@ -196,6 +197,79 @@ auth response
...
+
+// Current URL is still https://example.com/blog
+
+// Second middleware function
+async function second(context, next) {
+  // Receives updated `context`
+  console.log(context.url.pathname) // this will log  "/"    
+  return next()
+}
```
sarah11918, line 268:
```suggestion
// First middleware function
async function first(_, next) {
  console.log(ctx.url.pathname) // this will log "/blog"
  // Rewrite to a new route, the homepage 
  // Return updated `ctx` which is passed to next function
  return next("/")
}

// Current URL is still https://example.com/blog

// Second middleware function
async function second(ctx, next) {
  // Receives updated `ctx`
  console.log(ctx.url.pathname) // this will log  "/"    
  return next()
}
```

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,60 @@ if (!isLoggedIn(cookie)) {
...
+Rewrites are also useful for SEO purposes and user experience. They allow you to display content that otherwise would require redirecting your visitor to a different page or would return a 404 status. One common use of rewrites is to show the same localized content for different variants of a language.
+
+The following example uses a rewrite to render the `/es/` version of a page when the `/es-CU/` (Cuban Spanish) URL path is visited. When a visitor navigates to the URL `/es-cu/articles/introduction`, Astro will render the content generated by the file `src/pages/es/articles/introduction.astro`.
+
+```astro title="src/pages/es-cu/articles/introduction.astro"
+---
+return Astro.rewrite("/es/articles/introduction")
+---
+```
+
```
sarah11918, line 320:
``````suggestion

Use context.rewrite in your endpoint files to reroute to a different page:

```js
export function GET(ctx) {
  if (!ctx.locals.allowed) {
    return ctx.rewrite("/")
  }
}
```

``````
We didn't actually show an endpoint rewrite here, so we probably should!

File: src/content/docs/en/guides/middleware.mdx [hunk 27]
```diff
@@ -196,6 +197,79 @@ auth response
...
+  // Return updated `ctx` which is passed to next function
+  return next("/")
+}
+
+// Current URL is still https://example.com/blog
+
+// Second middleware function
+async function second(ctx, next) {
+  // Receives updated `ctx`
+  console.log(ctx.url.pathname) // this will log  "/"
```
sarah11918, line 266:
```suggestion
  console.log(context.url.pathname) // this will log "/blog"
  // Rewrite to a new route, the homepage 
  // Return updated `context` which is passed to next function
  return next("/")
}

// Current URL is still https://example.com/blog

// Second middleware function
async function second(context, next) {
  // Receives updated `context`
  console.log(context.url.pathname) // this will log  "/"    
```

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,70 @@ if (!isLoggedIn(cookie)) {
...
+Use `context.rewrite()` in your endpoint files to reroute to a different page:
+
+```js title="src/pages/api.js"
+export function GET(ctx) {
+  if (!ctx.locals.allowed) {
+    return ctx.rewrite("/")
```
sarah11918, line 326:
```suggestion
export function GET(context) {
  if (!context.locals.allowed) {
    return context.rewrite("/")
```

src/content/docs/en/guides/routing.mdx
```diff
@@ -296,6 +296,70 @@ if (!isLoggedIn(cookie)) {
...
+<div>...</div>
+``` 
+
+You can also conditionally rewrite based on an HTTP response status, for example to display a certain page on your site when visiting a URL that doesn't exist:
+
+```js title="src/middleware.mjs"
+export const onRequest = async (ctx, next) => {
+  const response = await next();
+  if (response.status === 404) {
+    return ctx.rewrite("/");
```
sarah11918, line 353:
```suggestion
export const onRequest = async (context, next) => {
  const response = await next();
  if (response.status === 404) {
    return context.rewrite("/");
```

PR #8915, opened 2024-07-24
Fix search category for integration guides

delucis:
This PR marks pages under `/guides/integrations-guide/*` as part of our “Learn” category for search purposes. This includes individual integration pages like `@astrojs/mdx` etc.

The category had previously been changed to “Recipes” when integration guides were grouped with our guides to deployment, CMS etc. but after the recent sidebar shuffle, restoring these to “Learn” makes sense.

Note that this change will only take effect after merging this PR and reindexing Algolia. That happens automatically every Monday, but we can run a manual reindex if this feels high enough priority.


sarah11918 (approved):
Thank you! I'd been looking for this, but in the wrong place. :sweat_smile:

PR #8942, opened 2024-07-29
Add information about browser compatibility to Installation guide

yanthomasdev:
#### Description (required)

This PR adds information about the browsers Astro supports by default through Vite.

### DIFF
--- a/src/content/docs/en/install-and-setup.mdx
+++ b/src/content/docs/en/install-and-setup.mdx
@@ -22,6 +22,10 @@ Prefer to try Astro in your browser? Visit [astro.new](https://astro.new/) to br
 - **Text editor** - We recommend [VS Code](https://code.visualstudio.com/) with our [Official Astro extension](https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode).
 - **Terminal** - Astro is accessed through its command-line interface (CLI).
 
+## Browser compatibility
+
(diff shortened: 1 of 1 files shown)


src/content/docs/en/install-and-setup.mdx
```diff
@@ -22,6 +22,10 @@ Prefer to try Astro in your browser? Visit [astro.new](https://astro.new/) to br
 - **Text editor** - We recommend [VS Code](https://code.visualstudio.com/) with our [Official Astro extension](https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode).
 - **Terminal** - Astro is accessed through its command-line interface (CLI).
 
+## Browser compatibility
+
+Astro is built with Vite, meaning it targets browsers with modern JavaScript support by default. For reference, you can see the [list of currently supported browser versions in Vite](https://vitejs.dev/guide/build.html#browser-compatibility).
```
sarah11918, line 27:
```suggestion
Astro is built with Vite which targets browsers with modern JavaScript support by default. For a complete reference, you can see the [list of currently supported browser versions in Vite](https://vitejs.dev/guide/build.html#browser-compatibility).
```
What do you think?

yanthomasdev: Sounds good!

sarah11918 (approved):
Amazing, thank you for jumping on this so quickly, Yan!

Issue #8929, opened 2024-07-26
Inaccurate Page interface for Pagination in Routing guide


ArmandPhilippot:
### 📚 Subject area/topic

Routing/Pagination

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/routing/
https://docs.astro.build/en/reference/api-reference/#paginate

### 📋 Description of content that is out-of-date or incorrect

The Routing guide displays a [complete API reference](https://docs.astro.build/en/guides/routing/#complete-api-reference) for the `page` prop related to Pagination. This reference is no longer accurate since https://github.com/withastro/astro/pull/11176. The documentation has been updated in [API reference](https://docs.astro.build/en/reference/api-reference/#the-pagination-page-prop) (see #8751) but not in `guides/routing.mdx`.

There is another mistake I think. The default for `page.size` seems to be `10` and not `25` (see [paginate.ts#L19](https://github.com/withastro/astro/blob/02c85b541241a07db45bf9e15717e111104898e5/packages/astro/src/core/render/paginate.ts#L19) and [paginate.ts#L80](https://github.com/withastro/astro/blob/02c85b541241a07db45bf9e15717e111104898e5/packages/astro/src/core/render/paginate.ts#L80)). It seems this error comes from [`astro/@types` comment](https://github.com/withastro/astro/blob/02c85b541241a07db45bf9e15717e111104898e5/packages/astro/src/%40types/astro.ts#L2707). I created a [Stackblitz reproduction](https://stackblitz.com/edit/astro-paginate-pagesize?file=src%2Fpages%2Fpokemon%2F[page].astro&on=stackblitz) for this. I'll make an update there as well.

I could update the Routing guide but I wonder if this section is relevant. Since we already have the interface in API Reference, it seems duplicated (and prone to errors when new features are added). I think a link to the API Reference could make more sense:

```diff
-#### Complete API reference
-
-```ts
-interface Page<T = any> {
-	/** result */
-	data: T[];
-	/** metadata */
-	/** the count of the first item on the page, starting from 0 */
-	start: number;
-	/** the count of the last item on the page, starting from 0 */
-	end: number;
-	/** total number of results */
-	total: number;
-	/** the current page number, starting from 1 */
-	currentPage: number;
-	/** number of items per page (default: 25) */
-	size: number;
-	/** number of last page */
-	lastPage: number;
-	url: {
-		/** url of the current page */
-		current: string;
-		/** url of the previous page (if there is one) */
[...]

Issue #6467, opened 2024-01-19
Revisiting consistency in documenting features that were previously experimental


VoxelMC:
### 📚 Subject area/topic

Experimental Features

### 📋 Page(s) affected (or suggested, for new content)

- **View Transitions - `Since` is current, but doesn't mention experimental.**
https://docs.astro.build/en/guides/view-transitions/
- **Astro Assets (`<Image />` and `<Picture />`) - Inconsistent `Since`**
https://docs.astro.build/en/guides/images/
- **Content Collections - Inconsistent `Since**
https://docs.astro.build/en/guides/content-collections/
- **Prerender API (`export const prerender = true | false`) - No `Since`**
https://docs.astro.build/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode
- **Dev Toolbar - No `Since`**
https://docs.astro.build/en/reference/configuration-reference/#dev-toolbar-options

### 📋 Description of content that is out-of-date or incorrect

Some features that have previously been added to Astro were first introduced as `experimental`.
In doing so, the addition of these features was documented as `experimental` when first introduced.
After the features were moved from `experimental` to officially supported, the documentation became inconsistent. 

In the case of **View Transitions**, the [`transition:persist`](https://docs.astro.build/en/guides/view-transitions/#maintaining-state) directive is documented to be introduced in `astro@2.10.0`. However, at this time, this would have been `experimental`. 
- Should this be documented as `astro@3.0.0`, which was when the featured moved out of `experimental`?

In **Astro Assets**, the `<Image />` component doesn't have a "since", while the `<Picture />` tag is documented as `astro@3.3.0`. Moreover, [some documented properties](https://docs.astro.build/en/guides/images/#densities) of the `Image` component are flagged as experimental, added in `3.3.0`.
- How should docs handle documenting when something is experimental and when it moves from experimental to official support?

Essentially, the documentation doesn't account for when something *was* experimental (if applicable). Those using older versions may believe that a feature is available when it was only experimental then.

sarah11918:
These are all great observations @VoxelMC , and your gut reactions here have been spot on!

I would say that yes, we can update any "Since"'s to their *official/non-experimental* version numbers.

I would also say that anything *missing* a Since at all could have them added, yes!

I don't see anything on the images currently being marked as Experimental though? Can you point out specifically what you mean there.

[...]

PR #8943, opened 2024-07-29
fix(api-reference): add missing info, types and `<Since />`

ArmandPhilippot:
This PR is a follow-up to my discussion with @sarah11918 in #8929.

#### Description (required)

I updated `api-reference.mdx` to add missing (or to update) information like:
* `<Since />` component,
* defaults,
* types for individual properties.

For this task, I tried to keep track of my references in each commit (with URLs) so that you can verify the accuracy of the information (this is why this PR has so many commits, I found it was easier to divide the file in parts).

I left out some types that seemed too complex to understand by reading the documentation. Also, I modified a type but I'm not sure its definition is clear from reading the documentation. I'll add a review comment.

If I haven't added a `<Since />` component it's either because I haven't found when the property/component was added or because it dates from v1 or earlier.

Additionally, I noticed some missing sections:
* all the properties of `Astro.cookies` was described except `merge`, I added it
* `getDataEntryById` is missing but `getEntryBySlug` is marked as deprecated in the documentation (I couldn't find a reference about this) so I don't know if we should add it (I guess it is also deprecated)

Finally, I removed:
* `Astro.canonicalURL` because it seems it was removed in `v2.0.0` (see https://github.com/withastro/astro/pull/5707 and we are currently in `v4.12.1`)

- Related: #8929 + #6467
- Suggested label: improve-documentation

#### Documentation "coding" style related

While translating the doc in French and while updating this file, I noticed that some parts of the files was inconsistent, more exactly when we have `<Since />` and/or `**Types:**`:
* Sometimes `<Since />` OR `**Types:**` are wrapped in `<p>...</p>` tags, other times it is not the case
* When `<Since />` AND `**Types:**` are both provided, sometimes they are in a block wrapped with `<p>...</p>`, other times they are separated (often we have a `<Since />` wrapped in `<p>...</p>` tags followed by or put after `**Types:**` and separated by a new line)

While updating this file I wasn't sure what to do. In [Astro Docs Docs](https://contribute.docs.astro.build/reference/custom-components/#examples-1), we have an example but I wasn't sure if it was a guideline or just an example.

So if you have any guidance about the formatting, I'll be happy to update this PR to make the documentation more consistent (at least in this file)! 😄


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1120,14 +1189,16 @@ This means collections **cannot** store a mix of content and data formats. You m
 
 #### `schema`
 
-**Type:** `TSchema extends ZodType`
+**Type:** `TSchema extends ZodType | (context: SchemaContext) => TSchema extends ZodType`
```
ArmandPhilippot, line 1192: I wasn't sure how to update this type. Is `TSchema extends ZodType` really meaningful for someone new to Astro (or Typescript)? Reading this, it seems weird to me since the other types are usable while here, it is an indication for those who understand Typescript. Even if it's not a real type, shouldn't we just stick to `ZodType`?

```
`ZodType | (context: SchemaContext) => ZodType`
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -206,28 +214,30 @@ Astro.response.headers.set('Set-Cookie', 'a=b; Path=/;');
...
 
+**Type:** `AstroCookies`
+
 `Astro.cookies` contains utilities for reading and manipulating cookies in [server-side rendering](/en/guides/server-side-rendering/) mode.
 
 ##### `get`
 
 <p>
-**Type:** `(key: string, options?: CookieGetOptions) => AstroCookie`
+**Type:** `(key: string, options?: CookieGetOptions) => AstroCookie | undefined`
```
ArmandPhilippot, line 224: The real type is `AstroCookieGetOptions`. Was the type renamed at some point (I couldn't find a reference) or was it to shorten the type declared in the documentation? (I found other examples where the `Astro` prefix is ​​missing)

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -52,14 +52,19 @@ Markdown files have the following interface:
 export interface MarkdownInstance<T extends Record<string, any>> {
   /* Any data specified in this file's YAML frontmatter */
 	frontmatter: T;
-  /* The file path of this file */
+  /* The absolute file path of this file */
 	file: string;
   /* The rendered path of this file */
 	url: string | undefined;
   /* Astro Component that renders the contents of this file */
 	Content: AstroComponent;
```
ArmandPhilippot, line 60: I hesitated to update this type (I believe it is present elsewhere as well). `AstroComponent` does not exist, the real type is `AstroComponentFactory`. Should I update all the occurrences or has the type been simplified for documentation purposes?

sarah11918:
Thank you @ArmandPhilippot , this looks great! I'm going to ask @delucis to review this one!

src/content/docs/en/reference/api-reference.mdx
delucis: I think using the accurate name is not a bad idea. Admittedly `AstroComponentFactory` is not a type we really document, so either way it doesn’t matter **_too_** much, but might as well use the same name we use in the code base in case anyone does want to search for it.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -52,14 +52,19 @@ Markdown files have the following interface:
...
+  /** Raw Markdown file content, excluding layout HTML and YAML frontmatter */
+	rawContent(): string;
+  /** Markdown file compiled to HTML, excluding layout HTML */
+	compiledContent(): string;
```
delucis, line 64: I believe these are only available for Markdown but **not** MDX. Do you think that should be mentioned here?

Similar to in the Markdown docs: https://docs.astro.build/en/guides/markdown-content/#exported-properties

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -52,14 +52,19 @@ Markdown files have the following interface:
 export interface MarkdownInstance<T extends Record<string, any>> {
```
delucis, line 52: I’m wondering if we should mention `MarkdownInstance` in the introductory line above. Currently there are no search results for it because code samples are excluded from indexing.

For example something like

> Markdown files loaded with `Astro.glob()` return the following `MarkdownInstance` interface:

src/content/docs/en/reference/api-reference.mdx
delucis: Again, I would say let’s use the full name. Very probable something was renamed and these docs were never updated.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -242,6 +252,14 @@ Invalidates a cookie by setting the expiration date in the past (0 in Unix time)
 
 Once a cookie is "deleted" (expired), `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options available when deleting a cookie are: `domain`, `path`, `httpOnly`, `sameSite`, and `secure`.
 
+##### `merge`
+
+<p>
+**Type:** `(cookies: AstroCookies) => void`
+</p>
+
+Merges a new `AstroCookies` instance into the current instance. Any new cookies will be added to the current instance, overwriting any existing cookies with the same name.
```
delucis, line 261: I think this probably needs an “and” to separate the two clauses as it is not the _new_ cookies that overwrite, but only the ones matching existing cookies.

```suggestion
Merges a new `AstroCookies` instance into the current instance. Any new cookies will be added to the current instance and any cookies with the same name will overwrite existing values.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -397,18 +415,12 @@ if (!isLoggedIn(cookie)) {
...
 ```
 
-### `Astro.canonicalURL`
-
-:::caution[Deprecated]
-Use [`Astro.url`](#astrourl) to construct your own canonical URL. 
-:::
-
-The [canonical URL][canonical] of the current page.
-
```
delucis, line 407: A note on this: we probably kept this around because there were some links to the `#astrocanonicalurl` heading slug. That said, I don’t think it’s still needed, because as you noted it was removed more than a year ago.

If we wanted to be responsible, we _could_ add a redirect for this case to `public/_redirects`:
```
/:lang:/reference/api-reference/#astrocanonicalurl   /:lang:/reference/api-reference/#astrourl
```

src/content/docs/en/reference/api-reference.mdx
delucis: Yes, I think your suggestion makes sense!

By the way, not required, but something we’ve done in the Starlight docs is actually link type definitions when we have them, i.e. here it might be:

```suggestion
**Type:** <code>ZodType | (context: <a href="#schemacontext">SchemaContext</a>) => ZodType</code>
```

delucis (changes requested):
Wow, thank you @ArmandPhilippot for this amazingly thorough work! Love to see stuff like the default pagination values specified — really helpful 💜


> So if you have any guidance about the formatting, I'll be happy to update this PR to make the documentation more consistent (at least in this file)! 😄

The examples in Astro Docs Docs are correct! Ideally, when there are multiple short items like Type + `<Since>`, these should be grouped together rather than in separate paragraphs.

You can see this style in use in our [configuration reference](https://docs.astro.build/en/reference/configuration-reference/#redirects) which is generated automatically from code:

Reference header reading “redirects” followed by type, default, and added in information grouped together visually


> `getDataEntryById` is missing but `getEntryBySlug` is marked as deprecated in the documentation (I couldn't find a reference about this) so I don't know if we should add it (I guess it is also deprecated)

Yes, if I’m not mistaken, they are both deprecated in favour of `getEntry()`.

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: Nice idea, this can be useful to better understand the return type! I'll check the other types (I'm thinking in particular of `AstroCookie` ).

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: Indeed, they are only available for Markdown. I had hesitated to add them but since this is `MarkdownInstance` and not `MDXInstance` I figured they should appear. But a `Markdown only` comment makes perfect sense!

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: The documentation doesn't seems to link to this section anymore, but external resources might so the redirection makes sense! I'll add it.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1279,13 +1449,14 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
...
-<p><Since v="2.5.0" /></p>
+<p>
 
 **Types:**
-
 - `(collection: string, contentSlugOrDataId: string) => CollectionEntry<collection>`
 - `({ collection: string, id: string }) => CollectionEntry<collection>`
 - `({ collection: string, slug: string }) => CollectionEntry<collection>`
+<Since v="2.5.0" />
+</p>
```
ArmandPhilippot, line 1459: I'm not sure about this one. Since we have a list of types the rendering might seem odd. I tried to place `<Since />` before `Types:` but we have a new line between the two. I think the Markdown parser ignores the `<p>` tag around because of the list inside. So I'm not sure what is the best option here.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1308,12 +1479,13 @@ See the `Content Collections` guide for examples of [querying collection entries
...
 
-<p><Since v="2.5.0" /></p>
+<p>
 
 **Types:**
-
 - `(Array<{ collection: string, id: string }>) => Array<CollectionEntry<collection>>`
 - `(Array<{ collection: string, slug: string }>) => Array<CollectionEntry<collection>>`
+<Since v="2.5.0" />
+</p>
```
ArmandPhilippot, line 1488: Like above: I'm not sure about this one. Since we have a list of types the rendering might seem odd. I tried to place `<Since />` before `Types:` but we have a new line between the two. I think the Markdown parser ignores the `<p>` tag around because of the list inside. So I'm not sure what is the best option here.

src/content/docs/en/reference/api-reference.mdx
delucis: Oh boy, yes, had no idea we’d allowed someone to have function overloads like this 😅 

I _think_ it might look ok if `<Since>` appears first in this case?

getEntry docs with added in version displayed above the list of valid types

src/content/docs/en/reference/api-reference.mdx
delucis: Ah, regarding the list nesting, we might need to use HTML for the whole markup potentially? Which is a bit painful. If indeed `<ul>` is valid inside `<p>`. I think so, but let me double check.

ArmandPhilippot: @delucis https://caninclude.glitch.me/caninclude?child=ul&parent=p It seems we cannot do this. 🫤

delucis: No, `<ul>` is indeed not valid inside a `<p>`, which should only contain phrasing content. Given this, maybe it’s best to flip the order but still not squash everything together into a single paragraph so spacing is consistent.

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: I felt like the spacing was bigger when I tried... I must have tested too quickly. So yes, it seems to be the best option when using multiple types!

delucis:
Re: https://github.com/withastro/docs/commit/35d1f4b6d342ae740814af0785bf44491a68a202 the newline made sure subsequent markup is treated as Markdown IIRC. In other words the following doesn’t work:
```md
<p>
**Type:** `undefined`
</p>
```
It renders the children as HTML directly:
```html
<p>
**Type:** `undefined`
</p>
```
But the new line lets the Markdown parser take over again:
```md
<p>

**Type:** `undefined`
</p>
```
Renders:
```html
```md
<p>
<strong>Type:</strong> <code>undefined</code>
</p>
```
```

> Even if the method is deprecated, shouldn't it be added?

Yes, probably. I asked the team about the status of `getDataEntryById()` — let’s see what they say!

File: src/content/docs/en/reference/api-reference.mdx (hunk 8 above)
Comment (ArmandPhilippot) [reply src/content/docs/en/reference/api-reference.mdx:1459]: It seems I'm struggling with this one. In all my tests, `Types:` is always rendered in its own `<p>` tag. So the spacing does not match your screenshot.

For example, the following should render both in a single `<p>` tag, but no... `Types:` is stubborn 😬 and has is own `<p>` tag in the end.

```mdx
<p>

<Since v="2.5.0" />
**Types:**
</p>
```

@delucis It was a live edit for your screenshot? The best I got is:

Illustration of the too important spacing between since and types

Is there any other trick I don't know about Markdown rendering? 😅 (Thanks by the way for the new line explanation! Erika had already told me about it for the MD images inside `figure`... but I'm not used to this syntax so I had forgotten... 👀 )

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1490,18 +1696,25 @@ export function onRequest (context, next) {
 
 ### `next()`
```
delucis, line 1697: I wonder if this should be nested under `onRequest()` given it is an argument of that function?

```suggestion
#### `next()`
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1490,18 +1696,25 @@ export function onRequest (context, next) {
...
 
-A function that intercepts (reads and modifies) the `Response` of a `Request` or calls the "next" middleware in the chain and returns a `Response`. For example, this function could modify the HTML body of a response.
+<p>
 
-This is an optional argument of `onRequest()`, and may provide the required `Response` returned by the middleware.
+**Type:** `(rewritePayload?: string | URL | Request) => Promise<Response>`<br />
+<Since v="4.13.0" />
```
delucis, line 1702: This one is a bit tricky. There has _always_ been a `next()` since middleware was first added, but the extra `rewritePayload` argument you can pass to it is new since `4.13.0` not quite sure how best to handle this situation. Might make most sense to move the mention inline with the final part about the `rewritePayload` instead of including it here?

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1490,18 +1696,25 @@ export function onRequest (context, next) {
...
+</p>
 
-<p><Since v="4.13.0" /></p>
+A function that intercepts (reads and modifies) the `Response` of a `Request` or calls the "next" middleware in the chain and returns a `Response`. For example, this function could modify the HTML body of a response.
 
-**Type:** `(rewritePayload?: string | URL | Request) => Promise<Response>`
+This is an optional argument of `onRequest()`, and may provide the required `Response` returned by the middleware.
 
 It accepts an optional URL path parameter in the form of a string, `URL`, or `Request` to [rewrite](/en/guides/routing/#rewrites) the current request without retriggering a new rendering phase.
```
delucis, line 1709: i.e. something like this:

```suggestion
Since Astro v4.13.0, `next()` accepts an optional URL path parameter in the form of a string, `URL`, or `Request` to [rewrite](/en/guides/routing/#rewrites) the current request without retriggering a new rendering phase.
```

delucis (commented):
Spotted a couple more things reading through the page!

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: Ah this one is my fault! It's [currently as you describe](https://docs.astro.build/en/reference/api-reference/#next), I haven't checked the history of the argument so I've pulled the information up so it's like the other sections. I'll revert this change!

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: I didn't read your suggestion before answering to the other comment. This suggestion seems better if we want to keep consistency with the other sections. The [current `Added in:`/`Type:` in the middle of the section](https://docs.astro.build/en/reference/api-reference/#next) seems a little odd.

So should I revert my changes or we use your suggestion?

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: I checked the history, it seems there is no particular reason for `next()` being at the same level. So I think we can safely move it under `onRequest`. However, it might seem strange to have only one subtitle in the section...

src/content/docs/en/reference/api-reference.mdx
delucis: I think I like my suggestion 😄 

I agree with you that it’s weird to have the “Added in” in the middle of the paragraph — and even weirder to have the type there. It makes it kind of unclear what is being referred to.

Alternatively, the current style could maybe make more sense with a subheading?

next() API reference section with separate heading for next(requestPayload)

But I think this overcomplicates things. Keeping the type clearly up top is preferable in my opinion.

src/content/docs/en/reference/api-reference.mdx
delucis: > it might seem strange to have only one subtitle in the section...

I think it’s OK. There is a question about why `context` doesn’t also have a heading maybe?

ArmandPhilippot: I guess it was tricky since the context object is already described above in `Endpoint contexts`. However, having two subheadings (one for `context` and another for `next()`) since seems preferable to me to avoid an isolated heading (`next()`). Maybe we could use something like:

An example of context and next function as subheading of onRequest

**Note:** I didn't check if `Since` could be applied to `context`, this is just an example.

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: Hum it could be confusing for users I think. They might think that the two are unrelated. I prefer your first suggestion and I think it's better to remove `Since` here. Usually `Since` applies to the section and not the parameter, so something like:

Example of the restructured next() section

ArmandPhilippot: I just realized that this is exactly what you are suggesting when combined with the comment on `Since`. (I must be tired 😅)  So yes, I'm okay with those changes!

src/content/docs/en/reference/api-reference.mdx
delucis: I like that a lot! And I don’t _think_ we need a `<Since>` for `context` as it was always available in middleware so covered by the main middleware one (and the availability of individual properties should be documented in the endpoint docs you linked to).

(I still don’t think we should display “Added in” in the heading for `next()`, but I guess that was not the point of this screenshot.)

ArmandPhilippot:
I think the previous points are all solved!

While I was rereading the page, I noticed that the `getImage` section could be improved. Currently the [`getImage` section in API reference](https://docs.astro.build/en/reference/api-reference/#getimage) is close to a copy-paste of the [section `Generating images with getImage()` in the Images guide](https://docs.astro.build/en/guides/images/#generating-images-with-getimage).

So I added [some changes](https://github.com/withastro/docs/pull/8943/files#diff-42959260ea642887b866e8e276a4520c1f112356eb9c15b31e510132a481aed6R1295-R1340) to differentiate a little the two sections:
* The function signature is now added in API reference (I had hesitated before because of the internal function which has a slightly different signature)
* The returned properties was incomplete, so I completed them
* I thought it might be useful to have the return type here instead of an object, so I updated the code block

I wonder if a link from `Generating images with getImage()` (Images guide) to `getImage()` (API reference) might be helpful. Something like:

```
<ReadMore>Learn more about [the `getImage()` return type](/en/reference/api-reference/#getimage).</ReadMore>
```

But we have almost the same information...

delucis:
Nicely spotted! Yes, I’d even say now that once we merge this we could also clean up the “Generating images…” section to be more focused on the practical usage part with a link to reference for the full type details.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1477,7 +1693,9 @@ Middleware allows you to intercept requests and responses and inject behaviors d
 
 ### `onRequest()`
 
-A required exported function from `src/middleware.js` that will be called before rendering every page or API route. It accepts two optional arguments: [context](#contextlocals) and [next()](#next). `onRequest()` must return a `Response`: either directly, or by calling `next()`.
+**Type:** `(context: APIContext, next: MiddlewareNext) => Promise<Response> | Response | Promise<void> | void`
+
+A required exported function from `src/middleware.js` that will be called before rendering every page or API route. It accepts two optional arguments: [context](#context) and [next()](#next). `onRequest()` must return a `Response`: either directly, or by calling `next()`.
```
delucis, line 1698: Small nit regarding the initial wording here:

```suggestion
A required exported function from `src/middleware.js` that will be called before rendering every page or API route. It receives two arguments: [context](#context) and [next()](#next). `onRequest()` must return a `Response`: either directly, or by calling `next()`.
```

Astro always passes the arguments to `onRequest` so they are not “optional arguments”. Whether a user uses them or not is up to them, but they will always be present if they do need them.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1488,19 +1706,36 @@ export function onRequest (context, next) {
...
+#### `next()`
+
+<p>
+
+**Type:** `(rewritePayload?: string | URL | Request) => Promise<Response>`<br />
+</p>
 
 A function that intercepts (reads and modifies) the `Response` of a `Request` or calls the "next" middleware in the chain and returns a `Response`. For example, this function could modify the HTML body of a response.
 
 This is an optional argument of `onRequest()`, and may provide the required `Response` returned by the middleware.
```
delucis, line 1729: Not changed in this PR, but taking advantage to suggest some changes as I don’t think this is particularly clear or accurate. This proposal also aligns  the phrasing a bit with the new `context` docs.

```suggestion
The second argument of `onRequest()` is a function that calls all the subsequent middleware in the chain and returns a `Response`. For example, other middleware could modify the HTML body of a response and awaiting the result of `next()` would allow your middleware to respond to those changes.
```

sarah11918:
Looking great, amazing work here! I trust Chris to make the final approval here, and will only say that I agree on scoping *this* PR to the api-reference page only! If you start looking on other pages, I'm sure you'll find more improvements you want to make, so let's make life easier on the translators and tackle these kinds of improvements one page at a time! :smile:

delucis (commented):
Woo, I think we’re pretty much there — amazing work @ArmandPhilippot! I left a couple of copy suggestions not strictly related to your changes, but thought we might as well get them in while we’re in this file.

sarah11918:
Also pointing out that there's a draft API reference for Actions in #8980 , so feel free to hop into that PR and get that API reference entry up to speed from the beginning!

ArmandPhilippot:
Thanks @delucis, the `onRequest` section seems clearer with your two suggestions! (and, while I'm at it, thanks for all your suggestions  ✨   )

I think we've covered this file. There are some missing types but adding them might be more harmful than helpful since they rely on Typescript generics:
* `Astro.glob()` and its overloads
* `props` and `params` in Astro global, context and `getStaticPath`

So unless you see any other improvements, I think we're good this time! 🎉

delucis (approved):
Agreed! Ready to merge — thanks again for the thorough and persistent work here @ArmandPhilippot 🙌

PR #8941, opened 2024-07-29
fix(guides/routing): update Page interface and add link to API reference

ArmandPhilippot:
#### Description (required)

* Update `page.size` default (see https://github.com/withastro/astro/pull/11561)
* Add `page.url.first` and `page.url.last` to the Page interface (see https://github.com/withastro/astro/pull/11176)
* Add a `ReadMore` link to the API reference page

#### Related issues & labels (optional)

- Closes #8929
- Suggested label: improve-documentation

PR #8922, opened 2024-07-25
i18n(fr): Add `reference/api-reference.mdx` ft. @ArmandPhilippot

thomasbnt:
#### Description (required)
Add `reference/api-reference.mdx` with @ArmandPhilippot for the second part. 

> [!NOTE]
> ~~Do not merge now, miss the second part from Armand.~~

#### Related issues & labels (optional)

- Suggested label: i18n

PR #8944, opened 2024-07-29
i18n(fr): Update `guides/routing.mdx` from #8941 and add `reference/api-reference.mdx` 

thomasbnt:
#### Description (required)
Update `guides/routing.mdx` from #8941

The link at 429 is good, see https://deploy-preview-8922--astro-docs-2.netlify.app/fr/reference/api-reference/#le-prop-page-de-pagination from https://github.com/withastro/docs/pull/8922

Add `reference/api-reference.mdx` with @ArmandPhilippot for the second part. 

- Suggested label: i18n


src/content/docs/fr/guides/routing.mdx
```diff
@@ -407,21 +407,27 @@ interface Page<T = any> {
...
 		next: string | undefined;
+		/** url de la première page (si la page actuelle n'est pas la première page) */
+		first: string | undefined;
+		/** url de la page suivante (si la page actuelle n'est pas la dernière) */
+		last: string | undefined;
 	};
 }
 ```
 
+<ReadMore>En savoir plus sur [le prop `page` de pagination](/fr/reference/api-reference/#le-prop-page-de-pagination).</ReadMore>
```
ArmandPhilippot, line 429: I haven't noticed it in `api-reference.mdx` but "*le prop*" sounds weird. Why not "*la prop*" instead of "*la propriété*" if we want to keep the abbreviation but this should be **la** and not **le** (in this PR and in #8922).

So, it should be:
```suggestion
<ReadMore>En savoir plus sur [la prop `page` de pagination](/fr/reference/api-reference/#la-prop-page-de-pagination).</ReadMore>
```

However, this PR might be merged before `api-reference.mdx` since this is a small one and #8922 is probably waiting for another reviewer. So I'm not sure we should use the translated anchor here. What do you think?

thomasbnt: > I haven't noticed it in `api-reference.mdx` but "_le prop_" sounds weird. Why not "_la prop_" instead of "_la propriété_" if we want to keep the abbreviation but this should be **la** and not **le** (in this PR and in #8922).
> 
> So, it should be:
> 
> However, this PR might be merged before `api-reference.mdx` since this is a small one and #8922 is probably waiting for another reviewer. So I'm not sure we should use the translated anchor here. What do you think?

yes, when I written this sentence, seems a little weird with `props`

thomasbnt: > However, this PR might be merged before api-reference.mdx since this is a small one and https://github.com/withastro/docs/pull/8922 is probably waiting for another reviewer. So I'm not sure we should use the translated anchor here. What do you think?
We can merge this one before #8922, but It won't interfere with the rest. It'll save you a second PR to change the link.

src/content/docs/fr/guides/routing.mdx
```diff
@@ -407,21 +407,27 @@ interface Page<T = any> {
...
 		next: string | undefined;
+		/** url de la première page (si la page actuelle n'est pas la première page) */
+		first: string | undefined;
+		/** url de la page suivante (si la page actuelle n'est pas la dernière) */
+		last: string | undefined;
 	};
 }
 ```
 
+<ReadMore>En savoir plus sur [la prop `page` de pagination](/fr/reference/api-reference/#la-prop-page-de-pagination).</ReadMore>
```
ArmandPhilippot, line 429: Since you choose `la propriété` instead of `la prop` in #8922 (see https://deploy-preview-8922--astro-docs-2.netlify.app/fr/reference/api-reference/#la-propri%C3%A9t%C3%A9-page-de-pagination) we need this update:
```suggestion
<ReadMore>En savoir plus sur [la prop `page` de pagination](/fr/reference/api-reference/#la-propriété-page-de-pagination).</ReadMore>
```

ArmandPhilippot (changes requested):
I'm ok to leave the translated anchor, however it needs an update because of the change in #8922

thomasbnt:
> I'm ok to leave the translated anchor, however it needs an update because of the change in #8922

yes, good!

ArmandPhilippot (approved):
LGTM, thanks @thomasbnt!

> [!note]
> The broken link will no longer be broken when #8922 is merged.

dreyfus92 (approved):
Thanks everybody! LGTM ✅

yanthomasdev (commented):
LGTM! Could you actually move the changes here to #8922, so we can merge without link issues? It's a bit confusing since all the PRs open depend on one another, but I think getting these two together would unblock that.

yanthomasdev (commented):
LGTM! Could you actually move the changes here to #8922, so we can merge without link issues? It's a bit confusing since all the PRs open depend on one another, but I think getting these two together would unblock that.

thomasbnt:
> LGTM! Could you actually move the changes here to #8922, so we can merge without link issues? It's a bit confusing since all the PRs open depend on one another, but I think getting these two together would unblock that.

[Merged this PR](https://github.com/withastro/docs/pull/8922) into this one.

yanthomasdev (approved):
LGTM! Good job peeps!

Issue #8855, opened 2024-07-18
Document passing information from a component to a function slot


Fryuni:
### 📚 Subject area/topic

Slots

### 📋 Suggested page

https://docs.astro.build/en/basics/astro-components/#slots

### 📋 General description or bullet points (if proposing new content)

- Slots can be a function. That allows information to be passed from component A to its parent, so the parent can decide which children to pass down to A.
- This pattern is shown as an example in the reference here: https://docs.astro.build/en/reference/api-reference/#astroslotsrender

### 🖥️ Reproduction of code samples in StackBlitz

https://stackblitz.com/edit/github-oqp5kf?file=src%2Fcomponents%2FMyList.astro,src%2Fpages%2Findex.astro

Soviut:
I originally asked about this because I use Vue's "scoped slots" a lot. https://vuejs.org/guide/components/slots.html#scoped-slots

It would be great if this pattern was given its own sub-section and named "Scoped slots". While the approach from Vue is very different, it would at least convey the intent.

Soviut:
For context, here's what I was demonstrating with Vue in the Discord

in the long term what i'm hoping for is this (note: Vue syntax)

**MyList.vue** (component)
```vue
<script setup lang="ts">
defineProps<{
  items: string[]
}>()
</script>

<template>
  <ul>
    <li v-for="(item, index) in items">
      <slot :item="item" :index="index" :total="items.length" />
    </li>
  </ul>
</template>
```
**Index.vue** (usage)
```vue
<MyList :items="items" v-slot="{ item, index, total }">
  <div>
    {{item}}
    <span>( {{index}} / {{total}} )</span>
  </div>
</MyList>
```
Note how the v-slot yields not only the `item` but also additional context like the `index` and `total` from within the component.

Here is a slightly modified Stackblitz that @Fryuni sent to me that does the same thing in Astro
https://stackblitz.com/edit/github-oqp5kf-ozddjd?file=src%2Fcomponents%2FMyList.astro,src%2Fpages%2Findex.astro

```astro
---
const { items } = Astro.props;
---
<ul>
  {items.map((item, index, list) => (
    <li set:html={Astro.slots.render('default', [item, index, list])} />
  ))}
</ul>
```

```astro
<MyList items={numbers}>
  {(item, index, list) => (
    <>
      Item {index}:
      <ul>
        <li>Value: {item}</li>
        <li>Last occurrence: {list.lastIndexOf(item)}</li>
      </ul>
    </>
[...]

PR #8946, opened 2024-07-29
Add reference for named slot functions

BryceRussell:
#### Description (required)

This PR partially addresses issue https://github.com/withastro/docs/issues/8855 by adding an example for named slot functions to the reference page for `Astro.slots.render()`

- Partially addresses: https://github.com/withastro/docs/issues/8855

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -539,8 +539,6 @@ import Shout from "../components/Shout.astro";
 
 Callback functions can be passed to named slots inside a wrapping HTML element tag with a `slot` attribute. This element is only used to transfer the callback to a named slot and will not be rendered onto the page.
 
-A callback 
-
 ```astro
 <Shout message="slots!">
   <fragment slot="message">
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -549,7 +549,7 @@ A callback
 </Shout>
 ```
 
-This tag cannot be named `Fragment`, `slot` or include a dash `-`. The tag is only used to transfer the callback to a named slot, it does not get rendered onto the page.
+Use a standard HTML element for the wrapping tag, or any lower case tag (e.g. `<fragment>` instead of `<Fragment />`) that will not be interpreted as a component.  Do not use the HTML `<slot>` element as this will be interpreted as an Astro slot.
 
 ### `Astro.self`
 
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -537,6 +537,18 @@ import Shout from "../components/Shout.astro";
 <!-- renders as <div>SLOTS!</div> -->
 ```
 
+Callback functions can be passed to named slots using a HTML tag with a `slot` attribute:
+
+```astro
+<Shout message="slots!">
+  <fragment slot="message">
+    {(message) => <div>{message}</div>}
+  </fragment>
+</Shout>
+```
+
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -537,6 +537,18 @@ import Shout from "../components/Shout.astro";
...
+
+```astro
+<Shout message="slots!">
+  <fragment slot="message">
+    {(message) => <div>{message}</div>}
+  </fragment>
+</Shout>
+```
+
+This tag cannot be named `Fragment`, `slot` or include a dash `-`. The tag is only used to transfer the callback to a named slot, it does not get rendered onto the page.
```
sarah11918, line 550:
```suggestion
Use a standard HTML element for the wrapping tag and not a component such as `<Fragment />`.  However, you cannot use the HTML `<slot>` element as this will be interpreted as an Astro slot. Lastly, the element name may not include a dash `-` to avoid conflicts with data attributes.
```

I'm wondering whether we can get away with just saying "Use a standard HTML element" which solves most of the naming restrictions. :smile:  Like, if they do choose a standard HTML element (that's not slot), we don't have to worry about them using Fragment (because it's a component, not an HTML element). I added a reason (not sure if accurate?) about why no dashes, but maybe we don't even need that sentence *if* we assume that a standard HTML element doesn't have dashes in the first place?

sarah11918 (commented):
This looks great, @BryceRussell ! Thank you so much for tackling this! See my comment below re: can we make advice easier if we suggest that the tag should be a standard HTML element.

I get that the naming of the tag is arbitrary, but I *wonder* whether presenting it as an element avoids some of the naming restrictions we don't want them to get tripped up by?

src/content/docs/en/reference/api-reference.mdx
BryceRussell: Ya, I like that, that makes sense. I took another look at the dash `-` restriction and it looks like a bug that can be fixed upstream, so we can just remove that restriction to simplify things:

```suggestion
Always use a standard HTML element for the wrapping tag and not a `<Fragment />` or component. However, you cannot use the HTML element `<slot>` as this will be interpreted as an Astro slot. 
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -537,6 +537,20 @@ import Shout from "../components/Shout.astro";
...
```
sarah11918, line 552:
```suggestion
Use a standard HTML element for the wrapping tag, or any lower case tag (e.g. `<fragment>` instead of `<Fragment />`) that will not be interpreted as a component.  Do not use the HTML `<slot>` element as this will be interpreted as an Astro slot.
```
I realized that I don't think `<fragment>` *is* a standard HTML element :sweat_smile: 

So, we should *either* do something like this, OR we should change the code sample to use a standard element!

BryceRussell: That example LGTM

I typically use a non-standard element like `<fragment>` for these examples to avoid confusion about the tag not getting rendered onto the page and looking very similar to other patterns inside `.astro` files like a `.map()`

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -537,6 +537,20 @@ import Shout from "../components/Shout.astro";
 <!-- renders as <div>SLOTS!</div> -->
 ```
 
+Callback functions can be passed to named slots inside a wrapping HTML element tag with a `slot` attribute. This element is only used to transfer the callback to a named slot and will not be rendered onto the page.
+
+A callback 
+
```
sarah11918, line 543:
```suggestion
```

sarah11918 (approved):
Looks great, thanks for this @BryceRussell ! Merge whenever you're happy with it!

PR #8952, opened 2024-07-30
fix(guides): update misleading copy in Netlify deploy guide

serhalp:
#### Description (required)

The wording "deploy your project using Netlify's Edge Functions" implies that it enables _rendering_ via Netlify Edge Functions, but this isn't true.

I removed that wording and replaced it with an attempted description of the actual behaviour.

See [inline comments describing the option](https://github.com/withastro/adapters/blob/6dd65a04c2afea941da34a1a1ce7b991f0768615/packages/netlify/src/index.ts#L167-L177).

- Suggested label: 3rd-party guide, improve documentation


src/content/docs/en/guides/deploy/netlify.mdx
```diff
@@ -21,7 +21,7 @@ Your Astro project is a static site by default. You don’t need any extra confi
 
 ### Adapter for SSR
 
-To enable SSR in your Astro project and deploy on Netlify, including using Netlify's edge functions:
```
serhalp, line 24: I also removed this, because it seemed too vague and possibly misleading

src/content/docs/en/guides/deploy/netlify.mdx
```diff
@@ -21,7 +21,7 @@ Your Astro project is a static site by default. You don’t need any extra confi
...
+To enable SSR in your Astro project and deploy on Netlify:
 
 Add [the Netlify adapter](/en/guides/integrations-guide/netlify/) to enable SSR in your Astro project with the following `astro add` command. This will install the adapter and make the appropriate changes to your `astro.config.mjs` file in one step.
```
delucis, line 26: I actually think we can probably collapse these two paragraphs as they basically repeat each other:

```suggestion
Add [the Netlify adapter](/en/guides/integrations-guide/netlify/) to enable SSR in your Astro project and deploy to Netlify with the following `astro add` command. This will install the adapter and make the appropriate changes to your `astro.config.mjs` file in one step.
```

src/content/docs/en/guides/deploy/netlify.mdx
```diff
@@ -50,7 +50,7 @@ If you prefer to install the adapter manually instead, complete the following tw
     });
     ```
 
-    You can also deploy your project using Netlify's Edge Functions by adding `edgeMiddleware: true` to the Netlify adapter's configuration:
+    You can also deploy your project's Astro Middleware using Netlify's Edge Functions by adding `edgeMiddleware: true` to the Netlify adapter's configuration:
```
delucis, line 53: Nit, but we usually don’t uppercase Astro features:

```suggestion
    You can also deploy your project's Astro middleware using Netlify's Edge Functions by adding `edgeMiddleware: true` to the Netlify adapter's configuration:
```

src/content/docs/en/guides/deploy/netlify.mdx
```diff
@@ -64,6 +64,8 @@ If you prefer to install the adapter manually instead, complete the following tw
         }),
       });
     ```
+
+    If you use Middleware to implement authentication, redirects, or similar things, you may need to enable this flag. Otherwise, your Middleware will not run at runtime for pre-rendered pages.
```
delucis, line 68: Same here re: capitalisation, but also suggesting flipping the wording for a more positive “do X to achieve Y” pattern rather than “do X to avoid Z”.

```suggestion
    
    Set `edgeMiddleware: true` if you want to run middleware for pre-rendered pages. This allows you to use middleware to implement authentication, redirects, or similar things, while still using static HTML output.
```

delucis (commented):
Thanks @serhalp — super helpful changes! I left a couple of small suggestions.

src/content/docs/en/guides/deploy/netlify.mdx
sarah11918: Great call, Chris!

For your consideration, I'll also throw this slight tweak out there to frontload the "to do X" part:

> To run middleware for pre-rendered pages, set `edgeMiddleware: true`. ...

delucis: Ah, great call, yes. I honestly initially aimed for that and didn’t quite land on a formulation I liked, but this works great. Will update my suggestion to match!

delucis (approved):
Thanks again for the suggestions @serhalp! Hope you don’t mind me tweaking the copy myself to get this merged.

PR #8955, opened 2024-07-30
Add guidance on how to use LibSQL with Astro DB

Fryuni:
#### Description

- Updates the Astro DB guide to include using LibSQL as the remote database.

- Docs for https://github.com/withastro/astro/pull/11385


src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,49 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+
+<p><Since pkg="@astrojs/db" v="0.14.0" /></p>
+
+Alternatively to Astro Studio, Astro DB can connect to any LibSQL server. You can use an offering from any platform that exposes the Hrana protocol of the server or host it yourself.
+
+To connecto Astro DB to a LibSQL instead of Studio, set the following environment variables:
+- `ASTRO_DB_REMOTE_URL`: the connection URL to your LibSQL server
+- `ASTRO_DB_APP_TOKEN`: the auth token to your LibSQL server
+
+The environment variables have to be set during `astro build`, `astro db push --remote` and during runtime if your project use Astro DB on SSR pages.
```
sarah11918, line 668: Can we explain or give an example of what someone needs to do to make sure these are set at the correct time? It's one thing to say they "have to be set during..", but it's helpful to show what they will need to do to make that happen. If simply setting the variables were enough, then this sentence wouldn't be necessary, right? So what else do they have to make sure they do?

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,49 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+
+Passing a separate connection URL on this search parameter to turn the DB into an embedded replica of another of another database. This should only be used with the schemes `file:` and `memory:`. The parameter must be URL encoded.
+
+Embedded replicas are a feature of LibSQL clients that enables a full synchronized copy of your database on a local file or in memory for ultra-fast reads. Writes are sent to a remote database defined on the `syncUrl` and synchronized with the local copy.
+
+For example, to have an in-memory embedded replica of a database on `libsql://your.server.io`, you can set the connection URL to `memory:?syncUrl=libsql%3A%2F%2Fyour.server.io`.
+
+### `syncInterval`
+
+Interval between embedded replicas synchronizations in seconds. By default it only synchronizes on startup and after writes.
```
sarah11918, line 699: Could also use an example people can see

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,49 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+LibSQL has native support for encrypted databases. Passing this search parameter will enable encryption using the given key:
+
+- `file:path/to/file.db?encryptionKey=your-encryption-key`
+- `libsql://your.server.io?encryptionKey=your-encryption-key`
+
+### `syncUrl`
+
+Passing a separate connection URL on this search parameter to turn the DB into an embedded replica of another of another database. This should only be used with the schemes `file:` and `memory:`. The parameter must be URL encoded.
+
+Embedded replicas are a feature of LibSQL clients that enables a full synchronized copy of your database on a local file or in memory for ultra-fast reads. Writes are sent to a remote database defined on the `syncUrl` and synchronized with the local copy.
```
sarah11918, line 693:
```suggestion
Embedded replicas are a feature of LibSQL clients that enables a full synchronized copy of your database on a local file or in memory for ultra-fast reads. Writes are sent to a remote database defined on the `syncUrl` and synchronized with the local copy.

Use this property to pass a separate connection URL to turn the DB into an embedded replica of another of another database. This should only be used with the schemes `file:` and `memory:`. The parameter must be URL encoded.
```
These read better when the order is swapped!

sarah11918 (commented):
Thanks @Fryuni , this is looking great! I can spot a few areas where I'd like to see more examples or have more descriptive help than "use", but this is pretty close to where I think it will end up! :raised_hands:

src/content/docs/en/guides/astro-db.mdx
ArmandPhilippot: I noticed a repeat that seems wrong to me: `replica of another of another database` > `replica of another database` (first paragraph in the original, or second paragraph in Sarah's suggestion).

src/content/docs/en/guides/astro-db.mdx
Fryuni: Setting the environment variables is enough, but they have to do it in all places being used. So, for example, when deploying to Vercel you need the environment variable set locally when you run `astro db push --remote` to apply the schemas _and_ on Vercel to connect to the DB.

Using Studio, the token only needs to be set on Vercel since the local run of `astro db push --remote` will use the user token.

sarah11918: So, would this be added to the command as a flag when run? Or just added in the appropriate places (one and done, like in a `.env` file or in a dashboard somewhere? I think this is what I'm getting at. 

"Set when you run" could mean "pre-set, and available somewhere" or something you need to provide each time.

Fryuni: I made the examples show the `.env` file so people put it there. There is multiple ways to do it along with security considerations, but that will depend on how their project and their connection to LibSQL is structured. I don't think there is much point going into details on Astro Docs vs following the docs of the hosting they are using

sarah11918 (commented):
This looks great, @Fryuni ! Just a tiny bit of light editing on the top.

See what you think re: my comment about the commands, and at some point, I do wonder whether some of this belongs on the `@astrojs/db` integration reference page? (like the sync stuff?) and we just link to it from here? But I think this is fine for now, unless you see a way you'd prefer to offload some documentation to the other page.

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,68 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
 
 </ReadMore>
 
+## LibSQL
+
+<p><Since pkg="@astrojs/db" v="0.14.0" /></p>
+
+Astro DB can connect to any LibSQL server from any platform that exposes the Hrana protocol of the server, or host it yourself.
```
sarah11918, line 662:
```suggestion
Astro DB can connect to any LibSQL server from any platform that exposes the Hrana protocol of the server, or can be self-hosted.
```

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,68 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+
+To connecto Astro DB to a LibSQL instead of Studio, set the following environment variables:
```
sarah11918, line 664:
```suggestion
To connect Astro DB to a LibSQL instance, set the following environment variables:
```
What's the noun here, is it "instance"?

sarah11918:
Thanks @Fryuni ! Literally two tiny nits to get just right grammatically, and we're done! This is amazing, thank you!

src/content/docs/en/guides/astro-db.mdx
Fryuni: I think "database" would be the most appropriate. At first I thought of server but it can be a local file or even in memory.

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,68 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+For example, to have a local file work as an embedded replica to an encrypted LibSQL server, you can set the following environment variables:
+
+```env title=".env"
+ASTRO_DB_REMOTE_URL=file://local-copy.db?encryptionKey=your-encryption-key&syncInterval=60&syncUrl=libsql%3A%2F%2Fyour.server.io
+ASTRO_DB_APP_TOKEN=token-to-your-remote-url
+```
+
+### URL scheme and host
+
+LibSQL supports both HTTP and WebSockets as the transport protocol for for a remote server. It also supports using a local file or an in-memory DB. Those can be configured using the following URL schemes in the connection URL:
```
yanthomasdev, line 681:
```suggestion
LibSQL supports both HTTP and WebSockets as the transport protocol for a remote server. It also supports using a local file or an in-memory DB. Those can be configured using the following URL schemes in the connection URL:
```

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,68 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+
+- `memory:` will use an in-memory DB. Host must be empty in this case.
```
yanthomasdev, line 683:
```suggestion
- `memory:` will use an in-memory DB. The host must be empty in this case.
```

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,68 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+
+```env title=".env"
+ASTRO_DB_REMOTE_URL=memory:?syncUrl=libsql%3A%2F%2Fyour.server.io
+```
+
+### `syncInterval`
+
+Interval between embedded replicas synchronizations in seconds. By default it only synchronizes on startup and after writes.
+
+This property is only used when `syncUrl` is also set. For example, to set an in-memory empbedded replica to synchronize every minute set the following environment variable:
```
yanthomasdev, line 713:
```suggestion
This property is only used when `syncUrl` is also set. For example, to set an in-memory embedded replica to synchronize every minute set the following environment variable:
```

yanthomasdev (commented):
Amazing work @Fryuni, I have a few suggestions. After applying those, LGTM!

sarah11918 (approved):
We did it! 🥳 This is gonna be so popular, @Fryuni . Can't wait!

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,68 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+
+<p><Since pkg="@astrojs/db" v="0.14.0" /></p>
+
+Astro DB can connect to any LibSQL server from any platform that exposes the Hrana protocol of the server, or can be self-hosted.
+
+To connect Astro DB to a LibSQL database, set the following environment variables:
+- `ASTRO_DB_REMOTE_URL`: the connection URL to your LibSQL server
+- `ASTRO_DB_APP_TOKEN`: the auth token to your LibSQL server
+
+The [commands for deploying and pushing changes](#deploy-with-a-studio-connection) to your database are the same when using LibSQL as when connecting to a Studio hosted database. However, both of your environment variables need to be set locally when running commands with the `--remote` option like `astro build` and `astro db push`.
```
lorenzolewis, line 668:
```suggestion
The [commands for deploying and pushing changes](#deploy-with-a-studio-connection) to your database are the same when using LibSQL as when connecting to an Astro Studio hosted database. However, both of your environment variables need to be set locally when running commands with the `--remote` option like `astro build` and `astro db push`.
```

Possibly make this a bit more explicit for those who may be linking to this specific section.

lorenzolewis (commented):
Optional nit

src/content/docs/en/guides/astro-db.mdx
sarah11918: Love it, thanks!

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,68 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
 
 </ReadMore>
 
+## LibSQL
+
+<p><Since pkg="@astrojs/db" v="0.14.0" /></p>
+
+Astro DB can connect to any LibSQL server from any platform that exposes the Hrana protocol of the server, or can be self-hosted.
```
notrab, line 662:
```suggestion
Astro DB can connect to any libSQL server from any platform that exposes the libSQL remote protocol of the server, or can be self-hosted.
```

We're moving away from calling it "Hrana", so not using it here would be a huge help 🫶

src/content/docs/en/guides/astro-db.mdx
```diff
@@ -654,6 +655,68 @@ When you're ready to deploy, see our [Deploy with a Studio Connection guide](#de
...
+
+To connect Astro DB to a LibSQL database, set the following environment variables:
```
notrab, line 664:
```suggestion
To connect Astro DB to a libSQL database, set the following environment variables:
```

We tend to refer to `libSQL` with a lowercase `l`. I hope this isn't too much trouble to find and replace all 🫶

src/content/docs/en/guides/astro-db.mdx
lorenzolewis: As a reader, this also makes it a lot easier to understand what it actually is in this context as well.

Fryuni: The name everywhere I see is LibSQL with capital L, is that changing too?

Fryuni: I looked again, and there seems to be a mix. I'll normalize our references to be `libSQL` but as a heads up @notrab , on Turso docs and the repos the links and titles are mostly using "LibSQL" while the content on the pages are using "libSQL".

sarah11918:
Thanks for jumping in here @notrab ! I'm glad this was caught in lots of time (a day... but for docs that's an *eternity* :joy: ) Will be released tomorrow! :raised_hands:

yanthomasdev (approved):
LGTM!

sarah11918:
!coauthor

PR #8981, opened 2024-08-01
Clarify framework dependency for JSX files

### LINKED ISSUE #8939: Request: Clarify "out-of-the-box" JSX/TSX support
### 📋 Explain your issue

I'm confused trying to understand the extent to which Astro, without the React (or another framework) integration, supports JSX/TSX syntax.

Support for `.jsx` and `.tsx` files is currently mentioned in the docs [here](https://docs.astro.build/en/guides/imports/#jsx--tsx) and [here](https://docs.astro.build/en/guides/imports/#supported-file-types) (both on the `guides/imports` page)

It is noted: 
> While Astro understands JSX syntax out-of-the-box, you will need to include a framework integration to properly render frameworks like React, Preact and Solid.

Could the extent to which the syntax/filetype is supported be explained a little more explicitly? I can see types for `JSXTransformConfig` and `JSXTransformFn` in the Astro source, but writing even the simplest JSX, such as `const testJsx = () => <p>Hello, world!</p>` leads to a build error like `Cannot find module 'react/jsx-dev-runtime' imported from <testJsx.jsx>`

In the docs-given example:
```JSX
import { MyComponent } from './MyComponent.jsx';
```
What JSX could be declared/implemented/returned by `MyComponent` that would not require a framework integration to be installed? 

It's late, so I appreciate I may not be firing on all cylinders, but I do wonder if this clarification may be a beneficial amendment to the docs. I appreciate your time; thanks!

kjkent:
#### Description (required)
Clarifies that a UI framework is required to import and render JSX/TSX files. My first PR here! I'm happy to make any amendments as seen fit :)

I just joined the Astro Discord, username is `footnuke` 

- Closes #8939
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/imports.mdx
+++ b/src/content/docs/en/guides/imports.mdx
@@ -15,10 +15,9 @@ The following file types are supported out-of-the-box by Astro:
 - Astro Components (`.astro`)
 - Markdown (`.md`, `.markdown`, etc.)
 - JavaScript (`.js`, `.mjs`)
-- TypeScript (`.ts`, `.tsx`)
+- TypeScript (`.ts`)
 - NPM Packages
 - JSON (`.json`)
-- JSX (`.jsx`, `.tsx`)
 - CSS (`.css`)
 - CSS Modules (`.module.css`)
 - Images & Assets (`.svg`, `.jpg`, `.png`, etc.)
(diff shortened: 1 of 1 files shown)


at-the-vr (approved):
LGTM 👍 Thank you for going all the way in with this one

delucis (approved):
Thank you @kjkent!

kjkent:
Thanks both!

"It ain't much, but it's honest work" :')

PR #8997, opened 2024-08-02
Overhaul `@astrojs/sitemap` docs

delucis:
This PR updates the docs for the `@astrojs/sitemap` integration.

My main aim was to be able to link directly to the information about linking to the sitemap in `robots.txt` for https://github.com/withastro/starlight/pull/2083 (to avoid duplicating these docs in the Starlight docs).

But while there I realised this page is fairly dusty and could use a small refresh.

Reorganised, reworded, and re-prioritised things.

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -82,66 +82,29 @@ export default defineConfig({
 
 ## Usage
 
-`@astrojs/sitemap` requires a deployment / site URL for generation. Add your site's URL under your `astro.config.*` using the `site` property. This must begin with `http:` or `https:`.
+`@astrojs/sitemap` needs to know your site’s deployed URL to generate a sitemap.
 
-```js title="astro.config.mjs" "'https://stargazers.club'"
+Add your site's URL as the [`site`](/en/reference/configuration-reference/#site) option in `astro.config.mjs`. This must begin with `http://` or `https://`.
+
+```js title="astro.config.mjs" {5}
 import { defineConfig } from 'astro/config';
 import sitemap from '@astrojs/sitemap';
 
 export default defineConfig({
-  // ...
   site: 'https://stargazers.club',
   integrations: [sitemap()],
+  // ...
 });
 ```
 
-Note that unlike other configuration options, `site` is set in the root `defineConfig` object, rather than inside the `sitemap()` call.
+With the sitemap integration configured, `sitemap-index.xml` and `sitemap-0.xml` files will be added to your output directory when building your site.
 
-Now, [build your site for production](/en/reference/cli-reference/#astro-build) via the `astro build` command. You will find both `sitemap-index.xml` and `sitemap-0.xml` in the `dist/` folder (or your custom [output directory](/en/reference/configuration-reference/#outdir) if set).
+`sitemap-index.xml` links to all the numbered sitemap files.
+`sitemap-0.xml` lists the pages on your site.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -82,35 +82,70 @@ export default defineConfig({
...
+  <url>
+    <loc>https://stargazers.club/second-page/</loc>
+  </url>
+</urlset>
+```
+</details>
+
+### Sitemap discovery
+
+You can make it easier for crawlers to find your sitemap with links in your site's `<head>` and `robots.txt` file.
```
sarah11918, line 133:
```suggestion
You can make it easier for crawlers to find your sitemap with links in your site's `<head>` or `robots.txt` file.
```
Just checking, would someone do *both* or only one of the following methods? If they're not supposed to add to both places (a conflict?) then I'd not use "and" this way. Could be swapping to "or" or could be extra context like:

> You can make it easier for crawlers to find your sitemap by including a link in either...

> To help crawlers discover your sitemap, you can add a link in one of the following locations...

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -119,54 +154,28 @@ Allow: /
 Sitemap: https://<YOUR SITE>/sitemap-index.xml
 ```
 
-To generate `robots.txt` dynamically, add a file named `robots.txt.ts` with the following code:
+You can also generate `robots.txt` dynamically if you want to reuse the `site` value from `astro.config.mjs`.
+To do so, create a `src/pages/robots.txt.ts` file with the following code:
```
sarah11918, line 158:
```suggestion
To generate a `robots.txt` file dynamically using the `site` value from `astro.config.mjs`, create the file `src/pages/robots.txt.ts` and add the following code:
```

sarah11918 (commented):
Wow, this looks so much better @delucis ! Thank you! Left two small comments!

src/content/docs/en/guides/integrations-guide/sitemap.mdx
delucis: Yes, doing both is not uncommon. Do you still think “or” is better in that case?

src/content/docs/en/guides/integrations-guide/sitemap.mdx
delucis: This removes the explanation for _why_ you might do this and almost makes it sound like you should use this dynamic approach as well as the static approach — which you shouldn’t. Maybe that should be clearer?

```suggestion
If you want to reuse the `site` value from `astro.config.mjs`, you can also generate `robots.txt` dynamically.
Instead of using a static file in the `public/` directory, create a `src/pages/robots.txt.ts` file and add the following code:
```

src/content/docs/en/guides/integrations-guide/sitemap.mdx
sarah11918: Ah, nope! If doing both is common, then it's fine as is!

src/content/docs/en/guides/integrations-guide/sitemap.mdx
sarah11918: Yeah, I was worried about whether the why was clear enough in my suggestion. I didn't think it made it sound like it would do both, but if my suggestion also introduces that, then I am happy to withdraw my suggestion.

Mostly, I was trying to avoid "to do so" because that feels like we really should be able to write around that. :smile:

sarah11918 (approved):
Approving! This is fantastic @delucis ! :partying_face:

PR #9011, opened 2024-08-05
Improve TypeScript "setup" section

sarah11918:
#### Description (required)

This adds some content and new headings to the TypeScript setup section to allow easier direct linking to different concepts.

#### Related issues & labels (optional)

- Closes #9010

PR #9016, opened 2024-08-05
i18n(fr): Update `guides/typescript.mdx` from #9011

thomasbnt:
#### Description (required)
Update `guides/typescript.mdx` from #9011

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/guides/typescript.mdx
+++ b/src/content/docs/fr/guides/typescript.mdx
@@ -35,7 +35,7 @@ De plus, nos modèles incluent un fichier `env.d.ts` dans le dossier `src` pour 
 /// <reference types="astro/client" />
 ```
 
-### Plugin de l'éditeur TypeScript
+### Plugin Typescript pour les éditeurs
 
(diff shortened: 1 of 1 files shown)


src/content/docs/fr/guides/typescript.mdx
```diff
@@ -33,7 +35,12 @@ De plus, nos modèles incluent un fichier `env.d.ts` dans le dossier `src` pour
 /// <reference types="astro/client" />
 ```
 
-Si vous n'utilisez pas VSCode, vous pouvez installer le [plugin Astro TypeScript](https://www.npmjs.com/package/@astrojs/ts-plugin) pour prendre en charge l'importation de fichiers `.astro` à partir de fichiers `.ts` (ce qui peut être utile pour la réexportation).
+### Plugin de l'éditeur TypeScript
```
ArmandPhilippot, line 38: Not necessarily a change to add, that's why I'm not adding a review but a simple comment.

This heading bothers me. From my understanding, this does not concern a particular editor but any editor. So I will tend to use the plural. I will also change the word order since it's a Typescript plugin rather than an editor plugin (the plugin is added in `tsconfig.json`).

```suggestion
### Plugin Typescript pour les éditeurs
```

But maybe I misunderstand the use of `@astrojs/ts-plugin`. So let me know if you think my suggestion makes more sense or not.

thomasbnt: Agreed! Thanks!

ArmandPhilippot (approved):
LGTM! Thanks @thomasbnt

yanthomasdev (approved):
LGTM! Thanks! 🙌

PR #9027, opened 2024-08-06
Replace rehype-minify-html with rehype-preset-minify in mdx docs

millette:
#### Description (required)

`rehype-minify-html` doesn't exist. Replace with equivalent module.


delucis (approved):
Thanks @millette! I could have sworn we’d fixed this, but I guess it was in the [main Markdown guide](https://docs.astro.build/en/guides/markdown-content/#markdown-plugins).

PR #9034, opened 2024-08-07
Capitalized "javascript"

yamanidev:

- Closes #
- Suggested label:


sarah11918 (approved):
Thank you for spotting and taking the time to come fix this! We appreciate the attention to detail, and are happy to welcome you to Team Docs!  🥳

sarah11918:
Note to translators:

I have checked, and multiple languages have copied the lower case, so this is NOT an English-only change, and I won't mark it as such. 

This will update our tracker for all languages as "needing update" so apologies if your language does not need this fix!

PR #9046, opened 2024-08-08
Update why-astro.mdx to indicate content collections can also be JSON data

kyleshevlin:
Indicate that content collections can be JSON data as well

#### Description (required)

Content collections are not restricted to Markdown (or MDX) files, so I thought it might be more accurate to indicate they can be JSON data as well.

- Closes #
- Suggested label: 

No related issue. Just suggesting an edit.

`@kyleshevlin` on Discord

### DIFF
--- a/src/content/docs/en/concepts/why-astro.mdx
+++ b/src/content/docs/en/concepts/why-astro.mdx
@@ -16,7 +16,7 @@ Some highlights include:
 - **[UI-agnostic](/en/guides/framework-components/):** Supports React, Preact, Svelte, Vue, Solid, Lit, HTMX, web components, and more.
 - **[Server-first](/en/basics/rendering-modes/):** Moves expensive rendering off of your visitors' devices.
 - **[Zero JS, by default](/en/basics/astro-components/):** Less client-side JavaScript to slow your site down.
-- **[Content collections](/en/guides/content-collections/):** Organize, validate, and provide TypeScript type-safety for your Markdown content.
+- **[Content collections](/en/guides/content-collections/):** Organize, validate, and provide TypeScript type-safety for your JSON or Markdown content.
 - **[Customizable](/en/guides/integrations-guide/):** Tailwind, MDX, and hundreds of integrations to choose from.
 
 ## Design Principles


src/content/docs/en/concepts/why-astro.mdx
```diff
@@ -16,7 +16,7 @@ Some highlights include:
 - **[UI-agnostic](/en/guides/framework-components/):** Supports React, Preact, Svelte, Vue, Solid, Lit, HTMX, web components, and more.
 - **[Server-first](/en/basics/rendering-modes/):** Moves expensive rendering off of your visitors' devices.
 - **[Zero JS, by default](/en/basics/astro-components/):** Less client-side JavaScript to slow your site down.
-- **[Content collections](/en/guides/content-collections/):** Organize, validate, and provide TypeScript type-safety for your Markdown content.
+- **[Content collections](/en/guides/content-collections/):** Organize, validate, and provide TypeScript type-safety for your JSON or Markdown content.
```
at-the-vr, line 19: should `YAML` be appended to this list as well? I see the format is called data-serialization

at-the-vr (commented):
LGTM 👍 Leaving 1 doubt to think for

src/content/docs/en/concepts/why-astro.mdx
kyleshevlin: I don't see an example of YAML on this docs page: https://docs.astro.build/en/guides/content-collections/

Unless you're considering Markdown frontmatter YAML?

If you can indeed have `.yaml` files as an alternative to `.json` files, perhaps there should be a table that identifies which extensions will work for which types of collections.

sarah11918:
Just a note that one of the reasons we've left this (in addition to trying to get both 4.15 and 5.0 beta prepped simultaneously! 😅 ) is that it's not *technically* correct that *content* collections can be JSON. We've been calling those *data* collections.

This line predates the existence of data collections, which yes, can also be YAML, and we're not sure how we want to muddy up this very short sentence right now... especially with the content layer also affecting our story here. If anything, content collections can also be MDX and Markdoc, but it's somewhat assumed that these are both ways of writing Markdown content.

So, we're not really in a position to figure out how to make this 100% accurate while incorporating your suggestion, and we honestly haven't had the time to properly consider it! That's just to keep you informed about what's happening!

sarah11918:
We do list all the file types here, fwiw! https://docs.astro.build/en/guides/content-collections/#what-are-content-collections

> A collection entry is any piece of content stored inside of your content collection directory. Entries can use content authoring formats including Markdown (.md) and MDX (.mdx using the MDX integration) or as one of two supported data formats: YAML (.yaml) and JSON (.json).

sarah11918:
@kyleshevlin I am never one to turn away a contributor!

If we close *this* PR, here's what I'm open to instead:

We have a `5.0.0-beta` branch in docs where we're starting to prepare docs for the 5.0 beta releasing in mid September. At that time, the Content Layer (which is now experimental) will be stable and will power the next gen of "content collections" which do *not* distinguish themselves by type, and avoid the problem we have here! It also introduces the option to have collections with remote content, which is more compelling than Markdown maybe?. I think *those* docs are a great chance to revamp that one liner to something more like...

> - **[Content collections](/en/guides/content-collections/):** Organize, validate, and provide TypeScript type-safety for your local and remote content.

We will have to rewrite our content collections page *anyway* (because in the new version of collections, `type: content` or `type: data` won't exist), so I'll make sure that the supported file types are front and centre on the main page.

If you would like to think about, and maybe make a PR to the upcoming branch's docs, I'd be happy to have that line be more appropriate for content layer! (You can read about content layer here: https://docs.astro.build/en/reference/configuration-reference/#experimentalcontentlayer

sarah11918:
I'll close this one for now, but my offer above still stands! :raised_hands:

PR #9048, opened 2024-08-09
Use Flightcontrol logo instead of wordmark

in-the-trees:
#### Description (required)

The logo fits better within the space

| Before  | After  |
|---|---|
|  image  |  image  |

- Suggested label: [design](https://github.com/withastro/docs/labels/design)


public/logos/flightcontrol.svg
```diff
@@ -1,3 +1 @@
...
-</svg>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 21"><g fill="white" fill-rule="nonzero"><path d="M19.195 8.35H10.52c-1.22 0-2.204.986-2.204 2.21 0 1.222.985 2.21 2.204 2.21h6.133L19.2 8.35h-.005Z"/><path d="M10.526 0C4.722 0 0 4.734 0 10.559v.476C0 15.647 2.968 19.566 7.082 21l2.111-3.669-1.665-.952a6.14 6.14 0 0 1-3.125-5.344v-.476c0-3.384 2.748-6.139 6.123-6.139h10.927L24 0H10.526Z"/></g></svg>
```
delucis, line 1: Ran this through [SVGOMG](https://jakearchibald.github.io/svgomg/) to shrink it even smaller!

```suggestion
<svg xmlns="http://www.w3.org/2000/svg" fill="#fff" viewBox="0 0 24 21"><path d="M19.2 8.3h-8.7c-1.2 0-2.2 1-2.2 2.3 0 1.2 1 2.2 2.2 2.2h6.2l2.5-4.5Z"/><path d="M10.5 0C4.7 0 0 4.7 0 10.6v.4c0 4.6 3 8.6 7 10l2.2-3.7-1.7-1A6.1 6.1 0 0 1 4.4 11v-.4c0-3.4 2.8-6.2 6.1-6.2h11L24 0H10.5Z"/></svg>
```

delucis (changes requested):
Thanks! Looks like this logo is new since that page was added — appreciate the update 🙌

public/logos/flightcontrol.svg
in-the-trees: I also ran it through SVGOMG, I guess I didn't select the optimal settings!

delucis: Ah, I guess I also manually moved the `fill` attribute to get rid of the need for the `<g>` and then tweaked the number precision. So not entirely automated!

delucis (approved):
Thanks again @in-the-trees — looks good!

public/logos/flightcontrol.svg
in-the-trees: The rest of the brand logos could do with similar treatment as well eventually. Given the number of them, it could probably save a few KB

PR #9054, opened 2024-08-10
add docs for meta attribute in Code component

jcayzac:
#### Description (required)

Documentation for https://github.com/withastro/astro/pull/11605

Discord: @jcayzac


sarah11918 (commented):
Thanks for submitting this along with your PR, @jcayzac !

I think we can do better than duplicating an entire huge example just to add the meta property, though. So I may decide this should just be a line in the main transformers section, and we just show one example including the `meta` tag. We do something similar in the config section on the Markdown page, where we just show one config example, and one of the plugins happens to have additional options passed.

I'm thinking maybe something like: 

```
[Shiki transformers](https://shiki.style/packages/transformers#shikijs-transformers) can optionally be applied to code by passing them in through the `transformers` property as an array. Since Astro v4.14.0, you can also provide a string for [Shiki's `meta` attribute](https://shiki.style/guide/transformers#meta) to pass options to transformers.
```

And then update the code sample to include both transformers in the array, plus the `meta` option, highlighting the two lines relevant to transformers (we use Expressive Code to do that in Docs :wink:)

```
astro title="src/pages/index.astro" {12-13}
```
What do you think about something like that?

sarah11918:
Oops, I see from the original PR that you're describing `meta` completely independent of the transformer usage. I think I need more context about how you see this feature used, as your examples are probably showing specific use cases without painting a general picture first. (Which is why I inferred incorrectly what the option was in general).

I asked for clarification in the PR there, or you can reply here!

If `meta` is a standalone attribute that has use outside of transformers, then it shouldn't be explained *within* transformers. I'll need a more general introduction to what `meta` is and is used for first, and then we can show examples. This will help me figure out where the documentation goes.

jcayzac:
> I'm thinking maybe something like:
> 
> ```
> [Shiki transformers](https://shiki.style/packages/transformers#shikijs-transformers) can optionally be applied to code by passing them in through the `transformers` property as an array. Since Astro v4.14.0, you can also provide a string for [Shiki's `meta` attribute](https://shiki.style/guide/transformers#meta) to pass options to transformers.
> ```
> 
> And then update the code sample to include both transformers in the array, plus the `meta` option, highlighting the two lines relevant to transformers

LGTM!

> (we use Expressive Code to do that in Docs 😉)
> 
> ```
> astro title="src/pages/index.astro" {12-13}
> ```

I did not understand that part. Expressive Code is not relevant to that documentation.

sarah11918:
Expressive Code is not a part of what we are documenting here, but it's how *we* have to format *our* code samples of your example here! I'm just pointing out that I find it coincidental that you have provided an example about line highlighting showing how to do it one way, but in order for us to actually document that, *we* have to use Expressive Code to do the highlighting of your example!  :smile:

sarah11918:
Thanks @jcayzac ! Can you please take a look over my changes and make sure they're correct? Anything else you notice or want to see different here?

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2108,11 +2108,11 @@ This component **does not** inherit the settings from your [Shiki configuration]
...
 
-[Shiki transformers](https://shiki.style/packages/transformers#shikijs-transformers) can optionally be applied to code by passing them in through the `transformers` property as an array:
+[Shiki transformers](https://shiki.style/packages/transformers#shikijs-transformers) can optionally be applied to code by passing them in through the `transformers` property as an array. Since Astro v4.14.0, you can also provide a string for [Shiki's `meta` attribute](https://shiki.style/guide/transformers#meta) to pass options to transformers.
 
 Note that `transformers` only applies classes and you must provide your own CSS rules to target the elements of your code block.
 
-```astro
+```astro title="src/pages/index.astro" {12-13}
 ---
 import { transformerNotationFocus } from '@shikijs/transformers'
```
sarah11918, line 2117:
```suggestion
import { transformerNotationFocus, transformerMetaHighlight } from '@shikijs/transformers'
```

sarah11918 (approved):
Approving for docs! @jcayzac , if you spot anything before this is released, feel free to return and let me know.

Otherwise, I'll consider this PR to be ready for our minor release on Thursday! Thank you for this contribution, and welcome to Team Docs!  :partying_face:

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2124,7 +2124,8 @@ console.log(foo + bar) // [!code focus]
 <Code
   code={code}
   lang="js"
-  transformers={[transformerNotationFocus()]} />
+  transformers={[transformerNotationFocus(),[transformerMetaHighlight()]}
```
jcayzac, line 2127: There is a bogus square bracket here.

```suggestion
  transformers={[transformerNotationFocus(), transformerMetaHighlight()]}
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2124,7 +2124,8 @@ console.log(foo + bar) // [!code focus]
...
+  meta="{1,3}" />
   
   <style is:global>
     pre.has-focused .line:not(.focused) {
```
jcayzac, line 2131: There is no style for `.line.highlighted` (see my original commit), so the code example doesn't really work (nothing happens when you add `meta`).

src/content/docs/en/reference/api-reference.mdx
jcayzac, line 2127: Because `transformerNotationFocus` blurs everything but the line marked with `// [!code focus]`, using it together with `transformerMetaHighlight` makes little sense IMHO.

Or at least the highlighted line should be the same as the focused line.

jcayzac:
Thanks @sarah11918, I commented on the issues I found.

src/content/docs/en/reference/api-reference.mdx
sarah11918: OK! Shall we just delete the NotationFocus transformer then? Would also be OK if you wanted to suggest two that did go together! Just let me know.

sarah11918: Great, thanks! Oops! We'll fix this when we decide exactly which transformer(s) we should show!

sarah11918, line 2127:
```suggestion
  transformers={[transformerMetaHighlight()]}
```
will commit only the one that makes sense, but not opposed to showing an array if there's a second one that works nicely!

jcayzac: @sarah11918 sorry it took me some time to circle back to this PR. I see it's merged now.

The example now doesn't work. I has style for a transformer that's not used, and doesn't have any style for the transformer it uses.

The fixed example is that of the original commit: https://github.com/withastro/docs/pull/9054/commits/975a6e1cf9018663a78e9f95adea4316c73ce53b#diff-42959260ea642887b866e8e276a4520c1f112356eb9c15b31e510132a481aed6R2147-R2165

PR #9064, opened 2024-08-12
Document new `astro:route:setup` hook

bluwy:
#### Description (required)

Adds documentation for the new `astro:route:setup` hook based on the core PR below.

The main use of this new hook is to configure route options dynamically (in this case, only the `prerender` option for now). See the core PR for more info!

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/integrations-reference.mdx
+++ b/src/content/docs/en/reference/integrations-reference.mdx
@@ -626,7 +626,7 @@ A list of all generated pages. It is an object with one property.
 
 **When:** In `astro dev`, whenever a route is requested. In `astro build`, while bundling and transforming a route file.
 
-**Why:** To update a route's option, i.e. `prerender` to configure [on-demand server rendering](https://docs.astro.build/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode).
+**Why:** To update a route's option, i.e. `prerender` to configure [on-demand server rendering](/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode).
 
 ```js
 'astro:route:setup'?: (options: { route: RouteOptions }) => void | Promise<void>;
@@ -640,7 +640,7 @@ An object with a readonly `component` property and a `prerender` property.
 
 The `component` property is a readonly string path relative to the project root, e.g `"src/pages/blog/[...slug].astro"`.
 
-The `prerender` property is used to configure [on-demand server rendering](https://docs.astro.build/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode) for a route. The value may be a `boolean` or `undefined` depending if there's an explicit `export const prerender` value set in the file. If the final value after running all the hooks is `undefined`, it will fall back to a prerender default based on the [`output` option](/en/reference/configuration-reference/#output).
+The `prerender` property is used to configure [on-demand server rendering](/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode) for a route. The value may be a `boolean` or `undefined` depending if there's an explicit `export const prerender` value set in the file. If the final value after running all the hooks is `undefined`, it will fall back to a prerender default based on the [`output` option](/en/reference/configuration-reference/#output).
 
 ### Custom hooks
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -619,6 +620,28 @@ A list of all generated pages. It is an object with one property.
 
 - `pathname` - the finalized path of the page.
 
+### `astro:route:setup`
+
+<p><Since v="4.14.0" /></p>
+
+**When:** In `astro dev`, whenever a route is requested. In `astro build`, while bundling and transforming a route file.
+
+**Why:** To update a route's option, i.e. `prerender` to configure [on-demand server rendering](/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode).
```
sarah11918, line 629:
```suggestion
**Why:** To set options for a route at build or request time, such as enabling [on-demand server rendering](/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode).
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -619,6 +620,28 @@ A list of all generated pages. It is an object with one property.
...
+
+```js
+'astro:route:setup'?: (options: { route: RouteOptions }) => void | Promise<void>;
+```
+
+#### `route` option
+
+**Type:** `RouteOptions`
+
+An object with a readonly `component` property and a `prerender` property.
```
sarah11918, line 639:
```suggestion
An object with a `component` property to identify the route and the following additional values to allow you to configure the generated route: `prerender`.
```
So, thinking future-proofing here, it would be nice to set this up for not having to *totally* reconstruct the sentence as you add more of these. Maybe something like that?

I know it feels a little artificial when there's only *one* configurable value, but I like the idea of something that sets us up for: identify the component, and then here's all the updates you can make to this file when actually creating the route.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -619,6 +620,28 @@ A list of all generated pages. It is an object with one property.
...
+
+The `component` property is a readonly string path relative to the project root, e.g `"src/pages/blog/[...slug].astro"`.
+
+The `prerender` property is used to configure [on-demand server rendering](/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode) for a route. The value may be a `boolean` or `undefined` depending if there's an explicit `export const prerender` value set in the file. If the final value after running all the hooks is `undefined`, it will fall back to a prerender default based on the [`output` option](/en/reference/configuration-reference/#output).
```
sarah11918, line 643:
``````suggestion

#### `route.prerender`

**Type:** `boolean`<br />
**Default:** `undefined`<br />
<Since v="4.14.0" />
</p>
The `prerender` property is used to configure [on-demand server rendering](/en/guides/server-side-rendering/#opting-in-to-pre-rendering-in-server-mode) for a route. 

If the route file contains an explicit `export const prerender` value, ...
Otherwise, ...

```js
function setPrerender() {
  return {
    name: 'set-prerender',
    hooks: {
      'astro:route:setup': ({ route }) => {
        if (route.component.endsWith('/blog/[slug].astro')) {
          route.prerender = true;
        }
      },
    },
  };
}
```
 
If the final value after running all the hooks is `undefined`, the route will fall back to a prerender default based on the [`output` option](/en/reference/configuration-reference/#output): prerendered for `hybrid` mode, and on-demand rendered for `server` mode.
``````

So, make this correct, but I was thinking maybe something like this format style?

sarah11918 (commented):
Thank you for making this so easy for me, @bluwy :muscle: 

I left a couple of quick suggestions for how I could envsion making this a bit more future-proof. See what you think, and whether this makes any sense!

src/content/docs/en/reference/integrations-reference.mdx
bluwy: Sounds good to me. I didn't word it that way because I don't see us probably adding a new field in a looong time, but I could be wrong 😄 But anyways I don't think it hurts to futureproof it now

bluwy:
I've added and updated the `route.prerender` section

PR #9082, opened 2024-08-14
fix: add reassign keyword to Dynamic Tags section in `astro-syntax.mdx`

ArmandPhilippot:
This PR is a follow-up to [Erika's issue on Discord](https://discord.com/channels/830184174198718474/853350631389265940/1273261496230744208):

> I was searching for "reassignement" and "hydration" but couldn't find anything

The [Dynamic Tags](https://docs.astro.build/en/basics/astro-syntax/#dynamic-tags) section shows an example of a reassignment but this keyword does not appear so we cannot find the page using the search form. This keyword is important since a reassigned component cannot use `client:` directives.

With "*hydration*" the page appears and I don't think we can improve the excerpt (we can't guess what the user hopes to find by using this keyword, this could be in another context).

So this PR will help make the page stand out in search results using the "*reassignment*" keyword.

#### Description (required)

* Rewords the first sentence in `Dynamic tags` section to help find the page using `reassignment` in the search form.
* Fixes a typo when escaping the `style` attribute in the `define:vars` bullet point: the double backtick must be around style, not inside. (Typo introduced by... me, sorry 😅  ... my first PR here 🎉 )

I don't know if that's enough. I never used Algolia so I don't know how it handles keywords. I'm not sure if this would work with `reassigning` or `reassigned` as well. Currently, using `reassignment` or `reassigning` does not display pages containing `reassigned`.

So I'm not sure we can support all variants but this should work with `reassign` and `reassignment`.

**Question:** Is there a way to generate and use a local index with Algolia? I wanted to test on the dev version, but it seems my changes don't affect the search form so I guess this is a shared index with the production version.

- Issue: [Erika issue on Discord](https://discord.com/channels/830184174198718474/853350631389265940/1273261496230744208)
- Suggested label:  improve documentation


sarah11918:
This looks great! I think it will work nicely.

> Question: Is there a way to generate and use a local index with Algolia? I wanted to test on the dev version, but it seems my changes don't affect the search form so I guess this is a shared index with the production version.

No, and in fact, we need to reindex Algolia manually to see any updates even in live docs (other than the regular weekly crawls) so you were never going to be able to test that. :smile:

sarah11918 (approved):
Thank you so much, looks great!

ArmandPhilippot:
> No, and in fact, we need to reindex Algolia manually to see any updates even in live docs (other than the regular weekly crawls) so you were never going to be able to test that. 😄

Ah ok, I thought there was a reindex triggered at build time (but not locally... I tested it just in case 😅 ). Thanks for the explanation!

yanthomasdev (approved):
LGTM!

PR #9083, opened 2024-08-14
Fix Algolia DocSearch result categorisation

delucis:
#### Description (required)

This PR explicitly sets `attributesToHighlight` in the DocSearch client, which brings back categorisation of search results. Not 100% clear what changed on Algolia’s end to require this, but this should at least address the search issues users have been reporting.

Shout out to @levimichael for helping us out with this!


sarah11918 (approved):
Thanks for figuring this out!

Issue #7961, opened 2024-04-20
Document View Transition events fields


Fryuni:
### 📚 Subject area/topic

View Transitions

### 📋 Suggested page

Either on the View Transitions page (https://docs.astro.build/en/guides/view-transitions/#lifecycle-events) or on some reference page

### 📋 General description or bullet points (if proposing new content)

Document the type of the View Transition events, which fields are available and what each of them represent.

`astro:page-load`, `astro:after-preparation` and `astro:after-swap` have no special field and are the most used events, this may lead users to incorrectly assume that none of the events have special properties.

Both `astro:before-preparation` and `astro:before-swap` have [the following properties](https://github.com/withastro/astro/blob/237e17c8feea244b06b00ceaed08cb16d3d6baeb/packages/astro/src/transitions/events.ts#L21-L27):
- `from: URL`: The URL of the page initiating the transition.
- `to: URL`: The URL transitioning to. (can be modified)
- `direction: Direction | string`: Direction of the transition, used to change to a different animation when going back in history.
- `navigationType: NavigationTypeString`: What kind of history navigation is being applied
	- `push`: New URL is added to the history
	- `replace`: New URL replaces the previous URL in the history
	- `traverse`: History remains unchanged, transition is moving through the history
- `sourceElement: Element | undefined`: Element that triggered the navigation, if it came from a DOM element. This is undefined when the navigations was initiated from the `navigate` API.
- `info: any`: ?? Unclear to me where this comes from, requires some investigation.
- `newDocument: Document`: Content that will be swapped in place of the current document (`document.documentElement`).

`astro:before-preparation` also has:
- `formData: FormData | undefined`: The data that will be sent on the request to load the next page if a form action triggered the navigation.
- `loader: () => Promise<void>`: The function that runs the next phase of the transition (loading the next page). Can be overridden to add extra behavior or to prevent the execution of the next phase.

`astro:before-swap` also has:
- `viewTransition: ViewTransition`: The View Transition element that allows you to hook into it's promises. On browsers that don't have native VT this will be an object implementing the same API.
- `swap: () => void`: The function that runs the next phase of the transition (swapping the new document in place of the old document). Can be overridden to add extra behavior or to prevent the execution of the next phase.

sarah11918:
Seems like a job for @martrapp to weigh in on? Martin, can you provide clarity/guidance here for whatever might be missing from docs perspective, and confirm that we have an exhaustive list here?

[...]

PR #9084, opened 2024-08-14
Document swapFunctions of astro:transitions/client

martrapp:
#### Description (required)

In 4.8.0 [10908](https://github.com/withastro/astro/pull/10908) we introduced building blocks for custom swap() functions. Up to now we didn't properly document and export them. We first wanted to see if they worked. They did. 

This PR is intended to discuss and close this technical debt.


sarah11918:
I'm wondering whether we should have a nice API entry for `swapFunctions` here: https://docs.astro.build/en/reference/api-reference/#viewtransitions- ?

Since it's something we import from `astro:transitions`, it feels like there should be a reference item for it?

martrapp:
> I'm wondering whether we should have ...

Yes, definitively. It could be part of #7961! But i would not add it to the API section before the events. You use the custom swap functions to redefine the swap function of the event object.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -642,6 +642,35 @@ At this point of the lifecycle, you could choose to define your own swap impleme
 </script>
 ```
 
+#### Building blocks for custom swap functions
+
+<p><Since v="4.15.0" /></p>
+
+Astro offers utility functions that can be used to define a custom swap implementation.
+The `swapFunction` object of the `astro:transitions/client` module provides several reusable building blocks. The following example demonstrates how to combine these blocks to re-create Astro's built-in swap implementation.
```
sarah11918, line 650:
```suggestion
The `swapFunction` object of the `astro:transitions/client` module defines a `swap` implementation with five utility functions to describe specific actions related to page elements, attributes, and scripts during the swap transition. These can be used directly to define a custom swap implementation.

The following example uses these six functions to Astro's built-in `swap` implementation:
``` 
This might not be entirely accurate, but what about something in this style?

sarah11918 (commented):
OK, so I am now *really* feeling the lack of a section in API reference for `astro:transitions` :smile:  But for now, for this addition, I think we can do this, then separately let's get a new section in the API reference page and we'll come back here and link to a proper full reference of all these things!

martrapp:
Thank you very much for your help Sarah! There is a reason why Astro Docs have such a good reputation 💜

sarah11918 (approved):
Thanks for the helpful polish on this, @martrapp ! Approving for docs! (Wait, this *is* docs...)

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -642,6 +642,36 @@ At this point of the lifecycle, you could choose to define your own swap impleme
...
+  swapFunctions.deselectScripts(doc);
+  swapFunctions.swapRootAttributes(doc);
+  swapFunctions.swapHeadElements(doc);
+  const restoreFocusFunction = swapFunctions.saveFocus();
+  swapFunctions.swapBodyElement(doc.body, document.body);
+  restoreFocusFunction();
+};
+
+...
+  ev.swap = () => mySwap(ev.newDocument);
```
sarah11918, line 668:
```suggestion
  event.swap = () => mySwap(event.newDocument);
```
Last tiny tiny thing: assuming that `ev` stands for `event`, we generally prefer to write these words out in full as they can help people who aren't used to standard abbreviations. This would be my *preference*, if correct!

martrapp: Good one! I had used `ev` for consistency with the rest of the page. Now everything is consistent again ;)

yanthomasdev (approved):
LGTM @martrapp, good work! 💜

PR #9109, opened 2024-08-15
[5.0] Update docgen script for new path to types

Princesseuh:

- Closes #
- Suggested label:

### DIFF
--- a/scripts/docgen.mjs
+++ b/scripts/docgen.mjs
@@ -12,7 +12,7 @@ const HEADER = `---
 # NOTE: This file is auto-generated from 'scripts/docgen.mjs'
 # Do not make edits to it directly, they will be overwritten.
 # Instead, change this file: https://github.com/withastro/astro/blob/main/packages/astro/src/%40types/astro.ts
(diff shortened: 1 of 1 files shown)


yanthomasdev:
I added the `merge-on-release` label since we depend on [#11715](https://github.com/withastro/astro/pull/11715) to be merged first. If we merge *now*, docgen will break since the main branch still doesn't have this new file structure.

sarah11918:
Thanks, @yanthomasdev -- I had removed that label because I *think* it's safe to merge into the 5.0 branch? That shouldn't affect our normal running of anything on main, right?

I will probably need @delucis to let me know how (if at all possible) we can pull this new `packages/astro/src/types/public/config.ts` into `5.0.0-beta` so we can keep that branch updated separately from our action to keep regular docs updated. But I don't think we should be blocked from adding this to the 5.0 branch,and that's where all the 5.0 config references will now be made anyway.

Princesseuh:
This is actually not needed anymore, I ended up pushing the change to the branch directly

PR #9115, opened 2024-08-15
I18n indonesia

rogerwenas-id:

- Closes #
- Suggested label:


yanthomasdev:
Hey @rogerwenas-id, thanks for this PR, but there are a couple of reasons that make me have to close this:

1. In our [i18n and Translation contributing guide](https://contribute.docs.astro.build/guides/i18n/) we state we currently aren't accepting new languages into docs for a bunch of reasons (open the link for an explanation).

2. A 316+ pages PR is very, very hard to review, for both our team and fellow Indonesians that could help you, although I do see this PR only has a few actually translated pages and the rest are placeholders. If you desire to contribute in the future, I'd recommend dividing these on a per-page basis, including only translated files.

If in the future we open Astro Docs for Indonesian translation, we'd be more than happy to receive your contribution. Again, thank you for the PR, but for both reasons in our process of adding new languages, and in our ability to review PRs, we cannot accept this.

PR #9121, opened 2024-08-16
Remove misleading reference to {once: true} from astro:page-load example

martrapp:
#### Description (required)

It's always good to know that you can add `{once: true}` as an option to `addEventListener`.

But for the `astro:page-load` example, it's a deal-breaker. The whole point here is that the code runs every time the page loads, not just once.
While people can grasp this pretty quickly, Kapa was completely caught off guard and insisted on using `once: true` as a best practice ;-)

This came up in a discord discussion: https://discord.com/channels/830184174198718474/1075516604428857344/1273972688788000811

### DIFF
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -664,14 +664,14 @@ An event that fires at the end of page navigation, after the new page is visible
 
 The `<ViewTransitions />` component fires this event both on initial page navigation for a pre-rendered page and on any subsequent navigation, either forwards or backwards.
 
-You can use this event to run code on every page navigation, or only once ever:
+You can use this event to run code on every page navigation. That is especially useful when it comes to setting up event listeners that would otherwise be lost during navigation.
 
-```astro "{ once: true }"
+```astro
 <script>
   document.addEventListener('astro:page-load', () => {
-    // This only runs once.
-    setupStuff();
-  }, { once: true });
+    // This runs on first page load and after every navigation.
+    setupStuff(); // e.g. add click event listeners
+  });
 </script>
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -664,14 +664,14 @@ An event that fires at the end of page navigation, after the new page is visible
 
 The `<ViewTransitions />` component fires this event both on initial page navigation for a pre-rendered page and on any subsequent navigation, either forwards or backwards.
 
-You can use this event to run code on every page navigation, or only once ever:
+You can use this event to run code on every page navigation. That is especially useful when it comes to setting up event listeners that would otherwise be lost during navigation.
```
delucis, line 667: I think we can turn the second sentence into an example:

```suggestion
You can use this event to run code on every page navigation, for example to set up event listeners that would otherwise be lost during navigation.
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -664,14 +664,14 @@ An event that fires at the end of page navigation, after the new page is visible
...
 
-```astro "{ once: true }"
+```astro
 <script>
   document.addEventListener('astro:page-load', () => {
-    // This only runs once.
-    setupStuff();
-  }, { once: true });
+    // This runs on first page load and after every navigation.
+    setupStuff(); // e.g. add click event listeners
```
delucis, line 673: I guess it doesn’t **have** to be click listeners, so we can simplify a bit here:

```suggestion
    setupStuff(); // e.g. add event listeners
```

delucis (commented):
Thanks for taking care of this @martrapp! Left a couple of small suggestions, but definitely clearer now 🙌

martrapp:
Thank you Chris!

dreyfus92 (approved):
LGTM ✅

PR #9130, opened 2024-08-19
Add timeout option to client:idle

ph1p:
#### Description

Adds documentation for https://github.com/withastro/astro/pull/11743


src/content/docs/en/reference/directives-reference.mdx
```diff
@@ -133,6 +133,15 @@ Load and hydrate the component JavaScript once the page is done with its initial
 ```astro
 <ShowHideButton client:idle />
 ```
+
+### `client:idle={{timeout}}`
+
+Optionally, a value for `timeout` can be passed to the underlying `requestIdleCallback`. It hydrates the component if the client is not idle, but specifies the maximum time to wait.
```
sarah11918, line 139:
```suggestion
### `timeout`

<p><Since v="4.15.0" /></p>

The maximum value to wait, in milliseconds, before hydrating the component, even if the page is not yet done with its initial load.

This value allows you to pass [the timeout option from the `requestIdleCallback()` specification](https://www.w3.org/TR/requestidlecallback/#the-requestidlecallback-method). This means you can delay hydration for lower-priority UI elements with more control to ensure your element is interactive within a specified time frame.
```
Just added our "Added in version" component to make this clear for which versions of Astro it's available! And, since we are defining a property, we prefer to start with its definition, so I just lightly rearranged some of the ideas here, and added some extra context similar to my edits in the implementation PR. 

Like I wrote in that PR, you're the expert, so please correct as necessary! But I think something like this fits the style of our documentation nicely.

sarah11918 (commented):
Thank you for these great docs @ph1p ! I've made some minor adjustments to the structure/style to fit with the rest of Astro docs, and have left a comment below!

src/content/docs/en/reference/directives-reference.mdx
ph1p: Thank you! :)

src/content/docs/en/reference/directives-reference.mdx
```diff
@@ -133,6 +133,19 @@ Load and hydrate the component JavaScript once the page is done with its initial
...
 <ShowHideButton client:idle />
 ```
+
+### `timeout`
+
+<p><Since v="4.15.0" /></p>
+
+The maximum value to wait, in milliseconds, before hydrating the component, even if the page is not yet done with its initial load.
+
+This value allows you to pass [the timeout option from the `requestIdleCallback()` specification](https://www.w3.org/TR/requestidlecallback/#the-requestidlecallback-method). This means you can delay hydration for lower-priority UI elements with more control to ensure your element is interactive within a specified time frame.
```
sarah11918, line 143:
```suggestion
The maximum time to wait, in milliseconds, before hydrating the component, even if the page is not yet done with its initial load.

This allows you to pass a value for [the `timeout` option from the `requestIdleCallback()` specification](https://www.w3.org/TR/requestidlecallback/#the-requestidlecallback-method). This means you can delay hydration for lower-priority UI elements with more control to ensure your element is interactive within a specified time frame.
```

sarah11918 (approved):
Happy with this one! Thank you again @ph1p , and welcome to Team Docs! :partying_face:

yanthomasdev (approved):
LGTM! Good job @ph1p!

sarah11918:
!coauthor

PR #9131, opened 2024-08-19
[i18nIgnore] Upgrade Starlight to 0.26.1

liruifengv:
#### Description (required)

Upgrade Starlight to 0.26.1

Hide the <Sponsors/> on mobile 

- https://github.com/withastro/starlight/pull/2150
- https://github.com/withastro/starlight/pull/1784
- https://github.com/withastro/starlight/pull/2051


src/components/starlight/Sidebar.astro
```diff
@@ -32,8 +32,7 @@ async function remap(i: SidebarEntry) {
 sidebar = await Promise.all(sidebar.map(remap));
 ---
 
-<SidebarSublist sublist={sidebar} />
-
+<Default {...Astro.props} sidebar={sidebar} />
 <Sponsors />
 
 <div class="md:sl-hidden">
```
delucis, line 38: I think it’s going to be a bit trickier than this because the sidebar already includes the mobile menu footer re-rendered below here, so on mobile this change results in:

Bottom of a mobile menu showing a footer with colour scheme and language options being rendered twice

Might need to remove the footer from this file? But not quite sure yet what to do with the sponsors 🤔

src/styles/custom.css
```diff
@@ -9,3 +9,8 @@
 		grid-template-columns: 1fr 1fr;
 	}
 }
+
+/* Hide the guideline for the final step in <Steps> lists. */
+.sl-steps > li:last-of-type::after {
+  background: transparent;
+}
```
delucis, line 16: I think we can use the new Starlight styles directly instead of overriding them. This style was changed to make the last step in `<Steps>` more obvious to users based on user feedback, so I think the new Starlight default makes sense for us in Astro Docs too.

```suggestion
```

src/content/docs/en/getting-started.mdx
```diff
@@ -12,9 +12,9 @@ hero:
     - text: Install Astro
       icon: rocket
       link: /en/install-and-setup/
-      variant: primary
     - text: Learn about Astro’s features
       icon: right-arrow
+      variant: minimal
```
delucis, line 17: Perhaps we can make this change across all languages and mark this PR as `i18nIgnore`? What do you think? Otherwise some languages will have a double “primary” button until they are updated, like this:
image

delucis (commented):
Thanks for tackling this @liruifengv!

src/components/starlight/Sidebar.astro
liruifengv: Yeah. Maybe we need to refactor Starlight's Sidebar component? So that it's easier for people to override it.

I tried to migrate the sl-sidebar-state-persist logic to the SideBarSublist component and it still seems to be working!

image

src/content/docs/en/getting-started.mdx
liruifengv: Ok. I will change all other languages and mark this PR as `i18nIgnore`

src/components/starlight/Sidebar.astro
delucis: The issue there is that `<SidebarSublist>` is rendered recursively, so having the script there means you will get it rendered for every sidebar group, which is not what we want.

The easiest quick fix might be to:
- Remove the extra mobile footer in our override
- Hide the `<Sponsors/>` on mobile

Then later we can figure out if sponsors on mobile is important and how to bring them back.

liruifengv: I've temporarily hidden the `<Sponsors/>` component on mobile.

PR #9133, opened 2024-08-19
feat(i18n): document new i18n logic

ematipico:
#### Description (required)

This PR documents `i18n.routing.fallbackType`, a new feature that will change how the fallback works.


src/content/docs/en/guides/internationalization.mdx
```diff
@@ -307,6 +307,29 @@ export default defineConfig({
...
+
+```js title="astro.config.mjs" ins={8}
+export default defineConfig({
+  i18n: {
+    defaultLocale: "en",
+    locales: ["es", "en", "fr"],
+    fallback: {
+      fr: "es"
+    },
+    fallbackType: "rewrite" // the value `"redirect"` is the default behaviour and value
```
yanthomasdev, line 326: nit: we use both the spelling `behavior` and `behaviour` here, we should default to the American spelling:
```suggestion
    fallbackType: "rewrite" // the value `"redirect"` is the default behavior and value
```

sarah11918 (commented):
@ematipico Take a look at what I've done and see what you think!

The existing Fallback content needed some revising anyway, because it described the existing behaviour of redirects everywhere.

Since this isn't a reference page, and we have full config reference items we can link to for `fallback` and `fallbackType`, this page can simply tell the story of what a fallback strategy looks like.

(Note, I will actually want to revisit the rest of this page in the same way at some point, but for now, let's just get the fallback in place!)

Let me know what you think of this, and whether seeing the big picture like this makes you want to e.g. move the option from `i18n.routing` and instead maybe rethink adding to `fallback` itself? When you look at what it takes to configure this, and how it is explained, you might think of a better way config could look to achieve a more complete fallback outcome. (setting in two places like this might seem a little clunky)

But, see what you think and let me know!

src/content/docs/en/guides/internationalization.mdx
```diff
@@ -286,27 +286,37 @@ With the above configuration:
...
+When a page in one language doesn't exist (e.g. a page that is not yet translated), instead of displaying a 404 page, you can choose to display fallback content from another `locale` on a per-language basis. This is useful when you do not yet have a page for every route, but you want to still provide some content to your visitors.
 
-For example, the configuration below sets `es` as the fallback locale for any missing `fr` routes. This means that a user visiting `example.com/fr/my-page/` will be redirected to and shown the content for `example.com/es/my-page/` instead of being taken to a 404 page when `src/pages/fr/my-page.astro` does not exist.
+Your fallback strategy consists of two parts: choosing which languages should fallback to which other languages ([`i18n.fallback`](/en/reference/configuration-reference/#i18nfallback)) and choosing whether to perform a [redirect](https://docs.astro.build/en/guides/routing/#redirects) or a [rewrite](https://docs.astro.build/en/guides/routing/#rewrites) to show the fallback content ([`i18n.routing.fallbackType`](/en/reference/configuration-reference/#i18nroutingfallbacktype) added in Astro v4.15.0).
```
sarah11918, line 293:
```suggestion
Your fallback strategy consists of two parts: choosing which languages should fallback to which other languages ([`i18n.fallback`](/en/reference/configuration-reference/#i18nfallback)) and choosing whether to perform a [redirect](/en/guides/routing/#redirects) or a [rewrite](/en/guides/routing/#rewrites) to show the fallback content ([`i18n.routing.fallbackType`](/en/reference/configuration-reference/#i18nroutingfallbacktype) added in Astro v4.15.0).
```

ematipico:
Thank you @sarah11918 , I approve your new edits!

sarah11918:
Noting that the link check should pass once the new config reference docs have been pulled in!

sarah11918 (approved):
I think we're in a good place here! :partying_face:

sarah11918:
!coauthor

PR #9148, opened 2024-08-20
fix: newline not being interpreted while adding .astro directory to gitignore file

ManorSailor:
This PR addresses an issue in the [Content Collection](https://docs.astro.build/en/guides/content-collections/#the-astro-directory) docs where running the given command in the tip callout does not work on Bash shell - version `5.1.16(1)`. Executing the command as-is renders the newline sequence within the `.gitignore` file, i.e., `\n.astro`

We need to use the `-e` flag to tell `echo` to interpret backslashes. As a sidenote, the same command works fine on `ZSH` without any changes.

- Closes  Didn't find any existing issue for this.
- Suggested label: `code snippet update` and/or `improve documentation`

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -38,10 +38,6 @@ The `.astro` directory will be updated for you automatically anytime you run the
 
 :::tip
 If you're using Git for version control, we recommend ignoring the `.astro` directory by adding `.astro` to your `.gitignore`. This tells Git to ignore this directory and any files inside of it.
-
-```bash
-echo -e "\n.astro" >> .gitignore
-```
 :::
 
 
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Thank you for pointing out the error in the script, and taking the time to come fix it, @ManorSailor ! We really appreciate the effort to help keep our docs accurate and helpful!  :raised_hands: 

We discussed this, and realized the *best* way to document this is to remove the script entirely as the specific steps needed to ignore the directory may work differently across different systems etc. It's sufficient to say that you should ignore this directory, and then let each user determine how they'd like to do that! (And, search for how to do that, if necessary.) This will also mean that any updates to this process will not require a change to our documentation. The goal remains the same.

Thank you again for this contribution to improve our advice here! We are happy to have this fix, and to welcome you to Team Docs!  :partying_face:

PR #9157, opened 2024-08-21
feat(docgen): Automatically add deprecated messages to removed errors

Princesseuh:
#### Description (required)

Astro developers keep not following my carefully crafted rules regarding removing errors. Which probably mean my rules are not so carefully crafted. This PR makes it so when an error is removed, a basic deprecated message is added automatically, so at least there's some indication that one could upgrade their Astro version to proceed further.


scripts/error-docgen.mjs
```diff
@@ -96,6 +100,29 @@ export async function run() {
...
+					filePath,
+					currentContent.replace(
+						'<DontEditWarning />',
+						['<DontEditWarning />', '', getDeprecatedText(true)].join('\n')
+					),
+					'utf8'
+				);
+			}
+		}
+	}
```
delucis, line 124: The `if` here is redundant right?

```suggestion
	for (const error of deprecatedErrors) {
		const filePath = `src/content/docs/en/reference/errors/${error}`;
		const currentContent = fs.readFileSync(filePath, 'utf8');

		// If the error got removed without a deprecation, add a deprecation notice
		if (!currentContent.includes(':::caution[Deprecated]')) {
			fs.writeFileSync(
				filePath,
				currentContent.replace(
					'<DontEditWarning />',
					['<DontEditWarning />', '', getDeprecatedText(true)].join('\n')
				),
				'utf8'
			);
		}
	}
```

src/content/docs/en/reference/errors/env-invalid-variable.mdx
```diff
@@ -13,6 +13,11 @@ import DontEditWarning from '~/components/DontEditWarning.astro'
 <DontEditWarning />
 
 
+:::caution[Deprecated]
+This error cannot be emitted by Astro anymore
```
delucis, line 17: I think this wording is maybe a tiny bit confusing, because some people landing here will literally have just had this error emitted by Astro. So it is possible, if you are on an old enough version.

Maybe something like this would work better? (And in the other places it is reused.)

```suggestion
This error was emitted by an older version of Astro and is no longer in use.
```

delucis (commented):
🫡 Thanks @Princesseuh!

src/content/docs/en/reference/errors/env-invalid-variable.mdx
sarah11918: So not only *are* they getting that error, but we also know they are running an older version of Astro. Maybe some kind of message like below is additionally helpful?

> This error is from an older version of Astro and is no longer in use. If you are unable to upgrade your project to a more recent version, then you can consult [unmaintained snapshots of older documentation][(/en/upgrade-astro/#older-docs-unmaintained) for assistance.

delucis (approved):

PR #9160, opened 2024-08-21
feat: astro:env stable

florian-lefebvre:
http://localhost:4321/en/guides/environment-variables/

- Closes #
- Suggested label:


src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -155,7 +156,7 @@ export default defineConfig({
 })
 ```
 
-These features are all available by default in Astro v5.0. Only `astro:env` requires any configuration to use.
+These features are all available by default in Astro v5.0. Only `astro:env` [requires some configuration](/en/guides/environment-variables/#astroenv) to be used.
```
sarah11918, line 159:
```suggestion
These features are all available by default in Astro v5.0. Only [`astro:env`](/en/guides/environment-variables/#astroenv) requires any configuration to be used.
```

src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -9,6 +9,7 @@ import { Steps } from '@astrojs/starlight/components';
 ## LINEAR ISSUES MARKED AS COMPLETED BUT STILL REQUIRING ATTENTION
 
 - (PLT-2347) Figure out where Zod comes from -- I THINK THIS WILL NEED DOCS, BUT NO TASK FOR IT
+- (PLT-2169) astro:env stable - This needs more docs
```
sarah11918, line 12:
```suggestion
```
We are making the new docs now! :smile:

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -5,9 +5,44 @@ i18nReady: true
 ---
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
-:::tip[Looking for astro&#58;env\?]
-Find out more about the [experimental `astro:env` API](/en/reference/configuration-reference/#experimentalenv) for type-safe environment variables!
-:::
+Astro supports 2 main ways for handling environment variables: the `astro:env` module and Vite's built-in support.
+
+## `astro:env`
```
sarah11918, line 10: Why did you not just copy the existing docs from the experimental section to start? In the worst case, just drop those in here and then we go from there!

sarah11918 (commented):
Quick initial review.. I didn't realized until I opened this up that it wasn't exactly ready for a *review* yet. :sweat_smile: 

See my comment below re: just reusing the material we already have, then working from there!

src/content/docs/en/guides/environment-variables.mdx
florian-lefebvre: Well I thought it was important to say there are 2 ways to do it for people not using astro:env just yet!

sarah11918: I meant, under the heading specifically for astro:env !  :smile:

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -9,38 +9,240 @@ Astro supports 2 main ways for handling environment variables: the `astro:env` m
...
+---
+
+<script>
+import { API_URL } from "astro:env/client"
+fetch(`${API_URL}/ping`)
+</script>
+```
 
-TODO: where to talk about integrations & adapters?
+**TODO: where to talk about integrations & adapters?**
```
sarah11918, line 51: If you mean for integration/adapter *authors*, then this can be a note with a link to one of the other API pages. I don't think this page should have information about writing integrations.

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -9,38 +9,240 @@ Astro supports 2 main ways for handling environment variables: the `astro:env` m
...
+
+   ```js
+   import { API_SECRET } from "astro:env/server"
+   ```
+
+   Worth nothing that by default, secrets are only validated at runtime. To validate secrets on start, check [validateSecrets](#validatesecrets-option).
+
+**Note:** Secret client variables are not supported because there is no safe way to send this data to the client. Therefore, it is not possible to configure both `context: "client"` and `access: "secret"` in your schema.
+
+### Data types
```
sarah11918, line 116: A lot of this is starting to get quite API referencey. Not everything is appropriate for a guide page. We should have an `astro:env` section on the API reference page for this content.

sarah11918:
Hey @florian-lefebvre , I'll be looking at this more closely today, but just pointing out that some of this content should end up in reference pages and NOT crammed all in here, which is more of a guide page.

As an example, see what I'm suggesting in #9133 - the guide content doesn't have to spell things out like an API reference (and shouldn't, because then we're duplicating content and our work!). Rather, it tells the story around using this and links to references where all the details of the items are kept.

So that's my first piece of advice, keep this for people learning about/setting this up for the first time, or coming back to confirm general usage, not looking up values etc. And  then we'll link to each item's reference location if they need to e.g. confirm types etc.

florian-lefebvre:
@sarah11918 tried my best to improve things!

sarah11918:
Just noting that I removed changes from the upgrade guide because I've already been working on that separately! This should also resolve conflicts on the environment variables page, so we can start fresh with that one!

sarah11918:
Hey @florian-lefebvre ! I committed an editing pass here that I think might be a more helpful organization structure? For example, the stuff about how to actually set values for environment variables was inside "Support for Vite" and you're gonna need to set these variables when using `astro:env`, too! :smile: 

So I tried to envision a flow for the entire page that is like;
- you have full Vite support, and some included defaults you can just use out of the box
- here's how you set them
- here's how you retrieve them
- Here's how you get type safe environment variables
- Intellisense (since it relates to both)

`getSecret()` absolutely needs an API reference entry on the API reference page. Is there anything else you'd put there with it? This section of this page is a little less fleshed-out, so let's start by getting a good API reference for this one, then we'll know what's already been said that we can just link to from this page. And this page can focus more on using it. E.g. maybe a better heading than `getSecret` for this page (since it's not the API reference) is "Retrieving secrets dynamically".

So see what you think about this, and tell me where / how it doesn't work so we can iterate on it!

florian-lefebvre:
@sarah11918

- I moved ts intellisense under vite support since it's not relevant for astro:env
- i moved getSecret to the api reference
- i still don't know what to do with the limitations, i don't know if there are similar cases in docs?

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -9,7 +9,7 @@ import ReadMore from '~/components/ReadMore.astro';
 
 Astro gives you access to [Vite's built-in environment variable support](#vites-built-in-support) and includes some [default environment variables for your project](#default-environment-variables) that allow you to access configuration values for your current project (e.g. `site`, `base`), whether your project is running in development or production, and more.
 
-Astro also provides [a virtual `astro:env` module](/en/reference/configuration-reference/#env) to use and organize your environment variables with type safety. It is available for use inside the Astro context (e.g. Astro components, routes and endpoints; UI framework components; middleware), and [managed with a schema in your Astro configuration](#astroenv).
+Astro also provides [a virtual `astro:env` module](/en/reference/configuration-reference/#env) to use and organize your environment variables with type safety. It is available for use inside the Astro context (e.g. Astro components, routes and endpoints; UI framework components; middleware), and [managed with a schema in your Astro configuration](#define-your-schema).
```
sarah11918, line 12: Note, I *intentionally* put the main `#astroenv` link here because we hadn't yet provided a link down to that section! The first link goes to the reference page, but I think people should be taken down to read *about* Astro env.

Open to suggestions if you'd like to change the wording, (it doesn't have to match exactly where the link takes you!) I just wanted to entice people, and let them get directly to that section. :smile:

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2302,6 +2302,39 @@ The `code` property accepts human-readable versions of all HTTP status codes. Th
...
+
+The `astro:env` API lets you configure a type-safe schema for environment variables you have set. This allows you to indicate whether they should be available on the server or the client, and define their data type and additional properties. For examples and usage instructions, [see the `astro:env` guide](/en/guides/environment-variables/#type-safe-environment-variables).
+
+### `getSecret()`
+
+<p>
+<Since v="5.0.0" />
+</p>
+
+The `getSecret()` helper function can be used for several things:
```
sarah11918, line 2319: Definition first, always, please!  :smile:

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2302,6 +2302,39 @@ The `code` property accepts human-readable versions of all HTTP status codes. Th
...
+      FOO, // boolean
+      getSecret
+   } from 'astro:env/server'
+
+   getSecret('FOO') // string | undefined
+   ```
+
+2. **Retrieve secrets not specified in your schema**: that can be useful if you need to get a secret depending on dynamic data, eg. a database or API.
+
+   :::note
```
sarah11918, line 2334: I think the word "use" is doing a lot of heavy lifting here. At first, I read it as "if you have environment variables at all" which is why I had trouble making sense of it. I think now you mean, "if you have to build logic around your environment variables" or "use your environment variables programmatically?"

Can you be more explicit about "use" (because *having* them in a project is also "using" them, and my first read was, "If I'm supposed to use getSecret if I have environment variables, what was all the other stuff about?)

Also, is it possible that this really doesn't require a note, but just a line somewhere in the description/explanation as an "advantage of `astro:env`? :smile: 

e.g.

> The `getSecret()` helper function can fetch the value of your environment variables even if you switch adapters, and therefore is more flexible than using `process.env`

src/content/docs/en/guides/environment-variables.mdx
florian-lefebvre: I feel like having the reference first is a bit odd as it could redirect people before looking at the guide

```suggestion
Astro also provides a way to [use and organize your environment variables with type safety](#type-safe-environment-variables). It is available for use inside the Astro context (e.g. Astro components, routes and endpoints, UI framework components, middleware), and managed with a schema in your Astro configuration. You can see learn more about the `astro:env` module configuration options [in the Configuration reference](/en/reference/configuration-reference/#env).
```

src/content/docs/en/reference/api-reference.mdx
florian-lefebvre:
```suggestion
The `getSecret()` helper function allows retrieving the raw value of an environment variable by its key. It has 2 usecases:
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2302,6 +2302,39 @@ The `code` property accepts human-readable versions of all HTTP status codes. Th
...
+   If you need to use environment variables, we recommend you use `getSecret()` instead of `process.env` (or equivalent) because its implementation is provided by your adapter: It means you won't need to update all your calls if you switch adapters. It defaults to `process.env` in dev and build.
+   :::
```
florian-lefebvre, line 2336:
```suggestion
If you need to retrieve environment variables programmatically, we recommend you use `getSecret()` instead of `process.env` (or equivalent) because its implementation is provided by your adapter: It means you won't need to update all your calls if you switch adapters. It defaults to `process.env` in dev and build.
```

src/content/docs/en/guides/environment-variables.mdx
sarah11918: Right, makes sense!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2302,6 +2302,37 @@ The `code` property accepts human-readable versions of all HTTP status codes. Th
...
+      FOO, // boolean
+      getSecret
+   } from 'astro:env/server'
+
+   getSecret('FOO') // string | undefined
+   ```
+
+2. **Retrieve secrets not specified in your schema**: that can be useful if you need to get a secret depending on dynamic data, eg. a database or API.
+
+If you need to retrieve environment variables programmatically, we recommend you use `getSecret()` instead of `process.env` (or equivalent) because its implementation is provided by your adapter: It means you won't need to update all your calls if you switch adapters. It defaults to `process.env` in dev and build.
```
sarah11918, line 2334:
```suggestion
The `getSecret()` helper function allows retrieving the raw value of an environment variable by its key.

For example, you can retrieve a boolean value as a string:

   ```js
   import {
      API_KEY, // boolean
      getSecret
   } from 'astro:env/server'

   getSecret('API_KEY') // string | undefined
   ```

This can also be useful to get a secret depending on dynamic data, such as from a database or API.

If you need to retrieve environment variables programmatically, we recommend using `getSecret()` instead of `process.env` (or equivalent). Because its implementation is provided by your adapter, you won't need to update all your calls if you switch adapters.  It defaults to `process.env` in dev and build.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -394,3 +394,51 @@ function createEdgeMiddleware(middlewareEntryPoint) {
...
+
+  const handler = (event, context) => {
+    // ...
+  };
+
+  return { handler };
+}
+```
+
+If you support secrets, calling `getSecret()` until `setGetEnv()` is called will throw a runtime error. This is important to keep in mind if your environment variables are tied to the request:
```
sarah11918, line 419:
```suggestion
If you support secrets, calling `getSecret()` before `setGetEnv()` is called will throw a runtime error. This is important to keep in mind if your environment variables are tied to the request:
```
should this be before?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -394,3 +394,51 @@ function createEdgeMiddleware(middlewareEntryPoint) {
     // emit a new physical file using your bundler
 }
 ```
+
+### envGetSecret
+
+To support [`astro:env` secrets](/en/guides/environment-variables/#getsecret), you need to update the `envGetSecret` adapter feature. Additionally, the `astro/env/setup` module allows you to provide an implementation for `getSecret()`. In your server entrypoint, call `setGetEnv()` as soon as possible:
```
sarah11918, line 400: Can we have a line explaining what this is before jumping into using it? Maybe matching the other entries (functionPerRoute was there before it was deleted in this branch. Maybe something like:

(I don't know if this should be `true`, or `"stable"` or what.)

``````suggestion
This is a feature to allow your adapter to retrieve secrets configured by users in `env.schema`.

Enable the feature by passing `true` to the adapter:

```js title="my-adapter.mjs" ins={9-11}
export default function createIntegration() {
  return {
    name: '@matthewp/my-adapter',
    hooks: {
      'astro:config:done': ({ setAdapter }) => {
        setAdapter({
          name: '@matthewp/my-adapter',
          serverEntrypoint: '@matthewp/my-adapter/server.js',
          adapterFeatures: {
              envGetSectret: true
          } 
        });
      },
    },
  };
}
```

The `astro/env/setup` module allows you to provide an implementation for `getSecret()`. In your server entrypoint, call `setGetEnv()` as soon as possible:
``````

**Review by sarah11918 — COMMENTED** [review]
I think we're finally happy with this????

File: src/content/docs/en/reference/adapter-reference.mdx [hunk 13]
```diff
@@ -394,3 +395,74 @@ function createEdgeMiddleware(middlewareEntryPoint) {
     // emit a new physical file using your bundler
 }
 ```
+
+### envGetSecret
+
+This is a feature to allow your adapter to retrieve secrets configured by users in `env.schema`.
+
+Enable the feature by passing any valid `SupportsKind` value to the adapter:
```
sarah11918, line 403:
```suggestion
Enable the feature by passing `envGetSectret: 'stable'` to the `adapterFeatures`:
```
Does this make more sense now?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -394,3 +395,74 @@ function createEdgeMiddleware(middlewareEntryPoint) {
...
+
+  const handler = (event, context) => {
+    // ...
+  };
+
+  return { handler };
+}
+```
+
+If you support secrets, calling `getSecret()` before `setGetEnv()` is called will throw a runtime error. This is important to keep in mind if your environment variables are tied to the request:
```
sarah11918, line 443:
```suggestion
If you support secrets, be sure to call `setGetEnv()` before `getSecret()` when your environment variables are tied to the request:
```
Is this correct to provide good guidance, in a postitive (ie. Do this!) way?

src/content/docs/en/reference/adapter-reference.mdx
florian-lefebvre: Well technically people could say it's in experimental, that's why i didn't want to say stable here

sarah11918: This is for the 5.0 beta docs though right? It's not experimental in these docs?

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -93,22 +121,181 @@ const data = fetch(`${import.meta.env.PUBLIC_POKEAPI}/pokemon/squirtle`);
...
+
+envField.enum({
+   // context & access
+   values: ['foo', 'bar', 'baz'],
+   optional: true,
+   default: 'baz',
+})
 ```
+
+Some of them come with more optional validation rules, check out the [source](https://github.com/withastro/astro/blob/main/packages/astro/src/env/schema.ts#L3-L45) to learn more.
```
sarah11918, line 252: Is there a reason we are not *documenting* these? I would rather point people to a docs link than source code.

src/content/docs/en/reference/adapter-reference.mdx
florian-lefebvre: I mean that's a different thing. Yes `astro:env` is stable but this option is for the adapter to tell its users if it supports `astro:env` well

src/content/docs/en/guides/environment-variables.mdx
florian-lefebvre: Because we have no good place for it! It's not a runtime API thing, it's a configuration reference thing but that one is fully automated. IMO most people will discover these options with autocompletion in their editor

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: AH, I was missing that part. OK, makes sense then!

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -93,22 +121,181 @@ const data = fetch(`${import.meta.env.PUBLIC_POKEAPI}/pokemon/squirtle`);
...
+
+   By default, secrets are only validated at runtime. You can enable validating private variables on start by [configuring `validateSecrets: true`](/en/reference/configuration-reference/#envvalidatesecrets).
+
+:::note
+**Secret client variables** are not supported because there is no safe way to send this data to the client. Therefore, it is not possible to configure both `context: "client"` and `access: "secret"` in your schema.
+:::
+
+### Data types
+
+There are currently four data types supported: strings, numbers, booleans and enums:
```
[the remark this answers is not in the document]
florian-lefebvre: there are not required, it's just the most common props. It was a middle ground between showing nothing (no codeblock) and showing every possible option

src/content/docs/en/guides/environment-variables.mdx
sarah11918: Right, so the docs should be in config reference!

sarah11918: I suggest we make a dummy `envField` entry in config reference locally in this branch, then go make one properly in the types/public file to fill it in.

sarah11918, line 252:
```suggestion
<ReadMore>For a complete list of validation fields, see the [`envField` API reference](/en/reference/configuration-reference/#envschema).</ReadMore>
```

src/content/docs/en/guides/environment-variables.mdx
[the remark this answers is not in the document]
florian-lefebvre: So no change required here isn't it?

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -93,22 +121,181 @@ const data = fetch(`${import.meta.env.PUBLIC_POKEAPI}/pokemon/squirtle`);
...
+export default defineConfig({
+  env: {
+    schema: {
+      // ...
+    }
+  }
+})
+```
+
+You can then [register variables as a string, number, or boolean](#data-types) using the `envField` helper. Define the [kind of environment variable](#variable-types) by providing a `context` (client or server) and `access` (secret or public) for each variable, and pass any additional properties such as `optional` or `default` in an object:
```
florian-lefebvre, line 148:
```suggestion
You can then [register variables as a string, number, enum or boolean](#data-types) using the `envField` helper. Define the [kind of environment variable](#variable-types) by providing a `context` (client or server) and `access` (secret or public) for each variable, and pass any additional properties such as `optional` or `default` in an object:
```

sarah11918: Yes, good call!

sarah11918 (approved):
Actually, realizing this can be merged now because the link to `#envschema` shouldn't break! It just won't have the information promised to be there yet. until the core PR is merged. :smile:

PR #9162, opened 2024-08-21
Fix import in Netlify adapter guide

delucis:
#### Description (required)

Since v4, the Netlify adapter has used a single `@astrojs/netlify` import, rather than the separate `static` one. This was likely overlooked when the docs were updated in https://github.com/withastro/adapters/pull/84.

I tested the import like this works as described, so it’s a simple typo fix effectively.


sarah11918 (approved):
Thank you, @delucis ! We really appreciate you taking the time to contribute a fix!  :rocket:

Issue #9125, opened 2024-08-16
Link to component prop types for getting props of built in components


sarah11918:
### 📚 Subject area/topic

Typing component props

### 📋 Suggested page

https://docs.astro.build/en/reference/api-reference/#built-in-components

### 📋 General description or bullet points (if proposing new content)

Add a link in the API reference section for our built-in components to our documentation about `ComponentProps`. Something like just adding the statement in bold to the existing text.

Built-in Components
Astro includes several built-in components for you to use in your projects. All built-in components are available in .astro files via import {} from 'astro:components';. **You can reference the `Props` of these components using the [`ComponentProps` type](/en/guides/typescript/#componentprops-type)**.

ManorSailor:
How about utilizing a custom components like _tip_ or _note_ instead of a bold statement?

sarah11918:
Sorry, I can see the misunderstanding. I made the statement bold so that it would be obvious what change to make! No, the text should not be bold, and it should not be any kind of tip or note! If it's just regular documentation, it does not belong in a note.

I'll fix that in the PR!

PR #9167, opened 2024-08-22
Update api-reference page to refer to `ComponentProp` type utility

ManorSailor:
#### Description (required)

I have made the text bold as suggested in the PR. Though, it looks like the odd-one-out among other texts. I addressed the same concern in [this](https://github.com/withastro/docs/issues/9125#issuecomment-2302629507) comment yesterday, but it was yet to receive a reply. If required, I will make the changes as needed.

- Closes: #9125 

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -2077,7 +2077,7 @@ export const onRequest = defineMiddleware(async (context, next) => {
 
 ## Built-in Components
 
-Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components';`.
+Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components';`. **You can reference the `Props` of these components using the [`ComponentProp` type](/en/guides/typescript/#componentprops-type) utility.**
 
 ### `<Code />`
 


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2077,7 +2077,7 @@ export const onRequest = defineMiddleware(async (context, next) => {
 
 ## Built-in Components
 
-Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components';`.
+Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components';`. **You can reference the `Props` of these components using the [`ComponentProp` type](/en/guides/typescript/#componentprops-type) utility.**
```
sarah11918, line 2080:
```suggestion
Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components';`.

You can reference the `Props` of these components using the [`ComponentProp` type](/en/guides/typescript/#componentprops-type) utility.
```

sarah11918 (commented):
Thank you, @ManorSailor ! My intention was not to make the text itself bold, but to show the added sentence. Yes, you're right, the text should not be bold. :smile: I'll remove the bold and then we're good to go!  :raised_hands:

PR #9181, opened 2024-08-23
Suggest using import.meta.glob

matthewp:
#### Description (required)

Replaces usage of `Astro.glob` with `import.meta.glob` instead.


src/content/docs/en/guides/imports.mdx
```diff
@@ -183,16 +183,17 @@ import logoUrl from '@assets/logo.png?url';
 These aliases are also integrated automatically into [VS Code](https://code.visualstudio.com/docs/languages/jsconfig) and other editors.
 
 
-## `Astro.glob()`
+## `import.meta.glob()`
```
matthewp, line 186: I guess there's more documentation on this page than I thought. What do you think about removing all of this except for the `#### Glob Patterns in Astro` section, which lists the limitations and links to the Vite docs?

sarah11918: I'll shorten the section a bit, but we do prefer our own explanation for our own users. So, I think we'd prefer to rework this page, and in fact on all our other non reference pages, link HERE instead of Vite. The Vite page isn't the most user- friendly, so we'll link there from here and from reference. But in our guides, where people are trying to use it in their project and understand how it works, we'll link here. If they want more referency at that point, this will take them to Vite.

sarah11918 (commented):
Thanks for this, @matthewp ! Only two kinds of small changes I've made below:

- Depending on the type of page, some of the links will link to our own Import page entry for `import.meta.glob()`. Only the more reference type pages will link back to Vite directly. This is to give our readers a friendlier explanation, and resources about glob patterns etc. if they're on more "learning" type pages like the guides and tutorial.
- The pages that do link back to vite include "Vite's" in the link so that it's clear readers will be taken off the page when they click, and that this is a "Vite thing" ultimately, so they understand that.

Otherwise, this PR is fine, and for Yan and the rest of Team Docs reading, yes, I am aware this means we will have to update the tutorial repo code.

I also think there are more `Astro.glob` references in docs. Some of them are error messages that need to be fixed at the source. But sometimes we will have written just the words "Astro.glob" e.g. in the in the frontmatter description of the tutorial 5-astro-api/1.mdx

So, I'm happy after committing these to do a bigger sweep, if you prefer! It might be easier for me to do, since I'll know whether I want to use `import.meta.glob()` or "importing with glob patterns" if it shows up somewhere in explanatory text.

sarah11918:
Noting that the main pages updated with new text content will be :+1: 
- imports guide (draft done)
- api reference

Most other pages should be updating examples, links etc.

sarah11918:
OK, @matthewp ! Thank you for your patience!

I went through and caught more instances, updated links etc. The **main** change occurs here, so if you don't mind previewing it to make sure this section seems OK, that would be great:

https://deploy-preview-9181--astro-docs-2.netlify.app/en/guides/imports/#importmetaglob

There is stuff there taken from both the `Astro.glob()` API reference (types) and a helpful pattern from the troubleshooting page. Both of these sections will be removed (Astro.glob eventually, troubleshooting now because it refers to a deprecated error), so it made sense to include the still relevant content all together here.

When that section looks good to you, I'm happy for this to be merged into the beta branch and the whole thing will be generally proofread/polished etc. before launch week, so no need for this to be perfect yet!

matthewp:
@sarah11918 look good to me!

src/content/docs/en/guides/imports.mdx
```diff
@@ -221,6 +232,96 @@ const components = await Astro.glob('../components/*.astro');
...
+
+```astro {6-7}
+---
+// src/components/featured.astro
+const { postSlug } = Astro.props;
+const pathToMyFeaturedPost = `src/pages/blog/${postSlug}.md`;
+
+const posts = await Astro.glob('../pages/blog/*.md');
+const myFeaturedPost = posts.find(post => post.file.includes(pathToMyFeaturedPost));
+---
```
ArmandPhilippot, line 249: From the context, I think we should use `import.meta.glob` instead of `Astro.glob` in this example. However, I'm not sure about the right way to use it.

I created a minimal example locally without adding Typescript (from the prompt) and it seems that, at least in VSCode, we still need to provide a type to `import.meta.glob` to avoid errors (e.g. `Property 'file' does not exist on type 'unknown'.ts(2339)`). I think this is because a `tsconfig.json` file is created in all cases.

So the example should be something like:

```suggestion
---
// src/components/featured.astro
interface Frontmatter {
  title: string;
  description?: string;
}

const { postSlug } = Astro.props;
const pathToMyFeaturedPost = `src/pages/blog/${postSlug}.md`;

const posts = Object.values(
  import.meta.glob<MarkdownInstance<Frontmatter>>("../pages/blog/*.md", { eager: true })
);
const myFeaturedPost = posts.find(post => post.file.includes(pathToMyFeaturedPost));
---
```

We could remove the type to simplify the example but my concern is that some users may copy/paste the code without understanding why their editor is complaining.

However using that updated example makes the next section (`Import type utilities`) a bit awkward since we are more or less repeating the `Markdown files` section example.

So I see three options:
* rewrite the example without Typescript type/interface but add the extra checks for `file`, `url` and `frontmatter.title` (but that might make the example a bit too complex).
* rewrite the example without Typescript type/interface and add a warning to inform users that the example is not complete and that without a Typescript type they'll need additional checks to make sure `file` and the other properties exist to avoid errors in their editor.
* reorganize the section to only use Typescript types and maybe add a note at the end to explain that `import.meta.glob` can be used without Typescript types but by doing it this way, the user will have to add additional checks to verify the shape.

I'm not sure which is the best choice here.

ArmandPhilippot (commented):
There is just one problem with one of the examples. I think it hasn't been updated during the transfer from another page. But I am not sure what the best approach is here.
Other than that, it looks good to me!

src/content/docs/en/guides/imports.mdx
sarah11918: I think it's fine to not use TypeScript in this example, as many of our examples currently don't. This example is more to illustrate the point of over searching and filtering, and it's not likely they'll copy this code because the variables won't be of much use to them.

However, I think I will probably remove it anyway for Reasons:tm: and I believe we show something similar in content collections about filtering a full collection fetch to only get *some* results.

I'm not sure how the `Astro.glob` missed the find and replace here though! I thought I had caught them all!

src/content/docs/en/guides/imports.mdx
```diff
@@ -221,6 +232,96 @@ const components = await Astro.glob('../components/*.astro');
...
+
+A common workaround is to instead import a larger set of files that includes all the files you need, then filter them:
+
+```astro {6-7}
+---
+// src/components/featured.astro
+const { postSlug } = Astro.props;
+const pathToMyFeaturedPost = `src/pages/blog/${postSlug}.md`;
+
+const posts = await Astro.glob('../pages/blog/*.md');
```
sarah11918, line 247:
```suggestion
const posts = await Object.values(await import.meta.glob("../pages/blog/*.md", { eager: true }));
```

src/content/docs/en/guides/imports.mdx
sarah11918: I committed the quick fix to remove Astro.glob, but ideally I'd also *like* for this example to maybe be fetching some images (like maybe using our dynamic imported images recipe?) and filtering for only "cat photos" or something?

Would you be interested in/willing to whip up an example here that isn't about fetching markdown files, but maybe images @ArmandPhilippot ? Again, I don't think TypeScript is necessary *here* to just show the idea of fetching more, then filtering.

ArmandPhilippot: @sarah11918 I can give it a try! 🫡  Since there is already an existing base, it will help me (never used `import.meta.glob` before)

sarah11918: Thank you! Again, I'm only looking really for a line showing an import, then a line that makes a new array of just a subset somehow. It doesn't need to be fancy, just illustrative of what we mean when we say "import more than you'll need, and narrow it down"  :smile:

src/content/docs/en/guides/imports.mdx
```diff
@@ -221,6 +232,96 @@ const components = await Astro.glob('../components/*.astro');
...
+const pathToMyFeaturedPost = `src/pages/blog/${postSlug}.md`;
+
+const posts = await Object.values(await import.meta.glob("../pages/blog/*.md", { eager: true }));
+const myFeaturedPost = posts.find(post => post.file.includes(pathToMyFeaturedPost));
+---
+
+<p>
+  Take a look at my favorite post, <a href={myFeaturedPost.url}>{myFeaturedPost.frontmatter.title}</a>!
+</p>
+```
```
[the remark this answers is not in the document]
sarah11918: OOh, something like that yes! I was thinking smaller scoped, like maybe:

```
const images = Object.values(import.meta.glob("/src/assets/*.{jpeg,jpg,png,gif}", { eager: true }));
const catImages = images.filter((img) => img.default.src.includes("cat"));
```

To just illustrate filtering after the fact vs trying to do an import based on a more specific search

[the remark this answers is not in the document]
ArmandPhilippot: I'm not sure what you mean by "smaller scoped". Without a full example (just the two lines) or a separated example?
````suggestion
```astro {6-7}
---
// src/components/featured.astro
const { postSlug } = Astro.props;
const pathToMyFeaturedPost = `src/pages/blog/${postSlug}.md`;

const posts = Object.values(import.meta.glob("../pages/blog/*.md", { eager: true }));
const myFeaturedPost = posts.find(post => post.file.includes(pathToMyFeaturedPost));
---

<p>
  Take a look at my favorite post, <a href={myFeaturedPost.url}>{myFeaturedPost.frontmatter.title}</a>!
</p>
```

The `import.meta.glob()` utility can be used with other file types, like images. The following example illustrates how to filter an array of imported images and then use only a subset:

```astro {6-7}
---
// src/components/cat-gallery.astro
const images = Object.values(import.meta.glob("/src/assets/*.{jpeg,jpg,png,gif}", { eager: true }));
const catImages = images.filter((img) => img.default.src.includes("cat"));
---

<p>Or check out my own collection of cat pictures:</p>
<ul>
  {
    catImages.map((img) => {
      const { height, src, width } = img.default;
      return (
        <li>
          <img height={height} src={src} width={width} />
        </li>
      );
    })
  }
</ul>
````

What I have done here:
* I put the cat gallery in a separate example with a paragraph to introduce this new example
* In the first example (`featured.astro`), I removed the two `await`. I haven't tested but Typescript tells me they are useless here (`'await' has no effect on the type of this expression.ts(80007)`). Also, according to [Vite documentation](https://vitejs.dev/guide/features#glob-import), when we use `eager` we don't get a `Promise`.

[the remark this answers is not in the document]
sarah11918: Alright, I appreciate the leg work here Armand. I chatted a bit with Team DX about this and after all this work, I think we're going to leave the original example for now! :sweat_smile:  I'm sorry for the extra work, but at least I wasn't keeping you from any French translations... :laughing:

sarah11918 (approved):
OK! I think we got everything worked out on this one, so merging in to the beta docs!

sarah11918:
!coauthor

PR #9191, opened 2024-08-25
docs: a code example is wrong

nadeemkhanrtm1:

- Closes #
- Suggested label:


TheOtterlord (commented):
Thanks for the PR! The tutorial is written following the [minimal](https://github.com/withastro/astro/tree/main/examples/minimal) (empty) template. The [`index.astro`](https://github.com/withastro/astro/blob/main/examples/minimal/src/pages/index.astro) file does use "Astro" has the heading, so it's correct as written.

PR #9200, opened 2024-08-26
feat: update `paginate()`

ematipico:
#### Description (required)

For Astro 5.0, it updates the docs for `paginate()`, and it adds docs for `page.url.last` and `page.url.first`.


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -989,23 +989,40 @@ The total number of pages.
...
+Get the URL of the previous page (will be `undefined` if on page 1). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.
 
 ##### `page.url.next`
 
 <p>
 **Type:** `string | undefined`
 </p>
 
-Get the URL of the next page (will be `undefined` if no more pages). If a value is set for [`base`](/en/reference/configuration-reference/#base), prepend the base path to the URL.
+Get the URL of the next page (will be `undefined` if no more pages). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.
```
delucis, line 1008: This was already the case in the existing docs, but taking advantage of this PR to tweak this and make it consistent with the others:

```suggestion
Get the URL of the next page (will be `undefined` if on the last page). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -989,23 +989,40 @@ The total number of pages.
...
+Get the URL of the first page (will be `undefined` if at the first page). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.
+
+##### `page.url.last`
+
+<p>
+  **Type:** `string`
+</p>
+
+Get the URL of the last page  (will be `undefined` if at the last page). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.
+
```
delucis, line 1025: I think “on” the page is probably more common than “at”:

```suggestion
##### `page.url.first`

<p>
  **Type:** `string`
</p>

Get the URL of the first page (will be `undefined` if on the first page). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.

##### `page.url.last`

<p>
  **Type:** `string`
</p>

Get the URL of the last page  (will be `undefined` if on the last page). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.

```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -989,23 +989,40 @@ The total number of pages.
...
+Get the URL of the current page (useful for canonical URLs). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.
 
 ##### `page.url.prev`
 
 <p>
 **Type:** `string | undefined`
 </p>
 
-Get the URL of the previous page (will be `undefined` if on page 1). If a value is set for [`base`](/en/reference/configuration-reference/#base), prepend the base path to the URL.
+Get the URL of the previous page (will be `undefined` if on page 1). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.
```
delucis, line 1000: And I guess we should probably also update this one for full consistency:

```suggestion
Get the URL of the previous page (will be `undefined` if on the first page). If a value is set for [`base`](/en/reference/configuration-reference/#base), the URL starts with that value.
```

delucis (commented):
Looks great @ematipico! Just left a few language nits.

sarah11918 (commented):
Looks great, thank you for picking up this PR @ematipico !

I'm wondering whether it might be helpful to include in the API reference for these that these are relative URL paths (not absolute URLs?). They didn't specify before, and maybe it's assumed because that's how we *use* links within a project. and we do say now "starts with the base" so I guess that's understood?

But, this PR is great as is!

Did you happen to check e.g. the routing page or elsewhere for code examples that use pagination, in case they might need updating? One of us will have to before merging this, but it's fine if it's me! :smile:  Just wondering whether you already did!

ematipico:
> Looks great, thank you for picking up this PR @ematipico !
> 
> I'm wondering whether it might be helpful to include in the API reference for these that these are relative URL paths (not absolute URLs?). They didn't specify before, and maybe it's assumed because that's how we _use_ links within a project. and we do say now "starts with the base" so I guess that's understood?
> 

Yeah, I think it's worth mentioning it, especially because we have some other APIs e.g. `Astro.url` that return an absolute URL. 

> But, this PR is great as is!
> 
> Did you happen to check e.g. the routing page or elsewhere for code examples that use pagination, in case they might need updating? One of us will have to before merging this, but it's fine if it's me! 😄 Just wondering whether you already did!

I checked the whole project, and I found only that guide where we show how to use paginate

sarah11918 (approved):
Approving, and making myself a note for `paginate()` to go in and specify that the URL includes `base` at some later point! Thank you both @ematipico and @kevinzunigacuellar for this feature!

PR #9218, opened 2024-08-28
feat(next): better tsconfig docs

florian-lefebvre:
#### Description (required)

Updates our tsconfig recommendations

- Closes #
- Suggested label: 5.0


sarah11918 (commented):
Looks good! Just a couple of little comments! Thanks @florian-lefebvre !

sarah11918 (approved):
Approving this!

sarah11918:
@florian-lefebvre I don't think this was released in alpha.3, so I'll let you decide whether you want to wait to merge this for whichever release it goes out in, or just do it now.

FYI, on the breaking change guides, I have been updating to show the alpha release number so that if people are looking at the upgrade guide, they can see which ones have or haven't been actually released yet and are just placeholders. So that way it's fine if this merges early.

PR #9255, opened 2024-08-29
Server islands

matthewp:
#### Description (required)

This includes:

- Updates to the directives reference page.
- Server islands guide
  - Note that this guide could also be a section of the Astro components page. It's longer than any section on that page, however. So it could also be linked to from that page and that's how you get to it.


sarah11918:
OK @matthewp ! I've been going through this closely and I *think* this could work well on the regular components page as an entry like "HTML components". I sketched out what I think that could look like vs editing your own page of content for now. See what you think if these were the "docs" for it: https://deploy-preview-9255--astro-docs-2.netlify.app/en/basics/astro-components/#server-island-components


I think it fits in there nicely, isn't *too* long since it's focused on basic usage, and also makes it easy to find/see right with other stuff about how components work.

(Since server islands are pretty easy to *do*, people probably won't be coming back to visit and look up things very much, so probably doesn't need an entire page in the sidebar? It's just a regular component with a directive. Since it uses slots and fallback content, it kind of does feel at home here.)

I'm not opposed to its own guide page, though! I felt I could get this in as part of our "basics" right in components so wanted to try that first. And, I kind of think that the interesting stuff is more the concept/effects, and so maybe you'd like a page that is written more like the blog post in the Concepts section, alongside Island architecture?

When I looked at what you wrote on the separate page, and in the 4.12 blog post, I think there's a lot there that feels more like a companion page to the Astro Islands. I think there's an argument to have a similar, conceptual, Server Islands page *there* that explains the benefits, performance etc. that we don't really get into in docs proper most of the time.

Anyway, see if you think this feels OK on the components page or not. I think that's a good place for it, as a first class docs citizen! And, it feels natural to update the existing example on that page of "what all you can do in the template" like this: 
image

src/content/docs/en/concepts/islands.mdx
```diff
@@ -84,4 +84,45 @@ Even better, you can tell Astro exactly how and when to render each component. I
...
+Add the [`server:defer` directive](/en/reference/directives-reference/#server-directives) to any component on your page to turn it into its own island:
+
+```astro title="src/pages/index.astro" "server:defer"
+---
+import Avatar from '../components/Avatar.astro';
+---
+<Avatar server:defer />
+```
+
+This allows your page's critical content to be rendered immediately as a shell, while deferring your most dynamic UI elements, such as user avatars or product reviews. This also means that a slower island, such as one connected to a legacy backend, won’t delay the rest of the page's personalized content from being available.
```
matthewp, line 102: Instead of saying "legacy backend", I'd say something about different API endpoints load at different speed and server islands allow them to all load in parallel. I'm afraid that seeing "legacy backend" will make some people think that the feature isn't valuable if you don't have legacy backends.

sarah11918:
```suggestion
This allows your page's critical content to be rendered immediately as a shell, while deferring your most dynamic UI elements, such as user avatars or product reviews. Different API endpoints load at different speeds, and server islands all load individually in parallel. A slower island won’t delay the rest of the page's personalized content from being available.
```

src/content/docs/en/concepts/islands.mdx
```diff
@@ -84,4 +84,45 @@ Even better, you can tell Astro exactly how and when to render each component. I
...
+- Dynamically backed by a database that changes infrequently, but more often than you deploy.
+- Personalized content, tailored to individual users.
+
+Astro's server island implementation happens mostly at build time, where component content is swapped out for a small script. When Astro builds your site, it will omit the component and inject a script in its place, with fallback content will be visible for a short amount of time before the dynamic islands are then loaded.
+
+This rendering pattern was built to be portable. It does not depend on any server infrastructure so it will work with any host you have, from a Node.js server in a Docker container, to the serverless provider of your choice.
+
+### Benefits of server islands
+
+The most obvious benefit of server islands is the ability to render the more highly dynamic parts of your page on the fly. This allows the outer shell and main content to be more aggressively cached, providing faster performance.
```
matthewp, line 116:
```suggestion
Server islands is the ability to render the more highly dynamic parts of your page on the fly, leaving the outer shell and main content to be more aggressively cached, providing faster performance.
```

sarah11918: So the push back I'd give here is that we don't really need to redefine server islands at this point. This section is about the benefits, and it should probably start like the client island section does, which is exactly this "The most obvious benefit".

Instead of removing that, if you had to fill in the blank to finish that (or a similar sentence), what would you say?

src/content/docs/en/concepts/islands.mdx
```diff
@@ -1,12 +1,12 @@
 ---
-title: Astro Islands
-description: "Astro Islands (aka Component Islands) are a pattern of web architecture pioneered by Astro. “Islands architecture” was first coined by Etsy's frontend architect Katie Sylor-Miller in 2019, and expanded on by Preact creator Jason Miller."
+title: Island Architecture
```
yanthomasdev, line 2: Any reason we are moving from Islands Architecture to Island Architecture? I generally feel the former makes more sense since the idea is that you'd generally have many islands in a page, vs. one. Silly, I know, but there's a good amount of blog posts, etc. using the former and it doesn't feel like a good wording break change to make.

src/content/docs/en/concepts/islands.mdx
```diff
@@ -84,4 +95,41 @@ Even better, you can tell Astro exactly how and when to render each component. I
...
 In Astro, it’s up to you as the developer to explicitly tell Astro which components on the page need to also run in the browser. Astro will only hydrate exactly what’s needed on the page and leave the rest of your site as static HTML. 
 
-**Islands are the secret to Astro’s fast-by-default performance story!**
+**Client islands are the secret to Astro’s fast-by-default performance story!**
+
+<ReadMore>Read more about [using JavaScript framework components](/en/guides/frameworkj-components/) in your project.</ReadMore>
+
+## Server islands
+
+Server islands are a way to move expensive or slow server-side code out of the way of the main rendering process, making it easy to combine high performance static HTML and dynamic server-generated components.
```
yanthomasdev, line 104:
```suggestion
Server islands are a way to move expensive or slow server-side code out of the way of the main rendering process, making it easy to combine high-performance static HTML and dynamic server-generated components.
```

src/content/docs/en/concepts/islands.mdx
```diff
@@ -84,4 +95,41 @@ Even better, you can tell Astro exactly how and when to render each component. I
...
+```astro title="src/pages/index.astro" "server:defer"
+---
+import Avatar from '../components/Avatar.astro';
+---
+<Avatar server:defer />
+```
+
+This breaks up your page with smaller areas of server-rendered content that each load in parallel.
+
+Your page's main content can be rendered immediately with placeholder content, such as a generic avatar, until your island's own content is available. With server islands, having small components of personalized content does not delay the rendering of an otherwise static page.
```
yanthomasdev, line 117:
```suggestion
Your page's main content can be rendered immediately with placeholder content, such as a generic avatar until your island's own content is available. With server islands, having small components of personalized content does not delay the rendering of an otherwise static page.
```

src/content/docs/en/concepts/islands.mdx
```diff
@@ -84,4 +95,41 @@ Even better, you can tell Astro exactly how and when to render each component. I
...
+
+This rendering pattern was built to be portable. It does not depend on any server infrastructure so it will work with any host, from a Node.js server in a Docker container, to the serverless provider of your choice.
```
yanthomasdev, line 119:
```suggestion
This rendering pattern was built to be portable. It does not depend on any server infrastructure so it will work with any host, from a Node.js server in a Docker container to the serverless provider of your choice.
```

src/content/docs/en/concepts/islands.mdx
```diff
@@ -84,4 +95,41 @@ Even better, you can tell Astro exactly how and when to render each component. I
...
+
+<h3>Benefits of server islands</h3>
```
yanthomasdev, line 121: I believe we intentionally want to hide this from the ToC? Pointing it out just in case it was an accident.

src/content/docs/en/guides/server-islands.mdx
```diff
@@ -0,0 +1,108 @@
+---
+title: Server islands
+description: Enable more fine-grained control over page load performance with Server islands.
```
yanthomasdev, line 3:
```suggestion
description: Enable more fine-grained control over page load performance with server islands.
```

src/content/docs/en/guides/server-islands.mdx
```diff
@@ -0,0 +1,108 @@
...
+const avatarURL = await getUserAvatar(userSession);
+---
+<img alt="User avatar" src={avatarURL} />
+```
+
+## Server island fallback content
+
+When using the `server:defer` attribute on a component to delay its rendering, you can "slot" in default loading content using the included named `"fallback"` slot.
+
+Your fallback content will be rendered along with the rest of the page initially on page load, and will be replaced with your component's own content when it is available.
```
yanthomasdev, line 41:
```suggestion
Your fallback content will be rendered along with the rest of the page initially on page load and will be replaced with your component's content when available.
```

src/content/docs/en/guides/server-islands.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+Server island implementation happens mostly at build-time where component content is swapped out for a small script.
+
+Each of the islands marked with `server:defer` is split off into its own special route which the script fetches at run time. When Astro builds your site it will omit the component and inject a script in its place, and any content you’ve marked with `slot="fallback"`.
+
+When the page loads in the browser, these components will be requested to a special endpoint that renders them and returns the HTML. This means that users will see the most critical parts of the page instantly. Fallback content will be visible for a short amount of time before the dynamic islands are then loaded.
+
+Each island is loaded independently from the rest. This means a slower island won't delay the rest of your personalized content from being available.
+
+This rendering pattern was built to be portable. It does not depend on any server infrastructure so it will work with any host you have, from a Node.js server in a Docker container, to the serverless provider of your choice.
```
yanthomasdev, line 72:
```suggestion
This rendering pattern was built to be portable. It does not depend on any server infrastructure so it will work with any host you have, from a Node.js server in a Docker container to the serverless provider of your choice.
```

yanthomasdev (commented):
Good work, here's a few of my nits/thoughts 🙌

src/content/docs/en/concepts/islands.mdx
sarah11918: Yes!

src/content/docs/en/concepts/islands.mdx
sarah11918: This might have been a mistake, or unintentional! We should probably revert to the phrase we've always used. We can change it back later if Fred really did intend this change.

PR #9261, opened 2024-08-29
Improve `handler()` reference docs

delucis:
#### Description (required)

This PR improves the `defineAction()` reference docs to better describe the signature of the `handler()` option and mention the `context` argument.

Also took the liberty of removing the extra `<Since>` components on `handler` and `input` sections — felt excessive to add those when they are properties of `defineAction()`, which means they were added at the same time.

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -2512,7 +2512,10 @@ export const server = {
 
 #### `handler()` property
 
+<p>
+
 **Type:** `(input, context) => any`
+</p>
 
 `defineAction()` requires a `handler()` function containing the server logic to run when the action is called. Data returned from the handler is automatically serialized and sent to the caller.
 
@@ -2522,7 +2525,10 @@ Return values are parsed using the [devalue library](https://github.com/Rich-Har
 
 #### `input` validator
 
+<p>
+
 **Type:** `ZodObject | undefined`
+</p>
 
 The optional `input` property accepts a Zod validator to validate handler inputs at runtime. If the action fails to validate, [a `BAD_REQUEST` error](#actionerror) is returned and the `handler` is not called.
 
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -2495,15 +2495,15 @@ Actions help you build a type-safe backend you can call from client code and HTM
 <Since v="4.15.0" />
 </p>
 
-The `defineAction()` utility is used to define new actions from the `src/actions/index.ts` file. This accepts a `handler()` function containing the server logic to run, and an optional `input` property to validate input parameters at runtime.
+The `defineAction()` utility is used to define new actions from the `src/actions/index.ts` file. This accepts a [`handler()`](#handler-property) function containing the server logic to run, and an optional [`input`](#input-validator) property to validate input parameters at runtime.
 
 ```ts
 export const server = {
-    getGreeting: defineAction({
+  getGreeting: defineAction({
     input: z.object({
       name: z.string(),
     }),
-    handler: async (input) => {
+    handler: async (input, context) => {
       return `Hello, ${input.name}!`
     }
   })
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2512,19 +2512,17 @@ export const server = {
 
 #### `handler()` property
 
-<p>
-<Since v="4.15.0" />
-</p>
+**Type:** `(input, context) => any`
+
+`defineAction()` requires a `handler()` function containing the server logic to run when the action is called. This function can return data that is automatically serialized and sent to the caller.
```
sarah11918, line 2517:
```suggestion
`defineAction()` requires a `handler()` function containing the server logic to run when the action is called. This function returns data that is automatically serialized and sent to the caller.
```
Is there a reason it's only "can return"? Can it be more definitive. Or..

> When data is returned, it is automatically serialized and sent to the caller.

sarah11918 (approved):
Amazing, thank you @delucis ! Just one comment for your suggestion, but excellent as is!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2495,15 +2495,15 @@ Actions help you build a type-safe backend you can call from client code and HTM
...
 export const server = {
-    getGreeting: defineAction({
+  getGreeting: defineAction({
     input: z.object({
       name: z.string(),
     }),
-    handler: async (input) => {
+    handler: async (input, context) => {
       return `Hello, ${input.name}!`
     }
```
ArmandPhilippot, line 2508: I think it's a good idea to show that context is also available, but shouldn't we update the example to include context?

This is a rough draft (I haven't taken the time to familiarize myself with Actions yet to find a better example) but maybe something like:
```suggestion
    handler: async (input, context) => {
      if (!input.name) context.redirect();
      return `Hello, ${input.name}!`
    }
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2512,19 +2512,17 @@ export const server = {
 
 #### `handler()` property
 
-<p>
-<Since v="4.15.0" />
-</p>
+**Type:** `(input, context) => any`
```
ArmandPhilippot, line 2515: This is a nitpicking but shouldn't we conserve the surrounding `<p>` tag? If tomorrow, a default `handler` is added having the `<p>` tag in place could encourage whoever updates the documentation to put the default in the same tag instead of adding a new line or adding another `<p>` tag for default. It's just a thought... nothing mandatory (I'm quibbling).

```suggestion
<p>

**Type:** `(input, context) => any`
</p>
```

src/content/docs/en/reference/api-reference.mdx
delucis: Maybe even?

```suggestion
`defineAction()` requires a `handler()` function containing the server logic to run when the action is called. Data returned from the handler is automatically serialized and sent to the caller.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2512,19 +2512,17 @@ export const server = {
...
 
-Return values are parsed using the [devalue library](https://github.com/Rich-Harris/devalue). This supports JSON values, along with instances of `Date()`, `Map()`, `Set()`, or `URL()`.
+Return values are parsed using the [devalue library](https://github.com/Rich-Harris/devalue). This supports JSON values and instances of `Date()`, `Map()`, `Set()`, and `URL()`.
 
 #### `input` validator
 
-<p>
-<Since v="4.15.0" />
-</p>
+**Type:** `ZodObject | undefined`
```
ArmandPhilippot, line 2525: nit: Same as above... maybe i'm quibbling too much 😄 

```suggestion
<p>

**Type:** `ZodObject | undefined`
</p>
```

ArmandPhilippot (commented):
A concern about the example since we now show that the context is available. And some nitpicking (or quibbling... not sure about the right word, sorry).

Other than that, looks good to me! I agree that repeating the Since may not be necessary and could be unhelpful if in another version another parameter is added (it's easier to spot a new parameter if the others don't have Since).

src/content/docs/en/reference/api-reference.mdx
delucis: `context.redirect()` is not available on this version of context 😅 

I also thought about this, but I mainly wanted to show it exists here without muddying the example. I think the reason `context` was excluded in the initial version is that it’s mostly for pretty advanced use cases (for example it allows you to read `context.locals` if you have some middleware), so I think it’s OK not to demonstrate it — just to mention it’s existence.

delucis:
Thanks @ArmandPhilippot — nitpicking and quibbling are both correct! (You could even do both simultaneously 😁)

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: Ah, I did say that I wasn't familiar enough with this feature. 😅 It's good for me so, no need for an overly complex example!

PR #9271, opened 2024-08-30
Update tutorial unit 4 by giving more explanation on h1 style in about.astro page

andrzeij:
…about.astro

#### Description (required)

For tutorial unit 4, it doesn't explain much about the effect of refactoring the about.astro page that it will replaces the local h1 style with h1 style from the BaseLayout. The visual cue is that the existing h1 colour is purple, but by applying the changes suggested in the tutorial will change the h1 colour to black.  

So I am suggesting to add the following note at Line 142:
(Note: using the BaseLayout to render your `about.astro` page means you will lose the style for h1. If you want to keep the h1 style, you need to pass the h1 style to BaseLayout as props via component attribute.)

- Closes #
- Suggested label: 

Yes, first time contributor. Please help me with any suggestion to make it better, thank you.

Discord username: andrzeij


src/content/docs/en/tutorial/4-layouts/1.mdx
```diff
@@ -139,7 +139,7 @@ You still have some Astro components repeatedly rendered on every page. It's tim
 
 ## Try it yourself - Use your layout everywhere
 
-**Refactor** your other pages (`blog.astro` and `about.astro`) so that they use your new `<BaseLayout>` component to render the common page elements.
+**Refactor** your other pages (`blog.astro` and `about.astro`) so that they use your new `<BaseLayout>` component to render the common page elements. (Note: using the BaseLayout to render your `about.astro` page means you will lose the style for h1. If you want to keep the h1 style, you need to pass the h1 style to BaseLayout as props via component attribute.)
```
sarah11918, line 142:
```suggestion
**Refactor** your other pages (`blog.astro` and `about.astro`) so that they use your new `<BaseLayout>` component to render the common page elements. (Note: using the BaseLayout to render your `about.astro` page means you will lose the `<style>` tag added to the `<head>` of this page. If you want to keep the custom `<h1>` style, move the style tag to the body of the page component.)
```
This is probably the solution I'd give at this stage in the tutorial! It is a simpler step, and also demonstrates that this style tag can live in the body of the component.

sarah11918 (commented):
Hi @andrzeij ! Thanks for taking the time to provide feedback on the tutorial! You're right that this is something that people may encounter, and I have a suggestion below for the action I'd suggest. Let me know what you think!

sarah11918 (approved):
Thank you again @andrzeij for contributing this! I just suggested what I think I would suggest to our new users in the tutorial as a remedy that I think is simpler and requires less explanation of how to do it.

We appreciate you taking the time to contribute and are happy to welcome you to Team Docs!  🥳

PR #9273, opened 2024-08-30
feat: integrations now receive a IntegrationRouteData

ematipico:
#### Description (required)

In the code PR, we changed the types received from the integrations. From `RouteData` to `IntegrationRouteData`. 

`IntegrationRouteData` has less information compared to `RouteData`.

However, I noticed that the docs of `RouteData` were very outdated, so I took the chance to update them with the same JSDocs we have in the code.

I would be great if both docs and code have the same description (more or less), so please lend an hand in the code PR too :)


src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -582,3 +582,67 @@ Know a good resource for Astro v4.0? [Edit this page](https://github.com/withast
...
+```diff
+-import type {RouteData} from 'astro';
++import type {IntegrationRouteData} from "astro"
+
+-function useRoute(route: RouteData) {
++function useRoute(route: IntegrationRouteData) {
+
+}
+```
+
```
sarah11918, line 604:
```suggestion

Please see the [API reference for IntegrationRouteData]() for full details of the fields now available.

```
Seems like this would be a good place to link to the reference so that people can check this out if they need more details!

src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -582,3 +582,67 @@ Know a good resource for Astro v4.0? [Edit this page](https://github.com/withast
 ## Known Issues
 
 Please check [Astro's issues on GitHub](https://github.com/withastro/astro/issues/) for any reported issues, or to file an issue yourself.
+
+
+### [changed]: `entryPoint` type inside the hook `astro:build:ssr`
+In Astro v4.x, the `entryPoint` type was `RouteData`.
```
sarah11918, line 588: I'm looking at the existing docs, and I can only find an entry for `entryPoints` (plural) in that hook -- is this that one? Is one of them wrong?

src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -582,3 +582,67 @@ Know a good resource for Astro v4.0? [Edit this page](https://github.com/withast
...
+```diff
+-import type {RouteData} from 'astro';
++import type {IntegrationRouteData} from "astro"
+
+-function useRoute(route: RouteData) {
++function useRoute(route: IntegrationRouteData) {
+
+}
+```
+
```
sarah11918, line 627:
```suggestion
```
Since the advice here is the same, it works for both astro and the vercel adapter to only mention it once, I think!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,12 +613,19 @@ You can reference the full `RouteData` type below, but the most common propertie
...
-interface RouteData {
-  /** Whether a given route is an HTML page or non-HTML endpoint */
-  type: 'page' | 'endpoint';
+interface IntegrationRouteData {
+	/**
+	 * The type of the route. It can be:
+   * 
+	 * - `page`: a route that lives in the file system, usually an Astro component
+	 * - `endpoint`: a route that lives in the file system, usually a JS file that exposes endpoints methods
+	 * - `redirect`: a route points to another route that lives in the file system
```
sarah11918, line 625:
```suggestion
	 * - `redirect`: a route that points to another route that lives in the file system
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -613,12 +613,19 @@ You can reference the full `RouteData` type below, but the most common propertie
...
+	 * - `fallback`: a route that doesn't exist in the file system that needs to be handled with other means, usually the middleware
```
sarah11918, line 626:
```suggestion
	 * - `fallback`: a route that doesn't exist in the file system and needs to be handled with other means, usually middleware
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -644,10 +651,37 @@ interface RouteData {
    */
   segments: { content: string; dynamic: boolean; spread: boolean; }[][];
   /**
-   * Function to render component in-place from a set of input data.
+	 * @param {any} data The optional parameters of the route
+	 * 
+	 * @description
```
sarah11918, line 656:
```suggestion
```
I think following the pattern here inside docs, we wouldn't have these lines!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -644,10 +651,37 @@ interface RouteData {
...
+	 * 
+	 * ```js
+	 * console.log(generate({ id: 'presentation' })) // will log `/blog/presentation` 
+	 * ```
    * This is typically for internal use, so call with caution!
-   */
+	 */
   generate: (data?: any) => string;
+  /**
+	 * Whether the route is prerendered or not
```
sarah11918, line 670:
```suggestion
	 * Whether the route is prerendered or not.
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -644,10 +651,37 @@ interface RouteData {
...
+	 */
+	prerender: boolean;
+  /**
+	 * The paths of the physical files emitted by this route. When a route **isn't** prerendered, the value is either `undefined` or an empty array.
```
sarah11918, line 674: Is "physical" the right adjective here, since files are not exactly concrete? If I forced you to choose another adjective, which one would you pick? :smile:

sarah11918 (commented):
This does look SOOOO much better ,thank you @ematipico !

I had a couple of questions and suggestions below just for tidying. I can also be in the astro PR to make sure these align!

src/content/docs/en/guides/upgrade-to/v5.mdx
ematipico: That's a typo, yes

src/content/docs/en/reference/integrations-reference.mdx
ematipico: Maybe we can rephrase it like this "The paths of the HTML files emitted by this route"

sarah11918 (approved):
This looks ready to go! We already had content in the upgrade guide for the breaking changes, so only the API reference here needs to be added!

PR #9301, opened 2024-09-04
feat(next): context.isPrerendered

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -1140,7 +1140,7 @@ You can use this property to understand which route is rendering your component.
 
 See also: [`Astro.routePattern`](#astroroutepattern)
 
-### `Astro.isPrerendered`
+### `context.isPrerendered`
 
 <p>
 
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -820,6 +820,20 @@ You can use this property to understand which route is rendering your component.
 
 See also: [`context.routePattern`](#contextroutepattern)
 
+### `Astro.isPrerendered`
+
+<p>
+
+**Type**: `string`<br />
+<Since v="5.0.0" />
+</p>
+
+A boolean representing whether or not the current page is prerendered.
+
+You can use this property to run conditional logic in middlewares, eg. to avoid accessing headers in prerendered pages.
+
+See also: [`context.isPrerendered`](#contextisprerendered)
+
 ## Endpoint Context
 
 [Endpoint functions](/en/guides/endpoints/) receive a context object as the first parameter. It mirrors many of the `Astro` global properties.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1126,6 +1140,20 @@ You can use this property to understand which route is rendering your component.
 
 See also: [`Astro.routePattern`](#astroroutepattern)
 
+### `Astro.isPrerendered`
```
ArmandPhilippot, line 1143: Except this:
```suggestion
### `context.isPrerendered`
```
Looks good to me!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -820,6 +820,20 @@ You can use this property to understand which route is rendering your component.
 
 See also: [`context.routePattern`](#contextroutepattern)
 
+### `Astro.isPrerendered`
+
+<p>
+
+**Type**: `string`<br />
```
sarah11918, line 827: Are these strings or boolean?

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -820,6 +820,20 @@ You can use this property to understand which route is rendering your component.
...
+<Since v="5.0.0" />
+</p>
+
+A boolean representing whether or not the current page is prerendered.
+
+You can use this property to run conditional logic in middlewares, eg. to avoid accessing headers in prerendered pages.
```
sarah11918, line 833:
```suggestion
You can use this property to run conditional logic in middlewares, for example, to avoid accessing headers in prerendered pages.
```
I'm not sure about the language of "avoid accessing headers".... that makes it sound like accessing headers a bad thing and you want to avoid it at all costs. Is that true? Or, is it that you *can't* access headers in prerendered pages? Is this a helpful alternative for something you can't otherwise do?

Just trying to get a sense of how this is useful/helpful to know how best to phrase it!

sarah11918 (commented):
A couple of quick questions, @florian-lefebvre !

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: Good catch, I read too quickly 😓 ([`boolean`](https://github.com/withastro/astro/blob/2fe97bc4fd855ca559664c578ace3f410134eae5/packages/astro/src/types/public/context.ts#L346)).

florian-lefebvre: copy pasting...my bad!

```suggestion
**Type**: `boolean`<br />
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1126,6 +1140,20 @@ You can use this property to understand which route is rendering your component.
 
 See also: [`Astro.routePattern`](#astroroutepattern)
 
+### `context.isPrerendered`
+
+<p>
+
+**Type**: `string`<br />
```
florian-lefebvre, line 1147:
```suggestion
**Type**: `boolean`<br />
```

src/content/docs/en/reference/api-reference.mdx
florian-lefebvre: Not sure how to phrase this! An example: it's useless to check if a user is authenticated (eg. using a cookie which uses headers) if you generate a static og image. Also, I found out today that stuff can break if you use headers in this situation so maybe "avoid" is the right term

sarah11918: OK, then I'm happy to go with this!

sarah11918 (approved):
Thank you @florian-lefebvre ! Happy with this one! :raised_hands:

PR #9308, opened 2024-09-05
Update docs for advanced APIs

Fryuni:
#### Description

- Updates the docs for the programmatic APIs for Astro commands `dev`, `build`, `preview` and `sync`.
- Document `DevServer`, the return type of `dev`
- Document `PreviewServer`, the return type of `preview`

#### Related issues & labels

- Found those mistakes and missing docs while working on the docs for https://github.com/Fryuni/inox-tools/pull/158

PR #9310, opened 2024-09-05
i18n(fr): update `cli-reference.mdx`

ArmandPhilippot:
#### Description (required)

* Add changes from #9308 to French translation of `cli-reference.mdx`
* Add missing space before colon for `Type` and `Default` in French translation (I also replaced `Défaut` with `Par défaut` because it seems more correct to me in French... and this is the syntax we use most often it seems).

- Suggested label: i18n


thomasbnt (approved):
LGTM!

yanthomasdev (approved):
LGTM! Good work!

PR #9314, opened 2024-09-05
use `connectedCallback` instead of `constructor`

mayank99:
#### Description (required)

Changed `constructor` to `connectedCallback`. It's less code and more correct.

The [custom elements spec](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-conformance) explicitly recommends implementing setup logic in `connectedCallback` rather than the constructor. See also: https://nolanlawson.com/2024/01/13/web-component-gotcha-constructor-vs-connectedcallback/


delucis (approved):
Thanks @mayank99! Appreciate you 💜

PR #9337, opened 2024-09-07
[mdx] Adds usage information, taken out of the Markdown guide

sarah11918:
#### Description (required)

Separates some content out from the Markdown page and puts it on the MDX page and the troubleshooting page.

I can't make the corresponding Markdown page changes here/yet because that's a lot of link resolving that I've already fixed on the beta docs branch. :smile:  

So that means this content will be duplicated until I can resolve them, but I think that's fine. It will give translators a chance to translate these early, and when the Markdown page is updated with content removed, it won't be a problem.

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/mdx.mdx
+++ b/src/content/docs/en/guides/integrations-guide/mdx.mdx
@@ -96,7 +96,7 @@ It also adds extra features to standard MDX, including support for Markdown-styl
 
 ### Using Exported Variables in MDX
 
-MDX supports using `export` statements to add variables to your MDX content.
+MDX supports using `export` statements to add variables to your MDX content or to export data to a component that imports it.
 
 For example, you can export a `title` field from an MDX page or component to use as a heading with `{JSX expressions}`:
 
@@ -106,19 +106,11 @@ export const title = 'My first MDX post'
 # {title}
 ```
 
-MDX files can also use an `export` statement to export data to a component that imports it.
+Or you can use that exported `title` in your page using `import` and `import.meta.glob()` statements:
 
-For example, you can export a `title` field from an MDX page or component.
-
-```mdx title="/src/pages/posts/post-1.mdx"
-export const title = 'My first MDX post'
-```
-
-This `title` will be accessible from `import` and `import.meta.glob()` statements:
-
-```astro title="// src/pages/index.astro"
+```astro title="src/pages/index.astro"
 ---
-const matches = await import.meta.glob('./*.mdx', { eager: true });
+const matches = await import.meta.glob('./posts/*.mdx', { eager: true });
 const posts = Object.values(posts);
 ---
 
@@ -129,7 +121,7 @@ const posts = Object.values(posts);
 
 The following properties are available to a `.astro` component when using an `import` statement or `import.meta.glob()`:
 
-- **`file`** - The absolute file path (e.g. `/home/user/projects/.../file.md`).
+- **`file`** - The absolute file path (e.g. `/home/user/projects/.../file.mdx`).
 - **`url`** - The URL of the page (e.g. `/en/guides/markdown-content`).
 - **`frontmatter`** - Contains any data specified in the file’s YAML frontmatter.
 - **`getHeadings()`** - An async function that returns an array of all headings (`<h1>` to `<h6>`) in the file with the type: `{ depth: number; slug: string; text: string }[]`. Each heading’s `slug` corresponds to the generated ID for a given heading and can be used for anchor links.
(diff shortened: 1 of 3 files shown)


src/content/docs/en/guides/integrations-guide/mdx.mdx
```diff
@@ -84,13 +84,136 @@ For other editors, use the [MDX language server](https://github.com/mdx-js/mdx-a
...
+
+For example, you can export a `title` field from an MDX page or component.
+
+```mdx title="/src/pages/posts/post-1.mdx"
+export const title = 'My first MDX post'
+```
+
+This `title` will be accessible from `import` and `import.meta.glob()` statements:
+
+```astro title="// src/pages/index.astro"
```
ArmandPhilippot, line 119:
````suggestion
```astro title="/src/pages/index.astro"
````

File: src/content/docs/en/guides/integrations-guide/mdx.mdx [hunk 2]
```diff
@@ -84,13 +84,136 @@ For other editors, use the [MDX language server](https://github.com/mdx-js/mdx-a
...
+---
+
+{posts.map(post => <p>{post.title}</p>)}
+```
+
+#### Exported Properties
+
+The following properties are available to a `.astro` component when using an `import` statement or `import.meta.glob()`:
+
+- **`file`** - The absolute file path (e.g. `/home/user/projects/.../file.md`).
```
ArmandPhilippot, line 132: nit: Since this is the MDX guide, I think we should replace the extension:
```suggestion
- **`file`** - The absolute file path (e.g. `/home/user/projects/.../file.mdx`).
```

src/content/docs/en/guides/integrations-guide/mdx.mdx
```diff
@@ -84,13 +84,136 @@ For other editors, use the [MDX language server](https://github.com/mdx-js/mdx-a
...
+- **`url`** - The URL of the page (e.g. `/en/guides/markdown-content`).
+- **`frontmatter`** - Contains any data specified in the file’s YAML frontmatter.
+- **`getHeadings()`** - An async function that returns an array of all headings (`<h1>` to `<h6>`) in the file with the type: `{ depth: number; slug: string; text: string }[]`. Each heading’s `slug` corresponds to the generated ID for a given heading and can be used for anchor links.
+- **`<Content />`** - A component that returns the full, rendered contents of the file.
+- **(any `export` value)** - MDX files can also export data with an `export` statement.
+
+### Using Frontmatter Variables in MDX
+
+The Astro MDX integration includes support for using frontmatter in MDX by default. Add frontmatter properties just as you would in Markdown files, and these variables are availble to use in the template, and as named properties when importing the file somewhere else.
```
ArmandPhilippot, line 141:
```suggestion
The Astro MDX integration includes support for using frontmatter in MDX by default. Add frontmatter properties just as you would in Markdown files, and these variables are available to use in the template, and as named properties when importing the file somewhere else.
```

src/content/docs/en/guides/integrations-guide/mdx.mdx
```diff
@@ -84,13 +84,136 @@ For other editors, use the [MDX language server](https://github.com/mdx-js/mdx-a
...
+
+MDX files can also use an `export` statement to export data to a component that imports it.
+
+For example, you can export a `title` field from an MDX page or component.
+
+```mdx title="/src/pages/posts/post-1.mdx"
+export const title = 'My first MDX post'
+```
+
+This `title` will be accessible from `import` and `import.meta.glob()` statements:
```
ArmandPhilippot, line 117: When I first saw the two code blocks with `export const title` and "*For example, you can export a `title` field from an MDX page or component*" I thought you forgot to remove one. After rereading, I understood the intention but I wonder if it can't be simplified to avoid duplication:
````suggestion
MDX supports using `export` statements to add variables to your MDX content or to export data to a component that imports it.

For example, you can export a `title` field from an MDX page or component to use as a heading with `{JSX expressions}`:

```mdx title="/src/pages/posts/post-1.mdx"
export const title = 'My first MDX post'

# {title}
```

Or you can use that exported `title` in your page using `import` and `import.meta.glob()` statements:
````
But I'm used to MDX, so my vision may be biased... Your approach may be better for a beginner. I'll let you judge!

**Review by ArmandPhilippot — COMMENTED** [review]
Looks good to me! I just notice two typo and one nitpicking. I also left a suggestion about a section which bothered me at first reading (but I'm not sure my suggestion is better).

File: src/content/docs/en/guides/integrations-guide/mdx.mdx [hunk 5]
```diff
@@ -84,13 +84,136 @@ For other editors, use the [MDX language server](https://github.com/mdx-js/mdx-a
...
+
+```astro title="// src/pages/index.astro"
+---
+const matches = await import.meta.glob('./*.mdx', { eager: true });
```
ArmandPhilippot, line 121: While rereading, I think this should be:
```suggestion
const matches = await import.meta.glob('./posts/*.mdx', { eager: true });
```
Based on the previous code block (with or without my suggestion), the posts are located in `/src/pages/posts/` and not at the root of `pages` directory.

src/content/docs/en/guides/integrations-guide/mdx.mdx
sarah11918: Yes, you're right. This was copied in separate chunks from different places and I didn't smooth the transition very well!

PR #9346, opened 2024-09-08
Use same heading IDs for all languages

Fryuni:
#### Description

- Use the same heading ID as English for all languages.
- Add script to update all heading IDs and links as more pages translations get up-to-date.

This is an experiment brought up over the weekend, we still have to talk about it.


sarah11918:
Hey, I'm sorry to burst your bubble Luiz, and it was a cool experiment. But especially with so many changes coming for Astro 5.0, we're not going to be able to play around with this idea any time soon.

So let's close this PR, and let's instead focus on one of the *goals* behind this idea: to identify translations that are too out of date to be helpful, and maybe should not be published in the version 5 docs if not properly updated.

PR #9374, opened 2024-09-10
Add loader reference to the beta docs

ascorbic:
#### Description (required)

Adds a loader API reference page

- Closes PLT-2514

### DIFF
--- a/src/content/docs/en/reference/loader-reference.mdx
+++ b/src/content/docs/en/reference/loader-reference.mdx
@@ -3,6 +3,371 @@ title: Astro Loader API
 i18nReady: true
 ---
 import Since from '~/components/Since.astro';
-import { FileTree } from '@astrojs/starlight/components';
 
-[publish your loader to NPM as a package](/en/reference/publish-to-npm/) to share with others and be included in our integrations library.
+Astro content collections help you organize and load data into your Astro project. This guide explains how to define a loader, and the API that loaders use to interact with the content layer.
+
+## What is a loader?
+
+Astro loaders allow you load data into [content collections](/en/reference/content-collections/), which can then be used in pages and components. The [built-in `glob()` and `file()` loaders](/en/guides/defining-collections/#built-in-loaders) are used to load content from the file system, and you can create your own loaders to load content from other sources. You can define a loader inline in your project's `src/content/config.ts` file, share one loader between multiple collections, or even [publish your loader to NPM as a package](/en/reference/publish-to-npm/) to share with others and be included in our integrations library.
+
+## Loader types
+
+Loaders can be defined either as a simple function that returns an array of entries, or with the more powerful object loader API.
+
+### Inline loaders
+An inline loader is an async function that returns an array or object containing entries. Use this for simple loaders, particularly those that are defined inline in the `src/content/config.ts` file. The function can be async, and must return either an array of entries that each contain a unique `id` field, or an object where each key is a unique ID and each value is the entry. Whenever the loader is invoked, it will clear the store and reload all the entries. 
+
+```ts
+const countries = defineCollection({
+  loader: async () => {
+    const response = await fetch("https://restcountries.com/v3.1/all");
+    const data = await response.json();
+    // Must return an array of entries with an id property
+    // or an object with IDs as keys and entries as values
+    return data.map((country) => ({
+      id: country.cca3,
+      ...country,
+    }));
+  },
+});
+```
+
+### Object loaders
+
+If a loader needs more control over the loading process it can use the object loader API. The loader is defined as an object with a `load()` method that is called at build time to fetch data and update the data store. It allows entries to be updated incrementally, or for the store to be cleared only when necessary. It can also define a schema for the entries, which can be used to validate the data and generate static types. 
+
+The recommended pattern is to define a function that accepts configuration options and returns the loader object, in the same way that you would normally define an Astro integration or Vite plugin. 
+
+
+```ts title=loader.ts
+
+import type { Loader, LoaderContext } from 'astro/loaders';
+
+// Define any options that the loader needs
+export function myLoader(options: { url: string }): Loader {
+  // Configure the loader 
+  const feedUrl = new URL(options.url);
+  // Return a loader object
+  return {
+    name: "my-loader";
+    // Called when updating the collection.
+    load: async (context: LoaderContext): Promise<void> => {
+      // Load data and update the store
+    }
+    // Optionally, define the schema of an entry. 
+    // It will be overridden by user-defined schema. 
+    schema: async () => z.object({
+      // ...
+    })
+  };
+}
+```
+
+## Loader API
+
+The API for [inline loaders](#inline-loaders) is very simple, and is shown above. This section shows the API for defining an object loader.
+
+### The `Loader` object
+
+The loader object has the following properties:
+
+#### `name`
+**Type**: `string`
+
+A unique name for the loader, used in logs and [for conditional loading](#integration-support).
+
+#### `load`
+**Type**: `(context: LoaderContext) => Promise<void>`
+
+An async function that is called at build time to load data and update the store. See [`LoaderContext`](#loadercontext) for more information.
+
+#### `schema`
+**Type**: `ZodSchema | Promise<ZodSchema> | (() => ZodSchema | Promise<ZodSchema>)`
+
+An optional [Zod schema](/en/guides/defining-collections/#defining-datatypes-with-zod) that defines the shape of the entries. It is used to both validate the data, and also to generate TypeScript types for the collection. If a function is provided, it will be called at build time before `load()` to generate the schema. You can use this to dynamically generate the schema based on the configuration options or by introspecting an API. 
+
+### `LoaderContext`
+This object is passed to the `load()` method of the loader, and contains the following properties:
+
+#### `collection`
+**Type**: `string`
+
+The unique name of the collection. This is the key in the `collections` object in the `src/content/config.ts` file.
+
+#### `store`
+**Type**: [`DataStore`](#datastore)
+
+A database to store the actual data. Use this to update the store with new entries. See [`DataStore`](#datastore) for more information.
+
+#### `meta`
+**Type**: `MetaStore`
+
+A simple KV store, designed for things like sync tokens and last-modified times. Use this to store metadata about the collection. The data is persisted between builds, but is only available inside the loader.
+
+```ts
+const lastModified = meta.get("lastModified");
+// ...
+meta.set("lastModified", new Date().toISOString());
+```
+
+#### `logger`
+**Type**: [`AstroIntegrationLogger`](/en/reference/integrations-reference/#astrointegrationlogger)
+
+A logger that can be used to log messages to the console. It is recommended to use this instead of `console.log` to ensure that logs are displayed correctly in the Astro UI. It will include the loader name in the log message. See [`AstroIntegrationLogger`](/en/reference/integrations-reference/#astrointegrationlogger) for more information.
+
+#### `config`
+**Type**: `AstroConfig`
+
+The full, resolved Astro configuration object with all defaults applied. See [the configuration reference](/en/reference/configuration-reference/) for more information.
+
+#### `parseData`
+**Type**: `(props: ParseDataOptions<TData>) => Promise<TData>`
+
+Validates and parses the data according to the collection schema. Pass data to this function to validate and parse it before storing it in the data store.
+
+```ts title=loader.ts {14-17}
+import type { Loader } from "astro/loaders";
+import { loadFeed } from "./feed.js";
+
+export function feedLoader({ url }): Loader {
+  const feedUrl = new URL(url);
+  return {
+    name: "feed-loader",
+    load: async ({ store, logger, parseData, meta, generateDigest }) => {
+      logger.info("Loading posts");
+      const feed = loadFeed(feedUrl);
+      store.clear();
+
+      for (const item of feed.items) {
+        const data = await parseData({
+          id: item.guid,
+          data: item,
+        });
+        store.set({
+          id,
+          data,
+        });
+      }
+    },
+  };
+}
+```
+
+#### `generateDigest`
+**Type**: `(data: Record<string, unknown> | string) => string`
+
+Generates a non-cryptographic content digest. This can be used to check if the data has changed. 
+
+```ts title=loader.ts {19,24}
+import type { Loader } from "astro/loaders";
+import { loadFeed } from "./feed.js";
+
+export function feedLoader({ url }): Loader {
+  const feedUrl = new URL(url);
+  return {
+    name: "feed-loader",
+    load: async ({ store, logger, parseData, meta, generateDigest }) => {
+      logger.info("Loading posts");
+      const feed = loadFeed(feedUrl);
+      store.clear();
+
+      for (const item of feed.items) {
+        const data = await parseData({
+          id: item.guid,
+          data: item,
+        });
+
+        const digest = generateDigest(data);
+
+        store.set({
+          id,
+          data,
+          digest,
+        });
+      }
+    },
+  };
+}
+```
+
+#### `watcher`
+**Type**: `FSWatcher`
+
+When running in dev mode, this is a filesystem watcher that can be used to trigger updates. See [`ViteDevServer`](https://vitejs.dev/guide/api-javascript.html#vitedevserver) for more information.
+
+```ts title="Extract from the file() loader" {8-13}
+return {
+  name: 'file-loader',
+  load: async ({ config, store, watcher }) => {
+    const url = new URL(fileName, config.root);
+    const filePath = fileURLToPath(url);
+    await syncData(filePath, store);
+
+    watcher?.on('change', async (changedPath) => {
+      if (changedPath === filePath) {
+        logger.info(`Reloading data from ${fileName}`);
+        await syncData(filePath, store);
+      }
+    });
+  },
+};
+```
+#### `refreshContextData`
+**Type**: `Record<string, unknown>`
+
+If the loader has been triggered by an integration, this may optionally contain extra data set by that integration. It is only set when the loader is triggered by an integration. See [integration support](#integration-support) for more information.
+
+```ts title=loader.ts {5-8}
+export function myLoader(options: { url: string }): Loader {
+  return {
+    name: "my-loader",
+    load: async ({ refreshContextData, store, logger }) => {
+      if(refreshContextData?.webhookBody) {
+        logger.info("Webhook triggered with body");
+        processWebhook(store, refreshContextData.webhookBody);
+      }
+      // ...
+    },
+  };
+}
+```
+
+### `DataStore`
+The data store is a loader's interface to the content collection data. It is a key value store, scoped to the collection, so a loader can only access the data for its own collection.
+
+#### `get`
+**Type**: `(key: string) => DataEntry | undefined`
+
+Get an entry from the store by its ID. Returns `undefined` if the entry does not exist. 
+
+```ts
+const existingEntry = store.get("my-entry");
+```
+
+The returned object has the following properties:
+```ts
+interface DataEntry<TData extends Record<string, unknown>> {
+	/** The ID of the entry. Unique per collection. */
+	id: string;
+	/** The parsed entry data */
+	data: TData;
+	/** The file path of the content, if applicable. Relative to the site root. */
+	filePath?: string;
+	/** The raw body of the content, if applicable. */
+	body?: string;
+	/** An optional content digest, to check if the content has changed. */
+	digest?: number | string;
+	/** The rendered content of the entry, if applicable. */
+	rendered?: RenderedContent;
+}
+```
+
+#### `set`
+
+**Type**: `(opts: Opts) => boolean`
+
+The main method to update the store with new entries. It takes an object with the following properties:
+
+- `id: string`: The ID of the entry. Must be unique per collection. If the ID already exists in the store, the entry will be replaced.
+- `data`: The data to store. This will be stored unchanged, so if it needs to be parsed or validated then it should be done before calling `set()`.
+- `body?: string`: The raw body of the content, if applicable. This is mostly relevant for file-based loaders, or those that also store rendered content.
+- `filePath?: string`: The file path of the content, if applicable. This only applies to file-based loaders, and is used to resolve paths such as images or other assets.
+- `digest?: string`: An optional content digest, to check if the content has changed. It is an opaque value, so you can use any method to generate it, but we recommended using the `generateDigest()` function. If provided it will be checked against the existing entry, and if it matches then the entry will not be updated and `set` will return false.
+- `rendered?: { html?: string, metadata?: Record<string, unknown> }`: The rendered HTML content of the entry, if applicable. If provided, users will be able to use the `<Content />` component to render the HTML. The metadata object is used for things like frontmatter or other metadata extracted from the content.
+
+Returns `true` if the entry was set. If `digest` was passed and matched an existing entry with the same ID, it will return `false` and not update the entry.
+
+```ts title=loader.ts {7-14}
+    for (const item of feed.items) {
+      const data = await parseData({
+        id: item.guid,
+        data: item,
+      });
+      const digest = generateDigest(data);
+      store.set({
+        id,
+        data,
+        rendered: {
+          html: data.description ?? "",
+        },
+        digest,
+      });
+    }
+```
+
+#### `entries`
+**Type**: `() => Array<[id: string, DataEntry]>`
+
+Get all entries in the collection as an array of key-value pairs.
+
+#### `keys`
+**Type**: `() => Array<string>`
+
+Get all the keys of the entries in the collection.
+
+#### `values`
+**Type**: `() => Array<DataEntry>`
+
+Get all entries in the collection as an array.
+
+#### `delete`
+**Type**: `(key: string) => void`
+
+Delete an entry from the store by its ID.
+
+#### `clear`
+**Type**: `() => void`
+
+Clear all entries from the collection.
+
+#### `has`
+**Type**: `(key: string) => boolean`
+
+Check if an entry exists in the store by its ID.
+
+
+## Integration support
+
+An integration can trigger an update to the content layer during `astro dev` using the `refreshContent` function passed to the [`astro:server:setup`](/en/reference/integrations-reference/#astroserversetup) hook. This can be used, for example, to register a webhook endpoint during dev, or to open a socket to a CMS to listen for changes.
+
+By default, `refreshContent` will refresh all collections. You can optionally pass a `loaders` property, which is an array of loader names. If provided, only collections that use those loaders will be refreshed. For example, A CMS integration could use this property to only refresh its own collections.
+
+You can also pass a `context` object to the loaders. This can be used to pass arbitrary data, such as the webhook body, or an event from the websocket. This data will be available in the loader's `refreshContextData` property.
+
+```ts title=my-integration.ts {19-22}
+ {
+    name: 'my-integration',
+    hooks: {
+        'astro:server:setup': async ({ server, refreshContent }) => {
+            // Register a dev server webhook endpoint
+            server.middlewares.use('/_refresh', async (req, res) => {
+                if(req.method !== 'POST') {
+                  res.statusCode = 405
+                  res.end('Method Not Allowed');
+                  return
+                }
+                let body = '';
+                req.on('data', chunk => {
+                    body += chunk.toString();
+                });
+                req.on('end', async () => {
+                    try {
+                        const webhookBody = JSON.parse(body);
+                        await refreshContent({
+                          context: { webhookBody },
+                          loaders: ['my-loader']
+                        });
+                        res.writeHead(200, { 'Content-Type': 'application/json' });
+                        res.end(JSON.stringify({ message: 'Content refreshed successfully' }));
+                    } catch (error) {
+                        res.writeHead(500, { 'Content-Type': 'application/json' });
+                        res.end(JSON.stringify({ error: 'Failed to refresh content: ' + error.message }));
+                    }
+                });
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -444,12 +444,17 @@ console.log(path) // URL
...
+**Why:** To update Vite server options and middleware, or enable support for refreshing the content layer.
 
 ```js
-'astro:server:setup'?: (options: { server: vite.ViteDevServer }) => void | Promise<void>;
+'astro:server:setup'?: (options: {
+  server: vite.ViteDevServer;
+  refreshContent: {
+    loaders?: Array<string>;
+    context?: Record<string, any>;
+  };
```
Fryuni, line 455: Docs below says this is a function. The example has an `await` so I imagine it is also async.
```suggestion
  refreshContent: (refreshOptions: {
    loaders?: Array<string>;
    context?: Record<string, any>;
  }) => Promise<void>;
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -471,6 +476,53 @@ export default {
...
+                        res.end(JSON.stringify({ error: 'Failed to refresh content: ' + error.message }));
+                    }
+                });
+            });
+        }
+    }
+}
+```
+
+The loader can then access the `refreshContextData` property to get the webhook body. See the [`refreshContextData`](/en/reference/loader-reference/#refreshcontextdata) property for more information.
```
Fryuni, line 523: This seems to be duplicated. The last line before the example says the same thing.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+  const feedUrl = new URL(options.url);
+  // Return a loader object
+  return {
+    name: "my-loader";
+    // Called when updating the collection.
+    load: async (context: LoaderContext): Promise<void> => {
+      // Load data and update the store
+    }
+    // Optionally, define the schema of an entry. 
+    // It will be overridden by user-defined schema.
```
Fryuni, line 58: I might have missed it, but where would a user define a schema?

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+
+#### `load`
+**Type**: `(context: LoaderContext) => Promise<void>`
+
+An async function that is called at build time to load data and update the store. See [`LoaderContext`](#loadercontext) for more information.
+
+#### `schema`
+**Type**: `ZodSchema | Promise<ZodSchema> | (() => ZodSchema | Promise<ZodSchema>)`
+
+An optional [Zod schema](/en/guides/defining-collections/#defining-datatypes-with-zod) that defines the shape of the entries. It is used to both validate the data, and also to generate TypeScript types for the collection. If a function is provided, it will be called at build time before `load()` to generate the schema. You can use this to dynamically generate the schema based on the configuration options or by introspecting an API.
```
Fryuni, line 87: > It is used to both validate the data, and also to generate TypeScript types for the collection.
> [...]
> or by introspecting an API. 

I think it is a bit weird to have the API introspection example when stating that this is used to generate the types because an introspected schema won't be typed.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+
+#### `store`
+**Type**: [`DataStore`](#datastore)
+
+A database to store the actual data. Use this to update the store with new entries. See [`DataStore`](#datastore) for more information.
+
+#### `meta`
+**Type**: `MetaStore`
+
+A simple KV store, designed for things like sync tokens and last-modified times. Use this to store metadata about the collection. The data is persisted between builds, but is only available inside the loader.
```
Fryuni, line 105: > The data is persisted between builds

It might be worth explaining where/how it is persisted. If my last modified time is persisted but the actual data that was read isn't, then it becomes a problem since the implementation won't fetch everything again.

For example, what should I configure in Turbo to store this information and the loaded entries?

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+    watcher?.on('change', async (changedPath) => {
+      if (changedPath === filePath) {
+        logger.info(`Reloading data from ${fileName}`);
+        await syncData(filePath, store);
+      }
+    });
+  },
+};
+```
+#### `refreshContextData`
```
Fryuni, line 215:
```suggestion

#### `refreshContextData`
```

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+        processWebhook(store, refreshContextData.webhookBody);
+      }
+      // ...
+    },
+  };
+}
+```
+
+### `DataStore`
+The data store is a loader's interface to the content collection data. It is a key value store, scoped to the collection, so a loader can only access the data for its own collection.
```
Fryuni, line 236: > scoped to the collection

Is the `meta` store not also scoped to the collection? It is not mentioned there.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+#### `set`
+
+**Type**: `(opts: Opts) => boolean`
+
+The main method to update the store with new entries. It takes an object with the following properties:
+
+- `id: string`: The ID of the entry. Must be unique per collection. If the ID already exists in the store, the entry will be replaced.
+- `data`: The data to store. This will be stored unchanged, so if it needs to be parsed or validated then it should be done before calling `set()`.
+- `body?: string`: The raw body of the content, if applicable. This is mostly relevant for file-based loaders, or those that also store rendered content.
+- `filePath?: string`: The file path of the content, if applicable. This only applies to file-based loaders, and is used to resolve paths such as images or other assets.
```
Fryuni, line 274: If not provided, the images and assets are resolved relative to what? Or are they not resolved at all?

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+        rendered: {
+          html: data.description ?? "",
+        },
+        digest,
+      });
+    }
+```
+
+#### `entries`
+**Type**: `() => Array<[id: string, DataEntry]>`
```
Fryuni, line 299: I'm just double-checking since this seems mostly inspired by JS' `Map`. Is this an actual array, or is it an iterator yielding those tuples?

src/content/docs/en/reference/loader-reference.mdx
florian-lefebvre: as a property of defineCollection

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+	/** An optional content digest, to check if the content has changed. */
+	digest?: number | string;
+	/** The rendered content of the entry, if applicable. */
+	rendered?: RenderedContent;
+}
+```
+
+#### `set`
+
+**Type**: `(opts: Opts) => boolean`
```
ArmandPhilippot, line 267: Nitpicking for the formatting.
Otherwise, does this property refer to [`MutableDataStore`](https://github.com/withastro/astro/blob/1c64ae304d3c008a776e98e48fd3ece8be7b1fb5/packages/astro/src/content/mutable-data-store.ts#L16C14-L16C30) because I can't find it in [`DataStore`](https://github.com/withastro/astro/blob/1c64ae304d3c008a776e98e48fd3ece8be7b1fb5/packages/astro/src/content/data-store.ts#L44). If so, I think the type should be updated to be clearer for readers:
```suggestion
<p>

**Type**: `(collectionName: string, key: string, value: unknown) => boolean`
</p>
```

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+        logger.info("Webhook triggered with body");
+        processWebhook(store, refreshContextData.webhookBody);
+      }
+      // ...
+    },
+  };
+}
+```
+
+### `DataStore`
```
ArmandPhilippot, line 235: nit: Like above, it's better with a new line after a heading
```suggestion
### `DataStore`

```

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+          digest,
+        });
+      }
+    },
+  };
+}
+```
+
+#### `watcher`
+**Type**: `FSWatcher`
```
ArmandPhilippot, line 194: nit:
```suggestion

<p>

**Type**: `FSWatcher`
</p>
```

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+### `LoaderContext`
+This object is passed to the `load()` method of the loader, and contains the following properties:
+
+#### `collection`
+**Type**: `string`
+
+The unique name of the collection. This is the key in the `collections` object in the `src/content/config.ts` file.
+
+#### `store`
+**Type**: [`DataStore`](#datastore)
```
ArmandPhilippot, line 98: nit:
```suggestion

<p>

**Type**: [`DataStore`](#datastore)
</p>
```

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+#### `schema`
+**Type**: `ZodSchema | Promise<ZodSchema> | (() => ZodSchema | Promise<ZodSchema>)`
+
+An optional [Zod schema](/en/guides/defining-collections/#defining-datatypes-with-zod) that defines the shape of the entries. It is used to both validate the data, and also to generate TypeScript types for the collection. If a function is provided, it will be called at build time before `load()` to generate the schema. You can use this to dynamically generate the schema based on the configuration options or by introspecting an API. 
+
+### `LoaderContext`
+This object is passed to the `load()` method of the loader, and contains the following properties:
+
+#### `collection`
+**Type**: `string`
```
ArmandPhilippot, line 93: nit:
```suggestion

<p>

**Type**: `string`
</p>
```

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: [You're right](https://github.com/withastro/astro/blob/40760a8ace13d8be56f3f164f97831c96565c7ee/packages/astro/src/types/public/integrations.ts#L196)!
nit: but we can just use `options` instead of `refreshOptions` (the context is already clear I think and and it shortens a little)

ArmandPhilippot (commented):
Ok so at first reading, I didn't dare quibble about the formatting (since apart from `api-reference` we hadn't reworked the other files) but since Sarah asked me to take a look... 😀 

Overall it's really just reformatting around `Type:`. (the surrounding `p` tag is mainly there to encourage future people who edit the document to place `Default:` or `Since` if needed in the same `p` tag and the empty line is here to prevent a possible bug due to Markdown being embed in HTML tags)
**Just be careful from `DataStore`** (sorry the suggestions are not necessarily in order, I forgot to add some comments): I might have made mistakes when changing the types! Maybe I looked at the wrong files or the types exposed to the public are a little different.
On the other hand, if I didn't make any mistakes, is the difference invisible to the user between `DataStore` and `MutableDataStore`? If it is visible, maybe these properties should be differentiated in the documentation...

I also left a suggestion regarding `DataEntry` as it seems that several properties use this type.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -471,6 +476,53 @@ export default {
 }
 ```
 
+#### `refreshContent` option
+
```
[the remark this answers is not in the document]
ArmandPhilippot: Now that I think about it, about Since, I'm not sure if `5.0.0` works or if we should use `5.0.0-beta` in this case.

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: Edited, thanks!

src/content/docs/en/reference/loader-reference.mdx
Fryuni: Ooh, right. The loaders are used inside the definition, not replacing it. That is what I got wrong.

src/content/docs/en/reference/integrations-reference.mdx
ascorbic: Thanks! Good spot.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+## Loader API
+
+The API for [inline loaders](#inline-loaders) is very simple, and is shown above. This section shows the API for defining an object loader.
+
+### The `Loader` object
+
+The loader object has the following properties:
+
+#### `name`
+**Type**: `string`
```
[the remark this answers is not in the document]
ascorbic: Why the `<p>` tags here?

src/content/docs/en/reference/loader-reference.mdx
ascorbic: It would. You can e.g. do an introspection query to an API endpoint, and then generate the schema from that. This will then in turn generate the TypeScript types from that schema.

src/content/docs/en/reference/loader-reference.mdx
ascorbic: Good point. It is actually persisted in the data store, but a different scope so it's not available in the collection. Would something like "This metadata is persisted between builds alongside the collection data, but is only available inside the loader" work?

src/content/docs/en/reference/loader-reference.mdx
ascorbic: It's an actual array

src/content/docs/en/reference/loader-reference.mdx
ascorbic: No, this has changed in the 5.0 branch. The `DataStore` interface is what was previous called `ScopedDataStore`. The previous `DataStore` is renamed to `ImmutableDataStore`.

See https://github.com/withastro/astro/blob/next/packages/astro/src/content/mutable-data-store.ts#L333

src/content/docs/en/reference/loader-reference.mdx
ascorbic: Good point. They're not resolved at all.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+
+Get all entries in the collection as an array of key-value pairs.
+
+#### `keys`
+**Type**: `() => Array<string>`
+
+Get all the keys of the entries in the collection.
+
+#### `values`
+**Type**: `() => Array<DataEntry>`
```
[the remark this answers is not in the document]
ascorbic: No, see above. You need to look at the `next` branch

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+
+Get all entries in the collection as an array.
+
+#### `delete`
+**Type**: `(key: string) => void`
```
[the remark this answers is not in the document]
ascorbic: See above

src/content/docs/en/reference/loader-reference.mdx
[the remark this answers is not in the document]
ascorbic: Ahh, missed the comment above.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,324 @@ title: Astro Loader API
...
+	/** The file path of the content, if applicable. Relative to the site root. */
+	filePath?: string;
+	/** The raw body of the content, if applicable. */
+	body?: string;
+	/** An optional content digest, to check if the content has changed. */
+	digest?: number | string;
+	/** The rendered content of the entry, if applicable. */
+	rendered?: RenderedContent;
+}
+```
```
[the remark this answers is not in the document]
ascorbic: Good idea. I think it could do with some more discussion too.

ascorbic:
@ArmandPhilippot Thanks for the detailed review! Addressing the point about the `DataStore` types, it seems you're looking at the wrong branch. In the 5.0 branch `DataStore` is the type previously called `ScopedDataStore`. See https://github.com/withastro/astro/pull/11914

ascorbic:
I've extracted the `DataEntry` type into another section, instead of documenting it inside the `get` and `set` methods.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+
+<p>
+
+**Type**: <code>(context: <a href="#loadercontext">LoaderContext</a>) => Promise&lt;void&gt;</code>
+</p>
+
+An async function that is called at build time to load data and update the store. See [`LoaderContext`](#loadercontext) for more information.
+
+#### `schema`
+**Type**: `ZodSchema | Promise<ZodSchema> | (() => ZodSchema | Promise<ZodSchema>)`
```
ArmandPhilippot, line 94: nit (formatting):
```suggestion

<p>

**Type**: `ZodSchema | Promise<ZodSchema> | (() => ZodSchema | Promise<ZodSchema>)`
</p>
```

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+
+An optional [Zod schema](/en/guides/defining-collections/#defining-datatypes-with-zod) that defines the shape of the entries. It is used to both validate the data, and also to generate TypeScript types for the collection. If a function is provided, it will be called at build time before `load()` to generate the schema. You can use this to dynamically generate the schema based on the configuration options or by introspecting an API. 
+
+### `LoaderContext`
+This object is passed to the `load()` method of the loader, and contains the following properties:
```
ArmandPhilippot, line 99: nit (formatting):
```suggestion

This object is passed to the `load()` method of the loader, and contains the following properties:
```

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+A database to store the actual data. Use this to update the store with new entries. See [`DataStore`](#datastore) for more information.
+
+#### `meta`
+
+<p>
+
+**Type**: `MetaStore`
+</p>
+
+A KV store scoped to the collection, designed for things like sync tokens and last-modified times. This metadata is persisted between builds alongside the collection data, but is only available inside the loader.
```
ArmandPhilippot, line 126: I think using the full words might make the sentence clearer for everyone:
```suggestion
A key-value store scoped to the collection, designed for things like sync tokens and last-modified times. This metadata is persisted between builds alongside the collection data, but is only available inside the loader.
```

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+
+<p>
+
+**Type**: `() => Array<[id: string, DataEntry]>`
+</p>
+
+Get all entries in the collection as an array of key-value pairs.
+
+#### `keys`
+**Type**: `() => Array<string>`
```
ArmandPhilippot, line 337: nit (formatting):
```suggestion

<p>

**Type**: `() => Array<string>`
</p>
```

src/content/docs/en/reference/loader-reference.mdx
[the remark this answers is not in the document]
ArmandPhilippot: Thanks for the edit (and sorry again), this one can be marked as resolved!

src/content/docs/en/reference/loader-reference.mdx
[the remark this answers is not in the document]
ArmandPhilippot: Thanks (and sorry again), this one can be marked as resolved!

ArmandPhilippot (commented):
A few more quibbles related to formatting and one about an abbreviation.

After this (and [my question about `addModuleImport`](https://github.com/withastro/docs/pull/9374#issuecomment-2343077538)), I will have nothing more to say! I'll let others speak. 😄 

Thanks @ascorbic for your edits (and your patience with my error 😓 ). Everything looks good to me (including the new `DataEntry` section)!

ascorbic:
@ArmandPhilippot thanks! And yes, you're right the `addModuleImport` JSDoc does need updating. The actual text is wrong too. I'll handle that in the repo.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+The returned object is a [`DataEntry`](#dataentry) object.
+
+#### `set`
+
+<p>
+
+**Type**: <code>(entry: <a href="#dataentry">DataEntry</a>) => boolean</code>
+</p>
+
+Adds an entry to the store, returning `true` if the entry was set. If `digest` was passed and matched an existing entry with the same ID, it will return `false` and not update the entry. See [`DataEntry`](#dataentry) for more information on the fields to pass. The entry should have been parsed using `parseData` before being stored. As well as applying any validations and transforms in the schema, this will also ensure that any images are correctly processed.
```
sarah11918, line 312:
```suggestion
Used after data has been [validated and parsed](#parseData) to add an entry to the store, returning `true` if the entry was set. This returns `false` when the [`digest`](#digest) property determines that an entry has not changed and should not be updated.
```
First paragraph: I feel like that was maybe a lot here? Does this adequately convey what's needed for *this* property with the link? Because we do mention that the `digest` property will search for matching entries, and I think that's adequately covered there?

Second paragraph: I always get nervous when I see "backwards documentation" :sweat_smile: (Oh, and you should have done this first....) I think this is important enough to put it right in the description!

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+An identifier for the entry, which must be unique within the collection. This is used to look up the entry in the store, and is the key used with `getEntry` for that collection. 
+
+#### `data`
+
+<p>
+
+**Type**: `Record<string, unknown>`
+</p>
+
+The actual data for the entry. When a user accesses the collection, this will have TypeScript types generated according to the collection schema. It the responsibility of the loader to use [`parseData`](#parsedata) to validate and parse the data before storing it in the data store: no validation is done when getting or setting the data.
```
sarah11918, line 406:
```suggestion
The actual data for the entry. When a user accesses the collection, this will have TypeScript types generated according to the collection schema.

It the loader's responsibility to use [`parseData`](#parsedata) to validate and parse the data before storing it in the data store: no validation is done when getting or setting the data.
```
Just separating paragraphs!

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+
+#### `filePath`
+
+<p>
+
+**Type**: `string | undefined`
+</p>
+
+A path to the file that is the source of this entry, relative to the root of the site. This only applies to file-based loaders, and is used to resolve paths such as images or other assets. If not set, then any fields in the schema that use [the `image()` helper](/en/guides/images/#images-in-content-collections) will be treated as [public paths](/en/guides/images/#where-to-store-images) and not transformed.
```
sarah11918, line 415:
```suggestion
A path to the file that is the source of this entry, relative to the root of the site. This only applies to file-based loaders, and is used to resolve paths such as images or other assets.

If not set, then any fields in the schema that use [the `image()` helper](/en/guides/images/#images-in-content-collections) will be treated as [public paths](/en/guides/images/#where-to-store-images) and not transformed.
```
Just breaking up paragraphs!

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+The raw body of the entry, if applicable. If the entry includes [rendered content](#rendered), then this field can be used to store the raw source. This is optional, and is not used internally.
+
+#### `digest`
+
+<p>
+
+**Type**: `string | undefined`
+</p>
+
+An optional content digest for the entry. This can be used to check if the data has changed. If the digest is passed when setting an entry, it will only update the entry if the digest does not match the existing entry. The format of the digest is up to the loader, but it should be a string that changes when the data changes. The [`generateDigest`](#generatedigest) function can be used for this.
```
sarah11918, line 433:
```suggestion
An optional content digest for the entry. This can be used to check if the data has changed.

When [setting an entry](#set), the entry will only update if the digest does not match an existing entry with the same ID.

The format of the digest is up to the loader, but it must be a string that changes when the data changes. This can be done with the [`generateDigest`](#generatedigest) function.
```
Alright, does this pair well with our `set` entry?

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+
+#### `rendered`
+
+<p>
+
+**Type**: `RenderedContent | undefined`
+</p>
+
+If the entry has been rendered to HTML, this field can be used to store an object with the rendered content and metadata. This can be used to store the rendered content of a markdown entry, for example, or HTML from a CMS. This field is optional, but if provided then the user can use [the `render()` function and `<Content />` component](/en/guides/working-with-collections/#rendering-body-content) to render the entry in a page.
```
sarah11918, line 442:
```suggestion
Stores an object with an entry's rendered content and metadata if it has been rendered to HTML. For example, this can be used to store the rendered content of a Markdown entry or HTML from a CMS.

If this field is provided, then [the `render()` function and `<Content />` component](/en/guides/working-with-collections/#rendering-body-content) are available to render the entry in a page.
```

sarah11918 (commented):
Whew! Great work everyone. I'm so happy with everyone's care and attention to detail in making our newest API page a model for going forward!

I've made a bunch of tiny changes here, like simply breaking up longer paragraphs, that I will commit directly to reduce noise.

In the case of ones where I'd like people to take a look, I'll leave those open and unresolved. But this is super close to being ready to go, and again, thanks for both the initial great content to work with, and the fabulous reviews to make this shine!

sarah11918 (commented):
Whew! Great work everyone. I'm so happy with everyone's care and attention to detail in making our newest API page a model for going forward!

I've made a bunch of tiny changes here, like simply breaking up longer paragraphs, that I will commit directly to reduce noise.

In the case of ones where I'd like people to take a look, I'll leave those open and unresolved. But this is super close to being ready to go, and again, thanks for both the initial great content to work with, and the fabulous reviews to make this shine!

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
 i18nReady: true
 ---
 import Since from '~/components/Since.astro';
-import { FileTree } from '@astrojs/starlight/components';
 
-[publish your loader to NPM as a package](/en/reference/publish-to-npm/) to share with others and be included in our integrations library.
\ No newline at end of file
+Astro content collections help you organize and load data into your Astro project. This guide explains how to define a loader, and the API that loaders use to interact with the content layer.
```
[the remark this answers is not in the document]
ascorbic: Great, thanks. I wasn't totally sure what tone to strike in this intro section.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+```
+
+#### `logger`
+
+<p>
+
+**Type**: [`AstroIntegrationLogger`](/en/reference/integrations-reference/#astrointegrationlogger)
+</p>
+
+A logger that can be used to log messages to the console. It is recommended to use this instead of `console.log` to ensure that logs are displayed correctly in the Astro UI. It will include the loader name in the log message. See [`AstroIntegrationLogger`](/en/reference/integrations-reference/#astrointegrationlogger) for more information.
```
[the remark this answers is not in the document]
ascorbic: Yeah, I'm not entirely sure what I was getting at there!

src/content/docs/en/reference/loader-reference.mdx
ascorbic: Looks good. Do you think it needs to call out the fact that it only does the conditional insert if the digest is provided. If it's unset, it will always update the entry and return true.

src/content/docs/en/reference/loader-reference.mdx
```diff
@@ -3,6 +3,450 @@ title: Astro Loader API
...
+    }
+    // Optionally, define the schema of an entry. 
+    // It will be overridden by user-defined schema. 
+    schema: async () => z.object({
+      // ...
+    })
+  };
+}
+```
+
```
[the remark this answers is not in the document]
ascorbic: How about this: https://github.com/withastro/docs/pull/9374/commits/e2f9b212072d85892d1cb391fcb19ff16532157c

src/content/docs/en/reference/loader-reference.mdx
sarah11918: I don't *think* we need to call that out? The field is optional, and it can't match if it's not provided in the first place? Let's go understated to start, then we'll find out if people are confused and need the call out!

src/content/docs/en/reference/loader-reference.mdx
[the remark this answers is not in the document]
sarah11918: Fantastic! :partying_face:

PR #9391, opened 2024-09-12
Rename ViewTransitions to ClientRouter

matthewp:
#### Description (required)

Renames `ViewTransitions` component to `ClientRouter` in all of the documentation.

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -2927,20 +2927,20 @@ const { Content } = await entry.render();
 <Content />
 ```
 
-### `<ViewTransitions />`
+### `<ClientRouter />`
 
 <p><Since v="2.9.0" /></p>
 
-Opt in to using view transitions on individual pages by importing and adding the `<ViewTransitions />` routing component to `<head>` on every desired page.
+Opt in to using view transitions on individual pages by importing and adding the `<ClientRouter />` routing component to `<head>` on every desired page.
 
 ```astro title="src/pages/index.astro" ins={2,7}
 ---
-import { ViewTransitions } from 'astro:transitions';
+import { ClientRouter } from 'astro:transitions';
 ---
 <html lang="en">
   <head>
     <title>My Homepage</title>
-    <ViewTransitions />
+    <ClientRouter />
   </head>
   <body>
     <h1>Welcome to my website!</h1>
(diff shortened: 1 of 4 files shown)


sarah11918:
Thanks @matthewp !

I will have to do a bunch of find and replace for regular text paragraphs and explanations, so just checking:

- Is it still the "View Transitions API" ? If not, what's the name of the API/module when we need to reference it.

- When we use it in a sentence, I'm assuming "enabling view transitions in your project" and "Astro's view transition support" should be changed to something like "enabling client-side routing in your project" and "Astro's client-side routing support"

- When we talk about **native browser support**, or anything to do with the browser, we probably still *should* say "view transitions"? Because browsers don't have "client-side routing support" ?

- Can you think of any other situations where saying "view transitions" would be the correct phrase to use?

matthewp:
I don't think anything really needs to change. The name of the thing being imported is all that has changed here. It still uses view transitions as a core part of the routing. It still attempts to mimic native routing as much as possible. Only the import has changed. We can quibble that some things are better worded different to differentiate the "real" view transitions MPA router vs ours, but those things could have / should have been worded differently before this change.

The view transition guide itself could probably use a bit of refactoring to talk about the choice between the MPA routing and using ClientRouter, but I don't see why that needs to happen in this change.

sarah11918:
Perfect, thank you!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -2221,10 +2221,14 @@ export const onRequest = defineMiddleware(async (context, next) => {
 })
 ```
 
-## View Transitions client API (`astro:transitions/client`)
+## View Transitions router API (`astro:transitions/client`)
```
sarah11918, line 2224:
```suggestion
## View Transitions Router API (`astro:transitions/client`)
```

sarah11918:
Thanks @matthewp ! Will merge this in so I can start incorporating, check for any links that might have pointed to the changed header etc.!

Note that the config reference docs need to be changed in the JS Doc in the Astro repo. If you get to it before I do, great. If not, I've made a note!

matthewp:
Those are already done in my PR: https://github.com/withastro/astro/pull/11980

sarah11918:
Perfect, yup, I see it now! They were also committed here, so I'll just undo these changes locally. All good!

PR #9410, opened 2024-09-14
Fix broken Content Layer API RFC link

jurajkapsz:
#### Description (required)

Fixes broken Content Layer API RFC link.


jurajkapsz:
Change made only in english version, if I understood the instructions correctly. Otherwise, the issue is present also in other translations.

delucis (approved):
Thanks for fixing this @jurajkapsz!

jurajkapsz:
Thanks @delucis , glad I could help.

PR #9443, opened 2024-09-18
Enable pagefind for draft docs

lorenzolewis:
#### Description (required)

POC to enable Pagefind for the beta site as the Algolia search is currently using the production site and doesn't match up with the content here.

- Closes #
- Suggested label:


lorenzolewis:
Two immediate issues discovered with this:

- No (easy) way to edit the ranking page results. For example, the Content Collections guide is the 11th result. Put in a discussion on the Starlight repo for this: https://github.com/withastro/starlight/discussions/2348
- No way to group individual results. Right now we have groups such as References, Learn, and Recipes to aid and give the user context when navigating results. This could potentially be solved with some hierarchy in the Pagefind search UI and also exposed in some way in Starlight, but I'm unfamiliar with this area.

sarah11918:
I will note that for the searches "content layer" and "loaders", this Page find version is far better. And, for these beta docs, maybe it's OK to optimize for that?

lorenzolewis:
Converting this to ready for review in case we would like to merge it. Note that this adds just over a minute to the build time when enabled.

TheOtterlord (approved):

PR #9449, opened 2024-09-19
fix: Update install-and-setup.mdx

Egpereira:
# Update install-and-setup.mdx

Update the "Install from the CLI wizard" section, on step 2, where it says "If you skipped the `npm install` step during the CLI wizard..." to a more identifiable name reflecting the current CLI wizard, where it says "Install dependencies?" instead of "npm install" and to not be focused on npm since other package managers are available. I'm not 100% sure if the question mark should be present or not though.

Also, the backticks were removed since I believe this text would not be considered appropriate inside a code tag.

Astro's CLI section entitled Install dependencies?


sarah11918 (approved):
Thank you so much! This is great. 💜

PR #9453, opened 2024-09-20
feat(markdown): document new option `langAlias`

ematipico:
#### Description (required)

Adds documentation for a new `shiki` option called `langAlias`

 #### For Astro version: `4.16.0`. See astro PR [#12039](https://github.com/withastro/astro/pull/12039).


src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -355,6 +355,12 @@ export default defineConfig({
       // Note: Shiki has countless langs built-in, including .astro!
       // https://shiki.style/languages
       langs: [],
+      // Add custom aliases for languages
+      // The key is the alias you'll use in your markdown files and the value is the language ID used by Shiki
```
sarah11918, line 359:
```suggestion
      // Map an alias to a Shiki language ID: https://shiki.style/languages#bundled-languages
```
I think this might be sufficient? Especially with the link below!

I do wonder about providing the languages link specifically, though. Because people might not realize that `js`, which everyone uses all the time, is itself an alias for `javascript` (the ID). So it might be helpful for them to have that list of language IDs!

sarah11918 (commented):
More options are always great! See my suggestion below!

sarah11918:
Noting that this PR looks good, but there might be something in the upstream PR re: whether or not `langs` is required. And if so, that will need to go here too. (If not, then this PR is fine!)

ematipico:
> Noting that this PR looks good, but there might be something in the upstream PR re: whether or not `langs` is required. And if so, that will need to go here too. (If not, then this PR is fine!)

@sarah11918  `langs` aren't required, Bjorn fixed the issue

sarah11918:
>  Bjorn fixed the issue

Of course he did! 😄

PR #9463, opened 2024-09-21
i18n(fr): fix inconsistent use of `props`/`prop`/`property` in French translation

ArmandPhilippot:
This PR follows up [a discussion we had on Discord](https://discord.com/channels/830184174198718474/971729943786561556/1286078277810782289) about consistency when translating `props`/`prop` here and in Starlight doc.

#### Description (required)

* Use feminine form when we use `props`
* Use `props` instead of `propriétés` when referring to components props
* Replace `prop` (singular) with `propriété` (in Astro doc, it seems `attribute` is mostly used for native HTML attributes)

In addition, in the updated file I added some other small changes:
* Replace quotation marks with [guillemet](https://en.wikipedia.org/wiki/Guillemet)
* Fix some typo
* Fix some erroneous translations
* Reword some sentences that sounded weird
* Translate in French some untranslated parts
* Add/remove blank lines to make line numbers match English version (it's easier to find your way when there are updates)

Finally, in `astro-components.mdx`, I also added a missing change (see #6793).

- Suggested label: i18n, consistency/formatting


src/content/docs/fr/guides/typescript.mdx
```diff
@@ -169,20 +169,20 @@ const { greeting = 'Hello', name } = Astro.props;
 
 <p><Since v="1.6.0" /></p>
 
-Astro est livré avec des types utilitaires intégrés pour les types d'objets courants. Ils sont disponibles sous l'endpoint `astro/types`.
+Astro est livré avec quelques types d'utilitaires intégrés pour les modèles courants de typage de propriété. Ceux-ci sont disponibles sous le point d'entrée `astro/types`.
```
ArmandPhilippot, line 172: I'm not proud of this translation... but the previous version was a bit wrong if we compare it with the original version (using `objets` when it's properties). If you have a better suggestion, I'm all ears!

thomasbnt (approved):
Good work!

yanthomasdev (approved):
LGTM! Amazing work! 🙌

PR #9464, opened 2024-09-21
i18n(fr): fix `tutorial/5-astro-api` wording

ArmandPhilippot:
#### Description (required)

In French and in this context, `tag` should be translated with `étiquette` and not with `balise`. Also, this word is mostly used in `3.mdx` so we should be consistent.

- Suggested label: i18n, consistency/formatting


yanthomasdev (approved):
LGTM!

PR #9465, opened 2024-09-22
[i18nIgnore] [reference] new components page

sarah11918:
#### Description (required)

This splits out the "Built-in components" section at the bottom of API reference into its own page. This makes them easier to find, and helps with the API reference page getting too long for Algolia to properly index


src/content/docs/en/reference/components-reference.mdx
```diff
@@ -0,0 +1,263 @@
+---
+title: Component Reference
```
yanthomasdev, line 2: It should be in plural, right?
```suggestion
title: Components Reference
```

sarah11918: Maybe even "Built-in Components Reference", since it's not a reference for generic components?

src/content/docs/en/guides/upgrade-to/v1.mdx
```diff
@@ -342,7 +342,7 @@ import { Prism } from '@astrojs/prism';
 ---
 ```
 
-Since the `@astrojs/prism` package is still bundled with `astro` core, you won't need to install anything new, nor add Prism as an integration! However, note that we _do_ plan to extract `@astrojs/prism` (and Prism syntax highlighting in general) to a separate, installable package in the future. See [the `<Prism />` component API reference](/en/reference/api-reference/#prism-) for more.
+Since the `@astrojs/prism` package is still bundled with `astro` core, you won't need to install anything new, nor add Prism as an integration! However, note that we _do_ plan to extract `@astrojs/prism` (and Prism syntax highlighting in general) to a separate, installable package in the future. See [the `<Prism />` component API reference](/en/reference/components-reference/#prism-) for more.
```
sarah11918, line 345:
```suggestion
Since the `@astrojs/prism` package is still bundled with `astro` core, you won't need to install anything new, nor add Prism as an integration! However, note that we _do_ plan to extract `@astrojs/prism` (and Prism syntax highlighting in general) to a separate, installable package in the future. See [the `<Prism />` component API reference](/en/reference/components-reference/#prism-) for more details.
```

src/content/docs/en/reference/components-reference.mdx
```diff
@@ -0,0 +1,263 @@
...
+---
+import Since from '~/components/Since.astro';
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
+import ReadMore from '~/components/ReadMore.astro';
+
+## Built-in Components
+
+Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components';`.
+
+You can reference the `Props` of these components using the [`ComponentProp` type](/en/guides/typescript/#componentprops-type) utility.
```
ArmandPhilippot, line 13:
```suggestion
You can reference the `Props` of these components using the [`ComponentProps` type](/en/guides/typescript/#componentprops-type) utility.
```

src/content/docs/en/reference/components-reference.mdx
```diff
@@ -0,0 +1,263 @@
...
+
+### `<Code />`
+
+```astro 'theme="dark-plus"' /wrap\b/ /(inline) \/>/
```
ArmandPhilippot, line 17: This has nothing to do with the split but as I was rereading I was wondering... Shouldn't `defaultColor` also be highlighted? (if the goal is to highlight the attributes cited in line 36)
````suggestion
```astro 'theme="dark-plus"' /wrap\b/ /(inline) 'defaultColor={false}' \/>/
````

Note: I'm not used to Expressive Code, so feel free to double-check that I'm using the correct syntax. I took over the one from `theme`...

File: src/content/docs/en/reference/components-reference.mdx [hunk 5]
```diff
@@ -0,0 +1,263 @@
...
+  code={code}
+  lang="js"
+  transformers={[transformerMetaHighlight()]}
+  meta="{1,3}" />
+  
+  <style is:global>
+    pre.has-focused .line:not(.focused) {
+      filter: blur(1px);
+    }
+  </style>
```
ArmandPhilippot, line 69: nit: The first list is a matter of taste I guess but, as for the rest, `<style>` should not be indented.
```suggestion
  meta="{1,3}"
/>
  
<style is:global>
  pre.has-focused .line:not(.focused) {
    filter: blur(1px);
  }
</style>
```

src/content/docs/en/reference/components-reference.mdx
```diff
@@ -0,0 +1,263 @@
...
+<Content />
+```
+
+### `<ViewTransitions />`
+
+<p><Since v="2.9.0" /></p>
+
+Opt in to using view transitions on individual pages by importing and adding the `<ViewTransitions />` routing component to `<head>` on every desired page.
+
+```astro title="src/pages/index.astro" ins={2,7}
```
ArmandPhilippot, line 230: nit: All the code examples on the page use highlighting rather than diff notation so I think this one should do the same.
````suggestion
```astro title="src/pages/index.astro" {2,7}
````
Was there a specific reason for using `ins`? I'm guessing it was a copy/paste from an experimental feature/changeset.

**Review by ArmandPhilippot — COMMENTED** [review]
Regarding the split, I have nothing to say... besides reducing the size of `api-reference.mdx`, I think it adds visibility for those only interested in components. So, LGTM!
But since I'm here... some nitpicking regarding code snippets. 😄

File: src/content/docs/en/reference/components-reference.mdx (hunk 6 above)
Comment (sarah11918) [reply src/content/docs/en/reference/components-reference.mdx:230]: I'm going to guess that the `ins` was used to emphasize the "add this to otherwise unchanged code, and you get view transitions." Probably an artefact of this having been written when the component was new, and so all the text was about "adding" view transitions. (I do remember at the time we were HEAVY on this language because we were emphasizing you didn't just "turn them on", you added them to every page you wanted them on, or in a broader component, like a layout.

We can consider now that this is just "reference" and not "omg, here's how you add this new thing!" 😄

Comment (ArmandPhilippot) [reply src/content/docs/en/reference/components-reference.mdx:230]: Afterwards, it's not necessarily annoying... it was more a question of consistency. But this can also be explained by the sentence above (“Opt in”) which perhaps "pushes more to action" than other descriptions so it can stay like that! (as I said, I was nitpicking 😄 ).

File: src/i18n/en/nav.ts [hunk 7]
```diff
@@ -111,6 +111,11 @@ export default [
 		slug: 'reference/directives-reference',
 		key: 'reference/directives-reference',
 	},
+	{
+		text: 'Built-in components',
```
VoxelMC, line 115:
```suggestion
		text: 'Built-in Components',
```
There once was a letter quite small,
That slipped through a crack in the wall.
It got lost in the code,
Walked many a road,
Until Vox made it quite a bit tall.

src/content/docs/en/reference/components-reference.mdx
```diff
@@ -0,0 +1,264 @@
...
+title: Built-in Components Reference
+i18nReady: true
+---
+import Since from '~/components/Since.astro';
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
+import ReadMore from '~/components/ReadMore.astro';
+
+## Built-in Components
+
+Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components';`.
```
VoxelMC, line 11:
```suggestion
Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components'`.
```
Not sure if the semi is necessary in an inline code example, but it's been a while!

src/content/docs/en/reference/components-reference.mdx
```diff
@@ -0,0 +1,264 @@
...
+
+You can reference the `Props` of these components using the [`ComponentProps` type](/en/guides/typescript/#componentprops-type) utility.
+
+### `<Code />`
+
+```astro 'theme="dark-plus"' /wrap\b/ /(inline) 'defaultColor={false}' \/>/
```
ArmandPhilippot, line 17: Sorry it seems I misplaced `defaultColor` in my previous suggestion... So both `inline` and `defaultColor` highlighting are broken. 😓 
````suggestion
```astro 'theme="dark-plus"' /wrap\b/ /(inline) \/>/ 'defaultColor={false}'
````
I think it should be good now... I don't know why I put it in the regex! 👀 (I should have done the review after and not before going to bed 😅 )

File: src/content/docs/en/reference/components-reference.mdx [hunk 10]
```diff
@@ -0,0 +1,264 @@
...
+```
+#### Properties
+
+- src (required)
+- alt (required)
+- width and height (required for `public/` and remote images)
+- format
+- quality
+- densities
+- widths
```
VoxelMC, line 152: These Properties don't follow the same format as the properties for `<Picture />`. Perhaps we should make them the same? I prefer the way of the Picture component, however, it might be lengthy here.

VoxelMC (commented):
Awesome, LGTM. Thank you!

This will make these more intuitive to find for sure!

A few tiny comments are all I have for ya.

src/i18n/en/nav.ts
sarah11918: A wild Voxel appears,
Feels like hasn't been spotted in years!
Cut to the chase,
With a fix to the case,
To a resounding chorus of cheers! 🎉

src/content/docs/en/reference/components-reference.mdx
sarah11918: Agreed. Looks a little funky.

src/content/docs/en/reference/components-reference.mdx
sarah11918: We do spell these out on the image page, and I think rather than overhaul here now we'll keep this for a future fix because:

- we do want to rework the image page, so we could swap the loooong list there for this one, but that will also mean checking links etc.
- we might even go the Starlight route and make a page for each component! Even if we don't, with these things being the only things on their own page, we can make some standardization choices across all the different components.

So since that's a BIG change, let's consider it for a future improvement of this page!

sarah11918:
!coauthor

src/content/docs/en/reference/components-reference.mdx
```diff
@@ -0,0 +1,265 @@
+---
+title: Built-in Components Reference
+i18nReady: true
+tableOfContents:
+  minHeadingLevel: 2
+  maxHeadingLevel: 2
```
sarah11918, line 6: Until we go through and structure this page and we decide what it will look like, temporary fix so that only the components show up in the TOC.

(The alternative would be to convert `###` to `<h3>`, which would also be fine! This way preserves our normal Markdown authoring.)

src/content/docs/en/reference/components-reference.mdx
```diff
@@ -0,0 +1,265 @@
...
+---
+import Since from '~/components/Since.astro';
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
+import ReadMore from '~/components/ReadMore.astro';
+
+Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files via `import {} from 'astro:components'`.
```
sarah11918, line 12: The heading that was here was redundant (pulled directly from the other page where it was simply a page section.)

This means we don't need it here, and all the components can be `##`.

sarah11918, line 12:
```suggestion
Astro includes several built-in components for you to use in your projects. All built-in components are available in `.astro` files and may require an import statement.
```

PR #9468, opened 2024-09-23
Update instructions to migrate from Studio to Turso

Fryuni:
#### Description

Following the current instructions to create a Turso DB from the exported dump using `turso db create --from-dump` will create all the tables required for the data without Astro DB being aware of those tables.
This causes a conflict later on when trying to push the schema since, as far as Astro DB knows, it hasn't created any table in the new DB. Trying to create them again fails and leaves the DB in an inconsistent state.


sarah11918 (approved):
Thank you!

PR #9471, opened 2024-09-23
fix: update ordered list numbering in `guides/deploy/fleek.mdx`

ArmandPhilippot:
#### Description (required)

Fix an ordered list numbering. (There were two number fours.)

It doesn't seem to affect the published page (the Steps component seems to fix the numbering). That said, it's best to fix the numbering now to avoid an error if `Steps` is ever changed.

- Suggested label: typo


yanthomasdev (approved):
Nice catch, thanks!

PR #9475, opened 2024-09-24
Update to SST v3

jayair:
Greetings Astronauts!

#### Description

Updating the SST tutorial with snippets and links for SST v3.

PR #9479, opened 2024-09-24
fix: line highlighting in `guides/deploy/sst.mdx`

ArmandPhilippot:
#### Description (required)

I'm not sure about this one, but it seems that the expressive code has not been updated in #9475.

See https://github.com/withastro/docs/pull/9475/files. The code block was:

````mdx
```ts {2} {4} title="sst.config.ts"
app.stack(function Site(ctx) {
  const bucket = new Bucket(ctx.stack, "public");
  const site = new AstroSite(ctx.stack, "site", {
    bind: [bucket],
  });

  ctx.stack.addOutputs({
    url: site.url,
  });
```
````

So the highlighting should match the lines:
* `const bucket` (`{2}`)
* `bind: [bucket]` (`{4}`)

With [the new version](https://docs.astro.build/en/guides/deploy/sst/#sst-components) the highlighting does not make sense.

However, `const bucket` is no longer on a single line. So I don't know if we should:
* remove the highlighting
* edit it to highlight the words and not the lines
* update the line numbers (what i did here)

- Suggested label: typo

### DIFF
--- a/src/content/docs/en/guides/deploy/sst.mdx
+++ b/src/content/docs/en/guides/deploy/sst.mdx
@@ -28,7 +28,7 @@ You can also read [the full Astro on AWS with SST tutorial](https://sst.dev/docs
 
 To use any [additional SST components](https://sst.dev/docs/), add them to `sst.config.ts`. 
 
-```ts {2} {4} title="sst.config.ts"
+```ts {1} {5} title="sst.config.ts"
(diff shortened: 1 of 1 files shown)


sarah11918:
!coauthor

sarah11918:
Great catch! Thank you @ArmandPhilippot and @Nin3lee !

PR #9488, opened 2024-09-25
Improve pagination documentation

### LINKED ISSUE #9442: Improvements to Pagination Documentation: Type Information, Argument Descriptions, and Redundant Sections
### 📚 Subject area/topic

Pagination - API Reference

https://docs.astro.build/en/reference/api-reference/#the-pagination-page-prop
https://docs.astro.build/en/reference/api-reference/#paginate
(Potentially) https://docs.astro.build/en/guides/routing/#complete-api-reference

A couple of days ago, I was implementing pagination in an Astro-based project on GitHub. While working on it, I faced a few challenges related to the documentation on pagination. I was able to resolve them by exploring the documentation, the Astro codebase, or a combination of both. After addressing the issues, I realized that these changes could be reflected in the documentation to prevent similar challenges for future users. Initially, I planned to create a PR right away, but the scope of this task felt slightly beyond the realm of "issue-less PRs." Therefore, I’ve created this issue first for approval.

On API Reference: 

Under [#Page Props](https://docs.astro.build/en/reference/api-reference/#the-pagination-page-prop):

1. Missing type information for `Page` prop
   - Just looking at the reference page, it is not immediately clear that `Page` is a `generic` type. 
   - Suggestion is to include a `Type` attribute under the title like: `Type: Page<T = any>`
2. Similarly, `page.data` property simply mentions `Array`. To be more precise, this could be updated to `Array[T]` following the changes mentioned above.
3. Incorrect mention of `page.data` prop as a function in its description. Perhaps, the intent was to mention either `paginate` or `getStaticPaths` function?
4. Lastly, one of the challenges I faced was typing the array returned by `page.data` when using Content Collections. It may have been my oversight, but it was not immediately obvious that I had to use `CollectionEntry` utility type. This could be worth mentioning in the reference page... or perhaps in the Pagination section under `Routing` guide.

Under [#paginate](https://docs.astro.build/en/reference/api-reference/#paginate):

1. Missing signature information for the `paginate` function
   - Looking at [PaginateFunction#L2834](https://github.com/withastro/astro/blob/8d4eb95086ae79339764d02215f84f4f21b96edc/packages/astro/src/%40types/astro.ts#L2834) type in the codebase, I realize why it was left out as the type is quite complex. I am not so well versed in Astro, so not sure how this could be tackled.
2. Missing  argument from the `paginate` list:
```mdx
paginate() has the following arguments:
    - pageSize - The number of items shown per page (10 by default)
    - params - Send additional parameters for creating dynamic routes
    - props - Send additional props to be available on each page
```
The reference mentions the `data` argument before the example, but it is missing from the list above. If the goal is to highlight optional arguments, then the wording could be improved.

On Routing guide, under [Pagination#Complete API Reference](https://docs.astro.build/en/guides/routing/#complete-api-reference):

This section seems redundant. Issue #8929 addressed the same concern, and the conclusion was to keep it due to the lack of a better solution. However, reading through the comments, it seems the pagination section previously didn't link back to the API reference page—perhaps the main reason for keeping it. Here are a few reasons why I think removing this section entirely might be beneficial:
- It adds to the maintenance cost, as any modifications need to be reflected across two pages. Failure to do so could again result in inaccurate information.
- The guide already mentions the most common properties. If the user wants to learn more, there is already a link to the reference page
- It doesn't seem to fit well in the "guide" section. Perhaps it's just me, but when I want to understand the structure of types, I typically consult the reference pages. This might not be the compelling argument, though.

As @sarah11918 suggested, we could move this type block into the reference page, as it does provide a useful high-level overview of the Page API. However, another argument can be made that it might be better to link to the actual type in the Astro codebase to avoid further maintenance costs.

_No response_

ManorSailor:
#### Description (required)

Following the discussion at #9442; This PR does the following:

1. Adds type info to `Page` prop.
   - As per [Page](https://github.com/withastro/astro/blob/8d4eb95086ae79339764d02215f84f4f21b96edc/packages/astro/src/%40types/astro.ts#L2804) `page` is a generic type. As a tradeoff, `TData`
   -  Where should we mention `TData` is of type `any`? `Type`? Or in description? Should we mention it at all? `TData` is generic enough...
name was decided to balance the documentation for TS & non-TS folks alike.
2. Adds the same type to `page.data` property
3. Adds type information for the `getStaticPaths` function
   - As the said function can be used not only for content collection, but also for data fetched via APIs... I believe adding it under the current section makes sense. I have a doubt though, how will it infer the types for data returned via API?
   - Added the block just above the caution block. Let me know how it looks.
4. & 5. The `paginate` function's type definition was too complex to be easily explained in the documentation. Therefore, I thought of describing the function signature in the docs instead.
6. Removed redundant `Page` type reference from the `Routing` guide
7. As suggested by @ArmandPhilippot, each mention of `CollectionEntry<collection>` has been replaced with `CollectionEntry<TCollectionName>` as specified by its documentation.

Anyways. That's all. Feel free to provide any inputs!

Special thanks to @ArmandPhilippot for his detailed responses. It cleared a lot of confusion.

- Closes #9442
- Suggested label: `improve-documentation`


src/content/docs/en/guides/routing.mdx
```diff
@@ -459,41 +459,6 @@ const { page } = Astro.props;
 {page.url.last ? <a href={page.url.last}>Last</a> : null}
 ```
 
-
-#### Complete API reference
```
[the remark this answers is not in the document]
ManorSailor: You raise a good point. We lose the convenience of glancing at all of the available properties, & having to navigate through the wall of text in the API Reference section... I agree that most of the properties are self explanatory. I can imagine that the some users may find it inconvenient to scroll all that just to see the available properties.

Considering the repercussions, keeping that block there seems warranted even though it adds to the maintenance costs.... Although, it still irks me. 😓 Surely, we can do better....

In hindsight, I realize that the actual purpose of that code block there was to serve as an _At a glance_ section to showcase all of the supported properties by `Page`. 

I was going to suggest that we simplify the type to a block which only mentions all of the available properties, but I realize that it would be much more work than simply copy-pasting the updated type from the codebase.

Should I revert back to its original state then?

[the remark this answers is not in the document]
sarah11918: I made a commit with an example of what I think could be a helpful structure to work from. What do you think?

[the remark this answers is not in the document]
sarah11918: though, we did lose this text which might go in the block itself for `data`

> array containing the page’s slice of data that you passed to the paginate() function

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1116,6 +1116,10 @@ The `getStaticPaths()` function should return an array of objects to determine w
 
 It can also be used in static file endpoints for [dynamic routing](/en/guides/endpoints/#params-and-dynamic-routing).
 
+:::tip
+When using TypeScript, use the [`InferGetStatic*`](/en/guides/typescript/#infer-getstaticpaths-types) type utilities to ensure type-safe access of your `params` and `props`.
```
ArmandPhilippot, line 1120: I'm a little late, I was waiting until the PR was no longer a draft to comment.

`InferGetStatic` does not exist and I don't think the asterisk is understandable to everyone. It seems to me that `GetStaticPaths` is enough to have both `Astro.params` and `Astrop.props` inferred properly. The two other helpers (`InferGetStaticParamsType` and `InferGetStaticPropsType`) can be useful in some cases but I don't think they are required here.
So I would suggest:

```suggestion
When using TypeScript, use the [`GetStaticPaths`](/en/guides/typescript/#infer-getstaticpaths-types) type utility to ensure type-safe access of your `params` and `props`.
```

src/content/docs/en/guides/routing.mdx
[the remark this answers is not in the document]
ArmandPhilippot: It seems we are coming back to https://github.com/withastro/docs/issues/8929#issuecomment-2255932662 😅 So, it's probably better to leave the block here. And, I agree with Sarah, we could replace `result` which is not descriptive enough with the sentence quoted!

sarah11918:
Just updated branch to re-trigger the Netlify deploy. We had some random wonkiness yesterday.

src/content/docs/en/reference/api-reference.mdx
ManorSailor: Thanks for the suggestion! How about the following phrasing:
```md
When using TypeScript, follow the [Infer `getStaticPaths` types](/en/guides/typescript/#infer-getstaticpaths-types) guide to ensure type-safe access of your `params` and `props`.
```

src/content/docs/en/guides/routing.mdx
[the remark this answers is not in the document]
ManorSailor: I think that adding the description for `data` property in the block would work perfectly fine.

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: Why not, all types are useful. If I insisted on `GetStaticPaths` it is perhaps because I find that this section of the Typescript guide seems to insist on the other two types (`InferGetStaticParamsType` and `InferGetStaticPropsType`) when only one (`GetStaticPaths`) is enough for most cases. But that might just be my perception...
So I'm fine with your suggestion, if you think it's more helpful!

ManorSailor: After looking into it, it appears that GetStaticPaths is sufficient for type safety. However, this raises the question of why there are additional type utilities included. Additionally, that section does not highlight the line where we add satisfies GetStaticPaths; it only mentions the InferGetStatic* types. It seems like there’s a new area for improvement here?Tthough I believe it falls outside the scope of this PR.

Considering this, I will go with your suggestion. It seems satisfactory.

ArmandPhilippot: Um yes I think there is a missing line among those highlighted. Maybe it was forgotten because the opening parenthesis is also important in this case. As for the usefulness of the other types... I don't know. I've never used them, but no doubt some people need them. Better excess types than missing types! 😄 
But you're right, it doesn't fit into this PR.

src/content/docs/en/guides/routing.mdx
```diff
@@ -494,6 +469,25 @@ interface Page<T = any> {
 }
 ```
 
+The following example template displays current information for the page along with links to navigate between pages:
```
ArmandPhilippot, line 472: I'm not sure `template` is needed here... I feel like it makes the sentence heavier... (I don't know if it makes sense in English) but I'm not the best person to discuss the wording.
```suggestion
The following example displays current information for the page along with links to navigate between pages:
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1217,28 +1221,35 @@ export async function getStaticPaths({ paginate }) {
 const { page } = Astro.props;
 ```
 
+`paginate()` has the following arguments:
+- `data` - Array of data to paginate
+- `options` - Optional object with following properties:
```
ArmandPhilippot, line 1226:
```suggestion
- `options` - Optional object with following the properties:
```

ArmandPhilippot (approved):
Some nitpicks about the wording. But I'm not the best person to judge so I'm not blocking the PR for that, LGTM!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1217,28 +1221,35 @@ export async function getStaticPaths({ paginate }) {
 const { page } = Astro.props;
 ```
 
+`paginate()` has the following arguments:
+- `data` - Array of data to paginate
```
sarah11918, line 1225:
```suggestion
- `data` - array containing the page’s data passed to the `paginate()` function
```

src/content/docs/en/guides/routing.mdx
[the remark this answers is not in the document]
sarah11918: OK, I think I'm good with where things are here!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1116,6 +1116,10 @@ The `getStaticPaths()` function should return an array of objects to determine w
 
 It can also be used in static file endpoints for [dynamic routing](/en/guides/endpoints/#params-and-dynamic-routing).
 
+:::tip
+When using TypeScript, use the [`GetStaticPaths`](/en/guides/typescript/#infer-getstaticpaths-types) type utility to ensure type-safe access of your `params` and `props`.
+:::
```
sarah11918, line 1121:
```suggestion
```
I agree with the comment from the issue about NOT putting this here. This is about `getStaticPaths()` and its type should be documented there.

Also, it's not a "tip" if it's "how things work."  😄

sarah11918:
OK, I think we're good to go here! Thank you both for your great work, and I'm going to get this merged to get helping people right away. 🚀

PR #9495, opened 2024-09-26
feat: createCodegenDir

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/integrations-reference.mdx
+++ b/src/content/docs/en/reference/integrations-reference.mdx
@@ -400,7 +400,7 @@ A function that creates the `<root>/.astro/integrations/<normalized_integration_
 
 It allows you to have a dedicated folder, avoiding conflicts with another integration or Astro itself. This directory is created by calling this function so it's safe to write files to it directly:
 
-```js
+```ts title="my-integration.ts"
 import { writeFileSync } from 'node:fs'
 
 const integration = {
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -386,6 +388,32 @@ The **`stage`** denotes how this script (the `content`) should be inserted. Some
...
+
+<p>
+
+**Type:** `() => URL;`<br />
+<Since v="5.0.0" />
+</p>
+
+A function that creates the `<root>/.astro/integrations/<normalized_integration_name>` folder and returns its path.
+
+It allows you to have a dedicated folder, avoiding conflicts with another integration or Astro itself. This directory is created by calling this function so it's safe to write files to it directly:
```
sarah11918, line 401:
```suggestion
A function that creates a `<root>/.astro/integrations/<normalized_integration_name>` folder and returns its path.

This allows your integration to have a dedicated configuration folder to retrieve during `astro:config:setup` hook, avoiding conflicts with another integration or Astro itself.

This directory is created by calling this function and only used by your integration, so it is safe to write files to it directly:
```
If this is accurate, it's a tiny bit more explicit about what's happening!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -386,6 +388,32 @@ The **`stage`** denotes how this script (the `content`) should be inserted. Some
...
+
+```js
```
sarah11918, line 403:
``````suggestion
```ts title="my-integration.ts"
``````
I think this would be an OK title? Would be nice to have one!

sarah11918 (approved):
Looks great, @florian-lefebvre ! Just made a couple of quick suggestions for your consideration, but nothing blocking! 🙌

PR #9497, opened 2024-09-26
feat: document decoding of params

ematipico:
#### Description (required)

This PR documents a new behaviour that changed in Astro v5


sarah11918 (commented):
Thank you for your patience @ematipico !

I made some tiny edits, and in both cases move the position of the content (so it's closer to the SSG/`getStaticPaths` stuff in the routing guide, and with the main Astro changes, not the other APIs, in the upgrade guide.

Please review to check on this, then you are happy and the feature is released, it's fine to merge!

PR #9505, opened 2024-09-27
Add info about propertiesToHash to image-service-reference.mdx

### LINKED ISSUE #9169: propertiesToHash missing from Image Service API Docs
### 📚 Subject area/topic

Image Service API

https://docs.astro.build/en/reference/image-service-reference/

According to [this PR](https://github.com/withastro/astro/pull/8984), the Image Service API has a property `propertiesToHash` which is used to control which properties are included in the filename hash of an optimised asset.

There doesn’t seem to be a mention on the page in the docs, although I believe there should be, since it’s important information to know when creating a custom image service that supports more options than the standard one.

On a related note, I believe the built-in `sharpService` is not using the `propertiesToHash`? I couldn’t find a mention in the source code. Is that intentional and if so, why?

**Edit:** looking at the patch that introduced the change, it seems like if an image service doesn’t define `propertiesToHash`, Astro will fall back on the default ones defined in `packages/astro/src/assets/consts.ts`. Perhaps this could also be added to the updated documentation.

_No response_

amxmln:
#### Description

Local image services have a property `propertiesToHash` which is missing from the docs. This change adds it along with a short explanation of what it does and what it falls back to if not passed.

- Closes #9169 
- Suggested label: add new  content

### DIFF
--- a/src/content/docs/en/reference/image-service-reference.mdx
+++ b/src/content/docs/en/reference/image-service-reference.mdx
@@ -122,7 +122,7 @@ const service: LocalImageService = {
 export default service;
 ```
 
-At build time for static sites and pre-rendered routes, both `<Image />` and `getImage(options)` call the `transform()` function. They pass options either through component attributes or an `options` argument, respectively. The transformed images will be built to a `dist/_astro` folder. Their file names will contain a hash of the properties passed to `propertiesToHash`. This property is optional and will fall back to `['src', 'width', 'height', 'format', 'quality']` if not set. If your custom image service has more options which change the generated images, you should add these to the array.
+At build time for static sites and pre-rendered routes, both `<Image />` and `getImage(options)` call the `transform()` function. They pass options either through component attributes or an `options` argument, respectively. The transformed images will be built to a `dist/_astro` folder. Their file names will contain a hash of the properties passed to `propertiesToHash`. This property is optional and will default to `['src', 'width', 'height', 'format', 'quality']`. If your custom image service has more options that change the generated images, add these to the array.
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Thank you for this PR! I'll ask @Princesseuh to give a technical review, and to confirm that this information is added in the best place!

Princesseuh (approved):
Looks great to me, thank you!

sarah11918 (approved):
We love it when issues work out like this! Thank you @amxmln , and welcome to Team Docs! :partying_face:

src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -116,12 +116,13 @@ const service: LocalImageService = {
...
 		};
-	}
+	},
+  propertiesToHash: ['src', 'width', 'height', 'format', 'quality'], 
 };
 export default service;
 ```
 
-At build time for static sites and pre-rendered routes, both `<Image />` and `getImage(options)` call the `transform()` function. They pass options either through component attributes or an `options` argument, respectively. The transformed images will be built to a `dist/_astro` folder.
+At build time for static sites and pre-rendered routes, both `<Image />` and `getImage(options)` call the `transform()` function. They pass options either through component attributes or an `options` argument, respectively. The transformed images will be built to a `dist/_astro` folder. Their file names will contain a hash of the properties passed to `propertiesToHash`. This property is optional and will fall back to `['src', 'width', 'height', 'format', 'quality']` if not set. If your custom image service has more options which change the generated images, you should add these to the array.
```
sarah11918, line 125:
```suggestion
At build time for static sites and pre-rendered routes, both `<Image />` and `getImage(options)` call the `transform()` function. They pass options either through component attributes or an `options` argument, respectively. The transformed images will be built to a `dist/_astro` folder. Their file names will contain a hash of the properties passed to `propertiesToHash`. This property is optional and will default to `['src', 'width', 'height', 'format', 'quality']`. If your custom image service has more options that change the generated images, add these to the array.
```

yanthomasdev (approved):
LGTM! Good work 🙌

PR #9513, opened 2024-09-30
feat: add a `<Meta />` component to document APIs

ArmandPhilippot:
This PR implements #9498. Feel free to check out this discussion for more context and some implementation details.

##### Changes

* Adds a `<Meta />` component to ease the authoring of the `Type:/Default:/Since` block
* Updates English documentation to use the new component
* Converts all `Signature:` to `Type:` in `dev-toolbar-app-reference.mdx` to be consistent (and to be able to use the component)
* Updates `docgen` script to use our new component

##### Supported Syntaxes

`<Meta />` can be used:
* without children if there is no type (e.g. `<Meta version="5.0.0" />`)
* on one line (e.g. ``<Meta>`number`</Meta>``) or adding line breaks (after `<Meta>` and before `</Meta>`) to make it easier to read/write when the type is long
* with HTML tags inside to use links

The available props are:
* `availableFor` (to support `integrations-guide/vercel.mdx`)
* `cli`
* `default`
* `pkg` and `version` (both are related to Since, without `version` the `pkg` do nothing)

Both `cli` and `default` properties can have backticks (Markdown syntax) to use inline code (e.g. ``<Meta default="`10`">`number`</Meta>``) so more complicated values like ``<Meta default='`"development"` when running `astro dev`, `"production"` when running `astro build`'>`` are also supported.

##### Not supported

1. Multiple types like [`getEntry()`](https://docs.astro.build/en/reference/api-reference/#getentry)

We could, but the rendering on smartphone is not great... You can see [some previews of my tests in the discussion](https://github.com/withastro/docs/discussions/9498#discussioncomment-10798357).

2. The syntax used in [`Collection Entry Type`](https://docs.astro.build/en/reference/api-reference/#collection-entry-type)

`Available for:` is already supported by the component. The problem is the different variants for the type (`Example type:`/` Example Types:`/`Type:` and which is sometimes on multiple lines). Should we add a new prop `exampleType`? Even if we do this, there remains the problem of multiple examples/types.

##### Future

1. We let active contributors update the documentation in their language to avoid errors
2. For languages ​​with fewer contributors, we can create a language-by-language PR with `i18nIgnore`
3. Once the documentation is updated in all languages, we can remove `<Since />` component OR update its syntax to use `dt/dd` and remove the duplicated logic from `<Meta />`

- Closes #9498
- Suggested label: consistency/formatting and feedback-improvement


src/i18n/en/ui.ts
```diff
@@ -39,6 +39,13 @@ export default {
 	'since.addedIn': 'Added in:',
 	'since.new': 'New',
 	'since.beta': 'Beta',
+	// Used in `<Meta>`
+	'meta.addedIn': 'Added in:',
+	'meta.badge.new': 'New',
+	'meta.badge.beta': 'Beta',
```
ArmandPhilippot, line 45: Since the `<Since />` logic was duplicated in `<Meta />`, I figured it might be better to duplicate the UI strings as well. But I can also reuse the existing ones if that's better.

scripts/docgen.mjs
```diff
@@ -148,10 +148,26 @@ function getDeprecatedAside(tag) {
 }
 
 /**
- * Get block of type, CLI command, default value, and version added in for the current comment.
+ * Retrieve the Meta component props as a string from the given data.
+ * @param {{ cli?: string; defaultValue?: string; version?: string }} data
+ */
+function getMetaComponentProps(data) {
+	const props = [];
+	if (data.cli) props.push('cli="`' + data.cli + '`"');
```
Fryuni, line 156: It seems weird to have hard-coded backticks surrounding this value just for it to be replaced inside of the component. Wouldn't it be better for the component to wrap anything passed in this field with `<code>`?

```suggestion
	if (data.cli) props.push(`cli="${data.cli}"`);
```

Then on the component:
```diff
- <dd class="meta-value" set:html={replaceBackticksWithCodeTags(cli)} />
+ <dd class="meta-value"><code set:html={cli}></code></dd>

**Review by Fryuni — COMMENTED** [review]
This looks great! I left some comments regarding readability and a piece of, apparently, redundant logic

File: src/components/Meta.astro [hunk 3]
```diff
@@ -0,0 +1,142 @@
...
+				<dt class="meta-label">{t('meta.type')}</dt>
+				<dd class="meta-value" set:html={trimmedType} />
+			</div>
+		) : null
+	}
+	{
+		cli ? (
+			<div class="meta-item">
+				<dt class="meta-label">{t('meta.cli')}</dt>
+				<dd class="meta-value" set:html={replaceBackticksWithCodeTags(cli)} />
```
ArmandPhilippot, line 89: This goes along with Fryuni's suggestion:
```suggestion
				<dd class="meta-value"><code set:html={cli}></code></dd>
```

sarah11918:
Hi Armand! We discussed this PR on Talking and Doc'ing today and decided that it was a great exploration, and maybe something to explore again in the future, but would take more discussion and work than we had planned for to make an implementation that everyone was happy with that is significantly better than the current authoring experience.

So we'll close this experiment for now, but are so thankful that you looked into this for us, and we are not against revisiting the idea sometime in the future! Right now, we're going to focus on our content structure/sidebar updates and then see where we are again! :raised_hands:

ArmandPhilippot:
No worries, I understand it wasn't the perfect solution! Thanks again for your feedback! 🫡

PR #9541, opened 2024-10-02
fix(guides): add error handling for sync errors in supabase guide

### LINKED ISSUE #9134: Infinite Loop redirection
Those two redirect can lead to an infinite loop redirection

Signin should only redirect on dashboard if the JWT is valid not only because not !

https://github.com/withastro/docs/blob/af56f405f3cfeb67044907c8114543c17f6130c3/src/content/docs/en/guides/backend/supabase.mdx?plain=1#L285

Dashboard should only redirect on signin if no error validating the tokens

https://github.com/withastro/docs/blob/af56f405f3cfeb67044907c8114543c17f6130c3/src/content/docs/en/guides/backend/supabase.mdx?plain=1#L331

setSession can return an error but also throw an error. this is a bad practice (I already report to Supabase team) but it is the case.
This mean that checking return error is not enough and the example should be rewrite like this maybe : 

```
...
const { cookies, redirect } = Astro;

const accessToken = cookies.get("sb-access-token");
const refreshToken = cookies.get("sb-refresh-token");

if (!accessToken || !refreshToken) {
  return redirect("/signin");
}
let setSession: AuthResponse;
try {
  setSession = await supabase.auth.setSession({
    refresh_token: refreshToken.value,
    access_token: accessToken.value,
  });
  if (setSession.error) {
    cookies.delete("sb-access-token", {
      path: "/",
    });
    cookies.delete("sb-refresh-token", {
      path: "/",
    });
    return redirect("/signin");
  }
} catch (error) {
  return redirect("/signin");
}
const email = setSession.data.user?.email;
...
```

You can produce the infinite loop by editing your cookie  with invalid data on sb-access-token for example.
it will create a server error by the way because of the unhandled throw but you have the point.

kevinzunigacuellar:
#### Description (required)

This PR adds a try catch block to handle sync errors in the supabase dashboard code snippet

- Closes #9134


sarah11918 (approved):
Thanks @kevinzunigacuellar ! I saw your comment about maybe eventually refreshing this guide more comprehensively now that we have env and actions, but I appreciate the immediate fix to catch errors!

yanthomasdev (approved):
LGTM!

PR #9563, opened 2024-10-04
change action name from addToCard to addToCart

Grahf0085:
I believe addtoCard is a small spelling error and the correct name for the action should be addToCart. Sorry if I misunderstood.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/actions.mdx
+++ b/src/content/docs/en/guides/actions.mdx
@@ -569,7 +569,7 @@ For example, you might return the name of a product in an `addToCart` action, ra
 import { defineAction } from 'astro:actions';
 
 export const server = {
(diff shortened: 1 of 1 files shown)


apatel369 (approved):
Nice catch! Welcome to Team Doc!

sarah11918 (approved):
Great catch @Grahf0085 ! Thanks so much for taking the time to make this fix, and again, welcome to Team Docs! 🥳

sarah11918:
!coauthor

sarah11918:
Note that this error is present in the two currently translated and up to date pages: `fr` and `ko` as the code sample I'm sure was just copy-pasted. So, we do want Lunaria to update these pages' status!

PR #9582, opened 2024-10-07
fix: add caution

ematipico:
#### Description (required)

More than once users were incorrectly using the containers API, I figured that adding a `caution` block could help user to understand which API to use, and when.

### DIFF
--- a/src/content/docs/en/reference/container-reference.mdx
+++ b/src/content/docs/en/reference/container-reference.mdx
@@ -68,7 +68,11 @@ For each official Astro integration, import and use the `getContainerRenderer()`
 
 For renderer packages outside the `@astrojs` npm org, look in their documentation for `getContainerRenderer()` or a similar function provided.
 
-When using `vite` (`vitest`, Astro integrations, etc.), the renderers are loaded with the function `loadRenderers()` from the virtual module `astro:container`. Outside `vite`, you'll have to load the renderers manually.
+When using `vite` (`vitest`, Astro integrations, etc.), the renderers are loaded with the function `loadRenderers()` from the virtual module `astro:container`.
+ 
+:::caution
+Outside `vite` or for on-demand usage, you'll have to [load the renderers manually](#adding-a-renderer-manually).
+:::
 
 The following example provides the necessary object to render an Astro component that renders a React component and a Svelte component: 
 


src/content/docs/en/reference/container-reference.mdx
```diff
@@ -68,7 +68,11 @@ For each official Astro integration, import and use the `getContainerRenderer()`
 
 For renderer packages outside the `@astrojs` npm org, look in their documentation for `getContainerRenderer()` or a similar function provided.
 
-When using `vite` (`vitest`, Astro integrations, etc.), the renderers are loaded with the function `loadRenderers()` from the virtual module `astro:container`. Outside `vite`, you'll have to load the renderers manually.
+When using `vite` (`vitest`, Astro integrations, etc.), the renderers are loaded with the function `loadRenderers()` from the virtual module `astro:container`.
+ 
+:::caution
+Outside `vite` or for on-demand usage, you'll have to [load the renderers manually](#adding-a-renderer-manually).
+:::
```
sarah11918, line 75:
```suggestion
Outside `vite` or for on-demand usage, you'll have to [load the renderers manually](#adding-a-renderer-manually).
```
Let's try without the big caution first, as this isn't like a security warning or anything, and see if just having it written explicitly here before the example is enough! If you still find people are struggling with this, then we can look at making it more visible!

sarah11918 (approved):
Seems like a great addition @ematipico ! My preference is always to avoid notes etc. for content that is essential, like this is. So I left a suggestion to try without a big flaming caution *first*, to see whether that's enough.

But the content is great!

src/content/docs/en/reference/container-reference.mdx
ematipico: I'm not really sure. Currently, we already have this phrase in production, and I still get users that use the container API incorrectly. :/

sarah11918: OK! I leave it to your choice! It's not my preference, but it's a quick fix and I'll put it on my plate to figure out if we can maybe do something else at a later time to make the different uses more clear. 

This might mean that the information is being laid out in the wrong way on the page, or that people are using the feature differently than we expect -- if using it *outside* of Vite is actually more common, then we could e.g. start with that section so that's what they see first!

But I'm OK if you want to merge the caution for now!

yanthomasdev (approved):

PR #9601, opened 2024-10-08
RSS: Add config reference links

bholmesdev:
#### Description (required)

Add links to the `@astrojs/rss` README for a complete configuration reference.

- Link from the step-by-step guide to call out all config options
- Link from the `items` rendering guide to call out the items reference

- Related: https://github.com/withastro/astro/pull/12157

### DIFF
--- a/src/content/docs/en/guides/rss.mdx
+++ b/src/content/docs/en/guides/rss.mdx
@@ -81,7 +81,7 @@ The RSS feed standard format includes the following values for each published it
 - `pubDate`: The date of publication of the entry. (optional)
 - `content`: The full content of your post. (optional)
 
-<ReadMore>[See `item` configuration reference](https://github.com/withastro/astro/tree/main/packages/astro-rss#items) for a complete list of options.</ReadMore>
+<ReadMore>See [`item` configuration reference](https://github.com/withastro/astro/tree/main/packages/astro-rss#items) for a complete list of options.</ReadMore>
 
 ### Using content collections
 
--- a/src/content/docs/en/guides/rss.mdx
+++ b/src/content/docs/en/guides/rss.mdx
@@ -101,7 +101,6 @@ export async function GET(context) {
       title: post.data.title,
       pubDate: post.data.pubDate,
       description: post.data.description,
-      customData: post.data.customData,
       // Compute RSS link from post `slug`
       // This example assumes all posts are rendered as `/blog/[slug]` routes
       link: `/blog/${post.slug}/`,
--- a/src/content/docs/en/guides/rss.mdx
+++ b/src/content/docs/en/guides/rss.mdx
@@ -67,7 +67,7 @@ The package [`@astrojs/rss`](https://github.com/withastro/astro/tree/main/packag
     ```
 </Steps>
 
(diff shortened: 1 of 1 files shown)


bholmesdev:
All very clear! Just removing `customData` since it's rarely used and best left to config reference

src/content/docs/en/guides/rss.mdx
```diff
@@ -77,7 +80,8 @@ The RSS feed standard format includes the following values for each published it
 - `link`: A URL to the original source of the entry. (optional)
 - `pubDate`: The date of publication of the entry. (optional)
 - `content`: The full content of your post. (optional)
-- `customData`: A field for including any extra data, such as other frontmatter properties from your blog posts.
+
+<ReadMore>See [`item` configuration reference](https://github.com/withastro/astro/tree/main/packages/astro-rss#items) for a complete list of options.</ReadMore>
```
yanthomasdev, line 84: I believe the option is actually `items`? I am not sure if you meant the configuration reference for an *individual* item, if so, then it should not contain backticks. Also, nit: missing "the".
```suggestion
<ReadMore>See the [`items` configuration reference](https://github.com/withastro/astro/tree/main/packages/astro-rss#items) for a complete list of options.</ReadMore>
```

src/content/docs/en/guides/rss.mdx
```diff
@@ -66,9 +67,11 @@ The package [`@astrojs/rss`](https://github.com/withastro/astro/tree/main/packag
...
 </Steps>
 
+<ReadMore>See the [`@astrojs/rss` README](https://github.com/withastro/astro/tree/main/packages/astro-rss) for the full configuration reference.</ReadMore>
+
 ## Generating `items`
 
-The `items` field accepts a list of RSS feed objects, which can be generated from content collections entries using `getCollection()` or from your page files using `pagesGlobToRssItems()`. 
+The `items` field accepts a list of RSS feed objects, which can be generated from content collections entries using `getCollection()` or from your page files using `pagesGlobToRssItems()`.
 
 The RSS feed standard format includes the following values for each published item:
```
sarah11918, line 76:
```suggestion
The RSS feed standard format includes metadata for each published item, including values such as:
```
So, my main concern is even with the link below, this sounds like an exhaustive, complete list. Anything to soften that up a bit here would be fine!

src/content/docs/en/guides/rss.mdx
```diff
@@ -66,9 +67,11 @@ The package [`@astrojs/rss`](https://github.com/withastro/astro/tree/main/packag
     ```
 </Steps>
 
+<ReadMore>See the [`@astrojs/rss` README](https://github.com/withastro/astro/tree/main/packages/astro-rss) for the full configuration reference.</ReadMore>
```
sarah11918, line 70: Now I **really** want this turned into an integration, so we just have the standard integration guide page in docs!!

src/content/docs/en/guides/rss.mdx
sarah11918: I believe Yan is correct here, it sounds like the intention is to point to the reference property!

src/content/docs/en/guides/rss.mdx
bholmesdev: Yup, totally agree!

sarah11918 (approved):
LGTM! Thanks for the update Ben!

PR #9571, opened 2024-10-05
Update category table for libraries on NPM

Fryuni:
#### Description

- Rename `collection` to `category` to match the name in the [integrations library page](https://astro.build/integrations)
- Add "Dev Toolbar" category:
	- Added as "Developer Tools" in https://github.com/withastro/astro.build/pull/878
	- Renamed to "Dev Toolbar" in https://github.com/withastro/astro.build/pull/906
- Add "Loader" and "Utilities" categories:
	- Added in https://github.com/withastro/astro.build/pull/1281
- Document default "Uncategorized" category:
	- Added in https://github.com/withastro/astro.build/pull/1281

PR #3457, opened 2023-06-08
Improve texts and remove "we"

cadamini:
#### What kind of changes does this PR include?

- Minor content fixes (according to the writing guide)

#### Description

1. Excited when I heard about Astro.
2. Headed over to your docs.
3. Found an interesting page tagged as stubs,
4. Checked the writing guide, 
5. Found this page, and on it, some sentences that could be simpler.

Edited the page content. I hope this helps.

PR #9608, opened 2024-10-08
i18n(fr): update `publish-to-npm.mdx`

ArmandPhilippot:
#### Description (required)

* Renames `Collection` to `Categories` and adds other changes from #9571
* Adds a new category from #9598
* Adds changes from #3457 (it seems that this PR has been forgotten)
* Replaces all `Paquet` occurrences with `paquet` (it's not a proper noun)

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/reference/publish-to-npm.mdx
+++ b/src/content/docs/fr/reference/publish-to-npm.mdx
@@ -17,7 +17,7 @@ Regardez [le modèle de composant de la communauté Astro](https://github.com/as
 
 ## Démarrage rapide
 
-Pour commencer à développer rapidement votre composant, nous avons un modèle déjà configuré pour vous.
+Pour commencer à développer rapidement votre composant, vous pouvez utiliser un modèle déjà configuré pour vous.
 
 ```bash
 # Initialise le modèle de composant Astro dans un nouveau répertoire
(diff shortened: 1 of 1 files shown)


src/content/docs/fr/reference/publish-to-npm.mdx
```diff
@@ -271,28 +269,34 @@ Vous avez besoin d'aide pour construire votre intégration, ou vous voulez simpl
...
 
-| Collection         | Mots-clés                                    |
-|--------------------|----------------------------------------------|
-| Accessibilité      | `a11y`, `accessibility`                      |
-| Adapteurs          | `astro-adapter`                              |
-| Analyse de données | `analytics`                                  |
-| CSS + UI           | `css`, `ui`, `icon`, `icons`, `renderer`     |
-| Frameworks         | `renderer`                                   |
-| Performance + SEO  | `performance`, `perf`, `seo`, `optimization` |
+Les paquets qui n'incluent aucun mot-clé correspondant à une catégorie seront affichés comme `Uncategorized`.
```
ArmandPhilippot, line 299: I didn't translate `Uncategorized` because it's between backticks and the [Integrations library](https://astro.build/integrations) is not available in French.
So, I also wonder if it is relevant to translate the name of the categories in the table above... Leaving them translated can help those who are looking to associate the right keywords but it can also be confusing if they are looking for the exact name in the library... I don't know which is better, so I translated the categories I added to the table since the others were already translated.

dreyfus92: It works as it is, thank you for taking the time in updating it🫡

dreyfus92 (approved):

PR #9611, opened 2024-10-09
Fix posts.map() return format in getStaticPaths function

developaul:
## Description
This PR fixes the return format of the `posts.map()` function in the `getStaticPaths` method. The previous format was causing issues with [describe the problem if you know it].

## Changes made
- Updated the return statement in `posts.map()` to use parentheses instead of square brackets
- Removed an unnecessary nested array


sarah11918 (approved):
Thank you for this @developaul ! We appreciate you taking the time to come and improve this. Welcome to Team Docs! 🥳

PR #9641, opened 2024-10-11
Document custom mode support

bluwy:
#### Description (required)

Add documentation for custom mode support. You can now pass any string to `--mode` and the `mode` inline config.

It also adds the `--devOutput` flag support for `astro build` to output a development-based build. Previously I don't believe there's an easy way, even if you had set `--mode development` (it may only partially work).

This is a feature and shouldn't have any breaking changes, so there's no entry in the migration guide.

- Closes #
- Suggested label:


src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -71,7 +71,7 @@ const isDev = import.meta.env.DEV;
 ### `.env` files
 Environment variables can be loaded from `.env` files in your project directory.
 
-You can also attach a mode (either `production` or `development`) to the filename, like `.env.production` or `.env.development`, which makes the environment variables only take effect in that mode.
+You can also attach a mode to the filename, like `.env.production`, `.env.development`, or `.env.staging`, which makes the environment variables only take effect in that mode. The mode can be changed by passing the [`--mode` flag](/en/reference/cli-reference/#--mode-string) to the `astro dev` or `astro build` commands (defaults to `"development"` and `"production"` respectively).
```
sarah11918, line 74:
``````suggestion
You can also add `.production`, `.development` or a custom mode name to the filename itself (e.g `env.testing`, `.env.staging`). This allows you to use different sets of environment variables at different times.

The `astro dev` and `astro build` commands default to `"development"` and `"production"` modes, respectively.  You can run these commands with the [`--mode` flag](#--mode-string) to pass a different value for `mode` and load the matching `.env` file.

This allows you to run the dev server or build your site connecting to different APIs:

```bash
# Run the dev server connected to a "staging" API
astro dev --mode staging

# Build a site that connects to a "production" API with additional debug information
astro build --devOutput

# Build a site that connects to a "testing" API
astro build --mode testing
```
``````
I really liked what you had in the changeset, Bjorn! What about something like this here?

ALSO, I think this entire section should go AFTER we show the basic .env file, that is the simple use case. I would put this after the example, and before the "read more"

File: src/content/docs/en/reference/cli-reference.mdx [hunk 2]
```diff
@@ -492,11 +506,14 @@ The inline config passed in this object will take highest priority when merging
...
 
-**Type:** `"development" | "production"`<br />
-**Default:** `"development"` when running `astro dev`, `"production"` when running `astro build`
+**Type:** `string`<br />
+**Default:** `"development"` when running `astro dev`, `"production"` when running `astro build`<br />
+<Since v="5.0.0" />
 </p>
 
-The mode used when building your site to generate either "development" or "production" code.
+The mode used when developing or building your site. It's passed to Vite that affects the value of `import.meta.env.MODE` and how `.env` files are loaded, which also affects the values of `astro:env`. See the [environment variables page](/en/guides/environment-variables/) for more details.
```
sarah11918, line 514:
```suggestion
The mode used when developing or building your site (e.g. `"production"`, `"testing"`).

This value is passed to Vite using [the `--mode` flag](#--mode-string) when the `astro build` or `astro dev` commands are run to determine the value of `import.meta.env.MODE`.This determines which `.env` files are loaded, and therefore the values of `astro:env`. See the [environment variables page](/en/guides/environment-variables/) for more details.
```
I might be making things up, but is this correct? 😄

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -200,7 +200,15 @@ Runs Astro's development server. This is a local HTTP server that doesn't bundle
 
 Builds your site for deployment. By default, this will generate static files and place them in a `dist/` directory. If [SSR is enabled](/en/guides/on-demand-rendering/), this will generate the necessary server files to serve your site.
 
-Can be combined with the [common flags](#common-flags) documented below.
+### Flags
+
+The command accepts [common flags](#common-flags) and the following additional flags:
```
sarah11918, line 205: Just a comment, but maybe at some point we stop splitting out common from all? Doesn't have to be in this PR of course!

sarah11918 (commented):
Looks great! @bluwy! I stole a bit from the PR changeset, and I think the environment variable stuff should become its own smaller section AFTER introducing variables in general (maybe we even give it a sub heading? ), so see what you think!

src/content/docs/en/reference/cli-reference.mdx
bluwy: Yeah I think that would be best. I notice our command flags don't overlap a lot actually, except like `--help`, `--verbose`, etc

sarah11918: If you'd *like* to do that here, feel free! Otherwise, I'll tackle this at some point separately!

bluwy: I think it might be best to tackle that separately 😅

src/content/docs/en/guides/environment-variables.mdx
bluwy: Looks great! I was afraid that it's getting a bit long, but if we put that in a separate section then it's probably fine then. I'll commit this change and re-sort it.

src/content/docs/en/reference/cli-reference.mdx
bluwy: Looks right to me!

src/content/docs/en/reference/cli-reference.mdx
sarah11918: Sure thing! Let's tackle the entire list separately!

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -492,11 +506,16 @@ The inline config passed in this object will take highest priority when merging
...
-**Default:** `"development"` when running `astro dev`, `"production"` when running `astro build`
+**Type:** `string`<br />
+**Default:** `"development"` when running `astro dev`, `"production"` when running `astro build`<br />
+<Since v="5.0.0" />
 </p>
 
-The mode used when building your site to generate either "development" or "production" code.
+The mode used when developing or building your site (e.g. `"production"`, `"testing"`).
+
+This value is passed to Vite using [the `--mode` flag](#--mode-string) when the `astro build` or `astro dev` commands are run to determine the value of `import.meta.env.MODE`. This also determines which `.env` files are loaded, and therefore the values of `astro:env`. See the [environment variables page](/en/guides/environment-variables/) for more details.
```
sarah11918, line 516:
```suggestion
This value is passed to Vite using [the `--mode` flag](#--mode-string) when the `astro build` or `astro dev` commands are run to determine the value of `import.meta.env.MODE`.

This also determines which `.env` file is loaded (e.g. `env.production`, `env.testing`, and therefore the values of `astro:env`. See the [environment variables page](/en/guides/environment-variables/) for more details.

Note that changing the `mode` does not change the kind of code transform handled by Vite and Astro:

- In `astro dev`, Astro will transform code with debug information.
- In `astro build`, Astro will transform code with the most optimized output and removes debug information.
- In `astro build --devOutput` (new flag), Astro will transform code with debug information like in `astro dev`. 

This enables various usecases like:

```bash
# Run the dev server connected to a "staging" API
astro dev --mode staging

# Build a site that connects to a "staging" API
astro build --mode staging

# Build a site that connects to a "production" API with additional debug information
astro build --devOutput

# Build a site that connects to a "testing" API
astro build --mode testing
```
``````
I really like the information from the changeset. Do you think it's helpful here, or too much?
```

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -82,6 +80,23 @@ DB_PASSWORD="foobar"
 PUBLIC_POKEAPI="https://pokeapi.co/api/v2"
 ```
 
+You can also add `.production`, `.development` or a custom mode name to the filename itself (e.g `env.testing`, `.env.staging`). This allows you to use different sets of environment variables at different times.
+
+The `astro dev` and `astro build` commands default to `"development"` and `"production"` modes, respectively.  You can run these commands with the [`--mode` flag](#--mode-string) to pass a different value for `mode` and load the matching `.env` file.
```
sarah11918, line 85:
```suggestion
The `astro dev` and `astro build` commands default to `"development"` and `"production"` modes, respectively. You can run these commands with the [`--mode` flag](#--mode-string) to pass a different value for `mode` and load the matching `.env` file.
```
Just an extra space!

sarah11918 (approved):
Looks great @bluwy ! Just had a question about adding more info in the CLI reference page. I'm not sure myself, so I'll leave it to you!

src/content/docs/en/reference/cli-reference.mdx
casungo: i think it can be really useful to have it in the docs!

sarah11918: lol, I must have really liked it! Serves me right for editing over two weeks :laughing:

sarah11918 (approved):
Looks great @bluwy  :raised_hands:

casungo (approved):
LGTM!

PR #9659, opened 2024-10-13
fix: update `cloudinary.mdx`

jsparkdev:
## replace `pnpm install` with `pnpm add`

The pnpm official documentation recommends using the `add` command instead of `install` when installing new packages as follows:

- https://pnpm.io/cli/add
- https://pnpm.io/cli/install

Currently it works the same without any differences, but the document clearly distinguishes these two commands.

This means that you have the flexibility to adapt to changes, such as changing or removing commands in future versions.

## other updates

- fix wrong external link
- fix wrong inline code highlight


sarah11918 (approved):
Thank you for always being on guard @jsparkdev ! 🫡

PR #9672, opened 2024-10-14
remove old-translations folder

sarah11918:
#### Description (required)

Removes the `old-translations/` folder (outside of `src/`!

At one point, we thought it would be helpful for translators to have access to recently removed or past translation files for quicker creating of new ones, using some of the same phrasing or terminology.

We decided not to do that anymore, so this folder goes! If helpful, translators can check past versions in git to see older files.


dreyfus92 (approved):
Nice housecleaning, @sarah11918! 🧹 It's great to tidy up and keep things streamlined. Thanks for taking care of this 😁

TheOtterlord (approved):

PR #9673, opened 2024-10-14
fix(5.0.0): nitpicking and suggestions regarding `on-demand-rendering.mdx`

ArmandPhilippot:
#### Description (required)

* Typo fix (`every`)
* Updates some code snippet due to some concerns about highlighting (I only updated from the previous/next example but maybe all examples should use the same format: either simple highlighting or `ins`)
* Updates the file name of a code snippet because the usage of `Astro.params` seemed wrong otherwise
* Adds a suggestion regarding `Astro.redirect` in `Response` subsection

- Suggested label: 5.0, 5.0.0-beta, code snippet update

### DIFF
--- a/src/content/docs/en/guides/on-demand-rendering.mdx
+++ b/src/content/docs/en/guides/on-demand-rendering.mdx
@@ -207,11 +207,11 @@ Astro.response.headers.set('Cache-Control', 'public, max-age=3600');
 
 #### Return a `Response` object
 
-You can also return a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object directly from any page using on-demand rendering. 
+You can also return a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object directly from any page using on-demand rendering either manually or with [`Astro.redirect`](/en/reference/api-reference/#astroredirect). 
 
-The example below returns a 404 on a dynamic page after looking up an ID in the database:
+The example below looks up an ID in the database on a dynamic page and either it returns a 404 if the product does not exist, or it redirects the user to another page if the product is no longer available, or it displays the product:
 
-```astro title="src/pages/product/[id].astro" {10-13}
+```astro title="src/pages/product/[id].astro" {10-13, 18}
 ---
 export const prerender = false; // Not needed in 'server' mode
 
@@ -226,27 +226,10 @@ if (!product) {
     statusText: 'Not found'
   });
 }
----
-<html>
-  <!-- Page here... -->
-</html>
-```
 
-#### Redirect to another page
-
-Instead of returning a `Response` object, you can also redirect the user to another page using `Astro.redirect`:
-
-```astro title="src/pages/product/[id].astro" {10}
----
-export const prerender = false; // Not needed in 'server' mode
-
-import { getProduct } from '../api';
-
-const product = await getProduct(Astro.params.id);
-
-// No product found
-if (!product) {
-  return Astro.redirect('/404', 404);
+// The product is no longer available
+if (!product.isAvailable) {
+  return Astro.redirect("/products", 301);
 }
 ---
 <html>
--- a/src/content/docs/en/guides/on-demand-rendering.mdx
+++ b/src/content/docs/en/guides/on-demand-rendering.mdx
@@ -57,7 +57,7 @@ For a **highly dynamic app**, after adding an adapter, [set your build output co
 
 Then, if needed, you can choose to prerender any individual pages that do not require a server to execute, such as a privacy policy or about page.
 
-```astro title="src/pages/about-my-app.astro" {2}
+```astro title="src/pages/about-my-app.astro" ins={2}
 ---
 export const prerender = true
 ---
--- a/src/content/docs/en/guides/on-demand-rendering.mdx
+++ b/src/content/docs/en/guides/on-demand-rendering.mdx
@@ -38,7 +38,7 @@ All other pages are statically-generated at build time!
 
 The following example shows opting out of prerendering in order to display a random number each time the endpoint is hit:
 
-```js title="src/pages/randomnumber.js" {1}
+```js title="src/pages/randomnumber.js" ins={1}
 export const prerender = false;
 
 export async function GET() {
@@ -53,11 +53,11 @@ export async function GET() {
 ```
 ### `'server'` mode
 
-For a **highly dynamic app**, after adding an adapter, [set your build output configuration to `output: 'server'`](/en/reference/configuration-reference/#output) to **server-render all your pages by default**. This is the equivalent of opting out of prerendering on evey page.
+For a **highly dynamic app**, after adding an adapter, [set your build output configuration to `output: 'server'`](/en/reference/configuration-reference/#output) to **server-render all your pages by default**. This is the equivalent of opting out of prerendering on every page.
 
 Then, if needed, you can choose to prerender any individual pages that do not require a server to execute, such as a privacy policy or about page.
 
-```astro title="src/pages/about-my-app.astro"
+```astro title="src/pages/about-my-app.astro" {2}
 ---
 export const prerender = true
 ---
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/on-demand-rendering.mdx
```diff
@@ -38,7 +38,7 @@ All other pages are statically-generated at build time!
 
 The following example shows opting out of prerendering in order to display a random number each time the endpoint is hit:
 
-```js title="src/pages/randomnumber.js" {1}
+```js title="src/pages/randomnumber.js" ins={1}
```
ArmandPhilippot, line 41: nit: The previous example use `ins={2}`, shouldn't we use the same visual information here? (or the other way around... not covered by my suggestion)

src/content/docs/en/guides/on-demand-rendering.mdx
```diff
@@ -53,11 +53,11 @@ export async function GET() {
 ```
 ### `'server'` mode
 
-For a **highly dynamic app**, after adding an adapter, [set your build output configuration to `output: 'server'`](/en/reference/configuration-reference/#output) to **server-render all your pages by default**. This is the equivalent of opting out of prerendering on evey page.
+For a **highly dynamic app**, after adding an adapter, [set your build output configuration to `output: 'server'`](/en/reference/configuration-reference/#output) to **server-render all your pages by default**. This is the equivalent of opting out of prerendering on every page.
 
 Then, if needed, you can choose to prerender any individual pages that do not require a server to execute, such as a privacy policy or about page.
 
-```astro title="src/pages/about-my-app.astro"
+```astro title="src/pages/about-my-app.astro" {2}
```
ArmandPhilippot, line 60: Same concern here about the highlighting. The next example uses "highlighting" but we don't use it here. Maybe we should use `ins` here and in the next example too? (I just updated from the next example)

src/content/docs/en/guides/on-demand-rendering.mdx
```diff
@@ -169,9 +169,9 @@ See more details about [`Astro.cookies` and the `AstroCookie` type](/en/referenc
 
 [`Astro.response`](/en/reference/api-reference/#astroresponse) is a standard [`ResponseInit`](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response#options) object. It can be used to set the response status and headers.
 
-The example below sets a response status and status text for a product listing page when the product does not exist:
+The example below sets a response status and status text for a product page when the product does not exist:
 
-```astro title="src/pages/my-product.astro" {10,11}
+```astro title="src/pages/product/[id].astro" {10,11}
```
ArmandPhilippot, line 174: From my understanding, `Astro.params` only works with path segments (e.g. `[id]`).

src/content/docs/en/guides/on-demand-rendering.mdx
```diff
@@ -211,7 +211,7 @@ You can also return a [Response](https://developer.mozilla.org/en-US/docs/Web/AP
 
 The example below returns a 404 on a dynamic page after looking up an ID in the database:
 
-```astro title="src/pages/[id].astro" {10-13}
+```astro title="src/pages/product/[id].astro" {10-13}
```
ArmandPhilippot, line 214: I only updated because that's the path I had used before... it wasn't really a problem, just a matter of consistency.

src/content/docs/en/guides/on-demand-rendering.mdx
```diff
@@ -232,6 +232,28 @@ if (!product) {
...
+// No product found
+if (!product) {
+  return Astro.redirect('/404', 404);
+}
+---
+<html>
+  <!-- Page here... -->
+</html>
+```
+
```
ArmandPhilippot, line 256: I'm not sure about this suggestion, that's why I made a separate commit. Since we present the different possible `Response` formats with on demand rendering, I thought that `Astro.redirect` also has its place here. Especially since the second parameter is only applicable with on demand rendering. On the other hand, this is perhaps not the best example... I took the one above.

JusticeMatthew: Doesn't returning a 404 response automatically navigate them to the 404 page anyway?

I'm not 100% sure here

ArmandPhilippot: Yes I think so too... that's why I said it might not be the best example. I did think of a 301, but I was looking for an example without making it too complicated; it didn't come to me at the time. I'll think about it! (well probably tomorrow, it's getting late for me ^^)

ArmandPhilippot: A quick idea that came to me: (I also added a link to `Astro.redirect` reference)
````suggestion
#### Redirect to another page

Instead of returning a `Response` object, you can also redirect the user to another page using [Astro.redirect](/en/reference/api-reference/#astroredirect):

```astro title="src/pages/product/[id].astro" {10}
---

import { UNRENEWED_PRODUCTS } from '../constants';
import { getProduct } from '../api';

const product = await getProduct(Astro.params.id);

// The product no longer exists
if (UNRENEWED_PRODUCTS.includes(product)) {
  return Astro.redirect('/products', 301);
}
---
<html>
  <!-- Page here... -->
</html>
```

````

Comment (JusticeMatthew) [reply src/content/docs/en/guides/on-demand-rendering.mdx:256]: What about just modifying the existing example slightly?

```ts
---

import { getProduct } from '../api';

const product = await getProduct(Astro.params.id);

// No product found
if (!product) {
  Astro.response.status = 404;
  Astro.response.statusText = 'Not found';
  // Or redirect the user to another page
  return Astro.redirect('/products', 301);
}
---
<html>
  <!-- Page here... -->
</html>
```

ArmandPhilippot: I think you're right, there's no need to add a new section! However, it seems to me that we always propose valid examples so we need another `if` for the code to be valid. I'm going to post a new commit with a new proposal for the ``Return a `Response` object`` section.

src/content/docs/en/guides/on-demand-rendering.mdx
sarah11918: Sure, let's make them all `ins`!

src/content/docs/en/guides/on-demand-rendering.mdx
sarah11918: Sorry, where would you place this example?

src/content/docs/en/guides/on-demand-rendering.mdx
sarah11918, line 60:
``````suggestion
```astro title="src/pages/about-my-app.astro" ins={2}
``````

sarah11918 (approved):
Thank you @ArmandPhilippot these changes look great to me! I made one `ins` suggestion where I could, and pointed out the one where I couldn't. But happy with this!

I will say, for the changes to the "Features" sections, which is basically copied from the existing `server-side-rendering` page in `main` branch, you can feel free to PR them to that page directly too, and this will let the translators work on that updated content in advance of this branch merging eventually. That would make that merge more just "copy paste" of that entire Features section, which I really hadn't intended to change here!

src/content/docs/en/guides/on-demand-rendering.mdx
ArmandPhilippot: To avoid any confusion I am talking about [the example in the section `Astro.params`](https://5-0-0-beta.docs.astro.build/en/reference/api-reference/#astroparams).
Sorry, I didn't detail as it's not related to this PR, it's just by checking `Astro.params` that I thought the location of the example seemed strange.
The example shows the use of `getStaticPaths` which is usable with static builds while the sentence just above talks about SSR builds. So what would seem more logical to me would be between the two sentences (`In static builds,...` > `getStaticPaths` example > `In SSR builds, ...`).
But thinking back, I guess that was to avoid isolating the "SSR builds" sentence.

ArmandPhilippot:
Thanks for your feedback @sarah11918! I'm going to make a PR to update the current doc! 🫡

sarah11918:
!coauthor

src/content/docs/en/guides/on-demand-rendering.mdx
ArmandPhilippot: I almost forgot about this comment, sorry. Yes I think it would be useful to show both uses here. 💯  I'm going to prepare a PR.

PR #9676, opened 2024-10-15
Add example of script re-execution after transition

### LINKED ISSUE #8000: View Transitions Script Behavior
### 📋 Explain your issue

The [/view-transitions/#script-re-execution](https://docs.astro.build/en/guides/view-transitions/#script-re-execution) section I think is not correctly directing users to the truly happy path. Ideally for most users a lifecycle event would be listened for in order to deal with script rexecution. `data-astro-rerun` should only be used when necessary, after considering using a lifecycle event. 

Commonly users might reach for `data-astro-rerun` when in fact they want to only run a callback on certain pages. This section could help users stay on the happy path by giving an example of checking if an element is on the page before running a function related to it like [this example](https://events-3bg.pages.dev/jotter/astro/scripts/#executing-code-on-selected-pages-only).

apatel369:
#### Description (required)

Add an example about executing scripts after page transitions 
Copied from Sarah's comment on #8000 

- Closes #8000


sarah11918:
So happy to have this in docs! 🥳 🎉

PR #9680, opened 2024-10-15
feat: document new entrypoints

ematipico:
#### Description (required)

This PR documents a new upcoming feature for Astro v5, where the entrypoints of some functions can now accept `URL` other than a string


ArmandPhilippot:
@sarah11918 I think there is only one example where `Added in:` was in the middle of the section. I finally found it, it was [`next()`](https://docs.astro.build/en/reference/api-reference/#next). Chris and I found this a little strange, so what [we decided was](https://github.com/withastro/docs/pull/8943#discussion_r1700943160):
* Write the type including the changes under the title
* Do not add `Since`
* Write in full "Since Astro v...," at the beginning of the sentence introducing the new type/parameter.

So I guess we could do the same thing here to be consistent? On the other hand these are two pages that have not yet been updated with [the API format](https://contribute.docs.astro.build/reference/api-references/).

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -83,6 +83,30 @@ The path to the file that exports the dev toolbar app.
 }
 ```
 
+<p><Since v="5.0.0" /></p>
+
+The function accepts a `URL` as `entrypoint`:
```
sarah11918, line 88:
```suggestion
The function also accepts a `URL` as `entrypoint`:
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -325,6 +329,27 @@ export const onRequest = defineMiddleware(async (context, next) => {
 });
 ```
 
+<p><Since v="5.0.0" /></p>
+
+The function accept a `URL` for `entrypoint`:
```
sarah11918, line 334:
```suggestion
The function also accepts a `URL` for `entrypoint`:
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -370,6 +395,17 @@ When publishing your package (`@fancy/dashboard`, in this case) to npm, you must
 }
 ```
 
+<p><Since v="5.0.0" /></p>
+
+The function accepts a `URL` for `entrypoint`:
```
sarah11918, line 400:
```suggestion
The function also accepts a `URL` for `entrypoint`: 
```

sarah11918 (approved):
Thanks for looking into this @ArmandPhilippot , and yes, I noticed that these don't seem to be updated in general for our new API reference. So, I would say especially for beta docs, we can merge as-is, and then look at updating these two pages eventually to match our format. Merging then changing is less chaotic when these pages aren't being immediately translated underneath us. 😄 

@ematipico , I'm just making all the wording consistent, then this one is fine to merge for the beta release!

PR #9685, opened 2024-10-15
Update aws.mdx github CI docs

shftlvch:
#### Description (required)

Update the docs for deploying to S3+CF to use `withastro/action@v3` instead of npm.

This is helpful if you use Bun or any other package manager.

Discord: shftlvch

- Suggested label: docs


kevinzunigacuellar:
Hey @shftlvch,

Thanks so much for submitting this PR! I really appreciate the time and effort you put into it. To keep things as clear as possible, I think it’s best if we stick with the original steps. While the action might work, it’s actually designed specifically for use within GitHub Pages, so using it outside of that context could be confusing.

PR #9687, opened 2024-10-15
split API reference into modules

sarah11918:
#### Description (required)

In an attempt to make the API reference easier to navigate adds a new reference section specifically for the Astro API and each module individually.

~~Basic infra only! No content changes or improvements attempted! We should use this PR to figure out file name/folder conventions to use! Consistent page titles etc.! No decisions are final!~~

TRANSLATORS:
I'm sorry that there will still be more work to do after this that will again update the content on some of these pages. And, there is more content changed here than I expected, but most of it is moving content around in sections! You can see the list below for things that I expect to continue to change here, but I did want to get the new pages created first, then we can tackle specific content independently!

- [x] redirects ? I've kept the name of the main API reference page so in the worst case people aren't getting 404s visiting content that has moved. Is having redirect from section headings to a new page practical?
- [x] one error message was force overwritten to pass link checks here; will need to be updated at source once we are confident in what changes this pr actually makes [PR](https://github.com/withastro/astro/pull/12298)
- [x] Page introductions. These were previously sections of a longer page, so they don't really *introduce the page* nicely at all yet; they start with another heading immediately after the page title etc.
- [x] Lunaria tracking - some translated pages have had links updated but there should be no other changes needed to those pages, so they should be ignored by the tracker. Only the English pages, and the nav should be tracked.

Summary of notes left open below for follow up work:

- remove "built in components" page entirely (already duplication now with the `<ViewTransitions>`, `<Image>`, and `<Picture>` components. Instead, content will go to relevant place in docs, perhaps with a new `astro:components` page 
- `assets` page needs full documentation of `<Image>` properties -- copy from guide to start, then actually make proper reference matching the rest of the entries
- for the imported components, and moving to the Components/Syntax page for things like `<Fragment>` that are not imported
- So far, only `middleware` documents a required export (vs things you import from the module). Other pages could also document exports, but currently don't. For the sake of this PR, we are not adding new entries but will revisit for consistency and can consider adding the "Exports" section to other pages like `astro:actions`, `astro:content` etc.
- sidebar nav huge design means not touching the section headers etc. for Reference. Sidebar will change in current project, so not addressed here
- Is `api-reference` still the right name for the "leftover page"? Work is ongoing as we speak to clean up and focus the Runtime API page, so naming not changed here (plus, means we don't need any redirects here because the old page still exists). This particular page will be updated separately.
- we noticed (by the fact there were no broken links) that the view transitions guide page does not at any point link to the `astro:transitions` reference. This should make us check that page to make sure it's not doing too much and trying to be reference itself instead of just linking out to reference. (Note, so many things were previously undocumented about the API that it could simply mean that content just wasn't covered in the guide either, so not necessarily a problem, but for sure a smell.)


src/content/docs/en/reference/errors/actions-returned-invalid-data-error.mdx
```diff
@@ -19,6 +19,6 @@ import DontEditWarning from '~/components/DontEditWarning.astro'
 Action handler returned invalid data. Handlers should return serializable data types, and cannot return a Response object.
 
 **See Also:**
--  [Actions handler reference](/en/reference/api-reference/#handler-property)
+-  [Actions handler reference](/en/reference/modules/actions-api-reference/#handler-property)
```
sarah11918, line 22:
```suggestion
-  [Actions handler reference](/en/reference/modules/astro-action/#handler-property)
```

src/i18n/en/nav.ts
```diff
@@ -119,6 +131,16 @@ export default [
...
 	{ text: 'Error Reference', slug: 'reference/error-reference', key: 'reference/error-reference' },
 
+	{ text: 'Astro API Reference', header: true, type: 'api', key: 'api-reference' },
+	{ text: 'The Astro Global', slug: 'reference/api-reference', key: 'reference/api-reference' },
+	{ text: 'astro:actions', slug: 'reference/modules/astro-actions', key: 'reference/modules/astro-actions' },
+	{ text: 'astro:assets', slug: 'reference/modules/astro-assets', key: 'reference/modules/astro-assets' },
+	{ text: 'astro:content', slug: 'reference/modules/astro-content', key: 'reference/modules/astro-content' },
+	{ text: 'astro:i18n', slug: 'reference/modules/astro-i18n', key: 'reference/modules/astro-i18n' },
+	{ text: 'astro:middleware', slug: 'reference/modules/astro-middleware', key: 'reference/modules/astro-middleware' },
+	{ text: 'astro:view-transitions', slug: 'reference/modules/astro-transitions', key: 'reference/modules/astro-transitions' },
```
ArmandPhilippot, line 141: A little renaming... which shouldn't involve links! 😄 
```suggestion
	{ text: 'astro:transitions', slug: 'reference/modules/astro-transitions', key: 'reference/modules/astro-transitions' },
```

sarah11918: D'oh! I thought I caught them all! Thank you! 🫡

src/content/docs/en/reference/api-reference.mdx
delucis, line 1: Should the `title` of this page change?

header of the API reference page

Here the title is “API Reference” which is the name of the whole _group_ in the sidebar. And the sidebar label is “The Astro Global”, but this page also documents endpoint context, `getStaticPaths()`, and `import.meta`.

Bit tricky to name because

- Only `Astro` is actually a “global” thing — and only exists in `.astro` components
- Context is documented alongside it because they’re very similar, but it isn’t global (it’s passed as an argument to endpoints)
- `getStaticPaths()` is something a _user_ writes rather than something Astro gives you, although Astro requires it to have a specific shape
- `import.meta` is also actually “global” — but exists everywhere (more or less), not just in `.astro` files

delucis, line 1: And while we’re on the subject of title: if we do indeed come up with a better title (and sidebar label?), we may as well also rename the URL to match and add a redirect.

delucis (commented):
I think this looks great! Definitely helpful to split these out into the new pages. 🙌 

No comment on the Lunaria side of things, but left a couple of notes on things that stood out to me.

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -0,0 +1,163 @@
+---
+title: Actions API Reference
+i18nReady: true
+---
+import Since from '~/components/Since.astro';
+import ReadMore from '~/components/ReadMore.astro';
+
+## Actions (`astro:actions`)
```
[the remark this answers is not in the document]
sarah11918: Two quick mocks:

With h2:
image

without, which would let us move all content up a level:
image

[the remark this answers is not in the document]
sarah11918: Noting that I went ahead and made this change, as I think it's better for page content!

src/content/docs/en/reference/api-reference.mdx
sarah11918: Yes, I realized haven't fully massaged this page yet. It's basically everything that is not a separate module with no content changes to the page (other than deletions). Great call.

For the purposes of this PR, we *could* just pick the least worst name for all this stuff, not change the content right now, and then revisit this page content separately. Or, we can also handle the content while we're here!

src/content/docs/en/reference/modules/astro-actions.mdx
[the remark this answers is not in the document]
delucis: Nice!

src/content/docs/en/reference/modules/astro-middleware.mdx
delucis, line 1: Wondering if it’s worth `astro:middleware` being a subheading here to distinguish `onRequest` (which is something you write and export in `middleware.js`) from `sequence()`, `createContext()`, and `trySerializeLocals()` (which are exported from `astro:middleware`)?

(Also calls into question whether this should be labelled `astro:middleware` in the sidebar if it’s a reference for how middleware works _including_ that module.)

src/content/docs/en/reference/modules/astro-transitions.mdx
delucis, line 1: Similar comment here: we have a subheading for “Lifecycle events” and should probably also have one to group those items which are actually exported by the `astro:middleware` module.

delucis (commented):
Love the new intros @sarah11918! Left some more notes. Sorry if some of the headings comments slightly contradict what we discussed previously. Going through things in more details made me realise that clearly labelling with module names may be more important than I initially realised (especially for certain pages).

Fryuni:
It seems a bit weird to me having a "Reference" group and then a bunch of reference pages outside of that group. The word "reference" is also repeated in multiple links.

What if we nest them to keep the context?

| Current | Nested and shortened |
|---|---|
| image | image |

delucis:
> It seems a bit weird to me having a "Reference" group and then a bunch of reference pages outside of that group. The word "reference" is also repeated in multiple links, but not all of them so it is inconsistent.

I agree, but for context, this version of the sidebar is almost certainly a very short-lived iteration. We’re heading in the direction of something more like: https://docs-sidebar-topics.netlify.app/reference/configuration-reference/

src/content/docs/en/reference/modules/astro-assets.mdx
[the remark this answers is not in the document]
sarah11918: Yes, this is a lonely page indeed! I think now that we're doing *this* restructuring, it's *probably* more helpful to have the components in their own module page? 

So, maybe we add an `astro:components` page for those imported components and move the other components to their respective pages? I think `<Content />` and `<Fragment>` are already documented on the Markdown and Components pages respectively as far as syntax and Markdown imports are concerned.

src/content/docs/en/reference/modules/astro-i18n.mdx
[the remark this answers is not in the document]
sarah11918: I think an import heading could work, even if it looks a little long. The advantage to removing a higher level h2 was not needing everything to be nested one more for no real good reason. Without the name in the heading though, do we run the risk of searching not turning up a good page? If Algolia is not returning search hits for content in code blocks, then is e.g. `astro:content` discoverable enough through search? Does it matter? Do we re-title the page to include the module name itself?

sarah11918:
@Fryuni and @martrapp - would y'all be so kind as to check in particular the `astro:transitions` module here? 

Most content in these pages is unchanged, just moved out into separate pages. BUT, because we noticed that some of these pages contain things you *import from* the module mixed with other reference content relevant to the feature, Chris and I decided to put an import section at the top of each page to distinguish. 

The transitions page here is by far the most complicated one, and there were some items in there I was making best guesses about. So I'd love corrections/suggestions here!

sarah11918:
> I agree, but for context, this version of the sidebar is almost certainly a very short-lived iteration.

Yes, just for more context, a whole new sidebar/nav structure is coming before the end of the year so we're not at all concerned about getting the sidebar right here. The purpose of this PR is almost exclusively just to separate out this one huge page into individual ones. We're not even *really* improving content at this stage (even though some is needed!) -- this is entirely structural for two reasons:

- immediately: Algolia cannot index this whole page and important reference items are not getting indexed!
- if we at least split up the pages now, with minimal content change, then it's much easier for the translators to "catch up" with almost entirely copy/paste. Languages can prepare for the upcoming structural changes by at least having the right pages in place now with less disruption when the new sidebar comes.

src/content/docs/en/reference/modules/astro-assets.mdx
at-the-vr, line 1: >From Talking & Doc'ing
https://docs.astro.build/en/guides/images/#images-in-astro-files & https://docs.astro.build/en/reference/components-reference/#image- happen to resemble similar content with `astro-assets.mdx`. A trimming might be required to avoid duplicity

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -0,0 +1,373 @@
...
+  reference,
+ } from 'astro:content';
+
+import type { 
+  CollectionEntry,
+  CollectionKey,
+  ContentCollectionKey,
+  DataCollectionKey,
+  SchemaContext,
+ } from 'astro:content';
```
at-the-vr, line 33: > Suggestion stolen from Luiz & Hideoo during Talking & Doc'ing 😆 

```suggestion
import { 
  z,
  defineCollection,
  getCollection,
  getEntry,
  getEntries,
  reference,
  type  CollectionEntry,
  type  CollectionKey,
  type  ContentCollectionKey,
  type  DataCollectionKey,
  type  SchemaContext,
 } from 'astro:content';
```

src/content/docs/en/reference/modules/astro-i18n.mdx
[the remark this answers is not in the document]
delucis: > If Algolia is not returning search hits for content in code blocks, then is e.g. `astro:content` discoverable enough through search?

Great point!

I like this new subheading. Even if for some it’ll be a bit long, it’s reference! It’s _supposed_ to be exhaustive!

sarah11918 (commented):
left a review comment instead of a single comment. Have to finish the review. Boooo

src/content/docs/en/reference/modules/astro-assets.mdx
sarah11918: Yes, intentional until we address the built in components page! Will live with a tiny bit of duplication for now!

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -343,7 +343,7 @@ const allBlogPosts = await getCollection('blog');
 const graceHopperProfile = await getEntry('authors', 'grace-hopper');
 ```
 
-Both functions return content entries as defined by the [`CollectionEntry`](/en/reference/api-reference/#collection-entry-type) type.
+Both functions return content entries as defined by the [`CollectionEntry`](/en/reference/modules/astro-content/#collection-entry-type) type.
```
sarah11918, line 346:
```suggestion
Both functions return content entries as defined by the [`CollectionEntry`](/en/reference/modules/astro-content/#collectionentry) type.
```

src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -61,7 +61,7 @@ const posts = Object.values(await import.meta.glob('../posts/*.md', { eager: tru
 
 When fetching data from your collections via helper functions, your Markdown's frontmatter properties are available on a `data` object (e.g. `post.data.title`). Additionally, `body` contains the raw, uncompiled body content as a string.
 
-<ReadMore>See the full [CollectionEntry type](/en/reference/api-reference/#collection-entry-type).</ReadMore>
+<ReadMore>See the full [CollectionEntry type](/en/reference/modules/astro-content/#collection-entry-type).</ReadMore>
```
sarah11918, line 64:
```suggestion
<ReadMore>See the full [CollectionEntry type](/en/reference/modules/astro-content/#collectionentry).</ReadMore>
```

src/content/docs/ja/guides/content-collections.mdx
```diff
@@ -337,7 +337,7 @@ const allBlogPosts = await getCollection('blog');
 const graceHopperProfile = await getEntry('authors', 'grace-hopper');
 ```
 
-どちらの関数も、[`CollectionEntry`](/ja/reference/api-reference/#collection-entry-type)型で定義されたコンテンツエントリーを返します。
+どちらの関数も、[`CollectionEntry`](/ja/reference/modules/astro-content/#collection-entry-type)型で定義されたコンテンツエントリーを返します。
```
sarah11918, line 340:
```suggestion
どちらの関数も、[`CollectionEntry`](/ja/reference/modules/astro-content/#collectionentry)型で定義されたコンテンツエントリーを返します。
```

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -438,7 +438,7 @@ const blogEntries = await getCollection('blog');
 
 A component can also pass an entire content entry as a prop. 
 
-If you do this, you can use the [`CollectionEntry`](/en/reference/api-reference/#collection-entry-type) utility to correctly type your components props using TypeScript.  This utility takes a string argument that matches the name of your collection schema, and will inherit all of the properties of that collection's schema.
+If you do this, you can use the [`CollectionEntry`](/en/reference/modules/astro-content/#collection-entry-type) utility to correctly type your components props using TypeScript.  This utility takes a string argument that matches the name of your collection schema, and will inherit all of the properties of that collection's schema.
```
sarah11918, line 441:
```suggestion
If you do this, you can use the [`CollectionEntry`](/en/reference/modules/astro-content/#collectionentry) utility to correctly type your components props using TypeScript.  This utility takes a string argument that matches the name of your collection schema, and will inherit all of the properties of that collection's schema.
```

src/content/docs/ja/guides/content-collections.mdx
```diff
@@ -432,7 +432,7 @@ const blogEntries = await getCollection('blog');
 
 コンポーネントは、コンテンツエントリー全体をプロパティとして渡すこともできます。
 
-この場合、[`CollectionEntry`](/ja/reference/api-reference/#collection-entry-type)ユーティリティを使用して、TypeScriptでコンポーネントのプロパティを適切に型付けできます。 このユーティリティは、コレクションスキーマの名前と一致する文字列引数を取り、そのコレクションのスキーマのすべてのプロパティを継承します。
+この場合、[`CollectionEntry`](/ja/reference/modules/astro-content/#collection-entry-type)ユーティリティを使用して、TypeScriptでコンポーネントのプロパティを適切に型付けできます。 このユーティリティは、コレクションスキーマの名前と一致する文字列引数を取り、そのコレクションのスキーマのすべてのプロパティを継承します。
```
sarah11918, line 435:
```suggestion
この場合、[`CollectionEntry`](/ja/reference/modules/astro-content/#collectionentry)ユーティリティを使用して、TypeScriptでコンポーネントのプロパティを適切に型付けできます。 このユーティリティは、コレクションスキーマの名前と一致する文字列引数を取り、そのコレクションのスキーマのすべてのプロパティを継承します。
```

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -0,0 +1,457 @@
...
+title: View Transitions API Reference
+i18nReady: true
+tableOfContents:
+  minHeadingLevel: 2
+  maxHeadingLevel: 6
+---
+import Since from '~/components/Since.astro';
+import ReadMore from '~/components/ReadMore.astro';
+
+<p><Since v="3.2.0" /></p>
```
sarah11918, line 11:
```suggestion
<p><Since v="3.0.0" /></p>
```

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -0,0 +1,457 @@
...
+
+## Imports from `astro:transitions`
+
+```ts
+import { ViewTransitions, fade, slide } from 'astro:transitions';
+```
+
+### `<ViewTransitions />`
+
+<p><Since v="2.9.0" /></p>
```
sarah11918, line 25:
```suggestion
<p><Since v="3.0.0" /></p>
```
(stable, not experimental)

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -0,0 +1,457 @@
...
+tableOfContents:
+  minHeadingLevel: 2
+  maxHeadingLevel: 6
+---
+import Since from '~/components/Since.astro';
+import ReadMore from '~/components/ReadMore.astro';
+
+<p><Since v="3.0.0" /></p>
+
+These modules provides functions to control and interact with the View Transitions API and client-side router.
```
delucis, line 13: Noun-verb agreement

```suggestion
These modules provide functions to control and interact with the View Transitions API and client-side router.
```

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -0,0 +1,381 @@
...
+  defineCollection,
+  getCollection,
+  getEntry,
+  getEntries,
+  reference,
+  type CollectionEntry,
+  type CollectionKey,
+  type ContentCollectionKey,
+  type DataCollectionKey,
+  type SchemaContext,
```
delucis, line 29: Forgot to remove these when splitting the types out into a separate subheading

```suggestion
```

src/content/docs/en/reference/modules/astro-middleware.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+**Type:** `APIContext`
+</p>
+
+The first argument of `onRequest()` is a context object. It mirrors many of the `Astro` global properties.
+
+<ReadMore>See [Endpoint contexts](/en/reference/api-reference/#endpoint-context) for more information about the context object.</ReadMore>
+
+### `next()`
+
```
delucis, line 108:
````suggestion
## Middleware exports

When defining your project’s middleware in `src/middleware.js`, export the following user-defined functions:

### `onRequest()`

**Type:** `(context: APIContext, next: MiddlewareNext) => Promise<Response> | Response | Promise<void> | void`

A required exported function from `src/middleware.js` that will be called before rendering every page or API route. It receives two arguments: [context](#context) and [next()](#next). `onRequest()` must return a `Response`: either directly, or by calling `next()`.

```js title="src/middleware.js"
export function onRequest (context, next) {
    // intercept response data from a request
    // optionally, transform the response
    // return a Response directly, or the result of calling `next()`
    return next();
};
```

Your `onRequest()` function will be called with the following arguments:

#### `context`

<p>

**Type:** `APIContext`
</p>

The first argument of `onRequest()` is a context object. It mirrors many of the `Astro` global properties.

<ReadMore>See [Endpoint contexts](/en/reference/api-reference/#endpoint-context) for more information about the context object.</ReadMore>

#### `next()`

````

delucis (commented):
Last couple of small changes!

delucis (approved):
We approve of ourselves!

Fryuni (approved):

PR #9688, opened 2024-10-15
i18n(de): update translation basics/astro-components.mdx

trueberryless:
#### Description

Translation of `basics/astro-components.mdx` into German language.

### DIFF
--- a/src/content/docs/de/basics/astro-components.mdx
+++ b/src/content/docs/de/basics/astro-components.mdx
@@ -12,7 +12,6 @@ Das Wichtigste an Astro-Komponenten ist, dass sie **nicht auf dem Client rendern
 
 Wenn deine Astro-Komponente clientseitige Interaktivität benötigt, kannst du [Standard-HTML-`<script>`-Tags](/de/guides/client-side-scripts/) oder [UI-Framework-Komponenten](/de/guides/framework-components/#interaktive-komponenten-hydratisieren) hinzufügen.
 
-
 ## Komponentenstruktur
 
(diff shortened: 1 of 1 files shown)


trueberryless:
Prettier is messing up a lot in this file. I'm new and would like to know how other translators that a working with Prettier too are handling these issues. This PR should only contain a few changes but Prettier formatted the whole file which unfortunately is not expected in many places. There is a `.prettierrc` file, so I thought it is possible to work with Prettier enabled on save, but this PR shows how much Prettier messes up the formatting especially on code blocks...

ArmandPhilippot:
Hi @trueberryless and welcome to Team Docs! 🎉  I don't speak German so I can't judge the translation. I landed here only out of curiosity.

To answer your question since I also translate (for the French version), I never noticed any problem with Prettier and without adding any modifications on my side if I remember correctly.
There is indeed a `.prettierrc` file but there is also a `.prettierignore` file which contains `*.mdx`. So normally Prettier should not format this file.

If you are using VS Code, you can use the `Output` tab, then select `Prettier` instead of `Tasks` (selected by default, it seems to me) in the dropdown and try to save again a `.mdx`. You should see "*File is ignored, skipping.*". If not, I don't understand why. Maybe a global configuration or a plugin (not Prettier's since I leave it activated while translating) that conflicts... Are you opening your editor at the root of `astro/docs`? If you only open the file to translate rather than the whole project, I think `.prettierignore` is not taken into account. Maybe that's the problem?

If you are using another editor/environment, feel free to give more information, it may help to understand the origin of the problem and help you next time! 😉

trueberryless:
omg, nice short story time:

first of all, 100% same setup so no diff but prettier still runs on *.md and *.mdx files...

I tried it just now (sry that it took so long) and it turns out there is a setting for prettier (I think it explains itself xD):

image

I'll explain it nonetheless (xD): If someone else runs into this issue, check your VSC settings and go to Prettier settings, scroll until you see the setting above but in the input field type the following so that it works with astro repo:

image

wish you all the best luck and fun for future translations, im gonna close this PR and make it new...

thx  @ArmandPhilippot for answering and investigeting my concerns very well!

PR #9702, opened 2024-10-16
fix(api-reference): add a code snippet for Astro.params with SSR

ArmandPhilippot:
#### Description (required)

Based on a discussion in #9673, I updated the `Astro.params` section with the following changes:
* Move the code snippet under `In SSR builds,` to `In Static builds`
* Add a new code snippet under `In SSR builds` to show an example specific to SSR

- Suggested label:  code snippet update


sarah11918 (approved):
Looks great! Love this change @ArmandPhilippot

PR #9715, opened 2024-10-17
Fix typo

kuldar:
Probably should be astro:content instead of astro:client

#### Description (required)

The docs had `astro:client` instead of `astro:content` in one place.

- Closes #
- Suggested label:


yanthomasdev (approved):
LGTM! Thanks @kuldar :raised_hands:

PR #9744, opened 2024-10-19
i18n(fr): Update `nav.ts`

pioupia:
#### Description (required)
Update `nav.ts` file because of missing and malformed translation key.

- Suggested label: i18n

I participate to the Hacktoberfest!


ArmandPhilippot (approved):
LGTM! Just a note: I guess the order doesn't matter, otherwise I think `reference/error-reference` should be after `guides/typescript`.

ArmandPhilippot:
Hi @pioupia! I was checking for any other inconsistencies in the menu... What do you think about updating `reference/adapter-reference` as well? All other translations start with `API` while this one ends with `API`, which is a bit strange. What do you think of `API de l'Adaptateur` or `API des Adaptateurs`?

pioupia:
> Hi @pioupia! I was checking for any other inconsistencies in the menu... What do you think about updating `reference/adapter-reference` as well? All other translations start with `API` while this one ends with `API`, which is a bit strange (well, there is also `reference/container-reference` but the page is not translated so it is perhaps less annoying). What do you think of `API de l'Adaptateur` or `API des Adaptateurs`?

Yeah, this is a good idea. On the `adapter-reference`, I think that the title is pretty bad "Adaptateur l'API de Astro". Maybe "Adaptateur de l'API d'Astro" is better.

Anyway, for this topic, I'm going to update the order even if it's not important, it will help us keep consistency between languages and design.
Also, I'll rename "Adaptateur API" in "API de l'Adaptateur" as you suggested above, and "Conteneur API" in "API de Conteneur".

ArmandPhilippot (approved):
LGTM! Thanks @pioupia, much better!

> I think that the title is pretty bad "Adaptateur l'API de Astro". Maybe "Adaptateur de l'API d'Astro" is better.

Yeah I saw that too... 😅 I think in this case it would be `API de l'Adaptateur d'Astro`, it still the Adapter API. But as you said, this is probably outside the scope of this PR.

yanthomasdev (approved):
LGTM!

PR #9703, opened 2024-10-16
docs: add better-auth guide for authentication

Bekacru:
Hey there! This PR adds a guide for using Better Auth with Astro for authentication. It’s just a starting point, so feel free to tweak it however you like. Better Auth is still in beta, but a stable release is coming soon, and it's pretty solid already. With Lucia being deprecated and Auth.js lack of support for astro, Better Auth could be a great open-source alternative for anyone looking to implement authentication in Astro.

PR #9747, opened 2024-10-19
i18n(fr): update `guides/authentication.mdx`

ArmandPhilippot:
#### Description (required)

Adds the Better Auth section in the French version of `guides/authentication.mdx` (see #9703)

- Suggested label: i18n


src/content/docs/fr/guides/authentication.mdx
```diff
@@ -188,6 +189,157 @@ const session = await getSession(Astro.request);
 - [`auth-astro` sur GitHub](https://github.com/nowaythatworked/auth-astro?tab=readme-ov-file#auth-astro)
 - [Documentation Auth.js](https://authjs.dev/)
 
+## Better Auth
+
+Better Auth est un framework indépendant d'authentification (et d'autorisation) pour TypeScript. Il fournit un ensemble complet de fonctionnalités prêtes à l'emploi et comprend un écosystème de plugins qui simplifie l'ajout de fonctionnalités avancées.
```
ArmandPhilippot, line 194: Not sure about my translation here... Would `agnostique` have been better than `indépendant`? It seemed strange to me in this context. So, if you have a better suggestion for `framework-agnostic authentication (and authorization) framework for TypeScript`, I'm all ears! 😄

pioupia: From the [Cambridge dictionary](https://dictionary.cambridge.org/dictionary/english/agnostic), "Agnostic" (in computer theme) means "A software that can be used with many different types of platform".
So `indépendant` works perfectly for me. And even if on your usage it has a double sense, it doesn't really matter, we touch the essential.

src/content/docs/fr/guides/authentication.mdx
```diff
@@ -188,6 +189,157 @@ const session = await getSession(Astro.request);
...
+ 
+const session = await auth.api.getSession({
+	headers: Astro.request.headers,
+});
+---
+
+<p>{session.user?.name}</p>
+```
+
+Vous pouvez également utiliser l'objet `auth` pour protéger vos routes à l'aide d'un middleware. L'exemple suivant vérifie si un utilisateur essayant d'accéder à une route de tableau de bord connectée est authentifié et le redirige vers la page d'accueil dans le cas contraire.
```
pioupia, line 318: I have some trouble with this sentence. Like "tableau de bord connectée" mean first of all "connected to internet", but the real purpose of this sentence is "logged-in".
I have no idea of how we can formulate this without being to boring (Réberbatif).
`L'exemple suivant vérifie si, un utilisateur essayant d'accéder à une route de tableau de bord nécessitant d'être connecté, est authentifié et le redirige vers la page d'accueil dans le cas contraire.`
But for me it's too much.

pioupia (approved):
All look good for me. Just need to open the discussion about one sentence.

src/content/docs/fr/guides/authentication.mdx
ArmandPhilippot: Yeah I know the meaning in English, but I wasn't sure if we also use this term in this sense in French so I was worried that `indépendant` didn't quite capture the definition.
On the other hand, I couldn't find a better word without becoming too wordy compared to the original version. So I preferred to indicate my doubt, it's always better to have a second opinion!
If you think we keep the same meaning, then that's fine with me!

src/content/docs/fr/guides/authentication.mdx
ArmandPhilippot: You're right, maybe it wasn't the right term! We could use `protégé`: either `accéder à une route de tableau de bord protégé` or `accéder à une route protégée de tableau de bord`. Does that seem clearer to you?

ArmandPhilippot: Ah I didn't see your edit. I must have opened the page before... So we agree!

yanthomasdev (approved):
LGTM!

PR #9748, opened 2024-10-19
i18n(ja): fix guides/images.mdx, output result was wrong [i18nIgnore]

Ryoga-exe:
#### Description (required)

Fix guides/images.mdx. 

Fixed an incorrect example in the output section of the [Astro Image Guide](https://docs.astro.build/ja/guides/images/#%E3%81%9D%E3%81%AE%E4%BB%96%E3%81%AE%E3%83%97%E3%83%AD%E3%83%91%E3%83%86%E3%82%A3). The example was incorrectly showing an `html` element, but it should output an `img` element instead, so I have replaced it accordingly.


yanthomasdev (approved):
Nice catch! LGTM!

PR #9751, opened 2024-10-20
Fix incorrect prop destructuring in tutorial

yusifsalam:
#### Description (required)

Fix incorrect prop destructuring in tutorial example.

- Closes #
- Suggested label:  tutorial

### DIFF
--- a/src/content/docs/en/tutorials/add-content-collections.mdx
+++ b/src/content/docs/en/tutorials/add-content-collections.mdx
@@ -208,7 +208,7 @@ The steps below show you how to extend the final product of the Build a Blog tut
     }
 
     const { post } = Astro.props;
-    const { Content } = await render(post );
+    const { Content } = await render(post);
     ---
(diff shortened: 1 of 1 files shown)


apatel369 (approved):
Nice catch!

src/content/docs/en/tutorials/add-content-collections.mdx
```diff
@@ -207,7 +207,7 @@ The steps below show you how to extend the final product of the Build a Blog tut
       }));
     }
 
-    const { entry } = Astro.props;
+    const { post } = Astro.props;
     const { Content } = await render(post );
```
sarah11918, line 211:
```suggestion
    const { Content } = await render(post);
```
Seems to be an unneeded space, while we're here?

sarah11918 (approved):
Thank you for spotting and taking the time to fix this! I just noticed a space that seemed off in the next line, so we can just fix both at the same time. Welcome to Team Docs! :partying_face:

PR #9755, opened 2024-10-21
feat: update Vercel imports

ematipico:
#### Description (required)

This PR updates all the `@astrojs/vercel/serverless` and `@astrojs/vercel/static` entrypoints to use the new upcoming one: `@astrojs/vercel`.

I also removed the `### Choosing a target` paragraph because there's no need anymore


sarah11918 (commented):
Everything on the Vercel page looks perfect, thank you @ematipico ! We will need an in the v5 upgrade guide too, since this will require people to update their code.

I'm thinking since we already have a place where `@astrojs/mdx` upgrade is required, we can maybe put this one right below as `@astrojs/vercel`?

image

ematipico:
No breaking changes are involved between `astro` and `@astrojs/vercel`, so there's no need to update that section of the website. The entrypoints `@astrojs/vercel/serverless` and `@astrojs/vercel/static` are *deprecated* but still valid.

ematipico:
Sure, but that section talks about **Astro dependencies**. The fact that we talk about `@astrojs/mdx` is because there was some code **in core** that was required by `@astrojs/mdx`. It's essentially an **internal** architecture mistake that we fixed in v5.

In this PR of mine, there's no correlation *inside Astro **core*** and the vercel adapter

sarah11918:
ok, then maybe it should be one of the breaking change items in the below section?

ematipico:
But there aren't breaking changes 😅 `@astrojs/vercel/serverless` **still** works. The breaking changes of the vercel adapter (and the others) are others, mostly caused by Astro v5. This isn't one of them.

sarah11918:
Yes, you're right! Not a breaking change since the deprecated code will still work. But it still means that there could be code that people have in their projects that is different from what's shown in docs with no mention of the way things used to be.

So I'll think about whether there's a good way to mention this in the upgrade guide, but that doesn't block this PR!

sarah11918 (approved):

Issue #9753, opened 2024-10-20
Observations on beta documentation about the new content loaders


Because789:
### 📚 Subject area/topic

Content loaders

### 📋 Page(s) affected (or suggested, for new content)

[Inline loaders](https://5-0-0-beta.docs.astro.build/en/guides/content-collections/#inline-loaders)
[Inline loaders](https://5-0-0-beta.docs.astro.build/en/reference/content-loader-reference/#inline-loaders)

### 📋 Description of content that is out-of-date or incorrect

I'm making myself familiar with the new content loaders. I'm sure, the docs on them are still a work in progress, here anyway some observations: 

- There is some redundant information (see provided links). Maybe the whole [Building a custom loader](https://5-0-0-beta.docs.astro.build/en/guides/content-collections/#building-a-custom-loader) could be moved to [Astro Content Loader API](https://5-0-0-beta.docs.astro.build/en/reference/content-loader-reference).
- This sentence gives me shudders `An inline loader is an async function that returns an array or object containing entries. Use this for simple loaders, particularly those that are defined inline in the src/content/config.ts file.` ([Astro Content Loader API](https://5-0-0-beta.docs.astro.build/en/reference/content-loader-reference).) I guess  `An inline loader is an async function that returns an array or object containing entries. Use this for simple loaders defined in the src/content/config.ts file.` could be enough. 

Edit: The `loader.ts` example in [Object loaders](https://5-0-0-beta.docs.astro.build/en/reference/content-loader-reference/#object-loaders) has several issues:

- Semicolon instead of comma line 9, missing comma line 14
-  `Cannot find name 'loadFeedData'.` line 13
- Missing `z` import line 18

sarah11918:
Thank you for pointing out the errors in the Object loaders example! I would welcome a PR to the beta branch to fix these! I am adding the "help wanted" label so that you or anyone in the community is welcome to contribute the updates.

And yes, all of our beta docs are very much a work in progress! :sweat_smile: As more people use this (experimental or stable-but-beta) feature, we see how people are using the feature in practice and what kind of documentation is helpful (and where). As you've identified, a loader can be as simple as an inline function or an involved setup. We specifically chose some redundancy to allow this content to be easily available to those creating inline loaders, who do not necessarily need to be taken to the full reference page (It is an extra page navigation, extra context switch and also contains information they may not need). The current  structure allows them to complete their content collections configuration without leaving the current guide.

[...]

PR #9777, opened 2024-10-23
Fix loader example in content-loader-reference

Because789:
This partly solves https://github.com/withastro/docs/issues/9753

- Closes #
- Suggested label:


sarah11918 (approved):
Thank you for jumping in to fix this code sample! It is very much appreciated!

PR #9778, opened 2024-10-23
Update markdown-content.mdx DE

flnzba:
added frontmatter.title

#### Description (required)

added frontmatter to post.title because could not recognize title from post without mentioning frontmatter

<ul>
  {posts.map(post => <li>{post.**frontmatter**.title}</li>)}
  
</ul>

DE

#### Related issues & labels (optional)

- Closes #
- Suggested label: 

YES

PR #9779, opened 2024-10-23
Markdown Content Change DE

flnzba:
#### Description (required)

reassigned posts because filter method can not be applied if posts is not reassigned

DE

- Closes #
- Suggested label:


yanthomasdev:
Sharing my comment from your other PR here: https://github.com/withastro/docs/pull/9778#issuecomment-2432850051

sarah11918:
See the other PR for explanation, but we are deleting this file instead of updating it because the translation has deviated too far from the original!

PR #9801, opened 2024-10-25
reorg of images and assets guide and reference content

sarah11918:
#### Description (required)

This PR 

- moves reference information out of the Images guide and into `astro:assets` (leaving the guide to be more guidey)
- deletes the `<Image>` and `<Picture>` components from the built-in components reference as they are now in `astro:assets` AND in the Images guide. ( also deletes the `<ViewTransitions>` component for the same reason)
- updates any necessary links, including manually overwriting some error messages here in this branch

Tasks
- [x]  Add full API reference headings to entries (merci, Armand!) 
- [x] Update error files at source that have been manually overwritten here once the exact links are known ([PR here](https://github.com/withastro/astro/pull/12331))
- [ ] Lunaria tracking statement


src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,187 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
-- densities
-- widths
+#### Image properties
 
-In addition to the properties above, the `<Image />` component accepts all properties accepted by the HTML `<img>` tag.
+The `<Image />` component accepts all properties accepted by the HTML `<img>` tag in addition to the properties described below.
 
-See more in the [Images Guide](/en/guides/images/#image--astroassets).
+##### src (required)
+
```
ArmandPhilippot, line 57:
```suggestion

<p>

**Type:** `ImageMetadata | string | Promise<{ default: ImageMetadata }>`
</p>

```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,187 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
+  <Image
+    src="https://example.com/remote-image.jpg"
+    alt="descriptive text"
+    width="200"
+    height="150"
+  />
+  ```
+
+##### alt (required)
+
```
ArmandPhilippot, line 99:
```suggestion

<p>

**Type:** `string`
</p>

```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,187 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
+Use the required `alt` attribute to provide a string of [descriptive alt text](https://www.w3.org/WAI/tutorials/images/) for images.
+
+If an image is merely decorative (i.e. doesn't contribute to the understanding of the page), set `alt=""` so that screen readers and other assistive technologies know to ignore the image.
+
+##### width and height (required for images in `public/`)
+
```
ArmandPhilippot, line 105:
```suggestion

<p>

**Type:** `number | undefined`
</p>

```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,187 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
+These properties define the dimensions to use for the image.
+
+When using images in their original aspect ratio, `width` and `height` are optional. These dimensions can be automatically inferred from image files located in `src/`. For remote images, add [the `inferSize` attribute set to `true`](#infersize) on the `<Image />` or `<Picture />` component or use [`inferRemoteSize()` function](#inferremotesize).
+
+However, both of these properties are required for images stored in your `public/` folder as Astro is unable to analyze these files.
+
+##### densities
+
+<p><Since v="3.3.0" /></p>
```
ArmandPhilippot, line 114:
```suggestion
<p>

**Type:** `(number | `${number}x`)[] | undefined`<br />
<Since v="3.3.0" />
</p>
```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,187 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
+  width="800"
+  height="450"
+  loading="lazy"
+  decoding="async"
+/>
+```
+
+##### widths
+
+<p><Since v="3.3.0" /></p>
```
ArmandPhilippot, line 153:
```suggestion
<p>

**Type:** `number[] | undefined`<br />
<Since v="3.3.0" />
</p>
```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,187 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
+  alt="A description of my image."
+  width="1600"
+  height="900"
+  loading="lazy"
+  decoding="async"
+/>
+```
+
+##### format
+
```
ArmandPhilippot, line 201:
```suggestion

<p>

**Type:** `ImageOutputFormat | undefined`
</p>

```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,187 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
+You can optionally state the [image file type](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#common_image_file_types) output to be used.
+
+By default, the `<Image />` component will produce a `.webp` file.
+
+##### quality
+
```
ArmandPhilippot, line 207:
```suggestion

<p>

**Type:** `ImageQuality | undefined`
</p>

```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,187 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
+`quality` is an optional property that can either be:
+- a preset (`low`, `mid`, `high`, `max`) that is automatically normalized between formats.
+- a number from `0` to `100` (interpreted differently between formats).
+
+##### inferSize
+
+<p><Since v="4.4.0" /></p>
```
ArmandPhilippot, line 214:
```suggestion
<p>

**Type:** `boolean`<br />
<Since v="4.4.0" />
</p>
```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -93,23 +262,53 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
-See more in the [Images Guide](/en/guides/images/#picture-).
+See the [Images Guide](/en/guides/images/#picture-) for usage.
 
-#### Properties
+#### Picture properties
 
-`<Picture />` accepts all the properties of the `<Image />` component, plus the following:
+`<Picture />` accepts all the properties of [the `<Image />` component](#image-properties), plus the following:
 
 ##### `formats`
```
ArmandPhilippot, line 272:
```suggestion

<p>

**Type:** `ImageOutputFormat[]`
</p>

```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -93,23 +262,53 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
 
-An array of image formats to use for the `<source>` tags. By default, this is set to `['webp']`.
+An array of image formats to use for the `<source>` tags. Entries will be added as `<source>` elements in the order they are listed, and this order determines which format is displayed. For the best performance, list the most modern format first (e.g. `webp` or `avif`). By default, this is set to `['webp']`.
 
 ##### `fallbackFormat`
```
ArmandPhilippot, line 276:
```suggestion

<p>

**Type:** `ImageOutputFormat`
</p>

```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -93,23 +262,53 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
 
-Format to use as a fallback value for the `<img>` tag. Defaults to `.png` for static images, `.gif` for animated images, and `.svg` for SVG files.
+Format to use as a fallback value for the `<img>` tag. Defaults to `.png` for static images (or `.jpg` if the image is a JPG), `.gif` for animated images, and `.svg` for SVG files.
 
 ##### `pictureAttributes`
```
ArmandPhilippot, line 280:
```suggestion

<p>

**Type:** `HTMLAttributes<'picture'>`
</p>

```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -157,3 +356,14 @@ type GetImageResult = {
   };
 }
 ```
+
+### inferRemoteSize()
+
+<p><Since v="4.12.0" /></p>
```
ArmandPhilippot, line 362:
```suggestion
<p>

**Type:** `(url: string) => Promise<Omit<ImageMetadata, 'src' | 'fsPath'>>`<br />
<Since v="4.12.0" />
</p>
```

ArmandPhilippot (commented):
Looks good to me! But since this is no longer a guide but a reference, I thought we could take the opportunity to add the property types. If you want to check:
* [`<Image />` types](https://github.com/withastro/astro/blob/e482762cb216681c8d70266e9427de8a61bd3fb0/packages/astro/src/assets/types.ts#L69)
* [`<Picture />` types](https://github.com/withastro/astro/blob/e482762cb216681c8d70266e9427de8a61bd3fb0/packages/astro/components/Picture.astro#L9)

The only thing that wasn't listed was `alt`, so I didn't copy/paste it, but other than `string` I don't think any other value is possible.

Maybe `densities` and `inferRemoteSize()` types are not clear enough for those who don't know Typescript. But not putting them on would have seemed odd I think.

src/content/docs/en/guides/images.mdx
```diff
@@ -91,191 +91,17 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
...
----
-<Image src="https://example.com/cat.png" inferSize alt="A cat sleeping in the sun." />
-```
-
-`inferSize` can fetch the dimensions of a [remote image from a domain that has not been authorized](#authorizing-remote-images), however the image itself will remain unprocessed.
-
-###### inferRemoteSize()
-
-<p><Since v="4.12.0" /></p>
+In addition to the properties above, the `<Image />` component accepts all properties accepted by the HTML `<img>` tag.
```
at-the-vr, line 102: This sentence happens to be duplicated in [Additional properties
](https://deploy-preview-9801--astro-docs-2.netlify.app/en/guides/images/#additional-properties). Could that be removed?

at-the-vr (commented):
Looking good already 🚀. Just 1 nit on my end

src/content/docs/en/guides/images.mdx
```diff
@@ -91,191 +91,17 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
...
 #### Properties
 
-##### src (required)
+- src (required)
+- alt (required)
+- width and height (required for `public/` and remote images)
+- format
+- quality
+- densities
+- widths
```
delucis, line 100: It feels weird to still include this “reference-y” if brief list of props here (and the same for `<Picture>` below). In general, although somme stuff has been moved, both thes `<Image>` and `<Picture>` sections still feel like reference entries to me: “This component exists. It does X. Code snippet. There are props.”

Is there a world where these get rewritten to follow a guide structure?

Like thinking use case-forward instead of API-forward something along the lines of this (assuming breaking up by component type still makes sense):

```diff
- ### `<Image />`
+ ### Display optimized images with the `<Image />` component

...

- ### `<Picture />`
+ ### Create responsive images with the `<Picture />` component

...

- ### `<img>`
+ ### Display unprocessed images with the HTML `<img>` tag
```

delucis (commented):
One comment on the future of the images guide, but looks good and +1 to Armand’s suggestions too 🚀

src/content/docs/en/guides/images.mdx
sarah11918: ooh, that looks lovely. OK, let's work that in!

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -47,19 +48,224 @@ import myImage from "../assets/my_image.png"; // Image is 1600x900
...
+
+When using images in their original aspect ratio, `width` and `height` are optional. These dimensions can be automatically inferred from image files located in `src/`. For remote images, add [the `inferSize` attribute set to `true`](#infersize) on the `<Image />` or `<Picture />` component or use [`inferRemoteSize()` function](#inferremotesize).
+
+However, both of these properties are required for images stored in your `public/` folder as Astro is unable to analyze these files.
+
+##### densities
+
+<p>
+
+**Type:** `(number | `${number}x`)[] | undefined`<br />
```
ArmandPhilippot, line 131: I think this is the type that is causing problems with `Check links`. Maybe we could use the `code` tag instead:
```suggestion
**Type:** <code>(number | `${number}x`)[] | undefined<code><br />
```

sarah11918:
OK, would love another pair of eyes on the Image Guide specifically! I removed/rearranged some stuff to make it more "guide"-y and let the reference stuff live only in reference!

Another pass, @at-the-vr , @ArmandPhilippot ?

src/content/docs/en/guides/images.mdx
```diff
@@ -89,197 +86,9 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
...
-import { inferRemoteSize } from 'astro:assets';
-const {width, height} = await inferRemoteSize("https://example.com/cat.png");
-```
-
-##### Additional properties
-
-In addition to the properties above, the `<Image />` component accepts all properties accepted by the HTML `<img>` tag.
+Additionally, the `<Image />` component accepts all properties accepted by the HTML `<img>` tag.
 
 For example, you can provide a `class` to the final `<img>` element.
```
ArmandPhilippot, line 93: I wonder if `Additionally` still makes sense now that we're not listing properties anymore.
```suggestion
The `<Image />` component accepts all properties accepted by the HTML `<img>` tag. For example, you can provide a `class` to the final `<img>` element.
```
I guess it follows on from the `ReadMore`... but scanning the page at first glance I didn't see a logical continuation with the previous paragraph. So, I figure `In addition`/`Additionally` is no longer useful.

src/content/docs/en/guides/images.mdx
```diff
@@ -457,13 +200,34 @@ For remote images, use the image's **full URL** as the `src` value:
...
+```astro title="src/components/BlogPostImage.astro"
+---
+import { Image } from 'astro:assets';
+
+const { src, ...attrs } = Astro.props;
+---
+<Image src={src} {...attrs} />
+
+<style>
+  img :global(img), svg {
```
ArmandPhilippot, line 225: It seems weird to me to target an image in an image... Doesn't the `<Image />` component return an `` tag in all cases? (whether the source is a jpg or an svg) So I would say we just need `img` here... but I am not sure (maybe `:global` is required).
```suggestion
  img {
```

ArmandPhilippot (approved):
It's even better than before! I just left two suggestions (don't worry, they are not links 😅 ), but feel free to ignore them, I'm not sure. One is about the wording, the other is a CSS code snippet that seems strange to me.

src/content/docs/en/guides/images.mdx
sarah11918: I will need to call in @delucis for assistance here!

src/content/docs/en/guides/images.mdx
ArmandPhilippot: Yes I think it's a question of perception. So your explanation makes sense.
For me, the `ReadMore` with the icon makes the sentence more of an "aside" than an extension so I attach the paragraph to the one before.
So I'm fine with keeping it as is, I just wanted to note it "just in case".

src/content/docs/en/guides/images.mdx
ArmandPhilippot: I've never used `<Image />` with an svg so I might be wrong... What seemed strange to me was mainly `img :global(img)`.

at-the-vr: +1 to Armand. SVGs are not processed in `<Image>` besides retaining `width` & `height` attributes. `:global` is related with [css modules](https://github.com/css-modules/css-modules/blob/master/docs/composition.md#exceptions) and is throwing me off too

src/content/docs/en/guides/images.mdx
sarah11918: Yup, I get that! Let me take one more shot at it for removing weirdness :smile:

sarah11918: @ArmandPhilippot one last try?  :smile:

src/content/docs/en/guides/images.mdx
delucis: Yeah the styles here don’t really make sense. You can’t have `</img>` and this style would create a selector like `img[data-astro-cid-XXX] img` so wouldn’t actually ever match what is generated by this component. The `svg` is also not relevant because even when using `<Image>` with an SVG file, you still get an `` tag, not an inline `<svg>`.

I think @ArmandPhilippot’s suggestion is correct. I did a quick test and it appears that `<Image>` inherits the containing component’s scoped classes, so just `img {}` will work well in this case — no need for `:global()` (which would also introduce other issues, as it would impact all images on the site).

src/content/docs/en/guides/images.mdx
ArmandPhilippot: @sarah11918  The flow seems better to me like this, thank you! 🙌🏽

src/content/docs/en/guides/images.mdx
sarah11918: Thank you all!

sarah11918:
I think we're all happy with this then! Just the Lunaria tracking statement, but that can wait (should wait?) until after I've had my espresso!  :laughing:

sarah11918:
`@lunaria-track:src/content/docs/en/**/*.mdx;src/i18n/en/nav.ts`

Fryuni:
I think we are missing links on the `<Image />` component in each section. For example, think of a user wanting to add a CMS-hosted image with 2x density who finds this page:
- They open the page and click on "Using Images from a CMS or CDN" which tells them to use the `<Image />` component, but it has no link to what it is or how to use it. It also mentions the markdown notation that is just below.
- The markdown section is great for the base use case 💅🏼 
- The user wants to customize it to set the density, then they need the Image component, which again doesn't have a link but tells them they need MDX
- The MDX section has an example, now they can use it. But the example only shows using `src` and `alt`, what do they need to set?
- Well there is a link to using the component on `.astro` files, if the user clicks on that then they see other examples using `src` and `alt` on an Astro file and finally a link to the `<Image />` component.
- But that link only goes to the next section. They need to scroll down and click on the "several components properties" link to find the references and there they have the densities.

So from the section about their use case to the section describing the option they need to read 5 sections and click through 3 links.

Couldn't we have at least one case in each section where `<Image />` links to the guide section or the reference?

sarah11918:
Thanks, taking a look at this journey now @Fryuni ! :raised_hands:

delucis (approved):
Fryuni’s suggestions are a good idea. Thumbs up from me for the rest of this though!

Fryuni:
Love it now 😍 

This one part might be too cluttered on mobile, but the info is there and it is at the end of the section so it doesn't bother me

Fryuni (approved):

sarah11918:
D'oh! the new content fits perfectly on *my* phone, but at least it's there! Let's do this!!

PR #9821, opened 2024-10-26
fix: wrong codeblock in `astro-content.mdx`

jsparkdev:
#### Description (required)

- update wrong codeblock in `astro-content.mdx`

- Closes #
- Suggested label:


delucis (approved):
Ah, very good point — thanks for catching these @jsparkdev 🚀

PR #9819, opened 2024-10-26
fix: remove unused link in `astro-actions.mdx`

jsparkdev:
#### Description (required)

- remove unused link in `astro-actions.mdx`

#### Related issues & labels (optional)

- Closes #
- Suggested label:

PR #9823, opened 2024-10-26
i18n(fr): create `reference/modules/astro-actions.mdx`

ArmandPhilippot:
#### Description (required)

Moves the `Actions` reference from `api-reference.mdx` to a new location `reference/modules/astro-actions.mdx` (see #9687 and #9819)

Note: I didn't remove the content in `api-reference.mdx` so at the moment the content is duplicated. I think it's better to do that in a separate PR to avoid conflicts.

- Suggested label: i18n


thomasbnt (approved):
LGTM!

PR #9827, opened 2024-10-26
i18n(fr): create `reference/modules/astro-i18n.mdx`

ArmandPhilippot:
#### Description (required)

Moves the Internationalization reference from `api-reference.mdx` to a new location `reference/modules/astro-i18n.mdx` (see #9687)

Note: I didn't remove the content in `api-reference.mdx` so at the moment the content is duplicated. I think it's better to do that in a separate PR to avoid conflicts.

- Suggested label: i18n


src/content/docs/fr/reference/modules/astro-i18n.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+```js title="middleware.js"
+import { defineMiddleware } from "astro:middleware";
+import { requestHasLocale } from "astro:i18n";
+
+export const onRequest = defineMiddleware(async (context, next) => {
+  if (requestHasLocale(context)) {
+    return next();
+  }
+  return new Response("Not found", { status: 404 });
```
thomasbnt, line 335:
```suggestion
  return new Response("Non trouvée", { status: 404 });
```

thomasbnt (changes requested):

thomasbnt (approved):

yanthomasdev (approved):
LGTM! Thanks!

PR #9880, opened 2024-10-30
catch Response in middleware

arty-name:
#### Description (required)

This PR describes a design pattern to reduce boilerplate code in pages. When some error condition in a function should always be handled by returning a specific `Response`, repeating this code in all the pages is cumbersome. [You can  instead catch a special subclass of `Error` in the middleware](https://github.com/withastro/roadmap/discussions/596).

### DIFF
--- a/src/content/docs/en/guides/server-side-rendering.mdx
+++ b/src/content/docs/en/guides/server-side-rendering.mdx
@@ -294,6 +294,44 @@ if (!product.isAvailable) {
 </html>
 ```
 
+#### Throw an `Error` with `Response` object
+
+Sometimes multiple pages in the project include identical code to check for a condition and return a specific `Response`. You can reduce such code repetition and bypass these checks in pages with the help of exceptions. Create a subclass of `Error` which contains a `Response`. Catch this exception in your middleware and return the `Response` instance from it.  Now you can throw this exception in your function instead of returning `null` or `undefined`.
+
+```astro title="src/ResponseError.js"
+export class ResponseError extends Error {
+  constructor(response) {
+    super(response.statusText)
+    this.response = response
+  }
+}
+```
+
+```astro title="src/middleware.js" {5}
+export async function onRequest(context, next) {
+  try {
+    return await next()
+  } catch (exception) {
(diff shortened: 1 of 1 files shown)


arty-name:
[The preview of the new content is here](https://deploy-preview-9880--astro-docs-2.netlify.app/en/guides/server-side-rendering/#throw-an-error-with-response-object)

sarah11918:
Hi @arty-name, thanks for wanting to improve our documentation!

Unfortunately, this isn't the kind of content we typically document. As I'm sure you can understand, we have to stick with documenting *how core Astro features work* or else we could end up documenting every practice/example under the sun. 😅 So, showing practices for reducing duplicate or boilerplate JavaScript code aren't really the mandate of Astro docs.

Tips like this can make great blog posts for self-publishing, or we'll often suggest working with https://astro-tips.dev/ to see if your content is a fit there! Unfortunately, it's not something for our official documentation so I'm closing this PR. But, we hope you'll find a place to share it with others.

arty-name:
Thank you for looking into this, @sarah11918! Fair enough, the documentation cannot include everything.

Would you also object to including a single sentence in the existing section? Something along the lines "You can also throw this Response and catch it in the middleware"…

sarah11918:
Hi, I have been discussing with the team and they feel like the middleware page itself covers this adequately. However we have been talking about what improvements on this page, maybe better linking to the middleware page we might want to make so that it's clearer what people can do with middleware and in which rendering situations (prerendering vs on demand). But, the goal of this page is simply t*o point out what becomes available with on demand rendering*, which includes some middleware features, not to fully document *using* middleware itself.

arty-name:
Thank you for your patience, @sarah11918!

I admit I don’t have a good overview of the documentation, but what you and the team say about middleware doesn’t seem really true to me as a consumer of the docs. I’d like to highlight to you this difference in perception.

The middleware page doesn’t mention catching of exceptions at all, and the option to create a new Response is hidden in an example of transformation, easy to gloss over. 

The SSR page was the only one I found which showcases the returning of manually created Responses. 

I don’t push for changes anymore and accept your decision.

PR #9882, opened 2024-10-30
clarify layout styling section of blog tutorial closes #9689

### LINKED ISSUE #9689: Styling part for the layout section of blog tutorial is a little bit confusing
### 📚 Subject area/topic

Layouts section in Blog tutorial

https://docs.astro.build/en/tutorial/4-layouts/1/

In the documentation, when giving the instructions for the `about.astro` page refactor, it says:

> (Note: using the BaseLayout to render your about.astro page means you will lose the `<style>` tag added to the `<head>` of this page. If you want to keep the custom `<h1>` style, move the style tag to the body of the page component.)

Even if we do this, the `h1` styles are not applied to the page because we are not using `h1` in `about.astro`, so the `BaseLayout`s styles are applied. I guess this is the default behaviour but in this case the documentation *might* be confusing for learners. So, maybe there should be a reminder about this is the default behaviour and either we use the element itself in the page or we do not see the styles are applied.

https://stackblitz.com/github/withastro/blog-tutorial-demo/tree/complete

Dillonpw:
#### Description (required)

Clarification to the blog tutorial, where layouts are introduced and how the implementation of layouts will effect existing styles as discussed in issue #9689 

- Closes # #9689 
- Suggested label:


sarah11918 (commented):
Thank you very much for picking this up and running with it @Dillonpw ! :raised_hands: 

I just formatted it in our Starlight "note" syntax. And, on the original issue, that note was meant to replace needing to say anything in the paragraph above the list, so I just went ahead and removed that while I was there! As soon as I can verify the preview looks fine with this formatting of a note inside a box (:sweat_smile: ), it'll be good to go!  :rocket:

PR #9888, opened 2024-10-30
Redistribute Built-in components and remove page

sarah11918:
#### Description (required)

This PR **removes the Built-in Components reference page** entirely in favour of redistributing the components to appropriate pages in docs.

(These components were not very similar to each other, did not all require imports, and in some cases were duplicate content existing in guides already.)

In the process:

- **Creates a new Syntax Highlighting guide** for both Markdown code block advice with the Code and Prism components which otherwise didn't have a super obvious place to go. This alleviates some of the content in the Markdown guide and describes all code syntax highlighting options together
- **Moves `markdown.shikiConfig` content to configuration reference** rather than having that entry link to the Markdown guide (in-branch overwrite right now, until we know what we want at the source)
- Makes the syntax highlighting content more "guide" like
- removes some "asides" in favour of proper documentation of "how things work"

Note: If we don't like the experiment of syntax highlighting as its own guide, this content can go back to the Markdown guide page. However, we would need to decide on a better place for the two Code and Prism components

UPDATE: The experimental changes above were well received on Talking and Doc'ing, so going with this overall structure for now!

- [x] Structure for new syntax higlighting page
- [x] file name and page title for new syntax highlighting page
- [x] deal with link fall out
- [x] remove component reference page from all `i18n/nav` pages that had it
- [x] LUNARIA TRACKING STATEMENT

Follow up:

- [ ] Make a PR for configuration reference (astro main branch) with the new Shiki config reference
- [ ] If someone wants to make it *actually* reference-y at some point/in the process, that's fine too!


src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -502,3 +378,24 @@ const {frontmatter} = Astro.props;
 You can also [style your Markdown](/en/guides/styling/#markdown-styling) in your layout component.
 
 <ReadMore>Learn more about [Markdown Layouts](/en/basics/layouts/#markdown-layouts).</ReadMore>
+
+## Fetching Remote Markdown
+
+**Astro does not include built-in support for remote Markdown outside of [experimental content collections](/en/reference/configuration-reference/#experimentalcontentlayer)!**
```
at-the-vr, line 384:
```suggestion
**Astro does not include built-in support for remote Markdown outside of [experimental content layer](/en/reference/configuration-reference/#experimentalcontentlayer)!**
```

at-the-vr (commented):
Got 1 typo as per usual, lotta redistribution 🚚 this month 😵

src/content/docs/en/guides/markdown-content.mdx
sarah11918: This is actually fine as is! The feature is still content collections, just the underlying API that powers them is experimental. The v5 docs will say "collections" just without the word experimental!

src/content/docs/en/guides/syntax-highlighting.mdx
```diff
@@ -0,0 +1,280 @@
...
+
+## Components for code blocks
+
+There are two Astro components available for `.astro` and `.mdx` files to render code blocks: [`<Code />`](#code-) and [`<Prism />`](#prism-)
+
+You can reference the `Props` of these components using the [`ComponentProps` type](/en/guides/typescript/#componentprops-type) utility.
+
+### `<Code />`
+
+This component This component is powered internally by Shiki. It supports all popular Shiki themes and languages as well as several other Shiki options such as custom themes, languages, [transformers](#transformers), and default colors.
```
ArmandPhilippot, line 132: typo:
```suggestion
This component is powered internally by Shiki. It supports all popular Shiki themes and languages as well as several other Shiki options such as custom themes, languages, [transformers](#transformers), and default colors.
```

ArmandPhilippot (commented):
I like the idea of a separate guide for syntax highlighting! Just spotted a typo and I left some thoughts regarding the content/structure of `guides/syntax-highlighting.mdx` >

I missed the beginning of Talking & Doc'ing so my thoughts may duplicate things that were said then.

The current structure bothers me a little:
* I feel like Prism is not very visible on the page. In particular, looking at the table of contents, it is not obvious that Prism is usable with Markdown (due to the subsections... Shiki, Shiki, Shiki :sweat_smile: ).
* We have 4 headings just to define a theme for Shiki while the heading to add a Prism theme does not appear in the table of contents.
* The section [Add a Prism stylesheet](https://deploy-preview-9888--astro-docs-2.netlify.app/en/guides/syntax-highlighting/#add-a-prism-stylesheet) is not only needed for the component `<Prism />` but also for using Prism with Markdown so its position doesn't seem ideal to me.

Could another page structure be more helpful? Maybe something like:

```
## Shiki
### With Markdown
// Shiki is the default, so you can directly use it... You can also configure it, see the reference...
#### Setting the theme used for code fences
##### Use a single theme
##### Use a dual theme
##### Use a custom theme
### With the `<Code />` component
#### Adding transformers
## Prism
### With Markdown
// You need to opt-in in your configuration file by doing...
### With the `<Prism />` component
### Setting a Prism theme
```

I don't know if this is the best structure and obviously the headings can be updated. This is just a draft based on the current sections.
My idea was not to add content, just to reorganize. So the Prism section is short but I was thinking of a structure like this so that:
* The user only interested in Prism or Shiki can easily jump to the right section without having to sort out what concerns him or not.
* We keep the distinction between usage with Markdown and usage of a component.
* Setting a Prism theme becomes visible in the table of contents and is no longer nested in either section (markdown/component).

sarah11918:
@ArmandPhilippot I do think that we *heavily* suggest using Shiki for a better experience (heck, everyone should really be using Expressive Code!) so there's a method to the current structure.

If you knew that we were subtly nudging and encouraging sticking with Shiki (which we have lots of configuration/features for, would that change your mind about the structure?

sarah11918:
Also, NOT grouping Shiki markdown and component together is VERY intentional, since people already assume that their `markdown.` config settings will apply to the `<Code />` component... which they don't! Because it's not used in Markdown!

ArmandPhilippot:
@sarah11918 It makes perfect sense! And I understood that Shiki is encouraged, that's why I clarified that I was not proposing to add content: it makes sense to develop the Shiki section more than Prism. No need for "artistic balance"! :smile:

My concern was mainly related to the first sentence: <q>Astro comes with built-in support for Shiki and Prism.</q>. Here is the scenario I imagined:
1. I get to the page, I read this.
2. Reaction: Cool, I'm more used to Prism, let's see how I can use it!
3. I look at the table of contents... and there it's not really clear. I see there is a component for Prism but apparently with Markdown you have to use Shiki instead... confusing.

So the current structure makes sense. Personally, I have no problem with it (I don't see why I would want to use Prism with Astro)! Rather, I was looking for how to solve a problem that might or might not happen. :sweat_smile:

Also I didn't notice that the link on `switch to Prism` was going to the stylesheet section... I don't know why. :eyes:

> One thought I did have was moving the Add a Prism theme out to ## level, at the bottom of the page, so it's kind of like an "addendum" vs owned entirely by the Prism component.

Yes, I think not nesting this section makes sense! Especially since we're returning the user from the Markdown section.
And maybe just doing that is enough to solve my fictional problem! After all this section recalls the two options (configuration and component) for using Prism.

Anyway, I'm waiting for the preview to be generated but I think your last change should be enough!

src/content/docs/en/guides/syntax-highlighting.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+### Customizing Shiki themes
+
+You can follow [Shiki's own theme documentation](https://shiki.style/themes) for more customization options for themes, light vs dark mode toggles, or styling via CSS variables.
+
+Astro code blocks are styled using the `.astro-code` class, so you will need to replace the `.shiki` class in the examples with `.astro-code`.
+
+## Components for code blocks
+
+There are two Astro components available for `.astro` and `.mdx` files to render code blocks: [`<Code />`](#code-) and [`<Prism />`](#prism-)
```
ArmandPhilippot, line 120: very small nit:
```suggestion
There are two Astro components available for `.astro` and `.mdx` files to render code blocks: [`<Code />`](#code-) and [`<Prism />`](#prism-).
```

ArmandPhilippot (approved):
LGTM! Now that Prism appears in the ToC I don't see any problem. 😉

src/content/docs/en/guides/syntax-highlighting.mdx
```diff
@@ -0,0 +1,227 @@
+---
+title: Code Syntax Highlighting
```
TheOtterlord, line 2:
```suggestion
title: Syntax Highlighting
```

src/i18n/en/nav.ts
```diff
@@ -68,6 +68,7 @@ export default [
 	{ text: 'CSS & Styling', slug: 'guides/styling', key: 'guides/styling' },
 	{ text: 'Images', slug: 'guides/images', key: 'guides/images' },
 	{ text: 'Fonts', slug: 'guides/fonts', key: 'guides/fonts' },
+	{ text: 'Code Syntax Highlighting', slug: 'guides/syntax-highlighting', key: 'guides/syntax-highlighting' },
```
TheOtterlord, line 71:
```suggestion
	{ text: 'Syntax Highlighting', slug: 'guides/syntax-highlighting', key: 'guides/syntax-highlighting' },
```

src/content/docs/en/guides/syntax-highlighting.mdx
```diff
@@ -0,0 +1,227 @@
...
+import { Steps } from '@astrojs/starlight/components';
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
+
+Astro comes with built-in support for [Shiki](https://shiki.style/) and [Prism](https://prismjs.com/). This provides syntax highlighting for:
+
+- all [code fences (\`\`\`)](#markdown-code-blocks) used in a Markdown or MDX file.
+- content within the [built-in `<Code />` component](#code-) (powered by Shiki) in `.astro` files.
+- content within the [`<Prism />` component](#prism-) (powered by Prism) in `.astro` files.
+
+Add [community integrations such as Expressive Code](https://astro.build/integrations/?search=expressive) for even more text marking and annotation options in your code blocks.
```
sarah11918, line 17:
```suggestion
Add [community integrations such as Expressive Code](https://astro.build/integrations/?search=syntax+highlight) for even more text marking and annotation options in your code blocks.
```

yanthomasdev:
@sarah11918, here is the official Lunaria statement:
```
@lunaria-track:src/content/docs/en/!(basics/astro-components.mdx);src/i18n/en/nav.ts
```

sarah11918:
Noting for Yan that even though I did have to update some i18n/nav files to remove the page being deleted, none of them will be updated because I did not add the new page to them, so I believe the original tracking statement is still correct!

PR #9899, opened 2024-10-31
fix: update Craft CMS title to be consistent with others guides

ArmandPhilippot:
#### Description (required)

Spotted during T&D: All other guides use ` & Astro` in their title, Craft CMS was the only exception. So I updated its title to be consistent with the others.


yanthomasdev (approved):
Nice catch, thanks!

PR #9900, opened 2024-10-31
revive recipes

sarah11918:
#### Description (required)

Quick fix for haunted recipe page

### DIFF
--- a/src/content/docs/en/recipes.mdx
+++ b/src/content/docs/en/recipes.mdx
@@ -1,7 +1,20 @@
 ---
-title: old recipes page
-description: old recipes page
+title: Official Astro recipes
+description: Short, focused how-to guides.
 i18nReady: false
 ---
 
-this is a fake placeholder page.
+import RecipesNav from '~/components/RecipesNav.astro';
+import ReadMore from '~/components/ReadMore.astro';
+
+See guided examples of adding features to your Astro project.
+
+## Official Recipes
+
+Astro's official recipes are short, focused how-to guides that walk a reader through completing a working example of a specific task. Recipes are a great way to add new features or behavior to your Astro project by following step-by-step instructions!
+
+<RecipesNav />
+
+## Community Resources
+
+<ReadMore>Find more recipes written and submitted by the community in our [community recipes list](/en/community-resources/content/#recipes-and-guides).</ReadMore>
--- a/src/content/docs/en/community-resources/content.mdx
+++ b/src/content/docs/en/community-resources/content.mdx
@@ -37,16 +37,6 @@ Check out the following courses and tutorials to learn more about Astro.
 
 ## Recipes and Guides
 
-import RecipesNav from '~/components/RecipesNav.astro';
-
-See guided examples of adding features to your Astro project.
-
-### Official Recipes
-
-Astro's official recipes are short, focused how-to guides that walk a reader through completing a working example of a specific task. Recipes are a great way to add new features or behavior to your Astro project by following step-by-step instructions!
-
-<RecipesNav />
-
 :::tip[Add your own!]
 Have you published a recipe or guide for working with Astro? [Edit this page](https://github.com/withastro/docs/edit/main/src/content/docs/en/community-resources/content.mdx) and add your link below in the appropriate section!
 :::


sarah11918:
All looks good to me, but need a place in the sidebar or to otherwise link to this! Now the haunted mystery page has content, but can't be accessed!

ArmandPhilippot:
Ah indeed, in the next version (your sidebar experiment) it would be simpler to put it in the sidebar I think but there...

Maybe:
1. Rename the `Community Resources` section to `Resources`
2. Rename the `Courses, Guides, and Recipes` to `Community Resources`
3. Add `Recipes` under the renamed `Resources`

But... This idea assumes that `Talks, Interviews, and Streams` always involves Astro members (which is therefore more or less "official")... which may not be the case (I haven't checked). If not, perhaps split the content out of `Talks, Interviews, and Streams` so that this page only lists contents where Astro members (at the time of recording at least) are involved and move the rest to the renamed `Community resources`.

Otherwise, maybe we can add `Recipes` to `Learn the basics` in the meantime... but it doesn't seem to be necessarily basic so it might not be better... Or perhaps add `Recipes` to `Maintain Your Project`?

For the moment I don't see any other alternatives. It doesn't seem to me that a link to `Recipes` would make more sense on one page than another...

sarah11918:
Thanks for the review @ArmandPhilippot ! As we discussed on the call, we may be doing away with all of this content except official recipes in the new design! (Sooner?) I think we are leaning towards we don't need to be people's search engines, although it's fair that if we can help some people, the content may not be harmful. (That said, we don't regularly review and update, so it may give more weight than we intend.)

To fix the immediate problem, which is that a page at `/lang/recipes` exists in a handful of other languages that would be difficult to quickly fix, adding back this content at `en/recipes.mdx` is by far the path of least resistance! That is the strategy here to avoid a massive i18n avalanche of debris.

But we then have the problem of "OK, but people need to be able to find the recipe content" so we need SOME place in the sidebar for it now! :sweat_smile:

sarah11918:
One possibility?

image

src/content/docs/en/recipes.mdx
```diff
@@ -1,7 +1,20 @@
 ---
-title: old recipes page
-description: old recipes page
+title: Official Astro recipes
```
ArmandPhilippot, line 2: Looks good! But I wonder if we should reword the title of the page:
* to avoid `<h1>Official Astro recipes</h1>` then `<h2>Official recipes</h2>` (this is the point that bothers me the most... but I can deal with it 😄 )
* to make the `Community resources` section fit better on the page (since they are not official)
* to reflect the name in the sidebar (even if it doesn't necessarily have to be word for word...)
```suggestion
title: How-to recipes
```

src/i18n/en/nav.ts
```diff
@@ -99,17 +99,19 @@ export default [
...
-	{
-		text: 'Talks, Interviews, and Streams',
-		slug: 'community-resources/talks',
-		key: 'community-resources/talks',
-	},
+	// {
+	// 	text: 'Talks, Interviews, and Streams',
+	// 	slug: 'community-resources/talks',
+	// 	key: 'community-resources/talks',
+	// },
```
sarah11918, line 114:
```suggestion
```

ArmandPhilippot:
One last point that requires confirmation since I'm not familiar enough with how Starlight works:

If I visit a recipe in the Netlify preview, both "How-to Recipes" and "Community Resources" links are highlighted. Is this due to the redirect being still "active" in the preview or is there something else to fix?
For example if I visit [Create a dev toolbar app](https://deploy-preview-9900--astro-docs-2.netlify.app/en/recipes/making-toolbar-apps/), in the sidebar I see:

astro-docs-recipes

I expect to see only "How-to Recipes" highlighted.

Other than that, LGTM! But I guess you're waiting for other opinions. 😅

sarah11918:
Converting to draft while I consider maybe handling this for v5 docs instead!

sarah11918:
Highlighting should be fixed with this next commit! Thank you so much for spotting Armand!

image
image

delucis (approved):
LGTM! Glad we are fixing the weird placeholder page even if all of this goes away with the v5 docs 🙌

src/content/docs/en/recipes.mdx
```diff
@@ -1,7 +1,20 @@
 ---
-title: old recipes page
-description: old recipes page
+title: How-to recipes
+description: Short, focused how-to guides.
 i18nReady: false
```
sarah11918, line 4:
```suggestion
i18nReady: true
```

PR #9906, opened 2024-11-03
Update aws.mdx

BayuPrasetyaUtomo:
#### Description (required)

- Add build settings for deployment on AWS Amplify using bun

- Closes #
- Suggested label:


TheOtterlord (commented):
Hi @BayuPrasetyaUtomo! Thanks for the PR!

Bun is not officially supported by Astro, and we don't include it in documentation examples. We do have a user-submitted recipe for [using Bun with Astro](https://docs.astro.build/en/recipes/bun/), but we've decided that for the time being, that is all and we will not be adding Bun as an additional supported package manager or runtime to the docs.

PR #9909, opened 2024-11-04
Update bun.mdx

Harold-Anderson:
Added link to my experiments with Bun vs pnpm (Astro 5) on Cloudflare Pages

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/recipes/bun.mdx
+++ b/src/content/docs/en/recipes/bun.mdx
@@ -135,3 +135,4 @@ To create your first Playwright test, follow the instructions for the rest of th
 Using Bun with Astro? Add your blog post or video to this page!
 
 - [Building a Cloudflare Pages site with Bun](https://blog.otterlord.dev/posts/hello-from-bun/) - blog post
+- [Comparing Bun to Pnpm on a Cloudflare Pages Site (Astro 5)](https://github.com/Harold-Anderson/acfo/blob/bun/README.md) - README.md file


TheOtterlord (commented):
Thanks for the PR @Harold-Anderson!

I took a look at your README, and a few things seemed a little strange.

First, the package manager / runtime you use shouldn't have an effect on CLS, since it doesn't affect the built output, or what the user is running. I did double check and I get 0 CLS on both preview sites listed.

It can have an effect on build speed (and response time if using that runtime for on-demand requests), but `bun run build` doesn't use the bun runtime by default in Vite projects, but instead node. To use the bun runtime, you need to use the [`--bun`](https://bun.sh/docs/cli/run#bun) flag. This probably explains why your build times were roughly the same (although yes, it's also not always the case that a bun build is faster).

I'll close this PR for now, as it's not entirely accurate, but if you end up trying bun again, please feel free to add your findings!

Harold-Anderson:
@TheOtterlord Thank you for the correction.  I have redeployed with ``bun run --bun build``

```
970 packages installed [2.75s]
--
Executing user command: bun run --bun build
$ astro build
```

The deployment is now clearly rendering differently.  https://15325ab1.acfo.pages.dev/

Would you like me to submit this as a PR?

TheOtterlord:
That's odd. Seems like any CSS that isn't the tailwind reset is not being loaded. It's probably a bug in Bun if it's only happening on the Bun branch. If you want, you could try getting help tracking it down in the [Astro Lounge](https://astro.build/chat) support forum (though there aren't that many Bun users).

Harold-Anderson:
My motivation was only to provide another datapoint, since your post is over a year old.  Since the runtime on Cloudflare is not bun, I was just interested in seeing if the build step was substantially faster, if the rendering was accurate, and if the compilation might have affected the speed of the rendering.  I suppose it could have been possible that bun (build) might have decided to preload my hero image, but that didn't happen.   As it happens, the build time is half of what pnpm's is, which is significant if you do a lot of builds.

I'm not going to track down the exact nature of the bug.  Just was interested in giving readers of the docs more info.  Let me know if you want me to submit the PR.

At some point, bun will be usable.  I've heard they have a lot of funding.  So I'm just staying up to date.

Harold-Anderson:
@TheOtterlord You were right it was tailwind.  Here is the bun deployment: https://878717ea.acfo.pages.dev/
I'll send you a new PR tomorrow. Thanks!

PR #9911, opened 2024-11-05
Add `svg` component reference

natemoo-re:
#### Description (required)

This PR adds a new section to the Images guide to introduce using `.svg` imports as components.

#### For Astro version: `5.x`. See astro PR [#12067](https://github.com/withastro/astro/pull/12067).


src/content/docs/en/guides/images.mdx
```diff
@@ -159,6 +159,37 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
 
 <ReadMore>See details about [the `<Picture />` component properties](/en/reference/modules/astro-assets/#picture-properties) in the `astro:assets` reference.</ReadMore>
 
+### Inline `.svg` files in the HTML
+
+<p><Since v="5.0.0" /></p>
+
+To inline local `.svg` files directly in your HTML, SVG files can be imported and used like any other [Astro components](/en/basics/astro-components/#component-based-design).
```
sarah11918, line 166:
```suggestion
To inline local `.svg` files directly in your HTML, import the SVG file to use as an Astro component. This process assigns an `<svg>` HTML tag to a component, and requires the same conventions (e.g. capitalization) as when [using dynamic tags](/en/basics/astro-syntax/#dynamic-tags).
```
I'm assuming this is how this works under the hood, and I think the link to the explanation and caveats there is more helpful, if true!

src/content/docs/en/guides/images.mdx
```diff
@@ -159,6 +159,37 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
...
+
+You can set attributes such as `width`, `height`, `fill`, `stroke`, and other common SVG attributes on the imported component. These attributes will automatically be applied to the underlying `<svg>` element. Properties passed to the component will override duplicate properties that exist in the original `.svg` file.
+
+
+```astro
+<Logo width={64} height={64} fill="currentColor" />
+```
+
+
+As a convenience, SVG components accept a `size` property. `size` is a shorthand which sets both the `width` and `height` properties to the same value.
```
sarah11918, line 184:
```suggestion
As a convenience, SVG components also accept a `size` property. `size` is a shorthand which sets both the `width` and `height` properties to the same value.
```
Such a tiny change, but feels more natural to me!

src/content/docs/en/guides/images.mdx
```diff
@@ -159,6 +159,37 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
...
+
+**Example:**
```
sarah11918, line 186:
```suggestion
The following example uses Astro's `size` property to set an equal width and height of 48 instead of setting the HTML `<svg>` attributes of `width` and `height` separately:
```

src/content/docs/en/guides/images.mdx
```diff
@@ -159,6 +159,37 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
...
+<p><Since v="5.0.0" /></p>
+
+To inline local `.svg` files directly in your HTML, SVG files can be imported and used like any other [Astro components](/en/basics/astro-components/#component-based-design).
+
+```astro
+---
+import Logo from '../assets/logo.svg';
+---
+
+<Logo />
```
sarah11918, line 173: Question: what happens if no attributes are passed? Is meta data from the `.svg` file used so that the image is basically presented as is? Are there any required attributes, or will this just work like this?

I'm not sure it's entirely necessary to actually state more here as long as someone can literally just do this and get SOME kind of image (even if not the way they want it to eventually look). Just want to make sure nothing is required, and everything else is an optional tweak of the logo file.

sarah11918 (commented):
People are going to love this @natemoo-re , thanks for chipping in on the docs for this feature! Some quick comments below!  :tada:

src/content/docs/en/guides/images.mdx
```diff
@@ -159,6 +159,37 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
 
 <ReadMore>See details about [the `<Picture />` component properties](/en/reference/modules/astro-assets/#picture-properties) in the `astro:assets` reference.</ReadMore>
 
+### Inline `.svg` files in the HTML
```
sarah11918, line 162:
```suggestion
### Inline `.svg` files in the HTML (experimental)
```
Looking at the source PR, it seems this will be behind an experimental flag?

CURRENTLY, **everything** about experimental features is only documented in config reference docs, but I know everyone hates that. So for v5 we can instead do something like this, but then we'll also need to link back to experimental config reference docs somehow to tell people how they can enable it.

OR **all** of this documentation exists only within the JSDOC of that PR.

src/content/docs/en/guides/images.mdx
stramel: If nothing is passed it will just use any attributes that are on the SVG file. I think this is the expected functionality.

stramel:
Should we also mention the default mode option in the config?

natemoo-re:
Made the updates that @sarah11918 suggested! Let me know if there's anything else I can do to help get this over the line!

sarah11918 (commented):
Thanks @natemoo-re and @stramel !

I also added the stuff for `mode` and now I will need your help because the section title title and intro paragraph are for "inlining svg" and ... now they're not *always* inlined!  😅 

I think with my suggestions here this is good (correct whatever is necessary!) EXCEPT that we probably need to now be more general about this section. So, maybe this is "SVG components"? And the intro is more like, "Astro provides support for local `.svg` files as components..." Does that framing sound right?

sarah11918:
Hey @natemoo-re and @stramel !

As I think this feature is likely to be released tomorrow, I made some executive decisions and put this together (moved now to the bottom of the page for easy extraction as this content will soon live on its own page instead of in this guide).

Would love at least a once-over that this content makes sense to you both, but no worries if this gets merged into beta docs with the feature before you've had a chance to review! They are just "beta" docs, and we can change 'em up real fast if and as we need to!

src/content/docs/en/guides/images.mdx
```diff
@@ -481,3 +480,95 @@ export default defineConfig({
 ## Community Integrations
 
 There are several third-party [community image integrations](https://astro.build/integrations?search=images) for optimizing and working with images in your Astro project.
+
+
+## Inline `.svg` files in the HTML (experimental)
```
natemoo-re, line 485:
```suggestion
## Using `.svg` files as components (experimental)
```

Is this a more accurate title?

src/content/docs/en/guides/images.mdx
```diff
@@ -481,3 +480,95 @@ export default defineConfig({
...
+
+ ```js
+{
+  experimental: {
+    svg: true,
+  },
+}
+```
+
+To use this feature, import an SVG file to use as an Astro component, passing any common SVG attributes. Astro also provides a `size` attribute to set equal `height` and `width` properties. This process assigns an `<svg>` HTML tag to a component, and requires the same conventions (e.g. capitalization) as when [using dynamic tags](/en/basics/astro-syntax/#dynamic-tags).
```
natemoo-re, line 499:
```suggestion
To use this feature, reference the default import of any local `.svg` file. Since this import is treated as an Astro component, it requires using the same conventions (e.g. capitalization) as when [using dynamic tags](/en/basics/astro-syntax/#dynamic-tags).
```

Simplified for clarity and avoids repetition.

src/content/docs/en/guides/images.mdx
```diff
@@ -481,3 +480,95 @@ export default defineConfig({
...
+
+```astro
+---
+import Logo from './path/to/svg/file.svg';
+---
+
+<Logo size={24} />
```
natemoo-re, line 506:
```suggestion
<Logo />
```

No props?

src/content/docs/en/guides/images.mdx
```diff
@@ -481,3 +480,95 @@ export default defineConfig({
...
+```
+
+### SVG component attributes
+
+You can set attributes such as `width`, `height`, `fill`, `stroke`, and other common SVG attributes on the imported component. These attributes will automatically be applied to the underlying `<svg>` element. Properties passed to the component will override duplicate properties that exist in the original `.svg` file.
```
natemoo-re, line 511:
```suggestion
You can pass props such as `width`, `height`, `fill`, `stroke`, and any other attribute accepted by the [native `<svg>` element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg). These attributes will automatically be applied to the underlying `<svg>` element. If a property is present in the original `.svg` file and is passed to the component, the value passed to the component will override the original value.
```

Clarity

src/content/docs/en/guides/images.mdx
sarah11918: Love it!

src/content/docs/en/guides/images.mdx
sarah11918: Ooh, much better. (Though, on its own page, we won't need a heading at all. But rather a page title and description).

I'm just about to commit the standalone Experimental Flag PRs, so you'll see what we need.

src/content/docs/en/guides/images.mdx
sarah11918: Crystal clear.

sarah11918:
PIVOT!

I'm going to commit @natemoo-re 's changes and update branch so that we have the new Experimental Flag infrastructure! 🎏 

Then, I'll move this into an experimental flag page, and you'll still have a chance to review before tomorrow's merge if necessary. Basically, instead of a heading, we'll need a page title and description, and possible sidebar label if we don't want the sidebar entry to match exactly the page title.

sarah11918:
Say hello to your new page!

image

natemoo-re:
Ooh, looks beautiful @sarah11918! Thanks for handling all the coordination, as always.

Consider this an "author is happy if Team Docs is happy" situation! 💜

PR #9896, opened 2024-10-31
Refactor docs to use Starlight’s built-in i18n tooling

delucis:
Closes DX-896

Updates the docs codebase to use Starlight’s built-in i18n tooling instead of our ancient, hand-crafted artisanal docs code.

- `ui.ts` and `docsearch.ts` translation dictionaries are now merged into a single YAML file per language in `src/content/i18n/`.
- Our components all now use Starlight’s `Astro.locals.t()` instead of the old `useTranslations()` and `<UIString>` we used to have — those old utilities are also removed in this PR.
- I removed our dusty `add-language` package script as it is now entirely out of date and given how rarely (if ever? 😅) we have used it, updating it is probably not worthwhile.
- Because I can’t contain myself, I also deleted a BUNCH more of our dependencies that were hanging around in the repo unnecessarily. Basically stuff that over time we’re not using anymore but forgot to uninstall.
- Companion PR to AD²: https://github.com/withastro/contribute.docs.astro.build/pull/31

### Todo

- [x] Need advice from @yanthomasdev about how to handle the move of translation resources in this PR for Lunaria to be happy
- [x] Need to make a companion PR to AD² to update guidance about translation
- [x] Need folks to test this to make sure things are working as expected across different languages

PR #9912, opened 2024-11-05
fix: consistent translation files in PR #9896

ArmandPhilippot:
> [!note]
> These changes target Chris' PR: #9896

#### Description (required)

This PR fixes an ["issue" I mentioned in Chris' PR](https://github.com/withastro/docs/pull/9896#issuecomment-2456897026). It:
* Removes the unused keys in the translation files (it seems to me that the english one is already up to date)
* Fixes a typo `voccabulary` >  `vocabulary`  (a remnant of the previous file)

@delucis I had started to create comments to suggest the lines to delete... Finally I told myself that a PR would be faster! And less “polluting” (3 comments per file... 39 comments in total 😅 ).

### DIFF
--- a/src/content/i18n/fr.yml
+++ b/src/content/i18n/fr.yml
@@ -1,14 +1,11 @@
 a11y.sectionLink: Titre de la section
 # Site settings
 site.title: Documentation Astro
-site.og.imageSrc: /default-og-image.png?v=1
-site.og.imageAlt: "Logo dAstro dans l'espace, avec une planète violette dans le style de saturne flottant à droite de limage."
 # Left Sidebar
 leftSidebar.sponsoredBy: Sponsorisé par
 # Right Sidebar
 rightSidebar.community: Communauté
 rightSidebar.joinDiscord: Rejoindre notre Discord
-rightSidebar.readBlog: Lire nos articles
 rightSidebar.contribute: Contribuer
 rightSidebar.translatePage: Traduire cette page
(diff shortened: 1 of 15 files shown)


src/content/i18n/fr.yml
```diff
@@ -1,14 +1,11 @@
 a11y.sectionLink: Titre de la section
 # Site settings
 site.title: Documentation Astro
-site.og.imageSrc: /default-og-image.png?v=1
-site.og.imageAlt: "Logo dAstro dans l'espace, avec une planète violette dans le style de saturne flottant à droite de limage."
```
delucis, line 5: Ah damn, looking at this line I wonder if I carelessly deleted `'` quotes in this file or if I just messed up this line. Noticing “dAstro” and “limage” although “l'espace” survived.

Maybe we got lucky though because I haven’t immediately spotted any others in keys we’re keeping.

delucis (approved):
Looks great and definitely appreciate this approach vs 39 GitHub comments to individually accept 😅 

Thanks for taking care of this @ArmandPhilippot!

src/content/i18n/fr.yml
ArmandPhilippot: Ah yes I had seen that at the time of my previous comment, I did not think to check in this PR if it happened elsewhere...
Apparently, there is only one other string concerned! (in the VF, I did not check the other languages 😅 )

PR #9942, opened 2024-11-07
i18n(de): update `astro-syntax.mdx`

trueberryless:
#### Description (required)

- update `astro-syntax.mdx`
- #9888 

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/de/basics/astro-syntax.mdx
+++ b/src/content/docs/de/basics/astro-syntax.mdx
@@ -77,7 +77,7 @@ Stattdessen verwende ein clientseitiges Skript, um den event handler hinzuzufüg
 
 Lokale Variablen können in JSX-ähnlichen Funktionen verwendet werden, um dynamisch generierte HTML-Elemente zu erzeugen:
 
-```astro title="src/components/DynamicHtml.astro" "{item}"
+```astro title="src/components/DynamicHtml.astro" "{tiere}"
 ---
 const tiere = ["Dog", "Cat", "Platypus"];
 ---
@@ -90,7 +90,7 @@ const tiere = ["Dog", "Cat", "Platypus"];
 
 Astro kann bedingt HTML anzeigen, indem es JSX-logische Operatoren und Ternär-Ausdrücke verwendet.
 
-```astro title="src/components/ConditionalHtml.astro" "visible"
+```astro title="src/components/ConditionalHtml.astro" "sichtbar"
 ---
 const sichtbar = true;
 ---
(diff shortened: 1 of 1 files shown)


jsparkdev:
I found a few things that were wrong.

brave_screenshot_github com (1)

`{item}` and `visiable` are words that don't exist, but want to highlight them. This causes the highlighting to not render. (lines 80, 93)

jsparkdev:
In the original English, `client:* hydration directives` is set as a link, but here, only `client:*` is set as a link.

This seems wrong to me, but I could be wrong, as different countries have different language usage.

If this is what you intended, please let me know!

brave_screenshot_deploy-preview-9942--astro-docs-2 netlify app

jsparkdev:
Other than the parts mentioned above, everything else seems to be perfect!

trueberryless:
Thank you so much for not only looking at the new content, but the whole page! Otherwise, I would never have found those mistakes!

You are right with the `client: *` link, the other word was not highlighted...

jsparkdev:
LGTM!

trueberryless:
Another cool tip, now that you're one of the maintainers:

You probably have seen many reviews from other people with over 1000 commits in 3 withastro repos, but I don't know if you know how to create a review. Follow this nice guide: 😉

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#starting-a-review

I'm just mentioning this because you posted the review as normal comments, so maybe I can help... 😊

trueberryless:
> Follow this nice guide: 😉
> 
> https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#starting-a-review


One important point, I always miss when starting a review myself: Click the "Start a review" button (Step 10 in the GitHug guide). It's usually green and somewhere at the bottom right of the "files changed" page. 😉

jsparkdev:
@trueberryless  Thanks! 

By the way, it looks like I can only add reviews to modified code in PR, so I added this review as a general comment!

I often forget to hit the green button to complete a review, haha.

trueberryless:
> By the way, it looks like I can only add reviews to modified code in PR, so I added this review as a general comment!

Oh yeah, I didn't think about this... Sorry 😔 
What HiDeoo does in this situation though, is just to add the comment at the last possible line and mention something like "can't add comment there cause out of review range"

jsparkdev:
Thanks for the tip, I'll try it next time :)

PR #9950, opened 2024-11-08
fix(v5): add missing import in content collection tutorial

ArmandPhilippot:
Following a request for help on Discord, I saw that an import was missing in `tutorials/content-collections.mdx` for v5. In addition, I took the opportunity to update two other code snippets.

#### Description (required)

* **Step 5:** the `glob` loader import was missing
* **Step 8:** `MarkdownPostLayout` was imported when it is not needed at this step, so I removed it
* **Step 9:** `MarkdownPostLayout` is only used in this step, so I highlighted the import in addition to its use which is already highlighted


sarah11918 (approved):
Looks great, thank you!! 🚀

PR #9958, opened 2024-11-08
feat(middleware): more explanation for `next()` rewrite

ematipico:
#### Description (required)

While working with Ben on its rewrite of Astro Actions, I came across a technical limitation which can't be fixed.

I thought that it would be helpful to explain this limitation in case users try to encounter it.

### DIFF
--- a/src/content/docs/en/guides/middleware.mdx
+++ b/src/content/docs/en/guides/middleware.mdx
@@ -248,6 +248,8 @@ The `next()` function accepts the same payload of [the `Astro.rewrite()` functio
 
 When you have multiple middleware functions chained via [sequence()](#chaining-middleware), submitting a path to `next()` will rewrite the `Request` in place and the middleware will not execute again. The next middleware function in the chain will receive the new `Request` with its updated `context`:
 
+Calling `next()` with this signature will create a new `Request` object using the old `ctx.request`, this means that trying to consume `Request.body`, either before or after this rewrite, will throw a runtime error. This can likely happen with [Astro Actions that use HTML forms](/en/guides/actions/#call-actions-from-an-html-form-action). In these cases it's advised to avoid a rewrite inside the middleware.    
+
 ```js title="src/middleware.js"
 // Current URL is https://example.com/blog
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks @ematipico ! I'll wait until @bholmesdev has looked this over! 🙌

src/content/docs/en/guides/middleware.mdx
```diff
@@ -248,6 +248,8 @@ The `next()` function accepts the same payload of [the `Astro.rewrite()` functio
 
 When you have multiple middleware functions chained via [sequence()](#chaining-middleware), submitting a path to `next()` will rewrite the `Request` in place and the middleware will not execute again. The next middleware function in the chain will receive the new `Request` with its updated `context`:
 
+Calling `next()` with this signature will create a new `Request` object using the old `ctx.request`, this means that trying to consume `Request.body`, either before or after this rewrite, will throw a runtime error. This can likely happen with [Astro Actions that use HTML forms](/en/guides/actions/#call-actions-from-an-html-form-action). In these cases it's advised to avoid a rewrite inside the middleware.
```
bholmesdev, line 251:
```suggestion
Calling `next()` with this signature will create a new `Request` object using the old `ctx.request`. This means that trying to consume `Request.body`, either before or after this rewrite, will throw a runtime error. This can likely happen with [Astro Actions that use HTML forms](/en/guides/actions/#call-actions-from-an-html-form-action). In these cases, we recommend handling rewrites from your Astro templates using `Astro.rewrite()` instead of using middleware.    
```

Could nits to clarify what users should do instead

sarah11918: Great note, Ben! Love your focusing on what you should do rather than what not to do and I agree with all your changes.  🙌 

Can I just clarify whether "This can likely happen" means:

- this often happens with
- the likely cause when this happens is

And whether "In these cases" refers to:
- the case of Astro Actions using HTML forms
- the case of a runtime error thrown (no matter what usage caused it)
- the case of an Astro Action using HTML forms that threw a runtime error (i.e. super specific!)

I think these are places were there's room to be a bit more explicit, and after that, great teamwork! 🫡

bholmesdev: Agreed this is ambiguous! "the likely cause when this happens" is what he's referring to. I'd change to this:

"This error is often raised when [Astro Actions with HTML forms](/en/guides/actions/#call-actions-from-an-html-form-action) are used.

sarah11918: Updating your suggestion above, and merging in so we can get this done! 🫡

PR #9992, opened 2024-11-12
Update dependencies

delucis:
#### Description (required)

This PR updates the docs site to use latest Astro & Starlight dependencies plus also bumps a few other related dependencies. It also _removes_ `astro-auto-import` entirely as it looks like we’re no longer using that directly.

I initially did this as part of #9896 but as that may take longer, I’d like to merge these updates separately to unblock other work that depends on the latest versions of things.


delucis:
Ahhhh… remembering why I did it as part of #9896 now 😅 

Updating these means fixing a bunch of i18n-related usage. Might as well wait for #9896 in that case :sadtaco:

PR #10002, opened 2024-11-13
Update Svelte integration v6 docs

bluwy:
#### Description (required)

Updates docs around new behaviour for `@astrojs/svelte` v6

- Closes #
- Suggested label:


src/content/docs/en/guides/integrations-guide/svelte.mdx
```diff
@@ -103,44 +103,28 @@ To use your first Svelte component in Astro, head to our [UI framework documenta
 
 This integration is powered by `@sveltejs/vite-plugin-svelte`. To customize the Svelte compiler, options can be provided to the integration. See the [`@sveltejs/vite-plugin-svelte` docs](https://github.com/sveltejs/vite-plugin-svelte/blob/HEAD/docs/config.md) for more details.
 
-### Default options
```
bluwy, line 106: Note there are now no default options set by Astro. It falls back to the behaviour in `@sveltejs/vite-plugin-svelte` by default now.

Which means the notes about `vitePreprocess` are no longer relevant, the user will have to always add that themselves if they use scss/stylus etc. Since `astro add svelte` has been adding those for users by default for a long time, I don't think this will break a lot of folks. I've noted this in the core PR changeset

src/content/docs/en/guides/integrations-guide/svelte.mdx
```diff
@@ -103,44 +103,28 @@ To use your first Svelte component in Astro, head to our [UI framework documenta
...
 };
 ```
 
-## Intellisense for TypeScript
+## Preprocessors
 
 <Since v="2.0.0" pkg="@astrojs/svelte" />
 
-If you're using a preprocessor like TypeScript or SCSS in your Svelte files, you can create a `svelte.config.js` file so that the Svelte IDE extension can correctly parse the Svelte files.
+If you're using SCSS or Stylus in your Svelte files, you can create a `svelte.config.js` file so that they are preprocessed by Svelte, and the Svelte IDE extension can correctly parse the Svelte files.
```
bluwy, line 127: I removed references to TypeScript here as Svelte 5 now handles TypeScript by default even without `vitePreprocess`.

src/content/docs/en/guides/integrations-guide/svelte.mdx
```diff
@@ -9,7 +9,7 @@ i18nReady: true
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 import Since from '~/components/Since.astro';
 
-This **[Astro integration][astro-integration]** enables server-side rendering and client-side hydration for your [Svelte](https://svelte.dev/) components. It supports Svelte 3, 4, and 5 (experimental).
+This **[Astro integration][astro-integration]** enables server-side rendering and client-side hydration for your [Svelte](https://svelte.dev/) 5 components. For Svelte 3 and 4 support, install `@astrojs/svelte@5` instead.
```
sarah11918, line 12: > For Svelte 3 and 4 support, install `@astrojs/svelte@5` instead

Just calling attention to it because it *looks* weird. Svelte 5 is the default. If you want 3 and  install `svelte@**5**` -- just checking this isn't a typo or something!

sarah11918 (commented):
Thanks, looks great @bluwy ! I only had one question below, but otherwise good to merge when the feature is released! :raised_hands:

sarah11918:
Oh! just calling out that we should have something in the v5 upgrade guide, too! Let's get that in this PR so it doesn't get forgotten!

src/content/docs/en/guides/integrations-guide/svelte.mdx
bluwy: It's referring to install `@astrojs/svelte`, which is an integration. And `@astrojs/svelte` v5 supports Svelte 3 and 4 only.

Svelte 5 support is added on `@astrojs/svelte` v6, which I didn't mention v6 here since it will be the latest.

sarah11918: Yup, figured this out in the other PR. What a pain, but understood!

Would be much nicer if our numbers aligned with theirs!

bluwy:
I suppose since this isn't related to Astro v5 we don't need to update the migration guide now? Currently the migration guide for `@astrojs/svelte` itself is in its changelog.

src/content/docs/en/guides/integrations-guide/svelte.mdx
casungo: > Would be much nicer if our numbers aligned with theirs!

Easy, just don't ship @astrojs/svelte v6 and update v5 until Svelte v6 drops. :trollface:

sarah11918:
Yup, that's when I thought this was related to Astro 5, too! This one's good to go!

sarah11918 (approved):
Ready to merge when the feature does!

casungo (approved):
LGTM!

PR #10008, opened 2024-11-13
Fix formatting CI

delucis:
#### Description (required)

https://github.com/withastro/docs/commit/6fb548796c6947feffc5a38c0595ea332e00e3d8 broke stuff because our CI task auto-formatting things removed a necessary type import. Turns out this was because `organize-imports-cli` is ancient and has never been updated for newer TS syntaxes like `satisfies`.

This PR removes import organization entirely and also updates Prettier and the Astro Prettier plugin.


HiDeoo (approved):
Looks good to me, tested locally `format:ci` no code was removed :tada:

PR #10020, opened 2024-11-14
fix: upgrade jsdoc-api

ascorbic:
#### Description (required)

Upgrades the jsdoc-api package, used in the docgen scripts. The current version has a bug that causes docgen to fail when a Mac tmp directory starts with an underscore. The latest version has a fix. It's a breaking change, so I've updated the usage too.

https://github.com/jsdoc2md/jsdoc-api/issues/19
https://github.com/jsdoc/jsdoc/issues/1712

- Closes #
- Suggested label:


delucis (approved):
Thanks for digging into this! I’m glad the pain was only `tmp`orary.

ascorbic:
That's a @bholmesdev -worthy review pun

bholmesdev:
It was a good at`tmp`t, but I’ve seen better from Chris 🙄

PR #10018, opened 2024-11-14
Update message format in `cli-reference.mdx`

Nin3lee:
#### Description (required)

Update message format in `cli-reference.mdx`

I notice the CLI message in the editor is going to be like this (Align to message's blank in the middle):

截屏2024-11-14 15 57 11

So I add the indent for code to fit the CLI's output message format.

#### Related issues & labels (optional)

- #9895 
- Suggested label: `consistency/formatting`

PR #9895, opened 2024-10-31
Update cli-reference.mdx

liruifengv:
#### Description (required)

Update cli-reference.mdx

- Add `create-key` command for cli help info
- Delete `--open` flag from cli help info. It's not in astro cli output help info.
- Move `--open` flag to Common flags from Global flags
- Add a section for how to get a list of flags for any command

#### Related issues & labels (optional)

- Closes #
- Suggested label:

PR #10021, opened 2024-11-14
i18n(fr): update `cli-reference.mdx`

ArmandPhilippot:
#### Description (required)

Adds changes from #9895 and #10018 to the French translation of `cli-reference.mdx`

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/reference/cli-reference.mdx
+++ b/src/content/docs/fr/reference/cli-reference.mdx
@@ -64,6 +64,7 @@ Commands
               add  Add an integration.
             build  Build your project and write it to disk.
             check  Check your project for errors.
+       create-key  Create a cryptography key
               dev  Start the development server.
              docs  Open documentation in your web browser.
              info  List info about your current Astro setup.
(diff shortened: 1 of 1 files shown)


src/content/docs/fr/reference/cli-reference.mdx
```diff
@@ -436,6 +474,10 @@ Active les journaux verbeux, ce qui est utile pour déboguer un problème.
 
 Active les journaux silencieux, ce qui lance le serveur sans aucune sortie dans la console.
 
+### `--open`
+
+Ouvre automatiquement l'application dans le navigateur au démarrage du serveur. Une chaîne d'URL complète (par exemple `--open http://example.com`) ou un chemin d'accès (par exemple `--open /about`) peut être transmis pour spécifier l'URL à ouvrir.
+
 ## Options globales
```
jsparkdev, line 481:
```suggestion
## Drapeaux mondiaux
```

It looks like the wrong heading was used.

I want to show Global Flags, but I don't know if the result of the French translation is correct :)

jsparkdev (commented):
Thank you!

src/content/docs/fr/reference/cli-reference.mdx
ArmandPhilippot: Thanks for the suggestion @jsparkdev!
For a literal translation, you would be right: flags should be translated with "drapeaux". But in this context (CLI), I don't think I've ever seen "drapeaux" used in French docs. "options" is more commonly used (and this is what the [French version of Wikipedia](https://fr.wikipedia.org/wiki/Interface_en_ligne_de_commande#Option) or the [PHP doc](https://www.php.net/manual/fr/features.commandline.options.php) uses for example). In other places one can also see "indicateurs" but I'd say it is less common than "option".

So I think the current version ("Options globales") is the correct translation for "Global Flags"!

jsparkdev: Oh I see, thank you for your response!

jsparkdev (approved):
LGTM!

Issue #10025, opened 2024-11-14
(v5) Tutorial update following template change


ArmandPhilippot:
### 📋 Explain your issue

First, sorry if I choose the wrong template... `Incorrect Content Report` didn't seem to fit so...

As discussed during Talking & Doc'ing yesterday (at least for me, the 14/11), the [tutorial for v5](https://5-0-0-beta.docs.astro.build/en/tutorial/0-introduction/) needs to be updated following the changes made by https://github.com/withastro/astro/pull/12083.

I haven't finished the tutorial yet, it took me longer than I thought. (This is the first time I do it, you have to excuse me. 🎉 Well, it's mainly taking notes that takes time! 😄 )

I stopped at <q>Add an RSS feed</q>. To avoid losing my notes, I preferred to share them now.

## First try with `pnpm create astro@beta`

I tried this approach first but, I agree with Chris. If we keep `pnpm create astro@latest`, there are too many things to update. Even by modifying two steps while initializing the project, from step 12 it is no longer possible to follow the tutorial without major changes.

## Second try with `pnpm create astro@beta -- --template minimal`

In summary, there are fewer changes but we will have problems with Typescript from Unit 5 onwards. I haven't looked for fixes yet, just noted the reported errors.

Also, I skipped the steps related to Github and Netlify because I don't see why they would be affected... and I didn't want to waste time with that. If needed, I'll redo the tutorial when I have more time.

1. On [Create your first Astro project](https://5-0-0-beta.docs.astro.build/en/tutorial/1-setup/2/)

* Step 1 needs to be updated with the template flag
* Step 4 and 5 needs to be removed

2. On [Create a blog post archive](https://5-0-0-beta.docs.astro.build/en/tutorial/5-astro-api/1/)

* Step 1, nit:

```diff
-const allPosts = Object.values(await import.meta.glob('./posts/*.md', { eager: true }));
+const allPosts = Object.values(import.meta.glob('./posts/*.md', { eager: true }));
```

We don't need `await`. In VS Code, this is underlined with three dots when we use it: `'await' has no effect on the type of this expression.ts(80007)`. So it's probably best not to include it.

* Step 2, we start having issues with Typescript!

```
{allPosts.map((post) => <li><a href={post.url}>{post.frontmatter.title}</a></li>)}
```

Both `url` and `frontmatter` are underlined and show the following error `Property 'url' does not exist on type 'unknown'.ts(2339)`.

[...]

PR #10033, opened 2024-11-15
remove View Transitions extend tutorial

sarah11918:
#### Description (required)

This removes one of the two "Extend the tutorial" tutorials as part of a larger restructuring of the tutorial/sidebar.

This removes the View Transitions tutorial but salvages some of the most-referred to content.

(Note: an example related to light/dark mode also exists in `astro:before-swap`, so the new example may not be needed/wanted. If we decide that the existing View Transitions page is helpful enough as it is, happy to not add any new content there!

Notably:

- removes the Extend with View Transitions tutorial page
- removes the Card at the end of the Build a blog tutorial suggesting this tutorial as a next step (there should not be any other existing internal links to this page)
- adds a redirect from this page to the View Transitions guide

Follow up to come:

- delete the "Add Content collections" tutorial and use that content for optional new module at the end of the existing tutorial


ArmandPhilippot (approved):
LGTM!

Regarding the tutorial, I don't know if you saw #10025 related to the T&D. I wanted to check if the Typescript errors I noted also exist in v4.... but I had a "slight" setback today.

PR #9344, opened 2024-09-08
new markdown content page

sarah11918:
Accompanies #9337 which updated the MDX page with info to be removed from the Markdown page.

Here's the new Markdown page with that content removed! Page has also been lightly edited and streamlined for better readability / flow.

I *had* been holding this for the beta 5.0 docs, since there are some specific changes that were relying on the new shape of docs (3 separate pages for content collections, content layer no longer experimental, etc.)

However, it turns out those are only a few minor things that I can easily update on the branch but have this page in docs now, in a mostly final form.

All the other files changed are just link updates.

PR #9335, opened 2024-09-07
Layouts page rewrite

sarah11918:
#### Description (required)

This lightly updates the `/basics/layouts.mdx` guide page to:

- Prepare for a rewrite of the existing Markdown page that will hold more of the Markdown-specific content (e.g. the removed code example of exported properties) so that the Layout page is less cluttered and mostly focused on the pattern
- Simplify by downplaying MDX and concentrating on Markdown (our built-in, first class citizen) (as the new Markdown page will, moving some content directly to the MDX page)

The only content temporarily *lost* in published docs as a result of this is the code example, but a version of that will exist in the new Markdown page, where it is more helpful and discoverable.

PR #10034, opened 2024-11-15
i18n(de): recreate `basics/layout.mdx`

trueberryless:
#### Description (required)

- delete the whole content because it was too out of date
- translate the english content from scratch

in theory this should update the translation of these PRs, but I think there are many more changes not tracked by Lunaria...
- #8763 
- #9333
- #9335 
- #9344 

(https://github.com/withastro/docs/commits/main/src/content/docs/en/basics/layouts.mdx?since=2024-07-10T18:34:38.000Z)


src/content/docs/de/basics/layouts.mdx
```diff
@@ -1,41 +1,51 @@
...
 
-Layouts werden normalerweise im Verzeichnis `src/layouts` deines Projekts abgelegt.
+Wenn eine Layoutkomponente jedoch eine Seitenhülle enthält, muss ihr `<html>`-Element das übergeordnete Element aller anderen Elemente der Komponente sein. Alle [`<style>`](/de/guides/styling/#styles-stile-in-astro) oder [`<script>`](/de/guides/client-side-scripts/#verwendung-von-script-in-astro) Elemente müssen von den `<html>` Tags umschlossen sein.
+
+Layout-Komponenten werden normalerweise in einem `src/layouts`-Verzeichnis in deinem Projekt platziert, aber das ist keine Voraussetzung; du kannst sie überall in deinem Projekt platzieren. Du kannst Layout-Komponenten sogar neben deinen Seiten platzieren, [indem du den Layout-Namen ein `_` voranstellst](/de/guides/routing/).
```
jsparkdev, line 19:
```suggestion
Layout-Komponenten werden normalerweise in einem `src/layouts`-Verzeichnis in deinem Projekt platziert, aber das ist keine Voraussetzung; du kannst sie überall in deinem Projekt platzieren. Du kannst Layout-Komponenten sogar neben deinen Seiten platzieren, [indem du den Layout-Namen ein `_` voranstellst](/de/guides/routing/#excluding-pages). 
```

The link seems to be different. If there is no heading that corresponds to the link, it seems like a good idea to keep the original link.

trueberryless: My thought process was:

- If there is no german file, link to the english heading.
- If there is a finished translation, link to the same locale heading.

But in this case, there is an outdated german version where this heading doesn't even exist yet... So linking to either the german or english heading will always land on top of the german page.

Is it now really better to link to the english heading, or should I already link to the german one (which I will translate in the future)?

jsparkdev: It's okay to keep it, it's just a suggestion.

The reason for the suggestion is that if you update the `routing.mdx` file after this PR is merged, you may forget that you need to update this file again.

If you add an English version of the heading, CI might throw an error when you update the `routing.mdx` file and realize that you need to update this file along with it.

trueberryless: Okay, this is a very inspiring way of keeping such things in mind! I wrote a TODO comment below this line, but I think I'll delete the TODO comment and update it. Thank you!

trueberryless:
I guess the CI check failing is okay, as mention [here](https://github.com/withastro/docs/pull/10034#discussion_r1850219476), @jsparkdev ?

sarah11918:
Just mentioning that this cannot merge with a failing check, because then all of docs will show a failing check because of this on the main branch itself, and it will affect everything!

**I would say to remove the link entirely**. The content doesn't exist on that page so it's not even a helpful link at all.

These docs are going to be retired soon, and almost all the new docs will be outdated anyway, so you'll probably HAVE to revisit this page. And, presumably at some point someone will update the routing page to actually include this heading, and at that point, the link can be readded.

Once v5 is published, this would be great to tackle: first update the routing page and then come back and add this link!

jsparkdev:
@trueberryless @sarah11918 Thanks for the nice comment, I guess my suggestion wasn't a good one 😂

jsparkdev (approved):
LGTM!

PR #10035, opened 2024-11-15
content collections tutorial

sarah11918:
#### Description (required)

This removes the individual "Extend the tutorial with Content Collections" tutorial and instead add a "bonus/optional" section to the existing blog tutorial

In particular:

- at this draft stage, provides TWO alternatives (we need to choose only one!) for adding this to the tutorial: as "6.4" or as an optional unit 7
- lightly edits text for context, now that it is part of the existing tutorial
- includes `create astro` updates introduced in https://github.com/withastro/astro/pull/12083 (minimal option must be chosen via template flag, is no longer a menu item; Typescript template is automatically set to `strict`)
- Includes a redirect
- updates the tutorial extension as a "next step" card at the end of the tutorial

Still to do:
- update any code samples as necessary now that the template uses the `strict` TypeScript setting
- choose a strategy for where this content is added (new unit? Bonus after Congratulations?)
- make sure the redirect is correct for our chosen location of the new content!

### DIFF
--- a/public/_redirects
+++ b/public/_redirects
@@ -31,7 +31,7 @@
 /:lang/guides/content/                               /:lang/guides/markdown-content/
 /:lang/community-resources/talks/                    /:lang/community-resources/content/
 /:lang:/reference/components-reference/              /:lang:/guides/syntax-highlighting/
-
+/:lang:/tutorials/add-content-collections/           /:lang:/tutorial/7-collections/
 # Very old docs site redirects
 # Once upon a time these URLs existed, so we try to keep them meaning something.
 /reference/renderer-reference                   /en/reference/integrations-reference/
--- a/src/content/docs/en/tutorial/0-introduction/1.mdx
+++ b/src/content/docs/en/tutorial/0-introduction/1.mdx
@@ -42,12 +42,6 @@ Hop into the support forum channel to ask questions, or say hi and chat in `#gen
 </details>
 
 <details>
-<summary>What can I do after I complete this tutorial?</summary>
-
-At the end of this tutorial, you will have a functioning blog using Astro's built-in file-based routing. You can enhance this project's final code by [refactoring it to manage your content with content collections](/en/tutorials/add-content-collections/) or [adding view transitions to customize page navigation](/en/tutorials/add-view-transitions/).
-</details>
-
-<details>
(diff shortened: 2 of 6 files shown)


ArmandPhilippot:
> at this draft stage, provides TWO alternatives (we need to choose only one!) for adding this to the tutorial: as "6.4" or as an optional unit 7

I'd say an unit 7 could be better:
* Because this part seems a bit long.
* If we want to encourage people to use Content Collections instead of the `layout:` property in the frontmatter of MD files. But, in both cases, it might seem weird after the "Congratulations" part.

Suggestion:

* `Make a Content Collection (optional)` could be a separate unit, just after `Check in: Unit 6 - Astro Islands`.
* We move `Congratulations` (6.3) in an unit 8.
* We add a CardLink at the beginning of unit 7 `Make a Content Collection (optional)`, with something like `Skip this optional unit`.

It's just an idea... maybe it can be improved. And it depends on the intention: to discourage the use of `layout:` or not.

sarah11918:
So one immediate problem with a Unit 7 (and holy cow, no unit 8!!) is how our tracker was designed to fit 6 units 😅 

Screenshot 2024-11-15 16 09 37

Also, as you yourself said, the tutorial is LONG. And all you can see when you start is the unit numbers, not for example, that the last one is optional... This is why I am *leaning* towards being in Chapter 6 (even if the unit title is a little weird for that) and specifically showing it AFTER congratulations in that way.

sarah11918:
> If we want to encourage people to use Content Collections instead of the layout: property in the frontmatter of MD files

Well, we want to give them the option of learning it, but think of how many Astro sites are 1 page portfolios that don't *need* a collection! (And a lot of new concepts specific to that feature.) That's in part why the tutorial emphasizes the file-based routing and had an additional add-on for those who *do* have sets of data to manage.

So, I would say it's not necessarily a given that the tutorial should push towards collections, but should make it available.

sarah11918:
Secret bonus track!! 😄

sarah11918:
OK, thanks to @ArmandPhilippot for evaluating the code samples and making the neccessary updates!

ArmandPhilippot (approved):
LGTM! 🚀

PR #10038, opened 2024-11-16
Ensure that the transition:animate="none" example is technically correct

### LINKED ISSUE #9977: The example for transition:animate:"none" does not work as advertised
### 📚 Subject area/topic

View Transitions

https://docs.astro.build/en/guides/view-transitions/#built-in-animation-directives

`transition:animate` will only work on the root element if it's used within a component that's shared across all pages.

The simplest solution is to use a common layout, here is the most minimal version:
```
---
// Layout.astro
---
<html transition:animate="none">
  <slot />
</html>
```

https://discord.com/channels/830184174198718474/1305515717101359104

here is a Stackblitz for the example from the support thread
https://stackblitz.com/edit/withastro-astro-emfhs5?file=src%2Fpages%2Findex.astro
not been able to turn off transitions in this example is inherent to Astro's approach and can not really be repaired.  

The fix in docs would be to use transition:animate="none" only in a layout, not directly on the `<html>` element of a layout-less `.astro` file. (this adds complexity to the example, but currently it is simply wrong)

martrapp:
#### Description (required)

Hy @sarah11918 👋🏼, after some thinking I managed to come up with a imho minimal change to the example that ensures it's technically correct without introducing further complexity. The most challenging part was resisting the urge to add excessive explanation and background details below the example. 

Would you please rephrase the text as you see fit for docs?

I hope this approach works, but I would also be fine with just adding `transition:name="root"` to the example without additional context. 🤓 

Closes #9977

### DIFF
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -182,7 +182,6 @@ import CommonHead from '../components/CommonHead.astro';
 </html>
 ```
 
-
 ### Customizing Animations
 
 You can customize all aspects of a transition with any CSS animation properties.
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -182,7 +182,6 @@ import CommonHead from '../components/CommonHead.astro';
 </html>
 ```
 
-Note that explicitly setting the [transition:name](#naming-a-transition) is necessary here, as the `<html>` element is directly used within in an individual page, not in a shared layout component.
 
 ### Customizing Animations
 
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -166,7 +166,7 @@ The example below produces a slide animation for the body content while disablin
 import CommonHead from '../components/CommonHead.astro';
 ---
 
-<html transition:animate="none">
+<html transition:name="root" transition:animate="none">
   <head>
     <CommonHead />
   </head>
@@ -181,6 +181,8 @@ import CommonHead from '../components/CommonHead.astro';
   </body>
 </html>
 ```
+
+Note that explicitly setting the [transition:name](#naming-a-transition) is necessary here, as the `<html>` element is directly used within in an individual page, not in a shared layout component.
 
 ### Customizing Animations
 


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -182,6 +182,8 @@ import CommonHead from '../components/CommonHead.astro';
 </html>
 ```
 
+Note that explicitly setting the [transition:name](#naming-a-transition) is necessary here, as the `<html>` element is directly used within in an individual page, not in a shared layout component.
```
sarah11918, line 185: So my thinking is, instead of explaining why naming is necessary here, the "naming a transition" section should specify when / why names are necessary!

It currently says:

image

The reason you mentioned above doesn't seem to be addressed in that. I would say, no explanation here. Just show the example with the `name="root"`. BUT maybe we need to add something to when names are necessary to explain this case? (If that is done, then I think this explanation stands on its own. Even if people see this and are like, "Why the name?" then they can check the docs for naming a transition and the explanation should make sense there.)

martrapp: A great, thank you very much!

That is a very good point! People can look up `transition:name` themselves when they are irritated here! 

Aside: The order of `name` and `animate` in the `<html>` element is arbitrary. I have no strong feeling which one is more natural for a reader looking for `transition:animate="none"`. I put `transition:animate="none"` at the end to make it stand out more clearly. But would be happy to re-order if that would be clearer.  

For improving the _Naming a transition_ section:
I never understood when automatically generated names would match. I recommend to always explicitly set names using `transition:name`. But that is probably not what we want to write in _Naming a transition_ ;-)

martrapp, line 185:
```suggestion
```

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -182,6 +182,7 @@ import CommonHead from '../components/CommonHead.astro';
 </html>
 ```
 
+
```
martrapp, line 185:
```suggestion
```

martrapp:
Thank you Sarah, mergin now!

>  If you feel that the "Naming a transition" section is inadequate.
Definitely not inadequate -- it already has the perfect explanation: Sometimes magic fails, and you’ve got to take matters into your own hands.

PR #10044, opened 2024-11-18
Run CI on all PRs

TheOtterlord:
#### Description (required)

Run CI & Lunaria on all pull requests.


delucis (approved):
Thanks @TheOtterlord! I believe `pull_request:` without options works, and if it doesn’t we’ll find out soon enough 😁

PR #10049, opened 2024-11-18
fix(v5): update tutorial

ArmandPhilippot:
> [!warning]
> This PR targets Sarah's PR #10035!

#### Description (required)

Following the discussion in #10025, this PR:
* Generalizes instructions for `create astro` in Unit 1
* Fixes Typescript error (`ts(2531)`) in Unit 3
* Fixes Typescript errors (`ts(18046)`) in Unit 5
* Fixes Typescript error (`ts(7006)`)  in Unit 5
* Removes `astro sync` command in Unit 7 and extended Unit 6
* Adds missing import for `render` in Unit 7 and extended Unit 6

Note for @sarah11918: I didn't directly suggest the changes on your PR since I can't add suggestions on unmodified parts. And it might be easier to review.

Except, my point 8, I think I covered everything. But:
* I still think there might be an "issue" with point 8 so **before merging**, please [read my detailed explanation](https://github.com/withastro/docs/issues/10025#issuecomment-2483627718)
* I'd like your opinion on the added types to satisfy Typescript (also covered in the link above)

And for anyone coming at this stage, you can [read the initial steps](https://github.com/withastro/docs/issues/10025#issuecomment-2480671139) in #10025.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/tutorial/1-setup/2.mdx
+++ b/src/content/docs/en/tutorial/1-setup/2.mdx
@@ -49,16 +49,16 @@ The preferred way to create a new Astro site is through our `create astro` setup
     </PackageManagerTabs>
 
 2. Confirm `y` to install `create-astro`
-3. When the prompt asks, "Where would you like to create your new project?" type in the name of a folder to create a new directory for your project, e.g.
+3. When the prompt asks you where to create the project, type in the name of a folder to create a new directory for your project, e.g.
 `./tutorial`
 
     :::note
     A new Astro project can only be created in a completely empty folder, so choose a name for your folder that does not already exist!
     :::
 
-4. When the prompt asks, "Would you like to install dependencies?" type `y`.
+4. When the prompt asks you whether or not to install dependencies, say yes.
 
-5. When the prompt asks, "Would you like to initialize a new git repository?" type `y`.
+5. When the prompt asks you whether or not to initialize a new git repository, say yes.
 </Steps>
 
 When the install wizard is complete, you no longer need this terminal. You can now open VS Code to continue.
--- a/src/content/docs/en/tutorial/3-components/4.mdx
+++ b/src/content/docs/en/tutorial/3-components/4.mdx
@@ -141,8 +141,8 @@ Adding a `<script>` tag provides client-side JavaScript to "listen" for a user e
     ```astro title="src/pages/index.astro" ins={2-6}
       <Footer />
       <script>
-        document.querySelector('.hamburger').addEventListener('click', () => {
-          document.querySelector('.nav-links').classList.toggle('expanded');
+        document.querySelector('.hamburger')?.addEventListener('click', () => {
+          document.querySelector('.nav-links')?.classList.toggle('expanded');
         });
       </script>
     </body>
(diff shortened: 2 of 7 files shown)


sarah11918:
This all looks amazing! 

Re: bringing in the new types...

What about something like the sentence AFTER the FIRST time this code is used:

Changing the existing (line 66 of unit 5-1):
> Your entire list of blog posts is now being generated dynamically, by mapping over the array returned by `import.meta.glob()`.

instead to:

```
Your entire list of blog posts is now being generated dynamically using Astro's built-in TypeScript support, by mapping over the array returned by `import.meta.glob()`.
```
We're just asking them to copy code in that step anyway. This gives enough of a nod to what's happening I think?

Similarly, for the tags needing to be strings:

In unit 5-3
- Should this type also occur in lines 145 and 155?
- line 148 could be updated like:

From:

> In your browser preview, verify that you can see your tags listed.

to instead:

```
In your browser preview, verify that you can see your tags listed. If any blog posts are missing tags, or they are improperly formatted, Astro's built-in TypeScript will show you errors so you can check and correct your code.
```

I think even just little bits like this each time the type is introduced should be fine? What do you think?

ArmandPhilippot:
> This gives enough of a nod to what's happening I think?
> [...]
> I think even just little bits like this each time the type is introduced should be fine? What do you think?

As you say, specifying Typescript leaves a hint for those who would like to know more. So it seems like a good compromise to me! I will add these changes.

> Should this type also occur in lines 145 and 155?

No everything is good here but thanks for pointing it out to me! I applied the change in the wrong place: line 171 was not a problem, the problem is line 289 (and 337 I think, I will retest before changing)!

Apparently, Typescript doesn't have a problem with `tags` when it is used directly. However, if it's a property of an object, it complains.

> OK, so weird?! 😅 Then, are we sure pubDate wasn't a string and unit 4 isn't wrong?? Or, maybe we don't even need this step, as I think date would have been the only possible type mismatch?

My first guess was that this was added in the previous steps to avoid a Typescript error, but no.

Maybe there was a change (I'm not familiar enough with the history of `Astro.glob` and `import.meta.glob`) or it was just to make sure the date was interpreted as a string? Since it's not wrapped in quotes? I don't know...

So yes, I think we don't need this step.
In another hand, as I said, I think this step (with a slight modification) helps to accentuate the benefits of content collections: we had a `string` (or at least an unknown type), now we are sure to obtain a `Date` so we can update our code to take advantage of this. But, yeah, we could have also done `new Date(frontmatter.pubDate).toLocaleDateString()` before. So I'm not sure what the best path to take here is.

sarah11918:
>  I think this step (with a slight modification) helps to accentuate the benefits of content collections: we had a string (or at least an unknown type), now we are sure to obtain a Date so we can update our code to take advantage of this. But, yeah, we could have also done `new Date(frontmatter.pubDate).toLocaleDateString()` before. So I'm not sure what the best path to take here is.

OK, let's update this to anything correct using `Date`, which will emphasize that we've set this to be a date in the schema. Even if it's basically the same as before, let's just use this as an opportunity to show something matching the schema. I think that works, and serves the purpose of reminding people to make sure you check your frontmatter stuff and that with a schema, we are now thinking in terms of types of data.

ArmandPhilippot:
I let you check the wording in my last commit, but I think we have now addressed all the issues noted in #10025!

sarah11918 (approved):
This is so fantastic, @ArmandPhilippot ! Really appreciate the thoughtfulness here! 🚀

PR #10058, opened 2024-11-19
create new sidebar section for feature flags

sarah11918:
#### Description (required)

This creates a new section in the sidebar for experimental and feature flags, which each get their own full docs page. This documentation will no longer live in the configuration reference file.

Notably:

- creates a feature flag index page with some additional content to make it a nice, full page
- creates an individual page for each feature flag
- copy/paste of text currently in the configuration reference page for each flag (no content changes -- just restructuring)
- ~~NO new sidebar headings, since new sidebar will merge before this PR~~

image

For review:

Naming conventions for file titles/sidebar entries:
- `contentIntellisense (experimental)`
- `Experimental contentIntellisense`
- `Intellisense for collections (experimental)`
- `Experimental Intellisense for collections`
- ???

Follow up TODOs:

- [x] - add to new sidebar
- [ ] - remove feature flag docs from the source file for config reference
- [ ] - handle two new incoming experimental flags (responsive images, sessions)


src/content/docs/en/reference/feature-flags/experimental-client-prerender.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: Experimiental client prerendering
```
delucis, line 2: Didn’t see “Experimiental” in your options, but I like it 😂

delucis (commented):
Seems fine! Re your question about naming: it might depend on how you’re imagining the sidebar for these. Leading with “experimental” or “legacy” would be repetitive and make scanning harder.

Personally, I’d imagined just two pages: one for legacy, one for experimental, so there wouldn’t be much of a sidebar question. I kind of feel like “feature flags” is not super user-facing but more how _we_ think of/implement things. So leading more directly with “Experimental” and “Legacy” could be more helpful.

sarah11918:
Ah, I was thinking of a page per flag, so either combining them under one unified "flags" heading, OR yes, I could have one heading for Experimental Features and one for Legacy Flags. But I think for maintenance going forward, I do want one page per feature.

src/content/docs/en/reference/feature-flags/legacy-collections.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: Legacy - Collections
```
delucis, line 2: The hyphen is weird here to me.

Should it be one of these?

> Legacy: Collections

> Legacy collections

src/content/docs/en/reference/feature-flags/legacy-collections.mdx
```diff
@@ -0,0 +1,46 @@
...
+sidebar:
+  label: Collections (Astro v2-v4)
```
delucis, line 4: I think this label might be a little confusing: it makes me thing this flag is available in v2–v4 NOT that this flag is available in v5 and reintroduces the v2–v4 behaviour.

src/content/docs/en/reference/feature-flags/experimental-client-prerender.mdx
```diff
@@ -0,0 +1,42 @@
+---
+title: Experimental client prerendering
+sidebar:
+  label: Client prerendering
+i18n: true
```
delucis, line 5: Should these be `i18nReady` in all of these files?

delucis (commented):
Spotted a few nit-picky details (along with the casing thing I already mentioned).

Looking through this again, I _still_ feel like these should be two pages (“Legacy” and “Experimental”) and not individual pages for each flag:

- Sidebar labelling can be tricky in this context as I think is shown by the `legacy.collections` flag in particular. It was hard to name the flag, but even harder to label in a sidebar. In a single “Legacy flags” page, using the actual flag name as a heading is much more justifiable.

- It’s going to look odd maintaining this when there aren’t one or other of these types. You either end up with an empty group (awkward) or remove the group and then have to add it all back again when adding a new flag (maintenance overhead).

- It’s just a lot of sidebar real estate _ahead_ of other useful pages for flags that are often quite minor. For example, in the current docs there are EIGHT experimental flags, including really minor stuff like `directRenderScript` (good luck labelling that in the sidebar 😅).

BUT that doesn’t have to block this PR if you’d prefer it didn’t. The content here is good in general and we can re-org before v5 is stable if necessary.

src/content/docs/en/reference/feature-flags/legacy-collections.mdx
sarah11918: Ah, OK, I was trying to figure out how to say "The collections from Astro versions 2 to 4" not with respect to when the flag was available.

I have gone through a few iterations in my mind:

Collections v1-- (the first "iteration" of collections)
Collections v2 / Collections (Astro v2) -- (the collections introduced in Astro v2)
Collections v2-v4 (the collections you used in versions of Astro 2 through 4)

src/content/docs/en/reference/feature-flags/experimental-client-prerender.mdx
sarah11918: Yes! Thank you!

src/content/docs/en/reference/feature-flags/legacy-collections.mdx
sarah11918: Yes, canonically I think we have preferred colons if using punctuation in the past, so def. `:` if punctuation ends up being used.

sarah11918:
re: how many pages: I am now thinking we split the difference!

How would you feel about a single, top level "Legacy flags" but still keeping Experimental Features as a top-level heading, with individual pages for each feature?

I do feel like some feedback we got from the devs is that maybe experimental features are unused because people don't know about them. And so for features we genuinely do want people to try and give feedback on, I like that they each have their own entry in the sidebar. They are also the easier things to name.

However, we don't really want to encourage legacy flag use; it's more a convenience/stop gap measure. So, I think it does make sense to just put THAT on one page, and even if there are none, there's still the intro about what legacy flags are/do.

sarah11918:
> but our other six experimental features in v4 are not really more than a paragraph or two each. 

Right, and in fact for v5 with things moving to stable, the features we will have will be two small ones but otherwise probably interesting/used ones:

- `contentIntellisense` (hold over)
- clientPrerender (hold over)
- `sessions` (new, exciting!)
- `responsiveImages` (new, exciting!)
- `svg` (new, exciting!)

So I think at least to start v5 it does make sense to give all these PRs a chance with their own page, then reevaluate from there! I can see the new ones getting chonky again, and it will be much easier to let each feature author mostly deal with their own page, we don't have to be one level indented in headings hierarchy...

But thank you! I appreciate the call out and discussion. Let's monitor this one!

PR #10061, opened 2024-11-19
Move content config file

ascorbic:
#### Description (required)

Once https://github.com/withastro/astro/pull/12475 is released, the content config file can be `src/content.config.ts`. This PR updates all filenames used in the docs to match. It's backwards-compatible, so I haven't added it to the migration guide

https://github.com/withastro/astro/pull/12475 
- Closes #
- Suggested label:


sarah11918:
Thanks, all looks good here! This should not merge until this is released in the next beta, so we'll hold onto it.

Also, I will take a larger scan at the Upgrade Guide instructions, because even though this is backwards compatible, this is the time people are updating their projects anyway. We **do** want to make sure we're showing instructions to update to whatever we're showing in the rest of docs anyway. It could get confusing if people rely on backwards compatibility for too long and then go check the docs and the examples look different from their projects. I'd rather treat that as "happy accident it didn't break when they didn't do it" rather than a strategy for them.

And just checking, what about people using the `legacy.collections` flag? We should probably add a mention there that if you have that flag, you *also* keep your content config in the old place (assuming that would work, because backwards compatibility of this feature)? 

Is it even *possible* for them to move their config file out of `src/content/`? Is there anything we have to worry about there? (And even if so, it might be nicer *here* to not make that suggestion, to further simplify the legacy vs not story?)

sarah11918 (commented):
PTAL bot wants official reviews to show I commented here!

PR #10062, opened 2024-11-19
Runtime API restructure

sarah11918:
#### Description (required)

WIP - experiment combining the `Astro.global` and `context` object entries since several are duplicates or only links up to the global entry anyway.

This also makes each entry more visible to see where we haven't sufficiently documented.

Notably:

Is this still the right page for:

- `getStaticPaths()` ?
- `Astro.slots` and `Astro.self` ?

Still figuring this page out!

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -9,124 +9,17 @@ import Since from '~/components/Since.astro';
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 import ReadMore from '~/components/ReadMore.astro';
 
-## `Astro` global
+When rendering a page, Astro provides a runtime API specific to the current render. This includes useful information such as the current page URL as well as APIs to perform actions like redirecting to another page.
 
-The `Astro` global is available in all contexts in `.astro` files. It has the following functions:
+In `.astro` components, this context is available from the `Astro` global object. Endpoint functions are also called with this same context object as their first argument, whose properties mirror many of the Astro global properties.
 
-### `Astro.glob()`
+Some properties are only available for pages rendered on demand, or may have limited functionality on prerendered pages.
 
-:::caution[Deprecated in v5.0]
-Use [Vite's `import.meta.glob`](https://vite.dev/guide/features.html#glob-import) to query project files. 
+The `Astro` global object is available to all `.astro` files. Use the `context` object in [endpoint functions](/en/guides/endpoints/) to serve static or live server endpoints and in [middleware](/en/guides/middleware/) to inject behavior when a page or end point is about to be rendered.
 
-`Astro.glob('../pages/post/*.md')` can be replaced with:
+## `Astro.props`
 
-`Object.values(await import.meta.glob('../pages/post/*.md', { eager: true }));`
-
-See the [imports guide](/en/guides/imports/#importmetaglob) for more information and usage.
-:::
-
-`Astro.glob()` is a way to load many local files into your static site setup.
-
-```astro
----
-// src/components/my-component.astro
-const posts = await Astro.glob('../pages/post/*.md'); // returns an array of posts that live at ./src/pages/post/*.md
----
-
-<div>
-{posts.slice(0, 3).map((post) => (
-  <article>
-    <h2>{post.frontmatter.title}</h2>
-    <p>{post.frontmatter.description}</p>
-    <a href={post.url}>Read more</a>
-  </article>
-))}
-</div>
-```
-
-`.glob()` only takes one parameter: a relative URL glob of which local files you'd like to import. It’s asynchronous, and returns an array of the exports from matching files.
-
-`.glob()` can't take variables or strings that interpolate them, as they aren't statically analyzable. (See [the imports guide](/en/guides/imports/#supported-values) for a workaround.) This is because `Astro.glob()` is a wrapper of Vite's [`import.meta.glob()`](https://vite.dev/guide/features.html#glob-import).
-
-:::note
-You can also use `import.meta.glob()` itself in your Astro project. You may want to do this when:
-- You need this feature in a file that isn't `.astro`, like an API route. `Astro.glob()` is only available in `.astro` files, while `import.meta.glob()` is available anywhere in the project.
-- You don't want to load each file immediately. `import.meta.glob()` can return functions that import the file content, rather than returning the content itself. Note that this import includes all styles and scripts for any imported files. These will be bundled and added to the page whether or not a file is actually used, as this is decided by static analysis, not at runtime.
-- You want access to each file's path. `import.meta.glob()` returns a map of a file's path to its content, while `Astro.glob()` returns a list of content.
-- You want to pass multiple patterns; for example, you want to add a "negative pattern" that filters out certain files. `import.meta.glob()` can optionally take an array of glob strings, rather than a single string.
-
-Read more in the [Vite documentation](https://vite.dev/guide/features.html#glob-import).
-:::
-#### Markdown Files
-
-Markdown files loaded with `Astro.glob()` return the following `MarkdownInstance` interface:
-
-```ts
-export interface MarkdownInstance<T extends Record<string, any>> {
-  /* Any data specified in this file's YAML frontmatter */
-	frontmatter: T;
-  /* The absolute file path of this file */
-	file: string;
-  /* The rendered path of this file */
-	url: string | undefined;
-  /* Astro Component that renders the contents of this file */
-	Content: AstroComponentFactory;
-  /** (Markdown only) Raw Markdown file content, excluding layout HTML and YAML frontmatter */
-	rawContent(): string;
-  /** (Markdown only) Markdown file compiled to HTML, excluding layout HTML */
-	compiledContent(): string;
-  /* Function that returns an array of the h1...h6 elements in this file */
-	getHeadings(): Promise<{ depth: number; slug: string; text: string }[]>;
-	default: AstroComponentFactory;
-}
-```
-
-You can optionally provide a type for the `frontmatter` variable using a TypeScript generic.
-
-```astro
----
-interface Frontmatter {
-  title: string;
-  description?: string;
-}
-const posts = await Astro.glob<Frontmatter>('../pages/post/*.md');
----
-
-<ul>
-  {posts.map(post => <li>{post.frontmatter.title}</li>)}
-</ul>
-```
-
-#### Astro Files
-
-Astro files have the following interface:
-
-```ts
-export interface AstroInstance {
-  /* The file path of this file */
-  file: string;
-  /* The URL for this file (if it is in the pages directory) */
-	url: string | undefined;
-	default: AstroComponentFactory;
-}
-```
-
-#### Other Files
-
-Other files may have various different interfaces, but `Astro.glob()` accepts a TypeScript generic if you know exactly what an unrecognized file type contains.
-
-```ts
----
-interface CustomDataFile {
-  default: Record<string, any>;
-}
-const data = await Astro.glob<CustomDataFile>('../data/**/*.js');
----
-```
-
-### `Astro.props`
-
-`Astro.props` is an object containing any values that have been passed as [component attributes](/en/basics/astro-components/#component-props). Layout components for `.md` and `.mdx` files receive frontmatter values as props.
+`props` is an object containing any values that have been passed as [component attributes](/en/basics/astro-components/#component-props). Layout components for `.md` and `.mdx` files receive frontmatter values as props.
 
 ```astro {3}
 ---
@@ -149,15 +42,36 @@ import Heading from '../components/Heading.astro';
 
 <ReadMore>Learn more about how [Markdown and MDX Layouts](/en/guides/markdown-content/#frontmatter-layout-property) handle props.</ReadMore>
 
-<ReadMore>Learn how to add [TypeScript type definitions for your props](/en/guides/typescript/#component-props).</ReadMore>
 
-### `Astro.params`
+`context.props` is an object containing any `props` passed from `getStaticPaths()`. Because `getStaticPaths()` is not used when building for SSR (server-side rendering), `context.props` is only available in static builds.
 
-`Astro.params` is an object containing the values of dynamic route segments matched for this request.
+```ts title="src/pages/posts/[id].json.ts"
+import type { APIContext } from 'astro';
+
+export function getStaticPaths() {
+  return [
+    { params: { id: '1' }, props: { author: 'Blu' } },
+    { params: { id: '2' }, props: { author: 'Erika' } },
+    { params: { id: '3' }, props: { author: 'Matthew' } }
+  ];
+}
+
+export function GET({ props }: APIContext) {
+	return new Response(
+    JSON.stringify({ author: props.author }),
+  );
+}
+```
+
+See also: [Data Passing with `props`](#data-passing-with-props)
+
+## `Astro.params`
+
+`params` is an object containing the values of dynamic route segments matched for this request.
 
 In static builds, this will be the `params` returned by `getStaticPaths()` used for prerendering [dynamic routes](/en/guides/routing/#dynamic-routes).
 
-```astro title="src/pages/posts/[id].astro"
+```astro title="src/pages/posts/[id].astro" "Astro.params"
 ---
 export function getStaticPaths() {
   return [
@@ -172,9 +86,28 @@ const { id } = Astro.params;
 <h1>{id}</h1>
 ```
 
+
+```ts title="src/pages/posts/[id].json.ts" "params.id"
+import type { APIContext } from 'astro';
+
+export function getStaticPaths() {
+  return [
+    { params: { id: '1' } },
+    { params: { id: '2' } },
+    { params: { id: '3' } }
+  ];
+}
+
+export function GET({ params }: APIContext) {
+	return new Response(
+    JSON.stringify({ id: params.id }),
+  );
+}
+```
+
 In SSR builds, this can be any value matching the path segments in the dynamic route pattern.
 
-```astro title="src/pages/posts/[id].astro"
+```astro title="src/pages/posts/[id].astro" "Astro.params"
 ---
 import { getPost } from '../api';
 
@@ -192,34 +125,145 @@ if (!post) {
 
 See also: [`params`](#params)
 
-### `Astro.request`
+## `Astro.url`
+
+<p>
+
+**Type:** `URL`<br />
+<Since v="1.0.0" />
+</p>
+
+`url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` URL string value. Useful for interacting with individual properties of the request URL, like pathname and origin. 
+
+Equivalent to doing `new URL(Astro.request.url)`.
+
+`url` will be `localhost` in dev mode if [site](/en/reference/configuration-reference/#site) is not configured for static sites, and for on-demand rendered sites using `server` or `hybrid` output.
+
+```astro
+<h1>The current URL is: {Astro.url}</h1>
+<h1>The current URL pathname is: {Astro.url.pathname}</h1>
+<h1>The current URL origin is: {Astro.url.origin}</h1>
+```
+
+You can also use `url` to create new URLs by passing it as an argument to [`new URL()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL).
+
+```astro title="src/pages/index.astro"
+---
+// Example: Construct a canonical URL using your production domain
+const canonicalURL = new URL(Astro.url.pathname, Astro.site);
+// Example: Construct a URL for SEO meta tags using your current domain
+const socialImageURL = new URL('/images/preview.png', Astro.url);
+---
+<link rel="canonical" href={canonicalURL} />
+<meta property="og:image" content={socialImageURL} />
+```
+
+## `Astro.site`
+
+<p>
+
+**Type:** `URL | undefined`
+</p>
+
+`Astro.site` returns a `URL` made from `site` in your Astro config. If `site` in your Astro config isn't defined, `Astro.site` won't be defined.
+
+`context.site` returns a `URL` made from `site` in your Astro config. If undefined, this will return a URL generated from `localhost`.
+
+## `Astro.clientAddress`
+
+<p>
+
+**Type:** `string`<br />
+<Since v="1.0.0" />
+</p>
+
+`clientAddress` specifies the [IP address](https://en.wikipedia.org/wiki/IP_address) of the request. This property is only available for routes rendered on demand and cannot be used on prerendered pages.
+
+```astro
+---
+export const prerender = false;
+const ip = Astro.clientAddress;
+---
+
+<div>Your IP address is: <span class="address">{ ip }</span></div>
+```
+
+```ts
+export const prerender = false;
+import type { APIContext } from 'astro';
+
+export function GET({ clientAddress }: APIContext) {
+  return new Response(`Your IP address is: ${clientAddress}`);
+}
+```
+
+## `Astro.generator`
+
+<p>
+
+**Type:** `string`<br />
+<Since v="1.0.0" />
+</p>
+
+`generator` provides the current version of Astro your project is running. This is a convenient way to add a [`<meta name="generator">`](https://html.spec.whatwg.org/multipage/semantics.html#meta-generator) tag with your current version of Astro. It follows the format `"Astro v5.x.x"`.
+
+```astro mark="Astro.generator"
+<html>
+  <head>
+    <meta name="generator" content={Astro.generator} />
+  </head>
+  <body>
+    <footer>
+      <p>Built with <a href="https://astro.build">{Astro.generator}</a></p>
+    </footer>
+  </body>
+</html>
+```
+
+```ts title="src/pages/site-info.json.ts"
+import type { APIContext } from 'astro';
+
+export function GET({ generator, site }: APIContext) {
+  const body = JSON.stringify({ generator, site });
+  return new Response(body);
+}
+```
+
+## `Astro.request`
 
 <p>
 
 **Type:** `Request`
 </p>
 
-`Astro.request` is a standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object. It can be used to get the `url`, `headers`, `method`, and even body of the request. 
+`request` is a standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object. It can be used to get the `url`, `headers`, `method`, and even body of the request. 
 
-```astro
+```astro wrap title="src/pages/index.astro" "Astro.request"
 <p>Received a {Astro.request.method} request to "{Astro.request.url}".</p>
-<p>Received request headers: <code>{JSON.stringify(Object.fromEntries(Astro.request.headers))}</code>
+<p>Received request headers:</p>
+<p><code>{JSON.stringify(Object.fromEntries(Astro.request.headers))}</code></p>
 ```
 
-See also: [`Astro.url`](#astrourl)
+```ts "request"
+import type { APIContext } from 'astro';
+
+export function GET({ request }: APIContext) {
+  return new Response(`Hello ${request.url}`);
+}
+```
 
 :::note
-With the default `output: 'static'` option, `Astro.request.url` does not contain search parameters, like `?foo=bar`, as it's not possible to determine them ahead of time during static builds. However in `output: 'server'` mode, `Astro.request.url` does contain search parameters as it can be determined from a server request.
+On prerendered pages, `request.url` does not contain search parameters, like `?foo=bar`, as it's not possible to determine them ahead of time during static builds. However, `request.url` does contain search parameters for pages rendered on-demand as they can be determined from a server request.
 :::
 
-### `Astro.response`
+## `Astro.response`
 
 <p>
 
 **Type:** `ResponseInit & { readonly headers: Headers }`
 </p>
 
-`Astro.response` is a standard `ResponseInit` object. It has the following structure. 
+`response` is a standard `ResponseInit` object. It has the following structure. 
 
  - `status`: The numeric status code of the response, e.g., `200`.
  - `statusText`: The status message associated with the status code, e.g., `'OK'`.
@@ -245,7 +289,7 @@ Astro.response.headers.set('Set-Cookie', 'a=b; Path=/;');
 ---
 ```
 
-### `Astro.cookies`
+## `Astro.cookies`
 
 <p>
 
@@ -253,9 +297,9 @@ Astro.response.headers.set('Set-Cookie', 'a=b; Path=/;');
 <Since v="1.4.0" />
 </p>
 
-`Astro.cookies` contains utilities for reading and manipulating cookies in [server-side rendering](/en/guides/on-demand-rendering/) mode.
+`cookies` contains utilities for reading and manipulating cookies in [server-side rendering](/en/guides/server-side-rendering/) mode.
 
-##### `get`
+### `get`
 
 <p>
 
@@ -264,7 +308,7 @@ Astro.response.headers.set('Set-Cookie', 'a=b; Path=/;');
 
 Gets the cookie as an [`AstroCookie`](#astrocookie) object, which contains the `value` and utility functions for converting the cookie to non-string types.
 
-##### `has`
+### `has`
 
 <p>
 
@@ -273,7 +317,7 @@ Gets the cookie as an [`AstroCookie`](#astrocookie) object, which contains the `
 
 Whether this cookie exists. If the cookie has been set via `Astro.cookies.set()` this will return true, otherwise it will check cookies in the `Astro.request`.
 
-##### `set`
+### `set`
 
 <p>
 
@@ -282,7 +326,7 @@ Whether this cookie exists. If the cookie has been set via `Astro.cookies.set()`
 
 Sets the cookie `key` to the given value. This will attempt to convert the cookie value to a string. Options provide ways to set [cookie features](https://www.npmjs.com/package/cookie#options-1), such as the `maxAge` or `httpOnly`.
 
-##### `delete`
+### `delete`
 
 <p>
 
@@ -293,7 +337,7 @@ Invalidates a cookie by setting the expiration date in the past (0 in Unix time)
 
 Once a cookie is "deleted" (expired), `Astro.cookies.has()` will return `false` and `Astro.cookies.get()` will return an [`AstroCookie`](#astrocookie) with a `value` of `undefined`. Options available when deleting a cookie are: `domain`, `path`, `httpOnly`, `sameSite`, and `secure`.
 
-##### `merge`
+### `merge`
 
 <p>
 
@@ -302,7 +346,7 @@ Once a cookie is "deleted" (expired), `Astro.cookies.has()` will return `false` 
 
 Merges a new `AstroCookies` instance into the current instance. Any new cookies will be added to the current instance and any cookies with the same name will overwrite existing values.
 
-##### `headers`
+### `headers`
 
 <p>
 
@@ -443,14 +487,15 @@ If true, the cookie is only set on https sites.
 
 Allows customizing how the cookie is serialized.
 
-### `Astro.redirect()`
+## `Astro.redirect()`
 
 <p>
 
 **Type:** `(path: string, status?: number) => Response`
+<Since v="1.5.0" />
 </p>
 
-Allows you to redirect to another page, and optionally provide an [HTTP response status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages) as a second parameter.
+`redirect()` returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object that allows you to redirect to another page, and optionally provide an [HTTP response status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages) as a second parameter.
 
 A page (and not a child component) must `return` the result of `Astro.redirect()` for the redirect to occur.
 
@@ -473,7 +518,16 @@ if (!isLoggedIn(cookie)) {
 ---
 ```
 
-### `Astro.rewrite()`
+```ts
+import type { APIContext } from 'astro';
+
+export function GET({ redirect }: APIContext) {
+  return redirect('/login', 302);
+}
+```
+
+
+## `Astro.rewrite()`
 
 <p>
 
@@ -481,7 +535,7 @@ if (!isLoggedIn(cookie)) {
 <Since v="4.13.0" />
 </p>
 
-Allows you to serve content from a different URL or path without redirecting the browser to a new page. 
+`rewrite()` allows you to serve content from a different URL or path without redirecting the browser to a new page. 
 
 The method accepts either a string, a `URL`, or a `Request` for the location of the path.
 
@@ -493,12 +547,28 @@ return Astro.rewrite("/login")
 ---
 ```
 
+```ts
+import type { APIContext } from 'astro';
+
+export function GET({ rewrite }: APIContext) {
+  return rewrite('/login');
+}
+```
+
 Use a `URL` type when you need to construct the URL path for the rewrite. The following example renders a page's parent path by creating a new URL from the relative  `"../"` path:
 
 ```astro title="src/pages/blog/index.astro"
 ---
 return Astro.rewrite(new URL("../", Astro.url))
 ---
+```
+
+```ts
+import type { APIContext } from 'astro';
+
+export function GET({ rewrite }: APIContext) {
+  return rewrite(new URL("../", Astro.url));
+}
 ```
 
 Use a `Request` type for complete control of the `Request` sent to the server for the new path. The following example sends a request to render the parent page while also providing headers:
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -192,34 +125,145 @@ if (!post) {
...
-**Type:** `Request`
+**Type:** `URL`<br />
+<Since v="1.0.0" />
+</p>
+
+`url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` URL string value. Useful for interacting with individual properties of the request URL, like pathname and origin. 
+
+Equivalent to doing `new URL(Astro.request.url)`.
+
+`url` will be `localhost` in dev mode if [site](/en/reference/configuration-reference/#site) is not configured for static sites, and for on-demand rendered sites using `server` or `hybrid` output.
```
trueberryless, line 140:
```suggestion
`url` will be `localhost` in dev mode if [`site`](/en/reference/configuration-reference/#site) is not configured for static sites, and for on-demand rendered sites using `server` or `hybrid` output.
```

I think this would be more readable if `site` is in a inline code block, like it's done below:

```
Astro.site` returns a `URL` made from `site` in your Astro config. If `site` in your Astro config isn't defined, `Astro.site` won't be defined.
```

trueberryless (commented):
I have quickly looked over the page (the first half more in depth).

I think moving all headings up one step is a good decision cause before everthing was kinda nested under `Astro global` which didn't provide much context. Generally speaking, it just looks cleaner...

I'm wondering if the h4 headings should be visible in ToC, maybe on this page it would be more beneficial to override it, but I also love if it is consistent across all pages, so we prob should just leave it as it is...

From the content point of few, I think I know the Astro basics too bad to comment on that. Let me read the docs first xD

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
+  <h1>{title}</h1>
+  <p>{date}</p>
+</div>
 ```
 
-A callback function passed as `<Shout />`’s child will receive the all-caps `message` parameter:
-
-```astro title="src/pages/index.astro"
+{/* ```astro /title=".+"/ /date=".+"/ */}
+```astro "title=" "date="
```
TheOtterlord, line 40: We can do something like this if we want the contents highlighted too.
``````suggestion
```astro /(?:title|date)="(?:\\"|.)*?"/
``````

File: src/content/docs/en/reference/api-reference.mdx [hunk 3]
```diff
@@ -1,1161 +1,608 @@
...
-    {(message) => <div>{message}</div>}
-  </fragment>
-</Shout>
+<Heading title="My First Post" date="09 Aug 2022" />
 ```
 
-Use a standard HTML element for the wrapping tag, or any lower case tag (e.g. `<fragment>` instead of `<Fragment />`) that will not be interpreted as a component.  Do not use the HTML `<slot>` element as this will be interpreted as an Astro slot.
-
-### `Astro.self`
+<ReadMore>Learn more about how [Markdown and MDX Layouts](/en/guides/markdown-content/#frontmatter-layout-property) handle props.</ReadMore>
```
TheOtterlord, line 48: should this be lower-case?
```suggestion
<ReadMore>Learn more about how [Markdown and MDX layouts](/en/guides/markdown-content/#frontmatter-layout-property) handle props.</ReadMore>
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
+`url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` URL string value. It is useful for interacting with individual properties of the request URL, like pathname and origin. 
 
-<p>
-**Type:** `(action: TAction) => ActionReturnType<TAction> | undefined`<br />
-<Since v="4.15.0" />
-</p>
+`Astro.url` is equivalent to doing `new URL(Astro.request.url)`.
 
-`Astro.getActionResult()` is a function that returns the result of an [Action](/en/guides/actions/) submission. This accepts an action function as an argument (e.g. `actions.logout`) and returns a `data` or `error` object when a submission is received. Otherwise, it will return `undefined`.
+`url` will be `localhost` in dev mode for all on-demand rendered routes, and for prerendered routes when [`site`](/en/reference/configuration-reference/#site) is not configured.
```
TheOtterlord, line 169: I think I understand this right. The second segment is also in relation to dev mode only? I found it a little confusing during the first read. Probably just me, but thought it was worth a mention

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
 
 ```astro title="src/pages/index.astro"
```
TheOtterlord, line 171: I saw example usage highlighted above. I'm guessing that's the pattern we want?
``````suggestion
```astro title="src/pages/index.astro" "Astro.url"
``````

File: src/content/docs/en/reference/api-reference.mdx [hunk 6]
```diff
@@ -1,1161 +1,608 @@
...
-### `Astro.callAction()`
-
-<p>
-<Since v="4.15.0" />
-</p>
-
-`Astro.callAction()` is a function used to call an Action handler directly from your Astro component. This function accepts an Action function as the first argument (e.g. `actions.logout`) and any input that action receives as the second argument. It returns the result of the action as a promise.
+You can also use `url` to create new URLs by passing it as an argument to [`new URL()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL).
 
 ```astro title="src/pages/index.astro"
```
TheOtterlord, line 179:
``````suggestion
```astro title="src/pages/index.astro" "Astro.url"
``````

File: src/content/docs/en/basics/astro-syntax.mdx [hunk 7]
```diff
@@ -173,3 +173,140 @@ In Astro, you can use standard HTML comments or JavaScript-style comments.
...
+`Astro.slots` contains utility functions for modifying an Astro component's slotted children.
+
+#### `Astro.slots.has()`
+
+<p>
+
+**Type:** `(slotName: string) => boolean`
+</p>
+
+You can check whether content for a specific slot name exists with `Astro.slots.has()`. This can be useful when you want to wrap slot contents, but only want to render the wrapper elements when the slot is being used.
```
yanthomasdev, line 191:
```suggestion
You can check whether content for a specific slot name exists with `Astro.slots.has()`. This can be useful when you want to wrap slot contents but only want to render the wrapper elements when the slot is being used.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
-	
-The value will start with a leading slash and look similar to the path of a page component relative to your `srcDir/pages/` folder without a file extension.
-	
-For example, the file `src/pages/en/blog/[slug].astro` will return `/en/blog/[slug]` for `context.routePattern`. Every page on your site generated by that file (e.g. `/en/blog/post-1/`, `/en/blog/post-2/`, etc.) shares the same value for `context.routePattern`. In the case of `index.*` routes, the route pattern will not include the word "index." For example, `src/pages/index.astro` will return `/`.
-	
-You can use this property to understand which route is rendering your component. This allows you to target or analyze similarly-generated page URLs together. For example, you can use it to conditionally render certain information, or collect metrics about which routes are slower.
+`site` returns a `URL` made from `site` in your Astro config. It returns `undefined` if you have not set a value for [`site`](/en/reference/configuration-reference/#site) in your Astro config.
 
-See also: [`context.routePattern`](#contextroutepattern)
+AN EXAMPLE WOULD BE NICE
```
TheOtterlord, line 199: Mostly adding a comment to prevent this being lost! Example is just from the RSS recipe
``````suggestion
```astro title="src/pages/index.astro" "Astro.site"
<link
    rel="alternate"
    type="application/rss+xml"
    title="Your Site's Title"
    href={new URL("rss.xml", Astro.site)}
/>
```
``````

File: src/content/docs/en/basics/astro-syntax.mdx [hunk 9]
```diff
@@ -173,3 +173,140 @@ In Astro, you can use standard HTML comments or JavaScript-style comments.
...
+    {(message) => <div>{message}</div>}
+  </fragment>
+</Shout>
+```
+
+Use a standard HTML element for the wrapping tag, or any lower case tag (e.g. `<fragment>` instead of `<Fragment />`) that will not be interpreted as a component.  Do not use the HTML `<slot>` element as this will be interpreted as an Astro slot.
+
+### `Astro.self`
+
+`Astro.self` allows Astro components to be recursively called. This behaviour lets you render an Astro component from within itself by using `<Astro.self>` in the component template. This can be helpful for iterating over large data stores and nested data-structures.
```
yanthomasdev, line 268:
```suggestion
`Astro.self` allows Astro components to be recursively called. This behavior lets you render an Astro component from within itself by using `<Astro.self>` in the component template. This can help iterate over large data stores and nested data structures.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
-import type { APIContext } from 'astro';
+`clientAddress` specifies the [IP address](https://en.wikipedia.org/wiki/IP_address) of the request. This property is only available for routes rendered on demand and cannot be used on prerendered pages.
 
-export function GET(context: APIContext) {
-  // ...
-}
-```
+<Tabs>
+  <TabItem label="Astro.clientAddress">
+    ```astro title="src/pages/ip-address.astro"
```
TheOtterlord, line 213:
``````suggestion
    ```astro title="src/pages/ip-address.astro" "Astro.clientAddress"
``````

File: src/content/docs/en/reference/api-reference.mdx [hunk 11]
```diff
@@ -1,1161 +1,608 @@
...
+    export const prerender = false; // Not needed in 'server' mode
+    const ip = Astro.clientAddress;
+    ---
 
-### `context.params`
+    <div>Your IP address is: <span class="address">{ ip }</span></div>
+    ```
+  </TabItem>
+  <TabItem label="context.clientAddress">
+    ```ts title="src/pages/ip-address.ts"
```
TheOtterlord, line 223:
``````suggestion
    ```ts title="src/pages/ip-address.ts" "clientAddress"
``````

File: src/content/docs/en/reference/api-reference.mdx [hunk 12]
```diff
@@ -1,1161 +1,608 @@
...
 
-### `context.url`
+`Astro.response` is used to set the `status`, `statusText`, and `headers` for a page's response.
 
-<p>
 
-**Type:** `URL`<br />
-<Since v="1.5.0" />
-</p>
+```astro
```
TheOtterlord, line 332:
``````suggestion
```astro "Astro.response"
``````

File: src/content/docs/en/reference/api-reference.mdx [hunk 13]
```diff
@@ -1,1161 +1,608 @@
...
+  Astro.response.statusText = 'Not found';
+}
+---
+```
 
-A [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `context.request.url` URL string value.
+Or to set a header:
 
-See also: [Astro.url](#astrourl)
+```astro
```
TheOtterlord, line 343:
``````suggestion
```astro "Astro.response"
``````

File: src/content/docs/en/reference/api-reference.mdx [hunk 14]
```diff
@@ -1,1161 +1,608 @@
...
+</Tabs>
+See also: [Data Passing with `props`](/en/reference/routing-reference/#data-passing-with-props)
 
-This component could then be used like this:
+### `params`
 
-```astro
----
-import NestedList from './NestedList.astro';
+`params` is an object containing the values of dynamic route segments matched for a route request. Its keys must match the [dynamic route parameters](/en/guides/routing/#dynamic-routes) from the the page file or endpoint route.
```
yanthomasdev, line 95:
```suggestion
`params` is an object containing the values of dynamic route segments matched for a route request. Its keys must match the [dynamic route parameters](/en/guides/routing/#dynamic-routes) from the page file or endpoint route.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
 
-See also: [Astro.clientAddress](#astroclientaddress)
+When using an on-demand rendering mode, status codes are supported. Astro will serve redirected requests with a default HTTP response status of `302` unless another code is specified.
 
+The following example redirects a user to a login page:
 
-### `context.site`
+<Tabs>
+  <TabItem label="Astro.redirect()">
+    ```astro title="src/pages/account.astro" {8}
```
TheOtterlord, line 369:
```suggestion
    ```astro title="src/pages/account.astro" {8} "Astro.redirect"
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
-</p>
+    // If the user is not logged in, redirect them to the login page
+    if (!isLoggedIn(cookie)) {
+      return Astro.redirect('/login');
+    }
+    ---
+    ```
+  </TabItem>
+  <TabItem label="context.redirect()">
+    ```ts
```
TheOtterlord, line 383:
``````suggestion
    ```ts "redirect"
``````

File: src/content/docs/en/reference/api-reference.mdx [hunk 17]
```diff
@@ -1,1161 +1,608 @@
...
-### `context.request`
+### `request`
 
 <p>
 
 **Type:** `Request`
 </p>
 
-A standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object. It can be used to get the `url`, `headers`, `method`, and even body of the request.
+`request` is a standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object. It can be used to get the `url`, `headers`, `method`, and even body of the request.
```
yanthomasdev, line 291:
```suggestion
`request` is a standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object. It can be used to get the `url`, `headers`, `method`, and even the body of the request. 
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
 
-**Type:** `(path: string, status?: number) => Response`<br />
-<Since v="1.5.0" />
-</p>
+Use a string to provide an explicit path:
 
-`context.redirect()` returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object that allows you to redirect to another page. This function is only available when building for SSR (server-side rendering) and should not be used for static sites.
+<Tabs>
+  <TabItem label="Astro.rewrite()">
+    ```astro title="src/pages/index.astro"
```
TheOtterlord, line 410:
```suggestion
    ```astro title="src/pages/index.astro" "Astro.rewrite"
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
 
-**Type:** `(rewritePayload: string | URL | Request) => Promise<Response>`<br />
-<Since v="4.13.0" />
+<Since v="2.4.0" />
 </p>
 
-Allows you to serve content from a different URL or path without redirecting the browser to a new page. 
-
-The method accepts either a string, a `URL`, or a `Request` for the location of the path.
+`locals` is an  object used to store and access arbitrary information during the lifecycle of a request. `Astro.locals` is an object containing any values from the `context.locals` object from a middleware. Use this to access data returned by middleware in your `.astro` files.
```
yanthomasdev, line 484:
```suggestion
`locals` is an object used to store and access arbitrary information during the lifecycle of a request. `Astro.locals` is an object containing any values from the `context.locals` object from a middleware. Use this to access data returned by middleware in your `.astro` files.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
+<Tabs>
+  <TabItem label="Astro.rewrite()">
+    ```astro title="src/pages/index.astro"
+    ---
+    return Astro.rewrite("/login")
+    ---
+    ```
+  </TabItem>
+  <TabItem label="context.rewrite()">
+    ```ts
```
TheOtterlord, line 417:
```suggestion
    ```ts "rewrite"
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
+export const onRequest: MiddlewareHandler = ({ locals }, next) => {
+  if (!locals.title) {
+    locals.title = "Default Title";
+  }
+  return next();
 }
 ```
 
-Use a `URL` type when you need to construct the URL path for the rewrite. The following example renders a page's parent path by creating a new URL from the relative  `"../"` path:
+`Astro.locals` is read only for Astro components and `context.locals` is read only for API endpoints:
```
yanthomasdev, line 499:
```suggestion
`Astro.locals` is read-only for Astro components and `context.locals` is read-only for API endpoints:
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
-     "x-custom-header": JSON.stringify(Astro.locals.someValue)
-   }
- }));
-}
-```
+**Type:** `string | undefined`<br />
+<Since v="3.5.0" />
+</p>
+
+`preferredLocale` is a computed value to find the best match between your visitor's brower language preferences and the locales supported by your site.
```
yanthomasdev, line 535:
```suggestion
`preferredLocale` is a computed value to find the best match between your visitor's browser language preferences and the locales supported by your site.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
+
+It is computed by checking the configured locales in your `i18n.locales` array and the locales supported by the users's browser via the header `Accept-Language`. This value is `undefined` if no such match exists.
```
yanthomasdev, line 537:
```suggestion
It is computed by checking the configured locales in your `i18n.locales` array and the locales supported by the user's browser via the header `Accept-Language`. This value is `undefined` if no such match exists.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
-
-The method accepts either a string, a `URL`, or a `Request` for the location of the path.
+`locals` is an  object used to store and access arbitrary information during the lifecycle of a request. `Astro.locals` is an object containing any values from the `context.locals` object from a middleware. Use this to access data returned by middleware in your `.astro` files.
 
-Use a string to provide an explicit path:
+Middleware functions can both read and write the values of `context.locals`:
 
-```ts
-import type { APIContext } from 'astro';
+```ts title="src/middleware.ts"
```
TheOtterlord, line 488:
```suggestion
```ts title="src/middleware.ts" "locals"
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
+  return next();
 }
 ```
 
-Use a `URL` type when you need to construct the URL path for the rewrite. The following example renders a page's parent path by creating a new URL from the relative  `"../"` path:
+`Astro.locals` is read only for Astro components and `context.locals` is read only for API endpoints:
+
+<Tabs>
+  <TabItem label="Astro.locals">
+    ```astro title="src/pages/Orders.astro"
```
TheOtterlord, line 503:
```suggestion
    ```astro title="src/pages/Orders.astro" "Astro.locals"
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
+    <h1>{title}</h1>
+    <ul>
+        {orders.map(order => {
+            return <li>{/* do something with each order */}</li>
+        })}
+    </ul>
+    ```
+  </TabItem>
+  <TabItem label="context.locals">
+    ```ts title="src/pages/hello.ts"
```
TheOtterlord, line 517:
```suggestion
    ```ts title="src/pages/hello.ts" "Astro.locals"
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1167,314 +614,321 @@ The route pattern responsible for generating the current page or route. In file-
...
-:::tip
-When using TypeScript, use the [`GetStaticPaths`](/en/guides/typescript/#infer-getstaticpaths-types) type utility to ensure type-safe access of your `params` and `props`.
-:::
+**Type:** <code>(key: string, options?: <a href="#astrocookiegetoptions">AstroCookieGetOptions</a>) => boolean</code>
+</p>
 
-:::caution
-The `getStaticPaths()` function executes in its own isolated scope once, before any page loads. Therefore you can't reference anything from its parent scope, other than file imports. The compiler will warn if you break this requirement.
-:::
+Whether this cookie exists. If the cookie has been set via `Astro.cookies.set()` this will return true, otherwise it will check cookies in the `Astro.request`.
```
yanthomasdev, line 649:
```suggestion
Whether this cookie exists. If the cookie has been set via `Astro.cookies.set()` this will return true, otherwise, it will check cookies in the `Astro.request`.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1167,314 +614,321 @@ The route pattern responsible for generating the current page or route. In file-
...
+    <h2>{post.frontmatter.title}</h2>
+    <p>{post.frontmatter.description}</p>
+    <a href={post.url}>Read more</a>
+  </article>
+))}
+</div>
+```
 
-##### `page.url.last`
+`.glob()` only takes one parameter: a relative URL glob of which local files you'd like to import. It’s asynchronous, and returns an array of the exports from matching files.
```
yanthomasdev, line 855:
```suggestion
`.glob()` only takes one parameter: a relative URL glob of which local files you'd like to import. It’s asynchronous and returns an array of the exports from matching files.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
 
-:::note
-The value of `locals` cannot be overridden at run time. Doing so would risk wiping out all the information stored by the user.  Astro performs checks and will throw an error if `locals` are overridden.
-:::
+**Type:** `string | undefined`<br />
+<Since v="3.5.6" />
+</p>
 
-See also: [`Astro.locals`](#astrolocals)
+The locale computed from the current URL, using the syntax specified in your `locales` configuration. If the URL does not contain a `/[locale]/` prefix, then the value will default to `i18n.defaultLocale`.
```
TheOtterlord, line 565: This is `undefined` if you have `routing.prefixDefaultLocale` set to `true`.

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,293 @@
...
+import ReadMore from '~/components/ReadMore.astro';
+
+There is no separate routing configuration in Astro. Every [supported page file](/en/basics/astro-pages/#supported-page-files) located within the special `src/pages/` directory creates a single page route, or multiple page routes dynamically when the file name contains a [parameter](#params).
+
+
+## `getStaticPaths()`
+
+**Type:** `(options: GetStaticPathsOptions) => Promise<GetStaticPathsResult> | GetStaticPathsResult`
+
+If a page uses dynamic params in the filename, that component will need to export a `getStaticPaths()` function.
```
yanthomasdev, line 19: I think the right thing would be to spell the full word (parameters) or then reference the property with code syntax, right?
```suggestion
If a page uses dynamic parameters in the filename, that component will need to export a `getStaticPaths()` function.
```

```suggestion
If a page uses dynamic `params` in the filename, that component will need to export a `getStaticPaths()` function.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
 <p>
-
 **Type:** `(action: TAction) => ActionReturnType<TAction> | undefined`<br />
 <Since v="4.15.0" />
 </p>
 
-`context.getActionResult()` is a function that returns the result of an [Action](/en/guides/actions/) submission. This accepts an action function as an argument (e.g. `actions.logout`), and returns a `data` or `error` object when a submission is received. Otherwise, it will return `undefined`.
+`getActionResult()` is a function that returns the result of an [Action](/en/guides/actions/) submission. This accepts an action function as an argument (e.g. `actions.logout`) and returns a `data` or `error` object when a submission is received. Otherwise, it will return `undefined`.
+
+```astro title="src/pages/index.astro"
```
TheOtterlord, line 576:
``````suggestion
```astro title="src/pages/index.astro" "Astro.getActionResult"
``````

File: src/content/docs/en/reference/routing-reference.mdx [hunk 32]
```diff
@@ -0,0 +1,293 @@
...
+
+This function is required because Astro is a static site builder. That means that your entire site is built ahead of time. If Astro doesn't know to generate a page at build time, your users won't see it when they visit your site.
```
yanthomasdev, line 21:
```suggestion
This function is required because Astro is a static site builder. That means that your entire site is built ahead of time. If Astro doesn't know how to generate a page at build time, your users won't see it when they visit your site.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,1161 +1,608 @@
...
+### `callAction()`
 
 <p>
 <Since v="4.15.0" />
 </p>
 
-`context.callAction()` is a function used to call an Action handler directly from your Astro component. This function accepts an Action function as the first argument (e.g. `actions.logout`) and any input that action receives as the second argument. It returns the result of the action as a promise.
+`callAction()` is a function used to call an Action handler directly from your Astro component. This function accepts an Action function as the first argument (e.g. `actions.logout`) and any input that action receives as the second argument. It returns the result of the action as a promise.
 
+```astro title="src/pages/index.astro"
```
TheOtterlord, line 597:
```suggestion
```astro title="src/pages/index.astro" "Astro.callAction"
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,293 @@
...
+The `getStaticPaths()` function should return an array of objects to determine which paths will be pre-rendered by Astro.
+
+It can also be used in static file endpoints for [dynamic routing](/en/guides/endpoints/#params-and-dynamic-routing).
+
+:::tip
+When using TypeScript, use the [`GetStaticPaths`](/en/guides/typescript/#infer-getstaticpaths-types) type utility to ensure type-safe access of your `params` and `props`.
+:::
+
+:::caution
+The `getStaticPaths()` function executes in its own isolated scope once, before any page loads. Therefore you can't reference anything from its parent scope, other than file imports. The compiler will warn if you break this requirement.
```
yanthomasdev, line 46:
```suggestion
The `getStaticPaths()` function executes in its own isolated scope once, before any page loads. Therefore you can't reference anything from its parent scope, other than file imports. The compiler will warn you if you break this requirement.
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,293 @@
...
+:::
+
+### `params`
+
+The `params` key of every returned object tells Astro what routes to build. The returned params must map back to the dynamic parameters and rest parameters defined in your component filepath.
+
```
yanthomasdev, line 52:
```suggestion
The `params` key of every returned object tells Astro what routes to build. The returned `params` must map back to the dynamic parameters and rest parameters defined in your component filepath.
```

TheOtterlord (commented):
Most of these add text highlighting, but there's a couple points relating to accuracy/missing content.

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,293 @@
...
+<h1>{id}</h1>
+```
+
+Then Astro will statically generate `posts/1`, `posts/2`, and `posts/3` at build time.
+
+### Data Passing with `props`
+
+To pass additional data to each generated page, you can also set a `props` value on every returned path object. Unlike `params`, `props` are not encoded into the URL and so aren't limited to only strings.
+
+For example, suppose that you generate pages based off of data fetched from a remote API. You can pass the full data object to the page component inside of `getStaticPaths`:
```
yanthomasdev, line 78: I think "based on" might be simpler?
```suggestion
For example, suppose that you generate pages based on data fetched from a remote API. You can pass the full data object to the page component inside of `getStaticPaths`:
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,293 @@
...
+  <h1>{id}: {post.title}</h1>
+  <h2>Category: {post.category}</h2>
+</body>
+```
+
+Then Astro will statically generate `posts/1` and `posts/2` at build time using the page component in `pages/posts/[id].astro`. The page can reference this data using `Astro.props`:
+
+### `paginate()`
+
+Pagination is a common use-case for websites that Astro natively supports via the `paginate()` function. `paginate()` will automatically generate the array to return from `getStaticPaths()` that creates one URL for every page of the paginated collection. The page number will be passed as a param, and the page data will be passed as a `page` prop.
```
yanthomasdev, line 128:
```suggestion
Pagination is a common use-case for websites that Astro natively supports via the `paginate()` function. `paginate()` will automatically generate the array to return from `getStaticPaths()` that creates one URL for every page of the paginated collection. The page number will be passed as a `param`, and the page data will be passed as a `page` prop.
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,293 @@
...
+Array of data returned from the `paginate()` function for the current page.
+
+##### `page.start`
+
+<p>
+
+**Type:** `number`
+</p>
+
+Index of first item on current page, starting at `0`. (e.g. if `pageSize: 25`, this would be `0` on page 1, `25` on page 2, etc.)
```
yanthomasdev, line 183:
```suggestion
Index of the first item on the current page, starting at `0`. (e.g. if `pageSize: 25`, this would be `0` on page 1, `25` on page 2, etc.)
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+##### `page.end`
+
+<p>
+
+**Type:** `number`
+</p>
+
+Index of last item on current page.
```
yanthomasdev, line 192:
```suggestion
Index of the last item on the current page.
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+##### `page.size`
+
+<p>
+
+**Type:** `number`<br />
+**Default:** `10`
+</p>
+
+How many items per-page.
```
yanthomasdev, line 202: This sounds more like a question, maybe rewording a bit does the trick? Also, makes it more consistent with `page.total`
```suggestion
The total number of items per page.
```

yanthomasdev (commented):
Good work!

src/content/docs/en/reference/api-reference.mdx
sarah11918: I actually changed this because I *didn't* want the contents highlighted. 😅  Do you think it looks better with?

src/content/docs/en/reference/routing-reference.mdx
sarah11918: Agreed it's from. And it's not "suppose that" :smile:

src/content/docs/en/reference/api-reference.mdx
TheOtterlord: nope, looks fine as is! was just confused by the comment. lgtm!

src/content/docs/en/reference/api-reference.mdx
sarah11918: I actually was concerned the original was confusing. 😅  Currently it says:

> Astro.url will be localhost in dev mode if [site](https://docs.astro.build/en/reference/configuration-reference/#site) is not configured for static sites, and for on-demand rendered sites using server or hybrid output.

I am assuming it's dev mode for all? I wouldn't be localhost in production, right?

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,855 +1,245 @@
...
-
-const result = Astro.getActionResult(actions.logout);
----
+`site` returns a `URL` made from `site` in your Astro config. It returns `undefined` if you have not set a value for [`site`](/en/reference/configuration-reference/#site) in your Astro config.
 
-<form action={actions.logout}>
-  <button type="submit">Log out</button>
-</form>
-{result?.error && <p>Failed to log out. Please try again.</p>}
+```astro title="src/pages/index.astro" "Astro.site"
```
sarah11918, line 199: Flagging that I had asked for an example here, and Reuben provided a placeholder one. Will this work, or do we need to change it?

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,855 +1,245 @@
...
+      ];
+    }
+
+    const { id } = Astro.params;
+    const { author } = Astro.props;
+    ---
+    ```
+  </TabItem>
+  <TabItem label="context.props">
+    ```ts title="src/pages/posts/[id].json.ts" "props:" {11-14}
```
ArmandPhilippot, line 71: nit: currently for the `context.props` example, everything is highlighted for `export function GET` except the closing brace. I think this should be included to avoid confusion if someone tries to copy only the highlighted part:
```suggestion
    ```ts title="src/pages/posts/[id].json.ts" "props:" {11-15}
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,855 +1,245 @@
...
+        { params: { id: '3' } }
+      ];
+    }
+    const { id } = Astro.params;
+    ---
+    <h1>{id}</h1>
+    ```
+  </TabItem>
+  <TabItem label="context.params">
+    ```ts title="src/pages/posts/[id].json.ts" "params:" {11-14}
```
ArmandPhilippot, line 116: nit: same as above, I think the closing brace should be included:
```suggestion
    ```ts title="src/pages/posts/[id].json.ts" "params:" {11-15}
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,855 +1,245 @@
...
 
-A callback function passed as `<Shout />`’s child will receive the all-caps `message` parameter:
+const post = await getPost(Astro.params.id);
 
-```astro title="src/pages/index.astro"
----
-import Shout from "../components/Shout.astro";
+// No posts found with this ID
+if (!post) {
+  Astro.redirect("/404")
```
ArmandPhilippot, line 147: It's time to fix my error... 👀 When I updated this part, I forgot the return statement: `Astro.redirect` should be returned!
```suggestion
  return Astro.redirect("/404")
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -861,301 +251,365 @@ A boolean representing whether or not the current page is prerendered.
...
+    ```ts
+    import type { APIContext } from 'astro';
+
+    export function GET({ rewrite }: APIContext) {
+      return rewrite(new Request(new URL("../", Astro.url), {
+      headers: {
+        "x-custom-header": JSON.stringify(Astro.locals.someValue)
+      }
+    }));
+    }
```
ArmandPhilippot, line 479: nit: indentation
```suggestion
    export function GET({ rewrite }: APIContext) {
      return rewrite(new Request(new URL("../", Astro.url), {
        headers: {
          "x-custom-header": JSON.stringify(Astro.locals.someValue)
        }
      }));
    }
```

sarah11918:
There's no rush @ArmandPhilippot , I will be fixing links FOREVER still!  😅

ArmandPhilippot:
@sarah11918 Oh you're used to it now, it'll go quickly... 😆

sarah11918:
THIS IS NOT GOING QUICKLY ARMAND

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1,855 +1,245 @@
...
-    </li>
-  ))}
-</ul>
-```
+**Type:** `URL`<br />
+<Since v="1.0.0" />
+</p>
 
-This component could then be used like this:
+`url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` URL string value. It is useful for interacting with individual properties of the request URL, like pathname and origin.
```
ArmandPhilippot, line 165: I wonder if the sentence is a bit "heavy" (hard to read) because of the repetition of `URL` (lower/upper). Would something like this work:
```suggestion
`url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` value. It is useful for interacting with individual properties of the request URL, like pathname and origin. 
```
That only takes away one... and I understand that this was to emphasize that the value is a string, but is it useful here? I don't think we need to know that the value used to compute `Astro.url` is a string... So maybe this help to lighten the sentence.

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: I think the example is fine!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -861,301 +251,365 @@ A boolean representing whether or not the current page is prerendered.
...
+    <h1>{title}</h1>
+    <ul>
+        {orders.map(order => {
+            return <li>{/* do something with each order */}</li>
+        })}
+    </ul>
+    ```
+  </TabItem>
+  <TabItem label="context.locals">
+    ```ts title="src/pages/hello.ts" "Astro.locals"
```
ArmandPhilippot, line 524: nit, highlighting. This the context tabs, so `Astro.locals` does not work, I think this should be:
```suggestion
    ```ts title="src/pages/hello.ts" "locals"
```

ArmandPhilippot (approved):
After reading again, I'd say: LGTM! Just noted another nitpick regarding a code snippet highlighting and maybe a sentence a bit hard to read. Well, there are also broken links... but I trust you on that! 🙌🏽 
Regarding the question in the first post:
* is the one about `astro-syntax`/`/basics/` still accurate? I haven't seen `basics`, so looks good to me.
* I don't have a strong opinion about the file renaming for `api-reference` so... as much as I know you like to correct links, I would say we can leave it like that... 😄

ArmandPhilippot:
> THIS IS NOT GOING QUICKLY ARMAND

You've lost your touch since the last PR... 😢

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+const { id } = Astro.params;
+---
+<h1>{id}</h1>
+```
+
+Then Astro will statically generate `posts/1`, `posts/2`, and `posts/3` at build time.
+
+### Data Passing with `props`
+
+To pass additional data to each generated page, you can also set a `props` value on every returned path object. Unlike `params`, `props` are not encoded into the URL and so aren't limited to only strings.
```
delucis, line 75: This would align this with my `params` suggestion above:

```suggestion
To pass additional data to each generated page, you can set a `props` value on each object in the array returned by `getStaticPaths()`. Unlike `params`, `props` are not encoded into the URL and so aren't limited to only strings.
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+
+For example, if you generate pages based from data fetched from a remote API, you can pass the full data object to the page component inside of `getStaticPaths`:
```
delucis, line 77: Nits:

```suggestion
For example, if you generate pages with data fetched from a remote API, you can pass the full data object to the page component inside of `getStaticPaths()`:
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,292 @@
...
+const {id} = Astro.params;
+const {post} = Astro.props;
+---
+<body>
+  <h1>{id}: {post.title}</h1>
+  <h2>Category: {post.category}</h2>
+</body>
+```
+
+Then Astro will statically generate `posts/1` and `posts/2` at build time using the page component in `pages/posts/[id].astro`. The page can reference this data using `Astro.props`:
```
delucis, line 123: Maybe this whole second chunk can be deleted? It feels like it’s not adding anything?
```suggestion
```

src/content/docs/en/reference/routing-reference.mdx
delucis, line 1: Maybe this page would also be the place to document `export const prerender`? I guess that’s kind of a “routing” concern?

delucis (commented):
First batch of comments! Still need to go through the new syntax and runtime reference pages.

Incredible work on the link fixing here. Wow… so many! 😅

astro.sidebar.ts
```diff
@@ -92,13 +92,14 @@ export const sidebar = [
...
 			group('reference.syntax', {
-				items: ['basics/astro-syntax', 'reference/directives-reference'],
+				items: ['reference/astro-syntax', 'reference/directives-reference'],
 			}),
 			'reference/configuration-reference',
 			'reference/cli-reference',
 			'guides/imports',
 			group('reference.runtime', {
 				items: [
+					'reference/routing-reference',
```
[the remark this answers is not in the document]
sarah11918: Done!

src/content/docs/en/reference/routing-reference.mdx
sarah11918: Done!

sarah11918:
Note: all Chris's suggestions committed via IDX; two open comments left unresolved.

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,312 @@
...
+
+:::caution
+The `getStaticPaths()` function executes in its own isolated scope once, before any page loads. Therefore you can't reference anything from its parent scope, other than file imports. The compiler will warn you if you break this requirement.
+:::
+
+### `params`
+
+The `params` key of each object in the array returned by `getStaticPaths()` tells Astro what routes to build.
+
+The keys in `params` must match the parameters defined in your component filepath. The value for each `params` object must match the parameters used in the page name. `params` are encoded into the URL, so only strings are supported as values.
```
delucis, line 99: “filepath” or “file path”? I get red squigglies for “filepath” 😄 

```suggestion
The keys in `params` must match the parameters defined in your component file path. The value for each `params` object must match the parameters used in the page name. `params` are encoded into the URL, so only strings are supported as values.
```

delucis (commented):
Round two on the routing API, I really do need to move on to those other pages don’t I 😅 

Prerender section is great!

delucis (commented):
Honestly amazing work on this. I think it’s night and day with how we were doing things previously 👏 

Hope all these comments aren’t too overwhelming. A lot of small little things tbh, which isn’t GitHub’s strong suit.

One overarching thought for the API reference page: we could use a [`syncKey`](https://starlight.astro.build/components/tabs/#sync-tabs) on the tabs if we wanted to link the `Astro.` and `context.` examples so they toggled together. Not super vital but there are a couple of spots with multiple examples that might benefit. Absolutely not a showstopper though. Can always happen later.

src/content/docs/en/reference/astro-syntax.mdx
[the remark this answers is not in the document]
sarah11918: Done!

sarah11918:
> we could use a [syncKey](https://starlight.astro.build/components/tabs/#sync-tabs) on the tabs

I could only find one entry that used multiple tabs (`rewrite()`), so I added that there! Otherwise, I had thought about it, but since the labels are the names of the entries, it's not a straight up match down the page.

ArmandPhilippot:
> I could only find one entry that used multiple tabs (`rewrite()`), so I added that there! Otherwise, I had thought about it, but since the labels are the names of the entries, it's not a straight up match down the page.

I think Chris meant across the whole page. So if I select `context` in `rewrite()`, all the other sections (`redirect()`, `generator`, etc.) show me the `context` tab directly. If someone is viewing the page with `Astro.` in mind, they are likely to only want to see `Astro.` tabs, and vice versa with `context`.
So exactly the same functioning as `PackageManagerTabs` (which uses the `package-managers` key), except that here it is `Astro` and `context`.

I think this is a good idea! I didn't add suggestions to the relevant places because, in that case, the key name (`rewrite`) may no longer be appropriate. But I can't find a meaningful word to connect the two... (same issue as the page title 😅 ). I did think of `context` (since the word is used in the title) but it could be confusing because of the context tab... Perhaps `render` or `render-context` or even `astro-context`?

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -0,0 +1,312 @@
...
+Every [supported page file](/en/basics/astro-pages/#supported-page-files) located within the special `src/pages/` directory creates a route. When the file name contains a [parameter](#params), a route can create multiple pages dynamically, otherwise it creates a single page.
+
+By default, all Astro page routes and endpoints are generated and prerendered at build time. [On-demand server rendering](/en/guides/on-demand-rendering/) can be set for individual routes, or as the default.
+
+## `prerender`
+
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `true` <br />
```
ArmandPhilippot, line 23: I already approved, but with Chris suggestions (🙌🏽 ) the content has changed a bit. Everything looks fine with a quick read (I didn't check for typo), but I'm a bit concern about this `Default:`.

Both `Per-page override` and `Switch to server mode` are under this `h2`. And while `true` is the default in static mode, this is not the case (from my understanding) with `server` mode. So it could be confusing.
I think we should either move `Default:` into subsections or use a slightly more complex `Default:`. Perhaps something like:
```suggestion
**Default:** `true` with `output: 'static'` or `false` with `output: 'server'` <br />
```

sarah11918: Thanks for commenting on this! I had wondered about this, because you're right, it's not the default in `server` mode. It *is* true that it's default **until you do something to change the nature of your project**, which is how the text was handling it. I wasn't sure whether we had a standard way of either/or 'ing it, or whether it was in fact actually helpful to present the default as `true` (because it is until something else happens.)

sarah11918:
> I think Chris meant across the whole page.

If you look at how the labels are implemented, it's not that simple! 😅  That's what I meant when I said it's not easy to get it sync'd all down the page. The tabs don't all share the same label because the tabs themselves are the `Astro.x` and `context.x` of the entry. So, there aren't duplicates!

`label` does double duty of syncing the tabs **and** it's what the visual tab label is: and I want the tabs visually to show the very specific occurrence of `Astro.` and `context.` the things. We would need to be able to designate a tab property as `Astro` or `context` while showing something different as the tab label. Right now, these are controlled by the same value.

src/content/docs/en/reference/routing-reference.mdx
ArmandPhilippot: I was hesitant about that. I understood the intention given the structure of the page, but I am now quite familiar with how it works (I think 😄 ). So I totally understand what you mean.
But for someone new it might be less obvious... especially if they quickly scan the page instead of reading the explanations in detail...
(When I see some questions asked, sometimes I just want to say "but have you read the doc"!? I admire your patience! 🙄) So I don't know what the best way to solve this is, but I think it could be problematic.

Regarding the standard way of doing it, I don't think so. But I'm sure there is another place where `Default:` is used as I suggested (two possible values, not the wording). But I can't find the page anymore... I don't think it's been edited, maybe moved. Or I'm not looking in the right place.

sarah11918: I had thought we had another place too, but I didn't immediately find it and didn't want to introduce inconsistency! If you can't find it either, maybe it didn't make it into the new docs!

I'm happy to go with your suggestion if you feel it's more helpful!

ArmandPhilippot: Ok, so I only just saw your comment now... (thanks the email notifications...) I was on the page, but it seems I'm having auto-refresh issues with Github the last few days (notifs in particular). Even after posting the link, your comment wasn't displayed... 

> I'm happy to go with your suggestion if you feel it's more helpful! My only slight nagging concern is that `output: 'static'` is a default that no one would set. And, it's def true by default default, even if they don't know what `output: 'static'` is. Like, this is _really_ the default. And I _do_ think there's value in it being presented that way.

Yeah, I understand your concern and honestly I don't have a perfect solution. Considering the way the content is presented, I would say it is fine the way it is. But then I think that some people read the doc too quickly and they risk stopping at the word in bold (`Default:`) without reading the rest and that could cause confusion.

So I thought I'd just mention it just in case. Now, especially since we're not sure what's best, I also agree to leave it as is, and if it's a problem, we'll update later.

sarah11918: Thank you for the conversation! I'm leaning towards leaving it like this and then waiting to see whether it causes confusion or more people mention it. It can always be changed!

I think part of why I have this opinion is the reason behind making the old hybrid behaviour the default is because too many people were simply enabling `output: 'server'` just to have ANY on-demand rendering features, even when they only needed it on one page. We're also trying to position the docs to nudge towards the behaviour of thinking of your site as prerendered, opting in per-route when you need it, and only when and if you realize you *really* need it, switching to `server` mode. I do think this framing helps that story as well!

delucis (approved):
Looking good @sarah11918! I bow down to your indomitable spirit in taming the reference wilderness.

I do agree with @ArmandPhilippot re: the “default” field for `prerender` potentially being confusing given it varies depending on mode, but will leave it to your discretion whether you want to address that or are happy leaving it as is.

PR #10077, opened 2024-11-20
feat(next): routes resolved hook

florian-lefebvre:
#### Description (required)

https://github.com/withastro/astro/pull/12329

- Closes #
- Suggested label:


src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -361,6 +361,42 @@ export default function createIntegration() {
 
 <ReadMore>Learn more about [the Adapter API](/en/reference/adapter-reference/) for building adapter integrations.</ReadMore>
 
+### Deprecated: `routes` on `astro:build:done` hook (Integration API)
+
+<SourcePR number="12329" title="feat(next): astro:routes:resolved"/>
+
+In Astro v4.x, you could only get routes in an integration from the `astro:build:done` hook.
+
+Astro v5.0 deprecates the `routes` array passed to this hook. Instead, it exposes a new `astro:routes:resolved` hook that runs before `astro:config:done`, and whenever a route changes in development. It has all the same properties of the deprecated `routes` list, except `distURL` which is only available during build. For this, you can use the newly exposed `assets` map:
```
sarah11918, line 370:
```suggestion
Astro v5.0 deprecates the `routes` array passed to this hook. Instead, it exposes a new `astro:routes:resolved` hook that runs before `astro:config:done`, and whenever a route changes in development. It has all the same properties of the deprecated `routes` list, except `distURL` which is only available during build. 

#### What should I do?

Remove any instance of `routes` passed to `astro:build:done` and replace it with the new `astro:routes:resolved` hook. Access `distURL` on the newly exposed `assets` map:
```
Just needed something to fit into this pattern. Can you update the sentence above to make it correct/helpful?

sarah11918 (commented):
Looks great @florian-lefebvre ! Just need to phrase this more like a "What should I do?" so I made a basic model, but you should update as necessary!

sarah11918:
Oh, and are we documenting this on the Integrations API page too? Is that coming later?

florian-lefebvre:
There you go :smile:

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+
+  /**
+	 * @param {any} data The optional parameters of the route
+	 *
+	 * @description
+	 * A function that accepts a list of params, interpolates them with the route pattern, and returns the path name of the route.
+	 *
+	 * ## Example
+	 *
+	 * For a route such as `/blog/[...id].astro`, the `generate` function would return something like this:
```
sarah11918, line 546:
```suggestion
	 * For the route `/blog/[...id].astro`, the `generate` function returns:
```
As much as possible, try to cut extra words that don't contain "meaning". I know you're showing a "possible example", so everything feels kind of uncertain. But the solution, when possible, is to define a thing, and then say THIS THING GIVES YOU THIS.  Whatever you need to state/assume to avoid being wishy washy is both more readable (more direct! fewer words that don't add to the meaning! easier to translate!) AND makes the reader feel more confident because YOU are giving examples that you are confident about.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+	 *
+	 * ```js
+	 * console.log(generate({ id: 'presentation' })) // will log `/blog/presentation`
+	 * ```
+	 */
+	generate: (data?: any) => string;
+
+  /**
+	 * Dynamic and spread route params
```
sarah11918, line 555:
```suggestion
	 * Dynamic and spread route params in the file path
```
Is this correct? When we can, try to *explain* in the definition vs just making people guess from the example what pattern they're supposed to learn.

("Params" are... params? Still feels a *little* circular to me, but at least even without the example, now someone could get closer to interpreting that statement.)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+	 * ex. "/pages/[lang]/[...slug].astro" will output the params ['lang', '...slug']
+	 */
+	params: string[];
+
+  /**
+	 * Output URL pathname where this route will be served
```
sarah11918, line 561: I noticed there's no example here, but there is for most other things. Is there a short helpful one we could put here?

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+	 * note: will be undefined for [dynamic] and [...spread] routes
+	 */
+	pathname?: string;
+
+  /**
+	 * Similar to the "params" field, but with more associated metadata. For example, for `/site/[blog]/[...slug].astro`, the segments are:
```
sarah11918, line 567:
```suggestion
	 * An array of metadata for each segment (separated by `/`) of the route pattern. Includes the `content` of the segment, as well as whether or not the path segment is dynamic and spread. For example, for `/site/[blog]/[...slug].astro`, the segments are:
```
OK, you might need to fix this up but this seemed more related to `pattern` in my mind than params, since params don't care about the static segments.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+	 * 3. `{ content: '...slug', dynamic: true, spread: true }`
+	 */
+	segments: RoutePart[][];
+
+  /**
+	 *
+	 * The type of the route. It can be:
+	 * - `page`: a route that lives in the file system, usually an Astro component
+	 * - `endpoint`: a route that lives in the file system, usually a JS file that exposes endpoints methods
+	 * - `redirect`: a route points to another route that lives in the file system
```
sarah11918, line 580:
```suggestion
	 * - `redirect`: a route that points to another route that lives in the file system
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+	 * - `fallback`: a route that doesn't exist in the file system that needs to be handled with other means, usually the middleware
```
sarah11918, line 581:
```suggestion
	 * - `fallback`: a route that doesn't exist in the file system and is handled with other means, usually middleware
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+	 */
+	type: RouteType;
+
+  /**
+	 * The route to redirect to. It holds information regarding the status code and its destination.
+	 */
+	redirect?: RedirectConfig;
+
+  /**
+	 * Whether the route comes from Astro core, an integration or the user's project
```
sarah11918, line 591:
```suggestion
	 * Whether the route is generated internally by Astro core, externally by an integration or by the user's project
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -809,6 +962,14 @@ interface IntegrationRouteData {
 }
 ```
 
+#### `assets` option
+
+<Since v="5.0.0" />
+
+**Type:** `Map<string, URL[]>`
+
+Contains URLs to output files paths, grouped by [`IntegrationResolvedRoute`](#integrationresolvedroute-type-reference) `pattern` property.
```
sarah11918, line 971:
```suggestion
Contains URLs to output files paths, grouped by the [`IntegrationResolvedRoute`](#integrationresolvedroute-type-reference) `pattern` property.
```
Does this need a "the"?

src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -361,6 +361,46 @@ export default function createIntegration() {
 
 <ReadMore>Learn more about [the Adapter API](/en/reference/adapter-reference/) for building adapter integrations.</ReadMore>
 
+### Deprecated: `routes` on `astro:build:done` hook (Integration API)
+
+<SourcePR number="12329" title="feat(next): astro:routes:resolved"/>
+
+In Astro v4.x, you could only get routes in an integration from the `astro:build:done` hook.
```
sarah11918, line 368:
```suggestion
In Astro v4.x, integrations accessed routes from the `astro:build:done` hook.
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+#### `routes` option
+
+**Type:** [`IntegrationResolvedRoute[]`](#integrationresolvedroute-type-reference)
+
+A list of all routes alongside their associated metadata.
+
+You can reference the full `IntegrationResolvedRoute` type below, but the most common properties are:
+
+- `entrypoint` - the input file path relative to the project root
+- `pathname` - the output file URL (undefined for routes using `[dynamic]` and `[...spread]` params)
```
sarah11918, line 493: I'm not a fan of defining twice. If these two are common, is there instead some kind of example with these two? something like:

The following example accesses the `entrypoint` and `pathname` ... 

And don't define them here? Is there some kind of minimal example usage that would work here?

florian-lefebvre: FYI I just copied and adapted what was in `astro:build:done` so maybe we can also update it there with you think is best!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,143 @@ const integration = {
...
+
+##### `IntegrationResolvedRoute` type reference
+
+```ts
+interface IntegrationResolvedRoute {
+	/**
+	 * The current **pattern** of the route. For example:
```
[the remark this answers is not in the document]
florian-lefebvre: All those jsdoc come directly from core, and are also in `astro:build:done`. So do you prefer not to update the jsdoc on that page at all, or update them in core and then reflect the changes here?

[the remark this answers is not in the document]
sarah11918: Ah, OK. How important is it that these are kept identical to core? This is public facing docs, and one could argue should be written like docs.

(I was even thinking, this is a LOT of documenting going on *inside* a code sample. Why aren't we writing normal docs here?)

In that case, I'm OK to merge exactly what's already matching, but then think about at some point in the future this page becomes an actual docs page, not just copy paste of whatever you've written in code that I have no control over the quality of. 😅

sarah11918 (commented):
@florian-lefebvre 

See my comment about being OK to copy these as-is since this is apparently just a reproduction of what's in code.

BUT, these don't feel like the best *docs* to me. (Why is docs just a copy of code comments and not "documentation"?) We can leave that for some potential future exploration, whether this page needs itself an overhaul! And we don't have to block merging for the feature!

src/content/docs/en/reference/integrations-reference.mdx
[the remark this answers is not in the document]
florian-lefebvre: Yeah I think that's a good idea :+1:

florian-lefebvre:
I think we can still take care of https://github.com/withastro/docs/pull/10077#discussion_r1850759057 in this one tho!

sarah11918:
Sure! Can you propose an example that would go there?

src/content/docs/en/reference/integrations-reference.mdx
florian-lefebvre: Gotta say I'm struggling to find a good simple example, the `routes` related logic is usually quite involved :sweat:. @Fryuni do you have an idea?

florian-lefebvre:
@sarah11918 i tried to add a little example, let me know what you think!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -452,10 +462,153 @@ const integration = {
...
+  routes: IntegrationResolvedRoute[];
+  logger: AstroIntegrationLogger;
+}) => void | Promise<void>;
+```
+
+#### `routes` option
+
+**Type:** [`IntegrationResolvedRoute[]`](#integrationresolvedroute-type-reference)
+
+A list of all routes with their associated metadata.
```
sarah11918, line 488:
```suggestion
A list of all routes with their associated metadata.

Example use:
```
How about a tiny statement that it's an example, then good to go??

sarah11918 (approved):
Just left the tiniest suggestion, but good to go whenever you're happy with it! (And, we can think about whether we want to improve this page overall in future!) :rocket:

PR #10079, opened 2024-11-20
add responsive images as an experimental flag page

sarah11918:
#### Description (required)

Creates a new experimental flag page for responsive images based on what was previously added to configuration-reference

Notably:

- ONLY covers the content from `experimental.resoponsiveImages` and not the `image.experimental*` features. (Is this how we want to do this? Leave those in regular configuration reference?)

- Links should fail until config reference merges with the `image.experimental*` properties. Will all work out tomorrow when I merge after feature release.

### DIFF
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -121,6 +121,7 @@ export const sidebar = [
 			group('reference.experimental', {
 				items: [
 					'reference/experimental-flags',
+					'reference/experimental-flags/responsive-images',
 					'reference/experimental-flags/client-prerender',
 					'reference/experimental-flags/content-intellisense',
 				],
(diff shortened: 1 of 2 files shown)


sarah11918:
Here's what your new page will look like @ascorbic ! Feel free to make this look like what you want before the feature merge tomorrow, and even after the fact we can quickly adjust if and as necessary!

PS You can now see what I mean about the additional config options. I'm open to moving them here, or leaving them in regular config reference. What do you think?

PPS I know there is still content in the public/types file that we'll remove at some point. I'm more concerned about getting this page the way you want it first!

ascorbic (approved):
Looks great. Thanks!

PR #10081, opened 2024-11-21
i18n(ko-KR): update `env-unsupported-get-secret.mdx`

jsparkdev:
#### Description (required)

- update `env-unsupported-get-secret.mdx`
- #10071

- Closes #
- Suggested label:


ArmandPhilippot (approved):
As I don't speak Korean, I first checked by comparing with other deprecated errors. I noticed you didn't use the same wording here and there: https://docs.astro.build/ko/reference/errors/action-query-string-invalid-error/ (for example).
But using a translator, the meaning seems the same so, LGTM!

PR #10095, opened 2024-11-22
one unified on-demand rendering page

sarah11918:
#### Description (required)

Lots of files changed and moving parts, but mostly because of links.

This consolidates the previous `basic/rendering-modes.mdx` and `guides/server-side-rendering.mdx` into one `guides/on-demand-rendering.mdx` page.

The changes to the underlying modes (`static` now allows on-demand rendering with an adapter, no `mode` change required) makes how we discuss this feature very different now, so this PR:

- has one new simplified page to describe on-demand rendering, and how to make that work
- removes in all languages the previous pages
- updates (so many!) translation links, *usually* to just the root of the new page. (The translated pages that link here will need more updating to the text around them, so the translators can choose an appropriate slug to match the English as they update the page. But at least any existing links point to this new page.)
- overwrites the config-ref and error files manually as needed to make link checks pass. These will be updated at the source eventually, but are hardcoded in the English for now to make sure everything works.

Follow up work:

- English config ref (1 tiny link) and (7) error messages will need updating at the source. At that time, I will also evaluate the wording around them, which in some cases will need a refresh as well (e.g. not referring to "server or hybrid mode")


sarah11918:
I AM VERY PROUD OF MYSELF FOR THIS ENTIRE PR AND **ONE** WRONG LINK!

image

yanthomasdev (approved):
LGTM!

PR #10103, opened 2024-11-25
Simplify `HTMLAttributes`' insertions using a regular expression

andreilgeorgescu:
#### Description (required)

The reason for this change is that the insertions were using two pieces of logic to match in [HTMLAttributes](https://docs.astro.build/en/guides/typescript/#built-in-html-attributes) where only one was needed.

A regular expression solved this issue and simplified the code a bit. This regular expression can also be further modified to match future changes more easily.

- Suggested label: [improve documentation](https://github.com/withastro/docs/labels/improve%20documentation)

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -177,7 +177,7 @@ Astro provides the `HTMLAttributes` type to check that your markup is using vali
 
 For example, if you were building a `<Link>` component, you could do the following to mirror the default HTML attributes for `<a>` tags in your component’s prop types.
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi! Thank you for taking the time to contribute a PR!

Unfortunately we are not accepting any changes to the v4 version of our docs at the moment, since we are very close to releasing v5 docs and have stopped contributions to this branch and are preparing it for long-term, unmaintained mode.

Many of our v5 docs are significantly different from the v4 version, so after release, if you would still like to propose this change, then we invite you to resubmit a new PR after v5 docs have been published if this change is still needed/relevant!

So I will close *this* PR, but do keep an eye out for v5 docs and I hope you'll consider contributing then to help make us better! 🥳

PR #10106, opened 2024-11-25
docs: Add information about dynamic static split server islands

Jacob-Roberts:
#### Description (required)

This PR is the documentation follow-on to [this PR](https://github.com/withastro/astro/pull/12320) which adds the ability to serve the server islands from a separate domain.

jake_roberts

### DIFF
--- a/src/content/docs/en/reference/integrations-reference.mdx
+++ b/src/content/docs/en/reference/integrations-reference.mdx
@@ -157,6 +157,8 @@ export default {
 }
 ```
 
+<p><Since v="5.x" /></p>
+
 This hook gives you the opportunity to add private config options that are exposed only to integrations.
 There are certain options that are not directly exposed to users, but are allowed to be written from integrations.
 
--- a/src/content/docs/en/reference/integrations-reference.mdx
+++ b/src/content/docs/en/reference/integrations-reference.mdx
@@ -157,6 +157,12 @@ export default {
 }
 ```
 
+This hook gives you the opportunity to add private config options that are exposed only to integrations.
+There are certain options that are not directly exposed to users, but are allowed to be written from integrations.
+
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -157,6 +157,14 @@ export default {
 }
 ```
 
+<p><Since v="5.x" /></p>
+
+This hook gives you the opportunity to add private config options that are exposed only to integrations.
+There are certain options that are not directly exposed to users, but are allowed to be written from integrations.
+
+For example, the astro config [`serverIslandDynamicBase`](https://github.com/withastro/astro/blob/30d720f519f991720ee0b7c328a4bf4bbf31760d/packages/astro/src/types/public/config.ts#L1979-L1986)
```
sarah11918, line 165: Heads up that we do not typically link to code outside of docs. Everything in docs should be fully explained so people don't have to leave the page. A full example would have to be provided here.

sarah11918 (commented):
Hi @Jacob-Roberts , thanks for submitting documentation for your feature! 🚀 

I took a look at the initial PR, and I haven't seen the code approved yet, and docs can never be edited before we know that the feature is complete and approved anyway, in case some details or implementation change during the code review process. So just letting you know this will have to sit and wait on the feature being accepted. (And, it's possible being so close to a v5 release that there may not yet be another beta release, in which case I don't know whether devs would accept a new feature without being able to at least release it in a beta first? Just a heads up on that, too! This is cutting it SUPER close for comfort! But, I'll wait and see what happens there!)

In the mean time, I've left a comment about needing an example though!

sarah11918 (changes requested):
Just putting a block on this PR so it doesn't accidentally get merged as this feature will not be released before v5.

sarah11918:
Ooh, it's letting me reopen... let me try this

Jacob-Roberts:
Thanks @sarah11918! I'll submit a new PR

PR #10111, opened 2024-11-27
Update Lunaria version and components

yanthomasdev:
#### Description (required)

Previously this PR added scripts to find outdated content. We decided not to use it but the other changes are still valid.

I took the opportunity to update Lunaria to the latest prerelease version (which fixes a few mistrackings when using tracker directives) and fixed a small dashboard paragraph with accidental quotes.

- Closes #
- Suggested label:


trueberryless (approved):
LGTM 🎉 

I have tested running the script locally and I have to say it took much longer than expected, but it works eventually! Remove option also works fine!

Also, it feels rough to be the translator for a language that's kinda outdated:

image

These are great additions if you ask me. Makes removing very old pages a lot easier... Good job!

sarah11918 (approved):
Yay! Updated Lunaria!

PR #10117, opened 2024-11-28
Correct location for creating Decap CMS admin HTML page

dasShounak:
#### Description (required)

The admin dashboard won't work if the Decap CMS HTML file is created within the public folder. Astro won't generate the route. Instead it should be created only in `src/pages` directory. Using `.html` extension will allow the use of the `window.CMS` object, which is useful for customizing the admin page.

Yes


sarah11918 (changes requested):
Hi @dasShounak , thank you for contributing a fix to docs! 🚀 

Unfortunately, we are on a code freeze for v4 Astro docs and are putting this branch into long-term, unmaintained status. So, that just means this PR will have to hang around for a little while until our v5 docs are published and we are accepting community contributions again!

Sorry for the inconvenience, and I'm putting a block on this PR so that it doesn't get accidentally merged before v5 docs are published. Please also understand that there may be conflicts at that time we have to resolve as the docs will be significantly updated. But, we'll find that out next week sometime! 😅 

Have a great weekend, and we'll see you in v5! 👋

sarah11918 (approved):
When checks pass, good to go! (Got a deploy preview on Netlify to rerun and work.)

sarah11918:
Got this deploy to work on Netlify, assuming it should be fine!

PR #10118, opened 2024-11-29
Fix fallback language remark plugin

HiDeoo:
#### Description (required)

This PR fixes the `remarkFallbackLang` remark plugin used to marks links to fallback language pages.

Repro:

1. Visit https://docs.astro.build/de/guides/cms/#kann-ich-astro-ohne-cms-nutzen
1. The `Inhalte zu erstellen` link should be marked and displayed as `Inhalte zu erstellen (EN)` as it's linking to a fallback language page in English.
   The `Integration` link in the previous section is correctly not marked as it's linking to a page in German.

This plugin broke with Astro `4.0.0-beta.0` due to to all the updates to the unified, remark and rehype dependencies in https://github.com/withastro/astro/pull/9138. It looks like the issue is located either in `hast-util-raw` or the `parse5` tokenizer (used internally) and it requires a raw node to either include an element or a whitespace (e.g. `&nbsp; (EN)` or `<span>&nbsp;(EN)</span>` instead of `&nbsp;(EN)` would work).

This PR workaround the issue by switching to a text node using the non-breaking space unicode character `U+00A0` instead of the HTML entity `&nbsp;` and I'll open later a proper issue on the related repo.

- Closes #
- Suggested label: bug

### DIFF
--- a/plugins/remark-fallback-lang.ts
+++ b/plugins/remark-fallback-lang.ts
@@ -30,8 +30,8 @@ export function remarkFallbackLang(): Plugin<[], Root> {
 			if (linkSourceFileName) return;
 
 			link.children.push({
-				type: 'html',
-				value: `&nbsp;(EN)`,
+				type: 'text',
+				value: `\u00A0(EN)`,
 			});
 		});
 	};


sarah11918 (approved):
Can confirm it's now working in the deploy preview! :partying_face:

delucis (approved):
Thank you a) for noticing this was broken, b) for fixing it, and c) for trekking through the entire unified.js cinematic universe on a quest for the infinity token 🙌

PR #10129, opened 2024-12-02
Add Google IDX as new official sponsor in left sidebar of docs

clearlyTHUYDOAN:
## Description

- Updates leftside bar navigation with Google Project IDX logo as our new Official Online Editor Partner


src/components/LeftSidebar/Sponsors.astro
```diff
@@ -24,6 +24,19 @@
 				fill="currentColor"></path>
 		</svg>
 	</a>
+	<a
+		href="https://idx.dev/?utm_source=astro&utm_medium=astro&utm_campaign=astro"
+		aria-label="Google IDX"
+	>
+	<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 214 52">
```
yanthomasdev, line 31: I believe this should fix the rendering and as an extra, improve accessibility.
```suggestion
	<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 214 52"  width="96" aria-hidden="true">
```

delucis, line 31: Needs a `width` attribute to size properly and the `aria-hiddent="true"` hides the SVG from screen readers (which get their label from the `aria-label` on the link element instead.

```suggestion
	<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 214 52" width="125" aria-hidden="true">
```

delucis (changes requested):
Close! Just needs a tiny tweak.

src/components/LeftSidebar/Sponsors.astro
```diff
@@ -24,6 +24,19 @@
 				fill="currentColor"></path>
 		</svg>
 	</a>
+	<a
+		href="https://idx.dev/?utm_source=astro&utm_medium=astro&utm_campaign=astro"
+		aria-label="Google IDX"
```
delucis, line 29: Oh, one thing I missed. The label here probably should be “Project IDX” to actually match the text in the logo, otherwise it would be a bit odd:
```suggestion
		aria-label="Project IDX"
```

delucis (approved):
LGTM @clearlyTHUYDOAN! Just noticed one detail, but I’ll go ahead and commit it myself.

PR #10136, opened 2024-12-03
Fix path typo on content collection reference page

pekochan069:
#### Description (required)

Fixing typo on content collection reference page
`src/content.config.ts` to `src/content/config.ts`

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/modules/astro-content.mdx
+++ b/src/content/docs/en/reference/modules/astro-content.mdx
@@ -34,9 +34,9 @@ import {
 <Since v="2.0.0" />
 </p>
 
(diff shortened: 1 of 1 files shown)


ArmandPhilippot:
Hi @pekochan069, thanks for taking the time to fix this. However, this is not a typo so we can't merge this PR!

Since v5 **released today**, `content.config.ts` is the new config file format used to define your collections! See: https://docs.astro.build/en/guides/upgrade-to/v5/#breaking-changes-to-legacy-content-and-data-collections

If you're still on v4, you can look for the v4 doc here: https://v4.docs.astro.build/en/reference/modules/astro-content/

PR #10142, opened 2024-12-03
Add the fact that the return value of getEntry is always nullable to upgrade guide

ascorbic:
#### Description (required)

One of the side effectys of the way we generate types with the content layer is that the return type for `getEntry` always includes `undefined`. This can cause type errors if you pass a static string key and assume the result is always defined. This PR adds this to the upgrade guide.

- Closes #
- Suggested label:


sarah11918 (approved):
Looks great, thanks!

PR #10169, opened 2024-12-04
Docs: dynamically-importing-images / Add uppercase glob pattern

tordans:
The glob pattern used in this guide is too restrictive when it comes to uppercase file endings.

Related: https://github.com/withastro/astro/issues/12621

### DIFF
--- a/src/content/docs/en/recipes/dynamically-importing-images.mdx
+++ b/src/content/docs/en/recipes/dynamically-importing-images.mdx
@@ -116,7 +116,7 @@ In this recipe, you will learn how to dynamically import your images using Vite'
     
     const { imagePath, altText, name, age } = Astro.props;
     const images = import.meta.glob<{ default: ImageMetadata }>('/src/assets/*.{jpeg,jpg,png,gif}');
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi! There could in fact be even more file extensions (e.g. `.webp`) and it's not up to a recipe to list every possible file extension. In step 2, only lower case extensions are shown, and it's a reasonable "up to the reader" to match the query to what is in your own folder.

Making the glob pattern longer isn't additionally helpful in this case, and I think adds more noise to the example. So I'm happy with the example as it is! Amending a recipe to suit one's own project is a reasonable expectation, and the more concise examples are, the more visible the ideas are that it's trying to communicate.

tordans:
I could prepare a PR to change it it to `jpeg,jpg,JPEG,JPG` which would be shorter but still explain the fact that glob is case sensitive. It would also signal more clearly, that the list in the example is not complete but just a suggestion. This would solve two issues I had with the guide…

- I assumed the list was "complete"
- I was not aware that glob is case sensitive

PR #10170, opened 2024-12-04
Add `--typescript` flag removal to v5 upgrade guide

HiDeoo:
#### Description (required)

As discussed on Discord, this PR adds a new section to the breaking changes of the v5 upgrade guide to document the removal of the `--typescript` flag from `create-astro`.

- Closes #
- Suggested label: 5.0

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v5.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v5.mdx
@@ -1164,9 +1164,9 @@ export default {
 
 <SourcePR number="12083" title="create-astro updates"/>
 
-The experience of creating a new Astro project using the `create astro` CLI command has been reworked.
-As Astro is TypeScript-only, the TypeScript question and its associated `--typescript` flag were often misleading and have been removed.
-The "Strict" preset is now the default, but it can still be changed manually in `tsconfig.json`.
+In Astro v4.x, it was possible to choose between Astro's three TypeScript settings when creating a new project using `create astro`, either by answering a question or by passing an associated `--typescript` flag with the desired TypeScript setting. 
+
+Astro 5.0 updates the `create astro` CLI command to remove the TypeScript question and its associated `--typescript` flag. The "strict" preset is now the default for all new projects created with the command line and it is no longer possible to customize this at that time. However, the TypeScript template can still be changed manually in `tsconfig.json`.
 
 #### What should I do?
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -1160,6 +1160,41 @@ export default {
 
 <ReadMore>Read more about [developing a dev toolbar app for Astro using the Dev Toolbar API](/en/reference/dev-toolbar-app-reference/).</ReadMore>
 
+### Removed: `create-astro` TypeScript question
+
+<SourcePR number="12083" title="create-astro updates"/>
+
+The experience of creating a new Astro project using the `create astro` CLI command has been reworked.
+As Astro is TypeScript-only, the TypeScript question and its associated `--typescript` flag were often misleading and have been removed.
+The "Strict" preset is now the default, but it can still be changed manually in `tsconfig.json`.
```
sarah11918, line 1169:
```suggestion
In Astro v4.x, it was possible to choose between Astro's three TypeScript settings when creating a new project using `create astro`, either by answering a question or by passing an associated `--typescript` flag with the desired TypeScript setting. 

Astro 5.0 updates the `create astro` CLI command to remove the TypeScript question and its associated `--typescript` flag. The "strict" preset is now the default for all new projects created with the command line and it is no longer possible to customize this at that time. However, the TypeScript template can still be changed manually in `tsconfig.json`.
```

sarah11918 (commented):
Thanks for jumping on this @HiDeoo ! See what you think of below.

src/content/docs/en/guides/upgrade-to/v5.mdx
HiDeoo: I was a bit hesitant on using the same pattern from other points:

- This is not strictly related to Astro 4 or Astro 5, this was released at the same time mostly
- This change was released in a minor of the `create-astro` package, not a major even tho it is a breaking change
- You can still install Astro 4 using the latest minor of `create-app` including this change

But maybe it's fine as it was released at the same time as Astro 5 and it's me nitpicking?

sarah11918: Yeah, I think it just confuses people more if it's NOT the same wording, maybe? The only people reading this guide are the people upgrading to v5 so I *think* it's most helpful just to keep it consistent? It also hides the fact that maybe this should have been included with v5? :smile: 

Could also....

"Astro 5.0 uses the updated `create astro` CLI command from v4.16 which removes..." if you feel strongly!

HiDeoo: You convinced me :thumbsup: I was mostly making sure the entire scope of what made me doubt was identified and we were on the same page.

Thanks for the great review :raised_hands:

sarah11918 (approved):

PR #10190, opened 2024-12-04
Added tip for Actions

hkbertoson:
#### Description (required)

Added a tip for Actions regarding how data is sent back.

- Closes #
- Suggested label:  improve documentation


src/content/docs/en/guides/actions.mdx
```diff
@@ -18,6 +18,15 @@ Use actions instead of API endpoints for seamless communication between your cli
 - Generate type-safe functions to call your backend from the client and even [from HTML form actions](#call-actions-from-an-html-form-action). No need for manual `fetch()` calls.
 - Standardize backend errors with the [`ActionError`](/en/reference/modules/astro-actions/#actionerror) object.
 
+
+<Aside type='tip'>
+Actions don’t just send back simple JSON. Instead, they return a special kind of data package that can handle things like dates, maps, sets, and URLs. This is made possible using a tool called the **devalue library**. However, because of this special format, you can't easily inspect the response from the network like you might with regular JSON.  
+
+If you want to understand more about how this library works, you can check it out [Devalue](https://github.com/Rich-Harris/devalue).  
+
+For details about the `handler()` property mentioned, take a look at [Handler Property](/en/reference/modules/astro-actions/#handler-property).
```
bholmesdev, line 27:
```suggestion
Actions return a custom data format that can handle Dates, Maps, Sets, and URLs [using the Devalue library.](https://github.com/Rich-Harris/devalue)

Because of this special format, you can't easily inspect the response from the network like you might with regular JSON. We recommend inspecting the `data` object returned by actions for debugging.

[See the `handler()` API reference](/en/reference/modules/astro-actions/#handler-property) for full details.
```

Reworded to clarify what users _should_ do when debugging. Also made the wording 10% more boring to quote @sarah11918 :)

I agree this isn't ideal, and I can try investigating other ways to give a preview of payloads in your network tab. May be a different `content-type` or a JSON object wrapper for the value.

bholmesdev (commented):
Some thoughts on wording. Thanks for taking `action` on this!

hkbertoson:
Got it updated!

src/content/docs/en/guides/actions.mdx
```diff
@@ -168,6 +168,10 @@ Now, all of your user actions are callable from the `actions.user` object:
 
 Actions return an object containing either `data` with the type-safe return value of your `handler()`, or an `error` with any backend errors. Errors may come from validation errors on the `input` property or thrown errors within the `handler()`.
 
+Actions return a custom data format that can handle Dates, Maps, Sets, and URLs [using the Devalue library.](https://github.com/Rich-Harris/devalue). Therefore, you can't easily inspect the response from the network like you can with regular JSON. For debugging, you can instead inspect the `data` object returned by actions.
```
sarah11918, line 171:
```suggestion
Actions return a custom data format that can handle Dates, Maps, Sets, and URLs [using the Devalue library](https://github.com/Rich-Harris/devalue). Therefore, you can't easily inspect the response from the network like you can with regular JSON. For debugging, you can instead inspect the `data` object returned by actions.
```

src/content/docs/en/guides/actions.mdx
```diff
@@ -703,3 +707,5 @@ if (searchQuery) {
 }
 ---
 ```
+
+
```
sarah11918, line 711:
```suggestion
```

sarah11918 (commented):
Thank you for submitting this helpful update @hkbertoson ! 🎉 

See what you think of the new placement of the content I'm suggesting here.

Notably:
- it was previously at the VERY BEGINNING, the FIRST THING SOMEONE READ, even before how to use an action. I don't think you meant for this to be of such supreme importance that someone cannot continue without knowing that info first? That is why I've moved it lower.
- Whenever new content is "how something works", it's typically *not* a note, tip, caution etc. It's part of the description/explanation. We typically use a "note" for true and helpful, related but maybe *slightly off the immediate topic* information that might trip you up if you don't know it. (But, this isn't really the place to document it.) A tip is a non-essential but helpful **thing you could do** (an action). In some worlds, this is a tip, since you're suggesting inspecting the data object. But given that the alternative is described as "difficult" (and not the happy path), this feels less "optional" and more like "how it should be done." To me, that's regular paragraph text and not a "tip you might want to consider."
-  The biggest task for incorporating new content into existing docs is to find out where it should *really* go. In this case, since it provides more context for the data object returned, and provides a link to the `handler()` API reference for more details, it seemed like it fit in the section about "handling returned data." Do you think this is the right place for it? Is there somewhere better?

So see what you think of my suggestion here, whether it's *accurate* (you're the subject matter expert here!), helpful, and seems to fit in this section. Make any suggestions you need to if it's not quite right yet! 🙌

hkbertoson:
I like the change. I think it does make sense where its at and fits the "flow" better.

sarah11918 (approved):
Great stuff! Let's get this into docs!

PR #10193, opened 2024-12-04
fix: remove experimental and update comments in `cli-reference.mdx`

ArmandPhilippot:
#### Description (required)

Adds two changes to `cli-reference.mdx`:
* the comment above each command was still using "`start` script" instead of `dev` so I updated the comments
* Astro Env was still marked as experimental in the `astro sync` section so I removed that mention


sarah11918 (approved):
Again, perfect, and so appreciated! :purple_heart:

PR #10206, opened 2024-12-05
Cleanup docgen workflows post-v5 release

delucis:
This PR removes the v5-beta workflows from our docgen CI scripts. These were running nightly to update our beta docs during the prerelease period, but are no longer needed (and fail because our v5 branches no longer exist).

I also updated a line logged in the config reference generation script that hadn’t been updated to match the new file location which we changed as part of v5.


sarah11918 (approved):
Looks perfect to me, thank you!

PR #10209, opened 2024-12-05
fix: remove useless escape characters in `pattern`

ArmandPhilippot:
#### Description (required)

To avoid confusion whether the editor automatically removes the character or ESlint displays a warning, I removed the useless escape character in the loader `pattern`.

#### Related issues & labels (optional)

- Related:
  * #10152
  * https://discord.com/channels/830184174198718474/853350631389265940/1314275896038785227
- Suggested label: code snippet update

PR #10213, opened 2024-12-05
fix: remove useless escape character in `content-collections.mdx`

ArmandPhilippot:
#### Description (required)

Same as #10209 but with a different pattern:
I haven't noticed this pattern in the previous PR because I was looking for an exact matching... but I checked `"**\/*.md"` and `"**/*.md"` are both valid.
So, to avoid confusion whether the editor automatically removes the character or ESlint displays a warning, I removed the useless escape character in some loader patterns in `content-collections.mdx`.


sarah11918 (approved):
Perfect, thank you!

yanthomasdev (approved):
Nice catch, thanks!

PR #10246, opened 2024-12-06
Fix call of render function in doc example

ancaemcken:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/markdown-content.mdx
+++ b/src/content/docs/en/guides/markdown-content.mdx
@@ -116,7 +116,7 @@ import {Content as PromoBanner} from '../components/promoBanner.md';
 import { getEntry, render } from 'astro:content';
 
 const product = await getEntry('products', 'shirt');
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -116,7 +116,7 @@ import {Content as PromoBanner} from '../components/promoBanner.md';
 import { getEntry, render } from 'astro:content';
 
 const product = await getEntry('products', 'shirt');
-const { Content } = await render();
+const { Content } = await product.render();
```
sarah11918, line 119:
```suggestion
const { Content } = await render(product);
```

sarah11918 (approved):
Thank you for taking the time to improve the docs! I just corrected the code sample (content collections are a little different now in v5, and your proposal would have been correct for v4), but they were absolutely missing `product`.

Welcome to Team Docs! 🥳

PR #10255, opened 2024-12-06
feat(link): update link in `guides/deploy/aws` because it now links to same page [i18nIgnore]

trueberryless:
#### Description (required)

The link `/en/guides/deploy/aws/#s3-static-website-hosting` can be shortened to just `#s3-static-website-hosting` because the heading is on the same page.

- Suggested label: link?


yanthomasdev (approved):
LGTM! Thanks!

PR #10277, opened 2024-12-08
Replace old <ViewTransition /> name with new one

bartosztrusinski:
#### Description

Replace old `<ViewTransition />` with new router name `<ClientRouter />`

discord username is trsx


yanthomasdev (approved):
Nice catch, thanks!

PR #10310, opened 2024-12-09
fix: remove unnecessary `await` before `import.meta.glob`

kidonng:
#### Description (required)

When used with `{ eager: true }`, [`import.meta.glob` no longer returns a promise](https://github.com/vitejs/vite/blob/ef7a6a35e6827b92445e5a0c2c0022616efc80dd/packages/vite/types/importGlob.d.ts#L66-L74) and thus no longer needs `await`.

- Closes #
- Suggested label:


TheOtterlord (approved):
Thanks for catching this!

PR #10315, opened 2024-12-09
Additional import.meta.glob fixes

TheOtterlord:
Fixes 2 more instances of `import.meta.glob`

### DIFF
--- a/src/content/docs/en/guides/imports.mdx
+++ b/src/content/docs/en/guides/imports.mdx
@@ -287,12 +287,13 @@ You can optionally provide a type for the `frontmatter` variable using a TypeScr
 
 ```astro
 ---
+import type { MarkdownInstance } from 'astro';
 interface Frontmatter {
     title: string;
     description?: string;
 }
 
-const posts = import.meta.glob<MarkdownInstance<Frontmatter>>('./posts/**/*.md');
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/imports.mdx
```diff
@@ -323,7 +324,7 @@ Other files may have various different interfaces, but `import.meta.glob()` acce
 interface CustomDataFile {
   default: Record<string, any>;
 }
-const data = await import.meta.glob<CustomDataFile>('../data/**/*.js');
+const data = import.meta.glob<CustomDataFile>('../data/**/*.js');
```
sarah11918, line 327: Just checking, normally we've been removing `await` when there is eager loading!

sarah11918 (approved):
Thank you for the fix!

src/content/docs/en/guides/imports.mdx
TheOtterlord: `await` is removed because `import.meta.glob` never returns a promise, but it returns a record of functions that return promises when called. There's no usage example, and we can't showcase this easily, but whether someone needs `eager: true` will probably come down to what they're importing and why. The Vite docs cover what eager does, and we link to that above.

yanthomasdev (approved):

PR #10336, opened 2024-12-11
Fix abbrevation duplication of GC project [i18nIgnore]

trueberryless:
#### Description (required)

I'm not sure if the P in `GCP` stands for project. If yes, then the word project is a duplicate, because if you don't use abbrevations it reads like: `Google Cloud Project project`

- Closes #
- Suggested label: ~duplication 🤣~


TheOtterlord:
GCP is "Google Cloud Platform", so it stands for "Google Cloud Platform project". In cases where it is redundant, I think you usually remove the part detached from the acronym, so the acronym is still recognisable.

trueberryless:
Thank you very much for the clarification! Didn't know what the P stands for...

PR #10338, opened 2024-12-11
Add dockerhub URL to registry examples in Google Cloud Deploy Guide

trueberryless:
#### Description (required)

Since Docker Hub is mentioned in the text above, I think it would make sense to include the Registry URL in the examples below...

- Closes #
- Suggested label:


sarah11918 (approved):
Sounds good!

PR #10332, opened 2024-12-10
[i18nIgnore] remove `await` from `import.meta.glob()`

sarah11918:
Changes all occurrences in all languages of `await import.meta.glob('*', { eager: true });` to remove `await` as it is not needed.

PR #10342, opened 2024-12-11
[I18nIgnore] fix(tutorial): remove a useless `await` in unit 5

ArmandPhilippot:
#### Description (required)

We have a remaining `await` in front of `Object.values` which is useless and makes the code snippets inconsistent across the tutorial, so I removed it.

- Related: #10332


sarah11918 (approved):
I swear they're multiplying behind our backs!

PR #9240, opened 2024-08-29
5.0.0 beta docs

delucis:
#### Description (required)

This is the WIP branch for the documentation changes for Astro v5, soon to be in beta.

Documentation may be incomplete, but feel free to use the deploy preview if you are trying out an early release of Astro v5: https://5-0-0-beta.docs.astro.build/

#### To do before merging

- [x] Remove `/` redirect added in 8b2b3ebac42b9c486966a56a1af20bd234e5de70
- [x] Remove v5 beta badge in site header added in 8e56929ec70294a010a3b943ee522721f15b3ccd
- [ ] Update [the docgen script](https://github.com/withastro/docs/blob/5.0.0-beta/scripts/docgen.mjs) to point to the new location for configuration types in the Astro repo in frontmatter etc. And remove beta docgen workflow.
- [x] Revert `editLink` change made in b3dea25cfa96604727cf047fe51bb2f4f2dc6e39.
- [x] Disable Pagefind and enable the `Search` component override changed in  ca7d812cee0fb28611099c665f77e0e1ad91a3bc

PR #10350, opened 2024-12-12
i18n(fr): update all the `recipes`

ArmandPhilippot:
#### Description (required)

Updates the French translations of the `recipes` with changes from #9240

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/recipes/bun.mdx
+++ b/src/content/docs/fr/recipes/bun.mdx
@@ -90,7 +90,7 @@ bunx --bun astro preview
 
 ## Ajouter SSR avec Bun
 
-Puisque Bun offre la [compatibilité API Node.js](https://bun.sh/docs/runtime/nodejs-apis), vous pouvez utiliser n'importe quel adaptateur Astro pour [rendre à la demande](/fr/guides/on-demand-rendering/) votre projet Astro :
+Puisque Bun offre la [compatibilité API Node.js](https://bun.sh/docs/runtime/nodejs-apis), vous pouvez utiliser n'importe quel adaptateur Astro pour [un rendu à la demande](/fr/guides/on-demand-rendering/) de votre projet Astro :
 
 ```bash
 bunx astro add vercel
--- a/src/content/docs/fr/recipes/bun.mdx
+++ b/src/content/docs/fr/recipes/bun.mdx
@@ -26,7 +26,7 @@ bunx create-astro@latest my-astro-project-using-bun
 ```
 
 :::tip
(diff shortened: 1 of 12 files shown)


src/content/docs/fr/recipes/bun.mdx
```diff
@@ -90,7 +90,7 @@ bunx --bun astro preview
 
 ## Ajouter SSR avec Bun
 
-Puisque Bun offre la [compatibilité API Node.js](https://bun.sh/docs/runtime/nodejs-apis), vous pouvez utiliser n'importe quel adaptateur Astro pour le [rendu côté serveur](/fr/guides/on-demand-rendering/) dans votre projet Astro :
+Puisque Bun offre la [compatibilité API Node.js](https://bun.sh/docs/runtime/nodejs-apis), vous pouvez utiliser n'importe quel adaptateur Astro pour [rendre à la demande](/fr/guides/on-demand-rendering/) votre projet Astro :
```
thomasbnt, line 93:
```suggestion
Puisque Bun offre la [compatibilité API Node.js](https://bun.sh/docs/runtime/nodejs-apis), vous pouvez utiliser n'importe quel adaptateur Astro pour [un rendu à la demande](/fr/guides/on-demand-rendering/) votre projet Astro :
```

thomasbnt: Maybe it's better? (Optionnal)

ArmandPhilippot: For me it's the same thing, so if you prefer that's fine with me. However, you need to update your suggestion: we need `de` before `votre projet Astro` so that it remains French. 😉

thomasbnt: Ah yes, if you want commit the suggestion. For the rest it's good for me. About https://github.com/withastro/docs/pull/10350#discussion_r1882010893, you will create another PR when `api-reference` is edited?

ArmandPhilippot: I don't think this is useful... It is the new link. But maybe we misunderstood each other. We have two options:
1. do not update the link, and remember to update it with `api-reference`... Since this is a code snippet and not a regular link, I'd say that we risk easily forgetting it...
2. update the link even if `#site` does not exist yet (the link is valid, but the user arrives at the top of the page instead of the correct section) and since this is not a regular link, it's not a problem with the checks.

For simplicity, I chose the second option. Just wanted to leave a note in case you check the link.

ArmandPhilippot: Ah, maybe it was the comment (`//`) that made you wonder. It's a code snippet, so the comment is expected here. No need to remove it later.

PR #10352, opened 2024-12-12
Add missing code fence

louisescher:
#### Description (required)

This PR fixes a code block which was missing the code fence, resulting in broken code highlights on the docs (see https://docs.astro.build/en/reference/routing-reference/#paginate)


yanthomasdev (approved):
Nice catch, thanks!

PR #10353, opened 2024-12-12
feat: getActionPath()

florian-lefebvre:
#### Description (required)

Documents a new `getActionPath()` helper

- Closes #
- Suggested label:


src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -301,3 +301,27 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+
+### `getActionPath()`
+
+<p>
+
+**Type:** `(action: ActionClient<any, any, any>) => string`
+<Since v="5.1.0" />
+</p>
+
+The `getActionPath()` utility is used to get the path to an action, taking base and trailing slash settings into account. This is useful when you want to call an action for more advanced usages, such as providing custom headers or not using the fetch API (eg. `navigator.sendBeacon`):
```
sarah11918, line 313: > used to get the path to an action,

What is "the path to an action"? I don't think we use this phrase/concept on this page elsewhere? Is this a file? A URL? And, we have to try really hard not to use the exact name as the definition 😅  ("Use `getActionPath()` to get the path of the action...") The better the variable name, the harder it is to write the description!) I'm also not sure what the commented part is showing. Is that the value returned from this `getActionPath()`? 

And just checking, is there a reason we're defining it as the variable and then using it in the `fetch` instead of e.g. `fetch(getActionPath(like.action),...`? Would this be common use? (It just puts one level of separation between what it is and where it's used this way, so I want to check that it might not be more helpful to see it as the argument in fetch directly.)

Also, I'm not sure how we're deciding "more advanced use"... is there a better way to describe when this is useful? It allows you to access something that just using `actions.like` doesn't? It allows you to pass options? Can we be specific about why this is useful?

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -301,3 +301,27 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+
+```ts
```
sarah11918, line 315: Can we have a "this example shows..." and then describe what is happening here?

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -301,3 +301,27 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+import { actions, getActionPath } from 'astro:actions'
+
+const path = getActionPath(actions.like) // '/_actions/like'
```
sarah11918, line 318: I'm not sure what the commented part is representing. Is that what is returned by `getActionPath(actions.like)`?

How is the example different using `getActionPath(actions.like)` vs just `actions.like`. You mention that often you can just use the action. What is happening here that we're benefiting from using `getActionPath()` or that wouldn't otherwise be possible?

sarah11918 (commented):
Thanks @florian-lefebvre ! I left some questions below to help tease out exactly what's happening.

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre: > What is "the path to an action"?

I guess technnically it's more "the pathname to the actions endpoint". For example, calling `actions.like` will under the hood make a fetch request to `/_actions/like`. This allows to get this pathname

> And just checking, is there a reason we're defining it as the variable

No it's not needed, we can shorten the example here. It was practical to show the output of the helper

> Also, I'm not sure how we're deciding "more advanced use"... is there a better way to describe when this is useful? It allows you to access something that just using actions.like doesn't?

I thought the 2 examples would be enough but let me detail them. Basically, calling `actions.like` is perfectly fine in most cases. But you can't:
- Pass custom headers (eg. for auth)
- Pass custom options (eg. `keepalive` from fetch)
- Call an action easily without using the fetch API (eg the beacon example)

---

Hope that helps?

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre: > I'm not sure what the commented part is representing. Is that what is returned by getActionPath(actions.like)?

Yes!

> How is the example different using getActionPath(actions.like) vs just actions.like. You mention that often you can just use the action. What is happening here that we're benefiting from using getActionPath() or that wouldn't otherwise be possible?

See the `keepalive` option. I guess it's not clear enough tho, so i'll probably use an example with `sendBeacon` instead

sarah11918 (commented):
ugh, so sorry I didn't realize these comments were pending!

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre:
```suggestion
The `getActionPath()` utility accepts an action and returns its URL pathname, taking base and trailing slash settings into account. This is useful when you want to call an action for more advanced usages, such as providing custom headers or not using the fetch API (eg. `navigator.sendBeacon`):
```

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre: Ha I see (I think)!
> How is this example different using getActionPath than not using it?
Calling an action directly does the following under the hood:
```js
await fetch(someInternalPathLogic, {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
  },
})
```
This is internal, which means you can't pass custom headers (eg. you want to protect your actions with a auth header) nor custom options (eg. keepalive which prevents the request from being killed).

Does that help?

florian-lefebvre (commented):
Gotta say I'm struggling to answer correctly your reviews :smile:

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: That is very helpful, thank you!

sarah11918:
> Gotta say I'm struggling to answer correctly your reviews 😄

Then they must be the right reviews! :sweat_smile:

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre: Is it worth showing 2 examples? Eg. one using the fetch API and one using the `navigator.sendBeacon` API

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: Yup, this is closer to what I was thinking of! 

I think also what I'm getting at is no where do we ever talk about this URL path when we describe how to use an action. People might not even realize there is a URL path in the first place! Normally, they are just "calling actions" that they have defined. So, if that's all you know, where does a "URL path" even come into play in that mental model?

What *I* am assuming is that Astro puts your defined actions somewhere accessible to call them, and this happens to be at a URL path, and this call is made quietly under the hood? (I could even be wrong.) If that's the case, then saying, "Hey, you can now access this URL path that you have no idea exists!" isn't the most helpful definition to the user.

So, something in the style of:

> The `getActionPath()` utility accepts an action and returns a URL path so you can execute an action call as a `fetch()` operation directly. This allows you to provide details such as custom headers when you call your action.

Again, making stuff up in case it's remotely true, but is there something we can say like this that is true? I think a description of this form is helpful

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -301,3 +301,42 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+
+<p>
+
+**Type:** `(action: ActionClient<any, any, any>) => string`
+<Since v="5.1.0" />
+</p>
+
+The `getActionPath()` utility accepts an action and returns a URL path so you can execute an action call as a `fetch()` operation directly. This allows you to provide details such as custom headers when you call your action.
+
+This example shows how to call an action with the `Authorization` header and the [`keepalive`](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive) option:
```
sarah11918, line 315:
```suggestion
This example shows how to call a defined `like` action passing the `Authorization` header and the [`keepalive`](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive) option:
```

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -301,3 +301,42 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+
+```ts {7,9}
```
sarah11918, line 317: Can we have a file title here? And, what do you think should be highlighted in the code sample? maybe just  `getActionPath(actions.like)` or would you do more?

sarah11918:
So last thing worth mentioning is we no longer mention taking base and trailing slash into account, but I'm wondering whether we have to? That could be "true but not particularly helpful to know" in this context? I mean, you get "the URL path you need for your project" and maybe that's enough? Thoughts?

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre: I'm not sure a title is relevant here because there's no context in this example. Should we add more like with a button event listener etc?

About highlighting, I'm unsure what we should emphasize. The use of  `getActionPath()`? The custom things it allows to provide?

sarah11918: Well, let's start with which file would you literally write this code in?

florian-lefebvre: mmh I think it's more likely to happen server side, so for example in a .astro component script (not frontmatter)

florian-lefebvre:
> last thing worth mentioning is we no longer mention taking base and trailing slash into account, but I'm wondering whether we have to? That could be "true but not particularly helpful to know" in this context? I mean, you get "the URL path you need for your project" and maybe that's enough? Thoughts?

"The URL path you need for your project" sounds more useful to me, people usually expect trailing slash and base to be taken into account

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: OK, then let's give these file names of `src/components/my-component.astro` and use the code fences at top and bottom (and just not show anything below the code fence). What about that?

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -301,3 +301,42 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+```ts {4}
+import { actions, getActionPath } from 'astro:actions'
+
+navigator.sendBeacon(
+  getActionPath(actions.like),
+  new Blob([JSON.stringify({ id: 'YOUR_ID' })], {
+    type: 'application/json'
+  })
+)
+```
```
florian-lefebvre, line 342:
``````suggestion
```astro title="src/components/my-component.astro" {8,10}
<script>
import { actions, getActionPath } from 'astro:actions'

await fetch(getActionPath(actions.like), {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    Authorization: 'Bearer YOUR_TOKEN'
  },
  body: JSON.stringify({ id: 'YOUR_ID' }),
  keepalive: true
})
</script>
```

This example shows how to call the same `like` action using the [`sendBeacon`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) API:

```astro title="src/components/my-component.astro" {5}
<script>
import { actions, getActionPath } from 'astro:actions'

navigator.sendBeacon(
  getActionPath(actions.like),
  new Blob([JSON.stringify({ id: 'YOUR_ID' })], {
    type: 'application/json'
  })
)
</script>
```
``````

File: src/content/docs/en/reference/modules/astro-actions.mdx [hunk 7]
```diff
@@ -301,3 +301,46 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+
+**Type:** `(action: ActionClient<any, any, any>) => string`
+<Since v="5.1.0" />
+</p>
+
+The `getActionPath()` utility accepts an action and returns a URL path so you can execute an action call as a `fetch()` operation directly. This allows you to provide details such as custom headers when you call your action.
+
+This example shows how to call a defined `like` action passing the `Authorization` header and the [`keepalive`](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive) option:
+
+```astro title="src/components/my-component.astro" {8,10}
```
sarah11918, line 317:
``````suggestion
```astro title="src/components/my-component.astro" {8,11}
``````

File: src/content/docs/en/reference/modules/astro-actions.mdx [hunk 8]
```diff
@@ -301,3 +301,46 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+<script>
```
sarah11918, line 318: Just double-checking you meant this to  now be in a `<script>` tag!

florian-lefebvre: Yes this is on purpose! This could is supposed to run on the client (altho it's possible on the server too, just seems like a more usecase)

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -301,3 +301,46 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+import { actions, getActionPath } from 'astro:actions'
+
+navigator.sendBeacon(
+  getActionPath(actions.like),
+  new Blob([JSON.stringify({ id: 'YOUR_ID' })], {
+    type: 'application/json'
+  })
+)
+</script>
+```
```
florian-lefebvre, line 346: @sarah11918 see https://github.com/withastro/astro/pull/12721/files#r1888930180
``````suggestion
```

:::note
Actions do not return plain JSON. Use `deserializeActionResult` from `astro:actions` if you need the returned data
:::

``````

Comment (sarah11918) [reply src/content/docs/en/reference/modules/astro-actions.mdx:346]: This will work in a pinch, but we don't actually have any code showing `deserializeActionResult` in use (not even the API reference). This could be a good place for that!

If you or @bholmesdev can whip something up here, then I think showing it in use is *better* than having a note that says "use it" with a link to something that similarly doesn't show how to use it.

Comment (florian-lefebvre) [reply src/content/docs/en/reference/modules/astro-actions.mdx:346]: I've not linked it on purpose, because the documented `deserializeActionResult` is the one returned from `getActionContext` above in the reference. This one we refer to is directly exported from `astro:actions`

Comment (sarah11918) [reply src/content/docs/en/reference/modules/astro-actions.mdx:346]: I think what we changed in the changeset works, actually! Let me suggest that here!

**Review by sarah11918 — APPROVED** [review]
Amazing, good to go @florian-lefebvre ! 🥳

File: src/content/docs/en/reference/modules/astro-actions.mdx [hunk 10]
```diff
@@ -301,3 +301,46 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+
+### `getActionPath()`
+
+<p>
+
+**Type:** `(action: ActionClient<any, any, any>) => string`
+<Since v="5.1.0" />
+</p>
+
+The `getActionPath()` utility accepts an action and returns a URL path so you can execute an action call as a `fetch()` operation directly. This allows you to provide details such as custom headers when you call your action. Then, you can [handle the custom-formatted returned data](https://docs.astro.build/en/guides/actions/#handling-returned-data) as needed, just as if you had called an action directly.
```
sarah11918, line 313:
```suggestion
The `getActionPath()` utility accepts an action and returns a URL path so you can execute an action call as a `fetch()` operation directly. This allows you to provide details such as custom headers when you call your action. Then, you can [handle the custom-formatted returned data](/en/guides/actions/#handling-returned-data) as needed, just as if you had called an action directly.
```

PR #10354, opened 2024-12-12
Add prefix to tab panels IDs to avoid clashes

yanthomasdev:
#### Description (required)

This PR fixes the occasion on which a tab panel ID conflicts with a page heading ID, so you cannot click to navigate to that specific heading from links.

For example, it currently occurs on the [`integrations` heading of Configuration Reference](https://docs.astro.build/en/reference/configuration-reference/#integrations), since we have a tab of the same name "Integrations" that gets the id `integrations`.


delucis (approved):
Thanks for fixing @yanthomasdev!

PR #10362, opened 2024-12-13

f33w:
# .default() custom validation in zod triggers error

The error occurs only using referenced collection. I made a demo:
https://stackblitz.com/edit/github-e7qgsrxj?file=src%2Fcontent.config.ts

My discord:
twel

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -290,7 +290,7 @@ defineCollection({
 
 ##### Zod schema methods
 
-All [Zod schema methods](https://zod.dev/?id=schema-methods) (e.g. `.parse()`, `.transform()`) are available, with some limitations. Notably, performing custom validation checks on images using `image().refine()` is unsupported.
+All [Zod schema methods](https://zod.dev/?id=schema-methods) (e.g. `.parse()`, `.transform()`) are available, with some limitations. Notably, performing custom validation checks on images using `image().refine()` is unsupported. Another validation check that is not supported is .default().
 
 #### Defining collection references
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks for letting us know about this! I will ask @ascorbic to take a look and let us know if this is expected behaviour that should be documented, or unexpected behaviour that should instead be fixed in collections!  (We don't update docs for e.g. bug fixes, we fix those! So I'll let Matt give us some guidance here.)

ascorbic:
Is `reference().default()` something that previously worked? I'm not sure what it's meant to be achieving.

f33w:
> Is `reference().default()` something that previously worked? I'm not sure what it's meant to be achieving.

It did. I was using it to specify the default author from the so named collection. Not a big deal, I am just manually selecting an author each time creating a post.

delucis:
> Is `reference().default()` something that previously worked? I'm not sure what it's meant to be achieving.

Yeah, basically the same as other Zod types, e.g. `z.string().default('Default')`, but in this case for specifying a default reference.

sarah11918:
So, if this worked before and does not now, is this perhaps rather a bug to investigate than something to change documentation?

ascorbic:
I think this is a bug

sarah11918:
Thanks Matt! @f33w , would you like to file an issue on the main https://github.com/withastro/astro repo for this?

Unfortunately, I don't think I can transfer a *PR* as an issue to another repo, so a new issue will have to be filed.

I will close this as a docs PR, since this is probably something to be fixed, not docs to be changed! Thank you again for reporting! 🙌

PR #10367, opened 2024-12-14
Update rss.mdx

jestsee:
#### Description (required)

replaced `slug` with `id` in accordance with the updated `getCollection` API introduced in Astro v5.

- Closes #
- Suggested label:


yanthomasdev (approved):
Nice catch, thanks!

PR #10370, opened 2024-12-15
Clarify content collection and page routing for Astro v5 

### LINKED ISSUE #10300: Content collection loader examples are inconsistent
### 📚 Subject area/topic

Content collections

- https://docs.astro.build/en/guides/content-collections/
- https://docs.astro.build/en/guides/upgrade-to/v5/
- https://docs.astro.build/en/tutorial/6-islands/4/
- https://docs.astro.build/en/guides/integrations-guide/mdx/
- https://docs.astro.build/en/reference/modules/astro-content/

The Astro v5 loader examples have a different file pattern depending on the page. The different examples are confusing, and it's not sure which pattern is the recommended default.

For example, the upgrade guide shows a loader pattern that excludes underscore:

```ts
loader: glob({ pattern: '**/[^_]*.md', base: "./src/data/blog" }),
```
The astro:content reference removes the underscore check:

```ts
loader: glob({ pattern: '**/*.md', base: './src/data/blog' }),
```

The MDX integration guide adds a backslash:

```ts
loader: glob({ pattern: "**\/*.{md,mdx}", base: "./src/blog" }),
```

I am willing to submit a PR if I know the recommended format.

_No response_

anaxite:
## Summary

This pull request reviews and edits wording around page and content collection routing to better explain file exclusion in 5.0.

## Description

In 5.0, content collections don't exclude any files by default. Particular globs are needed to exclude a file, which isn't necessarily clear. This change may also create confusion when comparing the behavior to page routing, which does exclude underscores.

This PR attempts to clarify the behavior change, and make content collection behavior stand out more compared to page routing. 

### Related issues & labels

- Closes #10300
- Suggested labels:
    - `code snippet update`
    - `improve documentation`


sarah11918:
This is looking pretty good @anaxite ! Is there a reason you still have this marked as a draft? Are you doing more with it?

Also noting that I will add something to the v5 upgrade guide in case people were using the built-in prefix trick that was supported in the old content collections API

anaxite:
I'm doing a last check, and then it'll be out of draft :)

sarah11918:
Noting that the Upgrade guide should also something like the following changes to the [Legacy collections introduction](https://docs.astro.build/en/guides/upgrade-to/v5/#legacy-v20-content-collections-api):

```
In Astro 4.x, content collections were defined, queried, and rendered using [the Content Collections API first introduced in Astro v2.0](https://astro.build/blog/introducing-content-collections/). All collection entries were local files within the reserved `src/content/` folder. Additionally, Astro's [file name convention to exclude building individual pages](/en/guides/routing/#excluding-pages) was built in to the Content Collections API.

Astro 5.0 introduces a new version of content collections using the Content Layer API which brings several performance improvements and added capabilities. While old (legacy) and new (Content Layer API) collections can continue exist together in this release, there are potentially breaking changes to existing legacy collections.

This release also removes the option to prefix collection entry file names with an underscore (`_`) to prevent building a route.
```

anaxite:
Alright, feel free to go ahead with the upgrade guide!

src/content/docs/en/guides/upgrade-to/v5.mdx
anaxite, line 1: Under #updating-existing-collections, on step 2, would it be worth adding a sentence like "Edit the `pattern` depending on what files you want to exclude." or is that too wordy? I'm thinking that's another signal that behavior has changed.

sarah11918: It's not a bad idea, but you're right, it's starting to get a bit much for something that probably most people will *not* have, and therefore an extra thought that might get put in their head? "Wait, exclude?"

Can we try as-is to start, as already a definite improvement?

anaxite: Sounds good. I like the addition!

PR #10377, opened 2024-12-16
feat: Add remote image caching behaviour docs

oliverlynch:
#### Description (required)

Currently the behaviour of Astro's remote image cache is undocumented, which makes discovering the feature difficult and its behaviour unclear. This change adds an explanation of the asset cache, as well as some technical details on how it functions, which should be helpful for anyone configuring a remote image server for Astro. 

Part of this change references the revalidation feature introduced in Astro PR [#12426](https://github.com/withastro/astro/pull/12426), however the rest documents existing functionality. 

- Closes #

#### For Astro version: `5.1`. See astro PR [#12426](https://github.com/withastro/astro/pull/12426).

Not yet merged - milestone.


ascorbic (commented):
This looks technically correct (which is the best kind of correct)

src/content/docs/en/guides/images.mdx
```diff
@@ -28,7 +28,20 @@ Store your images in the `public/` folder if you want to avoid any processing or
 You can also choose to store your images remotely, in a [content management system (CMS)](/en/guides/cms/) or [digital asset management (DAM)](/en/guides/media/) platform.
 Astro can fetch your data remotely using APIs or display images from their full URL path.
 
- For extra protection when dealing with external sources, Astro's image components and helper function will only process (e.g. optimize, transform) images from [authorized image sources specified in your configuration](#authorizing-remote-images). Remote images from other sources will be displayed with no processing.
+For extra protection when dealing with external sources, Astro's image components and helper function will only process (e.g. optimize, transform) images from [authorized image sources specified in your configuration](#authorizing-remote-images). Remote images from other sources will be displayed with no processing.
+
+#### Asset Caching
```
Princesseuh, line 33: Might be interesting to note that local images are also cached.

Princesseuh:
It might be interesting to mention where the images are cached, sometimes users want to know what to need to keep between builds for caching (here you can link to `cacheDir` in the config)

src/content/docs/en/guides/images.mdx
sarah11918: @Princesseuh Do you think maybe an entire section on Image Caching would make sense?

Princesseuh: I would say so! But it might be confusing for some users, since it's a bit more advanced than the rest and it's not something you interact with directly

oliverlynch:
I've moved the new documentation to its own section "Asset Caching", and mentioned how local images are also cached. 

I added a mention of `cacheDir` in the body text instead of just the readmore link -
Is it okay to write the default `cacheDir` value here, or should I avoid repeating it and have people click through to the configuration reference?

I also expanded on revalidation, explaining that the remote server requires support for the `If-Modified-Since` and `If-None-Match` headers for revalidation to work. While this is a bit technical, if someone is having difficulty getting revalidation working it is probably best for this to be clearly documented.

src/content/docs/en/guides/images.mdx
```diff
@@ -475,6 +475,26 @@ export default defineConfig({
 });
 ```
 
+## Asset Caching
+
+Astro stores processed image assets in a cache directory during site builds, for both local and [remote images from authorized sources](/en/guides/images/#authorizing-remote-images). By preserving the cache directory between builds, processed assets are reused, improving build time and bandwidth usage.
+The default cache directory is `./node_modules/.astro`, however this can be changed using the `cacheDir` configuration value.
```
sarah11918, line 481:
```suggestion
Astro stores processed image assets in a cache directory during site builds, for both local and [remote images from authorized sources](/en/guides/images/#authorizing-remote-images). By preserving the cache directory between builds, processed assets are reused, improving build time and bandwidth usage.

The default cache directory is `./node_modules/.astro`, however this can be changed using the `cacheDir` configuration value.
```

src/content/docs/en/guides/images.mdx
```diff
@@ -475,6 +475,26 @@ export default defineConfig({
...
+
+<ReadMore>See the configuration reference for [`cacheDir`](/en/reference/configuration-reference/#cachedir)</ReadMore>
```
sarah11918, line 483:
```suggestion
Astro stores processed image assets in a cache directory during site builds for both local and [remote images from authorized sources](/en/guides/images/#authorizing-remote-images). By preserving the cache directory between builds, processed assets are reused, improving build time and bandwidth usage.

The default cache directory is `./node_modules/.astro`, however this can be changed using the [`cacheDir`](/en/reference/configuration-reference/#cachedir) configuration setting.
```
The `cacheDir` entry isn't very exciting 😅 , so it feels like a little much to give it an entire "Read More" link!

src/content/docs/en/guides/images.mdx
```diff
@@ -475,6 +475,26 @@ export default defineConfig({
...
+### Remote Images
+
+Remote images in the asset cache are managed based on [HTTP Caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching), and respect the [Cache-Control header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) returned by the remote server.
+Images are cached if the Cache-Control header allows, and will be used until they are no longer [fresh](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#fresh_and_stale_based_on_age).
+
+#### Revalidation
+
+<p><Since v="5.1.0" /></p>
+
+[Revalidation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#validation) reduces bandwidth usage and build time by checking with the remote server if an expired cached image is still up-to-date. If the server indicates that the image is still fresh, the cached version is reused, otherwise the image is redownloaded.
```
sarah11918, line 494:
```suggestion
[Revalidation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#validation) reduces bandwidth usage and build time by checking with the remote server whether an expired cached image is still up-to-date. If the server indicates that the image is still fresh, the cached version is reused, otherwise the image is redownloaded.
```
tiny grammar nit, but in general, this is all super well written and I really appreciate it!

src/content/docs/en/guides/images.mdx
```diff
@@ -475,6 +475,26 @@ export default defineConfig({
...
+
+Revalidation requires the remote server to send [Last-Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified) and/or [Etag (entity tag)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) headers with its responses, and to support the [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) and [If-None-Match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) headers.
```
sarah11918, line 496:
```suggestion
Revalidation requires that the remote server send [Last-Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified) and/or [Etag (entity tag)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) headers with its responses. This feature is available for remote servers that support the [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) and [If-None-Match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) headers.
```
I know it looks worse in Markdown, but still tried to split this up into two more manageable chunks.

sarah11918 (commented):
Looks fantastic, @oliverlynch ! I really appreciate the great docs here, and the going above and beyond to document more than just your feature! 💜 

Had some small suggestions for your consideration, and can't wait to get this in with the next minor release! 🥳

oliverlynch:
Thanks @sarah11918 for the suggestions, I've committed them as-is.

sarah11918 (approved):
Fantastic @oliverlynch ! Welcome to Team Docs! :partying_face:

PR #10379, opened 2024-12-16
i18n(fr): update `guides/testing.mdx`

ArmandPhilippot:
#### Description (required)

The French translation of `guides/testing.mdx` was marked as outdated due to #9240 but the changes are not applicable (`island` > `islands`) so instead I updated the code snippets to make them consistent:
* a page to test was translated but the expectation was not translated
* a page to test and an expectation was both translated but in a different way...

- Suggested label: i18n


yanthomasdev (approved):
LGTM! Thanks!

PR #10393, opened 2024-12-16
Fix content.config.ts location in docs

Rushmore75:
The path of the new content collection config file has change since 4. The docs said it moved to `src/content.config.ts` however, it ***actually*** moved to just `content.config.ts`. When the file exists in under `src/` it causes issues with collections not being found.

I see that you want changes in only one language, however this is just a file path change so I've done it everywhere it existed. If you truly need it broken into just `en` I can do that, or really anyone with `sed`

- Suggested label: code-snippet-update, improve documentation


Rushmore75:
it looks like `src/content.config.ts` *right*, but the docs are still confusing, closing

PR #10394, opened 2024-12-16
Fix content.config.ts location in docs [i18nIgnore]

Rushmore75:
The path of the new content collection config file has change since 4. The docs said it moved to `src/content.config.ts` however, it ***actually*** moved to just `content.config.ts`. When the file exists in under `src/` it causes issues with collections not being found.

I see that you want changes in only one language, however this is just a file path change so I've done it everywhere it existed. If you truly need it broken into just `en` I can do that, or really anyone with `sed`

- Suggested label: code-snippet-update, improve documentation


sarah11918:
Hi! I'm sorry you're having trouble with your collections being found, but the correct location for the `content.config.ts` is absolutely within `src/`

It might be that moving your file outside of `src/` makes it not seen by Astro at all, and with some of our built-in backwards compatibility, this might coincidentally be working because this file was in fact not necessary in v4.

So I will close this as incorrect, and can recommend you try to make your project work knowing that the file must be in `src/`. If you need assistance, or someone to help troubleshoot your code, you are welcome to post in our `#support` channel in the Astro Discord. It's also possible you've discovered a bug! If so, you can submit an issue to the main Astro repo with a reproduction so they can look into it.

And for future reference, please do follow the instructions and submit in one language only. This requirement is clearly stated in the template.

PR #10410, opened 2024-12-17
fix: update hook signature

ematipico:
#### Description (required)

The documentation is incorrect, and the `updateConfig` is for **vite**, not Astro.

I wonder if we should change its name...

I noticed the mistake in https://github.com/withastro/astro/issues/12372


sarah11918 (approved):
Thank you for fixing @ematipico !

rbsummers:
Hi there! I think this PR, while necessary, won't close the referred issue because we need some changes in the astro package to have the config matter for `vite.build`. I've opened a PR to do that but for now I have only manually tested it, I'll need more time to figure out the testing setup

rbsummers:
Hi @sarah11918,

It was a tricky situation because the feature was implemented (the function `updateConfig` was in the code, probably because it mimicked some other function available & well documented in a different hook), but it did nothing and it was not properly documented. My PR (which @ematipico merged a bit ago, thank you!) fixed the first point, and this one addresses the second.

If merging this makes the doc change immediately, maybe we need to hold on a bit until the version with the fix (5.0.10 I guess?) is released and then merge.

ematipico:
> If merging this makes the doc change immediately, maybe we need to hold on a bit until the version with the fix (5.0.10 I guess?) is released and then merge.

Fixing the API signature should be done regardless of the fix of the code.

sarah11918:
@ematipico You have my approval to merge whatever is correct *now*! I just wasn't sure what was happening. 😄 

I agree, if this PR is entirely correct *now*, my suggestion would be to merge it!

PR #10418, opened 2024-12-17
Add need for adapter to server islands guide

sarah11918:
#### Description (required)

This PR adds explicit mention that server islands require an adapter installed (in three places).

### DIFF
--- a/src/content/docs/en/guides/on-demand-rendering.mdx
+++ b/src/content/docs/en/guides/on-demand-rendering.mdx
@@ -21,7 +21,7 @@ On-demand rendering on the server at request time is also known as **server-side
 
 To render any page on demand, you need to add an **adapter**. Each adapter allows Astro to output a script that runs your project on a specific **runtime**: the environment that runs code on the server to generate pages when they are requested (e.g. Netlify, Cloudflare).
 
-You may also wish to add an adapter even if your site is entirely static and you are not rendering any pages on demand. For example, the [Netlify adapter](/en/guides/integrations-guide/netlify/) enables Netlify's Image CDN, and the [Vercel adapter](/en/guides/integrations-guide/vercel/) enables services such as web analytics.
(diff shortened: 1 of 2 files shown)


src/content/docs/en/guides/on-demand-rendering.mdx
```diff
@@ -21,7 +21,7 @@ On-demand rendering on the server at request time is also known as **server-side
 
 To render any page on demand, you need to add an **adapter**. Each adapter allows Astro to output a script that runs your project on a specific **runtime**: the environment that runs code on the server to generate pages when they are requested (e.g. Netlify, Cloudflare).
 
-You may also wish to add an adapter even if your site is entirely static and you are not rendering any pages on demand. For example, the [Netlify adapter](/en/guides/integrations-guide/netlify/) enables Netlify's Image CDN, and the [Vercel adapter](/en/guides/integrations-guide/vercel/) enables services such as web analytics.
+You may also wish to add an adapter even if your site is entirely static and you are not rendering any pages on demand. For example, the [Netlify adapter](/en/guides/integrations-guide/netlify/) enables Netlify's Image CDN, and [server islands](/en/guides/server-islands/) require an adapter to use `server:defer` on a component.
```
yanthomasdev, line 24: I am not so sure about the wording, at first I got confused because it sounded like the adapter is what needs to use `server:defer` on a component, not that you need an adapter to use server islands. Or it might just be a brain fart on my side, it happens.

Maybe something like this is more clear? 

```suggestion
You may also wish to add an adapter even if your site is entirely static and you are not rendering any pages on demand. For example, the [Netlify adapter](/en/guides/integrations-guide/netlify/) enables Netlify's Image CDN, and to use [server islands](/en/guides/server-islands/) you also require an adapter to allow you to use `server:defer` on components.
```

yanthomasdev (commented):
Just a small detail, but should be good to go after you address it

src/content/docs/en/guides/on-demand-rendering.mdx
sarah11918: lol, yours seems more convoluted to me, so we might need an impartial judge here! 💜

yanthomasdev: yeah, that works for me!

yanthomasdev (approved):

PR #10448, opened 2024-12-18
feat(`integration-guide`): make some changes to the `guides/integration-guide/index.mdx`

trueberryless:
#### Description (required)

When I translated the `guides/integration-guide/index.mdx` page I noticed some "inconsistencies" (mostly missing dots and Steps) but maybe it's intended not to use Steps in some scenarios. Please let me know if you would like to have such changes, I can delete some changes if only the missing dots are what you wish for!

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/index.mdx
+++ b/src/content/docs/en/guides/integrations-guide/index.mdx
@@ -6,9 +6,9 @@ sidebar:
 i18nReady: true
 ---
 
-import IntegrationsNav from '~/components/IntegrationsNav.astro'
-import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
-import { Steps } from '@astrojs/starlight/components'
+import IntegrationsNav from '~/components/IntegrationsNav.astro';
+import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
+import { Steps } from '@astrojs/starlight/components';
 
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/index.mdx
```diff
@@ -76,16 +76,36 @@ It's even possible to add multiple integrations at the same time!
...
 ### Manual Installation
 
 Astro integrations are always added through the `integrations` property in your `astro.config.mjs` file.
 
 There are three common ways to import an integration into your Astro project:
+
+<Steps>
 1. [Install an npm package integration](#installing-an-npm-package).
-2. Import your own integration from a local file inside your project.
+2. Import your own integration from a local file inside your project.
```
trueberryless, line 108: I'm not sure, what symbol was there instead of a space, but VS Code marked it yellow, so I replaced it with a space...

sarah11918: 3 different ways are not steps. This is one of those cases where we are listing items, not dictating things that you have to do in order. In fact, if you do any one of these, you would NOT do any of the others! 😄

sarah11918 (commented):
Except for the list of things that shouldn't be steps, everything else here looks good! 💪

src/content/docs/en/guides/integrations-guide/index.mdx
trueberryless: Understood 🫡

yanthomasdev (approved):
LGTM! Thanks!

PR #10451, opened 2024-12-18
Simplify link imports in `guides/imports.mdx`

trueberryless:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/de/guides/imports.mdx
+++ b/src/content/docs/de/guides/imports.mdx
@@ -1,11 +1,8 @@
 ---
-title: Importe Referenz
-description: Erfahre, wie du verschiedene Dateitypen in Astro importieren kannst.
+title: Importe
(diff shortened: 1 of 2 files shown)


sarah11918:
Hi @trueberryless ! Can you be sure to title this like a translation, since much of what I see proposed here has to do with changing the German language. I see one English link change, and that should really be its own PR, but it's so tiny it's fine to include it here.

trueberryless:
> Hi @trueberryless ! Can you be sure to title this like a translation, since much of what I see proposed here has to do with changing the German language. I see one English link change, and that should really be its own PR, but it's so tiny it's fine to include it here.

I justed noticed as well, I didnt intend to include the German tranlation in here, `git add .` screwed me up... 

There is already an existing PR for German translation, so I'm gonna remove the de file changes here, and leave the PR with the small change open nonetheless, if it's okay...

sarah11918 (approved):
This works, thanks!

PR #10456, opened 2024-12-19
docs(content-collections): remove stale anti-use-case

rayhanadev:
#### Description (required)

With the advent of custom loaders in Astro v5, the "non-standard" data anti-use-case is invalid (yay!). This should be removed from documentation to prevent confusion :)

Discord Username: rayhanadev

- Closes #
- Suggested label: [improve documentation](https://github.com/withastro/docs/labels/improve%20documentation)


sarah11918 (commented):
Yay indeed, @rayhanadev !!  😄 

While we no longer have to worry about the data format now, I *did* like keeping the reference to alternative loading methods someone might possibly want or need. So I think we can sneakily combine that with the updated in real time!

In discussing these options, the team also came up with another reason to not use a collection with a loader, so what about throwing that in here, too? What do you think?

rayhanadev:
makes sense to me :D

PR #10467, opened 2024-12-19
i18n(fr): create `reference/legacy-flags.mdx`

ArmandPhilippot:
#### Description (required)

Translates in French the file `reference/legacy-flags.mdx` added in #9240

> [!NOTE]
> In French, I don't think we use `drapeaux` for `flags` in this context so I hesitated between `indicateurs` and `options`. I think the last one is more commonly used and this matches the term we use in CLI reference.

- Suggested label: i18n


thomasbnt:
> In French, I don't think we use `drapeaux` for `flags` in this context so I hesitated between `indicateurs` and `options`. I think the last one is more commonly used and this matches the term we use in CLI reference.

I used `flags` in french in https://github.com/withastro/docs/pull/10469. :+1:

ArmandPhilippot:
> I used `flags` in french in #10469. 👍

I'd say we should be consistent across the docs to avoid confusion the reader. So I think we should agree on a single term...
I checked on Starlight docs and I see `option` is also used in two places (https://starlight.astro.build/fr/manual-setup/#configurer-les-collections-de-contenu and https://starlight.astro.build/fr/reference/overrides/#id) where `flag` is used in the VO. And it's also what we have [in CLI reference](https://docs.astro.build/fr/reference/cli-reference/#options) for `Common flags`.
But also have other pages in Astro Docs that uses `drapeau` and `indicateur` (some integrations and upgrade guides for example) ... But I don't see any `flag` in Algolia search (in the VF I mean).

thomasbnt:
I think, `flags`makes it easier for developers to understand, and even in French. :+1:

thomasbnt (approved):
LGTM 🚀

ArmandPhilippot:
> I think, flagsmakes it easier for developers to understand, and even in French. 👍

I never saw `flag` in a French documentation (but I have to say that I don't read much documentation in French so... 😅 ). I'm not sure it has more meaning for French people... I checked what other docs are using for `flag`:
* React uses `drapeau` (for example `le drapeau isPending` in https://fr.react.dev/reference/react/useTransition#marking-a-state-update-as-a-non-blocking-transition)
* Vue uses `indicateur` (for example https://fr.vuejs.org/api/compile-time-flags.html) and `drapeau` (https://fr.vuejs.org/api/sfc-script-setup.html#default-props-values-when-using-type-declaration)
* In the context of CLI only, Wikipédia uses `option` (https://fr.wikipedia.org/wiki/Interface_en_ligne_de_commande#Option) otherwise it uses `drapeau` (https://fr.wikipedia.org/wiki/Drapeau_(informatique))

So I'd say we should translate it... but maybe `option` is not the best translation? I don't know...

ArmandPhilippot:
Yeah... 😅
In the CLI context, we also have [PHP that uses `option` in French for `flag`](https://www.php.net/manual/fr/features.commandline.options.php). In other contexts, they also use [`option`](https://www.php.net/manual/fr/function.fdf-set-flags.php) and [`drapeau`](https://www.php.net/manual/fr/function.uopz-flags.php).
Nobody knows how to translate it... 😆 

IMO, `option` is the most significant term in both contexts (`option de ligne de commande`, `option de configuration`...).

yanthomasdev (approved):
LGTM! Good job!

PR #10471, opened 2024-12-19
Fix typo in configuration example for Netlify Blob sessions

sledsworth:
#### Description (required)

Netlify blob example had `adapter` incorrectly nested in the `experimental`  property instead of at the root of the configuration object.

Discord: sledsworth


sarah11918 (approved):
Ooh, great spotting @sledsworth ! We appreciate you taking the time to come fix this, and are happy to welcome you to Team Docs! 🥳

PR #10472, opened 2024-12-19
i18n(fr): Update `reference/modules/astro-actions` from #10447

thomasbnt:
#### Description (required)
Update `reference/modules/astro-actions` from #10447

- Suggested label: i18n


ArmandPhilippot (approved):
Oh I missed that one yesterday, LGTM!

PR #10492, opened 2024-12-20
i18n(fr): create `content-loader-reference.mdx`

ArmandPhilippot:
#### Description (required)

Translates in French the new file `content-loader-reference.mdx` added in #9240

> [!NOTE]
> I translated `object loader` with `chargeur sous forme d'objet` because it seems to me this is the loader format and not a `chargeur d'objets` (not the same meaning)... but if you have better wording, feel free to suggest!

- Suggested label: i18n


thomasbnt (approved):
LGTM !

PR #10493, opened 2024-12-20
Modernize dependencies

delucis:
#### Description (required)

Updates various docs dependencies that are getting a bit dusty including removing some (like `node-fetch` that are no longer needed at all).

##### Removed

- `node-fetch` (no longer needed in newer Node versions)
- `eslint-plugin-react` (we don’t have any JSX to lint)
- `remark-directive` (not used directly anymore — was used for asides but that’s now handled by Starlight)

##### Major

- `@11ty/eleventy-fetch`: 3.0.0 => 5.0.1 (removes `node-fetch` under the hood)
- `@types/node`: 18.6.4 => 20.17.10 (we’re running on Node 20 everywhere now)
- `eslint`: 8.29.0 => 9.17.0 (switched to the newer “flat” config format)
  - `@typescript-eslint/eslint-plugin` was removed in favour of `typescript-eslint`
  - `@typescript-eslint/parser`: 5.46.1 => 8.18.1
  - `astro-eslint-parser`: 0.16.0 => 1.1.0
  - `eslint-plugin-astro`: 0.21.0 => 1.3.1
  - **Added:** `@eslint/js` and `globals`
- `domhandler`: 4.3.1 => 5.0.3 (types for `htmlparser2`)
- `htmlparser2`: 7.2.0 => 9.1.0 (used for DOM parsing in link checker CI, required a tweak to import style as they changed package format)
- `p-retry`: 5.1.1 => 6.2.1 (used for our cached fetches, v6 only dropped old Node version support)
- `sharp`: 0.32.6 => 0.33.5

##### Minor

`pnpm` itself: 9.3.0 => 9.15.1

- `@actions/core`: 1.9.0 => 1.11.1 (used in link checker CI and Tuesday translation bot)
- `@docsearch/css` & `@docsearch/js`: 3.5.1 => 3.8.2
- `fast-glob`: 3.2.11 => 3.3.2
- `prettier`: 3.3.3 => 3.4.2
- `typescript`: 5.0.2 => 5.6.2 (required a few minor tweaks to get type checks passing)

#### Other changes

- Removed two script utilities we no longer use (pre-Lunarian relics)

#### Not updated

- `canvas-confetti` — recent releases add new features we don’t need and increase bundle size
- `remark-smartypants` — causes type errors for some reason, didn’t investigate why


eslint.config.mjs
```diff
@@ -0,0 +1,56 @@
...
+
+	{
+		languageOptions: {
+			ecmaVersion: 'latest',
+			sourceType: 'module',
+			parser: tseslint.parser,
+		},
+		plugins: {
+			'@typescript-eslint': tseslint.plugin,
+		},
```
HiDeoo, line 23: I don't think the config enables any type-checked rules, and this also does not use the legacy format which required the `plugins` field but this is [not the case anymore](https://typescript-eslint.io/getting-started/typed-linting) with a flat config, so I think this entire section could be omitted.

HiDeoo (approved):
Looks good to me and seems to be working as expected locally through the CLI and Visual Studio Code (only tested these 2). Only left a small comment to potentially remove a little bit more config.

eslint.config.mjs
delucis: Ah, very possible, thank you for the feedback! I found the migration guidance pretty overwhelming and hard to piece together across the different plugins. Will try omitting this.

delucis: Yeah, doesn’t seem to be needed testing locally — neither were a couple of the rules I had there based on the old config, so I removed those too. We can always add things back if we hit problematic lint errors later.

sarah11918 (approved):
Tested the deploy preview with all of the usual suspects!

PR #10501, opened 2024-12-21
Remove code that relies on v4 cookie redirects

tylerlwsmith:
#### Description (required)

Astro 5 removed introduced a breaking change where [failed actions no longer redirect back to the page where a form was submitted](https://docs.astro.build/en/guides/upgrade-to/v5/#changed-actions-submitted-by-html-forms-no-longer-use-cookie-redirects).

This PR modifies an "actions" example that relied on the v4 behavior, making it work with Astro v5.

- Closes `N/A`


sarah11918 (approved):
Thank you for catching this @tylerlwsmith , and for taking the time to come update the code sample! We really appreciate it, and are happy to welcome you to Team Docs! 🥳

PR #10509, opened 2024-12-23
Fix param object key in content-collections.mdx

jhuleatt:
#### Description (required)

The sample as-is resulted in the error "[ERROR] Missing parameter: post"

- Closes #
- Suggested label:


ArmandPhilippot:
Hi @jhuleatt! Are you sure you have copied this snippet in a file named `[id].astro`?

Looking at the current version, I don't see anything wrong with the code snippet. The given path is `src/pages/posts/[id].astro` so `id` [is the correct param name](https://docs.astro.build/en/guides/routing/#dynamic-routes), not `post` as suggested...

Maybe you tried this snippet in a file named `[post].astro` so you get an error... which is expected. In that case, you're right, you need to update the param name in your file, not in the docs!

PR #10544, opened 2024-12-27
fix: add wait to code sample

justinnoel:
#### Description

When calling an action from a server endpoint, an `await` is required or `data` and `error` will be `undefined`.


sarah11918 (approved):
Thanks so much for catching this @justinnoel and taking the time to add the fix. Andwelcome you to Team Docs 🥳 (But, how is this your first contribution?? 😄 )

justinnoel:
> But, how is this your first contribution?? 

I only built a few simple things with Astro all those years ago. Then, I spent a lot of time with Hono. Now, I'm back to dabbling with Astro. It's amazing what y'all have done with it!!!

PR #10594, opened 2025-01-03
[markdown-content] clearer distinction between imported Markdown and collections data

### LINKED ISSUE #10361: Improve Clarity on Accessing Headings from Markdown Files via Content Collections
### 📚 Subject area/topic

Markdown / Content Collections

https://docs.astro.build/en/guides/markdown-content
https://docs.astro.build/en/guides/content-collections

While working with Astro content collections, I ran into confusion about how to access `headings` from Markdown files. Since `getHeadings()` is listed under [Available Properties](https://docs.astro.build/en/guides/markdown-content/#available-properties) on the Markdown docs page, I initially assumed it would work with entries from `getCollection()`. When it didn’t, I looked for alternatives.

I explored third-party resources, but many referenced outdated methods like `getHeaders()` or suggested that `headings` should be a key on the entry object itself. After more digging in the Astro docs, I finally found that headings are accessible via the `render()` function, as explained in the [Content Collections](https://docs.astro.build/en/guides/content-collections/#rendering-body-content) docs. However, since I started with the Markdown page, I didn’t realize this connection right away. The link from [The \<Content /\> Component](https://docs.astro.build/en/guides/markdown-content/#the-content--component) section to the Content Collections page wasn’t enough to signal that `headings` were addressed there.

Based on this experience, I think small clarifications could help others avoid the same confusion.

#### 2. Update the text in the \<Content /\> Component section:

  Update the link text pointing to the Content Collections page to explicitly mention headings (italics for emphasis):
>"You can similarly render the HTML content _and headings_ of a Markdown collection entry..."

#### 3. Update the code example in the \<Content /\> Component section:

  Include `headings` in the destructuring of the `render()` return, consistent with the Content Collections example:
```ts
const { Content, headings } = await render(product);
```

I think these small updates could help future developers quickly identify the right approach for accessing headings in content collections. If they resonate with the docs team, I'd be happy to make a PR.

Thanks!

_No response_

sarah11918:
Addresses an issue where it was difficult to find information about accessing headings data from Markdown content collections entries.

The page was written before collections, and therefore emphasized exported Markdown properties even though this information is only appropriate when importing Markdown files. More relevant information is available on the Contents Collections guide page, and in reference, but someone may first look at the Markdown guide even when their Markdown files are part of a collection.

This does not duplicate content found elsewhere, but attempts to:
- empahsize in multiple places that you may be working with Markdown that came from either a file import or a content collection (there's a difference!)
- adds more contextual links throughout, notably adds a link to the `render()` reference on the Content Collections guide page which contains helpful information in understanding how headings are accessed differently in Content Collections

Closes #10361

### DIFF
--- a/src/content/docs/en/guides/markdown-content.mdx
+++ b/src/content/docs/en/guides/markdown-content.mdx
@@ -31,7 +31,7 @@ Local Markdown can be imported into `.astro` components using an `import` statem
 
 If you have groups of related Markdown files, consider [defining them as collections](/en/guides/content-collections/). This gives you several advantages, including the ability to store Markdown files anywhere on your filesystem or remotely.
 
-Collections use content-specfic, optimized API for [querying and rendering your Markdown content](#markdown-from-content-collections-queries) instead of file imports. Collections are intended for sets of data that share the same structure, such as blog posts or product items. When you define that shape in a schema, you additionally get validation, type safety, and Intellisense in your editor.
+Collections use content-specfic, optimized APIs for [querying and rendering your Markdown content](#markdown-from-content-collections-queries) instead of file imports. Collections are intended for sets of data that share the same structure, such as blog posts or product items. When you define that shape in a schema, you additionally get validation, type safety, and Intellisense in your editor.
 
 <ReadMore>See more about [when to use content collections](/en/guides/content-collections/#when-to-create-a-collection) instead of file imports.</ReadMore>
 
--- a/src/content/docs/en/guides/markdown-content.mdx
+++ b/src/content/docs/en/guides/markdown-content.mdx
@@ -21,11 +21,17 @@ For additional functionality, such as including components and JSX expressions i
 
 ## Organizing Markdown files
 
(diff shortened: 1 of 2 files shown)


src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -21,11 +21,17 @@ For additional functionality, such as including components and JSX expressions i
...
 
 If you have groups of related Markdown files, consider [defining them as collections](/en/guides/content-collections/). This gives you several advantages, including the ability to store Markdown files anywhere on your filesystem or remotely.
 
-Collections also allow you to use content-specfic, optimized API for querying and rendering your content. Collections are intended for sets of data that share the same structure, such as blog posts or product items. When you define that shape in a schema, you additionally get validation, type safety, and Intellisense in your editor.
+Collections use content-specfic, optimized API for [querying and rendering your Markdown content](#markdown-from-content-collections-queries) instead of file imports. Collections are intended for sets of data that share the same structure, such as blog posts or product items. When you define that shape in a schema, you additionally get validation, type safety, and Intellisense in your editor.
```
sarah11918, line 34:
```suggestion
Collections use content-specfic, optimized APIs for [querying and rendering your Markdown content](#markdown-from-content-collections-queries) instead of file imports. Collections are intended for sets of data that share the same structure, such as blog posts or product items. When you define that shape in a schema, you additionally get validation, type safety, and Intellisense in your editor.
```
just matching verb tense!

PR #10595, opened 2025-01-03
[sitemap.mdx] update `lastmod` type

### LINKED ISSUE #10355: `lastmod` of SitemapItem object is not String type
### 📚 Subject area/topic

Integrations

https://docs.astro.build/ja/guides/integrations-guide/sitemap/#serialize

[The documentation](https://docs.astro.build/ja/guides/integrations-guide/sitemap/#serialize) states that `lastmod` is:

> lastmod (ISO formatted date, String type)

However, upon inspecting the [implementation code](https://github.com/withastro/astro/blob/main/packages/integrations/sitemap/src/index.ts#L33), `lastmod` is being handled as a Date type.

However since this Date type is based on Universal Time (UTC), it's not possible to directly specify the `lastmod` in a specific timezone (e.g., Japan Standard Time (JST)) using a string representation. It is my view that the implementation could be changed to consider the possibility of allowing the String type to be specified.

_No response_

sarah11918:
#### Description (required)

Updates the type as described in the issue.

Note: although they have also submitted a feature request to update `lastmod`, the docs should be correct in the mean time so I'm just making the edit for what is currently correct!

- Closes #10355

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -302,7 +302,7 @@ It receives as its parameter a `SitemapItem` object that can have these properti
 
 * `url` (absolute page URL). This is the only property that is guaranteed to be on `SitemapItem`.
 * `changefreq`
-* `lastmod` (ISO formatted date, `String` type)
+* `lastmod` (`Date`)
 * `priority`
 * `links`.
 


src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -302,7 +302,7 @@ It receives as its parameter a `SitemapItem` object that can have these properti
 
 * `url` (absolute page URL). This is the only property that is guaranteed to be on `SitemapItem`.
 * `changefreq`
-* `lastmod` (ISO formatted date, `String` type)
+* `lastmod` (`Date`)
```
ArmandPhilippot, line 305: The type linked in the issue doesn't seem to be the right one.

The `lastmod` property when using defining the integration (`integrations: [sitemap({lastmod: new Date()})]`) is effectively a `Date` object.

However, here, we're describing the `lastmod` property from `SitemapItem` passed to `serialize`. The type is different! It comes from the  `sitemap` package where [`lastmod` is declared as a `string`](https://github.com/ekalinin/sitemap.js/blob/0af656e6a4a7b1403c9b3af23603261bd9cf94d3/lib/types.ts#L328).

So I think the original version is correct: `lastmod` is ``(ISO formatted date, `String` type)``

**In case anyone wants to check:** In the implementation, [l13](https://github.com/withastro/astro/blob/73a078835eb92a05c3f681ee025c93d6db85b907/packages/integrations/sitemap/src/index.ts#L13) we pick some properties from [SitemapItemLoose imported from the sitemap package](https://github.com/withastro/astro/blob/73a078835eb92a05c3f681ee025c93d6db85b907/packages/integrations/sitemap/src/index.ts#L4C57-L4C73)

sarah11918: Good enough for me! Thank you for checking! I'll close this and the issue. 💜

PR #10598, opened 2025-01-03
[content-collections] quick fixes

sarah11918:
This is the "quick fix" to at least get some minimal helpful/correct info into content collections but does not entirely solve 10147 (not linking so it won't close)

- Adds that collections can be YAML
- notes that we use `micromatch` for matching in `pattern`


ArmandPhilippot (approved):
LGTM!

PR #10611, opened 2025-01-04
Pin Algolia Docsearch to 3.5.1

delucis:
This PR reverts the Algolia Docsearch dependencies to 3.5.1

#10493 updated these to 3.8.2, which @hippotastic and @OliverSpeir reported introduced issues in Safari on iOS.

I’m not 100% sure what in that update would cause this and can’t reproduce myself, so opening this PR to see if a revert would help.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
   "devDependencies": {
     "@11ty/eleventy-fetch": "^5.0.1",
     "@actions/core": "^1.11.1",
-    "@docsearch/css": "^3.8.2",
-    "@docsearch/js": "^3.8.2",
+    "@docsearch/css": "3.5.1",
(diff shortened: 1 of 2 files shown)


OliverSpeir:
https://github.com/algolia/docsearch/issues/2243

delucis:
> [algolia/docsearch#2243](https://github.com/algolia/docsearch/issues/2243)

Ah, good find! Version does 3.5.1 predate that issue (3.6.0 was the latest release at that time), but I guess we’ll see if this helps or not…

OliverSpeir:
This help for me, on iOS 18.2

hippotastic:
This also solves the issue for me. I can scroll down and still open the search in this version.

PR #10612, opened 2025-01-05
Add note about ids having to be strings to content collection docs

louisescher:
#### Description (required)

This PR adds a sentence to clarify that the IDs of content collections always have to be strings. This is undocumented behavior and lead to confusion, see [Custom content loader: `entry missing id` when using an id from the source data](https://discord.com/channels/830184174198718474/1319456296176582727) on the Astro Discord server.

- Related to [this PR on the main repo](https://github.com/withastro/astro/pull/12892)

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -223,7 +223,7 @@ Schemas enforce consistent frontmatter or entry data within a collection through
 
 Schemas also power Astro's automatic TypeScript typings for your content. When you define a schema for your collection, Astro will automatically generate and apply a TypeScript interface to it. The result is full TypeScript support when you query your collection, including property autocompletion and type-checking.
 
-Every frontmatter or data property of your collection entries must be defined using a Zod data type:
+Every frontmatter or data property of your collection entries must be defined using a Zod data type. When defining your schema, make sure that if you provide an `id` property, it is a string:
 
 ```ts title="src/content.config.ts" {6-11,15-19}
 import { defineCollection, z } from 'astro:content';


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -223,7 +223,7 @@ Schemas enforce consistent frontmatter or entry data within a collection through
 
 Schemas also power Astro's automatic TypeScript typings for your content. When you define a schema for your collection, Astro will automatically generate and apply a TypeScript interface to it. The result is full TypeScript support when you query your collection, including property autocompletion and type-checking.
 
-Every frontmatter or data property of your collection entries must be defined using a Zod data type:
+Every frontmatter or data property of your collection entries must be defined using a Zod data type. When defining your schema, make sure that if you provide an `id` property, it is a string:
```
ascorbic, line 226: This should say that an id field is required, so if you provide a schema it must include the id field.

louisescher: @ascorbic Would
```
When defining your schema, make sure that you provide an `id` property and that it is a string:
```
be fine?

sarah11918, line 226: This feels a little off here. This is currently talking about the schema including "Every frontmatter or data property" (ie properties of the data itself) and the example below does not show defining `id`. In fact, when we describe the built in `glob` loader we say `id` is automatically generated, but we say nothing about needing to define it in a schema. So, it's confusing (even to me!) -- obviously you don't always put the `id` in the schema (we didn't for `glob()`. So, when do you need to put it in the schema?

In any case, I think we could intro this a bit better with something like:

```suggestion
Every frontmatter or data property of your collection entries must be defined using a Zod data type. Your content loader may provide this schema for you based on the data you are importing. If not, you must define your own schema as a property of your collection.

The example below shows defining a schema based on frontmatter properties in local Markdown files that are loaded by Astro's built-in `glob()` loader:
```

This adds the idea that some loaders define the schema for you. But, even in the example below, we're defining a schema but *not* `id`.

So it feels weird to describe defining `id` when that's not a part of the basic intro to what schemas are. We need some kind of linkage, something like (this is not correct because I don't know when this is necessary):

> In addition to defining properties of your data entries in your schema, you can type the entry `id` itself as `z.string()`. This may be useful/necessary when the entry is not automatically generated for you by your loader...."

And, maybe this even comes AFTER the code example if it's kind of an edge case?

Again, my issue is that it's confusing to talk about typing `id` when our example does not show doing that, so this can't just be one sentence tacked on but needs some smoothing to make it make sense.

louisescher: @sarah11918 I do agree, it feels off to put this here. However, at no point in the docs is it mentioned that the `ID` must be a string. That's really the intent behind this change, but I had no better spot to put this. Maybe this is just a case of saying nothing would be better since the PR to Astro improves the error?

sarah11918: @louisescher I agree, and looking at the error message, it's less about **defining your `id` as a string` (as in, in the schema), and more about making sure IDs *are* strings, I think?

I think the error message is sufficient and would suggest closing this PR, I think? How about you?

louisescher: I agree with the location you proposed in your edit, that seems like a nice spot to put it. I'll revert the original change and add a sentence to that part, as well as changing the "Example Type: ..." to "Type: `string`". Would that be OK?

sarah11918: If you're going to document a type, then you need @ArmandPhilippot 's permission, not mine! 😅 

Here's the official pattern we use: https://contribute.docs.astro.build/reference/api-references/

ArmandPhilippot: 😄 Well, at first sight, I'd say if `id` is necessarily a string, yes, using `Type: string` instead of the `Example type:` makes sense to me! But, I'm a bit concerned about the location.

If I understand the issue correctly, the problem lies rather at the time of declaring the schema (with a custom loader?) than at the time of receiving the collection entries.
It seems to me that the `CollectionEntry` type matches the returned entries. So I'm not sure it will prevent errors... I'd say the correct location would be [`id` in `DataEntry`](https://docs.astro.build/en/reference/content-loader-reference/#id), but the type is already set to `string`. So I'm a bit confused about what to do. I feel like the docs is already defining the right type... and maybe the error covered by the linked PR is enough to inform the user?

louisescher: The reason I proposed a change to docs in the first place is because I know a lot of people only read the guides and not the actual reference. I felt like having a small sentence clarifying that only strings are accepted might save some headaches but after going through this I agree that the error itself is probably enough. Feel free to close this PR!

sarah11918:
Alright, thanks for everything everyone! I think we've decided that the error message works!

We can always revisit this topic if it turns out the error message isn't sufficient!

PR #10614, opened 2025-01-05
Update build-forms-api.mdx

my-astro:
Heya!
Thanks to @TheOtterlord my component is now working. If you guys think it is appropriate add-on, than small note for newbies like myself will be helpful 🫶

P.s. Feel free to tweak the wording to your taste


src/content/docs/en/recipes/build-forms-api.mdx
```diff
@@ -24,6 +24,8 @@ This recipe shows you how to send form data to an API endpoint and handle that d
     This example sends a JSON object with a message back to the client.
 
     ```ts title="src/pages/api/feedback.ts" "request.formData()" "post"
+    // uncomment the line below if you use static rendering (default)
+    // export const prerender = false;
```
sarah11918, line 28:
```suggestion
    export const prerender = false; // Not needed in 'server' mode
```
We have a standard way of saying this!

sarah11918 (approved):
Thank you @my-astro for taking the time to contribute back the solution that helped you! We really appreciate concern for helping others in the future.

I just made a slight adjustment to the text since we do have a standard way of expressing this in other code examples. But otherwise, this is good to go and we're happy to welcome you to Team Docs! 🥳

PR #10621, opened 2025-01-06
Update page footer to add sponsor link

delucis:
#### Description (required)

- This PR updates our page footer component to replace the feedback widget with a link to sponsor Astro on OpenCollective.
- It also removes the feedback widget from tutorial pages and deletes that code to reduce maintenance.
- While here, icons have been tweaked as well:
  - The contribute link now has a rocket icon instead of an open book
  - The sponsor link uses a heart icon previously used by the community link
  - The community link uses a “multiple people” icon

#### Before

Docs footer with links for Contribute, Feedback, and Community

#### After

Docs footer with links for Contribute, Community, and Sponsor


sarah11918 (approved):
I think this looks great! We haven't been getting much feedback lately, and much of it has been spam. This cleans up the code and I think better aligns the icons!

TheOtterlord (approved):

PR #10649, opened 2025-01-08
Added checklist component to the optional content collection tutorial.

kannansuresh:
#### Description

Added checklist component to the optional content collection tutorial.

Checklist message

**I can use content collections to manage groups of similar content for better performance and organization.**

Discord username: kannansuresh

##### Screenshot

CleanShot 2025-01-08 at 16 51 48@2x

### DIFF
--- a/src/content/docs/en/tutorial/6-islands/4.mdx
+++ b/src/content/docs/en/tutorial/6-islands/4.mdx
@@ -8,6 +8,7 @@ i18nReady: true
 ---
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 import Box from '~/components/tutorial/Box.astro';
+import Checklist from '~/components/Checklist.astro';
 import MultipleChoice from '~/components/tutorial/MultipleChoice.astro';
 import PreCheck from '~/components/tutorial/PreCheck.astro';
 import Option from '~/components/tutorial/Option.astro';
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
This looks great, @kannansuresh ! I'm sure people will appreciate being able to complete the full circle. 😄 

Very nice choice of wording for the checklist, too. Thank you so much for this wonderful contribution, and welcome to Team Docs! 🥳

src/content/docs/en/tutorial/6-islands/4.mdx
```diff
@@ -347,3 +348,11 @@ The tutorial blog project includes an RSS feed. This function must also use `get
     ```
 
 For the full example of the blog tutorial using content collections, see the [Content Collections branch](https://github.com/withastro/blog-tutorial-demo/tree/content-collections) of the tutorial repo.
+
+<Box icon="check-list">
+
+## Checklist
+<Checklist key="content-collections">
```
sarah11918, line 355:
```suggestion
<Checklist>
```
In fact, with only one checklist per page, we don't need any key at all!

sarah11918:
@kannansuresh - It turns out, we don't need a `key` at all if there is only one checklist on a page. (They were in use in an earlier version where there were multiple checklists on the same page.)

So, I'll merge this PR, but I think there is one other place (`6-islands/index`) where key is still used on the checklist component, but isn't necessary.

If you'd like to make a second PR to simply remove that property, that would be really helpful! 🎉

kannansuresh:
> If you'd like to make a second PR to simply remove that property, that would be really helpful! 🎉

Done 
#10655

PR #10671, opened 2025-01-09
[issue template] Remove "other" option

sarah11918:
#### Description (required)
 
On Talking and Doc'ing we decided to streamline the new issue flow and while we'll figure out more improvements, we don't need an "Other"!


sarah11918:
Going to do this all in one PR!

PR #10688, opened 2025-01-10
Updated docs to reflect that getEntry returns undefined

hkbertoson:
#### Description (required)

Updated Content Collections to reflect that `getEntry` returns undefined. 
Feedback is greatly appreciated since I am still fairly new on writing documention.  

- Closes #
- Suggested label:


sarah11918:
Thanks for identifying an issue and bringing it up @hkbertoson 

I'll ask @ascorbic to weigh in here, because I really don't love that our "basic, here's what it looks like" example has error checks and even different ways to do things in it! I don't think the guide page is the place for this unless we absolutely expect all of that to be written every time someone renders content. @ascorbic , is that really the idea?

We *could* have a line or section that mentions you may want to consider error handling/checking for entries, with its *own* separate example. Or, this just links to the reference page where there is a section on how to do this. 

But, this seems wrong as "Here's how you do this!" unless this *really* is "how we expect you to do it, every time, checking for whether things exist first." So I'll need Matt to let me know how he wants to guide people to use the feature! Hang tight Hunter until he gives his guidance! 🙌

hkbertoson:
Sure thing. Thanks! @sarah11918

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -409,12 +409,29 @@ Once queried, you can render Markdown and MDX entries to HTML using the [`render
 import { getEntry, render } from 'astro:content';
 
 const entry = await getEntry('blog', 'post-1');
+if(!entry){
+  // Handle Error
```
sarah11918, line 413:
```suggestion
  // Handle Error, for example:
  throw new Error('Could not find blog post 1');
```

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -409,12 +409,29 @@ Once queried, you can render Markdown and MDX entries to HTML using the [`render
...
+if (!entry) {
+  //Handle Error
+}
+```
+2. Non-Null Assertion Operator
+```ts
+// Use the ! operator to assert the value is non-null
+const { Content, headings, remarkPluginFrontmatter } = await render(entry!);
+```
+
```
sarah11918, line 434:
```suggestion
```
Added context to the example in a minimal way

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -409,12 +409,29 @@ Once queried, you can render Markdown and MDX entries to HTML using the [`render
 import { getEntry, render } from 'astro:content';
 
 const entry = await getEntry('blog', 'post-1');
+if(!entry){
```
sarah11918, line 412:
```suggestion
if (!entry) {
```

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -209,10 +209,31 @@ A function to compile a given entry for rendering. This returns the following pr
 ---
 import { getEntry, render } from 'astro:content';
 const entry = await getEntry('blog', 'entry-1');
+
+
+// Handle undefined return
+if (!entry) {
+  //Handle Error
+}
```
sarah11918, line 217:
```suggestion

if (!entry) {
   // Handle Error, for example:
  throw new Error('Could not find blog post 1');
}
```

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -209,10 +209,31 @@ A function to compile a given entry for rendering. This returns the following pr
...
+```
+2. Non-Null Assertion Operator
+```ts
+// Use the ! operator to assert the value is non-null
+const { Content, headings, remarkPluginFrontmatter } = await render(entry!);
+```
+
```
sarah11918, line 235:
```suggestion
```

sarah11918 (commented):
On Talking and Doc'ing we discussed how to include this in a minimal, helpful way. We think this, plus adding to the type reference that `getEntry()` can returned undefined should be sufficient!

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -209,10 +209,16 @@ A function to compile a given entry for rendering. This returns the following pr
...
+if (!entry) {
+   // Handle Error, for example:
+  throw new Error('Could not find blog post 1');
+}
 const { Content, headings, remarkPluginFrontmatter } = await render(entry);
 ---
 ```
 
+
```
sarah11918, line 221:
```suggestion
```

sarah11918 (approved):
Thanks so much for this great improvement @hkbertoson ! It's so appreciated that experiences from support threads make it into docs! 🙌

louisescher (approved):

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -147,7 +147,7 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
 
 <p>
 
-**Type:** `( collection: string, id: string ) => CollectionEntry<collection> | ({ collection: string, id: string }) => CollectionEntry<collection>`
+**Type:** `((collection: string, id: string) => Promise<CollectionEntry<collection> | undefined>) | (({ collection: string, id: string }) => Promise<CollectionEntry<collection> | undefined>)`
```
Fryuni, line 150: This might be too big to be legible in a single line. Do we have a multi-line syntax for types anywhere that we could replicate here?

delucis: Not really unfortunately. The usual pattern we’d follow instead would be to have something like ``**Type:** [`TypeName`](#typename)`` and show it as a type in a code block with a subheading. But usually that’s for things like options interfaces etc., not just a function type.

sarah11918: @ArmandPhilippot Can you think of an example off the top of your head? On desktop for me, it looks like this:

image

I think one problem is I'm not convinced that each of those two pieces will themselves fit on their own single line, so it's going to look icky either way. But, happy for someone to suggest something that they can figure out that looks good!

sarah11918: Also happy for any maintainer to merge once they either have improved, or resigned themselves to, the above! 😅

ArmandPhilippot: Yeah I see the same thing but... not really, no. Can we drop support for Legacy collections? 😆 Another "solution" would be to use a list under `Type:` so that each is on its own line... (instead of the union) but that's not ideal either.

ArmandPhilippot: I have a little more time than before, so I took the opportunity to take a screenshot to illustrate what I was saying:
Astro getEntry type using a list instead of an union
I don't have a strong opinion on this... I'm not sure what is more readable/understandable.

sarah11918: I mean, I guess that's *better*? What does everyone think? I'm fine with 2 lines (which unfortunately looks like it's gonna be 3 in practice?) Not opposed to the bullet points here!

louisescher: LGTM, way more readable

sarah11918 (approved):
@ArmandPhilippot Happy for you to make the two line change and merge!

src/content/docs/en/reference/modules/astro-content.mdx
ArmandPhilippot, line 150: Since there are no objections:
```suggestion
**Types:**
* `(collection: string, id: string) => Promise<CollectionEntry<collection> | undefined>`
* `({ collection: string, id: string }) => Promise<CollectionEntry<collection> | undefined>`
```

Issue #10532, opened 2024-12-27
The docs and the starter project seem to disagree about putting everything in html tags in layouts.


MikeMcDonald83:
### 📚 Subject area/topic

layouts

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/basics/layouts/

### 📋 Description of content that is out-of-date or incorrect

In the docs about layouts, it says "However, if a layout component does contain a page shell, its <html> element must be the parent of all other elements in the component. All [<style>](https://docs.astro.build/en/guides/styling/#styling-in-astro) or [<script>](https://docs.astro.build/en/guides/client-side-scripts/#using-script-in-astro) elements must be enclosed by the <html> tags."

In the barebones project after 'npm create astro', the Layout.astro file has the style tag outside of the html tag. I don't know if this is actually inconsistent. Maybe the style tag is being used in a different way in the starter. I just noticed it while I was learning Astro.

(If this issue should have gone elsewhere, my bad.)

sarah11918:
Thanks for bringing this up and I'll look into it! I know at one point we had a technical restriction that caused us to add that line to docs, but the basics template is brand new for v5 and maybe that restriction doesn't exist anymore?

sarah11918:
It does seem like this is not currently an issue, so I think the fix is to remove this sentence entirely from the Layouts page!

Will add `help-wanted` and `good first issue` and we would love a PR to remover this sentence/paragraph!

jsparkdev:
This issue has been resolved in #10690.

MikeMcDonald83:
Sorry if I was supposed to do something. I don't know much about PRs and stuff. Seems like it's resolved though. Thanks!

sarah11918:
No need to apologize @MikeMcDonald83 ! You did the right thing by reporting it, then the community can work together on a fix... which we did! All went perfectly as it should. 💜

jsparkdev:
Thank you all. This issue appears to be resolved, so I close it.

PR #10690, opened 2025-01-11
docs: update `layouts.mdx`

jsparkdev:
#### Description (required)

- update `layouts.mdx`
- removed the wording that doesn't match the current version of Astro and modified the example code blocks.
- #10532

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/basics/layouts.mdx
+++ b/src/content/docs/en/basics/layouts.mdx
@@ -14,7 +14,7 @@ We conventionally use the term "layout" for Astro components that provide common
 
  But, there is nothing special about a layout component! They can [accept props](/en/basics/astro-components/#component-props) and [import and use other components](/en/basics/astro-components/#component-structure) like any other Astro component. They can include [UI frameworks components](/en/guides/framework-components/) and [client-side scripts](/en/guides/client-side-scripts/). They do not even have to provide a full page shell, and can instead be used as partial UI templates.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/basics/layouts.mdx
```diff
@@ -45,12 +45,13 @@ const { title } = Astro.props;
     </article>
     <Footer />
   </body>
-  <style>
-    h1 {
-      font-size: 2rem;
-    }
-  </style>
 </html>
+
```
sarah11918, line 49: Let's NOT change this to be outside of the `<html>` tag, even though we discovered it's possible. Just because it *can* be done, doesn't mean that it should! 😅 

I think our main example should still show the most common, supported use case!

sarah11918 (commented):
Thanks @jsparkdev ! Removing the sentence is perfect, but I commented below that I don't think we should change our example from the common use!

src/content/docs/en/basics/layouts.mdx
jsparkdev: Good idea. I've fixed it.

sarah11918 (approved):
Perfect, thank you! 💜

sarah11918:
Ugh, will look into the failed Netlify deploy and see whether it was just a fluke!

jsparkdev:
It seems to be working now. Thank you!

PR #10704, opened 2025-01-14
feat: document external redirects

ematipico:
#### Description (required)

This PR documents the upcoming new feature, where external redirects are supported.


sarah11918 (commented):
Thanks @ematipico ! I made a commit here because it was otherwise awkward to suggest here in GitHub, but I don't think this needs to be a standalone section (where it's easy to read all the stuff saying "redirects are to another route...." for lots of paragraphs and then see, "Oh, or you can do this."

See what you think of this edit, and can we work with something like this?

src/content/docs/en/guides/routing.mdx
```diff
@@ -291,6 +298,24 @@ When running `astro build`, Astro will output HTML files with the [meta refresh]
...
+  redirects: {
+    '/blog': "https://example.com/blog"
+    '/news': {
+      status: 302,
+      destination: "https://example.com/news"
+    }
+  }
+});
+```
+
```
sarah11918, line 318:
```suggestion
```
If we think the above works, I would not have this section on its own.

sarah11918 (approved):
Approving for the next minor release! :tada:

yanthomasdev (approved):

Issue #10703, opened 2025-01-14
Provide documentation on generating custom IDs using the glob loader


MoritzLost:
### 📚 Subject area/topic

Content Collections

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/content-collections/

### 📋 Description of content that is out-of-date or incorrect

The `glob` loader generates IDs automatically based on the file path, but transforms those to lowercase and applies some other custom replacements:

https://github.com/withastro/astro/blob/c30070b9271e4c494e7cbf3a1c45515782034911/packages/astro/src/content/loaders/glob.ts#L35-L46
https://github.com/withastro/astro/blob/1072c76703b725919cf52f63cf254a3fba90c030/packages/astro/src/content/utils.ts#L403-L408

This is a breaking change from the behaviour of the `slug` property in Astro 4.

The `glob` loader accepts a `generateId` function to customize the way those IDs are generated, but that argument isn't documented anywhere. The page on Content Collections should be expanded:

- Mention how the ID is generated by default (e.g. that it uses GitHub Slugger), and maybe what caveats come with that (for example, that uppercase will be transformed to lowercase by default).
- Mention the possibility to use a custom `generateId` function, preferably with some example code.

For example, to retain uppercase letters in the ID, I used this:

```js
loader: glob({
    pattern: '**/*.json',
    base: './src/content/foo',
    generateId: ({ entry }) => entry.replace(/\.json$/, ''),
}),
```

PR #10710, opened 2025-01-15
Content Loader API: include docs for the built-in loaders

### LINKED ISSUE #10147: Improve built-in loader docs
### 📋 Explain your issue

Right now we document `glob()` and `file()` in a single [“Built-in loaders” subheading](https://docs.astro.build/en/guides/content-collections/#built-in-loaders), but we don’t have dedicated docs anywhere for each of these. Importantly this means we’re not documenting the glob pattern stuff in detail, including what pattern matching looks like.

#### Example use case

A user was trying to match a pattern that included a `+`. I suspect that may be interpreted in a specific way by the glob implementation, but docs don’t show or link to how glob patterns work.

ArmandPhilippot:
#### Description (required)

Updates the [Content Loader API page](https://docs.astro.build/en/reference/content-loader-reference/) to include a new section for the built-in loaders ([`glob`](https://github.com/withastro/astro/blob/main/packages/astro/src/content/loaders/glob.ts) and [`file`](https://github.com/withastro/astro/blob/main/packages/astro/src/content/loaders/file.ts)):
* provide information for the different properties including `generateId` which was missing in the docs,
* include examples with different use-cases for each loader
* add links from the Content Collection guide to the new reference entries
* add YAML support on both pages for the glob loader

- Closes #10147, #10703


src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -495,3 +495,137 @@ The format of the `RenderedContent` object is:
...
+
+**Type:** `(options: GenerateIdOptions) => string`
+</p>
+
+A callback function used to generate an ID for an entry. It accepts an object as parameter with the following properties:
+* `entry` - the path to the entry file, relative to the base directory
+* `base` - the base directory [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL)
+* `data` - the parsed, unvalidated data of the entry
+
+The function should return a unique string per entry in a collection.
```
ArmandPhilippot, line 572: I wonder if we should add more details about the implementation to explain why the ID should/must be unique ([Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)).

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -495,3 +495,137 @@ The format of the `RenderedContent` object is:
...
+```
+
+#### `fileName`
+
+<p>
+
+**Type:** `string`
+</p>
+
+Sets the path to the file to load, relative to the content directory.
```
ArmandPhilippot, line 613: I used the [JSDoc comment](https://github.com/withastro/astro/blob/0968069aaf509cea67e2bc2b03b1265327d14451/packages/astro/src/content/loaders/file.ts#L19) but I wonder if this is accurate: `relative to the content directory`. Based on the examples from the [Content Collections guide](https://docs.astro.build/en/guides/content-collections/#defining-the-collection-loader), it seems to me the path is relative to the root directory instead. Which makes more sense, especially since the content directory is no longer mandatory...

sarah11918:
Sorry I've been busy the last couple of days and haven't replied, but this is fantastic! I thought I was going to have to get on this myself today or tomorrow. 😄 

Was thinking we could look at this on Talking and Doc'ing tomorrow! 🎉

sarah11918:
Discussed on talking and doc'ing:

Content and comprehensiveness are excellent! This is the content we want, just need to incorporate it into the rest of docs fully:

- move this section to after What is a loader?
- following, is geared towards creating your own loader (vs built in)
- rename those following headings e.g. Object loader API
- link to these items from content collections guide now that we have reference items
- still general text edits to be made (e.g. `pattern` should mention glob)
- micromatch docs aren't super great -- should we link to a glob playground too (check original issues for resources) so they can confirm their `pattern` is valid? (getting the syntax perfect is now more important with collections v5, so the more guidance the better)

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -14,6 +14,140 @@ Astro loaders allow you to load data into [content collections](/en/guides/conte
...
```
ArmandPhilippot, line 131: I used the [JSDoc comment](https://github.com/withastro/astro/blob/0968069aaf509cea67e2bc2b03b1265327d14451/packages/astro/src/content/loaders/file.ts#L19) but I wonder if this is accurate: relative to the content directory. Based on the examples from the [Content Collections guide](https://docs.astro.build/en/guides/content-collections/#defining-the-collection-loader), it seems to me the path is relative to the root directory instead. Which makes more sense, especially since the content directory is no longer mandatory...

ArmandPhilippot:
I think I made all the updates. I posted again one of my original comments (content has moved) because I'm still not sure this is accurate. Other than that, I think it's ready to you to check the wording when you'll have time @sarah11918! I have reworded some sentences mentioned during T&D but... well, I'm not the expert here. :smile:

**Note:** I had to update a translation to fix a link: `zh-cn/reference/modules/astro-content.mdx`.

src/content/docs/en/reference/content-loader-reference.mdx
sarah11918: I think you're right here!

liruifengv:
`zh-cn` LGTM! I will update this link after `Content Loader reference` was translated.

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Okay since we agree I think it's safe to update:
```suggestion
Sets the path to the file to load, relative to the root directory.
```

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -83,11 +83,11 @@ The Content Layer API allows you to fetch your content (whether stored locally i
...
 
-The `glob()` loader creates entries from directories of Markdown, MDX, Markdoc, or JSON files from anywhere on the filesystem. It accepts a `pattern` of entry files to match using glob patterns supported by [micromatch](https://github.com/micromatch/micromatch#matching-features), and a base file path of where your files are located. Each entry's `id` will be automatically generated from its file name. Use this loader when you have one file per entry.
+The [`glob()` loader](/en/reference/content-loader-reference/#glob-loader) creates entries from directories of Markdown, MDX, Markdoc, JSON or YAML files from anywhere on the filesystem. It accepts a `pattern` of entry files to match using glob patterns supported by [micromatch](https://github.com/micromatch/micromatch#matching-features), and a base file path of where your files are located. Each entry's `id` will be automatically generated from its file name. Use this loader when you have one file per entry.
```
sarah11918, line 88:
```suggestion
The [`glob()` loader](/en/reference/content-loader-reference/#glob-loader) creates entries from directories of Markdown, MDX, Markdoc, JSON, or YAML files from anywhere on the filesystem. It accepts a `pattern` of entry files to match using glob patterns supported by [micromatch](https://github.com/micromatch/micromatch#matching-features), and a base file path of where your files are located. Each entry's `id` will be automatically generated from its file name. Use this loader when you have one file per entry.
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -14,6 +14,140 @@ Astro loaders allow you to load data into [content collections](/en/guides/conte
...
+
+### `glob()` loader
+
+<p>
+
+**Type:** <code>(options: GlobOptions) => <a href="#the-loader-object">Loader</a></code><br />
+<Since v="5.0.0" />
+</p>
+
+The `glob()` loader creates entries from directories of files from anywhere on the filesystem. The supported file types are Markdown, MDX, Markdoc, JSON and YAML files.
```
sarah11918, line 29:
```suggestion
The `glob()` loader creates entries from directories of files from anywhere on the filesystem. The supported file types are Markdown, MDX, Markdoc, JSON, and YAML files.
```

sarah11918 (approved):
Just a couple of Oxford commas, but this looks outstanding, Armand! 🥳  Thank you so much for taking this on! 💪

ArmandPhilippot:
Wow, thanks @sarah11918! I wasn't sure my rewording would be enough. 🙌🏽 

Do we need a Lunaria directive here? I'm asking because I updated a translation and while I've observed how things have been done in the past, I'm not sure I understand when it's necessary/how it works.

yanthomasdev:
> Do we need a Lunaria directive here?


Yes, @ArmandPhilippot! The zh-cn translation of the page isn't 100% updated with English, so we'd like to keep the outdated status. If it were completely updated, it wouldn't be necessary. 

Here's the directive you can paste into the commit body during merging:
```
@lunaria-track:src/content/docs/en/guides/content-collections.mdx;src/content/docs/en/reference/content-loader-reference.mdx;src/content/docs/en/reference/modules/astro-content.mdx
```

ArmandPhilippot:
Ok, so basically in these cases I just have to remember to list the paths of the English versions separated by a semicolon (with the prefix of course). Thanks for the explanations, @yanthomasdev! 🫡 (I saw these directives, but didn't look at them more closely before... 😅 )

Well, let's see! 🤞🏽

PR #10715, opened 2025-01-16
Make `implementation PR` can be translated

liruifengv:

- Closes #
- Suggested label:


delucis (commented):
Thank you for this @liruifengv!

Looks great but I noticed one small detail. The Chinese version exhibits an extra wide gap between the `:` and the PR title:

image

This is because the Chinese `：` character already includes built-in whitespace but the `SourcePR.astro` template _also_ includes a space between the UI string and the PR title, so we end up with `：` + ` `.

I think maybe we want to remove that space from the `SourcePR.astro` template and include it in the translated UI strings when necessary?

liruifengv:
Good idea. I do that.

sarah11918 (approved):

PR #10120, opened 2024-11-30
fix: make tutorial consistent with `blog-tutorial-demo` (Typescript errors)

ArmandPhilippot:
#### Description (required)

Fixes two more Typescript errors in the unit 6 of the tutorial. (see https://github.com/withastro/blog-tutorial-demo/pull/33)

#### Related issues & labels (optional)

- Suggested label: code snippet update

PR #10719, opened 2025-01-16
Removed redundant nullish coalescing operator

chasoft:
#### Description (required)

This PR removes the redundant nullish coalescing operator (??) from the code that retrieves the theme from localStorage. The nullish coalescing operator is unnecessary in this context because the existence of the theme in localStorage is already checked.

- Suggested label: enhancement

### DIFF
--- a/src/content/docs/en/tutorial/6-islands/2.mdx
+++ b/src/content/docs/en/tutorial/6-islands/2.mdx
@@ -104,10 +104,11 @@ To add interactivity to an Astro component, you can use a `<script>` tag. This s
       :global(.dark) .moon { fill: white; }
     </style>
 
-    <script is:inline>
+    <script>
       const theme = (() => {
-        if (typeof localStorage !== 'undefined' && localStorage.getItem('theme')) {
-          return localStorage.getItem('theme');
+        const localStorageTheme = localStorage?.getItem("theme") ?? '';
+        if (['dark', 'light'].includes(localStorageTheme)) {
+          return localStorageTheme;
         }
         if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
           return 'dark';
--- a/src/content/docs/en/tutorial/6-islands/2.mdx
+++ b/src/content/docs/en/tutorial/6-islands/2.mdx
@@ -107,7 +107,7 @@ To add interactivity to an Astro component, you can use a `<script>` tag. This s
     <script is:inline>
       const theme = (() => {
         if (typeof localStorage !== 'undefined' && localStorage.getItem('theme')) {
-          return localStorage.getItem('theme') ?? "light";
+          return localStorage.getItem('theme');
         }
         if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
           return 'dark';
--- a/src/content/docs/en/tutorial/6-islands/2.mdx
+++ b/src/content/docs/en/tutorial/6-islands/2.mdx
@@ -104,7 +104,7 @@ To add interactivity to an Astro component, you can use a `<script>` tag. This s
       :global(.dark) .moon { fill: white; }
     </style>
 
-    <script>
+    <script is:inline>
       const theme = (() => {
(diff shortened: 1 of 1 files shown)


ArmandPhilippot:
Hi @chasoft, and thanks for taking the time to fix the docs!

I'm a bit confused here because I remember adding that (in #10120) because Typescript was complaining without. So, I tested again to be sure it was safe to remove the nullish coalescing operator (some changes might have happen since), and it doesn't seem to be the case every time.

First I used the default branch (`complete`) of the [blog-tutorial-demo](https://github.com/withastro/blog-tutorial-demo). This branch doesn't use the nullish coalescing operator and I don't get any Typescript errors. I was surprised.

Then I remembered that we were working on the `content-collections` branch (and we've updated the tutorial based on that branch) so I switched. Now, `src/components/ThemeIcon.astro` uses the operator. If I remove it, then Typescript complains:

Typescript error ts(2345) on window.localStorage.setItem

To be honest, I don't understand why there is a difference between the two branches. This is the same component. But since the error may be raised, I'm not sure it's safe to remove the operator from the docs...

delucis:
Don't know if it helps but in general I'd avoid the double call to get stuff from local storage and do something like this instead: 
```js
if (typeof localStorage !== 'undefined') {
  const theme = localStorage.getItem('theme');
  if (theme) return theme; 
}
```

ArmandPhilippot:
I was just confused about Typescript behavior on two files that seem identical (once the operator removed)... Refactoring didn't come to my mind. 😄 

But yes, this is another solution and it might also be easier to understand while preventing Typescript errors/nullish coalescing operator.

ArmandPhilippot:
So Fryuni has found the difference between the two branches. 🙌🏽  One uses `is:inline` while the other not. Same as the tutorial... I missed that, sorry. So, you're right it's safe to remove that here because Typescript will not parse that code!

But I think Chris' suggestion is a good idea to prevent the error if someone forgot the `is:inline` directive while following the tutorial. Do you want to make the update @chasoft?

chasoft:
> So Fryuni has found the difference between the two branches. 🙌🏽 One uses `is:inline` while the other not. Same as the tutorial... I missed that, sorry. So, you're right it's safe to remove that here because Typescript will not parse that code!
> 
> But I think Chris' suggestion is a good idea to prevent the error if someone forgot the `is:inline` directive while following the tutorial. Do you want to make the update @chasoft?

Thank you for your response. I'll take another look at this by tomorrow.

src/content/docs/en/tutorial/6-islands/2.mdx
```diff
@@ -104,10 +104,11 @@ To add interactivity to an Astro component, you can use a `<script>` tag. This s
       :global(.dark) .moon { fill: white; }
     </style>
 
-    <script is:inline>
+    <script>
```
ArmandPhilippot, line 107: We still need `is:inline` to prevent anything to render until it finishes executing (so to avoid [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content)), the `content-collections` branch is missing that:
```suggestion
    <script is:inline>
```

delucis:
I think we do want to keep the `is:inline` attribute or did I misunderstand?

`is:inline` ensures the scripts runs _where it was written_ instead of being bundled and executed as an async module. Bundling this script would cause a flash of unstyled content, so the `is:inline` usage is intentional.

ArmandPhilippot (commented):
Thanks @chasoft, your refactoring works with and without Typescript parsing so looks good to me!
I only left a suggestion to add back `is:inline` which is needed to prevent FOUC (we miss that in the `content-collection` branch).
After that, looks ready to merge! Welcome to Team Docs! 🚀 

And you're very welcome to apply those changes to the `blog-tutorial-demo`. Thanks for willing to do that! 🙌🏽

ArmandPhilippot (approved):
Thanks for the update! LGTM!

PR #10722, opened 2025-01-16
i18n(fr): Update `guides/fonts.mdx` from #10717

thomasbnt:
#### Description (required)
Update `guides/fonts.mdx` from #10717

```diff
+ sidebar:
+   label: Fonts
``` 

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/guides/fonts.mdx
+++ b/src/content/docs/fr/guides/fonts.mdx
@@ -1,7 +1,7 @@
 ---
 title: Utiliser des polices personnalisées
 sidebar:
-  label: Fonts
+  label: Polices d'écriture
 description: >-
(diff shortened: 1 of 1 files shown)


pioupia (approved):
LGTM!

src/content/docs/fr/guides/fonts.mdx
```diff
@@ -1,5 +1,7 @@
 ---
 title: Utiliser des polices personnalisées
+sidebar:
+  label: Fonts
```
ArmandPhilippot, line 4: Personally I would have translated it. Especially since we're using the translated version in the page:
```suggestion
  label: Polices d'écriture
```

PR #10730, opened 2025-01-19
More links to tutorials for beginners in docs

mseibert:
Added other tutorials for beginners. It includes links to paid offerings. But I am in no way affiliated with any of these and genuinely believe that they do add value for readers.

#### Description (required)

I need to educate our employees in the company with Astro. We were a WordPress shop so far. So I have been searching for offerings with videos for them to onboard themselves. This is good content that all other newbies in the Astro community could benefit from. That's why I suggest to add it to the official docs.

- Suggested label: docs, tutorials, videos

Username in Astro Discord: martinseibert (legacy: mseibert#0410)

### DIFF
--- a/src/content/docs/en/tutorial/6-islands/3.mdx
+++ b/src/content/docs/en/tutorial/6-islands/3.mdx
@@ -71,7 +71,12 @@ Congratulations on completing the Astro blog tutorial! Share your achievement wi
 
 ## Next Steps
 
-You can enhance this project's final code with one of our tutorial extensions, or start your next Astro project!
+There are many more tutorials out there: 
+   1. Free YouTube videos: [1](https://youtu.be/e-hTm5VmofI?si=UHCR_OwX4YZXIKqf), [2](https://youtu.be/XoIHKO6AkoM?si=yMEREmelpz9Q6U3O) 
+   2. Paid online courses: [1](https://www.udemy.com/course/astrojs-101-build-blazing-fast-frontends/), [2](https://learnastro.dev/), [3](https://astrocourse.dev/), [4](https://www.udemy.com/course/astro-the-complete-guide/), [5](https://www.udemy.com/course/astrojs-101-build-blazing-fast-frontends/)
+
+Or you can enhance this project's final code with one of our tutorial extensions, or start your next Astro project!
+
 <CardGrid>
   <LinkCard
     title="Start a new Astro Project"


mseibert:
The changes can be seen here: https://deploy-preview-10730--astro-docs-2.netlify.app/en/tutorial/6-islands/3/

src/content/docs/en/tutorial/6-islands/3.mdx
```diff
@@ -71,7 +71,12 @@ Congratulations on completing the Astro blog tutorial! Share your achievement wi
 
 ## Next Steps
 
-You can enhance this project's final code with one of our tutorial extensions, or start your next Astro project!
+There are many more tutorials out there: 
+   1. Free YouTube videos: [1](https://youtu.be/e-hTm5VmofI?si=UHCR_OwX4YZXIKqf), [2](https://youtu.be/XoIHKO6AkoM?si=yMEREmelpz9Q6U3O)
```
trueberryless, line 75:
```suggestion
   - Free YouTube videos: [Astro Web Framework Crash Course](https://youtu.be/e-hTm5VmofI?si=UHCR_OwX4YZXIKqf), [Astro Quick Start Course | Build an SSR Blog](https://youtu.be/XoIHKO6AkoM?si=yMEREmelpz9Q6U3O) 
```

First, because the list doesn't have to do anything with ordering, I strongly recommend using an unordered list.

Second, I think we should not just list them as 1, 2 without any context what's behind the link. At least include the title of the videos.

src/content/docs/en/tutorial/6-islands/3.mdx
```diff
@@ -71,7 +71,12 @@ Congratulations on completing the Astro blog tutorial! Share your achievement wi
 
 ## Next Steps
 
-You can enhance this project's final code with one of our tutorial extensions, or start your next Astro project!
+There are many more tutorials out there: 
+   1. Free YouTube videos: [1](https://youtu.be/e-hTm5VmofI?si=UHCR_OwX4YZXIKqf), [2](https://youtu.be/XoIHKO6AkoM?si=yMEREmelpz9Q6U3O) 
+   2. Paid online courses: [1](https://www.udemy.com/course/astrojs-101-build-blazing-fast-frontends/), [2](https://learnastro.dev/), [3](https://astrocourse.dev/), [4](https://www.udemy.com/course/astro-the-complete-guide/), [5](https://www.udemy.com/course/astrojs-101-build-blazing-fast-frontends/)
```
trueberryless, line 76:
```suggestion
   - Paid online courses: [1](https://www.udemy.com/course/astrojs-101-build-blazing-fast-frontends/), [2](https://learnastro.dev/), [3](https://astrocourse.dev/), [4](https://www.udemy.com/course/astro-the-complete-guide/), [5](https://www.udemy.com/course/astrojs-101-build-blazing-fast-frontends/)
```

Same applies here. I would never click on any of those links if they look like this: 1, 2, 3, 4, 5

Like where should I start? What can I expect? I suggest that you at least change the numbers to some context.

trueberryless (changes requested):
Generally speaking, it feels not very Astro-ish to link to paid stuff, but please correct me if the docs sometimes do that.

Looking a little closer, I have to request some changes (the necessary ones can be seen below).

Overall, I also suggest that you don't just list the items, but rather use the CardGrid component to present them in a better way (you can scroll down the file to see examples).

And last but not least, the first sentence after Next steps feels to harsh, give more context. Just writing, there are a lot resources feels very distant from the reader imho...

src/content/docs/en/tutorial/6-islands/3.mdx
mseibert: I get your point. Above you said that it's not "Astro-ish" to link paid offerings. That's why I boiled it down to numbers. I do get the link / connection to a next course. But I do not offer any big SEO-keyword inheriting Google-juice. It's not super active, yet it's there. That was my approach.

mseibert:
> Generally speaking, it feels not very Astro-ish to link to paid stuff, but please correct me if the docs sometimes do that.
I get the point. Yet, the content is valuable. I have spent quite some time doing research and trying these.
 
> Overall, I also suggest that you don't just list the items, but rather use the CardGrid component to present them in a better way (you can scroll down the file to see examples).
How about a new page with video courses. Then we could also get rid of the short 1, 2, 3, because it would have room to be seen. And the likeliness of others adding new stuff as it appears would be helpful as well.

> And last but not least, the first sentence after Next steps feels to harsh, give more context. Just writing, there are a lot resources feels very distant from the reader imho...
I get the point. Will update this.

src/content/docs/en/tutorial/6-islands/3.mdx
mseibert: - Okay, an unordered list it will be. No problem.
- I did not want to give the linked pages to much "Google juice" but I agree that some words could make it simpler for the reader as well. Will change that.

sarah11918:
> How about a new page with video courses.

Hi! we did in fact have an entire page in the v4 version of the docs, but for a variety of reasons it was an intentional decision to remove this. So we won't be adding content like this back at this time, but we hope you and your team are having a great time learning Astro! 🚀

mseibert:
Thanks for clarifying! Understood and appreciate the quick response.

PR #10731, opened 2025-01-19
Update v5.mdx

severindupouy:
Add more explicit instructions about replacing `slug` by `id`

#### Description (required)

For non-English speakers like me, the instructions given didn't seem explicit enough. I spent a couple of hours before I realized that I also had to rename the files.

- Suggested label:  improve or update documentation

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v5.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v5.mdx
@@ -141,30 +141,22 @@ See the instructions below for updating an existing content collection (`type: '
     });
     ```
 
-3. **Change references from `slug` to `id`**. Content layer collections do not have a reserved `slug` field. Instead, all updated collections will have an `id`. You may also need to **update dynamic routing file names** to match a changed getStaticPaths() parameter:
+3. **Change references from `slug` to `id`**. Content layer collections do not have a reserved `slug` field. Instead, all updated collections will have an `id`:
 
-    - rename dynamic routing files
-  
-      ```sh
-      git mv src/pages/[...slug].astro src/pages/[id].astro
-      ```
-  
-    - use `id` in `getStaticPaths` functions :
-  
-      ```astro ins={7} del={6}
-      // src/pages/[id].astro
-  
-      ---
-      export async function getStaticPaths() {
-        const posts = await getCollection('blog');
-        return posts.map((post) => ({
-          params: { slug: post.slug },
-          params: { id: post.id },
-          props: post,
-        }));
-      }
-      ---
-      ```
+    ```astro ins={7} del={6}
+    // src/pages/[slug].astro
+    ---
+    export async function getStaticPaths() {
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -141,21 +141,30 @@ See the instructions below for updating an existing content collection (`type: '
...
+      export async function getStaticPaths() {
+        const posts = await getCollection('blog');
+        return posts.map((post) => ({
+          params: { slug: post.slug },
+          params: { id: post.id },
+          props: post,
+        }));
+      }
+      ---
+      ```
```
ArmandPhilippot, line 167: Hi! Thanks for taking the time to make the docs better!
Would the following suggestion have been clearer?
````suggestion
3. **Change references from `slug` to `id`**. Content layer collections do not have a reserved `slug` field. Instead, all updated collections will have an `id`:

    ```astro ins={7} del={6}
    // src/pages/[slug].astro
    ---
      return posts.map((post) => ({
        params: { slug: post.slug },
        params: { slug: post.id },
        props: post,
      }));
    }
    ---
    ```
You can also update the dynamic routing file names to match the value of the changed getStaticPaths() parameter.
````
Renaming the dynamic routing file is not necessary. You can still use `[slug]` or `[...slug]`, only the value inside `params` must be updated because the content collection no longer return `post.slug` but `post.id`.

I don't know if this is better. However, I feel like introducing an upgrade step that isn't necessary (`git mv`) might be more confusing for other people.

Comment (severindupouy) [reply src/content/docs/en/guides/upgrade-to/v5.mdx:167]: Hi,
You're welcome.
your suggestion is ok for me. Indeed maybe it's clearer.

Should I modify my PR ?

File: src/content/docs/en/guides/upgrade-to/v5.mdx [hunk 2]
```diff
@@ -141,21 +141,22 @@ See the instructions below for updating an existing content collection (`type: '
...
       return posts.map((post) => ({
         params: { slug: post.slug },
-        params: { id: post.id },
+        params: { slug: post.id },
         props: post,
       }));
     }
     ---
     ```
+You can also update the dynamic routing file names to match the value of the changed getStaticPaths() parameter.
```
ArmandPhilippot, line 159: Sorry, it seems I poorly formatted my suggestion. I think it's an indentation issue:
```suggestion
    You can also update the dynamic routing file names to match the value of the changed getStaticPaths() parameter.
```
So I'm going to commit this change so that the deployment preview can work.

That's said, I still want another look at this so there probably won't be any approval or changes by tomorrow (it's still the weekend!).

ArmandPhilippot (approved):
Thanks for your patience and help in improving the docs, LGTM! Welcome to Team Docs! 🎉 
I'm just adding one last commit to add backticks around a function.

PR #10741, opened 2025-01-21
i18n(fr): Update `reference/modules/astro-content.mdx` from #10688

thomasbnt:
#### Description (required)
Update `reference/modules/astro-content.mdx` from #10688

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/reference/modules/astro-content.mdx
+++ b/src/content/docs/fr/reference/modules/astro-content.mdx
@@ -147,7 +147,9 @@ const draftBlogPosts = await getCollection('blog', ({ data }) => {
 
 <p>
 
-**Type :** `( collection: string, id: string ) => CollectionEntry<collection> | ({ collection: string, id: string }) => CollectionEntry<collection>`
+**Types :**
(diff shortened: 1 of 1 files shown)


src/content/docs/fr/reference/modules/astro-content.mdx
```diff
@@ -209,6 +211,11 @@ Une fonction permettant de compiler une entrée donnée pour le rendu. Elle renv
 ---
 import { getEntry, render } from 'astro:content';
 const entry = await getEntry('blog', 'entry-1');
+
+if (!entry) {
+   // Manipuler l'erreur, par exemple :
```
ArmandPhilippot, line 216: nit: What do you think of:
```suggestion
   // Gérer l'erreur, par exemple :
```
I feel like `manipuler` means we get an error (i.e. an object) and use it whereas here we just need to handle the missing entry (we could use a redirect). Not sure if I'm clear. 😅

thomasbnt: yeah good!

PR #10790, opened 2025-01-24
docs: update `guides/deploy/cloudflare.mdx`

vrabe:
#### Description (required)

The steps to deploy with git are still strange.

1. All these steps are to set up a new project on Cloudflare Pages. Step 1 seems redundant.
2. It seems that selecting an account isn't done in "Workers" page. I can't verify that because I don't have multiple accounts.
3. We don't have to explicitly select "Overview". It shows up after clicking "Workers & Pages".
4. "Create application" button seems to be renamed to "Create":

cloudflare

- Closes #


sarah11918:
Thank you for contributing this! I will ask for feedback from others who use Cloudflare to confirm the steps! Perhaps @alexanderniebuhr is able to verify here?

alexanderniebuhr:
I think for users with multiple accounts this is different. I'll check tomorrow my time :)

hkbertoson:
Just checked it on my Cloudflare Account with Multiple accounts and it looks just fine to me.

src/content/docs/en/guides/deploy/cloudflare.mdx
```diff
@@ -27,13 +27,13 @@ To get started, you will need:
...
 
 <Steps>
-1. Set up a new project on Cloudflare Pages.
+1. Push your code to your git repository (GitHub, GitLab).
 
-2. Push your code to your git repository (GitHub, GitLab).
+2. Log in to the Cloudflare dashboard and select your account.
 
-3. Log in to the Cloudflare dashboard and select your account in **Account Home** > **Compute (Workers)** > **Workers & Pages** > **Overview**.
+3. Go to **Account Home** > **Compute (Workers)** > **Workers & Pages**.
```
jsparkdev, line 34: I don't think you need to go to the `Account Home` menu, let me know if you have any other thoughts.

```suggestion
3. Go to **Compute (Workers)** > **Workers & Pages**.
```

alexanderniebuhr:
Yeah as I said, for multi accounts you need to select the account after login, but I think that is already described. However I just attached the Cloudflare Dashboard flow as a screencast, for others to compare with the written steps, hope that helps:


https://github.com/user-attachments/assets/138fda2d-551b-400b-9a58-47fe4ce9474a

src/content/docs/en/guides/deploy/cloudflare.mdx
vrabe: @jsparkdev You're right.  

In [Cloudflare Docs](https://developers.cloudflare.com/pages/get-started/git-integration/#connect-your-git-provider-to-pages):

> Pages offers support for [GitHub](https://github.com/) and [GitLab](https://gitlab.com/). To create your first Pages project:
>
> 1. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com/) and select your account.
> 2. In Account Home, select **Workers & Pages**.
> 3. Select **Create application** > **Pages** > **Connect to Git**.

It seems that after login/selecting an account, you are in Account Home. Do we need to point out this?

```suggestion
3. In Account Home, select **Compute (Workers)** > **Workers & Pages**.
```

Or just apply your change?

jsparkdev: @vrabe 

It seems right to describe it in the official documentation :)

sarah11918:
Thanks for jumping in everyone!!

I will note, this is why we try to describe *other* services, dashboards etc. as **generally as possible** because we can't control (and often don't know) when these update. As evidenced by the maintenance burden this is causing now.  😅 

I would be fine with a PR suggestion that describes what you need to do *accurately*, but I would prefer as much of a generic, future proof solution as possible.

e.g. If you have to set something on "Workers and Pages", then it's not *our* job to say exactly how to get there. Their own docs should explain where Workers and Pages is in their UI and how to get there. In fact, if there's no guarantee that that setting will always even be found at "Workers and Pages", it is most future proof to just specify the setting that needs configuring, and someone can find that setting *through Cloudflare's own documentation* about setting that value. The frequency with which a setting, action, or option name changes is much lower than the frequency with which the dashboard or UI could get updated or rearranged.

So with that in mind, I'll ask you to give a fresh look over these instructions and see what, if anything, can be generalized for future-proof instructions! It's not our job to teach Cloudflare's current dashboard/UI -- only to let Cloudflare users know what must be configured to work with an Astro project!

vrabe:
If we want it general, maybe we need to change it back to the original style. Something like:

```
<Steps>
1. Push your code to your git repository (GitHub, GitLab).

2. Set up a new project on Cloudflare Pages. Use the following build settings:

    - **Framework preset**: `Astro`
    - **Build command:** `npm run build`
    - **Build output directory:** `dist`

3. Click the **Save and Deploy** button.
</Steps>
```

sarah11918:
That honestly looks amazing to me! Does that satisfy the original need for updating this page?

vrabe:
I don't mind changing to that.

sarah11918 (approved):
SHIP. IT. 🥳

PR #10803, opened 2025-01-25
i18n(es) Update markdoc.mdx

AitorMT:
Removed `Configura el elemento HTML raíz`. This section does not exist in English


yanthomasdev (changes requested):
Thanks @AitorMT, but I see there's a few more differences between English and Spanish here, could you take a look at this list of commits and see the changes that weren't applied? https://github.com/withastro/docs/commits/main/src/content/docs/en/guides/integrations-guide/markdoc.mdx?since=2024-05-20T04:50:47.000Z

Not sure if you're aware, but you can always check the [Astro i18n status](https://i18n.docs.astro.build/) to get a proper list of missing commits!

AitorMT:
I did not know about the existence of this page, thank you very much for providing it, as soon as I have a moment I will update the missing information.

AitorMT:
I have reviewed the changes and in principle everything is already updated, I don't know if I am missing something.

yanthomasdev (approved):
LGTM! Thanks!

PR #10810, opened 2025-01-26
Update middleware.mdx

ruohki:
#### Description (required)

Updated the env.d.ts section to reflect the current working implementation

discord: ruohki


src/content/docs/en/guides/middleware.mdx
```diff
@@ -141,13 +141,15 @@ export const onRequest = (context, next) => {
...
+    namespace App {
+        interface Locals {
+            user: {
+                name: string
+            },
+            welcomeTitle: () => string,
+            orders: Map<string, object>
+        }
     }
 }
```
ArmandPhilippot, line 154: Hi @ruohki and thanks for taking the time to help making the docs better!

I don't know if you followed [the discussion in Discord](https://discord.com/channels/830184174198718474/830184175176122389/1333174974365700246) but I don't think we should use `declare global` here because without imports it is working as expected. But I agree, we should show how to use an imported type as it seems to be confusing. Perhaps something like:
```suggestion
namespace App {
    interface Locals {
        site: {
            title: string
        },
        welcomeTitle: () => string,
        orders: Map<string, object>,
        user: import("@supabase/supabase-js").User
    }
}
```

ruohki: Hmm so maybe we could show both versions and mention that if someone is using imports to use the global approach?

ArmandPhilippot: Thanks, that looks better since I think the example should remain an ambient module declaration!

Regarding the format (as a commented property or not, or if we need to mention that it is Typescript behavior), I'm not sure what is best so I'll let Sarah (or someone else) give their opinion since it's kind of new content. She'll probably be able to give better advice! 😄 I will post a comment with a bit more context for reviewers.

ArmandPhilippot:
PR's context:
This issue has already been discussed earlier on Discord and [Fryuni already gives great explanations](https://discord.com/channels/830184174198718474/853350631389265940/1307714929088008223) regarding the use of imports in `.d.ts` files. It's a Typescript behavior but it seems to confuse some users.

This issue was raised again yesterday, so (after ruohki's PR) the [idea was to provide an example](https://discord.com/channels/830184174198718474/830184175176122389/1333174974365700246) on how to do imports in a `.d.ts` to avoid turning it into a normal module and using `declare global`.

I think adding a property could be enough to show that, but is it clear enough? I know we avoid documenting the behavior of third-party technologies, but should we add more context here?

sarah11918:
A comment from maintainer's discussion:

> Actual docs rather than a comment might be better? Plus also maybe an example with an Astro type rather than an external dependency?

sarah11918:
We discussed on Talking and Doc'ing and would love an example that includes both kinds of types together in one, in a realistic situation where people could be defining one themselves while also importing something else!

sarah11918:
We discussed on Talking and Doc'ing and would love an example that includes both kinds of types together in one, in a realistic situation where people could be defining one themselves while also importing something else!

ruohki:
Okay, think i could provide a small example (which i already did with an astro:db example that got rejected :D)

sarah11918:
Maybe next time will be your lucky day! 😅

ruohki:
> Maybe next time will be your lucky day! 😅

Is this the birth of the astro:user-examples repo?

ArmandPhilippot:
Since the PR was opened this issue has been addressed two more times, so I'm leaving a suggestion in case it can help move things forward. Perhaps we could use something like:

```ts
type User = {
  id: number;
  name: string;
};

declare namespace App {
  interface Locals {
    user: User;
    welcomeTitle: () => string;
    orders: Map<string, object>;
    i18n: import("astro").AstroConfig["i18n"];
  }
}
```

Well, `import("astro").AstroConfig["i18n"]` is probably not really useful, especially with the arrival of the [serialized configuration](https://docs.astro.build/en/reference/experimental-flags/serialized-configuration/) but I haven't found any other types in Astro packages/virtual modules that might make sense...

Otherwise, we can replace the `i18n` key with a type coming from a dummy local file. For example, [Lucia Auth](https://lucia-auth.com/sessions/cookies/astro#session-validation) uses:
```ts
session: import("./lib/server/session").Session | null;
```

I'm not an `Astro.locals` user (yet), so there may be better use cases. Just trying to help unlock this PR... 😄

sarah11918:
Just pinging this for freshness! I see Armand has left another comment. Are you still interested in working on this @ruohki ?

ruohki:
Ah sorry i lost track of this issue... Let me check things next week.

sarah11918:
Just a friendly, "Going through docs PRs after being busy with Launch Week" ping on this one! :smile:

sarah11918:
Just consider this a friendly ping @ruohki , and I know life can get in the way sometimes. Just let us know if you're no longer interested/available, no worries!

ArmandPhilippot:
To prevent this PR from getting stuck any longer, I just pushed an updated example where we use both a local type and an imported type using an inline import. We can always update this later with a more helpful example if needed!
Thanks @ruohki for bringing the issue! 🫡

sarah11918:
Updating branch to trigger another deploy. Can't immediately tell why the first one failed.

sarah11918 (approved):
Thanks everyone for working together to add this helpful update! Welcome to Team Docs, @ruohki ! :partying_face:

PR #10813, opened 2025-01-27
Added my "Gatsby to Astro" migration article

alvinometric:
I did the polite thing and didn't put my article at the top of the list 😄


sarah11918:
I appreciate not putting it at the *top* of the list, but I did just yesterday go through these for "balance", putting cards of roughly the same size next to each other, and this kind of throws that off a little with a cascading effect!

Looking quickly, I do think the end of the list is the best place at the moment, as there doesn't seem to be another good choice to pair it up with. So please move it there, and yay, thanks for sharing!

alvinometric:
No problem, thanks Sarah 🙏

PR #10822, opened 2025-01-28
fix: add undocumented options in `integrations-reference.mdx`

ArmandPhilippot:
I was looking at the [Integrations reference page](https://docs.astro.build/en/reference/integrations-reference/) and noticed some formatting issues... Since Florian is about to give a talk on integrations, I figured it was a good time to do a more thorough check instead! Also, I was worried that you would be bored after `5.2`. 😄 

#### Description (required)

Sorry the description isn't really succinct but I'm trying to keep track of the why and when for each point.

**In `upgrade-to/v5.mx`:**
* Fixes a mismatch between paths in `addDevToolbarApp` code snippets (the path is not the same as the one used in the code block's title)

**In `integrations-reference.mdx`:**

* Consistent formatting:
  * Updates the `Type:`/`Since` block to be consistent with the [formatting of API reference](https://contribute.docs.astro.build/reference/api-references/)
  * Updates the indentation in some code snippets (most of them use 2 spaces)
  * Updates some titles to explicitly show that they are functions (e.g. `addMiddleware` > `addMiddleware()`)
  * Replaces `+` with ExpressiveCode notation to highlight added lines in `Allow installation with astro add`
  * Makes `injectRoute` and `injectTypes` consistent with the other options. These were the only objects expanded in the API quick reference, so:
    * it might be better to be consistent with the others
    * since `InjectedRoute` and `InjectedType` can be imported, it could be nice to show the types there
    * this reduces the horizontal scroll which might be easier to read on mobile.

* Existing content:
  * Adds the missing `refreshContent`, `middlewareEntryPoint` and `pages` options in `Quick API Reference`
  * Adds the missing `logger` option in the type preview of some hooks
  * Adds a link from the `Hooks` section to the `AstroIntegrationLogger` since each hook has a `logger` option
  * Reorders the hooks to match the real order (tested using a custom integration and `console.log`)
  * Fixes the `Previous hook`/`Next hook` links (some was wrong/outdated or the anchor did not match the link)
  * Collects some types into a new section
  * Fixes some types
    * `addDevToolbarApp` have been updated in the other pages, but not in this one... `string` was removed in https://github.com/withastro/astro/pull/11987)
    * in `injectRoute`, `pattern` was displayed twice with a different type (one of them should be `prerender`)
    * `addWatchFile` is a function

* New content:
  * Adds the missing `buildOutput` option in `astro:config:done` (added in https://github.com/withastro/astro/pull/11824, so v5)
  * Adds the missing `toolbar` option in `astro:server:setup`  (added in https://github.com/withastro/astro/pull/7821, so v4.7.0)
  * Adds documentation for `astro:build:setup` options
  * Adds documentation for `astro:build:generated` options
  * Adds documentation for `astro:build:ssr` options
  * Break up `IntegrationResolvedRoute` and `IntegrationRouteData` types to avoid too long blocks of code and to make it look more like documentation (see the issue reported by [Fryuni on Discord](https://discord.com/channels/830184174198718474/853350631389265940/1329482701513687161) )

**Remaining question:** the `Integrations API` page becomes longer than the current [Render context page (formerly API reference](https://docs.astro.build/en/reference/api-reference/)) we split. Could a section in the sidebar like `Runtime API` make sense here? Not necessarily now, but the page might get a bit too big for Algolia (~1680 lines, IIRC the former API Reference had a bit more than 2000 lines).

- Suggested label: `add new content`, `consistency/formatting` and `improve or update documentation`


sarah11918:
Amazing, thank you so much for this initiative Armand!

As you say, in draft so no rush. But would be great for at least @florian-lefebvre and @Fryuni to have a look over when it's ready before I do!

You could also post this in the `#integrations` channel in Discord to see whether any of our integrations builders are interested in commenting on this!

ArmandPhilippot:
Oh definitely! There still some work to do here (mainly document what is not documented) so I'll try to do that with my knowledge but yeah, once done a review from our integrations expert would be much appreciated! 😄

florian-lefebvre:
Feel free to ping me once you're ready!

ArmandPhilippot:
I think I've covered all the updates!

So when you'll have time @florian-lefebvre and @Fryuni, feel free to make suggestions! I'm not sure if my descriptions are correct and perhaps you will have better examples to show. Thanks! 🙌🏽 

I think the key points to check here are:
* those listed in `New content:` (see the top post) since it is new content and I had a bit of trouble finding information (especially in the `astro:build:ssr` hook).
* if I didn't make a mistake with the order of the hooks (it seems the order was outdated). I checked with a custom integration and using `console.log()`.

But if you have time to look at the whole thing, that's also appreciated! 😄 

I also left some questions in the top posts. The last one is maybe more for Sarah, so here are the ones where you might have an opinion:
1. In the [`astro:config:setup`](https://deploy-preview-10822--astro-docs-2.netlify.app/en/reference/integrations-reference/#command-option) hook, should we precise that `command` also uses `sync` as value when running `astro check` (since `sync` is called under the hood)?
2. In [`route option`](https://docs.astro.build/en/reference/integrations-reference/#route-option) we have [route.prerender](https://docs.astro.build/en/reference/integrations-reference/#routeprerender) but we are not showing the `route.component` property, is it on purpose? (both were available at the time of writing the documentation, see
https://github.com/withastro/docs/pull/9064) In doubt I did not add it.

sarah11918:
Noting: I will come in *after* we've had technical reviews!

Thanks again for the great effort, Armand! 🙌

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -622,28 +645,37 @@ interface IntegrationResolvedRoute {
 'astro:config:done'?: (options: {
   config: AstroConfig;
   setAdapter: (adapter: AstroAdapter) => void;
-  injectTypes: (injectedType: { filename: string; content: string }) => URL;
+  injectTypes: (injectedType: InjectedType) => URL;
```
florian-lefebvre, line 648: Should we add a link?

florian-lefebvre (commented):
This looks pretty good! I'm not familiar with most build hooks so I trust you on those!

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Well, in the other "quick overviews" there was no links (e.g. `AstroAdapter` above) so I didn't add any here. Usually, the links are rather in the detailed properties. Here, it would be in `injectTypes() option` but the [type is already expanded](https://github.com/withastro/docs/blob/8dd180fe458bfa7f90c543c1098cad2bd124df09/src/content/docs/en/reference/integrations-reference.mdx?plain=1#L676).

I haven't touched this section, but if you think a link might be helpful we could slightly update the description:
```md
This function receives an `InjectedType` object with the following properties:
* `filename`, used to generate a file at ...
* `content`,  to create the body of the file from valid TypeScript.
```
And we could add the link on `InjectType` object?

Or:

```md
This function receives an object with the following properties:
// Same as above
```
So no link here, but we replace the `Type:` to show `InjectedType` with a link instead of the expanded object.

florian-lefebvre: Alright the current diff is fine then!

ArmandPhilippot:
Thanks for you review @florian-lefebvre! I based myself on:
* the JSDocs when available
* the PRs to gather some information
* when there were none, I tried to understand what was accessible in the object to write a description... Not ideal, but if you didn't see anything shocking I guess I managed to get by! 😅 

1. Oh, good to know! So yes, no need to update that.
2. I can't answer your question... but okay, I'll add the `component`property to the PR!

Fryuni:
There was some discussion about changing the name `component` prior to v5 because it is not always a component. I thought we had changed it to something else.

florian-lefebvre:
We changed it on `routes:resolved`, I completely missed the `route:setup` hook

ArmandPhilippot:
Oh okay that's why `IntegrationResolvedRoute` uses `entrypoint` while we have `component` in [`RouteOptions`](https://github.com/withastro/astro/blob/c497491cfe6a03f6a2118392d198ee1738754629/packages/astro/src/types/public/integrations.ts#L14-L27) and [`RouteData`](https://github.com/withastro/astro/blob/c497491cfe6a03f6a2118392d198ee1738754629/packages/astro/src/types/public/internal.ts#L42-L144).

Actually, I read the description of `route option` too quickly. We already had a description for `component` but in the parent section not on its own like `prerender`, hence the confusion. So I updated that to be consistent.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -548,7 +548,14 @@ const integration = {
...
 
-The `component` property is a readonly string path relative to the project root, e.g `"src/pages/blog/[slug].astro"`.
+##### `route.component`
+
+<p>
+**Type:** `string`<br />
+<Since v="4.14.0" />
+</p>
+
+The `component` property is a readonly string path relative to the project root, e.g `"src/pages/blog/[slug].astro"`. Its value helps you identify the route being set up.
```
florian-lefebvre, line 558: I think this is not always true, eg. for injected routes that reference package exports

ArmandPhilippot: You mean when the injected routes comes from `node_modules`? The path would not be `./node_modules/...`?

Reading [the JSDoc](https://github.com/withastro/astro/blob/c497491cfe6a03f6a2118392d198ee1738754629/packages/astro/src/types/public/integrations.ts#L15-L19), it seems a relative path is always the expected format, right? Looking at the [hook implementation](https://github.com/withastro/astro/blob/c497491cfe6a03f6a2118392d198ee1738754629/packages/astro/src/integrations/hooks.ts#L650), I see `route.component` being used in the logger... But I have no idea what else to expect than a path to a file. 😅 

Do you have an example of what this might be (or a direction where I could look)?

Fryuni:
> We changed it on `routes:resolved`, I completely missed the `route:setup` hook

We should update it on the others and mark the `component` field as deprecated then. We'll have to carry it for an entire major now 😅

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Yeah I understand that, but I have to admit that I am having a bit of trouble understanding the problem with the description (except that the property name is wrong but it is the name currently in use and we are documenting what currently exists).
But again, I'm not the integration expert here... So maybe some things that seem obvious to you, aren't obvious to me due to a lack of knowledge in this context.

I had tested on my own website where I have both static routes, dynamic routes, endpoints, and injected routes (but which do not refer to `node_modules`) and in every case, using `console.log(route.component)`, I see:
```
src/pages/en/index.astro
src/pages/en/[slug].astro
src/pages/feed.xml.ts
src/components/atoms/button/button.stories.astro
```
(`stories.astro` are my injected routes if you are wondering)

So it seems to me the description matches what I get. But I haven't tested with an import from a package, so maybe that's the problem...

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -453,164 +510,137 @@ It allows you to have a dedicated folder, avoiding conflicts with another integr
...
-	 *  Source component URL
-	 */
-	entrypoint: RouteData['component'];
+**Previous hook:** [`astro:route:setup`](#astroroutesetup)
 
-	/**
-	 * Whether the route is prerendered or not
-	 */
-	isPrerendered: RouteData['prerender'];
+**Next hook:** [`astro:config:done`](#astroconfigdone)
```
Fryuni, line 602: A bit tricky here. This hook also runs whenever routes change during dev mode, so it can run after `astro:config:done`.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -453,164 +510,137 @@ It allows you to have a dedicated folder, avoiding conflicts with another integr
...
 
-	/**
-	 * The {@link IntegrationResolvedRoute} to redirect to. It's present when `IntegrationResolvedRoute.type` is `redirect`.
-	 */
-	redirectRoute?: IntegrationResolvedRoute;
+**When:** In `astro dev`, it also runs if a file based route changes (added/removed/updated).
```
Fryuni, line 604:
```suggestion
**When:** In `astro dev`, it also runs on every change to a file based route (added/removed/updated).
```

I think it is better to note that it can run multiple times, not once in case an update happens.

Fryuni (commented):
Looks great. Left some small notes for us to discuss.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -33,41 +33,62 @@ interface AstroIntegration {
...
+    }) => void | Promise<void>;
+    'astro:build:generated'?: (options: {
+      dir: URL;
+      logger: AstroIntegrationLogger;
     }) => void | Promise<void>;
     'astro:build:done'?: (options: {
+      pages: { pathname: string }[];
       dir: URL;
       /** @deprecated Use the `assets` map and the new `astro:routes:resolved` hook */
       routes: IntegrationRouteData[];
```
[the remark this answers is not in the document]
ArmandPhilippot: With content collections, we had [`getEntryById`](https://v4.docs.astro.build/en/reference/modules/astro-content/#getdataentrybyid) and [`getEntryBySlug`](https://v4.docs.astro.build/en/reference/modules/astro-content/#getentrybyslug) in docs even though they were obsolete. So I think it's the same here and as long as they are available they should be in the documentation?
But perhaps we don't need to show that in the `Quick API Reference`.

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Now that I'm testing again, it seems a bit inconsistent (even with `astro:route:setup`). Something to do with the cache perhaps (just a guess)? Or my website not being well optimized, I have some custom setup that I need to improve... 😅 

The first run seemed a bit mixed up:
```
...
22:58:28 - astro:route:setup
22:58:28 - astro:route:setup
22:58:28 - astro:routes:resolved
22:58:28 - astro:config:done
22:58:28 - astro:server:setup
22:58:28 [types] Generated 0ms
22:58:28 [content] Syncing content
22:58:28 [content] Synced content
22:58:28 - astro:route:setup
22:58:28 - astro:route:setup
...
22:58:29 - astro:server:start
 astro  v5.1.10 ready in 1574 ms

22:58:29 watching for file changes...
22:58:29 - astro:route:setup
22:58:29 - astro:route:setup
...
22:58:29 - astro:route:setup
22:58:29 - astro:routes:resolved
22:58:29 - astro:route:setup
...
22:58:29 - astro:routes:resolved
```
The next two runs, I see:
```
23:04:33 - astro:route:setup
23:04:33 - astro:route:setup
23:04:33 - astro:routes:resolved
23:04:33 - astro:config:done
23:04:33 - astro:server:setup
23:04:33 [types] Generated 0ms
23:04:33 [content] Syncing content
23:04:33 [content] Synced content
23:04:33 - astro:server:start
 astro  v5.1.10 ready in 1883 ms

// No other hook calls after
```
So what is described...

However, with these three tests, I don't see the hook being called again while navigating... But perhaps my test is not accurate enough. I'm using something similar to: https://blog.logrocket.com/understanding-astro-integrations-hooks-lifecycle/#examining-hooks-evaluation-order (with the missing hooks)
Also this is a static website, it's possible that with SSR it works differently... So great call!

src/content/docs/en/reference/integrations-reference.mdx
florian-lefebvre: Yeah I was thinking about a package import

ArmandPhilippot: So I tried to find an existing package that I could use to test... there is `astrojs/web-vitals` but it still mentions Astro Studio, so I thought maybe it wasn't the best solution...

I went with a fake package (inspired by a [fake fixture in the core repo](https://github.com/withastro/astro/blob/f98e29d141ef5202cde6add62d451f55dab8cd00/packages/astro/test/fixtures/custom-404-injected-from-dep/astro.config.mjs)) and pnpm workspaces:
```
packages:
  - "fake-packages/*"
```

Using `logger.info(route.component)`, I see `[astro-route-setup] fake-packages/custom-404-pkg/404.astro`.
So it's using a relative path, right? However, perhaps because this is a workspace package, it's using the actual path instead of a path from `node_modules`... But I guess for an installed package, it will be `node-modules/fake-packages/...`.

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot, line 602: After doing more testing, perhaps we could add something like:
```suggestion
**Next hook:** [`astro:config:done`](#astroconfigdone) when setting up the server in "dev" mode or during production builds
```
In addition to your rewording for `When:`.

ArmandPhilippot: After doing more testing, perhaps we could add something like:
```suggestion
**Next hook:** [`astro:config:done`](#astroconfigdone) when setting up the server in "dev" mode or during production builds
```
In addition to your rewording for `When:`.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -453,164 +510,137 @@ It allows you to have a dedicated folder, avoiding conflicts with another integr
...
-<Since v="5.0.0" />
+<p><Since v="4.14.0" /></p>
 
 **Previous hook:** [`astro:config:setup`](#astroconfigsetup)
 
-**Next hook:** [`astro:config:done`](#astroconfigdone)
+**Next hook:** [`astro:routes:resolved`](#astroroutesresolved)
 
-**When:** In `astro dev`, it also runs if a file based route changes (added/removed/updated).
+**When:** In `astro dev`, whenever a route is requested. In `astro build`, while bundling and transforming a route file.
```
ArmandPhilippot, line 531: Is this part correct: `whenever a route is requested`? Or is it just me who is having trouble understanding the nuance (not a native speaker)?
I mean, from my tests, `astro:route:setup` in dev mode is not call again when a route is requested (while browsing the website). However, like `astro:routes:resolved`, `it also runs on every change to a file based route (added/removed/updated).` (taking Fryuni suggestion)

src/content/docs/en/reference/integrations-reference.mdx
florian-lefebvre: mmh i think it will remain a relative path. We're good then!

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: > So I guess you're right! But still, my first run seems a bit inconsistent... I don't know if it's my setup.

Your first run will cause a sync, which is adding files, so they run again in response to those events.

Fryuni: Perhaps this:
```suggestion
**Next hook:** [`astro:config:done`](#astroconfigdone) (only during setup)
```

It's simpler and applies for dev and build.

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Ok, so does this seem more accurate to you:
```suggestion
**When:** Before bundling starts. In `astro dev`, it also runs on every change to a file based route (added/removed/updated).
```
Or I misunderstand how Vite works and there is another difference between `dev`/`build`?

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: I'm fine with your suggestion!

> Your first run will cause a sync

Yes that's what I thought afterwards... I was just confused because I didn't expect a difference. Thanks for the confirmation!

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: That is... ok...

Technically there is no bundling at all during `dev`, but I don't think that matters here.

ArmandPhilippot: Oh yes! I haven't paid enough attention to the wording... I was hoping to find a description short enough to gather the two. In dev, it's more a module graph setup if I'm not mistaken (sorry if there are any approximations... I've dug into the SSR side a bit in the past but I'm not really familiar with the technical details of how Vite works under the hood)

Does something like that seems better/short enough?
```suggestion
**When:** In `astro build`, before bundling starts. In `astro dev`, while building the module graph and on every change to a file based route (added/removed/updated).
```

src/content/docs/en/reference/integrations-reference.mdx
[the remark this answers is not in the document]
ArmandPhilippot: Just pushed a commit to remove this option from the `Quick API Reference` So I'm marking this as resolved, feel free to undo the resolution if you think this needs further discussion!

src/content/docs/en/reference/integrations-reference.mdx
florian-lefebvre: @Fryuni you probably know more about this than me

src/content/docs/en/reference/integrations-reference.mdx
Fryuni: That's great!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -453,164 +508,137 @@ It allows you to have a dedicated folder, avoiding conflicts with another integr
...
-                console.log(projectRoutes)
-            },
+##### `route.component`
+
+<p>
+**Type:** `string`<br />
+<Since v="4.14.0" />
+</p>
+
+The `component` property is a readonly string path relative to the project root, e.g `"src/pages/blog/[slug].astro"`. Its value helps you identify the route being set up.
```
sarah11918, line 556:
```suggestion
The `component` property indicates the entrypoint that will be rendered on the route. You can access this value before the routes are built to configure on-demand server rendering for that page.
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -701,7 +754,53 @@ export default {
...
+A function that takes an event name as first argument and a callback function as second argument. This allows you to receive a message from a dev toolbar app with data associated to that event.
+
+##### `onAppInitialized()`
+
+<p>
+
+**Type:** `(appId: string, callback: (data: Record<string, never>) => void) => void`<br />
+</p>
+
+A function fired when a dev toolbar app is initialized. The first argument is the id of the app that was initialized. The second argument is callback function to run when the app is initialized.
```
sarah11918, line 783:
```suggestion
A function fired when a dev toolbar app is initialized. The first argument is the id of the app that was initialized. The second argument is a callback function to run when the app is initialized.
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -701,7 +754,53 @@ export default {
...
+A function fired when a dev toolbar app is toggled on or off. The first argument is the id of the app that was toggled. The second argument is a callback function providing the state to execute when the application is toggled.
+
+##### `send()`
+
+<p>
+
+**Type:** `<T>(event: string, payload: T) => void`<br />
+</p>
+
+A function that takes an event name as first argument and a payload as second argument. The payload can be any serializable data. This allows you to send a message to the dev toolbar that an app can listen for.
```
sarah11918, line 801:
```suggestion
A function that sends a message to the dev toolbar that an app can listen for. This takes an event name as the first argument and a payload as the second argument which can be any serializable data.
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -908,159 +1225,107 @@ You can reference the full `IntegrationRouteData` type below, but the most commo
...
+**Methods available** to write to terminal:
+- `logger.info("Message")`;
+- `logger.warn("Message")`;
+- `logger.error("Message")`;
+- `logger.debug("Message")`;
 
-export default defineConfig({
-  integrations: [setPrerender()],
-});
+All the messages are prepended with a label that has the same value of the integration.
```
HiDeoo, line 1278:
```suggestion
All the messages are prepended with a label that has the same value as the name of the integration.
```

Not sure if a word is missing there, it's old/moved content, but the "value of the integration" feels a bit weird to me.

JusticeMatthew (approved):
Review on talking and docing was great

Also checked for typos and came up empty

Great job 😄

Fryuni (approved):

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Yeah I think it sounds better, thanks for noticing that!

PR #10858, opened 2025-01-31
add documentation to inline svg

Its-Just-Nans:
#### Description (required)

Add doc for inlining svg

- https://github.com/withastro/roadmap/discussions/679
- https://github.com/withastro/roadmap/discussions/679#discussioncomment-6924267
- https://vitejs.dev/guide/assets.html#importing-asset-as-string
- https://docs.astro.build/en/reference/experimental-flags/svg/

### DIFF
--- a/src/content/docs/en/guides/images.mdx
+++ b/src/content/docs/en/guides/images.mdx
@@ -261,6 +261,22 @@ export default defineConfig({
 });
 ```
 
+
+### Inlining svg in `.astro` files
+
+For tiny svg, inlining them as base64 can speed up loading times.
+
+```astro
+---
+import svgAsStr from "../image.svg?raw";
+
+const svgBase64 = `data:image/svg+xml;base64,${btoa(svgAsStr)}`;
+---
+
+<img src={svgBase64} alt="A sleeping cat." >
+```
+
+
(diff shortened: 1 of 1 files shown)


sarah11918:
Thank you for providing more helpful context about using inline SVGs! As this is documentation for an experimental, and actively developing feature, it is not intended to be perfectly complete to allow for the greatest flexibilty while the feature is being worked on and may yet still change in significant ways. And we do encourage people to go check the roadmap proposals for more information, relying on docs only for basic implementation at this stage.

I checked the linked comments, and I see much of this conversation takes place in an open proposal for a feature, and discussion of a temporary workaround. Normally, we will only add to documentation when this is a **supported, recommended** procedure to follow that we don't expect to change too much.

So, while this is a solution, I'm not sure from that discussion whether we would consider this something to document *yet* as our official method, or whether a different proposed solution is in the works.

I'll ask @Princesseuh whether she would suggest adding this information to our formal documentation at this time!

sarah11918:
I'm going to close this PR as it references very old conversations, and there is an open [SVG RFC](https://github.com/withastro/roadmap/pull/1035) where you can discuss the current proposal for using SVG and whether this suggestion fits with that implementation.

PR #10791, opened 2025-01-24
feat: experimental serialized configuration

ematipico:
#### Description (required)

This PR creates a new page to explain the upcoming `astro:config`, highlighted in the roadmap PR https://github.com/withastro/roadmap/pull/1106

 #### For Astro version: `5.2.0`. See astro PR [#13084](https://github.com/withastro/astro/pull/13084).

PR #10872, opened 2025-02-03
fix: misplaced sentence in `serialized-configuration.mdx`

ArmandPhilippot:
#### Description (required)

This PR moves a sentence that seems misplaced in `serialized-configuration.mdx`.

Looking at #10791, it looks like content was added in the middle of a sentence so I think the following is now misplaced: `so attempting to use them wil raise an error.`

See the original commit: https://github.com/withastro/docs/pull/10791/commits/19abd8087e514a5127506f8e9d14ad97b7247c39


sarah11918 (approved):
Thank you!

PR #10883, opened 2025-02-04
i18n(pt-br): Update content-i18n-pt-BR.yml

avilyre:
#### Description (required)

In the Not Found page, the current 'take me home' in Portuguese translation, doesn't make sense, it sounds like "I want to go back to a physical house". So it's common to translate something like "Home Page" to "Página inicial" and then it sounds like "I will be back to the home page".

Here is the current translation:

Current "Me level para casa" -> "Me leve para página inicial"
Captura de Tela 2025-02-04 às 16 51 39

Result:
Captura de Tela 2025-02-04 às 17 28 32


src/content/i18n/pt-BR.yml
```diff
@@ -31,7 +31,7 @@ recipesLink.plural: Receitas relacionadas
 # 404 Page
 404.title: Não encontrado
 404.content: Esta página não está no nosso sistema solar.
-404.linkText: Me leve para casa.
+404.linkText: Página inicial.
```
yanthomasdev, line 34: The original is "Take me home", I guess we can still keep the original context but not translate "home" directly. Like this:
```suggestion
404.linkText: Me leve para página inicial.
```

avilyre: That makes sense

yanthomasdev (approved):
LGTM! Thanks!

PR #10913, opened 2025-02-07
Add missing server context from cloudflare example

robmen:
#### Description (required)

The Cloudflare example using `astro;env` did not work because it was missing the context. Adding `/server` fixed it for me, so I thought I'd send the change your way. If there is a better fix, please feel free to close this PR with a better one. 😄 

- Closes #
- Suggested label:


sarah11918:
Thanks for this @robmen ! I suspect this is exactly the change needed, but I'll tag @florian-lefebvre in to also confirm whether we should be updating [this reference page/example](https://docs.astro.build/en/reference/modules/astro-env/#imports-from-astroenv) so that we're *always* importing from either `/client` or `/server` everywhere we show it!

florian-lefebvre (approved):
Perfect thanks!

florian-lefebvre:
Yes the reference should be updated to, to use `astro:env/server`

sarah11918:
Thanks Florian! 

@robmen Would you be able to *also* add the change to the reference page I linked above in my last comment?

robmen:
It took me a while to get back to this but I finally made the other requested update.

src/content/docs/en/reference/modules/astro-env.mdx
```diff
@@ -19,7 +19,7 @@ The `astro:env` API lets you configure a type-safe schema for environment variab
 ```js
```
florian-lefebvre, line 19: @sarah11918 I wonder if should remove this codeblock entirely. There's another one below that is more useful

sarah11918: I'll ask @ArmandPhilippot to confirm, but normally this section is "everything available to import from the module". Where I guess it's a bit odd is that a user can configure other things that get imported from here, but we obviously can't know them ahead of time.

I would say we would typically keep this, to show the API. Do you agree Armand?

ArmandPhilippot: Yeah, this module is a bit special because it only has `getSecret` and we can import other "unfrozen" things... But, yes, I agree I think we should stay consistent with the other modules. When navigating from page to page within the same set, I think it can be helpful to keep the same structure.

sarah11918: Thank you! Then I say let's get that money!!

sarah11918 (approved):

sarah11918:
Well, we let this PR sit so long that we actually fixed the Cloudflare limitation! 😅

But, we still have the other `/server` to add, so thank you, Rob! 🫡

PR #10915, opened 2025-02-07
feat: add docs for experimental sessions auto-config

ascorbic:
#### Description (required)

The experimental sessions config is changing in 5.3. The Node and Netlify adapters will now automatically configure storage for you. The `experimental.session` value is changing to a boolean flag, and the configuration is moving to a top-level `session` object.

See https://github.com/withastro/astro/pull/13145 for the implementation


src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -35,54 +35,65 @@ The [`session` object](#sessions-api) allows you to interact with the stored use
 
 ## Enabling experimental sessions
 
-In this early release of `experimental.sessions`, Astro's official server adapters have not yet been updated to provide a default session driver based on the platform's default storage feature. Until these are built in, you will have to specify a `driver` yourself in the `experimental.sessions` configuration object in addition to [adding an adapter](/en/guides/on-demand-rendering/#add-an-adapter).
-
-The following sections show the recommended drivers for each adapter. (These will be automatically configured for you in the stable release of this feature.) Alternatively, you can specify [any supported driver from unstorage](https://unstorage.unjs.io/drivers/). 
+To enable sessions, set the `experimental.session` flag to `true`. Sessions only work on pages with on-demand rendering, so you need to [install an adapter](/en/reference/config/#adapter) that supports on-demand rendering, and ensure that any pages that use sessions are set to `prerender: false`, or `output` is set to `server` in the Astro config.
```
ArmandPhilippot, line 38: I'm not sure if you wanted to link to the [Configuration reference](https://deploy-preview-10915--astro-docs-2.netlify.app/en/reference/configuration-reference/#adapter) or the [On-demand rendering guide](https://deploy-preview-10915--astro-docs-2.netlify.app/en/guides/on-demand-rendering/#add-an-adapter), but this should fix the links issues:
```suggestion
To enable sessions, set the `experimental.session` flag to `true`. Sessions only work on pages with on-demand rendering, so you need to [install an adapter](/en/guides/on-demand-rendering/#add-an-adapter) that supports on-demand rendering, and ensure that any pages that use sessions are set to `prerender: false`, or `output` is set to `server` in the Astro config.
```
(I chose the guide link because it was the link we had before, but feel free to update)

ArmandPhilippot (commented):
I left a suggestion to fix the links issues. Otherwise, LGTM but I'm not the wording expert here. 😄 

Some other thoughts:

* I wonder, should we precise that the top-level `session` config was added in `5.3`? I mean its experimental, maybe it's enough to leave it as is. But it might help if someone was already using this feature to understand what changed. I'm not sure how we handle that with experimental features. So it's just a thought.

* Should [`session.cookie`](https://deploy-preview-10915--astro-docs-2.netlify.app/en/reference/experimental-flags/sessions/#sessioncookie) be updated as well? It's no longer under `experimental`, right?

* It wasn't documented before, but looking at your PR I saw the [`ttl` option](https://github.com/withastro/astro/blob/3ddeaf0d3d7feeb59705f82e978981f1e7726356/packages/astro/src/core/config/schema.ts#L548) in addition to `driver`, `options` and `cookies`. Should we document this as well?

ascorbic:
Thanks @ArmandPhilippot! I've done a bit of a rewrite and reorganised the options, moving `cookie` and adding `ttl`.

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,73 +22,141 @@ This experimental feature allows you to store and access items such as login sta
...
+:::
+
 ## Using sessions
 
-Once you have configured a driver, you can use the `session` object to interact with the session. This object is accessible as `Astro.session` in your Astro components and pages and on the `context` object in API endpoints, middleware and actions. The API is the same in all cases.
+Once you have configured a driver, you can use the `session` object to interact with the session. The object is accessible as `Astro.session` in your Astro components and pages and on the `context` object in API endpoints, middleware and actions. The API is the same in all cases. 
+
+The session is generated automatically when it is first used and can be regenerated at any time with [`Astro.session.regenerate()`](#sessionregenerate) or destroyed with [`Astro.session.destroy()`](#sessiondestroy).
 
 In most cases you will only need to use [`Astro.session.get()`](#sessionget) and [`Astro.session.set()`](#sessionset). For other available methods, see the [API section](#sessions-api).
```
yanthomasdev, line 161: not from this PR, but it doesn't hurt to fix this comma as well.
```suggestion
In most cases, you will only need to use [`Astro.session.get()`](#sessionget) and [`Astro.session.set()`](#sessionset). For other available methods, see the [API section](#sessions-api).
```

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,83 +22,151 @@ This experimental feature allows you to store and access items such as login sta
...
+#### `session.cookie`
+
+<p>
+
+**Type:** `string` | `object`<br />
+**Default:** `astro-session`<br />
+
+</p>
+
+Configures the session cookie. This cookie is set in the response when a session is generated. No actual user data is stored in the cookie – just an ID that is used to identify a user's session. The `session.cookie` option can be used to set options for this cookie. You can either provide a `string`, which will be used as the cookie name, or an object which allows additional options:
```
sarah11918, line 112: This feels more convoluted than what was at the old "cookies" entry. (Although, I do like the first sentence as a descriptor of the option!) Just checking that this is an intentional change that you don't like what already existed, which I think felt clearer?

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,83 +22,151 @@ This experimental feature allows you to store and access items such as login sta
...
+  }
+```
+
+#### `session.ttl`
+
+<p>
+  
+**Type:** `number`<br />
+**Default:** `undefined`<br />
+
```
sarah11918, line 134: We should have a `<Since />` here, probably?

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,83 +22,151 @@ This experimental feature allows you to store and access items such as login sta
...
+</p>
+
+An optional default time-to-live for session values, in seconds.
```
sarah11918, line 137:
```suggestion
An optional default time-to-live expiration period for session values, in seconds.
```
Adjust the wording if that's not exactly right, but the definition of a value should ideally be more than a restatement of its name. (ttl = ttl.) So, throwing the notion of an "expiration period" up there instead!

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,83 +22,151 @@ This experimental feature allows you to store and access items such as login sta
...
+<p>
+  
+**Type:** `number`<br />
+**Default:** `undefined`<br />
+
+</p>
+
+An optional default time-to-live for session values, in seconds. 
+
+By default, session values persist until they are deleted or the session is destroyed. If you would prefer the values to expire after a certain period, you can set a `ttl` option. You can set this global default `session.ttl` option, or you can pass a `ttl` option to [`session.set()`](#sessionset). A value passed to `session.set()` will override the global default for that entry.
```
sarah11918, line 139:
```suggestion
By default, session values persist until they are deleted or the session is destroyed, and do not automatically expire because a particular amount of time has passed. Set `session.ttl` to add a default expiration period for your session values. Passing a `ttl` option to [`session.set()`](#sessionset) will override the global default for that individual entry.
```
Just trying to tighten up and make a bit more direct. Check for accuracy/nuance and adjust as desired!

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,83 +22,151 @@ This experimental feature allows you to store and access items such as login sta
...
+```js title="astro.config.mjs" {3}
+  {
+    session: {
+      ttl: 60 * 60, // 1 hour
+    }
   }
 ```
 
+:::note
+Setting a value for `ttl` does not delete the value from storage unless it is accessed after the time has passed. In that case, it will be treated as if it does not exist and the value will be deleted.
```
sarah11918, line 150:
```suggestion
Setting a value for `ttl` does not automatically delete the value from storage after the time limit has passed.

Values from storage will only be deleted when there is an attempt to access them after the `ttl` period has expired. At this time, the session value will be undefined and only then will the value be deleted. 
```
Just trying to make this a bit more explicit. (Accuracy, nuance, etc...)

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,83 +22,151 @@ This experimental feature allows you to store and access items such as login sta
...
+    session: {
+      ttl: 60 * 60, // 1 hour
+    }
   }
 ```
 
+:::note
+Setting a value for `ttl` does not delete the value from storage unless it is accessed after the time has passed. In that case, it will be treated as if it does not exist and the value will be deleted. 
+
+Individual drivers may also support a `ttl` option that will automatically delete sessions after the specified time. See the driver documentation for more information.
```
sarah11918, line 152:
```suggestion
Individual drivers may also support a `ttl` option that will automatically delete sessions after the specified time. See your chosen driver's documentation for more information.
```

sarah11918 (commented):
Thanks @ascorbic ! Looking great! Here are my initial thoughts!

src/content/docs/en/reference/experimental-flags/sessions.mdx
ascorbic: The reason I changed it was because previously this was in a separate section about cookies, but I removed that because I was moving the docs for the option into a section with other options, and once that was gone the cookies section didn't make much sense to remain. I then pulled in part of the wording from the deleted section.

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -10,8 +10,8 @@ import ReadMore from '~/components/ReadMore.astro';
 
 <p>
 
-**Type:** `SessionConfig`<br />
-**Default:** `undefined`<br />
+**Type:** `boolean`<br />
```
sarah11918, line 13: Just noting, maybe for @ArmandPhilippot that although `experimental.sessions` was introduced in 5.1, it was a config object and not a boolean at that time. Only in 5.3 is it now a boolean.

For an experimental feature I'm *less* concerned about the distinction, but if you have something off the top of your head that perfectly captures this, I'm listening! 😄

ArmandPhilippot: Unfortunately I have no idea... When I wrote this down I was hoping we had a similar case before so we would know how to handle this. :sweat_smile:

Updating `Since` here may cause more confusion than anything else... So, since this is experimental, I think we can afford to leave `5.1`...

If we really want to note this change, perhaps it would make more sense in [Enabling experimental sessions](https://deploy-preview-10915--astro-docs-2.netlify.app/en/reference/experimental-flags/sessions/#enabling-experimental-sessions). Something like:
``Before 5.3, you had to define an object with a driver to enable sessions. Now, you need to set the experimental.session flag to `true`.``
But starting with what we had before does not sound like a good idea. And it doesn't seem to fit properly between the code snippet and the last paragraph. So I guess we can omit this info.

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,83 +22,151 @@ This experimental feature allows you to store and access items such as login sta
...
+    session: {
+      driver: "upstash",
+    },
+    experimental: {
+      session: true,
+    },
+  }
+```
+
 ### Driver options
```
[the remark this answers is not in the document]
ArmandPhilippot: If I understand correctly, `Driver options` are options available for the chosen driver. So it may vary according to the user choice.
We could use `session.options` to be consistent or just "Options", but the user might think these are options for the sessions when they are the driver options (so tied to the configured driver).

src/content/docs/en/reference/experimental-flags/sessions.mdx
```diff
@@ -22,83 +22,151 @@ This experimental feature allows you to store and access items such as login sta
...
+        base: "sessions",
       },
     },
+    experimental: {
+      session: true,
+    },
+  }
+```
+
+### Other session options
```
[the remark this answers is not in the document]
ArmandPhilippot: I think it was to make the distinction between the static options defined by Astro and the options offered by the configured driver. But, yes, perhaps the heading is not necessary. I don't have a strong opinion on this.

src/content/docs/en/reference/experimental-flags/sessions.mdx
ArmandPhilippot: `session.ttl` was already available in `5.1` (or at least before 5.3, I haven't checked the history) but not documented. So, I think we don't need a `Since` here since it was introduced with the feature?

sarah11918: Huh, the docs for it are new. Does this mean it existed but was never documented? If so, oops, but carry on!

src/content/docs/en/reference/experimental-flags/sessions.mdx
sarah11918: Yeah, exactly. You're spot on. I was looking to see whether you had a rabbit to pull out of your hat, but otherwise, I wouldn't go out of my way to mention it. EXPERIMENTAL! EXPERIMENTAL! 😄

src/content/docs/en/reference/experimental-flags/sessions.mdx
ArmandPhilippot: Yes, Matt added it after my initial review:
> It wasn't documented before, but looking at your PR I saw the [ttl option](https://github.com/withastro/astro/blob/3ddeaf0d3d7feeb59705f82e978981f1e7726356/packages/astro/src/core/config/schema.ts#L548) in addition to driver, options and cookies. Should we document this as well?

PR #10920, opened 2025-02-08
Replace the Astro.glob API deprecated in the tutorial

wsafight:
#### Description (required)

There are many places in the tutorial that use the deprecated Astro.glob instead of import.meta.glob.

Astro Discord:  jumpjump_02073


yanthomasdev:
Thanks @wsafight, I'll point out these files have other updates they need, and for that reason we leave them for the native translators of each language to update accordingly.

PR #10924, opened 2025-02-08
fix: update wrong links

jsparkdev:
#### Description (required)

- update wrong links

- Closes #
- Suggested label:


ArmandPhilippot (approved):
Nice catch! I hadn't checked the external links, thanks!

PR #10942, opened 2025-02-11
migrate: replace deprecated output: server refrence with SSR

Keshav-writes-code:
as of astro 5, the output: server is optional and not needed

#### Description (required)

So, the firebase and astro getting started documentation says that you need a astro project with config output: "server" which is no longer needed as of astro 5. so removing it to not cause any confusion for future readers

Just replaced the output:"server" refrence in the top of the document to point to "SSR" enabled project instead  

discord: gdm_music

### DIFF
--- a/src/content/docs/en/guides/backend/google-firebase.mdx
+++ b/src/content/docs/en/guides/backend/google-firebase.mdx
@@ -21,7 +21,7 @@ See our separate guide for [deploying to Firebase hosting](/en/guides/deploy/goo
 ### Prerequisites
 
 - A [Firebase project with a web app configured](https://firebase.google.com/docs/web/setup).
-- An Astro project with [SRR for on-demand rendering](/en/guides/on-demand-rendering/) enabled.
+- An Astro project with an adapter [for on-demand rendering](/en/guides/on-demand-rendering/) enabled.
 - Firebase credentials: You will need two sets of credentials to connect Astro to Firebase:
   - Web app credentials: These credentials will be used by the client side of your app. You can find them in the Firebase console under *Project settings > General*. Scroll down to the **Your apps** section and click on the **Web app** icon.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/backend/google-firebase.mdx
```diff
@@ -21,7 +21,7 @@ See our separate guide for [deploying to Firebase hosting](/en/guides/deploy/goo
 ### Prerequisites
 
 - A [Firebase project with a web app configured](https://firebase.google.com/docs/web/setup).
-- An Astro project with [`output: 'server'` for on-demand rendering](/en/guides/on-demand-rendering/) enabled.
+- An Astro project with [SRR for on-demand rendering](/en/guides/on-demand-rendering/) enabled.
```
sarah11918, line 24:
```suggestion
- An Astro project with an adapter [for on-demand rendering](/en/guides/on-demand-rendering/) enabled.
```
our more standard wording!

sarah11918:
@Keshav-writes-code Are you still interested in continuing with this PR? It cannot be merged as is because the entire guide is written for a project with `'server'` output. So, we cannot simply remove that prerequisite. It **is** a prerequisite **for the instructions that follow**! (Even though `'server'` is not required to use Firebase.)

In order to update this PR and make the suggested change you want to the prerequisites, you will need to also update the code examples for every page that assumes `'server'` on demand rendering by default to include `export const prerender = false // Not needed in 'server' mode`. You can see this in action in other code snippets on the docs in the [section about on demand rendering features](https://docs.astro.build/en/guides/on-demand-rendering/#on-demand-rendering-features) as an example of how we do this.

sarah11918:
Hi @Keshav-writes-code ! Just reminding you of my comment above! Unfortunately, this PR requires a bit more work if you want to continue with it. Please let me know if you're still interested in doing so! (Details are in my earlier comment.)

sarah11918:
I will close this PR due to inactivity for now, but if you'd like to pick it up again later and continue with the requested changes needed, you can reopen it at any time!

PR #10944, opened 2025-02-12
feat: explain the limitations of serialization

### LINKED ISSUE #9557: Document limitations of island prop serialization
### 📚 Subject area/topic

Islands

### 📋 Suggested page

- guides/framework-components
- guides/server-islands

The framework components guide documents the fact that you can't pass functions to client components, but it doesn't document any of the other restrictions. It also doesn't document any of these for server islands. Values must be serializable, and the following types are supported:  plain object, number, string, Array, Map, Set, RegExp, Date, BigInt, URL, Uint8Array, Uint16Array, Uint32Array, Infinity. Circular references and functions are not supported.

_No response_

branberry:
#### Description (required)

Documents the limitation of props for framework components, and server island components. 

- Closes #9557
- Suggested label:


sarah11918:
Thanks for jumping on this issue, Brandon! 🙌 It's looking pretty good, but I think this is now verging into "not a caution" territory, so I'm going to think about about making it just regular body text or at most a "note." Just waiting for some other feedback on this, but very happy to have this!!

src/content/docs/en/guides/framework-components.mdx
```diff
@@ -119,10 +119,15 @@ import Counter from '../components/Counter.svelte';
...
-This is because functions can't be _serialized_ (transferred from the server to the client) by Astro.
+
+:::caution
+You can pass a function as a prop to a framework component, but this only works during server rendering. If you try to use the function in a hydrated component (for example, as an event handler), an error will occur.
+
+This is because functions cannot be _serialized_ (transferred from the server to the client) by Astro. Objects with circular references are also not serializable.
+
+Props provided to hydrated components must be serializable. The following prop types are supported: 
+`plain object`, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
 :::
```
sarah11918, line 131:
```suggestion

Props that are passed to framework functions must be [serialized](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): translated into a format suitable for transfer over a network, or storage. However, for performance and/or security reasons, Astro does not serialize every type of data structure. Therefore, there are some limitations on what can be passed as props.

Notably, you can only pass a function as a prop to a framework component during server rendering. If you try to use the function in a hydrated component (for example, as an event handler), an error will occur. Similarly, passing an object with circular reference is not supported.

The following prop types are supported: 
`plain object`, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
```
OK, after some chat with the maintainers, this is my first pass at more generic docs! Will ask people to weigh in and poke holes in this, tell me exactly how it's wrong so we can make it better! 🙌 

cc @Fryuni @martrapp @OliverSpeir 

(PS I'm not necessarily thrilled with the existing content (middle paragraph here) that we're incorporating, so improving the old content that was here is fair game, too!)

branberry: Great! Thank you so much for doing this. I'll take some time to look into rewriting the existing content as well!

branberry:
Hey Sarah! I rewrote the middle sentence, and I moved it around a bit. Hopefully that seems to be a meaningful improvement!

sarah11918 (commented):
Thank you for taking another pass at the older, existing content that I do think can be improved. It's very helpful!

So given what you proposed, I tried to dig even a little deeper. I do not know that what I wrote is correct, but I think it's the *structure/level of detail* I'd like to have here. So if you can help make the content actually correct, then I think we have a winner!

src/content/docs/en/guides/framework-components.mdx
```diff
@@ -119,13 +119,16 @@ import Counter from '../components/Counter.svelte';
...
-This is because functions can't be _serialized_ (transferred from the server to the client) by Astro.
-:::
 
 
+Props that are passed to framework functions must be [serialized](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): translated into a format suitable for transfer over a network, or storage. However, for performance and/or security reasons, Astro does not serialize every type of data structure. Therefore, there are some limitations on what can be passed as props.
+
+The following prop types are supported: 
+`plain object`, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
+
+Props that are not serializable can still be passed to a framework components during server rendering. Framework components that are hydrated require that props are serializable.
```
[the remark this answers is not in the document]
branberry: I think this is largely accurate. I would say maybe instead of `Non-supported data structures` that we could say `non-serializable data structures`, just so that there's no ambiguity.

src/content/docs/en/guides/framework-components.mdx
```diff
@@ -119,13 +119,16 @@ import Counter from '../components/Counter.svelte';
...
 ```
 
-:::caution[Passing functions as props]
-You can pass a function as a prop to a framework component, but it only works during server rendering. If you try to use the function in a hydrated component (for example, as an event handler), an error will occur.
 
-This is because functions can't be _serialized_ (transferred from the server to the client) by Astro.
-:::
 
 
+Props that are passed to framework functions must be [serialized](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): translated into a format suitable for transfer over a network, or storage. However, for performance and/or security reasons, Astro does not serialize every type of data structure. Therefore, there are some limitations on what can be passed as props.
```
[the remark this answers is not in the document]
branberry: Thank you, Sarah! This looks a lot better, I appreciate the help! I Good call out on the contradiction between paragraph 1 and 3.

I think there's something I missed, though. 

For paragraph 1, I believe we want to say the props *should* be *serializable*. The distinction being that there's no action a user needs to take (which saying it must be serialized implies). The props either are able to be serialized, or they are not. The limitation is inherent with the data type. For example, a function is not serializable, therefore it cannot be passed to interactive client components like you mentioned.

Let me update the PR with some changes, and incorporate the ones you've added.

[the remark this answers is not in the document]
sarah11918: > For paragraph 1, I believe we want to say the props should be serializable. 

This was the distinction that our maintainers discussed to move away from that wording! Apparently, functions *can* be serialized, it's just that Astro chooses not to. That was the basis of my wording re: Astro doesn't serialize everything. So I don't think that's a change that should be made

src/content/docs/en/guides/framework-components.mdx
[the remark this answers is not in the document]
sarah11918: Again, see my point above: even functions ARE serializable (if you're determined enough 😄 ), it's that Astro chooses not to support serializing them. So again, that's not the nuance I was given by our maintainers and the basis for my wording it that way,

src/content/docs/en/guides/framework-components.mdx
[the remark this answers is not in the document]
branberry: Hi Sarah! My apologies the misunderstanding, that makes sense! I really appreciate the feedback. With that said, I believe what you've added looks great!

src/content/docs/en/guides/server-islands.mdx
```diff
@@ -33,6 +33,15 @@ const avatarURL = await getUserAvatar(userSession);
 ---
 <img alt="User avatar" src={avatarURL} />
 ```
+:::caution[Server Island Prop Limitations]
+You can pass a function as a prop to a server island component, but this only works during server rendering. If you try to use the function in a hydrated component (for example, as an event handler), an error will occur.
```
Fryuni, line 37: > If you try to use the function in a hydrated component (for example, as an event handler), an error will occur.

This example seems weird to me. Server islands are not hydrated (at least not in the sense we use anywhere else in the docs for that word, it doesn't add handlers and bindings for interactivity to existing DOM nodes). They get replaced with the content of a separate request when loaded.

The fallback/original content doesn't require any hydration and neither does the code after the swap. After the swap there isn't even any JS related to the server island in the document, contrary to hydration, which is also a win for server island, and might give people the idea that server islands are heavier on their site than it really is.

src/content/docs/en/guides/server-islands.mdx
```diff
@@ -33,6 +33,15 @@ const avatarURL = await getUserAvatar(userSession);
...
+
+This is because functions cannot be _serialized_ (transferred from the server to the client) by Astro. Objects with circular references are also not serializable.
```
Fryuni, line 39: I think this should have the same wording regarding serialization as the client islands. Even more so here to not say that the limitation is regarding sharing information with the client because that is an implementation detail. It could send the information between instances of the server (which might remove some other limitations we currently have) and it would have the same problem.

Fryuni (commented):
The new text for client islands looks great and is technically correct. Good job @branberry !!

I left some comments about the server islands page, mostly it needs the same improvements as the other page.

sarah11918:
Thanks for this additional helpful context, Fryuni! 

(And nice job to get a "technically correct" out of Fryuni, @branberry ! 😄 Would you be willing to tackle the extra sections mentioned above?)

src/content/docs/en/guides/server-islands.mdx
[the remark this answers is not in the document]
branberry: Hey @Fryuni! Thanks again. I'll steal some of what you wrote here as it was very helpful for me to better understand the limitation, specifically this point: `Meaning you can pass functions around on a server island, but never to a server island.`

sarah11918:
Just pinging this for freshness to see if you're still working on this @branberry ! 🙌

branberry:
Hi @sarah11918 ! Thank you for pinging me. Apologies, this has fallen off of my radar. I'll pick this back up soon. Again, sorry about the delay with this one.

sarah11918:
Consider this a friendly ping @branberry ! I know life gets in the way, so you can let us know if you're no longer interested/available if it's helpful to take it off your plate!

branberry:
Hi Sarah, thank you for the very thoughtful ping. I've been meaning to get back around to this, but like you said, life has a way of getting in the way of things 😅 If this could be taken off my plate, that would be great. My apologies for the lack of communication.

sarah11918:
Gentle ping @Fryuni to see if you're still interested in working on this!

sarah11918:
Ok, this was a great PR that needed some love! I took a shot at addressing @Fryuni 's comments so we could maybe "git 'er done" as NWTWWHB (Not Worse Than What We Had Before).

If what I wrote isn't *incorrect*, I'm going to suggest we merge so we can get this helpful content in docs! It can always be updated/improved again later!

src/content/docs/en/guides/server-islands.mdx
```diff
@@ -34,6 +34,15 @@ const avatarURL = await getUserAvatar(userSession);
...
 ```
 
+:::caution[Server Island Prop Limitations]
+Props provided to server island components must be [serializable](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): able to be translated into a format suitable for transfer over a network, or storage. Additionally, for performance and/or security reasons, Astro does not serialize every type of serializable data structure. Therefore, there are some limitations on what can be passed as props to a server island.
+
+Notably, functions cannot be passed to components marked with `server:defer` as they cannot be serialized. Objects with circular references are also not serializable.
+
+The following prop types are supported: 
+`plain object`, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
+:::
```
sarah11918:
```suggestion
### Passing props to server islands

Props provided to server island components must be [serializable](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): able to be translated into a format suitable for transfer over a network, or storage. Additionally, for performance and/or security reasons, Astro does not serialize every type of serializable data structure. Therefore, there are some limitations on what can be passed as props to a server island.

Notably, functions cannot be passed to components marked with `server:defer` as they cannot be serialized. Objects with circular references are also not serializable.

The following prop types are supported: 
`plain object`, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
```
Suggesting this becomes a heading, which can be directly linked to, vs a caution!

src/content/docs/en/guides/framework-components.mdx
```diff
@@ -121,13 +121,16 @@ import Counter from '../components/Counter.svelte';
...
 ```
 
-:::caution[Passing functions as props]
-You can pass a function as a prop to a framework component, but it only works during server rendering. If you try to use the function in a hydrated component (for example, as an event handler), an error will occur.
 
-This is because functions can't be _serialized_ (transferred from the server to the client) by Astro.
-:::
 
 
+Props that are passed to interactive framework components [using a `client:*` directive](/en/reference/directives-reference/#client-directives) must be [serialized](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): translated into a format suitable for transfer over a network, or storage. However, for performance and/or security reasons, Astro does not serialize every type of data structure. Therefore, there are some limitations on what can be passed as props to hydrated components.
```
ascorbic:
```suggestion
Props that are passed to interactive framework components [using a `client:*` directive](/en/reference/directives-reference/#client-directives) must be [serialized](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): translated into a format suitable for transfer over a network, or storage. However Astro does not serialize every type of data structure. Therefore, there are some limitations on what can be passed as props to hydrated components.
```
It's mostly for technical reasons rather than perf or security

src/content/docs/en/guides/framework-components.mdx
```diff
@@ -121,13 +121,16 @@ import Counter from '../components/Counter.svelte';
...
+
+The following prop types are supported: 
+`plain object`, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
```
ascorbic:
```suggestion
plain object, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
```
The others are actual types, but "plain object" isn't. It could be "plain `object`" but I think it's better without it.

src/content/docs/en/guides/server-islands.mdx
```diff
@@ -34,6 +34,15 @@ const avatarURL = await getUserAvatar(userSession);
 <img alt="User avatar" src={avatarURL} />
 ```
 
+### Passing props to server islands
+
+Props provided to server island components must be [serializable](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): able to be translated into a format suitable for transfer over a network, or storage. Additionally, for performance and/or security reasons, Astro does not serialize every type of serializable data structure. Therefore, there are some limitations on what can be passed as props to a server island.
```
ascorbic:
```suggestion
Props provided to server island components must be [serializable](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): able to be translated into a format suitable for transfer over a network, or storage. Additionally, Astro does not serialize every type of serializable data structure. Therefore, there are some limitations on what can be passed as props to a server island.
```

src/content/docs/en/guides/server-islands.mdx
```diff
@@ -34,6 +34,15 @@ const avatarURL = await getUserAvatar(userSession);
...
+
+Notably, functions cannot be passed to components marked with `server:defer` as they cannot be serialized. Objects with circular references are also not serializable.
+
+The following prop types are supported: 
+`plain object`, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
```
ascorbic:
```suggestion
plain object, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
```

src/content/docs/en/guides/framework-components.mdx
```diff
@@ -121,13 +121,16 @@ import Counter from '../components/Counter.svelte';
...
 ```
 
-:::caution[Passing functions as props]
-You can pass a function as a prop to a framework component, but it only works during server rendering. If you try to use the function in a hydrated component (for example, as an event handler), an error will occur.
 
-This is because functions can't be _serialized_ (transferred from the server to the client) by Astro.
-:::
 
 
+Props that are passed to interactive framework components [using a `client:*` directive](/en/reference/directives-reference/#client-directives) must be [serialized](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): translated into a format suitable for transfer over a network, or storage. However Astro does not serialize every type of data structure. Therefore, there are some limitations on what can be passed as props to hydrated components.
```
yanthomasdev: comma nit
```suggestion
Props that are passed to interactive framework components [using a `client:*` directive](/en/reference/directives-reference/#client-directives) must be [serialized](https://developer.mozilla.org/en-US/docs/Glossary/Serialization): translated into a format suitable for transfer over a network, or storage. However, Astro does not serialize every type of data structure. Therefore, there are some limitations on what can be passed as props to hydrated components.
```

src/content/docs/en/guides/framework-components.mdx
```diff
@@ -121,13 +121,16 @@ import Counter from '../components/Counter.svelte';
...
+
+The following prop types are supported: 
+plain object, `number`, `string`, `Array`, `Map`, `Set`, `RegExp`, `Date`, `BigInt`, `URL`, `Uint8Array`, `Uint16Array`, `Uint32Array`, and `Infinity`
+
+Non-supported data structures passed to components, such as functions, can only be used during the component's server rendering and cannot be used to provide interactivity. For example, passing functions to hydrated components is not supported because Astro cannot pass functions from the server in a way that make them executable on the client.
```
yanthomasdev:
```suggestion
Non-supported data structures passed to components, such as functions, can only be used during the component's server rendering and cannot be used to provide interactivity. For example, passing functions to hydrated components is not supported because Astro cannot pass functions from the server in a way that makes them executable on the client.
```

yanthomasdev (commented):
Good work, a few nits from me.

Fryuni (approved):

PR #10947, opened 2025-02-12
Document new HEAD handling

ascorbic:
#### Description (required)

Adds a line to the endpoint docs describing https://github.com/withastro/astro/pull/13210

#### Related issues & labels (optional)

- Closes #
- Suggested label:

PR #10760, opened 2025-01-22
Docs update for includeFiles and excludeFiles options on netlify adapter

dfdez:
**Description:**
Docs update for https://github.com/withastro/astro/pull/13194.

> [!NOTE]  
> This PR remains in draft status pending the merge of these options into the adapter. I will mark it as ready for review once the changes are deployed to production.

PR #10950, opened 2025-02-13
5.3.0 Minor release docs

sarah11918:
Includes the following PRs

#10760 
#10915 
#10947 
#10952 
#10953

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/netlify.mdx
+++ b/src/content/docs/en/guides/integrations-guide/netlify.mdx
@@ -9,6 +9,7 @@ category: adapter
 i18nReady: true
 ---
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
+import Since from '~/components/Since.astro';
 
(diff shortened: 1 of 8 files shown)


src/content/docs/en/guides/integrations-guide/netlify.mdx
```diff
@@ -244,6 +245,89 @@ With [fine-grained cache control](https://www.netlify.com/blog/swr-and-fine-grai
...
+```js title="astro.config.mjs" ins={8-9, 12-13}
+import { defineConfig } from 'astro/config';
+import netlify from '@astrojs/netlify';
+
+export default defineConfig({
+  output: 'server',
+  adapter: netlify({
+    includeFiles: [
+      './data/**/*.json',
+      './config/?(dev|prod).yml'
```
louisescher, line 321:
```suggestion
      './config/?{dev,prod}.yml'
```

I believe this is the wrong syntax? You can try this out on https://globster.xyz

ascorbic: I think it is correct, but I've removed it because it's possibly confusing

PR #10951, opened 2025-02-13
Update view-transitions.mdx

alexgagnon:
#### Description (required)

Add additional note about behavior in browsers that don't support View Transitions.

- Closes #
- Suggested label: 
https://github.com/withastro/astro/issues/13064

Yes

alexgagnon


sarah11918:
Thanks for making a PR to update the docs, @alexgagnon ! This is a helpful piece of information to add!

I've made an edit where I think your ideas could best go. (I didn't delete your note at this time, but my addition is meant to replace your note.)

Firstly, it's never great docs when we have multiple notes in a row! 😅 We can't just have a bunch of stuff competing for people's attention at the top of the page! So, we always try to work content in where on the page it belongs. You correctly identified, and even linked to, the appropriate section, so this is really the place where your information goes! 🙌 

Instead of a note, I've added more context to the explanation of fallback content, and added another example to illustrate. (It might not be the best one, so please suggest something better!). I think the new version gets your information across, and shows being explicit about transitions, relying on Astro's built-in fallback `animate` setting to kick in. And then, the existing example is a second one, to show if you want to override the default fallback. This feels like the right way to add the new information you want to bring.

I'll also note that it's not practical to be listing which browsers do or don't have support while things are constantly changing. Especially when these are not updates that we control or necessarily even know about, it puts our docs at risk of being outdated when someone *else* updates their feature set. So, we would not list browsers, but instead only refer to "browsers without support." I thought somewhere we had a link to a "current browser support status page" but I don't see it in this guide page of our docs. That would be a thing we could link to so people can always see the most updated status of browser support. But we are not going to try to keep up with any kind of list ourselves. 😅 

So see what you think of the general shape of my suggestion, and maybe suggest a more realistic example, and let's figure out how to not use a note at the top of the page, but get all the info you want right here!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -445,7 +448,14 @@ You can opt out of router transitions on any individual form using the `data-ast
 
 ## Fallback control
 
-The `<ClientRouter />` router works best in browsers that support View Transitions (i.e. Chromium browsers), but also includes default fallback support for other browsers. Even if the browser does not support the View Transitions API, Astro will still provide in-browser navigation using one of the fallback options for a comparable experience.
+The `<ClientRouter />` router works best in browsers that support View Transitions (i.e. Chromium browsers), but also includes default fallback support for other browsers. Even if the browser does not support the View Transitions API, Astro's client router can still provide in-browser navigation using one of the fallback options. Depending on browser support, you may need to explicitly set the `name` or `animate` [`transition` directives](#transition-directives) on the elements you wish to navigate for a comparable experience across all browsers.
```
sarah11918, line 451:
```suggestion
The `<ClientRouter />` router works best in browsers that support View Transitions (i.e. Chromium browsers), but also includes default fallback support for other browsers. Even if the browser does not support the View Transitions API, Astro's client router can still provide in-browser navigation using one of the fallback options.

Depending on browser support, you may need to explicitly set the `name` or `animate` [`transition` directives](#transition-directives) on the elements you wish to navigate for a comparable experience across all browsers:
```

alexgagnon (commented):
lgtm

alexgagnon:
Sorry I missed your note, I'll re review when I get a chance

sarah11918:
> However, I do feel having some more explicit information about browser differences at the top is important.

I get that, and in fact, I think maybe the first bit could instead be rewritten because a lot has changed since these docs were first written... both in Astro, and in what native browser view transitions do.

```
 Astro supports view transitions with just a few lines of code. View transitions update your page content without the browser’s normal, full-page navigation refresh and provide seamless animations between pages. Where [browser support for the View Transition API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API#browser_compatibility) is lacking, Astro allows you to [control fallback strategies](#fallback-control) for the best possible experience for all visitors.  
```

What about an intro something like that?

alexgagnon:
Yeah that's perfect :). Would you like me to update the PR or close it and it can be updated in the main repo?

sarah11918:
@alexgagnon Update it here! So that means that this PR would have an updated initial paragraph, as well as new content to the fallback section, and we'd be all good I think!

sarah11918 (approved):
Alright, I think this looks great and is much improved! Thank you @alexgagnon for your efforts and patience, and welcome to Team Docs! 🥳

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -445,7 +444,16 @@ You can opt out of router transitions on any individual form using the `data-ast
 
 ## Fallback control
 
-The `<ClientRouter />` router works best in browsers that support View Transitions (i.e. Chromium browsers), but also includes default fallback support for other browsers. Even if the browser does not support the View Transitions API, Astro will still provide in-browser navigation using one of the fallback options for a comparable experience.
+The `<ClientRouter />` router works best in browsers that support View Transitions (i.e. Chromium browsers), but also includes default fallback support for other browsers. Even if the browser does not support the View Transitions API, Astro's client router can still provide in-browser navigation using one of the fallback options.
+
+Depending on browser support, you may need to explicitly set the `name` or `animate` [`transition` directives](#transition-directives) on the elements you wish to navigate for a comparable experience across all browsers:
```
sarah11918, line 449:
```suggestion
Depending on browser support, you may need to explicitly set the `name` or `animate` [`transition` directives](#transition-directives) on the elements you wish to animate for a comparable experience across all browsers:
```

PR #10954, opened 2025-02-13
[nightly.yml] update PR title for ci config reference docs

sarah11918:
#### Description (required)

This PR updates the nightly auto-generated ci PR title for configuration reference docs to include the word "configuration."

We used to only have one action to pull in "reference" docs, but since we added the "error reference" docs separately, now we have "reference" and "error reference". 

I'm requesting that we instead use "configuration reference" and "error reference" so that I have an easier time telling them apart. 😅


delucis (approved):
LGTM!

PR #10819, opened 2025-01-28
Add Starlight `llms.txt` plugin

delucis:
#### Description (required)

Experimental PR that adds https://delucis.github.io/starlight-llms-txt/ with some configuration to see if the result could be helpful to anyone.

PR #10955, opened 2025-02-13
Update Astro & co.

delucis:
#### Description (required)

This PR updates Astro, Starlight, and the Expressive Code plugin we use to the latest versions.

#10819 also had done this for earlier versions, but splitting this out so we can get it merged more quickly.


dreyfus92 (approved):
🫡

sarah11918 (approved):
Checked a random sampling of different pages, and everything looks fine to me!

PR #10963, opened 2025-02-14
Fix malformed code

korbindeman:
#### Description (required)

A small snippet of the code in the "Content collections" doc is malformed. I have fixed it.


yanthomasdev (approved):
LGTM! Nice catch!

PR #10962, opened 2025-02-14
Replace `entry.render()` with `render(entry)` in `recipes/` & `integrations-guide`

Nin3lee:
#### Description (required)

I notice that entries no longer have a `render()` method, so I replace `entry.render()` with `render(entry)` and update the import sentences as well in following `recipes/` & `integrations-guide`:

- `reading-time.mdx`
- `modified-time.mdx` (include `params` changed)
- `markdoc.mdx`
- `tailwind-rendered-markdown.mdx`

#### Related issues & labels (optional)

- #10521 
- Suggested label: code snippet update

PR #10966, opened 2025-02-15
i18n(ko-KR): update `markdoc.mdx`

jsparkdev:
#### Description (required)

- update `markdoc.mdx`
- #10962

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/ko/guides/integrations-guide/markdoc.mdx
+++ b/src/content/docs/ko/guides/integrations-guide/markdoc.mdx
@@ -117,10 +117,10 @@ Markdoc 파일은 콘텐츠 컬렉션에서만 사용할 수 있습니다. `.mdo
 
 ```astro title="src/pages/why-markdoc.astro"
 ---
-import { getEntryBySlug } from 'astro:content';
+import { getEntry, render } from 'astro:content';
 
-const entry = await getEntryBySlug('docs', 'why-markdoc');
-const { Content } = await entry.render();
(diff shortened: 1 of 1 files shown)


ArmandPhilippot:
Hi @jsparkdev! Not related to this update, but while I was comparing the Korean page with the English one, I noticed a missing heading on the Korean version: https://docs.astro.build/en/guides/integrations-guide/markdoc/#advanced-markdoc-configuration

ko: https://github.com/withastro/docs/blob/d91478b598cbb407585e7c37af31e1b590cf6c2f/src/content/docs/ko/guides/integrations-guide/markdoc.mdx?plain=1#L533-L535
en: https://github.com/withastro/docs/blob/d91478b598cbb407585e7c37af31e1b590cf6c2f/src/content/docs/en/guides/integrations-guide/markdoc.mdx?plain=1#L537-L541

Could you add it to the PR? Thanks!

ArmandPhilippot:
Oh it seems we also miss that in French... 👀 (I should have checked there as well 😄 ) Maybe there was a Lunaria directive at some point that didn't change the status of the file!?

jsparkdev:
@ArmandPhilippot 

Thank you. Good point!

I guess the tracker didn't detect the change.

ArmandPhilippot (approved):
LGTM, thanks for the quick update!

Well, this is weird. It seems that heading was added in #7744 ... In French, we have a heading but level 3 instead of level 2 (that's probably why I didn't notice that before) and not quite the same wording. 👀

PR #10983, opened 2025-02-16
i18n(fr): fix a heading in `integrations-guide/markdoc.mdx`

ArmandPhilippot:
#### Description (required)

Fixes a heading in the French translation of `integrations-guide/markdoc.mdx`. It seems we have an inconsistency in wording and level compared to the English version.

The heading was added in #7744 ... we had a title but level 3 instead of level 2 and we missed the `advanced` part.

- Suggested label: i18n


thomasbnt (approved):
Lgtm

Issue #10945, opened 2025-02-12
Show examples of `env.d.ts` as a module


thegatesdev:
### 📚 Subject area/topic

TypeScript support, middleware, environment variables

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/guides/environment-variables/#intellisense-for-typescript
https://docs.astro.build/en/guides/middleware/#middleware-types

### 📋 Description of content that is out-of-date or incorrect

The documentation examples show the `env.d.ts` file as being a normal TypeScript file, not a module.
This caused two problems for me:

**1:** To get the types in this file to actually apply, I had to add the "src" directory to the tsconfig "include" paths:
```diff
- "include": [".astro/types.d.ts", "**/*"],
+ "include": [".astro/types.d.ts", "**/*", "src"],
```
It looks like non-module scripts are not automatically included otherwise. Although I often see "src/**/*" being used, which also solves this problem, using "**/*" is the default in Astro.

**2:** You cannot import types from other modules using the `import { ... } from` syntax. This is not really a problem for environment variable definitions, because they don't include complex types. But it's a different story with context locals:
```ts
namespace App {
  interface Locals {
    myLocalType: LocalType<unknown>;
  }
}
```
Having the above definition and using `Astro.locals.apolloClient` fails to get the correct type (and thus gives me ESLint warnings) because it isn't imported.

I suggest changing the examples so that this file becomes a module by default, just like the following example:
```ts
import type { ApolloClient } from "@apollo/client/core";

declare global {
  interface ImportMetaEnv {
    readonly MY_VARIABLE: string;
  }

  interface ImportMeta {
    readonly env: ImportMetaEnv;
  }

  namespace App {
    interface Locals {
      myLocalType: LocalType<unknown>;
    }
  }
}
```

Thank you!

ArmandPhilippot:
Hi! Thanks for taking the time to report this issue.

> I had to add the "src" directory to the tsconfig "include" paths:

Are you sure this is needed? It seems weird to me since `**/*` should match `src` as well... And, without imports, using `"include": [".astro/types.d.ts", "**/*"]` works fine for me.

Don't you think this is your other change that fixes your issue?

> as being a normal TypeScript file, not a module.

This is not a normal Typescript file, but an ambient module. So, the problem with your approach is that your `env.d.ts` will no longer be an ambient module but a regular module because of the import. That's why you need to add `declare global`, which can be used in any normal Typescript file.
[...]

PR #10987, opened 2025-02-18
Update middleware.mdx

Odas0R:
The previous syntax was not correct

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/middleware.mdx
+++ b/src/content/docs/en/guides/middleware.mdx
@@ -141,14 +141,19 @@ export const onRequest = (context, next) => {
 To type the information inside `Astro.locals`, which gives you autocompletion inside `.astro` files and middleware code, declare a global namespace in the `env.d.ts` file:
 
 ```ts title="src/env.d.ts"
-declare namespace App {
+declare global {
+  namespace App {
+    /**
+     * Used by middlewares to store information, that can be read by the user via the global `Astro.locals`
+     */
     interface Locals {
-        user: {
-            name: string
-        },
-        welcomeTitle: () => string,
-        orders: Map<string, object>
+      user: {
+        name: string;
+      };
+      welcomeTitle: () => string;
+      orders: Map<string, object>;
     }
+  }
 }
 ```
 


ArmandPhilippot:
Hi! Thank you @Odas0R for taking the time to submit a PR!

The previous syntax is correct. Using `declare global` will turn the file from an ambient module into a regular module (see also #10945). So this is not the solution we want to show in the docs.

But, you're right, we should have an example with an imported type as it seems to be a recurring issue! We'd like to show an example where people could define a type themselves while importing something else with an inline import to keep the file as an ambient module.

In fact, we have an open PR that aims to fix this, so I'm afraid we'll have to close this one. But please feel free to chime in on #10810 if you have a suitable example that could help move this PR forward!

In your case I think you can also avoid `declare global` by using something like:
```ts
namespace Astro {
  type CustomDirectives = import("./elements").AstroClientDirectives

  interface ClientDirectives extends CustomDirectives {
    // ...
  }
}
```

Odas0R:
Hi!

For me, 

```ts
	namespace App {
		/**
		 * Used by middlewares to store information, that can be read by the user via the global `Astro.locals`
		 */
		interface Locals {
			user: User;
		}
	}

```

Does not work. LSP doesn't catch it up. I had to add declare global { } on top of it.

ArmandPhilippot:
Where does `User` come from? Is it an external module? If so, this should work I think:
```ts
namespace App {
  /**
   * Used by middlewares to store information, that can be read by the user via the global `Astro.locals`
   */
  interface Locals {
    user: import("external-module").User;
  }
}
```

Odas0R:
Still doesn't work :')

I have the env.d.ts,

e.g

```ts
/// <reference path="../.astro/types.d.ts" />

import type {User} from "./core/lib/auth/user";

interface ImportMeta {
  env: Record<string, string>;
}

namespace App {
  /**
   * Used by middlewares to store information, that can be read by the user via the global `Astro.locals`
   */
  interface Locals {
    user: import("./core/lib/auth/user").User;
    token: import("./core/lib/auth/token").Token;
  }
}
```

ArmandPhilippot:
Hmm weird. Are you sure the issue is not related to `import type {User} from "./core/lib/auth/user";` being present on top of your file? I think Typescript might be confused because of that...

I made a quick repro, and it seems to work fine:
https://stackblitz.com/edit/astro-locals-typescript?file=src%2Ftypes%2Ffake.ts,src%2Fenv.d.ts,src%2Fpages%2Findex.astro&on=stackblitz

Well, Stackblitz seems to have trouble with Intellisense when using `Astro.locals.user` in `src/pages/index.astro` but if you download the project and open it in VSCode, everything seems to work as expected.

Odas0R:
> I think Typescript might be confused because of that...

Goddam, that was it.

Typical TS moment.

Thank you for your time and help!

ArmandPhilippot:
> It's probably a bug?

I don't think so. With the top-level import (even if it's not used), Typescript treats the file as a regular module, so `declare global` is required. Without the top-level import, the file remains an ambient module, so it's able to resolve types correctly.
Well, that's my interpretation, I could be wrong/inaccurate. I don't consider myself an expert in these things. 😄 

Glad you could resolve your issue! 🎉

PR #10993, opened 2025-02-19
fix: update cli commands in `testing.mdx`

jsparkdev:
#### Description (required)

- replace `pnpm dlx create-*` with `pnpm crete`.
- use the CLIs used in the official documentation for each technology as much as possible.
- update text that was not rendering because it was written in the wrong place.


ArmandPhilippot (approved):
LGTM, thanks!

PR #11019, opened 2025-02-21
feat: document new CLI flag

ematipico:
#### Description (required)

It documents a new CLI flag called `--allowed-hosts`.

I am not very good at this part, I just referenced the Vite documentation. 

 #### For Astro version: `5.4.0`. See astro PR [#13278](https://github.com/withastro/astro/pull/13278)


src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -432,6 +432,12 @@ Sets which network IP addresses the dev server and preview server should listen
 Do not use the `--host` flag to expose the dev server and preview server in a production environment. The servers are designed for local use while developing your site only.
 :::
 
+### `--allowed-hosts`
+
+<p><Since v="5.4.0" /></p>
+
+The hostnames that Astro is allowed to respond to. Refer to the [vite documentation](https://vite.dev/config/server-options.html#server-allowedhosts) for more information.
```
ArmandPhilippot, line 439: Perhaps we can add a little more information based on your changeset:
```suggestion
Specifies the hostnames that Astro is allowed to respond to in `dev` or `preview` modes. Can be passed a comma-separated list of hostnames to allow or `true` to allow any hostname.

Refer to the [Vite documentation](https://vite.dev/config/server-options.html#server-allowedhosts) for more information.
```
I also specified with which modes the flag works to prevent confusion if someone try to use this with the `build` command.

sarah11918: Fantastic edit, Armand! I would just suggest "list of hostnames or `true`" (vs the extra "to allow") since we've said that word a bunch of times already, but your edits are great!

sarah11918 (commented):
Fun fact: I just had to configure vite.server.allowedHosts yesterday while working on my site so that Gitpod could display the preview in the browser 😄 

Suggesting we accept Armand's edit (mostly word for word) if that sounds good to you, @ematipico !

src/content/docs/en/reference/cli-reference.mdx
ArmandPhilippot: Looks better, thanks! The two `to allow` almost next to each other was indeed a bit "heavy". I'll update the suggestion.

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -432,6 +432,14 @@ Sets which network IP addresses the dev server and preview server should listen
 Do not use the `--host` flag to expose the dev server and preview server in a production environment. The servers are designed for local use while developing your site only.
 :::
 
+### `--allowed-hosts`
+
+<p><Since v="5.4.0" /></p>
+
+Specifies the hostnames that Astro is allowed to respond to in `dev` or `preview` modes. Can be passed a comma-separated list of hostnames or `true` to allow any hostname.
+
+Refer to the [Vite documentation](https://vite.dev/config/server-options.html#server-allowedhosts) for more information.
```
sarah11918, line 441:
```suggestion
Refer to [Vite's `allowHosts` feature](https://vite.dev/config/server-options.html#server-allowedhosts) for more information, including security implications of allowing hostnames.
```
Noting that at the Vite link (and, when researching allowed hosts in general), there tend to be warnings *not* to just enable `true` for security reasons. 

Do we similarly want guidance/warnings about this ourselves? A hint of something that exists in the Vite docs like above? Nothing?

ematipico: I leave the decision to you. In my initial design, I wanted to defer the documentation to `vite`, but if that's not the case, then I think we should add a warning

sarah11918 (approved):
Looks great! (And, now I'll be able to remove a `vite` config in my own project... 😄 )

yanthomasdev (approved):

PR #11031, opened 2025-02-23
fix: update `install-and-setup.mdx`

jsparkdev:
#### Description (required)

update incorrect indentation and whitespace and case sensitivity

PR #11044, opened 2025-02-24
i18n(fr): update `install-and-setup.mdx`

ArmandPhilippot:
#### Description (required)

Adds changes from #11031 to the French translation of `install-and-setup.mdx` (Astro was already capitalized, so this only adds the other changes).

- Suggested label: i18n


yanthomasdev (approved):
LGTM!

PR #11051, opened 2025-02-25
Document new programmatic APIs and options

Fryuni:
#### Description

- Document API changes from withastro/astro#12052

### DIFF
--- a/src/content/docs/en/reference/programmatic-reference.mdx
+++ b/src/content/docs/en/reference/programmatic-reference.mdx
@@ -144,7 +144,7 @@ Returns a `Promise` that resolves once all pending requests have been fulfilled 
 
 <p>
 
-**Type:** `(inlineConfig: AstroInlineConfig) => Promise<void>`
+**Type:** `(inlineConfig: AstroInlineConfig, options?: BuildOptions) => Promise<void>`
 </p>
 
 Similar to [`astro build`](/en/reference/cli-reference/#astro-build), it builds your site for deployment.
@@ -156,6 +156,37 @@ await build({
   root: "./my-project",
 });
 ```
+
+### `BuildOptions`
+
+```ts
+export interface BuildOptions {
+	devOutput?: boolean;
+	teardownCompiler?: boolean;
+}
+```
+
+#### `devOutput`
+
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `false`
+</p>
+
+Output a development-based build similar to code transformed in `astro dev`. This can be useful to test build-only issues with additional debugging information included.
+
+#### `teardownCompiler`
+
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `true`
+</p>
+
+Teardown the compiler WASM instance after build. This can improve performance when building once, but may cause a performance hit if building multiple times in a row.
+
+When building multiple projects in the same execution (e.g during tests), disabling this option can greatly increase performance and reduce peak memory usage at the cost of higher sustained memory usage.
 
 ## `preview()`
 
@@ -246,4 +277,59 @@ import { sync } from "astro";
 await sync({
   root: "./my-project",
 });
+```
+
+## `mergeConfig()`
+
+<p>
+
+**Type:** `<T extends AstroConfig | AstroInlineConfig>(config: T, overrides: DeepPartial<T>) => T`
+</p>
+
+Imported from `astro/config`, merges a partial Astro configuration on top of an existing, valid, Astro configuration.
+
+- General options are overriden.
+- Integrations and remark plugins are appended to the existing list.
+- Vite options are merged using [Vite's own `mergeConfig` function](https://vite.dev/guide/api-javascript#mergeconfig) with default `isRoot` flag.
+- Lazy options passed as functions are wrapped into new functions that recursively merge the return values from both configurations with these same rules.
+
+```ts
+import { mergeConfig } from "astro/config";
+
+const config = mergeConfig(
+  {
+    integrations: [tailwind()],
+  },
+  {
+    integrations: [mdx()],
+  }
+);
+```
+
+## `validateConfig()`
+
+<p>
+
+**Type:** `(userConfig: any, root: string, cmd: string): Promise<AstroConfig>`
+</p>
+
+Imported from `astro/config`, validates an object as if it was exported from `astro.config.mjs` and imported by Astro.
+
+
+It takes the following arguments: 
+- The configuration to be validated.
+- The root directory of the project.
+- The Astro command that is being executed (`build`, `dev`, `sync`, etc.)
+
+The returned promise resolves to the validated configuration, filled with all default values appropriate for the given Astro command.
+
+```ts
+import { validateConfig } from "astro/config";
+
+const config = await validateConfig({
+  integrations: [tailwind()],
+}, "./my-project", "build");
+
+// defaults are applied
+await rm(config.outDir, { recursive: true, force: true });
 ```


src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +277,59 @@ import { sync } from "astro";
...
+
+## `mergeConfig()`
+
+<p>
+
+**Type:** `<T extends AstroConfig | AstroInlineConfig>(config: T, overrides: DeepPartial<T>) => T`
+</p>
+
+Imported from `astro/config`, merges a partial Astro configuration on top of an existing, valid, Astro configuration.
+
```
sarah11918, line 290:
```suggestion

The merge behaves according to the following rules:

```
Do we need some kind of sentence to introduce the bullet points maybe? (And, maybe that's not the right sentence -- just off the top of my head. It seems like we should set some context for "list of bullet points.")

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +277,59 @@ import { sync } from "astro";
...
+- General options are overriden.
```
sarah11918, line 291:
```suggestion
- General options in the existing config are overriden.
```
(Also, I'm not sure what "general" options are? Is that a specific thing? If not, is there a proper name for these options? "top-level"?)

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +277,59 @@ import { sync } from "astro";
...
+- Integrations and remark plugins are appended to the existing list.
+- Vite options are merged using [Vite's own `mergeConfig` function](https://vite.dev/guide/api-javascript#mergeconfig) with default `isRoot` flag.
```
sarah11918, line 293:
```suggestion
- Vite options are merged using [Vite's own `mergeConfig` function](https://vite.dev/guide/api-javascript#mergeconfig) with the default `isRoot` flag.
```

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +277,59 @@ import { sync } from "astro";
...
+- Lazy options passed as functions are wrapped into new functions that recursively merge the return values from both configurations with these same rules.
```
sarah11918, line 294: Similar comment here, is "lazy options" something that is generally understood as a thing? I'm assuming this means "passing a function to set an option, which therefore needs to be worked into a process" vs "passing an option directly by providing a new config object to merge with the existing one"?

I'm trying to figure out the exact parsing of this to know how best to evaluate this line. Is it... ?

- Lazy options (WHICH MEANS passed as functions which will set a config value vs set one directly)
- Lazy options THAT HAPPEN TO BE PASSED AS FUNCTIONS (but they could totally be passed a different way)
- something entirely else?

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +277,59 @@ import { sync } from "astro";
...
+
+```ts
```
sarah11918, line 296: Can we have a line explaining what this example is supposed to be showing? We've talked about a bunch of different things above including passing a function. Should we also show an example of a plugin or even function being passed here, or is that going to look too complicated?

Something like: 

The following example uses `mergeConfig` to append first, Astro's Tailwind and then MDX integrations to an existing config.

(I'm inferring from the example that if you have multiple integrations to add, you have to add multiple integrations separately? You can't combine it into one `integrations: []` array? This is maybe worth mentioning somewhere, too, that every single new thing you want to merge into an existing config must be passed as a separate (top-level?) object?)

Seeing this all now, I'm also tempted to reorder the entry as:

```
## `mergeConfig()`

<p>Type stuff</p>

Imported from `astro/config`, merges a partial Astro configuration on top of an existing, valid, Astro configuration.

Code Example so people just *see* it immediately

Each config value you wish to add or overwrite must be added as its own, individual config object.
For example, multiple integrations must be added separately, in the order in which you want to append them to the existing config. 

(ASIDE: what happens if you want an integration added earlier?? Does this ever happen/matter??)

Merging behavior follows these rules + list of bullet points
```

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +277,59 @@ import { sync } from "astro";
...
+
+
+It takes the following arguments: 
+- The configuration to be validated.
+- The root directory of the project.
+- The Astro command that is being executed (`build`, `dev`, `sync`, etc.)
+
+The returned promise resolves to the validated configuration, filled with all default values appropriate for the given Astro command.
+
+```ts
```
sarah11918, line 326: Noting that unlike the previous entry, I'm not going to suggest re-ordering here to put the code sample earlier because unlike before, the earlier text explains arguments, return value etc.

Previously, the additional text wasn't as much *what you absolutely needed to know before writing the code example* but a subsequent description of how that was going to work.

This feels like the correct order here, and if that means you'd prefer not to reorder above to have them match, that's fine.

(But, if I were looking at each one on its own, `mergeConfig()` feels like something where I'd naturally put the example earlier, because it's easier to grok just by looking at it, whereas someone looking up `validateConfig()` probably is checking "what are all the arguments? etc." and needs that info sooner.)

sarah11918 (commented):
Looks fantastic, Fryuni!!

Some comments below, including that we should include `<Since />`  for the things you're adding new. I know you said some stuff was previously undocumented? If any of these that you're adding are that, we *could* go back and look for the version they were introduced. But at the very least, let's identify what's new here.

sarah11918:
(Also noting that you can check #11019 for an example of filling in the PR template for a minor feature! It should have been included in the template with things you could uncomment, with a spot for the astro PR, the minor release version etc.)

src/content/docs/en/reference/programmatic-reference.mdx
Fryuni: Anything that doesn't match one of the other rules. Even if they are nested.

src/content/docs/en/reference/programmatic-reference.mdx
Fryuni: First option. Some options allow passing either the value or a function returning the value. For example:

```ts
defineConfig({
  server: {
    headers: {
      foo: 'bar',
    },
  },
})

defineConfig({
  server: ({command}) => ({
    headers: {
      astroCommand: command,
    },
  }),
})
```

src/content/docs/en/reference/programmatic-reference.mdx
sarah11918: OK, then should this point come *last*? As in:

- All other options override the existing config

src/content/docs/en/reference/programmatic-reference.mdx
sarah11918: Great, thanks for the clarification! Then maybe something like one of these just to avoid the ambiguous parsing...

- Lazy options (passed as functions) are wrapped into new functions ... (if you want to emphasize the term "lazy options" because that's particularly meaningful)
- Options configured by passing functions are wrapped into new functions ... (if you don't care about "lazy" and instead want to describe what the config looks like)

Also just want to check on the different "effect" or end result of this vs. just passing a value directly:
- is there an example of a situation where passing a function results in something different than overriding an existing config with a directly passed value?
- How is knowing that this process creates a new, recursive function helpful to the reader? Would they do anything differently because they know this? (In other words, is this more a "fun fact" / implementation detail than something people need to know to make sure they merge configs properly?
- Are there any particular gotchas/caveats/warnings about passing a config option this way that people should be aware of? (Other than I guess not all options support this method?) Does this "slow down" the process, or resolve itself in ways that might conflict with anything else?

src/content/docs/en/reference/programmatic-reference.mdx
Fryuni: I'll update the example to show a case with every item.

src/content/docs/en/reference/programmatic-reference.mdx
Fryuni: > How is knowing that this process creates a new, recursive function helpful to the reader? Would they do anything differently because they know this? (In other words, is this more a "fun fact" / implementation detail than something people need to know to make sure they merge configs properly?

Most "merge" functions in the wild would just replace the function with the new one instead of merging the result. That is the behavior of a recursive use of the spread operator (`{...current, ...overrides}`)

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +277,94 @@ import { sync } from "astro";
...
+## `mergeConfig()`
+
+<p>
+
+**Type:** `<T extends AstroConfig | AstroInlineConfig>(config: T, overrides: DeepPartial<T>) => T`
+</p>
+
+Imported from `astro/config`, merges a partial Astro configuration on top of an existing, valid, Astro configuration.
+
+The return value is a valid Astro config combining both values such that:
```
sarah11918, line 291:
```suggestion
`mergeConfig()` accepts two Astro config objects, and the return value is a valid Astro config combining both values such that:
```
Is this correct to say? At first, I didn't realize you needed to give it two different configs to merge; I thought it was "merge whatever you already have with this new thing I'm giving you", so it's maybe helpful to say this out in words (not just in the type declaration?)

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +277,94 @@ import { sync } from "astro";
...
+
+- Arrays are concatenated (including integrations and remark plugins).
+- Objects are merged recursively.
+- Vite options are merged using [Vite's own `mergeConfig` function](https://vite.dev/guide/api-javascript#mergeconfig) with the default `isRoot` flag.
+- Options that can be provided as functions are wrapped into new functions that recursively merge the return values from both configurations with these same rules.
+- All other options override the existing config.
+
+```ts
```
sarah11918, line 299: This example looks great!

src/content/docs/en/reference/programmatic-reference.mdx
Fryuni: The first argument is a config, the second is any part of a config, everything is optional on it (that is the `DeepPartial` in the types). Sending another option!

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -246,4 +279,96 @@ import { sync } from "astro";
...
+
+<p>
+
+**Type:** `<T extends AstroConfig | AstroInlineConfig>(config: T, overrides: DeepPartial<T>) => T`
+<Since v="5.4.0" />
+</p>
+
+Imported from `astro/config`, merges a partial Astro configuration on top of an existing, valid, Astro configuration.
+
+`mergeConfig()` accepts an Astro config object and a partial config (any portion of a config, everything is optional), and returns a valid Astro config combining both values such that:
```
sarah11918, line 294:
```suggestion
`mergeConfig()` accepts an Astro config object and a partial config (any set of valid Astro config options), and returns a valid Astro config combining both values such that:
```
I like it! If I tweak it a tiny bit, would this work?

Fryuni: Yes

PR #11055, opened 2025-02-25
Use `PackageManagerTabs` for CLI commands in `guides/environment-variables`

### LINKED ISSUE #10935: Using different environments -- command is erroneous
### 📚 Subject area/topic

environment variables

https://docs.astro.build/en/guides/environment-variables/#setting-environment-variables

In order to use different env-files, current documentation states the following command
"astro dev --mode staging"
where 'staging' refers to an env-file '.env.staging'.

This command does not work. Imho the reason for that bug is the command is not complete.
When changed to
"astro dev -- --mode staging" [in my case prefixed with an "npm run " i now can switch between different env-files via the mode-option.

In Discord you find the same error, see https://discord.com/channels/830184174198718474/1306650092886954107

_No response_

ArmandPhilippot:
#### Description (required)

Updates the [`.env files` section](https://docs.astro.build/en/guides/environment-variables/#env-files) in `guides/environment-variables` to use the `PackageManagerTabs` component.

`npm` requires an extra `--` before any flag so the previous format was confusing for `npm` users.

- Closes #10935

### DIFF
--- a/src/content/docs/en/guides/environment-variables.mdx
+++ b/src/content/docs/en/guides/environment-variables.mdx
@@ -89,16 +89,44 @@ The `astro dev` and `astro build` commands default to `"development"` and `"prod
 
 This allows you to run the dev server or build your site connecting to different APIs:
 
-```bash
-# Run the dev server connected to a "staging" API
-astro dev --mode staging
+<PackageManagerTabs>
+ <Fragment slot="npm">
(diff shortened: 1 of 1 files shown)


delucis (approved):
Great catch! Thank you @ArmandPhilippot 💖

PR #11064, opened 2025-02-26
fix: update prettier setup in `editor-setup.mdx`

jsparkdev:
#### Description (required)

- update the install and run CLI commands to respect the [Prettier documentation](https://prettier.io/docs/install).
- according to the [Prettier documentation](https://prettier.io/docs/configuration), the `JSON` format is the first to be documented. So I don't think it's necessary to use a different format.

if you have a different opinion on this, please let me know!

### DIFF
--- a/src/content/docs/en/editor-setup.mdx
+++ b/src/content/docs/en/editor-setup.mdx
@@ -98,11 +98,11 @@ To add support for formatting `.astro` files outside of the editor (e.g. CLI) or
       </Fragment>
     </PackageManagerTabs>
 
-2. Create a `.prettierrc.json` config file at the root of your project and add `prettier-plugin-astro` to it.
+2. Create a `.prettierrc` configuration file (or `.prettierrc.json`, `.prettierrc.mjs`, or [other supported formats](https://prettier.io/docs/configuration)) in the root of your project and add `prettier-plugin-astro` to it.
 
     In this file, also manually specify the parser for Astro files.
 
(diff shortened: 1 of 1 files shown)


ArmandPhilippot (approved):
LGTM!

I'm a bit torn about `--save-exact` but it's a Prettier recommendation:

> Even a patch release of Prettier can result in slightly different formatting, so you wouldn’t want different team members using different versions and formatting each other’s changes back and forth.

So I guess it makes sense to follow their recommendations since it can confuse less experienced people.

> the JSON format is the first to be documented

If we choose to use the first config file Prettier looks for, shouldn't we use `.prettierrc` (without extension) instead of `.prettierrc.json`?

jsparkdev:
@ArmandPhilippot 

Thanks!

Would it be better to specify `.prettierrc` in the sentence, and indicate that other extensions are available?

ArmandPhilippot:
I'm too slow sorry... 😆 I was saying: We could say something like `Create a [Prettier config file](https://prettier.io/docs/configuration) at the root` to avoid listing the different formats and just pick a filename for the code snippet. But your solution works too I think.

jsparkdev:
Merge the current commit? Or is there something else that needs to be updated? If so, please let me know!

ArmandPhilippot (approved):
I believe you can merge, LGTM!

jsparkdev:
@ArmandPhilippot 

Thanks for the great feedback.

PR #11065, opened 2025-02-26
feat(vercel): update `exclude` API for upcoming new feature

ematipico:
#### Description (required)

This PR updates the Vercel integration guide by adding a new supported example for `isr.exclude`.

The suggestion of the change was provided by @sarah11918 in the PR. Please note that the version number proposed was incorrect. My suggestion refers to the version number of `@astrojs/vercel`.


src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -277,7 +277,11 @@ export default defineConfig({
             // A secret random string that you create.
             bypassToken: "005556d774a8",
             // Paths that will always be served fresh.
-            exclude: [ "/api/invalidate", "/posts/[...slug]" ]
+            exclude: [
+              '/preview', 
+              '/auth/[page]',
+              /^\/api\/.+/ // Regular expressions supported since v8.1.0
```
ascorbic, line 283: Maybe avoid confusion by saying `@astrojs/vercel@8.1.0`?

ematipico: I mean, it's the page of the Vercel integration, I believe there's no confusion 😅

sarah11918: ugh, there is totally going to be confusion. lol I don't love how long that makes it, but we do need to say it.

sarah11918, line 283:
```suggestion
              /^\/api\/.+/ // Regular expressions supported since @astrojs/vercel@8.1.0
```
I think we can get away without back ticks because this is a code block anyway and won't render as inline code? Not opposed either way.

sarah11918 (commented):
Thanks for getting the version of the *integration* right, Ema! I do think Matt's right... some people will read and think version of Astro, so we probably do need to specify the integration itself. Does the below work?

ematipico:
Updated the version wording and the highlighting

sarah11918 (approved):
Thank you, looks great!

PR #11075, opened 2025-02-27
fix: update `dev-toolbar.mdx`

jsparkdev:
#### Description (required)

- The terms `development toolbar`, `developer toolbar`, and `dev toolbar`, which all mean the same thing, have been unified into `dev toolbar`, which is the most commonly used term in this document.
- Added missing spaces. Changed `enabled:true` to `enabled: true`

### DIFF
--- a/src/content/docs/en/guides/dev-toolbar.mdx
+++ b/src/content/docs/en/guides/dev-toolbar.mdx
@@ -9,7 +9,7 @@ While the dev server is running, Astro includes a dev toolbar at the bottom of e
 
 This toolbar includes a number of useful tools for debugging and inspecting your site during development and can be [extended with more dev toolbar apps](#extending-the-dev-toolbar) found in the integrations directory. You can even [build your own toolbar apps](/en/recipes/making-toolbar-apps/) using the [Dev Toolbar API](/en/reference/dev-toolbar-app-reference/)!
 
-This toolbar is enabled by default and appears when you hover over the bottom of the page. It is a dev tool only and will not appear on your published site.
+This toolbar is enabled by default and appears when you hover over the bottom of the page. It is a development tool only and will not appear on your published site.
 
 ## Built-in apps
 
--- a/src/content/docs/en/guides/dev-toolbar.mdx
+++ b/src/content/docs/en/guides/dev-toolbar.mdx
@@ -1,15 +1,15 @@
 ---
 title: Dev toolbar
-description: A guide to using the developer toolbar in Astro
+description: A guide to using the dev toolbar in Astro
 i18nReady: true
 ---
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/dev-toolbar.mdx
```diff
@@ -1,15 +1,15 @@
...
 ---
 import RecipeLinks from "~/components/RecipeLinks.astro";
 
-While the dev server is running, Astro includes a development toolbar at the bottom of every page in your local browser preview.
+While the dev server is running, Astro includes a dev toolbar at the bottom of every page in your local browser preview.
 
 This toolbar includes a number of useful tools for debugging and inspecting your site during development and can be [extended with more dev toolbar apps](#extending-the-dev-toolbar) found in the integrations directory. You can even [build your own toolbar apps](/en/recipes/making-toolbar-apps/) using the [Dev Toolbar API](/en/reference/dev-toolbar-app-reference/)!
 
-This toolbar is enabled by default and appears when you hover over the bottom of the page. It is a development tool only and will not appear on your published site.
+This toolbar is enabled by default and appears when you hover over the bottom of the page. It is a dev tool only and will not appear on your published site.
```
sarah11918, line 12:
```suggestion
This toolbar is enabled by default and appears when you hover over the bottom of the page. It is a development tool only and will not appear on your published site.
```
This isn't naming the toolbar itself, and should be spelled out in full!

sarah11918 (commented):
Thanks for making things consistent! I will note that since "dev" is an abbreviation, it is usual for the first use to be the full word, so people know what the word is/means, even if then the abbreviation is most commonly used. (We always do this with acronyms like CMS/content management system.) This would be why the page title was written out in full, to follow this pattern.

It's also generally OK to mix a full word and its abbreviation or acronym for sentence variety! This kind of repetition of both the full word and the acronym is can be helpful to establish the connection between the longer and shorter words or phrases. And, sometimes sentences will sound very repetitive with the exact same phrase over and over again. 

So, the lack of complete consistency isn't a *problem* here, but since "dev toolbar" is almost like a "name", I do think most of your changes here are fine. (There's just one time that's it's not used for the toolbar itself, and I point that out below.) 🚀

src/content/docs/en/guides/dev-toolbar.mdx
jsparkdev: Good catch. Thanks!

jsparkdev:
@sarah11918 

Thank you!

So should I keep line 8 as `development`?
Or should I update something else?

sarah11918 (approved):
I think all the other changes in this PR to `dev` are fine! Good to go!

PR #11077, opened 2025-02-27
fix(tutorial): update the highlighting color for added code

### LINKED ISSUE #11039: TypeError when copying code from the blog tutorial
### 📚 Subject area/topic

"Build a blog" introductory tutorial

https://docs.astro.build/en/tutorial/5-astro-api/2/#use-props-in-dynamic-routes

Step 2 of "Use props in dynamic routes" instructs us to insert the following code into our `[tag].astro` file: 
`const filteredPosts = posts.filter((post: any) => post.frontmatter.tags?.includes(tag));`
(after defining `posts` this way: `const { posts } = Astro.props;`)

However, when I put this line into my file, it results in a TypeError: "Cannot read properties of undefined (reading 'filter')"
Likewise, VS Code gives the warning: "Property 'filter' does not exist on type 'never'"

It seems to me that the author probably meant to use `allPosts` here, which was defined earlier in this page as a glob import, rather than `posts`. However, one more complication is that the tutorial code has us define `allPosts` inside of `getStaticPaths()`, which leaves it unavailable to reference in these later lines. To make this section of the tutorial work, I also had to edit the code from the earlier section "Create pages dynamically" and move the declaration of `allPosts` out into this file's global scope.

After making those changes, it runs without error, but I'm unsure if it's actually still doing what the tutorial intended it to do. I assume there was some kind of good reason for the author to declare `allPosts` inside the function originally?

_No response_

ArmandPhilippot:
#### Description (required)

Updates the highlighting color in a code snippet of the tutorial. We used two different colors for added code which caused confusion for some.

|Before|After|
|---|---|
|before|after|

- Closes #11039

### DIFF
--- a/src/content/docs/ja/guides/dev-toolbar.mdx
+++ b/src/content/docs/ja/guides/dev-toolbar.mdx
@@ -1,0 +1,85 @@
+---
+title: 開発ツールバー
+description: Astroでの開発ツールバーについてのガイド。
+i18nReady: true
+---
+import RecipeLinks from "~/components/RecipeLinks.astro";
+
+開発サーバーが実行されている間、Astroは各ページの下部付近にデベロッパーツールバーを表示します。
+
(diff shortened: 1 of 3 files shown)


yanthomasdev (approved):
LGTM!

PR #11112, opened 2025-03-01
Update actions.mdx

vladstudio:
There's an error on line 157 - the object needs to be spread.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/actions.mdx
+++ b/src/content/docs/en/guides/actions.mdx
@@ -154,7 +154,7 @@ import { user } from './user';
 
 export const server = {
   myAction: defineAction({ /* ... */ }),
-  user,
+  ...user,
 }
 ```
 


ArmandPhilippot:
Hi, and thanks for taking the time to fix the docs! However, I'm not sure this change is correct.
If we use the spread syntax, what is below the code snippet will be wrong since the `user` key will no longer exist:

```md
Now, all of your user actions are callable from the actions.user object:

*    actions.user.getUser()
*    actions.user.createUser()
```

When you say there is an error, are you sure you are using `actions.user.getUser()` and not `actions.getUser()`?

vladstudio:
Oh oops! I was not careful enough indeed, and was trying to access `actions.getUser()`. Sorry for the confusion, and thanks for catching.

PR #11114, opened 2025-03-01
Fix Broken Link in Zerops Guide

nermalcat69:
#### Description (required)

The current doc link for zerops yml is broken as zerops docs routes were recently changed.

- Closes #
- Suggested label:


sarah11918 (approved):
Thanks for updating, and keeping docs correct! :muscle:

PR #11127, opened 2025-03-03
fix: type annotation for `render()` in `astro:content`

### LINKED ISSUE #11123: Runtime API: Incorrect `render()` Type in `astro:content`
### 📚 Subject area/topic

Runtime API / astro:content

https://docs.astro.build/en/reference/modules/astro-content/#render

In the documentation, the `render()` function is defined as `() => Promise<RenderedEntry>`, which is incorrect—it should take an entry parameter.

I want to fix this, but I'm uncertain about the correct entry type.

In the generated type definitions, the function is declared as:

```typescript
function render<C extends keyof AnyEntryMap>(
		entry: AnyEntryMap[C][string],
	): Promise<RenderResult>;
```

Given this, should the entry type be annotated as `CollectionEntry<CollectionKey>`, `CollectionEntry<TCollectionName>`, `TCollectionEntry`, or something else?

maxchang3:
#### Description

The `render()` function was previously annotated as `() => Promise<RenderedEntry>`, which is incorrect—it should take an `entry` parameter.  

This PR updates the type annotation to reflect the correct function signature:  
`(entry: CollectionEntry) => Promise<RenderedEntry>`.  

#### Related issues

- Closes #11123

### DIFF
--- a/src/content/docs/en/reference/modules/astro-content.mdx
+++ b/src/content/docs/en/reference/modules/astro-content.mdx
@@ -197,7 +197,7 @@ const enterpriseRelatedPosts = await getEntries(enterprisePost.data.relatedPosts
 
 <p>
 
-**Type:** `() => Promise<RenderedEntry>`
+**Type:** `(entry: CollectionEntry) => Promise<RenderedEntry>`
 <Since v="5.0.0" />
 </p>
 


ArmandPhilippot (approved):
Nice catch, thanks for fixing that! 🙌🏽

PR #11146, opened 2025-03-04
fix(content-collections): improve `reference()` docs and minor fixes

### LINKED ISSUE #10774: Explain how the `reference()` helper function transforms the data
### 📚 Subject area/topic

Content Collections

https://docs.astro.build/en/guides/content-collections/#defining-collection-references
https://docs.astro.build/en/guides/content-collections/#accessing-referenced-data

The documentation for the `reference()` utility function is lacking some info. From the description, I only assumed it adds type-safety to referenced items (possibly validating that an item with that ID exists, still not sure if it actually does that or if that's even technically possible). However, it also _transforms_ the property.

For example, if I define a property using the reference helper:

```js
schema: z.object({
    crop: reference('crops'),
}),
```

And then specify an ID in a collection item:

```json
{
    "crop": "grapevine"
}
```

Now if I access that property, it's not a string any more, but an object containing both the collection and id:

```js
myItem.data.crop; // { collection: 'crops', id: 'grapevine' }
```

This is what allows me to use `getEntry` without specifying a collection:

```js
const crop = await getEntry(myItem.data.crop);
```

I was surprised by this and actually thought there was a bug (https://github.com/withastro/astro/issues/13052), because I didn't asume the `reference()` utility would transform the value of the property.

### Suggested improvements

- The documentation on defining collection references should mention that the `reference()` function transforms the property to an object, and how to work with that in a template
- The documentation on accessing referenced data should explain that it's possible to use `getEntry` without specifying a collection because of the above-mentioned transforms. Right now, this is easy to miss in the code example.

_No response_

ArmandPhilippot:
#### Description (required)

Updates both `guides/content-collections` and `reference/modules/astro-content`:
* add explanations regarding the `reference()` output (#10774)
* add missing imports in code snippets
* update the `CollectionKey` example with I believe a more useful example (the previous example was basically the same thing as the built-in `getCollection` function... and using the same name so it could be a recursive function)

- Closes #10774

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -295,7 +295,7 @@ All [Zod schema methods](https://zod.dev/?id=schema-methods) (e.g. `.parse()`, `
 
 Collection entries can also "reference" other related entries. 
 
-With the `reference()` function from the Collections API, you can define a property in a collection schema as an entry from another collection. For example, you can require that every `space-shuttle` entry includes a `pilot` property which uses the `pilot` collection's own schema for type checking, autocomplete, and validation.
+With the [`reference()` function](/en/reference/modules/astro-content/#reference) from the Collections API, you can define a property in a collection schema as an entry from another collection. For example, you can require that every `space-shuttle` entry includes a `pilot` property which uses the `pilot` collection's own schema for type checking, autocomplete, and validation.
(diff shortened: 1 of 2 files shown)


sarah11918 (approved):
Pretty fantastic, Armand! 💪 Thank you for this!

PR #11149, opened 2025-03-04
Update Astro to latest

debiru:
#### Description (required)

This PR updates esbuild and the packages that depend on esbuild.

- [esbuild (<= 0.24.2) enables any website to send any requests to the development server and read the response
](https://github.com/advisories/GHSA-67mh-4wv8-2f99)

Changed to use esbuild 0.25.0

The `overrides` section is for the `tsm` package.

- [Astro 5.4.0 release](https://github.com/withastro/astro/pull/13318)
- [Astro 5.4.1 release](https://github.com/withastro/astro/pull/13337)
- [tsm dependencies](https://github.com/lukeed/tsm/blob/master/package.json)


package.json
```diff
@@ -84,6 +84,9 @@
         "react-dom",
         "vite"
       ]
+    },
+    "overrides": {
+      "esbuild": "^0.25.0"
```
delucis, line 89: Thanks for looking into this @debiru! This shouldn‘t be needed though. Can you explain why you added it?

delucis: Oh, sorry, I skipped over the line in the PR comment:

> The `overrides` section is for the `tsm` package.

Do you think this upgrade is necessary? `tsm` is only used to execute a few scripts we have in CI, not run esbuild’s dev server, so I suspect it’s likely harmless to leave in place.

debiru: @delucis 

> Do you think this upgrade is necessary? tsm is only used to execute a few scripts we have in CI, not run esbuild’s dev server, so I suspect it’s likely harmless to leave in place.

That is certainly true from that perspective.

I ran `pnpm audit` locally and even updated the `tsm` dependency package to avoid the warning, which is why I included the diff.

If you say it is unnecessary, it may be unnecessary. Should I not add this `overrides` section?

Note: The original writing style would no longer result in a version newer than 0.25.0, so this has been corrected in 65d3959. 

#### The result of `pnpm audit`

| key | value |
|:---:|---|
| moderate            | esbuild enables any website to send any requests to the development server and read the response |
| Package             | esbuild                                                |
| Vulnerable versions | <=0.24.2                                               |
| Patched versions    | >=0.25.0                                               |
| Paths               | . > tsm@2.3.0 > esbuild@0.15.18                        |
| More info           | https://github.com/advisories/GHSA-67mh-4wv8-2f99      |

1 vulnerabilities found
Severity: 1 moderate

delucis: `npm audit` isn't perfect because often a vulnerability only exists in a specific scenario, so if you're not using code in that way, the vulnerability doesn't impact you. 

If I understood correctly, that seems to be the case here? Just updating Astro would be enough to solve the vulnerable case (running a dev server). It is safe to continue depending on the older, "vulnerable" version of `esbuild` via `tsm`, because that does not run a dev server, so is not impacted by the vulnerability.

debiru: @delucis 

Thank you. I think that is true.

The question that remains is whether it is ok to keep getting warnings when running `pnpm audit`.

If this is deemed not to be a problem, then I will remove the `overrides` section.

... I would remove the `overrides` section, as it seemed to say that there is no problem.

delucis: Yes, I think that's fine, thank you! We don't usually run the audit CLI, and are happy to address packages on a case by case basis, rather than for everything that gets reported that way.

louisescher (approved):

PR #11157, opened 2025-03-05
i18n(fr): update `guides/content-collections` & `modules/astro-content`

ArmandPhilippot:
#### Description (required)

Adds changes from #11146 to the French translations of `guides/content-collections` and `reference/modules/astro-content`.

- Suggested label: i18n


thomasbnt (approved):
Lgtm! Thanks!

PR #11158, opened 2025-03-05
feat(assets): document remote utilities

ematipico:
#### Description (required)

This is a follow-up PR of the code PR https://github.com/withastro/astro/pull/13355

I added the new remote pattern utilities at the end of the page. I used the description that we have in the code as a starting point, and of course we can change them.

Questions:
- Do we want to provide an example?
- I used a code block to show the signature of the functions, instead of simple code words like we do for the hooks function. I felt that having more colour would help. Happy to change it and make this consistent.

As a tangent, Astro exposes **more functions** from the specifier `astro/assets/utils` that we don't document. However, I believe that **not all functions** are meant to be exposed to userland. I want to check with the team first, and then we can do follow-up PRs to add the missing functions.

### DIFF
--- a/src/content/docs/en/reference/image-service-reference.mdx
+++ b/src/content/docs/en/reference/image-service-reference.mdx
@@ -277,9 +277,10 @@ import {
 
 ### `isRemoteAllowed()`
 
-<p><Since v="4.0.0" /></p>
-
-**Type**: `(src: string, { domains, remotePatterns }: {domains: string[], remotePatterns: RemotePattern[] }): boolean`
+<p>
+**Type:** `(src: string, { domains, remotePatterns }: {domains: string[], remotePatterns: RemotePattern[] }): boolean`<br />
+<Since v="4.0.0" />
+</p>
 
 Determines whether a given remote resource, identified by its source URL, is allowed based on specified domains and remote patterns.
 
@@ -301,9 +302,10 @@ console.log(`Is the remote image allowed? ${isAllowed}`);
 
 ### `matchHostname()`
 
-<p><Since v="4.0.0" /></p>
-
-**Type**: `(url: URL, hostname?: string, allowWildcard = false): boolean`
+<p>
+**Type:** `(url: URL, hostname?: string, allowWildcard = false): boolean`<br />
+<Since v="4.0.0" />
+</p>
 
 Matches a given URL's hostname against a specified hostname, with optional support for wildcard patterns.
 
@@ -326,9 +328,10 @@ console.log(`Does the hostname match with wildcard? ${isMatchWithWildcard}`); //
 
 ### `matchPathname()`
 
-<p><Since v="4.0.0" /></p>
-
-**Type**: `(url: URL, pathname?: string, allowWildcard = false): boolean`
+<p>
+**Type:** `(url: URL, pathname?: string, allowWildcard = false): boolean`<br />
+<Since v="4.0.0" />
+</p>
 
 Matches a given URL's pathname against a specified pattern, with optional support for wildcards.
 
@@ -352,9 +355,10 @@ console.log(`Does the pathname match with wildcard? ${isMatchWithWildcard}`); //
 
 ### `matchPattern()`
 
-<p><Since v="4.0.0" /></p>
-
-**Type**: `(url: URL, remotePattern: RemotePattern): boolean`
+<p>
+**Type:** `(url: URL, remotePattern: RemotePattern): boolean`<br />
+<Since v="4.0.0" />
+</p>
 
 Evaluates whether a given URL matches the specified remote pattern based on protocol, hostname, port, and pathname.
 
@@ -379,9 +383,10 @@ console.log(`Does the URL match the remote pattern? ${isPatternMatched}`); // Ou
 
 ### `matchPort()`
 
-<p><Since v="4.0.0" /></p>
-
-**Type**: `(url: URL, port?: string): boolean`
+<p>
+**Type:** `(url: URL, port?: string): boolean`<br />
+<Since v="4.0.0" />
+</p>
 
 Checks if the given URL's port matches the specified port. If no port is provided, it returns `true`.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -259,3 +259,71 @@ export default defineConfig({
...
+## Utilities
+
+Astro exposes a number of helper functions that can be used to develop a custom image service. These utilities can be imported from `astro/assets/utils`:
+
+```ts
+import { matchProtocol } from "astro/assets/utils";
+
+matchProtocol(new URL("https://example.com"), "https"); // returns true
+
+```
```
ArmandPhilippot, line 272: I think we should move this example under the right section, but maybe we could replace it with the list of available imports like we do in [`astro:assets`](https://docs.astro.build/en/reference/modules/astro-assets/#imports-from-astroassets) for example:
````suggestion
```ts
import { 
    isRemoteAllowed,
    matchHostname,
    matchPathname,
    matchPattern,
    matchPort,
    matchProtocol
} from "astro/assets/utils";
```
````

**Review by ArmandPhilippot — COMMENTED** [review]
Looks good! I wasn't aware of these utilities!

However, I think it's best to stay consistent with the other API references, so I would say we should:
* use the [API reference notation](https://contribute.docs.astro.build/reference/api-references/) instead of code blocks for the type, which also allow us to add a `Since` in the same block if new utilities are added later
* move the `matchProtocol` example under the corresponding section
* and, yes, if we can think of some use cases, I think it would be nice to have an example for each utility to add more colours/help users (but not mandatory... we can always add one later if a user reports that they have trouble understanding how to use it)

The first type is a bit long but I think it's still suitable:
|Before|After|
|---|---|
|Type as code block|API notation and example|

(Obviously my example is neither pretty nor correct because I edited it quickly in the browser... 😅 )

Ideally, I think we should also update the hooks section with API notation. We've done this on other API reference pages, but this page hasn't been updated yet. However, we can do that later! I'm just saying this for context (ie. we shouldn't use the Hooks section as an example of the correct format).

File: src/content/docs/en/reference/image-service-reference.mdx [hunk 2]
```diff
@@ -259,3 +259,180 @@ export default defineConfig({
...
+    matchPort,
+    matchProtocol
+} from "astro/assets/utils";
+```
+
+### `isRemoteAllowed()`
+
+<p><Since v="4.0.0" /></p>
+
+**Type**: `(src: string, { domains, remotePatterns }: {domains: string[], remotePatterns: RemotePattern[] }): boolean`
```
ArmandPhilippot, line 282:
```suggestion
<p>
**Type:** `(src: string, { domains, remotePatterns }: {domains: string[], remotePatterns: RemotePattern[] }): boolean`<br />
<Since v="4.0.0" />
</p>
```

src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -259,3 +259,180 @@ export default defineConfig({
...
+const isAllowed = isRemoteAllowed(url.href, { domains, remotePatterns });
+
+console.log(`Is the remote image allowed? ${isAllowed}`);
+```
+
+### `matchHostname()`
+
+<p><Since v="4.0.0" /></p>
+
+**Type**: `(url: URL, hostname?: string, allowWildcard = false): boolean`
```
ArmandPhilippot, line 306:
```suggestion
<p>
**Type:** `(url: URL, hostname?: string, allowWildcard = false): boolean`<br />
<Since v="4.0.0" />
</p>
```

src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -259,3 +259,180 @@ export default defineConfig({
...
+// Match with wildcard
+const isMatchWithWildcard = matchHostname(testURL, hostnameToMatch, true);
+console.log(`Does the hostname match with wildcard? ${isMatchWithWildcard}`); // Output: true
+```
+
+### `matchPathname()`
+
+<p><Since v="4.0.0" /></p>
+
+**Type**: `(url: URL, pathname?: string, allowWildcard = false): boolean`
```
ArmandPhilippot, line 331:
```suggestion
<p>
**Type:** `(url: URL, pathname?: string, allowWildcard = false): boolean`<br />
<Since v="4.0.0" />
</p>
```

src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -259,3 +259,180 @@ export default defineConfig({
...
+const wildcardPathname = '/images/*';
+const isMatchWithWildcard = matchPathname(testURL, wildcardPathname, true);
+console.log(`Does the pathname match with wildcard? ${isMatchWithWildcard}`); // Output: true
+```
+
+### `matchPattern()`
+
+<p><Since v="4.0.0" /></p>
+
+**Type**: `(url: URL, remotePattern: RemotePattern): boolean`
```
ArmandPhilippot, line 357:
```suggestion
<p>
**Type:** `(url: URL, remotePattern: RemotePattern): boolean`<br />
<Since v="4.0.0" />
</p>
```

src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -259,3 +259,180 @@ export default defineConfig({
...
+const isPatternMatched = matchPattern(testURL, remotePattern);
+
+console.log(`Does the URL match the remote pattern? ${isPatternMatched}`); // Output: true
+```
+
+### `matchPort()`
+
+<p><Since v="4.0.0" /></p>
+
+**Type**: `(url: URL, port?: string): boolean`
```
ArmandPhilippot, line 384:
```suggestion
<p>
**Type:** `(url: URL, port?: string): boolean`<br />
<Since v="4.0.0" />
</p>
```

src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -259,3 +259,180 @@ export default defineConfig({
...
+// Check a URL without explicitly providing a port (defaults to true if port is undefined)
+const isPortMatch3 = matchPort(testURL1);
+console.log(`Does the port match (no port specified)? ${isPortMatch3}`); // Output: true
+```
+
+### `matchProtocol()`
+
+<p><Since v="4.0.0" /></p>
+
+**Type**: `(url: URL, protocol?: string): boolean`
```
ArmandPhilippot, line 414:
```suggestion
<p>
**Type:** `(url: URL, protocol?: string): boolean`<br />
<Since v="4.0.0" />
</p>
```

ArmandPhilippot (approved):
I only left suggestions to update the Type/Since parts because we want them in the same block.
Other than that, looks great! Thanks for the update!

PR #11161, opened 2025-03-05
update File() loader explanation closes #11027

### LINKED ISSUE #11027: File Loader API unclear
### 📚 Subject area/topic

Content Loader API

https://docs.astro.build/en/reference/content-loader-reference/#file-loader

Currently the `file()` loader API reference contains this line:

> The `file()` loader creates entries from a single file that contains an array of objects with a unique `id` field, or an object with string keys.

What this line leaves out is that if you do use an object with string keys, the values must also be objects 

This caused some confusion when a user attempted to do something like this:

```json
{
  "entry1": "foo",
  "entry2": "bar",
  "entry3": "baz"
}
```

Which resulted in a schema error expecting an object but receiving a string

A reproduction was created here by the discord member that brought it to my attention
https://github.com/anmpog/content-collection-issue-repro

Dillonpw:
#### Description (required)
Updates the first line of the file loader descriptive paragraph as discussed in issue #11027 by clarifying the keys and values.

- Closes #11027


src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -99,7 +99,7 @@ By default it uses [`github-slugger`](https://github.com/Flet/github-slugger) to
 <Since v="5.0.0" />
 </p>
 
-The `file()` loader creates entries from a single file that contains an array of objects with a unique `id` field, or an object with string keys. It supports JSON or YAML, and you can provide a custom `parser` for data files it cannot parse by default.
+The `file()` loader creates entries from a single file that contains an array of objects with a unique id field, or an object with `id`'s as keys and `entries` as values. It supports JSON or YAML, and you can provide a custom `parser` for data files it cannot parse by default.
```
sarah11918, line 102:
```suggestion
The `file()` loader creates entries from a single file that contains an array of objects with a unique `id` field, or an object with IDs as keys and entries as values. It supports JSON or YAML, and you can provide a custom `parser` for data files it cannot parse by default.
```
I think the unique `id` field should be written as code, but I don't think `entries` is a special property needed by the object, so I think the rest is just in plain language.

PR #11156, opened 2025-03-05
fix: update `routing.mdx`

jsparkdev:
#### Description (required)

- Update codeblocks
- Update whitespace and double quotes
- Add inline code 

I removed lines [316](https://github.com/withastro/docs/pull/11156/files#diff-d5e3bc536d8257c16397086fee639a17012a6127b4cc1e8321a1034136fecf9dL316) and [366](https://github.com/withastro/docs/pull/11156/files#diff-d5e3bc536d8257c16397086fee639a17012a6127b4cc1e8321a1034136fecf9dL366) as they were not needed, but let me know if you have any other ideas.

PR #11162, opened 2025-03-06
i18n(ko-KR): update `routing.mdx`

jsparkdev:
#### Description (required)

- update `routing.mdx`
- #11156
- update Korean translation guide


ArmandPhilippot (approved):
This was a lot of Korean changes 😅 but as a non native speaker, LGTM! Feel free to merge!

While looking at the preview I noticed that the [`<Since />` text](https://github.com/withastro/docs/blob/20b9e5d2628815431f5d6d7ca6247b46f00470a7/src/content/i18n/ko.yml#L12) (`Added in:`) is not translated, is this intentional or an oversight?

jsparkdev:
My mistake! I'll update and merge this, thanks.

PR #11193, opened 2025-03-10
i18n(zh-cn): imporve Manual Setup zh-cn tranlate

peng:
#### Description (required)

I think `Manual Setup` translate `手动安装` is better. `手动设置` I always understand that I need to manually set something. 

Discord username: peng_zhi

- Suggested label:  i18n


liruifengv (approved):
LGTM!

PR #11196, opened 2025-03-11
fix: update some codeblocks

jsparkdev:
#### Description (required)

- Added missing double quotes, semicolons.
- Removed unnecessary `async` keyword.
- Fixed incorrect code highlighting.
- Updated invalid whitespace.
- Formatted code to improve readability.

### DIFF
--- a/src/content/docs/en/guides/data-fetching.mdx
+++ b/src/content/docs/en/guides/data-fetching.mdx
@@ -16,7 +16,7 @@ This fetch call will be executed at build time, and the data will be available t
 
 💡 Pass fetched data to both Astro and framework components, as props.
 
-```astro /await fetch\\(.*?\\)/
+```astro /await fetch\\(.*?\\);/
 ---
 // src/components/User.astro
 import Contact from "../components/Contact.jsx";
@@ -49,7 +49,7 @@ The `fetch()` function is also globally available to any [framework components](
 import type { FunctionalComponent } from 'preact';
 
 const data = await fetch('https://example.com/movies.json').then((response) =>
-  response.json()
+  response.json();
 );
 
 // Components that are build-time rendered also log to the CLI.
@@ -70,7 +70,8 @@ Astro can also use `fetch()` to query a GraphQL server with any valid GraphQL qu
 
 ```astro title="src/components/Film.astro" "await fetch"
 ---
-const response = await fetch("https://swapi-graphql.netlify.app/.netlify/functions/index",
+const response = await fetch(
+  "https://swapi-graphql.netlify.app/.netlify/functions/index",
   {
     method: "POST",
     headers: { "Content-Type": "application/json" },
@@ -87,7 +88,9 @@ const response = await fetch("https://swapi-graphql.netlify.app/.netlify/functio
         id: "ZmlsbXM6MQ==",
       },
     }),
-  });
+  }
+);
+
 
 const json = await response.json();
 const { film } = json.data;
--- a/src/content/docs/en/guides/routing.mdx
+++ b/src/content/docs/en/guides/routing.mdx
@@ -433,18 +433,10 @@ You can use the `paginate()` function to generate these pages for an array of va
 // src/pages/astronauts/[page].astro
 export function getStaticPaths({ paginate }) {
   const astronautPages = [
-    {
-      astronaut: "Neil Armstrong",
-    },
-    {
-      astronaut: "Buzz Aldrin",
-    },
-    {
-      astronaut: "Sally Ride",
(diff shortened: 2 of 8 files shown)


src/content/docs/en/guides/routing.mdx
```diff
@@ -427,26 +428,31 @@ Paginated route names should use the same `[bracket]` syntax as a standard dynam
...
+    {
+      astronaut: "Buzz Aldrin",
+    },
+    {
+      astronaut: "Sally Ride",
+    },
+    {
+      astronaut: "John Glenn",
+    },
+  ];
```
ArmandPhilippot, line 448: Maybe each item could be on one line since there is only one property:
```suggestion
  const astronautPages = [
    { astronaut: "Neil Armstrong" },
    { astronaut: "Buzz Aldrin" },
    { astronaut: "Sally Ride" },
    { astronaut: "John Glenn" },
  ];
```
I mean that on desktop (1920x1080) the code snippet takes up almost the entire height of the screen, so if we reduce this space, the need to scroll on mobile will be reduced. But it's just a nit!

src/content/docs/en/guides/typescript.mdx
```diff
@@ -282,12 +282,16 @@ Astro includes helpers for working with the types returned by your [`getStaticPa
 
 You can get the type of [`Astro.params`](/en/reference/api-reference/#params) with `InferGetStaticParamsType` and the type of [`Astro.props`](/en/reference/api-reference/#props) with `InferGetStaticPropsType`:
 
-```astro title="src/pages/posts/[...id].astro" {2,14-15}
+```astro title="src/pages/posts/[...id].astro" {2-6,18-19}
```
ArmandPhilippot, line 285: I wonder if we shouldn't also highlight `GetStaticPaths` here:
````suggestion
```astro title="src/pages/posts/[...id].astro" {2-6,18-19} "satisfies GetStaticPaths"
````
This is a bit tricky since the whole function would need to be updated (because of the parentheses) but highlighting only `satisfies GetStaticPaths` might still be a good hint.

I mean most of the time, people don't even need `InferGetStaticParamsType` and `InferGetStaticPropsType` if they use `GetStaticPaths` so I think it shouldn't be the least visible type.

File: src/content/docs/en/guides/typescript.mdx [hunk 3]
```diff
@@ -282,12 +282,16 @@ Astro includes helpers for working with the types returned by your [`getStaticPa
 
 You can get the type of [`Astro.params`](/en/reference/api-reference/#params) with `InferGetStaticParamsType` and the type of [`Astro.props`](/en/reference/api-reference/#props) with `InferGetStaticPropsType`:
```
ArmandPhilippot, line 283: For the same reason as the highlighting, perhaps we could update this sentence:
```suggestion
You can get the type of [`Astro.params`](/en/reference/api-reference/#params) with `InferGetStaticParamsType` and the type of [`Astro.props`](/en/reference/api-reference/#props) with `InferGetStaticPropsType` or you can use `GetStaticPaths` to infer both at once:
```

ArmandPhilippot (commented):
LGTM, I just left a few nits not necessarily related to your changes but while updating...

src/content/docs/en/guides/routing.mdx
jsparkdev: Great idea, thank you.

src/content/docs/en/guides/typescript.mdx
jsparkdev: Thanks, highlighting the newly added types sounds like a great idea!

src/content/docs/en/guides/typescript.mdx
jsparkdev: Thank you. You've added a clearer explanation.

jsparkdev:
@ArmandPhilippot 

Thanks for adding a clearer explanation of types!

ArmandPhilippot (approved):
LGTM, thanks! 🙌🏽

src/content/docs/en/guides/data-fetching.mdx
```diff
@@ -49,7 +49,7 @@ The `fetch()` function is also globally available to any [framework components](
 import type { FunctionalComponent } from 'preact';
 
 const data = await fetch('https://example.com/movies.json').then((response) =>
-  response.json()
+  response.json();
```
jsparkdev, line 52: If you update by referencing the changes in this PR, please include the changes in [this PR](https://github.com/withastro/docs/pull/11205/files) as well! Thank you!

PR #11213, opened 2025-03-12
Fixes type syntax generated by our config reference script

delucis:
#### Description (required)

This PR fixes a subtle issue with the docgen script used to generate our configuration reference.

The script uses `jsdoc-api` to parse JSDoc comments. This converts certain types to use JSDoc’s old-school Closure-style type notation. For example, `string[]` becomes `Array.<string>`, which looks a little odd these days with most people expecting TypeScript’s type syntax.

This PR adds a hotfix to remove the unexpected `.` in generic types like this fixing several instances of `Array.<...>` and `Record.<...>` in our config reference.

(I did not include the changes to the page itself in this PR as there are several things in flight there that will need handling anyway in #11212.)

H/T @ArmandPhilippot for spotting this issue

### DIFF
--- a/scripts/docgen.mjs
+++ b/scripts/docgen.mjs
@@ -157,9 +157,13 @@ function getCommentProperties(comment) {
 	if (comment.kind !== 'heading' && !comment.type && !typerawFlag) {
 		throw new Error(`Missing @docs JSDoc tag: @type or @typeraw`);
 	}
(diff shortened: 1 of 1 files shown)


ArmandPhilippot (approved):
LGTM! 🙌🏽 
I checked the current Config reference page, we only have this pattern in `Type:` so it should be safe!

PR #11226, opened 2025-03-14
fix: add missing frontmatter prop in `heading-id-compat.mdx` [i18nIgnore]

ArmandPhilippot:
#### Description (required)

Unless I'm missing something, there's no reason for [`experimental-flags/heading-id-compat`](https://docs.astro.build/en/reference/experimental-flags/heading-id-compat/) not to be translated. Currently this file does not appear in the [Lunaria dashboard](https://i18n.docs.astro.build/) (e.g. under `Français (fr)`). I think it's because the `i18nReady` property is missing in the frontmatter so I added it.

- Suggested label: i18n?


yanthomasdev (approved):
Nice catch, thanks!

PR #11230, opened 2025-03-14
fix(guide/storyblok): adjust storyblok/astro import

michelm117:
#### Description (required)

I fixed an incorrect import statement in the Storyblok guide. The current code incorrectly imports Storyblok as a default export:

```js
import storyblok from "@storyblok/astro";
```

This causes an error as '@storyblok/astro' does not provide a default export. I've updated it to use the correct named import:

```js
import { storyblok } from "@storyblok/astro";
```

This change ensures the code example works correctly for users following the guide.

- Suggested label: `bug fix` `documentation`


yanthomasdev (approved):
Nice catch, thank you! And welcome to Team Docs! 🙌

PR #11231, opened 2025-03-15
fix: update `view-transitions.mdx`

jsparkdev:
#### Description (required)

- Fixed wrong links and added missing links
- Fixed invalid whitespace, typos
- Updated codeblocks for readability


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -125,20 +125,33 @@ Not all state can be preserved in this way. The restart of CSS animations and th
 You can also [manually identify corresponding elements](#naming-a-transition) if the island/element is in a different component between the two pages.
 
 ```astro title="src/pages/old-page.astro" "Video" 'transition:name="media-player"'
```
ArmandPhilippot, line 127: We also need to fix the highlighting:
````suggestion
```astro title="src/pages/old-page.astro" "video" 'transition:name="media-player"'
````

File: src/content/docs/en/guides/view-transitions.mdx [hunk 2]
```diff
@@ -429,41 +440,41 @@ The `<ClientRouter />` router will trigger in-page transitions from `<form>` ele
...
 
 ```astro title="src/components/Form.astro"
-<form action="/contact" method="POST" enctype="application/x-www-form-urlencoded">
-  <!-- -->
-</form>
+<form
+  action="/contact"
+  method="POST"
+  enctype="application/x-www-form-urlencoded"
+>
```
ArmandPhilippot, line 447: I'm not sure we should remove the closing tag here:
```suggestion
>
  <!-- -->
</form>
```
I think usually we show valid syntax so the tag should be closed even if this not the main purpose of the example.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -429,41 +440,41 @@ The `<ClientRouter />` router will trigger in-page transitions from `<form>` ele
...
 ```
 
 You can opt out of router transitions on any individual form using the `data-astro-reload` attribute:
 
 ```astro title="src/components/Form.astro"
 <form action="/contact" data-astro-reload>
```
ArmandPhilippot, line 453: Same here, I think we should keep the closing tag:
```suggestion
<form action="/contact" data-astro-reload>
  <!-- -->
</form>
```

ArmandPhilippot (commented):
I left a few nits, otherwise looks good to me!

Although I'm not sure what happened with the `<Steps />` component in the `Client-side navigation process` section. In the live version of the docs, each step is wrapped in a `<p>` tag, making them more spaced out (like on other pages, e.g. https://docs.astro.build/en/tutorial/1-setup/4/#create-a-repository-on-github). However, in the preview, there is no `<p>` tag, so the steps look a bit squashed. See:
* https://deploy-preview-11231--astro-docs-2.netlify.app/en/guides/view-transitions/#client-side-navigation-process
* https://docs.astro.build/en/guides/view-transitions/#client-side-navigation-process

jsparkdev:
@ArmandPhilippot 

Thanks for the nice comment.

As you suggested, each step in `<Steps />` would look better if we reinstated it and added spacing between them. :+1: 

I removed the closing tag for `<form>` because I didn't think it was necessary in a code block that describes the usage of an element, and I use the same method in other code blocks. (See image below)

a
bb
cc
dd


If I were to add the closing tag back in, I would also update the elements shown in the above image.

That's how I see it, but I'm sure there's a better option, what do you think?

ArmandPhilippot:
@jsparkdev Oh you're right, I didn't pay attention to that... So, yeah, I'm fine with that. 😄 I marked these two suggestions as resolved!

ArmandPhilippot (approved):
LGTM, thanks!

PR #11243, opened 2025-03-17
Update environment-variables.mdx - Add caution from Vite docs about i…

filipekgregorio:
#### Description (required)

Add a caution note about imports breaking type augmentation in env.d.ts files when working with ImportMetaEnv. This issue is documented in Vite's documentation but not explicitly in Astro's docs, even though Astro uses Vite under the hood.

- Suggested label: `documentation`
- Suggested label: `typescript`

Hello, i'm filipe.k.gregorio on Discord.

### DIFF
--- a/src/content/docs/en/guides/environment-variables.mdx
+++ b/src/content/docs/en/guides/environment-variables.mdx
@@ -52,7 +52,7 @@ interface ImportMeta {
 :::caution
 Imports will break type augmentation
 
-If the ImportMetaEnv augmentation does not work, make sure you do not have any import statements in env.d.ts. This issue is also documented in [Vite's TypeScript documentation](https://vitejs.dev/guide/features.html#client-types) for the same reason.
+If the ImportMetaEnv augmentation does not work, make sure you do not have any import statements in env.d.ts. This issue is also documented in [Vite's TypeScript documentation](https://vite.dev/guide/env-and-mode#intellisense-for-typescript) for the same reason.
 :::
 
 ## Default environment variables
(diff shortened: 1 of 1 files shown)


ArmandPhilippot:
Hi @filipekgregorio and thanks for taking the time to fix the docs! However, I'm not quite sure your suggestion is correct...

First, I believe the link used in your caution should be https://vite.dev/guide/env-and-mode.html#intellisense-for-typescript because I didn't find anything related to that in the other page...

Then, top-level imports are indeed problematic because the module is no longer ambient... But you can use inline imports in your `.d.ts` file. The following works for example:

```ts
declare namespace App {
  interface Locals {
    i18n: import("astro").AstroConfig["i18n"];
  };
}

interface ImportMetaEnv {
  readonly DB_PASSWORD: string;
};

interface ImportMeta {
  readonly env: ImportMetaEnv;
};
```

So I think this is related to this opened issue #10945 (and by extension this PR #10810 for the `middleware.mdx` page) but feel free to correct me if I'm wrong. If so, we should indeed add something about imports, but I'm not sure this caution is the right solution here.

src/content/docs/en/guides/environment-variables.mdx
```diff
@@ -49,6 +49,12 @@ interface ImportMeta {
 }
 ```
 
+:::caution
+Imports will break type augmentation
```
vrabe, line 53: Hi! Just a typo.

```suggestion
Imports will break type augmentation.
```

filipekgregorio:
Hi @ArmandPhilippot, thank you so much for taking the time to review my contribution and for catching the issue with imports! I’ve learned from your insights and now understand how inline imports work—they avoid breaking global type augmentation and keep the file ambient.

With this clarified, I’ll go ahead and close the pull request.

A big thanks to @vrabe as well for spotting the typo—it’s always great to have fresh eyes on things.

Cheers, Filipe Gregório

PR #11257, opened 2025-03-19
Fix heading level in directives reference

delucis:
#### Description (required)

This is a tiny PR that fixes a heading level in the directives reference. The `timeout` heading is describing a parameter of the preceding `client:idle` heading but is at the same level currently. This PR nests that one level lower.


ArmandPhilippot (approved):
Nice catch! 🙌🏽

PR #11280, opened 2025-03-20
Images: add info about on-demand rendering

mb21:
Images guide: add sentence about on-demand rendering

Because nowhere on the entire page (and entire docs as far as I could tell) does it say anything about how images behave in on-demand rendering. And the `<Image />` component has what I feel is a very non-intuitive default in on-demand rendering mode. See https://github.com/withastro/astro/issues/13468

Probably I would even also change:

> We recommend that local images are kept in `src/`

to:

> Unless you've turned on on-demand rendering, we recommend that local images are kept in `src/`

Because on our on-demand rendered page, using the `public` folder instead of transforming the image every time speeds up the request from ~500ms to ~80ms for a 160kB image.


src/content/docs/en/guides/images.mdx
```diff
@@ -91,6 +91,13 @@ import myImage from '../assets/my_image.png'; // Image is 1600x900
   loading="lazy"
   alt="A description of my image."
 />
+
+<!-- Output rendered on demand-->
+<!-- src will use an endpoint generated-->
```
sarah11918, line 96:
```suggestion
<!-- src will use an endpoint generated on demand-->
```

sarah11918:
Thank you for calling our attention to the differences in rendering the `<Image />` component on demand!

I checked with some other maintainers who agreed that instead of putting this information in the section about the public folder, we could do a better job describing the `<Image />` component itself. 

So, I made a commit with something more along the lines we'd be thinking of to be most helpful in docs. I'll need some reviews on this, but this is something to get us started!  🙌

ArmandPhilippot (approved):
Seems clear to me! 🙌🏽

PR #11296, opened 2025-03-24
Fix code in actions.mdx example

dario-c:
if check was missing a `!`

Before the example would return a 403 if the user HAD a user-session cookie, instead of when the user didn't have it

discord: dar.io_19940


delucis (approved):
Thank you for catching and fixing this @dario-c!

yanthomasdev (approved):
LGTM! Nice catch! 🙌

PR #11311, opened 2025-03-27
Update scoped styles documentation in guides

dominik-ilja:
#### Description (required)
Previous docs state that scoped styles don't increase specificity which is inaccurate. Scoped styles will increase class specificity by 1.

The imported style in the second example will have a specificity of 0-0-2. This will not override the scoped style of 0-1-1. The example is updated to use an ID selector which will override the scoped style with its specificity of 1-0-0.

### DIFF
--- a/src/content/docs/en/guides/styling.mdx
+++ b/src/content/docs/en/guides/styling.mdx
@@ -30,12 +30,12 @@ Astro `<style>` CSS rules are automatically **scoped by default**. Scoped styles
 This CSS:
 ```astro title="src/pages/index.astro"
 <style>
-  h1 { 
-    color: red; 
+  h1 {
+    color: red;
   }
 
-  .text { 
-    color: blue; 
+  .text {
+    color: blue;
   }
 </style>
 ```
@@ -214,7 +214,7 @@ import 'package-name/styles.css';
 <html><!-- Your page here --></html>
 ```
 
-If your package **does _not_ suggest using a file extension** (i.e. `package-name/styles`), you'll need to update your Astro config first! 
+If your package **does _not_ suggest using a file extension** (i.e. `package-name/styles`), you'll need to update your Astro config first!
 
 Say you are importing a CSS file from `package-name` called `normalize` (with the file extension omitted). To ensure we can prerender your page correctly, add `package-name` to [the `vite.ssr.noExternal` array](https://vite.dev/config/ssr-options.html#ssr-noexternal):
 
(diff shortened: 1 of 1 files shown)


sarah11918 (commented):
Thanks for bringing this to our attention, @dominik-ilja ! You're right, our wording was from before we updated the default `styleScopingStrategy` from `where` to `attribute`.

I've made some suggestions below, because it's not really correct to say "scoped styles do/don't increase specificity" because it depends what the user has configured. So instead, I think the wording below prompts people to the config value which has explanations of each mode, and when specificity does or doesn't increase with scoped styling.  See if you think something like those changes work!

dominik-ilja:
Hey! I went ahead and made those changes. I was originally unaware that you could configure the scoped styles. Thank you for bringing that up.

src/content/docs/en/guides/styling.mdx
```diff
@@ -302,9 +302,11 @@ Astro CSS rules are evaluated in this order of appearance:
 - **imported styles**
 - **scoped styles** (highest precedence)
 
-### Scoped Styles 
+### Scoped Styles
+
+Depending on your chosen value for [`scopedStyleStrategy`](/en/reference/configuration-reference/#scopedstylestrategy), scoped styles may or may not increase the [CLASS column specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Specificity#class_column).
 
-Using [scoped styles](#scoped-styles) does not increase the _specificity_ of your CSS, but they will always come last in the _order of appearance_. They will therefore take precedence over other styles of the same specificity. For example, if you import a stylesheet that conflicts with a scoped style, the scoped style's value will apply:
+That being said, [scoped styles](#scoped-styles) will always come last in the order of appearance. These styles will therefore take precedence over other styles of the same specificity. For example, if you import a stylesheet that conflicts with a scoped style, the scoped style’s value will apply:
```
sarah11918, line 309:
```suggestion
However, [scoped styles](#scoped-styles) will always come last in the order of appearance. These styles will therefore take precedence over other styles of the same specificity. For example, if you import a stylesheet that conflicts with a scoped style, the scoped style’s value will apply:
```
This is more our language style, and simpler, more common words are easier to read (and to translate!)

src/content/docs/en/guides/styling.mdx
```diff
@@ -325,10 +327,10 @@ import "./make-it-purple.css"
 </div>
 ```
 
-If you make the imported style _more specific_, it will have higher precedence over the scoped style:
+Scoped styles will be overwritten if the imported style is _more specific_. The style with a higher specificity will take precedence over the scoped style:
```
sarah11918, line 330:
```suggestion
Scoped styles will be overwritten if the imported style is more specific. The style with a higher specificity will take precedence over the scoped style:
```
Just also taking the opportunity to remove italics from older docs when we come across them, as it's no longer our style to use them!

src/content/docs/en/guides/styling.mdx
```diff
@@ -214,7 +214,7 @@ import 'package-name/styles.css';
 <html><!-- Your page here --></html>
 ```
 
-If your package **does _not_ suggest using a file extension** (i.e. `package-name/styles`), you'll need to update your Astro config first! 
+If your package **does _not_ suggest using a file extension** (i.e. `package-name/styles`), you'll need to update your Astro config first!
```
sarah11918, line 217:
```suggestion
If your package **does not suggest using a file extension** (i.e. `package-name/styles`), you'll need to update your Astro config first!
```

sarah11918 (approved):
Thank you again for this super helpful contribution, @dominik-ilja , and welcome to Team Docs! 🥳

PR #11318, opened 2025-03-28
fix: broken async await chain

stoneG:
#### Description (required)
image

Ran into this `RollupError` when completing the tutorial and setting up my RSS feed.
Fixed the example - Tested on my local build.

sitong on Discord 👋


florian-lefebvre (approved):

sarah11918 (approved):
Thank you for taking the time to correct this example, @stoneG !  🙌 

When this is merged, our translation system will mark the other languages as needing updates. You are welcome to contribute an updated translation, following our process for localization. (If you look at other docs PRs, you'll notice the convention for PR titles to indicate which language etc.) But, we only update one language at a time in PRs, so no need to wait on this PR as we do not include localizations in an update to English docs.

So that means we can immediately welcome you to Team Docs! 🥳

PR #11331, opened 2025-03-29
i18n(fr): update `guides/client-side-scripts.mdx`

ArmandPhilippot:
#### Description (required)

I noticed a heading in the French version that shouldn't be in the ToC ([`Import local scripts`/`Importer des scripts locaux`](https://docs.astro.build/fr/guides/client-side-scripts/#importer-des-scripts-locaux)) so while fixing that, I also:
* translated some missing parts (mainly in code snippets)
* fixed some typo
* reworded some parts/headings to make reading smoother and more accurate (`processing` !== `bundling` so we shouldn't use `regroupement` but `traitement` for the scripts heading)

- Suggested label: i18n


thomasbnt (approved):
LGTM! Thanks! 🚀

yanthomasdev (approved):
LGTM!

PR #11333, opened 2025-03-30
docs: update reference() documentation to match Astro v5 behavior

### LINKED ISSUE #11303: Collections reference() API docs do not match current behavior
### 📚 Subject area/topic

Content Collections API Reference

https://docs.astro.build/en/reference/modules/astro-content/#reference

> This accepts a collection name and validates the entry identifier(s) specified in your content frontmatter or data file before transforming the reference into an object containing the collection name and the reference id.

No validation of entries occurs in v5, as originally reported in https://github.com/withastro/astro/issues/13268. Moreover, without some alternative explanation, this may leave developers wondering what reason there is to use the `reference()` API at all in v5.

From discussion in the original reported issue, this is apparently a regression to be tackled in v6, though I do not know where that is tracked since the issue in question has been closed (as "not planned", which sends the opposite message) and locked.

_No response_

augustocesarperin:
#### Description (required)

This PR updates the documentation for the `reference()` function to clarify its behavior in Astro v5.

The documentation now correctly explains that `reference()` no longer validates at build time. I added a note about validation happening at runtime with `getEntry()`/`getEntries()` and explained why `reference()` is still useful despite this change.

- Closes #11303


src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -88,7 +88,11 @@ A `loader` is either an object or a function that allows you to load data from a
 <Since v="2.5.0" />
 </p>
 
-The `reference()` function is used in the content config to define a relationship, or "reference," from one collection to another. This accepts a collection name and validates the entry identifier(s) specified in your content frontmatter or data file before transforming the reference into an object containing the collection name and the reference id.
+The `reference()` function is used in the content config to define a relationship, or "reference," from one collection to another. This accepts a collection name and transforms the reference into an object containing the collection name and the reference id.
+
+> **Note:** In Astro v5, `reference()` no longer validates entry identifiers at build time. Validation happens at runtime when using `getEntry()` or `getEntries()`, which may return `undefined` if the referenced entry is invalid.
+>
+> Despite this change, `reference()` remains valuable for its cleaner syntax and improved TypeScript support compared to manually specifying collection references.
```
sarah11918, line 95: I've been thinking about a good way to construct this. A couple things about our docs style:

- we really try to avoid "notes" etc. for "just how things work". That should be interwoven in proper documentation, it's not "aside" or "tangential" to documenting the feature.
- We do not reference past versions of Astro, only what is true now. (Someone new to docs does not care about the history of the project).

I think a solution here might be, instead of this note here, add the following **after** the existing code sample showing configuration. (And, right **before** the final sentence to see the guide for more information:

``````
Validation of referenced entries happens at runtime when using `getEntry()` or `getEntries()`: 

```astro title="src/pages/[posts].astro
// if a referenced entry is invalid, this will return undefined.
const authors = await getEntries(blogPost.data.relatedPosts);
```
``````

This both accurately describes what happens and shows an example of it in use, which may then make it apparent why it's a handy helper for writing collection queries. (Compared to the longer example needed that Armand provided in the linked issue.)

Since this is part of the API reference, and not a part of our "guides" (which are more based on using/explaining/teaching), the info here should be for people looking things up quickly as a reference. We typically wouldn't "hype" a feature here, or talk about its benefits, but save that for the guide page.

I think this is minimal enough to satisfy that condition, and straightforward for people checking on this item. Do you think this would have been enough to help you if it had been in docs?

Comment (augustocesarperin) [reply src/content/docs/en/reference/modules/astro-content.mdx]: Thanks, Sarah! I've updated the PR with the changes you suggested. And yes, I think that would have been enough. Let me know if anything else is needed!

**Review by sarah11918 — APPROVED** [review]
This lookg great @augustocesarperin ! Thanks so much for working with us and helping to get this just right. And, welcome to Team Docs! 🥳

File: src/content/docs/en/reference/modules/astro-content.mdx [hunk 2]
```diff
@@ -113,6 +114,12 @@ const authors = defineCollection({
 
 export const collections = { blog, authors };
 ```
+Validation of referenced entries happens at runtime when using `getEntry()` or `getEntries()`:
+
+```astro title="src/pages/[posts].astro"
+// if a referenced entry is invalid, this will return undefined.
+const authors = await getEntries(blogPost.data.relatedPosts);
```
sarah11918, line 121:
```suggestion
const relatedPosts = await getEntries(blogPost.data.relatedPosts);
```

oops!

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -113,6 +114,12 @@ const authors = defineCollection({
 
 export const collections = { blog, authors };
 ```
+Validation of referenced entries happens at runtime when using `getEntry()` or `getEntries()`:
+
+```astro title="src/pages/[posts].astro"
+// if a referenced entry is invalid, this will return undefined.
+const relatedPosts = await getEntries(blogPost.data.relatedPosts);
```
sarah11918, line 121: @ArmandPhilippot Can you please verify this? I tried to modify the example you gave in the linked issue to match with the schema above, which only uses a single `author` but multiple `relatedPosts`.

ArmandPhilippot: This looks correct to me!

sarah11918: Woo! We'll get this merge train moving then! Thank you!

PR #11346, opened 2025-04-02
i18n(zh-cn): Update `prerender-client-address-not-available.mdx`

Nin3lee:
#### Description (required)

Update `prerender-client-address-not-available.mdx`

I translate the error message `${name}` => `<名称>` to keep the sentence fully translated.

- #11308
- Suggested label: i18n


yanthomasdev (approved):
LGTM! Thanks @Nin3lee, just a heads up `<NAME>` in MDX will make it think its a component and cause an error, so you have to escape it to not break the build

PR #11347, opened 2025-04-02
Add session.load

ascorbic:
#### Description (required)

Adds the session.load() method to the experimental flag docs

- Closes #
- Suggested label:


sarah11918 (approved):
Looks great! I do really like the example used in the blog post, and am wondering whether at some future date we could add a similar code example for *all* of these, but not blocking this PR!

PR #11361, opened 2025-04-04
fix(css): prevent social buttons from stretching in footer (tutorial)

mohammedyh:
#### Description (required)

Adding `align-items: center` to the social buttons CSS ruleset because without it the buttons are stretched to the height of the footer, making them appear too large.


sarah11918:
Thanks for the suggestion! As you've discovered, people can add their own CSS to customize the look of the site. For the basic tutorial, we are not too concerned about the final look of the product, because people will customize it if they want to take it further!

The purpose is just to illustrate Astro concepts, so, we are just keeping things as simple as we can. It is not necessary to change the tutorial for this, so we'll be closing this PR. But I hope the tutorial helped you learn about Astro! :rocket:

PR #11365, opened 2025-04-04
docs: replace await.then() in `routing-reference.mdx`

viniciusdeliz:
#### Description (required)

syntax mixing async/await and chained promise methods found in `routing-reference.mdx`

the change also conforms with the format of the `paginate()` sample which sits a few lines below

### DIFF
--- a/src/content/docs/en/reference/routing-reference.mdx
+++ b/src/content/docs/en/reference/routing-reference.mdx
@@ -153,7 +153,8 @@ For example, if you generate pages with data fetched from a remote API, you can 
 ```astro title="src/pages/posts/[id].astro" {8}
 ---
 export async function getStaticPaths() {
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -153,7 +153,8 @@ For example, if you generate pages with data fetched from a remote API, you can
 ```astro title="src/pages/posts/[id].astro" {8}
```
ArmandPhilippot, line 153: We also need to update the highlighting since we have one more line:
````suggestion
```astro title="src/pages/posts/[id].astro" {9}
````

ArmandPhilippot (approved):
Looks good to me, thanks for catching this! 🙌🏽 I just added a suggestion that I'll commit to update the highlighting as well.

PR #11373, opened 2025-04-05
fix: update indent and spaces in `guides/deploy/cloudflare.mdx`

jsparkdev:
#### Description (required)

- Add whitespace to improve readability
- Update invalid indentation in codeblocks

### DIFF
--- a/src/content/docs/en/guides/deploy/cloudflare.mdx
+++ b/src/content/docs/en/guides/deploy/cloudflare.mdx
@@ -20,7 +20,9 @@ This guide includes:
 <ReadMore>Read more about [using the Cloudflare runtime](/en/guides/integrations-guide/cloudflare/) in your Astro project.</ReadMore>
 ## Prerequisites
 
-To get started, you will need a Cloudflare account. If you don’t already have one, you can create a free Cloudflare account during the process.
+To get started, you will need:
+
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/deploy/cloudflare.mdx
```diff
@@ -20,27 +20,27 @@ This guide includes:
 <ReadMore>Read more about [using the Cloudflare runtime](/en/guides/integrations-guide/cloudflare/) in your Astro project.</ReadMore>
 ## Prerequisites
 
-To get started, you will need:
-
-- A Cloudflare account. If you don’t already have one, you can create a free Cloudflare account during the process.
-
+To get started, you will need a Cloudflare account. If you don’t already have one, you can create a free Cloudflare account during the process.
```
sarah11918, line 23: I would keep this the way it was, since that's the pattern we use in all the other guides!

sarah11918 (commented):
Thank you @jsparkdev ! The only change I would revert is the Prerequisites section, since that's the pattern we use in all other deploy guides. 

Otherwise, good to merge, thank you for spotting! 💜

jsparkdev:
@sarah11918 

Thanks, I've reverted that part back to the original.

sarah11918 (approved):

yanthomasdev (approved):
LGTM! Thanks!

PR #11374, opened 2025-04-05
fix: add inline codeblock to filenames in `adapter-reference`

jsparkdev:
#### Description (required)

- Add inline codeblock to improve readability


sarah11918 (approved):
Thank you!

PR #11384, opened 2025-04-07
fix(css): prevent `LearnAstroAd` from being cut off when TOC is long

### LINKED ISSUE #11378: LearnAstroAd component is cut off when TOC is long
### 📝 Issue Description

When the Table of Contents in the Right sidebar is so long that you have to scroll down to the bottom, the very bottom of the Coding In Public ad component is cut off at the bottom. (The bottom of the outline is not visible.

https://docs.astro.build/en/reference/configuration-reference/
https://docs.astro.build/en/guides/styling/
etc.

### 🤔 Expected Behavior

The expected behaviour is that the bottom spacing matches the rest of the pages (where you do not have to scroll down to see the end of the table of contents)

Image

### 👀 Current Behavior

On these pages with long TOCs, we get this instead:

Image

### 🖥️ Browser

Chrome 134.0.6998.130

### 📄 Additional Information

This component is found in the docs repo here:
https://github.com/withastro/docs/blob/main/src/components/RightSidebar/LearnAstroAd.astro

The sidebar component itself (that holds both the TOC and the Ad component) is here: 
https://github.com/withastro/docs/blob/main/src/components/starlight/PageSidebar.astro

(And, checking the sidebar component now, I guess this *might* have something to do with it? 😅 )

Image

mingjunlu:
#### Description

- This PR adjusts CSS to prevent `LearnAstroAd` component from being cut off when the table of contents is long

#### Related issues & labels

- Closes #11378
- Suggested label: `design`

#### Demonstration

| Before | After |
| ------ | ----- |
| long-toc-before | long-toc-after |

### DIFF
--- a/src/components/starlight/PageSidebar.astro
+++ b/src/components/starlight/PageSidebar.astro
@@ -31,7 +31,7 @@ const Banner = LearnAstroAd || StarlightBanner;
 	.right-sidebar-panel {
 		flex-direction: column;
 		justify-content: space-between;
-		height: 100%;
+		min-height: 100%;
 		padding: 1rem var(--sl-sidebar-pad-x);
 	}
 	.sl-container {


delucis (approved):
Great work @mingjunlu — thank you for fixing this 🥳

PR #11396, opened 2025-04-08
Update flotiq.mdx

palmiak:
It adds a community article about using Astro and Flotiq together. It focuses on using the Content Layer API.

#### Description (required)

Just adds a resource that might be helpful for others.

mpalmowski

PR #11400, opened 2025-04-09
i18n(fr): update `guides/cms/flotiq.mdx`

ArmandPhilippot:
#### Description (required)

Adds changes from #11396 to the French translation of `guides/cms/flotiq.mdx`.

> [!NOTE]
> I don't think we have a precedent for the use of `x` between two names... and I'm not sure how to translate it or if this common enough to leave it as it is. If I had to translate it, I'll probably go with `Utiliser Flotiq avec Astro`... So happy to hear ideas if `x` doesn't seem suitable!

- Suggested label: i18n


yanthomasdev (approved):
LGTM! Thanks!

Issue #10843, opened 2025-01-30
Tutorial seems out of sync with latest


flcdrg:
### 📝 Issue Description

The tutorial instructions seem outdated or are incomplete.

The page says to run `npm create astro@latest -- --template minimal`, and then press 'y' to create the template.

But the output of this command presents a menu, and I'm confused. 

![Image](https://github.com/user-attachments/assets/c440ea52-c671-4dc2-bb57-b26b4dc7ae0e)

On the one hand there is a default to select, but the tutorial is titled 'Build a blog' so should I be picking the blog template? I don't want to make the wrong choice as possibly this will mean I go down the wrong path with the tutorial!

Additionally, the tutorial says "When the prompt asks you where to create the project"

But I never get asked where to create the project - it seems to use 'minimal' as the directory name

### 📋 On which page(s) it occurs

https://docs.astro.build/en/tutorial/1-setup/2/

### 🤔 Expected Behavior

Instructions match what the tooling provides

### 👀 Current Behavior

Choices that aren't mentioned in the tutorial

### 🖥️ Browser

Firefox

delucis:
Hi @flcdrg! Could you share what OS and Terminal app you’re using? I just double checked the command and saw the expected behaviour:

<img width="478" alt="terminal output showing create astro being run with a template flag and that template being used correctly" src="https://github.com/user-attachments/assets/8adabd63-b892-47f4-a68b-02581cd18c8f" />

But I vaguely remember someone reporting something similar in the past for certain platforms related to the `--` separator npm requires. I can’t remember now but it may have been some platforms needed a `---` separator with _three_ hyphens instead of two?

flcdrg:
Windows 11, Windows Terminal (running PowerShell)

flcdrg:
FYI I figured out that this seems to be PowerShell related. If I escape the double dash, then it appears to install the minimal template and prompt for the directory. eg.

```powershell
npm create astro@latest `-- --template minimal
```

flcdrg:
oh.. and now it looks like the index.astro content is as per the tutorial!

So I assume "--template minimal" is generating different content to if you use "A basic, minimal starter" from the prompt.

delucis:
> So I assume "--template minimal" is generating different content to if you use "A basic, minimal starter" from the prompt.

Correct — the “basics” one in the prompt is a notch up from “minimal” 😁 

> this seems to be PowerShell related

[...]

PR #11423, opened 2025-04-11
Update 1.mdx

intellimat:
Changed `<body></body>` for `<Layout></Layout>`  in about.astro as the new astro setup wizard generate a Layout.astro

Place of the error:
https://docs.astro.build/en/tutorial/2-pages/1/

#### Description
When beginners follow the "Build a blog" tutorial and create about.astro, the content of that page shouldn't include `<body</body>` tag as the new Astro setup wizard generate a layout.astro which already has `<body><slot/></body>`.
The `about.astro` page should be rendered as `<slot/>`, hence the proposed change.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/tutorial/2-pages/1.mdx
+++ b/src/content/docs/en/tutorial/2-pages/1.mdx
@@ -42,10 +42,10 @@ Right now, your "About" page should look exactly the same as the first page, but
 
 Edit the HTML content to make this page about you.
 
-To change or add more content to your About page, add more HTML element tags containing content. You can copy and paste the HTML code below between the existing `<body></body>` tags, or create your own.
+To change or add more content to your About page, add more HTML element tags containing content. You can copy and paste the HTML code below between the existing `<Layout></Layout>` tags, or create your own.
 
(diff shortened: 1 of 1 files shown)


ArmandPhilippot:
Hi @intellimat, thanks for taking the time to fix this but... I don't think this is true?

By "the new astro setup wizard", do you mean when running the command [from unit 1](https://docs.astro.build/en/tutorial/1-setup/2/#launch-the-astro-setup-wizard)? (ie. `npm create astro@latest -- --template minimal`)
The minimal template shouldn't include a `Layout` component (and the tutorial is based on that template, not the one you can choose from the CLI). Are you on Windows with Powershell? If so, you might need to escape the command parameters. Can you try:
```sh
npm create astro@latest --- --template minimal
```

Seems related to #10843

sarah11918:
Thanks everyone! Yes, this is an issue with Powershell and not the tutorial content itself.

So, I'm going to close this PR in favour of continuing the discussion on the Issue page. There are a few different directions we can take, but I'll comment there, and y'all are welcome to head over there!

PR #11425, opened 2025-04-11
i18n(es): translate `astro-middleware.mdx`

JuanPabloDiaz:
translate to spanish

### DIFF
--- a/src/content/docs/es/reference/modules/astro-middleware.mdx
+++ b/src/content/docs/es/reference/modules/astro-middleware.mdx
@@ -1,0 +1,118 @@
+---
+title: Referencia de la API de Middleware
+sidebar:
+  label: "astro:middleware"
+i18nReady: true
+tableOfContents:
+  minHeadingLevel: 2
+  maxHeadingLevel: 6
+---
+
+import Since from '~/components/Since.astro';
+import ReadMore from '~/components/ReadMore.astro';
+
+<p><Since v="2.6.0" /></p>
+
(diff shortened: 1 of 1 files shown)


src/content/docs/es/reference/modules/astro-middleware.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+#### `next()`
+
+<p>
+**Tipo:** `(rewritePayload?: string | URL | Request) => Promise<Response>`<br />
+</p>
+
+El segundo argumento de `onRequest()` es una función que llama a todos los middlewares subsiguientes en la cadena y devuelve una `Response`. Por ejemplo, otro middleware podría modificar el cuerpo HTML de una respuesta y esperar el resultado de `next()` permitiría a tu middleware responder a esos cambios.
+
+Desde Astro v4.13.0, `next()` acepta un parámetro de ruta URL opcional en forma de cadena, `URL` o `Request` para [reescribir](/es/guides/routing/#reescrituras) la solicitud actual sin volver a desencadenar una nueva fase de renderizado.
```
JuanPabloDiaz, line 118: no se porque genera error. debería cambiar la ruta?

```suggestion
Desde Astro v4.13.0, `next()` acepta un parámetro de ruta URL opcional en forma de cadena, `URL` o `Request` para [reescribir](/es/guides/routing/#rewrites) la solicitud actual sin volver a desencadenar una nueva fase de renderizado.
```

yanthomasdev: The heading links should match the translation as well, yes. In this case, its because the Routing page in Spanish is outdated and doesn't have this heading, so for that we'd need another PR updating Routing and then we could have this one merged.

PR #11428, opened 2025-04-11
i18n(es): translate `serialized-configuration.mdx`

JuanPabloDiaz:
translate to spanish


yanthomasdev (commented):
Thanks @JuanPabloDiaz, this file does not exist anymore and has been changed to another place, could you make a new PR for the new page? The content is similar but with some changes:
https://docs.astro.build/en/reference/modules/astro-config/#imports-from-astroconfigclient

PR #11435, opened 2025-04-11
fix: update type of CollectionEntry ids

fedetibaldo:
#### Description (required)

The legacy content collections were generating an union of all possible ids. This is no longer the case--it's just a string. This change reflects that.

Yes. @fedetibaldo is my handle on Discord


sarah11918:
Thanks for your patience, @fedetibaldo , while we're a little behind with docs PRs while we've been getting things ready and out for Launch Week!

I'm going to ask @ArmandPhilippot to review this PR for you. I *suspect* that it is correct as stated: the "Example Type" is meant to give an example as to what the (yes, string) might look like. I believe the type given is shown as an example to be more helpful than just saying that the type will be `string`, but I will let Armand continue with this PR as he has been helping us standardize our reference types/examples!

ArmandPhilippot (approved):
This is indeed correct. I believe the `Example Type` is rather what we used with the legacy API and we forgot to update that. I think this change will be more useful because it doesn't imply that autocomplete can be expected here.

Thanks @fedetibaldo and sorry for the delay! Welcome to Team Docs. 🥳

PR #11456, opened 2025-04-15
Update _redirects

wpinfusion:
Redirect sessions guide

#### Description (required)

https://docs.astro.build/en/reference/experimental-flags/sessions/ is 404, and should be redirected to https://docs.astro.build/en/guides/sessions/


sarah11918:
This is great, thank you so much for reporting... and fixing!

We in fact have *two more* experimental features released today, so I'm going to add them here, too

PR #11460, opened 2025-04-15
fix: change access signature to `Astro.session`

ematipico:
#### Description (required)

This PR fixes the call of `Astro.session` and `ctx.session`. The `session` object *might* be `undefined`, so we need the `?` to gracefully handle its access.


yanthomasdev (approved):
LGTM! Thanks!

PR #11484, opened 2025-04-18
feat(fonts): add note about google options

florian-lefebvre:
#### Description (required)

Adds a note about google options

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -200,6 +200,8 @@ provider: fontProviders.fontsource()
 provider: fontProviders.google()
 ```
 
+Accepts [some options](https://github.com/unjs/unifont/blob/main/src/providers/google.ts#L10-L26).
+
 </TabItem>
 
 </Tabs>


src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -200,6 +200,8 @@ provider: fontProviders.fontsource()
 provider: fontProviders.google()
 ```
 
+Accepts [some options](https://github.com/unjs/unifont/blob/main/src/providers/google.ts#L10-L26).
```
sarah11918, line 203: So, one issue is I don't *love* that this doesn't *look like* it's part of the tab content:

image

Since there are only a few options it accepts anyway, is there a good way to put this content right in the code block itself, and maybe not even link out? That would make it more clear that this content belongs *with* the Google provider, and saves people a click.

florian-lefebvre: I think a challenge is that it accepts an optional config, so I don't want to make it look like the config is required.

> and maybe not even link out

That means that anytime the unifont provider implementation is updated, we should update our docs too which doesn't sound ideal.

---

I don't have strong opinions but yeah I'm not sure either what shape that should take

sarah11918: OK, then the sentence has to be way more explicit about what it's referring to, and where it's linking.

Will make a new suggestion

sarah11918: What about something closer to:

`````suggestion
Additionally, the `google()` font provider accepts all custom options available for the [unifont Google `ProviderOption`](https://github.com/unjs/unifont/blob/main/src/providers/google.ts#L10-L26):

```
provider: fontProviders.google(
	glyphs: //show an example of this being used.
)
```
``````
This way, there's still nicer separation between the end of the `<Tab>` content and the next line, and it shows a helpful example of configuring one such option?

File: src/content/docs/en/reference/experimental-flags/fonts.mdx [hunk 2]
```diff
@@ -200,6 +200,16 @@ provider: fontProviders.fontsource()
 provider: fontProviders.google()
 ```
 
+Additionally, the `google()` font provider accepts all options available for the [unifont Google `ProviderOption`](https://github.com/unjs/unifont/blob/main/src/providers/google.ts#L10-L26):
+
+```
```
florian-lefebvre, line 205:
``````suggestion
```js
``````

sarah11918 (approved):
I think that looks great! Merge when ready! :saluting_face:

PR #11491, opened 2025-04-19
Fix: Adds inline to Fonts Tailwind v4 instructions

### LINKED ISSUE #11490: Tailwind @theme + new Fonts needs `inline`
### 📚 Subject area/topic

Fonts

https://docs.astro.build/en/reference/experimental-flags/fonts/

Step 2

You can see in the following image how we currently recommend adding the fonts to the TW4 `@theme` directive

Image

Per the tailwind docs, when referencing other CSS variables, you should add `inline` to that directive like so `@theme inline {}`

https://tailwindcss.com/docs/theme#referencing-other-variables

_No response_

JusticeMatthew:
#### Description (required)

Changes `@theme {}` instructions for tailwind to `@theme inline {}` per the tailwind docs:
<https://tailwindcss.com/docs/theme#referencing-other-variables>

- Closes #11490

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -116,7 +116,7 @@ body {
 
     <TabItem label="Tailwind CSS 4.0">
 
-    ```css title="src/styles/global.css" ins={4}
+    ```css title="src/styles/global.css" ins={4} ins="inline"
     @import 'tailwindcss';
 
     @theme inline {
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -119,7 +119,7 @@ body {
     ```css title="src/styles/global.css" ins={4}
     @import 'tailwindcss';
 
-    @theme {
+    @theme inline {
         --font-sans: var(--font-roboto);
     }
     ```


src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -119,7 +119,7 @@ body {
     ```css title="src/styles/global.css" ins={4}
     @import 'tailwindcss';
 
-    @theme {
+    @theme inline {
         --font-sans: var(--font-roboto);
     }
     ```
```
TheOtterlord, line 125: If we do want to recommend `inline` for this reason or another, we should probably highlight it in this example.

TheOtterlord (commented):
Confirmed this works. Although, I'm not sure there's any other benefit than shipping one fewer css variables?

src/content/docs/en/reference/experimental-flags/fonts.mdx
JusticeMatthew: Perhaps a comment like

To prevent duplicate variable declarations, use inline (link to tw docs)

I'm not even sure if that's the right explanation though

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -119,7 +119,7 @@ body {
     ```css title="src/styles/global.css" ins={4}
```
sarah11918, line 119:
``````suggestion
    ```css title="src/styles/global.css" ins={4} ins="inline"
``````
Reuben might have meant highlight as in make the code sample highlight the word inline! If the Tailwind docs suggest this usage, then we don't need any comment about it I don't think. We just tell people to use it! But updating our markdown block to annotate this might be helpful to draw people's attention to it!

JusticeMatthew: Oh I see, fancy starlight magic, that makes sense 😄

sarah11918:
If @florian-lefebvre approves this change from a technical standpoint, then it's fine by me!

florian-lefebvre (approved):
I think it's great

PR #11501, opened 2025-04-22
Update routing.mdx to reflect the need for adapters and prerender binding in SSR mode

Zafkiel45:
I added a change to the “SSR Mode” topic, because an error occurred when I tried to create dynamic routes using SSR instead of SSG. The error in question complains about the lack of “getStaticPaths()” when I try to create the build with dynamic routes without a “getStaticPaths()”. The solution I found in programming communities was to add an extra line, disabling “pre-rendering” in conjunction with an integrated SSR adapter.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/cms/preprcms.mdx
+++ b/src/content/docs/en/guides/cms/preprcms.mdx
@@ -142,7 +142,7 @@ Your root directory should include these new files:
 
 #### Creating individual blog post pages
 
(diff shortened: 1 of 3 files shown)


sarah11918 (commented):
Thank you very much for this contribution @Zafkiel45 ! 

Adding `export const prerender = false;` is a helpful improvement to this code example! We do have a standard wording we use because this line isn't *always* required (it depends on your `output` setting: `static` or `server`). So I just added that to your line.

Also, we are trying to use the phrase `on-demand rendering` instead of SSR wherever we can, so I updated the section heading too!

Thank you for taking the time to make docs better, and welcome to Team Docs! 🥳

PR #11504, opened 2025-04-22
view-transitions: add warning about Firefox video breakage with animate

ericswpark:
#### Description (required)

The Astro view transition script has a limitation where video playback on the new page will fail on Firefox, since it is using Javascript to emulate view transitions on Firefox. This should be documented so that users are aware that video playback won't work with swapped pages on Firefox if the `animate` fallback is used.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -483,6 +483,8 @@ import { ClientRouter } from "astro:transitions";
 
 :::note[Known limitations]
 The `initial` browser animation is not simulated by Astro. So any element using this animation will not currently be animated.
+
+On Firefox using `animate` will cause video playback on the new page to break. See [this GitHub issue](https://github.com/withastro/astro/issues/13477) for more information.
 :::
 
 ## Client-side navigation process


sarah11918:
Thanks for this suggestion, @ericswpark !

I will ask @martrapp to review your note in case there is anything else he thinks should accompany this, perhaps some other resource he knows about!

martrapp:
Hi @ericswpark 👋🏼,

Yes, there are situations where videos will not load when using fallback="animate" (which is the default) on Firefox.

That said, your suggestion feels a bit too broad to me as it stands. The issue you are referring to comes from a really cool and pretty complex site. On the other hand, we also have examples where client router navigation to a page with a video works just fine in Firefox using the default fallback. Do you have a current issue that I can help with?

I would lean more toward framing the known limitations like this:

- View Transition API's group animations that morph one image into another are not simulated.

- Unlike native view transitions that blend exit with entry animations, the simulation shows the exit animation on the old page, and after that starts the entry animation on the new page.

- The simulation might interfere with the loading of video elements.

- The initial value for transition: animate is not simulated, so any element relying on it will not animate for now.

ericswpark:
Hi @martrapp ,

I actually have a broken page on my personal website that's reproducible! Please see this page:

https://ericswpark.com/pages/purdue/cs354/

If you click on the "Running XINU on UTM" link on Firefox, the embedded video will not load until a refresh.

I was trying to figure out what in Firefox was broken until I realized it was just the JS fallback.

martrapp:
> I do understand it's broad but it's much better than nothing IMHO since it'll save others time debugging why the sites behave differently on Firefox.

Definitively.

> https://ericswpark.com/pages/purdue/cs354/

Do you have a repo for me?

ericswpark:
@martrapp unfortunately that particular website has a private repo because of sensitive data in the commit history (I haven't had the chance to look through and clean it). But the fallback used is the default and the behavior is consistent with the description of the linked issue.

martrapp:
If the problem is in the commit history, could you send me a zip of the checked out files?

ericswpark:
@martrapp I don't understand what you mean by this. The problem is with the currently deployed site. Is it not reproducible?

For reference, here's what I'm doing (within one directory):

`content.md`:

```mdx
Some text here

<video src="/pages/purdue/cs354/xinu-on-utm-walkthrough.mp4" controls></video>

Some other text here
```

`index.astro`:

```tsx
---
import "@/styles/content-page.css";

import Layout from "@/layouts/Layout.astro";

const title = "CS 354 - Running XINU on UTM";

// Load Markdown content page
import { Content } from "./content.md";
---

<Layout title={title}>
    <div class="container">
        <h1 transition:name="purdue-cs354-utm">{title}</h1>

        <Content />
    </div>
</Layout>
```

And the video file is in the `public/` directory.

By navigating back and forth between this page and other pages on the same website with view transitions the breakage should be observed.

martrapp:
Hi @ericswpark I took the information you gave me and made a minimal reproduction.
Everything works just fine with `fallback="animate"` on Firefox. 
See https://stackblitz.com/~/github.com/martrapp/withastro-astro-uuvplqdk

I would have been happy to help you with your own issue as well, but it seems you preferred to handle it on your own, which is perfectly fine. In the end, it just turned out that the sentence 'On Firefox using animate will cause video playback on the new page to break.' is not correct, so we should not include it in the docs.

martrapp:
Hi @ericswpark, I added a comment to [astro/13477](https://github.com/withastro/astro/issues/13477)
Maybe, that could be of help for your project and help others with similar behavior.

Closing this PR now as the statement isn't correct in its general form and the topic is better handled in the issues.

PR #11506, opened 2025-04-23
fix: update cloudflare integration guide

jsparkdev:
#### Description (required)

update cloudflare integration guide

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -207,7 +207,7 @@ More information about the custom route patterns can be found in [Cloudflare's r
 **Default:** `undefined`
 </p>
 
-Configure additional routes to be generated on demand by the Cloudflare adapter in the `routes.extend.include` array.
+Configures additional routes to be generated on demand by the Cloudflare adapter in the `routes.extend.include` array.
 
 #### routes.extend.exclude
 
@@ -216,14 +216,14 @@ Configure additional routes to be generated on demand by the Cloudflare adapter 
 **Default:** `undefined`
 </p>
 
-Configure routes to be excluded from on-demand rendering in the `routes.extend.exclude` array. These routes will be prerendered and served statically instead, and will not invoke the SSR function. Additionally you can use this option to serve any static asset (e.g. images, fonts, css, js, html, txt, json, etc.) files directly without routing the request through the SSR function.
+Configures routes to be excluded from on-demand rendering in the `routes.extend.exclude` array. These routes will be prerendered and served statically instead, and will not invoke the SSR function. Additionally you can use this option to serve any static asset (e.g. images, fonts, css, js, html, txt, json, etc.) files directly without routing the request through the SSR function.
 
 ```js title="astro.config.mjs"
 export default defineConfig({
   adapter: cloudflare({
     routes: {
       extend: {
-        include: [{ pattern: '/static' }], // Route a prerended page to the SSR function for on-demand rendering
+        include: [{ pattern: '/static' }], // Routes a prerended page to the SSR function for on-demand rendering
         exclude: [{ pattern: '/pagefind/*' }], // Use Starlight's pagefind search, which is generated statically at build time
       }
     },
@@ -452,7 +452,7 @@ The Cloudflare `workerd` runtime supports imports of some [non-standard module t
 
 - `.wasm` or `.wasm?module`: exports a [`WebAssembly.Module`](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Module) that can then be instantiated
 - `.bin`: exports an [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) of the raw binary contents of the file
-- `.txt`: Exports a string of the file contents
+- `.txt`: exports a string of the file contents
 
 All module types export a single default value. Modules can be imported both from server-side rendered pages, or from prerendered pages for static site generation.
 
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -155,7 +155,7 @@ Defines the path to the Wrangler configuration file. If no value is set, it trac
 **Default:** `undefined`
 </p>
 
-Sets the [Cloudflare environment](https://developers.cloudflare.com/workers/wrangler/environments/) to use. You must select the environment defined in the Wrangler configuration file, otherwise an error occurs.
+Sets the [Cloudflare environment](https://developers.cloudflare.com/workers/wrangler/environments/) to use. You must select an environment defined in the Wrangler configuration file, otherwise an error occurs.
 
 #### `platformProxy.persist`
 <p>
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -135,35 +135,39 @@ Proxies provided by this are a best effort emulation of the real production. Alt
...
 #### platformProxy.enabled
 <p>
-**Type:** `{ enabled?: boolean }`<br />
-**Default:** `{ enabled: false }`
+**Type:** `boolean`<br />
+**Default:** `true`
 </p>
 
-The `enabled` property allows you to enable the Cloudflare runtime in `astro dev`.
+Determine whether to enable the Cloudflare runtime in development mode.
```
sarah11918, line 142:
```suggestion
Determines whether to enable the Cloudflare runtime in development mode.
```
We typically (although I just checked, not always! 😅 ) use this form of the verb.

My thinking is that a definition answers either:

"What is this?" when the point is to say what something *is*
-  An array of allowed hosts
- The base path to deploy to

OR

"What does this do?" often with an assumed ["This"]:
- [This] Specifies the output target for builds
- [This] Enables security measures for an Astro website

I'm checking in config reference and of course I see right next to each other:
- Specify a mapping of redirects
- Specifies the output target for builds

So, clearly we have some work to do here, too! I see both in use on this page. (Enables, Determines... but then later Configure routes, Configure identical routes...

Let's make THIS page all verbs with the (s) for now, so that this one page is internally consistent, and we'll add this to our pile of "let's go through all the docs and standardize..." pile 😄

sarah11918 (commented):
Amazing updates, @jsparkdev ! Thank you so much for this PR! I appreciate the consistency, and love the definitions being turned into... definitions!

See my comments below for some specifics, which are mostly "I guess we're going to have to make some decisions and standardize throughout" 😅 

You've identified some great improvements here, which are going to force us to do more work... but then have even greater improvements! 💪

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -135,35 +135,39 @@ Proxies provided by this are a best effort emulation of the real production. Alt
...
 
 #### platformProxy.configPath
 <p>
-**Type:** `{ configPath?: string }`<br />
-**Default:** `{ configPath: 'wrangler.toml' }`
+**Type:** `string`<br />
+**Default:** `wrangler.toml`, `wrangler.json`, `wrangler.jsonc`
```
ArmandPhilippot, line 147: I'm wondering if we shouldn't use `undefined` instead? I think it would match what we describe below: "if no value is set, ...". And, if it seems a bit odd to have multiple files as default since property expects a single path.
```suggestion
**Default:** `undefined`
```

ArmandPhilippot (commented):
A tiny nit regarding a default, otherwise great job! 🙌🏽

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
jsparkdev: I've been thinking about this a lot, and it sounds good. Thanks!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
sarah11918: I just "unresolved" the conversation for two reasons @jsparkdev 👍 

1. I think there are more "no-s" verbs on this page we could standardize while we're here! Can you e.g. get the "Configure routes" while you're already here?
2. I'd like to use this comment tomorrow on Talking & Doc'ing, and this makes it easier to see. 😄

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -91,13 +91,13 @@ The Cloudflare adapter accepts the following options:
...
 <p>
-**Type:** `true | false`<br />
+**Type:** `boolean`<br />
 **Default:** `true`
 </p>
 
 Enables [imports of `.wasm`, `.bin`, and `.txt` modules](#cloudflare-module-imports).
 
-This functionality is enabled by default. If you'd like to disable it, set `cloudflareModules: false`.
+This functionality is enabled by default. If you'd like to disable it, set `false`.
```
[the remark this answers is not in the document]
sarah11918: Ugh, did GitHub eat my comment? Unresolving to make this easier to see on Talking and Doc'ing tomorrow, too!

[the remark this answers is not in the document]
jsparkdev: Here it is!

https://github.com/withastro/docs/pull/11506#discussion_r2055880838

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -91,13 +91,13 @@ The Cloudflare adapter accepts the following options:
 ### `cloudflareModules`
```
ArmandPhilippot, line 91: I found another inconsistent thing on this page. 😅 This is the only option on the page wrapped with backticks. I checked other pages to see what we usually do but, it seems we are inconsistent:
* other integrations pages and the [Render context reference](https://deploy-preview-11506--astro-docs-2.netlify.app/en/reference/api-reference/) wrap all options in backticks (e.g. [Netlify](https://deploy-preview-11506--astro-docs-2.netlify.app/en/guides/integrations-guide/netlify/#includefiles))
* the [configuration reference](https://deploy-preview-11506--astro-docs-2.netlify.app/en/reference/configuration-reference/) doesn't... but I guess this is because this file is autogenerated. Some other pages doesn't use backticks too like [`astro:assets`](https://deploy-preview-11506--astro-docs-2.netlify.app/en/reference/modules/astro-assets/#src-required)

So not sure what we should decide here (backticks seems more frequent...), but it would be nice if all the options in this page use the same formatting. So either we remove the backticks here, or we add them on other options.

jsparkdev: Thanks! I think it would be better to use backticks for all options.

sarah11918: I agree. Backticks for all! I know we can do them in config-reference, it's probably just the annoying writing experience and lack of a preview that makes us not always think about it there.

sarah11918 (approved):
OK, I think this looks great! And we can use it to update more going forward! Thank you again!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -125,45 +125,49 @@ export default defineConfig({
...
+#### `platformProxy.environment`
 <p>
-**Type:** `{ environment?: string }`<br />
-**Default:** `{ environment: undefined }`
+**Type:** `string`<br />
+**Default:** `undefined`
 </p>
 
-The `environment` property allows you to define which [environment configuration](https://developers.cloudflare.com/workers/wrangler/environments/) in your Wrangler configuration to use.
+Sets the [Cloudflare environment](https://developers.cloudflare.com/workers/wrangler/environments/) to use. You must select the environment defined in the Wrangler configuration file, otherwise an error occurs.
```
yanthomasdev, line 158: small nit, from what I understood you can have several environments set in a Wrangler configuration, so saying "the environment defined" sounds a bit misleading, as if you can only have one environment to select from. 
```suggestion
Sets the [Cloudflare environment](https://developers.cloudflare.com/workers/wrangler/environments/) to use. You must select an environment defined in the Wrangler configuration file, otherwise an error occurs.
```

yanthomasdev (commented):
Good work @jsparkdev, I have just a small comment/nit for you!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
jsparkdev: Thank you!

PR #11508, opened 2025-04-23
fix(guides/imports): update Node.js built-ins section

### LINKED ISSUE #7785: Imports guide recommends a buggy use of `import.meta.url`
### 📚 Subject area/topic

Imports

https://docs.astro.build/en/guides/imports/#node-builtins

The “Node Builtins” section in the Imports guide is three years old and probably misleading.

It suggests building a relative path to another file like this:
```js
const url = new URL('../../package.json', import.meta.url);
```

However, `import.meta.url`’s value varies between dev and build, for example during dev, a page at `src/pages/dir/index.astro` has an `import.meta.url` of `src/pages/dir/index.astro`. However, during a build this might be something like `dist/chunks/prerender_CX4g21Zw.mjs` instead.

In simple cases this technique can appear to work. For example, in `src/pages/index.astro` the folder depth is usually the same between dev and build and if you’re loading something at the root level like `package.json`, you will end up in the right place.

But in pages or components in subdirectories, or when trying to reference siblings in the file structure, you’re very likely to hit inconsistencies between dev and build (not to mention SSR uses of `node:fs` where you might even be referencing a file outside your project which is missing completely when deployed).

https://stackblitz.com/edit/github-9wjfwu?file=src%2Fpages%2Findex.astro

This example shows the documented technique apparently working in dev mode.

If you quit the dev server (<kbd>Ctrl + C</kbd>) and preview the build instead by running `npm run build && npm run preview`, the values in the sub directory page will be off.

ArmandPhilippot:
#### Description (required)

Updates the `Node.js builtins` section in `guides/imports` to fix an incorrect example and to improve the wordings:
* replace the example with the one provided by Fryuni in the issue
* reword the intro to avoid blaming people that use Node.js built-ins
* remove incorrect statement 

I don't think I was in this (these? 😅 ) T&D or I don't recall... so maybe the idea was to not mention adapters at all? I'm referring to this comment: https://github.com/withastro/docs/issues/7785#issuecomment-2371273677

Note: I'm not a native speaker but I believe `built-in`/`builtin` are the same... I used `built-in` because it seems to be the syntax used in other pages (I checked with Algolia). Should we update the heading as well?

- Closes #7785 (I think after a year it's time to close this one 😄 )

### DIFF
--- a/src/content/docs/en/guides/imports.mdx
+++ b/src/content/docs/en/guides/imports.mdx
@@ -361,21 +361,24 @@ Astro supports loading WASM files directly into your application using the brows
 
 ## Node Builtins
 
-We encourage Astro users to avoid Node.js builtins (`fs`, `path`, etc.) whenever possible. Astro is compatible with multiple runtimes using [adapters](/en/guides/on-demand-rendering/). This includes [Deno](https://github.com/denoland/deno-astro-adapter) and [Cloudflare Workers](/en/guides/integrations-guide/cloudflare/) which do not support Node builtin modules such as `fs`.
+Astro supports Node.js built-ins, with some limitations, using Node’s newer `node:` prefix. There may be differences between development and production, and some features may be incompatible with on-demand rendering. Some [adapters](/en/guides/on-demand-rendering/) may also be incompatible with these built-ins modules or require configuration to support a subset (e.g., [Cloudflare Workers](/en/guides/integrations-guide/cloudflare/) or [Deno](https://github.com/denoland/deno-astro-adapter)).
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Looks great! We'll review this tomorrow on Talking and Doc'ing... and close another issue! 💪

sarah11918 (approved):
Well, we didn't get to this on Talking and Doc'ing, but LGTM! Will happily take others reviews, but probably this is at the very least NWTWWHB. So, I think you can feel comfortable merging this one! Thank you for picking it up! 💪

PR #11510, opened 2025-04-23
[issue template] improve distinction between issue and discussion topics

sarah11918:
#### Description (required)

This helps to guide people to making discussions for topics that are not errors/incorrect statements to be fixed in docs. It emphasizes that improvements (not just new content) should be made as discussion topics, so that critical errors are easier to spot and triage.

### DIFF
--- a/.github/ISSUE_TEMPLATE/CONTENT.yml
+++ b/.github/ISSUE_TEMPLATE/CONTENT.yml
@@ -1,7 +1,7 @@
 name: "Incorrect Content Report 📄"
 description: Report a correction to docs.
 labels: [
-  "improve documentation"
+  "improve or update documentation"
 ]
 body:
   - type: markdown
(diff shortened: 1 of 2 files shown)


louisescher (approved):
LGTM

trueberryless (approved):
This phrasing is much better, it has confused me in the past too and I just couldn't decide sometimes. This should make it easier for people.

I generally am against not allowing blank issues, but as Astro grows, I think it makes sense to direct people into creating what I call "full" issues instead of "half empty" ones...

.github/ISSUE_TEMPLATE/CONTENT.yml
```diff
@@ -1,5 +1,5 @@
 name: "Incorrect Content Report 📄"
-description: Report an issue with existing content.
+description: Report a correction to docs.
 labels: [
   "improve documentation"
```
sarah11918, line 4:
```suggestion
  "improve or update documentation"
```

PR #11511, opened 2025-04-23
feat(fonts): warn if local font is in public dir

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -598,6 +598,10 @@ src: ["my-package/SomeFont.ttf"]
 
 </Tabs>
 
+:::caution
+We recommend not to store fonts in [`/public`](/en/reference/configuration-reference/#publicdir) as this will result in duplicated files in your build output. Instead, store them in [`/src`](/en/reference/configuration-reference/#srcdir) for example.
+:::
+
 You can also specify a [tech](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src#tech) by providing objects:
 
 ```js


ArmandPhilippot:
The content seems right to me. But perhaps we don't need a callout here and we should say that in a paragraph above the `Tabs`:

> Defines the [font sources](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src). The value can be a path relative to the root, a package import or a URL. URLs are particularly useful if you inject local fonts through an integration.
>
> Fonts should be stored in another location than [`/public`](/en/reference/configuration-reference/#publicdir) as this will result in duplicated files in your build output. You can store them inside [`/src`](/en/reference/configuration-reference/#srcdir) for example.

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -598,6 +598,10 @@ src: ["my-package/SomeFont.ttf"]
 
 </Tabs>
 
+:::caution
+We recommend not to store fonts in [`/public`](/en/reference/configuration-reference/#publicdir) as this will result in duplicated files in your build output. Instead, store them in [`/src`](/en/reference/configuration-reference/#srcdir) for example.
```
sarah11918, line 602:
```suggestion
We recommend not putting your font files in [the `public/` directory](/en/reference/configuration-reference/#publicdir). Since Astro will copy these files into that folder at build time, this will result in duplicated files in your build output. Instead, store them somewhere else in your project, such as in [`src/`](/en/reference/configuration-reference/#srcdir).
```
I made up a reason, because I think having extra context is helpful here. Can you make this accurate?

sarah11918 (commented):
I'm not totally sure this rises to a caution, but I don't hate it so we'll let it stand! 😄 

I think the comment is in the right spot, and just left an idea for maybe adding a bit of an explanation!

florian-lefebvre:
It *may not* require a callout there but I think it's helpful, especially for people migrating from manual usage to this API

PR #11513, opened 2025-04-24
Update experimental responsive image docs

ascorbic:
#### Description (required)

Updates the experimental flag docs to reflect the changes in https://github.com/withastro/astro/pull/13677. It also expands the docs to explain the feature in more detail, including explanations of _why_ to use it, and a note about potential impact on build times. It adds a note on use with Tailwind 4.

- Closes #
- Suggested label:


src/content/docs/en/reference/experimental-flags/responsive-images.mdx
```diff
@@ -14,7 +14,13 @@ import Since from '~/components/Since.astro'
 <Since v="5.0.0" />
 </p>
 
-Enables support for automatic responsive images in your project.
+Enables support for automatic responsive images in your project. 
+
+The term [responsive images](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Responsive_images) refers images that work well on different devices. This particularly applies to images that resize to fit their container, and that can be served in different sizes depending on the device's screen size and resolution. There are a number of additional properties that can be set to control how the image is displayed, but these can be complicated to handle manually. Incorrect handling of these properties can lead to images that are slow to download or that are not displayed correctly, and are one of the most common causes of poor Core Web Vitals and Lighthouse performance scores.
```
sarah11918, line 19:
```suggestion
The term [responsive images](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Responsive_images) refers images that work well on different devices. This particularly applies to images that resize to fit their container, and that can be served in different sizes depending on the device's screen size and resolution.

There are a number of additional properties that can be set to control how the image is displayed, but these can be complicated to handle manually. Incorrect handling of these properties can lead to images that are slow to download or that are not displayed correctly. This is one of the most common causes of poor Core Web Vitals and Lighthouse performance scores.
```

src/content/docs/en/reference/experimental-flags/responsive-images.mdx
```diff
@@ -33,22 +41,37 @@ Additionally, Astro's image components can receive [responsive image props](#res
 
 Images in your `public/` folder are never optimized, and responsive images are not supported.
 
-## Responsive image configuration settings
+:::note
+Enabling responsive images will generate additional image sizes for all affected images. For prerendered pages this happens during the build so may increase the build time of your project, especially if you have a large number of images. For pages rendered on-demand the images are generated as-needed, so this has no impact on build times but may increase the number of transformations performed. Depending on your image service this may incur additional costs.
```
sarah11918, line 45:
```suggestion
Enabling responsive images will generate additional image sizes for all affected images. For prerendered pages this happens during the build so may increase the build time of your project, especially if you have a large number of images.

For pages rendered on-demand the images are generated as-needed, so this has no impact on build times but may increase the number of transformations performed. Depending on your image service this may incur additional costs. 
```

src/content/docs/en/reference/experimental-flags/responsive-images.mdx
```diff
@@ -33,22 +41,37 @@ Additionally, Astro's image components can receive [responsive image props](#res
...
-## Responsive image configuration settings
+:::note
+Enabling responsive images will generate additional image sizes for all affected images. For prerendered pages this happens during the build so may increase the build time of your project, especially if you have a large number of images. For pages rendered on-demand the images are generated as-needed, so this has no impact on build times but may increase the number of transformations performed. Depending on your image service this may incur additional costs. 
+:::
+
+## Image layout
+
+In order to generate the correct `srcset` and `sizes` attributes, the `<Image />` and `<Picture />` components need to know how the image should resize when its container changes size. This is done by setting the `layout` prop, or `image.experimentalLayout` default. The supported values are:
+
+- `constrained` - The image will scale to fit the container, maintaining its aspect ratio, but will not exceed the specified `width` and `height`, or the image's original dimensions.
```
sarah11918, line 52:
```suggestion
- `constrained` - The image will scale to fit the container, maintaining its aspect ratio, but will not exceed the specified `width` and `height`, or the image's original dimensions. Use this for images that should be responsive, but only when smaller than the original (or a specified) size.
```
I like how the others had a "Use this/this is useful" so something like this maybe?

src/content/docs/en/reference/experimental-flags/responsive-images.mdx
```diff
@@ -33,22 +41,37 @@ Additionally, Astro's image components can receive [responsive image props](#res
...
+- `full-width` - The image will scale to fit the width of the container, maintaining its aspect ratio. Use this for hero images or other images that should take up the full width of the page.
+- `fixed` - The image will maintain the requested dimensions and not resize. The `srcset` will be generated for the requested image size, plus 2x resolution for high-density displays if the source image is large enough.
```
sarah11918, line 54:
```suggestion
- `fixed` - The image will maintain the requested dimensions and not resize. The `srcset` will be generated for the requested image size, plus 2x resolution for high-density displays if the source image is large enough. Use this when you prefer to see your image cropped at smaller sizes instead of showing a smaller version of the whole picture.
```
OK, I don't love this, but I do think an explanation of the container making only part of the image visible, at its original size, is helpful here for the Use when.../Useful for...

sarah11918 (commented):
Looks great! Just a few comments, mostly just breaking text up into more paragraphs.

I made two suggestions for the new image layout options list, because I think it's helpful that all of them have a "use for/useful for" ... these might not be the best choices, but I think if we can put something concrete in there, that will help people choose!

src/content/docs/en/reference/experimental-flags/responsive-images.mdx
ascorbic: I clearly need to explain this better, because this isn't what it does! It just displays them with no resizing at all, like a normal img tag with no styling.

src/content/docs/en/reference/experimental-flags/responsive-images.mdx
ascorbic: I've done new wording, including saying that if they're unsure they should choose this, and that it matches the default Tailwind styles

src/content/docs/en/reference/experimental-flags/responsive-images.mdx
ascorbic: I've rewritten this to try and make it clearer. lmk if it makes sense!

sarah11918 (approved):
Thanks! Appreciate the extra detail, and I think it's helpful to have the extra guidance for choosing! LGTM (I think the patch was already released, so this merges any time now?)

sarah11918:
Also merge #11532 when this is merged!

PR #11543, opened 2025-04-28
i18n(fr): update `experimental-flags/responsive-images.mdx`

ArmandPhilippot:
#### Description (required)

Adds changes from #11513 to the French translation of `experimental-flags/responsive-images.mdx`.

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/reference/experimental-flags/responsive-images.mdx
+++ b/src/content/docs/fr/reference/experimental-flags/responsive-images.mdx
@@ -16,6 +16,14 @@ import Since from '~/components/Since.astro'
 
 Active la prise en charge des images réactives automatiques dans votre projet.
 
+Le terme [images réactives](https://developer.mozilla.org/fr/docs/Web/HTML/Guides/Responsive_images) désigne les images qui fonctionnent bien sur différents appareils. Cela s'applique particulièrement aux images qui sont redimensionnées pour s'adapter à leur conteneur et qui peuvent être présentées dans différentes tailles en fonction de la taille et de la résolution de l'écran de l'appareil.
+
+Il existe un certain nombre de propriétés supplémentaires qui peuvent être définies pour contrôler la façon dont l'image est affichée, mais celles-ci peuvent être compliquées à gérer manuellement. Une mauvaise gestion de ces propriétés peut conduire à des images lentes à télécharger ou qui ne s'affichent pas correctement. Il s’agit de l’une des causes les plus courantes des mauvais scores de performances Core Web Vitals et Lighthouse.
+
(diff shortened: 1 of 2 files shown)


src/content/docs/fr/reference/experimental-flags/responsive-images.mdx
```diff
@@ -16,6 +16,14 @@ import Since from '~/components/Since.astro'
 
 Active la prise en charge des images réactives automatiques dans votre projet.
 
+Le terme [images réactives](https://developer.mozilla.org/fr/docs/Web/HTML/Guides/Responsive_images) désigne les images qui fonctionnent bien sur différents appareils. Cela s'applique particulièrement aux images qui sont redimensionnées pour s'adapter à leur conteneur et qui peuvent être présentées dans différentes tailles en fonction de la taille et de la résolution de l'écran de l'appareil.
```
ArmandPhilippot, line 19: This link refers to a page where "responsive images" is translated `images adaptatives` so it feels a bit weird to use `images réactives` as anchor... In an other hand, this is the translation we use in the docs.

I checked what is done elsewhere and I didn't find any "most used" translation. In [wiktionary](https://fr.wiktionary.org/wiki/responsive) they say:
> Le terme « réactif » s'utilise lorsque le code Web de la page est unique et est interprété de manière dynamique par le logiciel d'affichage[[1]](https://fr.wiktionary.org/wiki/responsive#cite_note-1). Le terme « adaptatif » s'utilise lorsque le logiciel sélectionne un code Web parmi plusieurs selon les caractéristiques du matériel[[2]](https://fr.wiktionary.org/wiki/responsive#cite_note-2).

Looking at the footnotes, it seems `adaptative` would be the recommendation in France (`recommandé officiellement par la Commission d'enrichissement de la langue française (France) `) while in Quebec they differentiate the two uses.

thomasbnt: We can leave it like _images adaptatives_, seems better than _images réactives_

thomasbnt, line 19:
```suggestion
Le terme [images adaptatives](https://developer.mozilla.org/fr/docs/Web/HTML/Guides/Responsive_images) désigne les images qui fonctionnent bien sur différents appareils. Cela s'applique particulièrement aux images qui sont redimensionnées pour s'adapter à leur conteneur et qui peuvent être présentées dans différentes tailles en fonction de la taille et de la résolution de l'écran de l'appareil.
```

ArmandPhilippot: Thanks, I also thanks it would make more sense. I'll accept your suggestion and update the rest of the page to use `image adaptative` instead, to be consistent. If this doesn't break links elsewhere I'll also make the change in `guides/images` otherwise I'll do that in another PR.

thomasbnt (approved):

PR #11549, opened 2025-04-28
fix(recipes/i18n): update code snippet to use collection `id`

ArmandPhilippot:
#### Description (required)

A code snippet in the `i18n` recipe was still using `slug` (legacy content collections) instead of `id` (content layer), so I replaced `page.slug` with `page.id`.


yanthomasdev (approved):
LGTM!

PR #11565, opened 2025-04-29
include Bun in documentation

jramma:

- Closes #
- Suggested label:


delucis:
Thanks for the PR @jramma! Astro doesn’t officially support Bun at the moment, this is why we don’t include it in our docs. I know some people do use Bun and Astro together successfully, but as we don’t have time to assure every aspect of Astro’s features works well on Bun, we don’t want to encourage its use in our official docs.

jramma:
Thanks for the clarification!  I’ll keep that in mind moving forward. Appreciate the quick feedback and all your work on Astro!

PR #11579, opened 2025-04-30
i18n(ja): Update `i18/ja.yml`

jp-knj:
#### Description (required)

Matches the latest English source (see: https://github.com/withastro/docs/blob/main/src/content/i18n/en.yml)  

- Closes #
- Suggested label: `i18n`


yanthomasdev (approved):
LGTM! Good job!

PR #11583, opened 2025-04-30
i18n(fr): update responsive image translation

ArmandPhilippot:
#### Description (required)

As discussed with @thomasbnt  in https://github.com/withastro/docs/pull/11543#discussion_r2063497285 other docs use `images adaptatives` instead of `images réactives` for `responsive images` so we should be consistent to avoid confusion (e.g. an external link to a page with another translation...).

- Suggested label: i18n


thomasbnt (approved):
LGTM

PR #11593, opened 2025-05-02
i18n(fr): update the French `i18n-guide`

ArmandPhilippot:
#### Description (required)

Updates the French `i18n-guide` to:
* turn it into a more comprehensive guide that includes Discord links and writing tips
* update the words list (remove obvious translations, update translations based on the current usage in the docs, adds new words)

The goals would be to:
* help newcomers get started
* make the translation across the docs more consistent
* provide a base to make the translation consistent in the Astro ecosystem (e.g. Starlight)
* give some references to previous discussions to keep a track both for newcomers and current translators in case of doubt

Next step: Scan the docs to update the translation when necessary based on the updated guide. I don't mind doing that, but first we need to agree on the translations!

- Suggested label: i18n


i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+#### Langage courant
+
+Certains mots ont un équivalent français qui devrait être utilisé uniformément dans les différentes traductions.
+
+| Anglais | Français | Explications/Référence (facultatif) |
+|---------|----------|-------------------------------------|
+| build | Suivant le contexte : création / construction / compilation | |
+| bundle | regroupement | |
+| changelog | journal des modifications | |
+| (the) CLI / (the) command line interface | (la) CLI / (l')interface en ligne de commande | |
```
ArmandPhilippot, line 64: Two things here:
* The previous guide was using `ILC` instead of `CLI`. I don't think I ever saw `ILC` in other docs but I don't usually read docs in French so I don't have a strong opinion... Currently, we don't use `ILC`.
* We use both `la CLI` and `le CLI` in the docs, it would be great to use a single form. I used `la CLI` here because if this is the abbreviation of `interface en ligne de commande`, `interface` is a feminine noun so it sounds consistent.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| header | en-tête | |
+| heading | titre ou en-tête | Nous n'avons pas de consensus (« titre » dans Astro Docs et « en-tête » dans Starlight). Précédente discussion : https://github.com/withastro/starlight/pull/2884#discussion_r1957379942 |
+| headless | sans-tête | |
+| hook | crochet | |
+| import | importation | |
+| issue | problème ou, en référence à la fonctionnalité de Github, ticket | [MDN](https://developer.mozilla.org/fr/docs/MDN/Community/Issues) utilise également « ticket » pour décrire la fonctionnalité de Github |
+| layout | mise en page | |
+| on-demand rendering | rendu à la demande | |
+| package | paquet | |
+| placeholder | paramètre fictif | |
```
ArmandPhilippot, line 80: Not sure what we should use here... Should we add more suitable translations? I saw some `emplacement réservé` in the current docs for example. I always find this odd but maybe it's only me!
I'd say it depends on the context. And sometimes no proposal will be suitable. So, it's a bit tricky. Happy to ear other opinions!

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| preset | préréglage | « préconfiguration » pourrait convenir mais dans certains cas la formulation devient lourde. Discuté dans https://github.com/withastro/starlight/pull/3126#discussion_r2046673972 |
+| prop | propriété | |
+| props | props ou propriétés | En référence à un composant privilégiez « props », autrement utilisez « propriétés » (voir [Concepts d'Astro](#concepts-dastro)) |
+| output | sortie ou affichage | |
+| re-render | rafraîchir / effectuer un nouveau rendu | |
+| recipes | recettes | |
```
ArmandPhilippot, line 86: The previous guide was using `méthodes`, but I think currently we always use `recettes`.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| server-side | coté serveur | |
+| template | modèle | |
+| to build | Suivant le contexte : créer / construire / compiler | |
+| to bundle | regrouper | |
+| to export | exporter | |
+| to fetch | récupérer | |
+| to import | importer | |
+| to render | rendre / afficher / restituer | |
+| to slot | inclure / injecter / insérer | |
+| to style | mettre en forme / appliquer des styles | |
```
ArmandPhilippot, line 103: I saw some `styliser` in the docs, but I think it's at best an anglicism or at worst a translation error. According to the [Larousse](https://www.larousse.fr/dictionnaires/francais/styliser/74966) or the [Wiktionary](https://fr.wiktionary.org/wiki/styliser) for example:
> Interpréter une forme naturelle en la simplifiant ou en la modifiant à des fins esthétiques.

So it doesn't sound right to me.

In [Starlight](https://starlight.astro.build/fr/guides/css-and-tailwind/), `CSS & Styling` is translated `CSS et mise en forme` which I think is the correct translation.

Happy to ear other thoughts!

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+
+#### Internes
+
+Tous les liens doivent rediriger vers la version française, pensez donc à remplacer `/en/` par `/fr/` même si la page de destination n'est pas encore traduite. Vous devez également pensez à traduire la partie après `#`, quand elle existe : elle correspond a l'id généré pour un titre sur la page. Vous pouvez utiliser le serveur de développement pour vérifier l'id généré.
+
+#### Externes
+
+* L'ancre du lien est toujours traduit.
+* Le lien ne change pas sauf s'il existe une version française. Il est donc préférable de vérifier le lien au moment de la traduction.
+* S'il n'existe pas de version française, il peut être utile de rajouter le code du langage (p. ex. `(en)`, à adapter suivant la langue) après le lien pour informer le lecteur que la page de destination est dans une autre langue.
```
ArmandPhilippot, line 147: There is also this... Currently, we use `(en)` when we indicate the language of a link but I wonder is this accessible? Should we use `(Anglais)` or `(en anglais)` instead?
The English docs use the full language name and Sarah also does that in monthly blog posts. Should we do the same? For example: https://docs.astro.build/en/guides/migrate-to-astro/from-docusaurus/#community-resources

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| back-end | Fait référence à la partie serveur d'une application. | Bien que « dorsal » soit une traduction possible, elle est rarement utilisée ailleurs. |
+| middleware | Désigne un logiciel qui permet de coordonner le fonctionnement de plusieurs logiciels. | Bien que « logiciel médiateur » et « intergiciel » soient les traductions recommandées, elles sont rarement utilisées ailleurs. |
+
+###  🔤 Traductions courantes
+
+#### Concepts et API d'Astro
+
+| Anglais | Français | Explications/Référence (facultatif) |
+|---------|----------|-------------------------------------|
+| Astro Islands | Îles Astro | |
```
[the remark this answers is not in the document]
thomasbnt: Agreed for `île`!

i18n-guides/français.md
thomasbnt: `Paramètre fictif`? Seems weird and not match the english word. Agreed for `emplacement réservé` for this one. I don't have another word.

i18n-guides/français.md
thomasbnt: Agreed!

i18n-guides/français.md
thomasbnt: Oh good catch! `(Anglais)` is better for accessibility! :o

thomasbnt:
All of your propositions @ArmandPhilippot seems better for the intern documentation.
- It's better explained
- More accessible

i18n-guides/français.md
ArmandPhilippot: Yeah, I should have put some links here. I looked elsewhere what was done and there is no real consensus...

* On [this MDN page](https://developer.mozilla.org/fr/docs/Web/HTML/Reference/Elements/input/search#ajouter_un_texte_indicatif), they use `text indicatif` (the usage is very specific and can't be used everywhere).
* On [Wiktionary](https://fr.wiktionary.org/wiki/placeholder), they say: `Marque substitutive, remplaçant, élément générique ou de substitution.`
* There is also this [question on StackExchange](https://french.stackexchange.com/questions/780/how-would-you-translate-placeholder-to-french) or this [discussion for the translation of Drupal](https://traduction.drupal.fr/terme/placeholder) with some other suggestions.
* The [Grand dictionnaire terminologique](https://vitrinelinguistique.oqlf.gouv.qc.ca/fiche-gdt/fiche/26532924/parametre-fictif) from Quebec recommends:
  * `paramètre fictif`
  * `emplacement réservé` (or `espace réservé`)
  * `marque substitutive`
  * `marque de réservation`
  In the definition, while not suggested as a translation they also use `paramètre substituable`.
  
What bothers me with `emplacement réservé`, it's the `reservé` part... For example `Une place de parking réservée` for me it's a place you can't use because this is for someone else. So I interpret `emplacement réservé` the same way. Whereas in most cases we mean that it can/should be replaced.

`fictif` doesn't bother me but maybe this is not meaningful enough... I thought I had found an example where this made sense, and that's why I used it here, but I can't find it anymore...

Otherwise, I like `substituable`... maybe `élément substituable`? With, maybe, a note that this can be adapted to the context?

i18n-guides/français.md
ArmandPhilippot, line 147: I'm unresolving to update the instruction! What do you think of:
```suggestion
* S'il n'existe pas de version française, il peut être utile de rajouter la langue après le lien (p. ex. `texte du lien (Anglais)`) pour informer le lecteur que la page de destination est dans une autre langue.
```

ArmandPhilippot:
Hi @HiDeoo, I'm pinging you here in case you want to take a look.

One the objective of updating the Astro Docs `i18n-guide` would be to "provide a base to make the translation consistent in the Astro ecosystem (e.g. Starlight)". Since this is mainly you who works on the French translation of Starlight, you might want to give your opinion here. No obligations of course, I just wanted to make sure you're aware of this if you're interested!

dreyfus92 (approved):

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| (the) CLI / (the) command line interface | (la) CLI / (l')interface en ligne de commande | |
+| client-side | côté client | |
+| code fences (`---`) | barrières de code | |
+| endpoint | point de terminaison | |
+| export | exportation | |
+| feedback | retour / réaction / commentaire | |
+| footer | pied de page | |
+| header | en-tête | |
+| heading | titre ou en-tête | Nous n'avons pas de consensus (« titre » dans Astro Docs et « en-tête » dans Starlight). Précédente discussion : https://github.com/withastro/starlight/pull/2884#discussion_r1957379942 |
+| headless | sans-tête | |
```
HiDeoo, line 73: Not something I ever had to translate, but I guess depending on the context, ways around this super literal translation could exist maybe, e.g. for an headless UI library, I guess one could use "sans interface graphique", "sans affichage", etc.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| hook | crochet | |
```
HiDeoo, line 74: Not translated in Starlight [on purpose](https://fr.wiktionary.org/wiki/hook) as "crochet" is too literal and the suggested ["Fonction de crochet"](https://fr.wiktionary.org/wiki/hook#Nom_commun_2) does not bring a single relevant search result.

i18n-guides/français.md
HiDeoo: I guess it highly depends on the context but most of the time, I would find a mix between "indicatif" and "par défaut".

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| scope | limité / délimité | |
+| scoped | à portée limitée | |
+| server-side | coté serveur | |
+| template | modèle | |
+| to build | Suivant le contexte : créer / construire / compiler | |
+| to bundle | regrouper | |
+| to export | exporter | |
+| to fetch | récupérer | |
+| to import | importer | |
+| to render | rendre / afficher / restituer | |
```
HiDeoo, line 101: Note that in Starlight, we tried to always use "afficher" and tried to avoid using "rendre".

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| to style | mettre en forme / appliquer des styles | |
+| to wrap | englober / envelopper | |
+| type safe | sûreté du typage | |
+| UI | UI | |
+| update | mise à jour | |
+| upgrade | mise à niveau | |
+| view transitions | transitions de vue | |
+
+> [!NOTE]
+> Ces traductions sont des recommandations. Il est parfois acceptable d'utiliser une autre traduction : elle plus adaptée au contexte ou elle rend la lecture plus fluide en évitant des répétitions.
```
HiDeoo, line 112:
```suggestion
> Ces traductions sont des recommandations. Il est parfois acceptable d'utiliser une autre traduction si elle plus adaptée au contexte ou elle rend la lecture plus fluide en évitant des répétitions.
```

Not sure, the `:` felt weird to me when reading the sentence.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+
+## 📝 Conseils stylistiques
+
+### Les titres
+
+Certains guides de style anglais préconisent l'utilisation de majuscules pour chaque mot d'un titre. Ces règles ne s'appliquent pas en français. Il est donc recommandé de n'utiliser une majuscule qu'en début de titre, à l'exception des noms propres et des composants.
+
+### Les guillemets
+
+Bien que difficilement accessibles sur un clavier, privilégiez les guillements francophones (`«»`) aux guillemets doubles (`""`).
```
HiDeoo, line 125: > En [français de France](https://fr.wikipedia.org/wiki/Fran%C3%A7ais_de_France), les guillemets sont comptés comme des ponctuations hautes, et en tant que telles appellent l’usage d’espaces fines insécables[[10]](https://fr.wikipedia.org/wiki/Guillemet#cite_note-Lexique2002-10),[[11]](https://fr.wikipedia.org/wiki/Guillemet#cite_note-11) après le guillemet ouvrant et avant le guillemet fermant.

[reference](https://fr.wikipedia.org/wiki/Guillemet#Guillemets_en_chevrons) 

I guess we should at least require a space, even if it's non-breaking.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+
+Mémo Unicode :
+* `«` : <kbd>U+00ab</kbd>
+* `»` : <kbd>U+00bb</kbd>
+
+### Code en ligne utilisé comme mot dans une phrase
+
+Dans la version anglaise, vous verrez régulièrement des mots entre accents graves `` ` ``. Cette syntaxe n'a pas toujours de sens en français, il est donc parfois préférable de remplacer le code par un mot français et, si besoin, de rajouter la version code à côté.
+
+**Par exemple :** ``The file `path`...`` devrait être traduit par `Le chemin du fichier...` ou ``Le chemin du fichier (`path`)...``.
```
HiDeoo, line 135: I personally think that if the English version explicitly choose to use inline code to also reference an option name or something specific, the translation should always include it too.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+
+### Liens
+
+#### Internes
+
+Tous les liens doivent rediriger vers la version française, pensez donc à remplacer `/en/` par `/fr/` même si la page de destination n'est pas encore traduite. Vous devez également pensez à traduire la partie après `#`, quand elle existe : elle correspond a l'id généré pour un titre sur la page. Vous pouvez utiliser le serveur de développement pour vérifier l'id généré.
+
+#### Externes
+
+* L'ancre du lien est toujours traduit.
```
HiDeoo, line 145:
```suggestion
* L'ancre (ou texte) du lien est toujours traduit.
```

It may not be super obvious what this means for less technical reader.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+### Les exemples
+
+Pour que l'exemple reste compréhensible pour les moins anglophones des lecteurs, il est important de traduire les commentaires. Il peut également être intéressant d'adapter les chemins et les noms des variables si vous jugez que ça facilite la compréhension.
+
+Lorsque vous modifiez un exemple, pensez à vérifier les arguments passés au bloc de code pour également [mettre à jour la mise en évidence](https://expressive-code.com/key-features/text-markers/#usage-in-markdown--mdx) si besoin.
+
+## 📚 Ressources
+
+* En cas de doute pour une traduction, n'hésitez pas à consulter ce qui est fait dans d'autres documentations (MDN, React, Vue, etc.).
+* [Vitrine linguistique de l'Office québécois](https://vitrinelinguistique.oqlf.gouv.qc.ca/) : Bien que destiné aux québecois, les recommandations françaises sont souvent affichées. En utilisant les « termes privilégiés », vous devriez trouver une traduction pouvant convenir à tout le monde.
```
HiDeoo, line 158: I would personally add https://www.culture.fr/franceterme

i18n-guides/français.md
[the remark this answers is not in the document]
ArmandPhilippot: Thanks, I didn't think of `îlet`! So, yeah, I also think `îlot` would be better to describe this!
```suggestion
| Astro Islands | Îlots Astro | Précédente discussion : https://github.com/withastro/docs/pull/11593/files#r2074910929 |
```
(and we would also need to check the links on Starlight when Astro Docs will be updated... maybe some headings will change)

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+Certains mots ont un équivalent français qui devrait être utilisé uniformément dans les différentes traductions.
+
+| Anglais | Français | Explications/Référence (facultatif) |
+|---------|----------|-------------------------------------|
+| build | Suivant le contexte : création / construction / compilation | |
+| bundle | regroupement | |
+| changelog | journal des modifications | |
+| (the) CLI / (the) command line interface | (la) CLI / (l')interface en ligne de commande | |
+| client-side | côté client | |
+| code fences (`---`) | barrières de code | |
```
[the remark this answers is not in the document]
ArmandPhilippot: Currently we use `barrières de code` and I think I also saw some `clôtures de code` and `barres de code`. With `barrières` and `clôtures` we kept the same meaning as `délimitation` I think, but I agree it sounds a bit odd. I've thought about it before, but I haven't come up with any better ideas...

I saw another other translations elsewhere:
* `Fenced code blocks` > `blocs de code balisés` (so, it would be `code balisé`, but `balisé` sounds even weirder to me)

Maybe:
* `délimitateur de code`
* `séparateur de code`
* `cloison de code`

Right now, I don't have more ideas. 😅

i18n-guides/français.md
ArmandPhilippot: Oh yeah, it seems obvious to me but you're right maybe we should be more explicit here! Would adding something in between be enough?
```suggestion
Bien que difficilement accessibles sur un clavier, privilégiez les guillements francophones (`« texte »`) aux guillemets doubles (`"texte"`).
```
Or do you think we should write it in plain text:
```suggestion
Bien que difficilement accessibles sur un clavier, privilégiez les guillements francophones (`«»`) aux guillemets doubles (`""`). Un espace est requis après le guillemet d'ouverture et avant celui de fermeture.
```

i18n-guides/français.md
ArmandPhilippot: Yeah I think so too. I guess was thinking of avoiding repeats in some cases but well, it would be at the appreciation of the translator so maybe we should just say:
```suggestion
**Par exemple :** ``The file `path`...`` devrait être traduit par ``Le chemin du fichier (`path`)...``.
```
And update the sentence above to remove `si besoin`.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+
+Bien que difficilement accessibles sur un clavier, privilégiez les guillements francophones (`«»`) aux guillemets doubles (`""`).
+
+Mémo Unicode :
+* `«` : <kbd>U+00ab</kbd>
+* `»` : <kbd>U+00bb</kbd>
+
+### Code en ligne utilisé comme mot dans une phrase
+
+Dans la version anglaise, vous verrez régulièrement des mots entre accents graves `` ` ``. Cette syntaxe n'a pas toujours de sens en français, il est donc parfois préférable de remplacer le code par un mot français et, si besoin, de rajouter la version code à côté.
```
ArmandPhilippot, line 133: Related to HiDeoo comment's below:
```suggestion
Dans la version anglaise, vous verrez régulièrement des mots entre accents graves `` ` ``. Cette syntaxe n'a pas toujours de sens en français, il est donc parfois préférable de remplacer le code par un mot français et de rajouter la version code à côté.
```

i18n-guides/français.md
ArmandPhilippot: Good call, I was hoping others would have other useful links! It doesn't come up often when I do researches so I forgot about this one... but it seems really similar to the Quebec one, so I'll add it to my bookmarks!

```suggestion
* [FranceTerme](https://www.culture.fr/franceterme)
* [Vitrine linguistique de l'Office québécois](https://vitrinelinguistique.oqlf.gouv.qc.ca/) : Bien que destiné aux québecois, les recommandations françaises sont souvent affichées. En utilisant les « termes privilégiés », vous devriez trouver une traduction pouvant convenir à tout le monde.
```

i18n-guides/français.md
ArmandPhilippot: I'm not fan of `sans-tête`, so I'm in favor to update our translation! I kept that because many websites use the literal translation (and so, we also use that currently).

I think your suggestion could work with `headless CMS` too (this is mainly for this we have `headless`). But, `sans affichage`  or `sans interface graphique` could me misleading: a CMS can be headless and have a graphic interface for administrator.

Maybe `sans interface utilisateur`, `sans interface frontale`, or since we don't translate `front-end` I guess we could use `sans front-end`?

i18n-guides/français.md
ArmandPhilippot: Yeah `fonction de crochet` sounds odd and `crochet` might not be meaningful enough. So it might be better to add it to the section "Autres mots non traduits".

I've done some more researches and I came across this [discussion on StackExchange](https://french.stackexchange.com/questions/6698/traduction-de-hook-dans-un-contexte-de-programmation) where the suggestion doesn't seem to bad to me:
* `point d'ancrage`
* `point de branchement`

At least it sounds better than `fonction de crochet`.

i18n-guides/français.md
ArmandPhilippot: > it highly depends on the context
Yeah, totally! I think this is the issue here. That's why it's tricky to suggest a translation. So maybe we should just omit this word and manage this on a case-by-case basis on PRs?

Because when I think about it again, even `emplacement réservé` could make sense. But, it depends on the context. Sometimes it sounds really weird and other time it's fine.

Or we could just say something: `Suivant le contexte : valeur indicative, valeur par défaut, emplacement réservé, élément subsistuable` with the hope that the translator will understand this is not word for word and he can adapt it in its translation.

i18n-guides/français.md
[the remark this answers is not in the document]
HiDeoo: > (and we would also need to check the links on Starlight when Astro Docs will be updated... maybe some headings will change)

Definitely. I _think_ for islands the only link may be an example linking to https://docs.astro.build/fr/concepts/islands but I could be wrong.

i18n-guides/français.md
```diff
@@ -1,65 +1,158 @@
...
+| Island architecture | Architecture en îles | |
+| Sessions | Sessions | |
+
+#### Langage courant
+
+Certains mots ont un équivalent français qui devrait être utilisé uniformément dans les différentes traductions.
+
+| Anglais | Français | Explications/Référence (facultatif) |
+|---------|----------|-------------------------------------|
+| build | Suivant le contexte : création / construction / compilation | |
```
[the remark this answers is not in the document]
HiDeoo: Looks perfect to me 👍

i18n-guides/français.md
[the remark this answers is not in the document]
HiDeoo: Oh, `délimitateur de code` sounds quite ok I think, at least to me it sounds way better than the other alternatives. Second one would probably be `séparateur de code` while all the others sounds weird to me

i18n-guides/français.md
HiDeoo: It's a tricky one as the admin UI would have a front-end 😅 It's also worth noting that headless is more and more used, probably not at the same level of framework for example, but definitely used a lot.

I'll try to do more research about this one and see if something better comes up.

i18n-guides/français.md
HiDeoo: They sound better indeed altho, searching them and even specifying a domain, e.g. "programmation" leads to no result at all so I think I would personally be in favor of keeping this one untranslated.

i18n-guides/français.md
HiDeoo: I think that could be a valid alternative indeed, good call. In Starlight, we mostly always use it in the context of what ends up showing on screen so limiting it to "afficher" makes sense but yeah, for other context I think any other one sounds better than "rendre".

i18n-guides/français.md
HiDeoo: I think the first one may be enough 👍 And with time, if we see it needs to be mentioned a lot, we could switch to the second version if ever needed.

i18n-guides/français.md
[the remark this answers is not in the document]
ArmandPhilippot:
```suggestion
| code fences (`---`) | délimitateur de code | |
```

i18n-guides/français.md
ArmandPhilippot: Oh yeah, it sounds better when I wrote it but now... you're right it's the same. 😅 I wish there was an official translation but, it's probably best to keep the english work than a literal translation that doesn't mean much.

i18n-guides/français.md
ArmandPhilippot: I'm okay with this, I'll move the term in the other section! Thanks for brainstorming with me!

i18n-guides/français.md
ArmandPhilippot: Yeah each time I read `rendu` I think of 3D rendering... ([`Rendu`](https://fr.wikipedia.org/wiki/Rendu) / [`Pré-rendu`](https://fr.wikipedia.org/wiki/Pr%C3%A9-rendu)). So, it's probably best to update this:
```suggestion
| to render | afficher / générer / restituer | |
```

We'll need to update `rendering` too to be consistent but, I wonder what to do with `renderer` used, for example, in:
* [`Container API`](https://docs.astro.build/en/reference/container-reference/#renderers-option) 
  > Importing a renderer makes both the server and client renderers available to your container.
* [`Integrations guide`](https://docs.astro.build/en/guides/integrations-guide/)
  > Unlock React, Vue, Svelte, Solid, and other popular UI frameworks with a renderer.
  
  (currently the French page looks outdated, we miss the renderer part... it's convenient. 😆 )

If we use `moteur de rendu`, it could feel odd, right? Maybe `générateur`?
* `L'importation d'un générateur rend les générateurs du serveur et du client disponibles pour votre conteneur.`
* `Débloquez React, Vue, Svelte, Solid et d'autres frameworks UI populaires avec un générateur.`

I don't know, it still sounds odd to me.

i18n-guides/français.md
```diff
@@ -1,65 +1,159 @@
...
+| header | en-tête | |
+| heading | titre ou en-tête | Nous n'avons pas de consensus (« titre » dans Astro Docs et « en-tête » dans Starlight). Précédente discussion : https://github.com/withastro/starlight/pull/2884#discussion_r1957379942 |
+| headless | sans-tête | |
+| import | importation | |
+| issue | problème ou, en référence à la fonctionnalité de Github, ticket | [MDN](https://developer.mozilla.org/fr/docs/MDN/Community/Issues) utilise également « ticket » pour décrire la fonctionnalité de Github |
+| layout | mise en page | |
+| on-demand rendering | rendu à la demande | |
+| package | paquet | |
+| placeholder | paramètre fictif | |
+| preset | préréglage | « préconfiguration » pourrait convenir mais dans certains cas la formulation devient lourde. Discuté dans https://github.com/withastro/starlight/pull/3126#discussion_r2046673972 |
```
ArmandPhilippot, line 81: Should we add `plugin` to the list? On Astro Docs I think we always uses `plugin` while on Starlight we use `module d'extension`... maybe we should stay consistent?
```suggestion
| plugin | module d'extension | |
| preset | préréglage | « préconfiguration » pourrait convenir mais dans certains cas la formulation devient lourde. Discuté dans https://github.com/withastro/starlight/pull/3126#discussion_r2046673972 |
```

i18n-guides/français.md
HiDeoo: Did a bit more research, asked a few friends but I'm starting to think like you, maybe until we can find an obvious better translation which is commonly used, maybe we should just keep it in English for the time being.

i18n-guides/français.md
HiDeoo: It's weird: `rendre` feels wrong to me for `to render`, but on the opposite side, `moteur de rendu` feels like the best for `renderer` 😵‍💫 

While doing more research, I also found that this is exactly the wording used on Wikipedia: `Moteur de rendu` → `afficher`

image

Maybe there is a world where we keep the difference?

ArmandPhilippot: I don't have any other suggestion so I'm okay to keep `moteur de rendu`!

But the inconsistency bothers me so I did a little more research and it seems `rendu` (without `moteur` in front) is also used elsewhere:
* on MDN, [`Rendre`](https://developer.mozilla.org/fr/docs/Web/Performance/Guides/How_browsers_work#rendre) (27 occurrences 😅 ) or [HTML](https://developer.mozilla.org/fr/docs/Web/HTML#sujets_avanc%C3%A9s)
* [Google Webmaster Tools](https://support.google.com/webmasters/answer/11626894?hl=fr) (well I don't if this is auto-translated or not) uses `Recherchez des éléments dans le rendu HTML`
* a [Razor page](https://learn.microsoft.com/fr-fr/aspnet/core/mvc/views/razor?view=aspnetcore-9.0#rendering-html) on Microsoft
* I also see this word uses on university (Poitiers) or school (Gobelins) websites 

Maybe we should avoid `rendu` when we have better words to describe what happens (and, above all, avoid to repeat it) but without explicitely discouraging it?

HiDeoo: I think that's an acceptable compromise and personally, it's only the verb "rendre" that bothers me for w/e reason :sweat_smile:

ArmandPhilippot: Oh okay, I think I better understand what you mean. Yeah, `rendre la page` sounds awful. So instead of using `rendre` we could add to the list `effectuer un rendu`:

```suggestion
| to render | afficher / générer / effectuer un rendu / restituer | |
```

i18n-guides/français.md
ArmandPhilippot: Okay so I think this is the last thing to tackle unless someone has something else to report! I looked at some examples where `placeholder` is used in the docs. Here's three examples:
* [`The <slot /> element is a placeholder for external HTML content`](https://docs.astro.build/en/basics/astro-components/#slots)
* [`Then, replace any placeholder “scripts” section of your package.json with the following:`](https://docs.astro.build/en/install-and-setup/)
* [`Your page's main content can be rendered immediately with placeholder content, such as a generic avatar until your island's own content is available.`](https://docs.astro.build/en/concepts/islands/#server-islands)

While `par défaut` could be suitable for the second one (and maybe better than the current translation `toute section réservée « scripts »`), it feels wrong for the first one and not accurate for the third one. I don't think `indicatif` sounds better here.
Currently for the first example we use `emplacement réservé pour` and for the third one we use `un contenu de remplacement`. They seem fine to me (well, I'd prefer `contenu substituable` for the third one but I think the meaning is here so it sounds right).

I'm not sure what to do here. With only three examples, we have three different possible translations. I'm not sure suggesting multiple translations would bring much value here because another example could use something else. So either:
* we remove that word from the list and leave it to the appreciation of reviewers
* keep it and either suggest some examples (because the note under the table suggests it's okay to use something else) or put something like "à traduire mais à adapter suivant le contexte".

i18n-guides/français.md
```diff
@@ -1,65 +1,160 @@
...
+| UI | UI | |
+| update | mise à jour | |
+| upgrade | mise à niveau | |
+| view transitions | transitions de vue | |
+
+> [!NOTE]
+> Ces traductions sont des recommandations. Il est parfois acceptable d'utiliser une autre traduction si elle est plus adaptée au contexte ou elle rend la lecture plus fluide en évitant des répétitions.
+
+> [!TIP]
+> Il est parfois utile pour faciliter la recherche d'un concept de proposer à certains endroits, avec parcimonie, deux versions couramment utilisées. Par exemple : `a headless CMS` pourrait être traduit de la façon suivante `un CMS sans-tête (ou « headless »)`.
```
ArmandPhilippot, line 116: Oh another thing while I think about it, since we moved headless, we probably should update this example:
```suggestion
> Il est parfois utile pour faciliter la recherche ou la compréhension d'un concept de proposer à certains endroits, avec parcimonie, deux versions. Par exemple : `the component props` pourrait être traduit de la façon suivante `les props (ou « propriétés ») du composant`.
```
We don't do that often, so except for `props` I didn't find another example...

HiDeoo: Good catch.

Can't think of another example right now, the only one coming to mind from the Starlight docs is:

> Les encarts (également connus sous le nom de « admonitions » ou « asides » en anglais)

Altho, it's super specific to Starlight so not as good as props I think. 👍

i18n-guides/français.md
ArmandPhilippot: Ah! I found an example where `fictif` is used... in Starlight! 😄 [Ajouter des entrées fictives](https://starlight.astro.build/fr/components/file-tree/#ajouter-des-entr%C3%A9es-fictives) which seems fine based on the English version `Add placeholder files and directories`. If don't think any other word would have make sense here.

So, maybe something like:
```suggestion
| placeholder | mot + par défaut / réservé / substituable / fictif | La traduction dépend fortement du contexte, les propositions précédentes ne sont que des exemples possibles. Précédente discussion : https://github.com/withastro/docs/pull/11593#discussion_r2071856903 |
```

ArmandPhilippot:
Okay this time I think everything has been addressed! So if you want to take another pass, you can also check how it looks with the [current file preview](https://github.com/withastro/docs/blob/518c3e1e7aacf71a5c9c3c7825daf295f2153142/i18n-guides/fran%C3%A7ais.md). It's not ideal on smartphone because of tables but... I don't think we have much choices.

cc @HiDeoo, @thomasbnt and @dreyfus92 (and thank you all for chime in here! 🙌🏽  )

HiDeoo (approved):
Did a final read-through and I'm personally happy with the changes, and as mentioned, even if we're forgetting a few things, "Ce glossaire est vivant : n'hésitez pas à l'enrichir quand le besoin se fait sentir !" so we can always update/add more terms later if needed.

Thanks a lot for kicking this off and for the great work on this :clap:

PR #11608, opened 2025-05-04
docs(en/editor-setup.mdx): IDX is now Firebase Studio

xPecoide:
#### Description (required)

Google changed the Project IDX name to "Firebase Studio", now with a new link that will be the standard one. I changed the reference in Astro Docs "Editors" page.


vrabe:
[IDX is Astro’s Official Online Editor Partner.](https://astro.build/blog/idx-official-online-editor-partner/) So the docs need to be changed may be more than this...

sarah11918:
Hi! Thank you for pointing this out @xPecoide ! And yes, @vrabe is correct that Google's Project IDX is an official sponsor and we have multiple references throughout docs.

We are actually discussing with them how they would best like to proceed with updating their content, logo, links etc. throughout our site, so I will close this PR at this time because this is something to be agreed upon with our sponsor, and we will be doing this as part of a larger update to make sure nothing is missed!

The current links still function (I am still using IDX myself with the old links no problem) , so there is no harm keeping the docs as is a little longer.

PR #11610, opened 2025-05-04
Document Astro.originPathname

TheOtterlord:
Document the `originPathname` context, implemented in https://github.com/withastro/astro/pull/12373 and first released in `astro@5.0.0-beta.8`


sarah11918:
Thanks @TheOtterlord !

I'm wondering whether there should be mention of/link to this in the `rewrite()` entry? (it's specific to rewrites only? not redirects?)

Also, given our description of rewrites:

>rewrite() allows you to serve content from a different URL or path without redirecting the browser to a new page.

It doesn't *seem* like `Astro.url.pathname` and `Astro.originPathname` would be different? I would expect `url` to match the URL in my browser bar, and if we haven't redirected the browser to a new path (only pulled content in from a new path), then do either of the descriptions for `redirect` or `url` have to change?

TheOtterlord:
> I'm wondering whether there should be mention of/link to this in the rewrite() entry? (it's specific to rewrites only? not redirects?)

Yes, this is specific to rewrites. I positioned it under `rewrite` since it's related, but we can also find a way mention it there if it's helpful.

> It doesn't seem like Astro.url.pathname and Astro.originPathname would be different? I would expect url to match the URL in my browser bar, and if we haven't redirected the browser to a new path (only pulled content in from a new path), then do either of the descriptions for redirect or url have to change?

`Astro.url.pathname` is the URL of the page you rewrite to, which is why `originPathname` exists (afaik). So `Astro.originalPathname` matches the path you'd see in the browser, whereas `Astro.url.pathname` would be the `/404` or whatever you called `rewrite()` with.

sarah11918:
So, my confusion is that this is what we say about rewrites in the routing guide:

> A rewrite allows you to serve a different route without redirecting the browser to a different page. **The browser will show the original address in the URL bar**, but will instead display the content of the URL provided to [Astro.rewrite()](https://docs.astro.build/en/reference/api-reference/#rewrite).

So, I'm trying to wrap my head around what I think the above says:

- someone requests URL-a (is this `originPath`? the original pathname requested, before rewrites were applied?)
- according to what we say about rewrites, the browser bar displays that you're on URL-a (what I'd expect `url.pathname` to show because that is the path you can see?) but shows the content that exists at URL-b.
- is URL-b the `url.pathname` since it's the location of the content you're seeing (but not actually the URL pathname of the page you're on right now? If that's the case, if `url.pathname` can be a different address then you are currently visiting, then I think that needs more explaining?

I get that we request one path and see the content for a different one, and that there's a "ask for this, get this" relationship. But, if the URL of the initial request, and the one shown in the browser are the same (because you didn't redirect, you just put different content at the requested path), then I would *expect* displaying `originPath` (the path I requested) and `url.pathname` (the URL I see in my browser tab, because I did in fact go there, and that wasn't changed on me) to be the same.

This is what I'm trying to figure out, and whether any other explanations need to change, because the example you added here makes me think that those would return two different URL paths, but I'm trying to figure out how that would be. 😅

sarah11918:
Discussed on Talking and Doc'ing 2025-06-05:

- should we have "Astro/context" tabs here too like the other entries?
- should we explicitly talk about a new rendering phase being triggered, like we do in the Middleware guide?
- shared the implementation PR for additional context https://github.com/withastro/astro/pull/12373
- this is a great reference entry, but realizing we haven't documented this anywhere, means we don't have any place where the more why/use case info is included. So, this can feel "accurate" and "useful" but still not feel "fully documented" and truly helpful to readers.

@TheOtterlord - see which of the above technical points (e.g. adding tabs for Astro/context) you want to include here, and we can consider whether maybe we need something added to the Middleware guide showing this used, but get this helpful, accurate reference merged!

ArmandPhilippot:
Related to T&D, I just found a use case in the current docs: https://docs.astro.build/en/guides/actions/#advanced-persist-action-results-with-a-session but we don't explicitly tell why we use `Astro.originPathname`. I guess this example would fail using `Astro.url.pathname` if someone had configured a rewrite?

Maybe we can do something around that regarding:
> we don't have any place where the more why/use case info is included

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -488,6 +488,21 @@ Use a `Request` type for complete control of the `Request` sent to the server fo
...
+**Type:** `string`<br />
+<Since v="5.0.0" />
+</p>
+
+`originPathname` defines the original pathname of the request, before rewrites were applied.
+
+```astro title="src/pages/index.astro"
+<p>The origin path is {Astro.originPathname}</p>
+<p>The rewritten path is {Astro.url.pathname}</p>
+```
```
sarah11918:
``````suggestion
<Tabs>
  <TabItem label="Astro.originPathname">
  	```astro title="src/pages/index.astro"
		<p>The origin path is {Astro.originPathname}</p>
		<p>The rewritten path is {Astro.url.pathname}</p>
		```
  </TabItem>
  <TabItem label="context.originPathname">
    ```ts title="src/middleware.ts"
  	// Redirect to the destination page on success
    return context.redirect(context.originPathname);
    ```
  </TabItem>
</Tabs>
``````
So, here's a super minimal version using the example Armand found that used `context.originPathname` that maybe matches the level of detail of the `.astro` example?

Comment (ArmandPhilippot) [reply src/content/docs/en/reference/api-reference.mdx]: I looked to see if I could find an example where using both `context.originPathname` and `context.url.pathname` would make sense, to have a 1:1 match between tabs, but without success... then I forgot to reply here. 😅 

Nit: it seems the other `context` snippets always wrap the example in a `GET` or `POST` function (API routes), so we could stay consistent even if this is a middleware example. And maybe, we can still show an example using both `context.originPathname` and `context.url.pathname`:

```ts
import { defineMiddleware } from 'astro:middleware';

export const onRequest = defineMiddleware(async (context, next) => {
  // If the URL was rewritten, do something in the middleware
  if (context.url.pathname !== context.originPathname) {
    // Then, redirect to the destination page on success
    return redirect(context.originPathname);
  }
  return next();
}
```

Not sure what could be done here in a real word example... but if we want to show the value is not always the same, I think this could work.

Otherwise, I think your suggestion could be enough and we can always revisit this later if needed.

sarah11918: Thank you for this @ArmandPhilippot ! 

@TheOtterlord would you like to take a look, and I leave the final call here to you!

sarah11918:
Gentle ping @TheOtterlord that you might have all the info here you're going to get, and I trust you to make a great decision and just merge!

theotterlord:
Thanks. This keeps getting lost in my notifications :sweat_smile:

Will get to this tonight!

sarah11918 (approved):
Looks fantastic! Thank you @theOtterlord :partying_face:

theotterlord:
Haha, oops. Was experimenting with other things in the background :sweat_smile:
Will undo that and merge this if it's good as-is

PR #11627, opened 2025-05-07
Update client-side-scripts.mdx

jazsouf:
#### Description

The 'constructor()' method is never mentioned in any piece of code in the page.
And from what I understand, it's the 'connectedCallback()' method that runs each time the browser finds a custom element in the DOM.


ArmandPhilippot (approved):
Nice catch, the code snippet has been updated in #9314 but we forgot to update the steps... So, thank you! And welcome to Team Docs 🥳

PR #11632, opened 2025-05-08
fix: update `fonts.mdx`

jsparkdev:
#### Description (required)

Clarified ambiguous wording.

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -551,7 +551,7 @@ If the associated font is a [variable font](https://developer.mozilla.org/en-US/
 weight: "100 900"
 ```
 
-If the value is set to `undefined`, Astro will try to infer the value based on the first [`source`](#src).
+When a value is not set, by default Astro will try to infer the value based on the first [`source`](#src).
 
 #### style
 
@@ -567,7 +567,7 @@ A [font style](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style):
 style: "normal"
 ```
 
-If the value is set to `undefined`, Astro will try to infer the value based on the first [`source`](#src).
+When a value is not set, by default Astro will try to infer the value based on the first [`source`](#src).
 
 #### src
 
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -444,7 +444,7 @@ display: "block"
 
 Determines when a font must be downloaded and used based on a specific [range of unicode characters](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range). If a character on the page matches the configured range, the browser will download the font and all characters will be available for use on the page. To configure a subset of characters preloaded for a single font, see the [subsets](#subsets) property instead.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -444,7 +444,7 @@ display: "block"
...
 
-This can be useful for localization to avoid unnecessary font downloads when a specific part of your website uses a different alphabet and will be displayed with a separate font. For example, a website that offers both English and Japanese versions can prevent the browser from downloading the Japanese font on English versions of the page that do not contain any of the Japanese characters provided in `unicodeRange`.
+This can be useful for localization to avoid unnecessary font downloads when a specific part of your website uses a different language and will be displayed with a separate font. For example, a website that offers both English and Japanese versions can prevent the browser from downloading the Japanese font on English versions of the page that do not contain any of the Japanese characters provided in `unicodeRange`.
```
sarah11918, line 447: I think the word "alphabet" was chosen specifically to refer to characters to be displayed, since we're talking about fonts. (Since e.g. English and French use mostly the same *alphabet*, just with some additional characters for accented letters in French.)

So I think alphabet may be a more precise distinction here, but I'll let Florian decide!

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -551,7 +551,7 @@ If the associated font is a [variable font](https://developer.mozilla.org/en-US/
 weight: "100 900"
 ```
 
-If left undefined, Astro will try to infer the value based on the first [`source`](#src).
+If the value is set to `undefined`, Astro will try to infer the value based on the first [`source`](#src).
```
sarah11918, line 554: I think the idea of "left as default" is important, because no one will actively "set" a value of `undefined`.

There could be a better way to say this though! Perhaps:

```suggestion
When a value is not set, by default Astro will try to infer the value based on the first [`source`](#src).
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -567,7 +567,7 @@ A [font style](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style):
 style: "normal"
 ```
 
-If left undefined, Astro will try to infer the value based on the first [`source`](#src).
+If the value is set to `undefined`, Astro will try to infer the value based on the first [`source`](#src).
```
sarah11918, line 570: Whatever we decide above, we'll do the same thing here!

sarah11918 (commented):
Thanks @jsparkdev - I left some comments but will want @florian-lefebvre to make the final decisions about the wording that best represents his feature/intention!

src/content/docs/en/reference/experimental-flags/fonts.mdx
florian-lefebvre: I'm not an expert of this CSS property but I believe alphabet is right there

src/content/docs/en/reference/experimental-flags/fonts.mdx
florian-lefebvre:
```suggestion
When the value is not set, by default Astro will try to infer the value based on the first [`source`](#src).
```
I would say it like this, I kept `the` since we refer to the current property (eg. weight)

jsparkdev: Thanks for good point!

src/content/docs/en/reference/experimental-flags/fonts.mdx
florian-lefebvre, line 447: Don't forget to revert this line

florian-lefebvre (approved):

sarah11918 (approved):
LGTM!

yanthomasdev (approved):
LGTM!

PR #11634, opened 2025-05-08
Fix import for the `astro:config/server` section in `astro-config.mdx`

notjb:
#### Description

Add proper import for the `astro:config/server` section in `astro-config.mdx`

#### First-time contributor to Astro Docs ✌️
username: `notjb`


sarah11918 (approved):
How did we never catch this before?? Thank you so much for fixing this up for us, @notjb , and welcome to Team Docs! 🥳

PR #11657, opened 2025-05-12
fix(astro-components): update syntax highlighting and add link

ArmandPhilippot:
#### Description (required)

Updates the `basics/astro-components.mdx` file to:

* fix a broken syntax highlighting (noticed while updating the French file, in French this is broken... in English this is invisible)

|Before|After|
|---|---|
|Broken syntax highlighting before|broken syntax highlighting after|

* update a syntax highlighting because the example is describing using `slot` with `Fragment` so I think it could be useful to highlight the closing tag too:

|Before|After|
|---|---|
|Incomplete syntax highlighting before|Incomplete syntax highlighting after|

* add a link to `is:inline` directive because in [HTML components](https://docs.astro.build/en/basics/astro-components/#html-components) we are referring twice to this directive so newcomers could ask themselves what it is.

- Suggested label:  code snippet update, typo/link/grammar - quick fix!


sarah11918 (approved):
Great catch, and we can use all the help with `slot` docs we can get! 🚀

PR #11680, opened 2025-05-15
fix: update `actions.mdx`

jsparkdev:
#### Description (required)

- Update codeblock indents
- Clarify what input refers to


src/content/docs/en/guides/actions.mdx
```diff
@@ -423,16 +423,16 @@ import { actions } from 'astro:actions';
 </form>
 ```
 
-Depending on your inputs, additional attributes on the `<form>` element may be necessary for proper schema validation with Zod. For example, to include file uploads, add `enctype="multipart/form-data"` to ensure that files are sent in a format correctly recognized by `z.instanceof(File)` inputs:
+Depending on type of `<input>`, additional attributes on the `<form>` element may be necessary for proper schema validation with Zod. For example, to include file uploads, add `enctype="multipart/form-data"` to ensure that files are sent in a format correctly recognized by `z.instanceof(File)`:
```
sarah11918, line 426:
```suggestion
Additional attributes on the `<form>` element may be necessary for proper schema validation with Zod. For example, to include file uploads, add `enctype="multipart/form-data"` to ensure that files are sent in a format correctly recognized by `z.instanceof(File)`:
```
I was trying to figure out how to avoid a sentence using both `<input>` and `type` as inline code so close together, then realized that maybe we can just simply to this? What do you think?

jsparkdev: Sounds good!

I don't think it's hard to understand what it means.

sarah11918 (approved):
LGTM! Thank you @jsparkdev ! 💜

yanthomasdev (approved):
LGTM!

PR #11688, opened 2025-05-15
feat: add explanation for Prettier plugin order to Editor setup page

### LINKED ISSUE #10877: Plugin required disclaimer for Prettier section of the editor setup docs
while trying to set up prettier with `"prettier-plugin-tailwindcss"` plugin, I faced an issue where the tailwind class sorting wasn't working .astro files but worked fine on .tsx files.

The issue turned out to be that because how  `"prettier-plugin-tailwindcss"`  behaves (it's required that it's the [last plugin in tailwind config](https://github.com/tailwindlabs/prettier-plugin-tailwindcss#compatibility-with-other-prettier-plugins)) the `prettier-plugin-astro` plugin needs to be installed
and the .prettierrc should look something like :
```ts
{
  "plugins": [
    "prettier-plugin-astro",
    "prettier-plugin-tailwindcss" // needs to be the last plugin in the array
  ],
  "overrides": [
    {
      "files": "*.astro",
      "options": {
        "parser": "astro"
      }
    }
  ]
}
```
The [Docs for the use of Prettier with Astro](https://docs.astro.build/en/editor-setup/#prettier) mentions:
If you’re using the [Astro VS Code Extension](https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode) or [the Astro language server within another editor](https://docs.astro.build/en/editor-setup/#other-code-editors), code formatting with Prettier is included."

but there's no mention that the `prettier-plugin-astro` would be required for .astro files when using the prettier tailwind plugin.

anaxite:
#### Description

This change adds an extra step to the Prettier setup instructions of `editor-setup.mdx`. The extra step discusses adding other Prettier plugins, making sure to mention the importance of their order. There is an example that specifically shows `prettier-plugin-tailwindcss` as the last one.

I went with an extra step to avoid cluttering step 2, and to try keeping concepts separate. 

#### Related issues & labels

- Closes #10877.

### DIFF
--- a/src/content/docs/en/editor-setup.mdx
+++ b/src/content/docs/en/editor-setup.mdx
@@ -116,7 +116,7 @@ To add support for formatting `.astro` files outside of the editor (e.g. CLI) or
     }
     ```
 
-3. Optionally, install other Prettier plugins for your project, and add them to the configuration file. These additional plugins may need to be listed in a specific order. For example, if you use Tailwind, `prettier-plugin-tailwindcss` must be the last Prettier plugin in the plugins array.
+3. Optionally, install other Prettier plugins for your project, and add them to the configuration file. These additional plugins may need to be listed in a specific order. For example, if you use Tailwind, `prettier-plugin-tailwindcss` must be [the last Prettier plugin in the plugins array](https://github.com/tailwindlabs/prettier-plugin-tailwindcss#compatibility-with-other-prettier-plugins).
 
     ```json title=".prettierrc"
     {
--- a/src/content/docs/en/editor-setup.mdx
+++ b/src/content/docs/en/editor-setup.mdx
@@ -116,7 +116,9 @@ To add support for formatting `.astro` files outside of the editor (e.g. CLI) or
     }
     ```
 
-3. Run the following command in your terminal to format your files.
+3. Optionally, install other Prettier plugins for your project, and add them to the configuration file. These additional plugins may need to be listed in a specific order. For example, if you use Tailwind, `prettier-plugin-tailwindcss` must be the last Prettier plugin in the plugins array.
+
+4. Run the following command in your terminal to format your files.
 
     <PackageManagerTabs>
       <Fragment slot="npm">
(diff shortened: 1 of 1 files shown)


src/content/docs/en/editor-setup.mdx
```diff
@@ -116,7 +116,26 @@ To add support for formatting `.astro` files outside of the editor (e.g. CLI) or
...
     ```
 
-3. Run the following command in your terminal to format your files.
+3. Optionally, install other Prettier plugins for your project, and add them to the configuration file. These additional plugins may need to be listed in a specific order. For example, if you use Tailwind, `prettier-plugin-tailwindcss` must be [the last Prettier plugin in the plugins array](https://github.com/tailwindlabs/prettier-plugin-tailwindcss#compatibility-with-other-prettier-plugins).
+
+    ```json title=".prettierrc"
+    {
+      "plugins": [
+        "prettier-plugin-astro",
+        "prettier-plugin-tailwindcss" // needs to appear after other Prettier plugins
```
sarah11918, line 125:
```suggestion
        "prettier-plugin-tailwindcss" // needs to be last
```
Something shorter here would avoid any horizontal scroll on the code snippet!

If this is a prettier file, and this is the `plugins` section, is there anything *other than* a Prettier plugin that could be listed here? Maybe we don't need to be so specific?

sarah11918 (commented):
This looks great, and thank you for jumping on this issue!

Just had one small comment about the code comment line. Maybe we can get that shorter?

sarah11918:
Looks like we all got busy, @anaxite ! 😂 I did leave a quick comment, which big if true! (If true, I think it's *better*?)

Otherwise, this one really should be able to get done, so let me know about that comment, and we'll get it merged in! 🙌

src/content/docs/en/editor-setup.mdx
anaxite: Good idea; I was being too careful at first.

anaxite:
Thanks for the reminder @sarah11918 ! I agree with your comment, and this PR should be good to go.

sarah11918 (approved):
Looks great @anaxite ! Thank you for tackling this one for us!

PR #11692, opened 2025-05-15
Update required Node version to 18.20.8

ascorbic:
#### Description (required)

Updates the minimum Node version to 18.20.8. Adds info to Cloudflare Pages guide about setting the Node version.

The updated requirements are implemented in https://github.com/withastro/astro/pull/13809 and will be released in the next minor

- Closes #
- Suggested label:


src/content/docs/en/tutorial/1-setup/1.mdx
```diff
@@ -28,18 +28,18 @@ You can access the command line through a local terminal program for your operat
 
 ### Node.js
 
-For Astro to run on your system, you will also need to have [**Node.js**](https://nodejs.org/en/) installed, version `v18.17.1` or `v20.3.0` or later. (`v19` is not supported.)
+For Astro to run on your system, you will also need to have [**Node.js**](https://nodejs.org/en/) installed, version `v18.20.8` or `v20.3.0`, `v22.0.0` or higher. (`v19` and `v21` are not supported.)
```
sarah11918, line 31: This sentence structure doesn't work as well here now that there are multiple versions "or higher" for each one of them.

What about something like:

```suggestion
For Astro to run on your system, you will also need to have a compatible version of [**Node.js**](https://nodejs.org/en/) installed. Astro supports **even-numbered** Node.js versions. The current minimum supported versions of each are:  `v18.20.8`, `v20.3.0`, and `v22.0.0`. (`v19` and `v21` are not supported.)
```

(Even if v18 must be *exactly* v.18.20.8, if there's no higher version that we *don't* support, I think the wording is fine for a more readable sentence?)

Alternatively, we go into bullet points (which weren't really needed before when it was simple enough to just say in a sentence):

```suggestion
For Astro to run on your system, you will also need to have a compatible version of [**Node.js**](https://nodejs.org/en/) installed. Astro supports **even-numbered** Node.js versions. The current supported versions are:

 - `v.18.20.8`
 - `v.20.3.0` or higher
 - `v.22.0.0` or higher 
```

src/content/docs/en/tutorial/1-setup/1.mdx
```diff
@@ -28,18 +28,18 @@ You can access the command line through a local terminal program for your operat
...
 ```sh
 node -v
 
 // Example output
-v18.17.1
+v18.20.8
 ```
 
-If the command returns a version number higher than `v18.17.1` or `v20.3.0` (excluding any `v19`), you're good to go!
+If the command returns a version number higher than `v18.20.8` or `v20.3.0` (excluding any `v19` or `v21`), you're good to go!
```
sarah11918, line 42:
```suggestion
If the command returns a version number supported by Astro, you're good to go!
```
Again, I don't think we can say this simply here like we could before. So maybe this is sufficient?

sarah11918 (commented):
Thanks for updating, Matt! Left some suggestions for maybe simplifying this now that there are more versions to describe.

yanthomasdev (approved):
LGTM! Thanks!

PR #11716, opened 2025-05-19
fix example `src/components/MarkdocFigure.astro` in markdoc.mdx

mrienstra:
Had `src: ImageMetadata | string;` but no `string` handling

Alternate solution would be to remove ` | string` from type, I instead opted to use the pattern shown in the example directly above.

Without these changes, `pnpm astro check` results in:

> error ts(2322): Type '{ src: string | ImageMetadata; alt: string; width: number; height: number; }' is not assignable to type 'IntrinsicAttributes & Props'.

PS: was tempted to make this a slightly more "kitchen sink" example, by throwing in an example of `<slot/>` (also functionally a nice thing to have, if someone reading the docs actually has a use for this particular component):

```astro
---
import { Image } from "astro:assets";

interface Props {
  src: ImageMetadata | string;
  alt: string;
  width: number;
  height: number;
}

const { src, alt, width, height } = Astro.props;
const hasCaptionSlot = Astro.slots.has("default");
---

<figure>
  {
    typeof src === "string" ? (
      <img {src} {alt} {width} {height} />
    ) : (
      <Image {src} {alt} {width} {height} />
    )
  }
  {
    hasCaptionSlot && (
      <figcaption>
        <slot />
      </figcaption>
    )
  }
</figure>
```

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/markdoc.mdx
+++ b/src/content/docs/en/guides/integrations-guide/markdoc.mdx
@@ -498,8 +498,14 @@ The following steps will create a custom Markdoc image tag to display a `<figure
     const { src, alt, width, height, caption } = Astro.props;
     ---
     <figure>
-        <Image {src} {alt} {width} {height}  />
-        {caption && <figcaption>{caption}</figcaption>}
+      {
+        typeof src === "string" ? (
+          <img {src} {alt} {width} {height} />
+        ) : (
+          <Image {src} {alt} {width} {height} />
+        )
+      }
+      {caption && <figcaption>{caption}</figcaption>}
     </figure>
     ```
 


TheOtterlord:
Thanks @mrienstra.

I went about reproducing this, since I remembered remote images only using a string for `src`. And yes, it's unfortunately just a weird type conflict, because strictly defining `src` (as a `string` or `ImageMetadata`) works fine.

image

This should probably be reported as a bug, since the best case is fixing this conflict. Making a note of what we might want to do if this isn't considered a bug though:

We can define `Props` as an intersection between `Image` props and custom props like `caption`.

```astro
---
import { Image, type LocalImageProps, type RemoteImageProps } from "astro:assets";

type Props = (LocalImageProps | RemoteImageProps) & {
  caption: string;
}

const { caption, ...imageProps } = Astro.props;
---

<figure>
  <Image {...imageProps} />
  {caption && <figcaption>{caption}</figcaption>}
</figure>
```

mrienstra:
Nice work-around, @TheOtterlord!

Do you want to open an issue (on main astro repo) for this or shall I?

Perhaps we should close this PR and open a docs issue to discuss the right solution. As I noted above, the example above this -- see https://docs.astro.build/en/guides/integrations-guide/markdoc/#custom-image-components -- uses a similar pattern. This prior example also _kind of_ implies you need to make a custom image component if you want `A picture of a cat` in `.mdoc` files to use the `<Image />` component (or "Astro’s [Image Service API](https://docs.astro.build/en/reference/image-service-reference/)" as it us explained [elsewhere](https://docs.astro.build/en/guides/images/#images-in-markdown-files)) to optimize your local images -- I'm pretty sure this works "out of the box" in `.mdoc` files, like it does in `.md` files ( https://docs.astro.build/en/guides/images/#images-in-markdown-files ), but I'll have to double-check this... That would be a weird omission.

sarah11918:
> Perhaps we should close this PR and open a docs issue to discuss the right solution.

Because this took a turn from the original intention, and because of how long this has been open without activity, I'm going to accept this decision to close it, and welcome a docs discussion! (If no one filed an issue with the main astro repo, then I also welcome you to do that!)

PR #11722, opened 2025-05-20
fix(guides/cms): update some FileTree and code snippets

ArmandPhilippot:
#### Description (required)

Updates some `guides/cms` pages to fix:
* some indentation in `<FileTree />` components (files wasn't under `src/` because of that)
* some code snippets title (we had an extra `"` before the title because the closing quote was missing)
* some code snippets languages
* a line marker in a code snippet to include the closing brace
* a typo in `sitecore.mdx`

In addition:
* in `preprcms.mdx`: rewrite the file trees because they were a bit off compared to what is described in the guide (ie. none of the files were under `src/`) and remove a leftover from #11501
* in `contentful.mdx`: replace `Server side rendering` with `On demand rendering` reusing the same wording used in Strapi.

I also noticed that [Keystatic](https://docs.astro.build/en/guides/cms/keystatic/#displaying-a-single-entry) uses the legacy content collection API but I preferred not to update it. I don't know if the CMS is compatible with the Content Layer API; their docs is also outdated. It seemed like it needed bigger changes (adding a step for `content.config.ts`?).

### DIFF
--- a/src/content/docs/en/guides/cms/contentful.mdx
+++ b/src/content/docs/en/guides/cms/contentful.mdx
@@ -428,9 +428,9 @@ const { content, title, date } = Astro.props;
 
 Navigate to http://localhost:4321/ and click on one of your posts to make sure your dynamic route is working!
 
-#### Server side rendering
+#### On-demand rendering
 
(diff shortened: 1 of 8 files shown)


yanthomasdev (approved):
LGTM! Good job! 🙌

PR #11743, opened 2025-05-22
docs(cloudflare): custom entryfile

alexanderniebuhr:
docs for https://github.com/withastro/astro/pull/13837

I'm not sure if we should a full section which explain the usage of this a bit more, or if it is enough to document the new options like I did? Also maybe the two options should be nested under one common prefix `entryfile.path` & `entryfile.exports`, however not to sure 🤔

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -305,7 +305,7 @@ Provide any additional defined named exports of your [custom entry file](#creati
 
 #### Creating a custom Cloudflare Worker entry file
 
-The custom entry file must export the `createExports()` function, with a `default` export including all the handlers you need.
+The custom entry file must export the `createExports()` function with a `default` export including all the handlers you need.
 
 The following example entry file registers a Durable Object and a queue handler:
 


alexanderniebuhr:
It's still in sync, but I will update it and ping you.

alexanderniebuhr:
cc @sarah11918 I've updated this due to changes in the core PR. Also add some comments.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,62 @@ export default defineConfig({
 });
 ```
 
+### `workerEntryPoint`
+<p>
+**Type:** `{path: string | URL, exports: string[]}`
+**Default:** `{ path: '@astrojs/cloudflare/entrypoints/server.js', exports: ['default'] }`
+<Since v="12.6.0" />
```
alexanderniebuhr: This is the adapter version, not core version. For now we used `<Since>` only for core versions. Not sure if the users might be confused now.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,62 @@ export default defineConfig({
...
+			},
+			async queue(batch, _env) {
+				let messages = JSON.stringify(batch.messages);
+    		console.log(`consumed from our queue: ${messages}`);
+			}
+		} satisfies ExportedHandler<Env>,
+		MyDurableObject: MyDurableObject,
+	}
+}
+```
```
alexanderniebuhr: All equally important, so not really any lines to highlight. Question if it would be better to just link to our [fixture](https://github.com/withastro/astro/pull/13837/files#diff-5fdadf72f60b4847de751747a474316912444f17a9ce443084bbda0afa038bff) if users need an example?

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
sarah11918: Great observation! We can actually specify a particular integration for the version! (See the [`<Since />` documentation](https://contribute.docs.astro.build/reference/custom-components/#since-component).

```suggestion
<Since v="12.6.0" pkg="@astrojs/cloudflare"/>
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
sarah11918: I always *prefer* not to have to link externally to code. Let's leave the example in.

But, given the explanation of the example, perhaps highlighting the `export function createExports(...)` line, as well as the `return handle(...)` makes sense for highlighting? These are the two thing you've mentioned explicitly.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,62 @@ export default defineConfig({
 });
 ```
 
+### `workerEntryPoint`
+<p>
+**Type:** `{path: string | URL, exports: string[]}`
```
```suggestion

**Type:** `{path: string | URL, exports: string[]}`
```
until we figure out definitively that this is no longer needed, let's keep consistent with everything else and add random extra line!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,62 @@ export default defineConfig({
...
+**Default:** `{ path: '@astrojs/cloudflare/entrypoints/server.js', exports: ['default'] }`
+<Since v="12.6.0" />
+</p>
+
+The `workerEntryPoint.path` option allows you to specify the path to a custom entry file. By default, the adapter uses a generic entry file, which only supports the `fetch` handler, but you can change it to support other [Cloudflare invocation handlers](https://developers.cloudflare.com/workers/observability/logs/workers-logs/#invocation-logs). This is useful if you want to use features that require other handlers (e.g. Durable Objects, Cloudflare Queues, Scheduled Invocations). The `workerEntryPoint.exports` option allows you to specify the exports from a custom entry file. By default, the adapter uses `["default"]`. If you have defined named exports in your custom entry file, you can specify them here. This is useful if you want to use (e.g. Durable Objects, RPC)
```
sarah11918: @alexanderniebuhr I know there's a lot going on here so hang in there...

From looking at the code for this feature, I think this subtle change in framing the feature is correct: "These items configure properties of the `workerEntryPoint`."

With configuration items, we don't typically talk about this being "something to customize" because that's what configuring *is for*. These are simply *properties that can be configured*. Whenever something has a default value, that means it exists whether or not you want to customize. (And, as I've done in my suggestion following, needs its own API entry as a heading!)

e.g. If we say: "Here's a feature to customize your entry file.." that makes it sound like it only exists or has a value *when you want to customize*. But, these have default values and these properties are used no matter what, so everyone needs to know what these are/mean! It's more like we are exposing the option to configure something that previously was hardcoded/unconfigurable, and would now say something like, "Here's the property for specifying your entry file. If not specified, it defaults to..."

So, with all that in mind (these are simply now config options that exist; **every** new part of the API needs its own proper documentation), I see this shaping up more like the structure below (obviously, I'll need you to help me make it accurate!):

``````suggestion

A configuration object to specify the [workerEntryPoint](https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/rpc/) for your Cloudflare Worker when you use the `astro build` command.

It allows you to optionally specify both a custom file `path` and `exports`:

```js title="astro.config.mjs"
import cloudflare from '@astrojs/cloudflare';

export default defineConfig({
	adapter: cloudflare({
		workerEntryPoint: {
			path: 'src/worker.ts',
			exports: ['default','MyDurableObject']
		}
	}),
});
```

### `workerEntryPoint.path`

<p>

**Type:** `string`<br />
**Default:** `@astrojs/cloudflare/entrypoints/server.js`
<Since v="12.6.0" pkg="@astrojs/cloudflare" />
</p>

The path to the entry file. This should be a relative path from the root of your Astro project.

By default, the adapter uses a generic entry file, which only supports the `fetch` handler.

To support other [Cloudflare invocation handlers](https://developers.cloudflare.com/workers/observability/logs/workers-logs/#invocation-logs), you can create a custom file to use as the entry point. This is useful if you want to use features that require other handlers (e.g. Durable Objects, Cloudflare Queues, Scheduled Invocations).

#### `workerEntryPoint.exports` 

<p>

**Type:** `[]`<br />
**Default:** `['default']`
<Since v="12.6.0" pkg="@astrojs/cloudflare" />
</p>

The exports to use for the entry file.

Provide any additional defined named exports of your entry file (e.g. `DurableObjects`) alongside your file's default exports. If not provided, only default exports will be included.
``````
^ I will note that I found `exports`'s behaviour a little difficult to explain, so that *might* be a code smell.

Will you **ever** not want default exports included here? The weird thing (about explaining how this works) is that if you don't specify, you get default. If you *do* specify, then you have to manually add back in default, or (I'm assuming) you *don't* get the default exports? 

I'm not familiar with the feature, or whether this makes sense/is possible, but it *seems* like a more user-friendly API would be to ALWAYS export default, no matter what, but this new option is instead e.g. `namedExports`  (defaulting to `[]`) and allows you to specify any *additional* named exports. That way, you don't have to remember to add `default` back in. (Assuming of course you'd always want the default exports of these files). It also means I don't have to say "by default you get default, but if you touch this at all, you have to add back in default!" (which is not a great story to write if I don't have to!)

File: src/content/docs/en/guides/integrations-guide/cloudflare.mdx [hunk 5]
```diff
@@ -248,6 +248,62 @@ export default defineConfig({
...
+
+export default defineConfig({
+	adapter: cloudflare({
+		workerEntryPoint: {
+			path: 'src/worker.ts',
+			exports: ['default','MyDurableObject']
+		}
+	}),
+});
+```
```
```suggestion
#### Creating a custom Cloudflare Worker entry file

The custom entry file must export the `createExports` function, with a `default` export including all the handlers you need.

The following example entry file exports all the necessary handlers required for a custom `console.log`:
```
Similarly, this now is *specifically* about the entry file itself, not the other properties. It lets the example have its own space/breathing room, and now I think it feels less of a concern to have it here!

sarah11918 (commented):
Great work on this, @alexanderniebuhr ! I really appreciate the attention you give to your docs, and you provide super helpful content!

See my (big) suggestion below that I think simply restructures/formats this to support your helpful content even more by making it scannable, look-up-able, and linkable. (And of course, needs your help to make sure it's all accurate and comprehensive!) 🙌

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
alexanderniebuhr: The suggestion of `namedExports` is great. I'll add it update the core and docs PR :)

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
...
+<p>
+
+**Type:** `[]`<br />
+**Default:** `['default']`
+<Since v="12.6.0" pkg="@astrojs/cloudflare" />
+</p>
+
+The exports to use for the entry file.
+
+Provide any additional defined named exports of your entry file (e.g. `DurableObjects`) alongside your file's default exports. If not provided, only default exports will be included.
```
```suggestion
#### `workerEntryPoint.namedExports` 

<p>

**Type:** `[]`<br />
**Default:** `['default']`
<Since v="12.6.0" pkg="@astrojs/cloudflare" />
</p>

An array of named exports to use for the entry file.

Provide any additional defined named exports of your [custom entry file](#creating-a-custom-cloudflare-worker-entry-file) (e.g. `DurableObjects`). If not provided, only default exports will be included.
```
Just a placeholder here in case this ends up being updated so that you never need to include "default" yourself!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
...
+
+```js title="astro.config.mjs"
+import cloudflare from '@astrojs/cloudflare';
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+	adapter: cloudflare({
+		workerEntryPoint: {
+			path: 'src/worker.ts',
+			exports: ['default','MyDurableObject']
```
``````suggestion
It allows you to optionally specify both a custom file `path` and `namedExports`:

```js title="astro.config.mjs"
import cloudflare from '@astrojs/cloudflare';

export default defineConfig({
	adapter: cloudflare({
		workerEntryPoint: {
			path: 'src/worker.ts',
			namedExports: ['MyDurableObject']
``````
Again, placeholder in case `exports` is changed to `namedExports`!

File: src/content/docs/en/guides/integrations-guide/cloudflare.mdx [hunk 8]
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
 });
 ```
 
+### `workerEntryPoint`
+<p>
+
+**Type:** `{path: string | URL, exports: string[]}`
+**Default:** `{ path: '@astrojs/cloudflare/entrypoints/server.js', exports: ['default'] }`
```
```suggestion
**Default:** `{ path: '@astrojs/cloudflare/entrypoints/server.js', namedExports: [] }`
```
@alexanderniebuhr this is what I wanted to check on! Is this just not updated for the new decision, or is `exports` here an internal (non-directly exposed) property that combines default and named exports? It seems weird to show that here instead of `namedExports`, which is what *is* exposed and configurable to the user, hiding the notion of default exports at all.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
...
+<Since v="12.6.0" pkg="@astrojs/cloudflare" />
+</p>
+
+An array of named exports to use for the entry file.
+
+Provide any additional defined named exports of your [custom entry file](#creating-a-custom-cloudflare-worker-entry-file) (e.g. `DurableObjects`). If not provided, only default exports will be included.
+
+#### Creating a custom Cloudflare Worker entry file
+
+The custom entry file must export the `createExports` function, with a `default` export including all the handlers you need.
```
```suggestion
The custom entry file must export the `createExports()` function, with a `default` export including all the handlers you need.
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
alexanderniebuhr: Yeah internally we merge `['default']` with `namedExports`, but the user can only configure `namedExports`, so I think it makes sense to only show what is exposed.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
 });
 ```
 
+### `workerEntryPoint`
+<p>
+
+**Type:** `{path: string | URL, exports: string[]}`
```
alexanderniebuhr:
```suggestion
**Type:** `{path: string | URL, namedExports: string[]}`
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
...
+<Since v="12.6.0" pkg="@astrojs/cloudflare" />
+</p>
+
+An array of named exports to use for the entry file.
+
+Provide any additional defined named exports of your [custom entry file](#creating-a-custom-cloudflare-worker-entry-file) (e.g. `DurableObjects`). If not provided, only default exports will be included.
+
+#### Creating a custom Cloudflare Worker entry file
+
+The custom entry file must export the `createExports()` function, with a `default` export including all the handlers you need.
```
yanthomasdev: comma nit
```suggestion
The custom entry file must export the `createExports()` function with a `default` export including all the handlers you need.
```

yanthomasdev (commented):
Just a small nit. As a bit of a follow-up, I feel like what people will look for is "how to use queues/durable objects/etc with Astro Cloudflare, so we might want to add a heading in the future that points to this part, but surely not the scope of this PR.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
...
+	const app = new App(manifest);
+	return {
+		default: {
+			async fetch(request, env, ctx) {
+				await env.MY_QUEUE.send("log");
+				return handle(manifest, app, request, env, ctx);
+			},
+			async queue(batch, _env) {
+				let messages = JSON.stringify(batch.messages);
+    		console.log(`consumed from our queue: ${messages}`);
```
ascorbic, line 335:
```suggestion
				console.log(`consumed from our queue: ${messages}`);
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
...
+			}
```
ascorbic:
```suggestion
			async queue(batch, _env) {
				let messages = JSON.stringify(batch.messages);
				console.log(`consumed from our queue: ${messages}`);
			}
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -248,6 +248,98 @@ export default defineConfig({
...
+
+An array of named exports to use for the entry file.
+
+Provide any additional defined named exports of your [custom entry file](#creating-a-custom-cloudflare-worker-entry-file) (e.g. `DurableObjects`). If not provided, only default exports will be included.
+
+#### Creating a custom Cloudflare Worker entry file
+
+The custom entry file must export the `createExports()` function, with a `default` export including all the handlers you need.
+
+The following example entry file exports all the necessary handlers required for a custom `console.log`:
```
ascorbic:
```suggestion
The following example entry file registers a Durable Object and a queue handler:
```

PR #11753, opened 2025-05-24
Clarify GitHub Pages `base` config and `withastro/action` usage

### LINKED ISSUE #10868: (en) `github.mdx`: The description of the text is discrepancy.
### 📚 Subject area/topic

guides/deploy

https://docs.astro.build/en/guides/deploy/github/

The leading slash in the value of the base parameter is being removed by #7481.

However, the text explains that the base parameter value must be prefixed with slash.

> The value for `base` should be your repository’s name starting with a forward slash, for example `/my-blog`. This is so that Astro understands your website’s root is `/my-repo`, rather than the default `/`.

Incidentally, when I tried deploying with GitHub Pages, I was able to deploy successfully with or without the slash. What exactly was the problem with #7418? And please think of a way to resolve this discrepancy.

_No response_

casungo:
## Description
This PR updates the English documentation for deploying Astro sites to GitHub Pages (`guides/deploy/github.mdx`) to address several points of confusion and provide clearer, more accurate instructions.

These changes aim to significantly reduce confusion for users deploying to GitHub Pages subpaths and improve the reliability of their setups by promoting best practices for path handling.

## Related issues & labels
- Closes #10868


sarah11918:
Hi @casungo ! Sorry this PR hasn't had much activity on it! I was hoping someone would be available to help go through this!

Firstly, there are conflicts to resolve. Would you mind handling those?

Secondly, the linked issue mentions that this did not fix someone's problem, but linked to a blog post showing adding a `nojekyll` file did, and I know that has been a solution to people in the past. I don't see this updated for that extra information in the issue, so I think it would be important to add that here.

Since I haven't had a chance to carefully go through this, I do want to first **only update what is incorrect** so that these instructions lead to success for someone. (Vs changing explanations / descriptions because they might be "better"). It is much easier to review a PR that only includes corrections for things that are wrong than one that *also* tries to improve it in various ways. And, since we're struggling to keep up with PRs right now, we need all the help we can get! 😅 

So, can we make sure this PR is *only* about correcting wrong or missing info, so that we can this reviewed and merged more quickly? For example, this PR changes statements that are correct like:

>  You can set up a custom domain by adding the following `./public/CNAME` file to your project:" 

to maybe "better" instructions like:

> You can set up a custom domain by adding a `CNAME` file to your project's `public/` directory:" 

but these don't fix a problem with incorrect statements at this time. I don't disagree that your statement is better! But, there are many of these "upgrades" that don't fix a problem, and that makes it harder to spot the things that actually *are* incorrect, and make this PR take longer to review. 😅  Right now, we really need to optimize for correcting inaccuracies!

Would you be willing to rework this PR in the above ways? (Resolve conflicts, figure out how to helpfully add the `nojekyll` information, and only focus on correcting things that are wrong/missing statements?) That would be super helpful, and we could get moving on this! (And, a later PR with "improvements" would always be welcome, where we can concentrate only on language, knowing that the basic steps all work!)

sarah11918:
(Just fixed the merge conflict)

sarah11918:
Hi casungo! I'm going to close this PR for now because there hasn't been any activity here in a while. And, like I mentioned before, I don't believe it addresses the one thing that solves the original issue, and this has just a little too much going on here that is focused on improvements vs the task of updating inaccurate information!

There is a discord thread where people are discussing how to handle the `.nojekyll` file, and I think that's going to end up being the fix this page needs first. Then, if there are improvements to be made, you're welcome to revisit the page! :raised_hands:

PR #11758, opened 2025-05-24
docs: add astro-loader-i18n to community libraries

openscript:
#### Description (required)

`astro-loader-i18n` is a **content loader** for internationalized content in [Astro](https://astro.build). It builds on top of Astro’s [`glob()` loader](https://docs.astro.build/en/reference/content-loader-reference/#glob-loader) and helps manage translations by detecting locales, mapping content, and enriching `getStaticPaths`.

It helps to translate routes, thus I want to add it to the community libraries. So far it's the only solution that uses a custom loader approach, compared to the other listed libraries.


sarah11918 (approved):
Looks great! Thank you for adding! 🚀

yanthomasdev (approved):
LGTM! Thanks @openscript 🙌

PR #11763, opened 2025-05-24
fix(migrate-to-astro): replace `getEntryBySlug` with `getEntry`

ArmandPhilippot:
#### Description (required)

We had migration guides still referring to `getEntryBySlug` instead of `getEntry`, so I replaced the two references.

- Suggested label: `typo/link/grammar - quick fix!` (not really a typo but certainly a quick fix!)


yanthomasdev (approved):
LGTM! Nice catch!

PR #11760, opened 2025-05-24
fix: update `astro-config.mdx`

tkskto:
#### Description (required)

fixed incorrect code example.

#### Related issues & labels (optional)

- Closes #11713
- Suggested label: `code snippet update`

This is my first-time contribute 🙌

Discord: tkskto

PR #11772, opened 2025-05-25
i18n(fr): Update `reference/modules/astro-config` from #11760

thomasbnt:
#### Description (required)
i18n(fr): Update `reference/modules/astro-config` from #11760

- Closes #
- Suggested label: i18n


ArmandPhilippot (approved):
LGTM!

PR #11776, opened 2025-05-26
chore: clarifying relation between Astro sessions & cookies

miguderp:
#### Description

The wording around Astro sessions can be a bit misleading, making you believe they work with 0 cookies.
This PR amends that, clarifying that a minimal cookie is created just to hold the session ID and that the actual data remains in the provider instead of the user’s browser.


miguderp:
Turns out the wording of cookie-less was mostly about the fact that [`load()`](https://docs.astro.build/en/reference/api-reference/#load) was added to bypass the cookie need whenever we handle the session ID ourselves

PR #11781, opened 2025-05-27
i18n(fr): improve the wording in `guides/ecommerce.mdx`

ArmandPhilippot:
#### Description (required)

Updates the French translations in `guides/ecommerce.mdx` to improve the wording (related to #11593):
* replace `en` with `Anglais` in community links
* add missing space after colon
* replace `construction` with `création` where more appropriate
* reword some parts to make the reading easier
* use French quotes instead of double quotes

In addition, and this is why I preferred to make a separate PR for this file:
* `overlay` was sometimes translated, omitted or untranslated. We had not covered this word in the i18n guide so, to be consistent, I translated it everywhere (and added where missing). I used `de superposition`/`superposé` (often with `fenêtre` in front) while the [Grand dictionnaire terminologique](https://vitrinelinguistique.oqlf.gouv.qc.ca/fiche-gdt/fiche/8368197/recouvrement) for example recommends `recouvrement` because doing a quick search in Google it seems using `superposition` is more instinctive.
* `checkout` is also a bit tricky but I think `processus de paiement` or `tunnel de commande` (marketing term) is more appropriate in most cases. While `caisse` or `commande` doesn't seem right.
* `inline`... we tend to use `en ligne` (literal translation) everywhere but maybe this is something we should cover in the i18n guide. Like `placeholder`, I don't think there is only one valid translation but I think `en ligne` is wrong almost all the time. In particular in this context. By consulting the docs links, I'd say in this context `inline` means directly integrated into the website instead of an overlay modal. That's why I used `intégré au site` most of the times.

So do not hesitate to comment on this if you have better suggestions!

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/guides/ecommerce.mdx
+++ b/src/content/docs/fr/guides/ecommerce.mdx
@@ -9,22 +9,22 @@ import ReadMore from '~/components/ReadMore.astro'
 
 Avec Astro, vous pouvez créer plusieurs options d'e-commerce, des liens de paiement aux pages de paiement hébergées, en passant par la création d'une boutique complète à l'aide d'une API de service de paiement.
 
-## Traitement des paiements
+## Traitement des paiements par superposition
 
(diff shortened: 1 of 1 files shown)


thomasbnt (approved):
Lgtm!

PR #11795, opened 2025-05-28
i18n(fr): add new words to the French `i18n-guide`

ArmandPhilippot:
#### Description (required)

I haven't finished rereading all the translations yet, but before I forgot, here are a few additional words that might be useful to add to the glossary. I'll put a comment where I think this could be discussed to avoid a too long description here.

I also updated some `Conseils stylistiques` and fixed an oversight (`île` > `îlot`) from the last time.

- Suggested label: i18n


i18n-guides/français.md
```diff
@@ -40,6 +40,7 @@ Il existe d'autres mots ne possédant pas de traduction officielle ou dont le te
...
 | hook | Désigne un point d'ancrage permettant à l'utilisateur d'exécuter du code à un endroit précis du code source. | « crochet » n'apporte pas suffisamment de sens et il est rarement traduit ailleurs. Précédente discussion : https://github.com/withastro/docs/pull/11593#discussion_r2074976678 |
 | headless | Désigne une architecture où le front-end et le back-end sont proposés séparément (p ex. un CMS pour gérer les contenus et une application Astro pour l'afficher). | Il n'existe pas de traduction officielle et nous pensons que « sans-tête » n'est pas suffisamment porteur de sens. Précédente discussion : https://github.com/withastro/docs/pull/11593#discussion_r2074967636 |
+| glob / globbing | Désigne une fonction permettant de faire correspondre des motifs à des noms de répertoire ou de fichier. Par extension, il désigne également la syntaxe utilisée. | Nous n'avons pas de traduction officielle pour ce terme et utiliser « global » serait une erreur de traduction. |
```
ArmandPhilippot, line 43: This is currently translated `global` in a lot of places... But, this is like a proper noun (initially the name of a function), so we shouldn't translate it. Using `global`/`globaux` sounds like a translation error to me.

i18n-guides/français.md
```diff
@@ -60,23 +61,31 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
...
 | import | importation | |
+| inline | au sein de / dans / intégré à | La traduction dépend fortement du contexte, les propositions précédentes ne sont que des exemples possibles. Il est préférable de trouver une alternative à « en ligne » dans la majorité des cas. |
 | issue | problème ou, en référence à la fonctionnalité de Github, ticket | [MDN](https://developer.mozilla.org/fr/docs/MDN/Community/Issues) utilise également « ticket » pour décrire la fonctionnalité de Github |
 | layout | mise en page | |
+| live | en ligne / opérationnel | « live » est généralement utilisé pour différencier développement/production donc « en direct » est rarement la bonne traduction. Il peut être préférable d'omettre ce mot si le contexte est clair. |
 | on-demand rendering | rendu à la demande | |
+| overlay | par superposition / (fenêtre) superposée | |
 | package | paquet | |
+| pattern | modèle / (en référence aux expressions régulières et à glob) motif | D'autres traductions peuvent être acceptables en fonction du contexte. Par exemple, « formule » semblait plus appropriée [dans le tutoriel](https://docs.astro.build/fr/tutorial/2-pages/3/#analyser-la-formule). |
```
ArmandPhilippot, line 88: When referring to regular expressions, it seems we should [use `motif`](https://fr.wikipedia.org/wiki/Expression_r%C3%A9guli%C3%A8re) instead of `modèle`. So, `glob patterns` should be `motifs glob` for example. It could help vary the wording when we have both `template` and `pattern` in the same page (currently we use `modèle` for both).

I also added a note with a link in explanations because sometimes it's fine/more appropriate to find another word but it seemed occasional so I didn't included the word in the translation recommendation.

i18n-guides/français.md
```diff
@@ -60,23 +61,31 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
 
 | Anglais | Français | Explications/Référence (facultatif) |
 |---------|----------|-------------------------------------|
+| assets | ressources | « actif » vient du [monde de la finance](https://www.culture.fr/franceterme/Resultats-de-recherche?q=asset&domaine=0). Il est préférable d'utiliser « [ressource](https://fr.wiktionary.org/wiki/asset) ». |
```
ArmandPhilippot, line 64: Currently we have mostly `ressources` but we have some `actifs`. As I pointed out in https://github.com/withastro/starlight/pull/3221, `actif` comes from the finance field and it is put forward due to crypto... When there is no connection with finance, we used to use `ressource` before. But maybe I'm just old. 👴🏽

i18n-guides/français.md
```diff
@@ -60,23 +61,31 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
...
 | bundle | regroupement | |
 | changelog | journal des modifications | |
 | (the) CLI / (the) command line interface | (la) CLI / (l')interface en ligne de commande | |
 | client-side | côté client | |
 | code fences (`---`) | délimitateur de code | |
+| docs | documentation | |
 | endpoint | point de terminaison | |
 | export | exportation | |
 | feedback | retour / réaction / commentaire | |
+| flag | option | |
```
ArmandPhilippot, line 76: Currently we have `option`, `indicateur` and `drapeau`. The last one is, IMO, a translation error. `indicateur` is not necessarily wrong but for CLI flags and configuration flags we usually use `option` ([`options en ligne de commande`](https://fr.wikipedia.org/wiki/Interface_en_ligne_de_commande#Option) / `options de configuration`) or eventually `paramètre`. I think in Starlight we only have `option` but I could be wrong.

i18n-guides/français.md
```diff
@@ -60,23 +61,31 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
...
 | footer | pied de page | |
 | header | en-tête | |
 | heading | titre ou en-tête | Nous n'avons pas de consensus (« titre » dans Astro Docs et « en-tête » dans Starlight). Précédente discussion : https://github.com/withastro/starlight/pull/2884#discussion_r1957379942 |
 | import | importation | |
+| inline | au sein de / dans / intégré à | La traduction dépend fortement du contexte, les propositions précédentes ne sont que des exemples possibles. Il est préférable de trouver une alternative à « en ligne » dans la majorité des cas. |
```
ArmandPhilippot, line 81: As pointed out in https://github.com/withastro/docs/pull/11781 (because it was confusing - `paiement en ligne` - as well as sounding incorrect), translating `inline` with `en ligne` seems wrong almost everywhere and even more in some places. I don't have a perfect translation unfortunately. I think this is something like `placeholder`. So I still included it in the list with some suggestions but if you have other ideas I'm all ears!

I think I saw a PR in Starlight from HiDeoo (?) where he was suggesting something else for `inline` because `en ligne` wasn't appropriate. So maybe he will have more ideas. (cc @HiDeoo )

i18n-guides/français.md
```diff
@@ -60,23 +61,31 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
...
 | issue | problème ou, en référence à la fonctionnalité de Github, ticket | [MDN](https://developer.mozilla.org/fr/docs/MDN/Community/Issues) utilise également « ticket » pour décrire la fonctionnalité de Github |
 | layout | mise en page | |
+| live | en ligne / opérationnel | « live » est généralement utilisé pour différencier développement/production donc « en direct » est rarement la bonne traduction. Il peut être préférable d'omettre ce mot si le contexte est clair. |
```
ArmandPhilippot, line 84: We have a lot of `en direct` but it doesn't sound right... Generally, `live` is used to make the distinction between dev/prod modes. I don't know if this is an "official" translation but for `live website` we used to say `site en ligne` so I used that in some places. In other places, I'm not sure what to use. Adding `opérationnel` or `en mode production` seems heavy and I wonder if it is not best to just omit the word when the context seems clear enough.

i18n-guides/français.md
```diff
@@ -96,12 +105,17 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
 | template | modèle | |
 | to build | Suivant le contexte : créer / construire / compiler | |
 | to bundle | regrouper | |
+| to commit | valider | Il s'agit de la traduction utilisée dans l'interface française de VS Code. |
 | to export | exporter | |
 | to fetch | récupérer | |
 | to import | importer | |
+| to push | transférer / envoyer | |
```
ArmandPhilippot, line 112: I was checking [Astro DB](https://docs.astro.build/fr/guides/astro-db/#pousser-les-sch%C3%A9mas-de-table) and we have some `Pousser` for `Push`. Maybe it's just me but it sounds odd to me. Shouldn't we use another word? I think `transférer` or `envoyer` (depending on the context) could work.

i18n-guides/français.md
```diff
@@ -96,12 +105,17 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
...
 | to render | afficher / générer / effectuer un rendu / restituer | |
 | to slot | inclure / injecter / insérer | |
+| to store | stocker / enregistrer / sauvegarder / conserver | |
```
ArmandPhilippot, line 115: `Stocker` is not necessarily wrong but we have a lot of `stocker` accross the docs and it feels a bit repetitive/not always the best word... I feel like we could vary the wording... So I only added that to suggest some other words...

i18n-guides/français.md
HiDeoo: I guess "glob" is usually a "function" while globbing is the process itself?

Maybe the sentence could also mention "caractères génériques", e.g. "Désigne une fonction ou processus permettant de faire correspondre des motifs à des noms de répertoire ou de fichier en utilisant des caractères géneriques (aussi connu sous le nom de « [appariement de formes](https://www.culture.fr/franceterme/terme/INFO82) »)"?

i18n-guides/français.md
```diff
@@ -60,23 +61,31 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
 
 | Anglais | Français | Explications/Référence (facultatif) |
 |---------|----------|-------------------------------------|
+| assets | ressources | « actif » vient du [monde de la finance](https://www.culture.fr/franceterme/Resultats-de-recherche?q=asset&domaine=0). Il est préférable d'utiliser « [ressource](https://fr.wiktionary.org/wiki/asset) ». |
+| breaking (changes) | (changements/modifications) avec rupture de compatibilité / cassants / de rupture | « changements majeurs » est acceptable pour éviter les répétitions tant qu'on garde la notion que quelque chose peut « casser » dans le titre/texte.|
```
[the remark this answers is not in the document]
HiDeoo: I've often used "Changements non rétrocompatibles" as it's the one used in the [French version of the semver spec](https://semver.org/lang/fr/).

1. le numéro de version MAJEUR quand il y a des changements non rétrocompatibles,
2. le numéro de version MINEUR quand il y a des ajouts de fonctionnalités rétrocompatibles,
3. le numéro de version de CORRECTIF quand il y a des corrections d’anomalies rétrocompatibles.

i18n-guides/français.md
HiDeoo: > I think in Starlight we only have option but I could be wrong.

I think so, it's based on the [Wikipedia usage](https://fr.wikipedia.org/wiki/Interface_en_ligne_de_commande#Option).

i18n-guides/français.md
HiDeoo: You may be thinking of this [section](https://starlight.astro.build/guides/site-search/#docsearch-configuration):

> The Starlight DocSearch plugin supports customizing the DocSearch component with the following inline options:

which I translated [here](https://github.com/withastro/starlight/pull/2836/files/ac58639256da372e462a0ae5f563dd5554d001e4#diff-46715b88964161097ab5d7a9a4dd1c130e56ad913615ca40916b837a5f1645fa):

> Le module d’extension DocSearch de Starlight supporte la personnalisation du composant DocSearch avec les options supplémentaires suivantes spécifiées dans la configuration du module d’extension :

I also linked to https://french.stackexchange.com/questions/54639/what-is-the-translation-of-inline-in-french

i18n-guides/français.md
HiDeoo: "Envoyer" or "Publier" are probably the one I've encountered the most so far.

i18n-guides/français.md
```diff
@@ -96,12 +105,17 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
...
+| to commit | valider | Il s'agit de la traduction utilisée dans l'interface française de VS Code. |
 | to export | exporter | |
 | to fetch | récupérer | |
 | to import | importer | |
+| to push | transférer / envoyer | |
 | to render | afficher / générer / effectuer un rendu / restituer | |
 | to slot | inclure / injecter / insérer | |
+| to store | stocker / enregistrer / sauvegarder / conserver | |
 | to style | mettre en forme / appliquer des styles | |
+| to support | prendre en charge | |
```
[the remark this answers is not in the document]
HiDeoo: Isn't this one [specific](https://www.academie-francaise.fr/supporter) to "cheering", e.g. in sport or politics?

HiDeoo (commented):
Amazing additions, left a few comments/questions but overall looks great.

i18n-guides/français.md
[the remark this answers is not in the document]
HiDeoo: Oh, my bad, I didn't spot the second anglicism in your link way down the page, and the one I had only talked about the first one.

Then, I guess prendre en charge/être compatible are the ones to use based on the context (I'm pretty sure I'll never remember this one 😅 )

i18n-guides/français.md
ArmandPhilippot: I like `publier`! I still think `transférer` could be valid in some places so maybe we could keep it but as last suggestion:
```suggestion
| to push | publier / envoyer / transférer | |
```
(It seems I forgot to post this before T&D 😅 )

i18n-guides/français.md
ArmandPhilippot: Let's update that:
```suggestion
| glob / globbing | Désigne une fonction ou processus permettant de faire correspondre des motifs à des noms de répertoire ou de fichier en utilisant des caractères géneriques (aussi connu sous le nom de « [appariement de formes](https://www.culture.fr/franceterme/terme/INFO82) ») | Nous n'avons pas de traduction officielle pour ce terme et utiliser « global » serait une erreur de traduction. |
```

i18n-guides/français.md
ArmandPhilippot: So, based on https://github.com/withastro/docs/pull/11795#discussion_r2113783815 ... maybe we should use `forme` instead of `motif`. Or should we accept the two? It seems the Wikipedia reference for `motif` is based on lecture notes. So if `motif` is teached, I guess it still right?

```suggestion
| pattern | modèle / (en référence aux expressions régulières et à glob) forme ou motif | D'autres traductions peuvent être acceptables en fonction du contexte. Par exemple, « formule » semblait plus appropriée [dans le tutoriel](https://docs.astro.build/fr/tutorial/2-pages/3/#analyser-la-formule). |
```

HiDeoo: Maybe "formules" and "motifs" like Wikipedia?

image

ArmandPhilippot: Oh I clicked on the footnote link but didn't see that. 😆 It was well hidden (14 occurences for `motif` compared to 1 for `formule` on the `Expression régulière` page 😅 ).

I think it sounds better than `forme`, I was just trying to be consistent with `appariement de formes`. So, I'm okay to use that instead!

```suggestion
| pattern | modèle / (en référence aux expressions régulières et à glob) motif ou formule | D'autres traductions peuvent être acceptables en fonction du contexte. Par exemple, « formule » semblait plus appropriée [dans le tutoriel](https://docs.astro.build/fr/tutorial/2-pages/3/#analyser-la-formule). |
```

i18n-guides/français.md
```diff
@@ -60,23 +61,31 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
...
 | Anglais | Français | Explications/Référence (facultatif) |
 |---------|----------|-------------------------------------|
+| assets | ressources | « actif » vient du [monde de la finance](https://www.culture.fr/franceterme/Resultats-de-recherche?q=asset&domaine=0). Il est préférable d'utiliser « [ressource](https://fr.wiktionary.org/wiki/asset) ». |
+| breaking changes | changements non rétrocompatibles / changements avec rupture de compatibilité | « changements majeurs » est acceptable pour éviter les répétitions tant qu'on garde la notion que quelque chose peut « casser » dans le titre/texte. Précédente discussion : https://github.com/withastro/docs/pull/11795#discussion_r2112817635 |
 | build | Suivant le contexte : création / construction / compilation. En référence à `astro build`, on privilégiera le terme compilation.  | |
 | bundle | regroupement | |
 | changelog | journal des modifications | |
 | (the) CLI / (the) command line interface | (la) CLI / (l')interface en ligne de commande | |
 | client-side | côté client | |
 | code fences (`---`) | délimitateur de code | |
```
ArmandPhilippot, line 71: It just came back to me: While updating the translations, I had a doubt between `délimitateur` and... `délimiteur`. It seems both have a similar meaning but `délimiteur` brings more results. So should we replace `délimitateur` with `délimiteur`, add `delimiteur` as acceptable translation or just stick with `délimitateur`?

Both seems right:
* [`délimitateur`](https://www.littre.org/definition/d%C3%A9limitateur): "Celui qui limite."
* [`délimiteur`](https://www.larousse.fr/dictionnaires/francais/d%C3%A9limiteur/23105): "Symbole utilisé pour séparer des suites adjacentes de bits ou de caractères au sein d'un ensemble de données."

I'd say it's better to have a single term to be consistent across the docs. But what do you think of `délimitateur` vs `délimiteur`? [Last time](https://github.com/withastro/docs/pull/11593#discussion_r2074931355) we hadn't thought of the latter.

ArmandPhilippot: It's only an interpretation so take it with a pinch of salt but Littré (the other link) is more a literary dictionary while Larousse is more "mainstream". So, perhaps `délimitateur` became obsolete at some point and a dictionary like Littré still considers it as valid while Larousse uses the new form...

So, yeah, it was also more or less my thinking. I don't mind `délimitateur` (after all it was my first instinct when you have suggested something around `délimitation`), but if its use is less common, maybe we should consider something else to avoid an eventual "burden" to the reader (ie. `délimitateur`, hmm shouldn't this be `délimiteur`!?).

PS: Thanks Google, I would have never guess that `avec des pincettes` would be `a pinch of salt`. 😅

i18n-guides/français.md
```diff
@@ -60,23 +61,31 @@ Certains mots ont un équivalent français qui devrait être utilisé uniformém
...
+| docs | documentation | |
```
ArmandPhilippot, line 72:
```suggestion
| deprecated | obsolète | |
| docs | documentation | |
```

This is a temporary suggestion. Now, I have updated again the `upgrade-to` guides, I remember another inconsistent translation across the docs: `deprecated`.

Currently we have `obsolète` and `déprécié` in Astro Docs. I think Starlight only use `déprécié` for `deprecated` and `obsolète` for `outdated`. I like the distinction. I don't think we have any `outdated` in Astro Docs but it still nice to consider this.

[FranceTerme](https://www.culture.fr/franceterme/Resultats-de-recherche?q=deprecate&domaine=0) doesn't suggest anything. [Wiktionary](https://fr.wiktionary.org/wiki/deprecated) suggests `obsolète` or `périmé` in a computer science context. The [Grand dictionnaire terminologique of Quebec](https://vitrinelinguistique.oqlf.gouv.qc.ca/fiche-gdt/fiche/26532619/deconseille) recommends `déconseillé` but also accept (in `Termes privilégiés`):
* `désapprouvé`
* `déprécié`

While the recommendation for `outdated` would be [périmé](https://vitrinelinguistique.oqlf.gouv.qc.ca/resultats-de-recherche?tx_solr%5Bq%5D=outdated&tx_solr%5Bfilter%5D%5B0%5D=type_stringM%3Abdl&tx_solr%5Bfilter%5D%5B1%5D=type_stringM%3Agdt&tx_solr%5Bfilter%5D%5B2%5D=terme%3A1) in a broader context.

In the [French Wikipedia page](https://fr.wikipedia.org/wiki/Obsolescence_(informatique)), `deprecation` is translated as `obsolescence` with a note:

> Une fonctionnalité obsolète est souvent dite deprecated voire « dépréciée » par [anglicisme](https://fr.wikipedia.org/wiki/Anglicisme).

(there is no footnotes regarding this statement... but I would tend to agree because we don't have an [equivalent meaning](https://www.cnrtl.fr/definition/d%C3%A9pr%C3%A9ci%C3%A9) but we are also late in finding French terms in this domain...)

While, the [English version of the Wikipedia page](https://en.wikipedia.org/wiki/Deprecation) specifies that:

> Not to be confused with [Depreciation](https://en.wikipedia.org/wiki/Depreciation) or [Obsolescence](https://en.wikipedia.org/wiki/Obsolescence).

And obviously, `obsolescence` is translated with the same word in French... So I'd say it is important to keep the distinction but `déprécié` would be an anglicism. If we avoid anglicism for plugins, support, etc. maybe we should consider something else. So, I think the Quebec recommendation makes sense here: `déconseillé` for `deprecated`. Unless you have other ideas to make the distinction clear while avoiding anglicisms?

For example:
> Cette fonctionnalité est déconseillée et sera supprimée dans une prochaine version.

I think this could work. The meaning seems here.

What do you think of this? (I think this was my last memory lapse 😄 ... that's why I preferred not to wait rereading every pages before this PR...)

ArmandPhilippot: Sorry if I wasn't clear about the example!  I would not suggest using the whole sentence, it was rather an example of the word used in a common sentence. I was influenced by the fact that I had just updated the `upgrade-to` guides to update the translation of `breaking changes`. 😅 So, for example, we have sentences like:
> Some deprecated features may temporarily continue to function until they are completely removed.

But, yes, I understand your concern... I hadn't thought of that! Good call!

* I don't think we have the notion of obsolescence in the docs, but `obsolète` doesn't seem quite right.
* I don't like `périmé`, I think it sounds odd `cette fonction est périmée`... 👀 
* Here's some other possibilities but I don't think this is necessarily better:
  * `dévalorisé`/`dévalorisation` ([Larousse](https://www.larousse.fr/dictionnaires/francais/d%C3%A9valoriser/24929))
  * `désuet`/`désuétude` ([Larousse](https://www.larousse.fr/dictionnaires/francais/d%C3%A9suet/24717), [Wiktionary](https://fr.wiktionary.org/wiki/d%C3%A9suet))
  * `déclassé`/`déclassement` ([Larousse](https://www.larousse.fr/dictionnaires/francais/d%C3%A9classer/22231))
  
But, then, in the [Larousse](https://www.larousse.fr/dictionnaires/francais/d%C3%A9pr%C3%A9cier/23902) we also have `déprécier`/`dépréciation` (not in this context, but the meaning seems okay)... Only, Wikipédia seems to say this is an anglicism.

So, I'm fine with `déprécié`/`dépréciation`! The key is to avoid any typo: [déprécation](https://www.larousse.fr/dictionnaires/francais/d%C3%A9pr%C3%A9cation/23899). 😆

i18n-guides/français.md
ArmandPhilippot: I'm making a suggestion for this, but I'll wait a bit before applying it. It would be nice to have other opinions!
```suggestion
| code fences (`---`) | délimiteur de code | Précédente discussion : https://github.com/withastro/docs/pull/11795#discussion_r2114303905 |
```

thomasbnt: Hello, `délimiteur` seems better to read.

i18n-guides/français.md
thomasbnt: `options` 🆗

PR #11824, opened 2025-06-02
Updated endpoint page 

hkbertoson:
#### Description (required)

Updated the Endpoints page to reflect that endpoints must be written in uppercase. 

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/endpoints.mdx
+++ b/src/content/docs/en/guides/endpoints.mdx
@@ -102,6 +102,8 @@ But, unlike in `static` mode, when you enable on-demand rendering for a route, t
 
 Your routes will be rendered on demand by default in `server` mode. In `static` mode, you must opt out of prerendering for each custom endpoint with `export const prerender = false`.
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks @hkbertoson for jumping on this from a Discord issue! 💜 

I took a look at the page, and there *is* a dedicated section for HTTP methods. So, maybe any additional content should end up there, as being most related to that section?

(Think about if you wanted to be able to link someone to this in support. You can probably get closer by linking directly to "HTTP methods" than "API Endpoints")

As Armand mentioned on Discord, every example in docs shows upper case letters, and the external link to define HTTP methods *also* only shows uppercase letters. So, there's an argument that this concept is already covered. 😅 But, I think if you want to add something, it's fine.

 I would just be strategic about *where* you add this, and pick the most relevant location so that if you want to link someone to this fact, they don't have to read it as a "random fun fact" at the end of a section. In fact, I bet you could even work this directly *in* to an existing sentence in HTTP method, the text is so related! See what you can do there!

PR #11825, opened 2025-06-03
Document `suppress` in `supportedAstroFeatures`

yanthomasdev:
#### Description (required)

This adds documentation for `suppress: "all"` and `suppress: "default"` in the `supportedAstroFeatures` property of `setAdapter()`.

- Core PR: https://github.com/withastro/astro/pull/13842/

### DIFF
--- a/src/content/docs/en/reference/adapter-reference.mdx
+++ b/src/content/docs/en/reference/adapter-reference.mdx
@@ -71,6 +71,7 @@ export type AdapterSupportsKind = 'unsupported' | 'stable' | 'experimental' | 'd
 export type AdapterSupportWithMessage = {
 	support: Exclude<AdapterSupportsKind, 'stable'>;
 	message: string;
+  suppress?: 'default' | 'all';
 };
 
 export type AdapterSupport = AdapterSupportsKind | AdapterSupportWithMessage;
@@ -419,6 +420,56 @@ export default function createIntegration() {
             sharpImageService: {
               support: 'limited',
               message: 'This adapter has limited support for Sharp, certain features may not work as expected.'
+            }
+          } 
+        });
+      },
+    },
+  };
+}
+```
+
+If the default log message sent along with your custom message can cause confusion, you can use `suppress: "default"` to suppress the default message:
+
+```js title="my-adapter.mjs" ins={13}
+export default function createIntegration() {
+  return {
+    name: '@matthewp/my-adapter',
+    hooks: {
+      'astro:config:done': ({ setAdapter }) => {
+        setAdapter({
+          name: '@matthewp/my-adapter',
+          serverEntrypoint: '@matthewp/my-adapter/server.js',
+          supportedAstroFeatures: {
+            sharpImageService: {
+              support: 'limited',
+              message: 'This adapter has limited support for Sharp, certain features may not work as expected.',
+              suppress: 'default'
+            }
+          } 
+        });
+      },
+    },
+  };
+}
+```
+
+You can also use `suppress: "all"` to suppress all messages for a specific feature, useful if it doesn't impact users in a specific context:
+
+```js title="my-adapter.mjs" ins={13}
+export default function createIntegration() {
+  return {
+    name: '@matthewp/my-adapter',
+    hooks: {
+      'astro:config:done': ({ setAdapter }) => {
+        setAdapter({
+          name: '@matthewp/my-adapter',
+          serverEntrypoint: '@matthewp/my-adapter/server.js',
+          supportedAstroFeatures: {
+            sharpImageService: {
+              support: 'limited',
+              message: 'This adapter has limited support for Sharp, certain features may not work as expected.',
+              suppress: 'all'
             }
           } 
         });


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -71,6 +71,7 @@ export type AdapterSupportsKind = 'unsupported' | 'stable' | 'experimental' | 'd
 export type AdapterSupportWithMessage = {
 	support: Exclude<AdapterSupportsKind, 'stable'>;
 	message: string;
+  suppress?: 'default' | 'all';
```
ascorbic:
```suggestion
	message: string;
	suppress?: 'default' | 'all';
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -428,6 +429,56 @@ export default function createIntegration() {
 }
 ```
 
+If the default log message sent along with your custom message can cause confusion, you can use `suppress: "default"` to suppress the default message:
```
sarah11918: I'm thinking of a structure like this (tell me what you think)

Up around line 375 at the beginning of this section, we have:

> When using these properties, Astro will:
> - run specific validation; 
> - emit contextual to the logs;

(Aside: should that be "emit contextual INFORMATION to the logs" ??)

> These operations are run based on the features supported or not supported, their level of support, the [desired amount of logging](#suppress), and the user's own configuration.

Then here, we can actually have a proper heading for this option:

```suggestion
### `suppress`

<p>

  **Type:** `default | all`<br />
  <Since v="5.9.0" />
</p>

An option to prevent showing some or all logged messages.

If Astro's default log message is confusing to the user in combination with your custom `message`, you can use `suppress: "default"` to suppress the default message and only log your message:
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -428,6 +429,56 @@ export default function createIntegration() {
...
+    hooks: {
+      'astro:config:done': ({ setAdapter }) => {
+        setAdapter({
+          name: '@matthewp/my-adapter',
+          serverEntrypoint: '@matthewp/my-adapter/server.js',
+          supportedAstroFeatures: {
+            sharpImageService: {
+              support: 'limited',
+              message: 'This adapter has limited support for Sharp, certain features may not work as expected.',
+              suppress: 'default'
```
```suggestion
              suppress: 'default' // Does not log the default message 'Cannot find Sharp'
```
I don't know what the default message is, but this is one of those "sounds like there's a story behind this example" 😄  - This could be a good way to illustrate what we mean by confusing or contradictory messages

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -428,6 +429,56 @@ export default function createIntegration() {
...
+            }
+          } 
+        });
+      },
+    },
+  };
+}
+```
+
+You can also use `suppress: "all"` to suppress all messages for a specific feature, useful if it doesn't impact users in a specific context:
```
```suggestion
You can also use `suppress: "all"` to suppress all messages for a specific feature. This is useful if these messages are unhelpful to users in a specific context. For example, if your adapter supports (but does not rely on) Sharp, you can choose to prevent logging any messages about Sharp:
```
I think again the text/code sample could explain some real-world context better here, maybe by naming this `@matthewp/my-adapter-that-does-not-use-sharp` or something like that.

Though, it might seem a little confusing that `sharpImageService` is *supported* but never needs any messages about it? So, can we maybe have an explanation (better than mine above "supports but does not rely on") of something where this would be used?

src/content/docs/en/reference/adapter-reference.mdx
yanthomasdev: > (Aside: should that be "emit contextual INFORMATION to the logs" ??)

probably, yeah. We found a typo!

src/content/docs/en/reference/adapter-reference.mdx
yanthomasdev: The default message is a bit long, not sure if it works well to include it, but it should be this:
`The adapter @matthewp/my-adapter has limited support for "sharp". Certain features may not work as expected.`

src/content/docs/en/reference/adapter-reference.mdx
ascorbic: Side-note, but I think it's better to use the `@example` scope for package names in examples, rather than a real one

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: Why is the default message the same as the custom message? (Sorry, not sure how this works)

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: We can slowly remove Matthew from the docs examples... 😄

src/content/docs/en/reference/adapter-reference.mdx
yanthomasdev: It's not the same message, it's more that the custom message I used as an example is not great, I guess its a bit hard to make a good example of this without being exactly the case we had in core 😅

yanthomasdev: I guess we could use something close enough? So we use a message like so:
```
message: 'The adapter has limited support for "sharp". It will be used for images during build time, but will not work at runtime.'
```

src/content/docs/en/reference/adapter-reference.mdx
yanthomasdev: I am not sure what you mean by "supports" but "does not rely on". Maybe my example above that its supported on build time but not runtime? It might not even need to be something about sharp support!

sarah11918: @yanthomasdev My question is, why is sharp listed as a supported Astro feature in the first place, but yet we don't want to show any error logs for it? What could be an example of a feature that *would* show up as a `supportedAstroFeature` but the adapter would want to silence all logs?

Like, in this case, the fact that Sharp may not work as intended seems like a helpful message to have in an adapter that supports Sharp. So, what's the reason for silencing it? That's more what I'm asking. The example shows how to suppress something, but I really don't know WHY I might do that in this particular example.

I think the example is more helpful not just showing where you set the thing, but if the example is actually a case where you WOULD set it, too.

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: OK, so then the point here would be we would want to suppress the default message because you wouldn't want the user to see both, because the two messages are similar anyway?

sarah11918:
Sorry, my previous commit merged before I changed the commit message! Previous commit:

- created a new `suppress` heading with type/version info
- referenced this section at the beginning of the Astro Features section
- updates all the `@matthewp` examples to `@example`

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -429,22 +429,32 @@ export default function createIntegration() {
...
-          name: '@matthewp/my-adapter',
-          serverEntrypoint: '@matthewp/my-adapter/server.js',
+          name: '@example/my-adapter',
+          serverEntrypoint: '@example/my-adapter/server.js',
           supportedAstroFeatures: {
             sharpImageService: {
               support: 'limited',
-              message: 'This adapter has limited support for Sharp, certain features may not work as expected.',
-              suppress: 'default'
+              message: 'message: 'The adapter has limited support for Sharp. It will be used for images during build time, but will not work at runtime.',
```
yanthomasdev:
```suggestion
              message: 'The adapter has limited support for Sharp. It will be used for images during build time, but will not work at runtime.',
```

src/content/docs/en/reference/adapter-reference.mdx
yanthomasdev: @sarah11918 In the example of `imageService`, if you set an `imageService` that isn't sharp, e.g. `"cloudflare"`, you aren't using sharp anymore in your project, so it becomes irrelevant to show that log every time you open the dev server, right?  Which was something someone made a issue for before.

Or, an integration author might want to add an option in their adapter to suppress it as a configuration property, so you don't get the log if you are aware of the limitation, for example. Having one less log about something *you know* makes it easier to see logs about things you *might not* when using the dev server, especially considering people generally have a small terminal window in their IDE, so the real state there is relevant.

I guess that's the most common reason why this is useful. I am having a hard time thinking of a good example that isn't super specific, and maybe we don't need a great example? This is for integration authors and not the average user.

sarah11918: Yeah, I think the text we have is fine, and am happy for it to merge as is.

I'm just trying to figure out why you would put that Sharp is a `supportedAstroFeature` in your adapter if you're going to use an imageService that isn't sharp. 😄

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -399,26 +400,86 @@ export default function createIntegration() {
...
+            }
+          } 
+        });
+      },
+    },
+  };
+}
+```
+
+You can also use `suppress: "all"` to suppress all messages for a specific feature. This is useful when these messages are unhelpful to users in a specific context. For example, you can choose to prevent logging any messages about Sharp:
```
```suggestion
You can also use `suppress: "all"` to suppress all messages about support for the feature. This is useful when these messages are unhelpful to users in a specific context, such as when they have a configuration setting that means they are not using that feature. For example, you can choose to prevent logging any messages about Sharp support from your adapter:
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -399,26 +400,86 @@ export default function createIntegration() {
...
+
+### `suppress`
+
+<p>
+
+  **Type:** `default | all`<br />
+  <Since v="5.9.0" />
+</p>
+
+An option to prevent showing some or all logged messages.
```
```suggestion
An option to prevent showing some or all logged messages about an adapter's support for a feature.
```

PR #11841, opened 2025-06-05
i18n(fr): improve the wording in already updated files (second pass)

ArmandPhilippot:
Sorry for the title not really explicit 😅 

#### Description (required)

This PR updates all the files already reviewed since #11593 to apply changes from the new i18n-guide update, see #11795
* replace `délimitateur` with `délimiteur`
* replace `support` with `prise en charge`
* remove `en direct` where it doesn't make sense
* update the use of `en ligne` for `inline` by either rewording the sentence or removing `en ligne` because I think the context is already clear no need to make the sentence cumbersome by repeating `dans le fichier`.
* reword a sentence in the Pelican guide to make reading smoother.

- Suggested label: i18n


thomasbnt (approved):
LGTM!

PR #11850, opened 2025-06-06
fix: update `csp.mdx`

jsparkdev:
#### Description (required)

Update incorrect order, method names, description formatting, etc.


src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -49,8 +48,8 @@ This element will have the `http-equiv="content-security-policy"` attribute, and
   <meta 
     http-equiv="content-security-policy" 
     content="
-      script-src 'self' 'sha256-somehash'; 
       style-src 'self' 'sha256-somehash';
+      script-src 'self' 'sha256-somehash';
```
ArmandPhilippot: Are your sure about the order? I was trying to find something in the source code... [I came across](https://github.com/withastro/astro/blob/a8e1c0a7402940e0fc5beef669522b315052df1b/packages/astro/src/runtime/server/render/csp.ts#L41):
```ts
return `${directives} ${scriptSrc} ${styleSrc}`;
```
So, `script` should be before `style` as before, no? Or maybe I'm looking in the wrong place 👀

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -236,8 +234,8 @@ After the build, the `<meta>` element will include your additional hashes in the
 <meta
   http-equiv="content-security-policy"
   content="
-    script-src 'self' 'sha256-generatedByAstro' 'sha384-scriptHash' 'sha512-scriptHash' 'sha256-scriptHash'; 
-    style-src 'self' 'sha256-generatedByAstro' 'sha384-styleHash' 'sha512-styleHash' 'sha256-styleHash';
+    style-src 'self' 'sha384-styleHash' 'sha512-styleHash' 'sha256-styleHash' 'sha256-generatedByAstro';
+    script-src 'self' 'sha384-scriptHash' 'sha512-scriptHash' 'sha256-scriptHash' 'sha256-generatedByAstro';
```
ArmandPhilippot: I haven't tested myself but looking at:
https://github.com/withastro/astro/blob/a8e1c0a7402940e0fc5beef669522b315052df1b/packages/astro/src/runtime/server/render/csp.ts#L7-L21
This seems correct. Assuming I understood the code right and `extraScriptHashes` are the hashes generated by Astro. So, nice catch!

ArmandPhilippot (commented):
Nice catches, @jsparkdev!
I just have a doubt regarding the order of `script-src` and `style-src`. Did you saw that by testing yourself?
Otherwise, all the other changes looks good to me!

src/content/docs/en/reference/experimental-flags/csp.mdx
ArmandPhilippot: Oh so this one LGTM as well. It seems I was looking at the wrong place: https://github.com/withastro/astro/blob/a8e1c0a7402940e0fc5beef669522b315052df1b/packages/astro/src/runtime/server/render/head.ts#L41-L47

ArmandPhilippot (approved):
It seems I was looking at the wrong place for the `script-src`/`style-src` order. So everything LGTM! Thanks! 🙌🏽

src/content/docs/en/reference/experimental-flags/csp.mdx
sarah11918: I'm asking for further clarification here!

jsparkdev: This part is correct.

> Are your sure about the order? I was trying to find something in the source code... [I came across](https://github.com/withastro/astro/blob/a8e1c0a7402940e0fc5beef669522b315052df1b/packages/astro/src/runtime/server/render/csp.ts#L41):
> 
> ```ts
> return `${directives} ${scriptSrc} ${styleSrc}`;
> ```
> 
> So, `script` should be before `style` as before, no? Or maybe I'm looking in the wrong place 👀

This part seems to refer to the `<style>`, `<link>`, and `<script>` tags, not the CSP meta tag.

> Oh so this one LGTM as well. It seems I was looking at the wrong place: [withastro/astro@`a8e1c0a`/packages/astro/src/runtime/server/render/head.ts#L41-L47](https://github.com/withastro/astro/blob/a8e1c0a7402940e0fc5beef669522b315052df1b/packages/astro/src/runtime/server/render/head.ts#L41-L47)

This appears to be a bug for script sources and style sources, and I requested a fix in https://github.com/withastro/astro/pull/13905.

> I'm asking for further clarification here!

jsparkdev:
@sarah11918 @ArmandPhilippot 

Thank you all for your reviews.

I've noticed myself that enabling `experimental.csp` always brings up the style source first, but this appears to be a bug.

If I'm right about the bug, I've requested a bug fix in https://github.com/withastro/astro/pull/13905 and will update the documentation once it's merged.

src/content/docs/en/reference/experimental-flags/csp.mdx
ArmandPhilippot: Ok, it seems I got even more confused by looking for the correct order in the code source... 😅 I was looking in the right place the first time. But I didn't see that the variables didn't contain the correct value when linking to this part! Nice catch, again!

So `script` comes before `style` as before but there was an error in the implementation. Sarah did well to ask further clarification on this! Great team work! 😄

PR #11868, opened 2025-06-09
chore: update CSP docs for v5.9.2

ematipico:
#### Description (required)

This PR updates the CSP docs where view transitions via `ClientRouter` aren't supported anymore

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/csp.mdx
+++ b/src/content/docs/en/reference/experimental-flags/csp.mdx
@@ -21,7 +21,7 @@ Enables support for [Content Security Policy (CSP)](https://developer.mozilla.or
 
 Enabling this feature adds additional security to **Astro's handling of processed and bundled scripts and styles** by default, and allows you to further configure these, and additional, content types.
 
-This experimental CSP feature has some limitations. Inline scripts are not supported out of the box, but you can [provide your own hashes](#hashes) for external and inline scripts. [Astro's view transitions](/en/guides/view-transitions/) using the `<ClientRouter />` aren't supported.
+This experimental CSP feature has some limitations. Inline scripts are not supported out of the box, but you can [provide your own hashes](#hashes) for external and inline scripts. [Astro's view transitions](/en/guides/view-transitions/) using the `<ClientRouter />` are not supported, but you can [consider migrating to the browser native View Transition API](https://events-3bg.pages.dev/jotter/astro-view-transitions/) instead if you are not using Astro's enhancements to the native View Transitions and Navigation APIs.
 
 :::note
 Due to the nature of the Vite dev server, this feature isn't supported while working in `dev` mode. Instead, you can test this in your Astro project using `build` and `preview`.
--- a/src/content/docs/en/reference/experimental-flags/csp.mdx
+++ b/src/content/docs/en/reference/experimental-flags/csp.mdx
@@ -1,5 +1,5 @@
 ---
-title: Experimental Content Security Policy
+title: Experimental Content Security Policy (CSP)
 sidebar:
   label: Content Security Policy
 i18nReady: true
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -21,7 +21,7 @@ Enables support for [Content Security Policy (CSP)](https://developer.mozilla.or
 
 Enabling this feature adds additional security to **Astro's handling of processed and bundled scripts and styles** by default, and allows you to further configure these, and additional, content types.
 
-This experimental CSP feature has some limitations. Inline scripts are not supported out of the box, but you can [provide your own hashes](#hashes) for external and inline scripts. Additionally, [Astro's view transitions](/en/guides/view-transitions/) using the `<ClientRouter />` are not yet fully supported: when navigating from one page to another, some styles may not be applied and some scripts may not be executed.
+This experimental CSP feature has some limitations. Inline scripts are not supported out of the box, but you can [provide your own hashes](#hashes) for external and inline scripts. Since v5.9.2, [Astro's view transitions](/en/guides/view-transitions/) using the `<ClientRouter />` are not supported anymore.
```
ascorbic: I'd just say it's not supported. We don't need to preserve the fact that it was supposedly supported but didn't work for a few days

ematipico:
```suggestion
This experimental CSP feature has some limitations. Inline scripts are not supported out of the box, but you can [provide your own hashes](#hashes) for external and inline scripts. [Astro's view transitions](/en/guides/view-transitions/) using the `<ClientRouter />` aren't supported.
```

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -21,7 +21,7 @@ Enables support for [Content Security Policy (CSP)](https://developer.mozilla.or
 
 Enabling this feature adds additional security to **Astro's handling of processed and bundled scripts and styles** by default, and allows you to further configure these, and additional, content types.
 
-This experimental CSP feature has some limitations. Inline scripts are not supported out of the box, but you can [provide your own hashes](#hashes) for external and inline scripts. Additionally, [Astro's view transitions](/en/guides/view-transitions/) using the `<ClientRouter />` are not yet fully supported: when navigating from one page to another, some styles may not be applied and some scripts may not be executed.
+This experimental CSP feature has some limitations. Inline scripts are not supported out of the box, but you can [provide your own hashes](#hashes) for external and inline scripts. [Astro's view transitions](/en/guides/view-transitions/) using the `<ClientRouter />` aren't supported.
```
```suggestion
This experimental CSP feature has some limitations. Inline scripts are not supported out of the box, but you can [provide your own hashes](#hashes) for external and inline scripts. [Astro's view transitions](/en/guides/view-transitions/) using the `<ClientRouter />` are not supported, but you can [consider migrating to the browser native View Transition API](https://events-3bg.pages.dev/jotter/astro-view-transitions/) instead if you are not using Astro's enhancements to the native View Transitions and Navigation APIs.
```
Just thought it would be nice to provide this option here, since it's possible some will value CSP more than the `<ClientRouter />`, and we are also updating the View Transitions page to encourage people to consider whether they still even need this component at all anymore,

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Experimental Content Security Policy
+title: Experimental Content Security Policy (CSP)
```
ematipico, line 2: The reason I added `CSP` is that it helps with the search. I searched CSP and couldn't find the page cc @sarah11918

sarah11918: Yes, this makes perfect sense! Thank you!

PR #11882, opened 2025-06-12
fix(csp): code examples

ematipico:
#### Description (required)

This PR does two things:
- it updates some examples where the URLs aren't supposed to have apostrophes
- updates some APIs, to be in line with https://github.com/withastro/astro/pull/13931


ematipico:
Hey @sarah11918 , the scope of the PR changed. Now it's just a fix for some examples. Here's the context https://github.com/withastro/astro/pull/13931#issuecomment-2969585731

I updated the PR with only fixes

sarah11918 (approved):
Thank you Ema!

PR #11884, opened 2025-06-13
feat: document `experimentalStaticHeaders`

ematipico:
#### Description (required)

This PR documents the newly added Astro feature `experimentalStaticHeaders`. The relative PR has been merged but not yet released.

Link to the PR https://github.com/withastro/astro/pull/13926

### DIFF
--- a/src/content/docs/en/reference/adapter-reference.mdx
+++ b/src/content/docs/en/reference/adapter-reference.mdx
@@ -672,7 +672,7 @@ When this feature is enabled, Astro will return a map of the `Headers` emitted b
   
 The value of the headers might change based on the features enabled/used by the application.
 
-For example, if CSP enabled, the `<meta http-equiv="content-security-policy">` element **isn't** added to the static page, instead its `content` is available in the `experimentalRouteToHeaders` map. 
+For example, if CSP enabled, the `<meta http-equiv="content-security-policy">` element is not added to the static page. Instead its `content` is available in the `experimentalRouteToHeaders` map. 
 
 ```js title="my-adapter.mjs" ins={9-11}
 export default function createIntegration() {


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -659,3 +659,42 @@ export default function createIntegration() {
...
+**Type:** `true | false`<br />
+<Since v="5.9.3" />
+</p>
+
+
+When this feature is enabled, Astro will return a map of the `Headers` emitted by the static pages. This map `experimentalRouteToHeaders` is available in the `astro:build:generated` hook.
+  
+The value of the headers might change based on the features enabled/used by the application.
+
+For example, if CSP enabled, the `<meta http-equiv="content-security-policy">` element **isn't** added to the static page, instead its `content` is available in the `experimentalRouteToHeaders` map.
```
```suggestion
For example, if CSP enabled, the `<meta http-equiv="content-security-policy">` element is not added to the static page. Instead its `content` is available in the `experimentalRouteToHeaders` map. 
```
Just a fun note for future historians wondering why this PR suggestion is regularly made when editing docs PRs: we typically **do not use bold text for individual word stress** within a sentence.

We sometimes do use bold, but it would normally be for a brief point that we don't want someone skimming to miss. The idea is that **bold calls attention to a key idea**, so like I've done here, it's typically a short "tl/dr" phrase.

One way to think about this is, if someone were skimming the text and they just saw ISN'T or NOT screaming out at them, that alone is not really helpful at conveying a thought.

It's actually sometimes helpful to receive these bold words in a PR, as a clue to reviewers that this might be an important or unexpected part of how something works! But we will always remove it from the final documentation.

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -659,3 +659,42 @@ export default function createIntegration() {
...
+
+### experimentalStaticHeaders
+
+
+<p>
+
+**Type:** `true | false`<br />
+<Since v="5.9.3" />
+</p>
+
```
sarah11918, line 670:
```suggestion
<p>

**Type:** `true | false`<br />
<Since v="5.9.3" />
</p>
```
Remove some blank lines

src/content/docs/en/reference/adapter-reference.mdx
ematipico: Very helpful, thank you!

PR #11889, opened 2025-06-15
new intro for view transitions

### LINKED ISSUE #10902: Add some hint that the <ClientRouter /> component is no longer Astro's preferred approach to view transitions.
### 📚 Subject area/topic

View Transitions

https://docs.astro.build/de/guides/view-transitions/

Right now, view transitions are prominently featured at the end of the 'Routing and Navigation' section in the docs. The page gives the impression that, even in 2025, the recommended way to enable view transitions in an Astro project is by adding the Astro-specific `<ClientRouter />` component.

This issue is intended as a nudge to follow through on [Fred's announcement](https://astro.build/blog/future-of-astro-zero-js-view-transitions/) from last year. The `<ClientRouter />` fka `<ViewTransitions />` was originally meant to allow Astro users early experiments with the then new View Transition API and bridge the time until more browsers offer cross-document view transitions. We shouldn't present the `<ClientRouter />` as a routing must-have. Instead, we should clarify upfront that it's a deprecated approach that doesn't align with Astro's "no JavaScript by default" philosophy. And that browser-native cross-document view transitions are the better alternative (if we agree on that).

I'm happy to leave the wording in docs to whoever is more involved with the docs.

From my perspective, these points might be helpful for Astro users:

- Astro provides a built-in component for cross-document view transitions, even for browsers that only support same-document transitions. (Note: Currently, all browsers either support both or neither.)

- It even enables a subset of view transitions for browsers with no native support at all. (Note: Right now, all major browsers except Firefox fully support native view transitions.)

- No need for FOMO: Despite the fancy name, the ClientRouter isn’t unlocking a groundbreaking new world of client-side routing. 

- There are still some things the ClientRouter handles better than native view transitions ([Firefox fallbacks, transition:persist, directional transitions, and animation functions](https://events-3bg.pages.dev/jotter/feature-comparison/)). When using native view transitions, you would need extra (third-party) scripts for these.

- But learning the still evolving web standard might be a better long-term investment than learning an Astro-specific subset variant.

- If your page relies on non-trivial scripts, integrating them with the ClientRouter can be a challenge.

This background information doesn't need to go into the docs. Maybe a brief note at the start of the page is enough to let people know they might be better off using CSS View Transitions instead of the `<ClientRouter />` component.

_No response_

martrapp:
#### Description (required)

Adding the introduction from #10902

- Closes #
- Suggested label: 

Closes #10902

### DIFF
--- a/src/content/docs/en/guides/view-transitions.mdx
+++ b/src/content/docs/en/guides/view-transitions.mdx
@@ -6,17 +6,25 @@ i18nReady: true
 import Since from '~/components/Since.astro'
 import { Steps } from '@astrojs/starlight/components'
 
-Astro supports view transitions with just a few lines of code. View transitions update your page content without the browser’s normal, full-page navigation refresh and provide seamless animations between pages. Where [browser support for the View Transition API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API#browser_compatibility) is lacking, Astro allows you to [control fallback strategies](#fallback-control) for the best possible experience for all visitors.  
+[View transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API) are animated transitions between different website views. They are a popular design choice for preserving visual continuity as visitors move between states or views of an application.
 
-Astro provides a `<ClientRouter />` routing component that can be added to a single page's `<head>` to control page transitions as you navigate away to another page. It provides a lightweight client-side router that [intercepts navigation](#client-side-navigation-process) and allows you to customize the transition between pages.
+[Browser-native, cross-document view transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API/Using#basic_mpa_view_transition) can be used in Astro to animate the navigation between documents in a multi-page app (MPA), often providing the experience of client-side routing of single-page applications. They don’t alter the core functionality of a multi-page application, nor do they affect any existing scripts or add additional JavaScript to your page load. They simply add animations.
 
-Add this component to a reusable `.astro` component, such as a common head or layout, for [animated page transitions across your entire site (SPA mode)](#full-site-view-transitions-spa-mode).
+For enhanced client-side routing and view transition features not yet fully supported by the View Transition API, Astro provides a built-in, lightweight component to enable client-side routing and turn your multi-page app into a [single-page app](#full-site-view-transitions-spa-mode) with smooth animations on navigation. 
 
-Astro's view transitions support is powered by the new [View Transitions](https://developer.chrome.com/docs/web-platform/view-transitions/) browser API and also includes:
+That comes with some benefits, like shared state across pages and persistent elements, and some drawbacks, such as needing to manually reinitialize scripts or state after navigation.
+
+Adding the `<ClientRouter />` [intercepts page navigation](#client-side-navigation-process) and gives you considerable control over this process; extends and enhances some View Transition/Navigation API features; and additionally allows you to [configure fallback strategies](#fallback-control) for when [native browser support is lacking](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API#browser_compatibility).
+
+However, as browser APIs and web standards evolve, using Astro's `<ClientRouter />` for this additional functionality will increasingly become unnecessary. We recommend keeping up with the current state of browser APIs so you can [decide whether you still need Astro's client-side routing](https://events-3bg.pages.dev/jotter/astro-view-transitions/) for the specific features you use.
+
+
+Astro's view transitions support is powered by the [View Transitions](https://developer.chrome.com/docs/web-platform/view-transitions/) browser API and also includes:
 
 - A few [built-in animation options](#built-in-animation-directives), such as `fade`, `slide`, and `none`.
 - Support for both forwards and backwards navigation animations.
 - The ability to fully [customize all aspects of transition animation](#customizing-animations), and build your own animations.
+- A way to carry HTML elements from the current page to the next during navigation.
 - The option to [prevent client-side navigation for non-page links](#preventing-client-side-navigation).
 - [Control over fallback behavior](#fallback-control) for browsers that do not yet support the View Transition APIs.
 - Automatic support for [`prefers-reduced-motion`](#prefers-reduced-motion).


sarah11918:
Thank you, Martin! I will ask for more reviews since I mostly crafted this draft originally!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -6,17 +6,25 @@ i18nReady: true
...
+[View transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API) are animated transitions between different website views. They are a popular design choice for preserving visual continuity as visitors move between states or views of an application.
 
-Astro provides a `<ClientRouter />` routing component that can be added to a single page's `<head>` to control page transitions as you navigate away to another page. It provides a lightweight client-side router that [intercepts navigation](#client-side-navigation-process) and allows you to customize the transition between pages.
+[Browser-native, cross-document view transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API/Using#basic_mpa_view_transition) can be used in Astro to animate the navigation between documents in a multi-page app (MPA), often providing the experience of client-side routing of single-page applications. They don’t alter the core functionality of a multi-page application, nor do they affect any existing scripts or add additional JavaScript to your page load. They simply add animations.
```
ArmandPhilippot: Should this be under a heading like "Choosing when to use native view transitions vs ClientRouter"?

What I mean is: the previous intro was already long and now it is even longer. Would a heading make it easier to read the page to get the right information and/or hold the reader's attention?

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -6,17 +6,25 @@ i18nReady: true
...
+For enhanced client-side routing and view transition features not yet fully supported by the View Transition API, Astro provides a built-in, lightweight component to enable client-side routing and turn your multi-page app into a [single-page app](#full-site-view-transitions-spa-mode) with smooth animations on navigation. 
 
-Astro's view transitions support is powered by the new [View Transitions](https://developer.chrome.com/docs/web-platform/view-transitions/) browser API and also includes:
+That comes with some benefits, like shared state across pages and persistent elements, and some drawbacks, such as needing to manually reinitialize scripts or state after navigation.
+
+Adding the `<ClientRouter />` [intercepts page navigation](#client-side-navigation-process) and gives you considerable control over this process; extends and enhances some View Transition/Navigation API features; and additionally allows you to [configure fallback strategies](#fallback-control) for when [native browser support is lacking](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API#browser_compatibility).
```
ArmandPhilippot: I'm not a native speaker so this is probably related, but would a list make reading more fluid (i.e., a shorter sentence that is easier to scan):
```suggestion
Adding the `<ClientRouter />`:
- [intercepts page navigation](#client-side-navigation-process) and gives you considerable control over this process;
- extends and enhances some View Transition/Navigation API features;
- allows you to [configure fallback strategies](#fallback-control) for when [native browser support is lacking](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API#browser_compatibility).
```
Note: I'm not sure if we should use semicolon or dot at the end. I think the standard in English is semicolon but we have a lot of list (like the one below) that use colon so... I'm not enough familiar with the right format to use.

ArmandPhilippot (approved):
The content looks great to me! But, I left two comments regarding the form: these are nitpicking so I'm approving and you're free to ignore them. 😄

What bothers me is that the intro is getting longer so I left some thoughts to help make scanning the page easier I think (but maybe it's not the best solution).
On other pages, the intro is often 1 or 2 paragraphs. So if it's the page's intro, people might be tempted to just skip it until the next heading while there is valuable information here.

src/content/docs/en/guides/view-transitions.mdx
sarah11918: Yes, the length of the intro (esp. compared to other pages) was something I noticed too! That's why I wanted more reviews. Your suggestion of another header is great!

src/content/docs/en/guides/view-transitions.mdx
sarah11918: We're not all standardized yet, but the format *I* typically use when creating new bulleted lists is this:

This is can be either a full sentence, or a sentence fragment that is completed by all bullet points below, ending in a colon:
- If this is a full sentence on its own, start with an uppercase letter and end with a period.
- If this *completes* a sentence fragment, start with a *lowercase* letter, and still end with a period.
- If this completes a sentence fragment and is SUPER SHORT (or is a simple list of examples instead of actually completing the sentence), no punctuation at all.

Semicolons look and feel formal, so I go with "each bullet either is or makes a complete thought, and therefore ends in a period." When the list replaces a normal list of items after a colon, no end punctuation. 

So in your case above, I'd end them ALL with periods. I'm going to commit this, then in conjunction with your heading idea, do a slight reordering, I think

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -25,6 +22,21 @@ Astro's view transitions support is powered by the new [View Transitions](https:
...
 
+## Differences between browser-native view transitions and Astro's `<ClientRouter />`
+
+[Browser-native, cross-document view transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API/Using#basic_mpa_view_transition) can be used in Astro to animate the navigation between documents in a multi-page app (MPA), often providing the experience of client-side routing of single-page applications. They don’t alter the core functionality of a multi-page application, nor do they affect any existing scripts or add additional JavaScript to your page load. They simply add animations.
+
+For enhanced client-side routing and view transition features not yet fully supported by the View Transition API, Astro provides a built-in, lightweight component to enable client-side routing and turn your multi-page app into a [single-page app](#full-site-view-transitions-spa-mode) with smooth animations on navigation. 
+
+That comes with some benefits, like shared state across pages and persistent elements, and some drawbacks, such as needing to manually reinitialize scripts or state after navigation.
+
+Adding Astro's `<ClientRouter />` component:
```
```suggestion
Adding Astro's built-in `<ClientRouter />` component:
```

sarah11918:
@ArmandPhilippot See what you think of this reordering to accommodate a heading!

ArmandPhilippot (approved):
Nice job with the reordering, I think this looks great! 🙌🏽

martrapp:
Adding the subheading was a great idea, I really like it this way!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -25,6 +22,21 @@ Astro's view transitions support is powered by the new [View Transitions](https:
...
+That comes with some benefits, like shared state across pages and persistent elements, and some drawbacks, such as needing to manually reinitialize scripts or state after navigation.
+
+Adding Astro's built-in `<ClientRouter />` component:
+- [intercepts page navigation](#client-side-navigation-process) and gives you considerable control over this process.
+- extends and enhances some View Transition/Navigation API features.
+- allows you to [configure fallback strategies](#fallback-control) for when [native browser support is lacking](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API#browser_compatibility).
+
+However, as browser APIs and web standards evolve, using Astro's `<ClientRouter />` for this additional functionality [will increasingly become unnecessary](https://astro.build/blog/future-of-astro-zero-js-view-transitions/). We recommend keeping up with the current state of browser APIs so you can [decide whether you still need Astro's client-side routing](https://events-3bg.pages.dev/jotter/astro-view-transitions/) for the specific features you use.
+
 ## Adding View Transitions to a Page
```
martrapp, line 40: Isn't this a bit misleading as you can't enable view transitions on a single page but only on a navigation A -> B, which requires the `<Clientrouter />` in A's and B's `<head>`.

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -25,6 +22,21 @@ Astro's view transitions support is powered by the new [View Transitions](https:
 By default, every page will use regular, full-page, browser navigation. You must opt in to view transitions and can use them either on a per-page basis or site-wide.
```
martrapp, line 22: Not sure if this would be a good idea, but we could link down to "per page" and "full site" from this note as a way to skip directly to the sections covering how to enable view transitions

src/content/docs/en/guides/view-transitions.mdx
martrapp: Isn't "Adding View Transitions to a Page" a bit misleading as you can't enable view transitions on a single page but only on a navigation A -> B, which requires the <Clientrouter /> in A's and B's <head>.

sarah11918: I was kind of wondering whether we take that part out entirely and just focus on site-wide transitions. It was written when we weren't sure how people were going to use them and wanted to make it super explicit that every page needed it in its head, but now, is it likely that people will NOT simply enable them throughout a site?

martrapp: So true!

sarah11918: @martrapp - See what you think of it now!

sarah11918 (approved):
omg I finally got all the checks to pass, @martrapp !! 😅 

I'm happy with this if you are, so it's yours to merge when you're satisfied!

src/content/docs/en/guides/view-transitions.mdx
martrapp: perfect.

ArmandPhilippot:
We might need a Lunaria directive because of changes in `pl/guides/actions.mdx`, right?

Edit: `@lunaria-track:src/content/docs/en/**/*.mdx` should work I think?

martrapp:
Read the final version once more. I think this turned out really great. It already started off well, and now it got even better. I am fully on board with it. And I am happy that our users now may have a clearer sense of their options.

Thank you so much for putting such care into it!

martrapp:
> We might need a Lunaria directive because of changes in `pl/guides/actions.mdx`, right?

Anything I have to be aware of before merging, @ArmandPhilippot?

ArmandPhilippot:
@martrapp  You just need to put `@lunaria-track:src/content/docs/en/**/*.mdx` in the commit description before merging and the `pl` file should be ignored (ie. not marked as up-to-date when it is not)!

PR #11896, opened 2025-06-18
fix(basics/ps): add an exception to 'src/pages' requirment

ghassanmas:
#### Description (required)

This doc regatrding that the project has to have `src/page` directory, is not accurate, this change adds an exception. 

 It's possible for the project to not have 'src/pages', if 'injectRoiute' is used, simliar to Starlight theme[^1].

  https://github.com/withastro/docs/blob/734084c6bc05be64a43c77f582540394f5ab8206/src/content/docs/en/reference/integrations-reference.mdx?plain=1#L415-L428

### DIFF
--- a/src/content/docs/en/basics/project-structure.mdx
+++ b/src/content/docs/en/basics/project-structure.mdx
@@ -81,7 +81,7 @@ While this guide describes some popular conventions used in the Astro community,
 Pages routes are created for your site by adding [supported file types](/en/basics/astro-pages/#supported-page-files) to this directory.
 
 :::caution
-`src/pages` is a **required** sub-directory in your Astro project. Without it, your site will have no pages or routes!
+`src/pages` is a **required** sub-directory in your Astro project. Without it, your site will have no pages or routes!, unless you use `injectRoiute()` [an optional callback of the setup hook of the integration API](/en/reference/integrations-reference/#injectroute-option)
 :::
 
 ### `src/components`


florian-lefebvre:
Hi, thanks for opening a PR! This page is meant to be introductory so we are hesitant to include such info **here**, that may confuse new users. Such info should be provided by an integration if it handles routing entirely.

Is there a specific reason you opened this PR in the first place? Eg. you struggled to setup something? Maybe there are other places where we can clarify this

ghassanmas:
> Hi, thanks for opening a PR! This page is meant to be introductory so we are hesitant to include such info here, that may confuse new users. Such info should be provided by an integration if it handles routing entirely.

Agree that it's an advaced use case, I think may be it can be a added as footnote? 

> Is there a specific reason you opened this PR in the first place? Eg. you struggled to setup something? Maybe there are other places where we can clarify this

 Yes, when using/comparing with [Starlight](https://github.com/withastro/starlight), as the project dosen't have 'src/pages` which contridict the statement

sarah11918:
Hi! As Florian said, this is a very basic, introductory page for people to understand their Astro project. If they have a template where the integration is taking over routing and handling `src/pages/index.astro` under the hood, then that template/integration will probably describe what about Astro's project structure is relevant and their routing choices. For example, the Starlight docs do describe how `src/pages/` is used for routes that they do not inject.

But in this basic project structure guide to introduce Astro, adding this extra information is potentially unhelpful to a new Astro user who may not understand the importance of this folder in their regular Astro project. It will not be relevant to someone *unless* they are using something like Starlight, which will have its own documentation, so we think it's best to not risk confusing someone on this page.

ghassanmas:
I think at least, the wording could be imporved to add something like _unless overriding advanced config_.  

My concern is that, the stateament is wrapped in _cautious qoute while it's not true_, in the most used theme/integration of the project.

PR #11899, opened 2025-06-19
fix: add `prerender = false` to Better Auth dynamic API route example

### LINKED ISSUE #11890: Missing prerender in better-auth docs
### 📚 Subject area/topic

Auth 

https://docs.astro.build/en/guides/authentication/#better-auth

The Docs tell us to create dynamic route in: src/pages/api/auth/[...all].ts

If prerender is not set to false you will get an Error: **`getStaticPaths()` function required for dynamic routes.**

so we must add ` export const prerender = false;`

```ts
import { auth } from "@libs/auth";

export const prerender = false;

export const ALL: APIRoute = async (ctx) => {
	console.log(ctx);
	ctx.request.headers.set("x-forwarded-for", ctx.clientAddress);
	return auth.handler(ctx.request);
};
```

_No response_

davidumoru:
#### Description (required)

Added `export const prerender = false;` to the Better Auth dynamic API route example in the authentication guide.  
This prevents Astro from trying to prerender the dynamic route, which would otherwise cause a build error.

- Closes #11890

Yes, this is my first contribution to Astro Docs.  
Discord username: `@notumoru`


sarah11918:
Hi @davidumoru ! Thank you for grabbing this issue and adding the extra line! Did you notice my comment in the issue that I *think* there are probably more code snippets than just that one that need this line? Did you check for those? If so, we can add them all at once here in this PR! 🙌

src/content/docs/en/guides/authentication.mdx
```diff
@@ -184,7 +184,9 @@ Configure your database table to store user data and your preferred authenticati
 ```ts title="src/pages/api/auth/[...all].ts"
 import { auth } from "../../../lib/auth"; // import your Better Auth instance
- 
+
+export const prerender = false; // Prevents static prerendering of this dynamic route
```
sarah11918:
```suggestion
export const prerender = false; // Not needed in 'server' mode
```
This is the line we use for these code snippets. We don't have to describe what the line does as much as we need to make it clear *when* it should be added, because another alternative is to have your entire app rendered on-demand by default!

davidumoru:
Only Better Auth uses a dynamic API route in the Authentication guide. Everything else is just .astro files, config, or middleware, so nothing else needs `prerender = false;`. Let me know if I missed anything, but I think we’re all set!

sarah11918:
Thanks for updating, but I still think this is needed in `.astro` pages? Like maybe here for example:

image

Anything that uses Sessions, Actions... it's not about dynamic routes themselves (those can be prerendered!) but about using an Astro feature on a page that requires server rendering, I think?

davidumoru:
Ah, got it, thanks for explaining! I’ll double-check the examples and update.

PR #11900, opened 2025-06-19
docs: add reference to Astro.slots utility functions in Components guide

davidumoru:
#### Description (required)

Added a `<ReadMore>` link to the `Astro.slots` utility functions in the Components guide, after the main Slots section.  
This makes it easier for users to discover advanced slot utilities when learning about slots in Astro components.

- Closes ##11871


sarah11918:
This looks great @davidumoru ! I'm trying to think of where to best "slot" this little note in so that it doesn't get lost. Since it doesn't relate to the specific headings, maybe this is best around line 222, as a conclusion to the main slots overview, and right *before* named slots? What do you think?

For it to be easily discoverable, we don't want people to e.g. need to go looking under "fallback content" in order to know that this exists.

davidumoru:
Thanks for the feedback, @sarah11918! I agree, placing it just before “Named Slots” makes it much more discoverable. I’ve moved it to that spot as recommended.

sarah11918 (approved):
Great! You're on a roll today! 🫡 Thank you again!

yanthomasdev (approved):
LGTM! Thanks!

PR #11908, opened 2025-06-19
docs: add note about non-deterministic sort order in collections

### LINKED ISSUE #11674: Explain that output order is not reliable when using `getCollection`
### Discussed in https://github.com/withastro/docs/discussions/11660

<div type='discussions-op-text'>

<sup>Originally posted by **ArmandPhilippot** May 12, 2025</sup>
Since https://github.com/withastro/astro/pull/13299, Astro no longer uses [`fast-glob`](https://www.npmjs.com/package/fast-glob) but [`tinyglobby`](https://www.npmjs.com/package/tinyglobby). Referring to a [Discord issue](https://discord.com/channels/830184174198718474/1371567015013122149), it seems there is a difference at least in the output order.

We should suggest that some kind of ordering can be necessary when using `getCollection()` because the order isn't predictable. A good place for this explanation would be in the [Querying Collections](https://docs.astro.build/en/guides/content-collections/#querying-collections) section.

We could use a similar wording than the one used in the [Upgrade to v5 guide](https://docs.astro.build/en/guides/upgrade-to/v5/#breaking-changes-to-legacy-content-and-data-collections):
> Sort order of generated collections is non-deterministic and platform-dependent. This means that if you are calling getCollection(), the order in which entries are returned may be different than before. If you need a specific order, you must sort the collection entries yourself.

davidumoru:
#### Description (required)

Added a note to the “Querying Collections” section of the Content Collections guide, explaining that the order of entries returned by `getCollection()` is non-deterministic and platform-dependent. This helps users understand that they should sort the results themselves if they need a specific order.

- Closes #11674


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -383,6 +383,10 @@ const poodleData = await getEntry('dogs', 'poodle');
 
 
 ```
+:::note
+Sort order of generated collections is non-deterministic and platform-dependent. This means that if you are calling getCollection(), the order in which entries are returned may be different than before. If you need a specific order, you must sort the collection entries yourself.
+:::
```
``````suggestion

The sort order of generated collections is non-deterministic and platform-dependent. This means that if you are calling `getCollection()` and need your entries returned in a specific order (e.g. blog posts sorted by date), you must sort the collection entries yourself:

```astro title="src/pages/blog.astro"

const posts = (await getCollection('blog')).sort(
	(a, b) => b.data.pubDate.valueOf() - a.data.pubDate.valueOf(),
);
```
``````

- updated the text because "different than before" is no longer meaningful if this isn't part of comparing old collections to new collections
- also thought a code example could be helpful here! (I grabbed this from the official Astro blog template)
- With this, the text is probably actually more readable if it's *not* in a note! Notes are for things that are "aside" of the regular explanation of how things work -- and this *is exactly* how it works, complete with an example, so I think it should be full proper text and not a HUGE note.

sarah11918 (commented):
Hi @davidumoru!  I was *sure* we had already written about this! Turns out, it's in the Upgrade to v5/from legacy collections guidance, but not here on this page. 😅 

Thank you for adding this helpful information in a place where more people will find it, and you'll find my suggestions below for making it fit within the flow of this new location!

davidumoru:
Thanks for the suggestions, just committed them. This version reads much better.

sarah11918 (approved):
Thank you so much, this is a fantastic addition to this page and will help a lot of people! We appreciate you taking the time to contribute, and are happy to welcome you to Team Docs! :partying_face:

PR #11901, opened 2025-06-19
5.10.0 Minor release docs

sarah11918:
Release docs for 5.10.0, including:

#11782 
#11741 
#11743 
#11902 
#11903 

NOTE: Some config reference docs were changed here and will need to be PR'd back at the source!

PR #11912, opened 2025-06-19
i18n(fr): update `configuration-reference`, `modules/astro-assets` & `guides/images`

ArmandPhilippot:
#### Description (required)

Those three files have crossed links so I've updated the French translation of these files in the same PR with changes from #11901 and #11911.

I also brought more changes to:
* `reference/modules/astro-assets` to fix a malformed Markdown bold syntax and to translate some code snippets
* `guides/images` to match the updated i18n guide (ie. `construction` > `compilation`, etc.) and to fix some translated inline code.

And I fixed links in other files because of some updates I made in `guides/images`.

- Suggested label: i18n


thomasbnt (approved):
LGTM!

PR #11920, opened 2025-06-20
fix(live-content-collections): increment `ol` + consistent verb tense

ArmandPhilippot:
#### Description (required)

In `experimental-flags/live-content-collections`:
* Fixes an ordered list: The rendered list is incremented but the MDX file numbers all elements with 1.
* Uses a consistent verb tense in the same list:  Not sure if we should use `[They] run` or `[it] runs` for the verb tense but currently we have a mix between the two. I guess the plural form makes more sense since we don't use the singular in the opening sentence.

- Suggested label: `consistency/formatting`

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/live-content-collections.mdx
+++ b/src/content/docs/en/reference/experimental-flags/live-content-collections.mdx
@@ -616,12 +616,12 @@ Live content collections have some limitations compared to build-time collection
 
 Live collections use a different API than current preloaded content collections. Key differences include:
 
-1. **Execution time**: Runs at request time instead of build time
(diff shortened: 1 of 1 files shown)


sarah11918:
I have looked at this so closely and every time "Run" sounds wrong but I'm convinced it's correct. 😂  Approved!

ArmandPhilippot:
😅 Well, if you prefer we could add a `s` everywhere. It just feels odd to have both tense currently! I have hesitated between with and without `s`... But we don't have a single `live content collection` (singular) above so I chose based on that.
You still have time to change your mind, I have other PRs to merge! 😄

jsparkdev (approved):
LGTM

PR #11923, opened 2025-06-21
fix: update `live-content-collections.mdx`

jsparkdev:
#### Description (required)

small updates `live-content-collections.mdx`


ArmandPhilippot (approved):
LGTM, nice catch for the highlighting I missed that!

The conflict is probably related to #11920 since we have updated the same list.

PR #11928, opened 2025-06-22
Update cloudflare.mdx

tobynguyen27:
#### Description (required)
According to this [docs](https://developers.cloudflare.com/workers/static-assets/routing/static-site-generation/#custom-404-pages)

Cloudflare Workers doesn't serve 404.html by default. For static websites, we need to add `"not_found_handling": "404-page"` to `wrangler.json` to tell Workers serve them

- Closes #
- Suggested label:


sarah11918:
Hi! I checked the docs at that link, and that appears to show this property as optional, and that it is for configuring a *custom* 404 page and overriding the defaults. I'm not sure we should include that here because it assumes that you have made a custom 404 page, and not everyone will have one.

This guide was recently submitted by someone at Cloudflare themselves, so they probably feel this is the best general advice. In Astro docs, we avoid documenting too much about how third party services work, but we do want to be helpful! 

I will ask for some other team opinions about adding this. If this is something that *most* people would do, then it can be reasonable to show it. But, this might be something that it is better if they read Cloudflare's own docs if they want this feature, so that they understand how to do it and use it.

alexanderniebuhr:
Thanks for the suggestion. This options should only be used when two requirements are met:

a) you are not using the Cloudflare adapter and have a 100% static page
b) you have at least a custom 404 page in ˋsrc/pagesˋ (not sure if Astro’s default one counts)

Since the deployment guide describes both approaches - using the adapter or non adapter - I think if we make sure this is added only to the correct tabs, we should recommend it.

src/content/docs/en/guides/deploy/cloudflare.mdx
```diff
@@ -69,7 +69,8 @@ To get started, you will need:
           // Update to today's date
           "compatibility_date": "2025-03-25",
           "assets": {
-            "directory": "./dist"
+            "directory": "./dist",
+            "not_found_handling": "404-page" // This option tells Workers to serve the 404.html
```
sarah11918: Thanks for the confirmation, Alex!

So it looks like the content is added in the correct place, but we need to be sure that those two conditions are met, otherwise this code sample will not work if someone copy and pastes it into their project.

So my original concern is still valid: what if someone has not created a custom 404 page? This advice is then not true in the basic/general case.

We could consider something like:

```suggestion
            "not_found_handling": "404-page" // If you have a custom `src/pages/404.astro` page
```
which sounds like it would technically not be correct code if they did not have a custom page, but at least the comment would make it clear what this line is doing and someone could find and remove it.

What does everyone think about this?

sarah11918: Hi @tobynguyen27 ! I'm "unresolving" this because "resolve" is like closing and ignoring. And, we can't ignore this!  We are still trying to figure out what to do here! 😄

sarah11918: @alexanderniebuhr Do you think this change would be OK/helpful to add here?

alexanderniebuhr: Yes I think it is a good idea :)

yanthomasdev (approved):
LGTM! Welcome to Team Docs @tobynguyen27 🙌

PR #11947, opened 2025-06-24
feat: node disable streaming option

florian-lefebvre:
#### Description (required)

Documents the upcoming `experimentalDisableStreaming` node adapter option

- Closes #
- Suggested label: 

https://github.com/withastro/astro/pull/14012

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/node.mdx
+++ b/src/content/docs/en/guides/integrations-guide/node.mdx
@@ -132,6 +132,28 @@ export default defineConfig({
 });
 ```
 
+### `experimentalStaticHeaders`
+
+<p>
+  **Type:** `boolean` <br />
+  **Default:** `false`<br />
+  <Since v="9.3.0"  pkg="@astrojs/node"/>
+</p>
+
+If enabled, the adapter will serve the headers of prerendered pages using the `Response` object when provided by Astro features, such as Content Security Policy.
+
+For example, when [experimental Content Security Policy](/en/reference/experimental-flags/csp/) is enabled, `experimentalStaticHeaders` can be used to add the CSP headers to the `Response` object instead of creating a `<meta>` element:
+
+```js title="astro.config.mjs" {9}
+  experimental: {
+    csp: true
+  },
+  adapter: node({
+    mode: 'standalone',
+    experimentalStaticHeaders: true,
+  })
+});
+```
 
 ## Usage
 
--- a/src/content/docs/en/guides/integrations-guide/netlify.mdx
+++ b/src/content/docs/en/guides/integrations-guide/netlify.mdx
@@ -327,6 +327,39 @@ export default defineConfig({
 });
 ```
 
+## Experimental features
+
+The following features are also available for use, but may be subject to breaking changes in future updates. Please follow the [`@astrojs/netlify` CHANGELOG](https://github.com/withastro/astro/tree/main/packages/integrations/netlify/CHANGELOG.md) carefully for updates if you are using these features in your project.
+
+### `experimentalStaticHeaders`
+
+<p>
+  **Type:** `boolean` <br />
+  **Default:** `false`<br />
+  <Since v="6.4.0"  pkg="@astrojs/netlify"/>
+</p>
+
+Enables specifying custom headers for prerendered pages in Netlify's configuration.
+
+If enabled, the adapter will save [static headers in the Framework API config file](https://docs.netlify.com/frameworks-api/#headers) when provided by Astro features, such as Content Security Policy.
+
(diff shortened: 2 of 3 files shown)


src/content/docs/en/guides/integrations-guide/node.mdx
```diff
@@ -106,6 +107,26 @@ export default defineConfig({
 });
 ```
 
+### `experimentalDisableStreaming`
+
+**Type:** `boolean` <br />
+<Since v="9.3.0" pkg="@astrojs/node" />
+
+Allow disabling [HTML streaming](/en/guides/on-demand-rendering/#html-streaming). This is useful for example if there are constraints from your host:
```
yanthomasdev:
```suggestion
Allows disabling [HTML streaming](/en/guides/on-demand-rendering/#html-streaming). This is useful, for example, if there are constraints from your host:
```

src/content/docs/en/guides/integrations-guide/node.mdx
```diff
@@ -106,6 +107,26 @@ export default defineConfig({
 });
 ```
 
+### `experimentalDisableStreaming`
+
+**Type:** `boolean` <br />
+<Since v="9.3.0" pkg="@astrojs/node" />
```
ArmandPhilippot, line 114: Without the `p` tag, there is no margins before the next line:
```suggestion
<p>
**Type:** `boolean` <br />
<Since v="9.3.0" pkg="@astrojs/node" />
</p>
```

src/content/docs/en/guides/integrations-guide/node.mdx
```diff
@@ -106,6 +107,28 @@ export default defineConfig({
...
 ```
 
+### `experimentalDisableStreaming`
+
+<p>
+**Type:** `boolean` <br />
+<Since v="9.3.0" pkg="@astrojs/node" />
+</p>
+
+Allows disabling [HTML streaming](/en/guides/on-demand-rendering/#html-streaming). This is useful, for example, if there are constraints from your host:
```
sarah11918: I didn't see "constraints from your host" in the original discussion as an advantage, and I'm not honestly sure what these might be. Should we be mentioning async components as a common use case for which these are an advantage? It would also be helpful if we knew exactly what problem this solves so people can be sure when NOT to do this: e.g. is this a thing that will help in some situations, but come at some other cost? We should help people decide if this is something they want/need.

sarah11918 (commented):
Looks great everyone! Thanks for working out all the details! 🙌 

Florian, I've left one comment because while this explains what it *does*, I think we could be more helpful in explaining when you would (or importantly, would not!) want this!

src/content/docs/en/guides/integrations-guide/node.mdx
florian-lefebvre:
```suggestion
Allows disabling [HTML streaming](/en/guides/on-demand-rendering/#html-streaming). This is useful, for example, if your host supports caching at the CDN level. This usually requires specific headers that streaming prevents setting (eg. `Content-Length`).

Keep in mind that opting out of streaming means you loose all of  its benefits, eg. sending HTML as fast as possible:
```

JusticeMatthew:
Provided feedback on T&D

florian-lefebvre:
That's because I force pushed, thanks for flagging!

src/content/docs/en/guides/integrations-guide/node.mdx
```diff
@@ -106,6 +107,30 @@ export default defineConfig({
...
+### `experimentalDisableStreaming`
+
+<p>
+**Type:** `boolean` <br />
+<Since v="9.3.0" pkg="@astrojs/node" />
+</p>
+
+Astro [streams HTML](/en/guides/on-demand-rendering/#html-streaming) by default, which helps with performance by allowing the user to see parts of the page as soon as possible. In most cases, you should keep using this behavior.
+
+However, you may have to disable HTML streaming, for example, if your host only supports non streamed HTML caching at the CDN level. You can opt out of the default behavior:
```
sarah11918:
```suggestion
Disables Astro's default [HTML streaming](/en/guides/on-demand-rendering/#html-streaming) for pages rendered on demand.

HTML streaming helps with performance and generally provides a better visitor experience. In most cases, disabling streaming is not recommended.

However, when you need to disable HTML streaming (e.g. your host only supports non-streamed HTML caching at the CDN level), you can opt out of the default behavior:
```
The extra context here is great! 🙌 

We still *do* want to make sure we start with a definition/explanation **of the feature itself**, though. So I've reorganized slightly. Plus, with the link describing what HTML streaming is/how it works, we don't need to duplicate with too much extra wordiness here with exactly *how* it's good, I think!

See what you think of this!

sarah11918 (approved):
LGTM! 🚀

PR #11955, opened 2025-06-28
Update framework-components.mdx

JusticeMatthew:
#### Description (required)

Informs users that additional configuration is required when mixing jsx components and provides a link to the related page

See discussion #11742 for more info

- Closes #11742

### DIFF
--- a/src/content/docs/en/guides/framework-components.mdx
+++ b/src/content/docs/en/guides/framework-components.mdx
@@ -84,7 +84,7 @@ There are several hydration directives available for UI framework components: `c
 
 ## Mixing frameworks
 
-You can import and render components from multiple frameworks in the same Astro component.
+You can import and render components from multiple frameworks in the same Astro component. [Additional configuration](https://docs.astro.build/en/guides/integrations-guide/react/#combining-multiple-jsx-frameworks) is required when mixing JSX frameworks.
 
 ```astro title="src/pages/mixing-frameworks.astro"
 ---


src/content/docs/en/guides/framework-components.mdx
```diff
@@ -84,7 +84,7 @@ There are several hydration directives available for UI framework components: `c
 
 ## Mixing frameworks
 
-You can import and render components from multiple frameworks in the same Astro component.
+You can import and render components from multiple frameworks in the same Astro component. [Additional configuration](/en/guides/integrations-guide/react/#combining-multiple-jsx-frameworks) is required when mixing JSX frameworks.
```
sarah11918:
```suggestion
You can import and render components from multiple frameworks in the same Astro component. 
```

I think we'd rather not clutter up the basic "here's how it works/what it looks like" before the code snippet, so we can quickly get this major point across.

I'd suggest something *between* the snippet and the caution, maybe like:

```
Astro will recognize and render your component based on its file extension. To distinguish between frameworks that use the same file extension, [additional configuration when rendering multiple JSX frameworks](/en/guides/integrations-guide/react/#combining-multiple-jsx-frameworks) (e.g. React and Preact) is required.
```
(also, one of the reasons for my rewording is thinking about what text is linked: "additional configuration" isn't as revealing about what the link is for as e.g. "additional configuration when rendering multiple JSX framworks" -- I always think about it as, what if the link text were my navigation/table of contents. Does it clearly tell me what is at the link the same way a section header should tell me what content is contained within that section?)

What do you think?

sarah11918 (commented):
Thank you for jumping on this issue! I think this is important and helpful! 🙌 

See below for my specific comments!

src/content/docs/en/guides/framework-components.mdx
JusticeMatthew: I think that's a great way to think about a link, very purpose driven design 😄 

Will update it 👍

sarah11918: @JusticeMatthew Are you still up for updating this? (I don't see another commit)

JusticeMatthew: Yes sorry, fell through the cracks with my computer dying 😅 (PC is repaird now)

Will update it today

JusticeMatthew:
Sorry for the delay @sarah11918 !

Suggested content added 👍

yanthomasdev (approved):

src/content/docs/en/guides/framework-components.mdx
```diff
@@ -84,7 +84,7 @@ There are several hydration directives available for UI framework components: `c
 
 ## Mixing frameworks
 
-You can import and render components from multiple frameworks in the same Astro component.
+You can import and render components from multiple frameworks in the same Astro component.
```
sarah11918:
```suggestion
You can import and render components from multiple frameworks in the same Astro component.
```

PR #11979, opened 2025-07-03
Fix Firefox sidebar content flash

### LINKED ISSUE #11611: Hidden tabs' contetnt flash in FF 137
### 📝 Issue Description

On page navigation, `<Sidebar>` flashes with all tabs

https://docs.astro.build/en/basics/astro-pages/
... practically all pages without splash layout

### 🤔 Expected Behavior

Non-active tabs should hide content and only show it when tab is clicked

### 👀 Current Behavior

https://github.com/user-attachments/assets/a188120b-670a-4bf6-b9e8-97e01615f572

### 🖥️ Browser

Firefox 137

### 📄 Additional Information

https://github.com/withastro/docs/blob/0b53033482167b75e1503252242084f4df5e213d/src/components/tabs/TabPanel.astro#L27-L40

HiDeoo:
As this was brought up in today's T&D, this PR fixes a flash of invalid active tab in both the sidebar and the tutorial tracker (they both use the same component).

After investigating, it looks like Firefox refuses to animate when the `from` keyframe [only contains the `display: none;` property](https://github.com/withastro/docs/blob/2bf83d22e659c3acec7d46393c63a19c8ec7d8e1/src/components/tabs/TabPanel.astro#L37-L39) (even tho it should use a discrete animation type).

This PR adds `content-visibility: hidden;` to the `from` keyframe, which seems to trigger the animation correctly in Firefox and does not seem to change any other behavior.

- Closes #11611

### DIFF
--- a/src/components/tabs/TabPanel.astro
+++ b/src/components/tabs/TabPanel.astro
@@ -36,6 +36,7 @@ const attributes: HTMLAttributes<'div'> = initial ? { 'data-initial': 'true' } :
 	@keyframes tab-panel-appear {
 		from {
 			display: none;
+			content-visibility: hidden;
 		}
 	}
 </style>
--- a/src/components/tabs/TabPanel.astro
+++ b/src/components/tabs/TabPanel.astro
@@ -35,6 +35,8 @@ const attributes: HTMLAttributes<'div'> = initial ? { 'data-initial': 'true' } :
 
 	@keyframes tab-panel-appear {
 		from {
+			/* `content-visibility` is set as well as `display` to work around a Firefox
+			   bug where animations containing only `display: none` don’t play. */
 			display: none;
 			content-visibility: hidden;
 		}


HiDeoo:
Here is a video showing me spamming `Cmd+R` on the live version and then on the deploy preview:


https://github.com/user-attachments/assets/2a48cb0e-51b1-4690-bda8-1d0075f7c532

src/components/tabs/TabPanel.astro
```diff
@@ -36,6 +36,7 @@ const attributes: HTMLAttributes<'div'> = initial ? { 'data-initial': 'true' } :
 	@keyframes tab-panel-appear {
 		from {
 			display: none;
+			content-visibility: hidden;
```
delucis, line 41: _Might_ be good to comment this given having both of these does look a bit strange if you’re not familiar/have forgotten the reasons?

```suggestion
			/* `content-visibility` is set as well as `display` to work around a Firefox
			   bug where animations containing only `display: none` don’t play. */
			display: none;
			content-visibility: hidden;
```

delucis (approved):
Awesome work investigating this and finding a very light touch fix @HiDeoo!

Tested vigorously in Firefox on my end and it works perfectly where the bug is clearly visible on the live site.

src/components/tabs/TabPanel.astro
HiDeoo: Excellent call, totally skipped my mind when I decided to push the fix and test in a real deploy preview :raised_hands:

PR #12012, opened 2025-07-10
Tutorial custom page title

HiDeoo:
#### Description (required)

Worked on during today's Talking & Doc'ing, this PR addresses an issue where some pages could end up having the same title, which is not ideal for SEO and user experience. 

This PR updates the titles of all the tutorial pages to be automatically prefixed by a translatable UI string (`tutorial.title.prefix`) using a route data middleware.

Tutorial pages can still manually specify their titles for cases where the automatic prefix is not desired using the `head` frontmatter property.

```md
---
type: tutorial
unitTitle: Something
title: Page title
head:
  - tag: title
    content: Custom title | Docs
---
```

- Closes #11766 (will need to be manually closed after merge as auto-closing does not work for discussions)


anaxite:
Came here from Talking & Doc'ing.

Quick summary of thoughts:
- I find that greather-than symbols really work for breadcrumbs.
- Colons or brackets in titles? Either work, but I found that colons are nice.

sarah11918:
Strong preference for colons! 😅

src/content/docs/en/tutorial/1-setup/index.mdx
```diff
@@ -8,6 +8,9 @@ description: >-
   Prepare your development environment, and create and deploy your first Astro
   site
 i18nReady: true
+head:
+  - tag: title
+    content: 'Build a blog tutorial: Unit 1 - Setup | Docs'
```
sarah11918, line 13: Just a quick note that we want this done for **all six** of the tutorial units (The check in page only) to match this format! (Remove "Check in:" and replace it with "Build a blog tutorial:")

If you're reading this PR, you could be contributing those changes!

HiDeoo:
Updated the PR and its description to reflect everything we talked about during T&D.

Something coming to mind is that once we're happy with the English version, I wonder if it could be worth to not immediately merge this PR, but instead ping translators to potentially translate the new UI string/titles in as many languages as possible before merging so search engines cannot end up indexing pages with an updated tutorial page title without a translated prefix? (I guess another possibility could also to tweak the code so that the automatic prefixing is applied only if the new UI string is available in the content language?)

Any thoughts on that?

delucis:
> another possibility could also to tweak the code so that the automatic prefixing is applied only if the new UI string is available in the content language?

That’s not a bad idea I’d say. We can use [`t.exists()`](https://starlight.astro.build/guides/i18n/#texists) for that right?

I think that’s preferable and then each language can translate in their own time.

sarah11918:
That idea sounds good! Was also thinking that in the worst case, for the tutorials that are already translated, we should *have* "Build a blog tutorial" in their language somewhere, and could add that in ourselves for the general case update.

But obviously prefer to let the translators in on the fun, and the solution that allows them to contribute as they are able seems ideal!

sarah11918:
Verified all the new page (tab) titles in Chrome are correct for English, and working as expected in French (e.g. `Build a blog tutorial: Embarquement: Unité 1 - Configuration | Docs`)

HiDeoo:
Perfect, updated the PR to include this new logic. I also added a quick French translation for the prefix, so we can test it out.

I'll note that this is the first translation that came to mind, so either we should remove it once we're done testing, or we should spend more than 5 seconds thinking about it and see what would be the best as a page title prefix for the tutorial in French (cc @ArmandPhilippot).

src/content/i18n/fr.yml
```diff
@@ -51,6 +51,7 @@ progress.done: Terminer
 # Tutorial Navigation
 tutorial.trackerLabel: Suivi du tutoriel
 tutorial.unit: Unité
+tutorial.title.prefix: "Tutoriel de création d'un blog : {{title}}"
```
ArmandPhilippot, line 54: The only alternative that comes to my mind - if we want to shorten it - would be `Tutoriel Créer un blog`. But without quotes around the tutorial title (ie. `Créer un blog`) it feels a bit odd.
So, I think your translation is just fine!

ArmandPhilippot (commented):
I checked the tutorial both in English and French (and somes pages in Spanish to see the difference) and this looks great! Good job 🙌🏽

sarah11918:
OK, so noting that now:

- English (canonical) ✅ 
- French (a translated tutorial that does have a value for `tutorial.title.prefix`) ✅ 
- German, Arabic etc: (a non-translated tutorial that does not have a value for `tutorial.title.prefix`) ✅  (these are completely English)
- Brazilian Portuguese, Russian etc: (a translated tutorial that does NOT have a value for `tutorial.title.prefix`) ❌
Screenshot 2025-07-11 08 49 22
These appear to be missing the new prefix entirely

src/content/i18n/fr.yml
vrabe: Isn't the English one the same? `Build a blog tutorial` and `"Build a blog" tutorial`.

HiDeoo:
Thanks for testing, if I'm not missing anything, this matches our latest requirements.

To give more details about the last point, this is expected to avoid a search engine temporarily indexing something like `Build a blog tutorial: Sobre este Tutorial | Docs`, so the current title `Sobre este Tutorial | Docs` will be used until a translated prefix is available.

sarah11918:
Gotcha! Then, looks like all is working as expected. Amazing job!

src/content/i18n/fr.yml
ArmandPhilippot: I don't know how it sounds for native English speakers while it sounds okay to me. However, in French I don't have the same feeling... maybe because of the words order...
I would write it `Tutoriel « Créer un blog » :` (capital letter for `tutoriel` because this is the beginning of the title, and for `créer` because we're quoting another title). Using French quotes seems odd in a `<title>` tag. I think HiDeoo's translation is more appropriate.

But that's only my opinion! So, I left this comment in case I was overthinking and that HiDeoo (or anaxite which is French too I believe?) had a different opinion on the matter! 😄

yanthomasdev (approved):
LGTM, amazing work! 🚀

PR #12022, opened 2025-07-14
Reference `<ClientRouter />` component rather than view transitions on Prefetch page

### LINKED ISSUE #12021: The Prefetch page should refer to the `<ClientRouter>` component rather than the concept of view transitions.
### 📚 Subject area/topic

Prefetch

https://docs.astro.build/en/guides/prefetch/#browser-support

The section states "When you use [View Transitions](https://docs.astro.build/en/guides/view-transitions/) on a page, prefetching will also be enabled by default".

This might be misleading, as prefetching is not enabled by using view transitions but by using the `<ClientRouter />` component.

_No response_

martrapp:
#### Description (required)

Astro´s `<ClientRouter />` component enables Astro´s prefetch feature by default.
Saying that _View Transitions_ set this default might be misleading.

- Closes #12021

### DIFF
--- a/src/content/docs/en/guides/prefetch.mdx
+++ b/src/content/docs/en/guides/prefetch.mdx
@@ -170,7 +170,7 @@ Make sure to only import `prefetch()` in client-side scripts as it relies on bro
 
 ## Using with View Transitions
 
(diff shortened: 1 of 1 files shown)


martrapp:
Wow, that was quick! Thank you, Sarah.

PR #12023, opened 2025-07-14
feat: experimental raw env values flag

florian-lefebvre:
#### Description (required)

Adds docs for `experimental.rawEnvValues`

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
+++ b/src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
@@ -1,0 +1,45 @@
+---
+title: Experimental raw environment variables values
+sidebar:
+  label: Raw environment variables values
+i18nReady: true
+---
+
+import Since from '~/components/Since.astro'
+
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `false`<br />
+<Since v="5.12.0" />
+</p>
+
+The `experimental.rawEnvValues` flag disables coercion of `import.meta.env` values that are populated from `process.env`.
+
+To disable coercion, set the flag to `true` in your Astro configuration:
+
+```js title="astro.config.mjs" ins={4-6}
+import { defineConfig } from "astro/config"
+
+export default defineConfig({
+  experimental: {
+    rawEnvValues: true,
+  }
+})
+```
+
+## Usage
+
+Currently, if an environment variable used through `import.meta.env` references a `process.env` variable whose value is `"true"`, `"false"`, `"1"`, or `"0"`, it will be coerced to respectively `true`, `false`, `1` and `0`. For example:
+
+- If `import.meta.env.FOO` is called and the project root contains a `.env` file with `FOO=false`, the value will be `"false"` (string)
+- If `import.meta.env.FOO` is called and the dev server is started with `FOO=false astro dev`, the value will be `false` (boolean)
+
+This behavior is a holdover from when Astro supported dynamic prerender values in routes, which was [removed in Astro v5](/en/guides/upgrade-to/v5/#removed-support-for-dynamic-prerender-values-in-routes).
+
+
+This experimental flag will keep the string value instead of turning into a boolean or a number. This aligns `import.meta.env`'s behavior in Astro with [Vite](https://vite.dev/guide/env-and-mode.html#env-variables).
+
+In a future major version, Astro will switch to not coercing `import.meta.env` values by default, but you can opt in to the future behavior early using the `experimental.rawEnvValues` flag.
+
+If you need coercion in Astro, we recommend you use [`astro:env`](/en/guides/environment-variables/).
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -141,13 +141,14 @@ export const sidebar = [
 			group('reference.experimental', {
 				items: [
 					'reference/experimental-flags',
+					'reference/experimental-flags/csp',
 					'reference/experimental-flags/fonts',
+					'reference/experimental-flags/live-content-collections',
 					'reference/experimental-flags/client-prerender',
 					'reference/experimental-flags/content-intellisense',
 					'reference/experimental-flags/preserve-scripts-order',
 					'reference/experimental-flags/heading-id-compat',
-					'reference/experimental-flags/csp',
-					'reference/experimental-flags/live-content-collections',
+					'reference/experimental-flags/raw-env-values',
 				],
 			}),
 			'reference/legacy-flags',


src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
```diff
@@ -0,0 +1,45 @@
...
+import Since from '~/components/Since.astro'
+
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `false`<br />
+<Since v="5.12.0" />
+</p>
+
+The `experimental.rawEnvValues` flag disables coercion of `import.meta.env` values that are populated from `process.env`.
```
sarah11918:
```suggestion
Astro allows you to configure a [type-safe schema for your environment variables](/en/guides/environment-variables/#type-safe-environment-variables), and converts variables imported via `astro:env` into the expected type.

However, Astro also converts your environment variables used through `import.meta.env` in some cases, and this can prevent access to some values such as the strings `"true"` (which is converted to a boolean value), and `"1"` (which is converted to a number).

The `experimental.rawEnvValues` flag disables coercion of `import.meta.env` values that are populated from `process.env`, allowing you to use the raw value.
```
Once again making things up, but can we have an introduction maybe something like this?

src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+To disable coercion, set the flag to `true` in your Astro configuration:
```
sarah11918:
```suggestion
To disable Astro's coercion on values used through `import.meta.env`, set the `experimental.rawEnvValues` flag to `true` in your Astro configuration:
```

src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
```diff
@@ -0,0 +1,45 @@
...
+- If `import.meta.env.FOO` is called and the dev server is started with `FOO=false astro dev`, the value will be `false` (boolean)
+
+This behavior is a holdover from when Astro supported dynamic prerender values in routes, which was [removed in Astro v5](/en/guides/upgrade-to/v5/#removed-support-for-dynamic-prerender-values-in-routes).
+
+
+This experimental flag will keep the string value instead of turning into a boolean or a number. This aligns `import.meta.env`'s behavior in Astro with [Vite](https://vite.dev/guide/env-and-mode.html#env-variables).
+
+In a future major version, Astro will switch to not coercing `import.meta.env` values by default, but you can opt in to the future behavior early using the `experimental.rawEnvValues` flag.
+
+If you need coercion in Astro, we recommend you use [`astro:env`](/en/guides/environment-variables/).
```
sarah11918:
``````suggestion
Enabling this experimental flag will no longer convert string values into booleans or numbers. This aligns `import.meta.env`'s behavior in Astro with [Vite](https://vite.dev/guide/env-and-mode.html#env-variables).

In a future major version, Astro will switch to not coercing `import.meta.env` values by default, but you can opt in to the future behavior early using the `experimental.rawEnvValues` flag and if necessary, [updating your project](#updating-your-project) accordingly.

### Updating your project

If you were relying on this coercion, you may need to update your project code to apply it manually:

```ts title="src/components/MyComponent.astro" del={1} ins={1}
const enabled: boolean = import.meta.env.ENABLED
const enabled: boolean = import.meta.env.ENABLED === "true"
```

If you need coercion in Astro, we recommend you use [`astro:env`](/en/guides/environment-variables/).
``````
How about something like this?

I don't think we need to go too far into a history lesson here. Rather, we can show what people might need to do to account for this change. This also helps us get a head start for when this becomes the default and we need to write the migration guide!

File: src/content/docs/en/reference/experimental-flags/raw-env-values.mdx [hunk 4]
```diff
@@ -0,0 +1,45 @@
+---
+title: Experimental raw environment variables values
+sidebar:
+  label: Raw environment variables values
```
sarah11918:
```suggestion
  label: Raw environment variables
```
For the label, this might be a little long? Would this still be meaningful enough?

sarah11918 (commented):
Here are my initial thoughts, Florian! See what you think/can do with this!

src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
florian-lefebvre: FYI I matched the structure of the heading compat id flag, but that's also fine by me!

src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+## Usage
+
+Enabling this experimental flag will no longer convert string values into booleans or numbers. This aligns `import.meta.env`'s behavior in Astro with [Vite](https://vite.dev/guide/env-and-mode.html#env-variables).
+
+In a future major version, Astro will switch to not coercing `import.meta.env` values by default, but you can opt in to the future behavior early using the `experimental.rawEnvValues` flag and if necessary, [updating your project](#updating-your-project) accordingly.
+
+### Updating your project
+
+If you were relying on this coercion, you may need to update your project code to apply it manually:
```
sarah11918, line 43: This example is great! Just checking though, you would *not* expect people to be updating any environment variables themselves? Are these things they would typically not have access to, so the real workaround is in fixing the logic to use the raw value?

florian-lefebvre: Yeah they have no control over the env var they receive, it will always be a string. So the only way really is to update were it's handled in code

yanthomasdev (approved):

adamhl8:
I think the mention of how this applies to variables "that are populated from `process.env`" is potentially confusing. 

Users who use something like Bun (which automatically loads `.env` files), [dotenvx](https://dotenvx.com/docs/package-managers/npm#astro), etc, will see the coercion behavior regardless.

Also, values from `.env` files *do* end up in `process.env`. It's just that Astro *only* coerces values that exist in the environment of the parent process (and therefore not values in `.env` files).

Correct me if I'm wrong, but this is the case because `envLoader` runs before Vite loads environment variables from `.env` files(?)

florian-lefebvre:
I think this mentions means exactly what the code is doing. The `fullEnv` comes from vite's `loadEnv`, which loads vars from `.env` files and `process.env`. The coercion logic only happens if the key is also present in `process.env`, hence this mention.

So values in `.env` do not end up in `process.env`

adamhl8:
Ah gotcha, thanks.

I thought that since Vite uses `dotenv` under the hood, it was populating `process.env` as well, but they only use it for [parsing](https://github.com/vitejs/vite/blob/6bc8bf634d4a2c9915da9813963dd80a4186daeb/packages/vite/src/node/env.ts#L50C20-L50C21) `.env` files.

Which is also why this was confusing to me because in my case `process.env` is being populated by Bun (and I thought *also* vite). So yeah makes sense; I think it's fair for the docs to assume "standard" tooling.

PR #12029, opened 2025-07-15
adds TOML support for built-in loaders

sarah11918:
#### Description (required)

Accompanies the new feature that adds TOML support to the `file()` and `glob()` loaders.

This PR:
- updates all lists of supported content collection entry file types to include TOML
- removes an example of writing a custom parser function to use TOML as a content collection entry

Considerations:
- There isn't really a great spot to mention that this is since v5.12? We could update lists to say e.g.:

> Collections stored locally in your project or on your filesystem can have entries of Markdown, MDX, Markdoc, YAML, TOML (added in `v5.12`), or JSON files:

### DIFF
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -142,13 +142,14 @@ export const sidebar = [
 			group('reference.experimental', {
 				items: [
 					'reference/experimental-flags',
+					'reference/experimental-flags/csp',
 					'reference/experimental-flags/fonts',
+					'reference/experimental-flags/live-content-collections',
 					'reference/experimental-flags/client-prerender',
 					'reference/experimental-flags/content-intellisense',
 					'reference/experimental-flags/preserve-scripts-order',
 					'reference/experimental-flags/heading-id-compat',
-					'reference/experimental-flags/csp',
-					'reference/experimental-flags/live-content-collections',
+					'reference/experimental-flags/raw-env-values',
 				],
 			}),
 			'reference/legacy-flags',
--- a/src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
+++ b/src/content/docs/en/reference/experimental-flags/raw-env-values.mdx
@@ -1,0 +1,50 @@
+---
+title: Experimental raw environment variables values
+sidebar:
+  label: Raw environment variables
+i18nReady: true
+---
+
+import Since from '~/components/Since.astro'
+
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `false`<br />
+<Since v="5.12.0" />
+</p>
+
+Astro allows you to configure a [type-safe schema for your environment variables](/en/guides/environment-variables/#type-safe-environment-variables), and converts variables imported via `astro:env` into the expected type.
+
+However, Astro also converts your environment variables used through `import.meta.env` in some cases, and this can prevent access to some values such as the strings `"true"` (which is converted to a boolean value), and `"1"` (which is converted to a number).
+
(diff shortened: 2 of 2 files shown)


aditsachde:
Thanks so much for making these updates, I think this looks good! Only note I have is that I updated the [changeset in my PR](https://github.com/withastro/astro/blob/02281a21142bc816f556520545fe844221f3f681/.changeset/icy-keys-tickle.md) to include a note about how the `file()` loader interacts with some quirks of the TOML format. I'm not sure if it's worth including in the docs – it seems intuitive enough to me, but I've also spent a lot of time staring at TOML, so I'm not a good judge there.

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -122,35 +122,17 @@ export const collections = { blog, dogs, probes };
 
 ##### `parser` function
 
-The `file()` loader accepts a second argument that defines a `parser` function. This allows you to specify a custom parser (e.g. `toml.parse` or `csv-parse`) to create a collection from a file's contents.
+The `file()` loader accepts a second argument that defines a `parser` function. This allows you to specify a custom parser (e.g. `csv-parse`) to create a collection from a file's contents.
 
-The `file()` loader will automatically detect and parse a single array of objects from JSON and YAML files (based on their file extension) with no need for a `parser` unless you have a [nested JSON document](#nested-json-documents). To use other files, such as `.toml` and `.csv`, you will need a to create a parser function.
+The `file()` loader will automatically detect and parse a single array of objects from JSON, YAML, and TOML files (based on their file extension) with no need for a `parser` unless you have a [nested JSON document](#nested-json-documents). To use other files, such as `.csv`, you will need a to create a parser function.
```
```suggestion
The `file()` loader will automatically detect and parse (based on their file extension) a single array of objects from JSON and YAML files, and will treat each top level table as an independent entry in TOML files. Support for these file types is built-in, and there is no need for a `parser` unless you have a [nested JSON document](#nested-json-documents). To use other files, such as `.csv`, you will need to create a parser function.
```
What about something like this, @aditsachde ??

sarah11918: OK, going with this so the feature can merge!

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -122,35 +122,17 @@ export const collections = { blog, dogs, probes };
 
 ##### `parser` function
 
-The `file()` loader accepts a second argument that defines a `parser` function. This allows you to specify a custom parser (e.g. `toml.parse` or `csv-parse`) to create a collection from a file's contents.
+The `file()` loader accepts a second argument that defines a `parser` function. This allows you to specify a custom parser (e.g. `csv-parse`) to create a collection from a file's contents.
 
-The `file()` loader will automatically detect and parse a single array of objects from JSON and YAML files (based on their file extension) with no need for a `parser` unless you have a [nested JSON document](#nested-json-documents). To use other files, such as `.toml` and `.csv`, you will need a to create a parser function.
+The `file()` loader will automatically detect and parse (based on their file extension) a single array of objects from JSON and YAML files, and will treat each top level table as an independent entry in TOML files. Support for these file types is built-in, and there is no need for a `parser` unless you have a [nested JSON document](#nested-json-documents). To use other files, such as `.csv`, you will need to create a parser function.
```
yanthomasdev:
```suggestion
The `file()` loader will automatically detect and parse (based on their file extension) a single array of objects from JSON and YAML files, and will treat each top-level table as an independent entry in TOML files. Support for these file types is built-in, and there is no need for a `parser` unless you have a [nested JSON document](#nested-json-documents). To use other files, such as `.csv`, you will need to create a parser function.
```

PR #12045, opened 2025-07-17
fix: Correct env.d.ts example

nzakas:
#### Description (required)

The example `env.d.ts` file in the docs doesn't actually work. I spent a lot of time tracking down the correct format.

- Refs https://github.com/withastro/astro/issues/7394
- Suggested label: bug

### DIFF
--- a/src/content/docs/en/guides/middleware.mdx
+++ b/src/content/docs/en/guides/middleware.mdx
@@ -147,12 +147,14 @@ type User = {
   name: string;
 };
 
-declare namespace App {
-  interface Locals {
-    user: User;
-    welcomeTitle: () => string;
-    orders: Map<string, object>;
-    session: import("./lib/server/session").Session | null;
+declare global {
+  namespace App {
+    interface Locals {
+      user: User;
+      welcomeTitle: () => string;
+      orders: Map<string, object>;
+      session: import("./lib/server/session").Session | null;
+    }
   }
 }
 ```


ArmandPhilippot:
Hi, and thanks for taking the time to contribute!

`declare global` shouldn't be required unless you're using top-level imports/exports in your `.d.ts` file. This why this example shows an inline import: in ambient modules, the types should be imported with a dynamic import rather than a top-level one to avoid "breaking" them. The issue you've linked suggest the same thing: https://github.com/withastro/astro/issues/7394#issuecomment-1896607172

When you use `declare global`, this is no longer an ambient module, and you could do the same in a regular `.ts`, alongside your code.

I must admit this behavior is a bit [hidden in Typescript docs](https://www.typescriptlang.org/docs/handbook/modules/reference.html#ambient-modules).

We have an ongoing discussion in the docs repo (#11759) because many people are confused by that if you're interested.

Are you sure you don't have any top-level import/export in your `.d.ts` file? Last time I checked the example was working as expected, but maybe something changed... or there are edge cases I'm not aware of.

nzakas:
I totally missed that we can't have top-level imports if the existing example is to work. Thanks for pointing that out.

I think this subtlety is pretty confusing, especially when most people just browse docs for a working example, modify it, then find it doesn't work. This is what happened to me and it took me way too much time to track down what apparently now wasn't even the right answer. The difference between a regular and ambient module is non-obvious, and honestly, I'm not sure what this behavior buys developers aside from extra cognitive overhead.

At a minimum, it seems like there should be some default ESLint rule shipped with Astro that catches this kind of thing. 

Thanks for the link. I don't have have enough time to continue investigating this path, I just hope something can be done to improve this experience.

PR #12054, opened 2025-07-20
i18n(ja): update installation steps in `1-setup/2.mdx`

lef237:
#### Description (required)

I fixed the Japanese document because its content was older than the English document. 

I also confirmed that the number of prompts has been reduced by actually testing it.

Reference: https://github.com/withastro/docs/blob/main/src/content/docs/en/tutorial/1-setup/2.mdx#launch-the-astro-setup-wizard


jp-knj (approved):
LGTM, Thank!

yanthomasdev (approved):
LGTM! Thanks and welcome to Team Docs!

PR #12055, opened 2025-07-20
Feat: Filter current page from BackendGuidesNav

2u841r:
#### Description (required)

What: Remove the current page from the backend services navigation component to prevent showing the page the user is already on.

Why: When viewing a specific backend service page (e.g., /guides/backend/turso/), it's illogical to show that same service in the navigation list. This improves UX by filtering out the current page and only showing other available backend services for navigation.

How: Added logic to extract the current page ID from the URL path and filter it out from the links array before passing to the CardsNav component.

Discord: 2u841r


sarah11918:
Hi! We have chosen to include all the services so that the grid looks the same no matter which page you are on.When the same services appear in the same place every time, it is less confusing as you might navigate through multiple options as you explore them.

2u841r:
then please consider adding active/selected indicator by css. thanks.

PR #12056, opened 2025-07-21
fix(content-loader-reference): update `store.set` usage in a snippet

ArmandPhilippot:
#### Description (required)

Someone has spotted in the core repo (https://github.com/withastro/astro/pull/14096) that the use of `store.set`  in the code snippet used for `renderMarkdown` was incorrect. They have corrected the changelog, but we should fix it here as well.
The [DataStore set function](http://localhost:4321/en/reference/content-loader-reference/#set) only takes one argument.


yanthomasdev (approved):
LGTM! Thanks!

PR #12061, opened 2025-07-21
docs: fix example domain and improve clarity in config guide

RezixDev:
- Corrected the example deployment domain from www.example/docs to www.example.com/docs
- Reworded the explanation about adding metadata to clarify the use of standard HTML <link> and <meta> tags

Suggested label: fix


sarah11918 (approved):
Thank you for these helpful fixes! :rocket:

PR #12063, opened 2025-07-22
Optimizely CMS Integration Information

jacobpretorius:
#### Description

- Add information about Optimizely CMS & Astro resources and sample implementations.

### DIFF
--- a/src/content/docs/en/reference/modules/astro-actions.mdx
+++ b/src/content/docs/en/reference/modules/astro-actions.mdx
@@ -78,7 +78,7 @@ If `input` is omitted, the `handler` will receive an input of type `unknown` for
 
 ##### Use with `accept: 'form'`
 
-If your action accepts form inputs, use the `z.object()` validator to automatically parse form data to a typed object. The following validators are supported for form data fields:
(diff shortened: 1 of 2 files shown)


sarah11918 (commented):
Thanks for your submission, @jacobpretorius !

There was a lot to tidy up and move around here, so it was easier for me to just do it in one quick pass vs a bunch of comments and suggestions!

Notably:

- I checked out Optimizely to see how they typically describe themeselves/their features. This is an example of about the level of detail we go into here, but you're the one who knows Optimizely, so you can adjust this to be most accurate/representative of the CMS as required!

- We prefer not to have several different links from one domain. There should be one, canonical entry point in most cases, so you'll have to select the best documentation link. Given the content of your community resources, it seems like the SaaS is what you have used, and have extra support for with the other resources, so I went with that one here.

- Similarly, since both your templates for static and server-rendered are housed within the same demo, it makes sense to link to it only once. The readme for the main project is way more helpful than the one at either of the project links, so it seems like a better entry point for people.

I would be happy to publish this now as is, but I want to wait for you to tell me whether there's now anything inaccurate here! Just confirm or suggest corrects, then we can get this ready for docs! :raised_hands:

jacobpretorius:
Hi @sarah11918, thanks for the help and revisions!

There's a lot of nuance with Optimizely products and especially the naming as the CMS itself is available in different configurations with different delivery channels. In the spirit of that, I've made a minor change to your update, please let me know if that's okay.

If it is, I think this is good to be merged.

yanthomasdev (approved):
LGTM! Thanks!

PR #12091, opened 2025-07-28
Client side scripts guide update

OliverSpeir:
#### Description (required)

Essentially the goal of this PR is to "modernize" this page, making it more focused and correct. 

We trim down some unnecessary examples, bring important information to the top of the page, and cut out some information that is no longer accurate and not really the docs' job to go over. 

A little background on the changes: 

Astro changed the way scripts work in v5 by "direct rendering" which inlines scripts and this page was written assuming that the scripts would always be externalized. Inline type module scripts do indeed block rendering, and instead of clarifying exactly how all that stuff works we cut most of it out because it's not astro specific behavior. 

These changes were discussed here: https://github.com/withastro/docs/discussions/11165

#### Additional notes from Sarah
- a few pages that link to now-changed section headers have been updated, and affects some translated pages, so we'll need to i18nIgnore those pages (**additional additional note from Yan**: in this case, a lunaria directive on merge to main! Here's the code: `@lunaria-track:src/content/docs/en/**/*.mdx`)
- Whatever we decide to call "opt out of processing/unprocessed scripts/inline scripts" will need to be updated on the View Transitions page (line 525) -- this should be the only link check that fails now


src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -123,6 +73,29 @@ To load scripts outside of your project's `src/` folder, include the `is:inline`
 <script is:inline src="https://my-analytics.com/script.js"></script>
 ```
 
+### Externalization
+
+When Astro builds your site for production deployment, it processes and minifies your scripts. When scripts are above 4kB in size, they will be made into external files using `<script type="module" src="...">`, while smaller ones are inlined directly as `<script type="module">` with the JavaScript code embedded. This approach provides a balance between the number of additional requests and the volume of JavaScript that can be cached between pages.
+
+You can configure the size at which scripts will be linked externally (in bytes) using the `assetsInlineLimit` vite build option. Note that this option affects stylesheet and image inlining as well.
```
OliverSpeir: > Note that this option affects stylesheet and image inlining as well

This wording comes from https://docs.astro.build/en/guides/styling/#bundle-control, I wasn't sure about duplicating it, and I also wasn't sure/aware that astro actually inlined images.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,70 +7,20 @@ i18nReady: true
 ---
 import ReadMore from '~/components/ReadMore.astro'
 
-You can add interactivity to your Astro components without [using a UI framework](/en/guides/framework-components/) like React, Svelte, Vue, etc. using standard HTML `<script>` tags. This allows you to send JavaScript to run in the browser and add functionality to your Astro components.
+You can add interactivity to your Astro components without [using a UI framework](/en/guides/framework-components/) like React, Svelte, Vue, etc. using `<script>` tags. This allows you to send JavaScript to run in the browser and add functionality to your components.
 
-## Client-Side Scripts
```
sarah11918, line 12: You've cut this entire section, but I'll note this section is very much in keeping with our docs style: we explain some use cases for the feature (in this case, script tags) and we show a basic example of what the feature *looks like* in Astro. Notably, I think showing adding an event listener is crucial here, because that is often when people realize, "Oh man, I gotta write vanilla JS!"

Maybe you feel the example is too long, or not realistic enough? (Maybe it should be a dark mode toggle? :joy:) But I do really think we need a short section *like this*, complete with example. 

Also open to a new section title if you don't like that, but one thing this particular heading does is emphasize the client-side nature of the page content, which helps with our [Astro component definition](https://docs.astro.build/en/basics/astro-components/) that "they are HTML-only templating components with no client-side runtime" -- which *could* lead people to think you'll never get ANY client-side interactivity with an Astro component. 

Will also note, the link to this page from that page's introduction uses the phrase "standard HTML script tags" which we should also update to match what's here!

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,70 +7,20 @@ i18nReady: true
...
-  // Add event listeners to fire confetti when a button is clicked.
-  buttons.forEach((button) => {
-    button.addEventListener('click', () => confetti());
-  });
-</script>
-```
-
-By default, Astro processes and bundles `<script>` tags, adding support for importing npm modules, writing TypeScript, and more.
-
-## Using `<script>` in Astro
```
sarah11918, line 36: Agree that this section feels unnecesary, if we've done a good job in the (reintroduced) section above!

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,70 +7,20 @@ i18nReady: true
...
-### Script processing
-
-By default, `<script>` tags are processed by Astro.
-
-- Any imports will be bundled, allowing you to import local files or Node modules.
-- The processed script will be injected at where it's declared with [`type="module"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).
-- TypeScript is fully supported, including importing TypeScript files.
-- If your component is used several times on a page, the script will only be included once.
-
-```astro title="src/components/Example.astro"
```
sarah11918, line 59: Will also note, this is *very* much Astro docs (Fred) style! The new bullet list looks fantastic, and is now our newer style. But I would tend to still keep this below (updated if necessary) in the interest of people who are just scanning.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,70 +7,20 @@ i18nReady: true
...
 
-The `type="module"` attribute makes the browser treat the script as a JavaScript module. This has several performance benefits:
-- Rendering is not blocked. The browser continues to process the rest of the HTML while the module script and its dependencies load.
-- The browser waits for HTML to be processed before executing module scripts. You do not need to listen for the "load" event.
-- `async` and `defer` attributes are unnecessary. Module scripts are always deferred.
-
-:::note
-The `async` attribute is valuable for normal scripts because it prevents them from blocking rendering. However, module scripts already have this behavior. Adding `async` to a module script will cause it to execute before the page has fully loaded. This is probably not what you want.
-:::
+- **TypeScript by default:** All scripts are processed as TypeScript.
```
sarah11918: List is fantastic, and updates are great!

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,70 +7,20 @@ i18nReady: true
...
+- **Import bundling:** Import local files or npm modules, which will be bundled together.
+- **Module injection:** The processed script is injected with [`type="module"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).
+- **Script deduplication:** If your component is used multiple times on a page, the script will only be included once.
+- **Automatic inlining:** If the script is small enough, Astro will inline it directly into the HTML to reduce the number of requests. See [Externalization](#externalization) for more details.
 
 ### Opting out of processing
```
sarah11918: Question: do you think this heading is the most helpful way to frame the content? Are people looking for/thinking "How can I opt-out of processing my scripts?" or are they more likely to encounter a script *not* being processed and asking *why* this is happening? 

Would something more like "Exceptions to script processing" (that is terrible, but something closer to that framing) serve both people who *want* to explicitly not have a script processed but also those who are trying to figure out why a script wasn't processed?

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -85,7 +35,7 @@ To prevent Astro from processing a script, add the `is:inline` directive.
 ```
 
 :::note
```
sarah11918: Feels like a great candidate for not being a note!  :smile: This is "how things work"

If the above is framed as "exceptions" or "unprocessed scripts" then this actually fits directly within the content, vs our framing now which is "here's how you do intentionally avoid script processing yourself... oh, and in some cases, Astro will do it for you when you might not expect it."

Simply annotating *when* scripts are *not* processed feels like it might be a win here, and give a nice support anchor link?

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -97,9 +47,9 @@ You may want to write your scripts as separate `.js`/`.ts` files or need to refe
 
 #### Import local scripts
 
-**When to use this:** when your script lives inside of `src/`.
+**When to use this:** When your script lives inside of `src/`.
 
-Astro will build, optimize, and add these scripts to the page for you, following its [script processing rules](#script-processing).
+Astro will build, optimize, and add these scripts to the page for you with TypeScript support and bundling.
```
sarah11918:
```suggestion
Astro will build, optimize, and add these scripts to the page for you with TypeScript support and bundling following its [script processing rules](#script-processing).
```
This may still be a helpful link! Some people enter the page here and so won't have read the above, and while `src` as an attribute doesn't mess with how scripts are handled, it wouldn't be clear from this that `src` is a special attribute.

Similarly, I can't suggest in the next section below where there are no changes, but I might consider adding it again like:

> This approach skips the [JavaScript processing, bundling, and optimizations](#script-processing) that are provided by Astro...

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -97,9 +47,9 @@ You may want to write your scripts as separate `.js`/`.ts` files or need to refe
 
 #### Import local scripts
 
-**When to use this:** when your script lives inside of `src/`.
+**When to use this:** When your script lives inside of `src/`.
```
sarah11918: Not sure I love this (original) format of when/when? But quite literally NWTWWHB! :joy:

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918: I can see the discussion on Discord re: clarifying inlined images, but as for the section duplication itself we have two options:

1. Duplicate, but ALSO go back to the CSS page and add the line that this will affect your scripts too and link here. (That would be a fine initial step, even if we decide option 2 may eventually be preferable.)
2. Figure out where to add a "Bundle Control" section (maybe we need a place for generic optimization advice?) and BOTH of these pages link there.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -123,6 +73,29 @@ To load scripts outside of your project's `src/` folder, include the `is:inline`
...
+    build: {
+      assetsInlineLimit: 1024,
+    }
+  }
+});
+```
+
+#### External script behavior
+
+External scripts with `type="module"` are automatically deferred by the browser and handled differently by Astro's [client router](/en/reference/modules/astro-transitions/#clientrouter). See the [View Transitions Guide](/en/guides/view-transitions/#script-behavior-with-view-transitions) for more information.
```
sarah11918:
```suggestion
External scripts with `type="module"` are automatically deferred by the browser and handled differently by Astro's [client router](/en/reference/modules/astro-transitions/#clientrouter-). See the [View Transitions Guide](/en/guides/view-transitions/#script-behavior-with-view-transitions) for more information.
```
Fix anchor link!

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -196,9 +168,7 @@ There are two advantages to using a custom element here:
 
 ### Pass frontmatter variables to scripts
 
-In Astro components, the code in [the frontmatter](/en/basics/astro-components/#the-component-script) between the `---` fences runs on the server and is not available in the browser. To send variables from the server to the client, we need a way to store our variables and then read them when JavaScript runs in the browser.
-
-One way to do this is to use [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) to store the value of variables in your HTML output. Scripts, including custom elements, can then read these attributes using an element’s `dataset` property once your HTML loads in the browser.
+In Astro components, [the frontmatter](/en/basics/astro-components/#the-component-script) code runs on the server and is not available in the browser. To pass server-side variables to client-side scripts, store them in [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) on HTML elements. Scripts can then access these values using the `dataset` property.
```
sarah11918:
```suggestion
In Astro components, [the frontmatter](/en/basics/astro-components/#the-component-script) code between the `---` fences runs on the server and is not available in the browser.

To pass server-side variables to client-side scripts, store them in [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) on HTML elements. Scripts can then access these values using the `dataset` property.
```
Feels like these are two key points and might be better scannable split up.

Also, remember anyone reading this section is probably newer to Astro and so repeating any additional *Astro* content (like the code fences) is likely good reinforcement. I totally agree with removing extraneous "how HTML works" and you've done that nicely here! But I wouldn't strip out the key Astro ideas that people might be new to.

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: #2 seems like the right call to me

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: Haha not sure why I did that, reverted it.

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: I went with "Unprocessed scripts" from your other comment

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: Added it back in

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: I dont think for now it needs to change, and I have since confirmed that astro does inline images when they are imported and used in .css files that are imported in .astro components, which is does automatically by vite and thus why it is affected here. 

In general some sort of centralized way to explain this and how assets work within astro in general would be cool but maybe not needed in this PR

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,14 +7,17 @@ i18nReady: true
...
+
+By default, Astro processes and bundles `<script>` tags with TypeScript support, npm module imports, and automatic optimization.
 
 ## Client-Side Scripts
 
 Scripts can be used to add event listeners, send analytics data, play animations, and everything else JavaScript can do on the web.
 
-```astro
-<!-- src/components/ConfettiButton.astro -->
+By default, Astro processes and bundles `<script>` tags, adding support for importing npm modules, writing TypeScript, and more.
```
OliverSpeir: this should probably be a link to the section below, commenting for T&D benefit

louisescher: We have this line twice right now, both in L18 and in L12?

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,14 +7,17 @@ i18nReady: true
 ---
 import ReadMore from '~/components/ReadMore.astro'
 
-You can add interactivity to your Astro components without [using a UI framework](/en/guides/framework-components/) like React, Svelte, Vue, etc. using standard HTML `<script>` tags. This allows you to send JavaScript to run in the browser and add functionality to your Astro components.
+You can add interactivity to your Astro components without [using a UI framework](/en/guides/framework-components/) like React, Svelte, Vue, etc. using `<script>` tags. This allows you to send JavaScript to run in the browser and add functionality to your components.
+
+By default, Astro processes and bundles `<script>` tags with TypeScript support, npm module imports, and automatic optimization.
```
sarah11918:
```suggestion
You can send JavaScript to the browser and add functionality to your Astro components through `<script>` tags written in the component template.

Scripts provide a way to add interactivity to your site without adding a [UI framework](/en/guides/framework-components/) like React, Svelte, or Vue to your project. This avoids the overhead of shipping framework JavaScript to the client.
```
It felt like maybe the focus on scripts here was stolen by mentioning all the famework stuff first, so what about something more like this framing for the intro? And, as Lou said, the same line is duplicated directly below, and I think maybe it fits better there than in an intro?

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918:
```suggestion
By default, [Astro processes and bundles `<script>` tags](#script-processing), adding support for importing npm modules, writing TypeScript, and more.
```

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -31,50 +34,19 @@ Scripts can be used to add event listeners, send analytics data, play animations
 </script>
 ```
 
-By default, Astro processes and bundles `<script>` tags, adding support for importing npm modules, writing TypeScript, and more.
-
-## Using `<script>` in Astro
```
sarah11918, line 36: Noting that this old heading is linked to from a few other pages, which is why the link check is failing, so this will require updating a few other files.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -31,50 +34,19 @@ Scripts can be used to add event listeners, send analytics data, play animations
...
+## Script processing
 
-In `.astro` files, you can add client-side JavaScript by adding one (or more) `<script>` tags.
+By default, `<script>` tags with no additional attributes other than `src` are processed by Astro in the following ways:
```
sarah11918:
```suggestion
By default, Astro processes `<script>` tags that contain no attributes (other than `src`) in the following ways:
```
The sentence gets a bit clunky here, so perhaps the parenthetical helps break up the sentence?

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -31,50 +34,19 @@ Scripts can be used to add event listeners, send analytics data, play animations
...
 
-In this example, adding the `<Hello />` component to a page will log a message to the browser console.
+- **TypeScript by default:** All scripts are processed as TypeScript.
+- **Import bundling:** Import local files or npm modules, which will be bundled together.
+- **Module injection:** The processed script is injected with [`type="module"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).
+- **Script deduplication:** If your component is used multiple times on a page, the script will only be included once.
+- **Automatic inlining:** If the script is small enough, Astro will inline it directly into the HTML to reduce the number of requests. See [Externalization](#externalization) for more details.
```
sarah11918, line 55:
``````suggestion
```astro title="src/components/Example.astro"
<script>
  // Processed! Bundled! TypeScript-supported!
  // Importing local scripts and Node modules works.
</script>
```

``````
This is the counterpart to the code block below, and I think is helpful to have here for scanning.

(And maybe now helps to show why "importing doesn't work" was used below, which we found unclear on its own!)

File: src/content/docs/en/guides/client-side-scripts.mdx [hunk 17]
```diff
@@ -31,50 +34,19 @@ Scripts can be used to add event listeners, send analytics data, play animations
...
 
-```astro title="src/components/Hello.astro"
-<h1>Welcome, world!</h1>
+### Unprocessed scripts
```
sarah11918, line 56:
```suggestion
### Inline scripts
```
I wonder if maybe this is the framing to take here?

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -31,50 +34,19 @@ Scripts can be used to add event listeners, send analytics data, play animations
...
-- `async` and `defer` attributes are unnecessary. Module scripts are always deferred.
-
-:::note
-The `async` attribute is valuable for normal scripts because it prevents them from blocking rendering. However, module scripts already have this behavior. Adding `async` to a module script will cause it to execute before the page has fully loaded. This is probably not what you want.
-:::
-
-### Opting out of processing
-
-To prevent Astro from processing a script, add the `is:inline` directive.
+Astro will not process your script tags if the `<script>` tag has any attribute other than `src`. Add the [`is:inline`](/en/reference/directives-reference/#isinline) directive to intentionally opt out of processing for a script.
```
sarah11918:
```suggestion
Adding any any attribute (other than `src`) to a `<script>` tag will create an inline script. It will be embedded directly inside your HTML and executed where it is rendered instead of processed and bundled by Astro.

You can add the [`is:inline`](/en/reference/directives-reference/#isinline) directive to intentionally opt out of processing for a script. This can be useful when you need more control over exactly when your script runs, for example, after certain DOM elements have loaded.
```
If we do think of these scripts as "inline scripts" vs "unprocessed scripts" then maybe a framing here is helpful. Bonus points: "inline scripts" are an actual thing, vs "unprocessed scripts" which feels more like a result/consequence/implementation detail? (Also, made up a reason here, but this is where we can add use cases for inline scripts.)

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -84,12 +56,27 @@ To prevent Astro from processing a script, add the `is:inline` directive.
 </script>
 ```
 
-:::note
-Astro will not process your script tags in some situations. In particular, adding `type="module"` or any attribute other than `src` to a `<script>` tag will cause Astro to treat the tag as if it had an `is:inline` directive.
-:::
+### Externalization
```
sarah11918: As we discussed on Talking and Doc'ing, maybe this whole section can be replaced with some kind of link to Vite, placed somewhere appropriate like where we talk about script processing?

> Astro's script processing and minification is handled by Vite. While normally not required, additional Vite configuration (e.g. [`vite.build.assetsInlineLimit`](https://vite.dev/config/build-options.html#build-assetsinlinelimit)) can be set in your Astro config file.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -84,12 +56,27 @@ To prevent Astro from processing a script, add the `is:inline` directive.
...
+export default defineConfig({
+  vite: {
+    build: {
+      assetsInlineLimit: 1024,
+    }
+  }
+});
+```
+
+#### External script behavior
```
sarah11918: Again wondering about the section heading here. It says "external scripts" which feels very broad, but then goes on to talk about "external scripts with type module" -- is that the only kind of external scripts? 

If the only content here is about being handled differently by the client router, maybe a section heading *about* scripts and the client router (even if it just mostly warns and links to view transitions) is a helpful anchor link to have, and descriptive of the content below?

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918: I'm further realizing that a link in the View Transitions page relies on this anchor link, and is linked from the text "inline scripts"! So, it does feel like maybe that's a vote in favour of that as an anchor link? I'm going to update THAT ONE to `/#inline-scripts` (even though it will still fail the check because this one hasn't been updated yet) while I fix the other failed links. Then, at least we'll only be left with the one we need to worry about.

src/content/docs/en/guides/client-side-scripts.mdx
bjohansebas: A small variation of @sarah11918’s proposal, although I also like that version quite a lot. 
```suggestion
You can send JavaScript to the browser and add functionality to your Astro components using `<script>` tags in the component template.

This allows you to create interactivity, handle events, or update content dynamically without integrating a [UI framework](/en/guides/framework-components/) like React, Svelte, or Vue, and without sending unnecessary framework JavaScript to the client.
```

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: the thing is that scripts are now inlined and processed, unless they are above 4kb in size total. So inlined isn't the best term to differentiate processed vs non processed scripts

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: for now I think we cut this and the whole externalization section as it wasn't there before so we aren't making things worse, and we can make it into a guide in future

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: there's no good link to replace it, I think for now we cut it out entirely and maybe do a guide or reference page

OliverSpeir:
I think beside final polishing etc what's left to do is 


1. Wording of the previous "Opt out of processing" heading which currently is "Unprocessed scripts"
    - The goal is to handle both the negative and the positive use case ( where they want to purposefully make it not processed, and where they want to figure out why wasn't their script processed ). The nuance is that now scripts are inlined by default, so you can have an inlined and processed script or if you don't want it processed you can set is:inline to make it inlined and not processed
3. Updating the files that referred to the old heading "Using `<script>` in Astro"
4. Update the files that refer to old "Opt out of processing" heading

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,18 +7,21 @@ i18nReady: true
 ---
 import ReadMore from '~/components/ReadMore.astro'
 
-You can add interactivity to your Astro components without [using a UI framework](/en/guides/framework-components/) like React, Svelte, Vue, etc. using standard HTML `<script>` tags. This allows you to send JavaScript to run in the browser and add functionality to your Astro components.
+You can send JavaScript to the browser and add functionality to your Astro components using `<script>` tags in the component template.
```
OliverSpeir, line 10: @bjohansebas @sarah11918 I modified y'alls suggestions for this line to be more terse, it's a bit of my own personal taste though so please weigh in again if you don't mind.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,18 +7,21 @@ i18nReady: true
...
+
+This allows you to create interactivity, handle events, or update content dynamically without integrating a [UI framework](/en/guides/framework-components/).
```
OliverSpeir: @bjohansebas @sarah11918 I modified y'alls suggestions for this line to be more concise, it's a bit of my own personal taste though so please weigh in again if you don't mind.

OliverSpeir:
For the sake of this PR I don't think we add the section parallel to [this style page section](https://docs.astro.build/en/guides/styling/#bundle-control), although I do think it should be documented somewhere/sometime, on talking and docing I got the sense that this was "reference" content and not "guide" content. 

For posterity the reason I think it should be documented is because Astro happens to use the `vite.build. assetsInlineLimit` config for deciding when to inline astro scripts and this isn't something that a user would assume this setting to affect as vite doesn't typically control whether a script gets inlined or not based on this.

from vite docs on the config option: 
> assets that are smaller than this threshold will be inlined as base64 URLs 
https://vite.dev/config/build-options.html#build-assetsinlinelimit

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918: OK, then back to the drawing board here! :sweat_smile: 

I'm still not *entirely* sold on "unprocessed scripts" (though I think that's the closest we've come), nor "opting out" as it was originally since I think this section also helps explain why scripts are unintentionally not processed.

Is there a more user-relevant term than "processed" to describe a script? What are they experiencing about those scripts? "Unbundled" (though, that might not cover it)? "Unoptimized"?

In this situation, we're using "processed" to mean scripts that do in fact still execute, just are not handled by Astro. So "unprocessed" almost feels too much like hinting at they might not actually work? We might "skip over them?"

I guess we'll keep thinking here!

sarah11918:
Also noting here that in a Discord thread we discussed the option of adding back that additional content as a recipe, as we've done for similar content:

> We could consider a Bundle Control *recipe*, which can then tie this action to a specific use case. We have a similar example where we used to show renaming output build file names right in a guide, but this was pretty advanced and we didn't necessarily want to suggest people actually do it unless they had a good reason to. So we removed it. Someone found it still in an old translation and thought it was helpful and could describe when it would be useful, so we added it back as a recipe here: https://docs.astro.build/en/recipes/customizing-output-filenames/
> 
> I could see doing something similar here, and then it can address both styles and scripts as necessary, itself could link to Vite, and then both pages could link to it?

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: Ya the distinction we are trying to make is whether Astro does anything to them. Normally html written in .astro files doesn't get changed, so we want to highlight that for scripts it does get changed, except in these distinct circumstances

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918:
```suggestion
Scripts add interactivity to your site, such as handling events or updating content dynamically, without the need for a [UI framework](/en/guides/framework-components/) like React, Svelte, or Vue. This avoids the overhead of shipping framework JavaScript and doesn't require you to know any additional framework to create a full-featured web site or application.
```
Thanks @OliverSpeir ! Just explaining the reason behind my edits, and my suggestion is just a way to get there, but can still be iterated upon!

- "this" is too vague to start here. (What's the antecedent: "script tags" (which is plural)? "JavaScript"? "the component template"?) For a basic page introduction, we have to be more explicit, I think, so we can deliberately lead the reader through a chain of thought with no doubt of where they're going. The introduction is not the place to skimp for the sake of brevity, since it has really important duties re: introducing the topic, grounding people in context, setting them up for what content is on this page, helping them figure out if they're in the right place... 

- we've never said "integrating a framework" and that usage feels off to me

- that list is written as an exhaustive list, and I'm not sure we mean to imply those are exactly and the only things possible, so I don't like framing it as "here are the 3 things you can do" (which again, in an introduction, is exactly how people will read this. If you were studying a new topic and taking notes, I guarantee that's how many people would interpret it.)

- we've lost one of the "benefits" (and in fact a question people often ask about!) if we don't mention not shipping framework JS to the browser.

- I also snuck in something I *wanted* to add in my first attempt, which is that you don't need to know anything but plain vanilla JS! Sorry, flagrantly bringing it up for consideration in the face of your attempt at brevity! :joy: But, I do want help saying this as concisely as possible!

OliverSpeir: Ok I trust the vision, thank you!

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,18 +7,21 @@ i18nReady: true
...
+
+Scripts add interactivity to your site, such as handling events or updating content dynamically, without the need for a [UI framework](/en/guides/framework-components/) like React, Svelte, or Vue. This avoids the overhead of shipping framework JavaScript and doesn't require you to know any additional framework to create a full-featured web site or application.
 
 ## Client-Side Scripts
 
 Scripts can be used to add event listeners, send analytics data, play animations, and everything else JavaScript can do on the web.
 
-```astro
-<!-- src/components/ConfettiButton.astro -->
+By default, [Astro processes and bundles `<script>` tags](#script-processing), adding support for importing npm modules, writing TypeScript, and more.
```
sarah11918:
```suggestion
The Astro `<script>` element is similar to the standard HTML script tag, but can provide extra [processing](#script-processing) that bundles multiple `<script>` tags, allows you to import npm packages, supports writing TypeScript, and more.


// perhaps even put the example here in between? 


<ReadMore>See [when your scripts will not be processed](#unprocessed-scripts) to troubleshoot script behavior, or to learn how to opt-out of this processing intentionally.</ReadMore>
```
@OliverSpeir Given our issue with "unprocessed scripts" below, maybe if we set this up here somehow then "Unprocessed scripts" fits a little more smoothly?

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918: OK, so see above something that might give a bit more context to "Unprocessed scripts"?

OliverSpeir: I think we want to keep the terminology consistent for the positive and the negative so if we say script processing we should say unprocessed scripts 

optimizing is good, bundled only covers part of it though 

im at a loss though for ideas sadly I will keep thinking though

src/content/docs/en/guides/client-side-scripts.mdx
OliverSpeir: This is great, I will commit it when back to computer, only thought is instead of saying can provide extra.. we should say does provide and then specify unless x

OliverSpeir: Also another thing I notice is that I'm trying to avoid saying supports typescript so that we can make it clear that the scripts are always treated as typescript

sarah11918: > we should say does provide and then specify unless x

This was my thinking with saying that it *can* provide processing, as a bonus, and then linking to "when will it not be processed." This lets us not clutter the basic definition of scripts with a bunch of exclusions here.

And, it doesn't *always* provide processing "by default" because there are exclusions, some of which you might not even know are happening (e.g. script size). So I think it's better not to say "by default" at all, but frame it as, "It's like HTML, but with the ability to be supercharged!" similar to the way we refer to Astro's templating language as a superset of HTML.

> make it clear that the scripts are always treated as typescript

a) How *important* is it for people to know this vs know they are allowed to write TypeScript? Who is that difference of nuance helping? If someone wouldn't write TS themselves, then is there any way not knowing this can hurt them? (And, we have to assure them that it's OK if they "only write JS") If they do know they can write TS, then won't they assume that their script is in fact treated like TS? I'm not opposed to adding it, but each bit of info we try to cram in makes things just that little bit clunkier (and less concise!) so it's important to consider the advantage of saying "everything that is true."


b) If it's crucial that people understand this, one thought is to include it in the definition (note, as attempting versions of this, I felt the clunkiness sinking in, which is why I stopped and wrote part a above, because it's a trade off):

e.g. squishing it upfront:
```
The Astro `<script>` element is similar to the standard HTML script tag, but is [treated as TypeScript](/en/guides/typescript/) and can provide extra [processing](#script-processing) that bundles multiple `<script>` tags, allows you to import npm packages, and more.
```

e.g. with another sentence
```
The Astro `<script>` element is similar to the standard HTML script tag, but can provide extra [processing](#script-processing) for your that bundles multiple `<script>` tags, allows you to import npm packages, and more. Like the code in your [Astro component script](/en/basics/astro-components/), [all JavaScript is treated as TypeScript](/en/guides/typescript/).
```

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918: Alright, let's go with "unprocessed scripts", especially in light of our addition earlier in the page!

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -522,7 +522,7 @@ When navigating between pages with the `<ClientRouter />` component, scripts are
 
 [Bundled module scripts](/en/guides/client-side-scripts/#script-processing), which are the default scripts in Astro, are only ever executed once. After initial execution they will be ignored, even if the script exists on the new page after a transition.
 
-Unlike bundled module scripts, [inline scripts](/en/guides/client-side-scripts/#opting-out-of-processing) have the potential to be re-executed during a user’s visit to a site if they exist on a page that is visited multiple times. Inline scripts might also re-execute when a visitor navigates to a page without the script, and then back to one with the script.
+Unlike bundled module scripts, [inline scripts](/en/guides/client-side-scripts/#inline-scripts) have the potential to be re-executed during a user’s visit to a site if they exist on a page that is visited multiple times. Inline scripts might also re-execute when a visitor navigates to a page without the script, and then back to one with the script.
```
```suggestion
Unlike bundled module scripts, [inline scripts](/en/guides/client-side-scripts/#unprocessed-scripts) have the potential to be re-executed during a user’s visit to a site if they exist on a page that is visited multiple times. Inline scripts might also re-execute when a visitor navigates to a page without the script, and then back to one with the script.
```

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918: Going to keep as is, and fix the other links accordingly

src/content/docs/en/guides/client-side-scripts.mdx
sarah11918: I think this is the single thing left to resolve before we're ready for a final boss review! :raised_hands:

yanthomasdev:
Added an additional note to @sarah11918's additional notes regarding i18n. I'll be back with a full review once we are done.

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,18 +7,21 @@ i18nReady: true
 ---
 import ReadMore from '~/components/ReadMore.astro'
 
-You can add interactivity to your Astro components without [using a UI framework](/en/guides/framework-components/) like React, Svelte, Vue, etc. using standard HTML `<script>` tags. This allows you to send JavaScript to run in the browser and add functionality to your Astro components.
+You can send JavaScript to the browser and add functionality to your Astro components using `<script>` tags in the component template.
+
+Scripts add interactivity to your site, such as handling events or updating content dynamically, without the need for a [UI framework](/en/guides/framework-components/) like React, Svelte, or Vue. This avoids the overhead of shipping framework JavaScript and doesn't require you to know any additional framework to create a full-featured web site or application.
```
yanthomasdev: website is one word
```suggestion
Scripts add interactivity to your site, such as handling events or updating content dynamically, without the need for a [UI framework](/en/guides/framework-components/) like React, Svelte, or Vue. This avoids the overhead of shipping framework JavaScript and doesn't require you to know any additional framework to create a full-featured website or application.
```

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -7,18 +7,21 @@ i18nReady: true
...
 
 ## Client-Side Scripts
 
 Scripts can be used to add event listeners, send analytics data, play animations, and everything else JavaScript can do on the web.
 
-```astro
-<!-- src/components/ConfettiButton.astro -->
+Astro automatically enhances the HTML standard `<script>` tag with bundling, TypeScript and more. See [how astro processes scripts](#script-processing) for more details.
```
yanthomasdev: comma nit
```suggestion
Astro automatically enhances the HTML standard `<script>` tag with bundling, TypeScript, and more. See [how astro processes scripts](#script-processing) for more details.
```

src/content/docs/en/guides/client-side-scripts.mdx
```diff
@@ -196,9 +171,9 @@ There are two advantages to using a custom element here:
 
 ### Pass frontmatter variables to scripts
 
-In Astro components, the code in [the frontmatter](/en/basics/astro-components/#the-component-script) between the `---` fences runs on the server and is not available in the browser. To send variables from the server to the client, we need a way to store our variables and then read them when JavaScript runs in the browser.
+In Astro components, [the frontmatter](/en/basics/astro-components/#the-component-script) code between the `---` fences runs on the server and is not available in the browser.
```
yanthomasdev: This sentence bothers me a bit. From the way it's written, it looks like "code" here refers to "the frontmatter" and not the code *in* the frontmatter.

Not sure the best way to write this, maybe like this?
```suggestion
In Astro components, the code in [the frontmatter](/en/basics/astro-components/#the-component-script) (between the `---` fences) runs on the server and is not available in the browser.
```

OliverSpeir: I guess technically we should call it component script not frontmatter or in that linked section it should say component script aka frontmatter but I like your suggestion it improves the clarity

PR #12095, opened 2025-07-29
add missing event listener boiler plate

martrapp:
#### Description (required)

The example for custom swap functions seemed to be missing a line.


ArmandPhilippot (approved):
Nice catch, Martin! 🫡

PR #12103, opened 2025-07-31
feat: fonts csp

florian-lefebvre:
#### Description (required)

Document new APIs that allow fonts to work with csp

- Closes #
- Suggested label: 

#### For Astro version: `???`. See astro PR [#14154](https://github.com/withastro/astro/pull/14154).

(Doesn't require a minor, but won't be ready for 5.13.0 so will release sometime after)


src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -265,6 +265,51 @@ export default defineConfig({
 });
 ```
 
+### `fontDirectiveResources`
+
+<p>
+
+**Type:** `string[]`<br />
+**Default:** `[]`<br />
```
sarah11918, line 273: Didn't we run into a similar tricky situation documenting something not too long ago where adding any value overrides (and removes) the default that needs to be added back in? 😅  What was that and how did we handle that?

I think what feels off here is that if you do nothing, by default you get the same result as if you added this property with `'self'`, right? So it's kind of like the default value is `'self'` and not nothing?

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -411,3 +456,30 @@ After the build, the `<meta>` element for this individual page will add your has
...
+---
+```
+
+After the build, the `<meta>` element for this individual page will add your source to the default `style-src` directive:
+
+```html
+<meta
+  http-equiv="content-security-policy"
+  content="
+    font-src https://fonts.cdn.example.com;
```
sarah11918: Is this the same situation where doing this overrides and removes `'self'`? If so, feels like that should also be mentioned, since it will remove the default?

sarah11918 (commented):
Hey Florian! Just a couple of quick initial questions here to start!

src/content/docs/en/reference/experimental-flags/csp.mdx
florian-lefebvre: Indeed! FYI I just copied the other properties so they would also need to be updated
```suggestion
**Default:** `["'self'"]`<br />
```

src/content/docs/en/reference/experimental-flags/csp.mdx
florian-lefebvre: In this case, it's merged with the config so we could show `'self'` if we wanted. Maybe @ematipico has an opinion on this!

sarah11918: OK, so if this merges that's probably easier to document! We would just make sure that we describe it as "additive"

florian-lefebvre:
@sarah11918 this is ready for review!

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -142,7 +142,7 @@ These properties are added to all pages and **completely override Astro's defaul
 <p>
 
 **Type:** `string[]`<br />
-**Default:** `[]`<br />
+**Default:** `["'self'"]`<br />
```
sarah11918, line 145: I am trusting you on the double set of quotation marks here!

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -265,6 +265,51 @@ export default defineConfig({
 });
 ```
 
+### `fontDirectiveResources`
+
+<p>
+
+**Type:** `string[]`<br />
+**Default:** `["'self'"]`<br />
+<Since v="5.12.7" />
```
sarah11918, line 274: Just checking whether this is releasing with the minor, or if before, then update this right before the patch release when known!

```suggestion
<Since v="5.14.0" />
```

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -265,6 +265,51 @@ export default defineConfig({
...
+  <meta
+    http-equiv="content-security-policy" 
+    content="
+      font-src 'self' https://fonts.cdn.example.com;
+    "
+  >
+</head>
+```
+
+When using [the experimental Fonts API](/en/reference/experimental-flags/fonts/), font resources will be injected by Astro (based on [`build.assetsPrefix`](/en/reference/configuration-reference/#buildassetsprefix)) as well as style [hashes](#hashes).
```
sarah11918, line 311: Just checking that this doesn't require experimental fonts to also be configured? When you say "when using the experimental fonts API" that means that all of the above is just generally true, even if not using. But now here, we are *additionally* specifying something that is true only when using experimental fonts, right?

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -265,6 +265,51 @@ export default defineConfig({
...
+      fontDirectiveResources: [
+        "'self'",
+        "https://fonts.cdn.example.com"
+      ]
+    }
+  }
+});
+```
+
+After the build, the `<meta>` element will instead apply your sources to the `font-src` directive:
```
sarah11918: I can see you were just copying the existing wording, which I must not have caught at the time, but is there a reason we can't just say:

```suggestion
After the build, the `<meta>` element will apply your sources to the `font-src` directive:
```

My first thought was "instead of what?" and I *think* we don't we need any "instead of" here, if Astro will just apply your configuration (as one would expect it to do)?

I don't think we need to say "instead of the default, you'll now have your configuration!" right? (Or is there something else I'm missing, because I might be!)

src/content/docs/en/reference/experimental-flags/csp.mdx
sarah11918, line 311: re: "font resources will be injected by Astro based on..." Would it be correct to say something like below, that I think is a bit more explicit?

```suggestion
When using [the experimental Fonts API](/en/reference/experimental-flags/fonts/), Astro will apply both your [`build.assetsPrefix`](/en/reference/configuration-reference/#buildassetsprefix) configuration as well as style [hashes](#hashes) to your font resources in the `<meta>` element.
```

And then this prompts the question... what if I'm *not* using experimental fonts?? Do I have to do anything special? Will this "just work" as expected, or do I need to set something manually because I'm not using experimental fonts? Do we need an "Otherwise, you will need to...." follow up statement if someone has `assets.buildPrefix` set, or needs hashes to apply?

Also just checking where it says in `hashes`:

> If you have external scripts or styles that aren’t generated by Astro, or inline scripts, this configuration option allows you to provide additional hashes to be rendered.

> After the build, the <meta> element will include your additional hashes in the script-src and style-src directives:

Do these now need to be extended to include fonts? Is there an analogous font line for this example? Are fonts simply an extension of styles and use their hashes, or are there separate font hashes?

(And noting that if this is only true when using experimental fonts, then, is there anything anyone needs to know in this section if they are NOT using experimental fonts?)

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -411,3 +456,30 @@ After the build, the `<meta>` element for this individual page will add your has
   "
 >
 ```
+
+### `insertFontResource`
+
+<p>
+
+**Type:** `(resource: string) => void`<br />
+<Since v="5.12.7" />
```
sarah11918, line 465: same note re: version, whatever it happens to be!

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -411,3 +456,30 @@ After the build, the `<meta>` element for this individual page will add your has
   "
 >
 ```
+
+### `insertFontResource`
```
sarah11918, line 460: Does this require experimental Fonts API, or work differently with or without it?

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -411,3 +456,30 @@ After the build, the `<meta>` element for this individual page will add your has
...
+
+Inserts a new resource to be used for the `font-src` directive.
+
+```astro
+---
+Astro.insertFontResource("https://fonts.cdn.example.com");
+---
+```
+
+After the build, the `<meta>` element for this individual page will add your source to the default `style-src` directive:
```
sarah11918: This looks good!

src/content/docs/en/reference/experimental-flags/csp.mdx
florian-lefebvre: Maybe this is not the right place for this info? But yeah basically there are 2 things:
1. The addition to the CSP API, that can be used on its own
2. Experimental fonts supporting CSP, which uses the new CSP API + injects styles as needed

florian-lefebvre: > And then this prompts the question... what if I'm not using experimental fonts?? Do I have to do anything special?

To recap:
- You're using experimental fonts: Astro will adds things to both `style.hashes` and `fontDirectiveResources` (taking `assets.buildPrefix`)
- You're not using experimental fonts: you can use `fontDirectiveResources` without any issues

So I'd say no.

> Do these now need to be extended to include fonts?

I don't think so? Fonts only have directives, not hashes

src/content/docs/en/reference/experimental-flags/csp.mdx
florian-lefebvre: It works without experimental fonts, and works the same with or without

src/content/docs/en/reference/experimental-flags/csp.mdx
florian-lefebvre:
```suggestion
After the build, the `<meta>` element for this individual page will add your source to the default `font-src` directive:
```

src/content/docs/en/reference/experimental-flags/csp.mdx
sarah11918: OK, but if Astro works differently when I am or am not using experimental fonts, and I probably want one particular outcome, then... how do I get that same desired result in both cases if Astro is working differently in each case? I'm not sure how they can work differently and yet "it's all good" in either case?

Like, is taking those things into consideration "good" or "bad"? (Is it more like "I probably want it consistent with how I have other things set up, and yay, using experimental fonts it's done for me so I don't have to do X, Y and Z manually when I'm not using experimental fonts!" or is this "Be careful! If you use experimental fonts, this will happen, unlike when you're not. You might have to undo it manually!")

florian-lefebvre: Ah I *think* I see what you mean now. Basically, experimental fonts are dealing with font directives and style hashes, because the way it works is it injects an inline script.

Now if you're not using experimental fonts, you may be just setting `fontDirectiveResources` and then doing things in a css file, in which case the style hashes section already covers that.

So I guess the main difference is this:
- When using experimental fonts, `fontDirectiveResources` may get things injected from `assets.buildPrefix`
- When using `fontDirectiveResources` without experimental fonts, you have to duplicate things from `assets.buildPrefix` if applicable

Does that help?

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -265,6 +265,51 @@ export default defineConfig({
 });
 ```
 
+### `fontDirectiveResources`
+
+<p>
+
+**Type:** `string[]`<br />
+**Default:** `["'self'"]`<br />
```
florian-lefebvre:
```suggestion
**Default:** `[]`<br />
```

PR #12104, opened 2025-08-01
Update from-docusaurus.mdx; en-only

LukeEades:
grammar

#### Description (required)

Changed from "to" to "with" in sentence "Both Astro and Docusaurus support MDX pages. You should be able to use your existing .mdx files to Astro". Changed for sake of grammatical correctness.

- Closes #
- Suggested label:


bjohansebas (approved):
LGTM

sarah11918 (approved):
This looks great, thank you! Some versions of these guides had "you should be able to *bring*... " which is probably what happened here. Thank you for taking the time to fix this, and welcome to Team Docs! :partying_face:

PR #12130, opened 2025-08-08
Add note that the bug appears only with trailingSlash: ignore

jp-knj:
#### Description (required)

This PR fixes a bug in the v5 upgrade guide's cookie-forwarding example for form actions. The example code was missing a path attribute when setting cookies, which causes actions to fail silently when using trailingSlash: "ignore" (the default with build.format: "directory").

The Problem:
Without an explicit path, cookies default to the request's directory path per RFC 6265. When a POST request to `/route` sets a cookie without a path, it's scoped to `/route`. The subsequent redirect to /route/ (after PR: https://github.com/withastro/astro/pull/13997) can't access this cookie, making it appear as if the action never executed.

- Related discussions: https://github.com/withastro/astro/issues/14177#issuecomment-3167708316
- Closes #
- Suggested label: docs


src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -735,6 +735,11 @@ export const onRequest = defineMiddleware(async (context, next) => {
 		context.cookies.set('ACTION_PAYLOAD', {
 			actionName: action.name,
 			actionResult: serializeActionResult(actionResult),
+		}, {
+			path: '/',      // Critical: ensures cookie works with trailingSlash: "ignore"
+			httpOnly: true,
+			sameSite: 'lax',
+			maxAge: 60
```
jp-knj: httpOnly, sameSite, maxAge follow best practice, it could remove

src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -735,6 +735,11 @@ export const onRequest = defineMiddleware(async (context, next) => {
 		context.cookies.set('ACTION_PAYLOAD', {
 			actionName: action.name,
 			actionResult: serializeActionResult(actionResult),
+		}, {
+			path: '/',      // Critical: ensures cookie works with trailingSlash: "ignore"
```
sarah11918:
```suggestion
			path: '/',
```
We don't typically point out when other necessary items in a code snippet are "critical", it's just, how it works. (In most code examples, if you don't follow them, it doesn't work. :smile: )

I'd love to hear opinions on this, and whether this is something worth pointing out in extreme detail vs just fixing in the code snippet.

src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -753,6 +758,10 @@ export const onRequest = defineMiddleware(async (context, next) => {
 })
 ```
 
+:::note[Cookie path configuration]
+The `path: '/'` option is critical when using `trailingSlash: "ignore"` (the default with `build.format: "directory"`). Without an explicit path, the cookie defaults to the request's directory path, which can cause cookies to be inaccessible after redirects between `/route` and `/route/`. Setting `path: '/'` ensures the cookie is available regardless of trailing slashes.
+:::
```
sarah11918: Similar to my comment above, I would need others to confirm that this really warrants specifically calling out. It's one thing to add a missing line to a code snippet. It's another to explain it in super detail just *because* it happened to be missing. Especially in an upgrade guide like this, which really just shows working code for the way Astro now works.

src/content/docs/en/guides/upgrade-to/v5.mdx
sarah11918: No problem with showing best practices, but we do want to make sure the emphasis here is on supporting how to deal with changed behaviour (since this is the upgrade guide); it's not meant to be a recipe teaching from scratch.

Will ask others to share if they find this helpful or potentially cluttering/distracting!

src/content/docs/en/guides/upgrade-to/v5.mdx
jp-knj: Thanks, Okay

src/content/docs/en/guides/upgrade-to/v5.mdx
sarah11918:
```suggestion
```
Discussed with the team and we're going to leave the note out, but keep the extra "best practice" in the example as being more helpful!

PR #12139, opened 2025-08-10
feat: add enum docs

coding-in-public:
per https://github.com/withastro/astro/pull/14190

#### Description (required)

[`@astrojs/db@0.17.0`](https://github.com/withastro/astro/pull/14190) just released support for enums.


src/content/docs/en/guides/integrations-guide/db.mdx
```diff
@@ -127,6 +127,28 @@ There are a few shared configuration values across all columns:
 - `unique` - Mark a column as unique. This prevents duplicate values across entries in the table.
 - `references` - Reference a related table by column. This establishes a foreign key constraint, meaning each column value must have a matching value in the referenced table.  
 
+The `text` column may optionally define an enum.
```
sarah11918:
```suggestion
A `text` column may optionally define an enum.
```
I think this might be more readable either as "a `text` column" (as in, any column defined as text) or
"the `text` column type" or something (as in, of the list above, the one that is `text`)?

It took me too long to figure out that "the" works (in a particular sense) here, vs "a" and that it wasn't just straight up a mistake, so I'd probably be clear about "the thing from the above list" if you're sticking with "the", or I'd switch to "a" and have this be read as "a text column that you define."  :smile:

sarah11918 (commented):
Haha, you're letting @Adammatthiesen off the hook for adding docs for his feature, @coding-in-public ! :joy: 

I've asked Adam to take a look to review this, but I only had one little language question on my end. Otherwise, it LGT thank you! (I think it's in the correct location, a more specific option after the ones common to all common types?)

I'll let the DB'ers handle this! (Any comments @Fryuni ?)

src/content/docs/en/guides/integrations-guide/db.mdx
```diff
@@ -127,6 +127,28 @@ There are a few shared configuration values across all columns:
...
+	columns: {
+		id: column.number({ primaryKey: true }),
+		name: column.text(),
+    rank: column.text({ enum: ['user', 'mod', 'admin'] }),
+	},
+});
+
+// Resulting type definition
+type UserTableInferInsert = {
+    id: string;
```
Adammatthiesen:
```suggestion
    id?: string;
```

This is optional due to the primary key

sarah11918:
( @Adammatthiesen You're allowed to have a higher standard than "not wrong" if you want! You can also have a different vision for these docs! Just letting you know! :smile: )

Adammatthiesen:
> ( @Adammatthiesen You're allowed to have a higher standard than "not wrong" if you want! You can also have a different vision for these docs! Just letting you know! 😄 )

Oh no, thats not what i meant, :sweat_smile: these changes will be correct for the current state of DB. might be thinking a bit too far ahead atm :sweat_smile: I have no other issues aside from what has been mentioned already

src/content/docs/en/guides/integrations-guide/db.mdx
Fryuni: I think it is worth adding in this paragraph what it means to "define an enum".

```suggestion
The `text` column may optionally define a list of string literals to serve as an enum. This will cause the generated types to be restricted to the given literal values. Although the type will reflect only the given values, **no runtime validation will be performed** so removal, addition and changing values should be handled in your code.
```

Are we talking directly to the reader in this page? If not then change the ending to "in the project's code."

coding-in-public:
I'll let you take it from here, Sarah. But happy to jump in if it would help :)

src/content/docs/en/guides/integrations-guide/db.mdx
```diff
@@ -127,6 +127,28 @@ There are a few shared configuration values across all columns:
 - `unique` - Mark a column as unique. This prevents duplicate values across entries in the table.
 - `references` - Reference a related table by column. This establishes a foreign key constraint, meaning each column value must have a matching value in the referenced table.  
 
+The `text` column may optionally define a list of string literals to serve as an enum. This will cause the generated types to be restricted to the given literal values. Although the type will reflect only the given values, **no runtime validation will be performed** so removal, addition and changing values should be handled in your code.
```
sarah11918:
```suggestion
A `text` column can optionally define a list of string literals to serve as an enum for generating types. However, **no runtime validation will be performed**. Removing, adding, and changing values should be handled in your project code.
```
@Fryuni I committed your suggestion as is, thank you! But I think something like above feels a bit more Astro Docs'y? Would the above still be correct/sufficient?

sarah11918:
Noting that this feature is already live and doesn't *have* to wait for the minor, but I've applied the labels/milestone to coordinate with the Astro milestone (which will discuss this in the blog post) to ensure it *at least* doesn't merge any *later* than the minor release!

src/content/docs/en/guides/integrations-guide/db.mdx
Fryuni: Yes, that is totally right.

PR #12140, opened 2025-08-11
i18n(fr): improve the wording in `recipes/*`

ArmandPhilippot:
#### Description (required)

Updates the French translation of all the recipes to improve the wording following changes in the French i18n guide:
* replace `plugin` with `module d'extension`
* replace `construction` with `compilation` where more appropriate
* replace `supporter` with `prendre en charge`
* replace `import` with `importation`
* replace `runtime` with `environnement d'exécution`
* replace `frontal` with `frontend` (do not translate)
* replace `endpoint`/`point d'accès` with `point de terminaison`
* replace `layout` (sometimes translated `modèle` and `disposition`) with `mise en page`
* replace `template` with `modèle`
* replace `île` with `îlot`
* replace `rendre` with `afficher`/`effectuer un rendu`
* replace `obsolète` with `déprécié`
* replace `styler`/`styliser` with `metttre en forme`
* replace `clé d'objet` with `nom de propriété` (same translation as MDN)
* use a consistent verb tense in steps
* use French quotes
* fix some typo (space before colon in inline code, `resources`)
* translate some examples
* small rewording to make the reading smoother

> [!NOTE]
> Regarding `bundle`, [as discussed on Discord](https://discord.com/channels/830184174198718474/971729943786561556/1400936653891309611), I don't think `regroupement` is the right translation. I'm not sure [`groupage`](https://www.larousse.fr/dictionnaires/francais/groupage/38421) is better. In both cases, I would expect `regroupement/groupage de quelque chose`, not a single name. And `paquet compilé` doesn't seem to convince anyone.
>* React keeps `bundle`
>* Vue uses `paquet`
>* MDN uses `paquet` (and `regroupement` for `bundling`)
>
> So I don't know how to translate it, and maybe we should keep thinking about it and update it later. In the meantime, I think we can keep `bundle` (instead of mixing translations as currently). I'd say NWTWWHB!

- Suggested label: i18n


dreyfus92 (approved):
LGTM 🤘🏻

PR #9328, opened 2024-09-07
Revamp Content section 5.0

sarah11918:
New content (& content collections) section for 5.0

This PR:
- updates the content collections guide, splitting it into a main landing page and two smaller guides
- updates Markdown guide for freshness, streamlined-ness, and simplifies by removing MDX content (which has, already in live published docs, been moved to the MDX page itself)
- updates a bunch of links that have shifted during the flight and other minor tweaks to give an overall "Content" refresh

Re: Content Collections (now, powered by a layer!) for @ascorbic 

1. Please review the three new content collections guide pages [preview here](https://deploy-preview-9328--astro-docs-2.netlify.app/en/guides/content-collections/):
content-collections.mdx
defining-collections.mdx
working-with-collections.mdx

You can ignore everything else! It's just links etc. around the rest of the site because the original page (and a couple of other related content / markdown pages) have changed. There's also some future proofing for when collections don't have to be in `src/content/`

By "review" I mean:
- correct any inaccuracies!
- point out any wrong nuances, takes, or nudges that don't feel right to you
- is anything key to *using collections* missing? We can always add more over time, but is there anything that's absolutely needed here or people are just not going to be successful?
- if you happen to notice language stuff, then all proofreading, grammar, formatting etc. are fine here, but it's not *required* to take a specific pass or pay super close attention to this since there will still be lots of available eyes to catch that stuff.

My main goal is to make sure that this guide faithfully and accurately represents both how to get set up and use collections, and the "vision" for collections.

**After, outside of this PR**:

2. I have added in this PR a shell for a new Loader API page. It seems like this should be a good place to document the Loader API in technical detail, and describe how to build one. **You don't need to contribute to this branch**. Once the guide content is fine, we can merge this branch and that shell page will exist. You can separately make a PR that will include:

- content for the Loader API page
- any updates to the API reference page, content collections section

Those two can be together in the same PR if you want, to get all the API content squared away.

My first priority is to get this PR merged into the beta branch so I can start fixing any conflicts because there were a lot of moving pieces in breaking up the content collections page! The API material can be added later!

PR #8705, opened 2024-06-30
feat: add docs for json schemas

alexanderniebuhr:
#### Description (required)

This PR adds docs for the following Core PR: https://github.com/withastro/astro/pull/11379

The feature was experimental before:
- https://github.com/withastro/astro/pull/10145 
- https://docs.astro.build/en/reference/configuration-reference/#experimentalcontentcollectionjsonschema

#### Related issues & labels (optional)

- Closes #
- Suggested label:

PR #12155, opened 2025-08-14
Add deleted section about json schema generation

alexanderniebuhr:
#### Description (required)

- This PR readds an section which was deleted by accident
- Initial release was in https://github.com/withastro/docs/pull/8705
- It got removed in https://github.com/withastro/docs/pull/9328

- needed for https://github.com/withastro/astro/pull/14221
- needed for https://github.com/withastro/astro/issues/13536


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -565,6 +565,37 @@ const { Content } = await render(post);
 Explore the `src/pages/` folder of the [blog tutorial demo code on GitHub](https://github.com/withastro/blog-tutorial-demo/tree/content-collections/src/pages) to see full examples of creating pages from your collections for blog features like a list of blog posts, tags pages, and more!
 :::
 
+## Enabling JSON Schema Generation
+
+<p><Since v="4.13.0" /></p>
+
+If you are working with collections of type `data`, Astro will auto-generate JSON schema files for your editor to get IntelliSense and type-checking. A separate file will be created for each data collection in your project based on your collections defined in `src/content/config.ts` using a library called [`zod-to-json-schema`](https://github.com/StefanTerdell/zod-to-json-schema#known-issues).
```
ArmandPhilippot: The config file has moved with the Content Layer API:
```suggestion
If you are working with collections of type `data`, Astro will auto-generate JSON schema files for your editor to get IntelliSense and type-checking. A separate file will be created for each data collection in your project based on your collections defined in `src/content.config.ts` using a library called [`zod-to-json-schema`](https://github.com/StefanTerdell/zod-to-json-schema#known-issues).
```

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -565,6 +565,37 @@ const { Content } = await render(post);
...
+
+This feature requires you to manually set your schema's file path as the value for `$schema` in each data entry file of the collection:
+
+```json title="src/content/authors/armand.json" ins={2}
```
ArmandPhilippot: nit: to make old and new behavior changes more obvious, `src/content` became `src/data` in this page. While this is valid, maybe we should stay consistent:
````suggestion
```json title="src/data/authors/armand.json" ins={2}
````

File: src/content/docs/en/guides/content-collections.mdx [hunk 3]
```diff
@@ -565,6 +565,37 @@ const { Content } = await render(post);
...
+```
+
+Alternatively, you can set this value in your editor settings. For example, to set this value in [VSCode's `json.schemas` setting](https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings), provide the path of files to match and the location of your JSON schema:
+
+```json
+{
+  "json.schemas": [
+    {
+      "fileMatch": [
+        "/src/content/authors/**"
```
ArmandPhilippot: nit: same as above, to make old and new behavior changes more obvious, `src/content` became `src/data` in this page, so maybe we should stay consistent:
```suggestion
        "/src/data/authors/**"
```

ArmandPhilippot (dismissed):
LGTM! I only left a fix (the config file) and two nits regarding paths.

src/content/docs/en/guides/content-collections.mdx
bjohansebas: I’ve been thinking about this for quite a while. I wonder which data-type collections could benefit from this, only `.json`? If it’s only JSON file collections, I think it would be better to say that it only benefits them instead of implying it’s more global.

Or, if this benefits more globally, like `.yaml` or `.toml` files, since those would be data-type collections (as far as I understand, I could be wrong), then improving the meaning of that “data-type collections” concept would work here, because I find it a bit confusing.

Just my two cents.

ArmandPhilippot: Nice catch, thanks! This is indeed the wording used with the previous API where `data` was a collection type. So we probably need to reword this part to clarify what a `data` collection is.

I quickly tested on my side, and the schema auto-generation works for JSON, YAML and TOML files. But now, I'm a bit confused because I tried with a collection using Markdown files and a JSON schema is also generated for those (e.g. `.astro/collections/posts.schema.json`). Which contradicts what we have in [Experimental Intellisense for content collections](https://docs.astro.build/en/reference/experimental-flags/content-intellisense/):
> When enabled, this feature will generate and add JSON schemas to the .astro directory in your project.

It seems the JSON schemas are always generated (for any file supported at least).
**EDIT:** However, it seems the `.astro/collections/collections.json` is only generated with the experimental flag enabled.

However, I guess there is no native support for validating a YAML or TOML file with a JSON schema (please correct me if I'm wrong) and, so, what is described here only work when using a JSON file. This can be extended to `.md`, `.mdx` and `.mkdoc` files using the experimental flag. Now, the grey zone to me is... is there a way to provide Intellisense to YAML and TOML file with a JSON schema?

And maybe we should rework this section to precise that the schema must be provided manually and:
* this is supported by default for JSON files,
* for `.md`, `.mdx` and `.mkdoc`, the user needs to enable the experimental flag
* figure out what can be done with YAML and TOML

sarah11918:
Will note that on Talking and Doc'ing last week, we briefly discussed **where on the page** this information makes sense to add because to this is does not fall under any of the existing 3 topics (Defining the collection, querying the collection generating routes from the collection) but is also not really "on the same hierarchy level" as those broad topics.

This is related to the file loader, but instead something (more advanced?) that you do with a collection *entry*, and in fact, maybe is in some way related to the `<h3>` "defining a custom ID" in that sense.

I think @delucis will be overseeing this PR, alongside and dependent upon his feature fix PR. So I wouldn't treat this as "ready to review" since it's not even correct content yet. It's just a cut and paste placeholder from the way collections worked pre-v5 right now (e.g. there is no `type` for collections anymore), but should be considered more like feature documentation to accompany the incoming PR.

We'll need to document how it's *supposed* to work with the content layer API now powering collections: does this only affect the `file()` loader loading a single JSON file, or also `glob()` if there are multiple JSON files for each entry? (The example here looks like the file is a single author, not an "authors" file? If we mean for this to be strictly a "one JSON file collection" then I think we might want the code example to show multiple author entries in the single file, and not be named after one author?) Does this only work for local JSON entries? (not remote JSON files that might be using a custom loader instead of our built-in loaders, but still loading and parsing JSON?)

There were a lot of things that were unclear about how this worked for new content collections when they were released in v5, which is why this section was removed at that time! It wasn't actually deleted by accident, but we didn't have confirmation for how this content needed to be updated at the time, so we could not include it. We instead went with the assumption that if this is still a thing, it will make itself an issue at some future date. THE FUTURE IS NOW! :smile:

src/content/docs/en/guides/content-collections.mdx
alexanderniebuhr: It does not only affect `.json` files, it also works for `.yaml` files for example

alexanderniebuhr:
> does this only affect the file() loader loading a single JSON file, or also glob() if there are multiple JSON files for each entry? 

Does affect both :)

src/content/docs/en/guides/content-collections.mdx
ArmandPhilippot: Oh okay, I thought YAML required to install something to be able to do that doing a very quick research. So, if it's supported by default, I guess we need to verify if TOML is also supported, because it was added more recently, or if we need to precise this only applies to JSON and YAML.

alexanderniebuhr: Well it depends on how your Editor handles it. But yeah for VSCode you need the redhat extension for "full yaml lsp" support. But other editors might have it built in. We shouldn't be editor specific here I think though.

alexanderniebuhr:
@delucis I think your core PR got merged, and I remember you wanted to look into this PR too. Any updates?

delucis:
Update the PR with a rework of the docs!

sarah11918 (approved):
I think this is great, and the best placement we have for it right now.

Since this is schema related, we *could* consider a short line (or tip?) somewhere within /en/guides/content-collections/#defining-the-collection-schema with an inline link about `using [JSON schemas for your collections](#collection-json-schemas) in your editor`.

But first task, is get this all on the page! 🙌

delucis:
> Since this is schema related, we _could_ consider a short line (or tip?) somewhere within /en/guides/content-collections/#defining-the-collection-schema with an inline link about `using [JSON schemas for your collections](#collection-json-schemas) in your editor`.

Well the one thing here is that this is potentially a bit confusing: these schemas are NOT exactly the same thing a those schemas. They’re derived _from_ them, but I was wondering if there could be some confusion about that, i.e. a JSON Schema is a specific file format used for validating JSON files, vs your collection schema which is a Zod schema you write.

sarah11918:
> Well the one thing here is that this is potentially a bit confusing:

None of this feels particularly ergonomic 😄 Happy just to have it documented, and we'll look at a larger pass at this page (e.g. we've talked about splitting up the two loaders into individual sections) to correspond to v6!

PR #12157, opened 2025-08-14
fix: update `sitemap.mdx`

jsparkdev:
#### Description (required)

fix wrong version in `sitemap.mdx`

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -259,7 +259,7 @@ export default defineConfig({
 
 **Type:** `string[]`<br />
 **Default:** `[]`<br />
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -259,7 +259,7 @@ export default defineConfig({
 
 **Type:** `string[]`<br />
 **Default:** `[]`<br />
-<Since v="3.5.0" pkg="@astrojs/sitemap" />
+<Since v="5.13.0" pkg="@astrojs/sitemap" />
```
sarah11918, line 262: Good spotting, but this actually *is* correct since it's v3.5 of the sitemap integration! (I have to pay very close attention when I do this for our integrations. :joy: )

A good way to tell is if `pkg` is included in the component, it's not just the major Astro version.

PR #12158, opened 2025-08-14
ci: use depot runners for CI workflows

ematipico:

- Closes #
- Suggested label:


delucis:
Updated branch to see if a second run would be different, but similar values: about 5m30s for our slowest job (Check Links). According to GitHub’s [Performance Metrics](https://github.com/withastro/docs/actions/metrics/performance?tab=jobs) this job currently averages 6m42s. So the depot runners are about 18% faster on that basis.

Given that we already wait on Netlify’s CI builds, we’ll probably not use depot for the docs repo and keep the exploration scoped to `astro`.

PR #12168, opened 2025-08-16
Remove `"default"` from `workerEntrypoint.namedExports` type

yanthomasdev:
#### Description (required)

This removes `"default"` from being part of `namedExports`' type, which wasn't correct. The `"default"` export is added *outside* of this array, which could confuse people into beliving that setting a custom `namedExports` meant adding `"default"` manually.

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -295,7 +295,7 @@ To support other [Cloudflare invocation handlers](https://developers.cloudflare.
 <p>
 
 **Type:** `[]`<br />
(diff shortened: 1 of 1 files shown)


yanthomasdev:
> e.g. "An array of additional named exports to use for the entry file" (I think the following paragraph takes care of explaining that default exports will exist if you don't add anything)

I know it sounds confusing @sarah11918, but the default export isn't a named export. The default export doesn't take a name, so you can import it without curly braces, different from named exports. And the default export can be exporting many functions, as is the case in the example below, but it is still considered one default export rather than export*s*.

sarah11918:
Ah, ok, that makes sense! Then yes, this all seems to check out!

PR #12173, opened 2025-08-19
feat: clarify partytown config

florian-lefebvre:
#### Description (required)

When working on a complex site using Partytown, we found the need for proxying based on environment variables. We struggled a lot because we thought everything had to be passed through the integration config, but it turns out it can also come from `window.partytown`.

Anyways, here are the additions:

- Document that `config` is serialized which means you can't access things outside functions scopes. Not sure how to convey that functions are also serialized (ie. not converted to `null`)
- Document that you can also use `window.partytown`, with a real world example

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/partytown.mdx
+++ b/src/content/docs/en/guides/integrations-guide/partytown.mdx
@@ -189,13 +189,13 @@ const proxyMap = {
 ---
 
 <script is:inline set:html={`
-window.partytown = {
-  resolveUrl: (url) => {
-    const proxyMap = ${JSON.stringify(proxyMap)};
-    url.hostname = proxyMap[url.hostname] || url.hostname;
-    return url;
-  },
-};
+  window.partytown = {
+    resolveUrl: (url) => {
+      const proxyMap = ${JSON.stringify(proxyMap)};
+      url.hostname = proxyMap[url.hostname] || url.hostname;
+      return url;
+    },
+  };
 `} />
 ```
 
--- a/src/content/docs/en/guides/integrations-guide/partytown.mdx
+++ b/src/content/docs/en/guides/integrations-guide/partytown.mdx
@@ -10,7 +10,7 @@ i18nReady: true
 ---
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 
-This **[Astro integration][astro-integration]** enables [Partytown](https://partytown.builder.io/) in your Astro project.
+This **[Astro integration][astro-integration]** enables [Partytown](https://partytown.qwik.dev/) in your Astro project.
 
 ## Why Astro Partytown
 
@@ -109,35 +109,11 @@ export default defineConfig({
 
 This mirrors the [Partytown config object](https://partytown.qwik.dev/configuration/).
 
-:::caution
-The `config` object is serialized and sent to the client. That means functions like `config.resolveUrl()` cannot use TypeScript, nor access anything outside of their scope.
-:::
-
-If you need to pass external data to functions when Partytown is initialized on the client, you can set the configuration through `window.partytown`:
-
-```astro title="Head.astro"
----
-const proxyMap = {
-  "connect.facebook.net": "my-proxy.com"
-};
----
-
-<script is:inline set:html={`
-window.partytown = {
-  resolveUrl: (url) => {
-    const proxyMap = ${JSON.stringify(proxyMap)};
-    url.hostname = proxyMap[url.hostname] || url.hostname;
-    return url;
-  },
-};
-`} />
-```
-
-It will be overriden by any config option coming from the integration.
+The `config` object is serialized, even [`config.resolveUrl()`](#configresolveurl), because it is sent to the client.
 
 ### config.debug
 
-Partytown ships with a `debug` mode; enable or disable it by passing `true` or `false` to `config.debug`. If [`debug` mode](https://partytown.builder.io/debugging) is enabled, it will output detailed logs to the browser console.
+Partytown ships with a `debug` mode; enable or disable it by passing `true` or `false` to `config.debug`. If [`debug` mode](https://partytown.qwik.dev/debugging) is enabled, it will output detailed logs to the browser console.
 
 If this option isn't set, `debug` mode will be on by default in [dev](/en/reference/cli-reference/#astro-dev) or [preview](/en/reference/cli-reference/#astro-preview) mode.
 
@@ -159,7 +135,7 @@ Third-party scripts typically add variables to the `window` object so that you c
 
 To solve this, Partytown can "patch" variables to the global window object and forward them to the appropriate script.
 
-You can specify which variables to forward with the `config.forward` option. [Read more in Partytown's documentation.](https://partytown.builder.io/forwarding-events)
+You can specify which variables to forward with the `config.forward` option. [Read more in Partytown's documentation.](https://partytown.qwik.dev/forwarding-events)
 
 ```js title="astro.config.mjs" {7}
 export default defineConfig({
@@ -174,6 +150,38 @@ export default defineConfig({
   ],
 });
 ```
+
+### config.resolveUrl
+
+Some third-party scripts may require proxying through `config.resolveUrl()`, which runs inside the service worker. [Read more in Partytown's documentation.](https://partytown.qwik.dev/forwarding-events)
+
+However since `config` is serialized (even functions), it has limitations:
+
+- It cannot reference anything outside of the function scope
+- It can only use JavaScript
+
+For advanced use cases, you may need to pass data to this function while initializing Partytown. To do so, you can set `resolveUrl()` on `window.partytown` instead of the integration config:
+
+```astro title="Head.astro"
+---
+const proxyMap = {
+  "connect.facebook.net": "my-proxy.com"
+};
+---
+
+<script is:inline set:html={`
+window.partytown = {
+  resolveUrl: (url) => {
+    const proxyMap = ${JSON.stringify(proxyMap)};
+    url.hostname = proxyMap[url.hostname] || url.hostname;
+    return url;
+  },
+};
+`} />
+```
+
+Note that the integration config will override `window.partytown` if you set a property in both.
+
 
 ## Examples
 
--- a/src/content/docs/en/guides/integrations-guide/partytown.mdx
+++ b/src/content/docs/en/guides/integrations-guide/partytown.mdx
@@ -152,12 +152,32 @@ export default defineConfig({
 
 ### config.resolveUrl
 
-Some third-party scripts may require proxying through `config.resolveUrl()`, which runs inside the service worker. [Read more in Partytown's documentation.](https://partytown.qwik.dev/forwarding-events)
+Some third-party scripts may require [proxying](https://partytown.qwik.dev/proxying-requests/) through `config.resolveUrl()`, which runs inside the service worker. You can set this configuration option to check for a specific URL, and optionally return a proxied URL instead:
 
-However since `config` is serialized (even functions), it has limitations:
+```js title="astro.config.mjs" {7-13}
+export default defineConfig({
+  // ...
+  integrations: [
+    partytown({
+      // Example: proxy Facebook's analytics script
+      config: {
+        resolveUrl: (url) => {
+          const proxyMap = {
+            "connect.facebook.net": "my-proxy.com"
+          }
+          url.hostname = proxyMap[url.hostname] || url.hostname;
+          return url;
+        },
+      }
+    }),
+  ],
+});
+```
 
-- It cannot reference anything outside of the function scope
-- It can only use JavaScript
+However since `config` is serialized, some limitations apply:
+
+- Functions passed to the `config` object cannot reference anything outside of the function scope.
+- Functions passed to the `config` object can only be written in JavaScript.
 
 In some advanced use cases, you may need to pass data to this function while initializing Partytown. To do so, you can set `resolveUrl()` on `window.partytown` instead of the integration config:
 
--- a/src/content/docs/en/guides/integrations-guide/partytown.mdx
+++ b/src/content/docs/en/guides/integrations-guide/partytown.mdx
@@ -201,7 +201,6 @@ window.partytown = {
 
 Note that the integration config will override `window.partytown` if you set a property in both.
 
-
 ## Examples
 
 * [Browse projects with Astro Partytown on GitHub](https://github.com/search?q=%22%40astrojs%2Fpartytown%22+path%3A**%2Fpackage.json\&type=code) for more examples!
--- a/src/content/docs/en/guides/integrations-guide/partytown.mdx
+++ b/src/content/docs/en/guides/integrations-guide/partytown.mdx
@@ -109,6 +109,32 @@ export default defineConfig({
 
 This mirrors the [Partytown config object](https://partytown.builder.io/configuration).
 
+:::caution
+The `config` object is serialized and sent to the client. That means functions like `config.resolveUrl()` cannot use TypeScript, nor access anything outside of their scope.
+:::
+
+If you need to pass external data to functions when Partytown is initialized on the client, you can set the configuration through `window.partytown`:
+
+```astro title="Head.astro"
+---
+const proxyMap = {
+  "connect.facebook.net": "my-proxy.com"
+};
+---
+
+<script is:inline set:html={`
+window.partytown = {
+  resolveUrl: (url) => {
+    const proxyMap = ${JSON.stringify(proxyMap)};
+    url.hostname = proxyMap[url.hostname] || url.hostname;
+    return url;
+  },
+};
+`} />
+```
+
+It will be overriden by any config option coming from the integration.
+
 ### config.debug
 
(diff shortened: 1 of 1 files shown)


sarah11918 (commented):
Some general comments to get started:

Does this seem like a place to have a new section heading perhaps? Section headings > asides most of the time! (They are descriptive, let the reader know what the topic is, can be linked to in support threads or when needing to explain...) I'm going off of "what most people think is worth calling out is something that could be handy to be able to link people directly to."

In any event, we typically wouldn't use a "caution" (your project is in danger, you might lose data, etc.) for "how stuff works" (which is just actual stuff to be documented). We prefer not to use any asides at all, (by definition, asides are inherently "beside the point" and tangential to the actual content) but for sure this doesn't rise to the level of you might lose your project. :smile: 

If we made this a heading here, what are we describing? And just confirming that your caution and your example are both related under one banner (e.g. understanding that the `config` object is serialized is relevant background information that sets the stage for accomplishing what you might want to do with the example)? Your note mentioned "`config.resolveURL()` cannot use TypeScript" but I see nothing on this page that comes close to mentioning anything like this so it feels very out of left field (and why I wasn't even sure that your caution and example are related to each other). I think a proper section heading where we can set the context for what this content is about will help here!

florian-lefebvre:
Yeah a new heading would probably be better! However I would probably leave the caution contents under the Configuration heading because that applies all the time, no matter if you try to do something advanced. What do you think?

Something a bit tricky to explain btw is this:
```ts
const x = 'bar'

config: {
	forward: ['foo'],
	resolveUrl: (url) => {
		console.log(x)
		return url
	}
}
// serialized as (no x)
{
	forward: ['foo'],
	resolveUrl: (url) => {
		console.log(x)
		return url
	}
}
```
So the problem with references is only in functions

sarah11918:
Is it a problem that we don't seem to document `config.resolveURL`? If it's something that can be in the config, shouldn't it be documented? Are there other `@astrojs/partytown` config options we don't document?

florian-lefebvre:
We support all https://partytown.qwik.dev/configuration/ and actually, `resolveUrl` is the only function so i could reuse this heading. But yeah other than that, the only special case we have is about `config.lib` that we use to serve partytown assets ootb

sarah11918:
A great! Looks like we do link to that, but the link is a redirect so we might as well update the config link at the same time! 🙌

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -109,6 +109,32 @@ export default defineConfig({
 
 This mirrors the [Partytown config object](https://partytown.builder.io/configuration).
```
sarah11918:
```suggestion
This mirrors the [Partytown config object](https://partytown.qwik.dev/configuration/).
```
Seems like this needs updating!

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -109,6 +109,32 @@ export default defineConfig({
...
 
+:::caution
```
sarah11918: We recently added some wording in docs in a couple of places about what "serialized" means (with respect to passing props). (Out of the scope of this PR, but this is making me think that it would be really nice to have ONE big place we can refer to this, since it seems to be cropping up in a bunch of places!)

This is what we have in the [UI frameworks guide](https://docs.astro.build/en/guides/framework-components/#passing-props-to-framework-components) and what we have in the [server islands guide](https://docs.astro.build/en/guides/server-islands/#passing-props-to-server-islands). Is there any kind of description we want to provide that is a bit more at this level of detail here maybe under something like a `### Partytown config object` heading nested under `## Configuration`? 

What I think is helpful is that it's explained what serialized means in the specific context, and doesn't give a specific example ("`config.resolveURL()` can't use TypeScript") and doesn't need squishing multiple concepts (like scope) into one sentence. If you were to imitate a paragraph similar to the passing props description we use, what would that look like?

BUT ALSO: remembering that not everything that is true belongs in docs 😅 ... just BECAUSE this is true, is it required "this is the first time I'm seeing/using the integration and configuring it" knowledge. OR, does someone only need this true knowledge *if* they are doing something advanced/customized outside of the norm? In which case, it could very well belong with the more advanced use case so that it doesn't clutter the simple act of configuring this integration for people who don't need to see it. AND we would make sure whatever heading is chosen for that setting makes it clear when THAT kind of configuration is necessary.

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -109,6 +109,32 @@ export default defineConfig({
...
+The `config` object is serialized and sent to the client. That means functions like `config.resolveUrl()` cannot use TypeScript, nor access anything outside of their scope.
+:::
+
+If you need to pass external data to functions when Partytown is initialized on the client, you can set the configuration through `window.partytown`:
```
sarah11918: OK, so here we definitely want a header with context. What is the main thing someone is trying to do here? Goal they're trying to achieve? (And, see above that maybe that earlier info about serialization happens here, which I'll demonstrate below.)

If this is an advanced use case (which I'm assuming since you said you wanted to split the info up from the previous statement since it applies all the time, not just in advanced cases), then it's worth making it stand out that this *is* an advanced case that someone might not have to worry about? People will see this in "Configuration" and could think this is something they need to know/do.

That means the header could also possibly indicate that this is advanced (if it is?) or sometimes we have a disclaimer in the initial sentence something like:

"In most cases, you will be able to configure the Partytown integration using the `config` object. However, this object is [serialized](mozilla link)... (stuff discussed by what was in your caution)"

Then we can have whatever content you feel is helpful, explained not needing to squish things into a note, and then this: "To pass external data to functions...." example fits nicely!

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -109,6 +109,32 @@ export default defineConfig({
...
+  resolveUrl: (url) => {
+    const proxyMap = ${JSON.stringify(proxyMap)};
+    url.hostname = proxyMap[url.hostname] || url.hostname;
+    return url;
+  },
+};
+`} />
+```
+
+It will be overriden by any config option coming from the integration.
```
sarah11918:
```suggestion
Configuration set in a client-side script will be overriden by any config options set by the Partytown integration's config object.
```
We can make this accurate, this is just a quick suggestion. But this statement is a little vague and if one thing is going to override another, we should be very explicit about what's happening! (No generic"it"; does "coming from" mean set in config? etc.)

Also, admittedly I don't know this integration or how to use it, but it feels weird to show something and then say, "but this thing I just showed you will be overriden" -- (which is also why we need to say exactly what's what, because maybe I have the wrong idea about what "it" is! But as written, we'd assume "it" is the stuff you set in the script, so "it" gets overridden.)

florian-lefebvre: So really it's looking like this:
```js
{
	...window.partytown,
	...configFromIntegration // always contains debug AFAIK
}
```

src/content/docs/en/guides/integrations-guide/partytown.mdx
florian-lefebvre: So a few things:
1. The serialization is different than UI framework props as you said:
	a. ui comp props: functions get serialized to null
	b. partytown config: functions are passed as is to the client ([source](https://github.com/QwikDev/partytown/blob/main/src/lib/utils.ts#L228-L233)). To be precise, the whole config is basically passed "as is" (ie. represented as string)
2. We can probably remove the `resolveUrl()` specific bit and put it under its own heading

src/content/docs/en/guides/integrations-guide/partytown.mdx
florian-lefebvre: I think this whole part (about `window.partytown`) can go under the new `config.resolveUrl()` heading (to be done still as part of this PR). Wdyt?

sarah11918:
@florian-lefebvre I think we're getting to the stage here where it's going to be helpful for you to actually write/update the text content with what you're thinking after our conversation so we can review it as actual docs content. 

Do you feel confident just taking another pass at the content here so we can start reviewing from that?

florian-lefebvre:
Yes definitely! Just wanted to see if we agreed on things but looks like it is, I'll try to work on it tomorrow otherwise next week

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -107,11 +107,13 @@ export default defineConfig({
 });
 ```
 
-This mirrors the [Partytown config object](https://partytown.builder.io/configuration).
+This mirrors the [Partytown config object](https://partytown.qwik.dev/configuration/).
+
+The `config` object is serialized, even [`config.resolveUrl()`](#configresolveurl), because it is sent to the client.
```
sarah11918: I really feel like this is jumping out as one of those "true, but..." Fun Facts that doesn't feel like it belongs here. And, the same information is added in the `config.resolveUrl` section, so it's duplicated content presented here without any extra content: it's not really offering a reason for why this fact is included here.

I honestly think this probably doesn't need to be here at all, and the new section can stand on its own. That's the place we're most concerned about making sure people know the config object is serialized, and where we can actually explain what we mean by serialized, and why that's a thing they might need to know or care about, in a place where they're reading about something where it very much matters.

```suggestion
```

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -149,6 +151,38 @@ export default defineConfig({
 });
 ```
 
+### config.resolveUrl
+
+Some third-party scripts may require proxying through `config.resolveUrl()`, which runs inside the service worker. [Read more in Partytown's documentation.](https://partytown.qwik.dev/forwarding-events)
```
sarah11918:
```suggestion
Some third-party scripts may require proxying through `config.resolveUrl()`, which runs inside the service worker. You can set this configuration option to check for a specific URL, and optionally return a proxied URL instead.
```

1. Reminder that this firstly has to sound like some kind of reference for what `config.resolveUrl()` *is*, which we haven't done yet! I just snagged wording from Partytown's own docs, but it might not be exactly right here so please do edit as necessary! Just make sure we're at least explaining what `resolveUrl` is/does, however briefly, in some way first here!

2. I feel like we *probably* should show an example of what this option looks like when set (remember, this is supposed to be our documentation of `config.resolveUrl()`!) Which means having an example where you *don't* need the advanced case. I'd do something like:
	> The following example configures checking whether the URL to be resolved is for Google Analytics, and then...
	> [code example of the partytown config]

	*Then* I think we can go on to the next paragraph about how it's serialized so therefore there are limitations.

3. (Hilariously, this is gonna be a SUPER long comment about just the link lol. Get ready!) 
	So, my first instinct was to do something like this, make the link inline so the text for what the link is about is descriptive:

	> Some third-party scripts may require [proxying through `config.resolveUrl()`](https://partytown.qwik.dev/forwarding-events), which runs inside the service worker.

	But I checked, and you're linking to "Forwarding events and Triggers" which mentions neither proxying nor `resolveURL` on that page. (Yet, there *is* a [Proxying requests](https://partytown.qwik.dev/proxying-requests/) page that *does* mention `resolveURL`! And, the `resolveUrl` config docs link to the proxying request page.) So I guess that's not what we should do here. 😅 

	I'm assuming you do want the Forwarding Events page linked because I can see it talks about using strings and serialization, so that's probably the relevant content. But then, maybe the link doesn't belong *right here*? It kind of blocks the flow of the point you're trying to make "some scripts require going through `config.resolveURL` -> but, Astro serializes -> so there are limitations".

 	And, it's not key that people go read and learn about this right now before moving on. The more likely scenario is that people already *do* know about this, and they are trying to read about how to properly use `resolveUrl` with the Astro adapter. I don't think reading this part of *our* docs is going to make them think, "Oh, cool, I should go learn about this new thing and try it in my project." It's the other way around!

	So, I would probably be more likely to include this link *later*, maybe even at the very end of this section. The intention is, "if you didn't figure out what you needed from this section in our docs here, go read all about it elsewhere." And then the link text just needs to be a bit more descriptive about what is at the link. So it could be something like:

> See [Partytown's documentation on forwarding events](https://partytown.qwik.dev/forwarding-events) for more information.

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -149,6 +151,38 @@ export default defineConfig({
...
+
+However since `config` is serialized (even functions), it has limitations:
+
+- It cannot reference anything outside of the function scope
+- It can only use JavaScript
```
sarah11918:
```suggestion
- It cannot reference anything outside of the function scope.
- It can only use JavaScript.
```
Firstly, just adding periods above. 

But also:
1. "It can only use JavaScript" means... as opposed to what? What *can't* it use that people might expect? TypeScript? Something else? (See my next point below too for how "it" and "use" are both extremely vague, so that makes this sentence feel even more like it's prompting more questions than it answers 😄 )

2. There is a lot of "it" going on. Do all the "it"'s refer to "the `config` object"? If so, is that exactly the thing you mean as a subject? e.g. "The config object can only use JavaScript" (does this object actually "use" anything?) Or, is there a more specific nuance like "Functions passed to the `config` object cannot reference anything outside of the function scope" or "Functions passed as config options like `resolveUrl` can only be written in JavaScript".

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -149,6 +151,38 @@ export default defineConfig({
...
+
+For advanced use cases, you may need to pass data to this function while initializing Partytown. To do so, you can set `resolveUrl()` on `window.partytown` instead of the integration config:
```
sarah11918:
```suggestion
In some advanced use cases, you may need to pass data to this function while initializing Partytown. To do so, you can set `resolveUrl()` on `window.partytown` instead of the integration config:
```
Great intro here!

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -149,6 +151,38 @@ export default defineConfig({
...
+
+```astro title="Head.astro"
```
sarah11918:
``````suggestion
```astro title="Head.astro" {8-14}
``````
I think these linkes should be the entire `window.partytown` block. Makes sense to highlight that, I think!

**Review by sarah11918 — COMMENTED** [review]
OK, I realized my review looks LOOONG but actually, this is very good and it's very close! 😅

The new structure here looks amazing, and I just have some polishing thoughts below.

File: src/content/docs/en/guides/integrations-guide/partytown.mdx (hunk 6 above)
Comment (florian-lefebvre) [reply src/content/docs/en/guides/integrations-guide/partytown.mdx]: Re point 3: I copied the wrong link...I wanted the proxying requests one!

File: src/content/docs/en/guides/integrations-guide/partytown.mdx (hunk 7 above)
Comment (florian-lefebvre) [reply src/content/docs/en/guides/integrations-guide/partytown.mdx]: > "Functions passed to the config object cannot reference anything outside of the function scope" AND "Functions passed as config options like resolveUrl can only be written in JavaScript".

Yeah exactly this!

File: src/content/docs/en/guides/integrations-guide/partytown.mdx [hunk 10]
```diff
@@ -149,6 +150,57 @@ export default defineConfig({
...
+      }
+    }),
+  ],
+});
+```
+
+However since `config` is serialized, some limitations apply:
+
+- Functions passed to the `config` object cannot reference anything outside of the function scope.
+- Functions passed to the `config` object can only be written in JavaScript.
```
sarah11918:
```suggestion
However since the `config` object is serialized when sent to the client, some limitations on functions passed to your configuration apply:

- Functions cannot reference anything outside of the function scope.
- Functions can only be written in JavaScript.
```
This is great! Just wondering if we can:
- remove some duplication by extracting "functions passed to config" and writing it above

- add the proper terminology to add one detail about serialization (e.g. "when sent to the client", "before being sent to the client", "and sent to the client", "for use on the client"... choose anything that is accurate!) ... I think one tiny extra bit of context here makes a stronger sentence!

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -149,6 +150,57 @@ export default defineConfig({
...
+---
+
+<script is:inline set:html={`
+window.partytown = {
+  resolveUrl: (url) => {
+    const proxyMap = ${JSON.stringify(proxyMap)};
+    url.hostname = proxyMap[url.hostname] || url.hostname;
+    return url;
+  },
+};
```
sarah11918:
```suggestion
  window.partytown = {
    resolveUrl: (url) => {
      const proxyMap = ${JSON.stringify(proxyMap)};
      url.hostname = proxyMap[url.hostname] || url.hostname;
      return url;
    },
  };
```
I just indented everything by one tab, because the preview was showing this: 
image

Can you double check this suggestion looks right, then otherwise, we're good to merge!

florian-lefebvre: Only issue I can see is we're mixing space and tabs, you may need to redo all the indentation :thinking:

sarah11918: Thanks, updated to all spaces and will commit like this. Assuming the preview looks good, then you can merge!

PR #12182, opened 2025-08-22
add tip to build forms recipe to consider actions

### LINKED ISSUE #10374: Update the build forms recipe to use actions instead.
### 📚 Subject area/topic

HTML forms, and Actions

https://docs.astro.build/en/recipes/build-forms/

This recipe uses standard handling of html. A section should be included to use actions instead because that's the updated way and a more cleaner way of handling forms in astro.

_No response_

sarah11918:
#### Description (required)

This PR adds a tip to the "Build HTML forms" recipe linking to actions as an alternate way to build forms with additional DX features.

This recipe predates actions, but is still valid for people who want to build HTML forms without actions. So it should not be replaced, but it is helpful to let people know here that actions exist.

- Closes #10374 

(Closing the issue with the proposed solution to link to the actions page since there isn't currently activity around updating/creating a new recipe. This doesn't preclude someone adding a new recipe using actions later!)

### DIFF
--- a/src/content/docs/en/recipes/rss.mdx
+++ b/src/content/docs/en/recipes/rss.mdx
@@ -163,7 +163,11 @@ This method is deprecated for all versions of Astro since v2.1.0, and cannot be 
 
 <p><Since v="1.6.14" /></p>
 
-The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
+Set the `content` key on `rss.items` to provide the full content of a post as HTML. This allows `@astrojs/rss` to make the full Markdown text of your post available to RSS feed readers. Images and links with full URL paths are also supported. However, images and internal links to other pages using relative paths are not.
+
+When rendering full post content, you will have to consider images, relative links, styles, scripts, and other elements beyond standard Markdown text that you may have in your posts. You may need to include additional logic in your `src/pages/rss.xml.js` endpoint to account for these, or to remove elements that are unnecessary for an RSS feed (e.g. those that are used only for styling or interaction on your website).
+
+You can see [one specific community implementation](https://github.com/delucis/astro-blog-full-text-rss/blob/latest/src/pages/rss.xml.ts) that addresses some of these concerns for an example of how to proceed.
 
 :::tip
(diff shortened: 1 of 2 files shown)


src/content/docs/en/recipes/build-forms.mdx
```diff
@@ -8,6 +8,10 @@ import { Steps } from '@astrojs/starlight/components';
 
 Astro pages that are rendered on demand can both display and handle forms. In this recipe, you'll use a standard HTML form to submit data to the server. Your frontmatter script will handle the data on the server, sending no JavaScript to the client.
 
+:::tip[Build forms with Astro actions]
+In v5.0, Astro added actions which provide several benefits over basic HTML forms including validating your form data and updating your UI based on the form submission. To use this method for building forms instead, see our [actions guide](/en/guides/actions/) to learn more about these features.
```
sarah11918:
```suggestion
In v4.15, Astro added actions which provide several benefits over basic HTML forms including validating your form data and updating your UI based on the form submission. To use this method for building forms instead, see our [actions guide](/en/guides/actions/) to learn more about these features.
```

src/content/docs/en/recipes/build-forms.mdx
```diff
@@ -8,6 +8,10 @@ import { Steps } from '@astrojs/starlight/components';
 
 Astro pages that are rendered on demand can both display and handle forms. In this recipe, you'll use a standard HTML form to submit data to the server. Your frontmatter script will handle the data on the server, sending no JavaScript to the client.
 
+:::tip[Build forms with Astro actions]
```
yanthomasdev: "Astro Actions" is capitalized in the main guide, so I generally think we should keep that convention when talking about the feature's "brand name"
```suggestion
:::tip[Build forms with Astro Actions]
```

bjohansebas (approved):
LGTM, thanks

src/content/docs/en/recipes/build-forms.mdx
sarah11918: Ah, I forgot we let Ben get away with this. Usually our features are not capitalized, only APIs and components. But you're right: we do capitalize when we say Astro Actions in full. (And otherwise, just "actions" when its on its own)

PR #12185, opened 2025-08-22
context about limited full post content support in RSS

### LINKED ISSUE #10946: RSS package doesn't support images
### Astro Info

```block
Astro                    v5.1.0
Node                     v23.1.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
                         @astrojs/mdx
```

_No response_

### Describe the Bug

Including full post content is RSS 101. [The recipe to that](https://docs.astro.build/en/recipes/rss/#including-full-post-content) doesn't produce valid image paths.

The recipe should generate content with valid image paths. Ideally it would with MDX too.

https://docs.astro.build/en/recipes/rss/#including-full-post-content

### Participation

- [ ] I am willing to submit a pull request for this issue.

sarah11918:
Adds additional context and guidance re: full post content in RSS feeds (what kinds of images are/aren't supported; things to consider; example to consult)

- Closes #10946

### DIFF
--- a/src/content/docs/en/recipes/rss.mdx
+++ b/src/content/docs/en/recipes/rss.mdx
@@ -163,7 +163,11 @@ This method is deprecated for all versions of Astro since v2.1.0, and cannot be 
 
 <p><Since v="1.6.14" /></p>
 
-The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
+The `content` key contains the full content of a post as HTML. This allows `@astrojs/rss` to make the full Markdown text of your post available to RSS feed readers. Images and links with full URL paths are also supported. However, images and internal links to other pages using relative paths are not.
+
+When rendering full post content, you will have to consider images, relative links, styles, scripts, and other elements beyond standard Markdown text that you may have in your posts. You may need to include additional logic in your `src/pages/rss.xml.js` endpoint to account for these, or to remove elements that are unnecessary for an RSS feed (e.g. those that are used only for styling or interaction on your website).
+
+You can see [one specific community implementation](https://github.com/delucis/astro-blog-full-text-rss/blob/17c14db9b4fd68a20f097f5ad8ae66edb2da1815/src/pages/rss.xml.ts) that addresses some of these concerns for an example of how to proceed.
 
 :::tip
 A package like [`sanitize-html`](https://www.npmjs.com/package/sanitize-html) will make sure that your content is properly sanitized, escaped, and encoded. In the process, such a package might also remove some harmless elements and attributes, so make sure to verify the output and configure the package according to your needs.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/recipes/rss.mdx
```diff
@@ -163,7 +163,11 @@ This method is deprecated for all versions of Astro since v2.1.0, and cannot be
 
 <p><Since v="1.6.14" /></p>
 
-The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
+The `content` key contains the full content of a post as HTML. This allows `@astrojs/rss` to make the full Markdown text of your post available to RSS feed readers. Images and links with full URL paths are also supported. However, images and internal links to other pages using relative paths are not.
+
+When rendering full post content, you will have to consider images, relative links, styles, scripts, and other elements beyond standard Markdown text that you may have in your posts. You may need to include additional logic in your `src/pages/rss.xml.js` endpoint to account for these, or to remove elements that are unnecessary for an RSS feed (e.g. those that are used only for styling or interaction on your website).
+
+You can see [one specific community implementation](https://github.com/delucis/astro-blog-full-text-rss/blob/17c14db9b4fd68a20f097f5ad8ae66edb2da1815/src/pages/rss.xml.ts) that addresses some of these concerns for an example of how to proceed.
```
delucis: Might make sense to avoid the permalink in favour of the branch name in this case? Otherwise if changes are made to the solution (and they probably should be as I haven’t updated that repo for a while 😅), docs readers following the link would always see the old code:

```suggestion
You can see [one specific community implementation](https://github.com/delucis/astro-blog-full-text-rss/blob/latest/src/pages/rss.xml.ts) that addresses some of these concerns for an example of how to proceed.
```

src/content/docs/en/recipes/rss.mdx
```diff
@@ -163,7 +163,11 @@ This method is deprecated for all versions of Astro since v2.1.0, and cannot be
 
 <p><Since v="1.6.14" /></p>
 
-The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
+The `content` key contains the full content of a post as HTML. This allows `@astrojs/rss` to make the full Markdown text of your post available to RSS feed readers. Images and links with full URL paths are also supported. However, images and internal links to other pages using relative paths are not.
```
delucis: I found this first sentence confusing. Going back and reading more of the guide, do I understand correctly that `content` here refers to the `content` key which a user _sets_ when passing stuff to the RSS package?

Reading that it “contains the full content”, makes it sound like it’s a property that already includes that stuff for us to use.

Maybe something like this?

```suggestion
Set the `content` key on `rss.items` to provide the full content of a post as HTML. This allows `@astrojs/rss` to make the full Markdown text of your post available to RSS feed readers. Images and links with full URL paths are also supported. However, images and internal links to other pages using relative paths are not.
```

sarah11918: Great improvement! I just updated your suggestion from "RSS items" to `rss.items` because it made me think at first I had to add something to each of my blog posts ("items") vs. adding on the `items` property! So I think the code syntax helps here!

delucis (approved):
I am content!

PR #12186, opened 2025-08-22
document that Font API downloads fonts

### LINKED ISSUE #11492: Add clarification that remote fonts are downloaded
### 📚 Subject area/topic

Fonts

https://docs.astro.build/en/reference/experimental-flags/fonts/

Add some clarification about how remote fonts are handled, specifically that they are downloaded to the project

This, in my case, actually caused the "remote" fonts to be *more* performant than when using the same fonts locally. 
(I presume this is due to them being downloaded to node_modules somewhere)

Additionally, this will clarify to users concerned about GDPR that the fonts are being downloaded and effectively "self-hosted"

_No response_

sarah11918:
Adds extra context about how the Fonts API works (downloads, caches fonts) and why this is good (without overpromising security)

- Closes #11492


bjohansebas (approved):
LGTM

PR #12188, opened 2025-08-22
tip to restart dev server if collections schema not recognized

### LINKED ISSUE #11188: Content Collections Guide Doesn't Mention To ReRun Astro
### 📚 Subject area/topic

Add Content To Your Site / Content Collections

https://docs.astro.build/en/guides/content-collections/

I was having a problem with the guide because I couldn't figure out how to get astro to load my collections, turns out it didnt find my collections because I hadn't restarted astro/ran astro sync

there should be a warning to remind the reader to rerun astro after adding a collection, imho

_No response_

sarah11918:
Ideally, this shouldn't be necessary, but this guidance follows what we do say in the tutorial.

Either way, the corresponding issue should be closed: either with this additional content, or by deciding that not recognizing the schema is an Astro bug and not a docs issue.

This PR is the solution in the first case!

- Closes #11188


src/content/docs/en/guides/content-collections.mdx
```diff
@@ -204,6 +204,10 @@ Schemas enforce consistent frontmatter or entry data within a collection through
 
 Schemas also power Astro's automatic TypeScript typings for your content. When you define a schema for your collection, Astro will automatically generate and apply a TypeScript interface to it. The result is full TypeScript support when you query your collection, including property autocompletion and type-checking.
 
+:::tip
+In order for Astro to recognize a new or updated schema, you may need to restart the dev server to define the `astro:content module`.
```
```suggestion
In order for Astro to recognize a new or updated schema, you may need to restart the dev server to define the `astro:content` module.
```

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -204,6 +204,10 @@ Schemas enforce consistent frontmatter or entry data within a collection through
 
 Schemas also power Astro's automatic TypeScript typings for your content. When you define a schema for your collection, Astro will automatically generate and apply a TypeScript interface to it. The result is full TypeScript support when you query your collection, including property autocompletion and type-checking.
 
+:::tip
+In order for Astro to recognize a new or updated schema, you may need to restart the dev server to define the `astro:content` module.
```
ArmandPhilippot: Since the issue was opened, we have updated the [CLI reference to include the shortcut](https://docs.astro.build/en/reference/cli-reference/#astro-dev), so I wonder should we mention it with a link perhaps?
```suggestion
In order for Astro to recognize a new or updated schema, you may need to restart the dev server or [sync the content layer](/en/reference/cli-reference/#astro-dev) (<code>s + enter</code>) to define the `astro:content` module.
```

ArmandPhilippot (approved):
A small (optional) suggestion, otherwise LGTM! 🙌🏽

src/content/docs/en/guides/content-collections.mdx
bjohansebas: Is that possible? Wow, incredible!

sarah11918: Great idea, but rather than just telling someone which buttons to push, we should also be telling them the name of the action they're performing, right? :smile: 

like, "...restart the dev server or [sync the content layer](https://docs.astro.build/en/reference/cli-reference/#astro-dev) (<code>s + enter</code>) to define..."

ArmandPhilippot: I agree that would be better 😄 I've updated the suggestion (and I also fixed the link to remove the domain 👀 ).

sarah11918:
Ok, our reservation about documenting a workaround for buggy behaviour aside, I think Armand has made this PR worth the price of admission by adding the shortcut that no one seems to know about! :joy: 

Besides, if we have a shortcut, it must be a normal thing to do, right? :sweat_smile: 

So I'm going to make the judgment call and say this is helpful, and we can revisit if it causes problems.

bjohansebas (approved):
LGTM

PR #5541, opened 2023-12-01
[i18nIgnore] Rename Dev Overlay to Dev Toolbar

matthewp:
#### Description (required)

Renames **Dev Overlay** to **Dev Toolbar** as well as **plugins** to **apps**.

#### Related issues & labels (optional)

For Astro 4

PR #12189, opened 2025-08-22
fix(dev-toolbar-app-reference): API reference format + new component

ArmandPhilippot:
#### Description (required)

Updates `dev-toolbar-app-reference.mdx` to:
* use [API reference format](https://contribute.docs.astro.build/reference/api-references/)
* add a missing `astro-dev-toolbar-select` component added in https://github.com/withastro/astro/pull/10591
* complete the `icons` list (see https://github.com/withastro/astro/blob/02366e9ce38df8e7362817c095ff05ae61dc7b56/packages/astro/src/runtime/client/dev-toolbar/ui-library/icons.ts#L25)

**Here's the details of the replacements:**
* Replaces `Signature` with `Type` for consistency. I think this is the only page using `Signature:`.
* Adds missing `Type`/`Default`/`Since`.
* Fixes the `init()` return type (can be a `Promise`, see https://github.com/withastro/astro/blob/15b55f34cb84ecfb99d2e76918a567a00bbb13f6/packages/astro/src/types/public/toolbar.ts#L53-L57).
* Fixes the `beforeTogglingOff()` return type (can be a `Promise` but can't be `void`, see https://github.com/withastro/astro/blob/15b55f34cb84ecfb99d2e76918a567a00bbb13f6/packages/astro/src/types/public/toolbar.ts#L58)
* Fixes an incorrect method name: `onInitialized` should be `onAppInitialized` (see https://github.com/withastro/astro/blob/02366e9ce38df8e7362817c095ff05ae61dc7b56/packages/astro/src/integrations/hooks.ts#L137)
* Slightly rewords the `addDevToolbarApp` text because the `icon` property is optional while others are required.
* Moves `toggle-style` in `astro-dev-toolbar-toggle` (see https://github.com/withastro/astro/blob/02366e9ce38df8e7362817c095ff05ae61dc7b56/packages/astro/src/runtime/client/dev-toolbar/ui-library/toggle.ts#L10)
* Adds `radio-style` to be consistent with other components (see https://github.com/withastro/astro/blob/02366e9ce38df8e7362817c095ff05ae61dc7b56/packages/astro/src/runtime/client/dev-toolbar/ui-library/radio-checkbox.ts#L6)
* Renames the `style` property in `astro-dev-toolbar-highlight` to use `highlight-style` and to stay consistent with other components (see https://github.com/withastro/astro/blob/02366e9ce38df8e7362817c095ff05ae61dc7b56/packages/astro/src/runtime/client/dev-toolbar/ui-library/highlight.ts#L102)
* Fixes a code snippet highlighting to include function closure

**Doubts:**
  * `addDevOverlayPlugin()` has been replaced with `addDevToolbarApp()` in `4.0.0`. However the signature has changed in `4.7.0`. So I'm not entirely sure if we should use `4.0.0` or `4.7.0` here. See:
    * https://github.com/withastro/astro/pull/9271
    * https://github.com/withastro/docs/pull/5541
    * https://github.com/withastro/astro/pull/10665
  * I set a default for `level` while the property is optional and the default seems to be set elsewhere [looking at the implementation](https://github.com/withastro/astro/blob/15b55f34cb84ecfb99d2e76918a567a00bbb13f6/packages/astro/src/runtime/client/dev-toolbar/helpers.ts#L36). So, should we use a default or not? Note, I've checked and `error` is still the default as described in the text.
  * Since `icon` is optional, should we precise a default? Without icon, `?` is displayed so I used that as value (see https://github.com/withastro/astro/blob/02366e9ce38df8e7362817c095ff05ae61dc7b56/packages/astro/src/runtime/client/dev-toolbar/entrypoint.ts#L211).

In case anyone want to check the types/components, here's a small app using almost all the APIs: https://stackblitz.com/edit/astro-dev-toolbar-app-example?file=my-toolbar-app%2Fmy-integration.ts,my-toolbar-app%2Fmy-app.ts&on=stackblitz (you can download it and open it in your editor to be able to see the signatures)

- Suggested label:  `add new content`, `consistency/formatting`


src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -35,10 +34,21 @@ export default () => ({
...
 ### `addDevToolbarApp()`
 
-A function available to [the `astro:config:setup` hook](/en/reference/integrations-reference/#astroconfigsetup) that adds dev toolbar apps. It takes an object with the following required properties to define the toolbar app: [`id`](#id), [`name`](#name), [`icon`](#icon), and [`entrypoint`](#entrypoint).
+<p>
+
+**Type:** `(entrypoint: DevToolbarAppEntry) => void`<br />
+<Since v="4.0.0" />
+</p>
+
+A function available to [the `astro:config:setup` hook](/en/reference/integrations-reference/#astroconfigsetup) that adds dev toolbar apps. It takes an object with the following required properties to define the toolbar app: [`id`](#id), [`name`](#name), and [`entrypoint`](#entrypoint). Optionally, you can also define an [`icon`](#icon) for you app.
```
sarah11918:
```suggestion
A function available to [the `astro:config:setup` hook](/en/reference/integrations-reference/#astroconfigsetup) that adds dev toolbar apps. It takes an object with the following required properties to define the toolbar app: [`id`](#id), [`name`](#name), and [`entrypoint`](#entrypoint). Optionally, you can also define an [`icon`](#icon) for your app.
```

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -62,6 +77,12 @@ The name of the app. This will be shown to users whenever the app needs to be re
 
 ### `icon`
 
+<p>
+
+**Type:** `<a href="#icons">Icon</a> | string`<br />
+**Default**: `"?"`
```
sarah11918, line 83: This threw me for a second! :joy:

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -35,10 +34,21 @@ export default () => ({
 
 ### `addDevToolbarApp()`
 
-A function available to [the `astro:config:setup` hook](/en/reference/integrations-reference/#astroconfigsetup) that adds dev toolbar apps. It takes an object with the following required properties to define the toolbar app: [`id`](#id), [`name`](#name), [`icon`](#icon), and [`entrypoint`](#entrypoint).
+<p>
+
+**Type:** `(entrypoint: DevToolbarAppEntry) => void`<br />
+<Since v="4.0.0" />
```
sarah11918, line 40: I agree with using v4.0.0 here, since that's when this was introduced!

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
```diff
@@ -270,11 +336,22 @@ app.toggleNotification({
...
 
 Indicates whether or not the app has a notification for the user. When `true`, the app icon will be highlighted. Conversely, when `false`, the highlight will be removed. If this property is not specified, `true` will be assumed.
 
-#### `level: 'error' | 'warning' | 'info'`
+#### `level`
+
+<p>
+
+**Type:** `'error' | 'warning' | 'info'`<br />
+**Default:** `'error'`
```
sarah11918, line 353: I agree that it looks like from the implementation that this is undefined if not set. 

Short of testing, it seems like this should be undefined, and the text below "If this property is not specified..." is incorrect?

ArmandPhilippot: The opposite, the text is right but it seems the "default" is set elsewhere in the implementation. So this is maybe more a "fallback" than a default (which would be `undefined`). But the result is the same. So I think, it should be fine if we show `'error'` as default.

In case anyone want to check, in the Stackblitz repro, you just need to delete the l164 in `./my-toolbar-app/my-app.ts`. The small warning triangle will be replaced with a red dot on the app button.

sarah11918: OK, then happy with this as is! If the guidance is accurate for the reader in practice, then that's what matters I think!

sarah11918 (approved):
Just checked the updated deploy preview and everything looks (visually) good on the page!

src/content/docs/en/reference/dev-toolbar-app-reference.mdx
bjohansebas: same
Image

Image

bjohansebas (approved):
LGTM! thanks

PR #12194, opened 2025-08-22
Updates CMS guide for Strapi 5

### LINKED ISSUE #11907: Update the Strapi guide to version 5 and use the official Strapi client library instead of a custom function
### 📚 Subject area/topic

Strapi CMS

https://docs.astro.build/fr/guides/cms/strapi/

The Astro guide for Strapi is outdated regarding the data format returned by the Strapi API. Since Strapi v5, the response no longer includes the attributes field. Additionally, it now uses a documentId field.
👉 See the [breaking change documentation](https://docs.strapi.io/cms/migration/v4-to-v5/breaking-changes/use-document-id) for more details.

Additionally, Strapi now provides an official client library that simplifies API calls and should be preferred over writing custom fetch logic.
📚 Official client documentation: [Strapi Client](https://docs.strapi.io/cms/api/client)

It would be better for the guide to be updated to reflect these changes and leverage the official client.

_No response_

sarah11918:
Updates the shape of the data returned from Strapi 5 per [Strapi's migration guide breaking changes](https://docs.strapi.io/cms/migration/v4-to-v5/breaking-changes/use-document-id)

Specifically:
- `id` is replaced by `documentId`
- there is no longer a nested `attributes` object for other data

This PR update the code blocks that show returned data and types, and additionally updates using the data in component templates because data is no longer on an `attributes` object.

According to the issue, there are other improvements that could be made for using Strapi 5 with Astro, but this at least updates the code blocks to show correct data.

- Closes #11907

### DIFF
--- a/src/content/docs/en/guides/cms/strapi.mdx
+++ b/src/content/docs/en/guides/cms/strapi.mdx
@@ -117,16 +117,14 @@ If you are using TypeScript, create the following Article interface to correspon
 
 ```ts title="src/interfaces/article.ts"
 export default interface Article {
-  id: number;
-  attributes: {
-    title: string;
-    description: string;
-    content: string;
-    slug: string;
-    createdAt: string;
-    updatedAt: string;
-    publishedAt: string;
-  };
+  documentId: number;
+  title: string;
+  description: string;
+  content: string;
+  slug: string;
+  createdAt: string;
+  updatedAt: string;
+  publishedAt: string;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/cms/strapi.mdx
```diff
@@ -293,23 +289,23 @@ const article = Astro.props;
...
 <html lang="en">
   <head>
-    <title>{article.attributes.title}</title>
+    <title>{article.title}</title>
   </head>
 
   <body>
     <main>
-      <img src={import.meta.env.STRAPI_URL + article.attributes.image.data.attributes.url} />
+      <img src={import.meta.env.STRAPI_URL + article.image.data.url} />
```
sarah11918, line 297: admittedly, I don't know for *sure* that I can just remove two `.attributes` like this? Is it safe to assume if one is gone, then they've also updated the shame of image data?

src/content/docs/en/guides/cms/strapi.mdx
```diff
@@ -350,23 +346,23 @@ try {
...
```
sarah11918, line 354: Here's the other place with image data. I'll see if I can confirm in Strapi's docs.

ArmandPhilippot (approved):
I'm not familiar with Strapi but, looking at their docs, this looks good to me! 🙌🏽

src/content/docs/en/guides/cms/strapi.mdx
ArmandPhilippot: Oh I didn't see your comments before approving. Yeah, this was also what I wondered. The breaking change guide is specific to the id (while showing flattened attributes), but a search using `attributes` doesn't give much more information... I'll try to see if I can find a better example of the new API too!

src/content/docs/en/guides/cms/strapi.mdx
sarah11918: OK, found a [code sample example](https://docs.strapi.io/cms/api/rest/guides/understanding-populate#example-with-populate) here that seems to use `image.data.url` and no mention of an `attributes` level, so I'm going with this!

PR #12229, opened 2025-08-27
Install tailwindcss rather than deprecated tailwind package

filiptammergard:
#### Description

When I followed these instructions I got this message in the terminal: 

>  WARN  deprecated tailwind@4.0.0: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.

Should use tailwindcss instead!

### DIFF
--- a/src/content/docs/en/guides/styling.mdx
+++ b/src/content/docs/en/guides/styling.mdx
@@ -447,17 +447,17 @@ In Astro `>=5.2.0`, use the `astro add tailwind` command for your package manage
 <PackageManagerTabs>
   <Fragment slot="npm">
     ```shell
-    npx astro add tailwind
+    npx astro add tailwindcss
     ```
   </Fragment>
   <Fragment slot="pnpm">
     ```shell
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi! Thank you for helping maintain the accuracy of our docs!

The current docs are correct in this case, though. When installing with `astro add`, the instruction is `tailwind` as we intend for this to be an alias for the package when installed this way. Given your error message, perhaps you were trying to `npm install` the package directly (which *is* `tailwindcss` as you correctly noted)?

I will close this PR as the docs do not require updating, but if you need assistance installing with this command, you can visit our community at https://astro.build/chat and they can help you determine whether there may in fact be a bug with the command and an issue should be filed on the main Astro repo. (And, in the case of a bug, we fix the bug instead of updating documentation!)

But it looks like you got Tailwind installed eventually, so I hope you're no longer blocked! Thank you for letting us know that you had a problem so we could confirm that our documentation is correct, just in case this was something that needed to be updated! 🙌

PR #2779, opened 2023-03-02
Add assets page

natemoo-re:
#### What kind of changes does this PR include?

- New or updated content

#### Description

- Adds **Assets (Experimental)** to sidebar to coincide with https://github.com/withastro/astro/pull/6344, https://github.com/withastro/roadmap/pull/500
- Explains public API surface of new `astro:assets` module

#### Open Questions

- How much do we want to integrate this into the Images guide? Is the top-level warning and pointing to the other page sufficient?

PR #12230, opened 2025-08-27
fix(image-service-reference): add API reference blocks + fix snippets

ArmandPhilippot:
#### Description (required)

* Adds `Type/Since` block to follow the [API references format](https://contribute.docs.astro.build/reference/api-references/).
* Fixes some code snippets that contained Typescript issues (because Astro uses Typescript, and we're showing Typescript files (`import type`))

You can check the details per heading below. I'm using an expandable content here because the list to track all the changes is quite long. 😅 
<details>
<summary><strong>Show details</strong></summary>

* External Services:
  * Adds missing type import
  * Adds a new condition in the if statement because `options.width` can be undefined
* Local Services:
  * Moves `AstroConfig["image"]` as type parameter to avoid overriding `imageConfig` type in each hook
  * Adds missing `params` variable in `parseURL` (see https://github.com/withastro/astro/blob/24b04c159ff2e5e1b15efcf8f3a77a131b5a9f92/packages/astro/src/assets/services/service.ts#L374)
  * Fixes `transform` method
    * Adds `async` because this is always asynchronous (and Typescript complains without it)
    * Removes the return type to be consistent with the others (this is not required)
    * Renames one of the `buffer` to prevent conflicts
    * Adds `await` before `mySuperLibraryThatEncodesImages` to justify the asynchronousity and because I think most of the time this would be an async call.
* getConfiguredImageService & imageConfig:
  * No updates to the code snippet because it would require ~20 additional lines to prevent Typescript errors... since we're linking to a working example below, I think it's okay to keep it as is.
* Hooks:
  * Adds `<Since />` for each hook. Most were added in 2.1.0 (see https://github.com/withastro/astro/pull/6344 and the related docs PR https://github.com/withastro/docs/pull/2779) and `validateOptions` was added in `2.1.4` (see https://github.com/withastro/astro/pull/6555)
  * `getURL()`:
    * Fixes the return type (can be a Promise, see https://github.com/withastro/astro/blob/24b04c159ff2e5e1b15efcf8f3a77a131b5a9f92/packages/astro/src/assets/services/service.ts#L46)
    * Fixes the code snippet showing the `ImageTransform` type (no `alt` but `fit` and `position`, see https://github.com/withastro/astro/blob/24b04c159ff2e5e1b15efcf8f3a77a131b5a9f92/packages/astro/src/assets/types.ts#L82-L93)
  * `parseUrl()`:
    * Fixes the return type (can be a Promise, see https://github.com/withastro/astro/blob/24b04c159ff2e5e1b15efcf8f3a77a131b5a9f92/packages/astro/src/assets/services/service.ts#L96-L99)
  * `getSrcSet()`:
    * Fixes the return type (`UnresolvedSrcSetValue` not `SrcSetValue`, see https://github.com/withastro/astro/blob/24b04c159ff2e5e1b15efcf8f3a77a131b5a9f92/packages/astro/src/assets/services/service.ts#L53-L56)
    * Replaces `SrcSetValue` with `UnresolvedSrcSetValue` in the code snippet, they are not exactly the same type (see https://github.com/withastro/astro/blob/24b04c159ff2e5e1b15efcf8f3a77a131b5a9f92/packages/astro/src/assets/types.ts#L57-L65)
  * `transform()`:
    * Fixes the return type (always a Promise and `OutputFormat` should be `ImageOutputFormat`, see https://github.com/withastro/astro/blob/24b04c159ff2e5e1b15efcf8f3a77a131b5a9f92/packages/astro/src/assets/services/service.ts#L104-L108)
  * `validateOptions()`:
    * Fixes the return type (can be a Promise, see https://github.com/withastro/astro/blob/24b04c159ff2e5e1b15efcf8f3a77a131b5a9f92/packages/astro/src/assets/services/service.ts#L75-L78)
* Utilities:
  * Adds a missing export (`emitImageMetadata`) in `astro/assets/utils`
  * Adds a missing `src` property in `generateTransformHash()` example
  * Updates the types format (in other pages, we use `=>` instead of `:` to represent the return type... so let's stay consistent?)

</details>

- Suggested label:  `code snippet update`, `consistency/formatting`

### DIFF
--- a/src/content/docs/ko/guides/deploy/railway.mdx
+++ b/src/content/docs/ko/guides/deploy/railway.mdx
@@ -1,0 +1,19 @@
+---
+title: Astro 사이트를 Railway에 배포하세요
+description: Railway 웹 인터페이스를 사용하여 Astro 사이트를 배포하는 방법
+sidebar:
+  label: Railway
+type: deploy
+i18nReady: true
+stub: true
+---
+
+[Railway](https://railway.com)는 서버부터 관측 가능성에 이르기까지 인프라 스택을 단일 확장 가능한 플랫폼으로 간소화하기 위해 구축된 배포 플랫폼입니다.
+
+## 공식 리소스
+
+[Astro 앱을 Railway에 배포하는 방법](https://docs.railway.com/guides/astro)
+
+## 커뮤니티 리소스
+
+[Astro 사이트를 Railway에 호스팅하는 방법](https://jacksmith.xyz/blog/how-to-host-astro-site-on-railway)
--- a/src/content/docs/zh-cn/guides/images.mdx
+++ b/src/content/docs/zh-cn/guides/images.mdx
@@ -10,9 +10,11 @@ import ReadMore from '~/components/ReadMore.astro';
 
 Astro 为你提供了多种在网站上使用图像的方法，无论它们是本地存储在你的项目内，还是链接到外部 URL，或者在 CMS 或 CDN 中管理的。
 
-Astro 提供 [image](#image-) 和 [picture](#picture-) 组件，[Markdown 图像语法](#markdown-文件中的图像) 处理，[SVG 组件](#svg-组件)，以及 [图像生成函数](#使用-getimage-生成图像) 来优化和/或转换你的图像。此外，你可以默认配置[自动调整响应式图像](#响应式图像行为)，或在单个图像和图片组件上设置响应式属性。
+Astro 提供内置的 [`<Image />`](#image-) 和 [`<Picture />`](#picture-) Astro 组件，[Markdown 图像语法](#markdown-文件中的图像)（`![]()`）处理，[SVG 组件](#svg-组件)，以及 [图像生成函数](#使用-getimage-生成图像) 来优化和/或转换你的图像。此外，你可以默认配置[自动调整响应式图像](#响应式图像行为)，或在单个图像和图片组件上设置响应式属性。
 
 在 `.astro` 或 Markdown 文件中，你可以选择使用原生 HTML 元素或 SVG 文件来引用图片，或者根据文件类型使用标准方式（例如在 MDX 和 JSX 中使用 `<img />`）。然而，Astro 不会对这些图片进行任何处理或优化。
+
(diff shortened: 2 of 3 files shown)


src/content/docs/en/reference/image-service-reference.mdx
```diff
@@ -173,58 +178,73 @@ For both types of services, `options` are the properties passed by the user as a
...
 
-**Required for local services; unavailable for external services**
+<p>
+
+**Type:** `(url: URL, imageConfig: AstroConfig['image']) => { src: string, [key: string]: any } | undefined | Promise<{ src: string, [key: string]: any }> | Promise<undefined>`<br />
+<Since v="2.1.0" />
+</p>
 
-`parseURL(url: URL, imageConfig: AstroConfig['image']): { src: string, [key: string]: any}`
+**Required for local services; unavailable for external services**
```
sarah11918:
```suggestion
**Required for local services only; unavailable for external services**
```
Just to be consistent?

ArmandPhilippot:
No worries, anyone could have reviewed this one I think! I mean there is no new content, I only:
* fixed some TypeScript errors in code blocks, without any functional changes. Copying and pasting into a new Astro project should be enough to check for any errors!
* collected types/since information. I don't think I missed any links to help review this! 😆 What I forgot to mention is that I replaced `LocalImageTransform` with the actual type (as we did in `transform`) because `LocalImageTransform` is not exported/public. That's why this is a bit more verbose than the source code (ie. `{ src: string, [key: string]: any }`).

So, thank you for taking the time for this one! 🙌🏽 

I don't see anyone in particular we should ping to look into this. Maybe JSPark if he has the time, since he's both technical and sufficiently thorough to spot any mistake/oversight?
I'll ping him, unless you can think of someone else. I don't think we need someone from `core` to do that... unless one of them gets bored. 😅

sarah11918:
OK @ArmandPhilippot I leave this completely to you! If you're happy with my having proofread it, and it LGTM, you are welcome to merge it as NWTWWHB! (I'm sure jspark will alert us of any errors when translating, and we can fix them then! 😂 )

So go ahead whenever you feel comfortable! 💜

PR #12237, opened 2025-08-28
new: Document `failOnPrerenderConflict` experimental flag

Fryuni:
#### Description (required)

Document new experimental flag to fail on multiple routes prerendering the same page instead of showing a warning.

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx
@@ -1,0 +1,32 @@
+---
+title: Experimental prerender conflict error
+sidebar:
+  label: Prerender conflict error
+i18nReady: true
+---
+
+import Since from '~/components/Since.astro'
+
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `false`<br />
+<Since v="5.14.0" />
+</p>
+
+Turns prerender conflict warnings into errors during the build process.
+
+Multiple dynamic routes can result in the same output path, for example `/blog/[slug]` and `/blog/[...all]` both could try to prerender the `/blog/post-1` path. In such cases, Astro will warn you about the conflict during build and render only the [highest priority route](/en/guides/routing/#route-priority-order) for the conflicting path. With this flag set, the build will instead fail immediately with an error.
+
+This will be the default behavior in Astro 6.
+
+Enable it in your `astro.config.mjs`:
+
+```js
+// astro.config.mjs
+defineConfig({
+	experimental: {
+		failOnPrerenderConflict: true,
+	},
+});
+```
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -151,6 +151,7 @@ export const sidebar = [
 					'reference/experimental-flags/heading-id-compat',
 					'reference/experimental-flags/static-import-meta-env',
 					'reference/experimental-flags/chrome-devtools-workspace',
+					'reference/experimental-flags/fail-on-prerender-conflict',
 				],
 			}),
 			'reference/legacy-flags',


src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx
```diff
@@ -0,0 +1,32 @@
...
+Enable it in your `astro.config.mjs`:
+
+```js
+// astro.config.mjs
+defineConfig({
+	experimental: {
+		failOnPrerenderConflict: true,
+	},
+});
+```
```
sarah11918:
``````suggestion
 Astro currently warns you during the build about any conflicts between multiple dynamic routes that can result in the same output path. For example `/blog/[slug]` and `/blog/[...all]` both could try to prerender the `/blog/post-1` path. In such cases, Astro renders only the [highest priority route](/en/guides/routing/#route-priority-order) for the conflicting path. This allows your site to build successfully, although you may discover that some pages are rendered by unexpected routes.

With this experimental flag set, the build will instead fail immediately with an error. This will require you to resolve any routing conflicts immediately, and will ensure that Astro is building your routes as you intend.

To enable this behavior, add the `experimental.failOnPrerenderConflict` feature flag to your Astro config:

```js title="astro.config.mjs" ins={4-6}
import { defineConfig } from "astro/config"

defineConfig({
	experimental: {
		failOnPrerenderConflict: true,
	},
});
```

## Usage

After enabling this flag, you may encounter errors about conflicting prerendered routes when you attempt to build your project. If this happens, you will need to update one or more of your [dynamic routes](/en/guides/routing/#dynamic-routes) to avoid ambiguous routing.
``````
This looks great!!

This is one big suggestion because I wanted to do some rearranging (more than actually editing). BUT, I did try to add some context as an example of "level of detail" I think might be useful.

Also, after this, I would consider showing a simple example of resolving a conflict. It's one thing to tell people "both of these routes are trying to render this page", but even better to show a typical conflict and how someone will need to change their project to be able to build!

**Review by sarah11918 — COMMENTED** [review]
Looks great, see my BIG comment for my BIG suggestion below! 😂

File: src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx (hunk 1 above)
Comment (Fryuni) [reply src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx]: Should we add that here or should we add a "troubleshooting" section on the routing page explaining the warning and how to fix it?

Comment (sarah11918) [reply src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx]: @Fryuni For *now*, while this is experimental, (and people flipping this on are the only people who are going to need it, and might *very well* need it!) it should go here!

All experimental documentation is contained to its one page. When this is no longer experimental and is the default, we can redistribute whatever content needed is on this page. (e.g. we might update the routing priority page with a one-liner, as you say, this could go in troubleshooting...)

Comment (Fryuni) [reply src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx]: Just to be clear. The experimental flag is to turn the warning into an error. The warning is for everyone.

Comment (sarah11918) [reply src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx]: Right, everyone gets the warning right now, but the flag means people's builds will fail.

Oh, I get you know -- you mean something in troubleshooting so people can fix the *warning* that they are *currently getting*! Actually, now maybe that does seem helpful. And we can link to it from the existing warning maybe? There's nothing someone would do to fix the existing warning that they wouldn't do to fix a broken build in this case, right? The advice is exactly the same?

I see what you were thinking. I *also* think that people who might have ignored the warning (maybe it didn't really matter to them or affect their project in a way they cared about) will now have failing builds, and these people need to realize this maybe even before deciding to enable the flag. So this still feels like important information about this feature in particular. So I might still include the "how to solve this" in the PR changeset for the minor release maybe, even if we put this in troubleshooting in the docs?

I could go either way on this PR, but now I think troubleshooting is not a bad idea, and might help more people sooner, so not opposed to adding it there! Great idea!

File: src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `false`<br />
+<Since v="5.14.0" />
+</p>
+
+Turns prerender conflict warnings into errors during the build process.
+
+Multiple dynamic routes can result in the same output path, for example `/blog/[slug]` and `/blog/[...all]` both could try to prerender the `/blog/post-1` path. In such cases, Astro will warn you about the conflict during build and render only the [highest priority route](/en/guides/routing/#route-priority-order) for the conflicting path. With this flag set, the build will instead fail immediately with an error.
```
yanthomasdev: small wording/flow nit.
```suggestion
Multiple dynamic routes can result in the same output path; for example, both `/blog/[slug]` and `/blog/[...all]` could try to prerender the `/blog/post-1` path. In such cases, Astro will warn you about the conflict during build and render only the [highest priority route](/en/guides/routing/#route-priority-order) for the conflicting path. With this flag set, the build will instead fail immediately with an error.
```

src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx
Fryuni: > There's nothing someone would do to fix the existing warning that they wouldn't do to fix a broken build in this case, right? The advice is exactly the same?

Correct. We had a discussion about this and decided to not make it the default in the future. It will be experimental to set it as an error and once stabilized it will be a normal flag, but the default will remain as a warning.

sarah11918: OK, updated my previous suggestion @Fryuni to:

- remove the comment about it becoming the default in Astro 6
- Write around Yan's earlier grammar/punctuation fix. Yan was, of course, grammatically correct; however, semi-colons are rarely our style in favour of multiple shorter sentences or moving clauses around to "write ourselves out" of the problem! 😄 

How close are we with this now?

sarah11918: Going to commit here so we have something updated to work off of! Can always reedit!

src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx
```diff
@@ -0,0 +1,32 @@
...
+Enable it in your `astro.config.mjs`:
+
+```js
+// astro.config.mjs
+defineConfig({
+	experimental: {
+		failOnPrerenderConflict: true,
+	},
+});
+```
```
sarah11918:
``````suggestion
Astro currently warns you during the build about any conflicts between multiple dynamic routes that can result in the same output path. For example `/blog/[slug]` and `/blog/[...all]` both could try to prerender the `/blog/post-1` path. In such cases, Astro renders only the [highest priority route](/en/guides/routing/#route-priority-order) for the conflicting path. This allows your site to build successfully, although you may discover that some pages are rendered by unexpected routes.

With this experimental flag set, the build will instead fail immediately with an error. This will require you to resolve any routing conflicts immediately, and will ensure that Astro is building your routes as you intend.

To enable this behavior, add the `experimental.failOnPrerenderConflict` feature flag to your Astro config:

```js title="astro.config.mjs" ins={4-6}
import { defineConfig } from "astro/config"

defineConfig({
	experimental: {
		failOnPrerenderConflict: true,
	},
});
```

## Usage

After enabling this flag, you may encounter errors about conflicting prerendered routes when you attempt to build your project. If this happens, you will need to update one or more of your [dynamic routes](/en/guides/routing/#dynamic-routes) to avoid ambiguous routing.
``````

src/content/docs/en/reference/experimental-flags/fail-on-prerender-conflict.mdx
Fryuni: Sorry for the lack of replies Sarah, going through some things. Feel free to commit any changes needed I'll try to get back to this as soon as I can.

PR #12241, opened 2025-08-29
ci: use depot runners for CI workflows

ematipico:
#### Description (required)

This PR updates the runtime APIs of CSP


ematipico:
PR merged

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -267,9 +267,9 @@ export default defineConfig({
 
 ## Runtime APIs
 
-You can customize the `<meta>` element per page via runtime APIs available from the `Astro` global inside `.astro` components, or the `APIContext` type in endpoints and middleware.
+You can customize the `<meta>` element per page via runtime APIs available from the `Astro.csp` global inside `.astro` components, or the `APIContext.csp` type in endpoints and middleware.
```
```suggestion
You can customize the `<meta>` element per page via runtime APIs available from the `Astro` global inside `.astro` components, or the `APIContext` type in endpoints and middleware.
```
I think it's fine to leave these as is! We often refer to these both as general concepts, and the headings below are named including the `csp`, so I think it's fine to say e.g. "`csp.insertDirective` is a available from the `Astro` global!

sarah11918 (approved):
Thanks for updating these Ema! Just one comment below, but this should be good whenever 5.13.5 is released!

PR #12247, opened 2025-08-30
Fix import in integration sample

ramdacxp:
#### Description (required)

Fixed typo in AlpineJS integration sample.

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
+++ b/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
@@ -102,7 +102,7 @@ The default export of this file should be a function that accepts an Alpine inst
 
 ```js title="astro.config.mjs"
 import { defineConfig } from 'astro/config';
-import alpine from '@astrojs/alpinejs';
+import alpinejs from '@astrojs/alpinejs';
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/alpinejs.mdx
```diff
@@ -106,7 +106,7 @@ import alpine from '@astrojs/alpinejs';
 
 export default defineConfig({
   // ...
-  integrations: [alpine({ entrypoint: '/src/entrypoint' })],
```
yanthomasdev, line 109: In this case, it's right since the default import is being named `alpine`  here, but since that would be different with the example above, let's change that as well!

yanthomasdev (approved):
LGTM! Thanks!

src/content/docs/en/guides/integrations-guide/alpinejs.mdx
ramdacxp: Yes, you're right. Consistency with the example above is nice. So one can copy that single line over to add a custom entrypoint on top of the default config created with `npx astro add alpinejs`.

PR #12250, opened 2025-09-01
feat(sitemap): add configurable XML namespaces support

jdcolombo:
This pull request adds documentation for the new `namespaces` configuration option to the sitemap integration guide. The update explains the purpose, type, and default values for this option, and provides a usage example.

**Sitemap integration documentation update:**

* Added a section describing the `namespaces` option, including its type, default values, and its role in controlling which XML namespaces are included during sitemap generation.
* Provided a code example in `astro.config.mjs` showing how to configure the `namespaces` option for the sitemap integration.

Updated docs for https://github.com/withastro/astro/pull/14285

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -486,7 +486,7 @@ The given configuration will generate sitemap files at `https://example.com/astr
 
 Controls which XML namespaces are included in the sitemap generation.
 
-By default, all namespaces default to `true` and you must explicitly set each one you want to change to `false`.
+By default, all configurable namespaces  (`news`, `xhtml`, `image`, and `video`) are included in your generated sitemap XML. To exclude a namespace from your sitemap generation, add a `namespaces` configuration object and set individual options to `false`:
 
 ```js title="astro.config.mjs" ins={8-14}
 import { defineConfig } from 'astro/config';
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -475,6 +475,31 @@ export default defineConfig({
 
 The given configuration will generate sitemap files at `https://example.com/astronomy-sitemap-0.xml` and `https://example.com/astronomy-sitemap-index.xml`.
 
+### `namespaces`
+
+**Type:** `{ news?: boolean; xhtml?: boolean; image?: boolean; video?: boolean; }`  
+**Default:** `{ news: true, xhtml: true, image: true, video: true }`
+
+Controls which XML namespaces are included in the sitemap generation.
+
+```js title="astro.config.mjs" ins={8}
+import { defineConfig } from 'astro/config';
+import sitemap from '@astrojs/sitemap';
+
+export default defineConfig({
+  site: 'https://example.com',
+  integrations: [
+    sitemap({
+      namespaces: {
+        news: true,
+        xhtml: true, 
+        image: true,
+        video: true,
+      },
+    }),
+  ],
+});
+
 ## Examples
 
 * The official Astro website uses Astro Sitemap to generate [its sitemap](https://astro.build/sitemap-index.xml).
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -486,7 +486,7 @@ The given configuration will generate sitemap files at `https://example.com/astr
 
 An object of XML namespaces to exclude from the generated sitemap.
 
-Excluding unused namespaces can help create cleaner, more focused sitemaps that are faster for search engines to parse and use less bandwidth. For example, if your site doesn't have news content, videos, or multiple languages, you can exclude those namespaces to reduce XML bloat.
+Excluding unused namespaces can help create more focused sitemaps that are faster for search engines to parse and use less bandwidth. For example, if your site doesn't have news content, videos, or multiple languages, you can exclude those namespaces to reduce XML bloat.
 
 By default, all configurable namespaces (`news`, `xhtml`, `image`, and `video`) are included in your generated sitemap XML. To exclude one or more of these namespaces from your sitemap generation, add a `namespaces` configuration object and set individual options to `false`:
 
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -498,11 +498,9 @@ export default defineConfig({
   site: 'https://example.com',
   integrations: [
     sitemap({
-      // Only include image and video namespaces
       namespaces: {
         news: false,
         xhtml: false, 
-        // image and video will be true by default
       }
     })
   ]
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -482,7 +482,9 @@ The given configuration will generate sitemap files at `https://example.com/astr
 
 Controls which XML namespaces are included in the sitemap generation.
 
-```js title="astro.config.mjs" ins={8}
+By default, all namespaces default to `true` and you must explicitly set each one you want to include to `false`.
+
+```js title="astro.config.mjs" ins={8-14}
 import { defineConfig } from 'astro/config';
 import sitemap from '@astrojs/sitemap';
 
@@ -490,11 +492,11 @@ export default defineConfig({
   site: 'https://example.com',
   integrations: [
     sitemap({
+      // Only include image and video namespaces
       namespaces: {
-        news: true,
-        xhtml: true, 
-        image: true,
-        video: true,
+        news: false,
+        xhtml: false, 
+        // image and video will be true by default
       },
     }),
   ],
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -501,6 +501,7 @@ export default defineConfig({
     })
   ]
 });
+```
 
 ## Examples
 
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -497,9 +497,9 @@ export default defineConfig({
         news: false,
         xhtml: false, 
         // image and video will be true by default
-      },
-    }),
-  ],
+      }
+    })
+  ]
 });
 
 ## Examples
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -482,7 +482,7 @@ The given configuration will generate sitemap files at `https://example.com/astr
 
 Controls which XML namespaces are included in the sitemap generation.
 
-By default, all namespaces default to `true` and you must explicitly set each one you want to include to `false`.
+By default, all namespaces default to `true` and you must explicitly set each one you want to change to `false`.
 
 ```js title="astro.config.mjs" ins={8-14}
 import { defineConfig } from 'astro/config';
--- a/src/content/docs/en/guides/integrations-guide/sitemap.mdx
+++ b/src/content/docs/en/guides/integrations-guide/sitemap.mdx
@@ -477,8 +477,12 @@ The given configuration will generate sitemap files at `https://example.com/astr
 
 ### `namespaces`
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,31 @@ export default defineConfig({
...
+export default defineConfig({
+  site: 'https://example.com',
+  integrations: [
+    sitemap({
+      namespaces: {
+        news: true,
+        xhtml: true, 
+        image: true,
+        video: true,
+      },
```
ascorbic: It would be good to demonstrate the behaviour when you use the object. i.e. how by default everything is true, but as soon as you pass an object, they all default to false so you have to specify all of the ones that you want to include

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,33 @@ export default defineConfig({
...
+    sitemap({
+      // Only include image and video namespaces
+      namespaces: {
+        news: false,
+        xhtml: false, 
+        // image and video will be true by default
+      },
+    }),
+  ],
+});
```
yanthomasdev: removed trailing commas
```suggestion
export default defineConfig({
  site: 'https://example.com',
  integrations: [
    sitemap({
      // Only include image and video namespaces
      namespaces: {
        news: false,
        xhtml: false, 
        // image and video will be true by default
      }
    })
  ]
});
```

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,33 @@ export default defineConfig({
...
+      // Only include image and video namespaces
+      namespaces: {
+        news: false,
+        xhtml: false, 
+        // image and video will be true by default
+      }
+    })
+  ]
+});
+
```
sarah11918, line 509:
``````suggestion
```

``````

File: src/content/docs/en/guides/integrations-guide/sitemap.mdx [hunk 4]
```diff
@@ -475,6 +475,34 @@ export default defineConfig({
 
 The given configuration will generate sitemap files at `https://example.com/astronomy-sitemap-0.xml` and `https://example.com/astronomy-sitemap-index.xml`.
 
+### `namespaces`
+
+**Type:** `{ news?: boolean; xhtml?: boolean; image?: boolean; video?: boolean; }`  
+**Default:** `{ news: true, xhtml: true, image: true, video: true }`
+
+Controls which XML namespaces are included in the sitemap generation.
```
sarah11918: Can we have a sentence here that explains *why* someone might want to do this? (See the other entries on this page like `customPages`) 

I can see from your implementation PR that you have said "This allows users to exclude unused namespaces like `xmlns:news` for cleaner sitemap files." Is that the main practical reason to have this option? These are namespaces that Astro would otherwise include, but now we have a configuration to NOT include them? And, is "looking cleaner" the advantage, or is there an associated performance advantage to having less included in the XML? Are there negative consequences they are trying to avoid? Is this personal preference? How do they decide?

Also, just more for my own understanding, is there a particular reason you've chosen these 4 options? Are there more that might eventually be added? In some cases, we'd even make separate `namespace.video` entries with a smaller sub section for each one. I think we can get away without that here, but we *may* consider doing that at some point, especially if there are enough of them that just looking at the type to figure out which ones are available seems a little difficult to read. 

I think you have nicely describe *what* this does, but our docs often try to have a tiny bit of the *why* so that people know *when* they might choose to use this config option!

Also, there is no standard set of XML namespaces, I believe? But, there's probably a list of ones that are generated by Astro by default (otherwise, why would we need a PR to allow you to exclude some of them?) I'm assuming that there are some/more that are NOT configurable to be removed? I don't know that we need to list them all, but it might be more appropriate to describe this feature as namespaces that are *excluded*, not included! (Since there are more included by default that you do not set and cannot explicitly include?) Something maybe like this for a first sentence (but make this accurate):

```suggest
An object of XML namespaces to exclude from the generated sitemap.
```

What would you suggest here so that this entry sounds more like the other config options?

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,34 @@ export default defineConfig({
 
 The given configuration will generate sitemap files at `https://example.com/astronomy-sitemap-0.xml` and `https://example.com/astronomy-sitemap-index.xml`.
 
+### `namespaces`
+
+**Type:** `{ news?: boolean; xhtml?: boolean; image?: boolean; video?: boolean; }`  
+**Default:** `{ news: true, xhtml: true, image: true, video: true }`
```
sarah11918:
```suggestion
<p>

**Type:** `{ news?: boolean; xhtml?: boolean; image?: boolean; video?: boolean; }` <br />
**Default:** `{ news: true, xhtml: true, image: true, video: true }` <br />
<Since v="3.6.0" pkg="@astrojs/sitemap" />
</p>
```
Thank you for adding the type information! Just formatting it here, and adding the version of the package number it was added in, in case people aren't up to date when using this!

(Note to @ArmandPhilippot , I guess we have our next page to check :sweat_smile: )

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,34 @@ export default defineConfig({
...
+
+Controls which XML namespaces are included in the sitemap generation.
+
+By default, all namespaces default to `true` and you must explicitly set each one you want to change to `false`.
```
sarah11918:
```suggestion
By default, all configurable namespaces  (`news`, `xhtml`, `image`, and `video`) are included in your generated sitemap XML. To exclude a namespace from your sitemap generation, add a `namespaces` configuration object and set individual options to `false`:
```
 Maybe something like this since we haven't mentioned (other than the type) what these options are in paragraph text?

We can also avoid mentioning both true and false as opposites if e.g. we say the *result* you are getting by default, and setting `false` to achieve the opposite result. This is helpful because no one is ever going to set `true` here, so we like to keep `false` as the only thing people see when skimming/scanning the page! (And, your code comment can mention `true` by default)

src/content/docs/en/guides/integrations-guide/sitemap.mdx
jdcolombo: Why someone would want to exclude namespaces:
The main practical reasons are:
- Excluding unused namespaces reduces XML bloat
- Smaller sitemap files are faster to parse by search engines
- Some SEO analysis tools prefer minimal sitemaps without unused declarations
- Particularly relevant for sites with very large sitemaps

Regarding the 4 namespace options:
These are the standard XML namespaces that Astro includes by default in sitemaps. 

- xmlns:news - Google News sitemaps
- xmlns:xhtml - Multilingual/alternate language links
- xmlns:image - Image sitemaps
- xmlns:video - Video sitemaps

You're right that we could potentially add more namespaces in the future, and separate subsections might be worth considering as the list grows.

Suggested documentation revision:
```

### `namespaces`

<p>

**Type:** `{ news?: boolean; xhtml?: boolean; image?: boolean; video?: boolean; }` <br />
**Default:** `{ news: true, xhtml: true, image: true, video: true }` <br />
<Since v="3.6.0" pkg="@astrojs/sitemap" />
</p>

An object of XML namespaces to exclude from the generated sitemap.

Excluding unused namespaces can help create cleaner, more focused sitemaps that are faster for search engines to parse and use less bandwidth. For example, if your site doesn't have news content, videos, or multiple languages, you can exclude those namespaces to reduce XML bloat.

By default, all configurable namespaces  (`news`, `xhtml`, `image`, and `video`) are included in your generated sitemap XML. To exclude a namespace from your sitemap generation, add a `namespaces` configuration object and set individual options to `false`:

```js title="astro.config.mjs" ins={8-14}
import sitemap from '@astrojs/sitemap';

export default defineConfig({
  site: 'https://example.com',
  integrations: [
    sitemap({
      // Only include image and video namespaces
      namespaces: {
        news: false,
        xhtml: false, 
        // image and video will be true by default
      }
    })
  ]
});
```

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,40 @@ export default defineConfig({
...
+**Type:** `{ news?: boolean; xhtml?: boolean; image?: boolean; video?: boolean; }` <br />
+**Default:** `{ news: true, xhtml: true, image: true, video: true }` <br />
+<Since v="3.6.0" pkg="@astrojs/sitemap" />
+</p>
+
+An object of XML namespaces to exclude from the generated sitemap.
+
+Excluding unused namespaces can help create cleaner, more focused sitemaps that are faster for search engines to parse and use less bandwidth. For example, if your site doesn't have news content, videos, or multiple languages, you can exclude those namespaces to reduce XML bloat.
+
+By default, all configurable namespaces  (`news`, `xhtml`, `image`, and `video`) are included in your generated sitemap XML. To exclude a namespace from your sitemap generation, add a `namespaces` configuration object and set individual options to `false`:
```
sarah11918:
```suggestion
By default, all configurable namespaces (`news`, `xhtml`, `image`, and `video`) are included in your generated sitemap XML. To exclude one or more of these namespaces from your sitemap generation, add a `namespaces` configuration object and set individual options to `false`:
```

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,40 @@ export default defineConfig({
...
+
+export default defineConfig({
+  site: 'https://example.com',
+  integrations: [
+    sitemap({
+      // Only include image and video namespaces
+      namespaces: {
+        news: false,
+        xhtml: false, 
+        // image and video will be true by default
```
sarah11918:
```suggestion
      namespaces: {
        news: false,
        xhtml: false, 
```
I think this code snippet works even without the comments! I had originally thought about keeping one or the other, but I'm not sure it's needed.

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,40 @@ export default defineConfig({
...
+<p>
+
+**Type:** `{ news?: boolean; xhtml?: boolean; image?: boolean; video?: boolean; }` <br />
+**Default:** `{ news: true, xhtml: true, image: true, video: true }` <br />
+<Since v="3.6.0" pkg="@astrojs/sitemap" />
+</p>
+
+An object of XML namespaces to exclude from the generated sitemap.
+
+Excluding unused namespaces can help create cleaner, more focused sitemaps that are faster for search engines to parse and use less bandwidth. For example, if your site doesn't have news content, videos, or multiple languages, you can exclude those namespaces to reduce XML bloat.
```
sarah11918:
```suggestion
Excluding unused namespaces can help create more focused sitemaps that are faster for search engines to parse and use less bandwidth. For example, if your site doesn't have news content, videos, or multiple languages, you can exclude those namespaces to reduce XML bloat.
```
I love this paragraph, thank you so much for adding it! I think we can stick to just "more focused" to keep this... more focused! 😂

sarah11918 (commented):
Really happy with where this landed! Just some tiny polishing and this will be good to go! Now that we know what our docs will say, it's easier to present the feature in the changeset of the implementation PR, so I'll move there next! 🚀

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: I didn't get a notification for that! 😮 I guess there are still a few pages like these, but yeah we should check the other props! I'll add this on my todo list. 😄

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -475,6 +475,38 @@ export default defineConfig({
...
+<Since v="3.6.0" pkg="@astrojs/sitemap" />
+</p>
+
+An object of XML namespaces to exclude from the generated sitemap.
+
+Excluding unused namespaces can help create more focused sitemaps that are faster for search engines to parse and use less bandwidth. For example, if your site doesn't have news content, videos, or multiple languages, you can exclude those namespaces to reduce XML bloat.
+
+By default, all configurable namespaces (`news`, `xhtml`, `image`, and `video`) are included in your generated sitemap XML. To exclude one or more of these namespaces from your sitemap generation, add a `namespaces` configuration object and set individual options to `false`:
+
+```js title="astro.config.mjs" ins={8-14}
```
ArmandPhilippot: nit: The highlighting is starting on the `namespaces` object, so it should end at the end of the object instead of the end of the config:
````suggestion
```js title="astro.config.mjs" ins={8-11}
````

yanthomasdev (approved):

PR #12264, opened 2025-09-04
feat(astro:i18n): add docs for some missing public utilities

ArmandPhilippot:
#### Description (required)

The [`i18nNoLocaleFoundInPath` error](https://docs.astro.build/en/reference/errors/i18n-no-locale-found-in-path/) suggests to use the `pathHasLocale` helper... but I couldn't find it in the docs. Actually we're missing a few utilities in the `astro:i18n` reference.

Adds documentation for:
* `normalizeTheLocale()` (since `astro:i18n` release? Maybe this was made public later but I couldn't find a PR suggesting this)
* `pathHasLocale()` (added in 4.6.0, https://github.com/withastro/astro/pull/10193)
* `toCodes()` (added in 4.0.0, https://github.com/withastro/astro/pull/9200)
* `toPaths()` (added in 4.0.0, https://github.com/withastro/astro/pull/9200)

I'm not quite sure about the use cases, so I added examples based on the [`getPathByLocale()`](https://docs.astro.build/en/reference/modules/astro-i18n/#getpathbylocale) one.

### DIFF
--- a/src/content/docs/en/reference/modules/astro-i18n.mdx
+++ b/src/content/docs/en/reference/modules/astro-i18n.mdx
@@ -40,6 +40,10 @@ import {
   notFound,
   middleware,
   requestHasLocale,
+  normalizeTheLocale,
+  pathHasLocale,
+  toCodes,
+  toPaths
  } from 'astro:i18n';
 ```
 
@@ -336,4 +340,112 @@ export const onRequest = defineMiddleware(async (context, next) => {
   }
   return new Response("Not found", { status: 404 });
 })
+```
+
+### `normalizeTheLocale()`
+
+<p>
+
+**Type:** `(locale: string) => string`
+</p>
+
+Replaces underscores (`_`) with hyphens (`-`) in the given locale before returning a lowercase version.
+
+```astro title="src/pages/index.astro"
+---
+normalizeTheLocale("it_VT") // returns `it-vt`
+---
+```
+
+### `pathHasLocale()`
+
+<p>
+
+**Type:** `(path: string) => boolean`<br />
+<Since v="4.6.0" />
+</p>
+
+Checks whether the given path contains a configured locale.
+
+This is useful to prevent errors before using an i18n utility that relies on a locale from a URL path.
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  i18n: {
+    locales: [
+      { codes: ["it-VT", "it"], path: "italiano" },
+      "es"
+    ]
+  }
+})
+```
+
+```astro title="src/pages/index.astro"
+---
+import { pathHasLocale } from "astro:i18n";
+
+pathHasLocale("italiano"); // returns `true`
+pathHasLocale("es"); // returns `true`
+pathHasLocale("it-VT"); // returns `false`
+---
+```
+
+### `toCodes()`
+
+<p>
+
+**Type:** `(locales: Locales) => string[]`<br />
+<Since v="4.0.0" />
+</p>
+
+Retrieves the configured locale codes for each locale defined in your configuration. When multiple codes are associated to a locale, only the first one will be added to the array.
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  i18n: {
+    locales: [
+      { codes: ["it-VT", "it"], path: "italiano" },
+      "es"
+    ]
+  }
+})
+```
+
+```astro title="src/pages/index.astro"
+---
+import { i18n } from "astro:config/client";
+import { toCodes } from "astro:i18n";
+
+toCodes(i18n!.locales); // ["it-VT", "es"]
+---
+```
+
+### `toPaths()`
+
+<p>
+
+**Type:** `(locales: Locales) => string[]`<br />
+<Since v="4.0.0" />
+</p>
+
+Retrieves the configured locale paths for each locale defined in your configuration.
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  i18n: {
+    locales: [
+      { codes: ["it-VT", "it"], path: "italiano" },
+      "es"
+    ]
+  }
+})
+```
+
+```astro title="src/pages/index.astro"
+---
+import { i18n } from "astro:config/client";
+import { toPaths } from "astro:i18n";
+
+toPaths(i18n!.locales); // ["italiano", "es"]
+---
 ```


sarah11918:
Thank you so much, @ArmandPhilippot ! I'm going to ask @ematipico to review this, make sure everything here should be in docs etc. before I look too closely! 🙌

sarah11918:
Noting that Ema is on a few weeks of holidays right now, so we'll get around to this one, eventually! 😄

ArmandPhilippot:
* I was checking for `normalizeThePath` and it is indeed exported from the [i18n module](https://github.com/withastro/astro/blob/7bf7195f8cb00f52022bd411f9c18ffcc6b345e3/packages/astro/src/i18n/index.ts#L237) but not from the [i18n virtual module](https://github.com/withastro/astro/blob/7bf7195f8cb00f52022bd411f9c18ffcc6b345e3/packages/astro/src/virtual-modules/i18n.ts#L12). So, you can't import it from `"astro:i18n"`. So, should we export it like the others?
* I'm not sure I get "moved away from internal code", do you mean this should be declared in `packages/astro/src/i18n/utils.ts` instead of `packages/astro/src/i18n/index.ts`?
* Yeah, that makes sense if they are public!

HiDeoo:
> - I was checking for `normalizeThePath` and it is indeed exported from the [i18n module](https://github.com/withastro/astro/blob/7bf7195f8cb00f52022bd411f9c18ffcc6b345e3/packages/astro/src/i18n/index.ts#L237) but not from the [i18n virtual module](https://github.com/withastro/astro/blob/7bf7195f8cb00f52022bd411f9c18ffcc6b345e3/packages/astro/src/virtual-modules/i18n.ts#L12). So, you can't import it from `"astro:i18n"`. So, should we export it like the others?

Good catch, we definitely only want to document what is public, I probably got confused during the T&D call.

> - I'm not sure I get "moved away from internal code", do you mean this should be declared in `packages/astro/src/i18n/utils.ts` instead of `packages/astro/src/i18n/index.ts`?

May need to double check if Ema had a specific idea in mind, but I guess the idea was to move public code away from `i18n/index.ts` so its content is no longer a mix of public and internal code.

src/content/docs/en/reference/modules/astro-i18n.mdx
```diff
@@ -337,3 +341,111 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+<p>
+
+**Type:** `(locale: string) => string`
+</p>
+
+Replaces underscores (`_`) with hyphens (`-`) in the given locale before returning a lowercase version.
+
+```astro title="src/pages/index.astro"
+---
+normalizeTheLocale("it_VT") // returns `it-vt`
```
ArmandPhilippot, line 358: Adds missing import in the code snippet:
```suggestion
import { normalizeTheLocale } from "astro:i18n";

normalizeTheLocale("it_VT") // returns `it-vt`
```

src/content/docs/en/reference/modules/astro-i18n.mdx
```diff
@@ -337,3 +341,113 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+**Type:** `(locale: string) => string`
+</p>
+
+Replaces underscores (`_`) with hyphens (`-`) in the given locale before returning a lowercase version.
+
+```astro title="src/pages/index.astro"
+---
+import { normalizeTheLocale } from "astro:i18n";
+
+normalizeTheLocale("it_VT") // returns `it-vt`
```
HiDeoo, line 358: I wonder if adding an example with `Astro.currentLocale`, as I assume it's most likely to be used in combination, could be relevant, e.g. `Astro.currentLocale` is `pt-PT` and you want the associated `pt-pt`?

src/content/docs/en/reference/modules/astro-i18n.mdx
```diff
@@ -337,3 +341,113 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+})
+```
+
+```astro title="src/pages/index.astro"
+---
+import { pathHasLocale } from "astro:i18n";
+
+pathHasLocale("italiano"); // returns `true`
+pathHasLocale("es"); // returns `true`
+pathHasLocale("it-VT"); // returns `false`
```
HiDeoo, line 394: While the description makes it obvious the `path` argument of `pathHasLocale()` is not a strict match to a locale `path` configuration (it checks if a segment of a path contains the locale path), I wonder if the example make that obvious.

Maybe something like?

```js
pathHasLocale('/es/blog/'); // returns `true`
```

src/content/docs/en/reference/modules/astro-i18n.mdx
ArmandPhilippot: Yeah, I wasn't totally sure about the use cases so I took inspiration from the other utilities for the example. But, yes, I think what you're suggesting makes sense!

To avoid another code snippet just for the config, maybe the following is enough:
```suggestion
normalizeTheLocale("it_VT") // returns `it-vt`
// Assuming the current locale is `"pt-PR"`:
normalizeTheLocale(Astro.currentLocale) // returns `pt-pt`
```

src/content/docs/en/reference/modules/astro-i18n.mdx
ArmandPhilippot: This one already had a JSDoc, so I admit that I haven't given much thought to the examples. 😀 
But, yes, again great suggestion!

```suggestion
pathHasLocale("italiano"); // returns `true`
pathHasLocale("es"); // returns `true`
pathHasLocale('/es/blog/'); // returns `true`
pathHasLocale("it-VT"); // returns `false`
```

src/content/docs/en/reference/modules/astro-i18n.mdx
HiDeoo: I would think so, altho I'll need to check when back at a computer because `Astro.currentLocale` is `string|undefined` and I don't know if this function accept a `Locale` (`string | undefined`) or just `string`. If it's just `string`, could be a bit more complex.

src/content/docs/en/reference/modules/astro-i18n.mdx
HiDeoo: Looks perfect to me, a good mix of everything possible considering the function does not care at all about position or anything, it just checks if a segment is a locale path 👍

src/content/docs/en/reference/modules/astro-i18n.mdx
ArmandPhilippot: Only `string` (`(locale: string) => string`), so yes it's not perfect.
That's said I doubt someone will copy paste the code block as is, this is not really useful. And, we write that the value is assumed, so maybe  this is enough to implicitly say this is written after `Astro.currentLocale !== undefined`?

HiDeoo: Ah yeah, thanks for checking. Indeed, I think that should be fine for such a "list of calls" type of example.

src/content/docs/en/reference/modules/astro-i18n.mdx
```diff
@@ -337,3 +341,116 @@ export const onRequest = defineMiddleware(async (context, next) => {
   return new Response("Not found", { status: 404 });
 })
 ```
+
+### `normalizeTheLocale()`
+
+<p>
+
+**Type:** `(locale: string) => string`
```
sarah11918, line 349: Do we need a `<Since />` here? Maybe it makes sense to use the default `4.0.0` rather than nothing?

sarah11918:
When y'all are happy with this, no need to wait for me!

src/content/docs/en/reference/modules/astro-i18n.mdx
ArmandPhilippot: I couldn't find a PR associated to this export, so I checked through VSCode/Gitlens the line history, and I can confirm this is here since the module creation (https://github.com/withastro/astro/issues/8974). So, it's more like `astro@3.5.0` (ie. like `astro:i18n`).

I don't have a strong opinion on whether each API should have a `<Since />` or not. I think it's clear that if there is no since, then it's here from the beginning (of the module or Astro according to the context).
But if we add `<Since />` for this one, we should update the already documented APIs to add it to them as well, for consistency.

ArmandPhilippot:
Noted, thanks! I was waiting for the [core's PR](https://github.com/withastro/astro/pull/14703) because this is new content but if you don't see anything wrong here, then I guess it's safe to merge it with Ema and Yan approval! 😄

yanthomasdev (approved):
LGTM!

ArmandPhilippot:
Since no one else has intervened since Yan's approval, I guess everyone is happy and this one is ready to be merged! If not, you still have time while the branch is updating. 😄

PR #12266, opened 2025-09-04
Update dependencies

delucis:
#### Description (required)

Updates ALL the deps!
Well, almost all — `canvas-confetti` has grown some features and has a larger bundle size these days, but we don’t need those, so I’ve left that at the old release. But everything else is updated to latest.

I did these step by step to let CI run and verify things were working. Where relevant, I ran stuff locally (e.g. linting/formatting) to check things were working as expected.

##### Major

- `pnpm` itself to the latest major release (9.15.1 => 10.15.1)
- `remark-smartypants` (2.1.0 => 3.0.2) (this introduces a type error with Astro for some reason, but works just fine)
- `globals` (15.14.0 => 16.3.0)
- `@astrojs/starlight` (0.34.4 => 0.35.2) 
- `@types/node` (20.17.10 => 22.18.1) — we’re using Node 22 in CI so this updates types to match that environment. No breaking changes for us.
- `sharp` (0.33.5 => 0.34.3)
- `starlight-llms-txt` (0.5.1 => 0.6.0) — added a feature, no breaking changes
- `p-retry` (6.2.1 => 7.0.0) — used in our cached fetch, drops old Node support
- `canvaskit-wasm` (0.39.1 => 0.40.0) — used for OG images, added some features, no breaking changes
- `htmlparser2` (9.1.0 => 10.0.0) — used  in link checking, no breaking changes for our usage

##### Minor

- `astro` (5.5.3 => 5.13.5)
- `@astrojs/sitemap` (3.3.0 => 3.5.1)
- `@11ty/eleventy-fetch` (5.0.1 => 5.1.0)
- `@eslint/js` (9.17.0 => 9.34.0)
- `@typescript-eslint/parser` (8.19.1 => 8.42.0)
- `eslint` (9.17.0 => 9.34.0)
- `typescript-eslint` (8.19.1 => 8.42.0)
- `prettier` (3.4.2 => 3.6.2) — I ran formatting locally and there’ll be two files with small formatting changes when CI runs on merge

##### Patch

- `@expressive-code/plugin-collapsible-sections` (0.41.2 => 0.41.3) 
- `hastscript` (6.0.3 => 6.0.4)
- `hast-util-select` (9.0.0 => 9.0.1)
- `jsdoc-api` (9.3.4 => 9.3.5)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -400,9 +400,6 @@ packages:
   '@emmetio/stream-reader@2.2.0':
     resolution: {integrity: sha512-fXVXEyFA5Yv3M3n8sUGT7+fvecGrZP4k6FnWWMSZVQf69kAq0LLpaBQLGcPR30m3zMmKYhECP4k/ZkzvhEW5kw==}
 
-  '@emnapi/runtime@1.3.1':
-    resolution: {integrity: sha512-kEBmG8KyqtxJZv+ygbEim+KCGtIq1fC22Ms3S4ziXmYKm8uyoLX0MHONVKwp+9opg390VaKRNt4a7A9NwmpNhw==}
-
   '@emnapi/runtime@1.5.0':
     resolution: {integrity: sha512-97/BJ3iXHww3djw6hYIfErCZFee7qCtrneuLa20UXFCOTCfBM2cvQHjWJ2EG0s0MtdNwInarqCTz35i4wWXHsQ==}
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Inspected the tutorial pages (you know I did... 🎊 ) and all looks good in the deploy preview!

sarah11918 (approved):
Forgot to approve based on testing tutorial and some other pages!

ArmandPhilippot:
I was really too slow 😄 but I didn't noticed anything broken!

Regarding:

> this introduces a type error with Astro for some reason, but works just fine

It could be the way they define the Remark plugin (looking quickly at their code). I think by default if the second `Plugin` type parameter is not explicit, `Node` would be assumed while Astro uses `Root` by default. Both types are not fully compatible.
So, this is not really an issue with `remark-smartypants`, neither Astro I think. I mean, using `Node` as default for `Plugin` seems odd to me (usually `Root` makes more sense IMO since the plugin runs on the whole tree). But there must be a reason for this.

So, yes, not a real issue! 🎉

delucis:
> I was really too slow

Or maybe we were too fast! 🙈 I wanted the Astro update for #12265 so merged with one approval, but I suspected it would be better to leave it for more eyes.

Glad you didn’t find anything broken though! Thanks for digging into that `remark-smartypants` issue.

PR #12268, opened 2025-09-06
Add Railway deploy guide stub to docs

jacksmithxyz:
Hello, this PR is to help start off the process of adding Railway into the deployment guides in the Astro docs. 

#### Description

Following the documentation at [Adding an Astro + X guide](https://contribute.docs.astro.build/guides/new-third-party-guide/), this PR:

- Adds the SVG logo for Railway into `public/logos`
- Adds the Railway SVG entry to the appropriate data file
- Creates a stub page for the Railway deployment guide

This is all for the `en` docs.

Thanks for reviewing!

Jack


src/content/docs/en/guides/deploy/railway.mdx
```diff
@@ -0,0 +1,20 @@
...
+description: How to deploy your Astro site using the Railway web interface.
+sidebar:
+  label: Railway
+type: deploy
+i18nReady: true
+stub: true
+---
+
+[Railway](https://railway.com) is a deployment platform built to help simplify
+the hosting and deployment of software products.
```
sarah11918:
```suggestion
[Railway](https://railway.com) is a deployment platform built to simplify your infrastructure stack from servers to observability with a single, scalable platform.
```
I took this from their home page. I appreciate that you didn't use ANY marketing speak! (And, I removed their "easy-to-use" line). I think this level of detail, mentioning the key/distinguishing features is helpful so that people have some kind of idea whether or not they want to explore further.

src/content/docs/en/guides/deploy/railway.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+## Official Resources 
+
+[Deploy an Astro App](https://docs.railway.com/guides/astro)
```
sarah11918:
```suggestion
[Railway guide to deploying an Astro app](https://docs.railway.com/guides/astro)
```
We don't have a totally standard way of saying this yet, but ideally this includes an indication that it comes from Railway or is located in Railway docs. So maybe something like this?

And, it doesn't have to be the exact page title of where you'll land! For the purposes of this guide, it's more important that it's representative of the content someone will find. So, we can adjust as necessary to best tell readers what they will find!

src/content/docs/en/guides/deploy/railway.mdx
jackwsmth: Thanks Sarah. I was unsure about taking anything directly, so it's great to have some input here.

src/content/docs/en/guides/deploy/railway.mdx
jackwsmth: Got you! Thanks for the feedback

jackwsmth:
Hey @sarah11918, I've set the `supports` value in src/components/DeployGuidesNav.astro to both `ssr` and `static` - I think this is correct.

The official guide mentioned in the stub sets up the deployment with the Node adapter for SSR, whilst the community guide which I have created uses Railpack which says:

```
 Detected Node
↳ Using npm package manager
↳ Deploying as astro static site
↳ Output directory: dist
```

Before creating a container for it. So I believe it supports both - but if I need to change anything here please let me know 🙂

src/content/docs/en/guides/deploy/railway.mdx
sarah11918: Yeah, I think places are generally OK if you take their specific taglines etc? It's good for their branding to have words reinforced in multiple places. And, we know that's how they like to refer to themselves, so it's pretty safe vs us saying something about them that they don't quite think is right. We can always change if they complain! But typically, when describing another project, it's common to use their own lingo to make sure you're presenting them in the light they would themselves.

The times I don't are when they ONLY have marketing statements, and never really define themselves well, so I have to piece together something and make my best guess about WHAT ARE YOU?!?! 😂

sarah11918:
> I believe it supports both

It would be nice to confirm this! I checked out your `modern-astro-starter` and I don't see an adapter included, so it can't be doing any on-demand rendering. But, it also doesn't include a link to see the template deployed anywhere so... does the template work? 😅 Do you have a functioning, entirely prerendered site hosted on Railway? If so, I'm comfortable showing both rendering options!

(It would also be a nice courtesy to include a demo link to the template deployed showing a working website, as people often want to check out the final version when choosing to use a template! That's what I would have checked just now to confirm that your template is functional!)

jackwsmth:
> Do you have a functioning, entirely prerendered site hosted on Railway? If so, I'm comfortable showing both rendering options!

My own website (https://jacksmith.xyz) is actually hosted this way, but I think it makes complete sense to first create a demo link for the template and make things feel a bit more polished on that front. I'll tag you once I've made some improvements!

sarah11918:
Sure thing! And to be clear, by "a demo link for the template" I mean something that is linked to in its GitHub repo's description or README. That wouldn't be added directly here in this PR, which should just link to that template on GitHub.

So, it doesn't *have* to hold up this PR if you don't want it to, but in general, templates are much more useful and helpful if they include a demo link. (For templates submitted to our theme catalogue for example, it's required.)

So I'll wait to hear from you before we do anything further with this PR! 🙌

jackwsmth:
Hey @sarah11918, I've updated my demo template repo slightly to ensure that it will deploy as a static site, no adapter, with a single click. I've added a `railway.json` file to the template repo to reflect this. 

I've also added the [demo link](https://modern-astro-starter-production.up.railway.app) in the GitHub repo for the template, although the site itself could do with a small amount of styling (although it is a template for people to kick off their own website with). 

I think I'm happy for the stub page to go through if you're in agreement 🙂

I can open a new PR  as I work through to add a full guide, and maybe we can go into more details about style and improvements here as well if that makes sense to you. 

Thanks!

sarah11918:
Thanks @jacksmithxyz , I think this all looks in order now! And yes, at your leisure you can consider working on a more full guide.

If so, please keep this fairly straightforward and make use of Railway's own documentation when possible. For example, instead of describing details or e.g. "where to go in a dashboard" to do something, we would tend to use a semantic link like `[enable X in your Railway project](link to railway's docs)` We shouldn't be documenting how to do/use Railway because they could change things, make this guide out of date, and we wouldn't really know.

Another way to keep these guides tight is to state Prerequisites and make assumptions. For example, a prerequisite could be "A Railway account configured for..." and then we don't even have to address it in our step-by-step! The guide here should focus as much as possible on just Astro-specific things (the things that maybe they won't find help with in Railway's own guides) like any configuration of an Astro project needed, special actions/files needed for an Astro project etc.

The two tricks of pre-requisites and simple steps that might even just be links to a Railway step can help keep this streamlined and easy to maintain! Happy to see what you come up with!

PR #12270, opened 2025-09-08
i18n(fr): improve the wording in errors references

ArmandPhilippot:
#### Description (required)

Updates the French translations of `error-reference.mdx` and most of `reference/errors` to improve the wording:
* make descriptions on `error-reference.mdx` match the title of individual errors pages
* translate all the titles on individual errors pages (actually I'm a bit torn, maybe none of the titles should be translated because users don't get a translated error message? well they can use the error code instead...)
* do not translate error quotes on individual error pages (users don't get a translated error message so it might be easier to search for them if the quote is not translated)
* fix a Markdown heading in `error-reference.mdx`
* fix a typo (an additional space actually) preventing the frontmatter to be correctly parsed in an individual error
* fix some mismatches with the English version
* revert translated inline codes
* use a consistent translation for `thrown` (I think `émise` or even `générée` sounds better than `lancée` or `apparaît`)
* translate `locale`
* and based on the French i18n guide updates:
  * replace `itinéraire` with `route`
  * replace `île` with `îlot`
  * do not translate `glob`
  * translate `pattern`
  * replace `supporté` with `pris en charge`
  * replace `rendre` with `effectuer le rendu`/`restituer`/`afficher` where more appropriate
  * replace `endpoint` and `point d'accès` with `point de terminaison`
  * replace `plugin` with `module d'extension`
  * replace `le CLI` with `la CLI`
  * replace `obsolète` with `dépréciée` for `deprecated`
  * replace `import` with `importation`
  * small rewording to make the reading smoother/avoid literal translation (e.g. `dans l'adaptateur`/`avec l'adaptateur`, I think the second sounds more natural)
  * some grammar nits (e.g. `locals` is a property (feminin noun), so verb agreement, and the same for the implicit "error" noun: `émis` should be `émise`)

- Suggested label: i18n


yanthomasdev (approved):
This one took a while to review; amazing work, LGTM!

ArmandPhilippot:
Yeah sorry, it was difficult to split it while editing `error-reference.mdx` at the same time. But maybe I should have done that after... and ship smaller PRs. But good news, after the last two opened PRs, I'm done with my `improve the wording` series. 😄 

Anyway, thanks a lot for reviewing it! ❤️

PR #12304, opened 2025-09-11
fix(sitemap): add API references to config options

ArmandPhilippot:
#### Description (required)

* Adds `Type`, `Default` and `Since` to the sitemap config options:
  * `filter`, `customPages` and `canonicalURL` were added in `0.1.0` which predates Astro v1.0.0, so no `Since` for them!
  * I couldn't find a PR/commit for `entryLimit`, `serialize` and `i18n` so I guess those also predate Astro v1.0.0
  * `changefreq`, `lastmod` and `priority` were added in `0.2.0` which was released as the same time as `astro@1.0.0-beta.48` (see https://github.com/withastro/astro/pull/3579)
  * others are listed in the [changelog](https://github.com/withastro/astro/blob/main/packages/integrations/sitemap/CHANGELOG.md)
* Removes the `i18n` types inside the description to avoid redundancy
* Incorporates both "Read more" links of the `i18n` option into the description (looking at the history - #954 - this was the format used in the README at the time and it was never updated)

I wasn't totally sure how to deal with `changefreq, lastmod, and priority` since they are all described in the same heading. I used an object to define the types but I'm open to other ideas!

- Suggested label:  `consistency/formatting`, `improve or update documentation`

This doesn't have to wait for #12250 but just noting the `integrations-guide/sitemap.mdx` file is also updated there (potential Git conflict?).

### DIFF
--- a/src/content/docs/en/guides/typescript.mdx
+++ b/src/content/docs/en/guides/typescript.mdx
@@ -93,7 +93,7 @@ To check that the plugin is working, create a `.ts` file and import an Astro com
 
 ### UI Frameworks
 
-If your project uses a [UI framework](/en/guides/framework-components/), additional settings depending on the framework might be needed. Please see your framework's TypeScript documentation for more information. ([Vue](https://vuejs.org/guide/typescript/overview.html#using-vue-with-typescript), [React](https://react-typescript-cheatsheet.netlify.app/docs/basic/setup), [Preact](https://preactjs.com/guide/v10/typescript), [Solid](https://www.solidjs.com/guides/typescript))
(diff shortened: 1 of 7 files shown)


ArmandPhilippot:
Sounds like a good plan to me!

I added a small update (I knew I had forgotten something): wrapping each option name with backticks! This matches what we describe in [API references](https://contribute.docs.astro.build/reference/api-references/) and the formatting used in the other PR. After that, I think this is ready to go!

sarah11918:
Ship it! 🫡

ArmandPhilippot:
Thanks, I wasn't sure if I should merge it after the small update or wait for a second look!

Spoiler: other integration guides are coming! Except the `db` one (a bit more tricky because of `zod`), I think I have everything ready but I didn't want to spoil your week-end so this will wait for next week. 😄

PR #12315, opened 2025-09-14
modernize eslint.config.mjs

InertSloth:
#### Description (required)

tseslint.config is deprecated and emits a warning, this changes to 9.x defineConfig() format, in this case should be backwards compatible.

Many astro users will peek at the official docs repo to see how they do things. So other astro projects may model after this eslint.config.mjs so it behooves us to make it pass as cleanly as possible. astro docs wether it wants to be or not will have the first time perception of being the "shining example" for how to do things.

First time contrib: Yes for PRs.

discord: inert.sloth

\- 🦥

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -2,9 +2,10 @@
 import eslint from '@eslint/js';
 import eslintPluginAstro from 'eslint-plugin-astro';
 import globals from 'globals';
+import { defineConfig } from 'eslint/config';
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks @InertSloth!

I dug up the [migration guide](https://typescript-eslint.io/packages/typescript-eslint/#migrating-to-defineconfig), just to double check this and it looks good 👍

PR #12316, opened 2025-09-15
Add "Accessing route data" section in `getStaticPaths` with `routePattern`

openscript:
Add section on accessing route data and example code for getStaticPaths.

#### Description (required)

This documents the [proposal](https://github.com/withastro/astro/pull/13520) to make `routePattern` available in `getStaticPaths`. With `routePattern` it's possible to dynamically calculate the `params` that `getStaticPaths` returns. Before it was not possible to access the `routePattern` because `getStaticPaths` is executed at a time when `Astro.routePattern` is not available.

#### For Astro version: `5.14`. See astro PR [#13520](https://github.com/withastro/astro/pull/13520).

Known as openscript in Discord.

### DIFF
--- a/src/content/docs/en/guides/images.mdx
+++ b/src/content/docs/en/guides/images.mdx
@@ -433,6 +433,33 @@ import Logo from '../assets/logo.svg';
 <Logo width={64} height={64} fill="currentColor" />
 ```
 
+#### `SvgComponent` Type
+
+<Since v="5.14.0" />
+
+You can also enforce type safety for your `.svg` assets using the `SvgComponent` type:
+
+```astro title="src/components/Logo.astro"
+---
+import type { SvgComponent } from "astro/types";
+import HomeIcon from './Home.svg'
+
+interface Link {
+	url: string
+	text: string
+	icon: SvgComponent
+}
+
+const links: Link[] = [
+	{
+		url: '/',
+		text: 'Home',
+		icon: HomeIcon
+	}
+]
+---
+```
+
 ### Creating custom image components
 
 You can create a custom, reusable image component by wrapping the `<Image />`  or `<Picture/>` component in another Astro component. This allows you to set default attributes and styles only once.


src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
 <h1>{id}: {post.name}</h1>
 ```
 
+### Accessing route data
```
sarah11918:
```suggestion
### `routePattern`
```
I think based on the format of below, (e.g. starting with something definition-y like "A string that can be returned..") it makes sense to use `routePattern` as the actual section header!

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
+
+<p>
+
+<Since v="5.14.0" />
```
sarah11918, line 178: @ArmandPhilippot  can maybe jump in here! 😄

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
+</p>
+
+A string that can be returned from [`getStaticPaths()`](#getstaticpaths) to access the current [`routePattern`](/en/reference/api-reference/#routepattern).
+
+If a page `src/pages/[...locale]/[files]/[slug].astro` is being rendered, the corresponding `routePattern` is `[...locale]/[files]/[slug]`:
```
sarah11918: If this is *identical* to `Astro.routePattern`, then I don't think we need to say this here. I think maybe it does fit when distinguishing `routePattern` from `params`, though?

Instead, it feels like maybe we could have something that explains the need/use case, maybe more like (just words here as an example of what I think could fit, make them accurate! 😅 )

```suggestion
This provides data from the [Astro render context](https://docs.astro.build/en/reference/api-reference/) that would not otherwise be available within the scope of `getStaticPaths()` and can be useful to calculate the `params` and `props` for each page route.
```

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
+
+Unlike `params`, which are concrete values for a page, `routePattern` always reflects the original dynamic segment definition in the file path.
```
sarah11918:
```suggestion
`routePattern` always reflects the original dynamic segment definition in the file path (e.g. `/[...locale]/[slug]`), unlike `params`, which are explicit values for a page (e.g. `/fr/post-1/`).
```
- not sure whether "explicit" is the right word? "static" maybe? But I don't think "concrete" is the right word. So anyone have suggestions here?
- does this example make sense? Do we need to show the initial `.astro` route for this to make sense? (I feel like `routePattern` might already do that itself? And here, the value is in showing how it is different from `params` maybe?)

Also, not as easy to suggest here on GitHub, but I might consider putting this *before* the previous line, so that the flow goes:

- defintion (a string returned...), with link to what `routePattern` actually is (so we don't have to overdefine here)
- what `routePattern` reflects, and how this is different from `params` (the more `getStaticPaths` specific concerns)
- why this is useful, how it can be used
- Code example that shows using it.

src/content/docs/en/reference/routing-reference.mdx
florian-lefebvre: Btw realistically it will be part of 5.15

sarah11918: Good to know, thanks! Feel free to update the version once we're sure it's not making it in 5.14!

src/content/docs/en/reference/routing-reference.mdx
ArmandPhilippot: I think it also makes it a little more explicit what the section contains. "route data" might be a bit vague (is it the route format? the data associated to the route, ie. `props`?). So, I agree with the renaming!

src/content/docs/en/reference/routing-reference.mdx
ArmandPhilippot: If we change the heading (previous suggestion), we'll also need to add:
```
**Type:** `string`<br />
```
before `Since`.

(Github doesn't like when Sarah calls me. I don't always get the notifications... 😅 I've double-checked my Github settings but I can't see anything wrong...)

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
 ```
 
+### Accessing route data
+
+<p>
+
+<Since v="5.14.0" />
+</p>
+
+A string that can be returned from [`getStaticPaths()`](#getstaticpaths) to access the current [`routePattern`](/en/reference/api-reference/#routepattern).
```
ArmandPhilippot: We don't return a string but an object. I think it sounds a bit off (maybe it's just me as a non native speaker).  So, maybe we should say something like:
```suggestion
A string accessible from [`getStaticPaths()`](#getstaticpaths) to access the current [`routePattern`](/en/reference/api-reference/#routepattern).
```
If this is not explicit enough, maybe ``accessible from `getStaticPaths()` arguments``.

src/content/docs/en/reference/routing-reference.mdx
ArmandPhilippot: * Both `explicit` and `static` sounds okay to me. I think I prefer `explicit` as a non native speaker.
* If you mean the example below, I agree, I don't think this looks like the best use case. The one in the implementation PR makes more sense to me. I'm adding a comment on the code snippet.
* Your suggested flow looks great to me!

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
+      params.push({ files: getMessage(locale, "files") });
+    });
+  }
+
+  return { params };
+}
+
+const { page } = Astro.props;
+---
+```
```
ArmandPhilippot: We might need a "valid" example here, I mean:
* `locales` is not defined
* we don't include `locale` and `slug` in `params`
* we don't return props, so the `Astro.props` line seems wrong here
* also I'm a bit confused because `routePattern.includes("[files]")` would always be `true` on `src/pages/[...locale]/[files]/[slug].astro`, this doesn't show how this could be useful I think.

I was trying to rewrite the example, but actually I don't really see a use case for using `routePattern` directly inside `getStaticPaths`. We could split and trim `routePattern` to retrieve `locale`, `files` and `slug` but... why doing that instead of hardcoding them on that path where they are expected to exist?

The only use case I can see is inside a reusable helper that would parse `routePattern` itself like the example you showed in the implementation PR. Are there other use cases? If not, maybe something like the following example is enough:
```astro
---
import { getLocalizedData } from "../../../utils/i18n";

export async function getStaticPaths({ routePattern }) {
  const response = await fetch('...');
  const data = await response.json();

  console.log(routePattern); // [...locale]/[files]/[slug]

  // Your custom helper will be responsible of parsing the `routePattern` value to return an array of objects matching
  // `{ params: { locale: "en", files: "localized-string", slug: "localized-slug" }, props: { /* your data */ } }`
  return data.flatMap((file) => getLocalizedData(file, routePattern));
}

const { locale, files, slug } = Astro.params;
---
```

What's your thoughts? Can you see an example where `routePattern` is actually useful inside `getStaticPaths`? And feel free to adjust the code! It seems you already have a great use case in mind and I think it makes more sense than the one described here.

src/content/docs/en/reference/routing-reference.mdx
openscript: I've copied this from the `paginate()` below the new section I'm adding. 

```md
A function that can be returned from [`getStaticPaths()`](#getstaticpaths) to divide a collection of content items into separate pages.
```

Should I leave it as it is for homogeneity or also adapt the `paginate()` to `accessible`?

ArmandPhilippot: Yeah I noticed that but `paginate()` can and must be returned:
```ts
export async function getStaticPaths({ paginate }) {
  /* ... */
  return paginate(data, { pageSize: 10 });
}
```
while this is not the case for `routePattern`! Hence my confusion here. So I think `accessible` would be more accurate.

openscript: Ah got it! Thank you for clarification.

openscript: I've shortened it to:

```md
A string parameter [`getStaticPaths()`](#getstaticpaths) to access the current [`routePattern`](/en/reference/api-reference/#routepattern).
```

So we don't have `access` twice in the same sentence.

ArmandPhilippot: Looks even better, thanks!

src/content/docs/en/reference/routing-reference.mdx
openscript: You are right, the example I initially provided is incomplete and misleading. The only use case I know of are helper functions. Maybe it even doesn't make sense to provide an example here? People will already know what it's used for, when they start looking for it. For the daily business with Astro, you probably wont need it. What do you think?

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,40 @@ const { post } = Astro.props;
...
+
+<p>
+
+**Type:** `string` <br />
+<Since v="5.15.0" />
+</p>
+
+A string parameter [`getStaticPaths()`](#getstaticpaths) to access the current [`routePattern`](/en/reference/api-reference/#routepattern).
+
+This provides data from the [Astro render context](https://docs.astro.build/en/reference/api-reference/) that would not otherwise be available within the scope of `getStaticPaths()` and can be useful to calculate the `params` and `props` for each page route.
```
sarah11918:
```suggestion
This provides data from the [Astro render context](/en/reference/api-reference/) that would not otherwise be available within the scope of `getStaticPaths()` and can be useful to calculate the `params` and `props` for each page route.
```
Just fixing the link!

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,40 @@ const { post } = Astro.props;
...
 
+### `routePattern`
+
+<p>
+
+**Type:** `string` <br />
+<Since v="5.15.0" />
+</p>
+
+A string parameter [`getStaticPaths()`](#getstaticpaths) to access the current [`routePattern`](/en/reference/api-reference/#routepattern).
```
sarah11918:
```suggestion
A string parameter of [`getStaticPaths()`](#getstaticpaths) to access the current [`routePattern`](/en/reference/api-reference/#routepattern).
```
We still need *some* word here so that it makes grammatical sense 😄 . What's the appropriate word here? is "parameter *of* `getStaticPaths()`" correct? *on*?

A string paramater available *in* `getStaticPaths()`? available *to* `getStaticPaths()`?

src/content/docs/en/reference/routing-reference.mdx
ArmandPhilippot: I'm a bit torn to be honest. Although an example does not directly show what to do with `routePattern` since we agree that this is mostly useful in an helper, maybe code snippets talk louder than words for some and seeing one will help them understand what it is and how they could use it? I think the other sections also have code snippets, so this would also help to be consistent.

src/content/docs/en/reference/routing-reference.mdx
ArmandPhilippot: Oh, good catch. 😅 Maybe `in`? Not sure. `of` makes be think this is a direct parameter of `getStaticPaths` (ie. `getStaticPaths(routePattern)`) while this is a property of the object available as first argument of `getStaticPaths`.

openscript: Then actually it's not a parameter, but a property in the parameter. How about the following?

```suggestion
A string property in [`getStaticPaths()`](#getstaticpaths) object parameter to access the current [`routePattern`](/en/reference/api-reference/#routepattern).
```

src/content/docs/en/reference/routing-reference.mdx
openscript: I see your point. I think your example fits this need, to give an idea, for what `routePattern` is intended here.

src/content/docs/en/reference/routing-reference.mdx
ArmandPhilippot: This sounds more accurate but maybe also a bit "heavy"? Maybe, considering the [getStaticPaths type](https://docs.astro.build/en/reference/routing-reference/#getstaticpaths), the following could work:
```
A property available in `getStaticPaths()` options to access the current `routePattern` as a string.
```

But I'm not the wording expert here so maybe Sarah will have a better idea than me. 😄

sarah11918: Let's go with that, at least it's English! 😅 

I'll commit it for now, but speak up if anyone decides there's a something better!

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
+---
+import { getLocalizedData } from "../../../utils/i18n";
+
+export async function getStaticPaths({ routePattern }) {
+  const response = await fetch('...');
+  const data = await response.json();
+
+  console.log(routePattern); // [...locale]/[files]/[slug]
+
+  // Your custom helper will be responsible of parsing the `routePattern` value to return an array (for each locale) of objects matching
```
sarah11918: OK, we might need to more explicitly mention that it's intended to be used inside helper functions (that go in `getStaticPaths()`? Right now, there's a lot going on here to understand when just skimming/glancing at it. 😄 

And, instead of a code comment to try to get that across, can we make this a sentence/paragraph to introduce the code sample, so people know what the heck they're looking at before they're faced with this?

So, something like (help me with the wording!):

> You can pass `routePattern` to `getStaticPaths` for use inside a custom helper function to return explicit route data. The following example uses `routePattern` in a custom `getLocalizedData` helper function to make `locale`, `files`, and `slug` available from `Astro.params`. Then, these explicit values can be used in your page template.

THEN, maybe the code comment can be a shortened version of the second line if you think showing the explicit data returned is helpful there?

(And, maybe show one line of the template, showing a use case? Do we think that is helpful here, or too much??)

Note: I might also suggest highlighting `routePattern` in this code snippet! and/or pointing out `getLocalizedData` somehow?

openscript: > THEN, maybe the code comment can be a shortened version of the second line if you think showing the explicit data returned is helpful there?

I think that's a good idea. What do you think @ArmandPhilippot?

> (And, maybe show one line of the template, showing a use case? Do we think that is helpful here, or too much??)

As this is more related with what URL a page can be rendered, it's not releated to the template or UI that much. I wouldn't add this.

> Note: I might also suggest highlighting routePattern in this code snippet! and/or pointing out getLocalizedData somehow?

I like this idea. So the the reader immediately sees, ah those are the special ones.

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
+
+**Type:** `string` <br />
+<Since v="5.15.0" />
+</p>
+
+A property available in [`getStaticPaths()`](#getstaticpaths) options to access the current [`routePattern`](/en/reference/api-reference/#routepattern) as a string.
+
+This provides data from the [Astro render context](/en/reference/api-reference/) that would not otherwise be available within the scope of `getStaticPaths()` and can be useful to calculate the `params` and `props` for each page route.
+
+`routePattern` always reflects the original dynamic segment definition in the file path (e.g. `/[...locale]/[slug]`), unlike `params`, which are explicit values for a page (e.g. `/fr/post-1/`).
```
openscript: I think we should change the route pattern to the same example `[...locale]/[files]/[slug]` like in the code block bellow.

src/content/docs/en/reference/routing-reference.mdx
ArmandPhilippot: Yes, I like Sarah's ideas (as always? 😆 ).

> You can pass routePattern to getStaticPaths for use

We don't pass anything, Astro passes it to the function for us. So maybe `access ... in` instead of `pass ... to`? But then, it sounds a bit redundant with what we already have at the beginning of the section. So, I don't know, maybe I'm just quibbling and that's fine!

> to make locale, files, and slug available from Astro.params. Then, these explicit values can be used in your page template.

This, and this also affects the generated routes which is maybe the first goal? So maybe we can insist on it: `Then, these explicit values can be used in your page template and will be used to generate the routes.`

Actually, nothing differs from what we describe in [`params`](https://docs.astro.build/en/reference/routing-reference/#params) except we don't "hard code" the values for each segment ​​and instead we rely on a custom helper to do that based on the `routePattern` value.

src/content/docs/en/reference/routing-reference.mdx
sarah11918: Yes, makes perfect sense to keep it consistent! Can you also then update the corresponding `params` value, please?

src/content/docs/en/reference/routing-reference.mdx
sarah11918: I want your quibbling, @ArmandPhilippot ! Are you willing to make a suggestion with, top to bottom, what you think that paragraph should be, and then I'll just make sure it's English. 😆 ?

ArmandPhilippot: Maybe a bit lengthy, but I was thinking at something like this before the example:
> The following example shows how to localize your route segments by parsing `routePattern` accessed from `getStaticPaths`. `routePattern` is passed to a custom `getLocalizedData` helper function, which returns an array of static paths. The [`params`](/en/reference/routing-reference/#params) object will set with explicit values for each route segment: `locale`, `files`, and `slug`. Then, these values will be used to generate the routes and can be used in your page template via the `Astro.params`.

Then, I guess the comment can be reduced to:
```
// Call your custom helper with `routePattern` to generate the static paths
```
or even without `routePattern` if we highlight it!

src/content/docs/en/reference/routing-reference.mdx
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
+
+```astro title="src/pages/[...locale]/[files]/[slug].astro"
```
sarah11918:
``````suggestion
The following example shows how to localize your route segments and return an array of static paths by passing `routePattern` to a custom `getLocalizedData()` helper function. The [params](https://github.com/en/reference/routing-reference/#params) object will be set with explicit values for each route segment: `locale`, `files`, and `slug`. Then, these values will be used to generate the routes and can be used in your page template via `Astro.params`.


```astro title="src/pages/[...locale]/[files]/[slug].astro" "routePattern" "getLocalizedData"
``````
OK, what do we think of this? Please do correct nuance, but this is my take on "simple English of what @ArmandPhilippot said"

File: src/content/docs/en/reference/routing-reference.mdx (hunk 9 above)
Comment (sarah11918) [reply src/content/docs/en/reference/routing-reference.mdx]: Going to resolve this one so we can pick it up in the other comment, and it's easier to read!

File: src/content/docs/en/reference/routing-reference.mdx [hunk 12]
```diff
@@ -170,6 +170,39 @@ const { post } = Astro.props;
...
+import { getLocalizedData } from "../../../utils/i18n";
+
+export async function getStaticPaths({ routePattern }) {
+  const response = await fetch('...');
+  const data = await response.json();
+
+  console.log(routePattern); // [...locale]/[files]/[slug]
+
+  // Your custom helper will be responsible of parsing the `routePattern` value to return an array (for each locale) of objects matching
+  // `{ params: { locale: "en", files: "localized-string", slug: "localized-slug" }, props: { /* your data */ } }`
```
sarah11918:
```suggestion
  // Call your custom helper with `routePattern` to generate the static paths

File: src/content/docs/en/reference/routing-reference.mdx (hunk 11 above)
Comment (ArmandPhilippot) [reply src/content/docs/en/reference/routing-reference.mdx]: LGTM, thanks!

File: src/content/docs/en/reference/routing-reference.mdx [hunk 13]
```diff
@@ -170,6 +170,41 @@ const { post } = Astro.props;
...
+
+```astro title="src/pages/[...locale]/[files]/[slug].astro" "routePattern" "getLocalizedData"
+---
+import { getLocalizedData } from "../../../utils/i18n";
+
+export async function getStaticPaths({ routePattern }) {
+  const response = await fetch('...');
+  const data = await response.json();
+
+  console.log(routePattern); // [...locale]/[files]/[slug]
```
sarah11918, line 198: Do we think we need this line?

src/content/docs/en/reference/routing-reference.mdx
sarah11918:
```suggestion
`routePattern` always reflects the original dynamic segment definition in the file path (e.g. `/[...locale]/[files]/[slug]`), unlike `params`, which are explicit values for a page (e.g. `/fr/fichiers/article-1/`).
```
Happy to make this change, but I'm not sure what would go in `files`?

openscript: The french translation of `files`: `fichiers`

openscript: Thank you so much @sarah11918!

src/content/docs/en/reference/routing-reference.mdx
openscript: I like this line, because as a reader I go through the example and I'm like.. huh.. okay, this is the `routePattern`.. I expect it looks like this. Then as a reader you have reassurance and confirmation what the value is you might expect. For me such lines always feel good, when I'm somewhat new to a topic.

sarah11918: Sold!

src/content/docs/en/reference/routing-reference.mdx
sarah11918: Thanks! OK, updated fichiers, but in your scenario, would someone therefore *also* want the slug to be `poste-1` do you think? @openscript

sarah11918: (Or, if these are not "blog posts" at all, do we want something entirely different here? `projects` and `projets` for example?)

openscript: Yes, I think it should be `poste-1` as well. This are blog post slugs.

I usually do this in my projects, so I can translate all segments of a path to a locale:

image

ArmandPhilippot: .In French, we use `article` (`poste` is rather for a job in a company or to designate a TV or a radio station).

sarah11918: OK, will commit `article`, and then any changes you want please suggest separately! Merci tout le monde!

sarah11918 (approved):
Approving for 5.14! :tada:

yanthomasdev (approved):
LGTM! Good job 🙌

PR #12319, opened 2025-09-15
fix(integrations-guide): add/format API references blocks

ArmandPhilippot:
#### Description (required)

* Updates the integrations guide to add or format the API reference blocks.
* Updates slightly two descriptions in `integrations-guide/db` to remove the inline types and to make the description more accurate based on the types.

This looks like a big update but it's not (each block is ~4-7 added lines) that's why I've updated all the integrations guides in one PR.

What you might want to double check is `foreignKeys` in `db`, `recmaPlugins` in `mdx` and the options in `partytown` (see the details for why). I think the choices are reasonable but you might have another opinion!

<details>
<summary>Show details</summary>

For each integration API, I added a link so you can easily check there is no errors. I've added some comments for some to explain my choice.

- In `alpinejs`:
  - `entrypoint`, see https://github.com/withastro/astro/pull/9802

- In `db`:
  - `columns`: thankfully the [type is exported from `@astrojs/db/types`](https://github.com/withastro/astro/blob/a791684e7a0fb35d71baf3df9dcb087ed721f35a/packages/db/src/core/types.ts#L59) so we don't need something more complex
  - [`indexes`](https://github.com/withastro/astro/blob/a791684e7a0fb35d71baf3df9dcb087ed721f35a/packages/db/src/core/types.ts#L71-L83): we could use `keyof ColumnsConfig` for `on`, but I think keeping `string` as we had before in the description is less confusing, so I reused that.
  - `foreignKeys`: this one is complex, we don't have a type to reference a single column defined in `columns`. In the description, we had `Column` but the type doesn't exist. I don't have a better idea so I reused that, NWTWWHB. And I updated the description to be more accurate (the [type allows a single column](https://github.com/withastro/astro/blob/a791684e7a0fb35d71baf3df9dcb087ed721f35a/packages/db/src/core/types.ts#L67-L70), this is not always an array).
  - `isDbError()`, see https://github.com/withastro/astro/pull/10498

- In `markdoc`: (technically speaking, both are not set to `false` by default but rather left undefined... which would have the same behavior as `false`: only `true` does something.)
  - `allowHTML`, see https://github.com/withastro/astro/pull/7597
  - `ignoreIndentation`, see https://github.com/withastro/astro/pull/8802

- In `mdx`:
  - `extendMarkdownConfig`, see https://github.com/withastro/astro/pull/5684
  - `recmaPlugins`, see https://github.com/withastro/astro/pull/5146 (note: `PluggableList` is not an Astro type but a `unified` one and I don't think we reexport it, but I don't have a better idea to describe the type)
  - `optimize`, see https://github.com/withastro/astro/pull/7151

- In `partytown`: this one is tricky because all the config options comes directly from `partytown`... so no `Since`, and it's possible that the types change if `partytown` is updated! But I still added the type for the few config options we describe based on https://github.com/QwikDev/partytown/blob/main/src/lib/types.ts#L398

- In `preact`:
  - `compat`, see https://github.com/withastro/astro/pull/3759

- In `solid-js`: I only formatted an API reference block

- In `vercel`: the missing `Since` are corrects, I only formatted the API reference blocks

- In `vue`:
  - `appEntrypoint`, see https://github.com/withastro/astro/pull/5075
  - [`jsx`](https://github.com/withastro/astro/blob/a791684e7a0fb35d71baf3df9dcb087ed721f35a/packages/integrations/vue/src/index.ts#L14): [the second type is a renamed imported type](https://github.com/withastro/astro/blob/a791684e7a0fb35d71baf3df9dcb087ed721f35a/packages/integrations/vue/src/index.ts#L4), so I think `object` is enough since we explain below what the object is.
  - [`devtools`](https://github.com/withastro/astro/blob/a791684e7a0fb35d71baf3df9dcb087ed721f35a/packages/integrations/vue/src/index.ts#L16): same, we explain the object type in "Customizing Vue DevTools" so I think this is enough

</details>

I haven't updated the other integrations guide because the API references were already there!

- Suggested label:  `consistency/formatting`, `improve or update documentation`


src/content/docs/en/guides/integrations-guide/db.mdx
```diff
@@ -172,12 +183,17 @@ This will generate a unique index on the `authorId` and `published` columns with
 
 The following configuration options are available for each index:
 
-- `on`: `string | string[]` - A single column or array of column names to index.
-- `unique`: `boolean` - Set to `true` to enforce unique values across the indexed columns.
-- `name`: `string` (optional) - A custom name for the unique index. This will override Astro's generated name based on the table and column names being indexed (e.g. `Comment_authorId_published_idx`). Custom names are global, so ensure index names do not conflict between tables.
+- `on` - A single column or array of column names to index.
+- `unique` - Set to `true` to enforce unique values across the indexed columns.
+- `name` (optional) - A custom name for the unique index. This will override Astro's generated name based on the table and column names being indexed (e.g. `Comment_authorId_published_idx`). Custom names are global, so ensure index names do not conflict between tables.
```
sarah11918: If `unique` and `boolean` are both optional, then we should be consistent about whether we include "(optional)" or not. Maybe slight preference for adding in, but not a strong one!

```suggestion
- `unique` (optional) - Set to `true` to enforce unique values across the indexed columns.
- `name` (optional) - A custom name for the unique index. This will override Astro's generated name based on the table and column names being indexed (e.g. `Comment_authorId_published_idx`). Custom names are global, so ensure index names do not conflict between tables.
```

src/content/docs/en/guides/integrations-guide/markdoc.mdx
```diff
@@ -8,10 +8,10 @@ githubIntegrationURL: 'https://github.com/withastro/astro/tree/main/packages/int
 category: other
 i18nReady: true
 ---
-import { FileTree } from '@astrojs/starlight/components';
+import { FileTree, Steps } from '@astrojs/starlight/components';
```
sarah11918, line 11:
```suggestion
import { FileTree } from '@astrojs/starlight/components';
```
Just checking, did you mean to add the Steps import here, since I don't see you adding the `<Steps />` component? (Maybe copy/paste?)

src/content/docs/en/guides/integrations-guide/partytown.mdx
```diff
@@ -110,7 +110,12 @@ export default defineConfig({
 This mirrors the [Partytown config object](https://partytown.qwik.dev/configuration/).
```
sarah11918:
```suggestion
This mirrors the [Partytown config object](https://partytown.qwik.dev/configuration/) and all options can be set in `partytown.config`. Some common configuration options for Astro projects are described on this page. 
```
So, we don't list *everything* that Partytown does here. This could mean that we said "mirrors" when only these items actually exist, and in the Astro integration you can only set the following options. OR all of the Partytown options are actually available, and we just only ever documented some of them.

This is a way we could address the latter situation. Another way we could address the later situation is to not document *any* of them, so we never get out of sync? 😅 

Can we confirm that our integration is always compatible with the latest version of Partytown, and there's never a risk of our integration not supporting a new config option if they add it? (Is this kind of a "passthrough" situation, or do we in fact create our own mirror config options for the integration?)

sarah11918 (commented):
This looks great, @ArmandPhilippot ! I've reviewed for syntax, and left thoughts about Partytown.

It would be great if someone had thoughts on the `db` and `mdx` (`recmaPlugins`) types, since that's out of my comfort zone!

src/content/docs/en/guides/integrations-guide/markdoc.mdx
ArmandPhilippot: We had two imports from `'@astrojs/starlight/components'` (`FileTree`, then before `ReadMore` we were importing `Steps`), so I gather the two in an unique import! This was just a nit, I don't think it changes anything...

src/content/docs/en/guides/integrations-guide/partytown.mdx
ArmandPhilippot: Your suggestion makes sense. Now, between showing some examples or none I'm a bit torn. I think it's nice to show updating a Partytown setting for less experienced people, but, yes, this means we could be out of sync. I'm not familiar with partytown so maybe they don't introduce breaking changes often, but this could happen.

Looking [at the integration code](https://github.com/withastro/astro/blob/a791684e7a0fb35d71baf3df9dcb087ed721f35a/packages/integrations/partytown/src/index.ts#L14), we [passthrough the config options](https://github.com/withastro/astro/blob/a791684e7a0fb35d71baf3df9dcb087ed721f35a/packages/integrations/partytown/src/index.ts#L33) except `debug` maybe.

One idea could be to use your suggestion and instead of headings for the options, we show a single code snippet updating some settings. The code snippet could become out of sync, but this is maybe less annoying presented this way (ie. this is an example, we don't describe how the options work).

ArmandPhilippot:
Thanks Sarah! I've answered for partytown. Now, for `mdx` and `db`, I'm not sure.

Maybe @Adammatthiesen (if you have time to look at it) would have an idea for the `foreignKeys` type since he's working on `db`.

For Adam, to avoid you reading all the details: In the description, we had `Column` to describe the [return type of `foreignKeys`](https://deploy-preview-12319--astro-docs-2.netlify.app/en/guides/integrations-guide/db/#foreignkeys) but the type doesn't exist. I don't have a better idea so I reused what we had. Do you know if there is a more explicit type here we could use? Or do you think it's fine even if the type doesn't exist? Any ideas are welcome!

But for `mdx` I don't know who we could ask... I hesitated to use `unified.PluggableList`, but I wasn't sure this would be more explicit for people.

src/content/docs/en/guides/integrations-guide/markdoc.mdx
sarah11918: Ah, I didn't notice, it just looked like Steps was added in the diff! Perfect, carry on! 🫡

src/content/docs/en/guides/integrations-guide/partytown.mdx
sarah11918: I think there's precedent for us doing something similar somewhere else in docs -- it's not *our* config, so we don't show everything, but we say here are some common ones that might be helpful.

Whether we do it by section heading (maybe we don't, if we're not strictly "API documenting") or by common use case examples/snippets, I think it's OK to show some commonly-used setups and then say "see the full Partytown documentation for all available options."

So if this page becomes more use-oriented than API oriented (because we're offloading the API documenting to Partytown), I think that's fine! We can think about what's *helpful* on this page vs being thorough/complete.

src/content/docs/en/guides/integrations-guide/mdx.mdx
```diff
@@ -326,13 +330,25 @@ export default defineConfig({
 
 ### `recmaPlugins`
 
+<p>
+
+**Type:** `PluggableList`<br />
```
sarah11918, line 335: > But for mdx I don't know who we could ask... I hesitated to use unified.PluggableList, but I wasn't sure this would be more explicit for people.

Let's ask @delucis !

(Armand's concern is regarding the entry for `recmaPlugins`: "PluggableList is not an Astro type but a unified one and I don't think we reexport it, but I don't have a better idea to describe the type.)

src/content/docs/en/guides/integrations-guide/partytown.mdx
ArmandPhilippot: It speaks to me, but I don't remember where!
An option could be to keep the existing structure but:
* add your suggestion because I think it makes things clearer
* remove the `Type`s
* and maybe instead of `config.*` as heading, use words to make it sound more like examples than an API (since this is not our API)? Something like:
  * Enabling debug mode
  * Forwarding variables
  * Proxying requests

sarah11918: Great idea!

Exactly that!

Adammatthiesen:
> Thanks Sarah! I've answered for partytown. Now, for `mdx` and `db`, I'm not sure.
> 
> Maybe @Adammatthiesen (if you have time to look at it) would have an idea for the `foreignKeys` type since he's working on `db`.
> 
> For Adam, to avoid you reading all the details: In the description, we had `Column` to describe the [return type of `foreignKeys`](https://deploy-preview-12319--astro-docs-2.netlify.app/en/guides/integrations-guide/db/#foreignkeys) but the type doesn't exist. I don't have a better idea so I reused what we had. Do you know if there is a more explicit type here we could use? Or do you think it's fine even if the type doesn't exist? Any ideas are welcome!
> 
> But for `mdx` I don't know who we could ask... I hesitated to use `unified.PluggableList`, but I wasn't sure this would be more explicit for people.

Yeah, im not sure where the original docs writer got that type from :sweat_smile: But the new description looks correct, since there really is no proper external type for that. (as currently ALL of the `astro:db` types are virtual only and cannot be directly referenced. I also plan to change this when i do my rebuild of DB... because it annoy's me :sweat_smile: )

ArmandPhilippot:
Thanks @Adammatthiesen 

> Yeah, im not sure where the original docs writer got that type from

I think the idea was to illustrate a single column configured inside `columns` in the absence of an actual type to use. And to be clear, I removed it from the description... but I'm reusing it inside `Type:` because we don't have an exported type for that and so I'm not sure what we could show here.

> ALL of the astro:db types are virtual only and cannot be directly referenced

Well, I was able to import `ColumnsConfig` from `@astrojs/db/types` while checking the types to use! 😄 (I'm using it for `indexes`). But, yeah, I wasn't able to find a meaningful type for `foreignKeys`...
So I guess the question would be should we use:
```
Type: { columns: string | string[]; references: () => Column | Column[]; }[]
```
even if `Column` doesn't exist. Or should we remove that type (while waiting for you to finish the DB rebuild 😀 ) because it's not really accurate? Do you think it's confusing/harmful to use that?

Adammatthiesen:
> Well, I was able to import `ColumnsConfig` from `@astrojs/db/types` while checking the types to use! 😄 (I'm using it for `indexes`). But, yeah, I wasn't able to find a meaningful type for `foreignKeys`... So I guess the question would be should we use:
> 
> ```
> Type: { columns: string | string[]; references: () => Column | Column[]; }[]
> ```
> 
> even if `Column` doesn't exist. Or should we remove that type (while waiting for you to finish the DB rebuild 😀 ) because it's not really accurate? Do you think it's confusing/harmful to use that?

yeah... there is a few very specific types available, but the primary table/column types are not at the moment.

For now probably fine to just keep the Mock Column type as long as we are explaining that the user is using a existing table's column (as shown in the code snippet example)

src/content/docs/en/guides/integrations-guide/partytown.mdx
sarah11918: So we're agreed. Which one of us is doing this....?  😂

ArmandPhilippot: Ah, it's not automatic? 😞 Okay, I'm going to commit your suggestion and add the other changes in a separate commit! 🫡

sarah11918:
@ArmandPhilippot Shall we NWTWWHB this one and get it merged? I know there was one outstanding question, but maybe we merge as-is then change if we find it's not helpful or have a better idea?

ArmandPhilippot:
@sarah11918 If someone is using a `recmaPlugin` (which is less known than remark/rehype), I assume they are already familiar with the unified ecosystem (and `PluggableList` should talk to them). So, I don't think the `recmaPlugin` question is really a blocker.
At worth, if someone find this confusing I'm sure we'll see a post about it on Discord or here and so we can find a better solution later if necessary.

So yes, I think we can merge this one! 🫡

PR #12323, opened 2025-09-16
Document experimental `useActionState()` helpers with Actions

yanthomasdev:
#### Description (required)

There are two APIs related to using Astro Actions with React 19 that have been experimental for quite a while and also **not** documented. Fun fact, I've found out this by rewatching [Ben's Astro Together talk](https://youtu.be/VkYQMhit_04?si=KLF3Izy__c1m5nia&t=945). 

Until this point, the only documentation we had about it is on the [3.4.0 `@astrojs/react` changelog](https://github.com/withastro/astro/blob/main/packages/integrations/react/CHANGELOG.md#minor-changes-8). 

This PR adds information about both `withState()` and `getActionState()` functions exported from `@astrojs/react`. We should discuss if we want to add anything about this to the Actions guide itself.

- Related: https://github.com/withastro/astro/pull/14386


src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +114,54 @@ To use your first React component in Astro, head to our [UI framework documentat
 * 💧 client-side hydration options, and
 * 🤝 opportunities to mix and nest frameworks together
 
+## Integrate Actions with `useActionState()` (experimental)
+
+[Astro Actions][astro-actions] can be integrated with [React's `useActionState()` hook](https://react.dev/reference/react/useActionState) to update client-side state when triggering actions during form submission.
```
sarah11918:
```suggestion
The `@astrojs/react` integration provides two experimental functions for use with [Astro Actions][astro-actions]: `experimental_withState()` and `experimental_getActionState()`.

These can be used with [React's `useActionState()` hook](https://react.dev/reference/react/useActionState) to update and read client-side state when triggering actions during form submission.
```
What about an introduction that front-loads the things we're documenting, and makes it clear that these are to be used with Actions? Then, we describe what they do.

src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +114,54 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+
+To do so, pass `experimental_withState()` receiving the action you want to trigger in the `useActionState()` hook:
```
sarah11918:
```suggestion
### `experimental_withState()`

<p>

**Type:** `(actionName: string) => void<br />
<Since v="4.3.2" />
</p>

You can pass `experimental_withState()` and the action you want to trigger to React's `useActionState()` hook as the form action function. The example below passes a `like` action to increase a counter along with an initial state of `0` likes.
```
(I don't know for sure what the type should be, but let's make a section for each one of these!)

src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +114,54 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+      <input type="hidden" name="postId" value={postId} />
+      <button disabled={pending}>{state} ❤️</button>
+    </form>
+  );
+}
+```
+
+The `experimental_withState()` function will match the action's types with React's expectations and preserve metadata used for progressive enhancement, allowing it to work even when JavaScript is disabled on the user's device.
+
+You can also access the state stored by `useActionState()` on the server in your action `handler` by calling `experimental_getActionState()` passing in the API context, and optionally, apply a type to the result:
```
sarah11918:
```suggestion
### `experimental_getActionState()`

<p>

**Type:** `(context: APIContext) => void<br />
<Since v="4.3.2" />
</p>

You can access the state stored by `useActionState()` on the server in your action `handler` with `experimental_getActionState()`. It accepts the Astro API context, and optionally, you can apply a type to the result. The example below gets the current value of likes from a counter, typed as a number, in order to create an incrementing `like` action.
```
Still kind of guessing, but maybe something like this?

sarah11918 (commented):
Thank you for plugging this hole, @yanthomasdev ! I'll need your help below, but see what you think about some framing like below!

src/content/docs/en/guides/integrations-guide/react.mdx
yanthomasdev: Looks good to me, maybe I would be a bit more explicit with the second sentence to be clear that these functions are only to be used with `useActionState()` and don't make sense otherwise. Also, changing the order a bit in the end
```suggestion
The `@astrojs/react` integration provides two experimental functions for use with [Astro Actions][astro-actions]: `experimental_withState()` and `experimental_getActionState()`.

These are used with [React's useActionState() hook](https://react.dev/reference/react/useActionState) to read and update client-side state when triggering actions during form submission.
```

src/content/docs/en/guides/integrations-guide/react.mdx
yanthomasdev: Action types are a bit complex, and in this case, that's an Action type that returns a React type of this hook. I am not sure there's an actual benefit to documenting these?

src/content/docs/en/guides/integrations-guide/react.mdx
yanthomasdev: The written text itself is good to me, the type is wrong, though. It returns a `Promise<T>` (just now noticed the example in the Changelog I used as a base is wrong, it needs to be awaited), T being the value you typed it to be; otherwise, it will try to infer and fallback to `any` (I think?).

The text felt a bit cramped, so I think we can make it a new paragraph when talking about the example?

```suggestion
### `experimental_getActionState()`

<p>

**Type:** `(context: APIContext) => Promise<T><br />
<Since v="4.3.2" />
</p>

You can access the state stored by `useActionState()` on the server in your action `handler` with `experimental_getActionState()`. It accepts the Astro API context, and optionally, you can apply a type to the result. 

The example below gets the current value of likes from a counter, typed as a number, in order to create an incrementing `like` action:
```

src/content/docs/en/guides/integrations-guide/react.mdx
sarah11918: OK, agree about the emphasis on `useActionState`! Can we be more brutal about the verbosity, and get away with this:


> These are used with [React's `useActionState()` hook](https://react.dev/reference/react/useActionState) to read and update client-side state when triggering actions during form submission.

src/content/docs/en/guides/integrations-guide/react.mdx
sarah11918: OK, then ditch the types? I would still have section headings here, since we're illustrating each one individually.

yanthomasdev: OK, I somewhat found the types it expects. It's a bit of a mouthful, though 😅 

```suggestion
### `experimental_withState()`

<p>

**Type:** `(action: FormFn<T>) => (state: T, formData: FormData) => FormFn<T>`<br />
<Since v="4.3.2" />
</p>

You can pass `experimental_withState()` and the action you want to trigger to React's `useActionState()` hook as the form action function. The example below passes a `like` action to increase a counter along with an initial state of `0` likes.
```

src/content/docs/en/guides/integrations-guide/react.mdx
yanthomasdev: LGTM, updated the suggestion above to have this so you can just commit that when satisfied

src/content/docs/en/guides/integrations-guide/react.mdx
sarah11918: That looks good to me if you're good with it!

sarah11918:
Just to be clear, the docs we're making here can merge at any time, right, because these already exist undocumented? But, after a minor release, we only need to remove `experimental_`?

I'll let you decide how close you think we are to stabilizing these, and whether you just want *one* PR for the stable features. (Will make the `<Since />` easier too, since we tend to use the stable version.)

yanthomasdev:
> Just to be clear, the docs we're making here can merge at any time, right, because these already exist undocumented? But, after a minor release, we only need to remove experimental_?

Yes, that's what I was going for @sarah11918. I think it's fine to make two PRs, it will need to wait for the next minor to be stabilized anyway.

src/content/docs/en/guides/integrations-guide/react.mdx
yanthomasdev: Let me know what you feel about this colossal type or if we should remove it @sarah11918!

sarah11918: Let's ask @ArmandPhilippot , who won't see this, so we'll need to ping him in Discord! 😆

ArmandPhilippot: This looks correct to me (maybe with `<T>` at the beginning?), but yes I don't know if this is really helpful... I wonder if we can't simplify a bit because we are currently documenting a React type if I understood correctly. Maybe something like:
```
**Type:** `<T>(action: (formData: FormData) => Promise<T>) => typeof useActionState`
```

What do you think Yan since you're more familiar with the API? If not, maybe removing the type is fine for this one.

yanthomasdev: We could also simplify by not "exploding" the types (example below), or maybe something between this and Armand's suggestion. What y'all think?
```
**Type:** `(action: FormFn<T>) => (state: T, formData: FormData) => FormFn<T>
```

ArmandPhilippot: Oh yeah, I didn't noticed this was the exact same function earlier! Your suggestion looks better. Approved!

I'm a bit confused with this API (nothing to do with the type, this looks good). I wanted to double check using the content of your changeset in a fresh project and I got different Typescript errors when I was trying to use `experimental_withState`!

I managed to get ride of those errors, but it seems what is described in the changelog (what you reused in your changeset, right?) is not fully compatible with `"astro/tsconfigs/strict"` (the default in v5):
* using `"astro/tsconfigs/strict"`, I had to make some adjustments in the two files to get ride of the errors
* using `"astro/tsconfigs/base"`, I had to make adjustments only in the component file (where we use `experimental_withState`) to get ride of the errors... and so I wonder if this works correctly because I'm not returning the same format in the `defineAction` handler.

One thing I discovered is that [using `accept: "form"`](https://docs.astro.build/en/guides/actions/#accepting-form-data-from-an-action), in `defineAction` might be a prerequisite (with `strict`) to make the action compatible with `experimental_withState`. And so, we might need to document that...

I'll post a comment on your changeset in the other PR with the changes I had to make to get ride of the errors because we might want to show a working function (at least without Typescript errors if what you had was working).

yanthomasdev: > One thing I discovered is that [using accept: "form"](https://docs.astro.build/en/guides/actions/#accepting-form-data-from-an-action), in defineAction might be a prerequisite (with strict) to make the action compatible with experimental_withState

Yes, I believe this function is only meant to be used in the case of form-submitted actions. For actions evoked like RPC it isn't necessary. Good point

src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +114,66 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+}
+```
+
+The `experimental_withState()` function will match the action's types with React's expectations and preserve metadata used for progressive enhancement, allowing it to work even when JavaScript is disabled on the user's device.
+
+### `experimental_getActionState()`
+
+<p>
+
+**Type:** `(context: APIContext) => Promise<T><br />
```
yanthomasdev:
```suggestion
**Type:** `(context: APIContext) => Promise<T>`<br />
```

src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +114,66 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+  );
+
+  return (
+    <form action={action}>
+      <input type="hidden" name="postId" value={postId} />
+      <button disabled={pending}>{state} ❤️</button>
+    </form>
+  );
+}
+```
```
yanthomasdev:
````suggestion
```jsx title="Like.tsx" ins={2,7}
import { actions } from 'astro:actions';
import { experimental_withState } from '@astrojs/react/actions';
import { useActionState } from "react";

export function Like({ postId }: { postId: string }) {
  const [state, action, pending] = useActionState(
    experimental_withState(actions.like),
    { data: 0, error: undefined }, // initial likes and errors
  );

  return (
    <form action={action}>
      <input type="hidden" name="postId" value={postId} />
      <button disabled={pending}>{state.data} ❤️</button>
    </form>
  );
}
```
````

File: src/content/docs/en/guides/integrations-guide/react.mdx [hunk 6]
```diff
@@ -114,6 +114,66 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+      postId: z.string(),
+    }),
+    handler: async ({ postId }, ctx) => {
+      const currentLikes = await experimental_getActionState<number>(ctx);
+      // write to database
+      return currentLikes + 1;
+    },
+  })
+};
+```
```
yanthomasdev, line 189:
````suggestion
```ts title="actions.ts" ins={3,11}
import { defineAction, type SafeResult } from 'astro:actions';
import { z } from 'astro:schema';
import { experimental_getActionState } from '@astrojs/react/actions';

export const server = {
  like: defineAction({
    input: z.object({
      postId: z.string(),
    }),
    handler: async ({ postId }, ctx) => {
      const currentLikes = await experimental_getActionState<SafeResult<any, number>>(ctx);
      // write to database
      return currentLikes + 1;
    },
  })
};
```
````

File: src/content/docs/en/guides/integrations-guide/react.mdx [hunk 7]
```diff
@@ -114,6 +115,75 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+}
+```
+
+The `experimental_withState()` function will match the action's types with React's expectations and preserve metadata used for progressive enhancement, allowing it to work even when JavaScript is disabled on the user's device.
+
+### `experimental_getActionState()`
+
+<p>
+
+**Type:** `(context: APIContext) => Promise<T>`<br />
```
ArmandPhilippot: When hovering, I see `ActionAPIContext` and it seems this is [a subset of the full context object](https://github.com/withastro/astro/blob/3c14936b4046062227925ec26c46689418cc3a22/packages/astro/src/actions/runtime/utils.ts#L25).
```suggestion
**Type:** `(context: ActionAPIContext) => Promise<T>`<br />
```
Hmm... okay I'm confused now 😅 I was checking if we had a link for the sentence below and now I read:
> Actions expose the full APIContext object
In: https://docs.astro.build/en/guides/actions/#authorize-users-from-an-action-handler

It seems this is the almost the full context object, we only omit `getActionResult` and `callAction`. Which makes sense I think.

src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +115,75 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+<Since v="4.3.2" />
+</p>
+
+You can access the state stored by `useActionState()` on the server in your action `handler` with `experimental_getActionState()`. It accepts the Astro API context, and optionally, you can apply a type to the result.
```
ArmandPhilippot: nit: maybe with a link? and we might not need to precise this is a subset as explained in my previous suggestion?
```suggestion
You can access the state stored by `useActionState()` on the server in your action `handler` with `experimental_getActionState()`. It accepts the [Astro API context](/en/reference/api-reference/#the-context-object), and optionally, you can apply a type to the result. 
```

src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +115,75 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+
+export const server = {
+  like: defineAction({
+    input: z.object({
+      postId: z.string(),
+    }),
+    handler: async ({ postId }, ctx) => {
+      const currentLikes = await experimental_getActionState<SafeResult<any, number>>(ctx);
+      // write to database
+      return currentLikes + 1;
```
ArmandPhilippot: `getActionState` return the full initial state (an object with both `data` and `errors`). So, here `currentLikes` is not a count but the state object.
In the changeset, I suggested `return (currentLikes.data ?? 0) + 1;` without considering showing how to handle errors because I think a minimal example was enough. But, maybe here we should?
```suggestion
      const { data: currentLikes = 0, error } = await experimental_getActionState<SafeResult<any, number>>(ctx);
      
      // handle errors
      if (error) throw(error);

      // write to database
      return currentLikes + 1;
```
What do you think?

src/content/docs/en/guides/integrations-guide/react.mdx
yanthomasdev: yeah, that's a bit confusing. I guess we should fix that page to something like:
> Actions expose a subset of the APIContext object

yanthomasdev: Thinking about it, I feel like we are fine keeping "the full APIContext" here and in that page, so people don't have to stress about what is and isn't available, when the only things omitted are the ones you *can't* to avoid nasty infinite loops and stuff.

I'll let @sarah11918 give her call before doing anything

src/content/docs/en/guides/integrations-guide/react.mdx
yanthomasdev: LGTM, totally forgot to fix that part

src/content/docs/en/guides/integrations-guide/react.mdx
sarah11918: I think it's fine to say subset, and accurate!

yanthomasdev: did you mean to comment this in the suggestion above instead, @sarah11918? 😆

sarah11918: You know it, @yanthomasdev ! (I had one job.)

sarah11918:
Let me know when it's time for me to look over after Yan/Armand have negotiated everything!

yanthomasdev:
I think we are in good shape for you to look @sarah11918. I don't have strong feelings regarding saying (full/subset) API context; either is fine.

src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +115,79 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+export const server = {
+  like: defineAction({
+    input: z.object({
+      postId: z.string(),
+    }),
+    handler: async ({ postId }, ctx) => {
+      const { data: currentLikes = 0, error } = await experimental_getActionState<SafeResult<any, number>>(ctx);
+      
+      // handle errors
+      if (error) throw(error);
```
ArmandPhilippot: I was hesitating between `console.error()` and `throw` and so... I left the parenthesis. 😅 
```suggestion
      if (error) throw error;
```

src/content/docs/en/guides/integrations-guide/react.mdx
```diff
@@ -114,6 +115,79 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+
+<p>
+
+**Type:** `(action: FormFn<T>) => (state: T, formData: FormData) => FormFn<T>`<br />
+<Since v="4.3.2" />
+</p>
+
+You can pass `experimental_withState()` and the action you want to trigger to React's `useActionState()` hook as the form action function. The example below passes a `like` action to increase a counter along with an initial state of `0` likes.
+
+```jsx title="Like.tsx" ins={2,7}
```
sarah11918:
``````suggestion
```jsx title="Like.tsx" ins={2,7} "useActionState"
``````
What do we think about this? Might also help emphasize how it's used!

sarah11918 (approved):
Looks great, thank you both Yan and Armand!

I left one (maybe good? maybe not?) suggestion for you to decide, and Armand has a fix to be committed, but then this is good to go! 🥳

PR #12325, opened 2025-09-16
Clarify that redirects can only be done at page level

mauriciabad:
#### Description (required)

Improving the documentation regarding redirects. It wasn't clear that they must be done at page level https://github.com/withastro/astro/issues/9188#issuecomment-1826146480

Also made this PR to improve the documentation on the error code that happens when a redirect is not done at page level: 
https://github.com/withastro/astro/pull/14391

### DIFF
--- a/src/content/docs/en/guides/routing.mdx
+++ b/src/content/docs/en/guides/routing.mdx
@@ -318,6 +318,8 @@ if (!isLoggedIn(cookie)) {
 ---
 ```
 
+Due to [HTML streaming](https://docs.astro.build/en/guides/on-demand-rendering/#html-streaming), redirects must be done at the page level, not inside components.
+
 ## Rewrites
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/routing.mdx
```diff
@@ -318,6 +318,8 @@ if (!isLoggedIn(cookie)) {
 ---
 ```
 
+Due to [HTML streaming](https://docs.astro.build/en/guides/on-demand-rendering/#html-streaming), redirects must be done at the page level, not inside components.
```
sarah11918:
```suggestion
Because Astro uses [HTML streaming](/en/guides/on-demand-rendering/#html-streaming) in on-demand rendering, redirects must be done at the page level, not inside child components.
```
Since pages *are* components, I think this might be more explicit?

Also fixes the link issue!

PR #12335, opened 2025-09-18
Clarify remote Markdown support in Astro

paul-sachs:
#### Description

Only built in loaders (file and glob) support built in markdown parsing. Since content collections now allow defining remote collections, this link is misleading.


sarah11918:
Thank you for this PR @paul-sachs ! We've been tracking down the specific support for Markdown in content collections, and I believe this phrase was added when we introduced [`renderMarkdown`](https://docs.astro.build/en/reference/content-loader-reference/#rendermarkdown) in the Content Loader API... but I think our original framing of built-in isn't really helpful here:

Yes, `renderMarkdown` is "built in" to Astro, but so that you can build your own custom loader, and that doesn't exactly feel super built-in to readers, I think! 😅 

So I think this phrasing needs a rewrite itself to discuss what is available/supported re: remote Markdown, and I think we should do away with saying "built in" and just be more explicit about the possibilities.

Would you like to try writing something here that expresses more clearly that there is a parser for remote Markdown to help you build a custom loader for content collections, but otherwise you will need to handle parsing remote Markdown yourself? We can workshop it together! Do you have any thoughts on making this more helpful?

paul-sachs:
@sarah11918 I can give it a try when I get some time though it'll take me some time to figure out exactly what we should say here.

I think my primary concern was making it obvious that the `renderMarkdown` function provided to custom content loaders does not support using the default markdown settings (using Astro.config markdown settings.) I'm actually not sure if that section really belongs here so much as perhaps an example showing how we might mimic that behaviour (as described in https://github.com/withastro/roadmap/discussions/1094)

sarah11918:
Thanks @paul-sachs and for sure did not mean to put this on you! It may very well be that removing the link as you initially proposed is the right move here. Just wanted to check to see whether you had thoughts on this! We'll be working on this one too 😄

src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -397,7 +397,7 @@ When using the frontmatter `layout` property, you must include the `<meta charse
 
 ## Fetching Remote Markdown
 
-Astro does not include built-in support for remote Markdown outside of [content collections](/en/guides/content-collections/).
+Astro does not include built-in support for remote Markdown.
 
 To fetch remote Markdown directly and render it to HTML, you will need to install and configure your own Markdown parser from NPM. This will not inherit from any of Astro's built-in Markdown settings that you have configured.
```
sarah11918:
```suggestion
Astro's internal Markdown processor is not available for processing remote Markdown.

To fetch remote Markdown for use in [content collections](/en/guides/content-collections), you can [build a custom loader](/en/guides/content-collections/#building-a-custom-loader) with access to a [`renderMarkdown()` function](/en/reference/content-loader-reference/#rendermarkdown).

To fetch remote Markdown directly and render it to HTML, you will need to install and configure your own Markdown parser from NPM. This will not inherit from any of Astro's built-in Markdown settings that you have configured.
```
OK, might have gone link-happy here, but what do we think about something like this @paul-sachs ? (Until such time as we actually do expose the internal Markdown parser?)

paul-sachs: Yeah this looks much better. I think the links will definitely help anyone else who runs into this rough corner.

src/content/docs/en/guides/markdown-content.mdx
```diff
@@ -397,7 +397,9 @@ When using the frontmatter `layout` property, you must include the `<meta charse
 
 ## Fetching Remote Markdown
 
-Astro does not include built-in support for remote Markdown outside of [content collections](/en/guides/content-collections/).
+Astro's internal Markdown processor is not available for processing remote Markdown.
+
+To fetch remote Markdown for use in [content collections](/en/guides/content-collections), you can [build a custom loader](/en/guides/content-collections/#building-a-custom-loader) with access to a [`renderMarkdown()` function](/en/reference/content-loader-reference/#rendermarkdown).
```
sarah11918:
```suggestion
To fetch remote Markdown for use in [content collections](/en/guides/content-collections/), you can [build a custom loader](/en/guides/content-collections/#building-a-custom-loader) with access to a [`renderMarkdown()` function](/en/reference/content-loader-reference/#rendermarkdown).
```

sarah11918:
OK, was waiting to see whether anyone else wanted to chime in here, but if you're happy and I'm happy, Paul, then let's just do this! Thanks again for raising this issue, and welcome to Team Docs 🥳

PR #12353, opened 2025-09-19
create v6 upgrade guide pagev6

florian-lefebvre:

- Closes #
- Suggested label:


sarah11918:
Changed the title to remove `i18n` so it's not filtered out of my view 😅  (plus, also more specific!)

src/content/docs/en/guides/internationalization.mdx
```diff
@@ -176,9 +176,9 @@ Set this option when all routes will have their `/locale/` prefix in their URL a
 
 Configures whether or not the home URL (`/`) generated by `src/pages/index.astro` will redirect to `/<defaultLocale>`.
```
sarah11918: So, lots of thoughts here:

1. This is the *guide* not the reference, so I don't think we have to adhere to option-based section headings and can focus more on use-cases
2. I *think* (from reading the other changes here) that this option is only used if `prefixDefaultLocale` is true, but yet you want to escape from that **for the home URL only**.

So, what if this entire section became focused on "opting out of redirects for the home URL" as a subheading of `prefixDefaultLocale`, something closer to:

- even if you *do* want default locale prefixed (e.g. all your routes to be `/en/page-1` when your default language is English, it is possible to opt-out of that behaviour for only the index page. This is useful when you want to have a site home that exists outside of your configured locale structure
- When `prefixDefaultLocale: true` is set, you can optionally configure `redirectToDefaultLocale` to control the redirect behaviour of `src/pages/index.astro` (e.g. whether or not it should redirect to `src/pages/en/index.astro`). This allows all your localized routes to be prefixed **except the home URL of your site**.

I already fear that this page feels too referency, so I think something like this will help us think of it more as a guide. What do you think about trying to shape it more like this?

florian-lefebvre: I'm not against it but isn't it a bit out of scope of this PR? I feel like the improvement could be done on main, then backported here

sarah11918: So, write it backwards, then update it here? 

Just thinking through how this works currently (and the fact that that's difficult to do probably is why we're changing it 😅 )

I think the story actually changes to be a more sensible one with this default, so it probably *is* in the scope of this PR. (I don't think the same story works as well backwards?)

So, I would say let's do that here, rather than break our brains explaining something that might cause an infinite loop anyway?

src/content/docs/en/guides/internationalization.mdx
```diff
@@ -170,15 +170,13 @@ Set this option when all routes will have their `/locale/` prefix in their URL a
 
 - URLs without a locale prefix, (e.g. `example.com/about/`) will return a 404 (not found) status code unless you specify a [fallback strategy](#fallback).
 
-### `redirectToDefaultLocale`
+#### Opting out of redirects for the home URL
 
 <p><Since v="4.2.0" /></p>
```
sarah11918:
```suggestion
```
Since this isn't the API reference for this, we don't need the Since!

src/content/docs/en/guides/internationalization.mdx
```diff
@@ -170,15 +170,13 @@ Set this option when all routes will have their `/locale/` prefix in their URL a
...
 
-Configures whether or not the home URL (`/`) generated by `src/pages/index.astro` will redirect to `/<defaultLocale>`.
+Even if you *do* want default locale prefixed (e.g. all your routes to be `/en/page-1` when your default language is English), it is possible to opt-out of that behaviour for only the index page. This is useful when you want to have a site home that exists outside of your configured locale structure
```
sarah11918:
```suggestion
Even with your default locale routes prefixed, it is possible to opt-out of that behaviour for only your site's index page. This is useful when you want to have a home page that exists outside of your configured locale structure.
```

src/content/docs/en/guides/internationalization.mdx
```diff
@@ -170,15 +170,11 @@ Set this option when all routes will have their `/locale/` prefix in their URL a
 
 - URLs without a locale prefix, (e.g. `example.com/about/`) will return a 404 (not found) status code unless you specify a [fallback strategy](#fallback).
 
-### `redirectToDefaultLocale`
+#### Opting out of redirects for the home URL
 
-<p><Since v="4.2.0" /></p>
+Even with your default locale routes prefixed, it is possible to opt-out of that behaviour for only your site's index page. This is useful when you want to have a home page that exists outside of your configured locale structure.
```
sarah11918:
```suggestion
Even with your default locale routes prefixed, this behaviour does not apply by default to your site's index page. This allows you to have a home page that exists outside of your configured locale structure, where all of your localized routes are prefixed except the home URL of your site.
```
Default is this now, right?

florian-lefebvre: Yeah I think that's right

src/content/docs/en/guides/internationalization.mdx
```diff
@@ -170,15 +170,11 @@ Set this option when all routes will have their `/locale/` prefix in their URL a
...
 
-Configures whether or not the home URL (`/`) generated by `src/pages/index.astro` will redirect to `/<defaultLocale>`.
-
-Setting `prefixDefaultLocale: true` will also automatically set `redirectToDefaultLocale: true` in your `routing` config object. By default, the required `src/pages/index.astro` file will automatically redirect to the index page of your default locale.
-
-You can opt out of this behavior by [setting `redirectToDefaultLocale: false`](/en/reference/configuration-reference/#i18nroutingredirecttodefaultlocale). This allows you to have a site home page that exists outside of your configured locale folder structure.
+When `prefixDefaultLocale: true` is set, you can optionally configure `redirectToDefaultLocale` to control the redirect behaviour of `src/pages/index.astro` (e.g. whether or not it should redirect to `src/pages/en/index.astro`). This allows all your localized routes to be prefixed **except the home URL of your site**.
```
sarah11918:
```suggestion
You can opt out of this behavior so that your main site URL will also redirect to a prefixed, localized route for your default locale. When `prefixDefaultLocale: true` is set, you can additionally configure `redirectToDefaultLocale: true`. This will ensure that `src/pages/index.astro` will redirect to `src/pages/[defaultLocale]/index.astro`.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -108,6 +108,31 @@ Some default behavior has changed in Astro v5.0 and your project code may need u
 
 In most cases, the only action needed is to review your existing project's deployment and ensure that it continues to function as you expect, making updates to your code as necessary. In some cases, there may be a configuration setting to allow you to continue to use the previous default behavior.
 
+### Changed: `i18n.routing.redirectToDefaultLocale` default value
+
+In Astro v5.0, `i18n.routing.redirectToDefaultLocale` default value as `true` and `i18n.routing.prefixDefaultLocale` default value was `false`. However, the resulting redirects could cause infinite loops.
```
```suggestion
In Astro v5.0, the `i18n.routing.redirectToDefaultLocale` default value was `true` and the `i18n.routing.prefixDefaultLocale` default value was `false`. However, the resulting redirects could cause infinite loops.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -108,6 +108,31 @@ Some default behavior has changed in Astro v5.0 and your project code may need u
...
+
+In Astro v6.0, `i18n.routing.redirectToDefaultLocale` now defaults to `false`. It can now only be used if `i18n.routing.prefixDefaultLocale` is set to `true`.
```
```suggestion
In Astro v6.0, `i18n.routing.redirectToDefaultLocale` now defaults to `false`. Additionally, it can now only be used if `i18n.routing.prefixDefaultLocale` is set to `true`.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -108,6 +108,31 @@ Some default behavior has changed in Astro v5.0 and your project code may need u
...
+
+#### What should I do?
+
+To get back to your existing behavior, review your Astro config to make sure properties have compatible values:
```
```suggestion
Review your Astro `i18n` config as you may now need to explicitly set values for `redirectToDefaultLocale` and `prefixDefaultLocale` to recreate your project's previous behavior.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -108,6 +108,31 @@ Some default behavior has changed in Astro v5.0 and your project code may need u
 
 In most cases, the only action needed is to review your existing project's deployment and ensure that it continues to function as you expect, making updates to your code as necessary. In some cases, there may be a configuration setting to allow you to continue to use the previous default behavior.
 
+### Changed: `i18n.routing.redirectToDefaultLocale` default value
+
+In Astro v5.0, the `i18n.routing.redirectToDefaultLocale` default value was `true` and the `i18n.routing.prefixDefaultLocale` default value was `false`. However, the resulting redirects could cause infinite loops.
```
```suggestion
In Astro v5.0, the `i18n.routing.redirectToDefaultLocale` default value was `true`. When combined with the `i18n.routing.prefixDefaultLocale` default value of `false`, the resulting redirects could cause infinite loops.
```
I think this better explains why we're talking about `prefixDefaultLocale` here.

src/content/docs/en/guides/internationalization.mdx
```diff
@@ -170,15 +170,11 @@ Set this option when all routes will have their `/locale/` prefix in their URL a
...
 
-Configures whether or not the home URL (`/`) generated by `src/pages/index.astro` will redirect to `/<defaultLocale>`.
-
-Setting `prefixDefaultLocale: true` will also automatically set `redirectToDefaultLocale: true` in your `routing` config object. By default, the required `src/pages/index.astro` file will automatically redirect to the index page of your default locale.
-
-You can opt out of this behavior by [setting `redirectToDefaultLocale: false`](/en/reference/configuration-reference/#i18nroutingredirecttodefaultlocale). This allows you to have a site home page that exists outside of your configured locale folder structure.
+You can opt out of this behavior so that your main site URL will also redirect to a prefixed, localized route for your default locale. When `prefixDefaultLocale: true` is set, you can additionally configure `redirectToDefaultLocale: true`. This will ensure that `src/pages/index.astro` will redirect to `src/pages/[defaultLocale]/index.astro`.
```
sarah11918:
```suggestion
You can opt out of this behavior so that your main site URL will also redirect to a prefixed, localized route for your default locale. When `prefixDefaultLocale: true` is set, you can additionally configure `redirectToDefaultLocale: true`. This will ensure that the home URL (`/`) generated by `src/pages/index.astro` will redirect to `/[defaultLocale]/`.
```
I realized that talking about the URL is probably the more important thing here!

florian-lefebvre:
Are we good to go?

PR #12358, opened 2025-09-20
Fix TSC5090 TypeScript compiler error

mondeja:
#### Description (required)

It seems that TypeScript added at some point a check to forbid non-explicit relative `paths` when `baseUrl` is not set. As `astro/tsconfigs/strict` does not define `baseUrl`, this can be reproduced in new astro projects following the Get Started guide and in the Astro docs itself.

To reproduce this problem:

1. Create an example project with `npm create astro@latest`.
2. `cd <project-name>`
3. Add the changes proposed by the [_Import Aliases_ section of the TypeScript setup guide](https://docs.astro.build/en/guides/typescript/#import-aliases) to _tsconfig.json_ file. 
4. Run `tsc`. The error is displayed:

```
$ tsc 
tsconfig.json:7:25 - error TS5090: Non-relative paths are not allowed when 'baseUrl' is not set. Did you forget a leading './'?
7       "@components/*": ["src/components/*"],
                          ~~~~~~~~~~~~~~~~~~

tsconfig.json:8:22 - error TS5090: Non-relative paths are not allowed when 'baseUrl' is not set. Did you forget a leading './'?
8       "@layouts/*": ["src/layouts/*"]
                       ~~~~~~~~~~~~~~~

Found 2 errors in the same file, starting at: tsconfig.json:7
```

Apply the fix proposed by this PR and the error disappears.

What I did find, which documents the historical behavior and discussions around it (and shows where the restriction comes from), are these relevant points:

- The historical decision that `paths` requires `baseUrl` is in the original discussion about path mappings in the repository (this entry explains the semantics and explicitly states that `paths` can only be used if `baseUrl` is defined):  
  https://github.com/microsoft/TypeScript/issues/5039

- Later proposals and discussions about allowing `paths` without `baseUrl` (showing that the requirement has been deliberate and debated since 2019):  
  https://github.com/microsoft/TypeScript/issues/31869

- Examples of issues/errors in ecosystems (Nuxt, Svelte, etc.) that report exactly this message in tooling around 2022–2024 (proving that the diagnostic appears in modern versions; e.g. a Nuxt issue showing the message with TypeScript v5.2.2):  
  https://github.com/nuxt/nuxt/issues/23157

mondeja.

### DIFF
--- a/src/content/docs/en/guides/authentication.mdx
+++ b/src/content/docs/en/guides/authentication.mdx
@@ -69,8 +69,9 @@ To install `auth-astro` manually, install the required package for your package 
 
 Then, apply the integration to your `astro.config.*` file using the `integrations` property:
 
-```ts title="astro.config.mjs" ins={2,7}
+```ts title="astro.config.mjs" ins={3,8}
 import { defineConfig } from 'astro/config';
+import netlify from '@astrojs/netlify';
 import auth from 'auth-astro';
 
 export default defineConfig({
--- a/src/content/docs/es/guides/authentication.mdx
+++ b/src/content/docs/es/guides/authentication.mdx
@@ -4,12 +4,14 @@ description: Una introducción a la autenticación en Astro
 i18nReady: true
 ---
 
+import { Steps } from '@astrojs/starlight/components'
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
+import UIFrameworkTabs from '~/components/tabs/UIFrameworkTabs.astro'
 import ReadMore from '~/components/ReadMore.astro'
 
 La autenticación y la autorización son dos procesos de seguridad que manejan el acceso a tu sitio web o aplicación. La autenticación verifica la identidad de un visitante, mientras que la autorización otorga acceso a áreas y recursos protegidos.
 
-La autenticación te permite personalizar áreas de tu sitio para personas que han iniciado sesión y proporciona la mayor protección para la información personal o privada. Las bibliotecas de autenticación (por ejemplo, [Lucia Auth](https://lucia-auth.com/), [Auth.js](https://authjs.dev/), [Clerk](https://clerk.com)) proporcionan utilidades para múltiples métodos de autenticación, como el inicio de sesión con correo electrónico y los proveedores de OAuth.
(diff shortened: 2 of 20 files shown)


ArmandPhilippot:
Thanks you for taking the time to open a PR to fix an issue in docs! But, I tried to reproduce the issue following your steps and using `"@components/*": ["src/components/*"]` works fine on my side. Unless you're able to provide a reproduction where this fails, I don't think we need those changes.

`baseUrl` was dropped in https://github.com/withastro/astro/pull/14064 because this is no longer required with paths aliases. And looking at the tests fixtures implemented in that PR, this seems to match what we described in the docs:
https://github.com/jp-knj/astro/blob/1a448231dc68d48b8b0532fb84f05f9dd13bfe4d/packages/astro/test/fixtures/alias-tsconfig-no-baseurl/tsconfig.json#L1-L7

mondeja:
Hi @ArmandPhilippot. Thanks for your initial review.

> Unless you're able to provide a reproduction where this fails, I don't think we need those changes.

I'm able to reproduce it on GitHub actions.

- Repository: https://github.com/mondeja/astro-docs-12358
- Run: https://github.com/mondeja/astro-docs-12358/actions/runs/17896035697/job/50882792680#step:6:7
- Workflow file: https://github.com/mondeja/astro-docs-12358/blob/e99012bd2c71b4be1a6da90e9336ae4b729f4dab/.github/workflows/ci.yml

Commands:

```
npm create astro@latest ./my-project -- --template minimal --yes
cd my-project
# tsc (OK)
tsc
# Edit tsconfig.json
jq '.compilerOptions.paths = {"@components/*": ["src/components/*"]}' tsconfig.json > tsconfig.tmp && mv tsconfig.tmp tsconfig.json
# Show new tsconfig.json content
cat tsconfig.json
# tsc (fails)
tsc
# Error: tsconfig.json(13,9): error TS5090: Non-relative paths are not allowed when 'baseUrl' is not set. Did you forget a leading './'?
echo $?
# 2
```

ArmandPhilippot (approved):
Oh sorry, maybe I was too quick earlier. I'm so used to run `astro check` in Astro projects that I misread the step 4. I didn't run `tsc`... I've only tried running the dev server and `astro check`. Both didn't complained and the path aliases was working just fine.

So, yes, I can see an error reported using `tsc` both with the previous steps and the new ones! I wonder why `astro check` didn't catch that and why this works (both with `dev` and `build`) without even a warning...

Looking at Typescript docs, it seems they [always use relative paths when declaring paths aliases](https://www.typescriptlang.org/tsconfig/#paths) and I can confirm this works.

So, the changes looks good to me. Thank you both for fixing the docs and your patience! Welcome to team Docs! 🎉

PR #12405, opened 2025-09-23
fix(configuration-reference): remove page slug from internal anchor links

ArmandPhilippot:
#### Description (required)

Fixes the `docgen` script to remove `/en/reference/configuration-reference/` from internal anchor links.

The changes was introduced by https://github.com/withastro/astro/pull/14411 because relative links do not work in editors.
Now, instead of `[redirects](#redirects)` we have `[redirects](/en/reference/configuration-reference/#redirects)` (for example).

Both works, but maybe we should clean the links as we do to make them relative instead of absolute?

(I don't have a strong opinion on this, it just seemed weird while I was updating the French translation. So feel free to tell me to close this one 😄 )

- Suggested label: `site improvement`?


sarah11918:
Thank you, Armand! 

OK, so as I see it what happened was:

- `[redirects](#redirects)` in docgen produced valid links of the form `[redirects](#redirects)` in our configuration reference page. (yay!)
- however, `[redirects](#redirects)` in docgen meant that editor hover links in the editor didn't work (boo!)
- So, we changed the docgen source to read `[redirects](/en/reference/configuration-reference/#redirects)` which shows editor hover links (yay!) and produces valid links in our configuration reference page (yay!)
- however, now the links in the configuration source page are the longer `[redirects](/en/reference/configuration-reference/#redirects)` format (boo!)

So this PR would allow us to type `[redirects](/en/reference/configuration-reference/#redirects)` in the docgen source, but the resulting code imported into the docs repo for generating the page itself would be `[redirects](#redirects)`. Did I get that summary correct?

I will let @delucis review this one when he returns! 😅

ArmandPhilippot:
The formatting is broken, but after checking with "Edit": this is exactly that! 😅

trueberryless:
Sarah, we need more yay!s:

> So this PR would allow us to type `[redirects](/en/reference/configuration-reference/#redirects)` in the docgen source, but the resulting code imported into the docs repo for generating the page itself would be `[redirects](#redirects)`.

(yay!)(yay!)

delucis (commented):
Code change LGTM! Did you try running this locally just to double check it works @ArmandPhilippot?

ArmandPhilippot:
@delucis Yes! I checked again and didn't spotted any unwanted changes or oversights!

delucis (approved):
Perfect, then I approve! Thanks @ArmandPhilippot 🎉

PR #12429, opened 2025-09-26
i18n(ko-KR): update `fonts.mdx`

jsparkdev:
#### Description (required)

- #12413

- Closes #
- Suggested label:


ArmandPhilippot:
Hi! I noticed #12326 didn't triggered a status change while updating the French version. It seems the Korean version is also using the wrong RFC link. :wink:

PR #12467, opened 2025-09-30
[v6] trailing slash never for endpoints with file extension

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/endpoints.mdx
+++ b/src/content/docs/en/guides/endpoints.mdx
@@ -47,6 +47,8 @@ import type { APIRoute } from "astro";
 export const GET: APIRoute = async ({ params, request }) => {...}
 ```
 
+Note that these endpoints can only be accessed without a trailing slash, regardless of your [`build.trailingSlash`](/en/reference/configuration-reference/#trailingslash) configuration.
+
 ### `params` and Dynamic routing
 
 Endpoints support the same [dynamic routing](/en/guides/routing/#dynamic-routes) features that pages do. Name your file with a bracketed parameter name and export a [`getStaticPaths()` function](/en/reference/routing-reference/#getstaticpaths). Then, you can access the parameter using the `params` property passed to the endpoint function:
--- a/src/content/docs/en/guides/endpoints.mdx
+++ b/src/content/docs/en/guides/endpoints.mdx
@@ -47,7 +47,7 @@ import type { APIRoute } from "astro";
 export const GET: APIRoute = async ({ params, request }) => {...}
 ```
 
-Note that these endpoints can only be accessed without a trailing slash, regardless of your [`build.trailingSlash`](/en/reference/configuration-reference/#trailingslash) configuration.
+Note that endpoints whose URLs include a file extension (e.g. `src/pages/sitemap.xml.ts`) can only be accessed without a trailing slash (e.g. `/sitemap.xml`), regardless of your [`build.trailingSlash`](/en/reference/configuration-reference/#trailingslash) configuration.
 
 ### `params` and Dynamic routing
 
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -89,6 +89,8 @@ Check that both your development environment and your deployment environment are
 
 ### Vite 7.0
 
+<SourcePR number="14445" title="feat: update vite"/>
+
 Astro v6.0 upgrades to Vite v7.0 as the development server and production bundler.
 
 #### What should I do?
@@ -287,6 +289,8 @@ In most cases, the only action needed is to review your existing project's deplo
 
 ### Changed: `i18n.routing.redirectToDefaultLocale` default value
 
+<SourcePR number="14406" title="feat(astro)!: update i18n.redirectToDefaultLocale default"/>
+
 In Astro v5.0, the `i18n.routing.redirectToDefaultLocale` default value was `true`. When combined with the `i18n.routing.prefixDefaultLocale` default value of `false`, the resulting redirects could cause infinite loops.
 
 In Astro v6.0, `i18n.routing.redirectToDefaultLocale` now defaults to `false`. Additionally, it can now only be used if `i18n.routing.prefixDefaultLocale` is set to `true`.
@@ -309,6 +313,26 @@ export default defineConfig({
 ```
 
 <ReadMore>Learn more about [Internationalization routing](/en/guides/internationalization/#routing).</ReadMore>
+value
+
+### Changed: endpoints with a file extension cannot be accessed with a trailing slash
+
+<SourcePR number="14457" title="feat!: trailing slash never for endpoints with file extension"/>
+
+In Astro v5.0, endpoints with a pathname ending with a file extension (eg. `/src/pages/sitemap.xml.ts`) could be accessed with a trailing slash (`/sitemap.xml/`) or without (`/sitemap.xml`), no matter the value of `build.trailingSlash`.
+
+In Astro v6.0, these endpoints can only be accessed without a trailing slash.
+
+#### What should I do?
+
+Review your links and remove any referencing an endpoint ending witha file a file extension and a trailing slash:
+
+```html del={1} ins={2} title="src/pages/index.astro"
+<a href="/sitemap.xml/">Sitemap</a>
+<a href="/sitemap.xml">Sitemap</a>
+```
+
+<ReadMore>Learn more about [`build.trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash).</ReadMore>
 
 ## Breaking Changes
 
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -313,7 +313,6 @@ export default defineConfig({
 ```
 
 <ReadMore>Learn more about [Internationalization routing](/en/guides/internationalization/#routing).</ReadMore>
-value
 
 ### Changed: endpoints with a file extension cannot be accessed with a trailing slash
 
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -331,7 +331,7 @@ Review your links and remove any referencing an endpoint ending with a file exte
 <a href="/sitemap.xml">Sitemap</a>
 ```
 
-<ReadMore>Learn more about [`build.trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash).</ReadMore>
+<ReadMore>Learn more about [`build.trailingSlash`](/en/reference/configuration-reference/#trailingslash).</ReadMore>
 
 ## Breaking Changes
 
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -324,7 +324,7 @@ In Astro v6.0, these endpoints can only be accessed without a trailing slash. Th
 
 #### What should I do?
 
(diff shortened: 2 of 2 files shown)


src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -309,6 +313,26 @@ export default defineConfig({
...
+
+<SourcePR number="14457" title="feat!: trailing slash never for endpoints with file extension"/>
+
+In Astro v5.0, endpoints with a pathname ending with a file extension (eg. `/src/pages/sitemap.xml.ts`) could be accessed with a trailing slash (`/sitemap.xml/`) or without (`/sitemap.xml`), no matter the value of `build.trailingSlash`.
+
+In Astro v6.0, these endpoints can only be accessed without a trailing slash.
+
+#### What should I do?
+
+Review your links and remove any referencing an endpoint ending witha file a file extension and a trailing slash:
```
ascorbic:
```suggestion
Review your links and remove any referencing an endpoint ending with a file extension and a trailing slash:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -309,6 +313,26 @@ export default defineConfig({
 ```
 
 <ReadMore>Learn more about [Internationalization routing](/en/guides/internationalization/#routing).</ReadMore>
+value
```
ascorbic: Is this a mistake?

florian-lefebvre: Yes! Not sure how it got there

```suggestion
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -310,6 +314,25 @@ export default defineConfig({
...
+#### What should I do?
+
+Review your links and remove any referencing an endpoint ending with a file extension and a trailing slash:
+
+```html del={1} ins={2} title="src/pages/index.astro"
+<a href="/sitemap.xml/">Sitemap</a>
+<a href="/sitemap.xml">Sitemap</a>
+```
+
+<ReadMore>Learn more about [`build.trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash).</ReadMore>
```
jsparkdev:
```suggestion
<ReadMore>Learn more about [`build.trailingSlash`](/en/reference/configuration-reference/#trailingslash).</ReadMore>
```

jsparkdev (changes requested):

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -310,6 +314,25 @@ export default defineConfig({
 
 <ReadMore>Learn more about [Internationalization routing](/en/guides/internationalization/#routing).</ReadMore>
 
+### Changed: endpoints with a file extension cannot be accessed with a trailing slash
+
+<SourcePR number="14457" title="feat!: trailing slash never for endpoints with file extension"/>
+
+In Astro v5.0, endpoints with a pathname ending with a file extension (eg. `/src/pages/sitemap.xml.ts`) could be accessed with a trailing slash (`/sitemap.xml/`) or without (`/sitemap.xml`), no matter the value of `build.trailingSlash`.
```
```suggestion
In Astro v5.0, custom endpoints whose URL ended in a file extension (e.g. `/src/pages/sitemap.xml.ts` ) could be accessed with a trailing slash (`/sitemap.xml/`) or without (`/sitemap.xml`), regardless of the value configured for `build.trailingSlash`.
```

UPDATED: does this make sense @florian-lefebvre

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -310,6 +314,25 @@ export default defineConfig({
...
+
+In Astro v6.0, these endpoints can only be accessed without a trailing slash.
```
```suggestion
In Astro v6.0, these endpoints can only be accessed without a trailing slash. This is true regardless of your `build.trailingSlash` configuration.
```
I'm assuming? If so we should say it!

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -310,6 +314,25 @@ export default defineConfig({
...
+
+#### What should I do?
+
+Review your links and remove any referencing an endpoint ending with a file extension and a trailing slash:
```
```suggestion
Review your links to your custom endpoints that include a file extension in the URL and remove any trailing slashes:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -310,6 +314,25 @@ export default defineConfig({
...
+
+```html del={1} ins={2} title="src/pages/index.astro"
+<a href="/sitemap.xml/">Sitemap</a>
+<a href="/sitemap.xml">Sitemap</a>
+```
+
+<ReadMore>Learn more about [`build.trailingSlash`](/en/reference/configuration-reference/#trailingslash).</ReadMore>
```
```suggestion
<ReadMore>Learn more about [custom endpoints](/en/guides/endpoints/).</ReadMore>
```
I think taking people to endpoints is more helpful here, since the config actually doesn't matter for this particular thing!

Also, I think we should say something *on* the endpoint guide about what the resulting URL will look like, and have a statement that regardless of `build.trailingSlash`, these URLs will not be accessible with a trailing slash.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -310,6 +314,25 @@ export default defineConfig({
 
 <ReadMore>Learn more about [Internationalization routing](/en/guides/internationalization/#routing).</ReadMore>
 
+### Changed: endpoints with a file extension cannot be accessed with a trailing slash
```
sarah11918, line 321: I *kind* of feel like this is maybe less a "changed default value" and more of a generic breaking change? And, I might phrase it as 

```suggestion
### Changed: Endpoints no longer produce URLs with trailing slashes
```

sarah11918 (commented):
Thanks Florian! Left some comments below to see what you think!

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: Unfortunately we have to be specific. This breaking change only applies to endpoints whose pattern contains a file extension, eg. `/sitemap.xml` and not `/api/submit`

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: As said in the review above, by file extension I mean within the resulting pattern, not the filename. Sure `sitemap.xml.ts` and `api/submit.ts` both have a file extension but it's not the case of their pattern

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: OK, gotcha, let me look over this again!

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: Gotcha, it wasn't clear to me that the file extension would be IN THE URL! Let me look at this again!

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: I might still put this in "breaking changes" vs "changed defaults"? But agreed, I think the original title is maybe as short as we can get it given the extra clarification!

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: Updated this one too. I think that one works now?

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: Still think this stands! It's an endpoint behavior, and probably worth noting on the endpoints page!

florian-lefebvre:
@sarah11918 just checked the docs and we call these routes "static file endpoints" if that helps!

sarah11918 (commented):
ugh... pending reviews strike again!

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: I would go further by linking to https://docs.astro.build/en/guides/endpoints/#static-file-endpoints, wdyt?

sarah11918:
> we call these routes "static file endpoints" if that helps!

Actually, I'm not entirely sure we do? I was rereading that page and it *looks* like we have two separate static / SSR route categories, but... the static one defines what these routes are, and the SSR one only describes extra functionality if the route is rendered on demand? (Though one could easily just jump to the later section for on-demand rendered routes, and not realize that we've explained what these are in the "static" section.) So, probably a case of adding SSR on after the fact and not getting the info hierarchy *quite* right there, tbh :sweat_smile:

florian-lefebvre:
That's how I understood this part:

> The .js or .ts extension will be removed during the build process, so the name of the file should include the extension of the data you want to create. For example, src/pages/data.json.ts will build a /data.json endpoint.

sarah11918:
Furthermore, I don't think this page shows *any* examples of an endpoint that does *not* contain a file extension before the `.js`/`.ts`... so a) maybe we should in general? and b) this extra info won't make sense unless we're VERY explicit about which ones cannot have the trailing slash (since we otherwise make no mention of differentiating these endpoints)

sarah11918:
> The .js or .ts extension will be removed during the build process, so the name of the file should include the extension of the data you want to create. For example, src/pages/data.json.ts will build a /data.json endpoint.

But, I think we *mean* for this to also be true about the Server Endpoints?

> Everything described in the static file endpoints section can also be used in SSR mode... This unlocks new features that are unavailable at build time

And no examples showing anything *without* an extension?
  
e.g. we never show how to make something like `/api/submit` ... is that kind of route only on-demand rendered?

src/content/docs/en/guides/endpoints.mdx
```diff
@@ -47,6 +47,8 @@ import type { APIRoute } from "astro";
 export const GET: APIRoute = async ({ params, request }) => {...}
 ```
 
+Note that these endpoints can only be accessed without a trailing slash, regardless of your [`build.trailingSlash`](/en/reference/configuration-reference/#trailingslash) configuration.
```
sarah11918:
```suggestion
Note that endpoints whose URLs include a file extension (e.g. `src/pages/sitemap.xml.ts`) can only be accessed without a trailing slash (e.g. `/sitemap.xml`), regardless of your [`build.trailingSlash`](/en/reference/configuration-reference/#trailingslash) configuration.
```
I think this is sufficient for the purposes of *this* PR! I don't think we can say "these" here because I don't think this page makes any distinction between routes that will contain file extensions in the URL vs not.

But at least if we say this here, it's NWTWWHB (even if we need to at some future point explain what we call the endpoints that DON'T have a file extension in the URL, and need to make it clear that these are really "built" the same way, but on-demand has more features available to it)

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: The only reason not to is that I (at least personally!) don't know that that is the proper distinction (static file endpoints) because the page doesn't otherwise discuss creating endpoints that don't have a file extension in the URL. So, that might not actually be the exact thing we are talking about?

src/content/docs/en/guides/endpoints.mdx
sarah11918: I think this one change, and we can merge this as NWTWWHB (with the other suggestion but just keeping generic endpoint link).

What do you think?

florian-lefebvre:
NWTWWHB for the win!

> we never show how to make something like /api/submit ... is that kind of route only on-demand rendered?

Yeah I don't think there's a case outside of SSR

sarah11918:
> Yeah I don't think there's a case outside of SSR

If so, I do think future works should be to explicitly show an example of that in the on-demand rendered section, but out of scope here!

PR #12469, opened 2025-09-30
[v6] clean deprecated APIs

florian-lefebvre:

#### Description (required)

I don't think docs need to be updated elsewhere because the things getting deprecated were already not documented anymore

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -279,6 +279,72 @@ const integration = () => {
 }
 ```
 
+### Removed: old `app.render()` signature (Adapter API)
+
+<SourcePR number="14462" title="feat: clean deprecated APIs"/>
+
+In Astro 4.0, the `app.render()` signature allowing to pass `routeData` and `locals` as optional arguments was deprecated in favor of a single optional `renderOptions` argument.
+
+Astro 6.0 removes this signature entirely.
+
+#### What should I do?
+
+Review your `app.render` calls and pass `routeData` and `locals` as properties of an object instead of as multiple independent arguments:
+
+```ts title="my-adapter/entrypoint.ts" del={1} ins={2}
+app.render(request, routeData, locals)
+app.render(request, { routeData, locals })
+```
+
+<ReadMore>Learn more about the [Adapter API](/en/reference/adapter-reference/).</ReadMore>
+
+### Removed: `handleForms` prop for the `<ClientRouter />` component
+
+<SourcePR number="14462" title="feat: clean deprecated APIs"/>
+
+In Astro 4.0, the `handleForms` prop of the `<ClientRouter />` component was deprecated.
+
+Astro 6.0 removes this prop entirely, as it had no impact on form submission anymore.
+
+#### What should I do?
+
+Remove the `handleForms` property from your `<ClientRouter />` component. It is no longer necessary.
+
+```astro title="src/pages/index.astro" del="handleForms"
+---
+import { ViewTransitions } from "astro:transitions";
+---
+<html>
+  <head>
+    <ViewTransitions handleForms />
+  </head>
+  <body>
+    <!-- stuff here -->
+  </body>
+</html>
+```
+
+<ReadMore>Learn more about [transitions with forms](/en/guides/view-transitions/#transitions-with-forms).</ReadMore>
+
+### Removed: `prefetch()` `with` option
+
+<SourcePR number="14462" title="feat: clean deprecated APIs"/>
+
+In Astro 4.8.4, the `with` option of the programmatic `prefetch()` function was deprecated.
+
+Astro 6.0 removes this option entirely, defaulting to `link` preloading.
+
+#### What should I do?
+
+Review your `prefetch()` calls and remove the `with` option:
+
+```ts
+prefetch('/about', { with: 'fetch' });
+prefetch('/about');
+```
+
+<ReadMore>Learn more about [prefetching](/en/guides/prefetch/).</ReadMore>
+
 ## Changed Defaults
 
 Some default behavior has changed in Astro v5.0 and your project code may need updating to account for these changes.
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,72 @@ const integration = () => {
...
+
+Remove the `handleForms` property from your `<ClientRouter />` component. It is no longer necessary.
+
+```astro title="src/pages/index.astro" del="handleForms"
+---
+import { ViewTransitions } from "astro:transitions";
+---
+<html>
+  <head>
+    <ViewTransitions handleForms />
```
ArmandPhilippot: This section is about `ClientRouter` so to avoid any confusion we should use that component rather than the removed one?
```suggestion
import { ClientRouter } from "astro:transitions";
---
<html>
  <head>
    <ClientRouter handleForms />
```

ArmandPhilippot (commented):
A small nit, otherwise this looks good. And I couldn't find any occurrences elsewhere in the docs. So yes, I think there is nothing else to update!

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: Ah yeah I copied the old diff, thanks!

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,72 @@ const integration = () => {
 }
 ```
 
+### Removed: old `app.render()` signature (Adapter API)
+
+<SourcePR number="14462" title="feat: clean deprecated APIs"/>
+
+In Astro 4.0, the `app.render()` signature allowing to pass `routeData` and `locals` as optional arguments was deprecated in favor of a single optional `renderOptions` argument.
```
```suggestion
In Astro 4.0, the `app.render()` signature that allowed passing `routeData` and `locals` as optional arguments was deprecated in favor of a single optional `renderOptions` argument.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,72 @@ const integration = () => {
...
+app.render(request, { routeData, locals })
+```
+
+<ReadMore>Learn more about the [Adapter API](/en/reference/adapter-reference/).</ReadMore>
+
+### Removed: `handleForms` prop for the `<ClientRouter />` component
+
+<SourcePR number="14462" title="feat: clean deprecated APIs"/>
+
+In Astro 4.0, the `handleForms` prop of the `<ClientRouter />` component was deprecated.
```
sarah11918: I was about to correct this, thinking you surely meant it was deprecated in 5.0, but nope! It has stuck around! :joy:

```suggestion
In Astro 4.0, the `handleForms` prop of the `<ClientRouter />` component was deprecated, as it was no longer necessary to opt in to handling `submit` events for `form` elements. This functionality has been built-in by default and the property, if still included in your project, silently had no impact on form submission.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,72 @@ const integration = () => {
...
+
+Astro 6.0 removes this prop entirely, as it had no impact on form submission anymore.
```
```suggestion
Astro 6.0 removes this prop entirely and it now must be removed to avoid errors in your project.
```
Notice that I'm providing a tiny bit of extra threat here, only because this has been around, doing literally nothing, for so long. Yeah, it's true of everything "removed" that it's not just "no longer needed" but like the user-facing problem is more "if you have it, stuff breaks because this doesn't exist anymore!" 

It's a judgement call how insistent we might be on any particular feature. I feel like one that's been allowed to silently not cause problems though is one we might want to give a little extra attention to in this way.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,72 @@ const integration = () => {
...
+
+#### What should I do?
+
+Remove the `handleForms` property from your `<ClientRouter />` component. It is no longer necessary.
```
```suggestion
Remove the `handleForms` property from your `<ClientRouter />` component if it exists. It has provided no additional functionality, and so removing it should not change any behavior in your project:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,72 @@ const integration = () => {
...
+</html>
+```
+
+<ReadMore>Learn more about [transitions with forms](/en/guides/view-transitions/#transitions-with-forms).</ReadMore>
+
+### Removed: `prefetch()` `with` option
+
+<SourcePR number="14462" title="feat: clean deprecated APIs"/>
+
+In Astro 4.8.4, the `with` option of the programmatic `prefetch()` function was deprecated.
```
```suggestion
In Astro 4.8.4, the `with` option of the programmatic `prefetch()` function was deprecated in favor of a more sensible default behavior that no longer required specifying the priority of prefetching for each page.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,72 @@ const integration = () => {
...
+
+Astro 6.0 removes this option entirely, defaulting to `link` preloading.
```
```suggestion
Astro 6.0 removes this option entirely and it is no longer possible to configure the priority of prefetching by passing the `with` option. Attempting to do so will now cause errors.

By default, Astro's prefetching now uses an automatic approach that will always try to use `<link rel="prefetch>` if supported, or will fall back to `fetch()`.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,72 @@ const integration = () => {
...
+
+#### What should I do?
+
+Review your `prefetch()` calls and remove the `with` option:
```
```suggestion
Review your `prefetch()` calls and remove the `with` option if it still exists:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -279,6 +279,74 @@ const integration = () => {
 }
 ```
 
+### Removed: old `app.render()` signature (Adapter API)
+
+<SourcePR number="14462" title="feat: clean deprecated APIs"/>
+
+In Astro 4.0, the `app.render()` signature that allowed passing `routeData` and `locals` as optional arguments was deprecated in favor of a single optional `renderOptions` argument.
+
+Astro 6.0 removes this signature entirely.
```
```suggestion
Astro 6.0 removes this signature entirely. Attempting to pass these separate arguments will now cause an error in your project.
```
maybe same reasoning as below?

PR #12477, opened 2025-10-02
[v6] remove context.rewrite in actions

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -375,7 +375,7 @@ prefetch('/about');
 
 <SourcePR number="14477" title="feat!: remove rewrite from action context"/>
 
-In Astro 5.5.6, the `ActionAPIContext.rewrite` method was deprecated because custom endpoints should be used where you might want to use rewrites.
+In Astro 5.5.6, the `ActionAPIContext.rewrite` method was deprecated because custom endpoints should be used instead of rewrites.
 
 Astro 6.0 removes the `rewrite()` method from `ActionAPIContext` entirely and it may no longer be used.
 


src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -371,6 +371,38 @@ prefetch('/about');
 
 <ReadMore>Learn more about [prefetching](/en/guides/prefetch/).</ReadMore>
 
+### Removed: `rewrite()` from Actions context
+
+<SourcePR number="14477" title="feat!: remove rewrite from action context"/>
+
+In Astro 5.5.6, `rewrite()` on Actions context was deprecated as it was not meant to be exposed here.
```
```suggestion
In Astro 5.5.6, the `ActionAPIContext.rewrite` method was deprecated as it was not meant to be exposed here.
```
I traced back to that original PR, and there wasn't any reason/description given there either, but in an ideal world we would be able to say something about what using this feature was/is... like, "Actions handlers were not intended to allow serving content from another page ..." (or explain why a rewrite in an action handler isn't safe, doesn't make sense... or often when we deprecate something it's in favour of something else.

 Do you know what I mean? Something that describes maybe how or when this might have been (improperly?) used that caused us to discover it was there and needed to remove it?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -371,6 +371,38 @@ prefetch('/about');
...
+
+Astro 6.0 removes this property entirely on Actions context.
```
```suggestion
Astro 6.0 removes the `rewrite()` method from `ActionAPIContext` entirely and it may no longer be used.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -371,6 +371,38 @@ prefetch('/about');
...
+    }),
+    handler: async (input, context) => {
+      context.rewrite('/')
+      // ...
+    }
+  })
+}
+```
+
+<ReadMore>Learn more about [rewrites](/en/guides/routing/#rewrites).</ReadMore>
```
sarah11918, line 404: I would only use rewrites as the link here if we actually *discuss/explain* the rewrite concept in the deprecation method above, and we think this really is informing someone more about redirects and not about actions.

In fact, this probably goes to https://docs.astro.build/en/reference/modules/astro-actions/#handler-property, which I notice describes this "mostly full context object" (so yeah, I think this is the good link?) BUT says it excludes *redirect*, but not rewrite! So, can we double check these reference docs too?

sarah11918 (commented):
Some thoughts, and I think maybe our reference docs for actions might be wrong?

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: I think the reason is that actions use endpoints under the hood, but you're not supposed to have access to so much power here. If you need a rewrite, use an endpoint basically. Maybe @ematipico has more context

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: I updated the actions reference, I think there was only one sentence to update. And yeah I thikn here it's more useful to link rewriting than actions

sarah11918: Just checking, we're linking to rewrites or to handling actions?

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: OK, I'll take more context if we have it and Ema has a great explanation, but non-blocking. I'd *prefer* something other than "oops, shouldn't have been here" like, "you were supposed to achieve this with X instead..." or "you should never do this because..."

I'll let you decide when you've figured out whether having something like that is actually doable, or you just want to merge it as at least correct upgrade instructions!

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: Linking to rewrites makes more sense to me, so I left it as is

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: What about something like this?
```suggestion
In Astro 5.5.6, the `ActionAPIContext.rewrite` method was deprecated because custom endpoints should be used where you might want to use rewrites.
```
I updated your comment! How about this? -Sarah

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -371,6 +371,38 @@ prefetch('/about');
 
 <ReadMore>Learn more about [prefetching](/en/guides/prefetch/).</ReadMore>
 
+### Removed: `rewrite()` from Actions context
+
+<SourcePR number="14477" title="feat!: remove rewrite from action context"/>
+
+In Astro 5.5.6, the `ActionAPIContext.rewrite` method was deprecated because custom endpoints should be used where you might want to use rewrites.
```
```suggestion
In Astro 5.5.6, the `ActionAPIContext.rewrite` method was deprecated because custom endpoints should be used instead of rewrites.
```
cleaner!

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -371,6 +371,38 @@ prefetch('/about');
...
+import { defineAction } from 'astro:actions';
+import { z } from 'astro:schema';
+
+export const server = {
+  getGreeting: defineAction({
+    input: z.object({
+      // ...
+    }),
+    handler: async (input, context) => {
+      context.rewrite('/')
```
sarah11918, line 397: So, we're saying remove this. How tricky would it be to show using an endpoint instead?

florian-lefebvre: Honestly I don't know how somehow would use a rewrite properly in an action, so I cannot find an example that makes sense. If we find one, we could easily "translate" the code but we'd have to figure how to make the diff make sense

sarah11918: OK, then we just call this good enough! (I think you mentioned that you didn't think people were doing this anyway...) 🫡

PR #12504, opened 2025-10-05
i18n(de): Translate astro-courses.mdx

randomguy-2650:
#### Description (required)

This PR translates the “Courses” (astro-courses.mdx) page to German.

- Suggested label: i18n, hacktoberfest-accepted

Hacktoberfest 🎃


src/content/docs/de/astro-courses.mdx
```diff
@@ -0,0 +1,51 @@
...
+:::tip[Unterstütze Astro beim Lernen]
+Nutze Astros Affiliate-Links, um bei unseren Bildungspartnern Rabatte zu erhalten – und ein Teil deines Kaufs fließt direkt in die Weiterentwicklung des Open-Source-Projekts Astro!
+:::
+
+Lerne Astro von vertrauenswürdigen Astro-Lehrkräften – mit Videolektionen, interaktiven Übungen und Projekten!
+
+<LinkCard
+  title="Lerne Astro mit Coding in Public"
+  href="https://learnastro.dev/?code=ASTRO_PROMO"
+  description="Ein interaktiver Premiumkurs mit über 150 Videolektionen, in dem du lernst, wie du Astros integrierte Tools und Funktionen nutzt."
```
trueberryless: Ich verwende hier immer gerne "welchem" da es sich flüssiger liest:

```suggestion
  description="Ein interaktiver Premiumkurs mit über 150 Videolektionen, in welchem du lernst, wie du Astros integrierte Tools und Funktionen nutzt."
```

randomguy-2650: Dazu würde ich erst noch Meinungen bekommen, weil irgendwie sehen beide komisch aus.

Vielleicht geht „indem“ statt „in dem“, aber ich bin noch nicht so sicher ob das Sinn ergibt.

---

I would rather get opinions on this because both look weird for some reason.

“indem” might work, instead of “in dem”, but not sure if this makes sense.

trueberryless: Hmm, I would not go with "indem" as it has a different meaning then, which would translate back to English like this which does not make sense:
> A premium interactive course with 150+ video lessons **by learning** how to use Astro’s built-in tooling and features.

And the suggestion to use "welchem" is little personal preference, which makes reading the text more clear tho, as it is not used as another word somewhere (in contrast to "der" which can also be used as an article...)

PR #12510, opened 2025-10-06
[v6] stabilize experimental.headingIdCompat

florian-lefebvre:

- Closes #
- Suggested label:


src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,57 @@ If you need more control over environment variables in Astro, we recommend you u
...
+will generate the following HTML:
+
+```html del={1} ins={2}
+<h2 id="picture"><code>&lt;Picture /&gt;</h2>
+<h2 id="picture-"><code>&lt;Picture /&gt;</h2>
+```
+
+If you want to keep the old ID generation for backward compatibility reasons, you can use a custom Rehype plugin:
+
+TODO:
```
florian-lefebvre: @delucis I'd need your help to write the code snippet here

delucis: Here’s a suggestion:

````suggestion
<Steps>

1. Install required dependencies:

   <PackageManagerTabs>
   <Fragment slot="npm">
   ```sh
   npm i github-slugger hast-util-heading-rank unist-util-visit hast-util-to-string
   ```
   </Fragment>
   <Fragment slot="pnpm">
   ```sh
   pnpm add github-slugger hast-util-heading-rank unist-util-visit hast-util-to-string
   ```
   </Fragment>
   <Fragment slot="yarn">
   ```sh
   yarn add github-slugger hast-util-heading-rank unist-util-visit hast-util-to-string
   ```
   </Fragment>
   </PackageManagerTabs>
   
2. Create a custom rehype plugin that will generate headings IDs like Astro v5:

   ```js title="plugins/rehype-slug.mjs"
   import GithubSlugger from 'github-slugger';
   import { headingRank } from 'hast-util-heading-rank';
   import { visit } from 'unist-util-visit';
   import { toString } from 'hast-util-to-string';

   const slugs = new GithubSlugger();

   export function rehypeSlug() {
     /**
      * @param {import('hast').Root} tree
      */
     return (tree) => {
       slugs.reset();
       visit(tree, 'element', (node) => {
         if (headingRank(node) && !node.properties.id) {
           let slug = slugs.slug(toString(node));
           // Strip trailing hyphen like in Astro v5 and below:
           if (slug.endsWith('-')) slug = slug.slice(0, -1);
           node.properties.id = slug;
         }
       });
     };
   }
   ```
   
3. Add the custom plugin to your Markdown configuration in `astro.config.mjs`:

   ```js title="astro.config.mjs" ins={2} ins="rehypeSlug"
   import { rehypeSlug } from './plugins/rehype-slug';
   
   export default defineConfig({
     markdown: {
       rehypePlugins: [rehypeSlug],
     },
   });
   ```

</Steps>
````

Some notes:

- This is not 100% 1:1 with what Astro does: Astro has a bunch more [complex logic](https://github.com/withastro/astro/blob/63ca266b9039ed241ee5257d1b2e6b2337a041c9/packages/markdown/remark/src/rehype-collect-headings.ts#L37-L67) for deciding what the string contents of a heading is, e.g. to support substituting frontmatter variables in MDX expressions. But reproducing that here would be pretty painful. For the most part the `toString()` utility should work for most people.

- We could drop the need for the `headingRank()` utility and write some custom logic for detecting a heading tag (Astro itself doesn’t use `hast-util-heading-rank`), but given the need for the other dependencies, I thought this one might as well be included too as it makes the code a little cleaner to read. It’s the same [approach `rehype-slug` uses](https://github.com/rehypejs/rehype-slug/blob/f90fc8f4b469b54b26c13ce4b754bd7e6da3e351/lib/index.js#L43).

- I didn’t preview this locally — could be I messed up some formatting in the GH editor.

- StackBlitz example where I tested this custom plugin: https://stackblitz.com/edit/github-jfpuphkx?file=src%2FContent.md,src%2Fpages%2Findex.astro,astro.config.mjs&on=stackblitz

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 2]
```diff
@@ -585,6 +586,57 @@ If you need more control over environment variables in Astro, we recommend you u
...
+```
+
+will generate the following HTML:
+
+```html del={1} ins={2}
+<h2 id="picture"><code>&lt;Picture /&gt;</h2>
+<h2 id="picture-"><code>&lt;Picture /&gt;</h2>
+```
+
+If you want to keep the old ID generation for backward compatibility reasons, you can use a custom Rehype plugin:
```
sarah11918: This seems like a pretty long example! I wonder whether we can do something like

```suggestion
If you want to keep the old ID generation for backward compatibility reasons, you can create a custom rehype plugin that will generate headings IDs like Astro v5.
```
And *either* leave it like that as an exercise left to the reader (😅) or, if there's a good *link* to building something like this, link to an example, or maybe even hide a potential example behind `<details>` like:

```
<details>
	<summary>See an example of how to build this with `github-slugger`</summary>
	instructions
</details>
```

sarah11918 (commented):
Looks good, but see my comment re: including how to build a custom rehype plugin inside this guide 😅

src/content/docs/en/guides/upgrade-to/v6.mdx
delucis: I don't think it can be an exercise left to the reader — they'd need to know the details of what changed and that's on us. 

Could be collapsed behind a details element though.

florian-lefebvre: Do we a have component for this kind of collapsible yet?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,120 @@ If you need more control over environment variables in Astro, we recommend you u
...
+   <Fragment slot="yarn">
+   ```sh
+   yarn add github-slugger hast-util-heading-rank unist-util-visit hast-util-to-string
+   ```
+   </Fragment>
+   </PackageManagerTabs>
+   
+2. Create a custom rehype plugin that will generate headings IDs like Astro v5:
+
+   ```js title="plugins/rehype-slug.mjs"
```
florian-lefebvre, line 669: Is there a reason why you used js instead of ts?

src/content/docs/en/guides/upgrade-to/v6.mdx
delucis: Starlight includes styles for `<details>` so you can just use that directly.

src/content/docs/en/guides/upgrade-to/v6.mdx
delucis: Mostly because that's what most of our examples do. But also because it means you can inline it in your config file if you like even if that is `.mjs`.

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: I'll give it a try

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
...
 
+### Changed: Markdown heading ID generation
+
+<SourcePR number="14494" title="feat!: stabilize experimental.headingIdCompat"/>
+
+In Astro 5.5, the `experimental.headingIdCompat` flag was introduced to make the IDs generated by Astro for Markdown headings compatible with common platforms like GitHub and npm, using the popular [`github-slugger`](https://github.com/Flet/github-slugger) package.
+
+In Astro 5.x, Astro had an additional default processing step that stripped a trailing hyphen from the end of IDs for headings ending in special characters.
+
+Astro 6.0 removes this experimental flag and makes this the new default behavior in Astro: trailing hyphen from the end of IDs for headings ending in special characters isn't stripped anymore.
```
yanthomasdev: There's a weird mix of singular and plural here. I guess we can go with plural all the way and make it flow better?
```suggestion
In Astro 5.x, Astro had an additional default processing step that stripped trailing hyphens from the end of IDs for headings ending in special characters.

Astro 6.0 removes this experimental flag and makes this the new default behavior in Astro: trailing hyphens from the end of IDs for headings ending in special characters aren't stripped anymore.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
...
+   export function rehypeSlug() {
+     /**
+      * @param {import('hast').Root} tree
+      */
+     return (tree) => {
+       slugs.reset();
+       visit(tree, 'element', (node) => {
+         if (headingRank(node) && !node.properties.id) {
+           let slug = slugs.slug(toString(node));
+           // Strip trailing hyphen like in Astro v5 and below:
```
yanthomasdev:
```suggestion
           // Strip trailing hyphens like in Astro v5 and below:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
 
 <ReadMore>Learn more about [environment variables](/en/guides/environment-variables/) in Astro, including `astro:env`.</ReadMore>
 
+### Changed: Markdown heading ID generation
+
+<SourcePR number="14494" title="feat!: stabilize experimental.headingIdCompat"/>
+
+In Astro 5.5, the `experimental.headingIdCompat` flag was introduced to make the IDs generated by Astro for Markdown headings compatible with common platforms like GitHub and npm, using the popular [`github-slugger`](https://github.com/Flet/github-slugger) package.
+
+In Astro 5.x, Astro had an additional default processing step that stripped trailing hyphens from the end of IDs for headings ending in special characters.
```
``````suggestion
In Astro 5.x, an additional default processing step to Markdown stripped trailing hyphens from the end of IDs for section headings ending in special characters. This provided a cleaner `id` value, but could lead to incompatibilities rendering your Markdown across platforms.

In Astro 5.5, the `experimental.headingIdCompat` flag was introduced to allow you to make the IDs generated by Astro for Markdown headings compatible with common platforms like GitHub and npm, using the popular [`github-slugger`](https://github.com/Flet/github-slugger) package.
``````

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 7]
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
...
+
+Astro 6.0 removes this experimental flag and makes this the new default behavior in Astro: trailing hyphens from the end of IDs for headings ending in special characters aren't stripped anymore.
```
``````suggestion
Astro 6.0 removes this experimental flag and makes this the new default behavior in Astro: trailing hyphens from the end of IDs for headings ending in special characters are no longer removed.
``````

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 8]
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
 
 <ReadMore>Learn more about [environment variables](/en/guides/environment-variables/) in Astro, including `astro:env`.</ReadMore>
 
+### Changed: Markdown heading ID generation
+
+<SourcePR number="14494" title="feat!: stabilize experimental.headingIdCompat"/>
+
+In Astro 5.5, the `experimental.headingIdCompat` flag was introduced to make the IDs generated by Astro for Markdown headings compatible with common platforms like GitHub and npm, using the popular [`github-slugger`](https://github.com/Flet/github-slugger) package.
+
```
```suggestion
```
The thing that will affect MOST people here is that the default behaviour has changed, not that some flag they probably weren't using is now stable. So, I think reordering this like below makes sense?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
...
+In Astro 5.x, Astro had an additional default processing step that stripped trailing hyphens from the end of IDs for headings ending in special characters.
+
+Astro 6.0 removes this experimental flag and makes this the new default behavior in Astro: trailing hyphens from the end of IDs for headings ending in special characters aren't stripped anymore.
+
+#### What should I do?
+
+If you were previously using this experimental feature, you must [remove this experimental flag from your configuration](#experimental-flags) as it no longer exists.
```
```suggestion
If you were previously using the experimental feature to enforce trailing hyphens, you must [remove this experimental flag from your configuration](#experimental-flags) as it no longer exists.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
...
+
+If you were previously using the `rehypeHeadingIds` plugin directly, remove the `headingIdCompat` option as it no longer exists:
```
```suggestion
If you were previously using the `rehypeHeadingIds` plugin directly to enforce compatibility, remove the `headingIdCompat` option as it no longer exists:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
...
+## `<Picture />`
+```
+
+will generate the following HTML:
+
+```html del={1} ins={2}
+<h2 id="picture"><code>&lt;Picture /&gt;</h2>
+<h2 id="picture-"><code>&lt;Picture /&gt;</h2>
+```
+
```
```suggestion
```
Again, same thing here. I think the big change to highlight is for people who previously did nothing, and now some default behaviour has changed on them

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
...
+<SourcePR number="14494" title="feat!: stabilize experimental.headingIdCompat"/>
+
+In Astro 5.5, the `experimental.headingIdCompat` flag was introduced to make the IDs generated by Astro for Markdown headings compatible with common platforms like GitHub and npm, using the popular [`github-slugger`](https://github.com/Flet/github-slugger) package.
+
+In Astro 5.x, Astro had an additional default processing step that stripped trailing hyphens from the end of IDs for headings ending in special characters.
+
+Astro 6.0 removes this experimental flag and makes this the new default behavior in Astro: trailing hyphens from the end of IDs for headings ending in special characters aren't stripped anymore.
+
+#### What should I do?
+
```
sarah11918, line 600:
``````suggestion

If you have manual links to headings, you may need to update the anchor link value with a new trailing hyphen. For example, the following Markdown heading:

```md
## `<Picture />`
```

will now generate the following HTML with a trailing hyphen in the heading `id`:

```html ins="-"
<h2 id="picture-"><code>&lt;Picture /&gt;</h2>
```

and must now be linked to as:

```markdown ins="-"
See [the Picture component](/en/guides/images/#picture-) for more details.
```

``````
I think the above way of just showing the hyphen added is more obvious? And, I included an example of writing the new link itself because THAT is actually what they will now have to do.

I also think this is going to be the main thing people have to worry about, so I put it first.

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 13]
```diff
@@ -585,6 +586,124 @@ If you need more control over environment variables in Astro, we recommend you u
...
+will generate the following HTML:
+
+```html del={1} ins={2}
+<h2 id="picture"><code>&lt;Picture /&gt;</h2>
+<h2 id="picture-"><code>&lt;Picture /&gt;</h2>
+```
+
+<details>
+
+<summary>If you want to keep the old ID generation for backward compatibility reasons, you can create a custom rehype plugin that will generate headings IDs like Astro 5.x. Click to expand.</summary>
```
```suggestion
If you want to keep the old ID generation for backward compatibility reasons, you can create a custom rehype plugin that will generate headings IDs like Astro 5.x. This will allow you to continue to use your existing anchor links without adding trailing hyphens.

<details>

<summary>Create a custom rehype plugin to strip trailing hyphens</summary>
```
I *think* that less text in the summary is a better reading experience? What do you think?

sarah11918 (approved):
LGTM!

PR #12514, opened 2025-10-06
Remove outdated ‘hybrid’ explanation from Spanish docs (configuration-reference.mdx)

Dinotaurent:
#### Description (required)

This PR removes the explanation about the `hybrid` mode in the Spanish documentation.

That content was removed from the Astro framework but was still present in the Spanish docs, which could cause confusion for readers.

This update aligns the Spanish documentation with the current English version.

Changes made **only in the Spanish documentation**.

---

- Suggested label: `i18n`, `docs`, `bug`

---

Yes, first-time contributor! 😊  
**Astro Discord:** @Dinotaurent

### DIFF
--- a/src/content/docs/es/reference/configuration-reference.mdx
+++ b/src/content/docs/es/reference/configuration-reference.mdx
@@ -141,7 +141,7 @@ Puedes personalizar el [código de estado de redirección](https://developer.moz
 ### output
 
 <p>
-**Tipo:** `'static' | 'server' | 'hybrid'`<br />
+**Tipo:** `'static' | 'server'`<br />
 **Por defecto:** `'static'`
 </p>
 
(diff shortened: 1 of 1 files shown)


yanthomasdev:
Thanks @Dinotaurent, I believe there are more changes to this file that need to be applied first. Here is a full list of commits in the English file since the last update to the translation: https://github.com/withastro/docs/commits/main/src/content/docs/en/reference/configuration-reference.mdx?since=2024-12-03T11:19:23.000Z

Feel free to check on our [i18n dashboard](https://i18n.docs.astro.build/)

sarah11918:
Thank you for fixing this quite outdated docs page, @Dinotaurent ! We really appreciate it!

As Yan pointed out, there are other things that could be updated too. 😅 Are you available to make this page *totally* up to date?? If not, (we know that might be a lot to ask you!), we can merge only your current change. But if we can get you to update all the changes, that would be amazing for us.

Please let us know what you prefer! Thanks!

Dinotaurent:
Hi, this is really confusing to me, as this is my first time "collaborating" on another project, and I'm still not entirely familiar with the technicalities and the correct way to work on these teams. And to make things even more complicated, English isn't my native language, and I only speak a little bit of it hahahah 🫠, so I really don't understand everything because I'm just getting started. Sorry, I'm a little afraid of damaging something. 😅

sarah11918:
@Dinotaurent No problem at all!! We are very happy to have the change you contributed, and we can let someone else make more updates. 💜 Thank you again, and welcome to Team Docs!

PR #12516, opened 2025-10-07
feat: document cli create-key command

florian-lefebvre:
#### Description (required)

`create-key` was shown on the server islands guide, but not on the CLI reference

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/cli-reference.mdx
+++ b/src/content/docs/en/reference/cli-reference.mdx
@@ -376,6 +376,10 @@ astro telemetry reset
 Add the `astro telemetry disable` command to your CI scripts or set the `ASTRO_TELEMETRY_DISABLED` environment variable.
 :::
 
+## `astro create-key`
+
+Generates a key to encrypt props passed to Server islands, see [how to reuse the encryption key](/en/guides/server-islands/#reusing-the-encryption-key).
+
 ## Common flags
 
 ### `--root <path>`


src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -376,6 +376,10 @@ astro telemetry reset
 Add the `astro telemetry disable` command to your CI scripts or set the `ASTRO_TELEMETRY_DISABLED` environment variable.
 :::
 
+## `astro create-key`
+
+Generates a key to encrypt props passed to Server islands, see [how to reuse the encryption key](/en/guides/server-islands/#reusing-the-encryption-key).
```
yanthomasdev: nit
```suggestion
Generates a key to encrypt props passed to server islands, see [how to reuse the encryption key](/en/guides/server-islands/#reusing-the-encryption-key).
```

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -376,6 +376,10 @@ astro telemetry reset
 Add the `astro telemetry disable` command to your CI scripts or set the `ASTRO_TELEMETRY_DISABLED` environment variable.
 :::
 
+## `astro create-key`
+
+Generates a key to encrypt props passed to server islands, see [how to reuse the encryption key](/en/guides/server-islands/#reusing-the-encryption-key).
```
sarah11918:
``````suggestion
Generates a key to encrypt props passed to server islands.

```shell
astro create-key
```

Set this key as the `ASTRO_KEY` environment variable (e.g. in a `.env` file) and include it in your CI/CD or host’s build settings when you need [a constant encryption key for your server islands](/en/guides/server-islands/#reusing-the-encryption-key) for situations like rolling deployments, multi-region hosting or a CDN that caches pages containing server islands.
``````
Something like this maybe?

sarah11918 (commented):
I think we show the shell command for everything else. Plus, grabbed a bit of copy from the page itself with a tiny bit of when/why/how to use this.

PR #12528, opened 2025-10-09
Update captcha.mdx

liyunze-coding:
#### Description

Fixed reCAPTCHA documentation error.

This guide uses reCAPTCHA v2 (not v3) and specifically the Invisible reCAPTCHA badge.

(Discord user: RythonDev)


sarah11918:
Hi! We believe this recipe does in fact use v3 and not v2. [Googles doc's](https://developers.google.com/recaptcha/docs/v3?hl=en#actions) say that reCAPTCHA v3 introduces a new concept: actions, which are used here.

So I will close this PR as I believe this is a v3 recipe. However, if you are able to provide justification and show us how this is a v2 recipe (I'm not familiar with reCAPTCHA, and we are going by the original submission that said it was v3, and checking Google's docs that does seem to match) then you are welcome to reopen and explain more why this should be v2!

PR #12530, opened 2025-10-10
Refactor contributors list & fetch wrapper [i18nIgnore]

HiDeoo:
#### Description (required)

The issue has been found and is related to a request to the GitHub API to fetch contributors returning a 403 status code.

The issue will not be fixed in this PR but I'm keeping the investigation notes here for reference and also as a reminder to refactor this part of the code to add more logging when fetching failures happen and also revisit the retry logic (the default settings exceeds the Netlify build time limit).

| Run | Status | Details |
| --- | ------ | ------------------------------------------------- |
| [1](https://app.netlify.com/projects/astro-docs-2/deploys/68e8ae679ebe2f000849b907)   | ❌     | Logs stopped when rendering "Contribute to Astro" |
| [2](https://app.netlify.com/projects/astro-docs-2/deploys/68e8b14637d9920008a14d3d)   | ❌     | Logs stopped when rendering ([`renderToString`](https://github.com/delucis/starlight-llms-txt/blob/main/packages/starlight-llms-txt/entryToSimpleMarkdown.ts#L183)) "Contribute to Astro" |
| [3](https://app.netlify.com/projects/astro-docs-2/deploys/68e8b43641a97c0008d39730)   | ❌     | Logs stopped when starting to [fetch contributors](https://github.com/withastro/docs/blob/main/src/util/getContributors.ts#L23) |
| [4](https://app.netlify.com/projects/astro-docs-2/deploys/68e8b91dbafb190008e76104)   | ❌     | [Fetch contributors](https://github.com/withastro/docs/blob/main/src/util/cachedFetch.ts#L25-L32) in a retry loop (default [10](https://github.com/sindresorhus/p-retry#retries)) with exponential backoff |
| [5](https://app.netlify.com/projects/astro-docs-2/deploys/68e8bcb7fe08ef0007c97edc)   | ❌     | `Bad response for https://api.github.com/repos/withastro/docs/contributors (403): Forbidden` |

- Closes #
- Suggested label:


delucis:
I confirmed this was a token issue and updating our token fixed deploys.

Nonetheless, some good action items we should look at coming out of this:

1. Move to collect GitHub contributor data in a GitHub workflow instead of with every build:

   - Remove the build-time calls to the GitHub API
   - Create a GitHub workflow that runs weekly (?) that fetches the data and stores it in the repo as JSON
   - Update the component to use that data instead

   Means everyone can render the component locally, and no need for tokens anywhere except for on GitHub where we get one for free.
   
2. Add logging to our fetch wrapper (which adds retries/caching) to ensure that cases like this will be easier to diagnose in the future.

HiDeoo:
Updated the PR based on your recommendations:

- Cleanup all logs I introduced for investigation.
- Move contributors data to a file content collection:
  - The file updated weekly as you suggested but we can discuss the schedule if needed.
  - We only store the data we really need (e.g. not the number of contributions which will make diffs a bit easier — I guess showing (and storing) contributors alphabetically rather than number of contributions would make this even easier if we wanted).
- I removed some unused code (`<TranslatorList/>`).
- I tweaked our `cachedFetch` utility:
  - I added logs on failed attempts to fetch data including the attempt number and retries left.
  - I tweaked the exponential backoff settings:

    The formula is: `Math.round(random * minTimeout * (factor ** (attempt - 1)))`
    With the defaults settings, this gives us: `Math.round(1 * 1000 * (2 ** (attempt - 1)));`

    | Attempt/Retry | Exponential Backoff Time |
    | --- | --- |
    | attempt 1 |  |
    | attempt 2 - retry 1 | 1s |
    | attempt 3 - retry 2 | 2s |
    | attempt 4 - retry 3 | 4s |
    | attempt 5 - retry 4 | 8s |
    | attempt 6 - retry 5 | 16s |
    | attempt 7 - retry 6 | 32s |
    | attempt 8 - retry 7 | 1m04s |
    | attempt 9 - retry 8 | 2m08s |
    | attempt 10 - retry9  | 4m16s |
    | attempt 11 - retry 10 | 8m32s |

    This is a lot and in case of failure, Netlify will timeout the build before we reach the max retries of 10.

    I set the max retries to `5` with a factor of `3` so the new settings give us:

    | Attempt/Retry | Exponential Backoff Time |
    | --- | --- |
    | attempt 1 |  |
    | attempt 2 - retry 1 | 1s |
    | attempt 3 - retry 2 | 2s |
    | attempt 4 - retry 3 | 9s |
    | attempt 5 - retry 4 | 27s |
    | attempt 6 - retry 5 | 1m21s |

    Concretely, this means the first attempts happens like before, very quickly. Then if it fails, we wait a lot more between retries, and just give up way earlier.

    Happy to discuss/adjust these settings of course.

HiDeoo:
> One thought though: we’re not using the `id` field right? Why not make the `login` value the `id` and just use that?

We actually [do](https://github.com/withastro/docs/pull/12530/files#diff-85688b72b20d6c6881180c8c80f1ca681a949707f542a8f36b9bf73c52cf845aR23) for the profile picture that's why I went with both of these fields being stored.

PR #12539, opened 2025-10-11
i18n(de): Update the v2 to v3 upgrade guide to follow the new i18n guide

randomguy-2650:
#### Description (required)

This PR updates the v2 to v3 guide (something that basically no one probably visits) to follow the new i18n guide. It also updates the title to just “v3.0” to match the English guide, from “Auf v3.0 aktualisieren” (≙ “Upgrade to v3.0”).

- Suggested label: i18n, hacktoberfest-accepted

Hacktoberfest 🎃


src/content/docs/de/guides/upgrade-to/v3.mdx
```diff
@@ -67,7 +67,7 @@ export default defineConfig({
 
 Diese Features sind jetzt standardmäßig verfügbar:
 
-- View Transitions für animierte Seitenübergänge und persistente Astro-Inseln.
+- View Transitions für animierte Seitenübergänge und persistente Astro-Islands.
```
trueberryless: I think "Astro Islands" (and "Astro Island") are the only use cases where we do not need the hyphen, as it is supposed to be translated as an "Astro Feature"...

```suggestion
- View Transitions für animierte Seitenübergänge und persistente Astro Islands.
```

src/content/docs/de/guides/upgrade-to/v3.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Auf Astro v3 aktualisieren
+title: v3.0
```
yanthomasdev: The title is still the full "Upgrade to Astro v(version)", the sidebar looks different because there's a property for it.

So, you should add to this and the v2 page's frontmatter. Could you do that? Then we are good to go.
```
sidebar:
  label: v3.0
```

PR #12542, opened 2025-10-12
fix: update `install-and-setup.mdx`

vrabe:
#### Description (required)

Update `install-and-setup.mdx`.

It seems that the space is intended? I'm not sure it's correct to remove the space.

I'm also not sure it's correct to add the `<ReadMore>`  component or not.

- Closes #


src/content/docs/en/install-and-setup.mdx
yanthomasdev, line 307: I guess this line should also have `<ReadMore>`

vrabe: I have added it.

PR #12552, opened 2025-10-14
docs: change responsive image global config reference link

trueberryless:
#### Description (required)

Change link on page https://docs.astro.build/en/guides/images/#responsive-image-behavior to link to `image.responsiveStyles` instead of `image.layout`

- Suggested label: **Hacktoberfest**

### DIFF
--- a/src/content/docs/en/guides/migrate-to-astro/from-nuxtjs.mdx
+++ b/src/content/docs/en/guides/migrate-to-astro/from-nuxtjs.mdx
@@ -489,7 +489,7 @@ See more about [Styling in Astro](/en/guides/styling/).
 
 ### Nuxt Image Plugin to Astro
 
-Convert any [Nuxt `<nuxt-img/>` or `<nuxt-picture/>` components](https://image.nuxtjs.org/components/nuxt-img) to [Astro's own image component](/en/guides/images/) in `.astro` or `.mdx` files, or to a [standard HTML `<img>`](/en/guides/images/#images-in-ui-framework-components) or `<picture>` tag as appropriate in your Vue components.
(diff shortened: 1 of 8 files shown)


trueberryless:
I was unsure too, so I asked in [Discord](https://discord.com/channels/830184174198718474/853350631389265940/1427664533694189669), but we can close it if it's right 👍

ArmandPhilippot:
Well, at first reading, I also had a doubt because of the Discord conversation. But reading it again, this sounds correct to me. So, maybe another opinion could be useful here!

@yanthomasdev, what made you agree this was incorrect on Discord?

If the wording is confusing, maybe we should update the wording rather than the link? Because I do think linking to `layout` rather than `responsiveStyles` was intentional.

src/content/docs/en/guides/images.mdx
```diff
@@ -323,7 +323,7 @@ Responsive images are images that adjust to improve performance across different
 
 With [responsive image properties](/en/reference/modules/astro-assets/#responsive-image-properties) applied to the `<Image />` or `<Picture />` components, Astro will automatically generate the required `srcset` and `sizes` values for your images, and apply the necessary [styles to ensure they resize correctly](#responsive-image-styles). 
 
-When this responsive behavior is [configured globally](/en/reference/configuration-reference/#imagelayout), it will apply to all image components and also to any local and remote images using [the Markdown `![]()` syntax](/en/guides/images/#images-in-markdown-files).
+When this responsive behavior is [configured globally](/en/reference/configuration-reference/#imageresponsivestyles), it will apply to all image components and also to any local and remote images using [the Markdown `![]()` syntax](/en/guides/images/#images-in-markdown-files).
```
trueberryless: Maybe something like that could help: 🤔

```suggestion
When this responsive behavior is [configured globally with `image.layout`](/en/reference/configuration-reference/#imagelayout), it will apply to all image components and also to any local and remote images using [the Markdown `![]()` syntax](/en/guides/images/#images-in-markdown-files).
```

ArmandPhilippot: Sounds good to me!

ArmandPhilippot (approved):
Thanks @trueberryless, I went ahead and merged your suggestion. LGTM!

PR #12572, opened 2025-10-16
feat(fonts): granular preload

florian-lefebvre:
#### Description (required)

Documents the addition of granular preloads to the `Font` component

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/netlify.mdx
+++ b/src/content/docs/en/guides/integrations-guide/netlify.mdx
@@ -248,6 +248,22 @@ With [fine-grained cache control](https://www.netlify.com/blog/swr-and-fine-grai
 standard caching headers like `CDN-Cache-Control` or `Vary`.
 Refer to the docs to learn about implementing e.g. time to live (TTL) or stale while revalidate (SWR) caching: https://docs.netlify.com/platform/caching
 
+### Skew Protection
+
+<p><Since v="6.6.0" pkg="@astrojs/netlify" /></p>
+
+Netlify's skew protection ensures that users accessing your site during a deployment continue to receive content from the same deploy version. The Netlify adapter automatically configures skew protection for Astro features like actions, server islands, view transitions, and prefetch requests by injecting the current deploy ID into internal requests. This prevents version mismatches between the client and server during active deployments.
+
+While Astro automatically adds the skew protection header for its built-in features, if you are making your own fetch requests to your site, you can include the header manually using the `DEPLOY_ID` environment variable:
+
+```js
+const response = await fetch('/api/endpoint', {
+  headers: {
+    'X-Netlify-Deploy-ID': import.meta.env.DEPLOY_ID,
+  },
+});
+```
+
 ### Including or excluding files from Netlify Functions
 
 When deploying an Astro site with on-demand rendering to Netlify, the generated functions automatically trace and include server dependencies. However, you may need to customize which files are included in your Netlify Functions.
--- a/src/content/docs/en/reference/adapter-reference.mdx
+++ b/src/content/docs/en/reference/adapter-reference.mdx
@@ -53,6 +53,18 @@ interface AstroAdapter {
 	args?: any;
 	adapterFeatures?: AstroAdapterFeatures;
 	supportedAstroFeatures: AstroAdapterFeatureMap;
+	client?: {
+		/**
+		 * Headers to inject into Astro's internal fetch calls (Actions, View Transitions, Server Islands, Prefetch).
+		 * Can be an object of headers or a function that returns headers.
+		 */
+		internalFetchHeaders?: Record<string, string> | (() => Record<string, string>);
+		/**
+		 * Query parameters to append to all asset URLs (images, stylesheets, scripts, etc.).
+		 * Useful for adapters that need to track deployment versions or other metadata.
+		 */
+		assetQueryParams?: URLSearchParams;
+	};
 }
 
 export interface AstroAdapterFeatures {


src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -325,6 +325,32 @@ import { Font } from 'astro:assets';
 <Font cssVariable="--font-roboto" preload />
 ```
 
+If not specified or set to `false`, no preload link will be emitted. If set to `true`, all possible links will be included.
```
ArmandPhilippot: Plural:
```suggestion
If not specified or set to `false`, no preload links will be emitted. If set to `true`, all possible links will be included.
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -325,6 +325,32 @@ import { Font } from 'astro:assets';
...
+
+#### Granular preloads
+
+<p>
+<Since v="5.15.0" />
+</p>
+
+If you need more control over which font file is preloaded, you can instead provide an array of objects:
+
+```astro title="src/components/Head.astro" {7-10}
```
ArmandPhilippot, line 342: We usually add the description before the code snippet. This is also consistent with the other examples on the page:
````suggestion
The following example will only preload font files with a `400` weight or a `normal` style in the `latin` subset:

```astro title="src/components/Head.astro" {7-10}
````

File: src/content/docs/en/reference/experimental-flags/fonts.mdx [hunk 3]
```diff
@@ -325,6 +325,32 @@ import { Font } from 'astro:assets';
...
+<Font
+  cssVariable="--font-roboto"
+  preload={[
+    { subset: 'latin', style: 'normal' },
+    { weight: '400' },
+  ]}
+/>
+```
+
+In this example, only font files that have a `400` weight, or a `normal` style in the `latin` subset will be preloaded.
```
ArmandPhilippot: Because we moved it to the top:
```suggestion
```

ArmandPhilippot (approved):
Thanks, LGTM. I left a few nits: a typo and we usually describe the code snippet before, not after.

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -325,6 +325,32 @@ import { Font } from 'astro:assets';
...
+
+<Font
+  cssVariable="--font-roboto"
+  preload={[
+    { subset: 'latin', style: 'normal' },
+    { weight: '400' },
+  ]}
+/>
+```
+
```
florian-lefebvre, line 355:
```suggestion

If a candidate is a file with a variable weight, it will be preloaded if the passed weight is within range. For example, a font file for font weight `100 900` will be included since `400` is within that range.

```

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -325,6 +325,34 @@ import { Font } from 'astro:assets';
 <Font cssVariable="--font-roboto" preload />
 ```
 
+If not specified or set to `false`, no preload links will be emitted. If set to `true`, all possible links will be included.
```
```suggestion
With the `preload` directive, the browser will immediately begin downloading all possible font links during page load. 
```
We don't actually show usage of setting this to "true" or "false", but rather including the directive or not. So, I think we can just talk about adding the directive?

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -325,6 +325,34 @@ import { Font } from 'astro:assets';
...
+
+#### Granular preloads
+
+<p>
+<Since v="5.15.0" />
+</p>
+
+If you need more control over which font file is preloaded, you can instead provide an array of objects:
```
```suggestion
You may not always want to preload every font link, as this can block loading other important resources, or may download fonts that are not needed for the current page. 

To selectively control which font files are preloaded, you can provide an array of objects describing any combination of font `weight`, `style`, or `subset` to preload.
```
Maybe explain the object a bit first?

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -325,6 +325,34 @@ import { Font } from 'astro:assets';
...
+<Font
+  cssVariable="--font-roboto"
+  preload={[
+    { subset: 'latin', style: 'normal' },
+    { weight: '400' },
+  ]}
+/>
+```
+
+If a candidate is a file with a variable weight, it will be preloaded if the passed weight is within range. For example, a font file for font weight `100 900` will be included since `400` is within that range.
```
sarah11918: I don't know what the word "candidate" means here (I sort of do, but it feels like it's a special term or jargon when it's not, so we should explain in more explicit terms.)

Do you mean something like:

```suggestion
Variable weight font files will be preloaded if any weight within its range is requested. For example, a font file for font weight `100 900` will be included when `400` is specified in a `preload` object.
```

But also, will ALL the weights 100 to 900 be dowloaded, or just the 400 in this case? I would want to specify that, too.

src/content/docs/en/reference/experimental-flags/fonts.mdx
ArmandPhilippot: Seeing the green lines, I see we have a lot of sentences starting with `if`. Maybe we should try to reword this part to improve the flow. What about something like:
```suggestion
You can get more control over which font file is preloaded using an array of objects. With a standard font, setting a weight, a style or a subset allows you to preload a font file matching exactly the provided configuration. With a variable font, the preloaded font file is a continuous range of font variations containing the provided configuration (e.g. for a `400` weight, the continuous range is `100 900`).
```
Make it accurate of course!

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -325,6 +325,34 @@ import { Font } from 'astro:assets';
...
+
```
ArmandPhilippot: If we move the variable fonts case above, we no longer need this.
```suggestion
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
florian-lefebvre: > But also, will ALL the weights 100 to 900 be dowloaded

A variable font file contains all weights for the given range so yeah they would all be downloaded, which is expected.

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -325,6 +325,36 @@ import { Font } from 'astro:assets';
...
 
+With the `preload` directive, the browser will immediately begin downloading all possible font links during page load. 
+
+#### Granular preloads
+
+<p>
+<Since v="5.15.0" />
+</p>
+
+You may not always want to preload every font link, as this can block loading other important resources, or may download fonts that are not needed for the current page.
```
yanthomasdev: comma nit
```suggestion
You may not always want to preload every font link, as this can block loading other important resources or may download fonts that are not needed for the current page. 
```

sarah11918: I thought about you during this comma, and thought you might allow it! 😂  Happy to remove!

PR #12596, opened 2025-10-20
feat(astro:actions): add docs for missing public APIs

### LINKED ISSUE #12570: Missing Actions API in the reference
### 📚 Subject area/topic

Actions

https://docs.astro.build/en/reference/modules/astro-actions/

See https://github.com/withastro/astro/pull/14538#discussion_r2434890386. Basically, some Actions APIs are only documented in the [Actions guide](https://docs.astro.build/en/guides/actions/), but not in the reference. Non exhaustive list of things that are missing:
- `orThrow`
- `querystring`

_No response_

ArmandPhilippot:
#### Description (required)

* Adds some missing documented imports to the list: `getActionContext` and `getActionPath`.
* Adds docs for `ACTION_QUERY_PARAMS` and `actions`.
* Adds an `astro:actions` types section with docs for `ActionAPIContext`, `ActionClient`, `ActionErrorCode`, `ActionReturnType`, and `SafeResult`.
* Adds a few more usage examples.
* Updates most of the types either to fix them or to add links to the new headings
  * I also simplified some types. For example, I replaced `ActionClient<any, any, any>` with `ActionClient` because this is always `<any, any, any>` and it seems to be an implementation detail rather than something useful to the end-user. And because we now have docs for the types, I think it makes it easier to understand.
  * Fixes the `defineAction()` type: `accept` was missing while we were documenting it under `input`. Not sure if there was a reason for that, maybe we should use the same heading level as the others to avoid any confusion?
  * To avoid long/unreadable types and because we have [a precedent](https://docs.astro.build/en/reference/modules/astro-content/#getentry), I used one `Types:` with a list instead of using an union.
* Moves the codes list from `code` (in `ActionError`) to the new `ActionErrorCode` heading. And, actually, replace the outdated list with a description of the expected format and an external link.
* Rewords some existing descriptions for consistency and to follow our [API references guide](https://contribute.docs.astro.build/reference/api-references/)... there are already a lot of changes so, while I'm here, I think a few more updates doesn't change much the work of translators (maybe reviewers tough, sorry!).
* Removes `Since` when the version is the same as the module
* Adds a few new links in the Actions guide (and fix `full` > `subset`).

##### Things we might want to consider

-  The `ACTION_ERROR_CODES` list was expanded in https://github.com/withastro/astro/pull/13769 (with some renaming). This list is quite long and I don't think this should be the role of Astro to describe them. Especially since some of them are defined as `reserved for future use` (e.g. [`PAYMENT_REQUIRED`](https://www.rfc-editor.org/rfc/rfc9110.html#name-402-payment-required)). If we think otherwise, happy to add back a full list instead.
- Should we precise where methods can be imported/used? For example, `defineAction()` and `getActionContext()` will throw on the client: https://github.com/withastro/astro/blob/main/packages/astro/src/actions/runtime/client.ts I think the current description could be enough, but maybe we want to be more explicit.

##### Not documented

Based on Florian's feedback, some of the available imports are not documented. I think it's fine (NWTWWHB) and they should be removed in v6.

<details>
<summary>See the list of undocumented imports</summary>
<pre><code>
import {
  type Actions,
  type ActionAccept,
  type ActionHandler,
  type AstroActionContext,
  type SerializedActionResult,
  ACTION_ERROR_CODES,
  ActionInputError,
  appendForwardSlash,
  astroCalledServerError,
  callSafely,
  deserializeActionResult,
  formDataToObject,
  getActionQueryString,
  serializeActionResult,
} from "astro:actions";
</code></pre>
</details>

##### Implementation links

There are a lot of changes, so I tried to keep track of the links to help you review the accuracy of the API blocks (type/since).
I haven't updated those list, so they include undocumented imports too.

<details>
<summary>Details of the "regular" imports:</summary>

- [`appendForwardSlash()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/internal-helpers/src/path.ts#L10): since `5.0.6`, see https://github.com/withastro/astro/pull/12657
- [`astroCalledServerError()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L366): no since, it's been there since the Actions stable release
- [`callSafely()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L214-L216): no since, it's been there since the Actions stable release
- [`defineAction()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/server.ts#L65-L80):
  - no since, it's been there since the Actions stable release
  - default: the "real" default for `accept` is `undefined`, but only `form` changes the behavior so I think it's safe to say `json` is the default here.
- [`deserializeActionResult()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L317): no since, it's been there since the Actions stable release
- [`formDataToObject()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/server.ts#L155-L158): no since, it's been there since the Actions stable release
- [`getActionContext()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/server.ts#L270): since `5.0.0`, see https://github.com/withastro/astro/pull/12373
- [`getActionPath()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/types/actions.d.ts#L4-L6): since `5.1.0`
- [`getActionQueryString()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L234): no since, it's been there since the Actions stable release
- [`isActionError()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L151): no since, it's been there since the Actions stable release
- [`isInputError()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L160-L162): no since, it's been there since the Actions stable release
- [`serializeActionResult()`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L257): no since, it's been there since the Actions stable release
- [`ACTION_ERROR_CODES`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L22): no since, it's been there since the Actions stable release
- [`ACTION_QUERY_PARAMS`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L18): no since, it's been there since the Actions stable release
- [`ActionError`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L116): no since, it's been there since the Actions stable release
- [`ActionInputError`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L187): no since, it's been there since the Actions stable release
</details>

<details>
<summary>Details of the type imports:</summary>

- [`Actions`](https://github.com/withastro/astro/blob/ca1d331526a2ed76f09f0c32e981b66dc5e9da76/packages/astro/src/actions/integration.ts#L43): no since, it's been there since the Actions stable release
- [`ActionAPIContext`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L17): no since, it's been there since the Actions stable release
- [`ActionAccept`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/server.ts#L33): no since, it's been there since the Actions stable release
- [`ActionClient`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/server.ts#L41): no since, it's been there since the Actions stable release
- [`ActionErrorCode`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L63): no since, it's been there since the Actions stable release
- [`ActionHandler`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/server.ts#L35-L37): no since, it's been there since the Actions stable release
- [`ActionReturnType`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/server.ts#L39): no since, it's been there since the Actions stable release
- [`AstroActionContext`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/server.ts#L241): since `5.0.0` under a different name. Since `5.4.2` with the current name, see https://github.com/withastro/astro/pull/13262
- [`SafeResult`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L177-L185): no since, it's been there since the Actions stable release
- [`SerializedActionResult`](https://github.com/withastro/astro/blob/93e8bd21225f16c253db4af34bd6ee4c9642974e/packages/astro/src/actions/runtime/shared.ts#L239-L255): no since, it's been there since the Actions stable release
</details>

- Closes #12570


ArmandPhilippot:
I think I've done all the changes I noticed here... Sorry for the Github diff: some content was already there but was moved.

I'm pinging @florian-lefebvre (no rush!) because I need a technical review to check the accuracy and if some things should be removed in v6, but anyone else is welcome to review this one!

I tried to keep a track of all the changes in the first post but there are a lot of changes so I'm not sure how readable/understandable it is. As explained in the issue, I noticed a lack of documentation for all the following imports:
```ts
import {
  ACTION_ERROR_CODES,
  ACTION_QUERY_PARAMS,
  ActionInputError,
  appendForwardSlash,
  astroCalledServerError,
  callSafely,
  deserializeActionResult,
  formDataToObject,
  getActionQueryString,
  serializeActionResult,
  type Actions,
  type ActionAPIContext,
  type ActionAccept,
  type ActionClient,
  type ActionErrorCode,
  type ActionHandler,
  type ActionReturnType,
  type AstroActionContext,
  type SafeResult,
  type SerializedActionResult,
} from "astro:actions";
```

So I would need a technical point of view for all of them, with some exceptions listed below:
* Among them, I think some shouldn't be documented:
  * `type Actions`: I think it is only meant to type the actions export and shouldn't be used directly by the user?
  * `astroCalledServerError`: I don't see a use case for it... this sounds like an internal function to me.
  * `callSafely`: I don't see a use case for it... this sounds like an internal function to me.
  * `ACTION_ERROR_CODES`: I think the constant itself is not useful for users, only the `ActionErrorCode` type built from the constant?
  * `ACTION_QUERY_PARAMS`: same I think this is internal and this is only useful through utilities like `getActionQueryString()`
* `deserializeActionResult` and `serializeActionResult` were already documented but as methods returned by `getActionContext`, not as imports from `astro:actions`. So I mostly moved around the docs here and duplicated the headings. I couldn't find a use case to prefer one over the other... so maybe this is something that should be updated? I mean either get them from `getActionContext` or as top level imports?

I added docs for:
* `appendForwardSlash` but if we move things in v6, I think this could be a good candidate? This doesn't seem related to actions specifically and so it sounds a bit weird to document it here.
* `ActionInputError` but I don't think there is a real use case for end-users because of `isInputError`

And because your issue was specifically about `orThrow` and `queryString` they are under [ActionClient](https://deploy-preview-12596--astro-docs-2.netlify.app/en/reference/modules/astro-actions/#actionclient). Feel free to suggest a more accurate description!

I'm not quite sure about the organization. I documented some properties under their types but maybe it could be more useful under the related function... But, we can think about this later!

(If you need direct links to the implementation, they are listed under a `details` in `Implementation links` in the first post)

EDIT: I was also not sure about an use case for `ActionReturnType`, so I'm not confident about the description...

florian-lefebvre:
Thanks a lot! Before reviewing the content, let's agree on what not to document so we can remove them more easily in v6. For me:

- `ACTION_ERROR_CODES`: do not document (internal)
- `ACTION_QUERY_PARAMS`: document. This is actually useful for advanced cases (I've used it)
- `ActionInputError`: do not document (internal)
- `appendForwardSlash`: do not document (internal)
- `astroCalledServerError`: do not document (internal)
- `callSafely`: do not document (internal)
- `deserializeActionResult`: do not document (probably better to have one source of truth with `getActionContext()`?)
- `formDataToObject`: do not document (internal)
- `getActionQueryString`: do not document (internal)
- `serializeActionResult`: do not document (probably better to have one source of truth with `getActionContext()`?)
- `type Actions`: do not document (we'll need to update the typegen in v6)
- `type ActionAPIContext`: document
- `type ActionAccept`: do not document (unless it's useful for something?)
- `type ActionClient`: document (advanced but useful for generics)
- `type ActionErrorCode`: document (and link to somewhere)
- `type ActionHandler`: maybe not document? I feel like that can be explained as part of `ActionClient`
- `type ActionReturnType`: document
- `type AstroActionContext`: do not document (getActionContext() is the same)
- `type SafeResult`: do not document (most people don't need it)
- `type SerializedActionResult`: do not document (most people don't need it)

That's what I'm thinking but I may be wrong, happy to discuss further! Am I missing something?

ArmandPhilippot:
Thanks! Except maybe `SafeResult` (see below my reasoning), everything makes sense to me.

> ACTION_QUERY_PARAMS: document. This is actually useful for advanced cases (I've used it)

Oh okay, I'll need your help to describe when this is useful then! 😅 

> type ActionAccept: do not document (unless it's useful for something?)

Probably not. It seems the `accept` property in `defineAction` is rather a static value.

> type ActionHandler: maybe not document? I feel like that can be explained as part of ActionClient

I'm not sure I see the relationship with `ActionClient`. But I agree the type might not be useful, we can describe it directly in the `handler()` property of `defineAction`!

> type AstroActionContext: do not document (getActionContext() is the same)

Makes sense! This is what I meant with "I documented some properties under their types but maybe it could be more useful under the related function...", so I agree this is probably not useful.

> type SafeResult: do not document (most people don't need it)

While users might not need to use it directly, maybe this is still useful to describe the format? ie:
```
Type: { data: TOutput, error: undefined } | { data: undefined, error: ActionError }
```
Because many functions uses this and so we can link to a single place?

florian-lefebvre:
> I'm not sure I see the relationship with ActionClient. But I agree the type might not be useful, we can describe it directly in the handler() property of defineAction!

Sorry yes that's what I meant!

> Because many functions uses this and so we can link to a single place?

SGTM

florian-lefebvre:
Re ACTION_QUERY_PARAMS: a common usecase IMO is with form actions. Submitting a form action appends a query param and to remove it, you could write a manual redirect or use `ACTION_QUERY_PARAMS.actionName` to avoid harcoding it, and make sure it doesn't break between releases

ArmandPhilippot:
Before you take a look at the content, I will clean up the content based on what we discussed and I'll add `ACTION_QUERY_PARAMS`. I'm putting this PR back to draft in the meantime and I'll ping you once this is ready!

ArmandPhilippot:
Okay, I think I've done all the updates here. I also updated the first post to actualize what has changed finally.
So, I think this is ready for you, @florian-lefebvre, to check, when you have time, the content and usage examples (I added some new ones in addition to the undocumented APIs/types)!

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -19,22 +19,26 @@ Actions help you build a type-safe backend you can call from client code and HTM
...
+import {
+  ACTION_QUERY_PARAMS,
+  ActionError,
   actions,
   defineAction,
-  isInputError,
+  getActionContext,
+  getActionPath,
   isActionError,
-  ActionError,
```
florian-lefebvre: Technically it's a class so maybe it belongs here?

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -72,13 +76,21 @@ Return values are parsed using the [devalue library](https://github.com/Rich-Har
...
-If your action accepts form inputs, use the `z.object()` validator to automatically parse form data to a typed object. All Zod validators are supported for form data fields (e.g. `z.coerce.date()`). Astro additionally provides special handling under the hood for your convenience for validating the following types of field inputs:
+<p>
+
+**Type:** `"form" | "json"`<br />
+**Default:** `json`
+</p>
+
+Defines the format expected by an action. If your action accepts form inputs, use `accept: "form"` with the `z.object()` validator to automatically parse `FormData` to a typed object.
+
+All Zod validators are supported for form data fields (e.g. `z.coerce.date()`). Astro additionally provides special handling under the hood for your convenience for validating the following types of field inputs:
```
florian-lefebvre: While we're here, could we document this as well? https://github.com/withastro/astro/issues/13562

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -123,83 +135,148 @@ export const server = {
...
+<script>
+import { actions } from 'astro:actions';
+
+async () => {
+  const { data, error } = await actions.myAction({ /* ... */ });
+}
+</script>
+```
+
+In order for Astro to recognize this property, you may need to restart the dev server or [sync the module](/en/reference/cli-reference/#astro-dev) (`s + enter`).
```
florian-lefebvre: 1. Is the link correct? Links to dev instead of sync
2. Not sure showing the shortcut helps here. It really just needs to run once

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -344,3 +446,170 @@ navigator.sendBeacon(
...
+  if (error) {
+    /* handle errors */
+  }
+})
+</script>
+```
+
+Alternatively, you can use it as an object giving you access to the `queryString` and an alternative `orThrow()` method.
+
+#### `queryString` property
```
florian-lefebvre, line 490: I don't know if it's worth documenting `toString()` which is an alias. It's convenient because that means you can use it in form like this:
```jsx
<form actions={actions.myAction}></form>
```
Which calls `toString()` under the hood

florian-lefebvre (commented):
Looking amazing! Left some minor comments

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: Sure, we can take advantage of this PR! Just to be sure I understood correctly the issue, we say:
> All other inputs can be validated using z.string()

But this doesn't guarantee the type to be a `string`, because internally empty strings are converted to `null`. Specifying an error message doesn't change the type, but it is useful to ensure that the user doesn't receive a confusing error message due to the internal type handling. Am I correct?

Do you think it's worth mentioning the null behavior or we just need to recommend using a custom error when using `z.string()`?

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: I took inspiration of the wording used for content collections (e.g. https://docs.astro.build/en/guides/content-collections/#defining-the-collection-schema), but, yeah, maybe the shortcuts are less relevant here.

Does the following works for you?
```suggestion
In order for Astro to recognize this property, you may need to restart the dev server to [sync the module](/en/reference/cli-reference/#astro-sync).
```

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: Hmm okay, so let me reword that to be sure I understood correctly.

Users can do what is currently described in `queryString`, but they don't need to because Astro already handles this for them internally (with `toString()`) when using `actions.myAction` directly?

If so, I don't think there's a need to explain the internal behavior (i.e., mention `toString()`), but yes, using `actions.myAction` seems like a better DX and we should recommend that instead. So, I'm wondering... is there a better use case for `queryString` that we can document?

Because if we say you can use `queryString` but you don't need it, Astro already does that for you, this might be confusing... and I wonder why we expose `queryString`?

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: Since this is a reference, page, I'd be tempted to spell out something like

```suggestion
In order for Astro to recognize this property, you may need to restart the dev server or [run the `astro sync` command](/en/reference/cli-reference/#astro-sync) (`s + enter`).
```

This way, people don't *have* to click to another page to find out *how* to sync the module, but the link is there for more info if just knowing the command is sufficient for them. They both know which command they need, and the shortcut to do it but can also go find its specific reference documentation, too.

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -19,22 +19,26 @@ Actions help you build a type-safe backend you can call from client code and HTM
...
 ### `defineAction()`
 
 <p>
-<Since v="4.15.0" />
+
+**Type:** <code>(\{ accept, input, handler \}) => <a href="#actionclient">ActionClient</a></code>
 </p>
 
-The `defineAction()` utility is used to define new actions from the `src/actions/index.ts` file. This accepts a [`handler()`](#handler-property) function containing the server logic to run, and an optional [`input`](#input-validator) property to validate input parameters at runtime.
+An utility to define new actions in the `src/actions/index.ts` file. This accepts a [`handler()`](#handler-property) function containing the server logic to run, and an optional [`input`](#input-validator) property to validate input parameters at runtime.
```
sarah11918:
```suggestion
A utility to define new actions in the `src/actions/index.ts` file. This accepts a [`handler()`](#handler-property) function containing the server logic to run, and an optional [`input`](#input-validator) property to validate input parameters at runtime.
```
You are BOTH French, and cannot be trusted with English indefinite articles lol

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -123,83 +135,148 @@ export const server = {
...
 
 <p>
 
-**Type:** <code>(error?: unknown | <a href="#actionerror">ActionError</a>) => boolean</code><br/>
-<Since v="4.15.0" />
+**Type:** `(error?: unknown) => boolean`
 </p>
 
-The `isInputError()` utility is used to check whether an `ActionError` is an input validation error. When the `input` validator is a `z.object()`, input errors include a `fields` object with error messages grouped by name.
+An utility used to check whether [an `ActionError`](#actionerror) is an input validation error. When the `input` validator is a `z.object()`, input errors include a `fields` object with error messages grouped by name.
```
sarah11918:
```suggestion
A utility used to check whether [an `ActionError`](#actionerror) is an input validation error. When the `input` validator is a `z.object()`, input errors include a `fields` object with error messages grouped by name.
```

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -123,83 +135,148 @@ export const server = {
...
 
 <p>
 
-**Type:** <code>(error?: unknown | <a href="#actionerror">ActionError</a>) => boolean</code><br/>
-<Since v="4.15.0" />
+**Type:** `(error?: unknown) => boolean`
 </p>
 
-The `isActionError()` utility is used to check whether your action raised an `ActionError` within the [handler property](/en/reference/modules/astro-actions/#handler-property). This is useful when narrowing the type of a generic error in a `try / catch` block.
+An utility used to check whether your action raised [an `ActionError`](#actionerror) within the [handler property](#handler-property). This is useful when narrowing the type of a generic error.
```
sarah11918:
```suggestion
A utility to check whether your action raised [an `ActionError`](#actionerror) within the [handler property](#handler-property). This is useful when narrowing the type of a generic error.
```

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -241,14 +316,41 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
+The name of the action. Useful to track the source of an action result during a redirect.
+
+##### `handler()`
+
+<p>
+
+**Type:** <code>() => Promise\<<a href="#saferesult">SafeResult</a>\></code>
+</p>
+
+A method allowing you to programmatically call the action to get the result.
```
sarah11918:
```suggestion
A method to programmatically call an action to get the result.
```
I *think* (though I didn't look at the broader context so I could be wrong) this is maybe to call *an/any* generic action? When you say *the* action, there must be one in question you're referring to.

It's much more common to use "a/an" unless you can definitively point to the specific thing this refers to in context. (Like, a config option 2 or 3 levels deep, where you're saying it refers to *the* main config option in question.) So, I might be wrong, but this is how I'd decide which to use here!

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -272,10 +374,10 @@ export const onRequest = defineMiddleware(async (context, next) => {
 
 <p>
 
-**Type:** `(result: SafeResult<any, any>) => SerializedActionResult`
+**Type:** <code>(res: <a href="#saferesult">SafeResult</a>) => SerializedActionResult</code>
 </p>
 
-`serializeActionResult()` will serialize an action result to JSON for persistence. This is required to properly handle non-JSON return values like `Map` or `Date` as well as the `ActionError` object.
+Serializes an action result to JSON for persistence. This is required to properly handle non-JSON return values like `Map` or `Date` as well as the `ActionError` object.
```
sarah11918, line 341: Such nice editing choices! 🙌

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -297,20 +399,20 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
 
 <p>
 
-**Type:** `(action: ActionClient<any, any, any>) => string`
+**Type:** <code>(action: <a href="#actionclient">ActionClient</a>) => string</code>
 <Since v="5.1.0" />
 </p>
 
-The `getActionPath()` utility accepts an action and returns a URL path so you can execute an action call as a `fetch()` operation directly. This allows you to provide details such as custom headers when you call your action. Then, you can [handle the custom-formatted returned data](/en/guides/actions/#handling-returned-data) as needed, just as if you had called an action directly.
+An utility that accepts an action and returns a URL path so you can execute an action call as a `fetch()` operation directly. This allows you to provide details such as custom headers when you call your action. Then, you can [handle the custom-formatted returned data](/en/guides/actions/#handling-returned-data) as needed, just as if you had called an action directly.
```
sarah11918:
```suggestion
A utility that accepts an action and returns a URL path so you can execute an action call as a `fetch()` operation directly. This allows you to provide details such as custom headers when you call your action. Then, you can [handle the custom-formatted returned data](/en/guides/actions/#handling-returned-data) as needed, just as if you had called an action directly.
```

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -344,3 +446,170 @@ navigator.sendBeacon(
 )
 </script>
 ```
+
+### `ACTION_QUERY_PARAMS`
```
sarah11918, line 411: This was really well-written, with a helpful use-case/example!

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -344,3 +446,170 @@ navigator.sendBeacon(
...
+<p>
+
+**Types:**
+* <code>(input?: any) => Promise\<<a href="#saferesult">SafeResult</a>\></code>
+* `{ queryString?: string; orThrow: (input?: any) => Promise<Awaited<TOutput>>; }`
+</p>
+
+Represents an action to be called on the client. You can use it as a function that accepts input data and returns a Promise with a [`SafeResult` object](#saferesult) containing the action result or validation errors.
+
+The following example handles error when incrementing the like count fails:
```
sarah11918:
```suggestion
The following example shows how you can provide error handling with an `if` statement when incrementing the like count fails:
```
I had to be careful not to say that the following code *handles the error*, because I don't think `/* handle errors */` counts! 😆

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: Haha, this totally slipped my mind! 👀

For my defense, I was taught that you should use "an" before a vowel... it was only "recently" that I discovered that this was wrong and that it was a sound issue. Wrong habits... 😅

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: This totally makes sense, and I would say the same in French. Thanks for noticing this!

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: Thank you! This made me notice my example does not work. 😅 (`link` is not defined) I'll commit a fix with your suggestions.

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -344,3 +446,170 @@ navigator.sendBeacon(
...
+
+This constant can be useful when you need to clean up URLs after a form submission. For example, you might want to remove action-related query parameters during a redirect:
+
+```ts title="src/pages/api/contact.ts" "ACTION_QUERY_PARAMS"
+import type { APIRoute } from "astro";
+import { ACTION_QUERY_PARAMS } from 'astro:actions'
+
+export const GET: APIRoute = ({ params, request }) => {
+  request.url.searchParams.delete(ACTION_QUERY_PARAMS.actionName);
+  request.url.searchParams.delete(ACTION_QUERY_PARAMS.actionPayload);
```
ArmandPhilippot: Declares `link` to make the code snippet valid:
```suggestion
export const GET: APIRoute = ({ params, request }) => {
  const link = request.url.searchParams;
  link.delete(ACTION_QUERY_PARAMS.actionName);
  link.delete(ACTION_QUERY_PARAMS.actionPayload);
```

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: Yeah, I wasn't sure how to describe that. 😆 Thanks, this is much better!

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre: Re-reading my suggestion in the issue, I'm not sure adding an error message works. I think it's better to talk about the null behavior

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre: I think `queryString` has something to do with the React 19 integration, can't remember the details tbh. But yeah maybe we shouldn't document it then because honestly the `toString()` behavior is better IMO

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: I'm not sure Sarah will be happy 😄  I added a caution because I couldn't figure how to integrate this in the current content. But we can think about this later.
I checked [the implementation](https://github.com/withastro/astro/blob/8ebc0d79756a73741ac01c4a09a864b60bb7207f/packages/astro/src/actions/runtime/server.ts#L213-L215) and I think this is correct, but let me know if I missed anything or how to make it more accurate!

(the caution is a little further down, `l103`)

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: Yeah, I saw in the implementation PR mentions of React but I couldn't find why this was needed. This code snippet was taken from the changeset. I didn't see anything in the RFC that could help...

If find it a bit annoying to document `orThrow` but not `queryString`. Because of Intellisense, users we'll see this property and might wonder what it is. So, I tried to came up with a slightly different example which I hope makes more sense!

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -344,3 +450,171 @@ navigator.sendBeacon(
...
+#### `queryString` property
+
+<p>
+
+**Type:** `string`
+</p>
+
+A string representation of the action that can be used to construct form action URLs. This can be useful when your form component is used in multiple places but you need to redirect to a different URL on submit.
+
+The following example uses `queryString` to construct a URL that will be used in the form `action` attribute:
```
ArmandPhilippot: Hmm, because we don't use `action` directly, maybe:
```suggestion
The following example uses `queryString` to construct a URL that will be passed to the form `action` attribute through a custom prop:
```

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: Don't think I didn't notice! 😅 

Reading this `accept` property, I'm *wondering* whether this is all in fact material for the guide, rather than reference? Like, a lot of what follows is about *using* the accept property, and not its type signature etc?

Then, maybe we don't need a "caution" to make this stand out, and it can be more like a regular sentence like in the validator above?

What do you think? This entry is rather long -- do you think it fits the "example in a reference" section, or should this be pulled out into a section about validating in the Actions guide?

ArmandPhilippot: Great call! Yes, this section is quite long and not really about `accept` so it makes sense to split it. I just pushed a commit with some changes:
* I moved most of the Zod behavior with `accept: "form"` to the Actions guide with a slight rewording to integrate it better
* I rewrote the `accept property` section in the reference to better integrate the sentence (no more caution 😁)

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -72,134 +76,168 @@ Return values are parsed using the [devalue library](https://github.com/Rich-Har
...
+**Default:** `json`
+</p>
 
-Extension functions including `.refine()`, `.transform()`, and `.pipe()` are also supported on the `z.object()` validator.
+Defines the format expected by an action:
+* Use `form` when your action accepts `FormData`.
+* Use `json`, the default, for all other cases.
 
-To apply a union of different validators, use the `z.discriminatedUnion()` wrapper to narrow the type based on a specific form field. This example accepts a form submission to either "create" or "update" a user, using the form field with the name `type` to determine which object to validate against:
+When your action accepts form inputs, the `z.object()` validator will automatically parse `FormData` to a typed object. All Zod validators are supported to validate your inputs. However, the output type is not guaranteed to match your `input` validator except when you use arrays or booleans. Any falsy values will be internally converted to `null`.
```
sarah11918:
```suggestion
When your action accepts form inputs, the `z.object()` validator will automatically parse `FormData` to a typed object. All Zod validators are supported to validate your inputs.

<ReadMore>Learn about [using validators with form inputs] in the Actions guide, including example usage and special input handling.</ReadMore>
```
What about just leaving ALL of that for the guide? It maybe feels more confusing here if the only real content is about special case/exceptions?

(If so, then delete the line below as well. I couldn't suggest this over a deleted line. Because GitHub.)

src/content/docs/en/guides/actions.mdx
```diff
@@ -301,6 +301,55 @@ export const server = {
 }
 ```
 
+### Using validators with form inputs
+
+When your action is [configured to `accept` form data](/en/reference/modules/astro-actions/#accept-property), you can use any Zod validators to validate your fields (e.g. `z.coerce.date()` for date inputs). Extension functions including `.refine()`, `.transform()`, and `.pipe()` are also supported on the `z.object()` validator.
```
sarah11918:
```suggestion
When your action is [configured to accept form data](/en/reference/modules/astro-actions/#accept-property), you can use any Zod validators to validate your fields (e.g. `z.coerce.date()` for date inputs). Extension functions including `.refine()`, `.transform()`, and `.pipe()` are also supported on the `z.object()` validator.
```
I wouldn't inline here since we're not specifically mentioning the property, just using the word "accept" (which happens to mirror the option's name) here in this sentence.

ArmandPhilippot:
Thanks @sarah11918 , I pushed a new commit with your two suggestions! I was a bit annoyed to have to repeat more or less the same information before, so I agree this makes more sense. 😄 I let you check the wording, but, yes, this looks better now!

sarah11918 (approved):
Love it! Thank you everyone! Let's Get That Money! 💰

ArmandPhilippot:
Thanks again everyone ❤️ , let's merge this! 🫡

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -72,134 +76,168 @@ Return values are parsed using the [devalue library](https://github.com/Rich-Har
...
+The following example creates a new `ActionError` when the user is not logged in:
+
+```ts title="src/actions/index.ts" {8-11} "ActionError"
+import { defineAction, ActionError } from "astro:actions";
+
+export const server = {
+  getUserOrThrow: defineAction({
+    accept: 'form',
+    handler: async (_, { locals }) => {
+      if (locals.user?.name !== 'florian') {
```
trueberryless, line 173: Sorry that I am little very late to the party and that I criticize here, but is this a good example for "Not authorized". Personally, I find it a little bit confusing that the user is unauthorized, when they are not "florian". I mean, I love such easter eggs in the docs, but I'm not sure if they are maybe too confusing here...

florian-lefebvre: Lol I completely missed that! I don't have strong opinions on this, I'll let others answer

ArmandPhilippot: I'm not an actions expert so I look at examples in the source code. 😅 I don't recall the exact example but I recall I adapt it a bit because the value seemed weird or something like that.

I guess we could update it with something like `locals.user?.role !== 'admin'` if you think this is better?

trueberryless: Sorry for the delayed answer, but personally I would find the admin check more fitting. However, as I think I might be the only one that was confused by it, we might as well just leave it and if more people complain, we can update it... Would that be fine for you?

ArmandPhilippot: Both work for me. But if you find this confusing, feel free to submit a PR to update the example! You may not be the only one, others might be hesitant to report this... So, as long as it's correct, an update can be helpful for others!

PR #12602, opened 2025-10-22
Refactor Starlight i18n config

HiDeoo:
#### Description (required)

This PR refactors the [Starlight i18n configuration](https://starlight.astro.build/reference/configuration/#locales) with 2 main objectives:

- Use the native Starlight i18n config approach, which is documented, rather than different levels of indirection generating such config.
- Make the configuration statically analyzable, e.g. for various tooling purposes, e.g. the [`Starlight Links` extension](https://github.com/HiDeoo/starlight-links).

_Unrelated: something I noticed but didn't want to do in this PR is that we can probably also disable the llms plugin during Astro's smoke tests._

- Closes #


delucis (approved):
Big stamp of Talking & Doc’ing approval! Thanks @HiDeoo 🙌

PR #12605, opened 2025-10-22
Add Railway guide to integrations

sarahkb125:
#### Description (required)

We are proposing a Railway guide - we have a template and guide on our site, but want to make it clear in the Astro docs.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/railway.mdx
+++ b/src/content/docs/en/guides/integrations-guide/railway.mdx
@@ -46,4 +46,4 @@ Once your Astro site is deployed on Railway:
 - **Monitor your deployment** - Use Railway's dashboard to monitor your site's performance and logs
 - **Set up custom domains** - Add your own domain
 - **Configure CI/CD** - Railway automatically deploys on every push to your main branch
-- **Scale as needed** - Railway automatically handles horizontal and vertical scaling based on your traffic
+- **Scale as needed** - Configure horizontal scaling with traffic as needed


src/content/docs/en/guides/integrations-guide/railway.mdx
```diff
@@ -0,0 +1,49 @@
+---
+type: integration
+title: '@astrojs/railway'
```
matthewp, line 3: There is no npm package for railway, so should this title be something else?

sarahkb125: what would you suggest? all the other integrations are of this format - happy to make it just "Astro on Railway" if you think that makes more sense.

ArmandPhilippot: The other integrations are owned by Astro and that's why they use this format.

I haven't read in details the guide at the moment, but it sounds like this is not an integration but a deploy guide? We already have a [page for Railway](https://docs.astro.build/en/guides/deploy/railway/). So, if I'm right, you might want to update the following file instead: https://github.com/withastro/docs/blob/main/src/content/docs/en/guides/deploy/railway.mdx

sarah11918:
Hi @sarahkb125 !

As Armand said, we already have a deployment guide for Railway in Astro docs that you can see here: https://docs.astro.build/en/guides/deploy/railway/

If you feel the need to update that content, you are welcome to make a PR to that page.

"Integrations" guide pages are only for official Astro built and maintained npm packages that are added to existing projects using an `astro add` command (e.g. `astro add react`). The place for content about deploying to external, third-party services is in our `Third-party services` menu and the Railway guide already exists.

Therefore, I'm closing this PR, but welcome you to make any adjustments to the Railway deployment guide if that needs updating!

PR #12616, opened 2025-10-24
[i18nIgnore] Localized v6 link failures related to Astro.glob()

sarah11918:
This addresses some link failures in translations on the v6 branch because:

- we recently removed a bunch of translated pages, so some other links in non-removed pages had localized slugs
- `Astro.glob()` is removed in v6, so some links that passed in current v5 docs (where it's still documented, though as deprecated) fail in v6 docs where this is removed entirely.


ArmandPhilippot (approved):
I was going to say we could just remove the "See also:" from the translated errors because they seem up to date, but they translate the error itself so, yeah, it's probably better to remove them entirely!

LGTM!

sarah11918:
Alright, thanks for the reviews! Let's watch some dominoes (hopefully) fall!

PR #12624, opened 2025-10-25
i18n(fr): update `integrations-guide/netlify.mdx`

ArmandPhilippot:
#### Description (required)

Adds changes from #12606 to the French translation of `integrations-guide/netlify`.

I wasn't totally sure how to translate `skew protection`. I checked some others docs and except Amazon's AWS documentation, I couldn't find anything. They use two different translations which sounds both pretty weird (`protection antiasymétrique` and `protection antioblique`) so I suspect this is auto-translated.
Since `skew protection` is related to `version skew` I think what I came up with makes sense but feel free to tell otherwise!

- Suggested label: i18n

### DIFF
--- a/src/content/docs/fr/guides/content-collections.mdx
+++ b/src/content/docs/fr/guides/content-collections.mdx
@@ -420,6 +420,8 @@ const { Content, headings } = await render(entry);
 <Content />
 ```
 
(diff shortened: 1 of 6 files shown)


thomasbnt (approved):
lgtm

PR #12625, opened 2025-10-25
perf: exclude llms plugin during smoke tests

trueberryless:
#### Description (required)

Disable the llms plugin during Astro's smoke tests (as @HiDeoo hints in #12602)
And fix a small typo.

Performance improvement (deleted `.cache/` and `.astro/` folder before both smoke runs):

| Before (with llms plugin)     | After (without llms plugin)      |
| image | image |

_I also noticed that the `▶ @astrojs/starlight/routes/static/index.astro` route builds for all languages no matter if `PUBLIC_TWO_LANG` is set to `true` or not, probably because its config (where it fetches the docs collection) is probably not configurable here. But that is just my assumption, I didn't look more closely into it_

- "Follow-up" #12602
- Suggested label: low effort

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -69,7 +69,10 @@ export default defineConfig({
 				},
 			],
 			disable404Route: true,
-			plugins: [starlightPluginSmokeTest(), starlightPluginLlmsTxt()],
+			plugins: [
+				starlightPluginSmokeTest(),
+				...(process.env.PUBLIC_TWO_LANG ? [] : [starlightPluginLlmsTxt()]),
+			],
 		}),
 		sitemap(),
 	],
--- a/src/languages.ts
+++ b/src/languages.ts
@@ -4,7 +4,7 @@ import { localesConfig, twoLocalesConfig } from '../config/locales';
 /** The current Starlight i18n configuration depending if we are running smoke tests or not. */
 const currentLocalesConfig = process.env.PUBLIC_TWO_LANG ? twoLocalesConfig : localesConfig;
 
-/** An array of all language currently configured. */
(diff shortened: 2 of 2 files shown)


HiDeoo:
Thanks for the PR.

We actually discussed that possibility when reviewing #12602 and deciced against it:

- This can technically be considered as part of the smoke test.
- It does not add that much time to the overall smoke test run.

> builds for all languages

This is expected in this case, but the Astro's smoke test does an extra step removing all translations except the expected ones: https://github.com/withastro/astro/blob/main/.github/workflows/ci.yml#L242-L243

Considering all this, I'll be closing this PR but thanks notheless for the contribution!

trueberryless:
Thank you so much for explaining the reasoning, HiDeoo! The decision definitely makes sense. 🚀

trueberryless:
@HiDeoo One thing I noticed was that we should maybe refactor some of the CIs because they still include `5.0.0-beta` and I am not sure if this is still needed: https://github.com/withastro/astro/blob/dacebaf8df6293558091a18123aed966e4e57415/.github/workflows/ci.yml#L226-L230

So a revert of https://github.com/withastro/astro/commit/7bcae4e3eea9aecc3480881088e334a77ae5c3e3 or maybe adaptation for Astro 6 could be useful? 🤔

HiDeoo:
This is mostly kept as an example as this doesn't hurt to have it here when not needed.

It's also have already been [updated in the `next` branch](https://github.com/withastro/astro/blob/next/.github/workflows/ci.yml#L226) to match [Astro Docs `v6` branch](https://github.com/withastro/docs/tree/v6).

PR #12637, opened 2025-10-28
feat: add experimetal svgo docs

azat-io:
#### Description (required)

Adds documentation for the new experimental SVG optimization feature.

Related to: withastro/astro#13880

- Added new page at `/reference/experimental-flags/svg/`
- Includes configuration examples and common use cases
- Documents SVGO integration and error handling

N/A.


astro.sidebar.ts
```diff
@@ -152,6 +152,7 @@ export const sidebar = [
 					'reference/experimental-flags/static-import-meta-env',
 					'reference/experimental-flags/chrome-devtools-workspace',
 					'reference/experimental-flags/fail-on-prerender-conflict',
+					'reference/experimental-flags/svg',
```
sarah11918:
```suggestion
					'reference/experimental-flags/svg-optimization',
```
Even if the experimental flag name doesn't change, this page should still be precisely named because we have other docs content about SVGs in general. (And, this will mean updating the file name of the reference page!)

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,187 @@
...
+        ]
+      }
+    }
+  }
+})
+```
+
+## Usage
+
+Once enabled, SVG optimization will automatically apply to all SVG files imported in your project:
```
sarah11918: Question: does this work if you import SVGs, but then use an `<svg>` element? Or does this only work if you happen to use the SVG component pattern shown below? Is it more about any imported file, or does it further depend on how it's used?

src/content/docs/en/reference/experimental-flags/svg.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+## Error handling
+
+If SVGO optimization fails for any reason, Astro will gracefully fall back to using the original, unoptimized SVG content. A warning will be logged to the console, but your build will continue without errors.
+
+## Performance considerations
+
+SVG optimization happens during the build process, not at runtime. This means:
+
+- **Development:** SVGs are not optimized during development for faster rebuild times
```
sarah11918: This doesn't feel like a cohesive list of three related things: dev, prod, then... something else? And not sure Performance Considerations will be the right heading, since that implies "Things you have to think about to manage performance" but really, this is just "how it works." Will think about this one!

src/content/docs/en/reference/experimental-flags/svg.mdx
```diff
@@ -0,0 +1,187 @@
...
+- **Production:** All SVGs are optimized once during the build
+- **No runtime overhead:** Optimized SVGs are served as static assets
+
+The optimization process can increase build times slightly, but results in smaller file sizes and faster page loads for your users.
+
+## Migration notes
```
sarah11918: I don't think this section is necessary at all. This is how all experimental features go, and most importantly, we don't want to call our shot by guessing what this feature will look like by the time it's stable. 

(If you look at my comments in the implementation PR, I'm still not even sure how it will look when the PR merges! 😂 )

sarah11918 (commented):
Thanks for this docs PR @azat-io ! It's wonderfully well-written and a pleasure to read!

I made some initial comments/thoughts here, because I still want some questions answered in the other PR first before we look at this one too closely. Just want to make sure we're clear on the feature before we hit the docs too hard, but it's now a work in progress! 🙌

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
azat-io: This is for components

astro.sidebar.ts
azat-io: Done ✅️️️️

src/content/docs/en/reference/experimental-flags/svg.mdx
azat-io: Done ✅️️️️

src/content/docs/en/reference/experimental-flags/svg.mdx
azat-io: Done ✅️️️️

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+```js title="astro.config.mjs"
+export default defineConfig({
+  experimental: {
+    svg: {
+      optimize: true,
+      svgoConfig: {
+        plugins: [
+          'preset-default',
+          {
+            name: 'removeHiddenElems',
```
sarah11918: I'm curious why this is required here, since looking at SVGO's docs, `removeHiddenElems` is included in `preset-default`. So why do we specifically need to mention and make this active?

Also from looking at their documentation, they show using `preset-default` to get that entire list by default (and then remove as you like) OR adding only individual plugins in an array. Does your feature have something comparable? Does it work if the plugins array is just a list of the individual plugins you *do* want, and does not include `preset-default`? 

We should have under `### svgoConfig` a `#### plugins` reference entry in either case, something like:

> #### `plugins`
> (type signature stuff)
> An array of [SVGO plugins](link to all the plugins) that will be used to optimize your SVG component imports.
> 
> This can include any plugins by ID name, including SVGO's `preset-default` collection of plugins. A plugin can optionally be passed as an object including both its `name` and `active` status, to enable or disable as necessary.

Then, here is also where we need to explain any other options that are not passed in the `plugins` array:

> #### Other configuration options
>
> You can also pass [other SVGO configuration options](ideally a link to where we can see these??), such as `floatingPoint` and ..., directly to your config object.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### `svgoConfig`
+
+**Type:** `Config` (SVGO configuration object)<br />
+**Default:** `{}`
+
+Configuration object passed directly to SVGO for customizing SVG optimization.
+
+See [SVGO documentation](https://svgo.dev/docs/preset-default/) for available options and plugins.
```
sarah11918: This link only appears to go to information about the `preset-default` and not plugins in general, nor any other available options. So, I'm wondering if https://svgo.dev/docs/plugins/ is maybe a better link to lead people to for plugins info. 
```suggest
This allows you to further control Astro's SVG optimization with [SVGO plugins](https://svgo.dev/docs/plugins/).
```

And also, I'm not sure exactly how to describe these "other options." Later, you show using `floatPrecision` outside of the plugins array and it's not obvious where that comes from, or what else is available to use. (Is that maybe how you pass paramaters to a plugin? In the SVGO documentation, it appears that `floatPrecision` is a parameter of a plugin, and not a separate thing itself.)

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+Control the precision of numeric values in path data:
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  experimental: {
+    svg: {
+      optimize: true,
+      svgoConfig: {
+        floatPrecision: 2
```
sarah11918: As mentioned above, this `svgoConfig` option appears to come out of nowhere. Is there a definitive list of everything you can pass to `svgoConfig`? This example shows no SVGO plugins being used at all, but I can't find in SVGO documentation a list of options that you can configure that are not related to plugins. We would need to document for people above in the `svgoConfig` section all the options you can pass to it.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+- In **development mode**, SVG files are not optimized to ensure faster rebuild times and a smoother development experience.
+- In **production builds**, all imported SVG files are optimized once during the build process, resulting in smaller file sizes.
+- There is **no runtime overhead** - optimized SVGs are served as pre-processed static assets.
+
+While the optimization process may slightly increase your build times, the result is smaller file sizes and faster page loads for your users.
+
+## Further reading
+
+- [SVGO documentation](https://svgo.dev/)
+- [SVGO preset-default plugins](https://svgo.dev/docs/preset-default/)
```
sarah11918:
```suggestion
```
Since we'll mention above a link to plugins in context, I think just the main documentation link is fine here.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+import Since from '~/components/Since.astro'
+
+<p>
+
+**Type:** `object`<br />
+**Default:** `undefined`<br />
+<Since v="5.8.0" />
+</p>
+
+This experimental feature enables automatic optimization of SVG assets using [SVGO](https://svgo.dev/) during build time.
```
sarah11918:
```suggestion
This experimental feature enables automatic optimization of your [SVG components](/en/guides/images/#svg-components) using [SVGO](https://svgo.dev/) during build time.
```
Just checking that this is true? Even without the `svgoConfig` object, SVGO *is* being used?

Also, making it clear that this feature is associated specifically with SVG components!

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+When enabled, all imported SVG files will be optimized for smaller file sizes and better performance while maintaining visual quality. This can significantly reduce the size of your SVG assets by removing unnecessary metadata, comments, and redundant code.
```
sarah11918:
```suggestion
When enabled, your imported SVG files used as components will be optimized for smaller file sizes and better performance while maintaining visual quality. This can significantly reduce the size of your SVG assets by removing unnecessary metadata, comments, and redundant code.
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+        ]
+      }
+    }
+  }
+})
+```
+
+## Usage
+
+Once enabled, SVG optimization will automatically apply to all SVG files imported in your project:
```
sarah11918:
```suggestion
No change to using SVG components is required to take advantage of this feature. With experimental `svg.optimization` enabled, all your SVG component import files will be automatically optimized:
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+```astro title="src/pages/index.astro"
+---
+import Logo from '../assets/logo.svg';
+---
+
+<Logo />
+```
+
+The SVG will be optimized during the build process, resulting in smaller file sizes in your production build.
```
sarah11918:
```suggestion
If SVGO optimization fails for any reason, Astro will gracefully fall back to using the original, unoptimized SVG content. A warning will be logged to the console, but your build will continue without errors.

Note that this optimization applies to every SVG component import in your project. It is not possible to opt out on a per-component basis.
```
The previous line was true, but already said! Since this section is focused on usage, we'd try to anticipate questions people might have about using it in their project. Like for example, can this be done on a per-component basis? (I'm assuming no, so I've written the above.) If you can think of any other kinds of questions people might have, this would be a good place to inform them of specific usage tips and tricks!

I don't think we need a specific heading about error handling because that implies there's something for people do to handle errors, vs just "here's what happens in case of an error." I think this content fits better right in here.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+      }
+    }
+  }
+})
+```
+
+## Error handling
+
+If SVGO optimization fails for any reason, Astro will gracefully fall back to using the original, unoptimized SVG content. A warning will be logged to the console, but your build will continue without errors.
+
```
sarah11918:
```suggestion
```
Suggest removing in favour of this line moving up into `## Usage` I think!

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+export default defineConfig({
+  experimental: {
+    svg: {
+      optimize: true
+    }
+  }
+})
+```
+
+## Configuration
```
sarah11918, line 49: I'd like to see this entire configuration section move down below `## Usage` so that people can immediately see "enable, then here's what affect it has on my existing project." I feel like a lot of people might turn this on and use it, then figure out whether they need or want to tweak customization later, so I think gets important information about what this means for your project and existing code up earlier!

It feels like a good flow here would be:

```
## Overview
## Usage
## Configuration
### optimize
### SVGO config
## Common Use Cases
(sub-headings)
## How it works
## Further Reading (maybe optional even? We already link to both the main and the docs site from this page. A *good* addition here could be if you know of a really awesome/canonical article on optimizing SVGs.)
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+---
+import Logo from '../assets/logo.svg';
+---
+
+<Logo />
+```
+
+The SVG will be optimized during the build process, resulting in smaller file sizes in your production build.
+
+## Common use cases
```
sarah11918, line 109:
```suggestion
## Common use cases 

SVGO provides an extensive [default plugin list](https://svgo.dev/docs/preset-default/) with opinionated optimizations that is more convenient than adding each plugin individually. However, you may need to customize it further for your needs. For example, it may remove items or clean up too aggressively for your situation.
```
We always prefer to have some text after a heading to explain what this section will focus on. So, I just invented the above which you are welcome to make accurate, but something *like* this should go here.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+### Preserve specific attributes
+
+You may want to preserve certain SVG attributes that SVGO removes by default:
```
sarah11918:
```suggestion
You may want to preserve certain SVG attributes, such as the `viewBox`, that SVGO removes by default:
```
Is this correct to say?

azat-io:
1. Restructured document: moved Usage before Configuration
2. Updated terminology: "SVG assets" → "SVG components" with link to guide
3. Enhanced Usage section: added error handling info, noted no per-component opt-out
4. Added `svgoConfig` documentation: `plugins` subsection and other options subsection
5. Updated SVGO link: from preset-default to general plugins page
6. Enhanced "Common use cases": added intro paragraph, replaced `removeHiddenElems` with `removeMetadata`
7. Simplified "Further reading": removed redundant links

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,225 @@
...
+        ]
+      }
+    }
+  }
+})
+```
+
+#### Other configuration options
+
+You can also pass [other SVGO configuration options](https://svgo.dev/docs/configuration/), such as `floatPrecision` and `multipass`, directly to your config object:
```
sarah11918: This is exactly what I'd want, but the link is 404'ing. So we'll need to find wherever this full configuration is listed! We prefer not to, but if we have to, we can even link to a file/specific lines in their repo where these options are listed. A docs link is better, but we absolutely need to make sure people can find what these options are, one way or another!

sarah11918:
This looks fantastic @azat-io , thank you! Just one broken link here, as mentioned, but then I think this is great!

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
azat-io: Done!

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,181 @@
...
+        }
+      ]
+    }
+  }
+})
+```
+
+### Other configuration options
+
+You can also pass [other SVGO configuration options](https://github.com/svg/svgo#configuration), such as `floatPrecision` and `multipass`, directly to your config object:
```
sarah11918:
```suggestion
You can also pass [other SVGO configuration options](https://github.com/svg/svgo/blob/66d503a48c6c95661726262a3068053c429b06a9/lib/types.ts#L335), such as `floatPrecision` and `multipass`, directly to your config object:
```
That link *still* didn't annotate the options available. I think this is the line in the code that mentions everything that is possible to put in your configuration. Assuming that they can configure all of these things, then let's use this link! (It's funny that I couldn't find this complete information anywhere in their documentation. How are people supposed to know what they can configure? 😄 )

azat-io: Done

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,181 @@
...
+i18nReady: true
+---
+
+import Since from '~/components/Since.astro'
+
+<p>
+
+**Type:** `boolean | object`<br />
+**Default:** `false`<br />
+<Since v="5.8.0" />
```
sarah11918:
```suggestion
<Since v="5.16.0" />
```
Is this coming in the next minor release @florian-lefebvre , or waiting for Astro 6?

florian-lefebvre: Coming in the minor!

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,181 @@
...
+---
+import Logo from '../assets/logo.svg';
+---
+
+<Logo />
+```
+
+The SVG will be optimized during the build process, resulting in smaller file sizes in your production build.
+
+If SVGO optimization fails for any reason, Astro will gracefully fall back to using the original, unoptimized SVG content. A warning will be logged to the console, but your build will continue without errors.
```
sarah11918: @florian-lefebvre Whatever we decide in https://github.com/withastro/astro/pull/13880 , needs to match here too!

sarah11918:
```suggestion
```
No longer true!

sarah11918 (approved):
LGTM! Thank you for all your work and responsiveness getting these docs done, @azat-io ! Assuming nothing about this feature changes, these docs will be ready to merge when 5.16.0 is released! 🥳

sarah11918:
@yanthomasdev - feel free to review at any time!

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -0,0 +1,179 @@
...
+<Since v="5.16.0" />
+</p>
+
+This experimental feature enables automatic optimization of your [SVG components](/en/guides/images/#svg-components) using [SVGO](https://svgo.dev/) during build time.
+
+When enabled, your imported SVG files used as components will be optimized for smaller file sizes and better performance while maintaining visual quality. This can significantly reduce the size of your SVG assets by removing unnecessary metadata, comments, and redundant code.
+
+To enable this feature with default settings, set it to `true` in your Astro config:
+
+```js title="astro.config.mjs" ins={4}
```
ArmandPhilippot: wrong line:
````suggestion
```js title="astro.config.mjs" ins={5}
````

File: src/content/docs/en/reference/experimental-flags/svg-optimization.mdx [hunk 22]
```diff
@@ -0,0 +1,179 @@
...
+<Logo />
+```
+
+The SVG will be optimized during the build process, resulting in smaller file sizes in your production build.
+
+Note that this optimization applies to every SVG component import in your project. It is not possible to opt out on a per-component basis.
+
+## Configuration
+
+You can pass a [SVGO configuration object](https://svgo.dev/docs/plugins/) to customize optimization behavior:
```
ArmandPhilippot: This doesn't show the configuration object? We should probably use the same link as in "Other configuration options":
```suggestion
You can pass a [SVGO configuration object](https://github.com/svg/svgo/blob/66d503a48c6c95661726262a3068053c429b06a9/lib/types.ts#L335) to customize optimization behavior:
```

PR #12322, opened 2025-09-15
Astro v6.0

sarah11918:
The branch for your docs for all Astro v6.0 PRs! (Astro Milestone: https://github.com/withastro/astro/milestone/37)

Deploy preview: https://v6.docs.astro.build/en/guides/upgrade-to/v6/

### Breaking changes (additions, removals, deprecations, stablizings...)
File to edit: v6 upgrade page: `src/content/docs/en/guides/upgrade-to/v6.mdx`

#### Use v5 upgrade guide as an example of format: 
- Deployed: https://docs.astro.build/en/guides/upgrade-to/v5/  
- Source file: `src/content/docs/en/guides/upgrade-to/v5.mdx`

#### Template

(More guidance available in the [Astro Docs Docs guide to breaking changes](https://contribute.docs.astro.build/upgrade-guides/breaking-changes/))

```
<SourcePR number="11111" title="PR title"/>

### [changed/removed/added/deprecated]: feature/name/title/description
In Astro v5.x, [ a statement in the past tense about what Astro did/used to do.
e.g.  bundled all assets into several different folders....]

Astro v6.0 [ a statement in the present tense about how Astro works now.
e.g. now bundles all assets into one single folder located in... ]

#### What should I do?
// Active sentences in the imperative (command) tense
// include diff code examples whenever appropriate

e.g. Update... / Remove... / Replace... / Create...

<ReadMore>Learn more about [relevant feature/topic](/en/.../...).</ReadMore>
```

### To-do

- [x] Merge #13363  to add redirects for experimental flag pages that no longer exist
- [x] Merge #13366 to remove temporary v6 warning 
- [x] Set up a `v5.docs.astro.build` subdomain to #13369 
- [x] Merge #13367 to uncomment out line 17 re: v5 docs, add to list of unmaintained older docs
- [x] Merge https://github.com/withastro/astro/pull/15732/ to update links in core repo to point to regular docs, not `v6`: JSDoc for config reference, error messages, changelog
- [x] (THEN - run Nightly Action to pull reference/error messages into v6 and merge #13370 and #13371 )
- [x] MERGE THIS PR with a LUNARIA STATEMENT from @yanthomasdev 

### Immediately after merging this PR
- [x] Merge #13372  to revert d3883e7e4d8961c41ea895cb2e64335948c34f81 before merging to `main`
- [x] Merge #13373  to pull astro main branch JSDoc into docs main branch
[...]

PR #12639, opened 2025-10-29
feat(reference): Update astro:assets reference for new Astro 6.0 behavior

Princesseuh:
#### Description (required)

In Astro 6, you can crop images and set fit and position without needing a layout. I'll admit, I don't fully understand how layouts work since I wasn't around when it was added, but hopefully I got it.

https://github.com/withastro/astro/pull/14629

### DIFF
--- a/src/content/docs/en/reference/modules/astro-assets.mdx
+++ b/src/content/docs/en/reference/modules/astro-assets.mdx
@@ -394,8 +394,6 @@ import myImage from '../assets/my_image.png';
 <Image src={myImage} alt="This will disable responsive images" layout="none" />
 ```
 
-###### Responsive Image Styles
-
 The value for `layout` also defines the default styles applied to the `<img>` tag to determine how the image should resize according to its container:
 
 ```css title="Responsive Image Styles"


florian-lefebvre:
You need to update the upgrade guide as well, see https://github.com/withastro/docs/pull/12322

sarah11918:
So based on this, if `fit` and `position` no longer require `layout`, then that "Responsive Images Properties" section probably isn't needed anymore, and `layout` itself could just be thrown up higher into the full list of image properties too, eh?

Like, this isn't true anymore:

>Setting the layout property on an `<Image />` or `<Picture />` component creates a responsive image and **enables additional property settings.**

That was the reason to have the section, so I'd suggest we just move the `layout` API reference up too, and remove that section.

Princesseuh:
I removed the section entirely. I'll try to think of something to put in the upgrade guide

sarah11918:
It's not going to let me suggest fixes for the links that are broken now that there is no more section called "Responsive Images."

In general, use the following ideas to fix these:
- occurrences of "responsive images" are just plain text, no link. They are a thing that exists, but they are (no longer) an *Astro* thing. So we use the phrase as regular terminology, but we don't document what they are.
- "responsive image properties" **no longer exist**. They are not a thing. They should not even be mentioned.

In *most* cases, just removing the link should be fine. In a case where e.g. it says "image properties AND responsive image properties", it's now just "image properties."

As for the `images.mdx` guide page that now has a broken link, I would remove the link and update the text there to be something like:

> With responsive image properties (`layout`, `fit`, `position`) applied to the `<Image />` or `<Picture />` components, ...

For the guide page, it's fine to be a little bit more explainy. These are generally known as properties you use when making images responsive.

If you have any doubt, when you update these, they will at least become changes I can make suggestions on top of!

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -322,104 +322,16 @@ fetchpriority="high"
...
-    loading="lazy"
-    decoding="async"
-  />
-</picture>
-```
-
-### Responsive image properties
-
-Setting the [`layout`](#layout) property on an [`<Image />`](#image-) or [`<Picture />`](#picture-) component creates a responsive image and enables additional property settings.
+Defines a responsive image and determines how the image should resize when its container changes size. Can be used to override the default configured value for [`image.layout`](/en/reference/configuration-reference/#imagelayout).
```
```suggestion
Determines how the image should resize when its container changes size. Can be used to override the default configured value for [`image.layout`](/en/reference/configuration-reference/#imagelayout).
```
This layout property no longer magically makes an image "responsive", so I think we can just say this!

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -715,6 +715,35 @@ If you need more control over environment variables in Astro, we recommend you u
 
 <ReadMore>Learn more about [environment variables](/en/guides/environment-variables/) in Astro, including `astro:env`.</ReadMore>
 
+### Changed: Cropping by default in default image service
+
+<SourcePR number="14629" />
```
```suggestion
<SourcePR number="14629" title="feat(assets): Always allow cropping and never upscale" />
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -715,6 +715,35 @@ If you need more control over environment variables in Astro, we recommend you u
...
+
+Astro's default image service now applies cropping by default without requiring setting the `fit` option.
```
florian-lefebvre: I think we usually have a sentence that says what was the behavior in v5, then another that explains the diff in v6

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -715,6 +715,35 @@ If you need more control over environment variables in Astro, we recommend you u
...
+import { Image } from 'astro:assets';
+import myImage from '../assets/photo.jpg';
+---
+<Image src={myImage} width={400} height={300} fit="contain" />
+<Image src={myImage} width={400} height={300} />
+```
+
+### Changed: Never upscale images in default image service
+
+<SourcePR number="14629" />
```
```suggestion
<SourcePR number="14629" title="feat(assets): Always allow cropping and never upscale" />
```

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre:
```suggestion
In Astro 5.0, the default image service would only apply cropping when the `fit` option was provided.

Astro 6.0 applies cropping by default without requiring setting the `fit` option.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -715,6 +715,35 @@ If you need more control over environment variables in Astro, we recommend you u
...
 
+### Changed: Cropping by default in default image service
+
+<SourcePR number="14629" title="feat(assets): Always allow cropping and never upscale" />
+
+Astro's default image service now applies cropping by default without requiring setting the `fit` option.
+
+#### What should I do?
+
+Normally no changes are needed. However, if you were previously setting `fit` to `contain` (its default value) in order to crop, you may now remove this option to get the same behavior:
```
```suggestion
No changes are needed to your existing cropped images as the `fit` property is still valid. However, if you were previously setting `fit` to `contain` (its default value) in order to crop your images, you may now remove this option and still achieve the same cropping behavior by specifying `width` and `height` alone:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -715,6 +715,35 @@ If you need more control over environment variables in Astro, we recommend you u
...
+---
+<Image src={myImage} width={400} height={300} fit="contain" />
+<Image src={myImage} width={400} height={300} />
+```
+
+### Changed: Never upscale images in default image service
+
+<SourcePR number="14629" title="feat(assets): Always allow cropping and never upscale" />
+
+In certain cases, Astro's default image service would upscale images when the requested dimensions were larger than the source image, which is often undesirable. This behavior has now been changed to never upscale images by default.
```
```suggestion
In Astro 5.0, the default image service would upscale images when the requested dimensions were larger than the source image.

Astro 6.0 removes this behavior: the default image service never upscales images.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -715,6 +715,35 @@ If you need more control over environment variables in Astro, we recommend you u
...
+
+#### What should I do?
+
+As upscaling images is typically undesirable, it is no longer possible to recreate the previous behavior in Astro itself. If you do need to upscale images, you may consider upscaling the images manually, or using a custom image service that supports upscaling.
```
```suggestion
Review your images and update dimensions as needed. If you do need to upscale images, you may consider upscaling the images manually or using a custom image service that supports upscaling.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -715,6 +715,35 @@ If you need more control over environment variables in Astro, we recommend you u
...
+
+<SourcePR number="14629" title="feat(assets): Always allow cropping and never upscale" />
+
+Astro's default image service now applies cropping by default without requiring setting the `fit` option.
+
+#### What should I do?
+
+Normally no changes are needed. However, if you were previously setting `fit` to `contain` (its default value) in order to crop, you may now remove this option to get the same behavior:
+
+```ts title="src/components/MyImage.astro" del={4} ins={5}
```
ArmandPhilippot: Fix highlighted lines:
````suggestion
```ts title="src/components/MyImage.astro" del={5} ins={6}
````

File: src/content/docs/en/guides/upgrade-to/v6.mdx (hunk 5 above)
Comment (sarah11918) [reply src/content/docs/en/guides/upgrade-to/v6.mdx]: (I updated Florian's suggestion!)

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 9]
```diff
@@ -715,6 +715,35 @@ If you need more control over environment variables in Astro, we recommend you u
...
+---
+import { Image } from 'astro:assets';
+import myImage from '../assets/photo.jpg';
+---
+<Image src={myImage} width={400} height={300} fit="contain" />
+<Image src={myImage} width={400} height={300} />
+```
+
```
sarah11918, line 738:
``````suggestion

If you were previously relying on omitting the `fit` property to avoid cropping, you can now prevent cropping by setting `fit="inside"`:

```ts title="src/components/MyImage.astro" del={5} ins={6}
---
import { Image } from 'astro:assets';
import myImage from '../assets/photo.jpg';
---
<Image src={myImage} width={400} height={300} />
<Image src={myImage} width={400} height={300}  fit="inside"/>
```

``````
I think from the implementation PR description this is also true? If so we should say something like this!

Comment (Princesseuh) [reply src/content/docs/en/guides/upgrade-to/v6.mdx:738]: It's the default value, so it wouldn't do anything to add it explicitly

File: src/content/docs/en/reference/modules/astro-assets.mdx
Comment (ArmandPhilippot) [inline src/content/docs/en/reference/modules/astro-assets.mdx:8]: nit: Maybe we should set the max level to 5? The `Responsive Image Styles` section looks a bit off to me in the table of contents:
`astro:assets` ToC

Comment (sarah11918) [reply src/content/docs/en/reference/modules/astro-assets.mdx:8]: I'm actually wondering whether that heading should be there at all now? Previously, we had a whole section for "responsive images" as we needed to be able to write about the group of properties enabled by setting a `layout`. Maybe now we don't need that heading at all as it's just the styles for the `layout` property?

(Otherwise, yes, changing the heading level makes sense). I think either solution works here.

File: src/content/docs/en/reference/modules/astro-assets.mdx [hunk 10]
```diff
@@ -322,13 +321,128 @@ fetchpriority="high"
...
+---
+import { Image } from 'astro:assets';
+import myImage from '../assets/my_image.png';
+---
+<Image src={myImage} alt="This will use constrained layout" width={800} height={600} />
+<Image src={myImage} alt="This will use full-width layout" layout="full-width" />
+<Image src={myImage} alt="This will disable responsive images" layout="none" />
+```
+
+###### Responsive Image Styles
```
```suggestion
```
We don't talk about "responsive images" anymore, so I think it makes sense to remove this heading. (And, will solve our TOC issue!)

D'oh, should remove another line around this, too.

```suggestion
```
We don't talk about "responsive images" anymore, so I think it makes sense to remove this heading. (And, will solve our TOC issue!)

PR #12642, opened 2025-10-29
chore: Remove Gitpod/Ona info

erbierc:
#### Description (required)

In Starlight, during a [PR](https://github.com/withastro/starlight/pull/3495) a discussion emerged about Gitpod and how it has become Ona, a pay-to-use AI-development platform. Gitpod was removed from the CONTRIBUTING.md file there and so I would like to continue with that for Astro docs as well.

- Closes #N/A
- Suggested label: 

Note: It would also be a good idea to remove any gitpod related files, such as [.gitpod.yml](https://github.com/withastro/docs/blob/main/.gitpod.yml).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,8 +101,7 @@ Our Docs are translated into several languages and we rely on automation to noti
 
 **Pages generated from outside sources**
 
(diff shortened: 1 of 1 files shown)


erbierc:
Done and done, absolutely take your time with the discussions!

PR #12643, opened 2025-10-30
i18n(pt-BR): Clarify zero JS statement in why-astro.mdx

thiagoprazeres:
Improved translation in "Zero JS, by default:"

- Closes #
- Suggested label:


yanthomasdev:
Thanks @thiagoprazeres, I appreciate making the PR! In this case, the sentence is correct and matches the original content. I see where your confusion came from, but the same could happen with this change by referring to another part of the sentence.

thiagoprazeres:
The sentence is incorrect and does not correspond to the original content. The sentence in Portuguese states that Astro makes your website slower, which makes no sense at all. Why would someone use a framework that slows down their website?

yanthomasdev:
Just to be clear, I had confirmed with another speaker, and they agreed the current translation is correct.

The original:
>Zero JS, by default: Less client-side JavaScript to slow your site down.

In the translation, "tornar seu site mais lento" refers to "JavaScript", as in, "there's less JavaScript to slow your site down" .

I hope that explains it.

PR #12647, opened 2025-10-31
Document new `ActionInputSchema` utility type

mauriciabad:
#### Description

Documentation for the new `ActionInputSchema` utility type I proposed in https://github.com/withastro/astro/pull/14698

I put the latest version +1 patch in the `<Since>` component, but I don't know the exact version yet.

Feel free to change the text if is not clear :)

I also native Spanish, if you want me to translate it.

### DIFF
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -152,6 +152,7 @@ export const sidebar = [
 					'reference/experimental-flags/static-import-meta-env',
 					'reference/experimental-flags/chrome-devtools-workspace',
 					'reference/experimental-flags/fail-on-prerender-conflict',
+					'reference/experimental-flags/svg-optimization',
 				],
 			}),
 			'reference/legacy-flags',
--- a/src/content/docs/en/reference/cli-reference.mdx
+++ b/src/content/docs/en/reference/cli-reference.mdx
@@ -206,9 +206,11 @@ Starts a local server to serve the contents of your static directory (`dist/` by
 
 This command allows you to preview your site locally [after building](#astro-build) to catch any errors in your build output before deploying it. It is not designed to be run in production. For help with production hosting, check out our guide on [Deploying an Astro Website](/en/guides/deploy/).
 
-Since Astro 1.5.0, the [Node adapter](/en/guides/integrations-guide/node/) supports `astro preview` for builds generated with on-demand rendering.
+The following hotkeys can be used in the terminal where the Astro preview server is running:
+- `o` + `enter` to open your Astro site in the browser.
+- `q` + `enter` to quit the preview server.
 
-Can be combined with the [common flags](#common-flags) documented below.
+The `astro preview` command can be combined with the [common flags](#common-flags) documented below to further control the preview experience.
 
 ## `astro check`
 
--- a/src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
+++ b/src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
@@ -1,0 +1,179 @@
+---
+title: Experimental SVG optimization
+sidebar:
+  label: SVG optimization
+i18nReady: true
+---
+
+import Since from '~/components/Since.astro'
+
+<p>
+
+**Type:** `boolean | object`<br />
+**Default:** `false`<br />
+<Since v="5.16.0" />
+</p>
+
+This experimental feature enables automatic optimization of your [SVG components](/en/guides/images/#svg-components) using [SVGO](https://svgo.dev/) during build time.
+
+When enabled, your imported SVG files used as components will be optimized for smaller file sizes and better performance while maintaining visual quality. This can significantly reduce the size of your SVG assets by removing unnecessary metadata, comments, and redundant code.
+
+To enable this feature with default settings, set it to `true` in your Astro config:
+
+```js title="astro.config.mjs" ins={5}
+import { defineConfig } from "astro/config"
+
+export default defineConfig({
+  experimental: {
+    svgo: true
+  }
+})
+```
+
+## Usage
+
+No change to using SVG components is required to take advantage of this feature. With experimental `svgo` enabled, all your SVG component import files will be automatically optimized:
+
+```astro title="src/pages/index.astro"
+---
+import Logo from '../assets/logo.svg';
+---
+
+<Logo />
+```
+
+The SVG will be optimized during the build process, resulting in smaller file sizes in your production build.
+
+Note that this optimization applies to every SVG component import in your project. It is not possible to opt out on a per-component basis.
+
+## Configuration
+
+You can pass a [SVGO configuration object](https://github.com/svg/svgo/blob/66d503a48c6c95661726262a3068053c429b06a9/lib/types.ts#L335) to customize optimization behavior:
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  experimental: {
+    svgo: {
+      plugins: [
+        'preset-default',
+        {
+          name: 'removeViewBox',
+          active: false
+        }
+      ]
+    }
+  }
+})
+```
+
+### `plugins`
+
+**Type:** `Array<string | PluginConfig>`<br />
+**Default:** `[]`
+
+An array of [SVGO plugins](https://svgo.dev/docs/plugins/) that will be used to optimize your SVG component imports.
+
+This can include any plugins by ID name, including SVGO's `preset-default` collection of plugins. A plugin can optionally be passed as an object including both its `name` and `active` status, to enable or disable as necessary.
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  experimental: {
+    svgo: {
+      plugins: [
+        'preset-default',
+        {
+          name: 'removeViewBox',
+          active: false
+        }
+      ]
+    }
+  }
+})
+```
+
+### Other configuration options
+
+You can also pass [other SVGO configuration options](https://github.com/svg/svgo/blob/66d503a48c6c95661726262a3068053c429b06a9/lib/types.ts#L335), such as `floatPrecision` and `multipass`, directly to your config object:
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  experimental: {
+    svgo: {
+      floatPrecision: 2,
+      multipass: true
+    }
+  }
+})
+```
+
+## Common use cases
+
+SVGO provides an extensive [default plugin list](https://svgo.dev/docs/preset-default/) with opinionated optimizations that is more convenient than adding each plugin individually. However, you may need to customize it further for your needs. For example, it may remove items or clean up too aggressively for your situation.
+
+### Preserve specific attributes
+
+You may want to preserve certain SVG attributes, such as the `viewBox`, that SVGO removes by default:
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  experimental: {
+    svgo: {
+      plugins: [
+        'preset-default',
+        {
+          name: 'removeViewBox',
+          active: false // Preserve viewBox attribute
+        }
+      ]
+    }
+  }
+})
+```
+
+### Remove specific elements
+
+You can configure plugins to remove specific unwanted elements like metadata or hidden layers. Note that many plugins are already included in `preset-default`, so you typically only need to configure their behavior:
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  experimental: {
+    svgo: {
+      plugins: [
+        'preset-default',
+        {
+          name: 'removeMetadata',
+          active: true
+        }
+      ]
+    }
+  }
+})
+```
+
+### Custom precision
+
+Control the precision of numeric values in path data:
+
+```js title="astro.config.mjs"
+export default defineConfig({
+  experimental: {
+    svgo: {
+      floatPrecision: 2
+    }
+  }
+})
+```
+
+## How it works
+
+SVG optimization happens during the build process, not at runtime:
+
+- In **development mode**, SVG files are not optimized to ensure faster rebuild times and a smoother development experience.
+- In **production builds**, all imported SVG files are optimized once during the build process, resulting in smaller file sizes.
+- There is **no runtime overhead** - optimized SVGs are served as pre-processed static assets.
+
+While the optimization process may slightly increase your build times, the result is smaller file sizes and faster page loads for your users.
+
+## Further reading
+
+- [SVGO documentation](https://svgo.dev/)


src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+
+### `ActionInputSchema`
+
+<p>
+
+**Type:** `ZodType`
+<Since v="5.15.4" />
+</p>
+
+A utility type that extracts the input's **zod schema type** from [an action handler](#defineaction). This can be useful if you need to reference an [action's validator]((#input-validator)) input type in your own type definitions.
```
sarah11918:
```suggestion
A utility type that extracts the input's **zod schema type** from [an action handler](#defineaction). This can be useful if you need to reference an [action's validator](#input-validator) input type in your own type definitions.
```
fix link syntax

sarah11918:
Thanks for submitting this documentation with your feature @mauriciabad ! I will also ask @florian-lefebvre to check in on these docs after everything is resolved in the Astro feature PR to make sure what's here matches the decisions made over there!

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
 Represents the result of an action call:
 * on success, `data` contains the output of the action and `error` is `undefined`.
 * on failure, `error` contains an [`ActionError`](#actionerror) with validation errors or runtime errors, and `data` is `undefined`.
+
+### `ActionInputSchema`
+
+<p>
+
+**Type:** `ZodType`
+<Since v="5.15.4" />
```
florian-lefebvre:
```suggestion
<Since v="5.16.0" />
```

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
 Represents the result of an action call:
 * on success, `data` contains the output of the action and `error` is `undefined`.
 * on failure, `error` contains an [`ActionError`](#actionerror) with validation errors or runtime errors, and `data` is `undefined`.
+
+### `ActionInputSchema`
+
+<p>
+
+**Type:** `ZodType`
```
florian-lefebvre: This is a generic type, IDK how we represent them usually here

ArmandPhilippot: Yeah, it's a bit tricky especially with Zod types. Currently `input` is typed as `ZodType | undefined`, so I think it's fine to use the same type here as long as the description links back to "input validator".

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+<Since v="5.16.0" />
+</p>
+
+A utility type that extracts the input's **zod schema type** from [an action handler](#defineaction). This can be useful if you need to reference an [action's validator](#input-validator) input type in your own type definitions.
```
sarah11918:
```suggestion
A utility type that extracts the input's Zod schema type from [an action handler](#defineaction). This can be useful to reference an [action's validator](#input-validator) input type in your own type definitions.
```
- no bold here, but would a link be helpful?

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+
+Even if the action [accepts `form`](#accept-property), `ActionInputSchema` lets you get the input type as `object` rather than [`FormData`](https://developer.mozilla.org/docs/Web/API/FormData).
```
sarah11918: It feels like this wording needs checking against the wording at `#accept-property` (does that wording need any updating?) The docs for `accept` say:

> When your action accepts form inputs, the z.object() validator will automatically parse FormData to a typed object. All Zod validators are supported to validate your inputs.

- is this still happening automatically?
- Does this happen *in addition* to you being able to get the input type as this new object?
- Do we need a link *up there* to this new entry to mention this option?
- It feels like maybe the link to `FormData` is best up there?

Just want to make sure that the docs are in sync with each other.

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+
+If the [`input` validator](#input-validator) is omitted, it returns `never`.
```
sarah11918: This says if `input` is not provided, "it" returns `never`. It's not clear to me what "it" is. (`ActionInputSchema`? Can it return `never`?) nor what `never` means in this context.

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+
+The following example uses `ActionInputSchema` on an action named `contact` to:
+* Retrieve the **zod schema type** for the input of the action.
```
sarah11918:
```suggestion
* Retrieve the Zod schema type for the input of the action.
```

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+* Retrieve the expected **input type** of the action's validator.
```
sarah11918:
```suggestion
* Retrieve the expected `input` type of the action's validator.
```

src/content/docs/en/reference/modules/astro-actions.mdx
florian-lefebvre: it means the return type of the `ActionInputSchema` generic type. and never means the `never` TS type

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: I don't think this means the `accept` docs is outdated. What happens when using `accept: "form"` is:
* we define a Zod schema to validate the data
* the action sends a `FormData` object
* the validator converts that to a different object shape
* `ActionInputSchema` can be used to retrieve the type of that different object shape

So basically, I'd say `ActionInputSchema` doesn't care about the input shape. All it's doing is extracting the type of the final object.

src/content/docs/en/reference/modules/astro-actions.mdx
ArmandPhilippot: Yeah, this means that what we are trying to do will `never` happen, the type can't be resolved. In this context, if the `input` validator is not defined, then this utility type can't resolve the `input` type because it doesn't exist and will "fall back" to `never`.

When `input` is not defined, if you assign `ActionInputSchema<typeof actions.contact>` to a variable and try to assign a value, Typescript will complain because no value is valid. This is a way to detect something is wrong. Here what's wrong is that `input` is not defined.

So, I'm not sure "returns `never`" is clear enough for people less familiar with Typescript. Maybe we should instead say something like: "This utility type  is only able to retrieve the type of an `input` when it is defined in your action.".

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+
+```astro title="src/components/Form.astro" {4}
```
ArmandPhilippot: Currently, we're highlighting an empty line:
````suggestion
```astro title="src/components/Form.astro" {5}
````

File: src/content/docs/en/reference/modules/astro-actions.mdx (hunk 4 above)
Comment (mauriciabad) [reply src/content/docs/en/reference/modules/astro-actions.mdx]: My reasoning to add the bold there was to put emphasis that the type returned is a ZodType, not an object. I thought that many users would quickly scan the text and assume this utility type returns the type of the input parameter of the action (it doesn't). 
We can remove if you think it's unnecessary.

File: src/content/docs/en/reference/modules/astro-actions.mdx (hunk 5 above)
Comment (mauriciabad) [reply src/content/docs/en/reference/modules/astro-actions.mdx]: Actually this paragraph is not necessary because the returned type already documented above. 
I added it thinking of possible doubts users would have, but it doesn't add any new info.

File: src/content/docs/en/reference/modules/astro-actions.mdx (hunk 6 above)
Comment (mauriciabad) [reply src/content/docs/en/reference/modules/astro-actions.mdx]: I think it's important that the text contains the keyword `never`. 

I wouldn't explain what `never` is, and if you think it's important, maybe instead link to some article or TS documentation.


```suggestion
Returns `never` when [`input` validator](#input-validator) is omitted.
```

src/content/docs/en/reference/modules/astro-actions.mdx
mauriciabad: I think this bolds are important because they highlight the returned data. The sentence is very specific and the important info may get lose in all the words.

mauriciabad:
Thanks everyone for the reviews, I added some comments and suggestions. Make any edits you want, for me all outcomes are okay.

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: It's our style to rarely use bold type in docs. If we start bolding everything, then nothing stands out. :smile: If this is a proper type, it can be written in inline code, or as I mentioned, if you feel it's helpful to link to a resource, then that gives people extra information. 

And if people are skimming, they only see "Zod Type" and not the meaning or context of that word. So bold is best used for an entire idea, not an individual word you would emphasize when e.g. speaking the sentence. That's why it's so rarely used! It's not often we want to bold an entire sentence, like the basic definition of something.

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: As I said, we rarely use bold in docs. All words are important! The sentence should be written to convey its meaning. I did update my suggestion to inline code `input` since it is something we document!

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+
+### `ActionInputSchema`
+
+<p>
+
+**Type:** `ZodType`
+<Since v="5.16.0" />
+</p>
+
+A utility type that extracts the input's Zod schema type from [an action handler](#defineaction). This can be useful to reference an [action's validator](#input-validator) input type in your own type definitions.
```
sarah11918:
```suggestion
A utility type that extracts the input's Zod schema type from [an action handler](#defineaction). This can be useful to reference an [action's `input` validator type](#input-validator) as an object in your own type definitions.
```
This matches the way we refer to the "input validator" in the earlier section, I think? And, since it links to the API reference for `input`, it feels like that should be a part of the linked text.

(Updated: also see my comment below if you want to remove the next paragraph and just sort of work that in here.)

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: I do like the emphasis on this entry (`ActionInputSchema`) being the first part of the sentence! And yes, I do think it should mention the type `never` being returned. This structure is clearer that this sentence is about this type entry. Works well!

src/content/docs/en/reference/modules/astro-actions.mdx
sarah11918: I made a suggestion above if the main idea here is that you can get this data as a separate object, we can work that into the earlier paragraph maybe? How does that sound? In that case then yes, we don't need this sentence. (But, no objections if people think adding that external link to `FormData` is helpful up there!)

sarah11918:
Should this also be added to the beginning of the types reference as something new to import from `astro:actions`?

Since these are all in alphabetical order, let's also move the entry up above `ActionReturnType` and `SafeResult` so that it's 3rd last.

image

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -575,3 +575,31 @@ type ContactResult = ActionReturnType<typeof actions.contact>;
...
+
+<p>
+
+**Type:** `ZodType`
+<Since v="5.16.0" />
+</p>
+
+A utility type that extracts the input's Zod schema type from [an action handler](#defineaction). This can be useful to reference an [action's `input` validator type](#input-validator) as an object in your own type definitions.
+
+Even if the action [accepts `form`](#accept-property), `ActionInputSchema` lets you get the input type as `object` rather than [`FormData`](https://developer.mozilla.org/docs/Web/API/FormData).
```
sarah11918:
```suggestion
```
I think the suggestion here was we don't need this paragraph anymore?

mauriciabad:
@sarah11918 feel free to directly commit the changes you want to make

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -546,6 +547,34 @@ button?.addEventListener('click', async () => {
...
```
sarah11918:
```suggestion
```
Decided to put this idea in the earlier paragraph

src/content/docs/en/reference/modules/astro-actions.mdx
```diff
@@ -546,6 +547,32 @@ button?.addEventListener('click', async () => {
...
 
+### `ActionInputSchema`
+
+<p>
+
+**Type:** `ZodType`
+<Since v="5.16.0" />
+</p>
+
+A utility type that extracts the input's Zod schema type from [an action handler](#defineaction). This can be useful to reference an [action's `input` validator type](#input-validator) as an object in your own type definitions.
```
florian-lefebvre:
```suggestion
A utility type that automatically infers the TypeScript type of an action's input based on its Zod schema. This can be useful to reference an action's [`input` validator type](#input-validator) as an object in your own type definitions.
```

sarah11918: @florian-lefebvre I edited your suggestion a bit! If you're OK with it, then I am too!

I don't think we need to link to `defineAction` if we're on the Actions reference page (so it's a bit distracting). Plus, I think the link to the `input` property itself is best where we're refering to it as a property, with inline code, vs just the "general idea of form input"

(Also, would be ok with "infers" instead of "determines" if you like that better!)

sarah11918 (approved):
Alright, this one's looking pretty good! Thanks everyone for their help with this one. Should be good to go for 5.16 this week!

sarah11918:
@yanthomasdev At your leisure.... :smile:

yanthomasdev (approved):
LGTM! Thanks!

PR #12648, opened 2025-10-31
docs: add note about environment variables with output: 'server'

allohamora:
#### Description (required)

I have added a missing note about bundling the environment variables during output: 'server' build, because it's unclear when you read the docs.

Example of the server output:
Image

Code what does it:
https://github.com/withastro/astro/blob/be566b3476260529645e92021957dd0ed502ce4b/packages/astro/src/env/vite-plugin-env.ts#L175

### DIFF
--- a/src/content/docs/en/guides/environment-variables.mdx
+++ b/src/content/docs/en/guides/environment-variables.mdx
@@ -285,6 +285,10 @@ There are three kinds of environment variables, determined by the combination of
 **Secret client variables** are not supported because there is no safe way to send this data to the client. Therefore, it is not possible to configure both `context: "client"` and `access: "secret"` in your schema.
 :::
 
+:::note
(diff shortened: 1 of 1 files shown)


florian-lefebvre:
Hi, thanks for opening a PR! Unfortunately I cannot approve it because its content is wrong. Let me explain:

> When using `output: 'server'`, `process.env` variables are bundled (hardcoded) into the server bundle during the `npm run build` phase

This is not exactly true. The conditions to make this happen are:
- You must use `astro:env` AFAIK
- You must use a public env var on a server rendered page

That being said, there's a bug there because we include too many things in this serialized object. I'll send a PR to only include the keys we need, thank you!

> This means environment variables become part of the compiled server code rather than being read from the environment at runtime. You will need to rebuild to have fresh `.env` values.

This is only true for public env vars when using `astro:env`, which is expected.

---

Sorry about that but thanks a lot for opening it tho! Allowed to spot a bug

PR #12658, opened 2025-11-03
Fix typos

deining:
This PR fixes typos I spotted in the project.

### DIFF
--- a/src/content/docs/en/guides/backend/xata.mdx
+++ b/src/content/docs/en/guides/backend/xata.mdx
@@ -104,7 +104,7 @@ const { records } = await xata.db.Posts.getPaginated({
   ))}
 </ul>
 ```
-It's important to note the SDK needs to be regenerated everytime your schema changes. So, avoid making changes to the generated files the Xata CLI creates because once schema updates, your changes will be overwritten.
+It's important to note the SDK needs to be regenerated every time your schema changes. So, avoid making changes to the generated files the Xata CLI creates because once schema updates, your changes will be overwritten.
 
 
 ## Official Resources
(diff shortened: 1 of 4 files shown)


deining:
> Thanks,

You are welcome.

> I think your right on both counts (I'm not entirely sure for `everytime`/`every time` because I'm not a native speaker and so I'd prefer another opinion).

Here comes another opinion:

```
Everytime should be written as two separate words: every time.
While some compound words like everywhere, everyday and everyone
have become commonplace in the English language,
everytime is not considered an acceptable compound word.
```

**Source:** https://www.grammarly.com/blog/commonly-confused-words/everytime-every-time/


> However, the `occurred` typo needs to be fixed in the core repo (https://github.com/withastro/astro/blob/main/packages/astro/src/core/errors/errors-data.ts) because the English error files are auto-generated as you can see at the top of each file:
> 
> > NOTE: This file is auto-generated from 'scripts/error-docgen.mjs'
> > Do not make edits to it directly, they will be overwritten.
> > Instead, change this file: https://github.com/withastro/astro/blob/main/packages/astro/src/core/errors/errors-data.ts
> > Translators, please remove this note and the  component.
> 
> So, the fixes will be overwritten in the next update if we apply them here. If you're able to make a PR on core instead, this would be helpful!

You are right, and I was aware of this, so the PR you requested for was already authored by me:

https://github.com/withastro/astro/pull/14711

> And, while this is a small fix, it's probably better to also remove the translations from this PR because we have an automated system to track files changes and when the core repo changes will be merged, this might be confusing for translators because this would already be fixed with this PR...

Done.

PR #12664, opened 2025-11-04
Document preview shortcuts

ghost:
#### Description (required)

Document shortcuts for `astro preview` 

- Closes #
- Suggested label: add-new-content, merge-on-release

### DIFF
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -152,6 +152,7 @@ export const sidebar = [
 					'reference/experimental-flags/static-import-meta-env',
 					'reference/experimental-flags/chrome-devtools-workspace',
 					'reference/experimental-flags/fail-on-prerender-conflict',
+					'reference/experimental-flags/svg-optimization',
 				],
 			}),
 			'reference/legacy-flags',
--- a/src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
+++ b/src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
@@ -1,0 +1,179 @@
+---
+title: Experimental SVG optimization
+sidebar:
+  label: SVG optimization
+i18nReady: true
+---
+
+import Since from '~/components/Since.astro'
+
+<p>
+
+**Type:** `boolean | object`<br />
+**Default:** `false`<br />
+<Since v="5.16.0" />
+</p>
+
+This experimental feature enables automatic optimization of your [SVG components](/en/guides/images/#svg-components) using [SVGO](https://svgo.dev/) during build time.
+
+When enabled, your imported SVG files used as components will be optimized for smaller file sizes and better performance while maintaining visual quality. This can significantly reduce the size of your SVG assets by removing unnecessary metadata, comments, and redundant code.
+
+To enable this feature with default settings, set it to `true` in your Astro config:
+
+```js title="astro.config.mjs" ins={5}
+import { defineConfig } from "astro/config"
+
+export default defineConfig({
+  experimental: {
+    svgo: true
+  }
+})
+```
+
+## Usage
+
+No change to using SVG components is required to take advantage of this feature. With experimental `svgo` enabled, all your SVG component import files will be automatically optimized:
+
+```astro title="src/pages/index.astro"
+---
+import Logo from '../assets/logo.svg';
+---
+
+<Logo />
+```
+
+The SVG will be optimized during the build process, resulting in smaller file sizes in your production build.
+
+Note that this optimization applies to every SVG component import in your project. It is not possible to opt out on a per-component basis.
+
+## Configuration
+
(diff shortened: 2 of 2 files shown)


src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -208,6 +208,10 @@ This command allows you to preview your site locally [after building](#astro-bui
 
 Since Astro 1.5.0, the [Node adapter](/en/guides/integrations-guide/node/) supports `astro preview` for builds generated with on-demand rendering.
 
+The following hotkeys can be used in the terminal where the Astro preview server is running:
+o + enter to open your Astro site in the browser.
+q + enter to quit the preview server.
```
ArmandPhilippot: This needs to be a list otherwise everything is rendered on a single line:
```suggestion
The following hotkeys can be used in the terminal where the Astro preview server is running:
- o + enter to open your Astro site in the browser.
- q + enter to quit the preview server.
```

ArmandPhilippot (commented):
I'm mentioning this in case anyone else has the same hesitation (ie. this could be a sign that this needs to move): I'm not sure if this should be displayed after or before the "Since Astro 1.5.0". I guess both works!

Except that hesitation, I left a suggestion due to a rendering issue but thanks, this looks good to me!

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -208,6 +208,10 @@ This command allows you to preview your site locally [after building](#astro-bui
 
 Since Astro 1.5.0, the [Node adapter](/en/guides/integrations-guide/node/) supports `astro preview` for builds generated with on-demand rendering.
 
+The following hotkeys can be used in the terminal where the Astro preview server is running:
+- o + enter to open your Astro site in the browser.
+- q + enter to quit the preview server.
+
 Can be combined with the [common flags](#common-flags) documented below.
```
sarah11918: Now this looks like *the hotkeys* can be combined with the common flags below. This list should go somewhere else to avoid confusion. I'd actually put it *before* the note about the Node adapter, which is pretty specific compared to general information about hotkeys.

sarah11918: And now that I think about it more, I don't even think we need that Node line at all? I think that was probably introduced when on-demand rendering was new. But it feels totally unnecessary now. What do you think, @ArmandPhilippot ? ("Since Astro v1, it does this thing that everyone totally expects it to do." :joy: )

ArmandPhilippot: Yeah, I think it's old enough to no longer say that about Node.

I wasn't totally sure about putting them in common flags though, because they are specific to `dev`/`preview` and both doesn't have the exact same hotkeys? (e.g. `s + enter` only works with `dev`).

But after reflection, it might simply be a matter of wording, and adding an "Hotkeys" section to the common flags would make them easier to find. Good idea!

sarah11918:
Just made a commit to:

- remove REAAAAALY old "fun fact"
- clarify that it's the `astro preview` command that can be combined with flags.

I think this is probably fine, but happy to have any opinions/suggestions to further edit this!

@jacobdalamb @ArmandPhilippot - Anything here you'd change or object to?

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -206,9 +206,11 @@ Starts a local server to serve the contents of your static directory (`dist/` by
 
 This command allows you to preview your site locally [after building](#astro-build) to catch any errors in your build output before deploying it. It is not designed to be run in production. For help with production hosting, check out our guide on [Deploying an Astro Website](/en/guides/deploy/).
 
-Since Astro 1.5.0, the [Node adapter](/en/guides/integrations-guide/node/) supports `astro preview` for builds generated with on-demand rendering.
+The following hotkeys can be used in the terminal where the Astro preview server is running:
+- o + enter to open your Astro site in the browser.
+- q + enter to quit the preview server.
```
ArmandPhilippot: I thought about the list items yesterday but I forgot we also need inline code here:
```suggestion
- `o` + `enter` to open your Astro site in the browser.
- `q` + `enter` to quit the preview server.
```

But, I also wonder if this is the best representation... Some might understand that as using both keys at the same time rather than in sequence (see: https://github.com/withastro/roadmap/discussions/1226#discussioncomment-14428321).

`o` + `enter` could be better than `o + enter`?

sarah11918: Yes, let's go with `o` + `enter`!

ghost:
LGTM

sarah11918:
Excellent! Seems we're good for 5.16 release here! It's in the milestone, and we'll merge it then.

bjohansebas (approved):
LGTM

PR #12673, opened 2025-11-05
refactor(adapter-reference): improve adapter API docs

ArmandPhilippot:
#### Description (required)

* Adds some missing `<Since />`.
* Adds missing info for `createExports()`: a second argument exists to retrieve data from `setAdapter()`.
* Adds docs for the missing `astro/app` APIs: `app.getAllowedDomains()`, `app.removeBase()`, `app.setCookieHeaders()`
* Fixes `app.match()` type: a second argument exists.
* Moves `envGetSecret` from "Adapter features" to "Astro features".
* Moves "Allow installation via astro add" to the end... not sure what is the best place for this.
* Replaces the types code block from [Building an adapter](https://docs.astro.build/en/reference/adapter-reference/#building-an-adapter) with proper documentation.
* Rewords the "Server entrypoint" heading and turns it into `h2`.
* Replaces `Exports`  > `createExports()` and `Start` > `start()` in "Build a server entrypoint" with some explanations about their shape.

<details>
<summary>See the relevant links to check types/default/since:</summary>

- `experimentalStaticHeaders`: added in https://github.com/withastro/astro/pull/13926
- `AdapterSupport` and `AdapterSupportWithMessage`: added in https://github.com/withastro/astro/pull/11806
  - `suppress` in `AdapterSupportWithMessage`: added in https://github.com/withastro/astro/pull/13842
- `previewEntrypoint`: added in https://github.com/withastro/astro/pull/5056
- `client`: added in https://github.com/withastro/astro/pull/14543
- `adapterFeatures`: added in https://github.com/withastro/astro/pull/7839
  - `buildOutput`: added in https://github.com/withastro/astro/pull/11942 (and https://github.com/withastro/astro/pull/11941 for the changesets)
- `supportedAstroFeatures`: added in https://github.com/withastro/astro/pull/7815
  - `envGetSecret`: added in https://github.com/withastro/astro/pull/10974
  - `i18nDomains`: added in https://github.com/withastro/astro/pull/9143 (there was an `i18n` property before, but since this uses another name, this is "new")
  - `sharpImageService`: added in https://github.com/withastro/astro/pull/11806
  - both `envGetSecret` and `i18nDomains` were added at the same time as the corresponding experimental feature, I'm not sure if we should use the stable release or not for them. Currently, I'm using the version of the experimental release.
  
 `astro/app` public APIs (only the missing ones):
* [`getAdapterLogger`](https://github.com/withastro/astro/blob/3d55c5d0fb520d470b33d391e5b68861f5b51271/packages/astro/src/core/app/index.ts#L137): added in https://github.com/withastro/astro/pull/8254 but I can't find it in the changelog... looking at the merge data, this was 2 days before [`v3.0.0`](https://github.com/withastro/astro/releases/tag/astro%403.0.0) so I think it's fine to use this version.
* [`getAllowedDomains`](https://github.com/withastro/astro/blob/3d55c5d0fb520d470b33d391e5b68861f5b51271/packages/astro/src/core/app/index.ts#L141): added in https://github.com/withastro/astro/commit/6ee63bfac4856f21b4d4633021b3d2ee059e553f which matches [`v5.14.2`](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#5142)
* [`removeBase`](https://github.com/withastro/astro/blob/3d55c5d0fb520d470b33d391e5b68861f5b51271/packages/astro/src/core/app/index.ts#L210): add in https://github.com/withastro/astro/pull/5290
* [`setCookieHeaders`](https://github.com/withastro/astro/blob/3d55c5d0fb520d470b33d391e5b68861f5b51271/packages/astro/src/core/app/index.ts#L507): added in https://github.com/withastro/astro/pull/4876
* [`setManifestData`](https://github.com/withastro/astro/blob/3d55c5d0fb520d470b33d391e5b68861f5b51271/packages/astro/src/core/app/index.ts#L206): added in https://github.com/withastro/astro/pull/7587

</details>

**Room for improvement:** except for `envGetSecret` that was already documented, the `Astro features` section is pretty succinct. I'm not sure how these features can/should be used, so I have not developed them.

##### Not documented
* `app.setManifestData` is publicly available when creating a new `App` (`astro/app`) but I couldn't find an usage in the source code so I wonder if this is a leftover that should be removed or something like that.

- Suggested label:  `add new content`, `consistency/formatting`, `improve or update documentation`

### DIFF
--- a/i18n-guides/正體中文.md
+++ b/i18n-guides/正體中文.md
@@ -1,154 +1,86 @@
-# Astro Docs 中文翻譯指南
+# Astro Docs 正體中文翻譯指南
 
-感謝你有興趣對 Astro 文件翻譯做出貢獻！  
-提交改動前，別忘了閱讀 [i18n 指南](https://contribute.docs.astro.build/guides/i18n/)，裡面包含了貢獻過程中需要知道的內容。  
-也歡迎加入 Astro 的 [Discord 社群](https://astro.build/chat)與其他貢獻者們交流！
+感謝你有興趣對 Astro 文件翻譯做出貢獻！
 
-本指南旨在整理一份統一的格式及風格指南，使文件讀起來更流暢並易於理解。  
+提交改動前，別忘了閱讀 [i18n 指南](https://contribute.docs.astro.build/guides/i18n/)，裡面包含了貢獻過程中需要知道的內容。
 
-編撰時參考了 [Vue 中文文档指南](https://github.com/vuejs-translations/docs-zh-cn/wiki/%E7%BF%BB%E8%AF%91%E9%A1%BB%E7%9F%A5)。
+此外，[Discord `docs-i18n` 頻道中的 `i18n-crew-zh` 討論串](https://discord.com/channels/830184174198718474/973149689610371093)是中文翻譯的討論區（目前未區分地區），歡迎先參與討論，避免重複其他人正在翻譯的頁面。
 
+## 本指南的目的
 
+本指南旨在定義書寫規範和統一的用語，讓讀者能流暢地理解內容，也使譯者能更簡單地完成翻譯。
 
-## 1. 術語和用詞
+## 中文書寫規範
 
-### 1.1 術語表
-> 部分參考：
-> - [《重編國語辭典修訂本》](https://dict.revised.moe.edu.tw/index.jsp)
-> - [臺灣在地化社群詞彙對照表](https://hackmd.io/@l10n-tw/glossaries)
-> - [大陸台灣計算機術語對照表](https://zh.wikibooks.org/zh-tw/大陸台灣計算機術語對照表)
+- 使用全形標點符號，使用直角引號。
+  - 斜線因全形在不同字體的外形差異較大，應使用半形，並於前後插入空格。
+- 拉丁字母、數字不使用全形。
+- 使用「你」而非「您」。
+- 選用臺灣正體漢字，例如「為」而非「爲」。
+- 專有名詞保留大小寫，無論出現於句中何處，例如「JavaScript」。
 
-|原文                   |推薦翻譯              |備註                    |
-|----------------------|---------------------|-----------------------|
-|library               |程式庫／函式庫         |                       |
-|integration           |整合                 |                       |
-|import                |引入／匯入            |                       |
-|package               |套件／軟體包          |                       |
-|routing               |路由                 |                       |
-|architecture          |架構                 |                       |
-|framework             |框架                 |                       |
-|reference             |參考                 |                       |
-|build                 |建構／建置            |                       |
-|deploy                |部署                 |                       |
-|component             |元件                 |                       |
-|migration             |遷移                 |                       |
-|asset                 |資產                 |                       |
-|docs                  |文件                 |                       |
-|documentation         |說明文件／技術文件     |                       |
-|project               |專案                 |                       |
-|configuration         |組態／設定            |                       |
-|community             |社群                 |                       |
-|hydration             |水合                 |                       |
-|islands               |群島                 |特指 Astro Islands      |
-|escape                |跳脫                 |特指 Escape Character   |
-|reactive              |反應式               |                       |
-|responsive            |響應式               |                       |
-|repository            |儲存庫               |                       |
-|experimental          |實驗性               |                       |
-|middleware            |中介層               |                       |
-|adapter               |配接器               |                       |
-|CLI                   |命令列               |                       |
-|terminal              |終端機               |                       |
-|codebase              |程式碼庫             |                       |
-|changelog             |變更日誌             |                       |
-|recipe                |操作指南             |                       |
-|server-side rendering |伺服器端算繪          |                       |
-|call                  |呼叫                 |                       |
+### 英數前後空格
 
+中文書寫時採用「[盤古之白](https://github.com/vinta/pangu.js)」，即中文和英文、數字間插入一個半形空格。
 
-### 1.2 中文用詞
-> 部分參考[《重編國語辭典修訂本》](https://dict.revised.moe.edu.tw/index.jsp)及[《異體字字典》](https://dict.variants.moe.edu.tw/variants/rbt/home.do)
+行內程式碼（`code`）前後也適用。
 
-|中文             |說明               |
-|----------------|-------------------|
-|你              |而不使用「您」       |
-|應用程式         |而不使用「應用」      |
-|舉例來說         |而不使用「舉個例子」等 |
-|其他             |並無「其它」一詞     |
+- 在漢字和拉丁字母、數字間插入一個半形空格。
+- 全形標點符號前後不應有空格。
+- 超連結前後不必加入空格，除非文字以半形字元開頭或結尾。
+- Markdown 標記開頭和結尾如要加上空格，應加在標記外部。
 
+## 術語表
 
-### 1.3 特指名詞及品牌名稱
-- 「JavaScript」或「ECMAScript」等詞語應**保留風格化寫法**，而非「Javascript」或「Ecmascript」
-- 「npm」及「webpack」等產品 / 品牌名稱的應**保留風格化寫法**，而非「Npm」或「Webpack」
+參考：
+- [臺灣在地化社群《詞彙對照表》](https://hackmd.io/@l10n-tw/glossaries)
+- [《資訊科技詞彙翻譯》](https://hackmd.io/@sysprog/it-vocabulary)
 
+| 原文                  | 譯文         | 備註                  |
+| --------------------- | ------------ | --------------------- |
+| adapter               | 配接器       |                       |
+| architecture          | 架構         |                       |
+| asset                 | 資產         |                       |
+| build                 | 建置         |                       |
+| call                  | 呼叫         |                       |
+| changelog             | 變更日誌     |                       |
+| CLI                   | 命令列       |                       |
+| codebase              | 程式碼庫     |                       |
+| community             | 社群         |                       |
+| component             | 元件         |                       |
+| configuration         | 組態         |                       |
+| deploy                | 部署         |                       |
+| docs                  | 文件         |                       |
+| documentation         | 文件         |                       |
+| escape                | 跳脫         | 特指 escape character |
+| experimental          | 實驗性       |                       |
+| framework             | 框架         |                       |
+| hydration             | 水合         |                       |
+| import                | 匯入         |                       |
+| integration           | 整合         |                       |
+| islands               | 群島         | 特指 Astro islands    |
+| library               | 函式庫       |                       |
+| middleware            | 中介層       |                       |
+| migration             | 遷移         |                       |
+| package               | 軟體包       |                       |
+| project               | 專案         |                       |
+| reactive              | 反應式       |                       |
+| recipe                | 操作指南     |                       |
+| reference             | 參考         |                       |
+| repository            | 儲存庫       |                       |
+| responsive            | 響應式       |                       |
+| routing               | 路由         |                       |
+| server-side rendering | 伺服器端算繪 |                       |
+| terminal              | 終端機       |                       |
 
-### 1.4 關於 MDN 中可查詢的前端術語
+### MDN 中可查詢的前端術語
+
 - 原則上，前端專業術語的翻譯優先**參考 MDN 中文譯文**，若無中文則**保留原文**。
 - 原文中的 MDN 文件連結建議轉換為 MDN 的中文譯文連結，但是鑒於 MDN 的中文翻譯也並不完善，所以：
-    - 當條目有中文譯文時，將其手動**變更為 MDN 中文連結**，其餘則**保留原文連結**
+  - 當條目有中文譯文時，將其手動**變更為 MDN 中文連結**，其餘則**保留原文連結**
 
+## 更新本指南
 
+本指南是現時社群翻譯貢獻者的共識，如有任何意見，歡迎在 Discord 的 `i18n-crew-zh` 討論串提出。
 
-## 2. 譯文格式
-
-### 2.1 標點符號
-> 參考[《重訂標點符號手冊》](https://language.moe.gov.tw/001/upload/files/site_content/m0001/hau/c2.htm)及[《W3C 中文排版需求》](https://w3c.github.io/clreq/#line_composition_rules_for_punctuation_marks)
-- 除特殊情況外，所有中文標點符號應為**全形字元**
-- 考慮到 `／` 在不同字體下算繪結果差異較大，分隔號應為前後包含空格的 `/`，如：`你好 / 世界`
-
-
-### 2.2 行內程式碼與關鍵字
-請使用**反單引號**（`）標記，例：
-```md
-呼叫 `window` 的 `alert()` 函式
-```
-
-
-### 2.3 中西文混排
-> 部分參考[《W3C 中文排版需求》](https://w3c.github.io/clreq/#chinese_and_western_mixed_text_composition)
-- 中文和英文之間**應**插入一個**半形空格**。如：`你好 Hello 世界`
-- 全形標點與英文之間**不須**插入空格。如：`你好世界（Hello World）`
-- 注意使用 Markdown 標記（如*斜體*、**粗體**或[超連結](#)等）時，應保證空格出現在欲分隔字詞的**外部**。如：
-```md
-遵守 **JavaScript 程式碼規範**非常重要
-遵守 [JavaScript 程式碼規範](#foo)非常重要
-詳見*中文 MDN* 的參考頁面
-詳見[中文 MDN](#bar) 的參考頁面
-```
-
-
-### 2.4 程式碼片段
-- 程式碼片段中的**註解應翻譯為中文**，其餘部分應**保留原文**
-- 日誌及執行結果等也應**保留原文**，以便和程式碼邏輯保持一致
-
-
-
-## 3. 翻譯建議
-
-### 3.1 調整語序為中文習慣
-例如：
-> So far we've only been binding to simple property keys in our templates. But Vue.js actually supports the full power of JavaScript expressions inside all data bindings.
-
-以上例句可譯為：「目前我們都只是在模板中綁定簡單的屬性鍵，但事實上 Vue.js 支援在任何資料綁定中使用 JavaScript 表達式」。
-
-
-### 3.2 省略部分在原文中僅作完成句式的詞語
-例如：
-> When Vue is updating a list of elements rendered with v-for, by default it uses an "in-place patch" strategy. If the order of the data items has changed, instead of moving the DOM elements to match the order of the items, Vue will patch each element in-place and make sure it reflects what should be rendered at that particular index.
-
-像是「to match the order of the items」這樣的短語可以在**保證語意不變**的情況下簡化。  
-
-以上例句可譯為：「Vue 預設按照『就地更新』的策略來更新透過 `v-for` 算繪的元素列表。當資料的順序改變時，Vue 不會隨之移動 DOM 元素的順序，而是就地更新每個元素，確保它反映應在指定索引位置上算繪的內容」。
-
-
-### 3.3 補充連接詞使句子更加連貫
-例如：
-> The mustache tag will be replaced with the value of the msg property from the corresponding component instance. It will also be updated whenever the msg property changes.
-
-可在句中加入如「然而」、「並且」或「即使……也……」等連接詞來使句子更加連貫。  
-
-以上例句的後半句可譯為：「每當 `msg` 屬性發生改變時，綁定處的內容便會隨之更新」。
-
-
-### 3.4 在尊重原文的前提下使讀者更容易理解
-
-> TODO
-
-### 3.5 譯文及原文行號應相對應
-由於 Astro 文件以 Markdown 撰寫，每一行形成一個自然段落，因此原則上應保持譯文及原文行號應相對應，以保證後續更新時不造成混淆。
-
-
-
-## 4. 關於本指南
-本指南**並非**硬性規定，而是代表中文文件翻譯團隊共識的建議。  
-如果你對本指南有任何想法或勘誤建議，歡迎發起 Issue 或是加入 [Discord 社群](https://astro.build/chat)中的 `#i18n-crew-zh` 頻道參與討論！
+術語表加入新項目時請依原文字母序排序，並保持原文全小寫，除非是縮寫。格式化表格時漢字視為 2 個半形字元寬。
--- a/public/logos/jekyllpad.svg
+++ b/public/logos/jekyllpad.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" stroke="#fc7414" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" viewBox="0 0 24 24">
+  <path d="M14.5 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V7.5z"/>
+  <path d="M14 2v6h6m-4 5H8m8 4H8m2-8H8"/>
+</svg>
--- a/src/components/RightSidebar/LearnAstroAd.astro
+++ b/src/components/RightSidebar/LearnAstroAd.astro
@@ -3,7 +3,7 @@ import { Image } from 'astro:assets';
 import CodingInPublic from './CodingInPublic.png';
 import { LinkButton } from '@astrojs/starlight/components';
 
-const offerUrl = 'https://learnastro.dev?code=ASTRO_PROMO';
+const offerUrl = 'https://learnastro.dev?code=ASTRO_HOLIDAY_PROMO';
 
 const idCount = (((Astro.locals as any)._ad_render_id as number | undefined) ??= 1);
 (Astro.locals as any)._ad_render_id++;
@@ -26,7 +26,7 @@ const headingId = `learn-astro-course-${idCount}`;
 			<span>•</span>
 			<span>Astro v5 ready</span>
 		</p>
-		<LinkButton href={offerUrl} class="cta" data-learn-astro-cta>Get 20% off</LinkButton>
+		<LinkButton href={offerUrl} class="cta" data-learn-astro-cta>Get 33% off</LinkButton>
 	</aside>
 </div>
 
--- a/src/content/docs/de/basics/layouts.mdx
+++ b/src/content/docs/de/basics/layouts.mdx
@@ -195,7 +195,7 @@ Ein Markdown-Layout hat Zugriff auf alle [verfügbaren Eigenschaften](/de/guides
 
 * Die Überschrifteninformationen (d.h. die Elemente `h1 -> h6`) sind über das Array `headings` verfügbar und nicht über die Funktion `getHeadings()`.
 
-* `file` and `url` are *also* available as nested `frontmatter` properties (i.e. `frontmatter.url` and `frontmatter.file`).
+* `file` und `url` sind *ebenfalls* verfügbar als verschachtelte `frontmatter` Eigenschaften (z. B. `frontmatter.url` und `frontmatter.file`).
 
 :::
 
--- a/src/content/docs/en/editor-setup.mdx
+++ b/src/content/docs/en/editor-setup.mdx
@@ -73,7 +73,7 @@ Installation instructions, editor integration, and additional information can be
 
 ### Biome
 
-[Biome](https://biomejs.dev/) is an all-in-one linter and formatter for the web. [Biome currently has partial support for `.astro` files](https://biomejs.dev/internals/language-support/#html-super-languages-support), and can be used to lint and format the frontmatter in `.astro` files.
+[Biome](https://biomejs.dev/) is an all-in-one linter and formatter for the web. [Biome currently has experimental support for `.astro` files](https://biomejs.dev/internals/language-support/#html-super-languages-support), and can be used to lint and format the frontmatter in `.astro` files.
 
 ### Prettier
 
--- a/src/content/docs/en/guides/cms/apostrophecms.mdx
+++ b/src/content/docs/en/guides/cms/apostrophecms.mdx
@@ -51,17 +51,17 @@ To connect Astro with your ApostropheCMS project, install the official Apostroph
 <PackageManagerTabs>
   <Fragment slot="npm">
   ```shell
-  npm install @apostrophecms/apostrophe-astro vite @astro/node
+  npm install @apostrophecms/apostrophe-astro vite @astrojs/node
   ```
   </Fragment>
   <Fragment slot="pnpm">
   ```shell
-  pnpm add @apostrophecms/apostrophe-astro vite @astro/node
+  pnpm add @apostrophecms/apostrophe-astro vite @astrojs/node
   ```
   </Fragment>
   <Fragment slot="yarn">
   ```shell
-  yarn add @apostrophecms/apostrophe-astro vite @astro/node
+  yarn add @apostrophecms/apostrophe-astro vite @astrojs/node
   ```
   </Fragment>
 </PackageManagerTabs>
--- a/src/content/docs/en/guides/cms/jekyllpad.mdx
+++ b/src/content/docs/en/guides/cms/jekyllpad.mdx
@@ -1,0 +1,21 @@
+---
+title: JekyllPad & Astro
+description: Manage your Astro site content with JekyllPad’s browser-based, GitHub-backed editor, no Git or Markdown hassles ever.
+sidebar:
+  label: JekyllPad
+type: cms
+service: JekyllPad
+i18nReady: true
+stub: true
+---
+
+[JekyllPad](https://www.jekyllpad.com) is a lightweight, browser-based CMS that connects directly to your GitHub repository. 
+
+It provides a modern WYSIWYG + Markdown editor, commits changes straight to your repo, and runs 100% client‑side, so you can manage content for Astro without setting up servers or learning Git.
+
+It is a client-side application that runs completely in user's browser. This means data, oauth and security tokens, all stays in user's browser. No installation required, works with any mobile or desktop browsers.
+
+## Official Resources
+
+- [JekyllPad CMS for Astro](https://www.jekyllpad.com/features/astro-headless-cms)
+
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -857,11 +857,17 @@ An Astro font provider is made up of two parts: the config object and the actual
 3. Add your custom provider to your font configuration.
 
     ```js title="astro.config.mjs" ins="myProvider()"
-    fonts: [{
-      provider: fontProviders.myProvider(),
-      name: "Custom Font",
-      cssVariable: "--font-custom"
-     }]
+    import { defineConfig } from "astro/config";
+    import { myProvider } from "./provider/config";
+
+    export default defineConfig({
+      experimental: {
+        fonts: [{
+          provider: myProvider(),
+          // ...
+        }]
+      }
+    });
     ```
 
 </Steps>
--- a/src/content/docs/en/reference/modules/astro-i18n.mdx
+++ b/src/content/docs/en/reference/modules/astro-i18n.mdx
@@ -40,6 +40,10 @@ import {
   notFound,
   middleware,
   requestHasLocale,
+  normalizeTheLocale,
+  pathHasLocale,
+  toCodes,
+  toPaths
  } from 'astro:i18n';
 ```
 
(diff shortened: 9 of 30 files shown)


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,228 @@ export default function createIntegration() {
...
+
+### `supportedAstroFeatures`
+
+<p>
+
+**Type:** `AstroAdapterFeatureMap`<br />
+<Since v="3.0.0" />
+</p>
+
+A map of Astro built-in features supported by the adapter. This allows Astro to determine which features an adapter is unable or unwilling to support so appropriate error messages can be provided.
```
ascorbic: "unable or unwilling" sounds a bit strange

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,228 @@ export default function createIntegration() {
...
+<p>
 
-```js
-export function handler(event, context) {
-  // ...
+**Type:** `string | URL`<br />
+<Since v="1.5.0" />
+</p>
+
+Defines the path of a module in the adapter's package that is responsible of starting up the built server when `astro preview` is run.
```
ascorbic:
```suggestion
Defines the path or ID of a module in the adapter's package that is responsible of starting up the built server when `astro preview` is run.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -174,11 +296,30 @@ export default function createIntegration() {
 }
 ```
 
-#### Start
+The `createExports()` second argument can be useful when you need to access build-time configuration in your server entrypoint. For example, your server entrypoint might need the directory path in the build output where the resources generated by Astro are located:
```
ascorbic:
```suggestion
The second argument passed to `createExports()` can be useful when you need to access build-time configuration in your server entrypoint. For example, your server entrypoint might need the directory path in the build output where the resources generated by Astro are located:
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -174,11 +296,30 @@ export default function createIntegration() {
...
+
+```js title="my-adapter/server.js"
+import { App } from 'astro/app';
+
+export function createExports(manifest, args) {
+  const assetsPath = args.assets;
+  /* ... */
+  return {};
+}
```
ascorbic, line 310: I think it would be good if this showed an example of an actual exported handler function rather than an empty object

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -190,10 +331,12 @@ export function start(manifest) {
 }
 ```
 
-#### `astro/app`
+### `astro/app`
 
 This module is used for rendering pages that have been prebuilt through `astro build`. Astro uses the standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) and [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) objects. Hosts that have a different API for request/response should convert to these types in their adapter.
 
+The `App` constructor accepts a mandatory SSR manifest and a second optional argument, defaulting to `true`, to enable or disable streaming.
```
ascorbic:
```suggestion
The `App` constructor accepts a required SSR manifest argument, and optionally an argument to enable or disable streaming, defaulting to `true`.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -211,14 +354,14 @@ export function start(manifest) {
...
-##### `app.render()`
+#### `app.render()`
 
 <p>
 
 **Type:** `(request: Request, options?: RenderOptions) => Promise<Response>`
 </p>
 
-This method calls the Astro page that matches the request, renders it, and returns a Promise to a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object. This also works for API routes that do not render pages.
+A method that accepts a mandatory `request` argument and an optional `RenderOptions` object. This calls the Astro page that matches the request, renders it, and returns a promise to a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object. This also works for API routes that do not render pages.
```
ascorbic:
```suggestion
A method that accepts a required `request` argument and an optional `RenderOptions` object. This calls the Astro page that matches the request, renders it, and returns a promise to a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object. This also works for API routes that do not render pages.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,24 +506,70 @@ if(app.match(request)) {
...
-{
-  "name": "example",
-  "keywords": ["astro-adapter"],
+<p>
+
+**Type:** `() => AstroIntegrationLogger`<br />
+<Since v="v3.0.0" />
+</p>
+
+Returns an [instance of the Astro logger](/en/reference/integrations-reference/#astrointegrationlogger) so that the adapter's runtime can use the same logger.
```
ascorbic: The duplication of "logger" is a bit awkward here

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -441,19 +630,58 @@ export default function createIntegration() {
...
-If Astro's default log message is redundant, or confusing to the user in combination with your custom `message`, you can use `suppress: "default"` to suppress the default message and only log your message:
+### `hybridOutput`
 
-```js title="my-adapter.mjs" ins={13}
+<p>
+
+**Type:** [`AdapterSupport`](#adaptersupport)
+</p>
+
+Defines whether the adapter is able to serve pages that are static or on-demand rendered.
```
ascorbic:
```suggestion
Defines whether the adapter is able to serve sites that include a mix of static and on-demand rendered pages.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -441,19 +630,58 @@ export default function createIntegration() {
...
+
+### `serverOutput`
+
+<p>
+
+**Type:** [`AdapterSupport`](#adaptersupport)
+</p>
+
+Defines whether the adapter is able to serve on-demand pages.
```
ascorbic:
```suggestion
Defines whether the adapter is able to serve on-demand rendered pages.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -476,34 +700,65 @@ export default function createIntegration() {
...
+
+  const handler = (event, context) => {
+    // ...
   };
+
+  return { handler };
+}
+```
+
+If you support secrets, be sure to call `setGetEnv()` before `getSecret()` when your environment variables are tied to the request:
```
ascorbic:
```suggestion
If the adapter supports secrets, be sure to call `setGetEnv()` before `getSecret()` when environment variables are tied to the request:
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -597,63 +852,109 @@ export default function createIntegration() {
...
+<p>
 
-setGetEnv((key) => process.env[key])
+**Type:** `boolean`<br />
+<Since v="5.9.3" />
+</p>
 
-export function createExports(manifest) {
-  const app = new App(manifest);
+When this feature is enabled, Astro will return a map of the `Headers` emitted by the static pages. This map `experimentalRouteToHeaders` is available in the [`astro:build:generated` hook](/en/reference/integrations-reference/#astrobuildgenerated).
```
ascorbic:
```suggestion
Defines if the adapter supports setting response headers for static pages. When this feature is enabled, Astro will return a map of the `Headers` emitted by the static pages. This map `experimentalRouteToHeaders` is available in the [`astro:build:generated` hook](/en/reference/integrations-reference/#astrobuildgenerated). The adapter could then use this to generate a `_headers` or other configuration file, for example.
```

I'm not sure about the wording, but it would be good to specify what this is *for*, not just what it does.

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: I'll update your suggestion before merging it: `support` > `supports`.

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: I reused the wording we had but I guess we could shorten the sentence if this isn't clear to you:
```suggestion
A map of Astro built-in features supported by the adapter. This allows Astro to determine which features an adapter supports so appropriate error messages can be provided.
```

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: My idea was not to reuse the same snippet in order to focus on `args`, but I agree this is a bit awkward. So I pushed a new commit, using line highlighting instead, let me know if that looks better to you.

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Yeah, it's not perfect. But I was also not a fan of:
> Returns an instance of the Astro logger so that the adapter's runtime environment can also use it.

I'll try to find another formulation.

src/content/docs/en/reference/adapter-reference.mdx
ascorbic: Yeah, that's better now

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: > Returns an instance of the Astro logger available to the adapter's runtime environment

???

src/content/docs/en/reference/adapter-reference.mdx
ascorbic: Yes, that's better

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: Maybe something like this?

```suggestion
Whether or not the adapter provides experimental support for setting response headers for static pages. When this feature is enabled, Astro will return a map of the `Headers` emitted by the static pages. This map `experimentalRouteToHeaders` is available in the [`astro:build:generated` hook](/en/reference/integrations-reference/#astrobuildgenerated) for generating files such as a `_headers` that allows you to make changes to the default HTTP header.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -597,63 +856,109 @@ export default function createIntegration() {
...
+  
+The value of the headers might change based on the features enabled/used by the application.
```
```suggestion
```
I think I'd keep these as one paragraph. (I can't make a suggestion across a deleted line, but I can remove the line between them.)

I also might put that paragraph *below* the code example, so it's not as far removed from the definition.

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -597,63 +856,109 @@ export default function createIntegration() {
...
 
-	return { default: { fetch } };
-}
-```
+**Type:** `"deprecated" | "experimental" | "limited" | "stable" | "unsupported"`
+</p>
+
+Defines the level of support for a feature by your adapter:
+* Use `"deprecated"` when your adapter deprecates a feature support before removing it completely in a future version.
+* Use `"experimental"` when your adapter adds support for a feature and issues or breaking changes are expected.
```
```suggestion
* Use `"experimental"` when your adapter adds support for a feature, but issues or breaking changes are expected.
```
What do you think? This felt like maybe a "but" instead of an "and"? Both are correct, but there's the possibility of reading "adds support for X and Y" before hitting the rest of the sentence and backtracking. I think "but" both separates the phrases cleanly, and adds extra nuance to the relationship.

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -597,63 +856,109 @@ export default function createIntegration() {
...
+* Use `"limited"` when your adapter only supports a subset of the full feature.
+* Use `"stable"` when the feature is fully supported by your adapter and stable.
```
```suggestion
* Use `"stable"` when the feature is fully supported by your adapter.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -695,16 +992,29 @@ export default function createIntegration() {
...
-        // for your virtual host of choice
-      },
     },
   };
 }
 ```
+
+## Allow installation via `astro add`
+
+[The `astro add` command](/en/reference/cli-reference/#astro-add) allows users to easily add integrations and adapters to their project. If you want _your_ adapter to be installable with this tool, **add `astro-adapter` to the `keywords` field in your `package.json`**:
```
```suggestion
[The `astro add` command](/en/reference/cli-reference/#astro-add) allows users to easily add integrations and adapters to their project.  To allow your adapter to be installed with this command, **add `astro-adapter` to the `keywords` field in your `package.json`**:
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -695,16 +992,29 @@ export default function createIntegration() {
...
+
+```json
+{
+  "name": "example",
+  "keywords": ["astro-adapter"],
+}
+```
+
+Once you [publish your adapter to npm](https://docs.npmjs.com/cli/v8/commands/npm-publish), running `astro add example` will install your package with any peer dependencies specified in your `package.json`. We will also instruct users to update their project config manually.
```
```suggestion
Once you [publish your adapter to npm](https://docs.npmjs.com/cli/v8/commands/npm-publish), running `astro add example` will install your package with any peer dependencies specified in your `package.json` and instruct users to update their project config manually.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -174,11 +296,34 @@ export default function createIntegration() {
 }
 ```
 
-#### Start
+The second argument passed to `createExports()` can be useful when you need to access build-time configuration in your server entrypoint. For example, your server entrypoint might need the directory path in the build output where the resources generated by Astro are located:
```
florian-lefebvre: Maybe link back to the args section?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,228 @@ export default function createIntegration() {
...
+
+### `adapterFeatures`
+
+<p>
+
+**Type:** `AstroAdapterFeatures`<br />
+<Since v="3.0.0" />
+</p>
+
+An object that enables specific features that must be supported by the adapter. These features will change the built output, and the adapter must implement the proper logic to handle the different output.
```
```suggestion
An object that specifies which [adapter features that change the build output](#adapter-features) are supported by the adapter.
```

It sounded a little odd to say "enables features that *must* be supported by the adapter". Is the above a correct statement? If you want to say more about implementing proper logic, I think that would be more appropriate actually *in* the "Adapter features" section. Here we're just defining what they are, and it seems the defining feature is they change the build output?

(Then, I wouldn't even have the readmore below because we've linked inline)

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -190,10 +335,12 @@ export function start(manifest) {
 }
 ```
 
-#### `astro/app`
+### `astro/app`
```
florian-lefebvre, line 336: Should we also talk about `NodeApp`?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,228 @@ export default function createIntegration() {
...
+<ReadMore>Learn more about the [available features](#adapter-features) that an adapter can activate.</ReadMore>
+
+### `args`
+
+<p>
+
+**Type:** `any`
+</p>
+
+A JSON-serializable value that will be passed to the adapter's server entrypoint at runtime. This is useful to pass an object containing build-time configuration (e.g. paths, secrets, etc.) to your server runtime code.
```
```suggestion
A JSON-serializable value that will be passed to the adapter's server entrypoint at runtime. This is useful to pass an object containing build-time configuration (e.g. paths, secrets) to your server runtime code.
```
"e.g." already means "a non-exhaustive list" so no need for "etc."!

florian-lefebvre:
Re `app.setManifestData`: I think it should be removed indeed, no match on GitHub either. I'll add it to the v6 milestone

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,228 @@ export default function createIntegration() {
...
+
+The following example defines an `args` object to pass to [the server entrypoint](#building-a-server-entrypoint) the directory path in the build output where the resources generated by Astro are located:
```
sarah11918: There is something not quite grammatical about this. So I tried to figure out what this was saying. When you break it down, it says:

```suggestion
The following example defines an `args` object with the directory path in the build output for the location of the resources generated by Astro:
```
Is `config.build.assets` a directory path? (it doesn't look like a *path* in the file tree/URL sense (`/*/*/), but I'm not familiar with this usage so it is probably fine. Just checking!) 

It's also true that it's being passed *to* the server entry point, but a) you already defined `args` as something that gets passed to a server entry point and b) that makes this sentence more complicated to include the thing that you are passing and what you're passing it *to*. So, it's probably clearer not to mention that so that this can focus on helping make sense of what `args` is/does.

So, is the above correct? (Because that's what you've said.) If not, how does it need to change?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,228 @@ export default function createIntegration() {
...
+<p>
 
-```js
-export function handler(event, context) {
-  // ...
+**Type:** `string | URL`<br />
+<Since v="1.5.0" />
+</p>
+
+Defines the path or ID of a module in the adapter's package that is responsible of starting up the built server when `astro preview` is run.
```
```suggestion
Defines the path or ID of a module in the adapter's package that is responsible for starting up the built server when `astro preview` is run.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,228 @@ export default function createIntegration() {
...
+    },
+  };
 }
 ```
 
-With the adapter API you achieve this by implementing `createExports` in your `serverEntrypoint`:
+## Building a server entrypoint
 
-```js
+Astro's adapter API attempts to work with any type of host, and gives a flexible way to conform to the host APIs.
```
```suggestion
You will need to create a file that executes during server-side requests to enable on-demand rendering with your particular host. Astro's adapter API attempts to work with any type of host, and gives a flexible way to conform to the host APIs.
```
Feels like *something* like this is kind of needed here? I just took phrases from various places, so make this make sense 😂

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,24 +510,70 @@ if(app.match(request)) {
...
+#### `app.getAllowedDomains()`
 
-## Astro features
+<p>
+
+**Type:** `() => Partial<RemotePattern>[] | undefined`<br />
+<Since v="5.14.2" />
+</p>
+
+Returns a list of permitted host patterns for incoming requests when using SSR [as defined in the user configuration](/en/reference/configuration-reference/#securityalloweddomains).
```
sarah11918: Can we say something besides SSR? Even if just "server-rendering"?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,24 +510,70 @@ if(app.match(request)) {
...
+#### `app.setCookieHeaders()`
+
+<p>
+
+**Type:** `(response: Response) => Generator<string, string[], any>`<br />
+<Since v="1.4.0" />
+</p>
 
-<p><Since v="3.0.0" /></p>
+Returns a generator that yields individual cookie header values from a `Response` object. This is used to properly handle multiple cookies that may have been set during request processing.
```
sarah11918, line 560:
```suggestion
Returns a generator that returns individual cookie header values from a `Response` object. This is used to properly handle multiple cookies that may have been set during request processing.
```
If we didn't say "yields", what could we replace it with? "provides" "produces" "outputs" "generates"?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -567,16 +826,16 @@ function createEdgeMiddleware(middlewareEntryPoint) {
...
-**Type:** `AdapterSupportsKind`
+**Type:** `"static" | "server"`<br />
+**Default:** `"server"`<br />
+<Since v="5.0.0" />
 </p>
 
-This is a feature to allow your adapter to retrieve secrets configured by users in `env.schema`.
-
-Enable the feature by passing any valid `AdapterSupportsKind` value to the adapter:
+Allows you to force a specific output shape for the build. This can be useful for adapters that only work with a specific output type, for instance, your adapter might expect a static website, but uses an adapter to create host-specific files. Defaults to `server` if not specified.
```
```suggestion
Allows you to force a specific output shape for the build. This can be useful for adapters that only work with a specific output type. For example, your adapter might expect a static website, but uses an adapter to create host-specific files. Defaults to `server` if not specified.
```

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot:
```suggestion
Returns an [instance of the Astro logger](/en/reference/integrations-reference/#astrointegrationlogger) available to the adapter's runtime environment.
```

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: I like Sarah's version but I'm not quite sure what `_headers` is. Reading Matt's suggestion I guess this is a configuration file so maybe we need to say it explicitly ("a `_headers` configuration file") to avoid confusion?

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Done in [aad56ad](https://github.com/withastro/docs/pull/12673/commits/aad56add7097dcb80c0ce61bd68cff1aadc00135)!

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Sounds good to me, I committed that separately to also remove the `ReadMore`: [10285f9](https://github.com/withastro/docs/pull/12673/commits/10285f94aa8e3c157a37f7ebe10d7ccedd357180)

src/content/docs/en/reference/adapter-reference.mdx
ascorbic: [`yield`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield) is the correct term when it's a generator

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: That's why I said "generating files such as a `_headers`. Apparently, in some cases it's a `.txt` file, but e.g. saw for Cloudflare that someone needed to remove the file extension. What if we just remove the "a"?

"for generating files such as `_headers` ..."

I followed the trail to get from https://docs.netlify.com/manage/routing/headers/ to https://developer.mozilla.org/en-US/docs/Glossary/HTTP_header and even had a draft that included the purpose of the HTTP header file, but ultimately, it's not our job to go that far and document that basic, I think, for someone reading the *Adapter* API?

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: OK, perfect then!

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: The `args` property was the hardest to understand (this is not the easiest word to look for in a codebase 😆), so this section may not be the most accurate.

> Is config.build.assets a directory path?

I'm honestly not sure what is the actual value. Reading the configuration reference, that could be only the directory name. I don't know if there is some kind of internal resolution to turn it in a relative/absolute path. 

So, your suggestion sounds good to me! I wasn't fan of my version but couldn't find a better wording. 😅 
I'm not committing it yet in case someone else want to answer your question/if there is anything to improve regarding that.

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: I took that from the configuration reference, but looking at the `security`, I think it's safe to use `on-demand rendering` here, to stay consistent. Good call!

```suggestion
Returns a list of permitted host patterns for incoming requests when using on-demand rendering [as defined in the user configuration](/en/reference/configuration-reference/#securityalloweddomains).
```

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: Maybe @ascorbic or @florian-lefebvre can clarify what `args` is: a path? a directory?

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Sure, good idea. I'm not sure how to integrate it in the current version, so maybe (while reusing Sarah's suggestion from the other place):
```suggestion
You can access the [`args`](#args) defined by your adapter through the second argument of `createExports()`. This can be useful when you need to access build-time configuration in your server entrypoint. For example, your server might need to identify where assets generated by Astro are located:
```

Edit: updated with Sarah's most recent suggestion

sarah11918:
I think everyone has done some amazing work supporting @ArmandPhilippot 's efforts! Only two points where I feel we really need to do a tiny bit better (called in both Matt and Florian to help), and once those feel good, this puppy LGTM!

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Good question! If you think this is useful for adapter builders, yes, we should mention it!
From what I see this extends the `App` so maybe this should be nested in `astro/app`? Unless users should prefer `NodeApp` when this is a Node environment?
I will check the additional methods to see how we can best integrate this!

src/content/docs/en/reference/adapter-reference.mdx
florian-lefebvre: Basically `args` can be anything (as long as it's serializable)! It's just data you want to get in `createExports()`, for example: https://github.com/withastro/astro/blob/7b7e037e435488227a361dcbe89a0a44d327c2c4/packages/integrations/node/src/server.ts#L16

src/content/docs/en/reference/adapter-reference.mdx
florian-lefebvre: We currently use it in the node and vercel adapters AFAIK

src/content/docs/en/reference/adapter-reference.mdx
ascorbic: I think "the directory path in the build output where the resources generated by Astro are located" is too confusing and shoulkd be removed. It's just an example of the sort of thing that could be passed in.

sarah11918:
```suggestion
The following example defines an `args` object to identify where assets generated by Astro are located:
```
Would this maybe be accurate and "good enough" then? Or 

"... defines an `args` object to provide additional information about assets generated by Astro"

?

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: Anyone? Suggestions for this?

ArmandPhilippot:
Thanks everyone for the really helpful feedbacks! ❤️ 

I added a new `astro/app/node` section with the methods overridden by this module and the additional ones.

While doing so, I noticed we document the methods available from `App` once created (ie. `const app = new App()`), not the static methods. Should we? I noticed:
- `App.getSetCookieFromResponse()`
- `App.validateForwardedHost()`

We also have some static methods for `NodeApp`:
- `NodeApp.createRequest()`
- `NodeApp.writeResponse()`

I'm not quite sure how to make the distinction between the static methods and the others with the current structure.

`Since` info:
* `nodeApp.render()` signature has changed in https://github.com/withastro/astro/pull/9199 so I set `<Since />` to this release.
* `nodeApp.headersMap` and `nodeApp.setHeadersMap()`: added in https://github.com/withastro/astro/pull/13972

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: I preferred to wait before committing it in case one of our adapter experts wanted to check the accuracy, but great idea! And this sounds correct to me.

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Resolved with [17f719a](https://github.com/withastro/docs/pull/12673/commits/17f719a535ad72372e167bec50858b98f70cf493)! I finally choose to put it at the same level as `astro/app`. Let me know if that sounds accurate to you (I tried to avoid duplicating the same info for `render()` and `match()`).

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Oh, I thought I had already answered, I probably forgot to click the button. 😅 

Thanks for the extra context, I wasn't aware this was an "expected" file. So yes, this sounds good to me! (I think I prefer without `a` before, but I'm not a native speaker so choose the version that sounds the more correct to you!)

ascorbic: Yeah, my wording was unclear! There was an implied [file] after "_headers". So to be explicit, it would be 
```
The adapter could then use this to generate a `_headers` file or other configuration file.
```

ascorbic: So, not "a (_headers) or (other configuration file)", but "a (_headers or other configuration) file"!

src/content/docs/en/reference/adapter-reference.mdx
ascorbic:
```suggestion
The following example defines an `args` object with a property that identifies where assets generated by Astro are located:
```

sarah11918:
@florian-lefebvre Are you available to give feedback on Armand's latest commit and questions?

florian-lefebvre:
I think we should document static methods as well, they are useful. It's fine to document these alongside the standard methods and just say they're static with a code example. For example usage of the node app specific static methods, I think this file is pretty good: https://github.com/withastro/astro/blob/5dbd59bdced080e3a9ae95ad91bbe737a2e3ba55/packages/integrations/node/src/serve-app.ts

ArmandPhilippot:
Thanks, I just pushed a new commit to document them. Please check the accuracy because I'm not really sure how this can be used (e.g. I wrap `validateForwardedHost` in a `start` function to justify where `allowedDomains` comes from... but should this be used here?).

I added `static method` each time for people reading the example too quickly as additional clue.

And, if anyone want to check the versions:
- `App.getSetCookieFromResponse()`: added in https://github.com/withastro/astro/pull/9661
- `App.validateForwardedHost()`: added in withastro/astro@6ee63bf which matches v5.14.2
- `NodeApp.createRequest()`: added in https://github.com/withastro/astro/pull/9661
- `NodeApp.writeResponse()`: added in https://github.com/withastro/astro/pull/9661

florian-lefebvre:
New static methods were just added to address a CVE FYI https://github.com/withastro/astro/pull/14743

ArmandPhilippot:
Oh thanks! I wouldn't have thought to check the latest version... I'll add them to the docs tomorrow!

ArmandPhilippot:
Thanks! And yeah, I'm glad we now have all those public APIs documented! But I must say, I didn't expected that when I started:
astro-adapter-diff
😅 

And yes, a review from @yanthomasdev is always useful, if you have any spare time! 😄 

Same for Florian or Matt, if you have time to check the accuracy of the new entries it would be helpful.
(No rush of course!)

florian-lefebvre (approved):
Looks pretty good! I think the the app methods could use more concrete examples perhaps (or maybe that's belong to a guide, idk) but not blocking at all! This PR is big enough already, thanks a lot!

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,226 @@ export default function createIntegration() {
...
+
+### `supportedAstroFeatures`
+
+<p>
+
+**Type:** `AstroAdapterFeatureMap`<br />
+<Since v="3.0.0" />
+</p>
+
+A map of Astro built-in features supported by the adapter. This allows Astro to determine which features an adapter supports so appropriate error messages can be provided.
```
yanthomasdev:
```suggestion
A map of Astro's built-in features supported by the adapter. This allows Astro to determine which features an adapter supports, so appropriate error messages can be provided.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,226 @@ export default function createIntegration() {
...
+    },
+  };
 }
 ```
 
-With the adapter API you achieve this by implementing `createExports` in your `serverEntrypoint`:
+## Building a server entrypoint
 
-```js
+You will need to create a file that executes during server-side requests to enable on-demand rendering with your particular host. Astro's adapter API attempts to work with any type of host, and gives a flexible way to conform to the host APIs.
```
yanthomasdev:
```suggestion
You will need to create a file that executes during server-side requests to enable on-demand rendering with your particular host. Astro's adapter API attempts to work with any type of host and gives a flexible way to conform to the host APIs.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -42,106 +42,226 @@ export default function createIntegration() {
...
+### `createExports()`
+
+<p>
+
+Type: `(manifest: SSRManifest, options: any) => Record<string, any>`
+</p>
+
+An exported function that takes an SSR manifest as its first argument and an object containing your adapter [args](#args) as its second argument. This should provide the exports required by your host.
+
+For example, some serverless hosts expect you to export an `handler()` function. With the adapter API you achieve this by implementing `createExports()` in your server entrypoint:
```
yanthomasdev:
```suggestion
For example, some serverless hosts expect you to export an `handler()` function. With the adapter API, you achieve this by implementing `createExports()` in your server entrypoint:
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -174,11 +294,34 @@ export default function createIntegration() {
...
+### `start()`
+
+<p>
+
+Type: `(manifest: SSRManifest, options: any) => Record<string, any>`
+</p>
+
+An exported function that takes an SSR manifest as its first argument and an object containing your adapter [args](#args) as its second argument.
 
 Some hosts expect you to *start* the server yourselves, for example by listening to a port. For these types of hosts, the adapter API allows you to export a `start` function which will be called when the bundle script is run.
```
yanthomasdev: comma nit
```suggestion
Some hosts expect you to *start* the server yourselves, for example, by listening to a port. For these types of hosts, the adapter API allows you to export a `start` function, which will be called when the bundle script is run.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -244,7 +389,7 @@ The `app.render()` method accepts a mandatory `request` argument, and an optiona
 Whether or not to automatically add all cookies written by `Astro.cookie.set()` to the response headers.
 
 When set to `true`, they will be added to the `Set-Cookie` header of the response as comma separated key-value pairs. You can use the standard `response.headers.getSetCookie()` API to read them individually.
```
yanthomasdev: hyphen nit
```suggestion
When set to `true`, they will be added to the `Set-Cookie` header of the response as comma-separated key-value pairs. You can use the standard `response.headers.getSetCookie()` API to read them individually.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,158 +508,292 @@ if(app.match(request)) {
 
 You can usually call `app.render(request)` without using `.match` because Astro handles 404s if you provide a `404.astro` file. Use `app.match(request)` if you want to handle 404s in a different way.
 
-## Allow installation via `astro add`
+By default, prerendered routes aren't returned, even if they are matched. You can change this behavior by using `true` as second argument.
```
yanthomasdev:
```suggestion
By default, prerendered routes aren't returned, even if they are matched. You can change this behavior by using `true` as the second argument.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,158 +508,292 @@ if(app.match(request)) {
...
-  <Since v="5.9.0" />
+**Type:** `(forwardedHost: string, allowedDomains?: Partial<RemotePattern>[], protocol?: string = 'https') => boolean`<br />
+<Since v="5.14.2" />
 </p>
 
-An option to prevent showing some or all logged messages about an adapter's support for a feature.
+Checks whether a `forwardedHost` matches any of the given [`allowedDomains`](/en/reference/configuration-reference/#securityalloweddomains). This static method accepts a third argument that allows you to override the host protocol, defaulting to `https`.
 
-If Astro's default log message is redundant, or confusing to the user in combination with your custom `message`, you can use `suppress: "default"` to suppress the default message and only log your message:
+The following example retrieves the `forwardedHost` from the headers and check if the host matches an allowed domain:
```
yanthomasdev:
```suggestion
The following example retrieves the `forwardedHost` from the headers and checks if the host matches an allowed domain:
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,158 +508,292 @@ if(app.match(request)) {
...
-      },
-    },
-  };
+<p>
+
+**Type:** `(hostname: string | undefined) => string | undefined`<br />
+<Since v="5.15.5" />
+</p>
+
+Validates a hostname by rejecting any name containing path separators. When an hostname is invalid, this static method will return `undefined`.
```
yanthomasdev:
```suggestion
Validates a hostname by rejecting any name containing path separators. When a hostname is invalid, this static method will return `undefined`.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,158 +508,292 @@ if(app.match(request)) {
...
 
 <p>
 
-**Type:** `boolean`
+**Type:** `(request: NodeRequest | Request, options?: RenderOptions) => Promise<Response>`<br />
+<Since v="4.0.0" />
 </p>
 
-Defines whether any on-demand rendering middleware code will be bundled when built.
+Extends [`app.render()`](#apprender) to also accept [Node.js `IncomingMessage`](https://nodejs.org/api/http.html#class-httpincomingmessage) objects in addition to standard `Request` objects as first argument. The second argument is an optional object allowing you to [control the rendering](#renderoptions).
```
yanthomasdev:
```suggestion
Extends [`app.render()`](#apprender) to also accept [Node.js `IncomingMessage`](https://nodejs.org/api/http.html#class-httpincomingmessage) objects in addition to standard `Request` objects as the first argument. The second argument is an optional object allowing you to [control the rendering](#renderoptions).
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,158 +508,292 @@ if(app.match(request)) {
...
+#### `nodeApp.headersMap`
+
+<p>
+
+**Type:** `NodeAppHeadersJson | undefined`<br />
+**Default:** `undefined`<br />
+<Since v="5.11.0" />
+</p>
+
+An array containing headers configuration. Each entry maps a pathname to a list of headers that should be applied for that route. This is useful for applying headers such as CSP directives to prerendered routes.
```
yanthomasdev:
```suggestion
An array containing the headers configuration. Each entry maps a pathname to a list of headers that should be applied for that route. This is useful for applying headers such as CSP directives to prerendered routes.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -363,158 +508,292 @@ if(app.match(request)) {
...
+
+#### `NodeApp.createRequest()`
+
+<p>
+
+**Type:** `(req: NodeRequest, options?: { skipBody?: boolean; allowedDomains?: Partial<RemotePattern>[]; }) => Request`<br />
+<Since v="4.2.0" />
+</p>
+
+Converts a NodeJS `IncomingMessage` into a standard `Request` object. This static method accepts an optional object as second argument allowing you to define if the body should be ignored, defaulting to `false`, and the [`allowedDomains`](/en/reference/configuration-reference/#securityalloweddomains).
```
yanthomasdev:
```suggestion
Converts a NodeJS `IncomingMessage` into a standard `Request` object. This static method accepts an optional object as the second argument, allowing you to define if the body should be ignored, defaulting to `false`, and the [`allowedDomains`](/en/reference/configuration-reference/#securityalloweddomains).
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -708,3 +1100,131 @@ export default function createIntegration() {
   };
 }
 ```
+
+The value of the headers might change based on the features enabled/used by the application. For example, if [CSP is enabled](/en/reference/experimental-flags/csp/), the `<meta http-equiv="content-security-policy">` element is not added to the static page. Instead its `content` is available in the `experimentalRouteToHeaders` map.
```
yanthomasdev:
```suggestion
The value of the headers might change based on the features enabled/used by the application. For example, if [CSP is enabled](/en/reference/experimental-flags/csp/), the `<meta http-equiv="content-security-policy">` element is not added to the static page. Instead, its `content` is available in the `experimentalRouteToHeaders` map. 
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -708,3 +1100,131 @@ export default function createIntegration() {
...
+
+### `AdapterSupportsKind`
+
+<p>
+
+**Type:** `"deprecated" | "experimental" | "limited" | "stable" | "unsupported"`
+</p>
+
+Defines the level of support for a feature by your adapter:
+* Use `"deprecated"` when your adapter deprecates a feature support before removing it completely in a future version.
```
yanthomasdev:
```suggestion
* Use `"deprecated"` when your adapter deprecates support for a feature before removing it completely in a future version.
```

ArmandPhilippot:
I agree with Florian, some examples could probably be improved, but without building an adapter to determine what constitutes a useful minimal code snippet, it's a bit difficult (for me) to trim what is necessary or not in the current adapters. Showing too much/too specific code could also be confusing for users I think.

At least the docs are now a bit more detailed for those who wish to build their own adapter! So, this is NWTWWHB and we can always improve the code snippets later if/when we get feedbacks.

I think this looks ready to be merged! Thank you all again for your help!

PR #12677, opened 2025-11-06
fix(astro-i18n): update wrong locale code

jsparkdev:
#### Description (required)

The locale code `pt-PR` does not seem to exist.

Based on my findings, the locales `pt-PT` and `pt-BR` exist, and given that the return value of the `normalizeTheLocale` function is `pt-pt`, it is assumed that `pt-PR` is a typo for `pt-PT`, and thus it is corrected.

- Closes #
- Suggested label:


ArmandPhilippot (approved):
Indeed this was a typo, nice catch! 🙌🏽

PR #12682, opened 2025-11-06
Update CodingInPublic ad for holiday sales

sarah11918:
#### Description (required)

Chris Pennington is offering extra holiday discounts for November/December, so this new promo code allows purchasers to take advantage of those sales while earning Astro a commission.

- Updates the promo code (URL) to `ASTRO_HOLIDAY_PROMO`
- Updates the amount of the discount to 33%

OPTIONAL: If we want to update the graphic to further emphasize holiday (Thanksgiving/Black Friday/Christmas) at any point, or to indicate that this is a special holiday promo, we are welcome to. But, we don't have to, and this at least provides the accurate discount amount.

### DIFF
--- a/src/content/docs/fr/reference/modules/astro-i18n.mdx
+++ b/src/content/docs/fr/reference/modules/astro-i18n.mdx
@@ -40,6 +40,10 @@ import {
   notFound,
   middleware,
   requestHasLocale,
+  normalizeTheLocale,
+  pathHasLocale,
+  toCodes,
+  toPaths
  } from 'astro:i18n';
 ```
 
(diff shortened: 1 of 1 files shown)


ArmandPhilippot (approved):
LGTM for the current update!
I am absolutely not a graphic designer so I can't help if we want an updated visual. 😅 This is a good idea to warn users something has changed though!

delucis (approved):
Approving as #NWTWWHB!

Would definitely be fun to jazz it up at some point, but this works for now — thanks @sarah11918 🙌

PR #12693, opened 2025-11-11
i18n(zh-TW): Revise glossary

lumynou5:
#### Description (required)

This adds some mistakable terms to the glossary to draw translators' attention and changes some translations.

There is a plan to change more, such as 水合 (hydration) due to its difficulty to understand, but discussion is needed.

Changes:
- 儲存庫 -> 版本庫 (repository): Use Git CLI translation.
- Adds 隨需算繪 (on-demand render)
- Adds 非完整頁面 (page partial)

Cc: @mingjunlu, @vrabe

### DIFF
--- a/i18n-guides/正體中文.md
+++ b/i18n-guides/正體中文.md
@@ -69,7 +69,7 @@
 | reactive              | 反應式       |                       |
 | recipe                | 操作指南     |                       |
 | reference             | 參考         |                       |
-| repository            | 儲存庫       |                       |
+| repository            | 版本庫       |                       |
 | responsive            | 響應式       |                       |
 | routing               | 路由         |                       |
 | server-side rendering | 伺服器端算繪 |                       |
--- a/src/content/docs/zh-tw/contribute.mdx
+++ b/src/content/docs/zh-tw/contribute.mdx
@@ -8,18 +8,18 @@ import ContributorList from '~/components/ContributorList.astro'
 我們歡迎任何規模的貢獻，無論你是初學者或專家。身為開源專案，我們相信應該要回饋貢獻者，因此很樂意對 PR、技術寫作、功能、想法提供指導。
 
 :::tip[未解決的 Issue]
-請瀏覽 [Astro 所有標記為 `help wanted` 及 `good first issues`，且未解決的 GitHub issue](https://github.com/search?q=org%3Awithastro+label%3A%22good+first+issue%22%2C%22help+wanted%22%2C+state%3Aopen&type=issues)，這些程式碼儲存庫包含：Astro 程式碼、文件、Starlight、Astro 網站、語言工具、GitHub Action、機器人、編譯工具……有很多地方可以貢獻！
+請瀏覽 [Astro 所有標記為 `help wanted` 及 `good first issues`，且未解決的 GitHub issue](https://github.com/search?q=org%3Awithastro+label%3A%22good+first+issue%22%2C%22help+wanted%22%2C+state%3Aopen&type=issues)，這些程式碼版本庫包含：Astro 程式碼、文件、Starlight、Astro 網站、語言工具、GitHub Action、機器人、編譯工具……有很多地方可以貢獻！
 :::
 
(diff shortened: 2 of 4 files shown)


vrabe:
I propose adding "on-demand rendering". It's a term introduced in Astro v5. I currently translate it as 隨需算繪.

lumynou5:
> I propose adding "on-demand rendering". It's a term introduced in Astro v5. I currently translate it as 隨需算繪.

I think [按](https://dict.revised.moe.edu.tw/dictView.jsp?ID=10458&la=0&powerMode=0) is perhaps better.

Also, 算繪 is a word I have doubts.  Since it doesn't mean to render something graphically but to generate HTML, I think we should choose another word.

vrabe:
The other common translation is 渲染. But it's also not so appropriate. What word do you want to choose?

lumynou5:
No, 渲染 is not *the* other.  "Render" is a word used on many different things, for example, audio, where 算繪 is obviously not appropriate.

I think we can just translate as though as it was "generate".

vrabe:
I think things to be generated are broader than things to be rendered. If we just simply replace render with generate, there will be some meaning loss.

It seems that there's no common translations for these front-end terms (including hydrate/hydration). If we want to translate them, we had better come up with the translation carefully.

lumynou5:
Sure we need.

What meaning will be lost?  I think "generating" destination files describes what really happens better than "drawing".  Or, we can create a new word if you have any idea; I guess "算" + verb (i.e. the same construction as "算繪") would be good but have no idea what to use.

Related discussion:
- https://t.me/l10n_tw/37610
- https://t.me/l10n_tw/40381
- https://t.me/l10n_tw/7962

yanthomasdev:
Hey 👋 

I see this PR got stale. I'll be closing it for now, but please feel free to open it again later!

lumynou5:
@vrabe Do you have any idea?  We can also leave it untouched.

Do you prefer 按, 隨, or 應 for "on demand"?

vrabe:
@lumynou5 No preference on those. If we can come up with a translation for "render", that would be great.

lumynou5:
How about 轉譯 (usually for "transpile" tho)?  The process of rendering here basically is transpiling source code to HTML, so is CSS preprocessors.

vrabe:
That will mislead readers because as you said 轉譯 is usually for "transpile". I think we can take the responses in this [blog post](https://medium.com/tenzblog/render-%E5%88%B0%E5%BA%95%E8%A9%B2%E6%80%8E%E9%BA%BC%E7%BF%BB-3fea26f1ec3f) into consideration.

lumynou5:
The post discusses rendering in computer graphics, and the responses probably do so.  Furthermore, 呈現 is very ambiguous as it's a common word, and 建構 might be mistaken for "construct".

I don't think 轉義 misleading reader much.  Since we're looking for a domain-specific translation, we need to take what it actually does into account; it's transpiling in this case.  Also, "render" can mean "translate" (to **render** sth. into a language), which is quite similar to the case.

vrabe:
Let's put it aside.

I‘m OK with both 儲存庫 and 版本庫. So this change looks good to me.

I think we can find other terms that are worth being added to the glossary and go back to "on demand rendering" / "render" later.

lumynou5:
I added 隨需算繪.  Change it after we find a better translation for "render".

"Hydration" is also one needed to discuss, but I don't have any idea for now.

I want to add "缺略" (or "非完整") for "partial".  I saw a page currently uses "局部", which is a bit different; page partials are imcomplete rather than parts of a whole thing.  What do you think?

lumynou5:
@yanthomasdev Could you reopen the PR so we can keep the discussion at the same place?

vrabe:
@lumynou5 It's incomplete. But I don't think we want to emphasize something is missing in that page component. So I think "缺略" is not that appropriate and "非完整" is better.

vrabe:
How about `schema -> 資料標準`? My inspiration is from [here](https://schema.gov.tw/).

lumynou5:
I think (資料)結構描述 is more descriptive.  資料標準 would make confusion: Is it "standard"? Or "specification"?

vrabe:
@lumynou5 I think it's better to open a discussion instead of discussing in a closed PR. So I am going to create that discussion. I'll ping you there.

PR #12696, opened 2025-11-11
feat(fonts): merge families

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -796,7 +796,7 @@ export default defineConfig({
 
 A font family is defined by a combination of properties such as weights and styles (e.g. `weights: [500, 600]` and `styles: ["normal", "bold"]`), but you may want to download only certain combinations of these.
 
-For greater control over which font files are downloaded, you can specify the same font (ie. with the same `cssVariable`, `name`, and `provider` properties) multiple times with different combinations and Astro will merge the results and download only the required files. For example, it is possible to download normal `500` and `600` while downloading only italic `500`:
+For greater control over which font files are downloaded, you can specify the same font (ie. with the same `cssVariable`, `name`, and `provider` properties) multiple times with different combinations. Astro will merge the results and download only the required files. For example, it is possible to download normal `500` and `600` while downloading only italic `500`:
 
 ```js
 // astro.config.mjs


src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -787,6 +787,41 @@ export default defineConfig({
...
 ```
 
+## Merging remote font families
+
+<p>
+
+<Since v="5.15.6" />
+</p>
+
+Using a single font family, it is not possible to only download font files for weights normal `500`, italic `500` and normal `600`. This is because a font family is defined by a combination of weights and styles, so you'd necessarily have to also download italic `600`.
```
sarah11918: Is the real change here that you can *define the same font multiple times*, and whatever you ask for will be combined and downloaded? (So, you can make multiple, smaller, more specific requests for what you want to download?)

You start by defining a very specific example case rather than the general idea here, so it makes it harder to understand what this change actually is.

florian-lefebvre: Yeah families will be merged if they share the same cssVariable, name and provider. I started with the example because that's how this feature came to be, based on feedback, but I agree it may not be ideal in terms of structure

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -787,6 +787,41 @@ export default defineConfig({
...
+For more advanced cases, families that have the same `cssVariable`, `name` and `provider` will be merged. That means you could achieve the desired result like this:
```
```suggestion
A font family is defined by a combination of weights and styles (e.g. `weights: [500, 600]` and `styles: ["normal", "bold"]`), but you may want to download only certain combinations of these.

For greater control over which font files are downloaded, you can specify the same font (ie. with the same `cssVariable`, `name`, and `provider` properties) multiple times with different combinations and Astro will merge the results and download only the required files. For example, it is possible to download normal `500` and `600` while downloading only italic `500`:

```
So this wording might need some help, but what about something like this? Is that representative of the feature?

sarah11918:
(Also noting it looks like the Astro PR is already merged and we didn't discuss its changeset!)

florian-lefebvre:
Oh I'm sorry! Since its a patch I completely forgot. We didn't released the ci PR yet so we can still edit the changeset (https://github.com/withastro/astro/blob/main/.changeset/tired-houses-sink.md)

florian-lefebvre:
I slightly edited your suggestion

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -787,6 +787,43 @@ export default defineConfig({
 });
 ```
 
+## Merging remote font families
```
sarah11918: I think the content looks good now! But I'm wondering whether this is the most helpful title for the section?

Will people be thinking, "I want to merge my remote font families. I wonder how to do that?" 😄 
The "merging" part is really an implementation detail, and it's not something the user does, but how Astro handles a config like this.

It seems to me that someone who wants to know about this might be thinking something more like...

- How do I  minimize the files I download?
- being ultra selective about which font files get downloaded
- further customizing download combinations
- "splitting" the font attributes I want

Can you think of a section heading that might be more along these lines? What do you think?

florian-lefebvre: Yeah I think it would frame it as "you need advanced control, look at this". Most people should/will not bother doing this

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -787,6 +787,43 @@ export default defineConfig({
...
+
+<p>
+
+<Since v="5.15.6" />
+</p>
+
+A font family is defined by a combination of weights and styles (e.g. `weights: [500, 600]` and `styles: ["normal", "bold"]`), but you may want to download only certain combinations of these.
```
florian-lefebvre:
```suggestion
A font family is defined by a combination of properties such as weights and styles (e.g. `weights: [500, 600]` and `styles: ["normal", "bold"]`), but you may want to download only certain combinations of these.
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
```suggestion
## Granular remote font configuration
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -787,6 +787,43 @@ export default defineConfig({
...
+## Granular remote font configuration
+
+<p>
+
+<Since v="5.15.6" />
+</p>
+
+A font family is defined by a combination of properties such as weights and styles (e.g. `weights: [500, 600]` and `styles: ["normal", "bold"]`), but you may want to download only certain combinations of these.
+
+For greater control over which font files are downloaded, you can specify the same font (ie. with the same `cssVariable`, `name`, and `provider` properties) multiple times with different combinations and Astro will merge the results and download only the required files. For example, it is possible to download normal `500` and `600` while downloading only italic `500`:
```
florian-lefebvre:
```suggestion
For greater control over which font files are downloaded, you can specify the same font (ie. with the same `cssVariable`, `name`, and `provider` properties) multiple times with different combinations. Astro will merge the results and download only the required files. For example, it is possible to download normal `500` and `600` while downloading only italic `500`:
```

PR #12701, opened 2025-11-12
[next] remove app.setManifestData()

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -463,7 +463,7 @@ Astro 6.0 removes this method entirely.
 
 #### What should I do?
 
-Remove any call to `app.setManifestData()`.
+Remove any call to `app.setManifestData()`. If you need to update the manifest, create a new `App` instance.
 
 <ReadMore>Learn more about the [Adapter API](/en/reference/adapter-reference/).</ReadMore>
 


florian-lefebvre:
I know this is very succinct but that's because I can't think of any good reason this would be used. There's not even a single public adapter using it on GitHub.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -453,6 +453,20 @@ app.render(request, { routeData, locals })
 
 <ReadMore>Learn more about the [Adapter API](/en/reference/adapter-reference/).</ReadMore>
 
+### Removed: `app.setManifestData()` (Adapter API)
+
+<SourcePR number="14758" title="chore(astro)!: remove app.setManifestData()"/>
+
+In Astro 5.0, the `app.setManifestData()` method was available on `App` and `NodeApp`, but was never used nor useful.
```
```suggestion
In Astro 5.0, the `app.setManifestData()` method was available on `App` and `NodeApp`, but is no longer used nor needed.
```
Poor method! This at least makes it sound like it was good for something at some point! 😂

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -453,6 +453,20 @@ app.render(request, { routeData, locals })
...
+
+Astro 6.0 removes this method entirely.
+
+#### What should I do?
+
+Remove any call to `app.setManifestData()`.
```
ArmandPhilippot: Yeah, I understand the struggle after trying to document it. 😄 
I guess if we wanted to be more verbose, we could say something like:
> If you were updating the manifest, you must now do that before creating an `App` or a `NodeApp`.

(since we pass the manifest to them, either from `createExports()` or `start()`)

But... without knowing any use cases, I'm not sure if this is helpful? So, with Sarah's suggestion this is probably fine as it is!

sarah11918: Is there *any* suggestion we can add here like, "Adapters do not need access to this method" or "Adapters can set manifest data in dev mode using...." which can reassure them? Would be nice if we had *something*

florian-lefebvre: I guess the closest would be something like this:
```suggestion
Remove any call to `app.setManifestData()`. If you need to update the manifest, create a new `App` instance.
```

sarah11918: Sounds good! (Even if no one needs to do this, it makes it sound/feel very comforting that there is a thing that can be done, and makes it look like we're taking care of people!)

PR #12706, opened 2025-11-13
fix(adapter-reference): small nitpicks (verb tense, indentation...)

ArmandPhilippot:
#### Description (required)

While translating in French, I noticed a few more fixes for the Adapter reference:
* verb tense in two locations
* redundant `.e.g.` with `etc.`
* consistent use of backticks for code
* consistent use of parenthesis for functions as inline code
* a code snippet was malformed
* the indentation in code snippets was inconsistent (some used 4 spaces, a mix of 2 and 4 spaces, or tabs)

- Suggested label:  `code snippet update`, `consistency/formatting`, `typo/link/grammar - quick fix!`

### DIFF
--- a/src/content/docs/en/reference/adapter-reference.mdx
+++ b/src/content/docs/en/reference/adapter-reference.mdx
@@ -166,7 +166,7 @@ export default function createIntegration() {
 **Type:** `URLSearchParams`
 </p>
 
-Defines the query parameters to append to all asset URLs (images, stylesheets, scripts, etc.). This is useful for adapters that need to track deployment versions or other metadata.
+Defines the query parameters to append to all asset URLs (e.g. images, stylesheets, scripts). This is useful for adapters that need to track deployment versions or other metadata.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -166,7 +166,7 @@ export default function createIntegration() {
 **Type:** `URLSearchParams`
 </p>
 
-Define the query parameters to append to all asset URLs (e.g. images, stylesheets, scripts, etc.). This is useful for adapters that need to track deployment versions or other metadata.
+Defines the query parameters to append to all asset URLs (images, stylesheets, scripts, etc.). This is useful for adapters that need to track deployment versions or other metadata.
```
```suggestion
Defines the query parameters to append to all asset URLs (e.g. images, stylesheets, scripts). This is useful for adapters that need to track deployment versions or other metadata.
```
Great call to remove the redundancy, but our standard way is to do it this way! :laughing:

ArmandPhilippot: Haha, I was hesitating! Although both works in French, I think I prefer the other way so I was influenced. 😀

PR #12709, opened 2025-11-14
Fix the phrasing `params provided to getStaticPaths()` in Routing docs

MareStare:
#### Description (required)

Fix for a grammar mistake.

Frankly, this section of the docs is a bit confusing. I don't understand why it says that `params` are provided **to** the function `getStaticPaths()` when this function takes zero parameters and can't capture outside variables. The example is even more confusing as it just uses `decodeURI("%5Bpage%5D")` on a static string literal. Maybe it makes sense to re-do/remove this section completely?

### DIFF
--- a/src/content/docs/ko/guides/cms/jekyllpad.mdx
+++ b/src/content/docs/ko/guides/cms/jekyllpad.mdx
@@ -1,0 +1,20 @@
+---
+title: JekyllPad & Astro
+description: JekyllPad의 브라우저 기반, GitHub 백엔드 편집기로 Git이나 Markdown의 번거로움 없이 Astro 사이트 콘텐츠를 관리하세요.
+sidebar:
+  label: JekyllPad
+type: cms
+service: JekyllPad
+i18nReady: true
+stub: true
+---
+
+[JekyllPad](https://www.jekyllpad.com)는 GitHub 저장소에 직접 연결되는 가벼운 브라우저 기반 CMS입니다.
+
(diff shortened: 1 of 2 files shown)


src/content/docs/en/guides/routing.mdx
```diff
@@ -88,7 +88,7 @@ Parameters can be included in separate parts of the path. For example, the file
 
 #### Decoding `params`
 
-The `params` provided to the function `getStaticPaths()` function are not decoded. Use the function [`decodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI) when you need to decode parameter values.
+The `params` provided to the function `getStaticPaths()` are not decoded. Use the function [`decodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI) when you need to decode parameter values.
```
delucis: I guess it might make more sense to describe it as “returned by” rather than “provided to”:

```suggestion
`params` returned by a `getStaticPaths()` function are not decoded. Use [`decodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI) when you need to decode parameter values.
```

Re: the static string thing — this is just an example, but I assume it’s demonstrating the concept for when you have e.g. a remote source of slugs that includes encoded parameters? It’s a pretty niche case, that’s for sure.

delucis (commented):
Thanks @MareStare! Left a suggestion that might help make this section a bit more correct?

src/content/docs/en/guides/routing.mdx
MareStare: Yeah, makes sense. Applied!

PR #12723, opened 2025-11-15
Add Codeberg Pages deploy docs

Wolffyhtl:
#### Description (required)

Add Codeberg Pages deploy docs in Chinese Simplified

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/deploy/codeberg.mdx
+++ b/src/content/docs/en/guides/deploy/codeberg.mdx
@@ -6,7 +6,7 @@ sidebar:
 type: deploy
 logo: codeberg
 supports: ['static']
-i18n: true
+i18nReady: true
 ---
 
 import { Steps } from '@astrojs/starlight/components';


clinasama:
????

sarah11918:
I  fixed the issues with the build and Steps component, however as I started to edit and gather links to Codeberg Pages, their main site claims that the project is in maintenance mode. I'm not sure whether we should be adding this guide at all since they have a large warning about not using it for important production sites.

What do people think? From: https://docs.codeberg.org/codeberg-pages/

image

src/content/docs/en/guides/deploy/codeberg.mdx
```diff
@@ -0,0 +1,133 @@
+---
+title: Deploy Your Astro Site to Codeberg Pages
+description: How to use Codeberg Pages to deploy your Astro site to the web.
+sidebar:
+   label: Codeberg Pages
+type: deploy
```
sarah11918, line 6:
```suggestion
type: deploy
logo: codeberg
supports: ['static']
i18n: true
```
updating for new syntax for deploy guides!

src/content/docs/en/guides/deploy/codeberg.mdx
```diff
@@ -0,0 +1,136 @@
...
+logo: codeberg
+supports: ['static']
+i18n: true
```
sarah11918:
```suggestion
i18nReady: true
```

sarah11918:
Just wanted to let you know @Wolffyhtl that we discussed whether we should include this guide in the Astro Docs and we've decided not to add this guide.

The Codeberg Pages project has limited maintenance resources (it's discouraged for use in production, and you have to get approval first before using it), so it doesn't seem right to include it here in our official documentation.

I think this content would be a great personal blog post so that someone searching can find an easy setup guide, but it doesn't seem like this is a project we should be promoting, since the project might struggle with a lot of new users who learn about it through Astro Docs. I'm sorry we can't publish it in our official documentation, but if you write a blog post, it can be shared in our community on our Discord `#showcase` channel, or submitted to [Astro Tips](https://astro-tips.dev).

PR #12728, opened 2025-11-17
Remove Sentry as sponsor

sarah11918:
#### Description (required)

Updates our sponsor section in the left sidebar to remove Sentry. (Sponsorship expired 2025-10-31)

### DIFF
--- a/src/content/docs/fr/guides/deploy/juno.mdx
+++ b/src/content/docs/fr/guides/deploy/juno.mdx
@@ -1,0 +1,173 @@
+---
+title: Déployer votre site Astro sur Juno
+description: Comment déployer votre site Astro sur le web à l'aide de Juno.
+sidebar:
+  label: Juno
+type: deploy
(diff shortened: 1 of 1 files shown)


src/components/LeftSidebar/Sponsors.astro
```diff
@@ -91,13 +90,6 @@ const instanceId = Math.floor(Math.random() * 100_000);
 	{/* Spacer element to push following logos onto a new row. */}
 	<div style="flex-basis: 100%"></div>
```
ArmandPhilippot, line 91: I'm not sure why this was added in the first place, but while resizing the browser to check if everything was fine in different resolution, this seems a bit weird. The Mux logo is on its own line with small viewports while there is room for it next to Cloudflare:

Image

Removing the `div` in the browser doesn't seem to produce bugs (edit: I checked with Firefox and Chromium), so maybe this is no longer necessary?

PR #12730, opened 2025-11-18
fonts: clarify adobe id loading

florian-lefebvre:
#### Description (required)

Our example of the adobe provider was wrong: `process.env` is not necessarily loaded by default. This PR updates this section to make it clearer and link to the relevant docs.

I also added some `hr` tags under font providers with text to make it standout more with the sentence below about custom font providers.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -152,7 +152,9 @@ body {
 
 ## Available remote font providers
 
-Astro re-exports most [unifont](https://github.com/unjs/unifont/) providers. The following have built-in support:
+Astro re-exports most [unifont](https://github.com/unjs/unifont/) providers. You can also [make a custom Astro font provider](#build-your-own-font-provider) for any unifont provider.
+
+The following providers have built-in support:
 
 - [Adobe](https://fonts.adobe.com/)
 - [Bunny](https://fonts.bunny.net/)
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/experimental-flags/fonts.mdx
ArmandPhilippot: I don't mind the `hr`, but I wonder if it wouldn't be better to put it above this line rather than only in some tabs?

I mean, when switching between tabs I found the inconsistency a bit confusing. This gives me the impression the text belong to tabs without the `hr` while this is not the case for the ones with `hr`.
||Before|After|
|---|---|---|
|Fontsource|Fontsource tab without hr|Fontsource tab with hr|
|Google|Google tab with hr|Google tab with hr|

I don't know, maybe this is just a me thing. 😅

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -212,6 +218,8 @@ provider: fontProviders.google({
 })
 ```
 
+---
```
sarah11918: Do not like :sweat_smile: : 

How about instead, if this add-on sentence is getting problematic for displaying, line 155 becomes something like:

```
Astro re-exports most [unifont](https://github.com/unjs/unifont/) providers. You can also [make a custom Astro font provider](#build-your-own-font-provider) for any unifont provider.

The following providers have built-in support:
```

And then remove the sentence in line 227. The very next thing after the tabs will be a `##`, so there's no confusion there.

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -167,9 +167,15 @@ To use a built-in remote provider, configure `provider` with the appropriate val
 <TabItem label="Adobe">
 
 ```js
-provider: fontProviders.adobe({ id: process.env.ADOBE_ID })
+provider: fontProviders.adobe({ id: 'your-id' })
 ```
 
+It is recommended to load the ID as an environment variable.
```
sarah11918: Since this will likely be the first entry people see (the default tab visible), this sentence should be clear it's about Adobe only. Maybe something like (and, it's ok to get opinionated!):

```suggestion
Pass the Adobe font provider an ID loaded as an [environment variable in your Astro config file](/en/guides/environment-variables/#in-the-astro-config-file)
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
florian-lefebvre: Yeah I like that!

PR #12731, opened 2025-11-18
fonts: updated default subset

florian-lefebvre:
#### Description (required)

The default subset is updated. I don't think we need to add a `Since` for this

- Closes #
- Suggested label:


florian-lefebvre:
I forgot to wait for the associated PR again...will be merged soon so shouldn't be a problem

PR #12732, opened 2025-11-18
feat: astro info --copy flag

florian-lefebvre:
#### Description (required)

This flag was never documented

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/cli-reference.mdx
+++ b/src/content/docs/en/reference/cli-reference.mdx
@@ -313,7 +313,7 @@ User preferences are scoped to the current project by default, stored in a local
 
 Using the `--global` flag, user preferences can also be applied to every Astro project on the current machine. Global user preferences are stored in an operating system-specific location.
 
-### Available preferences
+<h3>Available preferences</h3>
 
 - `devToolbar` — Enable or disable the development toolbar in the browser. (Default: `true`)
 - `checkUpdates` — Enable or disable automatic update checks for the Astro CLI. (Default: `true`)


src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -297,6 +297,13 @@ Adapter                  none
 Integrations             @astrojs/starlight (v0.35.3)
 ```
 
+### Flags
+
+
```
florian-lefebvre, line 303:
```suggestion
```

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -297,6 +297,12 @@ Adapter                  none
 Integrations             @astrojs/starlight (v0.35.3)
 ```
 
+### Flags
+
```
ArmandPhilippot, line 301: The other flags use an introductory sentence:
```suggestion

Use the following flags to customize the behavior of the command.

```
(I chose the plural to be future-proof, but we can use the singular if you prefer.)

src/content/docs/en/reference/cli-reference.mdx
```diff
@@ -216,7 +216,7 @@ Runs diagnostics (such as type-checking within `.astro` files) against your proj
 
 This command is intended to be used in CI workflows.
 
-<h3>Flags</h3>
+### Flags
```
ArmandPhilippot: I think using `h3` instead of `###` was deliberate here: this prevents the heading to be displayed in the table of contents.
But comparing with the live version, I can see we don't do that for `astro build` so yeah it's probably better to stay consistent! 😄

sarah11918: oooh, I'm not sure I like any of those flags at all :sweat_smile: Especially when usually they're immediately visible after only a sentence if you click on the main entry. They're not hard to find! 

Any objections to making them ALL h3?

ArmandPhilippot: Not from me, as long as it remains consistent! 😁

florian-lefebvre: If you don't mind, that can be a follow up PR!

sarah11918: It could be, but you've already kind of polluted this PR by updating multiple sections here. :smile: 

As long as it works out in the end!

florian-lefebvre: ahah fair enough :sweat_smile:. I'm not sure entirely what you want tho, what would the `### Flags` heading become? If flags themselves are h3s?

ArmandPhilippot: I think the idea is only to swap Markdown syntax with HTML to hide them in the ToC:
```diff
-### Flags
+<h3>Flags</h3>
```
(including the one that is visible in the live website)

sarah11918: Yes exactly. We don't want any of these "Flags" entries to show up in the TOC, so we'll use the HTML element instead of Markdown syntax to label them as headings. That's all!

(It's how Starlight works -- only Markdown headings (`###`) actually show up in the right sidebar TOC. We use HTML `<h>` elements when we want to be sneaky and avoid something ending up in the table of contents, ruining our beautifully crafted page outline!)

PR #12734, opened 2025-11-18
fix(astro:transitions): add missing docs for available public APIs

ArmandPhilippot:
#### Description (required)

- Updates `astro:transitions` reference to:
  - Fix `<Since />` in `<ClientRouter />` (added in v5, https://github.com/withastro/astro/pull/11980)
  - Fix the [`transitionEnabledOnThisPage`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/router.ts#L36-L37) type
  - Fix types in "Lifecycle events attributes":
    - [`info`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L27) is typed as `any` and this matches what we describe ("arbitrary data").
    - [`direction`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L24) is typed as `string` (because `string` is wider than `Direction`) and using `Direction` was a bit confusing given the description ("anything else").
  - Add `Type:`/`<Since />` for "Lifecycle events"
  - Add [`fallback`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/components/ClientRouter.astro#L5) as available prop for `<ClientRouter />` (hidden in the ToC)
  - Add default return value for [`getFallback()`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/router.ts#L134)
  - Add missing docs for public APIs added in https://github.com/withastro/astro/pull/9090:
    - [`createAnimationScope`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/runtime/server/transition.ts#L118-L129) (edit: removed from docs because no one seems to use it at the moment and this will be deprecated),
    - [`isTransitionBeforePreparationEvent`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L70-L107),
    - [`isTransitionBeforeSwapEvent`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L113-L143),
    - [`TRANSITION_BEFORE_PREPARATION`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L5),
    - [`TRANSITION_AFTER_PREPARATION`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L6),
    - [`TRANSITION_BEFORE_SWAP`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L7),
    - [`TRANSITION_AFTER_SWAP`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L8),
    - [`TRANSITION_PAGE_LOAD`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/events.ts#L9),
    - [`type NavigationTypeString`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/types.ts#L3),
    - [`type TransitionBeforePreparationEvent`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/client.d.ts#L154),
    - [`type TransitionBeforeSwapEvent`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/client.d.ts#L156)
  - Add docs for [`type Direction`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/types.ts#L2) and [`type Fallback`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/types.ts#L1) added in https://github.com/withastro/astro/pull/8571
  - Add some `<ReadMore />` around existing content
  - Specify when the lifecycle event attributes are available (for users not coming from the event itself, this can be confusing without that)
  - Slightly reword some existing content

- Updates "View Transitions" guide to:
  - Add some links to the reference
  - Replace `navigate()` options with a link (we already document them, the same way, in the reference)
  - Replace the `history options` link

#### Not documented

- [`ClientRouterModule`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/client.d.ts#L126), [`TransitionModule`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/client.d.ts#L121) from `astro:transitions`: although importable those are internal types
- [`EventModule`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/client.d.ts#L147), [`TransitionRouterModule`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/client.d.ts#L135), [`TransitionSwapFunctionModule`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/client.d.ts#L160) from `astro:transitions/client`: although importable those are internal types
- [`type Options`](https://github.com/withastro/astro/blob/ad3265daf8651c6e15243f5ba85165da5ca4652d/packages/astro/src/transitions/types.ts#L4-L10): I think it makes more sense to details the properties under `navigate()` so I wasn't sure how to document the type. I left it in the available imports from `astro:transitions/client` though.
- The `TransitionAnimation`, `TransitionAnimationPair`, `TransitionAnimationValue`, `TransitionDirectionalAnimations` types (currently partially [displayed in the guide](https://docs.astro.build/en/guides/view-transitions/#customizing-animations)) because they are exported from `astro`, not `astro:transitions`.


src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -93,16 +110,80 @@ import { slide } from 'astro:transitions';
 <div transition:animate={slide({ duration: '0.4s' })} />
 ```
 
+### `createAnimationScope()`
```
ArmandPhilippot: This whole section and the code snippet might need a review from someone used to view transitions (and Astro's implementation). This is my understanding but maybe I'm wrong here. 😅 
Also, the code snippet is a bit heavy, but if I understand correctly, we need all of this to have a working example?

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -278,6 +358,287 @@ Stores the element in focus on the current page and returns a function that when
...
+
+### `TRANSITION_BEFORE_PREPARATION`
+
+<p>
+
+**Type:** `'astro:before-preparation'`<br />
+<Since v="3.6.0" />
+</p>
+
+A constant to avoid writing the `astro:before-preparation` event name in plain text when you define an event.
```
ArmandPhilippot, line 389: This is probably for Sarah: I struggled a bit to describe the constants (this one and the followings). I'm not sure this is the best description, but I think those are useful to document.

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -373,26 +736,24 @@ The document for the next page in the navigation. The contents of this document
...
+**Type:** `string`
 </p>
 
-The direction of the transition.
-- `forward`: navigating to the next page in the history or to a new page.
-- `back`: navigating to the previous page in the history.
-- Anything else some other listener might have set.
+The direction of the transition. This can be a predefined [`Direction`](#direction) or anything else some other listener might have set.
+
+In the [`astro:before-preparation` event](#astrobefore-swap-event), the value is writable and accepts any `string`. In the [`astro:before-swap` event](#astrobefore-swap-event), the value is readonly.
```
ArmandPhilippot: Is it true ("some other listener might have set")? I'm not familiar enough with view transitions but from my understanding, this can only be defined inside an `astro:before-preparation` event so the current description might be confusing?

I reused the wording we had, and added a sentence below to clarify that, although common to both, this attribute does not have the same purpose depending on the event.

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -416,20 +777,20 @@ The URL of the page being navigated to. This property can be modified, the value
 
 <p>
 
-**Type:** `FormData | undefined`
+**Type:** `FormData | undefined`<br />
+**Available in:** [`astro:before-preparation` event](#astrobefore-preparation-event)
```
ArmandPhilippot, line 785: I wasn't sure if this should be plain text or part of the API block, but I think this is helpful to add this information for someone getting there from a direct link, an AI, etc.

ArmandPhilippot:
Anyone is welcome to review this PR, but I'll ping @matthewp and @martrapp because they know better the view transitions API. If one of you two has the time to check the accuracy of the new sections this would be helpful! No rush!

**Context:** Some imports weren't documented, at least in the reference page. I added docs for them and fixed some types/since while I was there. The details is available in the first post. You're free to check the whole PR, but your help is especially useful for [`createAnimationScope()`](https://deploy-preview-12734--astro-docs-2.netlify.app/en/reference/modules/astro-transitions/#createanimationscope) and the [`direction` attribute](https://deploy-preview-12734--astro-docs-2.netlify.app/en/reference/modules/astro-transitions/#direction-1) (see [my review comment](https://github.com/withastro/docs/pull/12734#discussion_r2539754453) for why I need you here).

Also, is there a reason for the `TransitionAnimation`, `TransitionAnimationPair`, `TransitionAnimationValue`, `TransitionDirectionalAnimations` types to be exported from `astro` instead of `astro:transitions`?
Not a real issue, I ask the question because it would have been convenient to have them in the same place so that they could be documented on the reference page, and avoid [displaying them as a code snippet](https://deploy-preview-12734--astro-docs-2.netlify.app/en/guides/view-transitions/#customizing-animations) in the guide.

martrapp:
Hi @ArmandPhilippot 👋🏼 thank you for taking care and fixing!

I'm afraid, some of these are long forgotten youthful mistakes and which we should mark them as deprecated:
- TRANSITION_AFTER_PREPARATION, TRANSITION_AFTER_SWAP, TRANSITION_BEFORE_PREPARATION, TRANSITION_BEFORE_SWAP, TRANSITION_PAGE_LOAD. 
When I introduced them, Matthew told me that Astro does typically not provide constants like this, but I missed the chance to take them out again.
- `isTransitionBeforePreparationEvent`, `isTransitionBeforeSwapEvent`: I found those narrowing function useful for Typescript. That was before I learned from Erika, how this is done right (types/public/view-transitions.ts)

ArmandPhilippot:
Thanks for the feedback, Martin! Well, even if we end up not documenting them because we prefer to deprecate them, this allows us to identify what we should clean up in a future version! The same thing happened with `astro:actions`: https://github.com/withastro/astro/issues/14608 😄

martrapp:
> Also, is there a reason for the TransitionAnimation, TransitionAnimationPair, TransitionAnimationValue, TransitionDirectionalAnimations types to be exported from astro instead of astro:transitions?

Good question, that never really stood out to me. @matthewp? 

> Direction: 

Yes, you are right, this can only be changed in a listener for `astro:before-preparation`. 

Details: For links and form submission, direction is automatically set to `forward`. History traversals set this to `forward` or `back`. The value is available in the TransitionBeforePreparationEvent and can be changed using a listener to `astro:before-preparation`. The value is used to control the animation direction right before `astro:before-swap` is dispatched. Listeners to `astro:before-swap` can still read the value, but in that event, it is write protected.

While normal navigation is `forward` and `back`, this limits you to two types of animation on navigation. So, yes, an idea was to support more than those two and hence allowing direction to have arbitrary values. We never pushed that much and integration with Astro scopes and the generated styles is very low-level.

Your `createAnimationScope` example looks good, but I'll need more time to review ;-)

florian-lefebvre:
Armand feel free to create a new issue on the main repo and assign it to the v6 milestone once you know what we should deprecated/remove!

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -93,18 +110,80 @@ import { slide } from 'astro:transitions';
...
+  },
+  south: {
+    new: { name: "slideFromBottom" },
+    old: { name: "slideFromTop" },
+  },
+});
+---
+
+<main data-astro-transition-scope={scope}>
+  <Fragment set:html={styles} />
```
martrapp:
```suggestion
  <style set:text={styles} />
```

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -93,18 +110,80 @@ import { slide } from 'astro:transitions';
...
+The following example creates a custom animation for a top-to-bottom navigable gallery, associates the direction in an event listener, and defines the CSS animations:
+
+```astro title="src/components/Main.astro"
+---
+import { createAnimationScope } from "astro:transitions";
+
+const { scope, styles } = createAnimationScope("gallery", {
+  north: {
+    new: { name: "slideFromTop" },
+    old: { name: "slideFromBottom" },
```
martrapp:
```suggestion
    old: { name: "slideToBottom" },
```

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -93,18 +110,80 @@ import { slide } from 'astro:transitions';
...
+  },
+  south: {
+    new: { name: "slideFromBottom" },
+    old: { name: "slideFromTop" },
```
martrapp:
```suggestion
    old: { name: "slideToTop" },
```

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -93,18 +110,80 @@ import { slide } from 'astro:transitions';
...
+<style>
+  @keyframes slideFromTop {
+    from { transform: translateY(-100%); }
+    to { transform: translateY(0); }
+  }
+
+  @keyframes slideFromBottom {
+    from { transform: translateY(100%); }
+    to { transform: translateY(0); }
+  }
```
martrapp:
```suggestion
  @keyframes slideFromTop {
    from { transform: translateY(-100%); opacity: 0 }
  }
  @keyframes slideFromBottom {
    from { transform: translateY(100%); opacity:0 }
  }
  @keyframes slideToTop {
    to { transform: translateY(-100%); opacity:0}
  }
  @keyframes slideToBottom {
    to { transform: translateY(100%); opacity: 0}
  }
```

martrapp:
> Your `createAnimationScope` example looks good, but I'll need more time to review ;-)

The main benefit of `createAnimationScope` is that it automatically creates the CSS for the fallback mode when the browser has no native support for view transitions. As of late 2025, current versions of all major browsers have native support.  

The method was very niche during the last two years, but now, it might become obsolete. Another deprecation candidate?

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: I hope you don't mind, I'm editing your suggestion to fix an indentation before committing all your suggestions! Thank you for you help! 🙌🏽

ArmandPhilippot:
I must say I'm late to the party... probably because I'm a Firefox user. 😆 I only try `<ClientRouter />` recently and I'm by no mean an expert in view transitions. So, thank you, Martin, for reviewing this, fixing things and giving suggestions for the future of this module!

If I summarize, all the new docs additions (not the small fixes) are no longer relevant and it's probably best to deprecate them? This means:
* `createAnimationScope()`
* `isTransitionBeforePreparationEvent()`
* `isTransitionBeforeSwapEvent()`
* `TRANSITION_BEFORE_PREPARATION`
* `TRANSITION_AFTER_PREPARATION`
* `TRANSITION_BEFORE_SWAP`
* `TRANSITION_AFTER_SWAP`
* `TRANSITION_PAGE_LOAD`

Given your explanations this seems reasonable.

I was checking on GitHub the use of `createAnimationScope()`, and [except a certain martrapp](https://github.com/martrapp/astro-vtbot-website/blob/fbdd5322c8c7d69431c3efdfe0777aef1cb454d0/src/layouts/AnimationMdxLayout.astro#L117) 🤔  it seems no one use this. I saw a few imports in other projects, but without actually using it!? So, maybe this one could stay undocumented if we deprecate it.

However, I was able to find a few occurrences for the other ones. So, having docs for them because they exist and are used might be okay for now.

I'll wait a bit before opening an issue on the core repo in case someone more familiar with view transitions in the framework team want to weight on this!

martrapp:
> I was checking on GitHub the use of createAnimationScope(), and [except a certain martrapp](https://github.com/martrapp/astro-vtbot-website/blob/fbdd5322c8c7d69431c3efdfe0777aef1cb454d0/src/layouts/AnimationMdxLayout.astro#L117) 🤔 it seems no one use this.

Haha, yes, I also found that one tonight ;-) There might still be some older Firefoxes that benefit from that, but time changes a lot of things.  

I'm sorry for he mess you've found and want to thank you very much for looking into this and fixing!

ArmandPhilippot:
For Martin, and anyone who wasn't in the Talking and Doc'ing event from today, I moved our schedule deprecation to a new section with a caution because we don't want people to think it's new and that they should use it.

One thing that bothered me with the new ToC is:
astro-docs-toc
(the parenthesis on a new line)

So, I added one more commit to wrap those headings with `<h4 />`. This way, we hide them from the ToC while keeping the `swapFunctions` subheadings.

This now looks ready for final reviews to me! cc @sarah11918

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -95,16 +111,22 @@ import { slide } from 'astro:transitions';
...
-  } from 'astro:transitions/client';
-</script>
+```ts
+import {
+  getFallback,
+  navigate,
+  supportsViewTransitions,
+  swapFunctions,
+  transitionEnabledOnThisPage,
+  /* The following are schedule for deprecation in v6: */
```
sarah11918:
```suggestion
  /* The following are scheduled for deprecation in v6: */
```

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -146,9 +168,7 @@ This option follows the [`history` option](https://developer.mozilla.org/en-US/d
 <Since v="3.5.0" />
 </p>
 
-A `FormData` object for `POST` requests.
-
-When this option is provided, the requests to the navigation target page will be sent as a `POST` request with the form data object as the content.
+When provided, the requests to the navigation target page will be sent as a `POST` request with the `FormData` object as the content.
```
sarah11918: Just checking that we are intentionally removing "A `FormData` object for `POST` requests." since we often start these with more of a "definition" of what things are.

I get that it's redundant since we do say the type, but we do typically use the first line to say what a thing *is* (vs. eg not what happens when this thing exists.) So, I'm checking!

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -212,18 +232,19 @@ Whether or not the current page has view transitions enabled for client-side nav
 
 <p>
 
-**Type:** `() => 'none' | 'animate' | 'swap'`<br />
+**Type:** <code>() => <a href="#fallback">Fallback</a></code><br />
 <Since v="3.6.0" />
 </p>
 
-Returns the fallback strategy to use in browsers that do not support view transitions.
+Returns the fallback strategy to use in browsers that do not support view transitions. This returns `animate` by default when the strategy has not been overridden.
```
sarah11918:
```suggestion
Returns the fallback strategy to use (`animate` by default) in browsers that do not support view transitions.
```
I think we can say this more concisely!

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -280,34 +301,337 @@ Stores the element in focus on the current page and returns a function that when
 
 Replaces the old body with the new body. Then, goes through every element in the old body that should be persisted and have a matching element in the new body and swaps the old element back in place.
 
+### Deprecated imports
+
+:::caution[Schedule for deprecation]
```
sarah11918:
```suggestion
:::caution[Scheduled for deprecation]
```

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -280,34 +301,337 @@ Stores the element in focus on the current page and returns a function that when
...
+The following imports are schedule for deprecation in v6. You can still use them in your project, but you may prefer to update your code now.
```
sarah11918:
```suggestion
The following imports are scheduled for deprecation in v6. You can still use them in your project, but you may prefer to update your code now.
```
Nice guidance!

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -280,34 +301,337 @@ Stores the element in focus on the current page and returns a function that when
...
+:::
+
+<h4>`isTransitionBeforePreparationEvent()`</h4>
```
sarah11918, line 311: I agree with hiding these from the TOC! Another option is to verify that we're actually making getting value from showing `<h4>`s at all. We could simply revert to the default of showing only `<h2>` and `<h3>`. But no objections here, especially since these are slated for deprecation. Happy that they will not show!

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -280,34 +301,337 @@ Stores the element in focus on the current page and returns a function that when
...
+
+<p>
+
+**Type:** `(value: any) => boolean`<br />
+<Since v="3.6.0" />
```
sarah11918, line 316: It's *probably* worth it to also include some kind of "deprecated" for each of these individually, in case someone just lands here from searching or something like that? 

Imagine someone ending up partway through this section, reading stuff like "this can be useful for..." and thinking great! and not realizing we don't recommend using these because the note we have about these being deprecated is no where in view.

We don't want to make these sound too appealing! 😂 

I'm not looking at what our precedent is, might just be a `:::caution`.

src/content/docs/en/reference/modules/astro-transitions.mdx
sarah11918: Going off of my earlier comment, where we don't want to make these sound too appealing anyway... it's also OK if these are "under"documented. I think at this point, our criteria is, "Is the statement true?" and we can worry less about whether it is "helpful" 😄 

Those words mean something, so I think we can consider it a win!

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -280,34 +301,337 @@ Stores the element in focus on the current page and returns a function that when
...
+
+A union of animation directions:
+- `forward`: navigating to the next page in the history or to a new page.
+- `back`: navigating to the previous page in the history.
+
+### `Fallback`
+
+<p>
+
+**Type:** `'none' | 'animate' | 'swap'`<br />
```
sarah11918, line 541: Earlier, we said `animate` was the default. Should we be specifying that default here?

(Or, should we have specified it as a default in the pre-reference there?)

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -280,34 +301,337 @@ Stores the element in focus on the current page and returns a function that when
...
+- `swap`: Astro will not attempt to animate the page. Instead, the old page will be immediately replaced by the new one.
+- `none`: Astro will not do any animated page transitions at all. Instead, you will get full page navigation in non-supporting browsers.
+
+<ReadMore>Learn more about [controlling the fallback strategy](/en/guides/view-transitions/#fallback-control) with the `ClientRouter`.</ReadMore>
+
+### `NavigationTypeString`
+
+<p>
+
+**Type:** `'push' | 'replace' | 'traverse'`<br />
```
sarah11918, line 556: similarly, does this have a default?

src/content/docs/en/reference/modules/astro-transitions.mdx
sarah11918: Agree!

src/content/docs/en/reference/modules/astro-transitions.mdx
```diff
@@ -212,18 +232,19 @@ Whether or not the current page has view transitions enabled for client-side nav
 
 <p>
 
-**Type:** `() => 'none' | 'animate' | 'swap'`<br />
+**Type:** <code>() => <a href="#fallback">Fallback</a></code><br />
```
ArmandPhilippot, line 237: Adds default:
```suggestion
**Type:** <code>() => <a href="#fallback">Fallback</a></code><br />
**Default:** `animate`<br />
```

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: I removed that because I thought both sentences could be merged:
> will be sent as a `POST` request with the `FormData` object

So, yes, this sounded a bit redundant to me. But, I get what you mean so I don't mind reverting those changes!

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: Yeah, I was hesitating. If we control that in the frontmatter instead, this means we hide both `navigate()` and `swapFunctions` subheadings. I think it can be useful to show them in the TOC? (because this is not too long) And so, only hiding the deprecated imports seemed better to me. But I don't have a strong opinion here.

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: I hesitated, but what you say makes sense. Maybe under `### Deprecated imports` this shouldn't be a caution but a regular sentence, and we use `caution` for each subheadings? I mean this is maybe too much to use a `:::caution` for both.

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: The nuance here is that it's a type. So, it doesn't have a default. This only defines the expected shape or, here, a set of allowed values. And user could import it to define the expected value for a prop: no default.

I used `Default`: for the `fallback` property on `ClientRouter` but, I was hesitant for `getFallback()`. I don't recall why it seemed better to add it in the text instead of the API block... even when I look at the implementation again. 😅  So, yes, we probably should add it there! I think that makes sense!

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: Same as above, because this is a type I don't think this is right to set a default here. And looking at [the property using it](https://deploy-preview-12734--astro-docs-2.netlify.app/en/reference/modules/astro-transitions/#navigationtype), I don't think a default applies here.

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: Ah, I shouldn't have done that while following T&D... this is the third `schedule` I got wrong. 😆

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: Reverted in [f6580ac](https://github.com/withastro/docs/pull/12734/commits/f6580ac3c91363f8d947aaf1df591a8324ca07af)

src/content/docs/en/reference/modules/astro-transitions.mdx
ArmandPhilippot: Updated as described in my previous comment in [1ad0374](https://github.com/withastro/docs/pull/12734/commits/1ad0374af4d829761ea62342d5be3ba05a65d247)

sarah11918: Yes, I think it looks great! Not too overwhelming!

PR #12735, opened 2025-11-18
[i18nIgnore] New deploy guide navigation component

sarah11918:
#### Description (required)

Updates our deployment guides navigation component to work like the other (newer) ones. In particular:

- replaces a hard-coded array that needs updating for every new guide with new deploy guide frontmatter properties (~~`service` name~~, `logo` to identify the logo icon used; `supports` array of deployment types)~~
- now sorts based on `id` and displays service based on `sidebar.label` (so, a sidebar label prop is now required, and can be fully displayed with localization e.g. "AWS avec SST" since this is now text provided by the translators in the MDX file itself)
- updates all existing deployment guides (including translations) to include these new properties
- updates the content collections config schema for deployment guides to include these new properties/requirements
- alphabetizes the entries, like our other navigation components
- renames some files (and adds redirects) for consistency of ordering in both the sidebar and the navigation grid itself (e.g. "AWS via SST" should show up alphabetically with "AWS" in both places)

TODOS:

- [x] redirect changed files

NICE TO HAVE FOLLOW UPS:
- [ ] CSS visual issue with badge labels

### DIFF
--- a/src/content/docs/zh-cn/guides/deploy/seenode.mdx
+++ b/src/content/docs/zh-cn/guides/deploy/seenode.mdx
@@ -4,6 +4,8 @@ description: 如何在 Seenode 上将你的 Astro 网站部署到网上。
 sidebar:
   label: Seenode
 type: deploy
+logo: seenode
+supports: ['ssr']
 i18nReady: true
 ---
 import { Steps } from '@astrojs/starlight/components';


src/components/DeployGuidesNav.astro
```diff
@@ -10,54 +12,22 @@ export interface Props {
...
+const links = enPages
+	.sort((a, b) => {
+		// Sort alphabetically.
+		return a.data.service.toLowerCase() > b.data.service.toLowerCase() ? 1 : -1;
+	})
+	.map((page) => {
+		const { service } = page.data;
+		const pageUrl = '/' + page.id.replace('en/', `${lang}/`) + '/';
+		const logo = isLogoKey(page.id.split('/').pop());
+		const { supports } = page.data;
```
sarah11918: Note: this `supports` is showing as a type error in my editor

HiDeoo: Didn't take a look at the PR yet but only commenting on this note, this may be a fluke or something in your editor as locally for me it's not the case:

image

sarah11918: A fluke in Firebase Studio?? :joy:

src/content/docs/en/guides/deploy/google-firebase.mdx
```diff
@@ -4,6 +4,8 @@ description: How to deploy your Astro site to the web using Google’s Firebase
 sidebar:
   label: Google Firebase
 type: deploy
+service: Google Firebase Hosting
```
HiDeoo:
```suggestion
service: Firebase
```

src/content/docs/de/guides/deploy/google-firebase.mdx
```diff
@@ -2,6 +2,10 @@
 title: Veröffentliche deine Astro-Website auf Google's Firebase Hosting
 description: Wie du deine Astro-Website auf Firebase-Hosting von Google im Internet veröffentlichst.
 type: deploy
+service: Firebase Hosting
```
HiDeoo:
```suggestion
service: Firebase
```

src/content/docs/fr/guides/deploy/google-firebase.mdx
```diff
@@ -4,6 +4,8 @@ description: Comment déployer votre site Astro sur le web en utilisant l'héber
 sidebar:
   label: Google Firebase
 type: deploy
+service: Google Firebase Hosting
```
HiDeoo:
```suggestion
service: Firebase
```

src/content/docs/ko/guides/deploy/google-firebase.mdx
```diff
@@ -4,6 +4,8 @@ description: Google의 Firebase Hosting을 사용하여 Astro 사이트를 웹
 sidebar:
   label: Google Firebase
 type: deploy
+service: Google Firebase Hosting
```
HiDeoo:
```suggestion
service: Firebase
```

src/content/docs/zh-cn/guides/deploy/google-firebase.mdx
```diff
@@ -4,6 +4,8 @@ description: 了解如何将你的 Astro 网页部署到 Google Firebase。
 sidebar:
   label: Google Firebase
 type: deploy
+service: Google Firebase Hosting
```
HiDeoo:
```suggestion
service: Firebase
```

src/content/docs/en/guides/deploy/cloudflare.mdx
```diff
@@ -4,6 +4,8 @@ description: How to deploy your Astro site to the web using Cloudflare
 sidebar:
   label: Cloudflare
 type: deploy
+service: Cloudflare Pages
```
HiDeoo:
```suggestion
service: Cloudflare
```

src/content/docs/fr/guides/deploy/cloudflare.mdx
```diff
@@ -4,6 +4,8 @@ description: Comment déployer votre site Astro sur le web en utilisant Cloudfla
 sidebar:
   label: Cloudflare
 type: deploy
+service: Cloudflare Pages
```
HiDeoo:
```suggestion
service: Cloudflare
```

src/content/docs/ko/guides/deploy/cloudflare.mdx
```diff
@@ -4,6 +4,8 @@ description: Cloudflare를 사용하여 Astro 사이트를 웹에 배포하는 
 sidebar:
   label: Cloudflare
 type: deploy
+service: Cloudflare Pages
```
HiDeoo:
```suggestion
service: Cloudflare
```

src/content/docs/zh-cn/guides/deploy/cloudflare.mdx
```diff
@@ -4,6 +4,8 @@ description: 如何使用 Cloudflare 将你的 Astro 网站部署到网络上
 sidebar:
   label: Cloudflare
 type: deploy
+service: Cloudflare Pages
```
HiDeoo:
```suggestion
service: Cloudflare
```

src/content/docs/de/guides/deploy/deno.mdx
```diff
@@ -4,6 +4,8 @@ description: Wie du deine Astro-Webseite mit Deno im Internet veröffentlichst.
 sidebar:
   label: Deno
```
sarah11918:
```suggestion
  label: Deno Deploy
```

src/content/docs/en/guides/deploy/deno.mdx
```diff
@@ -4,6 +4,8 @@ description: How to deploy your Astro site to the web using Deno.
 sidebar:
   label: Deno
```
ainurx:
```suggestion
  label: Deno Deploy
```

src/content/docs/fr/guides/deploy/deno.mdx
```diff
@@ -4,6 +4,8 @@ description: Comment déployer votre site Astro sur le web en utilisant Deno.
 sidebar:
   label: Deno
```
ainurx:
```suggestion
  label: Deno Deploy
```

src/content/docs/ko/guides/deploy/deno.mdx
```diff
@@ -4,6 +4,8 @@ description: Deno를 사용하여 Astro 사이트를 웹에 배포하는 방법.
 sidebar:
   label: Deno
```
ainurx:
```suggestion
  label: Deno Deploy
```

src/content/docs/zh-cn/guides/deploy/deno.mdx
```diff
@@ -4,6 +4,8 @@ description: 如何将你的 Astro 站点通过 Deno 部署上线。
 sidebar:
   label: Deno
```
ainurx:
```suggestion
  label: Deno Deploy
```

trueberryless:
Just noting down some visual bug, HiDeoo discovered in Talking and Doc'ing:

image

Potential fix could be to increase the minimal width of individual cards [here](https://github.com/withastro/docs/blob/a42ce7ed0baef8255df6df935ceed427d4f7b4f3/src/components/NavGrid/Grid.astro#L15). 🤔 I didn't check tho, where else this component is used. Maybe some additional CSS would be better here like `overflow-wrap: break-word` as [mentioned in this comment in a table scenario](https://github.com/withastro/starlight/issues/1321#issuecomment-3370480589)

src/data/logos.ts
```diff
@@ -26,6 +26,7 @@ export const logos = LogoCheck({
 	github: { file: 'github.svg', padding: '0.125em 0.125em 0.1375em' },
 	gitlab: { file: 'gitlab.svg', padding: '0' },
 	'google-cloud': { file: 'google-cloud.svg', padding: '.1875em' },
+	firebase: { file: 'firebase.svg', padding: '.1875em' },
 	'google-firebase': { file: 'firebase.svg', padding: '.1875em' },
```
sarah11918, line 30: I can't tell if this is Firebase Studio being Firebase Studio... I don't think we are relying on `google-firebase` anymore, but I get a type error when I remove this. It's possible this is being used by the Firebase backend guide? It's possible Firebase Studio can't effectively clear a cache and update?

In an ideal world, we don't need this, and I don't know if it's just my setup or if we're still using this somewhere.

public/_redirects
```diff
@@ -39,6 +39,9 @@
 /:lang/reference/experimental-flags/svg/				     /:lang/guides/images/
 /:lang/reference/experimental-flags/serialized-configuration/				 /:lang/reference/modules/astro-config/
 /:lang/reference/experimental-flags/responsive-images/				     /:lang/guides/images/
+/:lang/guides/deploy/sst/														/:lang/guides/deploy/aws-via-sst/
+/:lang/guides/deploy/flightcontrol/									/:lang/guides/deploy/aws-via-flightcontrol/
+
```
sarah11918:
```suggestion
/:lang/guides/deploy/google-firebase/						/:lang/guides/deploy/firebase/
```

src/data/logos.ts
ArmandPhilippot: It's not a Firebase issue. Well, it is but in our docs. 😄 
In `BackendGuidesNav.astro`, we use:
```ts
const logo = isLogoKey(page.id.split('/').pop());
```
And because the backend guide is still named `google-firebase.mdx`, we're checking if `google-firebase` is a valid logo key. When we remove `google-firebase` from `src/data/logo.ts` this is no longer the case.
So, either we also rename the backend file or we don't rename the logo key yet (and we would need to replace all the `logo: firebase` changes with `logo: google-firebase`).

HiDeoo: I _think_ we may have hit that exact issue during the call and decided against renaming the icon in this PR in the end? Or am I confusing this one with another one?

ArmandPhilippot: Yes, this is the same issue. 😅 If I recall what we said is that updating the other types of guide was maybe out of scope for this PR? But if the issue is just renaming the file and we're fine with that, this is a quick fix I think.

sarah11918: Yeah, I think this is a good "quick fix" for now -- we have two entries, one that satisfies each type of guide. When we reconfigure back-end guides to match this setup, we can remove the extraneous `google-firebase`

sarah11918: @delucis Can you weigh in on whether this "hack" seems OK for the scope of this PR?

sarah11918: UPDATE: I'm updating backend guides to work the same way, so if we merge this as-is, with two entries (`firebase` and `google-firebase`) then I should immediately have another PR to merge that will update all the backend guides and will delete the duplicate `google-firebase` entry.

sarah11918:
Also noting that we will need @yanthomasdev to help with a Lunaria statement here!

I *believe* that the only files that should show as "changed and needing updating" are the guides where `sidebar.label`:
- changed (e.g. "Deno" is now "Deno Deploy")
- contains extra text other than the service name (e.g. AWS **via** SST) because we'd now like these translated

yanthomasdev:
@sarah11918 I believe this is the code we want! It's only Flightcontrol, SST, Firebase, and Deno that changed, right?
```
@lunaria-track:src/content/docs/en/guides/deploy/aws-via-flightcontrol.mdx;src/content/docs/en/guides/deploy/aws-via-sst.mdx;
src/content/docs/en/guides/deploy/firebase.mdx;‎src/content/docs/de/guides/deploy/deno.mdx
```

ArmandPhilippot:
Yes, I think these are the only ones. But I wonder if the followings should be reported as updated:
* Firebase: this is the service name and, currently, I think every languages use `Google Firebase`
* Deno Deploy: same here, it seems to be the service name, so nothing to update? (I checked a few languages and it seems no one translate "Deno Deploy" inside the guide)

So, maybe the only ones worth tracking are Flightcontrol and SST because of "via"?

But if we prefer to make sure it's okay, yes, that sounds right to me!

sarah11918:
I realized that the `firebase.mdx` file changed names, so I *think* maybe best to leave that one in, just in case? 😅 But agreed, as no one seems to translate "Deploy" in Deno Deploy, I'll remove that one from the statement.

ArmandPhilippot:
Ah yes, good call! I hadn't thought about the filename! Sounds good to me!

PR #12740, opened 2025-11-20
feat: add doc for web components in MDX

Its-Just-Nans:
Related to https://github.com/withastro/roadmap/discussions/1260

#### Description (required)

Add doc for web components in MDX


sarah11918:
Hi! We've been discussing this, and this isn't really a happy path/standard Astro pattern that we'd recommend, so it's not quite right for our docs.

Something like this would make an interesting blog post as something you've discovered that people can learn from, or something to submit to [Astro Tips](https://astro-tips.dev)! But we won't be including it as reccomended usage in Astro Docs.

Its-Just-Nans:
Hi

Not sure to understand why it's not recommended but thanks for the explanation!

PR #12748, opened 2025-11-20
fix: remove binding assets

DevVali:


DevVali:
It might happen that this removal is needed in other parts of the sample config code as well, I've only used the static setup though so can't say for the other ones.

src/content/docs/en/guides/deploy/cloudflare.mdx
```diff
@@ -77,7 +77,6 @@ To get started, you will need:
           "name": "my-astro-app",
           "compatibility_date": "YYYY-MM-DD", // Update to the day you deploy
           "assets": {
-            "binding": "ASSETS",
             "directory": "./dist",
```
ArmandPhilippot, line 80: From what I see in Cloudflare docs, removing this line would be incorrect in some cases:
* [static websites](https://developers.cloudflare.com/workers/framework-guides/web-apps/astro/#if-you-have-a-static-site) don't need this
* [on-demand rendered websites](https://developers.cloudflare.com/workers/framework-guides/web-apps/astro/#if-your-site-uses-on-demand-rendering) need this

So, maybe we should use something like this instead:
```suggestion
            // Uncomment the following line when your site uses on-demand rendering:
            //"binding": "ASSETS",
            "directory": "./dist",
```

ArmandPhilippot (commented):
Thanks @DevVali for taking the time to fix our docs. It would have been helpful if you had uses our PR template and added a description to let us know why this is needed. Fortunately, Oliver opened a similar PR with a bit more context: https://github.com/withastro/docs/pull/12754

I left a suggestion because I don't think removing this line is the right solution here.

src/content/docs/en/guides/deploy/cloudflare.mdx
OliverSpeir: This is in the static tab of the static SSR switcher @ArmandPhilippot

sarah11918: I will note (not that I know anything about Cloudflare, just going through this document!):

The line above says that running `astro add cloudflare` (i.e. adding the adapter) will install this file for you. IF YOU ARE NOT USING THE ADAPTER (i.e. you have a static site) only THEN will you need to create this file. 

So, it seems to me that this file only needs to be correct for the case where you would manually add this for static sites? And we don't need to cater this file for on-demand rendered sites?

(The only exception would be if you install the adapter FOR a static site for some reason, and the file added by `astro add` needs updating? but I think we suggest not doing that?)

What we *could* clarify is in the "if you are not using the adapter" part, and maybe it could be more like:

> ... if you are not using the adapter because you are deploying a static site with no routes rendered on-demand, you'll need to create it yourself.

And then make this file correct only and exactly for static sites?

@OliverSpeir thoughts on this?

sarah11918: > This is in the static tab of the static SSR switcher

Ooh, thanks for pointing that out, Oliver. But now I'm confused. Can you run a Cloudflare on-demand site without the adapter? 😅 Why are we showing the SSR tab at all?

(If it's an "in case this didn't get added for you" then the text is all wrong introducing it!)

OliverSpeir: you need the Cloudflare adapter to deploy on demand on Cloudflare

I thought that we might be able to avoid using the tab switcher too at one point but, this deploy guide covers the full spectrum of deploying on Cloudflare:

1. Using adapter on workers 
2. Not using adapter on workers 
3. Using adapter on pages 
4. Not using adapter on pages 


There is a scenario where someone has a static site on Cloudflare deployed fine and then later they want to add a server island and need to see what to set the config up as, or the opposite they stop using they adapter and need to see what the config is

ArmandPhilippot: I totally missed the tabs. 🤦🏽 Thanks Oliver for noticing that and for the detailed explanation! So, even though it was confusing at first, I think this one change is enough.

PR #12754, opened 2025-11-21
Fix config for static workers section of cloudflare deploy guide

OliverSpeir:
Current config creates this error:

```sh
✘ [ERROR] Cannot use assets with a binding in an assets-only Worker.

  Please remove the asset binding from your configuration file, or provide
  a Worker script in your configuration file (`main`).
```


ArmandPhilippot:
Thanks, @OliverSpeir! It seems someone else has sent a PR to fix the same thing (https://github.com/withastro/docs/pull/12748) so I'll close this one. But, feel free to chime in the other PR. I left a suggestion there because I don't think this is the correct fix: it seems on-demand rendered websites still need that line according to Cloudflare docs.

PR #12759, opened 2025-11-21
[i18nIgnore] Update backend guides to use `logo` instead of `service`

sarah11918:
If wanted, we can update backend guides to accompany #12735

(This PR can change base to `main` to publish separately after 12735 is merged, or can merge into that one as this PR has both updated components)

Notes:

- There should be no translation updates needed (unlike in the Deploy Guide PR) here! Nothing re: `sidebar.label` changed in any of the guides here!

- This PR doesn't include support for `tags` (like the deploy guides) because we don't currently use a non-`minimal` version of this component. As a follow up, we could standardize and allow *all* components to have a minimal vs showing badges view, even if we don't currently display badges on any other kind of nav grid.


src/components/BackendGuidesNav.astro
```diff
@@ -1,26 +1,23 @@
...
 const links = enPages
-	.sort((a, b) => {
 		// Sort alphabetically.
-		return a.data.service.toLowerCase() > b.data.service.toLowerCase() ? 1 : -1;
-	})
+	.sort((a, b) => (a.id > b.id ? 1 : -1))
 	.map((page) => {
-		const { service } = page.data;
+		const { logo, sidebar } = page.data;
+		if (!sidebar.label) throw new Error('Deploy guides must always include a sidebar label.');
```
ArmandPhilippot:
```suggestion
		if (!sidebar.label) throw new Error('Backend guides must always include a sidebar label.');
```

Except this tiny nit, this looks great! 🙌🏽

public/_redirects
```diff
@@ -35,13 +35,15 @@
...
 /:lang/guides/rss/                                   /:lang/recipes/rss/
 /:lang/basics/astro-syntax/                          /:lang/reference/astro-syntax/
-/:lang/reference/experimental-flags/sessions/				 /:lang/guides/sessions/
-/:lang/reference/experimental-flags/svg/				     /:lang/guides/images/
-/:lang/reference/experimental-flags/serialized-configuration/				 /:lang/reference/modules/astro-config/
-/:lang/reference/experimental-flags/responsive-images/				     /:lang/guides/images/
-/:lang/guides/deploy/sst/														/:lang/guides/deploy/aws-via-sst/
-/:lang/guides/deploy/flightcontrol/									/:lang/guides/deploy/aws-via-flightcontrol/
-/:lang/guides/deploy/google-firebase/						/:lang/guides/deploy/firebase/
+/:lang/reference/experimental-flags/sessions/        /:lang/guides/sessions/
```
sarah11918, line 38: Noting that these earlier ones that appear changed is just a matter of using consistent spaces. This PR only introduces two new redirects (at the bottom)

src/content/docs/en/guides/backend/firebase.mdx
```diff
@@ -4,7 +4,7 @@ description: Add a backend to your project with Firebase
 sidebar:
   label: Firebase
 type: backend
-service: Firebase
+logo: firebase
```
sarah11918, line 7: Noting that the `sidebar.label` for these was already "Firebase", so no label changes (and therefore no translation considerations!) needed here.

PR #12760, opened 2025-11-21
Small build performance improvement

delucis:
#### Description (required)

This PR optimizes how we access some remote data during builds to speed them up.

**Before:** each `<Version>` and `<Since>` component call a `cachedFetch()` function to request package metadata from the npm registry. This resulted in 16 requests for fairly large JSON files like https://registry.npmjs.org/astro/latest.

**After:** a content collection loads the same metadata using a single `fetch()` call using the https://npm.antfu.dev service (see https://github.com/antfu/fast-npm-meta). This single fetch returns data ~4x smaller than just _one_ of the previous npm registry calls, so we use ~64x less bandwidth and avoid request overhead by only making one request instead of multiple.

While here, I also moved to use the same approach for the contributor data used in our `<FacePile>` component on the docs landing page. This meant I could remove our `cachedFetch()` utility and reduce our dependencies by 7 packages and 570 KB.

In some measurements on Netlify it looks like this might add up to a 15–20s saving in build times. Modest but still handy!


src/components/starlight/Hero/FacePile.astro
ArmandPhilippot: I guess we could now use `CollectionEntry<"astroContributors">["data"]` to avoid repeating the shape in two different places.

ArmandPhilippot (approved):
I checked different pages and couldn't spot any issues with our `<Since />` component. I also tested the custom error locally out of curiosity. Well done, and nice saving! 🙌🏽

src/components/starlight/Hero/FacePile.astro
delucis: Good idea. Did this and also a small refactor to avoid the extra `map()` in 16ef1de46aa569fa37c5f382a8583ad192b13c69.

src/content.config.ts
```diff
@@ -152,4 +152,42 @@ export const collections = {
...
+				'@astrojs/rss',
+				'@astrojs/sitemap',
+				'@astrojs/solid-js',
+				'@astrojs/svelte',
+				'@astrojs/vercel',
+				'@astrojs/vue',
+				'astro',
+			];
+			// See https://github.com/antfu/fast-npm-meta
+			const url = `https://npm.antfu.dev/${packages.join('+')}`;
```
HiDeoo, line 178: Did not find any restrictions regarding the usage of the "example server", I guess we will find if it's ever an issue 😅 

Question: could we end up in a weird state, e.g. right after a new release where we get cached outdated versions?  E.g. we merge a random PR, the endpoint is hit, versions get cached for 15mn, and right after Astro release a new version and we merge a docs PR for a minor?

delucis: > could we end up in a weird state, e.g. right after a new release where we get cached outdated versions? E.g. we merge a random PR, the endpoint is hit, versions get cached for 15mn, and right after we merge a docs PR for a minor?

I think it’s possible, yes, but not sure how much it matters. It looks like there is an undocumented `force` query param that can be used to bypass the cache although I’d be a bit hesitant to use it. Maybe there could be an argument for using it for production builds only or something? According to their docs, regular requests are cached for 15 minutes, forced requests only 30 seconds.

HiDeoo: Interesting, but yeah, I agree, seems a bit agressive for everything. Considering we're aware of this potential "issue", maybe we can just wait and see if it ever happens in the end, and if we ever get reports regarding it?

delucis: Yeah, hopefully in reality it’s not so common? I guess all of the following would need to be true:

- Production docs are merged <15 minutes after a new npm release
- Production docs are merged <15 minutes after a deploy that ran _before_ the new npm release

So in reality probably like merging ~8 minutes or less after a new npm release. Plus the other criteria is that this matters at all for a deploy which should really only be minor and major releases I guess? So it’s maybe <20x per year that even have this risk.

HiDeoo: Yeah, I was more asking the question regarding a small probability I imagined based on a quick read of their readme, rather than an important concern :thumbsup: Thanks for the clarification!

HiDeoo (approved):
Amazing refactor :clap: :clap:

Checked the preview and also tested locally, everything looks great to me!

PR #12778, opened 2025-11-23
i18n(de): refactor: remove inconsistencies (tutorial)

trueberryless:
#### Description

Go through all common inconsistencies in Discord (i18n-de channel) and apply them to all wrong occurances.
Also update with changes from #12763.

I basically did all 3 points from https://github.com/withastro/docs/pull/12502#issuecomment-3393453425

Please review @randomguy-2650 and maybe @louisescher if you both have the time.

After merging into `de/tutorial` branch, I think we may be able to merge, I didn't find any other changes to the English tutorial since 1. October with a little script 🤷

### DIFF
--- a/src/content/docs/de/tutorial/4-layouts/1.mdx
+++ b/src/content/docs/de/tutorial/4-layouts/1.mdx
@@ -105,7 +105,7 @@ Du hast noch einige Astro-Komponenten, die auf jeder Seite wiederholt gerendert 
 
 Das `<slot />`-Element ermöglicht es, **Kindelemente** einzufügen (oder „einzuschieben“), die zwischen den öffnenden und schließenden `<Component></Component>`-Tags in jede `Component.astro`-Datei geschrieben werden.
 
-## Übergib seiten-spezifische Werte als Props
+## Übergib seitenspezifische Werte als Props
 
 <Steps>
 6. Übergib den Seitentitel von `index.astro` an dein Layout über ein Komponenten-Attribut:


src/content/docs/de/tutorial/4-layouts/1.mdx
```diff
@@ -103,7 +103,7 @@ Du hast noch einige Astro-Komponenten, die auf jeder Seite wiederholt gerendert
     ```
 </Steps>
 
-Das `<slot />`-Element ermöglicht es, **Kind-Inhalte** einzufügen (oder „einzuschieben“), die zwischen den öffnenden und schließenden `<Component></Component>`-Tags in jede `Component.astro`-Datei geschrieben werden.
+Das `<slot />`-Element ermöglicht es, **Kindelemente** einzufügen (oder „einzuschieben“), die zwischen den öffnenden und schließenden `<Component></Component>`-Tags in jede `Component.astro`-Datei geschrieben werden.
 
 ## Übergib seiten-spezifische Werte als Props
```
randomguy-2650:
```suggestion
## Übergib seitenspezifische Werte als Props
```

trueberryless: excellent 👌

louisescher:
I think we should still attempt to get the tutorial merged. Trying to perfect this would take weeks if not months, and if there are major inconsistencies we will find them eventually! I think the state this is in is good enough.

trueberryless:
> I found another inconsistency, although there unfortunately might be more, as the tutorial consists of loads of pages.

I mean, I used search and replace with some RegEx to make it more manageable. We do not aim for perfection, remember if we spot something we can fix it later as well, doesn't have to be perfect from the beginning, well-done is enough...

> Side note: I used to expect it to be done by the end of October, but it clearly took way longer than I anticipated… maybe it still might not be done by December! 🫤

I think we could think about merging after this is merged, but if you want to take a closer look, feel free. From my side it's approved ✅

yanthomasdev:
As Sarah wrote in our Contributing Guide:
> Our first responsibility is to produce something that is accurate, clear, and helpful. Docs are never done, and constant improvement is the name of the game! This is where we can comfortably rely on #NWTWWHB (Not worse than what we had before) and strive to get better over time!

Let's get this full German tutorial merged @trueberryless @randomguy-2650 @louisescher! I believe there might be a few things to update from recent PRs that you can handle easily with one more PR, and then we should be good to go.

yanthomasdev (approved):

PR #12781, opened 2025-11-24
Remove unnecessary double dashes in npm commands

MarcusCaspeco:
#### Description (required)

`astro dev -- --mode dev` does not work, `astro dev --mode dev` does work. Not sure if this was changed recently and docs weren't updated, but this had me stuck for a while.

Not sure about --devOutput but since it's documented as a CLI command to astro I assume that the double dashes should be removed for that as well.

- Suggested label: Improve or update documentation 

No need for first time contributor, I am @MarcusOtter on work account 🕵️

### DIFF
--- a/src/content/docs/en/guides/environment-variables.mdx
+++ b/src/content/docs/en/guides/environment-variables.mdx
@@ -96,7 +96,7 @@ This allows you to run the dev server or build your site connecting to different
     npm run astro dev --mode staging
 
     # Build a site that connects to a "production" API with additional debug information
-    npm run astro build -- --devOutput
+    npm run astro build --devOutput
 
     # Build a site that connects to a "testing" API
     npm run astro build --mode testing


ArmandPhilippot:
Hi Marcus, I checked using the minimal template on Linux and this seems to work for me:
```
❭ npm run astro dev -- --mode dev

> astro-escape-cli@0.0.1 astro
> astro dev --mode dev

15:56:36 [types] Generated 0ms
15:56:36 [content] Syncing content
15:56:36 [content] Synced content

 astro  v5.16.0 ready in 51 ms

15:56:36 watching for file changes...
```

The escape characters before the options (e.g. `--`) was a requirement for PowerShell + npm users. But, IIRC, there have indeed been recent changes on PowerShell side... I think it was to make the escape characters unnecessary, not to make the command invalid when using them? 😅 

Since I don't have PowerShell, can you confirm the environment used for your test? I think on Linux/Mac the escapement was never a requirement (but it was working). So, now, if it fails with PowerShell, maybe we should apply your changes here (and probably on other pages?). Otherwise, I wonder why it's not working for you...

delucis:
IIRC it’s not only PowerShell in this case. PowerShell needed _double_ double dashes (i.e. `npm run astro dev -- -- --mode dev`), but I believe `--` at least always used to be required with `npm` on all platforms to indicate that the options are to be passed to the command being run and not directly to `npm` itself.

ArmandPhilippot:
Oh you're right, I mixed up things! I read that as if it were a double `--`. 🤦🏽  If I try without `--`, I get:
```
npm warn "dev" is being parsed as a normal command line argument.
npm warn Unknown cli config "--mode". This will stop working in the next major version of npm.
```

So, `--` seems to be required for npm...

MarcusCaspeco:
Ahh, right, this is my bad, I forgot this documentation is in the context of `npm run`! (as you can see in my PR description 😅)

I was adding it in my package.json like this, providing the arguments directly to the Astro CLI:
```json
{
    "scripts": {
        "dev": "astro dev --mode dev"
    }
}
```

(and then just `npm run dev`)

In which case `astro dev -- --mode dev` would not work, since I'm passing it to the CLI directly, not through npm.

I know you usually have to `-- --` when using `npm run ...`, like you mentioned Chris, I think I just forgot to read what these docs were telling me🤦 

Sorry! I'll blame monday morning brain

PR #12782, opened 2025-11-24
Tweak Learn Astro ad for Black Friday

delucis:
#### Description (required)

Updates the copy on the Learn Astro ad to promote the extra reduced discount of 33%

A banner ad from the Astro docs. The call-to-action button now reads “Black Friday! Get 33% off”.

### DIFF
--- a/src/components/NavGrid/CardsNav.astro
+++ b/src/components/NavGrid/CardsNav.astro
@@ -59,5 +59,6 @@ const currentPage = new URL(Astro.request.url).pathname;
 
 	.tags {
 		gap: 0.5rem;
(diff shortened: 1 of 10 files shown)


trueberryless:
Sorry to ask the stupid question, but where is this ad visible?

I can only see this one in the sidebar:

image

---

EDIT: Nvm, I visited the one page where the other ad is visible: https://deploy-preview-12782--astro-docs-2.netlify.app/en/install-and-setup/ But how does it end up there? The only place where `RandomizedAd.astro` is used is in the footer, no? 🤔

PR #12604, opened 2025-10-22
[v6] New content collections guide includes live collections

sarah11918:
#### Description (required)

This PR updates the Content Collections guide so that it describes and differentiates the two types of collections we will have stable in v6: build-time and live. It also updates entries in the `astro:content` module reference page and the Content Loader API page:

- [Guide](https://deploy-preview-12604--astro-docs-2.netlify.app/en/guides/content-collections/) - focuses on what collections are; how to configure/define; how to use data in a collection (querying, rendering, generating routes)
- [`astro:content` module reference](https://deploy-preview-12604--astro-docs-2.netlify.app/en/reference/modules/astro-content/) - API reference for things imported from `astro:content`
- [Content Loader API reference](https://deploy-preview-12604--astro-docs-2.netlify.app/en/reference/content-loader-reference/) - API reference for Content Loader (including Build-time collections and Live Collections) including explaining the basic format/functions of loaders and how to build them.

This incorporates content from the current existing [Experimental Live Collections reference page](https://docs.astro.build/en/reference/experimental-flags/live-content-collections/) into the new main content collections guide page.

TODO:

- [x] `LiveData*` types are still mostly placeholders
- [x] check for needed updates to Content collections guide now that we don't talk about these "types" of loaders anymore (inline, object, live)

My thoughts throughout this PR:

- It is important to name and distinguish the two types of collections (build-time and live)

- the two built-in build time loaders (`glob()` and `file()`) now have their own, clear, discoverable sections.

- The natural order of "setting up a collection" is still: define it, add a loader, most likely add a schema; then you're able to query/filter and render; oh, and you probably are going to want dynamic pages based on your entries. So, I've kept that overall structure. Build time and live collections are split out into discrete sections, each following that pattern, based on feedback.

- Did I say I added a LOT of internal page linking?? One big challenge here is that setting up a simple blog with local Markdown files is much less complicated than needing to build your own live loader. Frankly, there's a crap ton of loader content, but you can't do *anything* until you have a loader so it feels difficult to offload it to the bottom of the page and just say "go off here to learn about loaders so I can show you the rest of content collections". (see next point)

[...]

PR #12790, opened 2025-11-25
fix(live-content-collections): update `cacheHint` code snippet

ArmandPhilippot:
#### Description (required)

Fixes how we define `cacheHint` in a code snippet of `experimental-flags/live-content-collections`. I reused the version we have for the v6 docs, so this also adds imports.

### DIFF
--- a/src/content/docs/de/tutorial/0-introduction/1.mdx
+++ b/src/content/docs/de/tutorial/0-introduction/1.mdx
@@ -1,12 +1,12 @@
 ---
 type: tutorial
 title: Über dieses Tutorial
-description: Wie du mit dem Tutorial "Baue deinen ersten Astro-Blog" anfangen kannst.
+description: Wie du mit dem Tutorial „Erstelle deinen ersten Astro-Blog“ anfangen kannst.
 i18nReady: true
 ---
+
 import Checklist from '~/components/Checklist.astro';
 import Box from '~/components/tutorial/Box.astro';
-
 
 ## Was muss ich wissen, um loszulegen?
 
@@ -16,7 +16,7 @@ Du kannst das gesamte Tutorial abschließen, indem du den Anweisungen folgst. As
 Außerdem brauchst du einen [GitHub](https://github.com)-Account (oder etwas Vergleichbares), um dein Projekt im Internet zu veröffentlichen. 
 
 <details>
-<summary>Wie kann ich die Checklisten am Ende jedes Kapitels nutzen?</summary>
+<summary>Wie kann ich die Checklisten am Ende jeder Lektion nutzen?</summary>
 
 Du hakst sie einfach ab! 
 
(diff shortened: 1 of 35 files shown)


src/content/docs/en/reference/experimental-flags/live-content-collections.mdx
```diff
@@ -556,13 +561,13 @@ export function myLoader(config): LiveLoader<MyData> {
           // All fields are optional, and are combined with each entry's cache hints
           // tags are merged from all entries
           // lastModified is the most recent lastModified of all entries and the collection
-          lastModified: new Date(item.lastModified),
+          lastModified: new Date(products.sort(byLastModified)[0].lastModified),
```
delucis: This is a bit of a nit but do we show `byLastModified()` as an implementation anywhere? If not, I wonder if it’s worth simplifying this to show less detail and express more of the meaning, i.e. something like

```suggestion
          lastModified: new Date(getMostRecentlyUpdated(products).lastModified),
```

delucis: Or even just this depending on what we think is most likely:
```suggestion
          lastModified: getLastModifiedDate(products),
```

ArmandPhilippot: Ah yes, sorry, I talked about this on a Discord thread and forgot to add more context here!

We received a feedback regarding this code snippet: `item` is not available in `cacheHint` and, so, it was unclear to them what should be used here.
But, actually, we already fix that snippet in the v6 docs (https://github.com/withastro/docs/pull/12604 / https://deploy-preview-12604--astro-docs-2.netlify.app/en/guides/content-collections/#cache-hints), so I was applying the same code to our live docs.

No, we don't show the implementation for `byLastModified`. So, yeah, your last suggestion is probably enough here!

src/content/docs/en/reference/experimental-flags/live-content-collections.mdx
```diff
@@ -538,13 +538,18 @@ console.log(entry?.data.displayPrice); // e.g., "$29.99"
 Live loaders can provide cache hints to help with response caching. You can use this data to send HTTP cache headers or otherwise inform your caching strategy.
 
 ```ts title="my-loader.ts"
+import type { LiveLoader } from "astro/loaders";
+import { byLastModified } from "./sort";
+import { loadStoreProduct, loadStoreProducts } from "./store";
```
ArmandPhilippot: If we go with a single function, I guess we can use a single fake file to import it:
```suggestion
import { loadStoreProduct, loadStoreProducts, getLastModifiedDate } from "./store";
```

delucis: Sounds good to me!

ArmandPhilippot:
There are three lines to update in the v6 PR: the imports and `getLastModifiedDate` instead of the `.sort()` method currently. But I made a suggestion in that PR since I know exactly what changed. 😄

PR #12791, opened 2025-11-25
Document URL sanitization for `navigate()` API

delucis:
#### Description (required)

This PR documents sanitizing user input if passing it to the `navigate()` API exported from the `astro:transitions` module. This is the same advice as required when using user input with the web standard [`navigation.navigate()`](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate) method.

Shout out to @zomaxsec for highlighting this!

### DIFF
--- a/src/content/docs/de/tutorial/0-introduction/1.mdx
+++ b/src/content/docs/de/tutorial/0-introduction/1.mdx
@@ -1,12 +1,12 @@
 ---
 type: tutorial
 title: Über dieses Tutorial
-description: Wie du mit dem Tutorial "Baue deinen ersten Astro-Blog" anfangen kannst.
+description: Wie du mit dem Tutorial „Erstelle deinen ersten Astro-Blog“ anfangen kannst.
 i18nReady: true
 ---
+
 import Checklist from '~/components/Checklist.astro';
 import Box from '~/components/tutorial/Box.astro';
-
 
 ## Was muss ich wissen, um loszulegen?
 
@@ -16,7 +16,7 @@ Du kannst das gesamte Tutorial abschließen, indem du den Anweisungen folgst. As
 Außerdem brauchst du einen [GitHub](https://github.com)-Account (oder etwas Vergleichbares), um dein Projekt im Internet zu veröffentlichen. 
 
 <details>
-<summary>Wie kann ich die Checklisten am Ende jedes Kapitels nutzen?</summary>
+<summary>Wie kann ich die Checklisten am Ende jeder Lektion nutzen?</summary>
 
 Du hakst sie einfach ab! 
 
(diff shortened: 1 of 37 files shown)


src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -446,6 +447,32 @@ You can opt out of router transitions on any individual form using the `data-ast
 <form action="/contact" data-astro-reload>
 ```
 
+### Navigating with user input
+
+The `navigate()` API does not perform sanitization on the URLs passed to it. If you are using user input to determine the URL to navigate to, you should validate the input before passing it to `navigate()`.
+
+For example, if implementing a `?redirect` query parameter, it would be bad if `?redirect=http://malicious-site.com` or even `?redirect=javascript:alert('Evil code')` could be used to navigate away from your site or execute arbitrary code.
```
```suggestion
For example, a `?redirect` query parameter could be used to navigate away from your site (`?redirect=http://malicious-site.com`) or to execute arbitrary code (`?redirect=javascript:alert('Evil code')`).
```
Maybe something like this? Could also tack on an "if..." at the end like:

> ... if any user-submitted input were accepted/used...

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -446,6 +447,32 @@ You can opt out of router transitions on any individual form using the `data-ast
...
+
+To implement this safely, you could ensure only a set of known paths can be redirected to:
```
```suggestion
One way to implement this safely is to ensure only a set of known paths can be redirected to:
```
Maybe feels a little more comforting to have a stronger "one way you can do this" vs "you *could*..." (I think that's what is intended here, that here's a potential solution to this, but there could be others.)

src/content/docs/en/guides/view-transitions.mdx
```diff
@@ -446,6 +447,32 @@ You can opt out of router transitions on any individual form using the `data-ast
...
+
+  if (allowedPaths.includes(redirect)) {
+    navigate(redirect);
+  }
+</script>
+```
+
+The exact kind of sanitization you need will depend on your site and what you want to allow.
+
+<ReadMore>Consider enabling Astro’s experimental [Content Security Policy feature](/en/reference/experimental-flags/csp/) to help protect against cross-site scripting (XSS) risks if using user input with the `navigate()` API.</ReadMore>
```
```suggestion
<ReadMore>Consider enabling Astro’s [experimental Content Security Policy feature](/en/reference/experimental-flags/csp/) to help protect against cross-site scripting (XSS) risks if using user input with the `navigate()` API.</ReadMore>
```
nit: *slight* preference for including experimental here, I think?

src/content/docs/en/guides/view-transitions.mdx
delucis: Yes, I think that’s nice — thanks Sarah!

Maybe an “if” clause on the end like this would make sense?

```suggestion
For example, a `?redirect` query parameter could be used to navigate away from your site (`?redirect=http://example.com`) or to execute arbitrary code (`?redirect=javascript:alert('Evil code')`) if the value is not sanitized before use.
```

sarah11918: Yes, looks good!

PR #12792, opened 2025-11-25
fix(astro:middleware): add docs for `astro/middleware` imports

ArmandPhilippot:
#### Description (required)

Updates the `astro:middleware` docs to:
* Add two sections for imports available from `astro/middleware`
  * `userDefinedLocales` added in https://github.com/withastro/astro/pull/8974
  * `defaultLocale` added in https://github.com/withastro/astro/pull/12151
  * `locals` added in https://github.com/withastro/astro/pull/12647 but there is no changeset ... the last v4 version was released in November, so I think it's safe to say v5 here.
* Move `createContext` and `trySerializeLocals` from `astro:middleware` to `astro/middleware` (not available in the colon version)
* Add `defineMiddleware()` type and update the description
* Update `sequence()` code snippet to make it valid and replace `_` with `context` to improve clarity.

**Things to consider:** (already answered by Sarah below and on Discord)
* Is it okay to documented `astro/middleware` in the same place given the page is named `astro:middleware`? If so, maybe we should add a sentence to explain when to use one over the other.
* We already document `defineMiddleware()` and `sequence()` in `astro:middleware` imports but the same functions are available from `astro/middleware`: should they be available from only one version? how to document both without repeating the same thing.

See also https://github.com/withastro/astro/pull/8101#pullrequestreview-1582088352 for extra context.

- Suggested label: `add new content`, `improve or update documentation`


sarah11918:
After some discussion:

> `astro/middleware` in the same place given the page is named astro:middleware?

Yes, let's do this every time there's a regular module associated with the virtual model. Maybe we can have a sentence like "The following are imported from the virtual content module" / "The following are imported from the regular content module" to introduce each code block of imports.

We checked, and there should be a corresponding virtual model for every one of these modules that isn't already covered by its own special API page (e.g. `astro/loaders` is covered in the Content Loader API page), so keeping these pages named for the virtual modules seems fine.

> should they be available from only one version? how to document both without repeating the same thing.

"Should" is something to take up with the framework team! If we're responsible for documenting what is currently, and if both entries would be *identical*, then it's OK I think to:

- show them in both import code blocks
- put the reference entry in the module that we expect more people are using (I suspect the virtual one?)
- make a section heading in the other module, too, but its only text is to link back to the one written in the other module, maybe like:

	```
	See [`defineMiddleware()`](#definemiddleware) from `astro:middleware`
	```

src/content/docs/en/reference/modules/astro-middleware.mdx
```diff
@@ -16,26 +16,30 @@ Middleware allows you to intercept requests and responses and inject behaviors d
...
 
 ### `defineMiddleware()`
 
-You can import and use the utility function `defineMiddleware()` to take advantage of type safety:
+<p>
+
+**Type:** `(fn: MiddlewareHandler) => MiddlewareHandler`
+</p>
+
+A function for defining a middleware function with type safety. When you use this utility, the [`context`](#context) and [`next()`](#next) arguments are automatically typed, and you get a Typescript error when you try to return a [value not supported](#onrequest) in your middleware.
```
sarah11918:
```suggestion
A function for defining a middleware function with type safety. When you use this utility, the [`context`](#context) and [`next()`](#next) arguments are automatically typed, and you will get a Typescript error if you try to return a [value not supported](#onrequest) in your middleware.
```

src/content/docs/en/reference/modules/astro-middleware.mdx
```diff
@@ -16,26 +16,30 @@ Middleware allows you to intercept requests and responses and inject behaviors d
...
 
-```ts
-// src/middleware.ts
+```ts title="src/middleware.ts"
 import { defineMiddleware } from "astro:middleware";
 
-// `context` and `next` are automatically typed
+// `context` is typed with `APIContext` and `next` is typed with `MiddlewareNext`
```
sarah11918: We might not need this comment at all, since it has been provided in text!

(Making a code comment *longer* generally seems less helpful, and more like a proper explanation is needed).

We don't actually say that `next` is typed with `MiddlewareNext` in the `next` entry itself, do we? Is it important to say that? I would think that info would be better suited there, and all someone needs to know at this point is that `next` and `context` are automatically typed? (Which you've said in the paragraph above)

ArmandPhilippot: Because "automatically typed" was now in the description, I was looking for a replacement... but you're right: no comment at all is probably better! I'll push a commit to remove it.

> We don't actually say that next is typed with MiddlewareNext in the next entry itself, do we? Is it important to say that?

Ah, good call! Indeed we use:
> (rewritePayload?: string | URL | Request) => Promise<Response>

I checked and this is actually what is assigned to the `MiddlewareNext` type. I don't think this is important to update that. But, we could  if we document this type and `MiddlewareHandler`:
> import type { MiddlewareHandler, MiddlewareNext } from "astro";

They come from `astro` and not `astro/middleware` so I left them undocumented before. But now, we have a precedent in the v6 docs for imports from `astro`, so we could add them here.

They are not really useful because we have `defineMiddleware()`, but maybe someone don't want to use this helper for some reasons and so, it's probably better to document them. Then, we could update the type in `next()` and link to the `MiddlewareNext` section.

I'll see what I can do here!

ArmandPhilippot: Done! I also moved the `context` and `next()` sections because the flow might be better if they are under the new `MiddlewareHandler` type. This way users don't need to go back and forth when reading the documentation.

src/content/docs/en/reference/modules/astro-middleware.mdx
```diff
@@ -16,60 +16,84 @@ Middleware allows you to intercept requests and responses and inject behaviors d
...
 
 ### `defineMiddleware()`
 
-You can import and use the utility function `defineMiddleware()` to take advantage of type safety:
+<p>
+
+**Type:** <code>(fn: <a href="#middlewarehandler">MiddlewareHandler</a>) => MiddlewareHandler</code>
+</p>
+
+A function for defining a middleware function with type safety. When you use this utility, the [`context`](#context) and [`next()`](#next) arguments are automatically typed, and you will get a Typescript error if you try to return a [value not supported](#middlewarehandler) in your middleware.
```
sarah11918:
```suggestion
A function for defining a middleware function with type safety. When you use this utility, the [`context`](#context) and [`next()`](#next) arguments are automatically typed, and you will get a Typescript error if you try to return a [value not supported in your middleware](#middlewarehandler).
```
I was trying to think of the best link text here. I don't think "value not supported" on its own is the *best* if it points to the middlewarehandler. It could be all of this, maybe? Thoughts?

src/content/docs/en/reference/modules/astro-middleware.mdx
```diff
@@ -80,45 +104,174 @@ This function can be used by integrations/adapters to programmatically execute t
...
+### `MiddlewareHandler`
+
+<p>
+
+**Type:** <code>(context: <a href="/en/reference/api-reference/">APIContext</a>, next: <a href="#middlewarenext">MiddlewareNext</a>) => Promise\<Response\> | Response | Promise\<void\> | void</code>
+</p>
+
+Represents an Astro middleware function. Middleware handlers receive two arguments and can either return a `Response` directly or call `next()` to invoke the next middleware in the chain. Alternatively, you can use [`defineMiddleware()`](#definemiddleware) to get type safety for your middleware.
+
+The following example imports the `MiddlewareHandler` type to get type safety in the `onRequest()` function:
```
sarah11918:
```suggestion
The following example imports the `MiddlewareHandler` type to get type safety in the [`onRequest()`](#onrequest) function:
```
Just checking whether there's a reason we're not linking here? Otherwise, I'd think we would?

src/content/docs/en/reference/modules/astro-middleware.mdx
ArmandPhilippot: Yeah, I wasn't satisfied with that link anchor but I wasn't sure how to improve that. So, I'm fine with your suggestion!

src/content/docs/en/reference/modules/astro-middleware.mdx
ArmandPhilippot: No reason, just an oversight 😅

PR #12802, opened 2025-11-27
docs: document breaking change requiring SSR environment for Astro component tests

Ntale3:
#### Description (required)
This PR documents a breaking change in Astro 6 where rendering `.astro` components in Vitest tests now requires using an SSR environment (`node`) instead of client environments (`jsdom`/`happy-dom`).

The temporary workaround added in Astro 5.x has been removed to ensure tests run in an environment that matches Astro's SSR build process.

**What's included:**
- Clear identification of who is affected (Vitest + Container API + client environment users)
- Explanation of what will happen without migration
- Step-by-step migration guide with code examples
- Workspace configuration example for projects needing multiple test environments

- Closes withastro/astro#14895


ematipico:
Something is wrong with the PR @Ntale3 

I believe the problem is that you created your branch from the `main` branch and then opened the PR towards the `v6` branch. Could you fix that?

delucis:
@ematipico Should be fixed: `v6` branch needed updating with latest commits from `main` and then I pulled those changes in here. So now this PR only shows the relevant changes 👍

Ntale3:
Thanks for the guidance @delucis let me work as guided

Issue #12789, opened 2025-11-25
Adapter API links to a deprecated type


sarah11918:
### 📚 Subject area/topic

Adapter API ( `routeData`)

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/reference/adapter-reference/#routedata

### 📋 Description of content that is out-of-date or incorrect

`routeData` in the Adapter API reference links to a type that is deprecated, and provides another link as a replacement. The original `routeData` entry should be updated to go directly to what's relevant.

<img width="884" height="648" alt="Image" src="https://github.com/user-attachments/assets/262ec6e4-58d8-4fb4-949c-5165682cb8b5" />

<img width="885" height="648" alt="Image" src="https://github.com/user-attachments/assets/109c920d-6d2c-407c-b750-6dc610c9be6d" />

<img width="884" height="649" alt="Image" src="https://github.com/user-attachments/assets/9b687c3a-6884-4aed-91c0-fd9cb51dc65c" />

@florian-lefebvre suggested this might mean documenting the internal `RouteData` type at https://github.com/withastro/astro/blob/abed9294ce698980f22f8a3f6695670151915d9b/packages/astro/src/types/public/internal.ts#L43

anaxite:
Had a look in Talking & Doc'ing!

ArmandPhilippot:
Hi @JusticeMatthew & @anaxite , you probably wanted to leave a comment on the PR not the issue https://github.com/withastro/docs/pull/12946 😉

JusticeMatthew:
I swapped it

Was distracted and just tapped the first link I saw on mobile 😅

PR #12809, opened 2025-11-28
docs: document Environment API breaking changes

matthewp:
#### Description (required)

Updates the v6 guide with breaking changes from the Environment API refactor

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -700,6 +700,49 @@ export const server = {
 
 <ReadMore>Learn more about [rewrites](/en/guides/routing/#rewrites).</ReadMore>
 
+### Removed: schema function signature (Content Loader API)
+
+In Astro 5.x, a content loader could choose to define a schema as a function instead of defining a Zod schema object for validation. This is useful to dynamically generate the schema based on the configuration options or by introspecting an API.
+
+Astro 6.0 removes this signature and introduces a new `createSchema()` property as a replacement for those who still want to dynamically define a schema in their content loader. 
+
+Providing a schema function in the old way will log a warning message that the loader's schema is being ignored, but otherwise the loader will continue to work as if no schema had been provided. In a future major version, loaders that provide a schema function will throw an error and cannot be used.
+
+#### What should I do?
+
+If you are building a content loader and using a function to dynamically return a collection `schema` property, you must remove your existing function and use the new `createSchema()` property to define your schema instead.
+
+For example, you can reproduce Astro's previous behavior by using `zod-to-ts` directly with `createSchema()` and any previous function logic:
+
+```ts del={11} ins={2,12-22}
+import type { Loader } from 'astro/loaders'
+import { createTypeAlias, zodToTs } from 'zod-to-ts'
+import { getSchemaFromApi } from './utils'
+
+function myLoader() {
+  return {
+    name: 'my-loader',
+    load: async (context) => {
+      // ...
+    },
+    schema: async () => await getSchemaFromApi(),
+    createSchema: async () => {
+      const schema = await getSchemaFromApi()
+      const identifier = 'Entry'
+      const { node } = zodToTs(schema, identifier)
+      const typeAlias = createTypeAlias(node, identifier)
+
+      return {
+        schema,
+        types: `export ${typeAlias}`
+      }
+    }
+  } satisfies Loader
+}
+```
+
+<ReadMore>Learn more about [`createSchema()`](/en/reference/content-loader-reference/#createschema) in the Content Loader API reference.</ReadMore>
+
 ### Experimental Flags
 
 Experimental flags allow you to opt in to features while they are in early development. Astro may also use experimental flags to test breaking changes to default behavior. The following experimental flags have been removed in Astro 6.0 and are now stable, or the new default behavior.
@@ -1147,6 +1190,35 @@ const driver = await manifest.sessionDriver?.();
 ```
 
 <ReadMore>Learn more about [the Adapter API](/en/reference/adapter-reference/).</ReadMore>
+### Changed: schema types are inferred instead of generated (Content Loader API)
+
+<SourcePR number="14759" title="feat: loader.createSchema()" />
+
+In Astro 5.x, the types for content collections were generated using `zod-to-ts` when provided by a content loader and not defined by a user-provided schema.
+
+Astro 6.0 removes this behavior: types are no longer generated using `zod-to-ts`. Instead, types are inferred.
+
+#### What should I do?
+
+If you are providing a `schema` in a content loader, you must use the [TypeScript' `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator):
+
+```ts del={3,11} ins={4,12}
+import type { Loader } from 'astro/loaders'
+
+function myLoader(): Loader {
+function myLoader() {
+  return {
+    name: 'my-loader',
+    load: async (context) => {
+      // ...
+    },
+    schema: z.object({/* ... */})
+  }
+  } satisfies Loader
+}
+```
+
+<ReadMore>Learn more about [defining loader schema types](/en/reference/content-loader-reference/#the-loader-object).</ReadMore>
 
 ## Community Resources
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -21,6 +21,3 @@
 
 <!-- If you are a member of the Astro Discord, please add your username in the description so we can welcome you there! -->
 <!-- https://astro.build/chat -->
-
-<!-- TAKING PART IN HACKTOBERFEST? LET US KNOW! -->
-<!-- See https://contribute.docs.astro.build/guides/hacktoberfest/ for more details. -->
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -8,7 +8,7 @@ runs:
       uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
 
     - name: Setup Node
-      uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
+      uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
       with:
         node-version: 24.11.0
         cache: pnpm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Install Tools & Dependencies
         uses: ./.github/actions/install
@@ -36,7 +36,7 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Install Tools & Dependencies
         uses: ./.github/actions/install
@@ -49,7 +49,7 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Install Tools & Dependencies
         uses: ./.github/actions/install
@@ -62,7 +62,7 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Install Tools & Dependencies
         uses: ./.github/actions/install
--- a/.github/workflows/contributors.yml
+++ b/.github/workflows/contributors.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Install Tools & Dependencies
         uses: ./.github/actions/install
--- a/.github/workflows/discord-i18n-ping.yml
+++ b/.github/workflows/discord-i18n-ping.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           fetch-depth: 0
 
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -12,7 +12,7 @@ jobs:
       NODE_OPTIONS: "--max_old_space_size=4096"
     steps:
       - name: Check out code using Git
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           ref: ${{ github.head_ref }}
           # Needs access to push to main
--- a/.github/workflows/lint-ja-docs.yml
+++ b/.github/workflows/lint-ja-docs.yml
@@ -23,7 +23,7 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Install Tools & Dependencies
         uses: ./.github/actions/install
--- a/.github/workflows/lunaria.yml
+++ b/.github/workflows/lunaria.yml
@@ -22,7 +22,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           # Necessary for Lunaria to work properly
           # Makes the action clone the entire git history
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -21,7 +21,7 @@ jobs:
             PR_TITLE: 'ci: update configuration reference docs'
     steps:
       - name: Check out code using Git
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           ref: ${{ matrix.TARGET_BRANCH }}
 
@@ -62,7 +62,7 @@ jobs:
             PR_TITLE: 'ci: update error reference docs'
     steps:
       - name: Check out code using Git
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           ref: ${{ matrix.TARGET_BRANCH }}
 
--- a/.github/workflows/pr-coauthor.yml
+++ b/.github/workflows/pr-coauthor.yml
@@ -13,5 +13,5 @@ jobs:
     if: ${{ github.event.issue.pull_request }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - uses: kevinzunigacuellar/coauthor-action@292b32bdbffc663431a2a37b3bfa597d83a74905 # v0.1.3
--- a/.github/workflows/welcome-bot.yml
+++ b/.github/workflows/welcome-bot.yml
@@ -13,7 +13,7 @@ jobs:
     name: Welcome First Time Contributors
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - uses: zephyrproject-rtos/action-first-interaction@58853996b1ac504b8e0f6964301f369d2bb22e5c
         with:
           repo-token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,8 +44,6 @@ Submitting an Issue is usually the first step to making a change. After an Issue
 
 Larger contributions to the docs are encouraged after participating in Issues and Discussions, as unsolicited material may not fit into our existing plans.
 
-**Participating in Hacktoberfest?** See how you can get recognized for your translation PRs and PR reviews in our [Hacktoberfest guide](https://contribute.docs.astro.build/guides/hacktoberfest/).
-
 ### Examples of Helpful GitHub New Issues
 
 - a particular explanation is confusing (with explanation)
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -90,6 +90,7 @@ export const sidebar = [
 					group('guides.upgrade.major', {
 						collapsed: true,
 						items: [
+							'guides/upgrade-to/v6',
 							'guides/upgrade-to/v5',
 							'guides/upgrade-to/v4',
 							'guides/upgrade-to/v3',
--- a/scripts/lib/linkcheck/base/page.ts
+++ b/scripts/lib/linkcheck/base/page.ts
@@ -18,10 +18,14 @@ export class HtmlPage {
 	 * Example: `/en/getting-started/`
 	 */
 	readonly pathname: string;
-
-	readonly dom: Document;
-
-	readonly anchors: Element[];
+	/**
+	 * A list of all anchor elements on the page.
+	 */
+	readonly anchors: Array<{
+		label: string;
+		name: string;
+		href: string;
+	}>;
 	/**
 	 * A list of unique link hrefs on the page.
 	 */
@@ -46,12 +50,11 @@ export class HtmlPage {
 	 */
 	readonly isRedirect: boolean;
 	/**
-	 * The element containing the page's main content.
+	 * Indicates whether the page contains content.
 	 *
-	 * Prefers the first `<article>` element, with a fallback to `<body>` if no article was found,
-	 * and finally `null` if the page even doesn't have a body.
+	 * Set to `false` if the page even doesn't have a `<body>` element (a partial or invalid page).
 	 */
-	readonly mainContent: Element | null;
+	readonly hasContent: boolean;
 	/**
 	 * The language of the page's main content.
 	 *
@@ -71,25 +74,34 @@ export class HtmlPage {
 
 	constructor({ html, href, pathname }: { html: string; href: string; pathname: string }) {
 		// Attempt to read the HTML file and parse its DOM
-		this.dom = parseDocument(html);
+		const parser = new DocumentParser(parseDocument(html));
 		this.href = href;
 		this.pathname = pathname;
 
 		// Provide commonly used data as properties
-		this.anchors = DomUtils.getElementsByTagName('a', this.dom, true);
+		this.anchors = DomUtils.getElementsByTagName('a', parser.dom, true).map((el) => ({
+			// Pass the strings through Buffer to allow Node to reallocate them into independent memory
+			// instead of using slices of the original large string containing the full HTML document.
+			//
+			// This reduces memory usage significantly, at time of writing, 2.1Gib -> 300MiB.
+			label: Buffer.from(DomUtils.innerText(el)).toString(),
+			name: el.attribs.name && Buffer.from(el.attribs.name).toString(),
+			href: el.attribs.href && Buffer.from(el.attribs.href).toString(),
+		}));
 
 		// Build a list of unique link hrefs on the page
-		this.uniqueLinkHrefs = [...new Set(this.anchors.map((el) => decodeURI(el.attribs.href)))];
+		this.uniqueLinkHrefs = [...new Set(this.anchors.map((el) => decodeURI(el.href)))];
 
 		// Build a list of hashes that can be used as URL fragments to jump to parts of the page
-		const anchorNames = this.anchors
-			.map((el) => el.attribs.name)
-			.filter((name) => name !== undefined);
-		const ids = this.findAll((el) => Boolean(el.attribs.id)).map((el) => el.attribs.id);
+		const anchorNames = this.anchors.map((el) => el.name).filter((name) => name !== undefined);
+		const ids = parser
+			.findAll((el) => Boolean(el.attribs.id))
+			// Same reason as above.
+			.map((el) => Buffer.from(el.attribs.id).toString());
 		this.hashes = [...anchorNames, ...ids].map((name) => `#${name}`);
 
 		// Check if the page redirects somewhere else using meta refresh
-		const metaRefreshElement = this.findFirst(
+		const metaRefreshElement = parser.findFirst(
 			(el) =>
 				el.tagName.toLowerCase() === 'meta' && el.attribs['http-equiv']?.toLowerCase() === 'refresh'
 		);
@@ -99,7 +111,7 @@ export class HtmlPage {
 		this.isRedirect = Boolean(this.redirectTargetUrl);
 
 		// Get the page's canonical URL (if any)
-		const linkCanonicalElement = this.findFirst(
+		const linkCanonicalElement = parser.findFirst(
 			(el) =>
 				el.tagName.toLowerCase() === 'link' && el.attribs['rel']?.toLowerCase() === 'canonical'
 		);
@@ -107,14 +119,16 @@ export class HtmlPage {
 			(linkCanonicalElement && new URL(linkCanonicalElement.attribs['href'])) || null;
 
 		// Attempt to find the page's main content element
-		this.mainContent =
-			this.findFirst((el) => el.tagName.toLowerCase() === 'main') ||
-			this.findFirst((el) => el.tagName.toLowerCase() === 'body');
+		const mainContent =
+			parser.findFirst((el) => el.tagName.toLowerCase() === 'main') ||
+			parser.findFirst((el) => el.tagName.toLowerCase() === 'body');
+
+		this.hasContent = Boolean(mainContent);
 
 		// Attempt to determine the main content language by traversing the tree upwards
 		// until we find an element with a `lang` attribute
 		const mainContentParentWithLang =
-			this.mainContent && this.findParent(this.mainContent, (el) => Boolean(el.attribs?.lang));
+			mainContent && parser.findParent(mainContent, (el) => Boolean(el.attribs?.lang));
 		this.mainContentLang = mainContentParentWithLang?.attribs.lang || null;
 
 		// Attempt to determine the page's pathname-based language
@@ -123,23 +137,6 @@ export class HtmlPage {
 		// Detect if this is a language fallback page
 		this.isLanguageFallback =
 			Boolean(this.pathnameLang) && this.pathnameLang !== 'en' && this.mainContentLang === 'en';
-	}
-
-	findFirst(test: (elem: Element) => boolean) {
-		return DomUtils.findOne(test, this.dom.children);
-	}
-
-	findAll(test: (elem: Element) => boolean) {
-		return DomUtils.findAll(test, this.dom.children);
-	}
-
-	findParent(start: Element, test: (elem: Element) => boolean) {
-		let el: Element | null = start;
-		while (el) {
-			if (test(el)) return el;
-			el = DomUtils.getParent(el) as Element;
-		}
-		return null;
 	}
 
 	/**
@@ -161,5 +158,26 @@ export class HtmlPage {
 		// Only return parts that look like a two-letter language code
 		// with optional two-letter country code
 		if (firstPathPart.match(/^[a-z]{2}(-[a-zA-Z]{2})?$/)) return firstPathPart;
+	}
+}
+
+class DocumentParser {
+	constructor(public readonly dom: Document) {}
+
+	findFirst(test: (elem: Element) => boolean) {
+		return DomUtils.findOne(test, this.dom.children);
+	}
+
+	findAll(test: (elem: Element) => boolean) {
+		return DomUtils.findAll(test, this.dom.children);
+	}
+
+	findParent(start: Element, test: (elem: Element) => boolean) {
+		let el: Element | null = start;
+		while (el) {
+			if (test(el)) return el;
+			el = DomUtils.getParent(el) as Element;
+		}
+		return null;
 	}
 }
--- a/scripts/lib/linkcheck/checks/good-link-label.ts
+++ b/scripts/lib/linkcheck/checks/good-link-label.ts
@@ -1,4 +1,3 @@
-import { DomUtils } from 'htmlparser2';
 import kleur from 'kleur';
 import { dedentMd } from '../../output.mjs';
 import { CheckBase, type CheckHtmlPageContext } from '../base/check';
@@ -20,15 +19,13 @@ export class GoodLabels extends CheckBase {
 		if (context.page.isLanguageFallback) return;
 
 		context.page.anchors.forEach((anchor) => {
-			const linkLabel = DomUtils.innerText(anchor)
-				.replace(/[\n\s\t]+/g, ' ')
-				.trim();
+			const linkLabel = anchor.label.replace(/[\n\s\t]+/g, ' ').trim();
 
 			if (!blocklist.has(linkLabel.toLowerCase())) return;
 
 			context.report({
 				type: GoodLabels.BadLabel,
-				linkHref: anchor.attribs.href,
+				linkHref: anchor.href,
 				annotationText: dedentMd`Found link label “${linkLabel}”.
 						Please use descriptive accessible text for labels instead
 						of short undescriptive labels like “here” or “read more”.`,
--- a/scripts/lib/linkcheck/steps/build-index.ts
+++ b/scripts/lib/linkcheck/steps/build-index.ts
@@ -51,7 +51,7 @@ function parsePage(pathname: string, options: LinkCheckerOptions): HtmlPage {
 		const htmlPage = new HtmlPage({ html, href, pathname });
 
 		// Do not allow pages without main content unless they are a redirect
-		if (!htmlPage.isRedirect && !htmlPage.mainContent)
+		if (!htmlPage.isRedirect && !htmlPage.hasContent)
 			throw new Error('Failed to find main content - page has no <article> or <body>');
 
 		// Do not allow pages without a main content "lang" attribute unless they are a redirect
--- a/src/content/docs/de/guides/deploy/firebase.mdx
+++ b/src/content/docs/de/guides/deploy/firebase.mdx
@@ -14,7 +14,7 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 
 [Firebase Hosting](https://firebase.google.com/products/hosting) ist ein Dienst der App-Entwicklungsplattform [Firebase](https://firebase.google.com/) von Google, der für die Veröffentlichung einer Astro-Website genutzt werden kann.
 
-Siehe unseren separaten Leitfaden zum [Hinzufügen von Firebase-Backend-Diensten](/de/guides/backend/google-firebase/) wie Datenbanken, Authentifizierung und Speicher.
+Siehe unseren separaten Leitfaden zum [Hinzufügen von Firebase-Backend-Diensten](/de/guides/backend/firebase/) wie Datenbanken, Authentifizierung und Speicher.
 
 ## Projektkonfiguration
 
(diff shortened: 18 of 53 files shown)


src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -97,6 +97,150 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
...
+
+```ts title="my-integration.mjs"
+{
+  hooks: {
+    'astro:build:setup': ({ vite }) => {
+      vite.environments.client.build.minify = false;
+    }
+  }
+}
+```
```
ArmandPhilippot: Maybe the two code snippets can be merged (seeing the Expressive Code directives, I guess this was the intent first?)

````suggestion
```ts title="my-integration.mjs" del={3-7} ins={8-10}
{
  hooks: {
    'astro:build:setup': ({ target, vite }) => {
      if (target === 'client') {
        vite.build.minify = false;
      }
    }
    'astro:build:setup': ({ vite }) => {
      vite.environments.client.build.minify = false;
    }
  }
}
```
````
(we might need to check the syntax highlighting but I think I got it right...)

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 2]
```diff
@@ -97,6 +97,150 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
...
+#### What should I do?
+
+**Path properties are now `URL` objects:**
+
+When accessing path properties like `srcDir`, `outDir`, `cacheDir`, `publicDir`, `buildClientDir`, and `buildServerDir`, they are now `URL` objects instead of URL strings. If you were treating them as strings, you'll need to handle the `URL` object:
+
+```ts del={1} ins={2}
+const srcPath = manifest.srcDir; // was a URL string like "file:///path/to/src"
+const srcPath = manifest.srcDir; // now a URL object
+```
```
ArmandPhilippot: I'm not sure if `href` is the right property here but because this is a diff, maybe it would be helpful to show how they can access the same value instead of showing two identical lines? Something like:
````suggestion
```ts del={2} ins={3}
// To retrieve the same format (e.g., "file:///path/to/src"), make the following change:
const srcPath = manifest.srcDir;
const srcPath = manifest.srcDir.href;
```
````

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 3]
```diff
@@ -97,6 +97,150 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
...
+```ts del={1,2} ins={3,4}
+const mappings = manifest.serverIslandMappings;
+const driver = manifest.sessionDriver;
+const mappings = await manifest.serverIslandMappings?.();
+const driver = await manifest.sessionDriver?.();
+```
+
+<ReadMore>Learn more about [the Adapter API](/en/reference/adapter-reference/).</ReadMore>
+
+### RouteData.generate() Removed
```
ArmandPhilippot: The heading format in the upgrade guide is reversed:
```suggestion
### Removed: `RouteData.generate()`
```

But... `RouteData` was kinda internal, right? (I mean [some props were reused](https://docs.astro.build/en/reference/integrations-reference/#integrationresolvedroute-type-reference), but users don't access `RouteData` directly? Or am I missing something?)
Looking at the updated file, [`generate()` still exists](https://github.com/withastro/astro/blob/3e172ef85b62a20e5ff5a74147f25ad6e7a3337f/packages/astro/src/integrations/hooks.ts#L690) for integrations authors but this no longer relies on `RouteData`? The signature looks different based on the diff.

So, unless I'm totally wrong here, maybe this should be about the signature change instead of the removal?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -97,6 +97,150 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
...
+
+Remove any calls to `route.generate()` in your code. This method is no longer needed:
+
+```ts del={1}
+const generated = route.generate(params);
+```
+
+<ReadMore>Learn more about [routing](/en/guides/routing/).</ReadMore>
+
+### Percent-Encoding Security Fix
```
ArmandPhilippot: If we want to be consistent, I guess this should be a `Changed:` heading?
```suggestion
### Changed: Percent-Encoding Security Fix
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -97,6 +97,150 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
...
+If you have route files with `%25` in the filename, rename them to use a different character:
+
+```bash del={1} ins={2}
+src/pages/test%25file.astro
+src/pages/test-file.astro
+```
+
+This restriction prevents encoding-based security bypasses where `%25` decodes to `%`, potentially leading to ambiguous or invalid encoding sequences.
+
+### astro:ssr-manifest Virtual Module Removed
```
ArmandPhilippot: The heading format in the upgrade guide is reversed:
```suggestion
### Removed: `astro:ssr-manifest` virtual module
```
(not sure about the heading case though...)

ArmandPhilippot (commented):
Thanks Matthew, this looks great. I left a few suggestions, mostly about headings/code snippet consistency.

My main concern is the "RouteData.generate() Removed" section. I don't think we document `RouteData` directly, but I can see it used in the Integrations page to transmit properties. Looking at the implementation PR, it seems `generate()` still exist under another format. So, maybe this is the change we want to document here? (but maybe I'm wrong, the implementation PR contains a lot of changes, I haven't checked all the details yet 😅 )

Also, while I was quickly looking at the implementation PR, I noticed we have two new properties for `setAdapter()`: we'll need to document them in the Adapter API guide under [Building an adapter](https://docs.astro.build/en/reference/adapter-reference/#building-an-adapter). We now have a subheading for each available prop, so we'll need to add `devEntrypoint` and `entryType` there.

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: Hmm sorry, it's seems the GitHub new experience is buggy again. Looking at the suggested diff, this is wrong: I meant to replace the two code snippets with this. So, this might require a manual edit... or, let me resolve this and try to make a new suggestion by switching back to the old experience!

ArmandPhilippot: Hoping this time the diff is okay (the preview was okay in both cases). I'm copying again my previous comment for anyone looking at this.

Maybe the two code snippets can be merged (seeing the Expressive Code directives, I guess this was the intent first?)
````suggestion
```ts title="my-integration.mjs" del={3-7} ins={8-10}
{
  hooks: {
    'astro:build:setup': ({ target, vite }) => {
      if (target === 'client') {
        vite.build.minify = false;
      }
    }
    'astro:build:setup': ({ vite }) => {
      vite.environments.client.build.minify = false;
    }
  }
}
```
````
(we might need to check the syntax highlighting but I think I got it right...)

File: src/content/docs/en/guides/upgrade-to/v6.mdx (hunk 3 above)
Comment (matthewp) [reply src/content/docs/en/guides/upgrade-to/v6.mdx]: I can't think of a use-case for why someone would use it. But it is receivable in the build:done event and probably other places, so technically it could be called, perhaps to like create a bunch of URLs based on params for post-processing stuff. I could see sitemap like integrations wanting to use it.

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 6]
```diff
@@ -97,6 +97,147 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
 
 If you are using Vite-specific plugins, configuration, or APIs, check the [Vite migration guide](https://vite.dev/guide/migration) for their breaking changes and upgrade your project as needed.
 
+## Environment API Changes
+
+Astro v6.0 introduces significant changes to how Astro manages different runtime environments (client, server, and prerender) using Vite's new Environments API. Integration and adapter maintainers should pay special attention to these changes.
```
sarah11918: Thanks for this @matthewp ! Letting you know that I'm working on this, because it's a lot of content especially at the beginning of the guide that probably only affects integration/adapter authors. I'm thinking we could handle it similar to our removing experimental flags: mention the changes and link to the item entry that shows up in its regular place below:  

image

So my suggestion would be to address the Vite Environment API briefly, like a dependency (it *kind* of is, as a thing we're using) maybe something like:

```suggestion
### Vite Environment API

<SourcePR number="14306" title="feat: integrate vite environments"/>

Astro v6.0 introduces significant changes to how Astro manages different runtime environments (client, server, and prerender) after an internal refactor to use [Vite's new Environments API](https://vite.dev/guide/api-environment).

Integration and adapter maintainers should pay special attention to changes affecting these parts of the Integration API and Adapter API:

- [integration hooks and HMR access patterns](#changed-integration-hooks-and-hmr-access-patterns-integration-api)
- [`SSRManifest` structure](#changed-ssrmanifest-interface-structure-adapter-api)
- [generating routes with `RouteData`](#removed-routedatagenerate-integration-api)
- [routes with percent-encoded percent signs (e.g. `%25`)](#removed-percent-encoding-in-routes)
- [astro:ssr-manifest` virtual module](#removed-astrossrmanifest-virtual-module-integration-api)
```

And then that's it, short and sweet, and those 5 entries of things that were changed/removed etc. we would add to the lists below. Part of my thinking is yes, these changes were all made when implementing Vite's API, but as breaking changes, they aren't otherwise in their own "category", and people still need to see ALL the breaking adapter/integration API changes. So, this makes people aware that there was a big refactor and lots of changes to those APIs as a result, but doesn't make e.g. someone who just needs to know how to update their content collections navigate through/past all this.

So, I'm working on a version that then moves the content below into place, and traces each of those changes back to the rest of docs to the thing that they relate to to make sure that e.g. removed things are in fact removed from existing docs.

That means that @ArmandPhilippot and I might have some questions to verify regular docs content too as we go through this! (I already suspect that `RouteData.generate` needs some attention, because I already came across something in an earlier PR with @florian-lefebvre that our docs currentlysay `RouteData` was deprecated and to use the `IntegrationsRouteData` (which is a subset of the internal `RouteData` I believe?)

So be prepared for some questions as we work through this, but I think something like my suggestion above might be a good way to handle this!

sarah11918:
I just added a commit to reorganize and start to put these in our format.

- These still need full editing and tracing through to docs to make sure all changes are reflected in regular v6 docs! 
- I have left some code comments in items for further work

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: Copy pasting from Discord
> Re "RouteData.generate()": I think it makes sense to document `RouteData` on the adapter reference but not on the integrations reference (because they have `IntegrationRouteResolved` or something, hence why it was marked as deprecated there). And about `generate()` in particular, I don't remember exactly but I think there's a utility instead now

matthewp:
@sarah11918 to understand what is next, it looks like you've made the changes from the comment yourself. Is there anything remaining for me to do? I need to find places in the regular docs related to these changes and update those? Anything else?

sarah11918:
@matthewp I hope it's not too inconvenient, since I needed to move things around it was easier for me to make a single commit. I normally prefer leaving suggestions, but that wouldn't have been efficient here!

What *you* can do:
- review the entries (that are moved) and make sure they are still accurate/correct, and/or make any editing suggestions you like!

What *I* will do with @ArmandPhilippot : 
- trace through each of these entries to look to see how they affect docs
- leave comments where we might need guidance on updating the rest of the docs

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -425,6 +439,64 @@ import {
 
 <ReadMore>Learn more about all utilities available in the [Actions API Reference](/en/reference/modules/astro-actions/).</ReadMore>
 
+### Removed: Percent-Encoding in routes
```
sarah11918, line 469: I don't believe any extra docs changes are needed for this.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -425,6 +439,64 @@ import {
...
+
+If you have route files with `%25` in the filename, rename them to use a different character:
+
+```bash del={1} ins={2}
+src/pages/test%25file.astro
+src/pages/test-file.astro
+```
+
+### Removed: `astro:ssr-manifest` virtual module (Integration API)
+{/* I can't find any record of this ever being documented anywhere in docs, but I am assuming it was deprecated at some point, and since it's removed, we don't need anyting in docs now */}
```
```suggestion
### Removed: `astro:ssr-manifest` virtual module (Integration API)
```
I don't believe we've ever showed using this module in docs. But, Armand found that this was publicly exposed. So, we can document its removal to be safe, but no changes to docs should be needed.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1053,97 @@ export function getStaticPaths() {
 
 <ReadMore>Learn more about [dynamic SSG routes with `getStaticPaths()`](/en/guides/routing/#static-ssg-mode).</ReadMore>
 
+### Changed: Integration hooks and HMR access patterns (Integration API)
```
sarah11918, line 1123: Things to check:
The use of `target` in 'astro:build:setup' is shown in a few places on the Integration API page (but not the Adapter API at all):

https://docs.astro.build/en/reference/integrations-reference/#quick-api-reference
In the Quick API reference code block:
- is the `target` parameter of 'astro:build:setup' removed entirely? Or does it still exist, it's just not used with `vite`?

https://docs.astro.build/en/reference/integrations-reference/#astrobuildsetup
In the API reference itself for /en/reference/integrations-reference/#astrobuildsetup
- also remove mention of `target` here? Is this still valid use, or has the parameter been removed entirely?

https://docs.astro.build/en/reference/integrations-reference/#target-option
API reference for `target` itself
- has this been updated or removed? Does it exist for the usage shown, or does that need updating too?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1053,97 @@ export function getStaticPaths() {
...
+Replace `server.hot.send()` with `server.environments.client.hot.send()`:
+
+```ts del={1} ins={2}
+server.hot.send(event)
+server.environments.client.hot.send(event)
+```
+
+<ReadMore>Learn more about [Vite Environments API](https://vite.dev/guide/api-environment) and [integration hooks](/en/reference/integrations-reference/#astrobuildsetup).</ReadMore>
+
+### Changed: `SSRManifest` interface structure (Adapter API)
```
sarah11918, line 1163: Things to check:

On the Integrations API page:
Is the typing `manifest: SerializedSSRManifest;` still correct in code samples (it's just that `SerializedSSRManifest` itself has changed)? Will note: we don't mention `SSRManifest` at all, we only ever refer to `SerializedSSRManifest`

https://docs.astro.build/en/reference/integrations-reference/#manifest-option
- we link [to source code](https://github.com/withastro/astro/blob/3b10b97a4fecd1dfd959b160a07b5b8427fe40a7/packages/astro/src/core/app/types.ts#L91-L109) to define that, rather than to something else in docs. We will have to verify this link (assuming that we do not move this into docs proper)

On the Adapter API page:
- we describe in types `(manifest: SSRManifest, .....)` (`start()` and `createExports()`), again, assuming that saying `manifest` is of type `SSRManifest` does not need updating just because `SSRManifest` might have changed

https://docs.astro.build/en/reference/adapter-reference/#envgetsecret
In describing `env` support, we also import and use `SSRManifest` type, which should be fine.

ALSO CHECKING re: `createExports()` function -- I know the implementation PR documents in a `@astrojs/cloudflare` major that this function is no longer the pattern to use for creating a custom entrypoint. BUT, we are not removing this function? It still exists and should be documented here, and still works the same? (It's just no longer used for that particular purpose)?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1053,97 @@ export function getStaticPaths() {
...
+
+```ts del={2} ins={3}
+// To retrieve the same format (e.g., "file:///path/to/src"), make the following change:
+const srcPath = manifest.srcDir;
+const srcPath = manifest.srcDir.href;
+```
+
+{/*  These shouldn't be a breaking change, since they're new. But they should be documented in the appropriate API page if these are publicly available! */}
+
+{/* STUFF FOR REGULAR DOCS, PROBABLY??
```
sarah11918: I don't believe any of this stuff is documented anywhere. I don't know whether we should? If it's new properties available to use, we normally wouldn't include that in a breaking changes guide (because adding new things generally isn't breaking). Is this the case?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1053,97 @@ export function getStaticPaths() {
...
+- `assetsDir: string` — Directory containing built assets
+- `serverLike: boolean` — Whether this is a server-like environment
+- `logLevel: LoggerLevel` — Logging level
+
+**Removed `hrefRoot` property:**
+
+The `hrefRoot` property is no longer available on the manifest.
+
+**Async methods for dynamic data:**
+
```
sarah11918: Oops! These got stuck in the code comment, and are likely breaking changes since they are removed or updated and should be reintroduced!

(It does feel weird to say "here are new changed properties for this thing we don't actually show in docs" though. @ArmandPhilippot do you think it's worth figuring out how to document this one thing in the docs itself, even if the page is not totally updated vs. linking to the code sample?)

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -425,6 +439,64 @@ import {
...
+```ts del={1} ins={2,3}
+import { manifest } from 'astro:ssr-manifest';
+import { srcDir, outDir, root } from 'astro:config/server';
+// Use srcDir, outDir, root, etc. for configuration values
+```
+
+<ReadMore>Learn more about [the `astro:config` virtual module](/en/reference/modules/astro-config/).</ReadMore>
+
+### Removed: `RouteData.generate()` (Integration API)
```
sarah11918: I think this is the one that will need the most figuring out from our end, so starting that thread here.

In the v5 upgrade guide, we had an entry about [replacing `RouteData` with `IntegrationRouteData`](https://docs.astro.build/en/guides/upgrade-to/v5/#changed-routedata-type-replaced-by-integrationsroutedata-integrations-api). We gave the instruction:

>Update your adapter to change the type of entryPoints from RouteData to IntegrationRouteData.

And maybe we didn't update the docs properly last time, or maybe this isn't related, but there are still [places where we show](https://docs.astro.build/en/reference/integrations-reference/#integrationresolvedroute-type-reference) ` entrypoint: RouteData['component'];`. As I think Armand mentioned earlier, I *think* `RouteData` is internal, or meant to be, and `IntegrationRouteData` might be the subset that is meant to actually be used?

But admittedly, I'm not familiar here, and maybe I need Armand to be asking the questions to clarify whether we mean to be showing `RouteData` *anywhere*? Whether we mean for everyone to be using `IntegrationRouteData` *everywhere* instead? These are questions I might not even be asking properly!

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: From my understanding, this only means the `if` condition is no longer required because somehow `vite.environments` will be able to infer the current environment. So, `target` is still there, with the same type.  I don't think we need any changes there.

(Because it's not necessarily easy to browse all the files in the PR, I pulled the branch locally to check the current code in that branch.)

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: > (it's just that SerializedSSRManifest itself has changed)?

`SSRManifest` changed and, indirectly, `SerializedSSRManifest` because this type is based on `SSRManifest`. But because we don't documented any of those types, I don't think the docs need an update here outside the upgrade guide.

> Will note: we don't mention SSRManifest at all, we only ever refer to SerializedSSRManifest

I don't think `SSRManifest` is directly available for integrations authors, only for adapters authors. So, I don't think this affects our current docs.

> We will have to verify this link (assuming that we do not move this into docs proper)

The link is still relevant. But, yes, this should probably be proper docs. I recall adding this link based on feedbacks but I don't recall the full context (too internal? only missing docs?). A lot happens since I've updated that page. 😄 

> On the Adapter API page:

Yeah, I don't think this affects the current docs because we don't document what `SSRManifest` is, so the changes are currently "invisible" regarding docs. Only the upgrade guide requires this info I think.

> re: createExports() function 

I don't think I can answer that one. I was already in an unfamiliar directory when I refactored that page. I noticed the same thing as you, but honestly, I don't know.

Regarding the recurrent question around `SSRManifest` and `SerializedSSRManifest`:
Because this is public, yes, we should consider documenting them... Not sure where though, because `SSRManifest` is only relevant for Adapters authors (I think?). So, linking from the Integration API to the Adapter API (to avoid repeating props description) might be a bit awkward. Something to think about, but not directly related to this PR. This can be done in `main` (with proper updates for v6), or in v6. No need to do this with this PR.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1053,97 @@ export function getStaticPaths() {
...
+Astro 6.0 changes the form of these path properties to `URL` objects instead of URL strings. With this change, several new properties are now available on the manifest.
+
+#### What should I do?
+
+If you were treating these path properties as strings, you'll now need to handle the `URL` object. For example, you will now need to access the `href` property of the `URL` object:
+
+```ts del={1} ins={2}
+const srcPath = manifest.srcDir; // was a URL string like "file:///path/to/src"
+const srcPath = manifest.srcDir; // now a URL object
+
```
ArmandPhilippot: I know the diff looks weird, but this should be fine. I have suggested a change to that code snippet but it seems this one was also bugged because of the Github new experience (which seems to be the default for me now? So I need to think to switch back each time I review...) so:
```suggestion
```

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: * The `srcDir` change should be here
* The `hrefRoot` removal too, even if it is not documented, because this is currently available for Adapters authors (at least)
* But, yes, I believe the new properties should not be there.

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: I'm not sure what you mean by "vs. linking to the code sample".

I don't know how many people use these properties, but because they are available in `manifest` we'll need to document them at some point. Now, I don't think this needs to be now and only mentioning this in the upgrade guide is enough?

Sorry, if I don't fully answer your question. As I said, I'm not sure I got it right. 😄

src/content/docs/en/guides/upgrade-to/v6.mdx
matthewp: I wouldn't think so.

src/content/docs/en/guides/upgrade-to/v6.mdx
matthewp: Yes, it was never talked about, just created for the `bling` project that briefly existed.

matthewp: I don't knowif you want me to accept the commit or not, so I'll let you do that.

src/content/docs/en/guides/upgrade-to/v6.mdx
matthewp: It might not have actually been removed yet but it has no functional use-case any more, so I would remove all documentation about it. Using vite environments config is the way to do what you used to use this for.

I'll make a note to make sure we remove it from the code.

src/content/docs/en/guides/upgrade-to/v6.mdx
matthewp: SerializedSSRManifest is for the `astro:build:ssr` hook, that has changed indirectly as @ArmandPhilippot says.

> BUT, we are not removing this function? It still exists and should be documented here, and still works the same?

Yes, this still exists and works exactly the same as before.

src/content/docs/en/guides/upgrade-to/v6.mdx
matthewp: Sounds right to me.

src/content/docs/en/guides/upgrade-to/v6.mdx
matthewp: I struggled with this too, these are deep APIs for adapters only, and not stuff we even intended for them to use other than to pass to the `App` class. I think we've only ever documented `new App(manifest)`. But since the adapter does get this object it might use it for other means. That makes it difficult to document in this way since we don't really know the use-cases.

Happy with whatever solution you prefer.

src/content/docs/en/guides/upgrade-to/v6.mdx
matthewp: yeah I think changing this to `IntegrationRouteData` is fair. The `RouteData` object is available to adapters via the manifest, but the most part `IntegrationRouteData` is where people are more likely to use this (in the hooks). It's mostly the same object so documenting that this property is removed still seems worthy to me.

ArmandPhilippot: I just checked: `RouteData` is importable but I don't know if this a side effect or not. When I look at a core file importing it, it comes from `types/public/internal.js`... So, yeah, this sounds like internal. 😅 
But this is available for users? I can see it in use in `RenderOptions` (Adapter API):
```ts
/**
 * **Advanced API**: you probably do not need to use this.
 *
 * Default: `app.match(request)`
 */
routeData?: RouteData;
```
So, someone could have used `routeData.generate()`.

I'm not sure if this is accessible elsewhere. But because this is importable and usable through an "Advanded API", documenting `RouteData.generate()` removal in the upgrade guide probably makes sense.

And, it doesn't seem the API available for Integrations authors has changed? I think I got confused when I was looking at the diff earlier:
```diff
-generate: route.generate,
+generate: getRouteGenerator(route.segments, trailingSlash),
```

But I can see, in the branch, that `IntegrationResolvedRoute` (the type replacing `IntegrationRouteData`) still uses the types we use in docs:
```
generate: (data?: any) => string;
```

So, I wasn't sure why my previous comment was resolved but, now I'm looking at the codebase and not only at the PR, I think this is okay.

> In the v5 upgrade guide, we had an entry about [replacing RouteData with IntegrationRouteData](https://docs.astro.build/en/guides/upgrade-to/v5/#changed-routedata-type-replaced-by-integrationsroutedata-integrations-api). We gave the instruction:

Yeah, I don't recall when `IntegrationRouteData` has been deprecated but maybe we should have used `IntegrationResolvedRoute`, here? (see also https://github.com/withastro/docs/issues/12789)
We don't have any `Since` for them... not sure if this because both were available for a long time or if this is an oversight from me (ie. I thought about using `<Since />` for functions, but not for types).

> As I think Armand mentioned earlier, I think RouteData is internal, or meant to be

From what I see now, `RouteData` is a bit tricky: this sounds like it was meant to be internal but got exposed at some point for advanced use case only? That's probably why we don't document it... But because this is importable and available for Adapter authors, maybe we should (that's also what Florian suggests in the issue I linked above)? And maybe this should be moved outside of `types/internal` in core?

> and IntegrationRouteData might be the subset that is meant to actually be used?

Not quite a subset, but it reuses some properties to avoid redeclaring the same types (probably because internally they are tied). So, from what I see `RouteData` is available in adapters and `IntegrationResolvedRoute` (the replacement of `IntegrationRouteData`) in integrations.

> But admittedly, I'm not familiar here, and maybe I need Armand to be asking the questions to clarify whether we mean to be showing RouteData anywhere? Whether we mean for everyone to be using IntegrationRouteData everywhere instead? These are questions I might not even be asking properly!

My answer is quite verbose but to summarize: I think the current docs for the upgrade guide is fine.
However, yes, I think there are some work to do around `RouteData`:
* is this internal? Should this type be moved outside `types/internal` to make it more clear if this is meant to be public or not?
* if this is not meant to be public, we probably need to update the `RouteData["something"]` references in docs to show the actual types?
* if this is meant to be public, we probably need to document `RouteData` somewhere to avoid linking to the code source directly and to give proper docs for the properties available there. Even if this is advanced use cases, if this is public I think we should have docs for this?

And final though: unless I missed something here, maybe the heading should refer to `Adapter API` instead of `Integration API`?

matthewp: It's meant to be public only in that its on the manifest which is available to adapters. Otherwise no, I don't believe so.

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: Thanks! The commit was just to remove my code comment here that was a placeholder, so I can totally do that!

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: OK, so if no one is going to actually *use* target ever (for any integration/adapter), then yeah, we'll remove it from the documentation for this change.

If the code happens to stick around for a bit of time and happens to work, then no big deal. We'll just never show it.

sarah11918: Will resolve this conversation because we can now see the change made and anyone can comment there if that's not the right thing to do!

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: OK, so what I'm getting from this is:

- the only changes are types that we don't actually document the shape of
- we maybe / probably *should* document these, but because we don't currently, docs themselves don't need to update
- Although we don't want people to use `createExports()` with the Cloudflare adapter anymore to create a custom entrypoint, we're not actually changing anything about the existing function so no docs for it need to change.

Which means, this entry is fine as is with no accompanying docs changes, and the docs are NWTWWHB!

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: So, this is what we currently show for `manifest` (and yes, as Matthew says, I think we only ever show using it as `new App(manifest)` which I guess means knowing these properties isn't super important?

image

That link to `SerializedSSRManifest` is external, and points to Astro source code: https://github.com/withastro/astro/blob/3b10b97a4fecd1dfd959b160a07b5b8427fe40a7/packages/astro/src/core/app/types.ts#L91-L109

(We really prefer whenever possible to NOT do that, and to create a docs entry somewhere instead. As we're finding, that might be a challenge and in the spirit of "not worse than what we had before", doesn't necessarily need to be done now.)

BUT if the code is changing, then it's possible that link changes in the new code? It might not be at those exact lines anymore? So, this link will need checking in the new `next` branch?

Does this also mean that maybe people don't need to know about the new properties right now? (Since they don't actually know about any of the properties anyway)? I think the "quick and easy" solution is not to document the new content for `manifest` anywhere (until we're ready to properly document it).

But, I will "uncomment out" the `href` and other stuff in these lines that is actual changed/removed stuff, since those are breaking changes.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1052,93 @@ export function getStaticPaths() {
...
+When accessing `serverIslandMappings` and `sessionDriver`, they are now async methods:
+
+```ts del={1,2} ins={3,4}
+const mappings = manifest.serverIslandMappings;
+const driver = manifest.sessionDriver;
+const mappings = await manifest.serverIslandMappings?.();
+const driver = await manifest.sessionDriver?.();
+```
+*/}
```
``````suggestion

If you were accessing the `hrefRoot` property, you will need to remove it as is no longer available on the manifest.

Update any use of `serverIslandMappings` and `sessionDriver`, they are now async methods:

```ts del={1,2} ins={3,4}
const mappings = manifest.serverIslandMappings;
const driver = manifest.sessionDriver;
const mappings = await manifest.serverIslandMappings?.();
const driver = await manifest.sessionDriver?.();
```
``````

File: src/content/docs/en/guides/upgrade-to/v6.mdx (hunk 12 above)
Comment (sarah11918) [reply src/content/docs/en/guides/upgrade-to/v6.mdx]: Making a commit that I think makes sense! 

But still want to check whether that link needs updating! (recopying so it's easier to spot)

image

That link to `SerializedSSRManifest` is: https://github.com/withastro/astro/blob/3b10b97a4fecd1dfd959b160a07b5b8427fe40a7/packages/astro/src/core/app/types.ts#L91-L109

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 16]
```diff
@@ -981,6 +1052,78 @@ export function getStaticPaths() {
...
+
+If you were treating these path properties as strings, you'll now need to handle the `URL` object. For example, you will now need to access the `href` property of the `URL` object:
+
+```ts del={2} ins={3}
+// To retrieve the same format (e.g., "file:///path/to/src"), make the following change:
+const srcPath = manifest.srcDir;
+const srcPath = manifest.srcDir.href;
+```
+
+
```
sarah11918, line 1182:
```suggestion
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1052,77 @@ export function getStaticPaths() {
...
+<ReadMore>Learn more about [Vite Environments API](https://vite.dev/guide/api-environment) and [integration hooks](/en/reference/integrations-reference/#astrobuildsetup).</ReadMore>
+
+### Changed: `SSRManifest` interface structure (Adapter API)
+
+<SourcePR number="14306" title="feat: integrate vite environments"/>
+
+In Astro 5.x, path properties of the `SSRManifest` interface like `srcDir`, `outDir`, `cacheDir`, `publicDir`, `buildClientDir`, and `buildServerDir` were URL strings.
+
+Astro 6.0 changes the form of these path properties to `URL` objects instead of URL strings. With this change, several new properties are now available on the manifest.
```
```suggestion
Astro 6.0 changes the form of these path properties to `URL` objects instead of URL strings. With this change, several new properties are now available on the manifest and others have been updated or removed.
```

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Good call, Sarah! The path hasn't changed, but the line numbers have...
```suggestion
**Type:** [`SerializedSSRManifest`](https://github.com/withastro/astro/blob/95a1969a05cc9c15f16dcf2177532882bb392581/packages/astro/src/core/app/types.ts#L126-L144)
```
(the diff looks like I'm updating a blank line, but this shouldn't be the case: I'm just replacing the link)

And while I was checking where was that line... I noticed we have a few links to the source code instead of docs. Something to consider for later.

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: Yeah, good call on this! I made a suggestion but in case you don't trust the new GitHub experience 😅 Here's the link matching the next branch: https://github.com/withastro/astro/blob/95a1969a05cc9c15f16dcf2177532882bb392581/packages/astro/src/core/app/types.ts#L126-L144

Edit: seeing your last comment... hmm okay this might not be accurate neither because the Environment API is not merged yet. Maybe: https://github.com/withastro/astro/blob/71d238229603811273e47afe78ec8eeb294f635d/packages/astro/src/core/app/types.ts#L157-L188

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -425,6 +439,63 @@ import {
...
+
+```ts del={1} ins={2,3}
+import { manifest } from 'astro:ssr-manifest';
+import { srcDir, outDir, root } from 'astro:config/server';
+// Use srcDir, outDir, root, etc. for configuration values
+```
+
+<ReadMore>Learn more about [the `astro:config` virtual module](/en/reference/modules/astro-config/).</ReadMore>
+
+### Removed: `RouteData.generate()` (Integration API)
```
```suggestion
### Removed: `RouteData.generate()` (Adapter API)
```
Also changed in the implementation PR changeset!

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: So what I take from this is:

- This upgrade guide entry is fine as is, and updated to reflect that it best relates to the Adapter API.

- Of Armand's summary questions probably the most relevant answer is `RouteData` is technically public but probably not intended to be *used* directly except maybe in advanced cases. Being public, it's probably worth eventually documenting `RouteData` on the Adapter API page, but this is a NWTWWHB situation we don't have to figure out here.

- The Integration API does show `RouteData["something"]` in reference entries in the docs, and this likely *should* be updated to `IntegrationResolvedRoute["something"]` in those places, and since this seems to only be relevant for integrations, I'd suggest we document it on this page too. (also NWTWWHB)

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -425,6 +439,63 @@ import {
...
+#### What should I do?
+
+Remove any calls to `route.generate()` in your code. This method is no longer needed:
+
+```ts del={1}
+const generated = route.generate(params);
+```
+
+{/* I don't think we show anything like this on the routing page, so maybe linking to Integration API is better? */}
+<ReadMore>Learn more about [routing](/en/guides/routing/).</ReadMore>
```
```suggestion
<ReadMore>Learn more about [the Adapter API](/en/reference/adapter-reference/).</ReadMore>
```

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: Regarding the last point, I don't think this is correct?
```ts
interface IntegrationResolvedRoute {
  pattern: RouteData['route'];
  patternRegex: RouteData['pattern'];
  entrypoint: RouteData['component'];
  isPrerendered: RouteData['prerender'];
 /* more things */
}
``` 

So, this shouldn't be replaced by `IntegrationResolvedRoute["something"]` (what I show above IS `IntegrationResolvedRoute`). What we could eventually do is: get the right type for each properties to replace `RouteData['something']`. Which I think makes sense since we don't document `RouteData` and so showing that is not really helpful. But, yes, in any cases this is NWTWWHB.

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: *shakes fist*... I can't seem to accept this, so I'll update the link in the file I have open already! 😄

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: Ah my edit was too late. We might need to check all the links to the source code once the Environment API is merged because right now between the `feat/environment-api` branch and `next` we don't have the same line numbers.

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: Ah, ok! Yes, anyway that's Future Armand's problem. 🫡

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: Thank you!! Updated and resolving!

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -425,6 +439,62 @@ import {
...
 
+### Removed: Percent-Encoding in routes
+
+<SourcePR number="14306" title="feat: integrate vite environments"/>
+
+In Astro 5.x, it was possible to include a percent-encoded percent sign (`%25`) in filenames.
+
+Astro 6.0 removes support for the characters `%25` in filenames for security reasons. This restriction prevents encoding-based security bypasses where `%25` decodes to `%`, potentially leading to ambiguous or invalid encoding sequences.
+
+
```
```suggestion
```

sarah11918:
Will wait for checks and Netlify so I can check the deploy preview visually, but otherwise, I think the content here seems fine now!

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -97,6 +97,20 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
...
+<SourcePR number="14306" title="feat: integrate vite environments"/>
+
+Astro v6.0 introduces significant changes to how Astro manages different runtime environments (client, server, and prerender) after an internal refactor to use [Vite's new Environments API](https://vite.dev/guide/api-environment).
+
+Integration and adapter maintainers should pay special attention to changes affecting these parts of the Integration API and Adapter API (full details included below with other breaking changes to these APIs):
+
+- [integration hooks and HMR access patterns](#changed-integration-hooks-and-hmr-access-patterns-integration-api)
+- [`SSRManifest` structure](#changed-ssrmanifest-interface-structure-adapter-api)
+- [generating routes with `RouteData`](#removed-routedatagenerate-integration-api)
```
```suggestion
- [generating routes with `RouteData`](#removed-routedatagenerate-adapter-api)
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -1008,6 +1077,77 @@ export function getStaticPaths() {
...
+**For dev toolbar and integration code accessing HMR:**
+
+Replace `server.hot.send()` with `server.environments.client.hot.send()`:
+
+```ts del={1} ins={2}
+server.hot.send(event)
+server.environments.client.hot.send(event)
+```
+
+<ReadMore>Learn more about [Vite Environments API](https://vite.dev/guide/api-environment) and [integration hooks](/en/reference/integrations-reference/#astrobuildsetup).</ReadMore>
```
```suggestion
<ReadMore>Learn more about the [Vite Environment API](https://vite.dev/guide/api-environment) and Astro [integration hooks](/en/reference/integrations-reference/#astrobuildsetup).</ReadMore>
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -1008,6 +1077,77 @@ export function getStaticPaths() {
...
+
+<SourcePR number="14306" title="feat: integrate vite environments"/>
+
+In Astro 5.x, path properties of the `SSRManifest` interface like `srcDir`, `outDir`, `cacheDir`, `publicDir`, `buildClientDir`, and `buildServerDir` were URL strings.
+
+Astro 6.0 changes the form of these path properties to `URL` objects instead of URL strings. With this change, several new properties are now available on the manifest and others have been updated or removed.
+
+#### What should I do?
+
+If you were treating these path properties as strings, you'll now need to handle the `URL` object. For example, you will now need to access the `href` property of the `URL` object:
```
```suggestion
If you were treating these path properties as strings, you will now need to handle the `URL` object. For example, you will now need to access the `href` property of the `URL` object:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -1008,6 +1077,77 @@ export function getStaticPaths() {
...
+
+```ts del={2} ins={3}
+// To retrieve the same format (e.g., "file:///path/to/src"), make the following change:
+const srcPath = manifest.srcDir;
+const srcPath = manifest.srcDir.href;
+```
+
+If you were accessing the `hrefRoot` property, you will need to remove it as is no longer available on the manifest.
+
+Update any use of `serverIslandMappings` and `sessionDriver`, they are now async methods:
```
```suggestion
Update any use of `serverIslandMappings` and `sessionDriver`. These are now async methods:
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -97,6 +97,20 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
...
+
+Astro v6.0 introduces significant changes to how Astro manages different runtime environments (client, server, and prerender) after an internal refactor to use [Vite's new Environments API](https://vite.dev/guide/api-environment).
+
+Integration and adapter maintainers should pay special attention to changes affecting these parts of the Integration API and Adapter API (full details included below with other breaking changes to these APIs):
+
+- [integration hooks and HMR access patterns](#changed-integration-hooks-and-hmr-access-patterns-integration-api)
+- [`SSRManifest` structure](#changed-ssrmanifest-interface-structure-adapter-api)
+- [generating routes with `RouteData`](#removed-routedatagenerate-integration-api)
+- [routes with percent-encoded percent signs (e.g. `%25`)](#removed-percent-encoding-in-routes)
+- [astro:ssr-manifest` virtual module](#removed-astrossr-manifest-virtual-module-integration-api)
```
```suggestion
- [`astro:ssr-manifest` virtual module](#removed-astrossr-manifest-virtual-module-integration-api)
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -452,6 +466,61 @@ import {
...
+
+<ReadMore>Learn more about [the `astro:config` virtual module](/en/reference/modules/astro-config/).</ReadMore>
+
+### Removed: `RouteData.generate()` (Adapter API)
+
+<SourcePR number="14306" title="feat: integrate vite environments"/>
+
+In Astro 5.x, routes could be generated using the `generate()` method on `RouteData`.
+
+Astro 6.0 removes `RouteData.generate` because route generation is now handled internally by Astro.
```
ArmandPhilippot: nit for consistency:
```suggestion
Astro 6.0 removes `RouteData.generate()` because route generation is now handled internally by Astro.
```

ArmandPhilippot (approved):
I left a small nit, but this LGTM! 🎉

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -452,6 +466,61 @@ import {
...
+src/pages/test-file.astro
+```
+
+### Removed: `astro:ssr-manifest` virtual module (Integration API)
+
+<SourcePR number="14306" title="feat: integrate vite environments"/>
+
+In Astro 5.x, the deprecated `astro:ssr-manifest` virtual module could still be used to access configuration values.
+
+Astro 6.0 removes the `astro:ssr-manifest` virtual module entirely. It is no longer used by integrations or internally by Astro. The manifest is now passed directly through integration hooks and adapter APIs rather than through a virtual module. For build-specific manifest data, use the `astro:build:ssr` integration hook which receives the manifest as a parameter.
```
yanthomasdev: comma nit
```suggestion
Astro 6.0 removes the `astro:ssr-manifest` virtual module entirely. It is no longer used by integrations or internally by Astro. The manifest is now passed directly through integration hooks and adapter APIs rather than through a virtual module. For build-specific manifest data, use the `astro:build:ssr` integration hook, which receives the manifest as a parameter.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -1008,6 +1077,77 @@ export function getStaticPaths() {
...
+
+<ReadMore>Learn more about the [Vite Environment API](https://vite.dev/guide/api-environment) and Astro [integration hooks](/en/reference/integrations-reference/#astrobuildsetup).</ReadMore>
+
+### Changed: `SSRManifest` interface structure (Adapter API)
+
+<SourcePR number="14306" title="feat: integrate vite environments"/>
+
+In Astro 5.x, path properties of the `SSRManifest` interface like `srcDir`, `outDir`, `cacheDir`, `publicDir`, `buildClientDir`, and `buildServerDir` were URL strings.
+
+Astro 6.0 changes the form of these path properties to `URL` objects instead of URL strings. With this change, several new properties are now available on the manifest and others have been updated or removed.
```
yanthomasdev:
```suggestion
Astro 6.0 changes the form of these path properties to `URL` objects instead of URL strings. With this change, several new properties are now available on the manifest, and others have been updated or removed.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -1008,6 +1077,77 @@ export function getStaticPaths() {
...
+
+If you were treating these path properties as strings, you will now need to handle the `URL` object. For example, you will now need to access the `href` property of the `URL` object:
+
+```ts del={2} ins={3}
+// To retrieve the same format (e.g., "file:///path/to/src"), make the following change:
+const srcPath = manifest.srcDir;
+const srcPath = manifest.srcDir.href;
+```
+
+If you were accessing the `hrefRoot` property, you will need to remove it as is no longer available on the manifest.
```
yanthomasdev: typo and extra comma to make it flow better with the additional "it" to it
```suggestion
If you were accessing the `hrefRoot` property, you will need to remove it, as it is no longer available on the manifest.
```

sarah11918:
This is so exciting!! I know the Environment API refactors have been a massive effort from the entire core framework team. I hope we've done y'all justice in docs, and will merge this into the v6 branch. There's lots of time before v6 if anyone spots anything in the upgrade guide after the fact, but let's all enjoy this being DONE!

matthewp:
Thanks @sarah11918 !

PR #12810, opened 2025-11-28
[v6] Update content loader reference

ArmandPhilippot:
#### Description (required)

**Content Loader API:**

- Small fixes:
  - indentation again in a code snippet, not easy to fix in GitHub reviews 🙄 
  - a link was redirecting to the homepage instead of a heading on the same page
  - typo
  - replace `content` with `htmlContent` in the `Article` interface
  - add parenthesis to headings using a function name
- Updates a sentence in "Distributing your loader" because we were only talking about live loaders but the same applies to build time loaders
- Updates "Creating live loader error types" example: the `constructor` takes an optional `code` but we never assign it while we're using it later
- Updates live loaders code snippets around `error` in `try...catch` block: `error` in the catch statement is inferred as `unknown`, so Typescript complains when using `error.message`. One solution is to use `error: any` if we want to avoid Typescript errors. But HiDeoo suggested a better approach!
- Updates `LiveLoader` type parameters (see below for the why... my explanation is a bit long 😁 )
- Adds missing types/link in types
- Improves (try at least) the differentiation between the `LiveLoader` object and type (based on a previous T&D).
- Removes `renderMarkdown()` mention in `LiveDataEntry.rendered`: this is only available for build-time loaders. This is still in the [RFC Unresolved questions](https://github.com/withastro/roadmap/blob/feat/live-loaders/proposals/0055-live-content-loaders.md#unresolved-questions).

**`astro:content`:**

* Fixes `getLiveCollection()`/`getLiveEntry()` code snippets
* Adds docs for `LiveDataCollectionResult` and `LiveDataEntryResult` (note: this is imported from `astro`, not `astro:content`)

**TODO:** check `astro:content` new descriptions/code snippets, this is more of a draft at the moment.

**Regarding `LiveLoader` filters/type parameters:**

In short: if you use Typescript and have filters, you must provide the types, don't forget them.

Now, the detailed explanation:

I built Astro v6 (instead of using the experimental flag, to be safe) and I used it in a minimal project to check our code snippets. If we want to show valid examples without Typescript errors (well, except for things not implemented like `fetchFromCMS`), the `LiveLoader` type is a bit more complex to handle than what we describe.

When I copy/paste the "Example live loader" code snippet:
* the destructured `filter` in `loadCollection` is inferred as `undefined`, which is fine: no Typescript errors (unless the user has a typed function instead of `fetchFromCMS`).
* the destructured `filter` in `loadEntry` is inferred as `never`, which means Typescript complains about `filter.id` (`Property 'id' does not exist on type 'never'.ts(2339)`).

Which is a bit odd because of:
```ts
export interface LoadEntryContext<TEntryFilter = never> {
    filter: TEntryFilter extends never ? {
        id: string;
    } : TEntryFilter;
```

It should fallback to `{id: string}` but it doesn't seem to work. Typescript still think it's `never`.

So, from my understanding, this means:
* In a JS file, without any type annotations, this is not an issue.
* In a TS file, `LiveLoader<Article>` means the loader doesn't accept any filters and so, we need to use:
  * `LiveLoader<Article, Record<string, any>>` if only `loadEntry()` accepts filters.
  * `LiveLoader<Article, never, Record<string, any>>` if only `loadCollection()` accepts filters.
  * `LiveLoader<Article, Record<string, any>, Record<string, any>>` if both accepts filters.

I also updated the "Error handling in live loaders" code snippets because:
- We can't use `undefined` as type parameter (`Type 'undefined' does not satisfy the constraint 'Record<string, any>'.ts(2344)`).
- If we use `never` instead, we can't show `filter` (ie. the loader doesn't accept any filters)
- If we don't want to provide a type, we'll need `Record<string, any>` or a custom one (e.g. imported alongside `MyData`).

Since the goal is to show error handling, I think we can use `never` here and we don't need to show `filter`.

I think systematically displaying the filters types each time `filter` is used will help people understand that they are required.

- To merge in: #12604
- Suggested label: v6

### DIFF
--- a/src/content/docs/en/reference/modules/astro-content.mdx
+++ b/src/content/docs/en/reference/modules/astro-content.mdx
@@ -507,6 +507,8 @@ const { entries: allProducts } = await getLiveCollection('products');
 ---
 ```
 
+<ReadMore>Learn how to [access live data](/en/guides/content-collections/#accessing-live-data) with guided explanations and example usage.</ReadMore>
+
 #### `error`
 
 <p>
@@ -582,6 +584,8 @@ import { getLiveEntry } from 'astro:content';
 const { entry } = await getLiveEntry('products', Astro.params.id);
 ---
 ```
+
+<ReadMore>Learn how to [access live data](/en/guides/content-collections/#accessing-live-data) with guided explanations and example usage.</ReadMore>
 
 #### `error`
 


src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -1108,9 +1119,19 @@ This section shows the API for defining a [live loader](#live-loaders).
 
 ### The `LiveLoader` object
 
-A generic type to provide [type safety in your loader](/en/guides/content-collections/#type-safety-in-live-loaders). This describes an object and accepts four type parameters to describe, in this order: your data structure, your filters when querying a collection, your filters when querying a single entry, and errors.
+<p>
+
+**Type:** `LiveLoader<TData, TEntryFilter, TCollectionFilter, TError>`
+</p>
+
+A loader function returns an object describing how to fetch data from your source. When using TypeScript, you can use the `LiveLoader` generic type to provide [type safety in your loader](/en/guides/content-collections/#type-safety-in-live-loaders). This type accepts the following type parameters, in this order:
```
sarah11918:
```suggestion
An object returned by a live loader function describing how to fetch data from your source.

Use the `LiveLoader` generic type to provide [type safety in your loader](/en/guides/content-collections/#type-safety-in-live-loaders). This type accepts the following type parameters, in this order:
```
Usually, our first sentence should be a definition/description of what the thing is, so maybe something like above? (Edit as necessary!)

Also, I think it's *very* likely that people building live loaders are using TypeScript 😅 So I think it's OK to just make that assumption and keep the wording more concise. (And, if they're not using TypeScript, then they probably don't care about type safety? So even if it's possible to do this in JavaScript, they wouldn't bother with the type anyway?) (Again, edit if more nuance is needed)

BUT, I'm not loving that this stuff about type seems to "interrupt" talking about the loader itself and its properties? Do we have precedence for something like the following:

> ### The live loader object
>
> An object... with the generic [`LiveLoader` type] (#link-to-somewhere-else-we-define-the-type). It has the following properties:
> #### `name`
> #### `loadCollection()`
> #### `loadEntry()`

Because it seems to me that if I want to know about the loader object, I care more about seeing what properties it is composed of. If I want to know about the type, and it needs a whole paragraph/bulleted list to explain, I'd want to see that when I'm looking up type info? What do you think?

ArmandPhilippot: I don't think there's a precedent for a generic type that accepts so many parameters. And that's why this is a bit tricky to describe both. 😅 
Generally, there is only one, which makes it easier to describe and, most often, removes the need to explain the type parameter, as it is automatically inferred from the given property!
Here, if we don't provide these types, there's no "magic" inference. Users must define them explicitly to have type safety.

> I'm not loving that this stuff about type seems to "interrupt" talking about the loader itself and its properties?

I considered having two different sections (`The Loader object` and `The Loader type`) but I don't think this is better? A subheading to explain the type parameters could make sense. I'm not sure yet...
So, I agree, I'm not sure this is the best way but I think we need something like that, somewhere. I'm trying to figure out how to integrate that description.

At least in my mind, I don't think this necessarily "interrupts talking about the loader": this is how the loader works. Each property depends on those types. When using Typescript, the object depends on the type. They have the same shape/they are not really different.

So, if "it's very likely that people building live loaders are using TypeScript": I'm not sure we should consider them as two different things? Users will need to know how to pass the correct type to each property.

So yes, there's still work to be done. And I'm open to any suggestions for better integrating this description! 😄

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -302,11 +302,19 @@ import { fetchFromCMS } from './cms-client.js';
...
 }
 
-export function articleLoader(config: { apiKey: string }): LiveLoader<Article> {
+interface FetchSingleArticleFilters {
+  id: string;
+}
+
+interface FetchArticlesFilters {
+  author?: string;
+}
```
HiDeoo: Any reason to not go with `EntryFilter` and `CollectionFilter`?

I personally find these names easier to reason with, e.g. `EntryFilter` for the `loadEntry/getLiveEntry` combo, and `CollectionFilter` for the `loadCollection/getLiveCollection` one?

Considering the generic names are also `TEntryFilter` and `TCollectionFilter`, a lot of people would just remove the `T` part and go with that so that would also match that use case.

If we want, it could even be be `ArticleEntryFilter` and `ArticleCollectionFilter`.

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -323,7 +331,7 @@ export function articleLoader(config: { apiKey: string }): LiveLoader<Article> {
             data: article,
           })),
         };
-      } catch (error) {
+      } catch (error: any) {
```
HiDeoo: It's fine to omit the type here, it'll fallback to `unknown`, and if we specifically want to annotate it, we should use `unknown` rather than `any`.

```suggestion
      } catch (error) {
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -351,7 +359,7 @@ export function articleLoader(config: { apiKey: string }): LiveLoader<Article> {
             html: article.htmlContent,
           },
         };
-      } catch (error) {
+      } catch (error: any) {
```
HiDeoo:
```suggestion
      } catch (error) {
```

Same here regarding the type.

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -419,13 +427,16 @@ export class MyLoaderError extends Error {
   constructor(message: string, public code?: string) {
     super(message);
     this.name = 'MyLoaderError';
+    if (code) {
+      this.code = code;
+    }
```
HiDeoo:
```suggestion
```

This shouldn't be needed as it's using [TypeScript parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) so the `public` before the parameter name makes it automatically a class property with the same name and value.

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -531,9 +542,9 @@ Live loaders can provide cache hints to help with response caching. You can use
 ```ts title="my-loader.ts"
 import type { LiveLoader } from "astro/loaders";
 import { loadStoreProduct, loadStoreProducts, getLastModifiedDate } from "./store";
-import type { MyData } from "./types";
+import type { MyData, MyProductFilters, MyProductsFilters } from "./types";
 
-export function myLoader(config): LiveLoader<MyData> {
+export function myLoader(config): LiveLoader<MyData, MyProductFilters, MyProductsFilters> {
```
HiDeoo: Same as before, I personally think that `EntryFilter` and `CollectionFilter` may be better here rather than relying on pluralization, e.g. 1 letter difference, to distinguish between the two.

If we want, it could be `ProductEntryFilter` and `ProductCollectionFilter` (and maybe switch `MyData` to `Product` so we can see they are all related, e.g. one is the data, one is the filter used with `getLiveCollection` and one is the filter used with `getLiveEntry`?

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -1108,9 +1119,19 @@ This section shows the API for defining a [live loader](#live-loaders).
...
+<p>
+
+**Type:** `LiveLoader<TData, TEntryFilter, TCollectionFilter, TError>`
+</p>
+
+A loader function returns an object describing how to fetch data from your source. When using TypeScript, you can use the `LiveLoader` generic type to provide [type safety in your loader](/en/guides/content-collections/#type-safety-in-live-loaders). This type accepts the following type parameters, in this order:
 
-This object contains the following properties:
+- **`TData`** (defaults to `Record<string, unknown>`): The data structure of each entry returned by the loader.
+- **`TEntryFilter`** (defaults to `never`): The filter object type accepted by [`getLiveEntry()`](/en/reference/modules/astro-content/#getliveentry). Use `never` when you don't support filtering single entries.
```
HiDeoo: As we mentioned it's consumed by users with `getLiveEntry()`, should we mention it's received as an argument in `loadEntry`?

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -1108,9 +1119,19 @@ This section shows the API for defining a [live loader](#live-loaders).
...
+- **`TCollectionFilter`** (defaults to `never`): The filter object type accepted by [`getLiveCollection()`](/en/reference/modules/astro-content/#getlivecollection). Use `never` when you don't support filtering collections.
```
HiDeoo: Same question as above for `loadCollection`?

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -1108,9 +1119,19 @@ This section shows the API for defining a [live loader](#live-loaders).
...
+- **`TError`** (defaults to `Error`): A [custom `Error` class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#custom_error_types) for more specific error handling in the loader.
```
HiDeoo: Should we mention that this custom `Error` class can be returned by the loader for more granular error handling?

HiDeoo (commented):
Left a few comments.

I'm personally not bother at all with the generic types position. If the type of an object accept multiple generic types, I definitely want to know about them immediately, even before knowing what the properties of the object are considering such types could impact the properties.

src/content/docs/en/reference/content-loader-reference.mdx
sarah11918: Fair enough! All you say makes sense, and Hideoo agrees, so I'm happy with what you think is best here.

It *is* a tricky case, and I do think we tend to start all API references with a statement that is more of a definition, so that would be my strongest feedback here. "The `LiveLoader` object" is kind of weird as a heading given that normally we'd either use the technical term, or something just in regular words like "The live loader object". So, I think we can give ourselves some leeway here. :smile: 

```suggestion
A live loader function returns an object with [three required live loader properties](#live-loader-properties) that describes how to fetch both single entries and an entire collection from your live data source.

Use the `LiveLoader` generic type to provide [type safety in your loader](/en/guides/content-collections/#type-safety-in-live-loaders). This type accepts the following type parameters, in this order:
```
Maybe something like this, and then a `#### Live loader properties` subheading instead of a sentence saying it accepts the following properties?

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Good call! No reasons, I'm just used to define more specific names to avoid renaming them when importing them in another file, in case of conflicts... But, I agree those names are not terrible (here, and after) and for the docs your first suggestion is probably enough!

I'll update that, thanks.

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Without a type annotation, this means Typescript complains:
Without any, Typescript complains

I think we tend to avoid this to prevent users complaints about these red squiggles. If we use `unknown`, the error remains (and as you said, no need since this is the default type). With `any`, we don't have this issue.

Now, I agree: it's better to use `unknown` than `any`, but this means we would need a condition like `if (error instanceof Error)` before using it... The idea was just to eliminate those Typescript errors for people less comfortable with it.

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Oh, I didn't knew that! Thanks.

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Yeah, the subheading is probably better than the sentence here! I'll update that!

Now, about the wording: technically none of the type parameters are required. But:
* not providing the first one is kinda weird, who doesn't want type safety? 😄 
* because of the Typescript issue I noticed, even `filter.id` can't be used in `loadEntry()` without a type parameter... which makes it kinda required because `getLiveEntry()` requires at least an id as second parameter.

So, yes, I think it's safe to say the first two are required! But, I'm not sure for the third one? I think it's okay if `loadCollection` doesn't accept any filters?

src/content/docs/en/reference/content-loader-reference.mdx
HiDeoo: Oh, I didn't realize that's why you changed the type. But yeah, that's how it should be done in TS with both our `astro/tsconfigs/strict` and `astro/tsconfigs/strictest` presets, e.g. `error instanceof Error ? error.message : String(error)`.

I definitely get the point. I wonder if we have some similar cases in the past where we eliminated some TS practices for people less confortable with it at the potential cost of not-recommended practices.

I guess another option could be to use `new MyLoaderError("Failed to load article", "LOAD_ERROR", { cause: error })` and have our `MyLoaderError` constructor accept a `public options?: ErrorOptions` as 3rd-argument. In this case, we forward whatever was thrown and don't interact with it, but in the end, I'm not sure if it makes it even easier 😅

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Oh nevermind, now I'm addressing feedback and I can see I misread that! 😅

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: I know we have a few `: any` in the tutorial (related to importing files with `import.meta.glob` I think). We might have some of them elsewhere... but I'm not sure.

Now, the tutorial and the reference is not really for the same audience, so it could make sense to use something better here! It's a bit verbose, but yeah I think your first suggestion is fine.
The second is maybe a bit advanced for that section? I mean, the idea is to show a live loader example, not to create a custom error. (my example was from a sandbox to test the APIs, in the docs we use `Error` here, but the Typescript error is the same)

Edit: But, since this is a normal `Error` here, I guess we could just use: `new Error('Failed to load articles', { cause: error })`. That was a good idea after all 😄 
I'll update that.

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -468,3 +468,169 @@ const blog = defineCollection({
...
+<ReadMore>Learn more about [error handling](/en/guides/content-collections/#error-handling) with guided explanations and example usage.</ReadMore>
+
+#### `cacheHint`
+
+<p>
+
+**Type:** <code><a href="/en/reference/content-loader-reference/#cachehint-2">CacheHint</a> | undefined</code>
+</p>
+
+An object providing a hint for how to cache this entry.
```
sarah11918:
```suggestion
An object providing data that can be used to set response headers to inform a caching strategy.
```
I'm not *totally* sure this is the right wording, just whipped it up quickly, but trying to avoid circular "`cacheHint` is a cache hint!" kind of definitions :smile:

src/content/docs/en/reference/modules/astro-content.mdx
```diff
@@ -468,3 +468,169 @@ const blog = defineCollection({
   }),
 });
 ```
+
+## `astro` types
```
sarah11918, line 472: This new section is so good! And so exciting! I am so happy about this!

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -1104,15 +1112,29 @@ If the entry has Markdown content then you can use the [`renderMarkdown()`](#ren
...
 
-A generic type to provide [type safety in your loader](/en/guides/content-collections/#type-safety-in-live-loaders). This describes an object and accepts four type parameters to describe, in this order: your data structure, your filters when querying a collection, your filters when querying a single entry, and errors.
+<p>
 
-This object contains the following properties:
+**Type:** `LiveLoader<TData, TEntryFilter, TCollectionFilter, TError>`
+</p>
 
-#### `name`
+A live loader function returns an object with [three required live loader properties](#live-loader-properties) that describes how to fetch both single entries and an entire collection from your live data source.
```
sarah11918:
```suggestion
A live loader function returns an object with [three required live loader properties](#live-loader-properties). In addition to providing a name for the loader, this object describes how to fetch both single entries and an entire collection from your live data source.
```
Had to get more verbose because of an ambigous subject for "defines" (which could make it sound like it should be "define", depending on how you read it.

This way, at least the number 3 also makes sense, I guess! :sweat_smile:

src/content/docs/en/reference/modules/astro-content.mdx
ArmandPhilippot: Glad you like it! It bothered me a bit because this is not `astro:content`, but yeah I think this makes sense. And, this is better than leaving things undocumented!

We might need similar sections in other modules where we don't show everything importable from `astro` but only the things relevant to the current API! I guess that leaves me with something to do later. 😆

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Sounds good to me! Earlier, I was confused because I was so focused on type parameters that I forgot the object was containing required properties. 😆

ArmandPhilippot:
I pushed one more commit I was about to push when I saw your approval: two read more links for `entries`/`entry` in `astro:content`. I don't think this change your approval, and the wording can always be updated in your PR so I guess I can mark this PR ready for review 😆 and merge it! Thank you both for your reviews ❤️

PR #12812, opened 2025-12-01
csp: improve docs

ematipico:
#### Description (required)

Some users have asked how, with CSP, can support `script-src-*` and `style-src-*`.

Well, the thing is, we already do! However, this information was "hidden", in the sense that we don't explicitly state that `resources` can be used to add the values supported by those directives.

This was due to my lack of knowledge.

This PR expands the docs of `resources`. 

- Suggested label: 
- Related support thread: https://discord.com/channels/830184174198718474/1443253672036536445

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/csp.mdx
+++ b/src/content/docs/en/reference/experimental-flags/csp.mdx
@@ -146,7 +146,7 @@ These properties are added to all pages and **completely override Astro's defaul
 <Since v="5.9.0" />
 </p>
 
-A list of valid sources for the `script-src` and `style-src` directives.
+A list of valid sources for the `script-src` and `style-src` directives, including [values for subclasses](#adding-values-for-subclasses).
 
 The `script-src` and `style-src` directives are handled by Astro by default, and use the `'self'` resource. This means that scripts and styles can only be downloaded by the current host (usually the current website).
 
@@ -187,6 +187,8 @@ After the build, the `<meta>` element will instead apply your sources to the `st
   >
 </head>
 ```
+
+### Adding values for subclasses
 
 You can also use this field to add valid values that belong to [`script-src-attr`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src-attr), [`script-src-elem`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src-elem), [`style-src-attr`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-attr) and [`style-src-elem`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-elem), such as `unsafe-hashes` and `unsafe-inline`.
 


src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -188,6 +188,47 @@ After the build, the `<meta>` element will instead apply your sources to the `st
...
+      scriptDirective: {
+        resources: [
+          "'unsafe-inline'"
+        ]
+      }
+    }
+  }
+});
+```
+
```
sarah11918, line 217: Can we have an "After the build" statement similar to line 177 to separate the code blocks and explain what happens?

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -188,6 +188,47 @@ After the build, the `<meta>` element will instead apply your sources to the `st
 </head>
 ```
 
+You can use this field to add valid values that belong to [`script-src-attr`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src-attr), [`script-src-elem`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src-elem), [`style-src-attr`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-attr) and [`style-src-elem`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-elem), such as `unsafe-hashes` and `unsafe-inline`.
```
sarah11918: If all these things had a collective name and we could have a `#####` section heading for them, that would be perfect here. If not, no worries! I just wondered whether these are a subclass of the main `script-src` and `style-src` that we could identify somehow

```suggestion
You can also use this field to add valid values that belong to [`script-src-attr`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src-attr), [`script-src-elem`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src-elem), [`style-src-attr`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-attr) and [`style-src-elem`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-elem), such as `unsafe-hashes` and `unsafe-inline`.
```

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -188,6 +188,47 @@ After the build, the `<meta>` element will instead apply your sources to the `st
...
+
+For example, you have an external library that adds inline styles to some HTML elements of the page. You can use `unsafe-inline` to tell the browser that those styles are safe to render.
```
```suggestion
For example, if you have an external library that adds scripts or inline styles to some HTML elements of the page, you can add `unsafe-inline` to tell the browser that they are safe to render.
```
Can we combine scripts and styles here in one sentence maybe? (This might need editing to be technically correct)

src/content/docs/en/reference/experimental-flags/csp.mdx
ematipico: They are "sub classes", but I can't come up with a header to group all of them. Suggestions are welcome

src/content/docs/en/reference/experimental-flags/csp.mdx
```diff
@@ -188,6 +188,47 @@ After the build, the `<meta>` element will instead apply your sources to the `st
...
+
+The same analogy is valid for scripts that are added by other libraries.
```
```suggestion
```
If we can just say styles and/or scripts above, then we won't need this.

ematipico:
@sarah11918 I applied your suggestion

PR #12818, opened 2025-12-02
[docs]: Clarify astro:env/server variable evaluation and secret handling

### LINKED ISSUE #12581: `astro:env` secrets may throw on static pages
### 📚 Subject area/topic

astro:env

https://docs.astro.build/en/guides/environment-variables/

See https://github.com/withastro/astro/issues/14504 for context. If you are in such situation:

- You have public and secret server vars
- You do not use secrets on prerendered pages
- You use public server vars on prerendered pages
- You do not provide secrets during the build since they're not technically required
- You do not have `env.validateSecrets: true`

Then an error will be thrown saying that your secrets are missing. That's because as soon as the `astro:env/server` module is imported, even secrets are evaluated. There's nothing we can do about it unfortunately.

Solution is to pass fake values during the build to avoid errors.

_No response_

fkatsuhiro:
#### Description (required)

Added documentation regarding `secret server variable` validation during the build process.
Since `astro:env/server` is evaluated on import, builds can fail if secrets are missing. This PR adds a section explaining this behavior and how to fix it using dummy values.

スクリーンショット 2025-12-03 23 31 59

- Closes #12581

### DIFF
--- a/src/content/docs/en/guides/environment-variables.mdx
+++ b/src/content/docs/en/guides/environment-variables.mdx
@@ -279,7 +279,9 @@ There are three kinds of environment variables, determined by the combination of
    import { API_SECRET } from "astro:env/server";
    ```
 
-   By default, secrets are only validated at runtime. You can enable validating private variables on start by [configuring `validateSecrets: true`](/en/reference/configuration-reference/#envvalidatesecrets).
+   By default, all secrets are validated whenever anything is imported from the `astro:env/server` module. This means, secrets may be validated even when they are not imported. You may need to [pass dummy environment variables](#setting-environment-variables) to satisfy this validation during the build.
+
+   You can also enable validating secrets on start by [configuring `validateSecrets: true`](/en/reference/configuration-reference/#envvalidatesecrets).
 
 :::note
 **Secret client variables** are not supported because there is no safe way to send this data to the client. Therefore, it is not possible to configure both `context: "client"` and `access: "secret"` in your schema.
@@ -319,33 +321,6 @@ envField.enum({
 ```
 
 <ReadMore>For a complete list of validation fields, see the [`envField` API reference](/en/reference/configuration-reference/#envschema).</ReadMore>
-
-### Secrets during the build
-
-**Secret server variables** are validated even during `astro build` because `astro:env/server` is evaluated on import.
-
-If your build process does not have access to these secrets, the build will fail. You can pass dummy values to the build command to satisfy the validation:
-
-<PackageManagerTabs>
- <Fragment slot="npm">
-    ```shell
-    # Example: passing a dummy value to build
-    MY_SECRET_KEY="dummy" npm run build
-    ```
- </Fragment>
- <Fragment slot="pnpm">
-    ```shell
-    # Example: passing a dummy value to build
-    MY_SECRET_KEY="dummy" pnpm build
-    ```
- </Fragment>
-  <Fragment slot="yarn">
-    ```shell
-    # Example: passing a dummy value to build
-    MY_SECRET_KEY="dummy" yarn build
-    ```
- </Fragment>
-</PackageManagerTabs>
 
 ### Retrieving secrets dynamically
 


sarah11918:
Hi @fkatsuhiro. Thank you for grabbing the issue and helping us fix it! Your contribution is well-written and easy to read, but I don't think this information belongs in a "caution" block.

We try to use as few notes (and especially cautions!) in docs as possible. If *everything* is a big note to call your attention, then *nothing* stands out! 😄 (And, it gives us more room to add things like examples, since your proposal *mentions* there is a way to address the problem, but doesn't show an example of how to do it!)

Instead, when we are simply describing "how something works", it should be in regular paragraph text. Sometimes this means a new section heading so that specific information about a subtopic is easy to find and link to. And sometimes, new information is not just "one new paragraph" or section, but means reading the whole section and seeing if this information is well presented when the topic is explained. It involves figuring out how to update existing text to make sure the necessary information is presented in the appropriate place, at the appropriate time.

So, how would you propose we update docs if you were not allowed to use any kind of note/caution? Could you try again without doing so?

fkatsuhiro:
> Hi @fkatsuhiro. Thank you for grabbing the issue and helping us fix it! Your contribution is well-written and easy to read, but I don't think this information belongs in a "caution" block.
> 
> We try to use as few notes (and especially cautions!) in docs as possible. If _everything_ is a big note to call your attention, then _nothing_ stands out! 😄 (And, it gives us more room to add things like examples, since your proposal _mentions_ there is a way to address the problem, but doesn't show and example of how to do it!)
> 
> Instead, when we are simply describing "how something works", it should be in regular paragraph text. Sometimes this means a new section heading so that specific information about a subtopic is easy to find and link to. And sometimes, new information is not just "one new paragraph" or section, but means reading the whole section and seeing if this information is well presented when the topic is explained. It involves figuring out how to update existing text to make sure the necessary information is presented in the appropriate place, at the appropriate time.
> 
> So, how would you propose we update docs if you were not allowed to use any kind of note/caution? Could you try again without doing so?

Thank you for your reply ! 
Sure, I will try this again.

fkatsuhiro:
@sarah11918 san
Thanks for the feedback! I've updated this PR.

1.  Removed the caution block and added a new section "Secrets during the build". 
2. Add a sample commands.

Could you take another look when you have time?

sarah11918:
Thank you, I absolutely will take a look! 🙌 

But, please remove all the translations and only submit changes to docs in English. We have a translation procedure for handling the other languages.

sarah11918:
@florian-lefebvre Can you review the content and placement here?

One thing I noticed is that earlier on the page we say this ("only validated at runtime"), which seems to contradict the content here (unless there is a further nuance not expressed)? Can you provide clarity?

image

fkatsuhiro:
> Thank you, I absolutely will take a look! 🙌
> 
> But, please remove all the translations and only submit changes to docs in English. We have a translation procedure for handling the other languages.

Thank you for letting me know about the translation procedure! 
I've removed all the non-English changes and kept only the English docs.

florian-lefebvre:
Mmmh tricky. I suppose it depends what we mean by `runtime`. When I wrote this, I think I meant "when it is imported" (VS when you enable `validateSecrets`, which happens earlier in a vite plugin).

So I'm not sure a new heading is necessary, adding it below the "secret server variables" bullet point might make more sense (and then we can discuss editing the content to fit better). Wdyt Sarah?

fkatsuhiro:
Thanks for your suggestion! 
That makes sense to keep it contextually closer. I'll wait for Sarah's input, and if you both agree, I'll update the PR to move the content inside the bullet point.

florian-lefebvre (approved):
T&D ftw!

Fryuni:
We were there when the secrets were discussed... At T&D!!!

JusticeMatthew:
I struggled to understand in Talking & Docing, but I was there

sarah11918:
We looked at this one in painful detail on Talking and Doc'ing because we thought we could work this in to our explanation of how secret server variables work. I think we did it!

PR #12824, opened 2025-12-03
[next]: content loaders breaking changes

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/content-loader-reference.mdx
+++ b/src/content/docs/en/reference/content-loader-reference.mdx
@@ -687,7 +687,7 @@ When you only need to provide a static schema, provide a Zod validation object u
 
 If present, it will be overridden by any Zod `schema` defined for the collection in the `src/content.config.ts` file.
 
-The returned `types` contents will be written to a TypeScript file, and must export a `Entry` type or interface:
+The returned `types` contents will be written to a TypeScript file, and must export an `Entry` type or interface:
 
 ```ts title="src/feed-loader.ts" ins={27-35}
 import type { Loader } from 'astro/loaders';


src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -218,7 +218,7 @@ export function myLoader(options: { url: string, apiKey: string }): Loader {
     schema: z.object({
       // ...
     })
-  };
+  } satisfies Loader;
```
florian-lefebvre, line 223: I updated examples to use `satisfies` but I think it could be worth emphasing somewhere that this is actually required (technically not but it's worth showing only one way of doing things). Thing is: I don't know where, ideas?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,45 @@ export const server = {
...
+If you are building a content loader and using the function signature of `schema`, remove it and replace it with the new `createSchema()` property:
+
+```ts del={11} ins={2,12-22}
+import type { Loader } from 'astro/loaders'
+import { createTypeAlias, zodToTs } from 'zod-to-ts'
+import { getSchemaFromApi } from './utils'
+
+function myLoader() {
+  return {
+    name: 'my-loader,
```
```suggestion
    name: 'my-loader',
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,45 @@ export const server = {
...
+    async load(context) {
+      // ...
+    },
+    schema: async () => await getSchemaFromApi(),
+    async createSchema() {
+      const schema = await getSchemaFromApi()
+      const identifier = 'Collection'
```
```suggestion
      const identifier = 'Entry'
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1020,36 @@ export function getStaticPaths() {
...
+
+If you are building a content loader, make you use [TypeScript' `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator):
+
+```ts del={3,11} ins={4,12}
+import type { Loader } from 'astro/loaders'
+
+function myLoader(): Loader {
+function myLoader() {
+  return {
+    name: 'my-loader,
```
```suggestion
    name: 'my-loader',
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +653,65 @@ An async function that is called at build time to load data and update the store
...
+
+**Type**: `() => Promise<{ schema: ZodSchema; types: string }>`
+<Since v="6.0.0" />
+</p>
+
+An optional async function that returns an object containing a [Zod schema](/en/guides/content-collections/#defining-datatypes-with-zod) and types. It is used to dynamically generate the schema based on the configuration options or by introspecting an API.
 
 If present, it will be overridden by any Zod `schema` defined for the collection in the `src/content.config.ts` file.
 
+The returned `types` contents will be written to a TypeScript file, and must export a `Collection` type or interface:
```
florian-lefebvre:
```suggestion
The returned `types` contents will be written to a TypeScript file, and must export a `Entry` type or interface:
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +653,65 @@ An async function that is called at build time to load data and update the store
...
 <p>
 
-**Type**: `ZodSchema | Promise<ZodSchema> | (() => ZodSchema | Promise<ZodSchema>)`
+**Type**: `ZodSchema`
 <Since v="5.0.0" />
 </p>
 
 An optional [Zod schema](/en/guides/content-collections/#defining-datatypes-with-zod) that defines the shape of the entries. It is used to both validate the data and also to generate TypeScript types for the collection.
 
-If a function is provided, it will be called at build time before `load()` to generate the schema. You can use this to dynamically generate the schema based on the configuration options or by introspecting an API.
```
ArmandPhilippot, line 660: I wonder if we should keep a similar sentence to inform users they should use either `schema` or `createSchema`, not both at the same time.

Maybe something like:
> When you need to dynamically generate the schema based on configuration options or by introspecting an API, use [`createSchema()`](#createschema) instead.

And, add something similar to `createSchema()`?

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +653,65 @@ An async function that is called at build time to load data and update the store
...
+          id,
+          data: item,
+        });
+        store.set({
+          id,
+          data,
+        });
+      }
+    },
+    generateSchema: async () => {
```
ArmandPhilippot:
```suggestion
    createSchema: async () => {
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +653,65 @@ An async function that is called at build time to load data and update the store
...
+      const schema = await getSchema();
+      const types = await getTypes();
```
ArmandPhilippot, line 721: I wonder if there is a way to make more obvious in the code snippet that `types` requires a type or interface named `Entry`. I thought of:
```suggestion
      const types = await getEntryInterface();
```
But I'm not sure if this is the best...
Does `types` can contain other types or only `Entry` is valid?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,45 @@ export const server = {
 
 <ReadMore>Learn more about [rewrites](/en/guides/routing/#rewrites).</ReadMore>
 
+### Removed: schema function signature (Content Loader API)
+
+In Astro 5.x, a content loader could define a schema as a function. This is useful to dynamically generate the schema based on the configuration options or by introspecting an API.
```
```suggestion
In Astro 5.x, a content loader could choose to define a schema as a function instead of defining a Zod schema object for validation. This is useful to dynamically generate the schema based on the configuration options or by introspecting an API.
```
Since defining a schema as a function probably isn't the most common use, just wanted something so that people could distinguish from what they are probably doing (and therefore, to give them a heads up about when this might not apply to them).

Would something like this work?

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Probably in [The `Loader` object](https://deploy-preview-12824--astro-docs-2.netlify.app/en/reference/content-loader-reference/#the-loader-object), we did something similar for [The `LiveLoader` object](https://deploy-preview-12824--astro-docs-2.netlify.app/en/reference/content-loader-reference/#the-liveloader-object). But I'm not quite sure about the wording at the moment.

I know using `AnyType` as return type or using `satisfies AnyType` doesn't produce the same things in Typescript, but is this something required for inference? If someone use `Loader` as return type does this break anything? Or this is really only for consistency?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,45 @@ export const server = {
...
+
+Astro 6.0 removes this signature and introduces a new `createSchema()` property as a replacement. Providing a schema function will log a warning message but this will throw in a future major version.
```
```suggestion
Astro 6.0 removes this signature and introduces a new `createSchema()` property as a replacement for those who still want to dynamically define a schema in their content loader. 

Providing a schema function in the old way will log a warning message that the loader's schema is being ignored, but otherwise the loader will continue to work as if no schema had been provided. In a future major version, loaders that provide a schema function will throw an error and cannot be used.
```
I think this is what's happening?

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,45 @@ export const server = {
...
+
+#### What should I do?
+
+If you are building a content loader and using the function signature of `schema`, remove it and replace it with the new `createSchema()` property:
```
```suggestion
If you are building a content loader and using a function to dynamically return a collection `schema` property, remove it and replace it with the new `createSchema()` property:
```

src/content/docs/en/reference/content-loader-reference.mdx
florian-lefebvre: Yeah using `satisfies` is mandatory in this case for type inference

src/content/docs/en/reference/content-loader-reference.mdx
florian-lefebvre: Good idea, I'll add that!

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,45 @@ export const server = {
...
+import { getSchemaFromApi } from './utils'
+
+function myLoader() {
+  return {
+    name: 'my-loader',
+    async load(context) {
+      // ...
+    },
+    schema: async () => await getSchemaFromApi(),
+    async createSchema() {
```
sarah11918: This example looks more complicated than the one in the `createSchema` reference docs. I'm not sure I'd understand looking at this whether e.g. all these things are absolutely needed for `createSchema`, then if I go to docs, that example is quite minimal. If anything, THIS should be the more minimal example here.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1020,36 @@ export function getStaticPaths() {
 
 <ReadMore>Learn more about [dynamic SSG routes with `getStaticPaths()`](/en/guides/routing/#static-ssg-mode).</ReadMore>
 
+### Changed: schema types are inferred instead of generated (Content Loader API)
+
+<SourcePR number="14759" title="feat: loader.createSchema()" />
+
+In Astro 5.x, if a content collection was using a content loader defining a static schema without a user schema overriding it, Astro would generate types using `zod-to-ts`.
```
```suggestion
In Astro 5.x, the types for content collections were generated using `zod-to-ts` when provided by a content loader and not defined by a user-provided schema.
```
This is not as good as it was! 😭 But the idea here is really about `zod-to-ts`, and in fact nothing has changed re: a user schema overrides anyway. So I *think* this can be distilled to "when we got your schema from your loader, `zod-to-ts` was used to generate types.

So, maybe something like this?

src/content/docs/en/reference/content-loader-reference.mdx
florian-lefebvre: The only constraint is to export this `Entry type`, that's it! It can import modules if needed, eg:

```ts
import { Recipe } from '@my/sdk'

// or export type Entry = {
export interface Entry {
	title: string;
	recipe: Recipe;
}
```

As long as these types can resolve (eg. the user has the package), should be good!

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1020,36 @@ export function getStaticPaths() {
...
+
+Astro 6.0 removes this behavior: types are not generated anymore but instead inferred.
```
```suggestion
Astro 6.0 removes this behavior: types are no longer generated using `zod-to-ts`. Instead, types are inferred.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1020,36 @@ export function getStaticPaths() {
...
+
+#### What should I do?
+
+If you are building a content loader, make you use [TypeScript' `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator):
```
```suggestion
If you are providing a `schema` in a content loader, you must use the [TypeScript' `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator):
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1020,36 @@ export function getStaticPaths() {
...
+    async load(context) {
+      // ...
+    },
+    schema: z.object({/* ... */})
+  }
+  } satisfies Loader
+}
+```
+
+<ReadMore>Learn more about the [object loader API](/en/reference/content-loader-reference/#object-loader-api).</ReadMore>
```
```suggestion
<ReadMore>Learn more about [defining loader schema types](/en/reference/content-loader-reference/#the-loader-object).</ReadMore>
```
Maybe this is a good link?

Updated to the loader object link. What text do you think it should be?

src/content/docs/en/reference/content-loader-reference.mdx
sarah11918: I will let Armand handle this!

ArmandPhilippot: We could maybe use a structure similar to that of `LiveLoader` to be able to both describe how to use the type and to have a parallel structure. Something like:
```md
### The Loader object

<p>

**Type:** `Loader`
</p>

A loader function returns an object with [two required properties](#loader-properties). In addition to providing a name for the loader, this object describes how to fetch the collection data.

Use the [Typescript `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) instead of a return type annotation to provide type safety inside your loader object and to preserve type inference when your loader is used in a collection.

#### Loader properties
```

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Hmm, I can't think of a nice solution here. The explicit `Entry` was nice in your upgrade example but for the reference, we probably don't want to show unnecessary dependencies.
Not sure adding a comment above is better. I mean some people look at the code snippet instead of reading, so I'm not sure they would read the comment. 😄 

So, yeah, unless you know a way to show that `types` contains `Entry`, maybe that's fine and we can try to improve that later if people report this isn't clear enough (even the description say what they should do)?

src/content/docs/en/guides/upgrade-to/v6.mdx
ArmandPhilippot: Ah, I thought the opposite. 😆 Well, not necessarily because of the complexity but because we show `Entry` as requirement. So, reading the code snippet makes it obvious to me while this isn't the case in the reference.

But, you're right that if using `zod-to-ts` is not a requirement, maybe we shouldn't show that? I don't know.
If I understand correctly, that's what Astro used to do. Now it's up to the user to do it, and `zod-to-ts` is a handy library for that.

sarah11918: I think my point is, I have no idea from reading this example whether I am *supposed* to add all of that (because that is how `createSchema` works, and it relies on you doing all those things) or what of this might be from some made up example that doesn't apply to me. And glancing (didn't review thoroughly) at the reference example, there's none of this there.

So, maybe since this example is the upgrade guide help, we could keep this example but introduce it like:

> Since Astro no longer uses `zod-to-ts` internally, you can reproduce your existing functionality by using it yourself with `createSchema`

Or something like that? To make it clear that this is how you can upgrade what you already have with this and in theory, it should be doing the same thing (just you using `zod-to-ts` instead of Astro doing it for you?)

src/content/docs/en/reference/content-loader-reference.mdx
florian-lefebvre: Sounds good to me! I'll let you make a suggestion through GitHub's UI

src/content/docs/en/reference/content-loader-reference.mdx
florian-lefebvre: Maybe we could improve the example by doing this:
```js
return {
	schema,
	types: `export type Entry = ${types}`
}
```
Just to make it clearer

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: Yeah my thought was to show the closest equivalent to what's in v5. So it involves dealing with `zod-to-ts` APIs, which are not pretty

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: FYI

Image

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: I'm not sure. The point is not about the schema but really `satisfies`. I think we'll be able to tell where to link once https://github.com/withastro/docs/pull/12824#discussion_r2585530554 is resolved

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Hopefully suggesting things on unchanged lines works fine, I had a few bugs with the GitHub new experience 😅 
```suggestion

<p>

**Type:** `Loader`
</p>

A loader function returns an object with [two required properties](#loader-properties). In addition to providing a name for the loader, this object describes how to fetch the collection data.

Use the [Typescript `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) instead of a return type annotation to provide type safety inside your loader object and to preserve type inference when your loader is used in a collection.

#### Loader properties
```
(the diff doesn't show that, but `l630` should contain "The loader object has the following properties:", so this suggestion replaces it.)

src/content/docs/en/reference/content-loader-reference.mdx
ArmandPhilippot: Sounds good to me!

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: It has been resolved, we should link to "the Loader object" heading instead

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: Gah, GitHub! I made this sound so nice 😅 Let me see if I can figure it out again

sarah11918: Updated my suggestion above!

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +663,69 @@ An async function that is called at build time to load data and update the store
...
+          id,
+          data,
+        });
+      }
+    },
+    createSchema: async () => {
+      const schema = await getSchema();
+      const types = await getTypes();
+
+      return { schema, types };
```
florian-lefebvre:
```suggestion
      return {
      	schema,
      	types: `export type Entry = ${types}`,
      };
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +663,69 @@ An async function that is called at build time to load data and update the store
...
+
+An optional async function that returns an object containing a [Zod schema](/en/guides/content-collections/#defining-datatypes-with-zod) and types. It is used to dynamically generate the schema based on the configuration options or by introspecting an API.
+
+When you need to provide a static schema, use [`schema`](#schema) instead.
 
 If present, it will be overridden by any Zod `schema` defined for the collection in the `src/content.config.ts` file.
 
+The returned `types` contents will be written to a TypeScript file, and must export a `Entry` type or interface:
+
+```ts title="src/feed-loader.ts" ins={27-32}
```
florian-lefebvre:
```suggestion
```ts title="src/feed-loader.ts" ins={27-35}
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,47 @@ export const server = {
...
+
+In Astro 5.x, a content loader could choose to define a schema as a function instead of defining a Zod schema object for validation. This is useful to dynamically generate the schema based on the configuration options or by introspecting an API.
+
+Astro 6.0 removes this signature and introduces a new `createSchema()` property as a replacement for those who still want to dynamically define a schema in their content loader. 
+
+Providing a schema function in the old way will log a warning message that the loader's schema is being ignored, but otherwise the loader will continue to work as if no schema had been provided. In a future major version, loaders that provide a schema function will throw an error and cannot be used.
+
+#### What should I do?
+
+If you are building a content loader and using a function to dynamically return a collection `schema` property, remove it and replace it with the new `createSchema()` property:
```
```suggestion
If you are building a content loader and using a function to dynamically return a collection `schema` property, you must remove your existing function and use the new `createSchema()` property to define your schema instead.

For example, you can reproduce Astro's previous behavior by using `zod-to-ts` directly with `createSchema()` and any previous function logic:
```
What about this?

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: going to resolve *this* comment because I made another one above!

src/content/docs/en/guides/upgrade-to/v6.mdx
sarah11918: updated my suggestion above, but is the text for it right?

sarah11918:
REMINDER that I haven't reviewed the actual `createSchema()` entry yet! But it looks like all the previous comments are quickly getting resolved, so I'll do that now!

src/content/docs/en/guides/upgrade-to/v6.mdx
florian-lefebvre: It's not really about "definining schema types", more about "loader type to get proper schema types". But that's too long

sarah11918: what about "more about loader schema types" ??

(it can be just in the general ballpark... like, you WERE just going to link to the whole darn API in the original!)

florian-lefebvre: lol yeah works for me!

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -625,7 +627,17 @@ This section shows the API for defining an [object loader](#object-loaders).
...
-The loader object has the following properties:
+
+<p>
+
+**Type:** `Loader`
+</p>
+
+A loader function returns an object with [two required properties](#loader-properties). In addition to providing a name for the loader, this object describes how to fetch the collection data.
+
+Use the [Typescript `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) instead of a return type annotation to provide type safety inside your loader object and to preserve type inference when your loader is used in a collection.
```
sarah11918:
```suggestion
Optionally, you can return a third property defining a schema to validate your collection entries. Use the [Typescript `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) instead of a return type annotation to provide type safety inside your loader object and to preserve type inference when your loader is used in a collection.
```
I think since we've said there are two required properties, but you can optionally provide a schema through a third property, this is a nice intro to the TypeScript info!

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +663,72 @@ An async function that is called at build time to load data and update the store
...
 
-**Type**: `ZodSchema | Promise<ZodSchema> | (() => ZodSchema | Promise<ZodSchema>)`
+**Type**: `ZodSchema`
 <Since v="5.0.0" />
 </p>
 
 An optional [Zod schema](/en/guides/content-collections/#defining-datatypes-with-zod) that defines the shape of the entries. It is used to both validate the data and also to generate TypeScript types for the collection.
 
-If a function is provided, it will be called at build time before `load()` to generate the schema. You can use this to dynamically generate the schema based on the configuration options or by introspecting an API.
+When you need to dynamically generate the schema based on configuration options or by introspecting an API, use [`createSchema()`](#createschema) instead.
```
sarah11918:
```suggestion
When you need to dynamically generate the schema at build time based on configuration options or by introspecting an API, use [`createSchema()`](#createschema) instead.
```
I think the "build time" part is still true, right? I think this is important so people don't get the wrong idea that dynamic === live/fresh/updating

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +663,72 @@ An async function that is called at build time to load data and update the store
...
 
 If present, it will be overridden by any Zod `schema` defined for the collection in the `src/content.config.ts` file.
 
+#### `createSchema()`
```
sarah11918:
```suggestion
##### `createSchema()`
```
This should be a subheading of 	`schema` right?

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +663,72 @@ An async function that is called at build time to load data and update the store
...
+
+<p>
+
+**Type**: `() => Promise<{ schema: ZodSchema; types: string }>`
+<Since v="6.0.0" />
+</p>
+
+An optional async function that returns an object containing a [Zod schema](/en/guides/content-collections/#defining-datatypes-with-zod) and types. It is used to dynamically generate the schema based on the configuration options or by introspecting an API.
```
sarah11918:
```suggestion
An optional async function that returns an object containing a [Zod schema](/en/guides/content-collections/#defining-datatypes-with-zod) and types. It is used to dynamically generate the schema at build time based on the configuration options or by introspecting an API.
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +663,72 @@ An async function that is called at build time to load data and update the store
...
+
+When you need to provide a static schema, use [`schema`](#schema) instead.
```
sarah11918:
```suggestion
When you only need to provide a static schema, provide a Zod validation object using [`schema`](#schema) instead.
```
I want to make this sound more like the normal/easy case.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,49 @@ export const server = {
...
+import { getSchemaFromApi } from './utils'
+
+function myLoader() {
+  return {
+    name: 'my-loader',
+    async load(context) {
+      // ...
+    },
+    schema: async () => await getSchemaFromApi(),
+    async createSchema() {
```
```suggestion
    createSchema: async () => {
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -604,6 +604,49 @@ export const server = {
...
+
+```ts del={11} ins={2,12-22}
+import type { Loader } from 'astro/loaders'
+import { createTypeAlias, zodToTs } from 'zod-to-ts'
+import { getSchemaFromApi } from './utils'
+
+function myLoader() {
+  return {
+    name: 'my-loader',
+    async load(context) {
```
```suggestion
    load: async (context) => {
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -981,6 +1024,36 @@ export function getStaticPaths() {
...
+If you are providing a `schema` in a content loader, you must use the [TypeScript' `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator):
+
+```ts del={3,11} ins={4,12}
+import type { Loader } from 'astro/loaders'
+
+function myLoader(): Loader {
+function myLoader() {
+  return {
+    name: 'my-loader',
+    async load(context) {
```
```suggestion
    load: async (context) => {
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -651,16 +663,72 @@ An async function that is called at build time to load data and update the store
...
+<Since v="6.0.0" />
+</p>
+
+An optional async function that returns an object containing a [Zod schema](/en/guides/content-collections/#defining-datatypes-with-zod) and types. It is used to dynamically generate the schema at build time based on the configuration options or by introspecting an API.
+
+When you only need to provide a static schema, provide a Zod validation object using [`schema`](#schema) instead.
+
+If present, it will be overridden by any Zod `schema` defined for the collection in the `src/content.config.ts` file.
+
+The returned `types` contents will be written to a TypeScript file, and must export a `Entry` type or interface:
```
sarah11918:
```suggestion
The returned `types` contents will be written to a TypeScript file, and must export an `Entry` type or interface:
```

PR #12828, opened 2025-12-03
Add docs for `astro/zod`

### LINKED ISSUE #12814: Document `astro/zod`
### 📚 Subject area/topic

Reference?

I'm not sure

We need to document `astro/zod` as other zod things are deprecated in v6

_No response_

ArmandPhilippot:
#### Description (required)

* Adds a new page to document `astro/zod` module.
* Replaces the Zod cheatsheet from the Content Collections guide with a link to the new reference page
* Replaces a few links to Zod GitHub/website with links to our reference page

**Context:** in v6, the other exports (e.g. from `astro:content`) will be removed and only `astro/zod` will be available to import Zod. So, I think it makes sense to document this module now.

**Things to think about:**
1. Do we want `astro/zod` in "Runtime API"?
  This sounded okay to me to add this module alongside the others modules (even in this is a `/` module, not a virtual one) but because this is a re-export maybe not and maybe "Other development APIs" is more appropriate?
2. How much do we want to document?
  In addition to re-exporting `z`, this also re-export `z` flattened... which contains a lot of things, so showing everything is probably too verbose and it's not our role to document Zod APIs. But, we could be a bit more verbose for things commonly used in Astro to link from our other pages.

- Closes #12814


sarah11918:
Before I think too much about this, does this mean we can/should move some of our Zod content OUT of the content collections guide and link here instead?

ArmandPhilippot:
Yes, I think so. For the "cheatsheet" I actually took inspiration on what we currently have in the content collections guide. So, I think the content collections cheatsheet is no longer required and we could link to that section.

I didn't add that to "Things to consider" (I will), but, yes, this was also something I thought about. Maybe we could link there in some places.

For Actions, this might be tricky though because there are [limitations related to actions](https://docs.astro.build/en/guides/actions/#using-validators-with-form-inputs). But another thing we could consider is adding sections like "Using zod with FEATURE" in that page to explain the limitations and link to that section from the related guide/reference.

I considered that for the `ZodType` we show in different APIs to explain what it is (through a link)... but I'm not sure. The type is complex, this is not our role to document it, and I'm not sure this would help more than just saying `ZodType`.

And no rush, we could eventually talk about this in T&D if there is nothing else. 😄

sarah11918:
That's true! We have a Talking and Doc'ing in... 14.5 hours! 😆

ArmandPhilippot:
I made a few updates to the page to develop a bit why this module is useful and how Zod works. This gives us a few headings to linked to in our current docs. And I replaces a few links to Zod with a link to the reference page instead.

Usually, we use the import name as heading, but I wonder if in this case we shouldn't be more verbose and replace `z` with `z utility`. Looking at the table of contents, this single letter is a bit awkward:

Table of Contents of the `astro/zod` page

While checking the Zod mentions in our docs to replace the links, I also found a [~2-3 years old commented docs](https://github.com/withastro/docs/blob/da07c114ca5164fba2299426a418a3c5314445e0/src/content/docs/en/recipes/build-forms-api.mdx?plain=1#L432-L456) in a recipe... If we don't use it, maybe it's time to remove it? 😅 This has been merged as is in https://github.com/withastro/docs/pull/2795

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -241,41 +241,11 @@ Astro uses [Zod](https://github.com/colinhacks/zod) to power its content schemas
 
 To use Zod in Astro, import the `z` utility from `"astro:content"`. This is a re-export of the Zod library, and it supports all of the features of Zod.
```
```suggestion
To use Zod in Astro, import the `z` utility from `"astro/zod"`. This is a re-export of the Zod library, and it supports all of the features of Zod. 
```
Should this be this?

ArmandPhilippot: > As we discussed with the core team today, we can probably go ahead and update all the code examples now

Ah thanks, I missed that part because of my Wifi signal. 🤖 

So, yes, if we replace all the imports we can say `astro/zod` here! I'll check the other occurrences.

src/content/docs/en/guides/content-collections.mdx
```diff
@@ -208,11 +212,12 @@ Schemas also power Astro's automatic TypeScript typings for your content. When y
...
 :::
 
-Every frontmatter or data property of your collection entries must be defined using a Zod data type:
+Every frontmatter or data property of your collection entries must be defined using a [Zod data type](/en/reference/modules/astro-zod/#common-data-type-validators):
 
-```ts title="src/content.config.ts" {6-11,15-19}
-import { defineCollection, z } from 'astro:content';
+```ts title="src/content.config.ts" {7-12,16-20}
+import { defineCollection } from 'astro:content';
 import { glob, file } from 'astro/loaders'; // Not available with legacy API
```
```suggestion
import { glob, file } from 'astro/loaders';
```
This can probably go, even now? (For sure we should make sure this is not in v6 (checked and yes, it's gone!), but I think it's safe to remove now.)

PR #12832, opened 2025-12-04
[i18nIgnore] Fix broken links after renaming `google-firebase`

sarah11918:
#### Description (required)

Updates breaking links since the `google-firebase` deploy guide was renamed to `firebase`


anaxite:
Came here from Talking & Doc'ing

PR #12834, opened 2025-12-04
chore: Reduce memory usage during link checking

Fryuni:
#### Description

Refactor of the `HtmlPage` class used in link checking to avoid keeping references to the entire DOM tree and HTML source after constructing the page from source.

Measured memory usage, average of 5 runs:
- Prior to these changes: 9.7 GiB
- Removing just the DOM tree references after constructor (first commit): 2.1 GiB
- Reallocating string slices into independent strings (second commit): 300 MiB

### DIFF
--- a/scripts/lib/linkcheck/base/page.ts
+++ b/scripts/lib/linkcheck/base/page.ts
@@ -50,12 +50,11 @@ export class HtmlPage {
 	 */
 	readonly isRedirect: boolean;
 	/**
-	 * Indicator whether the page has a main content block.
+	 * Indicates whether the page contains content.
 	 *
-	 * Prefers the first `<article>` element, with a fallback to `<body>` if no article was found,
-	 * and finally set to `false` if the page even doesn't have a body (a partial or invalid page).
+	 * Set to `false` if the page even doesn't have a `<body>` element (a partial or invalid page).
 	 */
-	readonly hasMainContent: boolean;
+	readonly hasContent: boolean;
 	/**
 	 * The language of the page's main content.
 	 *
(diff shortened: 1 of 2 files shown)


scripts/lib/linkcheck/checks/good-link-label.ts
delucis: Easy lint failure fix here:

```suggestion
```

trueberryless:
Respect 😎 

image

scripts/lib/linkcheck/base/page.ts
```diff
@@ -99,22 +114,24 @@ export class HtmlPage {
...
 
 		// Attempt to find the page's main content element
-		this.mainContent =
-			this.findFirst((el) => el.tagName.toLowerCase() === 'main') ||
-			this.findFirst((el) => el.tagName.toLowerCase() === 'body');
+		const mainContent =
+			parser.findFirst((el) => el.tagName.toLowerCase() === 'main') ||
+			parser.findFirst((el) => el.tagName.toLowerCase() === 'body');
+
+		this.hasMainContent = Boolean(mainContent);
```
delucis: Should we actually call this something else? Given this also includes `<body>` it’s more like `hasContent` I guess rather than `hasMainContent`?

scripts/lib/linkcheck/base/page.ts
```diff
@@ -46,12 +50,12 @@ export class HtmlPage {
 	 */
 	readonly isRedirect: boolean;
 	/**
-	 * The element containing the page's main content.
+	 * Indicator whether the page has a main content block.
 	 *
 	 * Prefers the first `<article>` element, with a fallback to `<body>` if no article was found,
-	 * and finally `null` if the page even doesn't have a body.
+	 * and finally set to `false` if the page even doesn't have a body (a partial or invalid page).
```
delucis: So then these docs would be:

```suggestion
	 * Indicates whether the page contains content.
	 *
	 * Set to `false` if the page doesn't even have a `<body>` element (a partial or invalid page).
```

scripts/lib/linkcheck/base/page.ts
Fryuni: Seems fair. I kept all the names and semantics that were already there, just changed what was need for the optimizations.

But we can totally change the name to make things clearer.

sarah11918:
Thanks for tackling this!

Fryuni:
!coauthor

PR #12835, opened 2025-12-04
i18n(de): Replace straight quote with smart quote in sidebar navigation

randomguy-2650:
#### Description

This PR replaces the straight quote in “Los geht's” with the smart quote (“Los geht’s”). SmartyPants does not apply here, so it would make sense to basically do “manual SmartyPants formatting” in this file.

#### Related labels

- Suggested label: i18n


yanthomasdev (approved):
LGTM

PR #12836, opened 2025-12-04
Address v6 link errors

sarah11918:
This PR addresses the links failures that were not previously caught when the link checker ran out of memory.  Ooops.

(Yes, some of these are error messages. We are updating error messages with changed links in the v6 branch, and will eventually reconcile with updating the Astro repo source.)


ArmandPhilippot (approved):
Wow, that's a lot of links that haven't been detected before. 😅 LGTM!

PR #12842, opened 2025-12-05
[i18nIgnore] update CMS nav component logic

sarah11918:
This PR updates our CMS nav component to match the logic of the recently updated deploy guide nav component. 

It is an analogous update to the  backend nav component done in  https://github.com/withastro/docs/pull/12759

Notably:
- updates the CMS navigation component to use a `logo` property for grabbing the logo and the `sidebar.label` property for generating the name under the icon
- updates the CMS guide schema to require a `logo`
- updates each guide individually (including translations) to replace `service` with `logo`, ensuring the value matches the value in `logos.ts` and checking that `sidebar.label` exists.

While I was there, I alphabetized `logos.ts` to make things easier going forward.

Note: no filenames changed, so no redirects needed

Future Consideration:

- I don't think we actually use logic depending on `stub` anymore for any of these guides? If not, then a followup PR could remove the `stub` property from every guide, of every kind.

### DIFF
--- a/src/components/CMSGuidesNav.astro
+++ b/src/components/CMSGuidesNav.astro
@@ -19,5 +19,5 @@ const links = enPages
 ---
 
 <section>
-	<CardsNav links={links} />
+	<CardsNav minimal links={links} />
 </section>


sarah11918:
> You mean `stub` not `slug`, right?

D'oh, I *do* mean `stub`! Once upon a time, we displayed a different layout if a guide was considered a "stub" -- it had a big note that said, "This is just a stub! You can contribute to it!"

image

I think we care a lot less about that now 😅

PR #12843, opened 2025-12-05
Add docs about vite modes and wrangler environments

wuzzeb:
#### Description (required)

Add documentation about configuring multiple astro modes/wrangler environments.
This came out of the discussion in https://github.com/withastro/astro/issues/14540


sarah11918:
Hi! For a PR of this size, we normally expect a docs issue to be filed so that you can properly present what the concern is with the existing docs and we can all work together to make sure we're creating content that achieves that goal.

I can see in the discussion that you were asked to create one. Please do so before creating a PR! That will allow you to fill in the template and provide the context we need from a docs perspective to be able to move forward.

I will close this PR (at least for now) so we can follow our normal procedure and start with the issue!

PR #12844, opened 2025-12-05
[i18nIgnore] Update media guide nav component logic

sarah11918:
This PR updates our media nav component to match the logic of the recently updated deploy guide nav component. 

It is an analogous update to the backend nav component done in  https://github.com/withastro/docs/pull/12759

Notably:
- updates the media navigation component to use a `logo` property for grabbing the logo and the `sidebar.label` property for generating the name under the icon
- updates the media guide schema to require a `logo`
- updates each guide individually (including translations) to replace `service` with `logo`, ensuring the value matches the value in `logos.ts` and checking that `sidebar.label` exists.

While I was updating the schema, I grouped all the third-party service schemas that work this way together. 

Note: Other nav components (e.g. integrations, migration guides) rely on the logos too, but their pages and schema are different enough (and, these pages aren't added by community members) that I don't think we need to worry about updating any more of these nav components.

Future work:

This PR completes updating all our third-party guide logic, so I can now update the contributing guide with definitive instructions for adding a new Astro + X guide.

### DIFF
--- a/src/components/MediaGuidesNav.astro
+++ b/src/components/MediaGuidesNav.astro
@@ -12,7 +12,7 @@ const links = enPages
 	.sort((a, b) => (a.id > b.id ? 1 : -1))
 	.map((page) => {
 		const { logo, sidebar } = page.data;
-		if (!sidebar.label) throw new Error('Backend guides must always include a sidebar label.');
+		if (!sidebar.label) throw new Error('Media guides must always include a sidebar label.');
 		const pageUrl = '/' + page.id.replace('en/', `${lang}/`) + '/';
 		return { title: sidebar.label, href: pageUrl, logo };
 	});


src/components/MediaGuidesNav.astro
```diff
@@ -1,23 +1,20 @@
...
-	.sort((a, b) => {
-		// Sort alphabetically.
-		return a.data.service.toLowerCase() > b.data.service.toLowerCase() ? 1 : -1;
-	})
+	// Sort alphabetically.
+	.sort((a, b) => (a.id > b.id ? 1 : -1))
 	.map((page) => {
-		const { service } = page.data;
+		const { logo, sidebar } = page.data;
+		if (!sidebar.label) throw new Error('Backend guides must always include a sidebar label.');
```
ArmandPhilippot:
```suggestion
		if (!sidebar.label) throw new Error('Media guides must always include a sidebar label.');
```

sarah11918: Darnit! As soon as I saw you left a comment, I knew exactly what it was.... :joy:

PR #12858, opened 2025-12-08
i18n(fr): update `concepts/why-astro.mdx`

ArmandPhilippot:
#### Description (required)

Updates the French translation of `concepts/why-astro` to:
* Add spaces before units, see https://fr.wikipedia.org/wiki/Espace_ins%C3%A9cable#En_France for example (sorry, I didn't find a link in English for this).
* Small rewording in the same section

This file was set as outdated because of #12840 but this doesn't apply to the French version.

- Suggested label: i18n


yanthomasdev (approved):
LGTM!

PR #12867, opened 2025-12-10
fix(astro:assets): add docs for missing imports and types

ArmandPhilippot:
#### Description (required)

- Moves `astro/assets/utils` from the Image Service API to `astro:assets` module because we're now documenting the slash versions alongside the virtual modules. No content has changed except the following:
  - Updates the line breaks/indentation while moving the existing content.
  - Simplifies the code snippets
  - Updates `matchProtocol()` description (default value)
  - Adds default for `matchPort()`
- Formats headings for consistent use of inline code
- Adds new imports for `astro:assets`:
  - [`getConfiguredImageService()`](https://github.com/withastro/astro/blob/7d721b25a206e507de45ed3d6efa7e6941c25e2b/packages/astro/src/assets/internal.ts#L25): added in https://github.com/withastro/astro/pull/6530
  - [`imageConfig`](https://github.com/withastro/astro/blob/7d721b25a206e507de45ed3d6efa7e6941c25e2b/packages/astro/client.d.ts#L51): added in https://github.com/withastro/astro/pull/8414
- Adds `astro:assets` types
  - [`LocalImageProps`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L236-L250)
  - [`RemoteImageProps`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L252-L286)
- Adds `astro/assets` imports:
  - [`baseService`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/services/service.ts#L131-L151)
  - `getConfiguredImageService()`: same as `astro:assets`
  - [`getImage()`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/index.ts#L1): because this is internal code we probably don't want to expose it? Maybe this should be the [`astro:assets` version](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/client.d.ts#L31) instead.
  - [`isLocalService()`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/services/service.ts#L16-L22)
- Adds `astro/assets` types (same as `astro:assets`)
- Adds `astro` types related to assets (instead of listing them alphabetically, I tried to group them by usage)
  - [`ImageInputFormat`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L7)
  - [`ImageMetadata`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L38-L47): added in https://github.com/withastro/astro/pull/6530
    - `orientation`: added in https://github.com/withastro/astro/pull/7637
  - [`ImageOutputFormat`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L8)
  - [`ImageQuality`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L6)
  - [`ImageQualityPreset`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L5)
  - [`ImageTransform`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L82-L93)
    - [`densities`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L86) and [`widths`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L85): added in https://github.com/withastro/astro/pull/8620
    - [`fit`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L90) and [`position`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L91): added in https://github.com/withastro/astro/pull/12377 Note that responsive images have been stabilized in https://github.com/withastro/astro/pull/13917 but this doesn't update `ImageTransform`, so saying "since v5" is probably fine?
  - [`UnresolvedImageTransform`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L70-L77)
  - [`GetImageResult`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/types.ts#L95-L104)
    - [`srcSet`](https://github.com/withastro/astro/blob/16f3994fdb83d1b3421491c00bfd5ac9f7e37a5c/packages/astro/src/assets/types.ts#L99-L102): added in https://github.com/withastro/astro/pull/8620
  - [`RemotePattern`](https://github.com/withastro/astro/blob/8cab2a4f7ee0cfbcf0ddaec0878da637e7854b9d/packages/internal-helpers/src/remote.ts#L1-L6): exported in https://github.com/withastro/astro/commit/6ee63bfac4856f21b4d4633021b3d2ee059e553f / https://github.com/withastro/astro/pull/14466
  - [`ImageServiceConfig`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/types/public/config.ts#L40-L43): added in https://github.com/withastro/astro/pull/6848
  - [`ImageService`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/services/service.ts#L14)
  - [`ExternalImageService`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/services/service.ts#L82-L83)
  - [`LocalImageService`](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/src/assets/services/service.ts#L90-L117)
- Updates some existing types to flatten the type when it depends on non-exported types.
- Removes the `GetImageResult` code snippet in `getImage()`
- Reword a section in Image Service API and in the Images guide to replace a type displayed as a code snippet with a link
- Adds more link to `astro:assets` from the Image Service API

### DIFF
--- a/src/content/docs/en/reference/container-reference.mdx
+++ b/src/content/docs/en/reference/container-reference.mdx
@@ -430,7 +430,7 @@ const json = await response.json();
 <Since v="4.16.6" />
 </p>
 
-Whether or not the Container API renders components as if they were [page partials](/en/basics/astro-pages/#page-partials). This is usually the behavior you want when rendering `components.boolean` so you can render components without a full page shell.
+Whether or not the Container API renders components as if they were [page partials](/en/basics/astro-pages/#page-partials). The default `true` setting renders the component in isolation without a full page shell.
 
 To render a component as a full Astro page, including `<!DOCTYPE html>`, you can opt-out of this behavior by setting `partial` to `false`:
 
--- a/src/content/docs/fr/guides/cms/payload.mdx
+++ b/src/content/docs/fr/guides/cms/payload.mdx
@@ -88,9 +88,9 @@ export default Posts;
     });
     ```
 
-    Une nouvelle collection appelée "Posts" apparaîtra dans votre tableau de bord PayloadCMS à côté de la collection "Users". 
+    Une nouvelle collection appelée « Posts » apparaîtra dans votre tableau de bord PayloadCMS à côté de la collection « Users ». 
 
-2. Entrez dans la collection "Posts" et créez un nouveau message. Après l'avoir enregistré, vous remarquerez que l'URL de l'API apparaît dans le coin inférieur droit.
+2. Entrez dans la collection « Posts » et créez un nouveau message. Après l'avoir enregistré, vous remarquerez que l'URL de l'API apparaît dans le coin inférieur droit.
 
 3. Avec le serveur de développement en cours d'exécution, ouvrez `http://localhost:4321/api/posts` dans votre navigateur. Vous devriez voir un fichier JSON contenant le message que vous avez créé en tant qu'objet.
     ```json
@@ -134,7 +134,7 @@ Par exemple, pour afficher une liste de titres d'articles et leur contenu :
 ---
 import HomeLayout from "../layouts/HomeLayout.astro";
 
-const res = await fetch("http://localhost:5000/api/posts") // http://localhost:4321/api/posts by default
+const res = await fetch("http://localhost:5000/api/posts") // http://localhost:4321/api/posts par défaut
 const posts = await res.json()
 ---
 
--- a/src/content/docs/fr/guides/upgrade-to/v4.mdx
+++ b/src/content/docs/fr/guides/upgrade-to/v4.mdx
@@ -11,7 +11,7 @@ Ce guide vous aidera à migrer d'Astro v3 vers Astro v4.
 
 Besoin de mettre à niveau un ancien projet vers la v3 ? Consultez notre [ancien guide de migration](/fr/guides/upgrade-to/v3/).
 
-Besoin de voir la documentation de la v3 ? Visitez cette [ancienne version du site de documentation (instantané de la v3.6 non maintenu)](https://docs-git-v3-docs-unmaintained-astrodotbuild.vercel.app/).
+Besoin de voir la documentation de la v3 ? Visitez cette [ancienne version du site de documentation (instantané de la v3.6 non maintenu)](https://web.archive.org/web/20231203051122/https://docs.astro.build/fr/getting-started/).
 
 ## Mettre à niveau Astro
 
@@ -109,7 +109,7 @@ Inspectez visuellement vos pages Markdown/MDX avant le déploiement pour vous as
 
 Les modifications suivantes sont considérées comme des modifications avec rupture de compatibilité dans Astro. Ces changements peuvent ou non fournir une rétrocompatibilité temporaire, et toute la documentation est mise à jour pour faire référence uniquement au code actuellement pris en charge.
 
-Si vous avez besoin de vous référer à la documentation d'un projet v3.x, vous pouvez parcourir cet [instantané (non maintenu) de la documentation datant d'avant la sortie de la v4.0](https://docs-git-v3-docs-unmaintained-astrodotbuild.vercel.app/).
+Si vous avez besoin de vous référer à la documentation d'un projet v3.x, vous pouvez parcourir cet [instantané (non maintenu) de la documentation datant d'avant la sortie de la v4.0](https://web.archive.org/web/20231203051122/https://docs.astro.build/fr/getting-started/).
 
 ### Renommé : `entrypoint` (API des intégrations)
 
--- a/src/content/docs/fr/guides/upgrade-to/v5.mdx
+++ b/src/content/docs/fr/guides/upgrade-to/v5.mdx
@@ -609,8 +609,6 @@ const { afficherAlerte } = Astro.props;
 
 Les changements suivants sont considérés comme des changements avec rupture de compatibilité dans Astro v5.0. Ces changements peuvent ou non assurer une rétrocompatibilité temporaire. Si vous utilisiez ces fonctionnalités, vous devriez peut-être mettre à jour votre code comme recommandé dans chaque entrée.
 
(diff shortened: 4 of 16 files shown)


ArmandPhilippot:
This should be ready to review now!
* the TOC is a bit heavy now, but if we hide the `h4` this would hide the `<Image />` properties...
* `getImage()` from `astro/assets` is internal code. I don't know if this is expected or if this should be the [`astro:assets` version](https://github.com/withastro/astro/blob/f949aa87a40ced70b7533a0aaa0b6bbf29a8dbe6/packages/astro/client.d.ts#L31) instead.

src/content/docs/en/guides/images.mdx
```diff
@@ -488,18 +488,7 @@ The [Astro template syntax](/en/reference/astro-syntax/) also supports writing a
...
-
-```ts
-interface ImageMetadata {
-  src: string;
-  width: number;
-  height: number;
-  format: string;
-}
-```
+Local images must be imported from the relative path from the existing `.astro` file, or you can configure and use an [import alias](/en/guides/imports/#aliases). They become available as an [`ImageMetadata` object](/en/reference/modules/astro-assets/#imagemetadata-1) giving you access to the image's `src` and other properties to use in the `<img>` tag.
```
sarah11918: I like link update, but I do wonder if keeping example type is helpful here (even if content is duplicated). One of the issues we have in explaining how images work is that if you're just using the HTML image tag, you have to rely on the object for its properties (e.g. `image.src`) vs if you use the components, you can just write `src`). The extra reinforcement here might be good, actually, and not force people to click away for something that is kind of short anyway?

What do you think?

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -548,51 +557,958 @@ Values match those of CSS `object-position`. Defaults to `center`, or the value
...
-It returns an object with the following type:
+### `inferRemoteSize()`
+
+<p>
+
+**Type:** <code>(url: string) => Promise\<Omit\<<a href="#imagemetadata-1">ImageMetadata</a>, 'src' | 'fsPath'\>\></code><br />
+<Since v="4.12.0" />
+</p>
+
+A function to infer the dimensions of remote images. This can be used as an alternative to passing the `inferSize` property.
```
```suggestion
A function to set the original `width` and `height` of a remote image automatically. This can be used as an alternative to passing the [`inferSize` ](#infersize) property.
```
I like linking back, since there's more description of the concept of inferring size there. And, tried to avoid the "`inferSize` infers size!" kind of definition. This is what we say above, so this should work here too?

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -548,51 +557,958 @@ Values match those of CSS `object-position`. Defaults to `center`, or the value
...
+//    width: 800,
+//    height: 600,
+//    format: 'jpg',
+//    orientation: undefined
+// }
+```
+
+### `emitESMImage()`
+
+:::caution[Deprecated]
```
sarah11918: ~~Just checking that we know this still exists as deprecated, and hasn't been removed for v6.~~ We checked, and it has been removed for v6!

So, for ease of reconciling with v6 docs, let's do like we did in another module page and show it as deprecated in the imports, and throw it to the bottom? Seems odd to show this one before the one that you should be using to replace it anyway. 😄

sarah11918 (commented):
OK, just a few comments here before we decide what to do about the table of contents!

This could mean showing fewer levels, I think this should include avoiding "Image properties" so we can just have the image component, then each property etc. and we might end up strategically hiding some of the "building an image service only" content.

 We'll figure that out!

src/content/docs/en/guides/images.mdx
ArmandPhilippot: I'm used to click on links when I want to know more (they are not decorative 😄 ), so I didn't think about this. But, you're right, this might be easier for most people. I'll push a commit for that!

ArmandPhilippot:
I believe we would need this because of the `ru` file:
```
@lunaria-track:src/content/docs/en/**.mdx
```
https://lunaria.dev/guides/tracking/#lunaria-track

PR #12872, opened 2025-12-11
astro build environments

ematipico:
#### Description (required)

In v6 Astro has three environments during the build, one more compared to the classic `ssr` and `client` Vite environments.

This PR adds a new section to the integration APIs to cover these environments. 

The relevant [PR 14998](https://github.com/withastro/astro/pull/14998) was merged in main `next`

### DIFF
--- a/src/content/docs/en/reference/integrations-reference.mdx
+++ b/src/content/docs/en/reference/integrations-reference.mdx
@@ -1209,7 +1209,8 @@ resolveId(id) {
     return 'export const foo = "bar"';
   }
 }
-  
+```
+
 ## Integration types reference
 
 ### `AstroIntegrationLogger`


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1190,6 +1190,18 @@ declare global {
...
 
+## Astro vite environments
+
+Since v6, Astro is based on `vite` v7 and it heavily uses its [environment APIs](https://vite.dev/guide/api-environment). 
+  
+Astro build environments:
+- `ssr`: the classic and default `ssr` environment of `vite`. Astro uses this environment to build dynamic routes.
+- `client`: the classic and default `client` environment of `vite`.  Astro uses this environment to generate client assets, such as scripts, styles, etc.
+- `prerender`: an environment that is created to build static pages.
+
```
ematipico, line 1196:
```suggestion
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1190,6 +1190,18 @@ declare global {
 
 Astro reserves the `astro:` prefix for future built-in hooks. Please choose a different prefix when naming your custom hook.
 
+## Astro vite environments
+
+Since v6, Astro is based on `vite` v7 and it heavily uses its [environment APIs](https://vite.dev/guide/api-environment). 
+  
+Astro build environments:
+- `ssr`: the classic and default `ssr` environment of `vite`. Astro uses this environment to build dynamic routes.
```
matthewp: It's also used for non-dynamic routes in dev.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1190,6 +1190,18 @@ declare global {
...
+- `client`: the classic and default `client` environment of `vite`.  Astro uses this environment to generate client assets, such as scripts, styles, etc.
```
matthewp: Nit: Calling this and `ssr` "classic" is a little confusing to me. I think maybe instead you are wanting to say that they are vite-defined or built-in or something to that effect.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1190,6 +1190,18 @@ declare global {
...
+- `prerender`: an environment that is created to build static pages.
+
+  
+If your integrations rely on `vite` environments and the Astro build, you might want to optimise based on these environments provided.
```
matthewp: I would link to Vite's docs on environment (which imo aren't great but its what we have).

I might also mention that the primary thing you do with environment is apply config.

src/content/docs/en/reference/integrations-reference.mdx
ematipico: Do you think it's worth having a small paragraph for dev? At line 1197 I mention **build**

matthewp: If the goal is to explain that Astro uses environments now I would probably explain it all together. If the goal is to explain how there's `prerender` as a special environment in the build, I would explain the reasoning for this documentation more. Like for example something like:

```md
Astro inherits the environments Vite provides by default, `ssr` and `client`. Additionally there is the the `prerender` environment provided by Astro. Here's what each of these environments are used for:
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1190,6 +1190,16 @@ declare global {
 
 Astro reserves the `astro:` prefix for future built-in hooks. Please choose a different prefix when naming your custom hook.
 
+## Astro vite environments
+
+Astro inherits the environments Vite provides by default, `ssr` and `client`. 
+
+Additionally there are two other environments that Astro creates:
+- `prerender` is an environment used during the `build` and it's used to build static pages.
+- `astro` is an environment used during the development, and it's used as "secondary" SSR environment when the Vite `ssr` environment [isn't a runnable dev environment](https://vite.dev/guide/api-environment-frameworks#runnabledevenvironment).
```
sarah11918:
```suggestion
- `astro` is an environment used during the development, and it's used as a "secondary" SSR environment when the Vite `ssr` environment [isn't a runnable dev environment](https://vite.dev/guide/api-environment-frameworks#runnabledevenvironment). 
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1190,6 +1190,16 @@ declare global {
...
 
+## Astro vite environments
+
+Astro inherits the environments Vite provides by default, `ssr` and `client`. 
+
+Additionally there are two other environments that Astro creates:
+- `prerender` is an environment used during the `build` and it's used to build static pages.
+- `astro` is an environment used during the development, and it's used as "secondary" SSR environment when the Vite `ssr` environment [isn't a runnable dev environment](https://vite.dev/guide/api-environment-frameworks#runnabledevenvironment). 
+
+If your integrations rely on [`vite` environments](https://vite.dev/guide/api-environment) and the Astro build, you might want to optimise based on these environments provided. One of the main usage of Vite environments, is the ability to run a configure plugins conditionally.
```
sarah11918:
``````suggestion
Astro's [Vite environments](https://vite.dev/guide/api-environment) allow you to optimize your integration's Vite plugins for different environments. One of the main uses of Vite environments is the ability to run and configure your integration's Vite plugins conditionally:

```js
resolveId(id) {
  if (id === '\0virtual:foo') {
    if (this.environment.name === 'client') {
      throw new Error('This is a server-only module');
    }
    return 'export const foo = "bar"';
  }
}
```
``````
Also asking @florian-lefebvre to help with wording here! And, maybe a small code example showing what it looks like to do something conditionally with plugins would probably be helpful here?

UPDATE: Updated to include the example here.

Comment (florian-lefebvre) [reply src/content/docs/en/reference/integrations-reference.mdx]: Wording LGTM! I think we could show an example with `resolveId()`:
```js
resolveId(id) {
	if (id === '\0virtual:foo') {
		if (this.environment.name === 'client') {
			throw new Error('This is a server-only module');
		}
		return 'export const foo = "bar"';
	}
}
```
I don't think we should go in depth, it's vite's docs responsability

ematipico: Yeah we shouldn't go deeper than that.

sarah11918: Yup, that minimal example looks perfect to me!

sarah11918: @ematipico I'll leave it up to you to accept or modify that suggestion!

sarah11918:
(not `merge-on-release` as this is already released. This can merge any time!)

PR #12891, opened 2025-12-13
i18n(fr): update `guides/backend/xata.mdx` & `guides/cms/craft-cms.mdx`

ArmandPhilippot:
#### Description (required)

Both `guides/backend/xata.mdx` and `guides/cms/craft-cms.mdx` are already up to date (still the same issue related to #12759 and #12842).

* Small rewording in `guides/backend/xata.mdx`: 
  * "serverless" is not in the [French i18n guide](https://github.com/withastro/docs/blob/main/i18n-guides/fran%C3%A7ais.md#autres-mots-non-traduits) but this is the same as "headless", we shouldn't translate it.,
  * "full-text search" should be translated with ["recherche plein texte"](https://fr.wikipedia.org/wiki/Recherche_plein_texte)
* Add a blank line to `guides/cms/craft-cms.mdx`

- Suggested label: i18n


yanthomasdev (approved):
LGTM!

PR #12892, opened 2025-12-13
fix(upgrade-to): update the link to the v3 snapshot

ArmandPhilippot:
🚨 **For translators:** You may also need to check #12871. This PR did not trigger a status change due to "broken link" in the PR title.

#### Description (required)

We have updated the v3 snapshot link in https://github.com/withastro/docs/pull/12871 because this is now a broken link but we missed other places. So, I updated the links.

One of them is inside commented docs, and given the text this should be a link to v4... Should we also uncomment that part even if we are a bit late? 😅 

- Suggested label: `typo/link/grammar - quick fix!`

### DIFF
--- a/src/content/docs/de/basics/astro-components.mdx
+++ b/src/content/docs/de/basics/astro-components.mdx
@@ -50,7 +50,7 @@ import IrgendeineAstroKomponente from '../components/IrgendeineAstroKomponente.a
 import IrgendeineReactKomponente from '../components/IrgendeineReactKomponente.jsx';
 import irgendwelcheDaten from '../data/pokemon.json';
 
-// Zugriff auf übergebene Props (Komponenteneigenschaften), wie z.B. `<X title="Hallo, Welt!" />`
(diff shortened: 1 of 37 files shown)


src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -609,7 +609,7 @@ const { showAlert } = Astro.props;
 
 The following changes are considered breaking changes in Astro v5.0. Breaking changes may or may not provide temporary backwards compatibility. If you were using these features, you may have to update your code as recommended in each entry.
 
-{/* If you need to refer to the documentation for a v4.x project, you can browse this [(unmaintained) snapshot of the docs from before v5.0 was released](https://docs-git-v3-docs-unmaintained-astrodotbuild.vercel.app/). */}
+{/* If you need to refer to the documentation for a v4.x project, you can browse this [(unmaintained) snapshot of the docs from before v5.0 was released](https://v4.docs.astro.build/en/getting-started/). */}
```
sarah11918:
```suggestion
```
You caught me again 😅

PR #12894, opened 2025-12-13
feat(next): harden vite links in upgrade guides

trueberryless:
#### Description

This PR pin points each Vite documentation link in the major upgrade guides to their specific version for better reproducibility. The Vite documentation offers subdomains for each major Vite version, so I think it makes sense to use them in the upgrade guides. Especially Vite major migration guide links should be pin pointed so they do not all point to the current v6->v7 migration guide.


sarah11918:
We typically do not go back and update previous guides like this unless it's causing us a problem or a link has failed. It is too much maintenance for pages that people are typically not going to look at, and I don't want to set this precedent.

PR #9978, opened 2024-11-11
Migrate Astro Docs to Lunaria v1 prerelease

yanthomasdev:
#### Description (required)

This PR moves the main Lunaria logic used from v0.1 to the current v1 prerelease. This is meant to unblock #9896 with the new `parameters` options for custom path parameters that will be necessary to move into the new built-in Starlight i18n, as well as make it easy for me to update Lunaria whenever a new bug arises (they're always on the horizon when it comes to my code)

- [X] Migrate Tuesday Bot
- [X] Migrate Lunaria dashboard
- [X] Migrate GitHub Action

To test locally, run `pnpm lunaria:build` and open `dist/lunaria/index.html` in your browser. Then, compare it with https://i18n.docs.astro.build/ to see if there are any differences.

Take in mind this is very early Lunaria v1, things like the dashboard in this PR and soon the updated Action are far from what is planned on the roadmap, so expect something better than that.

PR #12896, opened 2025-12-13
Lunaria: fix configuration

urban-adeininger:
In the Lunaria config file, you will find option `ignoredKeywords`. According to lunaria [documentation](https://lunaria.dev/reference/configuration/#ignorekeywords), the correct spelling is `ignoreKeywords`, however.
This PR fixes that issue.

### DIFF
--- a/lunaria.config.ts
+++ b/lunaria.config.ts
@@ -123,7 +123,7 @@ export default defineConfig({
 	],
 	tracking: {
 		localizableProperty: 'i18nReady',
-		ignoredKeywords: [
+		ignoreKeywords: [
 			'lunaria-ignore',
 			'typo',
 			'en-only',


ArmandPhilippot:
Thanks, but this is not a typo and the configuration is correct: we use a prerelease of Lunaria v1 since https://github.com/withastro/docs/pull/9978 and if you check the `v1` branch, you can find [`ignoredKeywords` under `tracking`](https://github.com/lunariajs/lunaria/blob/f07e1a3c031192878091bd729c4b70c17fe47ea8/packages/core/src/config/schema.ts#L86-L91). The Lunaria docs website is not yet updated for the v1 because this is only a prerelease, so I can understand the confusion here. 😄

PR #12903, opened 2025-12-14
move bold text + remove extra ** in v6 migration page

KTrain5169:
#### Description (required)

I removed an extra ** in the v6 migration page. I also moved the bold text since I felt that simply "and some existing backwards compatibility that was not previously behind a flag" being bold wasn't explicit enough.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -1084,7 +1084,7 @@ If you have manual links to headings, you may need to update the anchor link val
 will now generate the following HTML with a trailing hyphen in the heading `id`:
 
 ```html ins="-"
-<h2 id="picture-"><code>&lt;Picture /&gt;</h2>
+<h2 id="picture-"><code>&lt;Picture /&gt;</code></h2>
 ```
 
 and must now be linked to as:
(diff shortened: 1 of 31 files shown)


trueberryless (approved):
Awesome catch 🚀
LGTM 🎉

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -300,7 +300,7 @@ Projects now containing these removed features will be unable to build, and ther
...
 
-**Astro v6.0 removes this previously deprecated Content Collections API support entirely, including the `legacy.collections` flag **and some existing backwards compatibility that was not previously behind a flag**. All content collections must now use [the Content Layer API introduced in Astro v5.0](https://astro.build/blog/content-layer-deep-dive/) that powers all content collections. **No backwards compatibility support is available.**
+Astro v6.0 removes this previously deprecated Content Collections API support entirely, **including the `legacy.collections` flag and some existing backwards compatibility that was not previously behind a flag**. All content collections must now use [the Content Layer API introduced in Astro v5.0](https://astro.build/blog/content-layer-deep-dive/) that powers all content collections. **No backwards compatibility support is available.**
```
```suggestion
Astro v6.0 removes this previously deprecated Content Collections API support entirely, including the `legacy.collections` flag **and some existing backwards compatibility that was not previously behind a flag**. All content collections must now use [the Content Layer API introduced in Astro v5.0](https://astro.build/blog/content-layer-deep-dive/) that powers all content collections. **No backwards compatibility support is available.**
```
Original bolding was intentional! We don't need to bold that a legacy flag is going away, but it's the "previously existing content not behind a flag" that is what we need people to notice/understand as it's unexpected.

PR #12905, opened 2025-12-15
Add Zod v3 → v4 codemod link to upgrade guide

HiDeoo:
#### Description

This PR updates the Astro 6 upgrade guide to include a link to the [community codemod](https://github.com/nicoespeon/zod-v3-to-v4) for migrating from Zod 3 to Zod 4.

I went with the minimal change to add this link right now, but I guess a few questions could be raised:

- I introduce it at the end of the Zod 4 section, I guess another option could be near the beginning of the section, after linking to the Zod migration guide.
- I only went with a link, not even a command example on how to run the codemod (it's available on the linked README).
- I only mention it in the Zod 4 section, not in the "Deprecated: astro:schema and z from astro:content" section even tho it can also be relevant there. Should it be mentioned there too?

#### Related issues & labels

- Closes #
- Suggested label: 6.0

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -24,19 +24,19 @@ Update your project's version of Astro to the latest version using your package 
   <Fragment slot="npm">
   ```shell
   # Upgrade Astro and official integrations together
-  npx @astrojs/upgrade
+  npx @astrojs/upgrade alpha
   ```
   </Fragment>
   <Fragment slot="pnpm">
   ```shell
   # Upgrade Astro and official integrations together
-  pnpm dlx @astrojs/upgrade
+  pnpm dlx @astrojs/upgrade alpha
   ```
   </Fragment>
   <Fragment slot="yarn">
   ```shell
   # Upgrade Astro and official integrations together
-  yarn dlx @astrojs/upgrade
+  yarn dlx @astrojs/upgrade alpha
   ```
   </Fragment>
 </PackageManagerTabs>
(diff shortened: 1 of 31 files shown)


src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -150,7 +150,9 @@ const blog = defineCollection({
 
 For the old behavior where defaults are parsed, use the new `.prefault()` method.
 
-These are only some of the many changes upgrading from Zod 3 to Zod 4. If you encounter any issues with your Zod schemas after upgrading to Astro 6, please consult the [Zod 4 changelog](https://zod.dev/v4/changelog) for complete upgrade guidance.
+These are only some of the many changes upgrading from Zod 3 to Zod 4.
+Additionally, a [community codemod](https://github.com/nicoespeon/zod-v3-to-v4), which can potentially automate some of these changes when migrating from Zod 3 to Zod 4, is also available.
+If you encounter any issues with your Zod schemas after upgrading to Astro 6, please consult the [Zod 4 changelog](https://zod.dev/v4/changelog) for complete upgrade guidance.
```
```suggestion
These are only some of the many changes upgrading from Zod 3 to Zod 4. If you encounter any issues with your Zod schemas after upgrading to Astro 6, please consult the [Zod 4 changelog](https://zod.dev/v4/changelog) for complete upgrade guidance.

Additionally, a [community codemod](https://github.com/nicoespeon/zod-v3-to-v4), which can potentially automate some of these changes when migrating from Zod 3 to Zod 4, is also available.
```
Feels like this comes better *after* the existing paragraph, tacked on at the end.

sarah11918:
Let's get this in like this for now!

- I could see posting both the upgrade link and the code mod right at the top, too (before the what should I do). Maybe something like:

```
### Zod 4

Astro v6.0 upgrades to [Zod 4 (CHANGELOG)](https://zod.dev/v4/changelog)), a major dependency update that may require changes to custom Zod schemas in your project. A [community Zod v3 to Zod v4 codemod](https://github.com/nicoespeon/zod-v3-to-v4) is also available.
```

- I think anyone dealing with upgrading the `z` schema stuff from other imports/exports will *also* end up here, so the codemod only here seems fine. BUT, we probably could update that entry to also link to the Zod 4 one? Just to be safe? It could be something like:

```
Astro 6.0 deprecates `astro:schema` and `z` from `astro:content` in favor of `astro/zod`. (Note that Astro 6.0 also [upgrades to Zod 4](#zod-4) and may require additional upgrades to your project.)
```

HiDeoo:
> Let's get this in like this for now!

Sounds like a good call to me (and also why I initially did not go with adding it to multiple places, etc but kept it as simple as possible) 👍

PR #12915, opened 2025-12-15
[v6]: add docs for `entryType` in Adapter API

ArmandPhilippot:
#### Description (required)

Adds docs for `entryType`, a new property added to `setAdapter()` in https://github.com/withastro/astro/pull/14306

The second added property (`devEntrypoint`) is now removed, see https://github.com/withastro/astro/pull/15005

- Suggested label: v6, add new content

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -152,6 +152,8 @@ For the old behavior where defaults are parsed, use the new `.prefault()` method
 
 These are only some of the many changes upgrading from Zod 3 to Zod 4. If you encounter any issues with your Zod schemas after upgrading to Astro 6, please consult the [Zod 4 changelog](https://zod.dev/v4/changelog) for complete upgrade guidance.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -246,6 +246,47 @@ export default function createIntegration() {
...
+### `entryType`
+
+<p>
+
+**Type:** `"legacy-dynamic" | "self"`<br />
+**Default**: `"legacy-dynamic"`<br />
+<Since v="6.0.0" />
+</p>
+
+Determines how the adapter's entrypoint is handled during the build. By default, Astro will build one and no additional configuration is required. You can provide your own entrypoint with `"self"` and by setting [`rollupOptions.input`](https://rollupjs.org/configuration-options/#input) with your entrypoint path.
```
```suggestion
Specifies whether the adapter's entrypoint is automatically created by Astro (default), or whether a custom entrypoint is used.

To provide your own entrypoint, configure `entryType: "self"` and set [`rollupOptions.input`](https://rollupjs.org/configuration-options/#input) with your custom entrypoint path.
```
Is it really something like this?

ArmandPhilippot: Yes, at least that's what I understand from looking at what's in the changeset and in the JSDoc!

I'll update your suggestion because I think you forgot to remove "by" from my wording ("and by set" > ""and set").

PR #12916, opened 2025-12-15
fix: improve wording partial reference section

### LINKED ISSUE #12914: Partial reference seems to include confusing `boolean`
### 📚 Subject area/topic

Page Partials

https://docs.astro.build/en/reference/container-reference/#partial-option

In the sentence:

> This is usually the behavior you want when rendering `components.boolean` so you can render components without a full page shell.

the "boolean" seems out of place. `components.boolean` is only mentioned this one time in the whole docs. Should it be `partial: false` instead?

As I am not 100% sure what it is supposed to mean, I created an issue. I traced the origin back to [this PR](https://github.com/withastro/docs/pull/9696#discussion_r1802946060) but couldn't determine it's correctness.

Thanks to Tim Gates for pointing out this possible typo!

_No response_

trueberryless:
#### Description

As discussed in #12914 there is a small typo in the [partial reference section](https://docs.astro.build/en/reference/container-reference/#partial-option) which we at the same time used to rewrite the sentence with @ArmandPhilippot so it sounds better 👍 

- Closes #12914

### DIFF
--- a/src/content/docs/en/reference/image-service-reference.mdx
+++ b/src/content/docs/en/reference/image-service-reference.mdx
@@ -284,6 +284,33 @@ export default defineConfig({
 });
 ```
 
+## Typing custom image service props
+
+<p><Since v="5.16.6" /></p>
+
+If your image service supports additional props in Astro’s `<Image>` component, `<Picture>` component, or the `getImage()` function, you can add types for these by extending the `Astro.CustomImageProps` interface.
+
+For example, to add a custom `blur` prop that your image service supports:
+
+```ts
+declare namespace Astro {
+  interface CustomImageProps {
+    /** Apply a Gaussian blur with this radius to the image. */
+    blur?: number;
+  }
+}
+```
+
+You can expose these types to users by making your image service an [Astro integration](/en/reference/integrations-reference/) and using the [`injectTypes()`](/en/reference/integrations-reference/#injecttypes-option) helper.
+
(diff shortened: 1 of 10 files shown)


src/content/docs/en/reference/container-reference.mdx
```diff
@@ -430,7 +430,7 @@ const json = await response.json();
 <Since v="4.16.6" />
 </p>
 
-Whether or not the Container API renders components as if they were [page partials](/en/basics/astro-pages/#page-partials). This is usually the behavior you want when rendering `components.boolean` so you can render components without a full page shell.
+Whether or not the Container API renders components as if they were [page partials](/en/basics/astro-pages/#page-partials). This is usually the behavior you want, as it allows you to render components without a full page shell.
```
sarah11918:
```suggestion
Whether or not the Container API renders components as if they were [page partials](/en/basics/astro-pages/#page-partials). The default `true` setting renders the component in isolation without a full page shell.
```
I think we can just stick to wording about what it does?

ArmandPhilippot: Agree, this sounds better! I'll edit your suggestion to remove an extra space then I'll apply your suggestion so this PR do net get stuck any longer. 😄

trueberryless: Thanks Armand! 

I agree and also would have removed the extra space! But I'm currently AFK 🙌

PR #12921, opened 2025-12-16
change resource sentence to say v6 instead of v5 in v6 migration guide

KTrain5169:
#### Description (required)

The end of the v6 migration guide said "Know a good resource for Astro v5.0?", but it should be v6.0 instead. This PR fixes that.

- Closes #
- Suggested label:


trueberryless:
If you want, you could also include the same v5 to v6 typo change under the "Breaking changes" section. In the intro it still says v5 👍

sarah11918:
Thanks @KTrain5169 , well spotted!

Unfortunately, this section shouldn't be shown yet because that link doesn't work. I usually comment this out when the docs go live. So thank you for the v6 update, and I'll hide this section for now (and, there aren't going to be any community resources yet this early anyway).

Thanks for the help!

PR #12933, opened 2025-12-17
i18n(ko-KR): update `astro-assets.mdx` and `image-service-reference.mdx`

jsparkdev:
#### Description (required)

- #12897 
- #12867

- Closes #
- Suggested label:


src/content/docs/en/reference/modules/astro-assets.mdx
yanthomasdev, line 1: I believe this change shouldn't be here. Can you make a separate PR for it?

jsparkdev:
@yanthomasdev

Thank you!

I've updated it.

PR #12954, opened 2025-12-18
Clarify Vercel adapter ISR caching behavior and reorganize configuration options

psd-coder:
#### Description (required)
We have pages with data coming from our CMS, which we invalidate using an on-demand invalidation mechanism, and data coming from third-party APIs, where we want to invalidate the page by setting the necessary `s-maxage` through `Cache-Control` header. Occasionally, to make the second approach work, you must specify an `expiration` value, which isn’t obvious from the docs.

The current docs say ISR “caches for the duration of your deployment,” which is technically accurate, but they don’t warn users that this can make explicit cache headers ineffective. This led to a long debugging process to find the root cause and understand that adding any `expiration` value restores `Cache-Control` behavior.

This PR improves the documentation for the ISR feature overall. It:
- Clarifies caching behavior by adding a caution section about `Cache-Control` behavior when expiration is omitted.
- Restructures the ISR configuration section to better explain different invalidation strategies. (The “Excluding paths from caching” section currently describes three unrelated features together: `exclude`, `bypassToken` for on-demand invalidation and `bypassToken` for draft mode).
- Fixes a typo.


src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -257,9 +257,14 @@ Note that ISR function requests do not include search params, similar to [reques
 
 By default, an ISR function caches for the duration of your deployment. You can further control caching by setting an expiration time, or by excluding particular routes from caching entirely.
 
+:::caution
```
sarah11918: Hi, thanks for your first docs PR!

Just a heads up that we typically only use cautions for dangerous situations (your project won't build, you might lose data) and not just when describing "how things work."

Feel free to include the paragraph above and rework both together entirely so that this reads more like an explanation of how caching works. That's what we normally do when we want to add more information or context! This doesn't have to be a "patch a note on existing text" situation. If we didn't explain it well, or completely, then our explanation should change!

psd-coder: Hi! Thanks for the clarification about cautions use. I incorporated warning right into text. Does it look good now?

src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -257,9 +257,11 @@ Note that ISR function requests do not include search params, similar to [reques
 
 By default, an ISR function caches for the duration of your deployment. You can further control caching by setting an expiration time, or by excluding particular routes from caching entirely.
 
+When no `expiration` value is set, cached ISR pages can only be invalidated through [on-demand invalidation](#on-demand-invalidation). In this configuration, any custom `Cache-Control` headers you set on these pages will be ignored by Vercel's edge cache. If you need `Cache-Control` headers to control caching behavior (for example, to set shorter TTLs on pages with third-party API data), you must configure an `expiration` value.
```
ArmandPhilippot: Any reasons to talk about `expiration` before the dedicated heading? I'm not familiar with Vercel but reading this, I wonder if this shouldn't be said after the heading and even merged with the current sentence. Something like:
> By default or when using `false`, the pages will stay cached as long as the current deployment is in production. This means cached ISR pages can only be invalidated through [on-demand invalidation](#on-demand-invalidation) and any custom `Cache-Control` headers you set on these pages will be ignored by Vercel's edge cache.
>
> To control caching behavior through `Cache-Control` headers (for example, to set shorter TTLs on pages with third-party API data), you have to change the length of time to cache routes by configuring an `expiration` value in seconds:

I can't make a proper suggestion because of the updated line below; GitHub says it can make a suggestion on deleted lines.

src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -289,16 +290,37 @@ export default defineConfig({
...
-              '/preview', 
-              '/auth/[page]',
-              /^\/api\/.+/ // Regular expressions supported since @astrojs/vercel@v8.1.0
+                '/preview',
+                '/auth/[page]',
+                /^\/api\/.+/ // Regular expressions supported since @astrojs/vercel@v8.1.0
             ]
         }
     })
 })
```
ArmandPhilippot, line 322: This was removed but this break the rendering:
````suggestion
})
```
````

**Comment by psd-coder** [comment]
Hi @ArmandPhilippot! Your suggestions make sense. I moved the explanation to the end of `Time-based invalidation` section and fixed the markup issue. I hope it looks better now.

File: src/content/docs/en/guides/integrations-guide/vercel.mdx [hunk 4]
```diff
@@ -259,7 +259,9 @@ By default, an ISR function caches for the duration of your deployment. You can
 
 ##### Time-based invalidation
 
-You can change the length of time to cache routes this by configuring an `expiration` value in seconds:
+You can change the length of time to cache routes by configuring an `expiration` value in seconds.
+
+By default or when using `false`, the pages will stay cached as long as the current deployment is in production. This means cached ISR pages can only be invalidated through [on-demand invalidation](#on-demand-invalidation). Note that in this case, any custom `Cache-Control` headers returned from your pages (for example, if you want to have shorter TTLs on pages with third-party API calls) won't work. To make them work, you have to set an `expiration` time.
```
ArmandPhilippot: If the goal is to highlight `expiration` is required to use `Cache-Control` headers, I think this can be shortened:
```suggestion
You can change the length of time to cache routes by configuring an `expiration` value in seconds. This will apply to all pages and you can override this duration for specific pages using [`Cache-Control` headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control) (e.g., on pages with third-party API calls).
```
If you wonder why I suggest to remove the other sentences:
* "_By default or when using `false`, the pages will stay cached as long as the current deployment is in production._": we already say that in the parent section, I don't think this is helpful to repeat it.
* "_This means cached ISR pages can only be invalidated through on-demand invalidation._": this doesn't sound accurate. They can be invalidated either with `bypassToken` or `expiration`. And if we update the sentence, then we're repeating what we say in the parent section.

src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -289,11 +290,33 @@ export default defineConfig({
...
+    })
+})
+```
+
+You can then invalidate a cached page by sending a HEAD or GET request to the page URL with the `x-prerender-revalidate` header set to your bypass token. See [Vercel's on-demand ISR documentation](https://vercel.com/docs/build-output-api/v3/features#on-demand-incremental-static-regeneration-isr) for details.
+
+
+##### Draft mode
+
+To bypass the ISR cache and render fresh content (e.g., for previewing unpublished CMS content), use [Vercel's Draft mode](https://vercel.com/docs/build-output-api/v3/features#draft-mode). This also requires a `bypassToken` configured as shown above.
```
ArmandPhilippot: A nit regarding "shown above". Because this is another section, I think a link to the previous section might be safer here (i.e. if/when content are reorganized).

And maybe we should explain the difference with the previous section. If I understood correctly, one use a `x-prerender-revalidate` header while the other use a `__prerender_bypass` cookie. Does the following sounds accurate to you?
```suggestion
To bypass the ISR cache and render fresh content (e.g., for previewing unpublished CMS content), use [Vercel's Draft mode](https://vercel.com/docs/build-output-api/v3/features#draft-mode). This requires [defining a `bypassToken`](#on-demand-invalidation) in your configuration and reusing its value in your pages to [set a cookie](/en/guides/on-demand-rendering/#cookies) named `__prerender_bypass`.
```

ArmandPhilippot (commented):
Thanks for the update! Sorry for these back-and-forths, but rereading this and looking at Vercel's docs, I wonder if, now, this is not even more confusing than your initial caution. We might need to split this information in different places.

I left a suggestion for the "Time-based invalidation" section to avoid repeating what we already describe in the parent section. But, if this doesn't sound enough to you, maybe `l258` can be updated with something like:
```diff
-By default, an ISR function caches for the duration of your deployment. You can further control caching by setting an expiration time, or by excluding particular routes from caching entirely.
+By default, an ISR function caches for the duration of your deployment. This means Vercel will ignore any custom `Cache-Control` headers. You can further control caching by setting an expiration time, by creating a bypass token, or by excluding particular routes from caching entirely.
```

This way, I think we don't need asides or to repeat the same information in different places.

I left another suggestion for the "Draft mode" section which is a bit short. I think describing how this is different from the "On-demand invalidation" section could be useful.

After that, I think we'll be in good shape to merge this!

psd-coder:
Hi! I like the suggestion for the Draft mode section (applied it!), but I’m not sure about other changes. The problem is that you might not need time based expiration at all. You can rely on on-demand invalidation for the most of the pages, but sometimes you require custom cache behavior. In this case you will specify `Cache-Control` header with necessary directives. And it won’t work… It isn’t really obvious you have to specify some value for `expiration` to make `Cache-Control` header work with on-demand invalidation. My goal was to highlight this in the docs. I simplified copy for the **Time-based invalidation** section. Does it sound better?

> "This means cached ISR pages can only be invalidated through on-demand invalidation.": this doesn't sound accurate. They can be invalidated either with bypassToken or expiration. And if we update the sentence, then we're repeating what we say in the parent section.

If I understand everything right, `bypassToken` doesn't invalidate cache. It just allows you bypass cached value, run function and get fresh page content. It's described here: https://vercel.com/docs/draft-mode

src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -259,7 +259,7 @@ By default, an ISR function caches for the duration of your deployment. You can
 
 ##### Time-based invalidation
 
-You can change the length of time to cache routes this by configuring an `expiration` value in seconds:
+You can change the length of time to cache routes by configuring an `expiration` value in seconds. Setting this value also allows making `Cache-Control` headers to work as expected (otherwise they are ignored).
```
ArmandPhilippot: If this is the default behavior, I think this should be said without parentheses. And because this is not our job to describe how Vercel works in details maybe a link could be helpful here.
What about the following:
```suggestion
By default, when ISR is enabled, routes use [Vercel's cache shielding](https://vercel.com/docs/incremental-static-regeneration#differences-between-isr-and-cache-control-headers) and any Cache-Control headers are ignored. Configuring an `expiration` value (in seconds) allows you to control how long routes are cached. This means Cache-Control directives set by your application are also respected.
```

src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -276,11 +276,10 @@ export default defineConfig({
 });
 ```
```
ArmandPhilippot: Because we're now talking about Cache Control headers and the example is unrelated, we'll need a sentence to introduces the example (and the comment is no longer useful).
````suggestion
The following example defines `expiration` to cache all pages on first request and save them for 1 day:

```js title="astro.config.mjs" {7-9}
import vercel from '@astrojs/vercel';

export default defineConfig({
  // ...
  adapter: vercel({
    isr: {
      expiration: 60 * 60 * 24,
    },
  }),
});
```
````

psd-coder:
Agree, both suggestions are good and give you more context. I applied them. Thanks for your help!

PR #12982, opened 2025-12-24
i18n(ru): add `guides/fonts.mdx`

wavy-cat:
#### Description (required)

Added Russian translation for the Fonts guide.

Discord: `wavycat`.

- Suggested label: `i18n`


sarah11918:
Thanks, @wavy-cat ! I'm sorry it took such a long time to get around to this PR! We rely on community volunteers who are native speakers for reviews and we just didn't get any in time... because, I'm afraid that we are very close to a v6 release and this page has been completely rewritten in the v6 docs! We would prefer not to have old, outdated information in v6, so not having any translation at all is actually better, and the English will be shown. 

So, I think it makes sense to close this PR, and we hope you'll considering translating in the future! For a language where we have trouble getting reviewers, I would suggest you wait until the v6 docs are published, because several docs pages will change significantly, and we might just be in the same position if you try again with another page. And, I don't want to put you through this again!

Thank you very much for your understanding, and hope to see more translations in the future!

PR #12981, opened 2025-12-23
Add Cloudflare Astro 6 docs

ascorbic:
#### Description (required)

This PR updates the Cloudflare integration guide to use the new Astro 6 APIs. It also adds a migration guide for users moving from Astro 5 to 6. It includes some info about manually configuring Cloudflare Pages (which is no longer automatic), but it doesn't yet have a very good guide to migrate from Pages to Workers. It links out to the guide in the Cloudflare docs, but I'd like to add something more Astro-specific in a follow-up PR.

#### Related issues & labels (optional)

- Closes #
- Suggested label:

PR #12995, opened 2025-12-29
docs for drop cloudflare pages pr

alexanderniebuhr:
This PR removes references to Cloudflare Pages because we will drop support for it, it also removes the routes sections as this is not needed anymore, since out of the box everything will be configured for the user.

cc @ascorbic not sure how much you have WIP for this page and if we might get conflicted

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -224,44 +224,6 @@ export default defineConfig({
 	}),
 });
 ```
-### `routes.extend`
-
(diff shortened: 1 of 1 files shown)


ArmandPhilippot:
@alexanderniebuhr Matt's PR is here https://github.com/withastro/docs/pull/12981 if you want to chime in it. I think the only difference is he doesn't remove everything in the "Routes" section.

alexanderniebuhr:
Okay let's go with his one then.

ArmandPhilippot:
Ah, I didn't see that you had already commented! 😅  Yeah, I think it makes sense to keep everything in the same place unless there is bigger changes to be made!

PR #13003, opened 2025-12-31
Add @gutenye/astro-i18next integration for i18n

gutenye:

- Closes #
- Suggested label:


sarah11918:
Thank you for this PR adding your own integration! In reviewing the existing list, we realized that there was at least one broken link, and we decided it's better to have only one link directly to the directory for all the i18n integrations so we are not duplicating work and maintenance here!

So I've updated to what I think is more helpful than a partial list of *some* options. I think this page is much better now! Thank you for bringing it to our attention. 🙌

PR #13012, opened 2026-01-03
Small code cleanup

delucis:
#### Description (required)

This PR removes some old code and unused dependencies:

- Path utilities we no longer use because Starlight handles stuff for us
- `<StudioHeading>` component which we no longer use (we had kept it around because it was still in some translated pages, but this is no longer the case)
- JSX config we had hanging around from when we used to use Preact
- An unused `export`


yanthomasdev (approved):
LGTM!

PR #13020, opened 2026-01-06
feat(fonts): clean types

florian-lefebvre:

- Closes #
- Suggested label:


src/content/docs/en/reference/experimental-flags/fonts.mdx
ArmandPhilippot: I think we also need this change:

```suggestion
**Type:** `FontProvider | "local"`
```

PR #13021, opened 2026-01-07
fix(integrations-reference): replace source code links with proper docs

ArmandPhilippot:
**Context:** We have a few links pointing to the source code to describe the types, instead of proper documentation. Either users need them and we should document the types and their properties, or the links to the source code should be removed.

#### Description (required)

* Removes a broken external link for `AddressInfo`. We could replace it with https://nodejs.org/api/net.html#serveraddress I guess, but we are already linking to the module page.
* Flattens the [`RouteOptions`](https://github.com/withastro/astro/blob/4eb6fc9f9a0804783377dc7dd3e73cb35af68085/packages/astro/src/types/public/integrations.ts#L14-L27) type: we already describes the properties so it sounded a bit redundant to describe the type as well.
* Replaces links to source code with docs for:
  * [`AstroIntegrationMiddleware`](https://github.com/withastro/astro/blob/4eb6fc9f9a0804783377dc7dd3e73cb35af68085/packages/astro/src/types/public/integrations.ts#L190-L193)
  * [`AstroRenderer`](https://github.com/withastro/astro/blob/4eb6fc9f9a0804783377dc7dd3e73cb35af68085/packages/astro/src/types/public/integrations.ts#L55-L62)
  * [`ClientDirectiveConfig`](https://github.com/withastro/astro/blob/4eb6fc9f9a0804783377dc7dd3e73cb35af68085/packages/astro/src/types/public/integrations.ts#L50-L53)
  * [`PageBuildData`](https://github.com/withastro/astro/blob/4eb6fc9f9a0804783377dc7dd3e73cb35af68085/packages/astro/src/core/build/types.ts#L17-L23): I described the properties under "`pages` option" because this type is not importable
    * `key`: added in https://github.com/withastro/astro/pull/10625
    * `styles`: added in https://github.com/withastro/astro/pull/6659
  * [`SerializedSSRManifest`](https://github.com/withastro/astro/blob/4eb6fc9f9a0804783377dc7dd3e73cb35af68085/packages/astro/src/core/app/types.ts#L126-L144) (not importable, based on `SSRManifest`)
    * `componentMetadata`: added in https://github.com/withastro/astro/pull/6638
    * `clientDirectives`: added in https://github.com/withastro/astro/pull/7074
    * `serverIslandNameMap`: added in https://github.com/withastro/astro/pull/11305
    * `key`: added in https://github.com/withastro/astro/pull/11535
* Add docs for the following importable types:
  * [`SSRManifest`](https://github.com/withastro/astro/blob/34f84c2437fd078e299a29eeb1f931c9f83c8d2e/packages/astro/src/core/app/types.ts#L54-L99): used by `SerializedSSRManifest` and in the Adapter API
    * `hrefRoot`: added in https://github.com/withastro/astro/pull/11489
    * `userAssetsBase`: added in https://github.com/withastro/astro/pull/13003
    * `trailingSlash`: added in https://github.com/withastro/astro/pull/9085
    * `buildFormat`: added in https://github.com/withastro/astro/pull/9739
    * `compressHTML`: added in https://github.com/withastro/astro/pull/7488
    * `assetsPrefix`: added in https://github.com/withastro/astro/pull/6862
    * `clientDirectives`: added in https://github.com/withastro/astro/pull/7074
    * `inlinedScripts`: added in https://github.com/withastro/astro/pull/10102
    * `componentMetadata`: added in https://github.com/withastro/astro/pull/6638
    * `pageModule`: added in https://github.com/withastro/astro/pull/7220
      * `onRequest`: added in https://github.com/withastro/astro/pull/8300
    * `serverIslandMap`: added in https://github.com/withastro/astro/pull/11305
    * `serverIslandNameMap`: added in https://github.com/withastro/astro/pull/11305
    * `i18n`: added in https://github.com/withastro/astro/pull/8974
    * `middleware`: added in https://github.com/withastro/astro/pull/9776
    * `actions`: added in https://github.com/withastro/astro/pull/13262
    * `checkOrigin`: added in https://github.com/withastro/astro/pull/10678
    * `sessionConfig`: added in https://github.com/withastro/astro/pull/12441
    * `cacheDir`: added in https://github.com/withastro/astro/pull/13084
    * `srcDir`: added in https://github.com/withastro/astro/pull/13084
    * `outDir`: added in https://github.com/withastro/astro/pull/13084
    * `publicDir`: added in https://github.com/withastro/astro/pull/13084
    * `buildClientDir`: added in https://github.com/withastro/astro/pull/13084
    * `buildServerDir`: added in https://github.com/withastro/astro/pull/13084
    * `csp`: added in https://github.com/withastro/astro/pull/13802
    * `internalFetchHeaders`: added in https://github.com/withastro/astro/pull/14543
  * [`SSRLoadedRenderer`](https://github.com/withastro/astro/blob/337bca1fff2671c3a171f57e6757685d95f8c2c9/packages/astro/src/types/public/internal.ts#L203-L205): used by `SSRManifest`
  * [`SSRLoadedRendererValue`](https://github.com/withastro/astro/blob/337bca1fff2671c3a171f57e6757685d95f8c2c9/packages/astro/src/types/public/internal.ts#L181-L201): used by `SSRManifest`
    * `supportsAstroStaticSlot`: added in https://github.com/withastro/astro/pull/7093
    * `renderHydrationScript`: added in https://github.com/withastro/astro/pull/6791
  * [`SSRManifestCSP`](https://github.com/withastro/astro/blob/337bca1fff2671c3a171f57e6757685d95f8c2c9/packages/astro/src/core/app/types.ts#L114-L123): used by `SSRManifest`, added in https://github.com/withastro/astro/pull/13802
  * [`SSRComponentMetadata`](https://github.com/withastro/astro/blob/967f69d5853d0ca429af4f12c5dbbb8836834b97/packages/astro/src/types/public/internal.ts#L288-L291): used by `SSRManifest` and `SerializedSSRManifest`
* Add docs for the following non-exported types:
  * [`RouteInfo`](https://github.com/withastro/astro/blob/967f69d5853d0ca429af4f12c5dbbb8836834b97/packages/astro/src/core/app/types.ts#L28-L39) described under `SSRManifest.routes`:
    * `styles`: added in https://github.com/withastro/astro/pull/6659
* Normalizes function notation in existing docs (e.g. `fileUrlToPath` > `fileUrlToPath()`)
* Adds links from the Adapter API reference to the Integrations API reference

**Note:** we still have some links to the source code in `addRenderer() option`, but for examples so this probably fine to keep them:
> Examples: [svelte](https://github.com/withastro/astro/blob/main/packages/integrations/svelte/src/index.ts), [react](https://github.com/withastro/astro/blob/main/packages/integrations/react/src/index.ts), [preact](https://github.com/withastro/astro/blob/main/packages/integrations/preact/src/index.ts), [vue](https://github.com/withastro/astro/blob/main/packages/integrations/vue/src/index.ts), [solid](https://github.com/withastro/astro/blob/main/packages/integrations/solid/src/index.ts)

Another thing I haven't done is provide usage examples. This is not always obvious with types and not really easy with nested properties without becoming too verbose... This could be an improvement for later because this PR already adds a lot of contents? I mean, NWTWWHB.

**TODO once merged**: check what changed in v6 (e.g. https://v6.docs.astro.build/en/guides/upgrade-to/v6/#changed-ssrmanifest-interface-structure-adapter-api) and update v6 docs

- Suggested label: add new contents


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -873,7 +873,7 @@ The loader can then access the `refreshContextData` property to get the webhook
 
 <p>
 
-**Type:** [`AddressInfo`](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules__types_node_net_d_._net_.addressinfo.html)
+**Type:** `AddressInfo`
```
sarah11918, line 876: Noting that this seems to be the corresponding updated link from Microsoft, which would mean not linking to Nodejs.org twice. Do we want this? Is it helpful? https://learn.microsoft.com/en-us/javascript/api/botframework-streaming/addressinfo

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -977,6 +977,57 @@ export default {
...
+
+###### `styles`
+
+<p>
+
+**Type:** `Array<{ depth: number; order: number; sheet: { type: 'inline'; content: string } | { type: 'external'; src: string } }>`<br />
+<Since v="2.4.0" />
+</p>
+
+A list of styles to render on the page. Each style contains its `depth` and display `order` on the page, as well as an indication of whether it should be applied as an inline or external style.
```
sarah11918: I'm wondering whether "depth" and "order" for CSS styles should be inline code, since these are not accessible/configurable properties? And, I think it's understood that order matters in CSS, but I had to go read the PR to figure out exactly what depth was referring to. I don't think "CSS depth" is A Thing:tm:. *Maybe* they don't even need to be mentioned at all?

Taken from the code comment in that PR:
> A higher depth means that the CSS comes from shared subcomponents. A lower depth means it comes directly from the top-level page.
>
> Can be used to sort stylesheets so that shared rules come first and page-specific rules come after.

So, since `depth` is not something a user will ever see, I wonder if it makes more sense to just describe rather than "property-check by name" depth and order?

```suggestion
A list of styles to render on the page, with ordering based on your [`scopedStyleStrategy`](/en/reference/configuration-reference/#scopedstylestrategy), as well as an indication of whether they should be applied as an inline or external style.
```
I'm not sure whether this is entirely accurate, but since `depth` really refers to `scopedness` (which the user CAN control, and read about), maybe a description closer to this is helpful? Is there something *like this* we can say?

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1071,6 +1122,77 @@ export default {
...
+
+###### `componentMetadata`
+
+<p>
+
+**Type:** <code>[string, <a href="#ssrcomponentmetadata">SSRComponentMetadata</a>][]</code><br />
+<Since v="2.1.7" />
+</p>
+
+Defines an array of key-value pairs where each entry is a tuple. The first element is the component identifier and the second is an object describing the build metadata.
```
sarah11918:
```suggestion
Defines an array of key-value pairs where the first element is the component identifier and the second is an object describing the build metadata.
```
I'm wondering whether it's helpful to define as a tuple for these? What helpful information/meaning do we lose if we just say this?

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1071,6 +1122,77 @@ export default {
...
+
+###### `clientDirectives`
+
+<p>
+
+**Type:** `[string, string][]`<br />
+<Since v="2.5.0" />
+</p>
+
+Defines an array of key-value pairs where each entry is a tuple. The first element is the directive name (e.g., `load`, `visible`) and the second is the directive's implementation code.
```
sarah11918:
```suggestion
Defines an array of key-value pairs where each entry is a tuple. The first element is the directive name (e.g. `load`, `visible`) and the second is the directive's implementation code.
```
Note that I did think about just listing them all, since there aren't that many of them. But, you can create your own custom client directives, in which case this list wouldn't be exhaustive. So, a couple of examples seems fine!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1146,7 +1268,7 @@ export default {
 **Type:** [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL)
 </p>
 
-A URL path to the build output directory. Note that if you need a valid absolute path string, you should use Node's built-in [`fileURLToPath`](https://nodejs.org/api/url.html#urlfileurltopathurl-options) utility.
+A URL path to the build output directory. Note that if you need a valid absolute path string, you should use Node's built-in [`fileURLToPath()`](https://nodejs.org/api/url.html#urlfileurltopathurl-options) utility.
```
sarah11918:
```suggestion
A URL path to the build output directory.

If you need a valid absolute path string, use the Node.js built-in [`fileURLToPath()`](https://nodejs.org/api/url.html#urlfileurltopathurl-options) utility.
```
I know you were only correcting the function syntax, but I don't necessarily trust the original wording here given these are older docs. I already took out "should", but I don't know the actual original intention.

As written, this implies that you might not always need a valid absolute path string... but in the cases when you do, use Node's utility. If in fact you ALWAYS need this and `dir` is only used in this way, then I would change the wording above further to be something like, "An absolute path can be generated using the Node.js built-in utility..." (which removes the "if you're going to use it like this" when you're pretty much always going to use it like this). AND, in that case a line break won't be as necessary. (Again, right now the second point is like "extra, bonus usage" and not the definition of the thing, so I prefer the greater separation between those two thoughts. If this is "how you're going to use it" then the two feel more related.)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1187,7 +1309,7 @@ export default {
...
```
sarah11918: See the comment above for the similar entry!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1344,9 +1475,108 @@ The example above will produce logs with `[astro-format]` by default, and `[astr
...
 ```
 
+### `AstroIntegrationMiddleware`
+
+<p>
+
+**Type:** `{ order: "pre" | "post"; entrypoint: string | URL; }`
+</p>
+
+Describes a [middleware added by an integration](#addmiddleware-option). This contains the following properties:
```
sarah11918: Just thinking aloud here that I'm not sure re: ending a sentence with a colon to describe an upcoming list when what follows is *not* a list, but lots of section headings.

I tried to see if we had precedence, and on the `astro:assets` page (thinking of the `<Image />` component properties) we say this:

> The `<Image />` component accepts the following listed properties and [responsive image properties](https://docs.astro.build/en/reference/modules/astro-assets/#responsive-image-properties) in addition to all properties accepted by the HTML  tag.

(ends with a period, full stop.)

Do you think we could get away with just not saying that? Does this flow well enough to just go right into the properties?

```suggestion
Describes a [middleware added by an integration](#addmiddleware-option).
```

(I'd thought about e.g. "Describes a middleware added by an integration to provide the necessary information to run it" but I'm not sure that last bit adds anything?)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1344,9 +1475,108 @@ The example above will produce logs with `[astro-format]` by default, and `[astr
...
+An object containing an [`onRequest()`](/en/reference/modules/astro-middleware/#onrequest) property defined with the project's middleware function when it exists.
+
+### `AstroRenderer`
+
+<p>
+
+**Type:** `{ name: string; clientEntrypoint?: string | URL; serverEntrypoint: string | URL; }`
+</p>
+
+Describes a [component framework renderer added by an integration](#addrenderer-option). This contains the following properties:
```
sarah11918: Same comment here? (and wherever it repeats) I wonder if we can get away without saying "This contains the following properties"?

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1344,9 +1475,108 @@ The example above will produce logs with `[astro-format]` by default, and `[astr
...
+
+**Type:** `string | URL`
+</p>
+
+Defines the import path of the code executed whenever the directive is used.
+
 ### `HookParameters`
 
-You can get the type of a hook’s arguments by passing the hook’s name to the `HookParameters` utility type. In the following example, a function’s `options` argument is typed to match the parameters of the `astro:config:setup` hook:
+You can get the type of a hook’s arguments by passing the hook’s name to the `HookParameters` utility type. In the following example, a function’s `options` argument is typed to match the parameters of the [`astro:config:setup` hook](#astroconfigsetup):
```
sarah11918:
```suggestion
You can get the type of a hook’s arguments by passing the hook’s name to the `HookParameters` utility type.

In the following example, a function’s `options` argument is typed to match the parameters of the [`astro:config:setup` hook](#astroconfigsetup):
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+Describes the build metadata of a component rendered by the server. This contains the following properties:
+
+#### `propagation`
+
+<p>
+
+**Type:** `'none' | 'self' | 'in-tree'`
+</p>
+
+A union of literals indicating whether the Astro runtime needs to wait for a component to render the head content:
```
sarah11918:
```suggestion
A description of how to render head content from this component, including whether the Astro runtime needs to wait for a component:
```
Feels like maybe we could come up with a more general description here, since the explanations below aren't ALL directly about waiting? Is there something like the above that works?

(It's just that "Do you wait?" "None!" "Self!" "In-tree"! don't feel like reasonable answers to this question. 😅 )

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+Determines whether the component contains the head content.
+
+### `SSRLoadedRenderer`
+
+<p>
+
+**Type:** `{ name: string; clientEntrypoint?: string | URL; ssr: SSRLoadedRendererValue; }`
+</p>
+
+Describes a loaded UI framework renderer. This is a subset of [`AstroRenderer`](#astrorenderer) with the following additional properties:
```
sarah11918: LOCKED AND LOADED! 😄 

I wonder whether there's a better word than "loaded" we can use, so that we don't circularly define "What's SSRLoadedRenderer?" "It's a loaded renderer!"

"Enabled" probably isn't quite right? "Available"? But maybe there's a word closer to that?

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+**Type:** `() => string`<br />
+<Since v="4.1.0" />
+</p>
+
+Returns a framework-specific hydration script that must be injected into the HTML before the first component that uses this renderer.
+
+### `SSRManifest`
+
+An object containing build configuration and project metadata that the server adapters use at runtime to serve on-demand rendered pages. This contains the following properties:
```
sarah11918: Just wanted to say that these are lovely definitions!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+Specifies the [configured prefix for Astro-generated asset links](/en/reference/configuration-reference/#buildassetsprefix).
+
+#### `renderers`
+
+<p>
+
+**Type:** <code><a href="#ssrloadedrenderer">SSRLoadedRenderer</a>[]</code>
+</p>
+
+An array containing information about each loaded UI framework renderer (e.g., React, Vue, Svelte).
```
sarah11918:
```suggestion
An array containing information about each loaded UI framework renderer (e.g. React, Vue, Svelte).
```
(This might be a French vs. English thing? 😄  If I'm wrong, then at least we're *consistently* wrong, because it's not our M.O. to include commas here.)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+#### `clientDirectives`
+
+<p>
+
+**Type:** `Map<string, string>`<br />
+<Since v="2.5.0" />
+</p>
+
+Defines a mapping of client directive names (e.g., `load`, `visible`) to their implementation code. This includes both [built-in client directives](/en/reference/directives-reference/#client-directives) and [custom client directives](/en/reference/directives-reference/#custom-client-directives).
```
sarah11918:
```suggestion
Defines a mapping of client directive names (e.g. `load`, `visible`) to their implementation code. This includes both [built-in client directives](/en/reference/directives-reference/#client-directives) and [custom client directives](/en/reference/directives-reference/#custom-client-directives).
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+#### `componentMetadata`
+
+<p>
+
+**Type:** <code>Map\<string, <a href="#ssrcomponentmetadata">SSRComponentMetadata</a>\></code><br />
+<Since v="2.1.7" />
+</p>
+
+Defines a mapping of component identifier to their build metadata. Each entry contains information about the `propagation` behavior and whether it contains head elements.
```
sarah11918:
```suggestion
Defines a mapping of component identifier to their build metadata. Each entry contains information about the [`propagation`](#propagation) behavior and whether it contains head elements.
```
If this is referring to the thing above, seems like a good place for a link!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+An [Astro middleware function](/en/reference/modules/astro-middleware/#onrequest) when defined in the user project.
+
+##### `renderers`
+
+<p>
+
+**Type:** <code><a href="#ssrloadedrenderer">SSRLoadedRenderer</a>[]</code>
+</p>
+
+An array of loaded UI framework renderer.
```
sarah11918:
```suggestion
An array of loaded UI framework renderers.
```
At the very least, needs an "s". If we manage to find an alternate text description of "loaded", then that could work here, too.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+#### `i18n`
+
+<p>
+
+**Type:** `SSRManifestI18n | undefined`<br />
+<Since v="3.5.0" />
+</p>
+
+Specifies the `i18n` resolved configuration when enabled in the project. This contains the following properties:
```
sarah11918:
```suggestion
Specifies the resolved [`i18n` configuration](/en/reference/configuration-reference/#i18n) when enabled in the project. This contains the following properties:
```
I think this is the proper word order? And, allows for better linking!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+##### `strategy`
+
+<p>
+
+**Type:** `"manual" | "pathname-prefix-always" | "pathname-prefix-other-locales" | "pathname-prefix-always-no-redirect" | "domains-prefix-always" | "domains-prefix-other-locales" | "domains-prefix-always-no-redirect"`
+</p>
+
+Defines the routing strategy used [based on the project configuration](/en/reference/configuration-reference/#i18nrouting). This determines how locales are handled in URLs and whether redirects occur.
```
sarah11918: I think where possible, we can make the link text actually include the config property. That avoids multiple links on this one page that all have the same link text, and is illustrative of the specific property in question (without making people click and go to another page to see which property we're talking about)

The following could be something either naming the property as a config option, or in words like: 

```suggestion
Defines the [i18n routing strategy](/en/reference/configuration-reference/#i18nrouting) configured. This determines how locales are handled in URLs and whether redirects occur.
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+Determines the [default locale configured in the project](/en/reference/configuration-reference/#i18ndefaultlocale).
+
+##### `fallback`
+
+<p>
+
+**Type:** `Record<string, string> | undefined`
+</p>
+
+Specifies a mapping of locales to their fallback locales as [defined in the project configuration](/en/reference/configuration-reference/#i18nfallback).
```
sarah11918:
```suggestion
Specifies a mapping of locales to their fallback locales as [configured in `i18n.fallback`](/en/reference/configuration-reference/#i18nfallback).
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+#### `actions`
+
+<p>
+
+**Type:** <code>() => Promise\<\{ server: Record\<string, <a href="/en/reference/modules/astro-actions/#actionclient">ActionClient</a>\>; \}\> | \{ server: Record\<string, <a href="/en/reference/modules/astro-actions/#actionclient">ActionClient</a>\>; \}</code><br />
+<Since v="5.4.2" />
+</p>
+
+An object, or a function that returns the same object, with a `server` property that maps action names to their callable functions.
```
sarah11918:
```suggestion
An object, or a function that returns an object, with a `server` property that maps action names to their callable functions.
```
We haven't *really* established something that this object would be the same as... so I think just this works fine!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+#### `checkOrigin`
+
+<p>
+
+**Type:** `boolean`<br />
+<Since v="4.6.0" />
+</p>
+
+Determines whether [origin checking is enabled in the configuration](/en/reference/configuration-reference/#securitycheckorigin).
```
sarah11918:
```suggestion
Determines whether [origin checking is enabled in the security configuration](/en/reference/configuration-reference/#securitycheckorigin).
```
Maybe makes this feel a little less circular? And gives a hint to what/why we're checking something?

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+#### `sessionConfig`
+
+<p>
+
+**Type:** <code><a href="/en/reference/configuration-reference/#session-options">SessionConfig\<TDriver\></a> & \{ driverModule?: () => Promise\<\{ default: () => unstorage.Driver \}\>; }</code><br />
+<Since v="5.1.0" />
+</p>
+
+An object containing the [resolved session configuration](/en/reference/configuration-reference/#session-options) and an additional `driverModule` property defining the module specifier.
```
sarah11918:
```suggestion
An object containing the [resolved session configuration](/en/reference/configuration-reference/#session-options) and an additional property defining the driver in use.
```
Like my earlier comment re: `depth` and `origin` in CSS, I'm not sure name checking it and explaining it as "defining the module specifier" is as helpful here as a more plain language approach when this isn't something people will directly interact with? The above might not be the totally correct way to say it, but is there something *like* that you think might work?

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+Specifies the headers that are automatically added to internal fetch requests made during rendering.
+
+### `SSRManifestCSP`
+
+<p>
+
+<Since v="5.9.0" />
+</p>
+
+Describes the [Content Security Policy configuration](/en/reference/experimental-flags/csp/). This contains the following properties:
```
sarah11918: So, I noticed this description sounds a lot like `csp` so I went to check how they are different, but then I noticed that `csp` is a `####` heading, and I tried to figure out what its parent is... and I got lost. 😅 

So this made me wonder whether we might at some point need a `#### parent.child` section heading strategy, because I think we have some real nesting in here that makes the parent hard to trace back to (these don't show up in the TOC anyway, so it's not like that would become a mess).

Then in saying that, I realized that *because* only the `<h3>`s are visible in the TOC, the proper one should be highlighted, no matter how far down the tree (or rabbit hole) we are.

Not sure what to do about all this, if anything, but just noting how I felt trying to navigate this page!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+#### `cspDestination`
+
+<p>
+
+**Type:** `'adapter' | 'meta' | 'header' | undefined`
+</p>
+
+A union of literals determining whether CSP directives should be injected as a `meta` element, as a response `header`, or by the [`adapter` when it supports setting response headers](/en/reference/adapter-reference/#experimentalstaticheaders).
```
sarah11918:
```suggestion
Specifies whether CSP directives should be injected as a `meta` element, as a response `header`, or by the [`adapter` when it supports setting response headers](/en/reference/adapter-reference/#experimentalstaticheaders).
```
Would this work? This might be one of those that's easier to define as "what it does" vs "what it *is*"... unless we say something like "A configuration that determines whether..." which I think would also be ok. (But the first is simpler.)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+Specifies a list of generated hashes for project scripts and [user-supplied hashes](/en/reference/experimental-flags/csp/#hashes) for external scripts.
+
+#### `scriptResources`
+
+<p>
+
+**Type:** `string[]`
+</p>
+
+Specifies a list of valid sources for project scripts.
```
sarah11918: Noting that I'm assuming this comes from both any configured script resources, as well as the runtime `insertScriptResource`, which is why maybe there's no link here.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1885,659 @@ A union of supported route types:
...
+
+#### `isStrictDynamic`
+
+<p>
+
+**Type:** `boolean`
+</p>
+
+Determines whether support for dynamic script injection is [enabled in the configuration](/en/reference/experimental-flags/csp/#strictdynamic).
```
sarah11918:
```suggestion
Determines whether support for [dynamic script injection is enabled in the configuration](/en/reference/experimental-flags/csp/#strictdynamic).
```
Just making the link text more explicit!

sarah11918 (commented):
I can't believe I finished this.

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Yes, this seems to be the same interface but this is used in the "botframework-streaming package" which doesn't sound related. So, this link would be more confusing than helpful I think?

But maybe, in the description, the current link to the module is not really useful and we should link directly to the relevant section instead. Something like:
```diff
-The address, family and port number supplied by the [Node.js Net module](https://nodejs.org/api/net.html).
+The address, family and port number supplied by the [`server.address()` method of the Node.js Net module](https://nodejs.org/api/net.html#serveraddress).
```

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: How this is useful for integrations authors... honestly, I'm not sure. Maybe if an integration wants to push some styles in a page? I don't think this is read-only since this is in the `astro:build:setup` ("this is your final chance to modify it. This can be useful for example to overwrite some defaults.").

> I'm wondering whether "depth" and "order" for CSS styles should be inline code, since these are not accessible/configurable properties?

Users can access them. `styles` is an array and each individual style is an object typed with `{ depth: number; order: number; sheet: { type: 'inline'; content: string } | { type: 'external'; src: string } }` (so we have `depth` and `order`). So an user could map over the existing `styles` or push a new one?

> since depth really refers to scopedness

I'm not entirely sure this is related to `scopedStyleStrategy`. I'd say this is more related to the project structure/the components tree? It doesn't matter if the user uses `attribute` or `where`, the nested components will still have a higher depth than the current component/page. At least, that's my understanding.

But, yes, we could better explain what is depth to avoid any confusion! Maybe:
```suggestion
A list of styles to render on the page. Each style contains its `depth` in the components tree and its display `order` on the page, as well as an indication of whether this should be applied as an inline or external style.
```

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Ah yeah, I never know when a comma is required or not. [The Chicago Manual of Style](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Abbreviations/faq0047.html) says to use a comma after, but [Merriam-Webster](https://www.merriam-webster.com/grammar/ie-vs-eg-abbreviation-meaning-usage-difference) says "is also usually followed by a comma" which implies this is not always the case. 😅 

So, I'm fine with removing it for consistency! And maybe we can reword based on your previous suggestion (ie. I don't think "tuple" is necessarily helpful here):
```suggestion
Defines an array of key-value pairs where the first element is the directive name (e.g. `load`, `visible`) and the second is the directive's implementation code.
```

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: Thanks for the updated explanation! I couldn't find where we defined `depth` for an integration author, so I assumed it was something that was autocalculated by Astro or something and not publicly exposed.

If it's something they can actually access and as you say override in the "last chance"... do we need a proper API entry for it itself? (and I guess also `order`)?

But, your updated suggestion looks good to me!

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: Was also going to joke, we could link to the webarchive of the old page 😆

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: It would be my fault 😱  It seems I wrote that in https://github.com/withastro/docs/pull/10822 😅 and I wasn't adding links for everything as I do now, so... I don't remember why. I must have found it somewhere because I'm sure I didn't invent it...  (ie. maybe I took inspiration from a JSDoc or PR somewhere) but I can't remember where it was.

But reading that again, what you're saying make sense! Given the example usage, I think this is poorly phrased. The idea was probably the user shouldn't use `${dir.pathname}/custom-file.js` if they want to use the `dir` option and instead rely on `fileUrlToPath()` to compute a valid path.

Maybe something like the following would be better?

```suggestion
A URL path to the build output directory.

The following example uses Node's built-in [`fileURLToPath()`](https://nodejs.org/api/url.html#urlfileurltopathurl-options) utility to compute a valid absolute path string for a file provided by the integration:
```

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: I think we currently have a mix around this in docs:
* at the end of the [`<Picture />` section](https://docs.astro.build/en/reference/modules/astro-assets/#picture-) we have a colon before the headings:

> <Picture /> accepts all the properties of [the <Image /> component](https://docs.astro.build/en/reference/modules/astro-assets/#image-), including [responsive image properties](https://docs.astro.build/en/reference/modules/astro-assets/#responsive-image-properties), plus the following:

* in `astro:middleware`, we also use a colon in [`CreateContext`](https://docs.astro.build/en/reference/modules/astro-middleware/#createcontext-1)

> An object to [create a context](https://docs.astro.build/en/reference/modules/astro-middleware/#createcontext) to be passed to an Astro middleware. This contains the following properties:

But, I agree with you. I don't think this is necessary, and your suggestion is enough!

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Ah, good call! I'm updating your comment to fix the typo and I'll merge this with all the other simple fixes once I'm done checking all your suggestions. 😄

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: I updated your suggestion because `defined configured` sounds odd, I think you forgot to remove `defined`!

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: > a `#### parent.child` section heading strategy

I think that makes sense! I was wondering the same... because each time we add new sections in between, we need to check if internal links are not broken.

For example, if we already have a `src` heading and we add a new section with the same name before, then the original `src` anchor becomes `#src-1` and so any links to `#src` is now wrong and we need to think about checking/updating every links...).

This could also make shared links more readable (ie. `#parentsrc` instead of `#src-1`).

So, I think this is a good call, and this would make every pages consistent with our Configuration Reference. But it would probably be better in a follow up PR to avoid a bigger diff? 😅 Or I can do that for the new sections, and update the existing ones later...

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Yes, but maybe we could reword that to be able to include links and be more descriptive.
```suggestion
Specifies a list of valid sources combining the [configured resources](/en/reference/experimental-flags/csp/#resources) and the [injected script resources](/en/reference/experimental-flags/csp/#cspinsertscriptresource).
```

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: > I assumed it was something that was autocalculated by Astro

Well, this is true for the default value! But because this is writable, I assume an integration could inject a style here (probably using `depth: 0` and `order: styles[styles.length - 1].order`). Now, I don't have any use cases in mind... I'm just saying this might be doable.

> do we need a proper API entry for it itself? (and I guess also order)?

The type for the individual `style` is not exported, I had to flatten it. So, while this can be done, maybe that's not an intended usage and we don't need to expand the description with a proper entry? I don't know honestly, maybe a magician would have a different opinion... 😄

sarah11918: Alright, let's just go with your updated suggestion then!

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: You can see that I wasn't inspired... 😅 While searching how this is used in core to find something better, I noticed this is also used to create an error if the MDX renderer is not available. So "UI framework" is not accurate.

> Describes a renderer available for the server to use.
> Describes a renderer accessible by the server in order to render a page.

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: I think that suggestion is pretty clear, yeah!

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: Yeah, something about a renderer that is available, accessible, ready to use... I *wonder* about "configured", since you'd need to install an integration for any of those renderers in question (UI frameworks or MDX)? (Unless of course this is also true for Markdown/`.astro` rendering, which doesn't need configuring and is available by default...)

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: oh, totally could be a follow up thing! That would be easier to review, too 😅 . Then, we could compare head-to-head and think about whether it's helpful.

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: I thought about that too, and almost suggested it. But we're very near the end and I got tired. 😅 

I also know that these links will change for v6, but certainly having them here, then breaking them, is a good way to remind us to update them. (Vs saying we'll wait and do it, but there's no failing test to force us to.) I do think the links and extra description here is helpful!

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Yeah, I wasn't quite sure how this can be used... But I guess there is two different usage here: in an integration or an adapter. Maybe an integration wouldn't access it but an adapter would?
But in both cases, I think your suggestion is fine and ready for the next batch! 😄

sarah11918:
OK @ArmandPhilippot I think I addressed all the comments you left me on my comments! Is there anything still to discuss or figure out, or is it all just a matter of updating now?

Thank you again for this AMAZING piece of work! I was actually *smiling* while reading some of it, it's sooooo much better than it was, and good in its own right!

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: I didn't see anything about Markdown, but I saw [`astro:jsx`](https://github.com/withastro/astro/blob/d8e64ef77ef364b1541a5d192bcff299135d3bc8/packages/astro/src/runtime/server/render/component.ts#L118) which I assume means the Astro renderer can be included here and this is not something users install separately. But I could be wrong.

ArmandPhilippot:
Thank you @sarah11918 , I know this was a big one to review! I think I have everything I need. 🙌🏽

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: I'm adding a similar suggestion here to apply it with the 2nd batch!
```suggestion
A URL path to the build output directory.

The following example uses Node's built-in [`fileURLToPath()`](https://nodejs.org/api/url.html#urlfileurltopathurl-options) utility to compute a valid absolute path string for a file provided by the integration before writing to it:
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1891,659 @@ A union of supported route types:
...
+Defines the name of the [server adapter](/en/guides/on-demand-rendering/#server-adapters) used for on-demand rendering.
+
+#### `routes`
+
+<p>
+
+**Type:** `RouteInfo[]`
+</p>
+
+A list of information about the routes available in this project. Each entry contains the following properties.
```
ArmandPhilippot, line 2023: This is the only "following properties" I kept because `routes` is an array and the subheadings are related to a single `RouteInfo`.

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1071,6 +1122,77 @@ export default {
 }
 ```
 
+##### Serialized SSR manifest properties
+
+A serialized version of the [`SSRManifest`](#ssrmanifest), accessible through the [`astro:build:ssr` hook](#astrobuildssr). This contains the same information as `SSRManifest`, with some properties converted to serializable formats.
```
ArmandPhilippot, line 1127: This has been reworded to avoid "the following properties:".

ArmandPhilippot:
@sarah11918  I think I addressed everything this time! I left two comments related to my manual edits to avoid you searching in the whole file. 😅 But feel free to reread everything if you get bored. 😁 (otherwise I made two separated commits to ease the reviewing)

I choose to not use "configured" because as I said it seems the JSX renderer for Astro can be in the `renderers` array, and maybe you're right Markdown could be there too!

src/content/docs/en/reference/integrations-reference.mdx
sarah11918: Love this!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1891,659 @@ A union of supported route types:
...
+Specifies a list of generated hashes for project scripts and [user-supplied hashes](/en/reference/experimental-flags/csp/#hashes) for external scripts.
+
+#### `scriptResources`
+
+<p>
+
+**Type:** `string[]`
+</p>
+
+Specifies a list of valid sources combining the [configured resources](/en/reference/experimental-flags/csp/#resources) and the [injected script resources](/en/reference/experimental-flags/csp/#cspinsertscriptresource).
```
sarah11918:
```suggestion
Specifies a list of valid sources combining the [configured script resources](/en/reference/experimental-flags/csp/#resources) and the [injected script resources](/en/reference/experimental-flags/csp/#cspinsertscriptresource).
```
I like the clarification here!

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1891,659 @@ A union of supported route types:
...
+Specifies a list of generated hashes for project styles and [user-supplied hashes](/en/reference/experimental-flags/csp/#hashes) for external styles.
+
+#### `styleResources`
+
+<p>
+
+**Type:** `string[]`
+</p>
+
+Specifies a list of valid sources for project styles.
```
sarah11918:
```suggestion
Specifies a list of valid sources combining the [configured style resources](/en/reference/experimental-flags/csp/#resources) and the [injected script resources](/en/reference/experimental-flags/csp/#cspinsertscriptresource).
```
Should this oen be updated, too?

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1891,659 @@ A union of supported route types:
...
+Specifies the [configured prefix for Astro-generated asset links](/en/reference/configuration-reference/#buildassetsprefix).
+
+#### `renderers`
+
+<p>
+
+**Type:** <code><a href="#ssrloadedrenderer">SSRLoadedRenderer</a>[]</code>
+</p>
+
+A list of renderers (e.g. React, Vue, Svelte) available for the server to use.
```
sarah11918:
```suggestion
A list of renderers (e.g. React, Vue, Svelte, MDX) available for the server to use.
```
Since we bothered to figure this out, I wonder if it makes sense to use MDX as one of the examples so this doesn't look like just UI frameworks?

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Ah good call, I thought about this then forgot! 😅

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1891,659 @@ A union of supported route types:
...
+
+#### `inlinedScripts`
+
+<p>
+
+**Type:** `Map<string, string>`<br />
+<Since v="4.5.0" />
+</p>
+
+Defines a mapping of scripts identifiers to their content for scripts that will be inlined in the HTML output.
```
yanthomasdev:
```suggestion
Defines a mapping of script identifiers to their content for scripts that will be inlined in the HTML output.
```

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1655,6 +1891,659 @@ A union of supported route types:
...
+
+#### `componentMetadata`
+
+<p>
+
+**Type:** <code>Map\<string, <a href="#ssrcomponentmetadata">SSRComponentMetadata</a>\></code><br />
+<Since v="2.1.7" />
+</p>
+
+Defines a mapping of component identifier to their build metadata. Each entry contains information about the [`propagation`](#propagation) behavior and whether it contains head elements.
```
yanthomasdev:
```suggestion
Defines a mapping of component identifiers to their build metadata. Each entry contains information about the [`propagation`](#propagation) behavior and whether it contains head elements.
```

PR #13032, opened 2026-01-11
fix: use danger in astro db page and remove it in heroku page

KTibow:
#### Description (required)

use the more severe `danger` instead of `caution` for a command that completely destroys the database

tested locally and danger was red while caution was yellow

- Closes #
- Suggested label:


ArmandPhilippot:
Thanks for the contribution but I'm not sure this is needed... Can you explain why you think a `danger` aside would be more visible than a `caution` one?

If a caution is not enough, I doubt `danger` will be more useful. And, in Astro Docs, we usually use a caution for the "data loss" context, see: https://contribute.docs.astro.build/reference/custom-components/#aside-component.

KTibow:
It's true that `caution` is much more used, but `danger` is already used once (on the heroku page to warn about heroku-static-buildpack) and it feels more appropriate to use a red box ("dangerous") than a yellow one (warns you but not universally applicable). If the proper scope would need to look at *all* `caution` blocks then it's fair to close this PR (and possibly change that danger to a caution?)

sarah11918:
It's true that we *rarely* use the `danger` aside, mostly because there was nothing you could really do in Astro that really counts as *that* dangerous until we introduced Astro DB :smile: 

This would possibly be the only time I could see us using it. It *is* a pretty dangerous situation, so for this one time, I'm OK with it.

That Heroku guide should totally NOT be using a `danger`. That should be a `note` at best, as it's not really *dangerous* at all, but maybe a caution at least makes people pause before following the instructions. I could go either way with that. If you'd like to also downgrade the Heroku aside in this PR, I'd be happy to accept that, too!

PR #13037, opened 2026-01-12
feat(fonts): family options

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -659,7 +659,7 @@ formats: ["woff2", "woff"]
 <p>
 
 **Type:** `Record<string, any>`<br />
-<Since v="5.16.10" />
+<Since v="5.16.12" />
 </p>
 
 An object to pass provider specific options. It is typed automatically based on the font family [provider](#provider):


src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -154,94 +154,139 @@ body {
 
 Astro exports a few built-in providers, based on [unifont](https://github.com/unjs/unifont/) providers. You can also [make a custom Astro font provider](#build-your-own-font-provider).
```
delucis, line 155: I know it’s not in this PR, but saw this while reviewing, so though I might as well suggest here too:

```suggestion
Fonts are loaded using a provider, which either downloads font files from a remote service or loads local font files from disk. Astro exports built-in providers for common services, based on [unifont](https://github.com/unjs/unifont/) providers. You can also [make a custom font provider](#build-your-own-font-provider).
```

Basically, “what is a provider?” is not necessarily super obvious currently as we have “Available providers: Astro exports […] providers, based on […] providers. You can also make a custom […] provider.” So I tried to add a tiny bit more context for people.

delucis: Another alternative suggestion:

```suggestion
Fonts are loaded using a provider, which either downloads font files from a remote service or loads local font files from disk. Astro exports built-in providers for common services. You can also [make a custom font provider](#build-your-own-font-provider).
```

Is it still even helpful to mention unifont if our provider syntax no longer 100% matches?

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -154,94 +154,139 @@ body {
...
 
-The following providers have built-in support:
-
-- [Adobe](https://fonts.adobe.com/)
-- [Bunny](https://fonts.bunny.net/)
-- [Fontshare](https://www.fontshare.com/)
-- [Fontsource](https://fontsource.org/)
-- [Google](https://fonts.google.com/)
-- [Google Icons](https://fonts.google.com/icons)
+To use a built-in remote provider, import `fontProviders` and configure [`provider`](#provider) with the appropriate value for your chosen font provider:
```
delucis, line 157: IIRC this will eventually also cover the local provider right? Just realising my first comment assumed that, but I guess it hasn’t been implemented yet.

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -154,94 +154,139 @@ body {
...
+      glyphs: ["a"]
+    }
+  }
+}
 ```
 
-</TabItem>
-
-</Tabs>
+Only available when resolving the new Material Symbols icons.
```
delucis: This seems important, so might be better to move it close to the heading?

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -981,16 +1143,73 @@ export default defineConfig({
...
+});
+```
+
+</TabItem>
+
+</Tabs>
+
 ### Supporting a 3rd-party unifont provider
 
 You can define an Astro font provider using a unifont provider under the hood:
```
delucis, line 1180: Probably not high priority given we’ve concluded almost none of these actually exist, but it looks like we could potentially provide a built-in “adapter” rather than people having to wire it up manually.
```js
provider: unifontToAstro(acmeProvider()),
```
Just a thought.

florian-lefebvre: We talked about this in various places FYI, quoting myself from the concerned PR:

> I considered exposing some kind of unifontToAstroFontProvider() util but I don't think it's worth doing. I'm not aware of any public 3rd-party unifont providers so I don't see the benefit for us to maintain such thing when we can just show an example in docs ([deploy-preview-13025--astro-docs-2.netlify.app/en/reference/experimental-flags/fonts#supporting-a-3rd-party-unifont-provider](https://deploy-preview-13025--astro-docs-2.netlify.app/en/reference/experimental-flags/fonts/#supporting-a-3rd-party-unifont-provider))

florian-lefebvre:
I think the comments about the local provider make sense but yeah it indeed isn't the right PR for them. So I'm gonna mark them as resolved for this PR and look at them when I unify providers later on this week or next week

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -154,94 +154,139 @@ body {
...
 
 ```js
-provider: fontProviders.googleicons({
-	glyphs: {
-		"Material Symbols Outlined": ["arrow_right", "favorite", "arrow_drop_down"]
-	}
-})
+{
+  // ...
+  provider: fontProviders.google(),
```
florian-lefebvre:
```suggestion
  provider: fontProviders.googleicons(),
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
florian-lefebvre: Not sure how to blend it properly with the existing sentences, but sure! FYI I copied it from unifont docs, where if was below the codeblock

src/content/docs/en/reference/experimental-flags/fonts.mdx
delucis: > I'm not aware of any public 3rd-party unifont providers so I don't see the benefit for us to maintain such thing

I wonder if there’s any point in us maintaining docs as well then? Doesn’t have to be decided in this PR, but I did read this section thinking a bit that no-one is ever going to be faced with needing to do this. (And if they do want to, it’s not impossible to figure it out just with the existing provider docs.)

florian-lefebvre: We talked about it during last T&D, and the conclusion was (IIRC): there are only 2 main usecases for this API which are private registries and unifont providers. Both are gonna be pretty niche anyways.

I personally think it's worth keeping (unifont is great!), it requires basically no maintenance from us.

src/content/docs/en/reference/experimental-flags/fonts.mdx
sarah11918, line 157: Especially if we're removing the convenient tabs, we should keep the list so people will be able to see at a glance what options are available. This way, it's a loooot of scrolling right now just even to learn what's available, let alone see instructions.


(Just checking, is there a good reason for changing from the tab set up? Are there just too many of them now? Did you think the Google one looked too long with more info? I kind of liked it!)

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -154,94 +154,139 @@ body {
...
+The provider comes with family specific [options](#options).
+
+#### experimental.glyphs
+
+<p>
+
+**Type:** `string[]`
+</p>
+
+Allows specifying a list of glyphs to be included in the font for each font family. This can reduce the size of the font file:
```
```suggestion
when resolving the new Material Symbols icons, allows specifying a list of glyphs to be included in the font for each font family. This can reduce the size of the font file:
```
Would this address Chris's comment maybe? And, then, show in the code blocks that Material Symbols is used, if it's required?

src/content/docs/en/reference/experimental-flags/fonts.mdx
sarah11918: (see my comment above)

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -602,6 +648,24 @@ An array of [font formats](https://developer.mozilla.org/en-US/docs/Web/CSS/Refe
 formats: ["woff2", "woff"]
 ```
 
+#### options
+
+<p>
+
+**Type:** `Record<string, any>`<br />
+<Since v="5.16.10" />
```
```suggestion
<Since v="5.16.12" />
```
Maybe!

src/content/docs/en/reference/experimental-flags/fonts.mdx
```suggestion
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
florian-lefebvre: We can keep the list there if we want, no strong opinion.

About tabs, I think it was just not as good as proper headings. Now that some providers have specific family options etc (eg. local fonts with `variants` soon-ish), it just makes sense to have headings. Also it shows in the TOC which is good IMO

florian-lefebvre:
```suggestion
To use a built-in remote provider, import `fontProviders` and configure [`provider`](#provider) with the appropriate value for your chosen font provider:

- [Adobe](#adobe)
- [Bunny](#bunny)
- [Fontshare](#fontshare)
- [Fontsource](#fontsource)
- [Google](#google)
- [Google Icons](#google-icons)

To use a built-in remote provider, import `fontProviders` and configure [`provider`](#provider) with the appropriate value for your chosen font provider:
```
(From Sarah: Just updated the first line because I think the existing text was actually helpful, and this introduces the "built-in" list!

sarah11918: I updated to combine the new text (which is good!) and the old list (which is also good!) What do you think about that combination now, as it stands?

florian-lefebvre: I'm confused, the edited suggestion starts and ends with the same exact sentence

sarah11918: Yeah, the only change here ends up being to now to remove "The following providers have built-in support:" and replace it with:

> To use a built-in remote provider, import `fontProviders` and configure [`provider`](#provider) with the appropriate value for your chosen font provider:

(Because the list comes back, and hasn't changed.) I think it all works out if you accept it, but if not, you can just update manually!

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -154,94 +154,147 @@ body {
...
-- [Google Icons](https://fonts.google.com/icons)
+- [Adobe](#adobe)
+- [Bunny](#bunny)
+- [Fontshare](#fontshare)
+- [Fontsource](#fontsource)
+- [Google](#google)
+- [Google Icons](#google-icons)
 
-To use a built-in remote provider, configure `provider` with the appropriate value for your chosen font provider:
+To use a built-in remote provider, import `fontProviders` and configure [`provider`](#provider) with the appropriate value for your chosen font provider:
```
sarah11918: Oh, I see what you mean now! the GitHub diff looked off. Sorry, I see this now! OK, lemme suggest fresh now.

PR #13047, opened 2026-01-14
feat(fonts)!: fontData instead of getFontData()

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/fonts.mdx
+++ b/src/content/docs/en/reference/experimental-flags/fonts.mdx
@@ -44,10 +44,10 @@ Then, add the `<Font />` component and site-wide styling in your `<head>`:
 
 ```astro title="src/components/Head.astro" ins={2,5,8-10}
 ---
-import { Font } from 'astro:assets';
+import { Font } from "astro:assets";
 ---
 
-<Font cssVariable='--font-roboto' preload />
+<Font cssVariable="--font-roboto" preload />
 
 <style>
 body {
@@ -90,7 +90,7 @@ body {
 
     ```astro title="src/components/Head.astro" ins={2, 5}
     ---
-    import { Font } from 'astro:assets';
+    import { Font } from "astro:assets";
     ---
 
     <Font cssVariable="--font-roboto" preload />
@@ -119,7 +119,7 @@ body {
     <TabItem label="Tailwind CSS 4.0">
 
     ```css title="src/styles/global.css" ins={4} ins="inline"
-    @import 'tailwindcss';
+    @import "tailwindcss";
 
     @theme inline {
         --font-sans: var(--font-roboto);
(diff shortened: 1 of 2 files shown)


src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -385,26 +385,26 @@ Variable weight font files will be preloaded if any weight within its range is r
 
 ## Accessing font data programmatically
 
-The `getFontData()` function is intended for retrieving lower-level font family data programmatically, for example, in an [API Route](/en/guides/endpoints/#server-endpoints-api-routes) or to generate your own meta tags.
+The `fontData` object is intended for retrieving lower-level font family data programmatically, for example, in an [API Route](/en/guides/endpoints/#server-endpoints-api-routes) or to generate your own meta tags.
```
sarah11918: An object doesn't "retrieve" data, the way a function does, so this feels slightly off (even if it's *where* you retrieve the info from). We could maybe update this sentence (not just change the function to an object) to describe the role the object plays?

Maybe something like the `fontData` object *contains* or *allows you to access/retrieve*? What do you think here?

```suggestion
The `fontData` object allows you to retrieve lower-level font family data programmatically, for example, in an [API Route](/en/guides/endpoints/#server-endpoints-api-routes) or to generate your own meta tags.
```

sarah11918: Sure, I think this works! (Sorry, noticing it's only a draft PR. I saw a short one I could review while waiting for the Cloudflare guide to update branch and the deploy preview to run! 😄 )

sarah11918:
(As I think I mentioned elsewhere, I didn't realize this was a draft when I reviewed it! I saw how tiny it was and thought I could squeeze it in waiting for other update branch/checks. 😅  So, I'll "un-approve" it, but only because it's not done!

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -385,26 +385,26 @@ Variable weight font files will be preloaded if any weight within its range is r
...
+The `fontData` object allows you to retrieve lower-level font family data programmatically, for example, in an [API Route](/en/guides/endpoints/#server-endpoints-api-routes) or to generate your own meta tags.
 
-### `getFontData()`
+### `fontData`
 <p>
 
-**Type:** `(cssVariable: CssVariable) => FontData[]`<br />
-<Since v="5.14.0" />
+**Type:** `Record<CssVariable, FontData>`<br />
+<Since v="5.16.10" />
```
```suggestion
<Since v="5.16.12" />
```
Assuming this will be 12 by now!

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -385,26 +385,26 @@ Variable weight font files will be preloaded if any weight within its range is r
...
 </p>
 
-Returns an array of `FontData` objects for the provided [`cssVariable`](#cssvariable-1), which contains `src`, `weight` and `style`.
+Contains an array of `FontData` objects by [`cssVariable`](#cssvariable-1), which contains `src`, `weight` and `style`.
```
sarah11918: I don't understand the word "by" in this context?

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -385,26 +385,26 @@ Variable weight font files will be preloaded if any weight within its range is r
...
-The `getFontData()` function is intended for retrieving lower-level font family data programmatically, for example, in an [API Route](/en/guides/endpoints/#server-endpoints-api-routes) or to generate your own meta tags.
+The `fontData` object allows you to retrieve lower-level font family data programmatically, for example, in an [API Route](/en/guides/endpoints/#server-endpoints-api-routes) or to generate your own meta tags.
 
-### `getFontData()`
+### `fontData`
 <p>
 
-**Type:** `(cssVariable: CssVariable) => FontData[]`<br />
-<Since v="5.14.0" />
+**Type:** `Record<CssVariable, FontData>`<br />
```
ArmandPhilippot: This is `FontDataRecord`, right? If so, it seems this should be:
```suggestion
**Type:** `Record<CssVariable, Array<FontData>>`<br />
```
(`export type FontDataRecord = Record<string, Array<FontData>>;`)

src/content/docs/en/reference/experimental-flags/fonts.mdx
ArmandPhilippot: Maybe with my suggestion for the type this makes a bit more sense?
Something like:
> An object where each key is a [`cssVariable`](#cssvariable-1) and the value is an array describing the associated fonts. Each entry contains the properties `src`, `weight`, and `style`.

```suggestion
An object where each key is a [`cssVariable`](#cssvariable-1) and the value is an array describing the associated fonts. Each entry contains the `src`, `weight`, and `style` properties.
```

src/content/docs/en/reference/experimental-flags/fonts.mdx
```diff
@@ -385,26 +385,26 @@ Variable weight font files will be preloaded if any weight within its range is r
...
 <p>
 
-**Type:** `(cssVariable: CssVariable) => FontData[]`<br />
-<Since v="5.14.0" />
+**Type:** `Record<CssVariable, Array<FontData>>`<br />
+<Since v="5.16.12" />
 </p>
 
-Returns an array of `FontData` objects for the provided [`cssVariable`](#cssvariable-1), which contains `src`, `weight` and `style`.
+An object where each key is a [`cssVariable`](#cssvariable-1) and the value is an array describing the associated fonts. Each entry contains the `src`, `weight`, and `style` properties.
```
ArmandPhilippot: I noticed we didn't specify this before, but `src` is an array. Should we mention it explicitly?

```suggestion
An object where each key is a [`cssVariable`](#cssvariable-1) and the value is an array describing the associated fonts. Each font is an object containing an array of `src` available for that font and the following optional properties: `weight` and `style`.
```

ArmandPhilippot (approved):
I left a nit regarding `src`, because we're using it in the code snippet and maybe it would be better to explicitly specify this is an array in the description?

And, I guess at one point we'll need to document `format` and `tech` too. But this is NWTWWHB, so LGTM. 😄 
```ts
export interface FontData {
	src: Array<{ url: string; format?: string; tech?: string }>;
	weight?: string;
	style?: string;
}
```

PR #13064, opened 2026-01-17
i18n(ja): update guides `dev-toolbar.mdx`

Yusaku01:
#### Description (required)

This PR's change is based on the following some commit:

- https://github.com/withastro/docs/commits/main/src/content/docs/en/guides/dev-toolbar.mdx?since=2025-02-27T11:29:15.000Z

"Developer" was changed to "Dev," but writing it as "デブ" (debu) in Japanese would be unfamiliar to readers and could cause misunderstanding, so I translated it as "開発" (development) instead.

- Closes #
- Suggested label: i18n

I'd appreciate it if you could review this when you have time🫡


jp-knj (approved):
LGTM 👍🏿

morinokami (approved):
LGTM! Thanks @Yusaku01 & @jp-knj 🚀

PR #13070, opened 2026-01-18
docs(react): add react compiler enablement instructions

erbierc:
#### Description (required)

I have been working on adding a config option for Astro's React integration which would enable React Compiler (withastro/astro#14955). This will be a draft for now, just to show a preview. I will add the config options after we get through the types.

- Closes #N/A
- Suggested label: add new content, minor release

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/react.mdx
+++ b/src/content/docs/en/guides/integrations-guide/react.mdx
@@ -280,6 +280,29 @@ export default defineConfig({
 });
 ```
 
+### Enable React Compiler
+
+Astro comes with an option to enable [React Compiler](https://react.dev/learn/react-compiler).
+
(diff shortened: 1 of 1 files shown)


sarah11918:
I know this feature is still in draft, as are the docs. But just leaving a comment that the docs look pretty good, and in keeping with the rest of the page (as it currently is)! It would just need to explicitly mention `<Since />` so that it's clear when this feature was added. But, I think it's well written and fits our form right now, so once it's accurate, it's likely good to go.

Note for @ArmandPhilippot that yes, this page doesn't have a proper API reference section (yet) and is more use-case oriented, so I would say this kind of docs section is keeping with the rest of the page.

sarah11918:
Hi! Since we are very close to v6 and all new contributions must be for Astro version 6.x, including v6 docs, I'm going to close this PR.

When you have your feature ready for v6, you can open a new PR to the new version of docs!

PR #13073, opened 2026-01-19
fix(integrations-reference): reorganize headings

ArmandPhilippot:
#### Description (required)

As discussed in https://github.com/withastro/docs/pull/13021#discussion_r2692505870 some headings are duplicated and because of nesting, it can be hard to follow to which parent they belong. This PR tries to reorganize/rename those headings to maybe improve the readability.

* Uses function notation for `createCodegenDir` (consistency)
* Fixes the headings level for "Page build data" properties (h6 > h5)
* Removes `SSRManifestCSP` from importable types and moves the headings under `SSRManifest.csp` (I don't think the `SSRManifestCSP` is useful on its own and moving the headings helps keeping all the `csp` headings tied)
* Uses `parent.child` heading notation for nested properties (from h3 to h6). This is now consistent with what we do under [`route` option](https://docs.astro.build/en/reference/integrations-reference/#route-option) and in the Configuration reference (e.g. [`security`](https://docs.astro.build/en/reference/configuration-reference/#security))
* Moves the `SerializedSSRManifest` properties description one level up to make it easier to use `parent.child` headings there

I left two comments for parts where I'm not sure this is better. Other opinions on this would be welcome!

- Suggested label:  consistency/formatting, improve or update documentation

### DIFF
--- a/src/content/docs/en/editor-setup.mdx
+++ b/src/content/docs/en/editor-setup.mdx
@@ -12,7 +12,7 @@ Customize your code editor to improve the Astro developer experience and unlock 
 
 ## VS Code
 
-[VS Code](https://code.visualstudio.com/) is a popular code editor for web developers, built by Microsoft. The VS Code engine also powers popular in-browser code editors like [GitHub Codespaces](https://github.com/features/codespaces) and [Gitpod](https://gitpod.io/).
+[VS Code](https://code.visualstudio.com/) is a popular code editor for web developers, built by Microsoft. The VS Code engine also powers popular in-browser code editors like [GitHub Codespaces](https://github.com/features/codespaces).
 
 Astro works with any code editor. However, VS Code is our recommended editor for Astro projects. We maintain an official [Astro VS Code Extension](https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode) that unlocks several key features and developer experience improvements for Astro projects.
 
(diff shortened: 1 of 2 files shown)


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -2022,7 +2019,7 @@ Defines the name of the [server adapter](/en/guides/on-demand-rendering/#server-
 
 A list of information about the routes available in this project. Each entry contains the following properties.
 
-##### `routeData`
+##### `RouteInfo.routeData`
```
ArmandPhilippot, line 2022: One caveat I see with `parent.child` headings is for properties typed as an array...

* `SSRManifest.routes` is typed with `RouteInfo[]`, so we can't use `SSRManifest.routes.routeData` here
* `RouteInfo` is not exported, but maybe it's still fine if we use `RouteInfo.routeData` here (and for the other properties)

src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -981,7 +981,7 @@ export default {
 
 An object describing how to build a page.
 
-###### `key`
+##### `PageBuildData.key`
```
ArmandPhilippot: The heading level was wrong, but I wonder if `PageBuildData.key` is better than `key` (same for the other properties). Like my first comment, the `parent.child` notation is tricky with arrays and objects, at least when the type is not exported.

sarah11918:
Maybe @TheOtterlord and @florian-lefebvre can weigh in on Armand's questions here?

src/content/docs/en/reference/integrations-reference.mdx
florian-lefebvre: For this one, I wonder if it would make more sense to say "pages it of type `PageBuilData[]`" and then link to a new section for this type

src/content/docs/en/reference/integrations-reference.mdx
florian-lefebvre: I think it's fine

src/content/docs/en/reference/integrations-reference.mdx
ArmandPhilippot: Now, I see the heading levels were correct, the issue was with `Page build data`... 😅 I'll push a commit to fix that.

So yes, I agree that would be better, at least to avoid such nesting which is hard to follow! But, the thing is `PageBuildData` is not exported, so we can't really include it in "Integration types reference" where we only show importable types. With the current structure, it seems the best place is to describe that under "pages option".

But maybe, to make the subheadings less confusing, "Page build data" should be "The `PageBuildData` object" or something like that?

ArmandPhilippot:
Thank you both! I just pushed a new commit based on Florian's comment to:
* fix the headings level related to `PageBuildData`
* replace the "Page build data" heading with "The `PageBuildData` object" to make the subheadings less awkward
* reword the description to avoid "The PageBuildData object" > "An object..."

I think this is ready to be merge if those changes looks fine to you!

PR #3495, opened 2023-06-13
[fix] revised typo

CBID2:
#### What kind of changes does this PR include?

- Minor content fixes (broken links, typos, etc.)

#### Description

- Revises a typo in the contributing guide. 
- Closes #3496

PR #13078, opened 2026-01-20
chore: remove gitpod/ona mention, update idx -> firebase studio

erbierc:
#### Description (required)

I have been slowly deleting mentions of Gitpod from the Astro codebase (as Gitpod is now Ona, an AI-powered paid vibe coding tool) - here are some examples: #12642 [#3495](https://github.com/withastro/starlight/pull/3495) - and just noticed that editor-setup.mdx still mentions it. I also updated the mention of IDX, which is now Firebase Studio - I moved it next to other editors, such as Stackblitz, as they all seem very similar to each other, so a separate mention of Firebase Studio did not seem like it was needed anymore. Open to feedback though!

- Closes #N/A

### DIFF
--- a/src/content/docs/en/editor-setup.mdx
+++ b/src/content/docs/en/editor-setup.mdx
@@ -53,8 +53,9 @@ Our amazing community maintains several extensions for other popular editors, in
 
 In addition to local editors, Astro also runs well on in-browser hosted editors, including:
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Thanks @erbierc ! Looks like we missed something in the Great Gitpod Purge. 😄 (We were pretty focused on https://astro.new at the time.)

I would still like to keep the separate listing for Firebase Studio though (just updating the link) because it really *is* a bit more full-featured, and more like (Open) VS Code where you can install VS Code extensions etc. So, can you revert the deletion, but just update the link there? Then I think we'll be good to go! :rocket:

erbierc:
Done! Wonder if there's anything else to be purged, I might investigate that later

vrabe:
This relates to #11608

PR #13079, opened 2026-01-21
[next]: make runtime csp optional

florian-lefebvre:
#### Description (required)

I didn't include an entry in the upgrade guide because this is a breaking change affecting a feature stabilized during the alpha

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/api-reference.mdx
+++ b/src/content/docs/en/reference/api-reference.mdx
@@ -1054,7 +1054,7 @@ Loads a session by ID. In normal use, a session is loaded automatically from the
 
 <p>
 
-**Type**: `AstroSharedContextCsp`
+**Type**: `object | undefined`
 <Since v="6.0.0" />
 </p>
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1054,7 +1054,7 @@ Loads a session by ID. In normal use, a session is loaded automatically from the
 
 <p>
 
-**Type**: `AstroSharedContextCsp`
+**Type**: `object | undefined`
```
florian-lefebvre, line 1057: Not sure what pattern we have been following, the type is now inline in the source:

```diff
-csp: AstroSharedContextCsp
-type AstroSharedContextCsp = {}
+csp: {}
```
This was done before the associated code PR btw

ArmandPhilippot: When the type is not too long, we use `{ prop1: type; prop2: type }`, but otherwise to avoid types on 3+ lines, we use exactly what you did!

florian-lefebvre:
Technically not out yet but I'm afraid of forgetting to merge it, and it doesn't hurt

PR #13107, opened 2026-01-23
[next] stable fonts

florian-lefebvre:
#### Description (required)

Some questions left:

- I left some `TODO:`s in the configuration reference. In the experimental page, there were headings for "Granular font configuration" and "Caching" that I feel would belong where I put them. But I don't feel like I can easily make a section for them there. So yeah not sure what to do
- I updated the fonts guide and added the `Fonts` API at the end, matching what we've done for `astro:env`. However here I'm wondering if it wouldn't make sense to have it first as it's easier and more powerful (vs. the env guide which has a bunch more things that are absolutely mandatory first)

- Closes #
- Suggested label:


src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,723 @@
+---
+title: Astro Font Provider API
+sidebar:
+  label: Font Provider API
+i18nReady: true
+tableOfContents:
+  minHeadingLevel: 2
+  maxHeadingLevel: 6
```
sarah11918:
```suggestion
```
Looking at this page, I think our normal TOC is probably fine!

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,723 @@
...
+</p>
+
+The [Fonts API](/en/guides/fonts/#fonts-api) allows you to access fonts in a unified way. Each family requires the use of an Astro Font Provider, which either downloads font files from a remote service or loads local font files from disk.
+
+## Built-in providers
+
+Astro exports built-in font providers from `astro/config`:
+
+```js
+import { fontProviders } from 'astro/config'
```
sarah11918, line 22: Does this mean we need an entry on the `astro:config` page for `fontProviders` @ArmandPhilippot

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -2135,3 +2135,305 @@ export default defineConfig({
 })
 ```
 
+## Fonts
```
sarah11918:
```suggestion
## font
```
(since this is the option prefix)

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -2135,3 +2135,305 @@ export default defineConfig({
...
+
+<p>
+
+**Type:** `Array<FontFamily>`<br />
+**Default:** `[]`<br />
+<Since v="6.0.0" />
+</p>
+
+Configures the Fonts API, which allows you to use fonts from your filesystem and various font providers through a unified, fully customizable, and type-safe API.
```
sarah11918: Check the other here like i18n and env. This should probably only be closer to this one line. The stuff below until the next reference entry probably goes in the guide!

src/content/docs/en/reference/configuration-reference.mdx
florian-lefebvre: I think we don't have a similar case yet in the config reference. The top property is an array of objects (fonts) but we document properties of each object (font)

src/content/docs/en/reference/font-provider-reference.mdx
ArmandPhilippot: Another ping that GitHub saw fit not to notify me of... 🙄 

I guess we could have a new section for `astro/config` imports but this one is a bit tricky because it contains a mix of things documented elsewhere:

```ts
import { defineConfig, envField, fontProviders, getViteConfig, mergeConfig, passthroughImageService, sharpImageService } from 'astro/config'
```

I'm not sure how helpful it is? Maybe as an overview of what is available for someone who don't know how to find the appropriate page?
But since we are not displaying the `astro/config` exports at the moment, we probably should do that in another PR, with the others, so that it doesn't seem strange to have only one export here.

src/content/docs/en/reference/configuration-reference.mdx
sarah11918: Hmm... so maybe "Font Options" (like we have for some of the other entries?)

florian-lefebvre: Not sure, that still sounds like an object. What about "Font families"?

ArmandPhilippot: I don't have a strong opinion, which is why I didn't reply at the same time as my previous comment 😁  , but if it helps here's my thoughts:
* `fonts` (plural, lowercase) could make sense and this is consistent with the previous headings
* `Font Options` also makes sense to me both for consistency and because this is a configuration option: either you configure that option with an array, or you don't enable it.
* `Font families` alone doesn't sound right/consistent to me, but if Sarah's suggestion bothers you, `Font Family Options` could be a good middle ground?

But, in any cases, I can see how this might sound confusing. I wonder if it might be due to the subheadings? Because at first glance (if we don't pay attention to plural/singular), this looks like an object.

Maybe we need an intermediate subheading (e.g. "The font family object") in between? Not sure if this is easy to do in the JSDoc though...
Or maybe, instead of `font.provider` we should use `FontFamily.provider` to highlight that it belongs to the way the option is configured, but that `fonts` is not an object.

This reminds me my last PR where I struggled a bit to find the right name with arrays. 😅

florian-lefebvre: My only concern about `FontFamily.provider` is it doesn't look consistent with the rest of the page. I agree it is tricky. But yeah the more I think about it, the more I like `fonts` and `font.x` after all. Let me both know what you think!

ArmandPhilippot: This doesn't look consistent because the other are not arrays... you should have built an object instead. 😄  But, yes, `fonts` sounds fine to me!

src/content/docs/en/guides/fonts.mdx
```diff
@@ -8,34 +8,57 @@ description: >-
...
+            src: ['./src/assets/fonts/DistantGalaxy.woff2'],
+            weight: 'normal',
+            style: 'normal'
+          }]
+        }
+      }]
+    });
+    ```
+
+3. Include styles using the [`<Font />`](/en/reference/modules/astro-assets/#font-) component in the head of your page, usually in a layout:
```
ArmandPhilippot: nit: Maybe a link for those who do not know about layouts?
```suggestion
3. Include styles using the [`<Font />`](/en/reference/modules/astro-assets/#font-) component in the head of your page, usually in a [layout](/en/basics/layouts/):
```

src/content/docs/en/guides/fonts.mdx
```diff
@@ -8,34 +8,57 @@ description: >-
...
+
+    ```astro title="src/components/Head.astro"
+    ---
+    import { Font } from "astro:assets";
+    ---
+
+    <Font cssVariable="--font-distant-galaxy" preload />
     ```
```
ArmandPhilippot: Because you say "in the head of your page, usually in a layout", and some people are new with web dev/Astro, I think this should be a Layout component with a minimal page shell:

````suggestion
    ```astro title="src/components/Layout.astro"
    ---
    import { Font } from "astro:assets";
    ---

    <html>
      <head>
        <Font cssVariable="--font-distant-galaxy" preload />
      </head>
      <body>
        <slot />
      </body>
    </html>
   ```
````

File: src/content/docs/en/guides/fonts.mdx
Comment (ArmandPhilippot) [inline src/content/docs/en/guides/fonts.mdx]: I know you haven't updated that part, but I feel like if we want a working example, we should import that Layout here:
```suggestion
   ---
   import Layout from "../components/Layout.astro";
   ---

   <Layout>
     <h1>In a galaxy far, far away...</h1>

     <p>Custom fonts make my headings much cooler!</p>
   </Layout>
```
(might be better to update that manually, not sure GitHub took all the lines given the preview...)

src/content/docs/en/guides/fonts.mdx
```diff
@@ -47,90 +70,252 @@ This example will demonstrate adding a custom font using the font file `DistantG
...
+        provider: fontProviders.fontsource(),
+        name: "Roboto",
+        cssVariable: "--font-roboto",
+      }]
+    });
+    ```
+
+3. Include styles using the [`<Font />`](/en/reference/modules/astro-assets/#font-) component in the head of your page, usually in a layout:
+
+    ```astro title="src/components/Head.astro"
```
ArmandPhilippot: We might want the same changes suggested before for the FontSource step?
(I'm not adding suggestions here because it seems my last suggestion requires manual editing, so it might be easier to update that section at the same time...)

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+The provider comes with family specific [options](/en/reference/configuration-reference/#fontoptions).
+
+#### experimental.glyphs
+
+<p>
+
+**Type:** `string[]`
+</p>
+
+when resolving the new Material Symbols icons, allows specifying a list of glyphs to be included in the font for each font family. This can reduce the size of the font file:
```
ArmandPhilippot: typo:
```suggestion
When resolving the new Material Symbols icons, allows specifying a list of glyphs to be included in the font for each font family. This can reduce the size of the font file:
```

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+
+Retrieves fonts from [Google](https://fonts.google.com/):
+
+```js
+provider: fontProviders.google()
+```
+
+The provider comes with family specific [options](/en/reference/configuration-reference/#fontoptions).
+
+#### experimental.glyphs
```
ArmandPhilippot: nit:
```suggestion
#### `experimental.glyphs`
```
(I know we don't have this on the configuration reference... but on other pages and even at the bottom of this page we use backticks for headings using code, so consistency 😄 )

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+  provider: fontProviders.google(),
+  options: {
+    experimental: {
+      glyphs: ["a"]
+    }
+  }
+}
+```
+
+#### experimental.variableAxis
```
ArmandPhilippot:
```suggestion
#### `experimental.variableAxis`
```

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+
+Retrieves fonts from [Google Icons](https://fonts.google.com/icons):
+
+```js
+provider: fontProviders.googleicons()
+```
+
+The provider comes with family specific [options](/en/reference/configuration-reference/#fontoptions).
+
+#### experimental.glyphs
```
ArmandPhilippot:
```suggestion
#### `experimental.glyphs`
```

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+
+Retrieves fonts from disk:
+
+```js
+provider: fontProviders.local()
+```
+
+The provider comes with required family specific [options](/en/reference/configuration-reference/#fontoptions).
+
+#### variants
```
ArmandPhilippot:
```suggestion
#### `variants`
```

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+          src: ["./src/assets/fonts/custom-700.woff2"]
+        }
+        // ...
+      ]
+    }
+  }]
+});
+```
+
+##### weight
```
ArmandPhilippot:
```suggestion
##### `weight`
```

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+
+If the associated font is a [variable font](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/Variable_fonts_guide), you can specify a range of weights:
+
+```js
+weight: "100 900"
+```
+
+When the value is not set, by default Astro will try to infer the value based on the first [`source`](#src).
+
+##### style
```
ArmandPhilippot:
```suggestion
##### `style`
```

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+
+A [font style](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style):
+
+```js
+style: "normal"
+```
+
+When the value is not set, by default Astro will try to infer the value based on the first [`source`](#src).
+
+##### src
```
ArmandPhilippot:
```suggestion
##### `src`
```

src/content/docs/en/guides/fonts.mdx
florian-lefebvre: Will apply once we're done with the local font section :+1:

src/content/docs/en/guides/fonts.mdx
```diff
@@ -8,129 +8,329 @@ description: >-
...
+    ---
+    import Layout from "../components/Layout.astro";
+    ---
+    <Layout>
+      <h1>In a galaxy far, far away...</h1>
+
+      <p>Custom fonts make my headings much cooler!</p>
+
+      <style>
+      h1 {
```
florian-lefebvre: @ArmandPhilippot I was thinking about doing the same for fontsource but that won't work as nicely I think. The style example there is for the `body` so that would usually go in the layout, not a page

src/content/docs/en/guides/fonts.mdx
```diff
@@ -8,129 +8,329 @@ description: >-
...
+      </style>
+    </Layout>
     ```
+
 </Steps>
 
 ## Using Fontsource
 
-The [Fontsource](https://fontsource.org/) project simplifies using Google Fonts and other open-source fonts. It provides npm modules you can install for the fonts you want to use.
+The [Fontsource](https://fontsource.org/) project simplifies using Google Fonts and other open-source fonts.
```
florian-lefebvre: The example is about the fontsource provider to match the old guide. I'm concerned that people may then miss all the other available providers

src/content/docs/en/guides/fonts.mdx
```diff
@@ -8,129 +8,329 @@ description: >-
...
+            weight: 'normal',
+            style: 'normal'
+          }]
+        }
+      }]
+    });
     ```
 
-3. Use the `font-family` value from the `@font-face` statement to style elements in your component or layout. In this example, the `<h1>` heading will have the custom font applied, while the paragraph `<p>` will not.
+3. Include styles using the [`<Font />`](/en/reference/modules/astro-assets/#font-) component in the head of your page, usually in a [layout](/en/basics/layouts/):
```
ArmandPhilippot: Is the following enough?
```suggestion
3. Include the [`<Font />`](/en/reference/modules/astro-assets/#font-) component in the head of your page, usually in a [layout](/en/basics/layouts/):
```
We don't really include styles in this step. I guess the idea was the Font component will generate the styles for the font-face, but this is an implementation detail and maybe we don't need to say that in the guide?

src/content/docs/en/guides/fonts.mdx
```diff
@@ -8,129 +8,329 @@ description: >-
...
 
-    ```astro title="src/pages/example.astro" {9-11}
+    ```astro title="src/components/Head.astro"
     ---
+    import { Font } from "astro:assets";
     ---
 
-    <h1>In a galaxy far, far away...</h1>
+    <Font cssVariable="--font-distant-galaxy" preload />
+    ```astro title="src/components/Layout.astro"
```
ArmandPhilippot: Sorry, it seems GitHub messed up again with my suggestions... Step 3 in "Using a local font file" mixes the old code and my suggestion... Do you want me to push a commit to fix it? (because I don't think a new suggestion will work better... 😅 )

src/content/docs/en/guides/fonts.mdx
```diff
@@ -8,129 +8,329 @@ description: >-
...
+      </body>
+    </html>
+   ```
 
-    <style>
-    h1 {
-      font-family: 'DistantGalaxy', sans-serif;
-    }
-    </style>
+4. Use the `font-family` value from the `@font-face` statement to style elements in your component or layout. In this example, the `<h1>` heading will have the custom font applied, while the paragraph `<p>` will not.
```
ArmandPhilippot: Rereading that, I wonder if talking about `@font-face` makes sense here. Maybe:
```suggestion
4. Use the `cssVariable` value of the `<Font />` component to define the font to use for elements in your component or layout. In the following example, the `<h1>` heading will have the custom font applied, while the paragraph `<p>` will not.
```

src/content/docs/en/guides/fonts.mdx
ArmandPhilippot: Ah, right, I didn't noticed the style was for the body. Looking again the step 4 for fontsource is also wrong:
> 4. Use the font-family value from the @font-face statement to style elements in your component or layout:

(we no longer have a `@font-face` statement...)

So, the current step 4 should be removed, and in step 5 (becoming step 4) we can replace:
```diff
-body {
+h1 {
```

I don't think this is important to keep the styles on body for the example?

src/content/docs/en/guides/fonts.mdx
ArmandPhilippot: Maybe instead of the tip at the end of the section, we could say that here? Something like:

```suggestion
Astro supports [several font providers](/en/reference/font-provider-reference/#built-in-providers) out of the box. For example, the [Fontsource](https://fontsource.org/) provider simplifies using Google Fonts and other open-source fonts.

To use Fontsource in your project:
```

src/content/docs/en/guides/fonts.mdx
```diff
@@ -13,21 +13,30 @@ import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
 
 This guide will show you how to add [web fonts](https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Web_fonts) to your project and use them in your components.
 
-Astro provides a way to use fonts [from your filesystem and various font providers](/en/reference/font-provider-reference/#built-in-providers) through a unified, [fully customizable](/en/reference/configuration-reference/#fonts), and type-safe API.
+Astro provides a way to use fonts from your filesystem and various font providers (e.g. Fontsource, Google) through a unified, [fully customizable](/en/reference/configuration-reference/#fonts), and [type-safe API](/en/reference/font-provider-reference/).
```
florian-lefebvre: I wouldn't link type safe to the font provider reference. I meant to say "type safe for you the user", not the font provider author

src/content/docs/en/guides/fonts.mdx
```diff
@@ -13,21 +13,30 @@ import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
...
 Web fonts can impact page performance at both load time and rendering time. This API helps you keep your site performant with automatic [web font optimizations](https://web.dev/learn/performance/optimize-web-fonts) including preload links, optimized fallbacks, and opinionated defaults. [See common usage examples](#examples).
 
 The Fonts API focuses on performance and privacy by downloading and caching fonts so they're served from your site. This can avoid sending user data to third-party sites, and also ensures that a consistent set of fonts is available to all your visitors.
 
-## Using a local font file
+## Configuring custom fonts
+
+Registering custom fonts for your Astro project is done through [the `fonts` option](/en/reference/configuration-reference/#fonts) in your Astro config.
+
+For each font you want to use, you must specify its name, a CSS variable, and an Astro font provider.
```
florian-lefebvre: Do we want to link each to their corresponding entry in the config reference?

src/content/docs/en/guides/fonts.mdx
sarah11918: OK, we can remove this link! I just didn't want people to need to leave the page just to see the list of providers when they're not ready for that page yet.

(It would be very tempting to click away to see what's supported at that point. We should try to keep them on this page to learn about fonts, first.)

src/content/docs/en/guides/fonts.mdx
sarah11918: Yes, we totally could!

sarah11918: I didn't make a link for [Astro font provider] because the next sentence immediately links to the built in ones, but I did add for name and css variable

src/content/docs/en/guides/fonts.mdx
```diff
@@ -23,7 +23,7 @@ The Fonts API focuses on performance and privacy by downloading and caching font
 
 Registering custom fonts for your Astro project is done through [the `fonts` option](/en/reference/configuration-reference/#fonts) in your Astro config.
 
-For each font you want to use, you must specify its name, a CSS variable, and an Astro font provider.
+For each font you want to use, you must specify its [name](/en/reference/configuration-reference/#fontname), a [CSS variable](/en/reference/configuration-reference/#fontcssvariable), and an Astro font provider.
```
florian-lefebvre, line 26: Maybe need to also add a link for the provider property?

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -0,0 +1,720 @@
...
+    cssVariable: "--font-custom"
+  }]
+});
+```
+
+</TabItem>
+
+<TabItem label="Provider options">
+
+```ts title="astro.config.ts" {8}
```
ArmandPhilippot: To highlight `token` (currently this is the closing braces):
````suggestion
```ts title="astro.config.ts" {7}
````

File: src/content/docs/en/reference/font-provider-reference.mdx [hunk 25]
```diff
@@ -0,0 +1,720 @@
...
+    cssVariable: "--font-custom"
+  }]
+});
+```
+
+</TabItem>
+
+<TabItem label="Family options">
+
+```ts title="astro.config.ts" {8-10}
```
ArmandPhilippot: This one is also offset:
````suggestion
```ts title="astro.config.ts" {7-9}
````

File: src/content/docs/en/reference/font-provider-reference.mdx [hunk 26]
```diff
@@ -0,0 +1,720 @@
...
+    cssVariable: "--font-custom"
+  }]
+});
+```
+
+</TabItem>
+
+<TabItem label="Provider options">
+
+```ts title="astro.config.ts" {8}
```
ArmandPhilippot: Same here:
````suggestion
```ts title="astro.config.ts" {7}
````

File: src/content/docs/en/reference/font-provider-reference.mdx [hunk 27]
```diff
@@ -0,0 +1,720 @@
...
+    cssVariable: "--font-custom"
+  }]
+});
+```
+
+</TabItem>
+
+<TabItem label="Family options">
+
+```ts title="astro.config.ts" {8-10}
```
ArmandPhilippot:
````suggestion
```ts title="astro.config.ts" {7-9}
````

src/content/docs/en/reference/font-provider-reference.mdx
florian-lefebvre: Should we create an issue for this so we don't forget?

ArmandPhilippot: I added that to my TODO list, so I'll get to it at some point (maybe next week). So, not necessarily? But if we want to give someone who wants to help a chance, yes, we can create one!

PR #13108, opened 2026-01-23
feat: add mention of possibility of perhaps async in the file() loader

Princesseuh:
#### Description (required)

For https://github.com/withastro/astro/pull/14932. The parser function can now be async, which is pretty neat if you want to fetch files every time you parse your .csv (maybe don't do that)

### DIFF
--- a/src/content/docs/en/reference/content-loader-reference.mdx
+++ b/src/content/docs/en/reference/content-loader-reference.mdx
@@ -28,7 +28,7 @@ Astro provides two built-in loaders to help you fetch your collections. Both off
 
 The `glob()` loader creates entries from directories of files from anywhere on the filesystem. The supported file types are Markdown, MDX, Markdoc, JSON, YAML, and TOML files.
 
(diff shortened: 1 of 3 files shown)


sarah11918:
Oops, forgot that the reference entry for `parser()` at [docs.astro.build/en/reference/content-loader-reference#parser](https://docs.astro.build/en/reference/content-loader-reference/#parser) needs updating (both in current docs and in v6 docs) to now show that this can be async.

I think @ArmandPhilippot suggested what this needs to look like in the other PR? https://github.com/withastro/astro/pull/14932#issuecomment-3756834417

ArmandPhilippot:
Yes, we would need to update the type with:
```
(text: string) => Record<string, Record<string, unknown>> | Array<Record<string, unknown>> | Promise<Record<string, Record<string, unknown>> | Array<Record<string, unknown>>>
```
to show in the reference that the parser can be async.

sarah11918:
Just noting that we still need *someone* to make the change Armand has suggested above! (Y'all really gonna make me open Firebase Studio for this.... 😆 )

PR #13139, opened 2026-01-25
reference for withastro/astro#15153

jcayzac:
#### Description (required)

This is the reference doc for withastro/astro#15153, for v6.


src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -323,6 +323,33 @@ fetchpriority="high"
...
+
+<p>
+
+**Type:** `string | undefined`<br />
+<Since v="6.0.0" />
+</p>
+
+By default, Sharp uses a black background when flattening an image. This property lets you specify a different background color.
+
+This is especially useful when requesting an output format that does not support transparency, e.g. `jpeg`:
```
```suggestion
The background color to use when flattening an image to transform it into the requested output `format`.

By default, Sharp uses a black background when flattening an image. Specifying a different background color is especially useful when transforming images with transparent backgrounds to a format that does not support transparency (e.g. `jpeg`):
```
Just a quick edit to be more in line with our style! Is anything incorrect here?

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -323,6 +323,33 @@ fetchpriority="high"
 
 These individual attributes can still be set manually if you need to customize them further.
 
+#### `background`
+
+<p>
+
+**Type:** `string | undefined`<br />
+<Since v="6.0.0" />
```
```suggestion
<Since v="5.17.0" />
```

sarah11918:
Ugh, I did not successfully switch this to the main branch, since this is for v5.17 minor and not v6.  Making a new PR to replace this one for simplicity 😅 #13149

sarah11918:
Pick this up in #13149

jcayzac:
> this is for our upcoming minor release, not waiting on the major

@sarah11918 Thanks! I keep getting confused between this PR and the SVG rasterization one (targeting v6), why is that? Am I this old already? 🧓🏼

PR #13141, opened 2026-01-25
fix: clarifying the staticpaths and endpoint props usage

### LINKED ISSUE #11785: Update `APIRoute` code snippets and clarify `props` in endpoints
### 📚 Subject area/topic

Endpoints

https://docs.astro.build/en/guides/endpoints/

1. **Update code snippets using `APIRoute`**

Erika was suggesting [on Discord](https://discord.com/channels/830184174198718474/853350631389265940/1377067027348590623):
> export const GET: APIRoute = async ({ params, request }) => {...}
should use satisfies instead

I'm not entirely sure why. I tried both but I don't see any differences... Maybe type inference in some cases?

2. **Misleading wording regarding `props`**

In [params and Dynamic routing](https://docs.astro.build/en/guides/endpoints/#params-and-dynamic-routing), we say:
> but because the endpoint is a function and not a component, [props](https://docs.astro.build/en/reference/routing-reference/#data-passing-with-props) aren’t supported.

In static mode, we can share `props` through the `APIContext` object using `getStaticPaths`. So, I think this is not entirely correct and could be confusing. Here's a small repro using a tweaked version of the code snippet we show above this sentence: https://stackblitz.com/edit/astro-endpoints-dynamic-routing?file=src%2Fpages%2Fapi%2F[id].json.ts,src%2Fpages%2Findex.astro&on=stackblitz

But, yeah, I guess we can't do that with on-demand rendered endpoints. So we might need to precise this only applies to on-demand mode.

https://stackblitz.com/edit/astro-endpoints-dynamic-routing?file=src%2Fpages%2Fapi%2F[id].json.ts,src%2Fpages%2Findex.astro&on=stackblitz

By clicking on the links, the console should log:
```
1 Chris
15:22:37 [200] /api/1.json 2ms
15:22:38 [200] / 27ms
2 Yan
15:22:39 [200] /api/2.json 4ms
15:22:39 [200] / 28ms
3 Elian
15:22:40 [200] /api/3.json 5ms
15:22:41 [200] / 18ms
```
`props` can be used with endpoints.

avilyre:
#### Description (required)

1 - This PR updates all endpoint code examples to use the `satisfies` operator with the `APIRoute` type. This ensures robust type safety;
2 - Clarifies how `props` flow between the endpoint and `getStaticPaths` function when using `static` mode for pages;

Docs reference: [params and Dynamic routing](https://docs.astro.build/en/guides/endpoints/#params-and-dynamic-routing)

#### Related issues & labels

- Closes #11785
- Suggested label:  improve or update documentation


src/content/docs/en/guides/endpoints.mdx
```diff
@@ -49,7 +49,7 @@ export const GET: APIRoute = async ({ params, request }) => {...}
 
 ### `params` and Dynamic routing
 
-Endpoints support the same [dynamic routing](/en/guides/routing/#dynamic-routes) features that pages do. Name your file with a bracketed parameter name and export a [`getStaticPaths()` function](/en/reference/routing-reference/#getstaticpaths). Then, you can access the parameter using the `params` property passed to the endpoint function:
+Endpoints support the same [dynamic routing](/en/guides/routing/#dynamic-routes) features that pages do. Name your file with a bracketed parameter name and export a [`getStaticPaths()` function](/en/reference/routing-reference/#getstaticpaths), that will be called before your endpoint and generate the paths for you. Then you can access the same generated parameters using the `params` property, but now in your endpoint too:
```
ArmandPhilippot: I'm not sure this change is required? This doesn't sound related to the issue, and I don't think this is clearer than before. Can you explain what the intent was?

src/content/docs/en/guides/endpoints.mdx
```diff
@@ -76,7 +76,7 @@ export function getStaticPaths() {
 }
 ```
 
-This will generate four JSON endpoints at build time: `/api/0.json`, `/api/1.json`, `/api/2.json` and `/api/3.json`. Dynamic routing with endpoints works the same as it does with pages, but because the endpoint is a function and not a component, [props](/en/reference/routing-reference/#data-passing-with-props) aren't supported.
+This will generate four JSON endpoints at build time: `/api/0.json`, `/api/1.json`, `/api/2.json` and `/api/3.json`. Dynamic routing with endpoints works the same as it does with pages, but `getStaticPaths()` is only supported in `static` mode, since it is called at build time.
```
ArmandPhilippot: I'm not sure this really fixes the issue. The idea was to make it clearer that you can pass props in static mode, but not with on-demand rendering. Maybe something like this:

```suggestion
This will generate four JSON endpoints at build time: `/api/0.json`, `/api/1.json`, `/api/2.json` and `/api/3.json`. Dynamic routing with endpoints works the same as it does with pages. In static mode, you can [pass props to the endpoint using `getStaticPaths()`](/en/reference/routing-reference/#data-passing-with-props). However, with on-demand rendering, since the endpoint is a function and not a component, passing props is not supported.
```

src/content/docs/en/guides/endpoints.mdx
avilyre: Not necessarily, it's just a suggestion to be clearer and reinforce the following text. I can remove it if it doesn't make sense 😄

✅ I updated with `satisfies` operator. If Erika is a core maintainer, then it really makes sense.

src/content/docs/en/guides/endpoints.mdx
```diff
@@ -39,32 +39,32 @@ export async function GET({ params, request }) {
 }
 ```
 
-You can also type your endpoint functions using the `APIRoute` type:
+You can also type your endpoint functions using the `APIRoute` type with the `satisfies` operator, to get a granular and type-safe endpoint:
```
ArmandPhilippot: This sounds a bit redundant, maybe the following is enough::
```suggestion
You can also get type safety in your endpoint functions using the `APIRoute` type with the `satisfies` operator:
```

src/content/docs/en/guides/endpoints.mdx
```diff
@@ -39,32 +39,32 @@ export async function GET({ params, request }) {
...
 
 ```ts
 
-export const GET: APIRoute = async ({ params, request }) => {...}
+export const GET = (async ({ params, request }) => { ... }) satisfies APIRoute;
```
ArmandPhilippot: nit: while we're here, we should probably use a comment to make the code valid:
```suggestion
export const GET = (async ({ params, request }) => { /* ... */ }) satisfies APIRoute;
```

src/content/docs/en/guides/endpoints.mdx
ArmandPhilippot: Yeah, I don't think this makes the section clearer, so it's probably best to revert that change:

```suggestion
Endpoints support the same [dynamic routing](/en/guides/routing/#dynamic-routes) features that pages do. Name your file with a bracketed parameter name and export a [`getStaticPaths()` function](/en/reference/routing-reference/#getstaticpaths). Then, you can access the parameter using the `params` property passed to the endpoint function:
```

src/content/docs/en/guides/endpoints.mdx
```diff
@@ -76,7 +76,7 @@ export function getStaticPaths() {
 }
 ```
 
-This will generate four JSON endpoints at build time: `/api/0.json`, `/api/1.json`, `/api/2.json` and `/api/3.json`. Dynamic routing with endpoints works the same as it does with pages, but because the endpoint is a function and not a component, [props](/en/reference/routing-reference/#data-passing-with-props) aren't supported.
+This will generate four JSON endpoints at build time: `/api/0.json`, `/api/1.json`, `/api/2.json` and `/api/3.json`. Dynamic routing with endpoints works the same as it does with pages. In `static` mode, you can [pass props to the endpoint using `getStaticPaths()`](/en/reference/routing-reference/#data-passing-with-props). However, with on-demand rendering, since the endpoint is a function and not a component, passing props is not supported
```
ArmandPhilippot:
```suggestion
This will generate four JSON endpoints at build time: `/api/0.json`, `/api/1.json`, `/api/2.json` and `/api/3.json`. Dynamic routing with endpoints works the same as it does with pages. In static mode, you can [pass props to the endpoint using `getStaticPaths()`](/en/reference/routing-reference/#data-passing-with-props). However, with on-demand rendering, since the endpoint is a function and not a component, passing props is not supported.
```

PR #13148, opened 2026-01-26
Change 'httpUrl' to 'serverUrl' in mcp_config.json

deepakness:
#### Description (required)

`httpUrl` is not supported in Antigravity so the MCP only works when you use `serverUrl`, so this must be updated.

Please make changes in all languages.


sarah11918:
Hi, thank you for this! These documents were just added very recently by another community member, and I've been trying to research in Google's documentation to know what the correct key to use is. Can you point me to some documentation on this so I can verify the change? Thank you! We really appreciate you helping us to keep the docs correct! 💪 

UPDATE: We reviewed the original PR that added this, and they had a link that did indeed show `serverUrl`, but their PR used `httpUrl` in error. That's good enough for me! Thank you, we are happy to have this fix! 🥳

PR #13150, opened 2026-01-26
document `retainBody` for glob() loader

sarah11918:
#### For Astro version: `5.17.0`. See astro [PR#15022](https://github.com/withastro/astro/pull/15022)

New content: 

- https://deploy-preview-13150--astro-docs-2.netlify.app/en/reference/modules/astro-content/#body
- https://deploy-preview-13150--astro-docs-2.netlify.app/en/reference/content-loader-reference/#retainbody

Note: as these are just reference entries, I *think* this should all transfer over to v6 just fine when merged in, but I'll keep an eye out when we update the v6 branch with main after the release.


src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -28,7 +28,7 @@ Astro provides two built-in loaders to help you fetch your collections. Both off
 
 The `glob()` loader creates entries from directories of files from anywhere on the filesystem. The supported file types are Markdown, MDX, Markdoc, JSON, YAML, and TOML files.
 
-This loader accepts an object with the following properties: `pattern`, `base` (optional), and `generateId` (optional).
+This loader accepts an object with the following properties: `pattern`, `base` (optional), `generateId` (optional), and `retainBody (optional)`.
```
ArmandPhilippot: nit:
```suggestion
This loader accepts an object with the following properties: `pattern`, `base` (optional), `generateId` (optional), and `retainBody` (optional).
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -44,6 +44,16 @@ const blog = defineCollection({
...
+const posts = defineCollection({
+  /* Retrieve all Markdown files in your blog directory and prevent
+   * the raw body of content files from being stored in the data store. */
+  loader: glob({
+    pattern: '**/*.md',
+    base: './src/content/blog',
+    retainBody: false
+  }),
+  schema: /* ... */
+});
```
ArmandPhilippot: even more nit: we already have `./src/data/blog` above for a blog collection... shouldn't we use another example?
```suggestion
const notes = defineCollection({
  /* Retrieve all Markdown files in your notes directory and prevent
   * the raw body of content files from being stored in the data store. */
  loader: glob({
    pattern: '**/*.md',
    base: './src/data/notes',
    retainBody: false
  }),
  schema: /* ... */
});
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -91,6 +101,25 @@ A callback function that returns a unique string per entry in a collection. It a
...
+<Since v="5.17.0" />
+</p>
+
+Whether or not to store the raw body of content files in the data store.
+
+When `retainBody` is `false`, [`entry.body`](/en/reference/modules/astro-content/#body) will be `undefined` instead of containing the raw file contents.
+
+Setting this property to `false` significantly reduces the deployed size of the data store and helps avoid hitting size limits for sites with very large collections.
+
+For Markdown files, the rendered body will still be available in the `entry.rendered.html` property, and the `entry.filePath` property will still point to the original file.
```
ArmandPhilippot: Does it makes sense to add links here? (to see the full shape, especially for `rendered`)?

```suggestion
For Markdown files, the rendered body will still be available in the [`entry.rendered.html` property](/en/reference/content-loader-reference/#rendered), and the [`entry.filePath` property](/en/reference/content-loader-reference/#filepath) will still point to the original file. 
```

PR #13165, opened 2026-01-28
Document renderMarkdown frontmatter parsing and fileURL option

matthewp:
#### Description (required)

- Adds frontmatter parsing to `renderMarkdown`
- Adds `fileURL` option to renderMarkdown as well.


src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -903,6 +903,23 @@ export function myLoader(settings) {
 }
 ```
 
+##### fileURL
```
ArmandPhilippot: nit: we usually use backticks for code even in headings
```suggestion
##### `fileURL`
```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -903,6 +903,23 @@ export function myLoader(settings) {
...
+
```
ArmandPhilippot, line 936: We'll need to indicate when this was added and the type:
```suggestion

<p>

**Type**: `URL`
<Since v="6.0.0" />
</p>

```

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -903,6 +903,23 @@ export function myLoader(settings) {
...
+Pass a `fileURL` option to allow relative image paths in the Markdown content to be resolved correctly:
```
ArmandPhilippot:
```suggestion
Specifies the file path to use for resolving relative image paths in Markdown content.

The following example uses the [configured root directory](/en/reference/configuration-reference/#root) to resolve image paths:
```
(I'm not Sarah, so there may be room for improvement, but this is consistent with our other API inputs.)

sarah11918: Thanks Armand! Yes, our format is to always start with a kind of "definition": a description of what the thing *is* or *does*. It answers the question for the reader, "What the heck is this thing??" 😄 

(Here's our documentation on the format we use for entries: https://contribute.docs.astro.build/reference/api-references/ )

florian-lefebvre:
Just noting that the associated code PR has been merged and released, so once this is approved it can be merged!

src/content/docs/en/reference/content-loader-reference.mdx
```diff
@@ -868,15 +868,15 @@ export function feedLoader({ url }) {
 
 <p>
 
-**Type**: `(markdown: string) => Promise<RenderedContent>`
+**Type**: `(content: string, options?: { fileURL?: URL }) => Promise<RenderedContent>`
 <Since v="5.9.0" />
 </p>
 
 Renders a Markdown string to HTML, returning a `RenderedContent` object.
```
sarah11918, line 904: Is this definition of what it returns still accurate, since it now also returns `metadata` and access to `frontmatter`? (We also say this in the very last line on the page)

matthewp: Yes, it's still the same, it always returned this shape it just didn't have the frontmatter in it before.

PR #13176, opened 2026-01-30
Update Nightly GitHub action to pull docs from astro repo

sarah11918:
#### Description (required)

The `withastro/astro` repo will soon use `main` for v6 beta code, and a `5-legacy` branch for v5 code. This updates the nightly GitHub action to pull in config reference and errors to represent these new source branches for code to make sure docs get to the right branch!

Don't merge until @florian-lefebvre makes the corresponding change in the astro repo!

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,10 @@
 <!-- Thank you for opening a PR! We really appreciate you taking the time to help out 🙌 -->
 
+<!-- 🚨 IMPORTANT INFO AS WE PREPARE FOR v6 🚨 -->
+<!-- We are only accepting emergency fixes for v5 docs (typos, links etc.) -->
+<!-- All new feature documentation PRs must use the `v6` branch  -->
+<!-- No new translations will be accepted until v6 is released -->
+
 #### Description (required)
 
 <!-- Please describe the change you are proposing, and why -->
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -14,13 +14,14 @@ jobs:
       matrix:
         include:
           # Default options for running with latest Astro code
-          - SOURCE_BRANCH: main
+          # switch this back to main source branch for normal operations after v6 merge!
+          - SOURCE_BRANCH: legacy-5
```
ArmandPhilippot: I think this should be:
```suggestion
          - SOURCE_BRANCH: 5-legacy
```
https://github.com/withastro/astro/tree/5-legacy

sarah11918: ugh, I specifically thought I typed that (I know Florian had mentioned the opposite.) Thank you!

ArmandPhilippot (approved):
Looks better now, and I think this is all we need! 🙌🏽

.github/workflows/nightly.yml
```diff
@@ -14,13 +14,14 @@ jobs:
       matrix:
         include:
           # Default options for running with latest Astro code
-          - SOURCE_BRANCH: main
+          # switch this back to main source branch for normal operations after v6 merge!
```
delucis: Probably makes no difference, but could be good to include `TODO:` here as the standard pattern and makes it easy to search for in the code base.

```suggestion
          # TODO: switch this back to main source branch for normal operations after v6 merge!
```

PR #13200, opened 2026-02-01
docs: use relative path for Decap CMS config to support subpath deployments

tushar-manna:
The docs currently reference the config file as:

"<link href="/admin/config.yml" ...>"

This works only when the site is hosted at the domain root.

On platforms like GitHub Pages project sites (e.g. username.github.io/repo), the leading slash causes Decap to request the config from:

  /admin/config.yml

at the domain root, which results in a 404.

Changing this to a relative path:

 <link href="config.yml" ...>

allows Decap to correctly resolve the config relative to /admin/index.html. This works for:
- GitHub Pages project sites
- Custom domains
- Local development

without breaking existing setups.


yanthomasdev (approved):
LGTM! Thanks

PR #13207, opened 2026-02-02
i18n(ja): update static-client-address-not-available.mdx

morinokami:
#### Description (required)

- [diff](https://github.com/withastro/docs/commits/main/src/content/docs/en/reference/errors/static-client-address-not-available.mdx?since=2023-08-21T17:51:08.000Z)
  - https://github.com/withastro/docs/pull/9240

- Suggested label: i18n


sarah11918:
Note for @yanthomasdev that this error message doesn't update from v5 -> v6, so seems safe!

PR #13208, opened 2026-02-02
Document React `appEntrypoint` prop

meyer:
#### Description (required)

This PR contains React integration docs updates for this PR: https://github.com/withastro/astro/pull/15368

I documented the new `appEntrypoint` prop and added specific details for accessing `Astro.locals` with the `locals` prop.

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/react.mdx
+++ b/src/content/docs/en/guides/integrations-guide/react.mdx
@@ -280,6 +280,60 @@ export default defineConfig({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


sarah11918:
I'm so sorry, but it looks like this is not making it in pre-v6, and our docs repo is going to update majorly after v6 so it will be safer to re-draft this if/when the feature is closer to being released. Closing for now, and the content can be added in a separate PR later when the feature is approved!

joaopalopes24:
@meyer Reopen this PR, please.

PR #13212, opened 2026-02-02
i18n(de): remove very outdated translation [i18nIgnore]

trueberryless:
#### Description

Cleaning up the German translation for the v6 release by deleting one file which was very outdated.

Related: #13213


sarah11918 (approved):
Thank you for making the tough calls! :saluting_face: To keep things in sync, can you please also make a v5 PR to avoid any conflicts?

trueberryless:
> Thank you for making the tough calls! 🫡 To keep things in sync, can you please also make a v5 PR to avoid any conflicts?

Thank you. There are some other files that are outdated, but I checked each one and they all just have smaller fixes or changes which would not harm the user experience as badly as this one.

I also fixed the broken link fragments in both PRs. Thanks for reminding me about the v5, I would have forgotten!

PR #13213, opened 2026-02-02
i18n(de): remove very outdated translation [i18nIgnore]

trueberryless:
#### Description

Cleaning up the German translation for the v6 release by deleting one file which was very outdated.

Related: #13212


sarah11918:
Ah, I see all the pages updated for now English links. Thank you!

PR #13220, opened 2026-02-03
[next] fix cloudflare custom entrypoint docs

florian-lefebvre:
#### Description (required)

The package we documented was incorrect, and the code sample could be improved

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -579,7 +579,7 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
     export default {
       async fetch(request, env, ctx) {
         await env.MY_QUEUE.send('log');
-        return await handle(request, env, ctx);
+        return handle(request, env, ctx);
       },
       async queue(batch, _env) {
         let messages = JSON.stringify(batch.messages);


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -579,9 +579,9 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
     export default {
       async fetch(request, env, ctx) {
         await env.MY_QUEUE.send('log');
-        return handler.fetch(request, env, ctx);
+        return await handle(request, env, ctx);
```
ArmandPhilippot: nit: I think we don't need the await in the return statement?

```suggestion
        return handle(request, env, ctx);
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -579,9 +579,9 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
...
       },
-      async queue(batch, _env) {
+      async queue(batch) {
```
ArmandPhilippot: nit: Looking at [the changeset](https://github.com/withastro/astro/blob/main/.changeset/cloudflare-entrypoint-breaking.md?plain=1), this still exists and I think `_env` was there to show this can take a second argument, even if this is not used in the example and so maybe it's better to keep it?

```suggestion
      async queue(batch, _env) {
```

ArmandPhilippot (approved):
Two non-blocking nits, otherwise LGTM! Thanks!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
florian-lefebvre: I think we may? At least it seems more correct to me given `fetch` above, otherwise that means the async function returns a promise. And `fetch` needs to be async because of the queue

sarah11918: Can we get someone to confirm? @alexanderniebuhr maybe? Otherwise, I'm happy if Armand is happy!

ArmandPhilippot: I just notice the changeset doesn't use `await`, and usually using `await` in a return statement is redundant. But this is a nit, I don't know how `handle()` works so... I'm fine if you prefer to keep it! 😄

florian-lefebvre: nvm it's fine! just checked

yanthomasdev (approved):
LGTM!

PR #13222, opened 2026-02-03
feat(v6): add docs for `astro/config` module

ArmandPhilippot:
#### Description (required)

* Adds docs for `astro/config` exports in our `astro:config` module reference:
  - `defineConfig()`: Not the easiest to describe... I was tempted by "Defines the configuration for" :smile:
  - `envField`,  I see two options (I chose the first one):
    1. We document everything here and we replace what's in Configuration reference with a link to that page. This declutters the configuration reference and we can more easily describe each `envField` methods.
    2. We only provide a link to the matching section in the Configuration reference.
  - `fontProviders`: pretty minimal because we already have a reference for all the providers
  - `getViteConfig()`: Added in https://github.com/withastro/astro/pull/4154 (v1, so no Since). I don't know if there are other use cases than Vitest, linking to the Testing guide is probably enough.
  - `mergeConfig()`: moved from [Programmatic Astro API](/en/reference/programmatic-reference/#mergeconfig) with some rewording
  - `passthroughImageService()`: I reused the wording we had in the Images guide.
  - `sessionDrivers`: the object is pretty big, so I think it's fine to keep it minimal and use links, as with `fontProviders`.
  - `sharpImageService()`: not really added, but this is available from this module since https://github.com/withastro/astro/pull/6995 and I'm not really quite sure about any use cases since Sharp is now the default image service.
  - `validateConfig()`: moved from [Programmatic Astro API](/en/reference/programmatic-reference/#mergeconfig) with some rewording

* Updates the `env.schema` configuration reference as a preview of what it might look like if we update the JSDoc (see `envField`). If we're happy with that, we'll need to make a PR in the core repo!

**Things to consider:**
* Should we add links to `mergeConfig()` and `validateConfig()` from the Programmatic Astro API page? I don't see any obvious place because the current content is focused on CLI commands executed programmatically. We might need to reword a bit the introduction if we want to add links.
* Are we okay with the configuration reference change? We can also keep the previous description and just add a link to `envField`.

- Suggested label: add new content, v6

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -566,10 +566,10 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
     }
     ```
 
-3. Update your custom worker entry file to use standard Worker syntax. Import the handler from `@astrojs/cloudflare/entrypoints/server` and export a standard Cloudflare Worker object, alongside any custom exports like Durable Objects:
+3. Update your custom worker entry file to use standard Worker syntax. Import the handler from `@astrojs/cloudflare/handler` and export a standard Cloudflare Worker object, alongside any custom exports like Durable Objects:
 
     ```ts title="src/worker.ts"
-    import handler from '@astrojs/cloudflare/entrypoints/server';
+    import { handle } from '@astrojs/cloudflare/handler';
     import { DurableObject } from 'cloudflare:workers';
 
     export class MyDurableObject extends DurableObject<Env> {
@@ -579,7 +579,7 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
     export default {
       async fetch(request, env, ctx) {
         await env.MY_QUEUE.send('log');
-        return handler.fetch(request, env, ctx);
+        return handle(request, env, ctx);
       },
       async queue(batch, _env) {
         let messages = JSON.stringify(batch.messages);
--- a/src/content/docs/en/guides/integrations-guide/netlify.mdx
+++ b/src/content/docs/en/guides/integrations-guide/netlify.mdx
@@ -408,23 +408,19 @@ Injects environment variables from your Netlify site into the development enviro
 
 This allows you to use the same values in development as you would in production. See [the Netlify docs on environment variables](https://docs.netlify.com/build/environment-variables/overview/) for more information, including how to use different variables for different environments.
 
-## Experimental features
-
-The following features are also available for use, but may be subject to breaking changes in future updates. Please follow the [`@astrojs/netlify` CHANGELOG](https://github.com/withastro/astro/tree/main/packages/integrations/netlify/CHANGELOG.md) carefully for updates if you are using these features in your project.
-
-### `experimentalStaticHeaders`
+### `staticHeaders`
 
 <p>
   **Type:** `boolean` <br />
   **Default:** `false`<br />
-  <Since v="6.4.0"  pkg="@astrojs/netlify"/>
+  <Since v="7.0.0"  pkg="@astrojs/netlify"/>
 </p>
 
 Enables specifying custom headers for prerendered pages in Netlify's configuration.
 
 If enabled, the adapter will save [static headers in the Framework API config file](https://docs.netlify.com/frameworks-api/#headers) when provided by Astro features, such as Content Security Policy.
 
-For example, when [Content Security Policy](/en/reference/configuration-reference/#securitycsp) is enabled, `experimentalStaticHeaders` can be used to add the CSP `headers` to your Netlify configuration, instead of creating a `<meta>` element:
+For example, when [Content Security Policy](/en/reference/configuration-reference/#securitycsp) is enabled, `staticHeaders` can be used to add the CSP `headers` to your Netlify configuration, instead of creating a `<meta>` element:
 
 ```js title="astro.config.mjs" {9}
 import { defineConfig } from 'astro/config';
@@ -435,11 +431,10 @@ export default defineConfig({
     csp: true
   },
   adapter: netlify({
-    experimentalStaticHeaders: true 
+    staticHeaders: true 
   })
 });
 ```
-
 
 ## Examples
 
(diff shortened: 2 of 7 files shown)


sarah11918:
> Should we add links to mergeConfig() and validateConfig() from the Programmatic Astro API page? I don't see any obvious place because the current content is focused on CLI commands executed programmatically. We might need to reword a bit the introduction if we want to add links.

So one way to think about this: The Programmatic API page is marked as "experimental" while the `astro:config` page is clearly not. Do we know the status of these functions? If we intend to imply they too are kind of "experimental" in terms of stability or support, then I think we could keep them on the Programmatic API page, especially since that's clearly their *use*. (And, even if they're perfectly stable, if their *purpose* is to execute an experimental feature, then I think that could also make sense.)

In either case, it seems we *could* do with an improved introduction, either linking lower in the page or to the `astro:config` page entries:

> If you need more control when running Astro, the "astro" package exports APIs to programmatically run the CLI commands. There are also two (experimental, if true?) `astro:config` helpers that can programmatically verify and merge configurations.

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -2078,7 +2078,10 @@ See our guide for more information on [environment variables in Astro](/en/guide
 <Since v="5.0.0" />
 </p>
 
-An object that uses `envField` to define the data type and properties of your environment variables: `context` (client or server), `access` (public or secret), a `default` value to use, and whether or not this environment variable is `optional` (defaults to `false`).
+Describes the environment variables to use in your project, where each key corresponds to the variable name and the value to the data type and validations [defined with `envField`](/en/reference/modules/astro-config/#envfield).
```
sarah11918:
```suggestion
Defines environment variables to be enforced by Zod validation and for which TypeScript support (e.g. autocompletion, type-safety) is available. Each key corresponds to the variable name and the value to the data type and validations [defined with `envField`](/en/reference/modules/astro-config/#envfield).
```
So, this wording may not be perfect *yet*, but the thinking behind my change is that these aren't necessarily ALL / EVERY environment variable used by your project, right? Only the ones you're explicitly adding the type-safety feature to?

(Otherwise, yes, I agree with your choice of redistributing the information! 🙌 )

ArmandPhilippot: Ah, you're right! I was thinking that if someone uses this feature, all their environment variables are probably defined that way. But I agree, it's best not to assume anything!

Your suggestion looks great to me!

ArmandPhilippot:
> Do we know the status of these functions?

In the source code, I saw nothing that would lead me to believe that these functions are experimental. They were exposed by Fryuni in https://github.com/withastro/astro/pull/12052

One thing that make me think those functions might be useful elsewhere is the changeset. More precisely the first part of this sentence:
> These helpers are particularly useful for integration authors and for developers writing scripts that need to manipulate Astro configurations programmatically.

But maybe I misunderstand that. Anyway, the primary use case is probably running the CLI programmatically so I agree to move them back! I'll push a commit for that, with your improved intro too!

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -4,7 +4,7 @@ i18nReady: true
 ---
 import Since from '~/components/Since.astro';
 
-If you need more control when running Astro, the `"astro"` package exports APIs to programmatically run the CLI commands.
+If you need more control when running Astro, the `"astro"` package exports APIs to programmatically run the CLI commands. There are also two [`astro:config`](/en/reference/modules/astro-config/) helpers that can programmatically verify and merge configurations.
```
sarah11918:
```suggestion
If you need more control when running Astro, the `"astro"` package exports APIs to programmatically run the CLI commands. There are also two `astro:config` helpers that can programmatically [validate](#validateconfig) and [merge](#mergeconfig) configurations.
```
What do you think about these links instead? It seems like it might be less helpful to send people to the module page rather than send them to the entries below on this page!

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -2078,7 +2078,10 @@ See our guide for more information on [environment variables in Astro](/en/guide
 <Since v="5.0.0" />
 </p>
 
-An object that uses `envField` to define the data type and properties of your environment variables: `context` (client or server), `access` (public or secret), a `default` value to use, and whether or not this environment variable is `optional` (defaults to `false`).
+Defines environment variables to be enforced by Zod validation and for which TypeScript support (e.g. autocompletion, type-safety) is available. Each key corresponds to the variable name and the value to the data type and validations [defined with `envField`](/en/reference/modules/astro-config/#envfield).
+
+Four data types are supported: string, number, enumeration, and boolean.Each type requires a `context` (client or server), an `access` level (public or secret), and additional validations, such as a `default` value and an indication of whether the variable is `optional` (defaults to `false`).
```
yanthomasdev: Missing whitespace
```suggestion
Four data types are supported: string, number, enumeration, and boolean. Each type requires a `context` (client or server), an `access` level (public or secret), and additional validations, such as a `default` value and an indication of whether the variable is `optional` (defaults to `false`).
```

src/content/docs/en/reference/programmatic-reference.mdx
```diff
@@ -304,61 +302,58 @@ import { mergeConfig } from "astro/config";
...
-**Type:** `(userConfig: any, root: string, cmd: string): Promise<AstroConfig>`
+**Type:** <code>(userConfig: any, root: string, cmd: string) => Promise\<<a href="/en/reference/configuration-reference/">AstroConfig</a>\></code><br />
 <Since v="5.4.0" />
 </p>
 
-Imported from `astro/config`, validates an object as if it was exported from `astro.config.mjs` and imported by Astro.
-
-
-It takes the following arguments: 
+Validates an object as if it was exported from `astro.config.mjs` and imported by Astro. This takes the following arguments:
```
yanthomasdev: I think this should be a "were" instead?
```suggestion
Validates an object as if it were exported from `astro.config.mjs` and imported by Astro. This takes the following arguments: 
```

yanthomasdev (commented):
Great work here, just two small final boss nits

PR #13223, opened 2026-02-03
Document new emitClientAsset function

matthewp:
#### Description (required)

`emitClientAsset` is a new function exported from `astro/assets/utils` that's useful for integration authors that want to emit an asset that should be available in the client asset folder (such as images).

#### For Astro version: `6.x`. See astro PR [#15345](https://github.com/withastro/astro/pull/15345).

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -566,10 +566,10 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
     }
     ```
 
-3. Update your custom worker entry file to use standard Worker syntax. Import the handler from `@astrojs/cloudflare/entrypoints/server` and export a standard Cloudflare Worker object, alongside any custom exports like Durable Objects:
+3. Update your custom worker entry file to use standard Worker syntax. Import the handler from `@astrojs/cloudflare/handler` and export a standard Cloudflare Worker object, alongside any custom exports like Durable Objects:
 
     ```ts title="src/worker.ts"
-    import handler from '@astrojs/cloudflare/entrypoints/server';
+    import { handle } from '@astrojs/cloudflare/handler';
     import { DurableObject } from 'cloudflare:workers';
 
     export class MyDurableObject extends DurableObject<Env> {
@@ -579,7 +579,7 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
     export default {
       async fetch(request, env, ctx) {
         await env.MY_QUEUE.send('log');
-        return handler.fetch(request, env, ctx);
+        return handle(request, env, ctx);
       },
       async queue(batch, _env) {
         let messages = JSON.stringify(batch.messages);
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -126,6 +126,7 @@ export const sidebar = [
 					'reference/modules/astro-env',
 					'reference/modules/astro-i18n',
 					'reference/modules/astro-middleware',
+					'reference/modules/astro-static-paths',
 					'reference/modules/astro-transitions',
 					'reference/modules/astro-zod',
 				],
--- a/src/content/docs/en/guides/cms/cloudcannon.mdx
+++ b/src/content/docs/en/guides/cms/cloudcannon.mdx
@@ -17,7 +17,7 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 [CloudCannon](https://cloudcannon.com) is a Git-based headless content management system that provides a visual editor for your content and UI components, providing a rich, live editing experience.
 
 :::tip
-If you're starting a project from scratch, you can use the [CloudCannon Astro Starter Template](https://cloudcannon.com/templates/astro-starter/) with commonly used CloudCannon features to cut down on setup time.
+If you're starting a project from scratch, you can use the [CloudCannon Astro Sendit Template](https://cloudcannon.com/templates/sendit/?ssg=astro) with commonly used CloudCannon features to cut down on setup time.
 :::
 
 ## Integrating with Astro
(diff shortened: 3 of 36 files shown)


src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -1111,6 +1112,32 @@ const metadata = await emitImageMetadata(imageId);
...
+### `emitClientAsset()`
+
+<p>
+
+**Type:** `(pluginContext: Rollup.PluginContext, options: Rollup.EmitFile) => string`<br />
+
+<Since v="6.0.0" />
+</p>
+
+Emits a client asset and tracks it for later movement to the client directory during SSR builds. This function is intended for integration authors who need to emit assets (such as images) from server-rendered content that should be available on the client.
```
sarah11918: Just trying to make this a little smoother if we can...

```suggestion
Emits a client asset that will be moved to the client directory for assets (e.g. `dist/client/_astro/`) during SSR builds. This function is intended for integration authors who need to emit assets (such as images) from server-rendered content that should be available on the client.
```
UPDATED MY SUGGESTION to make the specific directory an *example* that at least people will understand, and can adjust as necessary for their situation, I think. What about that?

---
If this is correct, I think we often spell out the location of this directory when we refer to it, at least once! (Then, it's fine to talk about "the client output directory" below).

"later movement" sounded a little strange, and I'm not sure whether it's the *tracking* that's really important here? Looking at the implementation PR, it seems like this is a way for Vite to analyze which assets need to be client assets. Is *how that happens* important (e.g. returning the asset handle identifier below? Will the integration author ever need to use that, or is it just interpreted by Vite?). Or is it only important THAT assets are specified this way?

If I shorten to the above, what necessary context would integration authors be missing? ("tracking for later movement" is just kind of a slightly weird thing to say in docs like this, so maybe if we start with that part removed, we'll think of something better to add back in, if it's needed!)

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -1111,6 +1112,32 @@ const metadata = await emitImageMetadata(imageId);
...
+```ts
+import { emitClientAsset } from 'astro/assets/utils';
+
+// Inside a Vite plugin's transform or load hook
+const handle = emitClientAsset(this, {
+  type: 'asset',
+  name: 'my-image.png',
+  source: imageBuffer,
+});
+// Returns the asset handle (e.g., "a]b2c3d4")
```
sarah11918: As mentioned above, it's not clear to me (and maybe just me!) what *use* this string is. Is the point that an integration author might need to reference or identify this asset with this string later? Is it that Vite understands what this handle means and knows that it should copy over to the build? (identifies it as an asset for Vite's manifest)? I think the later is maybe an implementation detail that's not necessary to document here, since the function will achieve the *purpose* the integration author wants? (The asset copied over).

But if you want to take the opportunity to mention how e.g. 
```suggestion
// Returns the asset handle (e.g., "a]b2c3d4")
// to identify it as an asset to be copied
```

Then that could give some context!

I'll have to leave it to you to let me know what makes sense here!

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -1111,6 +1112,32 @@ const metadata = await emitImageMetadata(imageId);
 // }
 ```
 
+
+### `emitClientAsset()`
+
+<p>
+
+**Type:** `(pluginContext: Rollup.PluginContext, options: Rollup.EmitFile) => string`<br />
```
ArmandPhilippot: We have precedent for linking to external docs (e.g. Vite), so as Sarah said this could be useful to know what is expected exactly.
Unfortunately, not all types are documented so this is tricky to do. I was able to find a link for `PluginContext` (hopefully, this is the same "plugin context") but I don't see anything for `EmitFile` (unless this is the union describe in https://rollupjs.org/plugin-development/#this-emitfile ?):
```suggestion
**Type:** <code>(pluginContext: <a href="https://rollupjs.org/plugin-development/#plugin-context">Rollup.PluginContext</a>, options: Rollup.EmitFile) => string</code><br />
```

src/content/docs/en/reference/modules/astro-assets.mdx
matthewp: The location depends on which adapter you're using. It's a combination of https://docs.astro.build/en/reference/configuration-reference/#buildclient and https://docs.astro.build/en/reference/configuration-reference/#buildassets. `dist/client/_astro` is accurate for the Node adapter, so if it's ok that this is a little hand-wavey and not precise, I think this change is fine.

"For later movement" is implementation detail and good to remove. What you're going for here is better for sure.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -591,6 +591,23 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
 
 The manifest is now created internally by the adapter, so it does not need to be passed to your handler.
 
+### Removed: `cloudflareModules` option
+
+The `cloudflareModules` adapter option has been removed because it is no longer necessary. Cloudflare natively supports importing `.sql`, `.wasm`, and other module types.
+
+Remove the `cloudflareModules` option from your Cloudflare adapter configuration if you were using it:
+
+```diff
```
``````suggestion
```ts title="astro.config.mjs" del={5}
``````
I think this should do it!

File: src/content/docs/en/guides/integrations-guide/cloudflare.mdx [hunk 5]
```diff
@@ -591,6 +591,23 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
...
+import cloudflare from '@astrojs/cloudflare';
+
+export default defineConfig({
+  adapter: cloudflare({
+-   cloudflareModules: true
```
```suggestion
    cloudflareModules: true
```

src/content/docs/en/reference/modules/astro-assets.mdx
matthewp: It's the return type of this function: https://rollupjs.org/plugin-development/#this-emitfile

And as you can see its used for all sorts of things, which I don't think we should document ourselves. Could just link to this page, would that be ok?

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -1111,6 +1112,32 @@ const metadata = await emitImageMetadata(imageId);
...
+
+<Since v="6.0.0" />
+</p>
+
+Emits a client asset and tracks it for later movement to the client directory during SSR builds. This function is intended for integration authors who need to emit assets (such as images) from server-rendered content that should be available on the client.
+
+Use this instead of `pluginContext.emitFile()` directly when working in a Vite plugin context and you need the emitted asset to be moved to the client output directory.
```
ArmandPhilippot: Not sure about the syntax, but a link to Rollup docs could be useful here I think:
```suggestion
Use this instead of [Rollup `pluginContext.emitFile()`](https://rollupjs.org/plugin-development/#this-emitfile) directly when working in a Vite plugin context and you need the emitted asset to be moved to the client output directory.
```

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -1111,6 +1112,32 @@ const metadata = await emitImageMetadata(imageId);
...
+
+```ts
+import { emitClientAsset } from 'astro/assets/utils';
+
+// Inside a Vite plugin's transform or load hook
+const handle = emitClientAsset(this, {
```
ArmandPhilippot: I guess `this` refers to "Vite plugin's transform or load hook" context. But extracted, I'm not sure this makes sense to users? Is there something else we can use here? Or should this be wrapped in a context where the use of `this` makes sense?

src/content/docs/en/reference/modules/astro-assets.mdx
sarah11918: @matthewp Thanks, updated my suggestion to see if that bridges the gap between handwavy and helpful. Do you think *that* works?

But also, I *think* this is a situation where being a little handwavy is fine? I *think* the helpfulness of seeing the `/_astro/` directory outweighs that this is a common, but particular case. (I discussed with Armand that we don't document the `/_astro/` directory specifically to just be able to link there, but maybe we should!)

Would also be OK with no directory! I'll let you make the final call!

src/content/docs/en/reference/modules/astro-assets.mdx
sarah11918: Sure! I love linking out to things!

src/content/docs/en/reference/modules/astro-assets.mdx
matthewp: Yeah definitely, inside of a example plugin is probably what i should do. will update.

src/content/docs/en/reference/modules/astro-assets.mdx
matthewp: done

src/content/docs/en/reference/modules/astro-assets.mdx
matthewp: updated to show a small vite plugin.

src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -1111,6 +1112,38 @@ const metadata = await emitImageMetadata(imageId);
...
+function myVitePlugin() {
+  return {
+    name: 'my-plugin',
+    transform(code, id) {
+      const handle = emitClientAsset(this, {
+        type: 'asset',
+        name: 'my-image.png',
+        source: imageBuffer,
+      });
+      // Returns the asset handle, see the [Rollup emitFile docs](https://rollupjs.org/plugin-development/#this-emitfile)
```
```suggestion
      // Returns the asset handle similar to `emitFile()`
```
Since we have a link to this above (and since Markdown links won't display within a code block) we could do something like this? Not sure if "similar to" is technically correct (enough)? I was even wondering if we could get away with saying "Returns the asset handle from `emitFile()`" to  imply the comparison?

But, let's find a way to suggest that this works just like `emitFile()` without the link! It's referenced above by name so I think as long as we SAY `emitFile()` then it's reasonable to expect people to be able to click earlier link for more info if they want it.

Other than that, this is great!

Issue #12943, opened 2025-12-17
Document building a preview entrypoint


florian-lefebvre:
### 📚 Subject area/topic

Adapter API

### 📋 Page(s) affected (or suggested, for new content)

https://docs.astro.build/en/reference/adapter-reference/#previewentrypoint

### 📋 Description of content that is out-of-date or incorrect

We do not document at all how to make an adapter support `astro preview`. We should probably have a section "Building a preview entrypoint" just like we have "Building a server entrypoint".

PR #13229, opened 2026-02-05
[next] improve adapter type docs

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/adapter-reference.mdx
+++ b/src/content/docs/en/reference/adapter-reference.mdx
@@ -294,6 +294,54 @@ export default function createIntegration() {
 }
 ```
 
+
+## Custom prerenderer
+
+<p><Since v="6.0.0" /></p>
+
+Adapters can provide a custom prerenderer to control how pages are prerendered by using the [`setPrerenderer()`](/en/reference/integrations-reference/#setprerenderer-option) function in the `astro:build:start` hook.
+
+The following example shows how an adapter can set a custom prerenderer:
+
+```js title="my-adapter.mjs"
+export default function createIntegration() {
+  return {
+    name: '@example/my-adapter',
+    hooks: {
+      'astro:build:start': ({ setPrerenderer }) => {
+        setPrerenderer((defaultPrerenderer) => ({
+          name: 'my-prerenderer',
+          async setup() {
+            // Start a preview server
+          },
+          async getStaticPaths() {
+            // Returns array of { pathname: string, route: RouteData }
+            return defaultPrerenderer.getStaticPaths();
+          },
+          async render(request, { routeData }) {
+            // request: Request, options: { routeData: RouteData }
+            // Custom rendering logic, e.g. make HTTP requests to a preview server
+            const response = await fetch(`http://localhost:4321${new URL(request.url).pathname}`);
+            return response;
+          },
+          async teardown() {
+            // Stop the preview server
+          }
+        }));
+      },
+      'astro:config:done': ({ setAdapter }) => {
+        setAdapter({
+          name: '@example/my-adapter',
+          serverEntrypoint: '@example/my-adapter/server.js',
+        });
+      },
+    },
+  };
+}
+```
+
+The factory function receives the default prerenderer, allowing you to wrap or extend its behavior. This is useful when you only need to customize specific aspects of prerendering.
+
 ## Building a server entrypoint
 
 You will need to create a file that executes during server-side requests to enable on-demand rendering with your particular host. Astro's adapter API attempts to work with any type of host and gives a flexible way to conform to the host APIs.
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -126,6 +126,7 @@ export const sidebar = [
 					'reference/modules/astro-env',
 					'reference/modules/astro-i18n',
 					'reference/modules/astro-middleware',
+					'reference/modules/astro-static-paths',
 					'reference/modules/astro-transitions',
 					'reference/modules/astro-zod',
 				],
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -591,6 +591,23 @@ If you were using a custom `workerEntryPoint` configuration in the adapter optio
 
 The manifest is now created internally by the adapter, so it does not need to be passed to your handler.
 
+### Removed: `cloudflareModules` option
+
+The `cloudflareModules` adapter option has been removed because it is no longer necessary. Cloudflare natively supports importing `.sql`, `.wasm`, and other module types.
+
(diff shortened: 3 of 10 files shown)


florian-lefebvre:
@ArmandPhilippot as part of this PR, I want to update the "Building a server entrypoint" section to use the new API. However, we still need to document `createExports` and `start`. What do you think we should do?

ArmandPhilippot:
To be honest, I'm not sure what those changes are and so, I'm trying to get a clearer picture first to be helpful.

If I understand correctly:
* "Building a server entrypoint" is only required with `legacy-dynamic` because of `serverEntrypoint`? And that's what you want to emphasize?
* Does that means `astro/app` and `astro/app/node` are only useful with `legacy-dynamic` or should these sections be moved because they are also useful for the preview entrypoint?

If they are useful for every entrypoints, maybe something like the following is better (and shapes things for #12943):

```
## Building a server entrypoint
... available when using `entryType: "legacy-dynamic"`...
### `createExport()`
### `start()`
## Entrypoints utilities
### `astro/app`
### `astro/app/node`
```

Does this help you? Did I misunderstand your question?

florian-lefebvre:
Basically building an adapter for legacy-dynamic or self looks pretty different, eg. we can't just rely on the section on `entryType` for the new API. What I was thinking was maybe have 2 sections, one on building an adapter but the old and the new API?

`astro/app` and `astro/app/node` may still be used by the new API

ArmandPhilippot:
Okay thanks, this is clearer! So, this is more an issue about "Building an adapter" than "Building a server entrypoint" and what's tricky is mixing guide and API reference?

What if:
* "Building an adapter" becomes a guide (what you want to describe, and yes subsections sounds good... not sure about the name yet, but maybe this will be easier once we'll have the content)
* everything in "Building an adapter" is moved in "Adapter types reference" and now we have a proper link for the `AstroAdapter` type
* we can keep "Building a server entrypoint" as a top-level heading (I think this makes sense if we have "Building a preview entrypoint" later) but we precise when this is useful (either `entryType: "legacy-dynamic"` or link back to a subsection in "Building an adapter")
* we move `astro/app` and `astro/app/node` in another section

Something like this could work I think:
```
What is an adapter?
Building an adapter
    ...AstroAdapter is now a reference in "Adapter types reference" and this becomes a guide to build an adapter with the two different approaches...
Building a server entrypoint
    ... available when using `entryType: "legacy-dynamic"` (or link back to a subsection in "Building an adapter")...
    createExports()
    start()
Entrypoints utilities (or something else if you have a better name to classify them)
    astro/app
    astro/app/node
Astro features
Adapter features
Adapter types reference
    AstroAdapter
        name
        serverEntrypoint
        supportedAstroFeatures
        adapterFeatures
        args
        client
        exports
        previewEntrypoint
        entryType
    AdapterSupport
    AdapterSupportsKind
    AdapterSupportWithMessage
Allow installation via astro add
```

sarah11918:
Will wait until you've chatted and this PR is undrafted to make my comments, but the name `legacy-dynamic` for an Astro default is kind of terrible, no? 😅 

Docs isn't a history lesson. They don't care that this is a holdover from something older. If it's the default, it should honestly have a more neutral name. All those notices now read like "only available with a legacy setting!"

florian-lefebvre:
Mmmh, yeah I guess my idea of revamping the guide is just too much for now. Maybe my PR as it is is enough

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -302,6 +307,8 @@ You will need to create a file that executes during server-side requests to enab
 
 An exported function that takes an SSR manifest as its first argument and an object containing your adapter [`args`](#args) as its second argument. This should provide the exports required by your host.
 
+This property is only available with [`entryType: "legacy-dynamic"`](#entrytype).
```
ArmandPhilippot: This is not really a property here, so maybe we should include that in another way:
```suggestion
An exported function that takes an SSR manifest as its first argument and an object containing your adapter [`args`](#args) as its second argument. This is only available with [`entryType: "legacy-dynamic"`](#entrytype) and should provide the exports required by your host.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -360,6 +367,8 @@ export function createExports(manifest, args) {
 **Type:** <code>(manifest: <a href="/en/reference/integrations-reference/#ssrmanifest">SSRManifest</a>, options: any) => Record\<string, any\></code>
 </p>
 
+This property is only available with [`entryType: "legacy-dynamic"`](#entrytype).
+
 An exported function that takes an SSR manifest as its first argument and an object containing your adapter [`args`](#args) as its second argument.
```
ArmandPhilippot: This should start with the definition:
```suggestion
An exported function that takes an SSR manifest as its first argument and an object containing your adapter [`args`](#args) as its second argument. This is only available with [`entryType: "legacy-dynamic"`](#entrytype).
```

ArmandPhilippot (commented):
Both are fine with me! We can scope this PR to the current changes.

But, I think we still need to make some changes on that page based on your feedback (e.g. `astro/app` is misplaced and this is the whole "Building a server entrypoint" section which is is only available with `entryType: "legacy-dynamic"`).

I can work on revamping the sections in another PR. And while I'm at it, maybe include #12943! Then, I'll ping you to check if this is what you had in mind and/or to do more improvements?

> All those notices now read like "only available with a legacy setting!"

If we have a subsection under "Building an adapter", we could use that link instead to avoid having "legacy" everywhere. 😅 But, yes, if can find a better name before v6 become stable that would be better!

florian-lefebvre:
Yeah then let's keep this one small, thank you!

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -62,6 +62,8 @@ Defines a unique name for your adapter. This will be used for logging.
 
 Defines the entrypoint for on-demand rendering.
 
+This property is only available with [`entryType: "legacy-dynamic"`](#entrytype).
```
sarah11918, line 65: Can we maybe update all these messages to say something like:

```suggestion
This property is not available if you have specified a `custom entrypoint`](#entrytype).
```
I still maintain it's terribly misleading and unhelpful to have the word "legacy" thrown around this whole page.

(I had also wondered about separating these relevant properties into a "building your own" section, like Armand suggested. But for now, this would be a quick fix.)

florian-lefebvre: I think the suggestion makes it more confusing unfortunately? I think it's okay as is for now altho the naming is bad, let's use the week to find a better name for this stuff (will create a discord thread)

sarah11918: OK, I trust you @florian-lefebvre ! You make the final call!

sarah11918:
(Triggering a new Netlify build. Both GitHub and Netlify are having trouble today!)

PR #13235, opened 2026-02-07
Added Vault CMS stub page.

davidvkimball:
#### Description (required)

Change: adding [Vault CMS](https://vaultcms.org) to the list.

- Suggested label: `3rd-party guide`

Discord username: `davidvkimball`

### DIFF
--- a/src/content/docs/es/getting-started.mdx
+++ b/src/content/docs/es/getting-started.mdx
@@ -5,9 +5,6 @@ i18nReady: true
 tableOfContents: false
 editUrl: false
 next: false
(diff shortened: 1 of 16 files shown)


sarah11918:
Note for translators: It is **OK** to translate this page in v5 docs!

sarah11918:
Will merge this in to v5 docs today, but remember we DO have a general v5 docs freeze on. So, **no updates to this page may be made before v6**. It can be translated for v5, but we are not otherwise accepting updates to docs pages!

PR #13245, opened 2026-02-10
[i18nIgnore] fix: i18n-guides/Japanese.md

Yusaku01:
#### Description (required)
- Colons are used fairly frequently in the English version, but the Japanese translation guide
   did not have an explicit rule on whether to keep them as colons or replace them with  something else.                                                                          
- In the existing Astro Docs Japanese translations, colons tend to be replaced with periods   
  (。), though a fair number of colons still remain.
- While this may seem like a minor difference to readers, having to mentally equate colons and
   periods while reading adds unnecessary cognitive load.
- There may be cases where a colon is appropriate, but in Japanese, the following approaches
  feel more natural:
    - Replacing with a period (。)
    - Supplementing with words like "以下..." (below/following) to direct the reader's attention
   to what comes next

For the reasons above, I created this PR.
I realize the timing may not be ideal given the ongoing v6 preparation, but without an explicit rule, we risk continuing to produce inconsistent translations that create a stressful reading experience for users🤔

#### Additional info

It might be worth illustrating this rule with conforming and non-conforming examples, similar to the "Spaces around alphanumeric characters (### 英数字前後のスペース)" section.


Yusaku01:
I've moved this to draft for now, as I don't think it's urgent.

i18n-guides/日本語.md
```diff
@@ -23,6 +23,7 @@ Astroの日本語翻訳にはAstroユーザーであり、ネイティブレベ
   - 使うことができます → 使えます
 - メニュー項目など表示領域が限られている箇所（たとえば`src/i18n/ja/ui.ts`のRight Sidebar）では、たとえば「このページを編集」などのように、可能であれば文末の「する」を省略します（「このページを編集する」とは訳さないでください）
   - ただし、「貢献する」のように目的語などがなく「名詞+する」だけの場合は「する」を省略しません
+- 基本的に英語版の例示に用いられる半角コロン（:）は、句点（。）に置き換え翻訳します
```
morinokami, line 26: この記述自体は問題ないと思いますが、下部にインデントして例を補ってあげたほうがよりわかりやすいかなと思いました（「たとえば〜という英文は、〜ではなく〜のように訳します」みたいな感じで）。

sarah11918:
Hi! Just checking on the status of this PR as there hasn't been any activity for a couple of weeks and we would like to freeze the v5 docs branch very soon!

Is any assistance needed here? @Yusaku01 @morinokami

morinokami:
@Yusaku01 Could you take a look at this when you get a chance? I think the fix based on my comment should be pretty quick. If you’re busy or won’t be able to get to it soon, I can handle the change myself, so feel free to let me know!

sarah11918:
@morinokami Please do handle this when you are able to! I would like really like to stop merges to the main branch as soon as I can! 🫡

i18n-guides/日本語.md
morinokami: @Yusaku01 こちら、自分の方で以下のように修正しておきますね:

```suggestion
- 基本的に英語版の例示に用いられる半角コロン（:）は、句点（。）に置き換え翻訳します
  - たとえば「Run the following command in your terminal to start the install wizard:」という英文は「以下のコマンドをターミナルで実行して、インストールウィザードを起動します。」のように訳します
```

morinokami:
@Yusaku01 I’ve added an additional comment. If there’s any problem with it, feel free to open another PR to update it after v6 is released.

@sarah11918 I tweaked the changes and it looks LGTM, so I’m going to merge it after CI passes!

morinokami:
@sarah11918 But would it be better to close it instead of merging? I’d like to get your opinion.

sarah11918:
@morinokami If this is a helpful change, then I am happy to let this one merge! I am only trying to stop new PRs right now because we have to make a cut off somewhere. 😄  But this one has been here for a while and I was expecting it to get merged. Thanks for handling it!

morinokami:
@sarah11918  Got it, thanks! Since it’s a helpful change, I’ll go ahead and merge it. 👍

PR #13246, opened 2026-02-10
support serverEntrypoint for self adapters

florian-lefebvre:
#### Description (required)

I know the docs are still not ideal but the changes are incremental. First this, then later on better naming and clear legacy path

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/reference/adapter-reference.mdx
+++ b/src/content/docs/en/reference/adapter-reference.mdx
@@ -62,8 +62,6 @@ Defines a unique name for your adapter. This will be used for logging.
 
 Defines the entrypoint for on-demand rendering.
 
-This property is only available with [`entryType: "legacy-dynamic"`](#entrytype).
-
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -62,8 +62,6 @@ Defines a unique name for your adapter. This will be used for logging.
 
 Defines the entrypoint for on-demand rendering.
 
-This property is only available with [`entryType: "legacy-dynamic"`](#entrytype).
```
ArmandPhilippot, line 65: If `serverEntrypoint` is now supported with `self`, does this mean that `createExports()` and `start()` are also supported (and should we also revert the changes made to these sections)?

florian-lefebvre: No it's only `serverEntrypoint` :+1:

PR #13247, opened 2026-02-10
Add featured CMS grid to navigation

delucis:
#### Description (required)

This PR updates the navigation component for CMS guides to be able to highlight featured platforms from our OSS sponsors. Starting with [CloudCannon](https://cloudcannon.com/) 🎉

### DIFF
--- a/src/content/docs/fr/guides/cms/cloudcannon.mdx
+++ b/src/content/docs/fr/guides/cms/cloudcannon.mdx
@@ -17,7 +17,7 @@ import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 [CloudCannon](https://cloudcannon.com) est un système de gestion de contenu headless basé sur Git qui fournit un éditeur visuel pour votre contenu et vos composants d'interface utilisateur, offrant une riche expérience d'édition en direct.
 
 :::tip
-Si vous démarrez un projet à partir de zéro, vous pouvez utiliser le [modèle de démarrage Astro de CloudCannon](https://cloudcannon.com/templates/astro-starter/) avec les fonctionnalités de CloudCannon couramment utilisées pour réduire le temps de configuration.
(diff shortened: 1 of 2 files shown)


src/data/logos.ts
```diff
@@ -15,7 +24,7 @@ export const logos = LogoCheck({
 	caisy: { file: 'caisy.svg', padding: '.05em' },
 	cleavr: { file: 'cleavr.svg', padding: '0.125em 0.125em 0.1375em' },
 	'clever-cloud': { file: 'clever-cloud.svg', padding: '.09em' },
-	cloudcannon: { file: 'cloudcannon.svg', padding: '.25em' },
+	cloudcannon: { file: 'cloudcannon.svg', padding: '.25em', bg: 'var(--sl-color-accent)' },
```
delucis: This got lost in a merge, but 55e08cb9784a9f9864cdb974fc407893a42ebc5e added the correct brand colour here:
```suggestion
	cloudcannon: { file: 'cloudcannon.svg', padding: '.25em', bg: '#034ad8' },
```

src/data/logos.ts
delucis: I also added a colour for Sanity apparently. Might as well bring that back too for completeness:
```suggestion
	sanity: { file: 'sanity.svg', padding: '.15em', bg: '#F03E2F' },
```

sarah11918:
Thanks @delucis and @HiDeoo ! I'm just checking in both light and dark, various screen sizes... one initial thing: do you think we still need "More CMS Guides" now that we have *both* Featured CMS and All CMS? Maybe we don't need "More" at all? Or do we think that's good from an organization/hierarchy standpoint? Just feels like a lot of titles in quick succession!

image

delucis:
> do you think we still need "More CMS Guides" now that we have _both_ Featured CMS and All CMS? Maybe we don't need "More" at all? Or do we think that's good from an organization/hierarchy standpoint? Just feels like a lot of titles in quick succession!

Yeah I also wondered this. I think _some_ kind of separation is good to clarify this is separate from the main content? Doesn’t have to be a heading necessarily though I guess. (We’d need to bump up the heading levels though if we did that.)

sarah11918:
I think it's fine for now! Technically, they are both "more". Happy to let you make the call, but it did stand out.

PR #13248, opened 2026-02-10
fix(v6): reorganize `@astrojs/sitemap` sections

ArmandPhilippot:
#### Description (required)

**Context:** `SitemapItem` is used by `serialize()` and `chunks`, it might be useful to describe it in a separate entry and use links to that section. While checking the available types, I noticed some others public types are not documented.

- Adds a new "types reference" section for `ChangeFreq`, `ChangeFreqEnum`, `LinkItem`, `SitemapItem`, and `SitemapOptions`
  - `LinkItem`:
    - `hreflang`/`lang`: they are [the same](https://github.com/ekalinin/sitemap.js/blob/723d8e7db5a52d5ce52c7539b35250608c058341/CHANGELOG.md#634) but [if `lang` is defined `hrefLang` will never be used](https://github.com/ekalinin/sitemap.js/blob/723d8e7db5a52d5ce52c7539b35250608c058341/lib/sitemap-item-stream.ts#L231)
- Adds a new "utilities reference" section for `ChangeFreqEnum`
- Updates `serialize()` description
- Fixes some code snippets related to `SitemapItem`:
  - `lastmod` should be a string (ISO formatted date)
  - `changefreq`: the value is valid, but Typescript doesn't like it... It seems that's why `ChangeFreqEnum` has been exported in https://github.com/withastro/astro/pull/7028 and so it might be better to use something that doesn't cause red squiggles

- Suggested label: v6, add new content, improve or update existing documentation


src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -602,6 +588,164 @@ export default defineConfig({
...
+Defines the priority of this URL relative to other URLs on your site. The value should be a number in the range from `0.0` to `1.0`.
+
+#### `SitemapItem.links`
+
+<p>
+
+**Type:** <code><a href="#linkitem">LinkItem</a>[] | undefined</code>
+</p>
+
+Defines a list of alternate pages including the current page.
```
yanthomasdev: small nit
```suggestion
Defines a list of alternate pages, including the current page.
```

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -346,34 +346,20 @@ export default defineConfig({
...
+### `serialize()`
 
 <p>
 
-**Type:** `(item: SitemapItem) => SitemapItem | Promise<SitemapItem | undefined> | undefined`
+**Type:** <code>(item: <a href="#sitemapitem">SitemapItem</a>) => SitemapItem | Promise\<SitemapItem | undefined\> | undefined</code>
 </p>
 
-A function called for each sitemap entry just before writing to a disk. This function can be asynchronous.
+A function called for each sitemap entry just before writing to disk. This function can be asynchronous and must return a [`SitemapItem`](#sitemapitem), whether touched or not, or `undefined` to remove it from the sitemap.
```
sarah11918: "touched" doesn't feel like a very docs-y word. What would be more precise here? "used", "implemented", etc.?

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -602,6 +588,164 @@ export default defineConfig({
...
+A union of valid values to specify the update frequency of an entry.
+
+### `LinkItem`
+
+<p>
+
+**Type:** `{ lang: string; hreflang?: string; url: string; }`
+</p>
+
+Describes an alternative version of the page.
```
sarah11918: This feels a little vague to me? I'm not sure what the alternative in question is, nor why I might need to specify this?

(Note: some of this may be a Sarah Skillz Issue, and that's fine! I just want to make sure)

Is alternative in opposition to "canonical" maybe? I would love to have more info here!

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -602,6 +588,164 @@ export default defineConfig({
...
+
+#### `LinkItem.lang`
+
+<p>
+
+**Type:** `string`
+</p>
+
+Specifies the language code supported by this version of the page. When a value is set, you do not need to also set [`hreflang`](#linkitemhreflang).
```
sarah11918, line 666: I'm going to assume that once we negotiate the earlier comment, then saying "this version of the page" will either make perfect sense, or we'll know exactly how to update it!

src/content/docs/en/guides/integrations-guide/sitemap.mdx
```diff
@@ -602,6 +588,164 @@ export default defineConfig({
...
+Specifies the language code supported by this version of the page. When a value is set, you do not need to also set [`lang`](#linkitemlang).
+
+#### `LinkItem.url`
+
+<p>
+
+**Type:** `string`
+</p>
+
+Specifies the fully-qualified URL for the version of this page for the specified language.
```
sarah11918: I'm not entirely sure "fully-qualified URL" is a known thing? Is there something we can link to here for a definition, or is there a way of saying it that we use elsewhere in docs? (I don't actually know exactly what you mean by this. Is it just a full, absolute URL?)

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: I would say "updated"? The function let us update the item properties.
```suggestion
A function called for each sitemap entry just before writing to disk. This function can be asynchronous and must return a [`SitemapItem`](#sitemapitem), whether updated or not, or `undefined` to remove it from the sitemap.
```

Otherwise "altered" maybe? Or "whether the object's properties have been modified or not"?

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: Actually, good call! "alternative" is probably superfluous here. This is used by `SitemapItem.links` (so I was trying to not repeat the same description) but this also contains the current version of the page. "Describes a version of the page" is probably enough, but I think you'll still find this confusing. 😄  But I lack of inspiration here...

`LinkItem` contains the data used to define [`<link rel="alternate" hreflang="" />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel#alternate). And "alternate" is a bit misleading, because the current version of the page should also be added.

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: Yes, just full/absolute! We could use "absolute" if you find this less confusing?
I don't think I have a good link for URL. This is often used for domain names (e.g. https://en.wikipedia.org/wiki/Fully_qualified_domain_name). So a fully-qualified URL would be a fully qualified domain name + the slug.

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: Maybe something like:
> Describes the URL of a page. This could be the default version of the document or one of its translations.

src/content/docs/en/guides/integrations-guide/sitemap.mdx
sarah11918: OK, that's a lot more clear, thanks! So, now my question is, how *important* is it to say "whether updated or not"? Do we have people asking/clamouring for ISR in sitemaps? 😅 What would we lose or be confusing IF we just said:

> ... a `SitemapItem` or `undefined` to exclude it from the sitemap.

This way, it feels obvious that yes, every item needs to either return proper sitemap entry stuff, or the things that are not to appear in the sitemap should return undefined. (Slight nuance shift from assuming something is there to "remove" from the user's perspective to simply say, this should not be in the sitemap). Again, could be a Sarah Skillz thing! What is missing from the above, so that we know what we NEED to add back?

src/content/docs/en/guides/integrations-guide/sitemap.mdx
sarah11918: Ooh, that feels way better!

src/content/docs/en/guides/integrations-guide/sitemap.mdx
sarah11918: If we use your suggestion above, are we cool with "this version of the page"? (I think so?)

src/content/docs/en/guides/integrations-guide/sitemap.mdx
sarah11918: OK, if it's in Wikipedia than it's for sure a known thing that people can figure out! Objection rescinded!

I'm not sure I love the "this version of the page" here though? Do you think it still fits?

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: Oh, to be honest, I just reused the wording we had (the sentence has moved and I added the `undefined` part) so I'm not opposed to remove that! 😄 But, if we remove it, I guess we loose the purpose of the function?

I mean, maybe it's the language barrier., but when I read `serialize` I don't immediately think "oh, so I can alter each item here". I think of it more like `JSON.stringify()` where you convert to another format... not the same format with updated values.

But maybe we can improve the description. Previously, I mostly trimmed it rather than rewriting it. Maybe something like:

```suggestion
Generates an editable representation of each sitemap entry before returning either a [`SitemapItem`](#sitemapitem) or `undefined` to remove it from the sitemap. This function can be asynchronous and is called for each sitemap entry just before writing to disk.
```

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: I'll merge this with the other changes, no need to tire the CI!
```suggestion
Describes the URL of a page. This could be the default version of the document or one of its translations.
```

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: Yes, I think it's now clearer what "a version" can be! Thanks for the help 😄

src/content/docs/en/guides/integrations-guide/sitemap.mdx
ArmandPhilippot: Yeah, you're right this is probably superfluous/confusing rather than helpful here:
```suggestion
Specifies the absolute URL of the page for the specified language.
```

edit: Actually, maybe we should also update that to use "absolute". If you find this confusing, it probably is for other people too, and people from various backgrounds use Astro. So it might be better to use a clearer term for everyone here!

src/content/docs/en/guides/integrations-guide/sitemap.mdx
sarah11918: Now we're talking!! That looks fantastic!

src/content/docs/en/guides/integrations-guide/sitemap.mdx
sarah11918: Yup, that reads great to me!

PR #13267, opened 2026-02-13
feat(v6): update cloudflare default image service

Princesseuh:
#### Description (required)

https://github.com/withastro/astro/pull/15435/ changes the default to use the bindings instead of compile because compile's dev behavior is not intuitive anymore (yet)


sarah11918:
Just noting that this will need an entry in the Cloudflare upgrade guide here too, where the change is clearly noted and with any upgrade guidance! https://v6.docs.astro.build/en/guides/integrations-guide/cloudflare/#upgrading-to-v13-and-astro-6

Normally when a default setting has changed, we say things like "inspect your deployed project for anything funky" in case this affects them. And if possible, we like to give advice on what to do to "revert to the previous behavior" (since presumably that was already working), which in this case seems to be explicitly setting `imageService: "compile"`

Edited to add: Will note that this won't go in the main v6 upgrade guide since it's just for Cloudflare, but you can see what other "Changed default" entries typically look like there for inspiration! https://v6.docs.astro.build/en/guides/upgrade-to/v6/#changed-defaults

OliverSpeir (approved):

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,17 +118,18 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
...
 </p>
 
-Determines which image service is used by the adapter. The adapter will default to `compile` mode when an incompatible image service is configured. Otherwise, it will use the globally configured image service:
+Determines which image service is used by the adapter. The adapter will default to `cloudflare-binding` mode when an incompatible image service is configured. Otherwise, it will use the globally configured image service:
 
 - **`cloudflare`:** Uses the [Cloudflare Image Resizing](https://developers.cloudflare.com/images/image-resizing/) service.
 - **`cloudflare-binding`:** Uses the [Cloudflare Images binding](https://developers.cloudflare.com/images/transform-images/bindings/) for image transformation. The binding is automatically provisioned when you deploy.
 - **`passthrough`:** Uses the existing [`noop`](/en/guides/images/#configure-no-op-passthrough-service) service.
-- **`compile`:** Uses Astro's default service (Sharp), but only on pre-rendered routes at build time. For pages rendered on demand, all `astro:assets` features are disabled.
+- **`compile`:** Transforms images at build time for pre-rendered routes. Uses passthrough for on-demand rendered pages.
```
sarah11918: This should still say what it *uses*, like all the other options, since this option is to define which image service is being used.

And, is the previous sentence now incorrect, that all `astro:assets` features are disabled for pages rendered on demand? That seemed a very clear way of stating it, so if it hasn't changed, I'm wondering why the wording has? Or do you want to emphasize the `passthrough` runtime behaviour? If so, then I'd write `passthrough` in code, something like:

```suggestion
- **`compile`:** Uses XXXXXXX to transform images at build time for prerendered routes. The noop `passthrough` option is configured for on-demand rendered pages.
```
(Also, note that we're standardizing docs around prerendered (no hyphen) as we get around to them, so might as well update that here!)

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,17 +118,18 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
...
 - **`custom`:** Always uses the image service configured in [Image Options](/en/reference/configuration-reference/#image-options). **This option will not check to see whether the configured image service works in Cloudflare's `workerd` runtime.**
+- **`{ build: 'compile', runtime?: 'cloudflare-binding' | 'passthrough' }`:** Same as `'compile'`, but allows configuring the runtime service used for on-demand rendered pages.
```
sarah11918: So, right now `'compile'` basically is the equivalent of `{ build: 'compile', runtime: 'passthrough'  }` right?

So, no one would ever *need* to set that explicitly? They'd only ever set `{ build: 'compile', runtime?: 'cloudflare-binding' }` ?

If so, I *might* be tempted to list out this option simply as something like:

```suggestion
- **`{ build: 'compile', runtime: 'cloudflare-binding'}`:** Uses `'compile'` at build time for prerendered routes, but uses `cloudflare-binding` as the runtime service used for on-demand rendered pages instead of the default noop `passthrough` service.
```
(The full options are shown in the type at the beginning of the listing, so we have our technical bases covered. But here we can show what someone would actually configure)

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -620,4 +621,10 @@ The Astro Cloudflare adapter now only supports deploying to Cloudflare Workers b
 
 If you need to continue using Cloudflare Pages, see [Using with Cloudflare Pages](#using-with-cloudflare-pages) for the required manual configuration.
 
+### Changed: `imageService` default
+
+The default value of `imageService` has changed from `'compile'` to `'cloudflare-binding'`. The `cloudflare-binding` service uses the [Cloudflare Images binding](https://developers.cloudflare.com/images/transform-images/bindings/) to transform images at runtime, and the binding is automatically provisioned when you deploy.
```
```suggestion
The default value of `imageService` has changed from `'compile'` to `'cloudflare-binding'` for an improved experience when working with images.

The `cloudflare-binding` service uses the [Cloudflare Images binding](https://developers.cloudflare.com/images/transform-images/bindings/) to transform images at runtime, and the binding is automatically provisioned when you deploy.
```
I don't necessarily love what I wrote, but it would be nice to have *some* statement of either why or what's good about the change! Please feel free to "fill in the blank" and correct as necessary!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -620,4 +621,10 @@ The Astro Cloudflare adapter now only supports deploying to Cloudflare Workers b
...
+
+To revert to the previous behavior, set `imageService: 'compile'` explicitly in your adapter config.
```
```suggestion
To revert to the previous behavior, where image transformation was only available on prerendered routes at build time, set `imageService: 'compile'` explicitly in your adapter config.
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
OliverSpeir: https://github.com/withastro/docs/pull/13267/commits/c7af404c310c1522b003770f1e15a22959c6b0ce 

The reason it just says locally instead of explaining exactly what it uses to transform is because its using sharp but through miniflare so its not exactly the same as it was before when we just said it uses sharp

sarah11918: But, it's still *using* something, right? ~~And, I don't see the above saying "local" anywhere~~ (sorry, can see the text was updated which isn't visible looking at this conversation view!), which certainly could be incorporated for more explanation! If this is about configuring which image service is used, I do think the option should be described in terms of what (service) is or isn't being used, something along the lines of:

 "Uses a combination of internal dependencies to transform images locally at build time for prerendered pages"

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,17 +118,18 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
 ### `imageService`
 
 <p>
-**Type:** `'passthrough' | 'cloudflare' | 'cloudflare-binding' | 'compile' | 'custom'`<br />
-**Default:** `'compile'`
+**Type:** `'passthrough' | 'cloudflare' | 'cloudflare-binding' | 'compile' | 'custom' | { build: 'compile', runtime?: 'cloudflare-binding' | 'passthrough' }`<br />
```
sarah11918, line 110:
```suggestion
**Type:** `'passthrough' | 'cloudflare' | 'cloudflare-binding' | 'compile' | 'custom' | 'compile-and-run'`<br />
```
So, I woke up this morning thinking, "If there really is only one reasonable configuration option for this thing, can we just give it a simple name (while using the implementation under the hood)?

This is a dummy name, but wouldn't it make sense to just do that?

OliverSpeir: The original idea was it opens the door to more in future, and acts as an invitation to anyone who wants to add more like custom to runtime

ascorbic: In that scenario, wouldn't you want the type to be something like `ImageService | { build: ImageService, runtime: ImageService }` rather than only allowing one?

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,17 +118,18 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
...
 - **`cloudflare-binding`:** Uses the [Cloudflare Images binding](https://developers.cloudflare.com/images/transform-images/bindings/) for image transformation. The binding is automatically provisioned when you deploy.
 - **`passthrough`:** Uses the existing [`noop`](/en/guides/images/#configure-no-op-passthrough-service) service.
-- **`compile`:** Uses Astro's default service (Sharp), but only on pre-rendered routes at build time. For pages rendered on demand, all `astro:assets` features are disabled.
+- **`compile`:** Transforms images locally at build time for pre-rendered routes. The noop `passthrough` option is configured for on-demand rendered pages.
 - **`custom`:** Always uses the image service configured in [Image Options](/en/reference/configuration-reference/#image-options). **This option will not check to see whether the configured image service works in Cloudflare's `workerd` runtime.**
+- **`{ build: 'compile', runtime: 'cloudflare-binding'}`:** Uses `'compile'` at build time for prerendered routes, but uses `cloudflare-binding` as the runtime service used for on-demand rendered pages instead of the default noop `passthrough` service.
```
```suggestion
It is also possible to configure your image service as an object, setting both a build time and runtime service independently. Currently, `'compile'` is the only available build time option. The supported runtime options are `'passthrough'` (default) and `'cloudflare-binding'`:

```js title="astro.config.mjs" ins={6}
import cloudflare from '@astrojs/cloudflare';

export default defineConfig({
  adapter: cloudflare({
    imageService: { build: 'compile', runtime: 'cloudflare-binding' }
  }),
});
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
sarah11918, line 134:
```suggestion
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,17 +118,28 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
...
+import { defineConfig } from 'astro/config';
+import cloudflare from '@astrojs/cloudflare';
+
+export default defineConfig({
+  adapter: cloudflare({
+    imageService: { build: 'compile', runtime: 'cloudflare-binding' }
+  }),
+});
 
 ```js title="astro.config.mjs" ins={6}
```
sarah11918, line 124: GitHub won't delete this code example (in favour of the one above) even though I've tried multiple times. It's not needed to have these two back to back. Furthermore, if `cloudflare-binding` is now the default, we for sure wouldn't show it in an example because no one would ever configure that.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
sarah11918: I'll not resolve this thread in case there's any response to Matt's comment here, but noting for the record that I do not intend for my suggestion here to be taken!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,17 +118,28 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
...
+Determines which image service is used by the adapter. The adapter will default to `cloudflare-binding` mode when an incompatible image service is configured. Otherwise, it will use the globally configured image service:
 
 - **`cloudflare`:** Uses the [Cloudflare Image Resizing](https://developers.cloudflare.com/images/image-resizing/) service.
 - **`cloudflare-binding`:** Uses the [Cloudflare Images binding](https://developers.cloudflare.com/images/transform-images/bindings/) for image transformation. The binding is automatically provisioned when you deploy.
 - **`passthrough`:** Uses the existing [`noop`](/en/guides/images/#configure-no-op-passthrough-service) service.
-- **`compile`:** Uses Astro's default service (Sharp), but only on pre-rendered routes at build time. For pages rendered on demand, all `astro:assets` features are disabled.
+- **`compile`:** Transforms images locally at build time for pre-rendered routes. The noop `passthrough` option is configured for on-demand rendered pages.
```
```suggestion
- **`compile`:** Uses a combination of internal dependencies to transform images locally at build time for prerendered routes. The noop `passthrough` option is configured for on-demand rendered pages.
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
OliverSpeir: Yes probably the ideal is { build: ImageService, dev: ImageService, runtime: ImageService }

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,28 +118,28 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
...
 import { defineConfig } from 'astro/config';
 import cloudflare from '@astrojs/cloudflare';
 
 export default defineConfig({
   adapter: cloudflare({
-    imageService: 'cloudflare-binding',
+    imageService: { build: 'compile', runtime: 'cloudflare-binding' }
   }),
 });
-```
```
sarah11918, line 134: We're missing closing ``` here!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,28 +118,28 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
...
 
 ### `sessionKVBindingName`
```
sarah11918, line 135: OK, the link checker is claiming this section doesn't exist, but clearly it does... it might be confused by the missing/open ``` so I'll try to add them in myself

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -129,7 +129,17 @@ Determines which image service is used by the adapter. The adapter will default
 - **`passthrough`:** Uses the existing [`noop`](/en/guides/images/#configure-no-op-passthrough-service) service.
 - **`compile`:** Transforms images locally at build time for pre-rendered routes. The noop `passthrough` option is configured for on-demand rendered pages.
 - **`custom`:** Always uses the image service configured in [Image Options](/en/reference/configuration-reference/#image-options). **This option will not check to see whether the configured image service works in Cloudflare's `workerd` runtime.**
-- **`{ build: 'compile', runtime: 'cloudflare-binding'}`:** Uses `'compile'` at build time for prerendered routes, but uses `cloudflare-binding` as the runtime service used for on-demand rendered pages instead of the default noop `passthrough` service.
+It is also possible to configure your image service as an object, setting both a build time and runtime service independently. Currently, `'compile'` is the only available build time option. The supported runtime options are `'passthrough'` (default) and `'cloudflare-binding'`:
```
yanthomasdev: hyphen nit, the second "build time" is being used as an adjective, so it gets hyphenated.
```suggestion
It is also possible to configure your image service as an object, setting both a build time and runtime service independently. Currently, `'compile'` is the only available build-time option. The supported runtime options are `'passthrough'` (default) and `'cloudflare-binding'`:
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -118,25 +118,26 @@ This functionality is enabled by default. If you'd like to disable, set `cloudfl
...
 - **`passthrough`:** Uses the existing [`noop`](/en/guides/images/#configure-no-op-passthrough-service) service.
-- **`compile`:** Uses Astro's default service (Sharp), but only on pre-rendered routes at build time. For pages rendered on demand, all `astro:assets` features are disabled.
+- **`compile`:** Uses a combination of internal dependencies to transform images locally at build time for prerendered routes. The noop `passthrough` option is configured for on-demand rendered pages.
 - **`custom`:** Always uses the image service configured in [Image Options](/en/reference/configuration-reference/#image-options). **This option will not check to see whether the configured image service works in Cloudflare's `workerd` runtime.**
+It is also possible to configure your image service as an object, setting both a build time and runtime service independently. Currently, `'compile'` is the only available build-time option. The supported runtime options are `'passthrough'` (default) and `'cloudflare-binding'`:
```
sarah11918, line 122:
```suggestion

It is also possible to configure your image service as an object, setting both a build time and runtime service independently. Currently, `'compile'` is the only available build-time option. The supported runtime options are `'passthrough'` (default) and `'cloudflare-binding'`:
```
Just need to make this a new paragraph to render properoy

PR #13270, opened 2026-02-16
feat(fonts): npm provider

florian-lefebvre:

- Closes #
- Suggested label:


src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -319,6 +320,112 @@ export default defineConfig({
 });
 ```
 
+### NPM
+
+Retrieves fonts from npm packages, either from locally installed packages in `node_modules` or from a CDN:
```
sarah11918:
```suggestion
Retrieves fonts from NPM packages, either from locally installed packages in `node_modules` or from a CDN:
```

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -319,6 +320,112 @@ export default defineConfig({
...
+#### Family options
+
+The provider comes with the following family-specific options that can be added in the [`font.options` object](/en/reference/configuration-reference/#fontoptions).
+
+##### `package`
+
+<p>
+
+**Type:** `string`<br />
+**Default:** Auto-detected or guessed from family name
```
sarah11918:
```suggestion
**Default:** Auto-detected or inferred from family name
```
I'm not entirely sure, but I think maybe using "infer" instead of "guess" feels more docs-y. 😄 

SPIN THAT WHEEL...

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -319,6 +320,112 @@ export default defineConfig({
...
+</p>
+
+The npm package name. When not specified, the provider will try to find the font family in known font package patterns or guess based on Fontsource conventions:
```
sarah11918:
```suggestion
The NPM package name. When not specified, the provider will try to find the font family in known font package patterns or infer based on Fontsource conventions:
```

src/content/docs/en/reference/font-provider-reference.mdx
```diff
@@ -319,6 +320,112 @@ export default defineConfig({
...
 
+### NPM
+
+Retrieves fonts from npm packages, either from locally installed packages in `node_modules` or from a CDN:
+
+```js
+provider: fontProviders.npm()
+```
+
+The provider automatically detects fonts from your `package.json` dependencies and can resolve fonts from packages like `@fontsource/*`, `@fontsource-variable/*`, and other known font packages.
```
sarah11918, line 331: Question, if you have multiple font packages in your `package.json` (would you ever?), then how does it choose/resolve?

The examples for the family options below don't show anything being passed to `fontProviders.npm()`, but yet specify e.g. the package version to use. Will it apply that package version to ALL fonts packages you have in your `package.json`? What if they all are on different current version numbers?

EDIT: Ah, I somehow skimmed over `package` too quickly! Is either `package` or `cdn` required so it's clear WHAT the source is, as appropriate?

florian-lefebvre:
FYI these docs are mostly copied from unifont so I can't take credits for writing good docs this time!

src/content/docs/en/reference/font-provider-reference.mdx
florian-lefebvre: Yes it can be manually specified if more appropriate. Otherwise let's say you don't use any extra option and pass `Roboto` as name, _I think_ it will be inferred as `@fontsource/roboto`. Does that answer your question(s)?

sarah11918: It does, but the *think* doesn't inspire me with a whole lotta confidence 😅 If it's what we have, it's what we have!

florian-lefebvre: I checked the source again and I'm pretty confident this is what it does!

PR #13253, opened 2026-02-11
improve naming of new adapter api

florian-lefebvre:
#### Related issues & labels (optional)

- Closes #
- Suggested label:

PR #13273, opened 2026-02-16
feat: add streaming arg to createApp()

florian-lefebvre:

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -104,17 +104,6 @@ Now, you can enable [on-demand rendering per page](/en/guides/on-demand-renderin
 
 The Cloudflare adapter accepts the following options:
 
-### `cloudflareModules`
-
-<p>
-**Type:** `boolean`<br />
-**Default:** `true`
-</p>
-
-Enables [imports of `.wasm`, `.bin`, and `.txt` modules](#cloudflare-module-imports).
-
-This functionality is enabled by default. If you'd like to disable, set `cloudflareModules` to `false`.
-
 ### `imageService`
 
 <p>


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -491,7 +491,7 @@ import { createApp } from 'astro/app/entrypoint';
 
 <p>
 
-**Type:** `() => App`
+**Type:** `(options?: { streaming: boolean }) => App`
```
florian-lefebvre: I chose not to document `streaming` in detail as it's already documented for `astro/app`, which is linked below

sarah11918: It's not really documented though? (I can't find it). There's a mention in text that streaming can be disabled, but it's not shown in the code example, so it's not even shown what it looks like to use it?

I think at least one of these two places should have *something* for streaming (and the content can look similar to what I'm proposing in the changeset now)

ArmandPhilippot: Yes, this is only a mention. I guess for now the easiest path would be to add a `streaming` subsection here and add a link to that section in the `astro/app` description.

But looking at `astro/app`, I wonder if we shouldn't consider having a `constructor` heading each time we document a class (and potentially subheadings for each arguments).
Since #13253, even `manifest` is no longer clear. Previously, the code snippet retrieved it via the `start()` function, but now we no longer show how the user can obtain it...

sarah11918:
Neither of us are happy anymore! (Just to be clear 😂 )

src/content/docs/en/reference/adapter-reference.mdx
florian-lefebvre: You can derail this PR if you want something more solid overall! I don't think that's a big deal if this PR is merged later than the code release

sarah11918:
Noting that now I think the changeset from the implementation PR has some text we can work with for creating a `streaming` entry!

florian-lefebvre:
So just restating that I will merge the code PR because I need it for other stuff, but we don't need to rush to get this PR ready in time for release. No one uses `createApp()` apart from us atm

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -491,7 +491,7 @@ import { createApp } from 'astro/app/entrypoint';
...
 </p>
 
 Returns an [`App` instance](#astroapp).
```
ArmandPhilippot: Using the changeset contents (no credits to me!):
```suggestion
Returns an [`App` instance](#astroapp).

#### `options.streaming`

<p>

**Type:** `boolean`<br />
**Default:** `true`
</p>

Defines whether HTML streaming is enabled. In most cases, disabling streaming is not recommended as it improves performance and generally provides a better visitor experience.

HTML streaming breaks a document into chunks to send over the network and render on the page in order. This normally results in visitors seeing your HTML as fast as possible but factors such as network conditions and waiting for data fetches can block page rendering.
```

sarah11918:
I am removing the `merge-on-release` tag because this *is* released now, and we are just getting around to the docs when we can!

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Add links for both the manifest and the streaming option:
```suggestion
The `App` constructor accepts a required [SSR manifest](/en/reference/integrations-reference/#ssrmanifest) argument, and optionally an argument to [enable or disable streaming](#optionsstreaming), defaulting to `true`.
```

ArmandPhilippot: I don't have a good suggestion here to prevent `manifest` to come from nowhere... Should the manifest be passed the same way described in "Passing build-time configuration"?

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: Hey, I like that! 😂 

I would probably *also* keep the "when/why you might want to disable it" and code sample part of the changeset too, since that's kind of important when explaining the feature's use?

(I'll let Florian comment on what is happy path/suggested usage that we want to show in docs)

florian-lefebvre:
I think it makes sense to stop documenting `astro/app` and `App` in favor of just `astro/app/entrypoint`. I didn't accept any suggestions because I feel like this comment will change quite a few things

sarah11918:
No need to accept suggestions, @florian-lefebvre ! Happy if you want to "start over" with what you think is most helpful on this page in general!

florian-lefebvre:
I am not fully satisfied with the header structure of `astro/app/entrypoint`. The existing content for `astro/app` (eg. `app.render()`) already goes down to h6s

ArmandPhilippot:
> already goes down to h6s

Welcome to our nightmare. 😄 

I agree this is a bit weird to have `createApp()` and all the methods at the same level... I don't have a good idea at the moment (I'll think about it) but I do have more questions...

Looking at the new structure, I realize this also means all the static methods are no longer relevant because we can't access them from `createApp()`, and we might need to update some descriptions. For example, `RenderOptions.addCookieHeader` currently links to `App.getSetCookieFromResponse()`...

So, are there any use cases for the following static methods (inherited from `BaseApp`)?
* `App.getSetCookieFromResponse()`
* `App.validateForwardedHost()`

Also, it seems the following static methods have been removed in v6 and so we'll need to remove them anyway:
* `App.sanitizeHost()`
* `App.validateForwardedHeaders()`

florian-lefebvre:
I don't have context on these static methods, I'm not sure if they are deprecated or removed. Given that, it may be better to keep the `astro/app` heading after all

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -491,31 +491,22 @@ import { createApp } from 'astro/app/entrypoint';
...
-addEventListener('fetch', event => {
-  event.respondWith(
-    app.render(event.request)
-  );
-});
-```
+Defines whether HTML streaming is enabled. In most cases, disabling streaming is not recommended as it improves performance and generally provides a better visitor experience.
 
-The following methods are provided:
+HTML streaming breaks a document into chunks to send over the network and render on the page in order. This normally results in visitors seeing your HTML as fast as possible but factors such as network conditions and waiting for data fetches can block page rendering.
```
``````suggestion

However, when you need to disable HTML streaming (e.g. your host only supports non-streamed HTML caching at the CDN level), you can opt out of the default behavior by passing `streaming: false` to `createApp()`:

```ts
import { createApp } from 'astro/app/entrypoint'

const app = createApp({ streaming: false })
```

``````
I know I'd said we should do this, but here's the thing I'm thinking of here!

**Comment by ArmandPhilippot** [comment]
I was moving some sections, locally, to see how this could be organized without looking odd and to workaround the `h6`s. I wonder if it would make sense to have `astro/app` in the "Runtime API" section. I know this is only used with adapters but... we already have `astro:static-paths` in there.

Here's a preview of both TOC:

|Adapter API TOC|`astro/app` TOC|
|---|---|
|Preview of the table of contents for the Adapter reference|Preview of the table of contents for astro/app|

In the Adapter API page I introduced a "Building adapter entrypoints" section in anticipation of the "Building a preview entrypoint" section. And "Passing build-time configuration" is probably for all entrypoints so we don't need to nest it under "Building a server entrypoint".

If we proceed this way:
* we probably need to think about a better name than "App API Reference" for the `astro/app` page
* even if we show `App` as an available import maybe we can just phrase the description to say this should be created with `createApp()`? Not ideal but with or without the reorganization I don't see how to separate the options from the methods returned... Unless we introduce some `Options` and `Returned methods` headings. Not possible at the moment because of the `h6`s though.
* (edit) there is also a path where we document `App` under `createApp()` with a heading like `The App instance` and we keep `import from astro/app` only if there is something else to document here

I also noticed there are a lot of available imports from `astro/app` we don't document. Do you know if we should Florian?

```js
import {
  App,
  AppPipeline, // Probably not
  BaseApp,  // Probably not
  createConsoleLogger,
  deserializeManifest,
  deserializeRouteData,
  deserializeRouteInfo,
  fromRoutingStrategy,
  serializeRouteData,
  serializeRouteInfo,
  toRoutingStrategy,
  type RenderErrorOptions,
  type RenderOptions,
  type RoutesList,
} from "astro/app";
```

I haven't pushed those changes yet because there is more work to do (linking, check/improve the descriptions) and I'm not sure we want this. But, if you're not against that idea:
* if you think a preview could be useful to see how it looks before spending more time on it, I can create a draft PR tomorrow even if some links are broken.
* I'll polish what I have before creating a PR (targeting Florian's branch? v6?)

florian-lefebvre:
I think having a dedicated page would make sense yeah!

> In the Adapter API page I introduced a "Building adapter entrypoints" section in anticipation of the "Building a preview entrypoint" section. And "Passing build-time configuration" is probably for all entrypoints so we don't need to nest it under "Building a server entrypoint".

No this is specific to the server entrypoint, the preview entrypoint cannot use the same mechanism

> I also noticed there are a lot of available imports from astro/app we don't document.

I don't know either! I have the feeling some of these are meant to be used internally so I would only document `App` for now

> I'll polish what I have before creating a PR (targeting Florian's branch? v6?)

Feel free to do it in this PR :+1:

sarah11918:
I guess a separate page would help future proof us for Algolia complaining the page is too long again 😅 

My *only* concern is that `astro/app` is only going to be used by integration/adapter authors, right? But for most people coming from e.g. React or a React-based framework, they are used to thinking of `App` as their "project entry point". So, people might think this is part of building an Astro site/"app" itself and we'd want to make it super clear that this is for integration authors.

Note that I don't think making it more visible it is necessarily a bad thing, though! I see Florian suggesting in support all the time, "Oh, you need this functionality, you can write an integration for that! So, integrations are a natural extension of Astro behaviour, and maybe one we *should* be encouraging! Again, my only concern would be people inadvertantly getting in there, getting confused etc. when they're just making a basic site, so clarity is key.

But if that entire section of the Adapter API page is/would be simply a string of API references *anyway*, then moving it off to the modules section does make sense, I think!

ArmandPhilippot:
Yeah, that was also my concern re: "think about a better name than "App API Reference" for the astro/app page".

But having one module documented here while all the others (including `astro:static-paths` which is for integrations/adapter too) have their own page is a bit odd. And, yes, I think this helps both being future-proof with Algolia and dealing with nested sections.

> No this is specific to the server entrypoint, the preview entrypoint cannot use the same mechanism

Oh, thanks for pushing back! I assumed they worked in the same way.

Alright, I'll polish what I have and I'll push the changes here. If we don't like them, we can always revert the commit!

ArmandPhilippot:
> So tiny! 😅

Well, not so tiny anymore. 😆 

There is probably some wording to polish but I think all the changes are here, including broken links fixes because of the reorganization!

To summarize what I've done:
* I moved all the `astro/app` docs to a new page in "Runtime APIs"
* I took the third path, ie. "we document App under createApp()" because I don't see how we can get `manifest` (to satisfy the `App` constructor) and we document what is public/what is useful for adapters authors, not what is helpful for core maintainers...
* And because of the point above, I removed all the static methods (anyway, two of them are already removed in core)... maybe we want to add something in the upgrade guide just in case?
* While I was there (ie. in the new `astro/app` page):
  * I updated the way we were introducing the module to call out this is for adapters authors
  * I added more links
  * I updated the `RenderOptions.addCookieHeader` description to remove "When this is false..." because it seems this edge case is only relevant for core maintainers now
  * I updated the `createApp()` description
  * I updated the last two code snippets to include `createApp()` (my thinking is, if people are interest in Node specifically, this might be useful to have a picture of how things unfold)
* I updated the "Building a server entrypoint" description to add links to the new page

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+  maxHeadingLevel: 3
+---
+import ReadMore from '~/components/ReadMore.astro';
+import Since from '~/components/Since.astro';
+
+This module helps you [build a server entrypoint for your adapter](/en/reference/adapter-reference/#building-a-server-entrypoint) while supporting pages rendered in development mode or that have been prebuilt through `astro build`.
+
+Astro uses the standard [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) and [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) objects. Hosts using a different API for requests/responses should convert to these types in their adapter. For example, Astro exposes [helpers to work with NodeJS](#imports-from-astroappnode).
+
+## `astro/app` types
```
florian-lefebvre: Not sure if types should come first, or if the "runtime" should instead

florian-lefebvre:
Looks pretty good!

src/content/docs/en/reference/modules/astro-app.mdx
ArmandPhilippot: Yeah, I get what you mean. I hesitated. I left it there because I thought it might be confusing for the user to go into `astro/app` and not see any "direct" import option from that module (if that makes sense 😅 ).
But looking back, I think you you're right and we can move them at the bottom!

sarah11918:
I just want to note, for the record, this was Florian's original PR:

image

Team Docs, FTW! 💪

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+  label: 'astro/app'
+i18nReady: true
+tableOfContents:
+  minHeadingLevel: 2
+  maxHeadingLevel: 3
+---
+import ReadMore from '~/components/ReadMore.astro';
+import Since from '~/components/Since.astro';
+
+This module helps you [build a server entrypoint for your adapter](/en/reference/adapter-reference/#building-a-server-entrypoint) while supporting pages rendered in development mode or that have been prebuilt through `astro build`.
```
sarah11918:
```suggestion
This module helps adapter authors [build a server entrypoint](/en/reference/adapter-reference/#building-a-server-entrypoint) while supporting pages rendered in development mode or that have been prebuilt through `astro build`. 

`astro/app` is used internally for [Astro's official server adapters](/en/guides/on-demand-rendering/#server-adapters), and is also publicly available for you to build a custom adapter for your specific runtime or deploy host.

:::tip[Publishing your adapter]
Use the appropriate keywords when you [publish your adapter to NPM](/en/reference/publish-to-npm/#categories) so that your adapter will be listed in our [Integrations Directory](https://astro.build/integrations/).
:::
```
I like replacing "you" with "adapter authors" so that in this case, it's *less* easy for regular Astro users to identify with this statement (and know they can ignore this). Tried to also emphasize "this is for building adapters" in some other ways.... Something like this maybe?

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+### `createApp()`
+
+<p>
+
+**Type:** <code>(options?: \{ streaming: boolean \}) => <a href="#the-app-instance">App</a></code><br />
+<Since v="6.0.0" />
+</p>
+
+Returns an [`App` instance](#the-app-instance) offering methods to work with standard Request and Response objects when [building an adapter's server entrypoint](/en/reference/adapter-reference/#building-a-server-entrypoint).
```
sarah11918:
```suggestion
Returns an [`App` instance](#the-app-instance) that includes methods to work with standard Request and Response objects when [building an adapter's server entrypoint](/en/reference/adapter-reference/#building-a-server-entrypoint).
```
"offering" is not quite right I don't think? I'm not sure what it should be, but flagging that word!

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+```
+
+##### `app.match()`
+
+<p>
+
+**Type:** <code>(request: Request, allowPrerenderedRoutes = false) => <a href="/en/reference/integrations-reference/#routedata">RouteData</a> | undefined</code>
+</p>
+
+Determines if a request is matched by the Astro app's routing rules.
```
sarah11918:
```suggestion
Determines whether a request is matched by the Astro app's routing rules.
```
nit: "If" is a hypothetical, or cause and effect ("If this happens, then...") "Whether or not" is preferred for describing boolean situations!

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+##### `app.getAllowedDomains()`
+
+<p>
+
+**Type:** <code>() => Partial\<<a href="/en/reference/modules/astro-assets/#remotepattern">RemotePattern</a>\>[] | undefined</code><br />
+<Since v="5.14.2" />
+</p>
+
+Returns a list of permitted host patterns for incoming requests when using on-demand rendering [as defined in the user configuration](/en/reference/configuration-reference/#securityalloweddomains).
```
sarah11918:
```suggestion
Returns a list of permitted host patterns for incoming requests when using on-demand rendering [as configured in `security.allowedDomains`](/en/reference/configuration-reference/#securityalloweddomains).
```
The link text on its own was a little vague about saying exactly where you were being taken to. And, this means that someone doesn't *have* to click the link if they already know/understand `security.allowedDomains` and don't need its full reference!

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+The following helpers are imported from the `node` directory in the app module:
+
+```js
+import {
+  createRequest,
+  writeResponse
+} from "astro/app/node";
+```
+
+This module is used in conjunction with [the methods provided by `createApp()`](#createapp) to convert a [NodeJS `IncomingMessage`](https://nodejs.org/api/http.html#class-httpincomingmessage) into a web-standard `Request` and stream a web-standard `Response` into a [NodeJS `ServerResponse`](https://nodejs.org/api/http.html#class-httpserverresponse).
```
sarah11918, line 184: Really nice work here! Love this intro to the module!

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+<p>
+
+**Type:** `boolean`<br />
+**Default:** `true`
+</p>
+
+Defines whether HTML streaming is enabled. In most cases, disabling streaming is not recommended as it improves performance and generally provides a better visitor experience.
+
+HTML streaming breaks a document into chunks to send over the network and render on the page in order. This normally results in visitors seeing your HTML as fast as possible but factors such as network conditions and waiting for data fetches can block page rendering.
```
yanthomasdev:
```suggestion
HTML streaming breaks a document into chunks to send over the network and render on the page in order. This normally results in visitors seeing your HTML as fast as possible, but factors such as network conditions and waiting for data fetches can block page rendering.
```

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+#### `RenderOptions.locals`
+
+<p>
+
+**Type:** `object`
+</p>
+
+The [`context.locals` object](/en/reference/api-reference/#locals) used to store and access information during the lifecycle of a request.
+
+The example below reads a header named `x-private-header`, attempts to parse it as an object and pass it to `locals`, which can then be passed to any [middleware function](/en/guides/middleware/).
```
yanthomasdev:
```suggestion
The example below reads a header named `x-private-header`, attempts to parse it as an object, and passes it to `locals`, which can then be passed to any [middleware function](/en/guides/middleware/).
```

src/content/docs/en/reference/modules/astro-app.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+### `createRequest()`
+
+<p>
+
+**Type:** <code>(req: NodeRequest, options?: \{ skipBody?: boolean; allowedDomains?: Partial\<<a href="/en/reference/modules/astro-assets/#remotepattern">RemotePattern</a>\>[]; \}) => Request</code><br />
+<Since v="6.0.0" />
+</p>
+
+Converts a NodeJS `IncomingMessage` into a standard `Request` object. This function accepts an optional object as the second argument, allowing you to define if the body should be ignored, defaulting to `false`, and the [`allowedDomains`](/en/reference/configuration-reference/#securityalloweddomains).
```
sarah11918: The "this function" sentence is too long here and a little difficult to parse. "ignoring the body, defaulting to false," and `allowedDomains` feel too crammed in for explaining this object... maybe because you have TWO ideas but the "defaulting to false" clause is on "equal footing" with them, making this look like more than it means to be. Can someone take another shot at this?

src/content/docs/en/reference/modules/astro-app.mdx
ArmandPhilippot: Ah, thanks for catching this! Yeah, this was one of the "wording that might need polishing". I wasn't satisfied with "offering" but couldn't think of something else yesterday. Your suggestion is much better!

src/content/docs/en/reference/modules/astro-app.mdx
ArmandPhilippot: Maybe something like:
> Converts a NodeJS `IncomingMessage` into a standard `Request` object. An optional object can be passed as a second argument to further control how the request is created. This is useful if you want to ignore the body (defaults to `false`) or pass the configured [`allowedDomains`](/en/reference/configuration-reference/#securityalloweddomains) to the request.

sarah11918: Much nicer! Let's do it!

ArmandPhilippot:
I love seeing how moving content helps to refine what we had! Great teamwork!

If everyone is happy, I think this one is ready to be merged! 🎉

sarah11918 (approved):
The world's best docs team! 🙌

sarah11918:
Just noting that we have the one `ru` page being updated here, so we'll need a Lunaria statement for that!

ArmandPhilippot:
This is for v6 though, so we'll need to deal with Lunaria on the v6 PR not here, right?

sarah11918:
Oh, forgot! Yes, should be fine! Thank you!

florian-lefebvre:
Alright then can this PR be merged? The associated code PR was released last week already so we're not blocked by that

ArmandPhilippot:
Yeah, Sarah approved it! I know I hijacked your PR but this is still your PR so I was waiting for you to merge it just in case you had second thoughts. 😄

PR #13277, opened 2026-02-17
feat: update `edgeMiddleware` references to `middlewareMode`

leekeh:
#### Description (required)

Update documentation to use `middlewareMode` instead of `edgeMiddleware`. Context is in linked PR.

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -185,7 +185,7 @@ export const collections = { blog };
 
 #### Defining custom IDs
 
-When using the [`glob()` loader](#the-glob-loader) with Markdown, MDX, Markdoc, JSON, or TOML files, every content entry [`id`](/en/reference/modules/astro-content/#id) is automatically generated in an URL-friendly format based on the content filename. This unique `id` is used to query the entry directly from your collection. It is also useful when [creating new pages and URLs from your content](#generating-routes-from-content).
+When using the [`glob()` loader](#the-glob-loader) with Markdown, MDX, Markdoc, JSON, or TOML files, every content entry [`id`](/en/reference/modules/astro-content/#collectionentryid) is automatically generated in an URL-friendly format based on the content filename. This unique `id` is used to query the entry directly from your collection. It is also useful when [creating new pages and URLs from your content](#generating-routes-from-content).
 
 You can override a single entry’s generated `id` by adding your own `slug` property to the file frontmatter or data object for JSON files. This is similar to the “permalink” feature of other web frameworks.
 
--- a/src/content/docs/en/reference/content-loader-reference.mdx
+++ b/src/content/docs/en/reference/content-loader-reference.mdx
@@ -117,7 +117,7 @@ By default it uses [`github-slugger`](https://github.com/Flet/github-slugger) to
 
 Whether or not to store the raw body of content files in the data store.
 
-When `retainBody` is `false`, [`entry.body`](/en/reference/modules/astro-content/#body) will be `undefined` instead of containing the raw file contents.
+When `retainBody` is `false`, [`entry.body`](/en/reference/modules/astro-content/#collectionentrybody) will be `undefined` instead of containing the raw file contents.
 
 Setting this property to `false` significantly reduces the deployed size of the data store and helps avoid hitting size limits for sites with very large collections.
 
--- a/src/content/docs/en/reference/modules/astro-actions.mdx
+++ b/src/content/docs/en/reference/modules/astro-actions.mdx
@@ -5,7 +5,7 @@ sidebar:
 i18nReady: true
 tableOfContents:
   minHeadingLevel: 2
-  maxHeadingLevel: 6
+  maxHeadingLevel: 3
 ---
 import Since from '~/components/Since.astro';
 import ReadMore from '~/components/ReadMore.astro';
@@ -277,7 +277,7 @@ export const onRequest = defineMiddleware(async (context, next) => {
 });
 ```
 
-##### `calledFrom`
+##### `action.calledFrom`
 
 <p>
 
@@ -286,7 +286,7 @@ export const onRequest = defineMiddleware(async (context, next) => {
 
 Whether an action was called using an RPC function or an HTML form action.
 
-##### `name`
+##### `action.name`
 
 <p>
 
@@ -295,7 +295,7 @@ Whether an action was called using an RPC function or an HTML form action.
 
 The name of the action. Useful to track the source of an action result during a redirect.
 
-##### `handler()`
+##### `action.handler()`
 
 <p>
 
@@ -488,7 +488,7 @@ button?.addEventListener('click', async () => {
 
 Alternatively, you can use it as an object giving you access to the `queryString` and an alternative `orThrow()` method.
 
-#### `queryString` property
+#### `ActionClient.queryString`
 
 <p>
 
@@ -510,7 +510,7 @@ feedbackUrl.search = actions.myAction.queryString;
 <FeedbackForm sendTo={feedbackUrl.pathname} />
 ```
 
-#### `orThrow()` property
+#### `ActionClient.orThrow()`
 
 <p>
 
--- a/src/content/docs/en/reference/modules/astro-assets.mdx
+++ b/src/content/docs/en/reference/modules/astro-assets.mdx
@@ -1041,7 +1041,7 @@ isRemoteImage(localImage); // Output: `false`
 
 <p>
 
-**Type:** <code>(src: <a href="#src-2">UnresolvedImageTransform['src']</a>) => Promise\<string | <a href="#imagemetadata-1">ImageMetadata</a>\></code><br />
+**Type:** <code>(src: <a href="#unresolvedimagetransformsrc">UnresolvedImageTransform['src']</a>) => Promise\<string | <a href="#imagemetadata-1">ImageMetadata</a>\></code><br />
 <Since v="4.0.0" />
 </p>
 
@@ -1308,7 +1308,7 @@ import type {
 
 Describes the result of the transformation after the call to [`getImage()`](/en/reference/modules/astro-assets/#getimage).
 
-#### `attributes`
+#### `GetImageResult.attributes`
 
 <p>
 
@@ -1317,7 +1317,7 @@ Describes the result of the transformation after the call to [`getImage()`](/en/
 
 Defines the additional HTML attributes needed to render the image (e.g. width, height, style).
 
-#### `options`
+#### `GetImageResult.options`
 
 <p>
 
@@ -1326,7 +1326,7 @@ Defines the additional HTML attributes needed to render the image (e.g. width, h
 
 Describes the transformation settings after validation.
 
-#### `rawOptions`
+#### `GetImageResult.rawOptions`
 
 <p>
 
@@ -1335,7 +1335,7 @@ Describes the transformation settings after validation.
 
 Describes the original transformation settings.
 
-#### `src`
+#### `GetImageResult.src`
 
 <p>
 
@@ -1344,7 +1344,7 @@ Describes the original transformation settings.
 
 The path to the generated image.
 
-#### `srcSet`
+#### `GetImageResult.srcSet`
 
 <p>
 
@@ -1354,7 +1354,7 @@ The path to the generated image.
 
 An object describing how to render the [`srcset` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#srcset).
 
-##### `values`
+##### `GetImageResult.srcSet.values`
 
 <p>
 
@@ -1363,7 +1363,7 @@ An object describing how to render the [`srcset` attribute](https://developer.mo
 
 An array of generated values where each entry includes a URL and a size descriptor. This can be used to manually generate the value of the `srcset` attribute.
 
-##### `attribute`
+##### `GetImageResult.srcSet.attribute`
 
 <p>
 
@@ -1381,7 +1381,7 @@ A value ready to use in the `srcset` attribute.
 
 Defines the options accepted by the image transformation service. This contains a required `src` property, optional predefined properties, and any additional properties required by the image service:
 
-#### `src`
+#### `ImageTransform.src`
 
 <p>
 
@@ -1390,7 +1390,7 @@ Defines the options accepted by the image transformation service. This contains 
 
 Defines the path to a local image in the `public` directory, the URL of a remote image, or the data from an imported image.
 
-#### `width`
+#### `ImageTransform.width`
 
 <p>
 
@@ -1399,7 +1399,7 @@ Defines the path to a local image in the `public` directory, the URL of a remote
 
 The width of the image.
 
-#### `height`
+#### `ImageTransform.height`
 
 <p>
 
@@ -1408,7 +1408,7 @@ The width of the image.
 
 The height of the image.
 
-#### `widths`
+#### `ImageTransform.widths`
 
 <p>
 
@@ -1418,7 +1418,7 @@ The height of the image.
 
 A list of widths to generate for the image.
 
-#### `densities`
+#### `ImageTransform.densities`
 
 <p>
 
@@ -1428,7 +1428,7 @@ A list of widths to generate for the image.
 
 A list of pixel densities to generate for the image.
 
-#### `quality`
+#### `ImageTransform.quality`
 
 <p>
 
@@ -1437,7 +1437,7 @@ A list of pixel densities to generate for the image.
 
 The desired quality for the output image.
 
-#### `format`
+#### `ImageTransform.format`
 
 <p>
 
@@ -1446,7 +1446,7 @@ The desired quality for the output image.
 
 The desired format for the output image.
 
-#### `fit`
+#### `ImageTransform.fit`
 
 <p>
 
@@ -1456,7 +1456,7 @@ The desired format for the output image.
 
 Defines a list of allowed values for the `object-fit` CSS property, extensible with any string.
 
-#### `position`
+#### `ImageTransform.position`
 
 <p>
 
@@ -1475,7 +1475,7 @@ Controls the value for the `object-position` CSS property.
 
 Represents an image with transformation options. This contains the same properties as the [`ImageTransform` type](#imagetransform) with a different `src` type and an additional `inferSize` property.
 
-#### `src`
+#### `UnresolvedImageTransform.src`
 
 <p>
 
@@ -1484,7 +1484,7 @@ Represents an image with transformation options. This contains the same properti
 
 The path to an image imported or located in the `public` directory, or the URL of a remote image.
 
-#### `inferSize`
+#### `UnresolvedImageTransform.inferSize`
 
 <p>
 
@@ -1505,7 +1505,7 @@ Determines whether the width and height of the image should be inferred.
 
 Describes the data collected during image import. This contains the following properties:
 
-#### `src`
+#### `ImageMetadata.src`
 
 <p>
 
@@ -1514,7 +1514,7 @@ Describes the data collected during image import. This contains the following pr
 
 The absolute path of the image on the filesystem.
 
-#### `width`
+#### `ImageMetadata.width`
 
 <p>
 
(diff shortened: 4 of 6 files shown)


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1287,6 +1293,16 @@ export default function createIntegration() {
 }
 ```
 
+### `MiddlewareMode`
```
sarah11918, line 913: @ArmandPhilippot I'll ask for your guidance here. Somehow it feels *less* helpful to have such a short entry here that actually tells people which values they can set as a middleware mode? I would *think* that that it's more valuable that in the entry above for the config option, people can very quickly see "oh, it's classic or edge!" rather than see "Oh, middleware mode is a middleware mode" then need to follow a trail to get here?

But, what do you think?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1287,6 +1293,16 @@ export default function createIntegration() {
...
+
+<p>
+
+**Type:** `"classic" | "edge"`<br />
+<Since v="6.0.0" />
```
sarah11918, line 918: And wherever we do document these values explicitly, is there a default?

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,29 +1041,35 @@ Defines whether the adapter supports image transformation using the built-in Sha
...
 
 <p>
 
-**Type:** `boolean`
+**Type:** [`MiddlewareMode`](#middlewaremode)<br />
+<Since v="6.0.0" />
 </p>
 
-Defines whether any on-demand rendering middleware code will be bundled when built.
+Defines in which way middleware is run.
```
sarah11918: This is a little vague, and maybe leads to explaining more about what's going on in each option below than might be necessary?

If there does happen to be a defining axis/split on which these two options are clearly different, it would be nice to get that upfront HERE.

(e.g., -- and this is just words as examples for adding more detail, not meant to be accurate -- "Defines when and how middleware is run" at least gives a hint that middleware might run at different times, so people are primed for thinking about choosing options below based on WHEN they'd like their middleware to run. That's still pretty general, so if there is any more precise difference, I'd gladly take something even more informative here!)

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,29 +1041,35 @@ Defines whether the adapter supports image transformation using the built-in Sha
...
+
+```js title="my-adapter.mjs" ins={9-11}
 export default function createIntegration() {
   return {
     name: '@example/my-adapter',
     hooks: {
       'astro:config:done': ({ setAdapter }) => {
         setAdapter({
           name: '@example/my-adapter',
-          entrypointResolution: 'auto',
```
ArmandPhilippot, line 682: If think Florian added that on purpose in #13253 because we want people to use the new API. So we should probably revert the deleted line.
(this means the hightlighted lines should also be updated)

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1074,18 +1080,17 @@ export default function createIntegration() {
...
-```js title="my-adapter.mjs" ins={16-21}
+```js title="my-adapter.mjs" ins={15-20}
 export default function createIntegration() {
   return {
     name: '@example/my-adapter',
     hooks: {
       'astro:config:done': ({ setAdapter }) => {
         setAdapter({
           name: '@example/my-adapter',
-          entrypointResolution: 'auto',
```
ArmandPhilippot, line 704: Same here.

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,29 +1041,35 @@ Defines whether the adapter supports image transformation using the built-in Sha
...
-**Type:** `boolean`
+**Type:** [`MiddlewareMode`](#middlewaremode)<br />
+<Since v="6.0.0" />
 </p>
 
-Defines whether any on-demand rendering middleware code will be bundled when built.
+Defines in which way middleware is run.
 
-When enabled, this prevents middleware code from being bundled and imported by all pages during the build:
+#### `classic`
```
ArmandPhilippot: Not everything is standardized (yet) but usually we don't create headings for each member of an union.

Maybe the description of both modes should be in the `middlewareMode` description, as part of Sarah's comment re: explaining what it does. And we can introduce the code snippets, with:
> The following example uses the `edge` mode...

See https://v6.docs.astro.build/en/reference/adapter-reference/#envgetsecret for example (not an union, but for how to introduce the code snippet).

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,29 +1041,35 @@ Defines whether the adapter supports image transformation using the built-in Sha
 
 A set of features that changes the output of the emitted files. When an adapter opts in to these features, they will get additional information inside specific hooks and must implement the proper logic to handle the different output.
 
-### `edgeMiddleware`
+### `middlewareMode`
 
 <p>
 
-**Type:** `boolean`
+**Type:** [`MiddlewareMode`](#middlewaremode)<br />
```
ArmandPhilippot: As Sarah said, we'll need a default here:
```suggestion
**Type:** [`MiddlewareMode`](#middlewaremode)<br />
**Default:** `"classic"`<br />
```

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: My take would be: Is this type useful so that the user can use it elsewhere (for example, a helper function) or is it only useful in this specific case? I would say the second one here, but I'm not an adapter builder so I could be wrong.

If there are use cases for the type, while I agree with Sarah, I think this can helpful to have an entry for it (so people know they can import it). Otherwise, yes, we probably don't need this and we can replace `MiddlewareMode` with `"classic" | "edge"` in the property type!

src/content/docs/en/reference/adapter-reference.mdx
ArmandPhilippot: Well, if we keep the type as an independent section, a type does not have a default. But, yes, `middlewareMode` (the property) should!

sarah11918: Right, that's what I meant, for the property yes! Thanks!

src/content/docs/en/reference/adapter-reference.mdx
leekeh: sorry, I didn't mean to, it was a copy-paste issue since i started working on the wrong branch 😓  I'll fix all of that.

src/content/docs/en/reference/adapter-reference.mdx
leekeh: Use case would mainly be for adapter developers, to expose this as a property on the adapter. I'd recommend using the type over hardcoding it, since it is going to change in the (near) future.

src/content/docs/en/reference/adapter-reference.mdx
leekeh: Fair point, I improved it with this:
`Determines at which stage of the page lifecycle the middleware is executed, and how the middleware code is emitted in the build output.
`
Let me know if it suffices.

src/content/docs/en/reference/adapter-reference.mdx
leekeh: I thought it would be helpful to separate them, but yes I was not sure if it was right to use headers 🤔 I can just use paragraphs, but I thought it might be clearer this way. But we can always try it out 👍

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,29 +1041,33 @@ Defines whether the adapter supports image transformation using the built-in Sha
...
```
ArmandPhilippot: Currently this is a link to the current section instead of the type:
```suggestion
**Type:** [`MiddlewareMode`](#middlewaremode-1)<br />
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,29 +1041,33 @@ Defines whether the adapter supports image transformation using the built-in Sha
...
+The `classic` mode matches the default behavior of Astro. On prerendered pages, middleware is run at build time, and when the page is requested, the middleware is not run again. On dynamic pages, middleware is run at request time only. The middleware code is part of your server bundle.
 
-```js title="my-adapter.mjs" ins={10-12}
+In `edge` mode, the middleware code can be deployed independently from the server bundle, for example, as an edge function.
+
+```js title="my-adapter.mjs" ins={9-11}
 export default function createIntegration() {
   return {
     name: '@example/my-adapter',
+    entrypointResolution: 'auto',
```
ArmandPhilippot: This is in the wrong place:
```suggestion
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,29 +1041,33 @@ Defines whether the adapter supports image transformation using the built-in Sha
...
     hooks: {
       'astro:config:done': ({ setAdapter }) => {
         setAdapter({
           name: '@example/my-adapter',
```
ArmandPhilippot, line 685: This is this where `entrypointResolution` should be located:
```suggestion
          name: '@example/my-adapter',
          entrypointResolution: 'auto',
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,29 +1041,33 @@ Defines whether the adapter supports image transformation using the built-in Sha
...
-Defines whether any on-demand rendering middleware code will be bundled when built.
+Determines at which stage of the page lifecycle the middleware is executed, and how the middleware code is emitted in the build output.
 
-When enabled, this prevents middleware code from being bundled and imported by all pages during the build:
+The `classic` mode matches the default behavior of Astro. On prerendered pages, middleware is run at build time, and when the page is requested, the middleware is not run again. On dynamic pages, middleware is run at request time only. The middleware code is part of your server bundle.
 
-```js title="my-adapter.mjs" ins={10-12}
+In `edge` mode, the middleware code can be deployed independently from the server bundle, for example, as an edge function.
+
+```js title="my-adapter.mjs" ins={9-11}
```
ArmandPhilippot: Because we now have `entrypointResolution`, this should be:

````suggestion
```js title="my-adapter.mjs" ins={10-12}
````

File: src/content/docs/en/reference/adapter-reference.mdx [hunk 12]
```diff
@@ -1074,7 +1078,7 @@ export default function createIntegration() {
 
 Then, consume the hook [`astro:build:ssr`](/en/reference/integrations-reference/#astrobuildssr), which will give you a `middlewareEntryPoint`, an `URL` to the physical file on the file system.
 
-```js title="my-adapter.mjs" ins={16-21}
+```js title="my-adapter.mjs" ins={15-20}
```
ArmandPhilippot: Because we now have `entrypointResolution` this should be:

````suggestion
```js title="my-adapter.mjs" ins={16-21}
````

File: src/content/docs/en/guides/integrations-guide/vercel.mdx [hunk 13]
```diff
@@ -442,20 +442,20 @@ export default defineConfig({
 
 ### Running Astro middleware on Vercel Edge Functions
 
-The `@astrojs/vercel` adapter can create an [edge function](https://vercel.com/docs/functions/edge-functions) from an Astro middleware in your code base. When `edgeMiddleware` is enabled, an edge function will execute your middleware code for all requests including static assets, prerendered pages, and on-demand rendered pages.
+The `@astrojs/vercel` adapter can create an [edge function](https://vercel.com/docs/functions/edge-functions) from an Astro middleware in your code base. When `middlewareMode` is set to `'edge'`, an edge function will execute your middleware code for all requests including static assets, prerendered pages, and on-demand rendered pages.
```
ArmandPhilippot: We probably want a link here, as in the Netlify guide:
```suggestion
The `@astrojs/vercel` adapter can create an [edge function](https://vercel.com/docs/functions/edge-functions) from an Astro middleware in your code base. When [`middlewareMode`](/en/reference/adapter-reference/#middlewaremode) is set to `'edge'`, an edge function will execute your middleware code for all requests including static assets, prerendered pages, and on-demand rendered pages.
```

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1105,6 +1109,7 @@ function createEdgeMiddleware(middlewareEntryPoint) {
 }
 ```
 
+
```
leekeh:
```suggestion
```

src/content/docs/en/reference/adapter-reference.mdx
sarah11918: Thanks, I think that's so much more helpful to the reader! I appreciate you digging a little deeper.

sarah11918 (approved):
Checked the deploy preview and that looks good to me! Thanks for your help writing great docs, @leekeh ! 💪

ArmandPhilippot:
Ah, it seems we have a broken link because of a translation. Adapter reference is not yet translated in Japanese, so `/ja/guides/integrations-guide/netlify` is using the English anchor.

If you're able to update:
https://github.com/withastro/docs/blob/3c074907afbb38b42531c136fd7909fabbc67ee6/src/content/docs/ja/guides/integrations-guide/netlify.mdx?plain=1#L103
with the following replacement:
```diff
-[`edgeMiddleware`オプション](/ja/reference/adapter-reference/#edgemiddleware)
+[`middlewareMode`オプション](/ja/reference/adapter-reference/#middlewareMode)
```
The tests should pass!

But, if you prefer we can also push a fix to your branch!

src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -1041,16 +1041,20 @@ Defines whether the adapter supports image transformation using the built-in Sha
 
 A set of features that changes the output of the emitted files. When an adapter opts in to these features, they will get additional information inside specific hooks and must implement the proper logic to handle the different output.
 
-### `edgeMiddleware`
+### `middlewareMode`
```
sarah11918, line 663: Noting that this is causing a link failure in `/ja/` where this is translated and the Netlify guide is linking here.

Would be great if @jp-knj or @morinokami could jump in here! We can translate *just* the heading, then use that as the anchor link on the Netlify page (but still mark this as [i18nIgnore] so that it's clear the pages do still need the regular updates.

One thing we *could* do that doesn't have to include any translations yet, since the Netlify guide will need to update *anyway* is that this PR could temporarily just update the `ja` adapter page to say (in English) "middlewareMode", and then we hard code the English slug in the broken link. When the translators go to update the Netlify page, they can *also* change the English heading and slug to translations. (note this would still require [i18nIgnore] in the subject because we would have touched those `/ja/` files)

sarah11918:
Armand and I have to stop meeting like this! 😄

src/content/docs/en/guides/integrations-guide/netlify.mdx
```diff
@@ -98,9 +98,9 @@ The [Netlify Blog post on Astro](https://www.netlify.com/blog/how-to-deploy-astr
 
 ### Running Astro middleware on Netlify Edge Functions
 
-Any Astro middleware is applied to pre-rendered pages at build-time, and to on-demand-rendered pages at runtime.
+By default, Astro middleware is applied to pre-rendered pages at build-time, and to on-demand-rendered pages at runtime.
```
yanthomasdev: comma nit
```suggestion
By default, Astro middleware is applied to pre-rendered pages at build-time and to on-demand-rendered pages at runtime.
```

src/content/docs/en/guides/integrations-guide/netlify.mdx
```diff
@@ -109,12 +109,12 @@ import netlify from '@astrojs/netlify';
...
   // ...
   adapter: netlify({
-    edgeMiddleware: true,
+    middlewareMode: 'edge',
   }),
 });
 ```
 
-When `edgeMiddleware` is enabled, an edge function will execute your middleware code for all requests including static assets, prerendered pages, and on-demand rendered pages.
+When `middlewareMode` is set to `'edge'`, an edge function will execute your middleware code for all requests including static assets, prerendered pages, and on-demand rendered pages.
```
yanthomasdev:
```suggestion
When `middlewareMode` is set to `'edge'`, an edge function will execute your middleware code for all requests, including static assets, prerendered pages, and on-demand rendered pages.
```

yanthomasdev (commented):
Thanks @leekeh, I have a few nits 🙌

sarah11918:
Hi @leekeh ! Just a reminder that we can't merge this PR with a broken link. I think Armand's solution is simpler than mine, so please just update the link anchor as described in his earlier comment here: https://github.com/withastro/docs/pull/13277#issuecomment-3927756964

leekeh:
> Thank you for fixing that link! 🙌🏽 LGTM! Now, we just need to wait for the implementation to be released before merging the docs!

Thanks 😄  Thanks also for your patience with me, I will try to stick to the rules better in my next contribution 🙂 This was my first one, it's really inspiring to see the level of quality you guys push and the clear guidelines that exist.

src/content/docs/ja/guides/integrations-guide/netlify.mdx
```diff
@@ -100,7 +100,7 @@ netlify deploy
 
 Astroミドルウェアは、ビルド時に事前レンダリングされたページに、実行時にオンデマンドでレンダリングされたページに適用されます。
 
-事前レンダリングされたページのリダイレクト、アクセス制御、またはカスタムレスポンスヘッダーを実装するには、[`edgeMiddleware`オプション](/ja/reference/adapter-reference/#edgemiddleware)を有効にして、Netlify Edge Functionsでミドルウェアを実行します。
+事前レンダリングされたページのリダイレクト、アクセス制御、またはカスタムレスポンスヘッダーを実装するには、[`middlewareMode`オプション](/ja/reference/adapter-reference/#middlewareMode)を有効にして、Netlify Edge Functionsでミドルウェアを実行します。
```
sarah11918:
```suggestion
事前レンダリングされたページのリダイレクト、アクセス制御、またはカスタムレスポンスヘッダーを実装するには、[`middlewareMode`オプション](/ja/reference/adapter-reference/#middlewaremode)を有効にして、Netlify Edge Functionsでミドルウェアを実行します。
```
Took me longer than I want to admit to figure out why middlewaremode didn't exist when it's CLEARLY there, but URLs never have upper case letters in them. 😅

sarah11918:
(Also updated branch to include the most recent Lunaria fixes)

sarah11918:
Thank you so much for your patience with us, @leekeh ! I *think* updating the URL and the branch should make all the tests pass here.

In any case, it's now OUR responsibility to get this into a merge-able shape and out whenever the feature is released. You've done your job, and we appreciate it! 🫡 

Welcome to Team Docs! 🥳

src/content/docs/en/guides/integrations-guide/vercel.mdx
```diff
@@ -442,20 +442,20 @@ export default defineConfig({
 
 ### Running Astro middleware on Vercel Edge Functions
 
-The `@astrojs/vercel` adapter can create an [edge function](https://vercel.com/docs/functions/edge-functions) from an Astro middleware in your code base. When `edgeMiddleware` is enabled, an edge function will execute your middleware code for all requests including static assets, prerendered pages, and on-demand rendered pages.
+The `@astrojs/vercel` adapter can create an [edge function](https://vercel.com/docs/functions/edge-functions) from an Astro middleware in your code base. When [`middlewareMode`](/en/reference/adapter-reference/#middlewaremode) is set to `'edge'`, an edge function will execute your middleware code for all requests including static assets, prerendered pages, and on-demand rendered pages.
```
yanthomasdev:
```suggestion
The `@astrojs/vercel` adapter can create an [edge function](https://vercel.com/docs/functions/edge-functions) from an Astro middleware in your code base. When [`middlewareMode`](/en/reference/adapter-reference/#middlewaremode) is set to `'edge'`, an edge function will execute your middleware code for all requests, including static assets, prerendered pages, and on-demand rendered pages.
```

PR #13278, opened 2026-02-17
feat(cloudflare): support vite plugins options

florian-lefebvre:

- Closes #
- Suggested label:


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -102,7 +102,16 @@ Now, you can enable [on-demand rendering per page](/en/guides/on-demand-renderin
 
 ## Options
 
-The Cloudflare adapter accepts the following options:
+The Cloudflare adapter accepts the following options from [`@cloudflare/vite-plugin`](https://developers.cloudflare.com/workers/vite-plugin/):
```
sarah11918, line 105: I'm going to guess that this link is the best one we have? On that page it says absolutely nothing about these options, but I also couldn't easily find anywhere these 6 options were all mentioned together. (The first 4 are mentioned on the specific API page - https://developers.cloudflare.com/workers/vite-plugin/reference/api/ but the last two are not)

florian-lefebvre: Yeah maybe we could link to https://developers.cloudflare.com/workers/vite-plugin/reference/api/#interface-pluginconfig but otherwise, I also failed to find docs for the rest

sarah11918: I'll let you make the call! The API reference page is maybe marginally more helpful in our immediate context of saying you can set these options? But the above link *is* the "entry page" to everything about the plugin itself, which is a perfectly reasonable thing to link to when mentioning it for the first time! If there *were* a more obvious link, we'd use it. Right now, I don't think there's a clear winner, so go with your gut!

florian-lefebvre:
@sarah11918 can I get your review on the associated changeset?

PR #13280, opened 2026-02-18
remove deprecated cloudflareModules

KTrain5169:
#### Description (required)

`cloudflareModules` was removed in v13 of the adapter but was left in the list of options on the Cloudflare adapter guide.
This PR removes it.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
+++ b/src/content/docs/en/guides/integrations-guide/cloudflare.mdx
@@ -104,6 +104,32 @@ Now, you can enable [on-demand rendering per page](/en/guides/on-demand-renderin
 
 The Cloudflare adapter accepts the following options:
 
+### `imageService`
+
+<p>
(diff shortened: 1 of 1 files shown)


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -104,43 +104,6 @@ Now, you can enable [on-demand rendering per page](/en/guides/on-demand-renderin
...
-import { defineConfig } from 'astro/config';
-import cloudflare from '@astrojs/cloudflare';
-
-export default defineConfig({
-  adapter: cloudflare({
-    imageService: 'cloudflare-binding',
-  }),
-});
-```
-
```
ArmandPhilippot, line 143: I think this removal is unwanted and only `cloudflareModules` should be removed? Can you revert this?

KTrain5169: oops, didn't catch that, sorry! will readd

PR #13282, opened 2026-02-18
Update inferSize fetching behavior

matthewp:
#### Description (required)

It's no longer true that `inferSize` will fetch unauthorized images. Allowing this exposes a vulnerability with internal services, so we now check `image.domains`.


sarah11918:
Already released with https://github.com/withastro/astro/pull/15569

PR #13293, opened 2026-02-22
clarify Astro's public API in semantic versioning docs

OliverSpeir:
Adds a sentence to the semantic versioning section clarifying that Astro's public API is defined by the features documented on the official Astro documentation.

>  For this system to work, you first need to declare a public API.

https://semver.org

### DIFF
--- a/src/content/docs/en/upgrade-astro.mdx
+++ b/src/content/docs/en/upgrade-astro.mdx
@@ -118,7 +118,7 @@ Documentation for older versions of Astro is not maintained, but is available as
 
 ## Semantic versioning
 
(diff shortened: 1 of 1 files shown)


sarah11918:
Hi Oliver! We're in a v5 docs freeze right now, so we're not updating English docs except emergency fixes like broken links. (I'm not sure if you filed this in such a way that you saw our PR template, but it clearly states there that we are not updating English docs right now.)

Also, I'm not sure this statement is necessary? Reading the original text, I don't believe the intention is that you need to publicly declare "this is a public API"? I think it means that their system *depends on there being* a public API in the first place, which as they say could be described through documentation (which we do).

So, I'm going to close this PR, but maybe you could open a discussion to see what others think about that statement and get some consensus? (Again, I don't read it as a formal declaration being needed at all, but obviously you did. So I guess we'll need more eyes!) Then, if needed, something can be done after we've released v6.

OliverSpeir:
😅 you miss every shot you don't take, sorry if it came across as inconsiderate

PR #13307, opened 2026-02-24
revert: change to shiki behaviour 

ematipico:
#### Description (required)

Reverts some changes on the upgrade guide. Plus, I added a limitation to the CSP. I know users asked for this, and we decided we want to support it, I added a paragraph about it


src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -516,7 +516,8 @@ Enabling this feature adds additional security to Astro's handling of processed
 This feature comes with some limitations:
 - External scripts and external styles are not supported out of the box, but you can [provide your own hashes](https://v6.docs.astro.build/en/reference/configuration-reference/#hashes).
 - [Astro's view transitions](https://v6.docs.astro.build/en/guides/view-transitions/) using the `<ClientRouter />` are not supported, but you can [consider migrating to the browser native View Transition API](https://events-3bg.pages.dev/jotter/astro-view-transitions/) instead if you are not using Astro's enhancements to the native View Transitions and Navigation APIs.
-- Shiki isn't currently supported. By design, Shiki functions using inline styles.
+- Shiki isn't currently supported. By design, Shiki functions using inline styles and it can't work with Astro infrastructure.
```
ArmandPhilippot: The changes on this file should be added to your core PR since this is auto-generated from https://github.com/withastro/astro/blob/9e47a0c6d3a30c6a827a4947e666ea49a389def3/packages/astro/src/types/public/config.ts#L664

I wonder if we should add a link to Prism as an alternative for people who want to use CSP but are blocked because of this. Something like:

```suggestion
- Shiki isn't currently supported. By design, Shiki functions use inline styles that cannot work with Astro CSP implementation. Consider [using `<Prism />`](https://v6.docs.astro.build/en/guides/syntax-highlighting/#prism-) when your project requires both CSP and syntax highlighting.
```

sarah11918: Agree with Armand's suggestion, AND yes, these changes need to be made in the core astro repo.

But also note that I have no objections to including config ref updates here so that they are in our config reference as the "desired state", and then the ci PR is our "test" that adding the JSDoc actually gave us what we wanted/expected. 😄

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -516,7 +516,8 @@ Enabling this feature adds additional security to Astro's handling of processed
...
+- `unsafe-inline` directives can't work with how Astro implements CSP. By default, Astro will emit hashes for all its bundled scripts (e.g. client islands), so `unsafe-inline` directives are automatically rejected by all modern browsers.
```
sarah11918: I'm just a little confused about the "so" (cause and effect) relationship stated here. "Astro emits hashes... SO browsers reject `unsafe-inline` directives"? Just checking whether this is in fact a *reason* for the rejection, or whether there's a clearer description we can make here.

Also adding a link we could use if we want to link externally to `unsafe-inline` for reference: https://content-security-policy.com/unsafe-inline/

sarah11918:
Noting that this appears to have already been released, so this is NOT `merge-on-release` and can merge whenever we are happy with it!

src/content/docs/en/reference/configuration-reference.mdx
ematipico: Apologies, I missed the comment at the top. I'll send a PR in core

src/content/docs/en/reference/configuration-reference.mdx
ematipico: I let you rephrase the paragraph then. If we have hashes or `nonce` to implement CSP, then the `unsafe-inline` won't work, as per browser specs. 

With Astro it's possible to add directives via configuration. If users add `unsafe-inline` in the configuration, it won't work

sarah11918: OK, I was just checking because I wasn't sure that it was the presence of hashes that mean `unsafe-inline` won't work, but it sounds like that *is* the case! Thanks!

sarah11918:
Config reference PR: https://github.com/withastro/astro/pull/15658/ for the updated JSDoc

PR #13315, opened 2026-02-26
Use a consistent headings format in `reference/modules`

ArmandPhilippot:
Nothing urgent, just some polishing before v6. Context:

In a [previous PR](https://github.com/withastro/docs/pull/13021#discussion_r2692505870), we decided we preferred `Type.prop` (or `object.prop`) as headings instead of `prop` both for readability and linking. We already use this format in some modules reference (e.g. [`envField`](https://v6.docs.astro.build/en/reference/modules/astro-config/#envfield)) but most of the modules were using the old format (e.g. [`action`](https://v6.docs.astro.build/en/reference/modules/astro-actions/#action)).

#### Description (required)

Updates all the `reference/modules` to use a consistent heading format (and fix some links because of the changes).

I'll probably do the same for the other references, but not in this PR to keep the scope small and easily reviewable (e.g. `api-reference.mdx` requires to fix some links in the translations too!).

### DIFF
--- a/src/content/docs/en/reference/modules/astro-actions.mdx
+++ b/src/content/docs/en/reference/modules/astro-actions.mdx
@@ -5,7 +5,7 @@ sidebar:
 i18nReady: true
 tableOfContents:
   minHeadingLevel: 2
-  maxHeadingLevel: 6
+  maxHeadingLevel: 3
 ---
 import Since from '~/components/Since.astro';
 import ReadMore from '~/components/ReadMore.astro';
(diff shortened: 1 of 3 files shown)


sarah11918:
This looks great! I'm just checking the deploy preview now because the only potential thing NOT to like about this is what it does to the TOC right sidebar:

- `astro:actions`: I suggest we take this back to a normal h2/h3 only
- `astro:assets`: LGTM! (and I think is already only h2/h3
- `astro:content`: I think I want to go back to h2/h3 here, too. It's not *terrible*, but I think these identifiers now mean we don't *have* to worry as much about exposing these
- `astro:middleware`: again, I think I want to go back to h2/h3 here too now!

Maybe this is helping us in general with our TOCs!

ArmandPhilippot:
Good call, much cleaner! And for imports/types reference this is probably enough levels in the TOC!

sarah11918:
Yeah, going to go out on a limb and suspect that ALL of these modules will look better with just h2/h3!

This one LGTM!

PR #13316, opened 2026-02-26
fix: consistent headings format in `api-reference.mdx`

ArmandPhilippot:
Same as #13315

For context: In a https://github.com/withastro/docs/pull/13021#discussion_r2692505870, we decided we preferred `Type.prop` (or `object.prop`) as headings instead of `prop` both for readability and linking. We already use this format in some modules reference (e.g. [envField](https://v6.docs.astro.build/en/reference/modules/astro-config/#envfield)) but most of the modules were using the old format (e.g. [action](https://v6.docs.astro.build/en/reference/modules/astro-actions/#action)).

#### Description (required)

Uses a consistent heading format in `api-reference.mdx`.

Note that we'll need to update the config file in core to fix a link. I'll check the other references before to avoid doing this multiple times.

### DIFF
--- a/src/content/docs/en/guides/content-collections.mdx
+++ b/src/content/docs/en/guides/content-collections.mdx
@@ -185,7 +185,7 @@ export const collections = { blog };
 
 #### Defining custom IDs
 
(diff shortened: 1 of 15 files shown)


sarah11918 (commented):
This one looks great, too!

ArmandPhilippot:
Thank you!

Just noting that I wait to check the other references before merging this in order to make a single PR in core if other files also affect the configuration reference! (this shouldn't take weeks 😄 )

ArmandPhilippot:
I'll check the links in the core repo and make a PR, so this one is ready to go!

PR #13317, opened 2026-02-26
fix: consistent headings format in `content-loader-reference.mdx`

ArmandPhilippot:
Same as https://github.com/withastro/docs/pull/13315 and #13316 

For context: In https://github.com/withastro/docs/pull/13021#discussion_r2692505870, we decided we preferred `parent.child` as headings instead of `child` both for readability and linking. We already use this format in some modules reference (e.g. [envField](https://v6.docs.astro.build/en/reference/modules/astro-config/#envfield)) but most of the modules were using the old format (e.g. [action](https://v6.docs.astro.build/en/reference/modules/astro-actions/#action)).

#### Description (required)

Updates `content-loader-reference.mdx` to use a consistent heading format (`parent.child`).

I also removed two headings (`Loader properties` and `Live loader properties`) because we usually don't use those... and I feel this is no longer necessary with the `parent.child` notation. (btw we had a nesting issue for one of them)

This one implies a lot of links updates... but I personally think this is nice (e.g. `#entries-1` vs `#livedatacollectionentries`).

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/netlify.mdx
+++ b/src/content/docs/en/guides/integrations-guide/netlify.mdx
@@ -98,9 +98,9 @@ The [Netlify Blog post on Astro](https://www.netlify.com/blog/how-to-deploy-astr
 
 ### Running Astro middleware on Netlify Edge Functions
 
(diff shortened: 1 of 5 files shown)


ArmandPhilippot:
Thanks! I take this as an explicit approval... If not, please scream while I update the branch. 😄

PR #13329, opened 2026-03-01
i18n(es): Add Checklist to the box

AitorMT:
#### Description (required)

The title “CheckList” has been added to the box, as this is how it appears in the English version.


sarah11918:
Sorry, we are no longer accepting translation updates to v5 docs! We are in a docs freeze for v5 now, and we will be accepting translations again after we release v6!

AitorMT:
When the version is re-released, can I request to open this branch again if there are no significant new changes? @sarah11918

sarah11918:
You can certainly make a new PR after v6 is released! But there are significant changes between the v5 and v6 docs, so your branch will certainly be very out of date by then. I would suggest making a brand new PR at that time.

PR #13333, opened 2026-03-01
i18n(es): Update 5.mdx

AitorMT:
#### Description (required)

A translation error has been fixed.


sarah11918:
Sorry, we are no longer accepting translation updates to v5 docs! We are in a docs freeze for v5 now, and we will be accepting translations again after we release v6!

PR #13336, opened 2026-03-02
Add preserveBuildClientDir adapter feature

matthewp:
#### Description (required)

Adds new `preserveBuildClientDir` option to the adapter features that preserves `config.build.client` on static builds.

### DIFF
--- a/astro.sidebar.ts
+++ b/astro.sidebar.ts
@@ -148,6 +148,7 @@ export const sidebar = [
 			group('reference.experimental', {
 				items: [
 					'reference/experimental-flags',
+					'reference/experimental-flags/route-caching',
 					'reference/experimental-flags/client-prerender',
 					'reference/experimental-flags/content-intellisense',
 					'reference/experimental-flags/chrome-devtools-workspace',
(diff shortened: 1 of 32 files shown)


src/content/docs/en/reference/adapter-reference.mdx
```diff
@@ -795,6 +795,38 @@ export default function createIntegration() {
 
 The value of the headers might change based on the features enabled/used by the application. For example, if [CSP is enabled](/en/reference/configuration-reference/#securitycsp), the `<meta http-equiv="content-security-policy">` element is not added to the static page. Instead, its `content` is available in the `routeToHeaders` map. 
 
+### `preserveDirectoryStructure`
+
+<p>
+
+**Type:** `boolean`<br />
+<Since v="6.0.0" />
```
sarah11918, line 805:
```suggestion
**Default:** `false`<br />
<Since v="6.0.0" />
```
Sounds like this is false by default. ( @armandphilippot I'm noticing `staticHeaders` above doesn't specify a default either, and probably should? Just want to check with you since you recently worked on this page though!)

ArmandPhilippot: Yeah, in both cases undefined or `false` behave the same. So, I think we can add a default here. IIRC, we've already done that in a few places!

sarah11918: Perfect, thanks! I think that should be all that's needed here then, to add `default: false` in both places!

matthewp: Ok, will update tomorrow.

PR #13337, opened 2026-03-03
fix: replace IDX mentions with Firebase Studio + remove Gitpod

ArmandPhilippot:
#### Description (required)

We replaced IDX with Firebase Studio and removed Gitpod in `editor-setup.mdx` in #13078 but we still have a lot of mentions of IDX and Gitpod.

v6 could be a great time to update that? I know there are redirects from IDX but this might be confusing for those who are not aware of the rebranding, especially in the tutorial.
I have not checked the steps in the tutorial because I believe this is just a rebranding and the interface should be similar.

- Suggested label: v6, typo/link/grammar - quick fix!

### DIFF
--- a/src/content/docs/en/reference/error-reference.mdx
+++ b/src/content/docs/en/reference/error-reference.mdx
@@ -48,6 +48,7 @@ The following reference is a complete list of the errors you may encounter while
 - [**RemoteImageNotAllowed**](/en/reference/errors/remote-image-not-allowed/)<br/>Remote image is not allowed
 - [**UnsupportedImageFormat**](/en/reference/errors/unsupported-image-format/)<br/>Unsupported image format
 - [**UnsupportedImageConversion**](/en/reference/errors/unsupported-image-conversion/)<br/>Unsupported image conversion
+- [**CannotOptimizeSvg**](/en/reference/errors/cannot-optimize-svg/)<br/>Cannot optimize SVG
 - [**PrerenderDynamicEndpointPathCollide**](/en/reference/errors/prerender-dynamic-endpoint-path-collide/)<br/>Prerendered dynamic endpoint has path collision.
 - [**PrerenderRouteConflict**](/en/reference/errors/prerender-route-conflict/)<br/>Prerendered route generates the same path as another route.
 - [**ExpectedImage**](/en/reference/errors/expected-image/)<br/>Expected src to be an image.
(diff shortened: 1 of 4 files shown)


src/content/docs/en/tutorial/1-setup/index.mdx
ArmandPhilippot, line 103: Should we keep that? 😄 I mean, we can always go back in the Git history if needed.

src/content/docs/en/guides/cms/frontmatter-cms.mdx
```diff
@@ -9,7 +9,7 @@ i18nReady: true
 ---
 import { FileTree } from '@astrojs/starlight/components';
 
-[Front Matter CMS](https://frontmatter.codes/) brings the CMS to your editor, it runs within Visual Studio Code, Gitpod, and many more.
+[Front Matter CMS](https://frontmatter.codes/) brings the CMS to your editor, it runs within Visual Studio Code.
```
ArmandPhilippot: I checked their website and GitHub repo and I can't find mentions of other editors, only VSCode (I guess the idea was any editor based on VSCode?). So, I just removed the end of the sentence instead of replacing with something else.

sarah11918: That was never a great sentence anyway, but it does claim:

Image

How about


```suggestion
[Front Matter CMS](https://frontmatter.codes/) brings the CMS to your editor, allowing you to create and preview content in real-time in Visual Studio Code.
```

sarah11918 (commented):
Agreed, seems like a great time!

(Fun fact, all my browser IDE links are still `idx` based 😆 )

src/content/docs/en/tutorial/1-setup/index.mdx
sarah11918: I mean, we can even *revert* to it... 😄  We can see whether anyone has any strong opinions on which one to show.

src/content/docs/en/guides/cms/frontmatter-cms.mdx
ArmandPhilippot: Oh, I'm blind! 😆 I mostly checked the docs... but, yeah, your suggestion sounds better. Thanks!

src/content/docs/en/tutorial/1-setup/index.mdx
ArmandPhilippot: I don't have a strong opinion... but this could make sense since we have also reinstated Stackblitz as the default option in [astro.new](https://astro.new/latest/).

PR #13351, opened 2026-03-06
feat(reference): document new `clientAddress`

ematipico:
#### Description (required)

This PR documents the upcoming new property of `CreateContext`

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -97,7 +97,7 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
 
 If you are using Vite-specific plugins, configuration, or APIs, check the [Vite migration guide](https://vite.dev/guide/migration) for their breaking changes and upgrade your project as needed.
 
-Using [Astro's `getViteConfig()` helper](/en/guides/testing/#vitest) requires Vitest v3.2 (Vitest v4 is not yet supported).
+Using [Astro's `getViteConfig()` helper](/en/guides/testing/#vitest) requires at least Vitest v3.2 or v4.1 beta 5.
 
 ### Vite Environment API
 


src/content/docs/en/reference/modules/astro-middleware.mdx
```diff
@@ -174,6 +174,19 @@ An object for storing arbitrary information from a middleware, accessible to the
...
+
+#### `CreateContext.clientAddress`
+
+<p>
+
+**Type:** `string`<br />
+<Since v="6.0.0" />
+</p>
+
+`clientAddress` specifies the [IP address](https://en.wikipedia.org/wiki/IP_address) of the request. It must be provided by the adapter using the context of the possible edge function.
```
ArmandPhilippot: nit, we usually don't start the sentence with the property name:
```suggestion
Specifies the [IP address](https://en.wikipedia.org/wiki/IP_address) of the request. It must be provided by the adapter using the context of the possible edge function.
```

src/content/docs/en/reference/modules/astro-middleware.mdx
ArmandPhilippot: We need to add the property under `CreateContext` too:
```suggestion
**Type:** `{ request: Request; params?: Params; userDefinedLocales?: string[]; defaultLocale: string; locals: App.Locals; clientAddress?: string }`<br />
```
If the object is growing, we can consider replacing the type with `object` instead...

src/content/docs/en/reference/modules/astro-middleware.mdx
```diff
@@ -174,6 +174,19 @@ An object for storing arbitrary information from a middleware, accessible to the
...
+
+If not provided, accessing `context.clientAddress` will throw an error.
```
ArmandPhilippot: Because this is under `CreateContext`, and not `createContext()` (the function) I wonder if using `context.` could be confusing for the user?

I don't have a strong opinion on this, so I'll let you decide what you find the most useful:

```suggestion
If not provided, accessing `clientAddress` will throw an error. 
```

ematipico:
Code PR has been merged

src/content/docs/en/reference/modules/astro-middleware.mdx
```diff
@@ -174,6 +174,19 @@ An object for storing arbitrary information from a middleware, accessible to the
...
+
+#### `CreateContext.clientAddress`
+
+<p>
+
+**Type:** `string`<br />
+<Since v="6.0.0" />
+</p>
+
+Specifies the [IP address](https://en.wikipedia.org/wiki/IP_address) of the request. It must be provided by the adapter using the context of the possible edge function.
```
sarah11918:
```suggestion
The [IP address](https://en.wikipedia.org/wiki/IP_address) of the request.

This must be provided by the adapter or platform from a trusted source (e.g. socket address, platform-provided header).
```
Armand had found this in the JSDoc for the feature. Does this convey the same meaning? I think it sounds a bit clearer, since I'm not sure what a "possible edge function" means.

matthewp: Yeah I think this is good!

PR #13354, opened 2026-03-06
docs: Add guidance regarding using getImage on the client

Princesseuh:
#### Description (required)

`getImage` on the client does not really work / lead to sub-optimal results, so we're disabling it completely in V6

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v6.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v6.mdx
@@ -97,7 +97,7 @@ Astro v6.0 upgrades to Vite v7.0 as the development server and production bundle
 
 If you are using Vite-specific plugins, configuration, or APIs, check the [Vite migration guide](https://vite.dev/guide/migration) for their breaking changes and upgrade your project as needed.
 
-Using [Astro's `getViteConfig()` helper](/en/guides/testing/#vitest) requires Vitest v3.2 (Vitest v4 is not yet supported).
+Using [Astro's `getViteConfig()` helper](/en/guides/testing/#vitest) requires at least Vitest v3.2 or v4.1 beta 5.
 
 ### Vite Environment API
 
--- a/src/content/docs/en/reference/modules/astro-middleware.mdx
+++ b/src/content/docs/en/reference/modules/astro-middleware.mdx
@@ -118,7 +118,7 @@ import type {
 
 <p>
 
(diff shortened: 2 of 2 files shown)


src/content/docs/en/reference/modules/astro-assets.mdx
```diff
@@ -638,7 +638,9 @@ Variable weight font files will be preloaded if any weight within its range is r
 </p>
 
 :::caution
-`getImage()` relies on server-only APIs and breaks the build when used on the client.
+`getImage()` relies on server-only APIs and will throw an error on the client. 
+
+If you need the resulting image URL client-side, pass the `src` from a server-rendered call to the client. See [generating images with `getImage()`](/en/guides/images/#generating-images-with-getimage) for an example.
```
```suggestion
If you need the resulting image URL client-side, you can [pass the `src` from a server-rendered `getImage()` call to the client](/en/guides/images/#generating-images-with-getimage).
```
When it's easy enough, we often inline the links with paragraph text. Fewer words to read/write. 😄  I think this works!

src/content/docs/en/guides/images.mdx
```diff
@@ -633,6 +633,24 @@ const allBlogPosts = await getCollection("blog");
 
 The `getImage()` function is intended for generating images destined to be used somewhere else than directly in HTML, for example in an [API Route](/en/guides/endpoints/#server-endpoints-api-routes). When you need options that the `<Picture>` and `<Image>` components do not currently support, you can use the `getImage()` function to create your own custom `<Image />` component.
```
sarah11918, line 635: Out of scope for this PR, but now that we have an example for using images on the client, it would probably be nice to have a short example here, too now! Maybe in an API route!

In which case, I'd probably then add a sub-heading for the section below like `### Using generating images on the client`

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -1430,6 +1430,36 @@ If you were previously relying on the fact that the image service would automati
...
+
+In Astro 5.x, calling `getImage()` from `astro:assets` on the client would silently fail or produce incorrect results.
+
+Astro 6.0 throws a runtime error when `getImage()` is called on the client.
+
+#### What should I do?
+
+Call `getImage()` on the server and pass the resulting `src` to the client instead:
+
+```astro title="src/components/ClientImage.astro"
```
```suggestion
```astro title="src/components/ClientImage.astro" {5, 8, 11}
```
Just did this by counting, but I think it's right...

PR #13364, opened 2026-03-09
Revert no-index in headers

sarah11918:
Reverts the headers used for the preview `v6.docs.astro.build`

Replaced by #13372

### DIFF
--- a/public/_headers
+++ b/public/_headers
@@ -5,7 +5,3 @@
   cache-control: max-age=31536000
   cache-control: immutable
   cache-control: public
-
-# TODO: Remove this before merging to main
-/*
-  x-robots-tag: no-index


sarah11918:
Sure, if that makes more sense! I *suspect* we are not going to keep a v6.docs branch around (or at least, for very long) because by the time we do want a dedicated v6.docs branch, it'll be when we're releasing v7 and in the meantime that branch will need to be constantly kept up with main (or mirrored I guess?)

Happy to let you decide which is the best path here!

delucis:
OK, plan of action:

1. Merge #12322 
2. Merge **this** PR (#13364 — crazy that there are 1,000 PRs between these 😄)
3. Create a temporary new v6 branch that will redirect with a `301` status code from any `v6.docs.astro.buid/*` URL to `docs.astro.build/*`. This will ensure any v6 links hanging around (e.g. v6 beta blog post, etc.) will keep working in the short-term.
4. (Medium term) Remove that temporary branch once we’re happy most people won’t hit it.
5. (Long term) Astro v7 and a new v6 branch to archive docs will be made

sarah11918:
I've also captured those steps in https://github.com/withastro/docs/pull/12322

sarah11918:
Note, after the v6 branch PR is merged, this will be merging to `main`. I'll handle that tomorrow!

PR #13369, opened 2026-03-09
Archive v5 docs

delucis:
#### Description (required)

This PR updates the `v5` branch which will be used to archive the v5 docs with changes equivalent to [what we do for our v4 archive](https://github.com/withastro/docs/compare/main...v4):

- Displays a warning banner on every page linking to the latest docs
- Uses `docs.astro.build` as the canonical `site` value to avoid this version showing up in search results

### DIFF
--- a/scripts/lint-linkcheck.ts
+++ b/scripts/lint-linkcheck.ts
@@ -65,7 +65,7 @@ class LinkChecker {
 
 // Use our class to check for link issues
 const linkChecker = new LinkChecker({
-	baseUrl: 'https://docs.astro.build',
+	baseUrl: 'https://v5.docs.astro.build',
 	buildOutputDir: './dist',
 	pageSourceDir: './src/content/docs',
 	checks: [


sarah11918:
Do we also need an i18nIgnore here (because this will merge into main)?

delucis:
> Do we also need an i18nIgnore here (because this will merge into main)?

This will merge in to `v5`, not `main`!

astro.config.ts
```diff
@@ -12,14 +12,9 @@ import { starlightPluginSmokeTest } from './config/plugins/smoke-test';
...
 
-/* https://docs.netlify.com/configure-builds/environment-variables/#read-only-variables */
-const NETLIFY_PREVIEW_SITE = process.env.CONTEXT !== 'production' && process.env.DEPLOY_PRIME_URL;
-
-const site = NETLIFY_PREVIEW_SITE || 'https://docs.astro.build/';
-
 // https://astro.build/config
 export default defineConfig({
-	site,
+	site: 'https://docs.astro.build/',
```
sarah11918, line 17: Just checking that this doesn't need a `v5.`

delucis: It does not. It intentionally points to the production docs site as the canonical source for search engines. This is the same as [what we do for our v4 archive](https://github.com/withastro/docs/compare/main...v4).

PR #13389, opened 2026-03-11
add prerenderEnvironment option to Cloudflare adapter docs

OliverSpeir:
#### Description (required)

Document `prerenderEnvironment` option added to cloudflare adapter in https://github.com/withastro/astro/pull/15711 

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/ko/guides/actions.mdx
+++ b/src/content/docs/ko/guides/actions.mdx
@@ -329,14 +329,14 @@ export const server = {
         // `type` 필드의 값이 `create`인 경우 일치합니다.
         type: z.literal('create'),
         name: z.string(),
-        email: z.string().email(),
+        email: z.email(),
       }),
       z.object({
         // `type` 필드의 값이 `update`인 경우 일치합니다.
         type: z.literal('update'),
         id: z.number(),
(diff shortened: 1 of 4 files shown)


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
sarah11918: If possible, this should probably be a `<Since v="x.x.0" pkg="@astrojs/cloudflare" />` version number!

Not blocking this PR, but if it's easy to track down and add here, would be appreciated. Otherwise, we send @ArmandPhilippot on a scavenger hunt... 🪖

ArmandPhilippot: Ah, indeed. This should be [`12.4.0`](https://github.com/withastro/astro/blob/main/packages/integrations/cloudflare/CHANGELOG.md#1240)
```suggestion
<Since v="12.4.0" pkg="@astrojs/cloudflare" />
```

matthewp: Ah, didn't see this, I added same thing in dcdde42

PR #13404, opened 2026-03-11
Update route caching documentation for production mode

cfjedimaster:
The fact that caching doesn't work in development is mentioned a bit later in the doc, but as this tripped me up I thought it made sense to add an even earlier note about this. Also, the fact that it is in prod only is mentioned in "Checking if caching is enabled" - I ignored this section when I was testing earlier as i didn't need to check, I just wanted to cache, so I was confused when it didn't work.

#### Description (required)

This helps flag the fact that caching is a production mode only feature.

I'm @raymondcamden in Discord.

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/route-caching.mdx
+++ b/src/content/docs/en/reference/experimental-flags/route-caching.mdx
@@ -122,7 +122,7 @@ Multiple calls to `cache.set()` within a single request are merged:
 
 Middleware, layouts, content loaders, and page code can each contribute cache directives independently.
 
-## Dev mode behavior
+### Dev mode behavior
 
 In dev mode, the cache API is available so that route code does not need conditional checks, but no actual caching occurs. `cache.enabled` is `false`, and `cache.set()` and `cache.invalidate()` are no-ops. To test your caching locally, build then preview your site.
 


sarah11918:
Thanks for commenting on these brand new docs, for a very new feature! 😄 

We try not to duplicate information that already exists on the page. I think this is an issue of finding the info, not it being missing. At the bottom of the page, we do have a section heading for `## Dev mode behavior`. Perhaps an option is to bring that right up under `## Usage` instead? So maybe a structure like:

```
Using Route Caching
├── Checking to see if cache is available
├── ...
├── Merge behavior
└── Dev mode behavior
```

Additionally, that section currently says:

> In dev mode, the cache API is available so that route code does not need conditional checks, but no actual caching occurs. cache.enabled is false, and cache.set() and cache.invalidate() are no-ops.

We could further add a sentence to that like:

```
To test your production site's caching locally, [build then preview your site](/en/develop-and-build/#build-and-preview-your-site).
```

What do you think of a solution more like this instead?

cfjedimaster:
I think that's fine - want me to edit the PR or... ?

sarah11918:
> want me to edit the PR or... ?

Yes, please!

cfjedimaster:
Change done - thank you!

src/content/docs/en/reference/experimental-flags/route-caching.mdx
```diff
@@ -122,6 +122,10 @@ Multiple calls to `cache.set()` within a single request are merged:
 
 Middleware, layouts, content loaders, and page code can each contribute cache directives independently.
 
+## Dev mode behavior
```
sarah11918:
```suggestion
### Dev mode behavior
```
This should be nested under usage I think!

PR #13408, opened 2026-03-11
feat: enable the Rust compiler

Princesseuh:
#### Description (required)

What the title says. In previous tests, this made the build 10% or so faster, so that's pretty good!


sarah11918:
Clicked around and tested various bits of the site (including tutorial UI) and didn't notice anything off!

morinokami:
The basic functionality seems to work fine, but when viewing the site in languages other than English, the "EN" label that appears next to links to untranslated pages in the sidebar menu seems to disappear in the Preview version.

Here is a comparison of the Japanese docs:

| Production | Preview |
|--------|--------|
| Screenshot 2026-03-12 at 23 13 42 | Screenshot 2026-03-12 at 23 15 38 |

Princesseuh:
Interesting, I'll take a look!

delucis:
Just a note on Shinya’s finding there in case you haven’t looked at it yet @Princesseuh, it seems to be related to style scoping. The `EN` is added using an `::after` CSS style. In the deployment using the Rust compiler this style **is** still present:

sources panel in browser dev tools showing a CSS rule adding the EN text content after links with the class fallback

But the component ID does not match what is in the DOM:

markup for the links we’re trying to target. they have an Astro generated hash class name that doesn’t match the one used in the CSS

Princesseuh:
This should be fixed now!

ArmandPhilippot:
Thanks for the reminder 😄 I didn't noticed any visual changes, but I noticed a change in the HTML which could impact some Starlight websites?

|Rust compiler|Go compiler|
|---|---|
|astro-docs-with-rust|astro-docs-without-rust|

This is the badges used in https://docs.astro.build/en/editor-setup/#other-code-editors

This is related to the workaround I was talking about yesterday: https://github.com/withastro/compiler/issues/1003#issuecomment-3426351769
Which is used in Starlight for `<Badge />` and `<Icon />`: https://github.com/withastro/starlight/pull/2924

This is a bit odd, because the space in the HTML output was visible before the workaround and this is no longer the case!? But, this might be something to monitor before enabling the Rust compiler by default!

delucis:
Amazing catch @ArmandPhilippot!

The source for that markup is here:

https://github.com/withastro/docs/blob/313627deda1a5e7d766376a233d4cd7f37b36285/src/content/docs/en/editor-setup.mdx#L45-L50

Like you say, might be related to https://github.com/withastro/compiler/issues/1003 and as you say, we had to [work around that in Starlight](https://github.com/withastro/starlight/blob/732c2d8883ac8cb1fa036fc63aba7edf61992de6/packages/starlight/user-components/Badge.astro#L20-L26), so I guess the Rust compiler breaks that workaround?

code comment saying “The fragment around the element is used as a workaround to avoid a trailing whitespace in the output.” and linking to the compiler issue above

Princesseuh:
@ArmandPhilippot Fixed in the last version! Let me know if you find anything else, but otherwise it seems like this might be ready for merging.

ArmandPhilippot (commented):
I tried to spot either visual changes or changes in the code source (whitespace that could become an issue) and didn't find anything.

I'm not approving explicitly because it would be great to run HiDeoo's tool before merging but this one looks ready to go for me. Great job here! 🙌🏽

PR #13433, opened 2026-03-12
Fix font-family syntax in fonts.mdx

pawelgrzybek:
#### Description (required)

The name of the custom property shouldn't be quoted.


ArmandPhilippot:
Thanks for the PR, but I don't think we need this change.

This is a CSS variable, using `var()` doesn't require quotes. See: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascading_variables/Using_custom_properties#referencing_custom_properties_with_var

If you have troubles using the Font API do not hesitate to create a support thread [on Discord](https://astro.build/chat), our community will be happy to help you!

PR #13457, opened 2026-03-14
Fix font styles from 'italics' to 'italic'

teinett:
Now:
```
styles: ["normal", "italics"],
```

Should be:
```
styles: ["normal", "italic"],
```


ArmandPhilippot (approved):
Nice catch, thank you!

edit: I updated the title because of Lunaria, we want the change to be applied to translations too!

PR #13466, opened 2026-03-14
remove CF Pages workaround

alexanderniebuhr:
#### Description (required)

- This Workaround somehow made it to the final docs. However after talking with Cloudflare and discussing internally we found out Pages is not supported and won't be supported, not even with a manual workaround.

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/en/guides/deploy/cloudflare.mdx
+++ b/src/content/docs/en/guides/deploy/cloudflare.mdx
@@ -128,7 +128,7 @@ If you're using Workers Builds:
 <Steps>
 1. Follow Steps 1-3 from the Wrangler section above.
 
-2. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com/) and navigate to `Compute > Workers & Pages`. Select `Create`.
+2. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com/) and navigate to `Compute > Workers & Pages`. Select `Create application`.
 
 3. Under `Import a repository`, select a Git account and then the repository containing your Astro project.
 


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -683,8 +646,6 @@ The Astro Cloudflare adapter now only supports deploying to Cloudflare Workers b
 
 <ReadMore>See Cloudflare's [migration guide from Pages to Workers](https://developers.cloudflare.com/workers/static-assets/migration-guides/migrate-from-pages/) for detailed migration instructions.</ReadMore>
```
ArmandPhilippot: If we no longer support Cloudflare Pages, I think this whole section should be reworded.

```suggestion
### Removed: Cloudflare Pages support

The Astro Cloudflare adapter no longer supports deployment on Cloudflare Pages. For the best experience and feature support, you should migrate to Cloudflare Workers.

<ReadMore>See Cloudflare's [migration guide from Pages to Workers](https://developers.cloudflare.com/workers/static-assets/migration-guides/migrate-from-pages/) for detailed migration instructions.</ReadMore>
```

Something like that. Do no hesitate to make it accurate or be a bit more verbose to explain why. I'm not familiar with Cloudflare.

Also, I have no confidence in the Github diff so it might be best to update that locally instead... I'm used to empty lines on non updated lines, but this is new. The preview looks really awful... My suggestion should only be applied to the "Deprecated: Cloudflare Pages support" section...

Edit: well the preview once applied is different from what I had before... but this is still wrong! So, yeah, if you agree with my suggestion, please make the changes locally!

alexanderniebuhr: Thanks, that makes sense. Made a commit with you as coauthor :)

ArmandPhilippot (approved):
Thank you! This file looks good to me (so, I'm approving to avoid blocking anything) but I think we would need to also update the deploy guide if the adapter no longer supports Pages (sorry, I missed that with my previous review), see https://docs.astro.build/en/guides/deploy/cloudflare/#cloudflare-pages

if it only consists of removing mentions of Cloudflare Pages mentions, we can handle that if you want! But, it would be helpful if someone familiar with Cloudflare scanned that page to check that nothing else needs to be deleted/modified.

alexanderniebuhr:
Let me quickly add the update to that page 😓

src/content/docs/en/guides/deploy/cloudflare.mdx
```diff
@@ -13,12 +13,11 @@ import { Steps } from '@astrojs/starlight/components';
 import StaticSsrTabs from '~/components/tabs/StaticSsrTabs.astro';
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro'
 
-You can deploy full-stack applications, including front-end static assets and back-end APIs, as well as on-demand rendered sites, to both [Cloudflare Workers](https://developers.cloudflare.com/workers/static-assets/) and [Cloudflare Pages](https://pages.cloudflare.com/). 
+You can deploy full-stack applications, including front-end static assets and back-end APIs, as well as on-demand rendered sites, to [Cloudflare Workers](https://developers.cloudflare.com/workers/static-assets/). 
 
 This guide includes:
 
 - [How to deploy to Cloudflare Workers](#cloudflare-workers)
```
ArmandPhilippot: I think the list no longer makes sense, we already have the aside below calling attention to Workers vs Pages.
```suggestion
```

src/content/docs/en/guides/deploy/cloudflare.mdx
ArmandPhilippot: Is the name still "Workers & Pages"?

alexanderniebuhr: Funny enough it actually still is, however we can improve the explanation a bit 😓 

image

alexanderniebuhr:
```suggestion
2. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com/) and navigate to `Compute > Workers & Pages`. Select `Create`.
```

src/content/docs/en/guides/deploy/cloudflare.mdx
```diff
@@ -125,14 +121,14 @@ After your assets are uploaded, Wrangler will give you a preview URL to inspect
...
-You can also use a CI/CD system such as [Workers Builds (BETA)](https://developers.cloudflare.com/workers/ci-cd/builds/) to automatically build and deploy your site on push. 
+You can also use a CI/CD system such as [Workers Builds](https://developers.cloudflare.com/workers/ci-cd/builds/) to automatically build and deploy your site on push. 
 
 If you're using Workers Builds:
 
 <Steps>
 1. Follow Steps 1-3 from the Wrangler section above.
 
-2. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com/) and navigate to `Workers & Pages`. Select `Create`.
+2. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com/) and navigate to `Compute > Workers & Pages`. Select `Create`.
```
alexanderniebuhr:
```suggestion
2. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com/) and navigate to `Compute > Workers & Pages`. Select `Create application`.
```

alexanderniebuhr:
Not waiting for another round of reviews, since I already got approval twice 😓

PR #13505, opened 2026-03-17
Align experimental SVGO documentation with v4

jpc-ae:
#### Description

The SVGO documentation was written prior to v3. One breaking change and one major change has since been made to that package. Specifically, [the `active` parameter is no longer accepted](https://svgo.dev/docs/migrations/migration-from-v2-to-v3/#active-plugins) (this was replaced by an overrides parameter in v3) and `removeViewBox` is no longer included in the default preset making removing it redundant.

The SVGO documentation is also rather sparce when it comes to documenting the config object, so further clarifications were added.

### DIFF
--- a/src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
+++ b/src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
@@ -128,7 +128,7 @@ export default defineConfig({
 })
 ```
 
-Setting `multipass` to true will run the optimization engine multiple times on each file to ensure no optimizations are missed. The `floatPrecision` parameter will be passed to all the many plugins that use it to control the number of decimal places to preserve. It can be overridden for a specific plugin by passing a value in its `params` property.
+The `multipass` option sets whether to run the optimization engine multiple times until no further optimizations are found. The `floatPrecision` option sets the number of decimal places to preserve globally, but can be overridden for a specific plugin by specifying a custom value in its `params` property.
 
 ## Common use cases
 


src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -93,36 +113,44 @@ export default defineConfig({
...
-      floatPrecision: 2,
-      multipass: true
+      multipass: true,
+      floatPrecision: 2
     }
   }
 })
 ```
 
+Setting `multipass` to true will run the optimization engine at least twice for each SVG until no further optimizations are performed. This can result in smaller file sizes at the expense of longer build times and may cause unwanted side-effects depending on how your plugins are configured.
```
FredKSchott: not sure we need to go so much into detail on these options, would probably want to combine these two paragraphs into one.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -152,13 +187,28 @@ export default defineConfig({
 
 ### Custom precision
```
FredKSchott: `floatPrecision` is documented above, no need to re-document here. safe to remove this entire section.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
jpc-ae: Fair enough. I've tried combining them.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
jpc-ae: I was considering that too. I've tried replacing it with information about the removeXMLNS and removeXlink plugins which are not default, but are recommended by SVGO for inline SVG.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -93,36 +113,42 @@ export default defineConfig({
...
     svgo: {
       plugins: [
-        'preset-default',
         {
-          name: 'removeViewBox',
-          active: false // Preserve viewBox attribute
+          name: 'preset-default',
+          params: {
+            overrides: {
+              inlineStyles: false, // Preserve style tags for csp hashing
```
yanthomasdev:
```suggestion
              inlineStyles: false, // Preserve style elements for CSP hashing
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -93,36 +113,42 @@ export default defineConfig({
...
-      floatPrecision: 2,
-      multipass: true
+      multipass: true,
+      floatPrecision: 2
     }
   }
 })
 ```
 
+Setting `multipass` to true will run the optimization engine multiple times on each file to ensure no optimizations are missed. The `floatPrecision` parameter will be passed to all the many plugins that use it to control the number of decimal places to preserve. It can be overridden for a specific plugin by passing a value in its `params` property.
```
yanthomasdev: I'll go an extra mile and say we can probably omit this paragraph as a whole. The link to the configuration options already explains what each of these options do. What do you think?

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -93,36 +113,42 @@ export default defineConfig({
...
+
 ## Common use cases
 
-SVGO provides an extensive [default plugin list](https://svgo.dev/docs/preset-default/) with opinionated optimizations that is more convenient than adding each plugin individually. However, you may need to customize it further for your needs. For example, it may remove items or clean up too aggressively for your situation.
+SVGO provides an extensive [default plugin list](https://svgo.dev/docs/preset-default/) with opinionated optimizations. Using this preset is more convenient than adding each plugin individually; however, you may need to customize it further for your needs. For example, it may remove items or clean up too aggressively for your situation, especially when using animations.
 
 ### Preserve specific attributes
 
-You may want to preserve certain SVG attributes, such as the `viewBox`, that SVGO removes by default:
+You may want to preserve certain SVG attributes, such as style tags, that SVGO inlines or removes by default:
```
yanthomasdev: The sentence is a bit inconsistent here, it mentions SVG attributes but then talks about "style tags", which I know is in the context of not inlining them into an attribute, but that's a bit counterintuitive. And I guess in this case it would also be an style *element*, since it has both an opening and closing tag.

What about this instead?
```suggestion
You may want to preserve certain SVG attributes and elements, such as `<style>`, that SVGO inlines or removes by default:
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -93,36 +113,42 @@ export default defineConfig({
...
       plugins: [
-        'preset-default',
         {
-          name: 'removeViewBox',
-          active: false // Preserve viewBox attribute
+          name: 'preset-default',
+          params: {
+            overrides: {
+              inlineStyles: false, // Preserve style tags for csp hashing
+              removeDesc: false // Keep tag regardless of contents
```
yanthomasdev:
```suggestion
              removeDesc: false // Keep element regardless of contents
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -139,26 +165,42 @@ export default defineConfig({
...
     }
   }
 })
 ```
 
-### Custom precision
+### Optimize for inlining in modern HTML5
 
-Control the precision of numeric values in path data:
+Inline SVG does not require the xmlns tag and can safely be converted to SVG 2. The `removeXMLNS` and `removeXlink` plugins are therefore recommended:
```
yanthomasdev: Correct me if I am wrong, but the `xmlns` I know is an attribute and not a tag, or is there also an `xmlns` element? Just making sure we are being consistent!

I did a bit of a wording change as well, let me know what you think!
```suggestion
Inline SVG does not require the `xmlns` attribute and can be safely converted to the SVG 2 specification. The `removeXMLNS` and `removeXlink` plugins are recommended for this purpose:
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -73,17 +78,32 @@ export default defineConfig({
 
 An array of [SVGO plugins](https://svgo.dev/docs/plugins/) that will be used to optimize your SVG component imports.
 
-This can include any plugins by ID name, including SVGO's `preset-default` collection of plugins. A plugin can optionally be passed as an object including both its `name` and `active` status, to enable or disable as necessary.
+This can include SVGO's [`preset-default`](https://svgo.dev/docs/preset-default/) plugin collection, individual built-in plugins or [custom plugins](https://svgo.dev/docs/plugins-api/). To use a plugin’s default configuration, simply add its name to the list. To override specific plugins in `preset-default`, use the overrides parameter. To override specific parameters of a plugin, pass them in an object along with the plugin name.
```
yanthomasdev: This is a lot for a single paragraph, I did some changes to make it easier to read, let me know what you think!

```suggestion
This can include SVGO's [`preset-default`](https://svgo.dev/docs/preset-default/) plugin collection, individual built-in plugins, or [custom plugins](https://svgo.dev/docs/plugins-api/). 

To use a plugin’s default configuration, add its name to the array. If you need more control, use the `overrides` parameter to customize specific plugins within `preset-default`, or pass an object with a plugin's `name` to override its individual parameters.
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
```diff
@@ -93,36 +113,42 @@ export default defineConfig({
...
   }
 })
 ```
 
+Setting `multipass` to true will run the optimization engine multiple times on each file to ensure no optimizations are missed. The `floatPrecision` parameter will be passed to all the many plugins that use it to control the number of decimal places to preserve. It can be overridden for a specific plugin by passing a value in its `params` property.
+
 ## Common use cases
 
-SVGO provides an extensive [default plugin list](https://svgo.dev/docs/preset-default/) with opinionated optimizations that is more convenient than adding each plugin individually. However, you may need to customize it further for your needs. For example, it may remove items or clean up too aggressively for your situation.
+SVGO provides an extensive [default plugin list](https://svgo.dev/docs/preset-default/) with opinionated optimizations. Using this preset is more convenient than adding each plugin individually; however, you may need to customize it further for your needs. For example, it may remove items or clean up too aggressively for your situation, especially when using animations.
```
yanthomasdev: I generally don't like the use of `;` in technical documentation like this, I'd rather reword and break it up in different sentences altogether, what do you think?

```suggestion
SVGO provides an extensive [default plugin list](https://svgo.dev/docs/preset-default/) with opinionated optimizations. While using this preset is more convenient than adding each plugin individually, you may need to customize it further. For example, it may remove items or clean up too aggressively for your situation, especially when using animations.
```

Otherwise, I would just say maybe we can just keep the previous iteration with the added context about animations:
```suggestion
SVGO provides an extensive [default plugin list](https://svgo.dev/docs/preset-default/) with opinionated optimizations that is more convenient than adding each plugin individually. However, you may need to customize it further for your needs. For example, it may remove items or clean up too aggressively for your situation, especially when using animations.
```

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
jpc-ae: Agreed!

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
jpc-ae: I like this too. Some of my text may have ended up clunky when I was trying to not change the original text too much, hah.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
jpc-ae: The problem is that the SVGO dev docs don't really document `floatPrecision` properly, and if I document that one here I might as well do both. I've submitted a docs PR to the SVGO team, but their maintainers are stretched thin it seems. I'd prefer to keep it for now.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
jpc-ae: Yep, that was me getting lost in the weeds. Great catch! I like the wording changes too.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
jpc-ae: Looks good. I see you prefer the oxford comma, which is fine.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
jpc-ae: Yeah, it was a bit clunky, but I couldn't think of a better way. I like your first suggestion, since it encourages developers to customise.

src/content/docs/en/reference/experimental-flags/svg-optimization.mdx
yanthomasdev: Right, I think once this feature is stable and your PR to them is merged we'll be able to remove this and link to their docs instead.

I would like to simplify this sentence a bit, what do you think?:
```suggestion
The `multipass` option will run the optimization engine multiple times per file until no more optimizations are found, and the `floatPrecision` option sets the number of decimal places to preserve across plugins. It can be overriden for a specific plugin by passing a custom value to its respective `params` property.
```

jpc-ae: How about the following? I think it works better as two sentences. Also overridden definitely has two Ds.

```suggestion
The `multipass` option sets whether to run the optimization engine multiple times until no further optimizations are found. The `floatPrecision` option sets the number of decimal places to preserve globally, but can be overridden for a specific plugin by specifying a custom value in its `params` property.
```

yanthomasdev: Sounds good to me, @jpc-ae! Teamwork makes dream work! 🚀

jpc-ae: Absolutely! Thanks for all the suggestions.

FredKSchott:
@jpc-ae thanks for the revisions! @yanthomasdev looks like this is ready for another review.

PR #13538, opened 2026-03-19
[i18nIgnore] Remove `llms.txt` files

delucis:
#### Description (required)

This PR removes the docs `llms.txt` files and updates our “Building Astro sites with AI tools” guide to remove mentions of them.

Currently we generate a range of Markdown files including `llms-full.txt` and `llms-small.txt` for every build. However, we’ve seen little uptake recently in usage — these files get very little traffic. Instead we can focus efforts on the [MCP server](https://github.com/withastro/docs-mcp) and perhaps in the future offer per-page Markdown content.

Apologies to anyone who was relying on these! AI trends move quickly 😢 

#### Related issues

- Closes #13006

#### Notes

Anecdotal obviously but here’s a quick example of potential build time benefits of skipping this:

two Netlify builds one taking 5m 6s, the other for this branch taking 4m 22s


ascorbic:
I don't think this should be done unless and until we have per-page markdown

FredKSchott:
LGTM! would be nice if we had analytics data on this but alas.

I don't consider per-page markdown related/dependent on this PR, no ones agents are requesting the `llm.txt` file because the docs page is returning HTML. Having that would have no impact on `llm.txt` usage or usefulness AFAIK.

delucis:
> would be nice if we had analytics data on this but alas.

We do! See the linked discussion: https://github.com/withastro/docs/discussions/13006

FredKSchott:
ah nice I hadn't read #13006, but how did you calculate the 0.01% of traffic number? AFAICT that wouldn't show up in fathom, and I couldn't find it in netlify

delucis:
> how did you calculate the 0.01% of traffic number? 

This was using Netlify’s newer “Observability” panel which provide more granular per-route insights. It’s limited to a 7-day window on our plan, but that felt like a big enough sample to get a rough idea.

Looks like relative usage has increased slightly since my post to ~0.3% of requests.

FredKSchott:
> `requests for files like this made up <0.1% of our traffic.`
Do you have an idea of total number of requests in last 7 days? Hard to put this into context.

But regardless, I'm +1 on removing.

delucis:
> Do you have an idea of total number of requests in last 7 days? Hard to put this into context.

Yeah, we can see a total # of requests. Just checked and it’s back down to ~0.1% again.

Roughly 5k requests to `llms.txt`, `llms-full.txt`, and `llms-small.txt` over the last 7 days. `llms-small.txt` accounts for >50% of those requests, which is interesting given in theory the `llms.txt` file is the entrypoint to discover the others — suggests some direct usage perhaps.

Of those ~5k, user-agent breakdown is:

- 35% GoogleBot
- 30% browser UAs
- 15% [ChatGPT-User](https://developers.openai.com/api/docs/bots)
- long tail of <2% UAs like `curl`, `node`, `opencode` and stuff like that

I’m not 100% sure how to interpret the browser UAs, but this suggests a majority of requests should perhaps be discounted as noise because e.g. crawlers and browser visits are not relevant usage.

One more number to put this into perspective: our MCP server has had more than 3 **million** requests in the same time span with UAs like Cursor and Claude Code dominating. Obviously `llms.txt` and the MCP are slightly different resources and I’d expect an MCP to be queried more frequently than `llms.txt`, but here the MCP usage is ~1,000x larger which seems significant.

ArmandPhilippot (approved):
It seems Fred has already implicitly approved this, but if you want an explicit one you have mine!

The discussion is the most upvoted and given the percentage of use, I don't think it's worth slowing down the build for that.

jlarmstrongiv:
Ahh, I’ve been using the llms.txt lately, surprised to see it disappeared today

delucis:
@jlarmstrongiv Sorry about that! More context in #13006.

PR #13557, opened 2026-03-20
Update clever-cloud.mdx

leo-clever-cloud:
Fixing a broken URL

- Closes #
- Suggested label:

### DIFF
--- a/src/content/docs/fr/guides/deploy/clever-cloud.mdx
+++ b/src/content/docs/fr/guides/deploy/clever-cloud.mdx
@@ -11,7 +11,7 @@ i18nReady: true
 import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 
-[Clever Cloud](https://clevercloud.com) est une plateforme cloud européenne qui fournit des services automatisés et évolutifs.
+[Clever Cloud](https://clever.cloud.com) est une plateforme cloud européenne qui fournit des services automatisés et évolutifs.
 
 ## Configuration du projet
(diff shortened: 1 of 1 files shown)


ArmandPhilippot:
Thank you, but it seems that this URL doesn't work either, and this one should be `https://www.clever.cloud/` instead?

And I can see this is not the only URL that requires an update, we have more `www.clevercloud.com` occurrences. Also, part of our process is to update the English version first as this is the source of truth, then, once merged, we update the translations to apply the changes.

src/content/docs/fr/guides/deploy/clever-cloud.mdx
```diff
@@ -11,7 +11,7 @@ i18nReady: true
 import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
 import PackageManagerTabs from '~/components/tabs/PackageManagerTabs.astro';
 
-[Clever Cloud](https://clevercloud.com) est une plateforme cloud européenne qui fournit des services automatisés et évolutifs.
+[Clever Cloud](https://clever.cloud.com) est une plateforme cloud européenne qui fournit des services automatisés et évolutifs.
```
yanthomasdev, line 14: This link doesn't work, I believe the new URL is `cleaver.cloud` instead!


```suggestion
[Clever Cloud](https://clever.cloud) est une plateforme cloud européenne qui fournit des services automatisés et évolutifs.
```

yanthomasdev (changes requested):
Thanks @leo-clever-cloud, can I ask you to also propagate this to the English version of this guide? Note that the link is wrong, I've added a suggestion explaining it.

PR #13571, opened 2026-03-23
New Ecosystem tab

FredKSchott:
Restructures the docs sidebar from 5 tabs to 4 tabs:
- **Tutorial (new):** Moves the tutorial into its own, new tab section.
- **Guide:** Brings together all guides about Astro into a single section. Absorbs the old "start" tab group, minus the tutorial.
- **Reference:** Unchanged.
- **Ecosystem (new):** Brings together all content about the Astro ecosystem into a single section. Combines integrations, third-party services, etc. 

My initial goal was to clean up our third-party + first-party integration content. As part of that, I'll create a single organizational structure that covers both, which means restructuring the docs sidebar from `5 -> 4` in a way that doesn't feel bolted on. The next PR will tackle the actual third-party guides themselves.

The only content change was combining "publishing to npm" with "integrations overview" into a new "Guides > Working with Integrations" guide. This felt natural, but also allowed us to cleanly merge the old "Integrations" tab group sections into the new "Ecosystem" by getting rid of the two "special" pages and moving them into a focused guide.

Translations were updated with the help of Opus 4.6. Because I was just moving content around and not creating anything net new, this felt like the right approach with minimal risk to translation errors, but @yanthomasdev LMK if any concerns.

Screenshot 2026-03-22 at 11 59 54 PM
Screenshot 2026-03-22 at 11 59 59 PM
Screenshot 2026-03-23 at 12 00 04 AM
Screenshot 2026-03-23 at 12 00 09 AM


FredKSchott:
yea it wasn't so bad until the URL changed for the integration guide. @yanthomasdev WDYT? I can also keep using the URL of the old integration guide, which would minimize the number of changes needed to lunaria.

yanthomasdev:
@FredKSchott there's no "changes" needed to Lunaria, we just need to give it a pointer to ignore all these link changes, so just adding `@lunaria-track:src/content/nav/en.ts` to the merge commit description. Just need to take some time to go through all files and make sure that's right

ArmandPhilippot:
I didn't want to give an immediate answer, I wanted to give myself time to adjust to the change (maybe I'm not completely finished with that... 👀 ).

This is not AI generated 😀 , but I know I can be quite verbose sometimes so I think some headings might help to make it easier to scan. And if that seems overwhelming, feel free to read "To conclude" first. 😅 

### Tutorial relocation

I think the tutorial being more visible is a good thing because I saw some users on Discord asking how/where they can find a tutorial (did they check the documentation first though?). But, as Chris pointed out this needs a bit more thinking.

### Merging "Publish to npm" with "Working with Integrations"

I'm a bit concerned about the "Publish to npm" and "Working with Integrations" merge. But maybe this is just a me thing! This sounds a bit weird to me when I read the following:
```md
## Publishing Your Integration to npm

Publishing an Astro component is a great way to reuse your existing work across your projects
```

In my head, "integrations" are kinda plugins for Astro and components are something different.

I know this was in the "Integrations" tab before but this didn't felt natural to me. However, because the page title was "Publish to npm" (ie. kinda generic) this made some sense.

Now, reading the first sentence under this heading feels even more wrong to me. I still think "Publish to npm" as a separate page made sense. And maybe, this shouldn't be tied to integrations given the contents.

I saw some people on Discord asking how to share their components across repo or in a monorepo. Maybe "Publish to npm" wasn't visible enough for them and the page, as a standalone page, deserves to be more visible, not merged. And maybe this shouldn't be only about "publish to npm" but kinda "share your work" (well, not that, but something like that).

### Overall feeling

I am not claiming that the current docs organization is perfect. If it were my website, I would organize some pages differently. For example, "Testing" is an important concept to me and I would put it in the same place as "Environment variables", not in "Third-party services"/"Ecosystem"... But, perhaps because I am not a native English speaker, I would say that I would more easily look for it in the "Third-party services" tab rather than in "Ecosystem".

But does the changes make life easier for everyone? That's more difficult to answer!
And, if we want to make more reorganization, we might need a wider range of feedback? Something like what was done for the v5 reorganization through T&D could be useful, but maybe using another format to gather more feedback.

### A thing that bothers me a little

This is not necessarily related to this PR... and I know Chris might disapprove because this is tied to URLs and according to [Hypertext Styles](https://www.w3.org/Provider/Style/URI) "Cool URIs don't change". 😄 

For me, URLs are kinda breadcrumbs (ie. reading `/en/guides/view-transitions` I know I'm browsing/I'll go to the View transitions guide in English language). There were already inconsistencies before (because of previous changes), but I feel like they're even more noticeable now. (e.g.  `/basics`, `/concept`, `/guides`, `/recipes`, and top-level URLs such as `/upgrade-astro` are all under "Guides").

If we plan to reorganize pages soon , I guess we shouldn't fix that... even if this is a bit confusing. 😅 And I know this is not easy to find permanent URLs when docs need to be reorganized quite often. So, this is more a "thing that bothers me" than a "thing that needs to be fixed".

(and so, I think the link renaming is a good thing! `guides/integrations` sounds cleaner to me than `/guides/integrations-guide`).

### To conclude

I don't mind trying something different if this can help our users! Except my concern about "Publish to npm" and what Chris raised about the tutorial, this looks good to me.

The changes are pretty minimal if we disregard the link renaming. I didn't notice anything wrong! So, once Chris concerns is addressed (and if I'm the only one thinking this about "Publish to npm"), I think we can try that.

Well, nothing wrong, except in French... It seems some non-breaking spaces were replaced with regular spaces which means, using the same resolution, some users can see:
|In the deploy preview|In the live docs|
|---|---|
|image|image|

Punctuation marks should not be isolated on a new line. But, not something blocking. I can always update that after! 🫡

FredKSchott:
> `Publishing an Astro component is a great way to reuse your existing work across your projects`

I generally agree, I tried to change as little content as possible in this PR, just combined the two guides together to keep scope managable, so my guess is that that was just how the old guide was written. I didn't introduce that term.

Can you lead a re-write in a follow-up PR if you're interested? That way we can keep scope here small given everything else I'm changing, but then treat it more like an enhancement post-merge.

ArmandPhilippot:
Yeah, I understand the intent to keep things small! Thanks for that!

My point was more... I'm not sure the merge is a good thing? Because the "Publish to npm" is not tied to integrations (according to my definition at least) but also addresses components published on npm. So, as it stands, I'm not sure how to improve the wording for a "Publishing Your Integration to npm" section. Everything under this heading is about publishing components to npm, not integrations.

I mean, either we need specific content for integrations here and we move the component-related content elsewhere (not sure where yet)... or a dedicated page for publishing on npm remains relevant.

But sure, if we can agree on what would be best, I don't mind making the changes afterwards!

ArmandPhilippot:
Ah, yeah, that page was merged today. I couldn't fix that directly through Github so I pushed a commit, I hope you don't mind Fred!

FredKSchott:
> IMO, I think we should just remove the translated integrations.mdx and let translators do that in their own pace without the risk of AI having made any mistakes with the nuance of spacing and so on that is very relevant in a few languages.

done! removed the translated integrations.md files

PR #13584, opened 2026-03-24
feat: document new babel options for Preact integration

Princesseuh:
#### Description (required)

Docs for https://github.com/withastro/astro/pull/15862

### DIFF
--- a/src/content/docs/en/guides/integrations-guide/preact.mdx
+++ b/src/content/docs/en/guides/integrations-guide/preact.mdx
@@ -177,7 +177,7 @@ Currently, the `compat` option only works for React libraries that export code a
 
 <p>
 
-**Type:** [`BabelOptions`](https://babeljs.io/docs/options)<br />
+**Type:** [`BabelOptions`](https://github.com/preactjs/preset-vite#babel-configuration)<br />
 <Since pkg="@astrojs/preact" v="5.1.0" />
 </p>
 


src/content/docs/en/guides/integrations-guide/preact.mdx
```diff
@@ -173,6 +173,33 @@ Check out the [`pnpm` overrides](https://pnpm.io/package_json#pnpmoverrides) and
 Currently, the `compat` option only works for React libraries that export code as ESM. If an error happens during build-time, try adding the library to `vite.ssr.noExternal: ['the-react-library']` in your `astro.config.mjs` file.
 :::
 
+### `babel`
+
+<p>
+
+**Type:** `BabelOptions`<br />
```
delucis: Named interfaces that don’t point anywhere tend not to feel very helpful to me (like “Thank you Astro docs, the type of the `babel` option is `BabelOptions`”) but because we don’t own the type we don’t really have more docs to link to. Maybe we can link to the Preact Vite plugin docs here to help?

```suggestion
**Type:** [`BabelOptions`](https://babeljs.io/docs/options)<br />
```

src/content/docs/en/guides/integrations-guide/preact.mdx
```diff
@@ -173,6 +173,33 @@ Check out the [`pnpm` overrides](https://pnpm.io/package_json#pnpmoverrides) and
...
+<Since pkg="@astrojs/preact" v="5.1.0" />
+</p>
+
+You can pass additional [Babel configuration options](https://babeljs.io/docs/options) to the Preact Vite plugin. This allows you to customize the Babel transformation applied to your Preact components.
+
+For example, to enable Babel to load `.babelrc` when processing your Preact components, pass the following object to the Preact integration:
```
delucis: Small style suggestion here:

```suggestion
For example, the following configuration tells Babel to load `.babelrc` when processing your Preact components:
```

src/content/docs/en/guides/integrations-guide/preact.mdx
```diff
@@ -173,6 +173,33 @@ Check out the [`pnpm` overrides](https://pnpm.io/package_json#pnpmoverrides) and
 Currently, the `compat` option only works for React libraries that export code as ESM. If an error happens during build-time, try adding the library to `vite.ssr.noExternal: ['the-react-library']` in your `astro.config.mjs` file.
 :::
 
+### `babel`
+
+<p>
+
+**Type:** [`BabelOptions`](https://babeljs.io/docs/options)<br />
```
delucis: I am a lemon and pasted the wrong link in my suggestion. This is the one I was thinking:
```suggestion
**Type:** [`BabelOptions`](https://github.com/preactjs/preset-vite#babel-configuration)<br />
```
But you can decide which you prefer. (This one felt less intimidating than the Babel docs and the usage example is closer to what an Astro–Preact user would actually do.)

src/content/docs/en/guides/integrations-guide/preact.mdx
```diff
@@ -173,6 +173,33 @@ Check out the [`pnpm` overrides](https://pnpm.io/package_json#pnpmoverrides) and
...
+<Since pkg="@astrojs/preact" v="5.1.0" />
+</p>
+
+You can pass additional [Babel configuration options](https://babeljs.io/docs/options) to the Preact Vite plugin. This allows you to customize the Babel transformation applied to your Preact components.
```
yanthomasdev, line 184: should we also change the Babel docs link here as well? @delucis

delucis: Not necessarily — this one could make sense I think? If we did think they should point to the same thing, then probably this link should be removed. But I can see a direct link to Babel options being helpful here.

yanthomasdev: Agree, just wanted to be sure!

PR #13586, opened 2026-03-25
feat: document fallbackRoutes being present on integration resolved routes

Princesseuh:
#### Description (required)

Documentation for https://github.com/withastro/astro/pull/15455

### DIFF
--- a/src/content/docs/en/reference/integrations-reference.mdx
+++ b/src/content/docs/en/reference/integrations-reference.mdx
@@ -1764,7 +1764,7 @@ When the value of `IntegrationResolvedRoute.type` is `redirect`, the value will 
 <Since v="6.1.0" />
 </p>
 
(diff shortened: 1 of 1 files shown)


src/content/docs/en/reference/integrations-reference.mdx
```diff
@@ -1755,6 +1756,17 @@ Determines whether the route use [on demand rendering](/en/guides/on-demand-rend
...
 
+#### `IntegrationResolvedRoute.fallbackRoutes`
+
+<p>
+
+**Type:** `IntegrationResolvedRoute[]`<br />
+<Since v="6.1.0" />
+</p>
+
+When the project uses [i18n with fallback routes](en/guides/internationalization/#fallback), the value will be a list of `IntegrationResolvedRoute` the route falls back to when the requested locale isn't available. The fallback content may be served as a redirect or rewrite depending on `i18n.routing.fallbackType`. Otherwise, the value will be an empty array.
```
delucis: Link fix and small suggestion to prefer plain language over the “IntegrationResolvedRoute” type name:

```suggestion
When the project uses [i18n with fallback routes](/en/guides/internationalization/#fallback), the value will be a list of the routes this route falls back to when the requested locale isn't available. The fallback content may be served as a redirect or rewrite depending on `i18n.routing.fallbackType`. Otherwise, the value will be an empty array.
```

PR #13598, opened 2026-03-26
i18n(fr): add `guides/integrations.mdx`

ArmandPhilippot:
#### Description (required)

* Creates the French translation of `guides/integrations.mdx` based on changes from #13571 and #13596
* Updates the links to `/guides/integrations` in other pages

This basically merges the old `integration-guides/index.mdx` and `reference/publish-to-npm.mdx` but while I was there:
* I've done some rewording
* I tried to make the headings consistent (e.g. "Créer..." then "Développement de..." vs "Créer..." then "Développer...", this looks better in the TOC I think)
* The table in "Catégories" was translating the UI strings used in https://astro.build/integrations/ filters... this can be confusing. So, I added the translation between parentheses. This is not ideal with small devices because of the table, but I don't think this looks too bad.

- Suggested label: i18n


src/content/docs/fr/guides/integrations.mdx
```diff
@@ -0,0 +1,588 @@
...
+  "version": "1.0.0",
+  "homepage": "https://github.com/utilisateur/projet#readme",
+  "type": "module",
+  "exports": {
+    ".": "./index.js",
+    "./astro": "./MonComposantAstro.astro",
+    "./react": "./MonComposantReact.jsx"
+  },
+  "files": ["index.js", "MonComposantAstro.astro", "MonComposantReact.jsx"],
+  "keywords": ["astro","astro-component", "...", "..."]
```
ld-web:
```suggestion
  "keywords": ["astro", "withastro", "astro-component", "...", "..."]
```

Should we put the same keywords as the original version ? Can be ignored if not !

src/content/docs/fr/guides/integrations.mdx
```diff
@@ -0,0 +1,588 @@
...
+     - nom-test-01.astro
+     - nom-test-02.astro
+     - nom-test-03.astro
+   </FileTree>
+</Steps>
+
+
+### Publier votre composant
+
+Une fois votre paquet prêt, vous pouvez le publier sur npm à l'aide de la commande `npm publish`. Si cela ne fonctionne pas, assurez-vous de vous être connecté via `npm login` et que votre fichier `package.json` est correct. Si la publication réussit, c'est terminé !
```
ld-web:
```suggestion
Une fois que votre paquet est prêt, vous pouvez le publier sur npm à l'aide de la commande `npm publish`. Si cela ne fonctionne pas, assurez-vous de vous être connecté via `npm login` et que votre fichier `package.json` est correct. Si la publication réussit, c'est terminé !
```

Wording, can be ignored.

src/content/docs/fr/guides/integrations.mdx
```diff
@@ -0,0 +1,588 @@
...
+
+Notez qu'aucune étape de compilation (`build`) n'a été nécessaire pour les paquets Astro. Tout type de fichier pris en charge nativement par Astro, comme `.astro`, `.ts`, `.jsx` et `.css`, peut être publié directement sans étape de compilation.
+
+Si vous avez besoin d'un autre type de fichier qui n'est pas pris en charge nativement par Astro, ajoutez une étape de compilation à votre paquet. Cet exercice avancé est laissé à votre discrétion.
+
+### Bibliothèque des intégrations
+
+Partagez votre travail en ajoutant votre intégration à notre [bibliothèque d'intégrations](https://astro.build/integrations/) !
```
ld-web: You probably already thought about that, but I can't find something that conveys "hard work" properly, and would maybe be closer to the original version 🤔

src/content/docs/fr/guides/integrations.mdx
ArmandPhilippot: Instead of starting from scratch I copied/pasted [what we had before in `publish-to-npm`](https://github.com/withastro/docs/pull/13571/changes#diff-fc562ea4d2977541a2c55ce0bc211ca4da5e3a6c28370121defab12aa075348a) and I wasn't expected something to fix there. 😅 So, yes, absolutely, nice catch!

src/content/docs/fr/guides/integrations.mdx
ArmandPhilippot: The ellipsis was intentional, but I don't mind reverting this!

src/content/docs/fr/guides/integrations.mdx
ArmandPhilippot: I haven't thought about this because I just took what we had before in `publish-to-npm` here. 😄 

But, now that I think about that... I personally don't have issues we just "travail" here and all I can think of is:
* "travail acharné" or "labeur" sounds a bit too much I think?
* maybe "vos efforts"?

Does this sounds better to you?

ld-web: I'm okay with "travail" as well, but I think it's a good idea to emphasize a bit more about the "hard work" in the original version.

Therefore, I would probably use something like "Partagez les fruits de votre travail" or "Partagez le résultat de vos efforts".

What do you think ?

ArmandPhilippot: I think "fruits de votre travail" could be misunderstood (ie. sharing the recognition/retribution/... and not the result).

But why not "résultat de vos efforts"!

```suggestion
Partagez le résultat de vos efforts en ajoutant votre intégration à notre [bibliothèque d'intégrations](https://astro.build/integrations/) !
```

ArmandPhilippot:
Looks like this is ready to go now! Thanks again for the reviews! 🙌🏽

PR #13674, opened 2026-04-05
i18n(ja): create craft cms page

fkatsuhiro:
#### Description (required)

Create ja translation : [craft cms page](https://github.com/withastro/docs/blob/main/src/content/docs/en/guides/cms/craft-cms.mdx)

- Suggested label: `i18n`


morinokami (approved):
LGTM!! Thanks!

PR #13704, opened 2026-04-09
Add notes about `glob()`'s normalization of `id`s in upgrade guides

OverflowCat:
#### Description

This PR updates the documentation to clarify that collection entry IDs are automatically lowercased when migrating legacy content collections to `glob()`. In legacy content collections, file names with uppercase characters could be used directly as IDs, while `glob()` normalizes them to kebab-case. This change is currently not clearly documented in the migration guide, which may lead to confusion for users.

```js
getEntry('data', 'testTest') // ❌
getEntry('data', 'testtest') // ✅
```

#### Related issues & labels

- This has been reported in https://github.com/withastro/astro/issues/13265#issuecomment-2666292571:

> Would still be useful to have that in the migration guide or content collections docs, as it's not expected.

### DIFF
--- a/src/content/docs/en/guides/upgrade-to/v5.mdx
+++ b/src/content/docs/en/guides/upgrade-to/v5.mdx
@@ -121,7 +121,7 @@ See the instructions below for updating an existing content collection (`type: '
 
 1. **Move the content config file**. This file no longer lives within the `src/content/` folder. This file should now exist at `src/content.config.ts`.
 
-2. **Edit the collection definition**. Your updated collection requires a `loader` which indicates both a folder for the location of your collection (`base`) and a `pattern` defining the collection entry filenames and extensions to match. (You may need to update the example below accordingly. You can use [globster.xyz](https://globster.xyz/) to check your glob pattern.) The option to select a collection `type` is no longer available.
+2. **Edit the collection definition**. Your updated collection requires a `loader` which indicates both a folder for the location of your collection (`base`) and a `pattern` defining the collection entry filenames and extensions to match. (You may need to update the example below accordingly. You can use [globster.xyz](https://globster.xyz/) to check your glob pattern.) The option to select a collection `type` is no longer available. Different from content collctions, the `glob()` loader normalizes `id`s to a kebab-cased slug by default, so hard-coded `id`s that contain capital letters (like `getEntry("UpperCase")`) in your code may break. You can pass a custom [`generateID()` helper function](/en/reference/content-loader-reference/#generateid) to override this behavior.
 
     ```ts ins={3,8} del={7}
     // src/content.config.ts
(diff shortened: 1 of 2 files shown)


src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -612,6 +612,8 @@ const blog = defineCollection({
   }),
 });
 ```
+
+Different from content collections, the `glob()` loader normalizes `id`s to kebab-cased slugs by default, so hard-coded `id`s including capital letters (like `getEntry("UpperCase")`) in your code may break. You can pass a custom [`generateID()` helper function](/en/reference/content-loader-reference/#generateid) to override this behavior.
```
ArmandPhilippot, line 616: What about something like:
```suggestion
By default, the `glob()` loader generates an ID by converting all uppercase letters to lowercase. Therefore, you may need to check and update your calls to `getEntry()` or [use the `generateId()` helper to restore the old behavior](/en/guides/content-collections/#defining-custom-ids).
```

* Unless I'm mistaken, what has changed is not really the slugs in kebab-case but the identifier in lowercase, so we should probably be more direct in drawing attention to this new behavior.
* We have an example in the content collections guide to restore the old behavior. So, it might be more helpful to use that link rather than the reference.

src/content/docs/en/guides/upgrade-to/v5.mdx
```diff
@@ -121,7 +121,7 @@ See the instructions below for updating an existing content collection (`type: '
...
+2. **Edit the collection definition**. Your updated collection requires a `loader` which indicates both a folder for the location of your collection (`base`) and a `pattern` defining the collection entry filenames and extensions to match. (You may need to update the example below accordingly. You can use [globster.xyz](https://globster.xyz/) to check your glob pattern.) The option to select a collection `type` is no longer available. Different from content collctions, the `glob()` loader normalizes `id`s to a kebab-cased slug by default, so hard-coded `id`s that contain capital letters (like `getEntry("UpperCase")`) in your code may break. You can pass a custom [`generateID()` helper function](/en/reference/content-loader-reference/#generateid) to override this behavior.
```
ArmandPhilippot, line 124: This step is becoming very long and hard to parse. And, the code snippet below doesn't display `generateId()`.
I wonder if it wouldn't be more readable to use the wording suggested for the v6 upgrade guide and place it below the code snippet.

ArmandPhilippot (commented):
Thank you, I think this can be helpful! I left two suggestions: I think we have a more helpful link and, I'm not quite sure about the current format for the v5 guide.

ArmandPhilippot:
I'm closing this due to the inactivity, thanks for trying. Feel free to reopen with an improved version if you want to!

PR #13164, opened 2026-01-28
fix(v6): update integrations reference with v6 changes

ArmandPhilippot:
#### Description (required)

Updates `integrations-reference.mdx` on v6 branch to reflect changes made on the core `next` branch, mostly in https://github.com/withastro/astro/pull/14306

##### Added

- [`SSRManifest.serverLike`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L76-L82)
- [`SSRManifest.serverIslandMappings`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L93)
  - `SSRManifest.serverIslandMap` becomes [`SSRManifest.serverIslandMappings.serverIslandMap`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L47-L50)
  - `SSRManifest.serverIslandNameMap` becomes [`SSRManifest.serverIslandMappings.serverIslandNameMap`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L47-L50)
- [`SSRManifest.sessionDriver()`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L98)
- [`SSRManifest.rootDir`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L105)
- [`SSRManifest.assetsDir`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L107)
- [`SSRManifest.devToolbar`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L111-L124)
- [`SSRManifest.logLevel`](https://github.com/withastro/astro/blob/cdb99c87c6404ec1dfb1ed0945b27e6fe7f81592/packages/astro/src/core/app/types.ts#L126)
[...]

PR #13715, opened 2026-04-10
Add docs for the Renderer API

### LINKED ISSUE #13199: Document the renderer API
### 📚 Subject area/topic

Other development APIs

Existing
- https://docs.astro.build/en/reference/integrations-reference/#addrenderer-option
- https://docs.astro.build/en/reference/integrations-reference/#astrorenderer
New page:
- https://docs.astro.build/en/reference/renderer-reference/

Currently we do not document in depth how to build a renderer, we only have types (which is better than nothing!). I think we should have a dedicated page for it, like other APIs

_No response_

ArmandPhilippot:
#### Description (required)

Adds a new reference page for the Renderer API:
* move some types from the Integration API to the new Renderer API (some have been reworded, not all)
* reuse a structure similar to the Adapter API to explain how to build a renderer
* clean up the Integration reference
  * `SSRManifest.pageModule.renderers` was removed in v6, the type was a leftover from #13164
  * we now describe twice `serverEntrypoint`/`clientEntrypoint` in the Renderer API, so we might not need a third description in the Integration API, and I'm not sure the React link was helpful? (ie. this was old content, maybe it was useful at that time)

Feedback welcome:
1. We don't want to show complex code snippets, but something that is enough to understand how this works. I tried to provide minimal code snippets based on the existing renderers but I'm not sure if this is too minimal or not.
2. It seems `/en/reference/renderer-reference` existed before, so I had to update the redirections. I'm not sure how Netlify handles them... maybe we want another slug for the page to avoid any conflicts?

3. I wonder if it would make sense to have a new sidebar group for integration-related APIs instead of lengthening the list in "Other development APIs"? Something like:
    Screenshot of the sidebar reorganization

- Closes #13199


src/content/docs/en/reference/renderer-reference.mdx
```diff
@@ -0,0 +1,330 @@
...
+i18nReady: true
+tableOfContents:
+  minHeadingLevel: 2
+  maxHeadingLevel: 3
+---
+import Since from '~/components/Since.astro';
+import { Tabs, TabItem } from '@astrojs/starlight/components';
+import ReadMore from '~/components/ReadMore.astro';
+
+Astro is designed to support any UI frameworks. This ability is powered by renderers, which are [integrations](/en/reference/integrations-reference/). See the [front-end frameworks guide](/en/guides/framework-components/) to learn how to use UI components from different frameworks in Astro.
```
matthewp:
```suggestion
Astro is designed to support any UI framework. This ability is powered by renderers, which are [integrations](/en/reference/integrations-reference/). See the [front-end frameworks guide](/en/guides/framework-components/) to learn how to use UI components from different frameworks in Astro.
```

src/content/docs/en/reference/renderer-reference.mdx
```diff
@@ -0,0 +1,330 @@
...
+
+## What is a renderer?
+
+A renderer is a special kind of [integration](/en/reference/integrations-reference/) that tells Astro how to detect and render component syntaxes it does not handle natively, such as [UI framework components](/en/guides/framework-components/). A renderer consists of two parts:
+- a server module imported during development and production builds to render components to HTML.
+- an optional client module imported in the browser to hydrate components with [client directives](/en/reference/directives-reference/#client-directives).
```
matthewp, line 20: great explanation of these two things

matthewp (approved):
Definitely much better than what we had before. Great job!

yanthomasdev (approved):
LGTM! Amazing work @ArmandPhilippot!

ArmandPhilippot:
If even Yan doesn't have nitpicks about the wording, I guess this one is ready to be shipped! Thank you both for the reviews!

I guess the sidebar reorganization (my third point in "Feedback welcome") can wait (NWTWWHB), especially since Fred wants to do some clean up around third-party guides. Perhaps he has other changes in mind.

PR #13786, opened 2026-04-22
docs(astro-assets): clarify getImage() options don't match all Image props

### LINKED ISSUE #13301: Incorrect description of UnresolvedImageTransform in getImage()
### 📚 Subject area/topic

Image service, getImage

https://docs.astro.build/en/reference/modules/astro-assets/#getimage

Image

According to the [docs](https://docs.astro.build/en/reference/modules/astro-assets/#getimage)
> This [getImage] takes an options object with the [same properties as the Image component](https://docs.astro.build/en/reference/modules/astro-assets/#image-) (except alt) and returns a [GetImageResult object](https://docs.astro.build/en/reference/modules/astro-assets/#getimageresult).

But hat's not the case. `getImage()` also doesn't accept `layout` property.

_No response_

nielskaspers:
## Summary

The reference docs state that \`getImage()\` takes an options object with \"the same properties as the Image component (except \`alt\`)\". That isn't accurate: \`getImage()\` accepts \`UnresolvedImageTransform\`, which does not include \`priority\` or \`layout\` either — those are \`<Image />\`-only props.

The same wording was duplicated in the \`LocalImageService\` / \`ExternalImageService\` \`getImage()\` entry, so both are updated.

## Issue

Fixes #13301

## Changes

- \`src/content/docs/en/reference/modules/astro-assets.mdx\`: reword the two \`getImage()\` descriptions to say \"a subset of the Image component properties (excluding \`alt\`, \`priority\`, and \`layout\`)\".

## Testing

- Docs-only change; no code or configuration touched.

### DIFF
--- a/src/content/docs/en/reference/modules/astro-assets.mdx
+++ b/src/content/docs/en/reference/modules/astro-assets.mdx
@@ -645,7 +645,7 @@ If you need the resulting image URL client-side, you can [pass the `src` from a 
 
 The `getImage()` function is intended for generating images destined to be used somewhere else than directly in HTML, for example in an [API Route](/en/guides/endpoints/#server-endpoints-api-routes). It also allows you to create your own custom `<Image />` component.
 
-This takes an options object with the [same properties as the Image component](#image-) (except `alt`) and returns a [`GetImageResult` object](#getimageresult).
+This takes an options object with a subset of the [Image component properties](#image-) (excluding `alt`, `priority`, and `layout`) and returns a [`GetImageResult` object](#getimageresult).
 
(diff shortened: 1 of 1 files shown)


ArmandPhilippot (changes requested):
Thanks for jumping on this but this isn't the right fix:
* We shouldn't have to list all the properties not included. This is a sign that something is wrong here.
* `getImage()` accepts a new property (`background`, not documented, we should) since this was written. So, this is no longer accurate.

And, we do not accept [automated PRs](https://agentscan.netlify.app/user/nielskaspers). Please follow our PR template...

ArmandPhilippot:
Closing due to inactivity.

PR #13798, opened 2026-04-29
feat: Add docs for image.followRedirects option

louisescher:
#### Description (required)

This PR adds a new reference documentation for the `image.followRedirects` option.

N/A


ArmandPhilippot:
Hi Lou, FYI this PR isn't necessary, you can read at the top of the file (which should be read-only):
> NOTE: This file is auto-generated from 'scripts/docgen.mjs'
> Do not make edits to it directly, they will be overwritten.
> Instead, change this file: https://github.com/withastro/astro/blob/main/packages/astro/src/types/public/config.ts

I mean it's okay to do that if you want to check the rendering but ultimately the content for that page will be pulled from the core repo. So, they either need to match exactly or this PR can be closed. 😉

louisescher:
Whoops, my bad! Can you tell I've never done this before? But sure, let's just say I wanted to check the rendering. That works 😆

louisescher:
So, we _do_ need docs! Reopening.

PR #13808, opened 2026-05-01
docs: use the Bun runtime

weakish:
By default, `bun run` starts the Astro dev server on the Node.js runtime.
This change adds instructions for using the Bun runtime instead.

Reasons:

1. Users running Astro with Bun may prefer to use the Bun runtime.
2. The CI and production environments may only include Bun, not Node.js; configuring Bun runtime keeps development, CI, and production consistent.
3. Bun’s official Astro guide also documents using the Bun runtime:
   https://bun.com/docs/guides/ecosystem/astro

### DIFF
--- a/src/content/docs/en/recipes/bun.mdx
+++ b/src/content/docs/en/recipes/bun.mdx
@@ -84,6 +84,25 @@ bun run build
 
 When the build is finished, run the appropriate preview command (e.g. `bun run preview`) in your terminal and you can view the built version of your site locally in the same browser preview window.
 
+### Runtime
+
+By default, Bun will run the Astro dev server with Node.js.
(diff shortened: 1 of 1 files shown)


ArmandPhilippot:
Thanks for willing to contribute but I'm not convinced we should add this to the documentation.

The reasons given don't change the fact that Astro doesn't support the Bun runtime. Adding this mention could encourage users down an unsupported path.

And, please read [our contributing guide](https://github.com/withastro/docs/blob/main/CONTRIBUTING.md). These kinds of addition should start with a [Discussion](https://github.com/withastro/docs/discussions). This is not a "very obvious documentation fixes like typos or broken links".

PR #13784, opened 2026-04-21
fix(sessions): explain how to override options at runtime

ArmandPhilippot:
#### Description (required)

* Updates `guides/sessions` to show how to override driver options at runtime.

  I took inspiration from the code snippet provided by Florian in https://github.com/withastro/roadmap/discussions/1214#discussioncomment-15454277 with some minor changes:
  * the example was mixing `.ts` and `.js`... not sure if this was intentional
  * I guess this was written before `sessionDrivers` stabilization and we don't need to use `unstorage` directly

* Updates `configuration-reference` to:
  * replace the [deprecated usage](https://docs.astro.build/en/guides/upgrade-to/v6/#deprecated-session-driver-string-signature) in "[Session Options](https://docs.astro.build/en/reference/configuration-reference/#session-options)"
  * add more links to the session guide to emphasize that runtime variables may require adjustment
  * reorganize a bit the "Session Options" section

Note that I updated `configuration-reference` here to check and validate this is what we want, but we'll need to do the same changes in the core repo.

#### Related issues & labels (optional)

- Closes #13028
- Suggested label: improve/update documentation

PR #13826, opened 2026-05-04
i18n(fr): update `configuration-reference.mdx`

ArmandPhilippot:
#### Description (required)

Adds changes from #13802, #13784, #13806, #13844, and #13849 to the French translations of `configuration-reference.mdx` and `integrations-reference.mdx`.

- Suggested label: i18n


src/content/docs/fr/reference/configuration-reference.mdx
```diff
@@ -1579,6 +1584,19 @@ Les options d'encodage par défaut transmises à `sharp().png()` lors de l'utili
...
+### image.dangerouslyProcessSVG
+
+<p>
+
+**Type :** `boolean`<br />
+**Par défaut :** `false`<br />
+<Since v="6.3.0" />
+</p>
+
+  Permet aux images SVG d'être traitées par le pipeline d'optimisation des images.
```
yanthomasdev: I think this whitespace wasn't meant to be here?
```suggestion
Permet aux images SVG d'être traitées par le pipeline d'optimisation des images.
```

ArmandPhilippot: Ah, nice catch! I must have accidentally pressed the tab key. 😅

trueberryless (commented):
Thanks for your translation!

LGTM 🎉

#### Further potential improvements (outside diff range)

- **`src/content/docs/fr/reference/integrations-reference.mdx` line 1424**: Heading structure mismatch: the original English file has **14** `h3` subsection(s) under this `h2` section, but this translation has **15**. Please check for missing or extra headings.

These are out of scope for this PR, so let's create follow-up PR(s) for them 👍

---

EDIT: Hi Armand! I'm testing a little project I am working on and it discovered that the French translation of `integrations-reference.mdx` still includes the `Importations de types dépréciées` section, which got removed in the English doc I think. That's what the message above is about...

ArmandPhilippot:
Ah, nice catch! It seems I forgot a change in #12322 ... I double checked; the line numbers now match. Thanks!

Edit: but the line number was wrong, the section was at line 2623 not at line 1424.

trueberryless:
> Ah, nice catch! It seems I forgot a change in #12322 ... I double checked; the line numbers now match. Thanks!
> 
> Edit: but the line number was wrong, the section was at line 2623 not at line 1424.

Thank you very much for the confirmation!
I only noticed because I am currently building a little tool to catch common "mistakes" for Astro docs: https://astro-docs-i18n-reviewer.trueberryless.org/

But it's still in the early phase, so thanks for the feedback with the line number 👍 

Edit: Looking closer at this specific case, depending on how you would determine the suggestion line, line 1424 is technically correct for saying: "this is the heading for the section which has a heading mismatch under it" (so it's the first possible line where the heading mismatch affects the missing / additional part)...

Another option would be to suggest on the last possible heading line of the mismatching section, I'm actually not sure what would be nearer in most cases... I'll have to think about it.

yanthomasdev (approved):
LGTM! Good work!

PR #13732, opened 2026-04-15
feat: svg optimizer

florian-lefebvre:
#### Description (required)

`svgOptimizer` replaces `svgo`. Choosing which version to use for the `Since` component is tricky: SVG optimization exists since 5.16 but the API changes a lot in 6.2. So I decided not to touch it

#### Related issues & labels (optional)

- Closes #
- Suggested label:

PR #13828, opened 2026-05-04
fix: update experimental `svgOptimizer` default value

ArmandPhilippot:
#### Description (required)

The `experimental-flags/svg-optimization` page has been updated in #13732 but we forgot to update the `Default:` value.

Since 6.2, the type is no longer `boolean | object` but `SvgOptimizer`, so it cannot be `false` by default... Actually, it [doesn't have a default value](https://github.com/withastro/astro/blob/7711e471ab8950d00f38ae1dd1acba451705dd51/packages/astro/src/core/config/schemas/base.ts#L544) but I guess using `undefined` is fine here.

- Suggested label: quick fix


yanthomasdev (approved):
LGTM

PR #13848, opened 2026-05-07
Document Cloudflare binding build-time image optimization

Daedalus-Icarus:
## Changes

- Documents the new compound `imageService` configuration for opting in to build-time image optimization with the Cloudflare Images binding.
- Clarifies that `imageService: 'cloudflare-binding'` keeps the existing runtime-only behavior.

## Related

- Supports withastro/astro#16194


src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -128,7 +128,7 @@ Determines which image service is used by the adapter. The adapter will default
 - **`compile`:** Uses the [configured image service](/en/reference/configuration-reference/#imageservice) or falls back to a combination of internal dependencies to transform images locally at build time for prerendered routes. The noop `passthrough` option is configured for on-demand rendered pages.
 - **`custom`:** Uses the [configured image service](/en/reference/configuration-reference/#imageservice) (defaults to Sharp) to process assets at build time for prerendered routes. Bundles the image service for runtime image handling, without checking its compatibility with Cloudflare's `workerd` runtime.
 
-It is also possible to configure your image service as an object, setting both a build time and runtime service independently. Currently, `'compile'` is the only available build-time option. The supported runtime options are `'passthrough'` (default) and `'cloudflare-binding'`:
+It is also possible to configure your image service as an object, setting both a build time and runtime service independently. The supported build-time options are `'compile'` and `'cloudflare-binding'`. The supported runtime options are `'passthrough'` (default) and `'cloudflare-binding'`:
```
alexanderniebuhr: The `'passthrough'` "(default)" claim no longer holds for the whole union. In withastro/astro#16194, `normalizeImageServiceConfig` resolves an omitted `runtime` as:

```js
runtimeService: config.runtime ?? (config.build === 'compile' ? 'passthrough' : config.build)
```

So with `build: 'cloudflare-binding'`, an omitted `runtime` defaults to `'cloudflare-binding'`, not `'passthrough'`.

```suggestion
It is also possible to configure your image service as an object, setting both a build time and runtime service independently. The supported build-time options are `'compile'` and `'cloudflare-binding'`. The supported runtime options are `'passthrough'` and `'cloudflare-binding'`. When `runtime` is omitted, it defaults to `'passthrough'` for `build: 'compile'`, and to `'cloudflare-binding'` for `build: 'cloudflare-binding'`:
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -141,6 +141,19 @@ export default defineConfig({
 });
 ```
 
+To opt in to build-time image optimization with the Cloudflare Images binding during `workerd` prerendering, set `build` to `'cloudflare-binding'`. The default string shorthand, `imageService: 'cloudflare-binding'`, keeps the existing runtime-only behavior.
```
alexanderniebuhr: Non-blocking: the changeset in withastro/astro#16194 notes that if the binding fails to transform an image at build time, the adapter falls back to the Node-side image service (e.g. Sharp). Worth a sentence here so users aren't surprised to see Sharp output despite opting into the binding:

```suggestion
To opt in to build-time image optimization with the Cloudflare Images binding during `workerd` prerendering, set `build` to `'cloudflare-binding'`. If the binding fails to transform an image at build time, the adapter falls back to the configured Node-side image service (e.g. Sharp). The default string shorthand, `imageService: 'cloudflare-binding'`, keeps the existing runtime-only behavior.
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -141,6 +141,19 @@ export default defineConfig({
...
+
+```js title="astro.config.mjs" ins={6}
+import { defineConfig } from 'astro/config';
+import cloudflare from '@astrojs/cloudflare';
+
+export default defineConfig({
+  adapter: cloudflare({
+    imageService: { build: 'cloudflare-binding', runtime: 'cloudflare-binding' }
```
alexanderniebuhr, line 158: Nit: `runtime: 'cloudflare-binding'` is the default when `build` is `'cloudflare-binding'` (see comment above), so this could be just `imageService: { build: 'cloudflare-binding' }`. Fine to keep the explicit form for clarity, but the surrounding prose shouldn't imply the explicit `runtime` is required for the binding to be used at runtime.

Also: since this documents unreleased behavior from a `minor` changeset in `@astrojs/cloudflare`, consider adding a `<Since v="x.y.z" pkg="@astrojs/cloudflare" />` marker to this new section once the release version is known, and holding this PR for merge-on-release (astro#16194 is still open).

alexanderniebuhr (commented):
Checked this against the implementation in withastro/astro#16194. The type signature and the opt-in/shorthand behavior descriptions match the code. One correction needed on the `runtime` default claim (inline suggestion below), plus two non-blocking notes.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
ArmandPhilippot: Your suggestion looks correct to me; thanks for catching that! I'm only quibbling about the form. As it stands (including your suggestion), it seems a bit difficult to read:

Updated docs with Alex suggestion

We should probably rework this paragraph now that there is more content. Maybe something like:

```suggestion
<p><Since v="14.2.0" pkg="@astrojs/cloudflare" /></p>

You can also configure the image service as an object by providing a build-time service (`compile` or `cloudflare-binding`) and, optionally, a runtime service (`passthrough` or `cloudflare-binding`).

When the runtime service is omitted, its default value depends on the configured build-time service. It falls back to `'passthrough'` for `build: 'compile'` and to `'cloudflare-binding'` for `build: 'cloudflare-binding'`.
```

Feel free to reword/update! But I think two paragraphs might be easier to parse:

Updated docs with my suggestion

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
ArmandPhilippot: I'm not familiar with how Cloudflare works so reading the whole section (ie. `imageService`) is a bit confusing. It's not clear to me this is new behavior. Isn't using `imageService: 'cloudflare-binding'` the same as using `imageService: { build: 'cloudflare-binding' }`?

I'm particularly confused because of:
> The default string shorthand, `imageService: 'cloudflare-binding'`, keeps the existing runtime-only behavior.

Is it true that `imageService: 'cloudflare-binding'` is only for the runtime service!?

If so, we should probably update the description above. Reading the following it's not clear to me this is runtime only:
> cloudflare-binding: Uses the [Cloudflare Images binding](https://developers.cloudflare.com/images/transform-images/bindings/) for image transformation. The binding is automatically provisioned when you deploy.

But then, I wonder why the two don't behave in the same way. If this is confusing for me ("new" to Cloudflare), I guess I would not be the only one. I have some concerns about the DX, and I wonder if the implementation could be "consistent.".

If it's not something fixable, then, maybe we should explain the difference instead of saying "keeps the existing runtime-only behavior"... and/or we should update the description for `cloudflare-binding` taking inspiration of `compile` and `custom`.

If it's the same... I'm not sure we need this sentence and the code snippet below.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
ArmandPhilippot: If we need this snippet (see my comment above), I agree with Alex, `build` only makes more sense:

```suggestion
    imageService: { build: 'cloudflare-binding' }
```

ArmandPhilippot (commented):
Thanks Alex for handling this PR! I particularly appreciate this as I am not familiar with Cloudflare and every time the options change it confuses me even more. 😅

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
ArmandPhilippot: Alright, I think I better understand how it works now, and it seems this is new behavior...

So, yes, we need Alex's suggestion. But, I would reorder the flow to say loud that the shorthand is not some syntactic sugar. And "configured Node-side image service" is not something we ever mention in docs... If this means the configured image service, we don't need to confuse users with extra words, and Sharp is not an example but the default.

Please correct me if I'm wrong! But, maybe something like the following would be clearer:

```suggestion
Using `imageService: 'cloudflare-binding'` enables image optimization using the [Cloudflare Images binding](https://developers.cloudflare.com/images/transform-images/bindings/) only at runtime.

To opt in to build-time image optimization with the binding during `workerd` prerendering, set `build` to `'cloudflare-binding'`. If the binding fails to transform an image at build time, the adapter falls back to the [configured image service](/en/reference/configuration-reference/#imageservice) (defaults to Sharp).
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
ArmandPhilippot: And yes, we need a `<Since />` as Alex pointed out, so I have updated my suggestion.

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -116,7 +116,7 @@ It also accepts the following:
 ### `imageService`
 
 <p>
-**Type:** `'passthrough' | 'cloudflare' | 'cloudflare-binding' | 'compile' | 'custom' | { build: 'compile', runtime?: 'cloudflare-binding' | 'passthrough' }`<br />
+**Type:** `'passthrough' | 'cloudflare' | 'cloudflare-binding' | 'compile' | 'custom' | { build: 'compile', runtime?: 'cloudflare-binding' | 'passthrough' } | { build: 'cloudflare-binding', runtime?: 'cloudflare-binding' | 'passthrough' }`<br />
```
ArmandPhilippot: Given we have the same options, I wonder if we can simplify that for docs:

```suggestion
**Type:** `'passthrough' | 'cloudflare' | 'cloudflare-binding' | 'compile' | 'custom' | { build: 'compile' | 'cloudflare-binding', runtime?: 'cloudflare-binding' | 'passthrough' }`<br />
```

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
```diff
@@ -141,6 +145,21 @@ export default defineConfig({
...
+
+To opt in to build-time image optimization with the binding during `workerd` prerendering, set `build` to `'cloudflare-binding'`. If the binding fails to transform an image at build time, the adapter falls back to the [configured image service](/en/reference/configuration-reference/#imageservice) (defaults to Sharp).
+
+```js title="astro.config.mjs" ins={6}
+import { defineConfig } from 'astro/config';
+import cloudflare from '@astrojs/cloudflare';
+
+export default defineConfig({
+  adapter: cloudflare({
+    imageService: { build: 'cloudflare-binding', runtime: 'cloudflare-binding' }
```
ArmandPhilippot, line 158: I noticed we still show `build` and `runtime` here. Have you changed your mind about removing `runtime`? It's not a problem if that's the case, I just want to make sure it wasn't marked as resolved by mistake.

ArmandPhilippot (approved):
I left one last nit to simplify the type and a non-blocking question. Otherwise, it seems ready to me. Thank you both!

src/content/docs/en/guides/integrations-guide/cloudflare.mdx
alexanderniebuhr: I think even if it is the default value, we should still be able to configure it. Maybe the code PR needs an update here.

ArmandPhilippot:
Thanks again! I'm updating the branch and merging this, so we have a single PR to merge for the 7.2 release.

PR #13875, opened 2026-05-12
fix(v7): remove Rust compiler page and use `alpha` in upgrade command

ArmandPhilippot:
#### Description (required)

* Updates the upgrade guide to use the `alpha` tag in the upgrade command
* Removes the Rust compiler from experimental flags

- Suggested label: v7


delucis (approved):
Thanks for cleaning those pages up @ArmandPhilippot!

PR #13890, opened 2026-05-14
Extend advanced routing docs with new features added by the RFC

matthewp:
#### Description (required)

Adds new docs for:

- `fetchFile` option
- extending Provider types
- the Fetchable. type.

Associated with https://github.com/withastro/astro/pull/16723

### DIFF
--- a/src/content/docs/en/reference/configuration-reference.mdx
+++ b/src/content/docs/en/reference/configuration-reference.mdx
@@ -111,6 +111,41 @@ See your hosting platform's documentation for more information. You cannot use A
 **See Also:**
 - build.format
 
+### fetchFile
+
+<p>
+
+**Type:** `string | null`<br />
+**Default:** `'app'`<br />
+<Since v="6.4.0" />
+</p>
+
+Customize the file used as the [advanced routing](/en/reference/experimental-flags/advanced-routing/) entrypoint inside your `srcDir`. Defaults to `'app'`, meaning Astro looks for `src/app.ts`.
+
+Set to a different string to use a different filename:
+
+```js
+export default defineConfig({
+  fetchFile: 'fetch.ts',
+  experimental: {
+    advancedRouting: true,
+  },
+});
+```
+
+With this configuration, Astro will look for `src/fetch.ts` instead of `src/app.ts`.
+
+Set to `null` to disable the advanced routing entrypoint entirely. This is useful if you already have a `src/app.ts` file used for other purposes:
+
+```js
+export default defineConfig({
+  fetchFile: null,
+  experimental: {
+    advancedRouting: true,
+  },
+});
+```
+
 ### redirects
 
 <p>
--- a/src/content/docs/en/reference/configuration-reference.mdx
+++ b/src/content/docs/en/reference/configuration-reference.mdx
@@ -111,40 +111,6 @@ See your hosting platform's documentation for more information. You cannot use A
 **See Also:**
 - build.format
 
-### fetchFile
-
-<p>
-
-**Type:** `string | null`<br />
-**Default:** `'app'`
-</p>
-
-Customize the file used as the [advanced routing](/en/reference/experimental-flags/advanced-routing/) entrypoint inside your `srcDir`. Defaults to `'app'`, meaning Astro looks for `src/app.ts`.
-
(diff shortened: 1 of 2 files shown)


src/content/docs/en/reference/experimental-flags/advanced-routing.mdx
```diff
@@ -265,6 +296,20 @@ try {
 }
 ```
 
+### Typing custom providers with `App.Providers`
```
ArmandPhilippot: I think this section should either be moved outside of `astro/fetch handler reference`, or be a `<h4 />` (at the same level as "Properties" and "Methods") because the children of `astro/fetch handler reference` are what can be imported:

Image

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -111,6 +111,40 @@ See your hosting platform's documentation for more information. You cannot use A
 **See Also:**
 - build.format
 
+### fetchFile
+
+<p>
+
+**Type:** `string | null`<br />
+**Default:** `'app'`
```
ArmandPhilippot: I don't why you removed `<Since />` in [25bc533](https://github.com/withastro/docs/pull/13890/commits/25bc533af328a0d9301cd9e853e8e5fe7f5d38b8) but this is new so we should have the version here:
```suggestion
**Default:** `'app'`<br />
<Since v="6.4.0" />
```

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -111,6 +111,40 @@ See your hosting platform's documentation for more information. You cannot use A
...
+</p>
+
+Customize the file used as the [advanced routing](/en/reference/experimental-flags/advanced-routing/) entrypoint inside your `srcDir`. Defaults to `'app'`, meaning Astro looks for `src/app.ts`.
+
+Set to a different string to use a different filename:
+
+```js
+export default defineConfig({
+  fetchFile: 'fetch.ts',
```
ArmandPhilippot: Is the `.ts` extension expected here? I mean the default do not have an extension so it looks a bit weird and looking at [the implementation](https://github.com/withastro/astro/blob/f2baf2ce0193f9973cb7b366ccd3b536ba7e4cad/packages/astro/src/core/fetch/vite-plugin.ts#L51-L52) it seems this should be `fetch` instead?

ArmandPhilippot (commented):
Looks good to me overall but I left a few comments (section out of place, missing since, and a code snippet that looks a bit odd to me).

And, I see you have updated `src/content/docs/en/reference/configuration-reference.mdx` here but the content is different from [`packages/astro/src/types/public/config.ts`](https://github.com/withastro/astro/pull/16723/changes#diff-3919010fb2f9bfd24993b91e720147659af609f7f770c14a9aab6fb2908f5fe3). I think this is great to show the two examples but we'll need to apply the same changes in `packages/astro/src/types/public/config.ts` since this file is auto-generated.

src/content/docs/en/reference/configuration-reference.mdx
ArmandPhilippot: Ah, I see Ema posted a comment on the implementation so this might not be useful!

matthewp: Because this won't be released in a minor, i'm not sure when it will be released. Do we add these for things released in a patch?

ArmandPhilippot: Yeah I understood that after Ema comment. 😄 But, yes we also add `<Since />` for patches. This helps users who have not yet upgraded their project to distinguish what is available to them and what is not.

src/content/docs/en/reference/configuration-reference.mdx
matthewp: The extension is optional, we just pass it into vite.resolve, so it can have the extension or not. I thought since it's called fetchFile it should be a filename, which includes the extension.

I'd rather not document that it's optional as this is an implementation detail that we could change in the future.

ArmandPhilippot (approved):
Thanks, LGTM!

yanthomasdev (approved):
LGTM! Good work!

PR #13903, opened 2026-05-17
docs: clarify Astro.url and Astro.params runtime behavior

### LINKED ISSUE #16313: URL Decoding in Dynamic Routes (SSR)
### Astro Info

```block
Astro                    v6.1.6
Vite                     v7.3.2
Node                     v22.22.0
System                   Linux (x64)
Package Manager          npm
Output                   server
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

I’m not sure whether this is a bug or a documentation issue.
The values of `Astro.url` and `new URL(Astro.request.url)` are different.

This seems to be related to the following change:
[https://docs.astro.build/en/guides/upgrade-to/v6/#removed-percent-encoding-in-routes](https://docs.astro.build/en/guides/upgrade-to/v6/#removed-percent-encoding-in-routes)

However, the API reference for `Astro.url` states:

> Astro.url is equivalent to doing new URL(Astro.request.url).

which does not match the current behavior.

Additionally, regarding `Astro.params`, there is no mention of decoding in the documentation:
[https://docs.astro.build/en/reference/api-reference/#params](https://docs.astro.build/en/reference/api-reference/#params)

`Astro.url` and `new URL(Astro.request.url)` should be identical (i.e., `%25` should not be decoded).

Currently, the `pathname` of `Astro.url` is partially decoded, which results in an invalid URL and makes it unusable for purposes such as redirects.

----

The same issue applies to `Astro.params`.
It should either not be decoded at all, or if decoding is intended, `decodeURIComponent` should be used instead of `decodeURI`.

Alternatively, it could be made configurable via `astro.config`.

https://stackblitz.com/edit/github-wipyksgd?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

Prayaksh:
## **Description (required)**

This PR clarifies the runtime behavior of URL decoding in Astro v6 dynamic routes.

The original issue was reported in the `withastro/astro` repository, and after maintainer confirmation it was identified as a documentation gap rather than a runtime bug.

In Astro v6, `Astro.url` returns a normalized and partially decoded URL, which may differ from `new URL(Astro.request.url)`. Additionally, `Astro.params` values are automatically decoded. This behavior was not clearly documented in the v6 upgrade guide section covering percent-encoded routes.

This update adds a runtime clarification under the “Percent-Encoding in routes” section to reflect the actual behavior and avoid confusion for users working with encoded dynamic route segments.

---

## **Related issues & labels (optional)**

* Closes withastro/astro#16313
* Suggested label: `documentation`

---


ArmandPhilippot (commented):
Thanks for your contribution!

If this is indeed a documentation issue (I wouldn't say "probably" means "confirmation"), this addition can be helpful but it doesn't fully fix the issue.

The issue mentions a misleading description in https://docs.astro.build/en/reference/api-reference/#url and potentially in https://docs.astro.build/en/reference/api-reference/#params

@ematipico can you confirm we should fix the docs? Or the fix mentioned by the bot could work and no docs change is required?

**Edit:** I hadn't followed the linked PR, it seems this might have been fixed in https://github.com/withastro/astro/pull/16808 ?

Prayaksh:
I realize now that I acted on the “probably” comment before the issue was fully confirmed. Thank you for clarifying. I’m still learning the process and will make sure to verify before updating the docs in the future.

ematipico:
The documentation update is still valid, because not all characters can be safely decoded e.g. `%25AB`. Even after my fix, there are still cases that are uncertain, so Astro play it safe and returns an error.

ArmandPhilippot:
Okay, thanks Ema!

And no problem @Prayaksh! Are you still willing to check if other pages require an update? It seems we at least need to update https://docs.astro.build/en/reference/api-reference/#url

Prayaksh:
Thanks @ArmandPhilippot !  I’ve updated the Astro.url and Astro.params API reference docs to clarify the normalization/decoding behavior as well.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -137,6 +137,8 @@ In static builds, this will be the `params` returned by `getStaticPaths()` used
 
 When routes are rendered on demand, `params` can be any value matching the path segments in the dynamic route pattern.
 
+`params` values are automatically decoded from route pathnames.
```
ArmandPhilippot: This doesn't seem to be correct? Reading this, I get the impression that everything is correctly decoded. But, the bot said:
> Since Astro.params are derived from this same normalized pathname (via getParams(routeData, pathname) at line 104), the corruption also affects param values.

And looking at the minimal repro in the linked PR, for `test+test.png`, I can see:

Image

Which means the bot is right and not everything is automatically decoded.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -165,7 +167,9 @@ See also: [`params`](/en/reference/routing-reference/#params)
 
 `url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` value. It is useful for interacting with individual properties of the request URL, like pathname and origin.
```
ArmandPhilippot: I think the first line you added can be here instead:
```suggestion
A normalized [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object derived from the current `request.url` value. It is useful for interacting with individual properties of the request URL, like pathname and origin. 
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -165,7 +167,9 @@ See also: [`params`](/en/reference/routing-reference/#params)
 
 `url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` value. It is useful for interacting with individual properties of the request URL, like pathname and origin. 
 
-`Astro.url` is equivalent to doing `new URL(Astro.request.url)`.
+`Astro.url` is a normalized `URL` instance derived from `Astro.request.url`.
+
+In Astro v6, the pathname is partially decoded during route normalization, so `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
```
ArmandPhilippot: Then, we don't need the first line, and "Since" is probably better than "In" because the behavior will be probably the same in Astro v7:

```suggestion
Since Astro v6, the pathname is partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -832,6 +832,13 @@ If you have route files with `%25` in the filename, rename them to use a differe
 src/pages/test%25file.astro
 src/pages/test-file.astro
 ```
+Astro v6 also changed how URLs and route parameters are handled at runtime.
+
+`Astro.url` returns a normalized and partially decoded URL, so it may differ from `new URL(Astro.request.url)`.
+
+`Astro.params` values are automatically decoded, and percent-encoded characters in pathnames or parameters (like `%25`) may appear decoded in `Astro.url` or `Astro.params`.
+
+If you need the raw request URL with no decoding, use `Astro.request.url` and parse it manually.
```
ArmandPhilippot: This doesn't seem to be the right place:
* In "What should I do?" we explain what the user should do based on the change describe above. Here we are explaining that something else has changed without explaining what the user should do.
* This is in the "Removed: Percent-Encoding in routes" section... but here we are not explaining something that has been removed but changed. Even, if this is tied to the same PR, it seems doesn't feel right.

I think we need a new section in "Breaking changes" instead. Probably something similar to this one https://docs.astro.build/en/guides/upgrade-to/v5/#changed-params-no-longer-decoded in the v5 upgrade guide.

ArmandPhilippot (commented):
Thanks for the update! But, I think the upgrade guide addition is misplaced and your description for `Astro.params` doesn't seem correct.

Prayaksh:
tried to change as suggested.

thanks!

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -163,9 +163,9 @@ See also: [`params`](/en/reference/routing-reference/#params)
 <Since v="1.0.0" />
 </p>
 
-`url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` value. It is useful for interacting with individual properties of the request URL, like pathname and origin. 
+A normalized [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object derived from the current `request.url` value. It is useful for interacting with individual properties of the request URL, like pathname and origin. 
 
-`Astro.url` is equivalent to doing `new URL(Astro.request.url)`.
+Since Astro v6, the pathname is partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
```
ematipico:
```suggestion
Since Astro v6, the pathname may be partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
```

It's not always the case, so maybe using "may" sounds better

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -832,6 +832,16 @@ If you have route files with `%25` in the filename, rename them to use a differe
...
+## Changed: Route pathname normalization
+
+Since Astro v6, route pathnames are partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
+
+This also affects route params derived from normalized pathnames.
+
+### What should I do?
+
+If you need access to the original encoded pathname, use `new URL(Astro.request.url).pathname` instead of `Astro.url.pathname`.
+
```
ArmandPhilippot: This should be an h3 (instead of h2) and h4 (instead of h3) and moved to the [Breaking changes](https://docs.astro.build/en/guides/upgrade-to/v6/#breaking-changes) section.

src/content/docs/en/reference/configuration-reference.mdx
```diff
@@ -517,6 +517,8 @@ In some specific contexts (e.g., applications behind trusted reverse proxies wit
 
 When not configured, `X-Forwarded-Host` headers are not trusted and will be ignored.
 
+When deploying behind reverse proxies or serverless platforms (for example Vercel), configure `allowedDomains` so Astro can safely trust forwarded host headers during origin validation.
+
```
ArmandPhilippot: As stated at the beginning of this file, it is read-only and auto-generated from the main repo. So, any change should be made in the `withastro/astro` repo instead.
But, anyway, this sounds unrelated to this PR or the issue.

ArmandPhilippot (commented):
Thanks for the update but, in addition to Ema suggestion, I see two issues:
* `configuration-reference.mdx` contains unrelated changes
* the upgrade guide section is misplaced:

Prayaksh:
Sorry for the confusion — my previous commit accidentally included an unrelated change. I've corrected the commit and updated the branch so the PR now only contains the intended documentation changes.

I also wasn't able to identify the source PR for this behavior change. If there's a specific PR number that should be referenced, please let me know.

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -1571,6 +1572,18 @@ If you want to keep the old ID generation for backward compatibility reasons, yo
 
 <ReadMore>Learn more about [Heading IDs](/en/guides/markdown-content/#heading-ids).</ReadMore>
 
+### Changed: Route pathname normalization
+
+<SourcePR number="???" title="???"/>
```
ArmandPhilippot: Maybe Ema could confirm as he has a better overall view of the changes that have taken place in the main repository, but I guess we can use the same PR as [Percent-Encoding in routes](https://docs.astro.build/en/guides/upgrade-to/v6/#removed-percent-encoding-in-routes).


```suggestion
<SourcePR number="14306" title="feat: integrate vite environments"/>
```

src/content/docs/en/guides/upgrade-to/v6.mdx
```diff
@@ -1571,6 +1572,18 @@ If you want to keep the old ID generation for backward compatibility reasons, yo
...
+
+Since Astro v6, route pathnames are partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
+
+This also affects route params derived from normalized pathnames.
+
+#### What should I do?
+
+If you need access to the original encoded pathname, use `new URL(Astro.request.url).pathname` instead of `Astro.url.pathname`.
```
ArmandPhilippot, line 1585: I guess this could be useful, but I think people will rather expect to know how to access the **decoded** path since this is the breaking change. Something like:

````suggestion
If you need access to the original encoded pathname, use `new URL(Astro.request.url).pathname` instead of `Astro.url.pathname`.

If you were previously relying on the automatic decoding, use [`decodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI):

```astro title="src/pages/[slug].astro"
---
const decodedPathname = decodeURI(Astro.url.pathname);
---

<p>You are currently browsing: {decodedPathname}</p>
```
````

File: src/content/docs/en/guides/upgrade-to/v6.mdx [hunk 10]
```diff
@@ -1571,6 +1572,18 @@ If you want to keep the old ID generation for backward compatibility reasons, yo
 
 <ReadMore>Learn more about [Heading IDs](/en/guides/markdown-content/#heading-ids).</ReadMore>
 
+### Changed: Route pathname normalization
+
+<SourcePR number="???" title="???"/>
+
+Since Astro v6, route pathnames are partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
```
ArmandPhilippot: Ema suggestion should be applied here too:

```suggestion
Since Astro v6, route pathnames may be partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -163,9 +163,9 @@ See also: [`params`](/en/reference/routing-reference/#params)
 <Since v="1.0.0" />
 </p>
 
-`url` is a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object constructed from the current `request.url` value. It is useful for interacting with individual properties of the request URL, like pathname and origin. 
+A normalized [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object derived from the current `request.url` value. It is useful for interacting with individual properties of the request URL, like pathname and origin. 
 
-`Astro.url` is equivalent to doing `new URL(Astro.request.url)`.
+Since Astro v6, the pathname may be partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments.
```
ArmandPhilippot: I think we're missing something to make it helpful to users:
```suggestion
Since Astro v6, the pathname may be partially decoded during route normalization. This means `Astro.url.pathname` may differ from `new URL(Astro.request.url).pathname` for percent-encoded route segments, and you may need to decode it yourself with [`decodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI).
```

ArmandPhilippot (commented):
Because this is a breaking change, I don't think users want to know how to access the encoded path, but rather the decoded one (see my suggestions).

I'm not sure about the implementation PR, maybe Ema have more info on this.

ematipico:
What's the status of this PR? Can I help with anything? People are filing bugs because the docs aren't up to date

Prayaksh:
I updated it. Check for any suggestions @ematipico. Thanks

ArmandPhilippot (approved):
Thanks for applying the last suggestions! And, welcome among docs contributors!

PR #13830, opened 2026-05-04
i18n(fr): add `experimental-flags/logger.mdx`

ArmandPhilippot:
#### Description (required)

Adds the French translation of `experimental-flags/logger.mdx` added in #13802

#### Related issues & labels (optional)

- Suggested label: i18n

PR #13907, opened 2026-05-18
feat: stable logger

ematipico:
#### Description (required)

This PR adds documentation for the logger APIs, which will be stable soon


src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1229,3 +1229,54 @@ content="
...
 >
 ```
+### `logger`
+
+<p>
+
+  **Type**: `object | undefined`
+  <Since v="6.4.0" />
+</p>
+
```
ArmandPhilippot, line 1239: We'll need a description here. Maybe something like:
```suggestion

An object that allows you to add additional logs during the rendering of a page. This is particularly useful with [on-demand routes](/en/guides/on-demand-rendering/) and when connecting log aggregation services (e.g. Kibana, Grafana, Loki).

```
But feel free to change the wording as you wish. Maybe my "this is useful" is not accurate and you see something better to say here.

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1229,3 +1229,54 @@ content="
...
+#### `error`
```
ArmandPhilippot: We now prefix methods everywhere (ie. `logger.error()` instead of `error()`). This helps to identify more easily where you are on the page, especially when objects are very nested. And, that allows us to avoid slugs like `#error-1`, `#error-2` which are not very meaningful when sharing a link. Also, for functions, we use parentheses as a suffix.
```suggestion
#### `logger.error()`
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1229,3 +1229,54 @@ content="
...
+
+Emits a message with `error` level.
+
+```astro title="src/pages/index.astro"
+---
+Astro.logger.error("Can't find the checkout ID.");
+---
+```
+  
+#### `warn`
```
ArmandPhilippot: Same as above:
```suggestion
#### `logger.warn()`
```

src/content/docs/en/reference/api-reference.mdx
```diff
@@ -1229,3 +1229,54 @@ content="
...
+</p>
+Emits a message with `warn` level.
+
+```astro title="src/pages/index.astro"
+---
+Astro.logger.warn("Can't find the checkout ID.");
+---
+```  
+  
+#### `info`
```
ArmandPhilippot: Same as above:
```suggestion
#### `logger.info()`
```

src/content/docs/en/reference/logger-reference.mdx
```diff
@@ -1,47 +1,105 @@
 ---
-title: Experimental logger
+title: Astro Logger API
 sidebar:
-  label: Logger
-i18nReady: true
+  label: Logger API
+  i18nReady: true
```
ArmandPhilippot: I'm surprised it didn't cause an error but `i18nReady` shouldn't be indented:
```suggestion
i18nReady: true
```

src/content/docs/en/reference/logger-reference.mdx
```diff
@@ -1,47 +1,105 @@
...
-**Type:** `object`<br />
-**Default:** `undefined`<br />
-<Since v="6.2.0" />
+  <Since v="6.4.0" />
 </p>
 
-Enables an experimental, custom logger that can be controlled by the user. 
 
-When provided, the user has total control over the logs emitted by Astro. Additionally, the feature comes with some built-in loggers that can be used via the new `logHandlers`.
+Astro allows to customize, change and compose loggers.
```
ArmandPhilippot: This seems a bit short as an introduction, especially since we don't have a guide. Being a little more verbose might help capture the attention of those who stumble upon this page just out of curiosity.

Maybe something like:
```suggestion
The Logger API provides greater control over Astro's logging infrastructure. This allows you to replace the default console output with custom logging implementations and connect to log aggregation services.

This API includes three ready-to-use loggers and allows you to integrate your own loggers and compose them.
```

src/content/docs/en/reference/logger-reference.mdx
```diff
@@ -1,47 +1,105 @@
...
+
+matchesLevel("error", "info");
+```
 
 ## Built-in loggers
 
-The feature comes with three built-in loggers.
+Astro offers built-in loggers that applications can use.
 
 ### `logHandlers.json`
```
ArmandPhilippot: for consistency:
```suggestion
### `logHandlers.json()`
```

src/content/docs/en/reference/logger-reference.mdx
ArmandPhilippot: for consistency:
```suggestion
### `logHandlers.node()`
```

ArmandPhilippot: for consistency:
```suggestion
### `logHandlers.console()`
```

src/content/docs/en/reference/logger-reference.mdx
```diff
@@ -298,26 +259,26 @@ The level of the message. Available variants:
...
 - `level`: the level of the message.
 - `label`: an arbitrary label assigned to the log message.
 - `newLine`: whether this message should add a trailing newline.
 
-## APIs reference 
+## APIs reference
 
 The following APIs can be imported from the `astro/logger` specifier.
 
 ### `matchesLevel`
```
ArmandPhilippot: for consistency:
```suggestion
### `matchesLevel()`
```

ArmandPhilippot (commented):
Already! Ahah, you move too fast for the French reviewers... I guess I can close https://github.com/withastro/docs/pull/13830 😅 

Speaking of that, the link error is because of the Korean translation. We usually delete the translations as well when we stabilize an experimental feature.

I know the content hasn't changed much... maybe we should consider renaming the translation and using a Lunaria directive in a case like this. But, I believe Junseong is able to easily browse the Git history to recover its translation so we can remove it for now.

I left a few suggestions to be consistent with our other references (a bit more verbose in some places, functions notation).

Is `Astro.logger` particularly useful with custom loggers? If so, maybe we should add a link to `/en/reference/api-reference/#logger` to improve its discoverability. If not, no change needed!

Also, I noticed we now have [`loggerConfig` in `SSRManifest`](https://github.com/withastro/astro/pull/16745/changes#diff-a5c1ab6b3e939c526b5a546426a2fd601d8d7b654fe206725d92d5f76a4750ab). We'll need a new section in https://docs.astro.build/en/reference/integrations-reference/#ssrmanifest

src/content/docs/en/reference/api-reference.mdx
ematipico: @ArmandPhilippot I updated your suggestion, what do you think?

ematipico:
> Is `Astro.logger` particularly useful with custom loggers? If so, maybe we should add a link to `/en/reference/api-reference/#logger` to improve its discoverability. If not, no change needed!

It's the whole point! Ahahaha. Dufus me that I completely overlooked this important connection. The important thing is that the runtime APIs are always at user's disposal regardless of a custom logger, but with a custom logger users have more control. 

My last commit added a small paragraph with the link

src/content/docs/en/reference/api-reference.mdx
ArmandPhilippot: Sounds good to me!

src/content/docs/en/reference/logger-reference.mdx
```diff
@@ -1,49 +1,110 @@
...
+        // log message somewhere and take the level into consideration
+      }
+    }
+  }
+}
+
+export default orgLogger;
+```
+
+You can now add your own logs during the rendering of a page using [the runtime APIs](en/reference/api-reference/#logger).
```
ArmandPhilippot: Broken link due to a missing slash:
```suggestion
You can now add your own logs during the rendering of a page using [the runtime APIs](/en/reference/api-reference/#logger).
```

> It's the whole point!

Ah, I thought it seemed odd that there wasn't a link. But since I haven't tested this feature yet, I wasn't sure where the best place to add one would be... 😆

src/content/docs/en/reference/logger-reference.mdx
ArmandPhilippot, line 195: I missed this one, parentheses for consistency:
```suggestion
### `logHandlers.compose()`
```

src/content/docs/en/reference/logger-reference.mdx
```diff
@@ -1,49 +1,110 @@
...
+You can now add your own logs during the rendering of a page using [the runtime APIs](/en/reference/api-reference/#logger).
+
+## Log level
+
+A level is an internal, arbitrary score assigned to each message. When a logger is configured with a certain level, only the messages with a level equal to or higher are printed.
+
+There are three levels, from the highest to the lowest score:
+1. `error`
+2. `warn`
+3. `info`
```
ArmandPhilippot, line 81: I was going to say I'm a bit confused by this because in `AstroLoggerLevel` we also have `debug` and `silent`. So, 5 and not 3 levels. But, I just saw the unresolved question in the RFC:

> debug info level has been kept out of the picture because we use [obug](https://npmx.dev/package/obug) for it, which is noisy and completely different from the rest of the logs. Should we keep it out? Docs should address the differences from normal logging.

It seems in the end we'll keep it? I guess this does not contribute to the score and this is a special case like `silent`, as this is not available in the runtime API. And, these two levels match the [`verbose`](https://docs.astro.build/en/reference/cli-reference/#--verbose) and [`silent`](https://docs.astro.build/en/reference/cli-reference/#--silent) flags. Am I right?

In any case, we should probably have something explaining they are also available and how they differ.

ArmandPhilippot (commented):
I was double checking, and I think this is almost ready.
* In the [`SSRManifest` reference](https://docs.astro.build/en/reference/integrations-reference/#ssrmanifest), we need a section for [`loggerConfig`](https://github.com/withastro/astro/pull/16745/changes#diff-a5c1ab6b3e939c526b5a546426a2fd601d8d7b654fe206725d92d5f76a4750ab). And, now we have a section for log levels, we should update the [`SSRManifest.logLevel`](https://docs.astro.build/en/reference/integrations-reference/#ssrmanifestloglevel) description with a link to that section.
* I left a question about log levels because `debug` and `silent` are also available but we do not explain how they differ.

**Edit:** I found another place https://docs.astro.build/en/reference/modules/astro-config/#imports-from-astroconfig we'll need to add `logHandlers` there. A good occasion to create a link to the Logger API there!

src/content/docs/en/reference/logger-reference.mdx
ematipico: They were kept out on purpose because:
- debug logging **isn't** astro logging. It involves `obug`, which also enables vite logging. It's a mess. I wanted to keep users out of it. Perhaps we could mention how our/users logging interact with debug logging
- silent logging isn't really a level that you want to configure and/or use, right? Specifically `Astro.logger.silent`. Perhaps we could expand the docs of `--silent` to explain that users logs are hidden too

ArmandPhilippot: Well, I was referring to levels because this is a value allowed by `AstroLoggerLevel`. Naming... 😄 But, yeah, I agree those are not levels. I just feel there is something missing and that could be confusing for the users too.

Maybe this section is fine as it is because this is about the real levels. And, maybe, the `AstroLoggerLevel` description should be reworked a little instead? I mean to link the possible values ​​and what they refer to. This way we can link back to the Log Level section for `info`, `warn`, and `error`, and to the CLI flags for `silent` and `debug`.

I'm not yet sure about the wording to make a suggestion (too late to have a clear head 😴 ) but it might make more sense than updating the Log level section...

But, yes, updating `--silent` sounds like a good idea!

ematipico: I'll look into it, we shouldn't provide silent and debug levels to the user

src/content/docs/en/reference/logger-reference.mdx
ArmandPhilippot: So, what I meant last night in https://github.com/withastro/docs/pull/13907#discussion_r3269438064 was that the `AstroLoggerLevel` section could look something like this:
```suggestion
### `AstroLoggerLevel`

<p>

**Type:** `'debug' |'info' |'warn' | 'error' | 'silent'`
</p>

Specifies the verbosity level of logs:
- `info`, `warn`, and `error`: defines the minimal [log level](#log-level) to print.
- `silent`: this is equivalent to the [`--silent` CLI flag](/en/reference/cli-reference/#--silent) and enables silent logging.
- `debug`: this is equivalent to the [`--debug` CLI flag](/en/reference/cli-reference/#--verbose) and enables verbose logging, including Vite logging.
```

ArmandPhilippot:
Ah, right, my French PR finally got merged. We'll need to remove it to fix the failing check!

matthewp (approved):

